PR #131, opened 2022-12-25
refactor, feat: clean Cargo.toml files, upgrade tch to 0.10

ChefKissInc:
Adds support for Metal Performance Shaders to the Tch backend

### DIFF
--- a/.github/workflows/publish-template.yml
+++ b/.github/workflows/publish-template.yml
@@ -25,5 +25,3 @@ jobs:
         run: ./ci/publish.sh ${{ inputs.crate }}
         env:
           CRATES_IO_API_TOKEN: ${{ secrets.CRATES_IO_API_TOKEN }}
-
-
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: publish
 
 on:
   push:
-    tags:        
+    tags:
       - 'v*'
 
 jobs:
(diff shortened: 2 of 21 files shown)


nathanielsimard (changes requested):
I would want to run the github action checks before merging, I think the github templates should be updated to trigger on `pull_request` instead of on `push`. It would be appreciated if you could update it so that we can test if it actually triggers for pull requests from public forks.

Otherwise great work and thank you for your contributions :smiley:

ChefKissInc:
> I would want to run the github action checks before merging, I think the github templates should be updated to trigger on `pull_request` instead of on `push`. It would be appreciated if you could update it so that we can test if it actually triggers for pull requests from public forks.
> 
> Otherwise great work and thank you for your contributions 😃

Done

ChefKissInc:
I also noticed a bunch of clippy warnings about inlining format arguments. I can send in another PR after this one to fix them if you'd like @nathanielsimard

nathanielsimard:
> I also noticed a bunch of clippy warnings about inlining format arguments. I can send in another PR after this one to fix them if you'd like @nathanielsimard

Yes why not, but the CI is supposed to fail when there are clippy warnings. What command did you run?

ChefKissInc:
> > I also noticed a bunch of clippy warnings about inlining format arguments. I can send in another PR after this one to fix them if you'd like @nathanielsimard
> 
> Yes why not, but the CI is supposed to fail when there are clippy warnings. What command did you run?

I'm actually on nightly rust, so the warnings may be new.
It's just cargo clippy

ChefKissInc:
I've actually already fixed all the warnings except this
Screenshot 2022-12-25 at 17 37 40
At
Screenshot 2022-12-25 at 17 38 15

PR #137, opened 2022-12-25
refactor: fix all clippy warnings

ChefKissInc:


ChefKissInc:
The commits look a bit weird because I reused the branch from the previous PR. But the diff is only for the fixed clippy warnings. Just do a squash merge I guess

PR #144, opened 2023-01-03
fixed expected lifetime parameter warning

SunnyWar:
This fixes the lint warning "expected lifetime parameter"


nathanielsimard:
@SunnyWar The continuous integration (CI) checks are failing. I tried reproducing the warnings on nightly and was not able to do so. Maybe you compiled with an outdated nightly version? 

Anyways, I think it is strange that we have to explicitly declare anonymous lifetimes, so I think it's safer to wait for the warnings to be present on rust stable.

PR #146, opened 2023-01-07
Impl display on Tensor

nopestack:
WIP implementation of std::fmt::Display on Tensor and BoolTensor

### DIFF
--- a/burn-tensor/src/tensor/base.rs
+++ b/burn-tensor/src/tensor/base.rs
@@ -5,6 +5,7 @@ use crate::tensor::ElementConversion;
 use crate::tensor::{Data, Distribution, Shape};
 use crate::BoolTensor;
 use std::convert::TryInto;
+use std::fmt::Display;
 use std::ops::Range;
 
 /// A tensor or a *n-dimensional* array.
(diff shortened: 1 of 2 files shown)


burn-tensor/src/tensor/base.rs
```diff
@@ -724,3 +725,22 @@ impl<const D: usize, B: ADBackend> Tensor<B, D> {
...
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let dims = self.shape().dims;
+
+        f.write_str("Tensor {\n")?;
+        f.write_fmt(format_args!("\tshape: {dims:?},\n"))?;
+        f.write_fmt(format_args!("\tdevice: {:?},\n", self.device()))?;
+        f.write_str("}\n")?;
+
+        // TODO: figure out how to infer backend name + dtype
```
nathanielsimard, line 741: `Backend::name() -> String` exists, so it would be pretty easy :)
Maybe `Backend::Elem` should implement display and should be used for dtype.

burn-tensor/src/tests/stats/basic.rs
```diff
@@ -13,4 +13,12 @@ mod tests {
         let data_expected = Data::from([[2.4892], [15.3333]]);
         data_expected.assert_approx_eq(&data_actual, 3);
     }
+
+    #[test]
+    fn tensor_should_impl_display() {
+        let data = Data::from([[0.5, 1.8, 0.2, -2.0], [3.0, -4.0, 5.0, 0.0]]);
+        let tensor = Tensor::<TestBackend, 2>::from_data(data);
+
+        // TODO: figure out where to add this test
```
nathanielsimard, line 22: Assert with a string, when you implement `std::fmt::Display`, the function `to_string` is automatically implemented and can be used to create a string.

nopestack:
Thanks for the feedback @nathanielsimard. I'll get back to this PR over the weekend as I free up a bit more time

Issue #159, opened 2023-02-05
Feature request: Mish activator


bioinformatist:
### Feature description

Hey, can I implement the [Mish activator](https://arxiv.org/abs/1908.08681) (of course as well as its unit test) for burn?

### Feature motivation

It does have better performance in our projects (implemented by Tensorflow in the past).

### (Optional) Suggest a Solution

I wanna do this job myself. Could I:smile:?

nathanielsimard:
Yes, go ahead, if you need help or anything just ask on the discord.

PR #160, opened 2023-02-10
feat(trait-TensorOps): add log1p

bioinformatist:
log1p will be needed for activation functions such as softplus

Resolves burn-rs/burn#159


bioinformatist:
Oh sorry that I fogot the issue #159 is for mish.

nathanielsimard (commented):
I think the only thing missing is the test for the log1p function in `burn-tensor`, but other than that good job!
You also have to format your code with the following to satisfy the CI:

```bash
cargo clippy --fix
cargo fmt --all
```

PR #164, opened 2023-02-16
Add feature to load config / state from binary slice.

twitchax:
These changes:
* Add a new `load_binary` function to the `Config` trait which allows for loading the config from a binary slice.
* Add a new `load_binary` function to the `State` impl which allows for loading the state from a binary slice.

The impetus of these changes are for cleanly shipping models for future inference.  For example, if you train a model, and then want to build the inference model into a new binary, you could include those binaries into the executable with `Include_bytes`.  The current implementation would require that you ship the model config and model state as separate files, which may be inconvenient for certain delivery methods.


nathanielsimard (approved):
Great job! It's a nice addition to the burn.

Issue #168, opened 2023-02-19
Make inference packages no_std compatible for ndarray backend


antimora:
### Feature description

`burn-core`, `burn-tensor`, and `burn-ndarray` should run with `no_std` attribute. 

### Feature motivation

Making inference packages (burn-core, burn-tensor, and burn-ndarray) `no_std` compatible gives developers ability to deploy models for inference mode to wasm, embedded systems and mobile (or any other system where OS might not be available. 

### (Optional) Suggest a Solution

General steps on how are outlined in this book's chapter on no_std: https://www.lurklurk.org/effective-rust/no-std.html. 

Specific steps on the current solution will be updated here (TBD).

antimora:
Completed the task. Closing as done.

antimora:
@spadarian, I worked on this PR to make Burn (deep learning framework) no_std compatible. Take a look: https://github.com/burn-rs/burn.

There is a section on no_std: https://github.com/burn-rs/burn#no_std-support

Here is an online demo that relies on no_std and it compiles with wasm: https://burn-rs.github.io/demo

We have CIs that test no_std, so we know Burn builds for no_std. 

For the inference, you'll need to use burn-ndarray backend. I am currently working on ONNX import part: https://github.com/burn-rs/burn/issues/204 but it will be a while to have the full set of ops.

spadarian:
I checked the no_std example... I still get a bunch of errors like this one (all relates to `spin`):

```bash
error[E0599]: no method named `compare_exchange_weak` found for struct `AtomicBool` in the current scope
   --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/spin-0.9.8/src/mutex/spin.rs:182:14
    |
180 |           while self
    |  _______________-
181 | |             .lock
182 | |             .compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed)
    | |             -^^^^^^^^^^^^^^^^^^^^^ method not found in `AtomicBool`
    | |_____________|
```

Maybe there is something wrong with my configuration?

antimora:
Yeah, I have seen this before. AtomicBool is not available for the older ARM but works for `thumbv7m-none-eabi`. Atomic CPU feature is needed for locking. Can you share which target you're attempting to build?

spadarian:
It's for the rp2040... so I think it is `thumbv6m-none-eabi`

antimora:
[...]

PR #169, opened 2023-02-19
Make burn-ndarray and burn-tensor no_std compatible (#168)

antimora:
I made `burn-ndarray` and `burn-tensor` `no_std` compatible. All tests are passing. However, when I run mnist training example, it fails with the following error. I need someone review my changes and suggest a fix. I have used debugger extensively to see the value, so I have some extra data to help with debugging. 

```
[mnist]$ RUST_BACKTRACE=1 cargo run --example mnist --features ndarray            
   Compiling burn-tensor v0.5.0 (/workspaces/burn/burn-tensor)
   Compiling burn-autodiff v0.5.0 (/workspaces/burn/burn-autodiff)
   Compiling burn-ndarray v0.5.0 (/workspaces/burn/burn-ndarray)
   Compiling burn-core v0.5.0 (/workspaces/burn/burn-core)
   Compiling burn-train v0.5.0 (/workspaces/burn/burn-train)
   Compiling burn v0.5.0 (/workspaces/burn/burn)
   Compiling mnist v0.1.0 (/workspaces/burn/examples/mnist)
    Finished dev [unoptimized + debuginfo] target(s) in 45.31s
     Running `/workspaces/burn/target/debug/examples/mnist`
=== PANIC ===
A fatal error happened, you can check the experiment logs here => '/tmp/burn-example-mnist/experiment.log'
=============
thread 'main' panicked at 'Slice end 785 is past end of axis of length 784', /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/ndarray-0.15.6/src/dimension/mod.rs:401:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
   2: ndarray::dimension::to_abs_slice
             at /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/ndarray-0.15.6/src/dimension/mod.rs:401:5
   3: ndarray::dimension::do_slice
             at /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/ndarray-0.15.6/src/dimension/mod.rs:430:30
   4: ndarray::impl_methods::<impl ndarray::ArrayBase<S,D>>::slice_axis_inplace
             at /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/ndarray-0.15.6/src/impl_methods.rs:647:22
   5: ndarray::impl_methods::<impl ndarray::ArrayBase<S,D>>::slice_move::{{closure}}
             at /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/ndarray-0.15.6/src/impl_methods.rs:538:17
   6: <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::for_each
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/slice/iter/macros.rs:211:21
   7: ndarray::impl_methods::<impl ndarray::ArrayBase<S,D>>::slice_move
             at /home/dev/.cargo/registry/src/github.com-1ecc6299db9ec823/ndarray-0.15.6/src/impl_methods.rs:535:9
   8: burn_ndarray::ops::tensor::<impl burn_tensor::tensor::ops::tensor::TensorOps<burn_ndarray::backend::NdArrayBackend<E>> for burn_ndarray::backend::NdArrayBackend<E>>::index
             at /workspaces/burn/burn-ndarray/src/ops/tensor.rs:296:21
   9: burn_ndarray::ops::maxpool::max_pool2d_backward_naive
             at /workspaces/burn/burn-ndarray/src/ops/maxpool.rs:82:37
  10: burn_ndarray::ops::module::<impl burn_tensor::tensor::ops::modules::base::ModuleOps<burn_ndarray::backend::NdArrayBackend<E>> for burn_ndarray::backend::NdArrayBackend<E>>::max_pool2d_with_indexes_backward
             at /workspaces/burn/burn-ndarray/src/ops/module.rs:109:32
  11: <burn_autodiff::ops::module::BackwardMaxPool<B,2_usize,4_usize> as burn_autodiff::graph::ops::base::BackwardRecordedOps<<B as burn_tensor::tensor::backend::base::Backend>::TensorPrimitive<4_usize>>>::backward_step
             at /workspaces/burn/burn-autodiff/src/ops/module.rs:335:21
  12: <burn_autodiff::graph::node::backward::BackwardNode<T> as burn_autodiff::graph::ops::base::RecordedOpsParent>::backward_step
             at /workspaces/burn/burn-autodiff/src/graph/node/backward.rs:79:9
  13: burn_autodiff::graph::node::backward::BackwardNode<Out>::backward
             at /workspaces/burn/burn-autodiff/src/graph/node/backward.rs:65:17
  14: burn_autodiff::tensor::ad::<impl burn_autodiff::tensor::base::ADTensor<_,B>>::backward
             at /workspaces/burn/burn-autodiff/src/tensor/ad.rs:15:9
  15: <burn_autodiff::backend::ADBackendDecorator<B> as burn_tensor::tensor::backend::base::ADBackend>::backward
             at /workspaces/burn/burn-autodiff/src/backend.rs:37:9
  16: burn_tensor::tensor::base::Tensor<B,_>::backward
             at /workspaces/burn/burn-tensor/src/tensor/base.rs:736:9
  17: <mnist::model::Model<B> as burn_train::learner::train_val::TrainStep<B,mnist::data::MNISTBatch<B>,burn_train::learner::classification::ClassificationOutput<B>>>::step
             at ./src/model.rs:87:32
  18: burn_train::learner::train_val::<impl burn_train::learner::base::Learner<M,O,TO,VO>>::train_step
             at /workspaces/burn/burn-train/src/learner/train_val.rs:158:24
  19: burn_train::learner::train_val::<impl burn_train::learner::base::Learner<M,O,TO,VO>>::fit
             at /workspaces/burn/burn-train/src/learner/train_val.rs:71:17
  20: mnist::training::run
             at ./src/training.rs:54:26
  21: mnist::ndarray::run
             at ./examples/mnist.rs:13:9
  22: mnist::main
             at ./examples/mnist.rs:50:5
  23: core::ops::function::FnOnce::call_once
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/ops/function.rs:507:5
=== PANIC ===
A fatal error happened, you can check the experiment logs here => '/tmp/burn-example-mnist/experiment.log'
=============
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: SendError { .. }', /workspaces/burn/burn-core/src/data/dataloader/multithread.rs:53:30
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
   2: core::result::unwrap_failed
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/result.rs:1791:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/result.rs:1113:23
   4: <burn_core::data::dataloader::multithread::MultiThreadDataLoader<O> as burn_core::data::dataloader::base::DataLoader<O>>::iter::{{closure}}::{{closure}}
             at /workspaces/burn/burn-core/src/data/dataloader/multithread.rs:51:25

```


burn-core/src/module/param/id.rs
```diff
@@ -1,6 +1,7 @@
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Hash, PartialEq, Eq, Clone, Serialize, Deserialize)]
+
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Serialize, Deserialize, Ord, PartialOrd)]
```
nathanielsimard, line 4: Why the `ParamId` needs to implement `Ord` and `PartialOrd`?

burn-ndarray/Cargo.toml
```diff
@@ -21,18 +21,29 @@ blas-openblas-system = [
...
 ]
 
+# Use randomly generated seed during build. This is useful for testing purposes.
+build-generated-seed = []
+
 [dependencies]
-burn-tensor = { version = "0.5.0", path = "../burn-tensor" }
-blas-src = { version = "0.8.0", default-features = false, optional = true }
-openblas-src = { version = "0.10.5", optional = true }
+blas-src = {version = "0.8.0", default-features = false, optional = true}
```
nathanielsimard, line 28: I would have a `no_std` and a `std` feature flag on all the crates that can be built with `no_std`.

burn-ndarray/Cargo.toml
```diff
@@ -21,18 +21,29 @@ blas-openblas-system = [
...
 
-ndarray = "0.15.6"
-libm = "0.2.6"
-derive-new = "0.5.9"
-rand = "0.8.5"
-num-traits = "0.2.15"
+num-traits = {version = "0.2.15", default-features = false, features = ["libm"]} # Default enables std
+spin = {version = "0.9.5", features = ["mutex", "spin_mutex"]} # using in place of use std::sync::Mutex;
+
+# Used to generate a seed for random at build time
```
nathanielsimard, line 41: What is the use case for generating a seed at build time?

burn-ndarray/src/backend.rs
```diff
@@ -20,7 +28,7 @@ impl Default for NdArrayDevice {
 
 #[derive(Clone, Copy, Default, Debug)]
 pub struct NdArrayBackend<E> {
-    _e: E,
+    phantom: PhantomData<E>,
```
nathanielsimard, line 31: What is `PhantomData`  does it remove the unecesasry element contained in the struct? If yes that's awesome.

burn-ndarray/src/element.rs
```diff
@@ -13,66 +15,140 @@ pub(crate) trait ExpElement {
...
-                let tmp = $tmp::sqrt(self as $tmp);
-                tmp as $elem
-            }
-        }
-    };
-}
-
 impl NdArrayElement for f64 {}
-impl_exp_elem!(f64);
+impl ExpElement for f64 {
```
nathanielsimard, line 19: I prefer having a macro for that kind of stuff, but it can be refactored later.

burn-ndarray/src/ops/conv.rs
```diff
@@ -84,8 +86,8 @@ fn conv2d_with_kernel<E: NdArrayElement>(
     let [k1, k2] = kernel.shape().dims;
     let [heigth, width] = x.shape().dims;
 
-    let heigth_new = f32::ceil((heigth - k1 + 1) as f32 / stride[0] as f32) as usize;
-    let width_new = f32::ceil((width - k2 + 1) as f32 / stride[1] as f32) as usize;
+    let heigth_new = ceilf((heigth - k1 + 1) as f32 / stride[0] as f32) as usize;
+    let width_new = ceilf((width - k2 + 1) as f32 / stride[1] as f32) as usize;
```
nathanielsimard, line 90: I would check here if the bug is not related to the ceiling function. It would be suprising, but hey we don't know.

burn-ndarray/src/ops/conv.rs
```diff
@@ -1,7 +1,9 @@
 use super::padding::apply_padding2d;
 use crate::{element::NdArrayElement, tensor::NdArrayTensor, NdArrayBackend, NdArrayDevice};
 use burn_tensor::{ops::TensorOps, Shape};
-
+use libm::ceilf;
+extern crate alloc;
```
nathanielsimard, line 5: I think `extern crate` should all be in `lib.rs`

burn-ndarray/src/ops/tensor.rs
```diff
@@ -51,10 +58,16 @@ impl<E: NdArrayElement> TensorOps<NdArrayBackend<E>> for NdArrayBackend<E> {
...
+        let mut rng: SmallRng;
+        rng = if let Some(rng_seeded) = seed.as_ref() {
+            rng_seeded.clone()
+        } else {
+            if cfg!(feature = "build-generated-seed") || cfg!(test) {
+                SmallRng::seed_from_u64(GENERATED_SEED)
+            } else {
+                panic!("Seeding from the OS automatically is not supported.")
+            }
         };
```
nathanielsimard, line 71: That's a reason why a `std` feature flag can be usefull. I would automatically generate a seed at compile time when the `no_std` flag is activated, which would avoid having to `panic` here.

burn-tensor/src/tensor/container.rs
```diff
@@ -1,17 +1,23 @@
...
+use alloc::boxed::Box;
+
+use alloc::collections::BTreeMap;
 
 /// Contains tensor of arbitrary dimension.
 #[derive(Debug)]
 pub struct TensorContainer<B: Backend, ID> {
-    tensors: HashMap<ID, Box<dyn Any + Send + Sync>>,
-    _b: B,
+    tensors: BTreeMap<ID, Box<dyn Any + Send + Sync>>,
```
nathanielsimard, line 13: Hmm, ok so I understand now why you needed `Ord` and `PartialOrd` for the ID struct. `HashMap` might still be more performant, so we may switch to a `HashMap` when `no_std` is not activated.

nathanielsimard (changes requested):
I'm not sure why you have a bug when running the `mnist` example, but I would focus on cleaning up a bit the code and introducing the `std` and `no_std` flags. It's often easier to debug after a bit of cleaning, and sometime the bugs just go away by themself 😅.

Otherwise good job for the PR, it will be very awesome to support a `no_std` environment 😃

burn-ndarray/src/element.rs
antimora: I thought keeping the macros but I would have introduced additional two (four would have been total) to generate 6 implementation. The new would have been needed for f32 which is supported as separate function in libm. Overall I think it was simpler to see code without macros.

burn-ndarray/src/ops/conv.rs
antimora: I'll run some tests. I also reverted these functions and the bug persisted.

burn-ndarray/src/ops/conv.rs
antimora: OK. I'll move these there.

burn-ndarray/src/ops/tensor.rs
antimora: We can do this for std but we should keep panic if Backend is not seeded (more explicit). The user will have an option to use `build-generated-seed` if needed.

burn-tensor/src/tensor/container.rs
antimora: OK. I'll add behind the feature flag.

burn-core/src/module/param/id.rs
antimora: This is needed for BTreeMap which is used instead of HashMap. HashMap requires randomization and not supported with no_std.

burn-ndarray/Cargo.toml
antimora: OK. I'll add `std` flag, which is the recommended way for such situations. Having `std` will bring in extra features (following additive nature of feature flags)

burn-ndarray/src/backend.rs
antimora: That's exactly what it does. Plus it's zero sized type. It has the effect of _e but clearer and without memory over head.

burn-ndarray/src/ops/tensor.rs
nathanielsimard: What is the use case when you want `no_std` and build without a generated seed?

burn-ndarray/Cargo.toml
antimora: When I removed auto-seeding from OS (no_std environment does not have it), the tests broke. Seeding for each test would have required more code changes. So I introduced `build-generated-seed` flag. But now I am thinking if we have `std` flag, which is set by default, then we still can make the tests work. 

It might worth keeping the build time seed. It could be useful in some environments where seeding is not available. 

It's your call really. We can remove it if you'd like.

burn-ndarray/src/backend.rs
antimora: https://doc.rust-lang.org/std/marker/struct.PhantomData.html

burn-ndarray/src/ops/tensor.rs
antimora: OK. To make consistent with `std` fallback behavior, i.e. get seed automatically, no_std should get from somewhere, which will be from the build time seed. I'll remove the `build-generated-seed` entirely.

antimora:
The latest commit fixes the training issue. I don't know how but using rand/SmallRng is the problem. I was able to switch to StdRng which works with no_std as well. 

The main change/fix in the latest commit is to use std flag. Next I'll work on burn-core to make it no_std compatible.

burn-ndarray/tests/no_std_checks.rs
```diff
@@ -0,0 +1,94 @@
+// This module contains tests to verify substituded functions are the same with no_std and std
```
nathanielsimard, line 1: I'm not sure we should keep this test. It may have been usefull to debug, but it essentially tests other crates.

nathanielsimard (changes requested):
Everything looks good, but I don't think we should keep the `libm` math operation tests.

The last thing to do would be to run clippy and cargo fmt:

```bash
cargo clippy --fix 
cargo fmt --all
```

burn-ndarray/tests/no_std_checks.rs
antimora: OK. I'll run the clippy and fmt. Also I'll remove the tests.

antimora:
Fixed the formatting issues, removed unnecessary tests and put back std::sync::Mutex for std feature flag (other wise spin::Mutex is used).

Ready for the review and merge. I am working on burn-core.

PR #172, opened 2023-02-22
perf: remove unecessary masking in dropout

nathanielsimard:


nathanielsimard:
It doesn't seem to work for now, I need to add some forward bacward tests.

PR #173, opened 2023-02-23
Make burn and burn-core packages no_std compatible (#168)

antimora:
I made all necessary changes to make `burn` and `burn-core` `no_std` compatible. I added a top level crate burn-no-std-tests to integrate all no_std parts. Currently it only contains one test (inference on mnist model). 

I pulled all common cargo versions into the top level Cargo.toml which helps managing no_std dependency compatibility. It was hard to make sure we had the same version and min features across no_std burn packages. I introduced `burn-common` package to hold common packages (for std and no_std modes). 

The next on my to do is to modify CI workflow to add test flags (no-default-features) and new targets. Also I'll add my local script that I run before commit. There are lots of testing and build need to be done before check in. I figured we should help dev with a simple script to test. Otherwise they'll have to rely on CI to check their work.


antimora:
I am fixing `burn-dataset` build issue. Sorry I missed this somehow.

antimora:
OK. burn-dataset build error is fixed. Ready to review.

Cargo.toml
```diff
@@ -14,5 +14,34 @@ members = [
...
+log4rs = "1.2.0"
+ndarray = {version = "0.15.6", default-features = false}# Default enables std        
+num-traits = {version = "0.2.15", default-features = false, features = ["libm"]}# Default enables std 
+rand = {version = "0.8.5", default-features = false, features = ["std_rng"]}# Default enables std   
+rand_distr = {version = "0.4.3", default-features = false}# Default enables std   
+serde = {version = "1.0.152", default-features = false, features = ["derive", "alloc"]}# Default enables std   
+serde_json = {version = "1.0.91", default-features = false}# Default enables std   
+spin = {version = "0.9.5", features = ["mutex", "spin_mutex"]}
+thiserror = "1.0.38"
+uuid = {version = "1.3.0", default-features = false}
```
nathanielsimard, line 47: I would separate the dependencies in blocks with comments about why it is need and maybe features comments comments.

README.md
```diff
@@ -174,6 +175,12 @@ fn main() {
 
 See this [example](https://github.com/burn-rs/burn/tree/main/examples/mnist) for a real usage.
 
+## no_std support
+
+Burn supports `no_std` with `alloc` for the inference mode with the NDArray backend. Simply disable the default features of the `burn` and `burn-ndarray` packages (minimum required to run the inference mode). See the [burn-no-std-tests](https://github.com/burn-rs/burn/tree/main/examples/burn-no-std-tests) package as a reference implementation. Additionally `burn-core` and `burn-tensor` packages support `no_std` with `alloc` if needed to direclty include them as dependencies (the `burn` package reexports `burn-core` and `burn-tensor`).
+
+Note, under the `no_std` mode, a random seed is generated during the build time if the seed is not initialized by by `Backend::seed` method. Additionally, [spin::mutex::Mutex](https://docs.rs/spin/latest/spin/mutex/struct.Mutex.html) is used in place of [std::sync::Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html) under the `no_std` mode.
```
nathanielsimard, line 182: Typo: by by `Backend::seed` => by `Backend::seed`

burn-common/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+authors = ["Dilshod Tadjibaev (@antimora)"]
+edition = "2021"
+license = "MIT/Apache-2.0"
+name = "burn-common"
+readme = "README.md"
+repository = "https://github.com/burn-rs/burn-common"
```
nathanielsimard, line 7: repository = "https://github.com/burn-rs/burn-common" => repository = "https://github.com/burn-rs/burn/tree/main/burn-common"

burn-common/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+version = "0.1.0"
```
nathanielsimard, line 8: For now I set the same version for all crates in the workspace, it's easier when deploying, so I would change it to 0.5.0. Maybe we could rethink a versionning strategy, but in another PR with the proper changes in the CI/CD pipeline.

burn-common/src/id.rs
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            get_seeded_rng()
+        };
+
+        let random_bytes: Bytes = rng.gen();
+        *seed = Some(rng);
+
+        let uuid = Builder::from_random_bytes(random_bytes).into_uuid();
+
+        uuid.as_hyphenated().to_string()
```
nathanielsimard, line 23: I'm not sure why we should use a seed here, is it to be `no_std` compatible?

burn-common/src/id.rs
```diff
@@ -0,0 +1,78 @@
...
+        for _i in 0..IDS_CNT {
+            assert_eq!(set.insert(IdGenerator::generate()), true);
+        }
+
+        assert_eq!(set.len(), IDS_CNT);
+    }
+
+    #[cfg(feature = "std")]
+    #[test]
+    fn thread_safety_test() {
```
nathanielsimard, line 57: I would extract `10` and `1000` and `10_000` in variables, like `num_threads` , `num_repeat` and `num_added`.

burn-common/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub mod id;
+pub mod rand;
+pub mod stub;
+
+extern crate alloc;
```
nathanielsimard, line 7: Do we need the alloc crate when std is activated? Does it change anything?

burn-common/src/stub.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    #[inline(always)]
+    pub fn write(&self) -> Result<RwLockWriteGuard<T>, alloc::string::String> {
+        Ok(self.inner.write())
+    }
+}
+
+// ThreadId stub when no std is available
+#[derive(Eq, PartialEq, Clone, Copy, Hash, Debug)]
+pub struct ThreadId(core::num::NonZeroU64);
```
nathanielsimard, line 52: I would like to have the same kind of `wapper struct` for our `IdGenerator`, maybe `BurnId` ?

burn-core/README.md
```diff
@@ -4,3 +4,12 @@ This crate should be used with [burn](https://github.com/burn-rs/burn).
 
 [![Current Crates.io Version](https://img.shields.io/crates/v/burn-core.svg)](https://crates.io/crates/burn-core)
 [![license](https://shields.io/badge/license-MIT%2FApache--2.0-blue)](https://github.com/burn-rs/burn-core/blob/master/README.md)
+
+
+## Feature Flags
```
nathanielsimard, line 9: 👌

burn-common/src/stub.rs
nathanielsimard: It should be done in another PR I think

burn-core/src/module/param/running.rs
```diff
@@ -1,14 +1,33 @@
...
     backend::{ADBackend, Backend},
     Data, Tensor,
 };
-use std::{
-    collections::HashMap,
-    sync::{Arc, Mutex, RwLock},
-    thread::ThreadId,
-};
+
+cfg_if::cfg_if! {
```
nathanielsimard, line 10: Is it the only place where `cfg_if` is used? I'm not a bit fan, it can easily be written without it.

```rust
#[cfg(feature = "std")] 
mod threading {
       use std::thread::ThreadId;
       use std::sync::{Mutex, RwLock};
       use std::collections::HashMap;

       #[inline(always)]
       fn get_thread_current_id() -> ThreadId {
           std::thread::current().id()
      }
}

#[cfg(not(feature = "std"))]
mod threading {
       use burn_common::stub::{Mutex, RwLock, ThreadId};
       use hashbrown::HashMap;

       #[inline(always)]
       fn get_thread_current_id() -> ThreadId {
           panic!("Current thread id is not available")
       }
}
use threading::*;
```

burn-no-std-tests/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+authors = [
+  "nathanielsimard <nathaniel.simard.42@gmail.com>",
+  "Dilshod Tadjibaev (@antimora)",
+]
+edition = "2021"
+license = "MIT/Apache-2.0"
+name = "burn-no-std-tests"
+readme = "README.md"
+repository = "https://github.com/burn-rs/burn-no-std-tests"
```
nathanielsimard, line 10: Fix the repository url like the one with burn-common.

burn-no-std-tests/README.md
```diff
@@ -0,0 +1,26 @@
+The `burn-no-std-tests` contains integration tests aimed to check `no_std` compatibility of `burn`, `burn-core`, `burn-tensor` and `burn-ndarray` packages.
```
nathanielsimard, line 1: I'm not sure we should create a package for that kind of tests, but instead building the examples without the `std` flag. It's not a decision we have to take right now, since this package will never be released, therefore easilly changed.

nathanielsimard (changes requested):
There are only minor changes, overall awesome job! Using the workspace dependencies is the right approach here so thanks for that!

burn-common/src/id.rs
antimora: nanoid() was pulling random numbers from OS's getrandom directly which always gives a fresh random numbers. It's like getting from_entropy() each time. Since no_std does not have access to getrandom, we use our pre-seeding. It's similar what you wrote for tensor.random(). nanoid() was a big sticking point when I was building for no_std targets. Plus this approach after because of amortization for subsequent ID generation. 

I am using UUID's v4 with random data. It's good for database ids. Any other version relies on time and globally unique sequence generator (MAC id). For our use case V4 is sufficient and on par with nanoid.

burn-common/src/stub.rs
antimora: OK. I can add TODO for this.

burn-no-std-tests/Cargo.toml
antimora: ok

burn-no-std-tests/README.md
antimora: OK. Let's brew this idea for a bit. Examples could serve as our integration tests. In fact, I wanted to have no_std example. Let's come back to this. I'll add TODO for this.

burn-common/src/id.rs
antimora: OK

burn-common/src/lib.rs
antimora: No, it does not change anything. Rust reexports core and alloc to std. no_std hides std name space. 

alloc crate is needed because the code references Vec and String from alloc. 

BTW, this is how NDArray does it: https://github.com/rust-ndarray/ndarray/blob/master/src/lib.rs#L115

(but I don't like that they re-export std as core because it's a lie and compilation errors look weird). 

This is also a recommended way in this book: https://www.lurklurk.org/effective-rust/no-std.html

Cargo.toml
antimora: OK. Agree. Done.

burn-common/Cargo.toml
antimora: We can keep the version number in Cargo's workspace definition just like this: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

burn-core/src/module/param/running.rs
antimora: Here is a the final version to record. The import's visibility needed pub(super) visibility. 

```rust

#[cfg(feature = "std")]
mod threading {
    pub(super) use std::collections::HashMap;
    pub(super) use std::sync::{Mutex, RwLock};
    pub(super) use std::thread::ThreadId;

    #[inline(always)]
    pub(super) fn get_thread_current_id() -> ThreadId {
        std::thread::current().id()
    }
}

#[cfg(not(feature = "std"))]
mod threading {
    pub(super) use burn_common::stub::{Mutex, RwLock, ThreadId};
    pub(super) use hashbrown::HashMap;

    #[inline(always)]
    pub(super) fn get_thread_current_id() -> ThreadId {
        panic!("Current thread id is not available")
    }
}

```

antimora: I like this approach because my vscode's rust-analyzer disables blocks if cfg yields false like this:

image

burn-no-std-tests/Cargo.toml
antimora: OK. I'll fix this. I discovered other packages have this way too and sometimes incorrect URLs.

antimora:
Fixed the issues (and more related). Ready for another review.

PR #178, opened 2023-02-26
Fix CI template that was causing error (burn-rs#168)

antimora:


antimora:
I submitted the PR but the build is failing. I am looking into this.

antimora:
The previous build error is fixed.  We can only build for wasm and arm targets, and not to test. That was the problem. Typo in my code.

Issue #180, opened 2023-02-26
burn-tch does not run/build with mac m1/m2 and linux aarch64


antimora:
**Describe the bug**
Build/test fails on burn-tch: `cargo test` on Mac M1/M2 and  aarch64 linux. 

**To Reproduce**
 on Mac M1/M2 or  aarch64 linux  run `cargo test` on ` burn-tch`

**OS Information**
```
bash: uname -a
Darwin MacBook-Pro.local 22.3.0 Darwin Kernel Version 22.3.0: 
Mon Jan 30 20:38:37 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T6000 arm64

```

```
[burn]$ uname -a
Linux e80e5571e6de 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022
 aarch64 aarch64 aarch64 GNU/Linux

```

antimora:
This could be related to: https://github.com/LaurentMazare/tch-rs/issues/542

antimora:
I filled a ticket with tch-rs team: https://github.com/LaurentMazare/tch-rs/issues/629

I found a solution that works for me. I'll codify my solution and check in.

antimora:
Here is a temporary workaround script (now as part of burn): https://github.com/burn-rs/burn/blob/main/scripts/fix-tch-build-arm64.py . This ticket will remain open till the problem is fixed in upstream libs: tch-rs and torch.

antimora:
The upstream has a fix for this but it's not released yet: https://github.com/LaurentMazare/tch-rs/pull/661

antimora:
These two PRs https://github.com/burn-rs/burn/pull/347 and https://github.com/burn-rs/burn/pull/345 close this ticket. We will have to remove the script once the release is made. We still can keep it because not all will upgrade at the same time.

PR #182, opened 2023-02-28
Add a workaround script for arm64 tch-rs build issue (#180)

antimora:


burn-tch/build.rs
```diff
@@ -0,0 +1,15 @@
+use std::env;
+fn main() {
```
nathanielsimard, line 2: This is so small, but I would like spaces between imports and functions 😅. I guess the `build.rs` is not formatted by `cargo fmt` ?

burn-tch/build.rs
```diff
@@ -0,0 +1,15 @@
...
+    // Temporary workaround for https://github.com/burn-rs/burn/issues/180
+    // Remove this once tch-rs upgrades to Torch 2.0 at least
+
+    if cfg!(all(target_arch = "aarch64", target_os = "macos")) {
+        let message = "Run scripts/fix-tch-build-arm64.py to fix environment variables for torch";
```
nathanielsimard, line 7: I would add the link to the issue in the comment, so it's clearer why this step is necessary.

nathanielsimard: And I guess most poeple using burn from crates.io won't have access to the script, so pointing to the link online is important.

nathanielsimard (changes requested):

burn-tch/build.rs
antimora: OK. Not sure why there is no space.

antimora:
Fixed the issues. Currently the CI is checking.

PR #197, opened 2023-03-06
Refactor/int backend

nathanielsimard:


burn-tensor/src/tensor/backend/base.rs
```diff
@@ -3,9 +3,51 @@ use alloc::string::String;
...
+/// which avoids copying the tensor's buffer. Note that it is still possible to mutate and
+/// reuse tensors' buffer without locking; see the next section on the Mutable API.
+///
+/// ### Mutable API
+///
+/// There is no mutable or inplace operation API to implement, but that does not mean that
+/// backends cannot support them. Using [try_unwrap](alloc::sync::Arc::try_unwrap) and
+/// [get_mut](alloc::sync::Arc::get_mut) allows backends to have access to an owned or mutable
+/// reference to their tensor buffer data structure if the tensor is not shared. In that case,
+/// backends can dispatch to their owned inplace operations for better performance.
```
antimora, line 46: Excellent explanation. I would love more insight like this. Perhaps, Burn deserves it's own Rust style book/documentation.

antimora (approved):
This is a lot of changes.  Thought I cannot comment on your design/approach, I looked through the code and didn't see any problems.

burn-tensor/src/tensor/backend/base.rs
nathanielsimard: I'll add more and more documentation like this. Maybe a Rust style book at some point will become necessary.

PR #238, opened 2023-03-16
Fix/batch norm

nathanielsimard:
Fix multiple problems that added up to bugs with batch norm:

- Fix batch norm 2d (problem with batch size reshaping) modified the tests with batch_size > 1 (that failed with the previous implementation).
- The `tch` crate is unsafe to use, even if not explicitly said so. Therefore, instead of wrapping the tensor handle inside an `Arc` and using the Rust rules for internal mutability, I'm wrapping the data pointer. I added some tests (that failed before the changes) where the original tensor would be changed even if multiple references were pointing to it.
- Updated the MNIST example to use batch norm.

### DIFF
--- a/burn-tch/Cargo.toml
+++ b/burn-tch/Cargo.toml
@@ -15,6 +15,8 @@ doc = ["tch/doc-only"]
 
 [dependencies]
 burn-tensor = {path = "../burn-tensor"}
+libc = "0.2.0"
+
 half = {workspace = true}
 rand = {workspace = true, features = ["std"]}
 log = {workspace = true}
--- a/burn-tch/Cargo.toml
+++ b/burn-tch/Cargo.toml
@@ -17,6 +17,7 @@ doc = ["tch/doc-only"]
 burn-tensor = {path = "../burn-tensor"}
 half = {workspace = true}
 rand = {workspace = true, features = ["std"]}
+log = {workspace = true}
 
 [target.'cfg(not(target_arch = "aarch64"))'.dependencies]
 tch = {version = "0.10.3"}
--- a/burn-tch/Cargo.toml
+++ b/burn-tch/Cargo.toml
@@ -19,7 +19,6 @@ libc = "0.2.0"
 
 half = {workspace = true}
 rand = {workspace = true, features = ["std"]}
-log = {workspace = true}
 
 [target.'cfg(not(target_arch = "aarch64"))'.dependencies]
 tch = {version = "0.10.3"}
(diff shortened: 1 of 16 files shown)


burn-tch/Cargo.toml
```diff
@@ -15,6 +15,8 @@ doc = ["tch/doc-only"]
 
 [dependencies]
 burn-tensor = {path = "../burn-tensor"}
+libc = "0.2.0"
```
antimora, line 18: Going sublevel, aren't we =)

examples/mnist/examples/mnist.rs
```diff
@@ -24,9 +24,15 @@ mod tch_gpu {
     pub fn run() {
         #[cfg(not(target_os = "macos"))]
         let device = TchDevice::Cuda(0);
+        #[cfg(not(target_os = "macos"))]
+        type FloatElement = burn::tensor::f16;
```
antimora, line 28: We need to test f16 on GPU without f16 native support. My linux  GPU (GeForce GTX 1080) does not have f16 native.  Maybe it works without overflow.

antimora (changes requested):
Wow, so many changes to fix one bug =)

Overall, it's looking good. I found a few minor issues that I could identify.

examples/mnist/examples/mnist.rs
nathanielsimard: I'll put f32 by default then, it's not a big deal :)

antimora (approved):
Looks good. 

I am running some training to synchronize models on for the web . It looks the number of params got doubled (to 633407). If it's okay with you, I'll adjust hyperparams a little.

burn-tch/src/tensor.rs
```diff
@@ -38,39 +69,51 @@ unsafe impl<E: tch::kind::Element, const D: usize> Send for TchTensor<E, D> {}
...
+    pub fn mut_ops<
+        F: Fn(&mut tch::Tensor) -> tch::Tensor,
+        EOut: tch::kind::Element,
+        const D_OUT: usize,
+    >(
+        &mut self,
+        func: F,
+    ) -> Option<TchTensor<EOut, D_OUT>> {
+        if Arc::strong_count(&self.storage) > 1 {
+            return None;
```
twitchax, line 82: Nice.

twitchax (approved):
LGTM.

The tests are going to help ensure the mutability rules are conserved going forward, which is a good peace of mind.

Issue #226, opened 2023-03-12
Proposal: a functional API for Burn


jmacglashan:
### Feature description

Build a functional API for burn, in which the gradient of the arguments of simple functions can be computed and in which
Rust structures themselves can be differentiable types.

You can find a repository of the proof of concept here:
https://github.com/jmacglashan/burn-functional

### Feature motivation
It is my personal opinion that for Rust to be a viable option for ML researchers,  Rust needs to have a first-class autodifferentiation library that keeps the key programming concepts close to the mathematical ideas researchers are thinking about: functions, gradients, and how you compute them. Inspired by previous AD efforts like [Swift for Tensorflow](https://github.com/tensorflow/swift) and [JAX](https://github.com/google/jax), I believe a functional API is one of the best ways to achieve that and can be accomplished by building on the strong and flexible foundation of Burn.

### Action items
The work I've done toward this API is a proof of concept. It's also something that could be defined in its own separate crate
without any other changes in Burn. However, I think it would be useful if Burn adopted an API like this, and I think
making it first class in Burn would go a long way towards getting ML researchers interested in using Rust.

As such, I'd like to hear feedback on this kind of API and interest in incorporating it into Burn. There are lots of rough edges in it right now. In fact, I learned how to use Rust Macros for the first time in working on it! Consequently, I would *not* immediately put it into Burn. But it is far enough along with enough examples of how to build on it that I think it's ready for feedback.

I'd like to know
* If there is interest in bringing an API like this into Burn
* What things you dislike about it
* Suggestions for how to improve it
* Milestones you'd want to reach before putting it in Burn proper, if there was interest in doing so.

If there is interest in putting this into Burn proper, I'd also be willing to spend time polishing it.

-------------------

# Description and explanation of proposal
Below I've included all my comments on the readme of the repo that shows how to use a proof of concept functional API for Burn and how the PoC works. The full repository can be found here: https://github.com/jmacglashan/burn-functional

## Computing the gradient of function argments
Let's start with a simple example, computing the gradient of a simple function that multiplies two tensors.
```rust
fn main() {
    type B = NdArrayBackend<f32>;  // choose a backend

    // Make some example random tensor data
    let x: Tensor<B, 2> = Tensor::random([3, 2], Distribution::Standard);
    let y: Tensor<B, 2> = Tensor::random([3, 2], Distribution::Standard);

    // Define a function we want to differentiate. with_grad will make a new function that computes the
    // the value and returns the gradients of its arguments
[...]

PR #250, opened 2023-03-24
Refactor/init modules

nathanielsimard:
Move the initialization of modules into an `init` method on their config struct. 

Init state should come later when #226 is more advanced.

A potential improvement would be to use a `config` function on the `Module` struct itself instead of using `new` on the `Config` struct.

```rust
let linear = LinearConfig::new(5, 5).with_bias(false).init(); // Now
let linear = Linear::config(5, 5).with_bias(false).init(); // After
```

To do this, we could need to move the `Config` derive to a more powerfull macro:

```rust
#[derive(Config)] // Now
struct LinearConfig {
...
}

#[config(module = "Linear")] // After
struct LinearConfig {
...
}
```

The second macro can receive parameters in it declaration telling us to create an alias function `Linear::config => LinearConfig::new`.

### DIFF
--- a/burn-core/src/module/state.rs
+++ b/burn-core/src/module/state.rs
@@ -338,7 +338,7 @@ mod tests {
     }
 
     pub fn create_model() -> nn::Linear<TestBackend> {
-        nn::Linear::<crate::TestBackend>::new(&nn::LinearConfig::new(32, 32).with_bias(true))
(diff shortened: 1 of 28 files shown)


antimora (approved):
Code looks a lot cleaner and less intimidating without seeing: `::new(&config` =)

PR #256, opened 2023-03-25
Text classification inference example

### LINKED ISSUE #165: Example: inference use case
Add an example on how to save the trained weights and how to load them.
This should include saving and loading config as well.

### LINKED ISSUE #254: How to inference text-classification and text-generation
How to inference text-classification and text-generation by generated model.bin ?

nathanielsimard:
How to run inference using text-classification example.

fix #254
fix #165

### DIFF
--- a/examples/text-classification/examples/text-classification-ag-news.rs
+++ b/examples/text-classification/examples/text-classification-ag-news.rs
@@ -1,14 +1,15 @@
-use burn::optim::{decay::WeightDecayConfig, momentum::MomentumConfig};
-use text_classification::{training::ExperimentConfig, AgNewsDataset};
+#[cfg(feature = "training")]
+fn training() {
+    use burn::optim::{decay::WeightDecayConfig, momentum::MomentumConfig};
+    use text_classification::{training::ExperimentConfig, AgNewsDataset};
 
-type Backend = burn_autodiff::ADBackendDecorator<burn_tch::TchBackend<f32>>;
+    type Backend = burn_autodiff::ADBackendDecorator<burn_tch::TchBackend<burn::tensor::f16>>;
 
-fn main() {
     let config = ExperimentConfig::new(
-        burn::nn::transformer::TransformerEncoderConfig::new(256, 512, 4, 4),
+        burn::nn::transformer::TransformerEncoderConfig::new(256, 1024, 8, 4).with_norm_first(true),
         burn::optim::SgdConfig::new(5.0e-3)
-            .with_weight_decay(Some(WeightDecayConfig::new(5e-5)))
-            .with_momentum(Some(MomentumConfig::new().with_nesterov(true))),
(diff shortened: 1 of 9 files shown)


examples/text-classification/examples/text-classification-ag-news.rs
```diff
@@ -1,14 +1,15 @@
-use burn::optim::{decay::WeightDecayConfig, momentum::MomentumConfig};
-use text_classification::{training::ExperimentConfig, AgNewsDataset};
+#[cfg(feature = "training")]
+fn training() {
+    use burn::optim::{decay::WeightDecayConfig, momentum::MomentumConfig};
+    use text_classification::{training::ExperimentConfig, AgNewsDataset};
 
-type Backend = burn_autodiff::ADBackendDecorator<burn_tch::TchBackend<f32>>;
+    type Backend = burn_autodiff::ADBackendDecorator<burn_tch::TchBackend<burn::tensor::f16>>;
```
antimora, line 6: f32 might be a safe choice for the examples. On mac it's definitely gets NaN loss. Also older Nvidia cards do not support f16. I yet to test on my NVidia GPU.

examples/text-classification/examples/text-classification-db-pedia.rs
```diff
@@ -1,14 +1,15 @@
-use burn::optim::{decay::WeightDecayConfig, momentum::MomentumConfig};
-use text_classification::{training::ExperimentConfig, DbPediaDataset};
+#[cfg(feature = "training")]
+fn training() {
+    use burn::optim::{decay::WeightDecayConfig, momentum::MomentumConfig};
+    use text_classification::{training::ExperimentConfig, DbPediaDataset};
 
-type Backend = burn_autodiff::ADBackendDecorator<burn_tch::TchBackend<f32>>;
+    type Backend = burn_autodiff::ADBackendDecorator<burn_tch::TchBackend<burn::tensor::f16>>;
```
antimora, line 6: Same comment regarding the f16.

antimora (changes requested):
I had a comment about f16 element type. I could test tomorrow if you'd like.

examples/text-classification/examples/text-classification-ag-news.rs
nathanielsimard: Yeah, but in the other hand I think it's important to demonstrate f16 capabilities, since most big models will be trained witht that. Maybe we should have a feature flag for that!

antimora: I agree we should demonstrate the capability. Putting the feature flag might be a good way doing it. Also it's more explicit and would make the users aware that they can use f16. We should mention this in the readme, and maybe in the main readme too? (someday in the book too: https://github.com/burn-rs/burn/issues/200) =D

Issue #138, opened 2022-12-26
Pretty print of Tensor


nathanielsimard:
### Feature description

Implement `std::fmt::Display` for `Tensor` and `BoolTensor` the same way for each `Backend` with basic metadata.
This could look something like the following, but it's just a draft.

```
Tensor {
  data: [[0., 0., 0.,  ..., 0., 0., 0.],
         [0., 0., 0.,  ..., 0., 0., 0.],
         [0., 0., 0.,  ..., 0., 0., 0.],
         ...,
         [0., 0., 0.,  ..., 0., 0., 0.],
         [0., 0., 0.,  ..., 0., 0., 0.],
         [0., 0., 0.,  ..., 0., 0., 0.]],
  shape:   [30, 30],
  device:  Cpu,
  backend: ndarray,
  dtype:   f32,
}
 ```

nopestack:
Is anyone already looking into this issue? If not I'd be happy to contribute.

nathanielsimard:
I don't know anyone who is working on this, so feel free to contribute.

nopestack:
@nathanielsimard what's the best place to put a test for this feature?

nopestack:
@nathanielsimard I've started working on https://github.com/burn-rs/burn/pull/146

I'd appreciate if you can point me to the right direction there.

nathanielsimard:
@nopestack I just replied in the draft pull request, the test is at the right place.

To display the data (values) we should use the `std::fmt::Display` implementation of the `Data` struct using the `to_string` method.
It's already implemented, but the output is just wrong. The refined implementation can be done in another pull request to reduce the scope.

sunny-g:
Took a stab [here](https://github.com/procyon-rs/showata/pull/29) to add [showata](https://github.com/procyon-rs/showata) support for `burn-tensor` for use in Jupyter; would be willing to collab! 

here's the current output which is likely too simple rn, but I'm quite new to ML/torch so 🤷 
![Screenshot 2023-03-11 at 12 40 55](https://user-images.githubusercontent.com/2055636/224505974-8bb519c6-1f5d-4607-af93-155beb5811f3.png)

nathanielsimard:
That's a good start! I think it's probably a good idea to start by implementing `core::fmt::Display` for the `Data` struct. I would be happy to review your work if you want to open a PR.

I'm not sure how we can generalize the function to any number of dimensions, but maybe a recursive strategy would work, since we only want to print nested 2D matrices after all.

[...]

PR #257, opened 2023-03-30
Pretty Print Tensors

agelas:
@nathanielsimard I'm new to Rust so I figured I'd give #138 a shot, Let me know if this is somewhere in the ballpark of what you were thinking. I only implemented a pretty print for 2D Tensor structs composed of Ints (or at least that's what I think I did 🙃), and I can generalize to other types later.


burn-tensor/src/tensor/api/base.rs
```diff
@@ -168,6 +168,43 @@ where
...
+{
+    fn to_nested_vec(&self) -> Vec<Vec<B::IntElem>> {
+        let data = self.to_data();
+        let mut result = vec![vec![B::IntElem::default(); self.dims()[1]]; self.dims()[0]];
+        for (i, val) in data.value.iter().enumerate() {
+            let row = i / self.dims()[1];
+            let col = i % self.dims()[1];
+            result[row][col] = *val;
+        }
+        result
```
antimora, line 183: This could be made as a separate standalone function that can take the multi dimensional tensor and return formatted string string. This way you don't need to worry about nested vectors. 

The challenging part is to convert the flat data into nested array. But I am sure it's doing. Here is an example output how numpy does:


```
>>> np.ones([15, 15,15, 15])
array([[[[1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         ...,
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.]],

        [[1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         ...,
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.]],

        [[1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         ...,
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.]],

        ...,

        [[1., 1., 1., ..., 1., 1., 1.],
         [1., 1., 1., ..., 1., 1., 1.],

```

antimora (changes requested):

burn-tensor/src/tensor/api/base.rs
```diff
@@ -168,6 +168,43 @@ where
...
+    }  
+}
+
+/// Pretty print 2D Int tensors
+impl<B> std::fmt::Display for Tensor<B, 2, Int>
```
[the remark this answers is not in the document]
agelas: I can certainly try

burn-tensor/src/tensor/api/base.rs
```diff
@@ -168,6 +168,43 @@ where
...
+where
+    B: Backend,
+    B::IntElem: std::fmt::Display,
+{   
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Tensor {{")?;
+        writeln!(
+            f,
+            " data: {:?},",
```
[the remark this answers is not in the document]
agelas: Sorry, not sure what you mean. I thought the whole point was to print the data. Or do you mean that there's no need for the actual text "data" to show up when printing?

burn-tensor/src/tensor/api/base.rs
agelas: My current strategy is to try and recursively format the tensor as a 2D vector for each dimension, indenting based on the dimensionality level. It's still a work in progress because I need to figure out how to properly slice up tensors.

burn-tensor/src/tensor/api/base.rs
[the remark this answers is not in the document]
antimora: Sorry, I was reviewing it using my phone. I thought I saw a data is being printed but you're reformatting it.

burn-tensor/src/tensor/api/base.rs
```diff
@@ -168,6 +173,93 @@ where
...
+        
+        let mut acc = String::new();
+        let mut multi_index = vec![0; D];
+        self.display_recursive(&mut acc, 0, &mut multi_index);
+        write!(f, "{}", acc)?;
+        writeln!(f, ",")?;
+        writeln!(f, "  shape:   {:?},", self.dims())?;
+        writeln!(f, "  device:  {:?},", self.device())?;
+        writeln!(f, "  backend: {:?},", B::name())?;
+        writeln!(f, "  dtype:   {:?},", "int")?; // this is probably cheating
```
agelas, line 258: @antimora @nathanielsimard Ok I think I have something that works, but still only for Int tensors. Is there a good way to get the dtype out of the Backend struct that I'm missing?

nathanielsimard: You have to specify the kind as a generic parameter to the tensor struct.

```rust
impl<B: Backend, const D: usize, K: BasicOps<B>> Tensor<B, D, K> {
   // K::Elem is an associative type, maybe we will it to implement core::fmt::Display
}
```
see https://github.com/burn-rs/burn/blob/main/burn-tensor/src/tensor/api/base.rs for more details.

burn-tensor/src/tensor/api/base.rs
antimora: @agelas @nathanielsimard 

I came across this [project serde-dim](https://github.com/RReverser/serde-ndim) that looks interesting. Maybe there is a something you can learn from it: https://github.com/RReverser/serde-ndim/blob/main/src/ser.rs . What this project is trying to accomplish is basically convert one dimensional data array into multi dimensional array. Pretty what you're trying to accomplish.

agelas: @antimora Looks interesting, I'll give it a look. But your description seems almost the opposite of what this PR is for, ie we're going from multi-dimensional array -> "1-dimensional" string representation and not the other way around.

antimora: > @antimora Looks interesting, I'll give it a look. But your description seems almost the opposite of what this PR is for, ie we're going from multi-dimensional array -> "1-dimensional" string representation and not the other way around.

The tensor data is stored as one-dimensional vector and shape contains the dimensions so you know how to decompose into rows and columns. The link I shared basically is trying to take the serialized data (1d) and deserialize into n-dimensional array by recursion. By coincidence I came across this code and I thought there might be something to learn. I haven't dived deep to see if this direct helps you but in principle by its description it should. I hope it does not confuse you =)

burn-tensor/src/tensor/api/base.rs
```diff
@@ -39,6 +39,11 @@ where
         K::shape(&self.primitive)
     }
 
+    /// Returns size of the current tensor.
+    pub fn size(&self) -> usize {
+        self.dims().iter().fold(1, |acc, &dim| acc * dim)
+    }
```
nathanielsimard, line 45: There is a similar method `num_elements` in the `Shape` struct, I would use it instead here, but I like the fold implementation, so we may update the `num_elements` function implementation to use `fold` instead of loops. `size` is a bit unclear as a function name, since size often return the `shape` in PyTorch, so we may update the name to `num_elements`.

burn-tensor/src/tensor/api/base.rs
```diff
@@ -168,6 +173,217 @@ where
     }
 }
 
+impl<B, const D: usize> Tensor<B, D, Int>
+where
+    B: Backend,
+{
```
nathanielsimard, line 179: I think you should make the function generic over the kind with `K: BasicOps`. The same code can be used for all tensor kinds.

burn-tensor/src/tests/stats/basic.rs
```diff
@@ -13,4 +13,58 @@ mod tests {
...
+        expected = "Tensor {\n  data: [[1.1, 2.2, 3.3], [4.4, 5.5, 6.6], [7.7, 8.8, 9.9]],\n  shape:   [3, 3],\n  device:  Cpu,\n  backend: \"ndarray\",\n  dtype:   \"float\",\n}";
+        assert_eq!(output, expected);
+
+        let data_bool = Data::from([[true, false, true], [false, true, false], [false, true, true]]);
+        let tensor_bool: burn_tensor::Tensor<TestBackend, 2, burn_tensor::Bool> =
+            Tensor::from_data(data_bool);
+
+        output = format!("{}", tensor_bool);
+        expected = "Tensor {\n  data: [[true, false, true], [false, true, false], [false, true, true]],\n  shape:   [3, 3],\n  device:  Cpu,\n  backend: \"ndarray\",\n  dtype:   \"bool\",\n}";
+        assert_eq!(output, expected);
```
nathanielsimard, line 40: I think it's a good idea to have tests for all tensor kind even if their implementation are the same, at least for the basic 2d case. However, I would put each test in its own function like `test_display_2d_float_tensor`.

nathanielsimard (changes requested):
This is starting to look good! I think we should make the implementation generic over the kind, and with the other minor requested changes we could merge the PR pretty soon.

burn-tensor/src/tensor/api/base.rs
agelas: Actually I didn't mean to leave this in- I needed something like this for an earlier implementation I was trying out. We can remove it, rename it, or do whatever honestly 🤣

burn-tensor/src/tensor/api/base.rs
agelas: Ah that's how you do it. I thought there was a better way than defining an `impl ` for everything. Thanks!

burn-tensor/src/tensor/api/base.rs
```diff
@@ -168,13 +173,103 @@ where
...
+}
+
 /// Trait that list all operations that can be applied on all tensors.
 ///
 /// # Warnings
 ///
 /// This is an internal trait, use the public API provided by [tensor struct](Tensor).
 pub trait BasicOps<B: Backend>: TensorKind<B> {
-    type Elem;
+    type Elem: 'static;
```
agelas, line 272: @nathanielsimard Will doing this impact anything? From my understanding, adding the `'static` means any type implementing `BasicOps` must have a static lifetime for its Elem type, which shouldn't be a problem since numeric types like i32, f32, etc., have a static lifetime..

burn-tensor/src/tensor/api/base.rs
```diff
@@ -212,6 +307,19 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
         rhs: Self::Primitive<D>,
     ) -> Tensor<B, D, Bool>;
     fn equal_elem<const D: usize>(lhs: Self::Primitive<D>, rhs: Self::Elem) -> Tensor<B, D, Bool>;
+    fn elem_type() -> &'static str {
+        if TypeId::of::<Self::Elem>() == TypeId::of::<f32>() {
```
agelas, line 311: @nathanielsimard Also here I assume I'm missing a bunch of other possibilities like f64 for float, or i8, i16, i128, etc. for ints. What else should I add?

burn-tensor/src/tests/stats/basic.rs
agelas: Fixed!

burn-tensor/src/tensor/api/base.rs
nathanielsimard: Yes this is fine!

burn-tensor/src/tensor/api/base.rs
nathanielsimard: You can use directly `core::any::type_name::<Seld::Elem>()` and all types will be supported :)

agelas: Do you want the more specific type that comes from `core::any::type_name::<Self::Elem>()`, or should I group the possible returns as int or float?

nathanielsimard: The more specific type is needed, but we could also add the tensor kind:

```
  data: [...],
  kind: Float, // We could add a function `name` in the trait TensorKind.
  elem: f32
```

agelas: Sounds good, just threw kind in as well.

agelas:
@nathanielsimard I think this is pretty much done unless you can think of anything else. I just merged main so the branch is all up to date as well.

nathanielsimard:
@agelas I think the only missing part is fixing the CI (mostly no_std support by importing stuff from alloc) and remove the `size` function from the tensor API.

agelas:
@nathanielsimard Ok should be good now, although locally I can't seem to build because the compiler is complaining about the `"use of unstable library feature 'unzip_option': recently added"` in `burn-core/src/optim/simple/adaptor.rs:86:63`. Not sure if that will trip up the CI as well.

nathanielsimard:
> @nathanielsimard Ok should be good now, although locally I can't seem to build because the compiler is complaining about the `"use of unstable library feature 'unzip_option': recently added"` in `burn-core/src/optim/simple/adaptor.rs:86:63`. Not sure if that will trip up the CI as well.

@agelas I'm curious, what is your local Rust version? Maybe I used a newly stabilized function, and I should update the minimum required Rust version in the readme. The CI is always using the latest stable Rust version.

agelas:
@nathanielsimard Oh I'm on 1.65.0, that might explain it.

nathanielsimard:
@agelas You can run clippy and cargo fmt with the lastest Rust to fix most problems.

```bash
cargo clippy --fix
cargo fmt --all
```

agelas:
@nathanielsimard clippy and fmt didn't seem to change anything. I added a few more imports from `alloc ` though hopefully that should fix it.

agelas:
@nathanielsimard Ok we're getting a little bit closer 😆. The test-burn-ndarray one is strange. I've noticed that sometimes the backend comes back as `ndarray`, sometimes as `tch`, do you know what might be up with that?

nathanielsimard:
> @nathanielsimard Ok we're getting a little bit closer laughing. The test-burn-ndarray one is strange. I've noticed that sometimes the backend comes back as `ndarray`, sometimes as `tch`, do you know what might be up with that?

I think, the tests in ndarray will have ndarray as backend and the ones in tch will have tch. But backends can be used as dev dependencies just to run the tests, so it might change. I would sugest to update the test to use the `TestBackend` name directly.

agelas:
> I think, the tests in ndarray will have ndarray as backend and the ones in tch will have tch. But backends can be used as dev dependencies just to run the tests, so it might change. I would sugest to update the test to use the `TestBackend` name directly.

Sorry, not quite sure what you mean by using the `TestBackend` name directly. Do you mean explicitly doing something like `let TestBackend = burn_ndarray::NdArrayBackend<f32>`?

nathanielsimard:
> Sorry, not quite sure what you mean by using the `TestBackend` name directly. Do you mean explicitly doing something like `let TestBackend = burn_ndarray::NdArrayBackend<f32>`?

No I mean in the test, instead of hardcoding the name of the backward in the expected string, maybe use a `format!("... backend: {:?}...", TestBackend::name())`

agelas:
@nathanielsimard I'm just gonna remove the doctest lol.

agelas:
@nathanielsimard alright this time I'm 99% sure it should pass

PR #258, opened 2023-03-31
Feat: add new tensor ops mask_scatter

bioinformatist:
Since there's no operation for filling elements by values from another tensor by mask, which will be needed for implementation of softplus, I have written this new operation.

However, considering my poor experience with automatic differentiation, please be careful with the mathematical details this time.

### DIFF
--- a/burn-autodiff/src/ops/tensor.rs
+++ b/burn-autodiff/src/ops/tensor.rs
@@ -672,6 +672,56 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
         }
     }
 
+    fn mask_scatter<const D: usize>(
+        tensor: ADTensor<B, D>,
+        mask: BoolTensor<B, D>,
+        source: ADTensor<B, D>,
+    ) -> ADTensor<B, D> {
+        #[derive(Debug)]
+        struct MaskScatter;
+
+        impl<B: Backend, const D: usize> Backward<B, D, 2> for MaskScatter {
+            type State = (BoolTensor<B, D>, Shape<D>, Shape<D>, B::Device);
+
+            fn backward(self, ops: Ops<Self::State, 2>, grads: &mut Gradients) {
+                let (mask, shape_lhs, shape_rhs, device) = ops.state;
+                let [mask_4lhs, mask_4rhs] = duplicate(&ops.parents, Some(mask));
+
+                binary::<B, D, D, D, _, _>(
+                    ops.parents,
+                    ops.node,
+                    grads,
+                    |grad| {
+                        let zeros = B::zeros(shape_lhs, &device);
+                        B::mask_scatter(grad, mask_4lhs.unwrap(), zeros)
+                    },
+                    |grad| {
+                        let zeros = B::zeros(shape_rhs, &device);
+                        B::mask_scatter(zeros, mask_4rhs.unwrap(), grad)
+                    },
+                );
+            }
+        }
+
+        match MaskScatter
+            .prepare([tensor.node, source.node], [tensor.graph, source.graph])
+            .statefull()
+        {
+            OpsKind::Tracked(prep) => prep.finish(
+                (
+                    mask.clone(),
+                    B::shape(&tensor.primitive),
+                    B::shape(&source.primitive),
+                    B::device(&source.primitive),
+                ),
+                B::mask_scatter(tensor.primitive, mask, source.primitive),
+            ),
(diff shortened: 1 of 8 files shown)


burn-tch/src/ops/tensor.rs
```diff
@@ -221,6 +221,20 @@ impl<E: TchElement> TensorOps<TchBackend<E>> for TchBackend<E> {
...
 
+    fn mask_scatter<const D: usize>(
+        tensor: TchTensor<E, D>,
+        mask: TchTensor<bool, D>,
+        source: TchTensor<E, D>,
+    ) -> TchTensor<E, D> {
+        TchTensor::binary_ops_tensor(
+            tensor,
+            source,
+            |tensor, source| tensor.f_masked_scatter_(&mask.tensor, source).unwrap(),
```
nathanielsimard, line 232: You may have to cast to `Kind::Bool` like with `mask_fill`, you may add a test to check if it's needed?

burn-autodiff/src/ops/tensor.rs
```diff
@@ -672,6 +672,56 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
...
+                    ops.parents,
+                    ops.node,
+                    grads,
+                    |grad| {
+                        let zeros = B::zeros(shape_lhs, &device);
+                        B::mask_scatter(grad, mask_4lhs.unwrap(), zeros)
+                    },
+                    |grad| {
+                        let zeros = B::zeros(shape_rhs, &device);
+                        B::mask_scatter(zeros, mask_4rhs.unwrap(), grad)
```
nathanielsimard, line 700: I don't believe this is correct, I think the gradient of the LHS should be the gradient of the input tensor with a mask scatter using the same mask as input, but filled with the gradients of the RHS instead of zeros. And it's the reverse for the RHS.

burn-tch/src/ops/tensor.rs
bioinformatist: It seems there's no need for casting when return value is a `TchTensor<TchElement>`, while those with `TchTensor<bool>` as return type should be converted to `tch::Kind::Bool` in clousures standing for mutable. And does it already covered in the tests across `burn-tensor` and `burn-tensor`?

I'm not sure if there's any misunderstanding:cold_sweat:

burn-autodiff/src/tests/mask.rs
```diff
@@ -23,4 +23,32 @@ mod tests {
...
+    fn should_diff_mask_scatter() {
+        let data_1 = Data::<f32, 2>::from([[1.0, 7.0], [2.0, 3.0]]);
+        let data_2 = Data::<f32, 2>::from([[4.0, 7.0], [2.0, 3.0]]);
+        let mask = Data::<bool, 2>::from([[true, false], [false, true]]);
+
+        let data_source = Data::<f32, 2>::from([[8.8, 8.8], [8.8, 8.8]]);
+
+        let tensor_1 = TestADTensor::from_data(data_1).require_grad();
+        let tensor_2 = TestADTensor::from_data(data_2).require_grad();
+        let tensor_source = TestADTensor::from_data(data_source);
```
nathanielsimard, line 37: We should also validate the gradients for `tensor_source` with `require_grad()`. Also, did you validate the output values with a similar PyTorch program (just to be sure)?. I'm not sure of the backward pass and maybe your first implementation was right!

nathanielsimard (changes requested):

burn-autodiff/src/tests/mask.rs
bioinformatist: Just let me have a try with PyTorch:blush:

burn-autodiff/src/ops/tensor.rs
nathanielsimard: In term of memory management, everything seems just fine :) _

burn-autodiff/src/tests/mask.rs
bioinformatist: I've checked the gradients in PyTorch for the same scenario:

image

It seems the previous version for backward pass was right, so I revert it back.

Also the `tensor_source` has been labelled as `require_grad()` in its related test now.

nathanielsimard (approved):
The last step is fixing the merge conflicts and the CI. The conflicts are probably about news functions that were added at the same place, so nothing hard to fix.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #259, opened 2023-03-31
Refactor Param wrapping only for Tensor

nathanielsimard:
I was working on refactoring the entire API based on discussion #226, but the task was too large and the solution was not satisfactory. However, it helped me understand the problem more deeply, and I am now planning to split the work into smaller chunks, with each piece of work improving the API. This pull request is part of that series.

* The `Optimizer` trait is now bound to a `Module` and a `Backend` instead of having associative types. This change is helpful since modules no longer have a backend, and as discussed in issue #226, optimizers should have a higher cohesion with modules.

* The `Module` trait no longer has a backend as an associative type, which allows any type to implement it. As proposed in the same issue, everything should implement the same basic trait, even constants, and this is now the case.

* The `Param` struct is only necessary to wrap `Tensor`, giving them a `ParamId`. While it may become unnecessary in future work, multiple steps need to be taken before that happens.


antimora (approved):
Looks good.

PR #262, opened 2023-04-01
Feat/record

nathanielsimard:
This PR will help with saving states. I added the `Record`, `Recorder` and `RecordSettings` trait to abstract *what* is saved and *how* it's saved. The `RecordSettings` trait only contains associative type, so everything is static.

✅ Decouple the element types used for training and inference from the ones used for serialization.
  * You can save the model weights in any element type even if the backend don't support it.
  
✅ Introduce a way to set the data version.
✅ Will enable optimizers and modules to define their record type with an associative type, giving them the control to define how they must be serialized.
✅ Add metadata to each serialized record. When it's possible, we augment the error message showing it.


antimora (changes requested):
One small typo that needs to be corrected. Overall it looks good. I need to try the new API for ONNX and see if it works for this use case.

antimora (approved):

antimora (approved):
Approved

PR #275, opened 2023-04-06
Fix CI workflows

### LINKED ISSUE #263: Need to upgrade CI actions soon: Node.js 12 actions are deprecated
See action result warning: https://github.com/burn-rs/burn/actions/runs/4589033931

image

### LINKED ISSUE #266: CI tasks should point to head template
Currently when template and ci task are changed at the same time, CI fails because it uses template from the main branch instead of head.

antimora:
1. workflows to point to templates from HEAD (Fixes #266)
2. Update workflows to use dtolnay/rust-toolchain (Fixes #263)

### DIFF
--- a/.github/workflows/publish-template.yml
+++ b/.github/workflows/publish-template.yml
@@ -13,12 +13,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: install rust
-        uses: actions-rs/toolchain@v1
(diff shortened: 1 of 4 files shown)


nathanielsimard (commented):
It seems OK, but are the tests working?

antimora:
> It seems OK, but are the tests working?

Pointing to the local file isn't working. According to this: https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow we should be able to specify local files. I am trying to fix it.

antimora:
Closing this because I could not get the local template to be used instead of repo/file@main format.

PR #276, opened 2023-04-07
Feat/lr scheduler

### LINKED ISSUE #142: Learning Rate Scheduler
### Feature description

Introduce the learning rate scheduler concept to control how the learning rate evolves during training.

nathanielsimard:
fix #142 

## Decisions

Optimizers should not have the learning rate in their configuration since it's not something opaque that should be hidden by implementations. Some frameworks use object oriented patterns with get and set to mutate the optimizer state during training when a learning rate scheduler is used. However I find this error prone and not very Rusty. Therefore, the learning rate is instead passed as parameter to the optimizer when performing a step. Custom training loop can do whatever they want to control the learning rate, but a common way is using a learning rate scheduler, though I created the `LearningRateScheduler` trait. It is quite similar to the `Optimizer` and `Module` trait such as it defines a `Record` associative type to save and load its state.

In order to add the learning rate metric, I had to add the `MetricMetadata` parameter in the `Metric` trait. Those metadata can be used to implement metrics which are not specific to a model output types.

## Questions

I'm still unsure if the `Optimizer` and `LearningRateSchedule` step method should be taking a mutable reference or taking ownership and returning itself.

I'm still unsure if we should use `learning_rate` or the common `lr` abreviation. Learning rate is quite long and since `lr` is commondly used in many frameworks, it might be the right name.


burn-core/src/optim/adam.rs
```diff
@@ -44,6 +41,7 @@ impl<B: Backend> SimpleOptimizer<B> for Adam<B> {
 
     fn step<const D: usize>(
         &self,
+        learning_rate: f64,
```
antimora, line 44: for consistency, should use `LearningRate` instead of 64.

burn-core/src/optim/adam.rs
```diff
@@ -173,14 +170,16 @@ mod tests {
     use crate::tensor::{Data, Distribution, Tensor};
     use crate::{nn, TestADBackend, TestBackend};
 
+    static LEARNING_RATE: f64 = 0.01;
```
antimora, line 173: Probably const is better since the value is used to initialize and not make it available globally. Also, for consistency, should use `LearningRate` instead of 64.

burn-core/src/optim/sgd.rs
```diff
@@ -58,6 +53,7 @@ impl<B: Backend> SimpleOptimizer<B> for Sgd<B> {
 
     fn step<const D: usize>(
         &self,
+        learning_rate: f64,
```
antimora, line 56: for consistency, should use `LearningRate` instead of 64.

burn-core/src/optim/sgd.rs
```diff
@@ -105,14 +101,16 @@ mod tests {
         TestADBackend, TestBackend,
     };
 
+    static LEARNING_RATE: f64 = 0.02;
```
antimora, line 104: Minor comment, since it's the test, about const vs static comment.

burn-core/src/optim/simple/adaptor.rs
```diff
@@ -43,9 +43,13 @@ where
 {
     type Record = HashMap<ParamId, AdaptorRecord<O, B::InnerBackend>>;
 
-    fn step(&mut self, module: M, mut grads: GradientsParams) -> M {
-        let mut mapper =
-            SimpleOptimizerMapper::<M, B, O>::new(&self.optim, &mut self.records, &mut grads);
+    fn step(&mut self, learning_rate: f64, module: M, mut grads: GradientsParams) -> M {
```
antimora, line 46: for consistency, should use `LearningRate` instead of 64.

burn-core/src/optim/simple/adaptor.rs
```diff
@@ -69,6 +73,7 @@ where
     optimizer: &'a O,
     records: &'a mut HashMap<ParamId, AdaptorRecord<O, B::InnerBackend>>,
     grads: &'a mut GradientsParams,
+    learning_rate: f64,
```
antimora, line 76: for consistency, should use `LearningRate` instead of 64.

burn-core/src/optim/simple/base.rs
```diff
@@ -19,6 +19,7 @@ where
     /// the saving and loading of recorded states.
     fn step<const D: usize>(
         &self,
+        learning_rate: f64,
```
antimora, line 22: for consistency, should use `LearningRate` instead of 64.

burn-train/src/callback/base.rs
```diff
@@ -14,4 +14,5 @@ pub struct LearnerItem<T> {
     pub epoch: usize,
     pub epoch_total: usize,
     pub iteration: usize,
+    pub learning_rate: Option<f64>,
```
antimora, line 17: for consistency, should use `LearningRate` instead of 64.

burn-train/src/metric/base.rs
```diff
@@ -1,3 +1,14 @@
+use burn_core::data::dataloader::Progress;
+
+/// Metric metadata that can be used when computing metrics.
+pub struct MetricMetadata {
+    pub progress: Progress,
+    pub epoch: usize,
+    pub epoch_total: usize,
+    pub iteration: usize,
+    pub learning_rate: Option<f64>,
```
antimora, line 9: for consistency, should use `LearningRate` instead of 64.

burn-train/src/metric/dashboard/base.rs
```diff
@@ -179,12 +193,12 @@ where
 }
 
 trait DashboardNumericMetric<T>: Send + Sync {
-    fn update(&mut self, item: &LearnerItem<T>) -> (MetricEntry, f64);
+    fn update(&mut self, item: &LearnerItem<T>, metadata: &MetricMetadata) -> (MetricEntry, f64);
```
antimora, line 196: for consistency, should use `LearningRate` instead of 64.

burn-train/src/metric/dashboard/base.rs
```diff
@@ -199,8 +213,8 @@ where
     M: Metric + Numeric + 'static,
     T: Adaptor<M::Input>,
 {
-    fn update(&mut self, item: &LearnerItem<T>) -> (MetricEntry, f64) {
-        let update = self.metric.update(&item.item.adapt());
+    fn update(&mut self, item: &LearnerItem<T>, metadata: &MetricMetadata) -> (MetricEntry, f64) {
```
antimora, line 216: for consistency, should use `LearningRate` instead of 64.

burn-train/src/metric/learning_rate.rs
```diff
@@ -0,0 +1,48 @@
...
+        )
+    }
+
+    fn clear(&mut self) {
+        self.state.reset()
+    }
+}
+
+impl Numeric for LearningRateMetric {
+    fn value(&self) -> f64 {
```
antimora, line 45: for consistency, should use `LearningRate` instead of 64.

antimora (changes requested):
Minor updates for consistency sake.

burn-train/src/metric/dashboard/base.rs
nathanielsimard: This is not the LearningRate, this is any numeric value that can be plotted on the terminal.

burn-train/src/metric/dashboard/base.rs
nathanielsimard: Same here

burn-train/src/metric/learning_rate.rs
nathanielsimard: Same here, it's just a numeric value

antimora (approved):

PR #283, opened 2023-04-10
Feat/add tensor checks

nathanielsimard:
# Example

```rust
let _ = tensor_2d.index([1..1, 0..2, 0..2]);
````

```bash
thread 'tests::index::tests::should_support_partial_indexing_2d' panicked at '=== Tensor Operation Error ===
  Operation: 'Index'
  Reason:
    1. The provided indexes array has a higher number of dimensions than the current tensor. The indexes array must be smaller or equal to the tensor number of dimensions. Tensor number of dimensions: 2, indexes array lenght 3. 
    2. The provided indexes array has a range where the start index is bigger or equal to its end. The range at dimension '0' starts at '1' and is greater or equal to its end '1'. Tensor shape [2, 3], provided indexes [1..1, 0..2, 0..2]. 
', /home/nathaniel/Projects/burn/burn-tensor/src/tensor/api/base.rs:165:9
```


antimora:
We need to have this pre-commit script so the developers can run before a commit: https://github.com/burn-rs/burn/issues/284

PR #285, opened 2023-04-11
fix: text classification example

nathanielsimard:


antimora (approved):
OK. I have tested both examples on Mac and they are working now.

PR #292, opened 2023-04-11
Fix/text gen example

nathanielsimard:


antimora (approved):

antimora:
Tested on Mac. It's working. I didn't wait it to finish because it takes way too long.

PR #300, opened 2023-04-17
Feat/conv stride

nathanielsimard:
- [x] Support stride convolution backward
- [X] Support transposed convolution forward

Support for transposed convolution backward will be done in a following PR.

Since transposed convolution backward is not implemented, it's not available as a module yet, it should be exposed when auto diff is fully supported.


burn-autodiff/src/ops/module.rs
```diff
@@ -158,6 +158,28 @@ impl<B: Backend> ModuleOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
...
 
+    fn conv_transpose2d(
+        _x: ADTensor<B, 4>,
+        _weight: ADTensor<B, 4>,
+        _bias: Option<ADTensor<B, 1>>,
+        _stride: [usize; 2],
+        _padding: [usize; 2],
+        _out_padding: [usize; 2],
+    ) -> ADTensor<B, 4> {
+        panic!("Transposed 2D convolution doesn't yet support backward.");
```
antimora, line 169: Something I learned recently: using `todo!` macro panics and some IDEs convey as a todo (maybe also useful to search the code)

antimora (approved):

PR #305, opened 2023-04-22
Gradient Clipping

agelas:
@nathanielsimard I'm creating a draft before I get too far into this to make sure this is heading the right direction. If you could take a look at my questions that would be much appreciated!


burn-core/src/grad_clipper/base.rs
```diff
@@ -0,0 +1,69 @@
...
+        match (self.clip_value, self.clip_norm) {
+            (Some(clip_value), None) => self.clip_by_value(grad, clip_value),
+            (None, Some(clip_norm)) => self.clip_by_norm(grad, clip_norm),
+            _ => panic!("Gradient Clipper must have one (and only one) clip_value or clip_norm."),
+        }
+    }
+
+    fn clip_by_value<B: Backend, const D: usize>(&self, grad: Tensor<B, D>, threshold: f32) -> Tensor<B, D> 
+    where
+        B::FloatElem: core::cmp::PartialOrd<f32> + From<f32>, // this doesn't feel right
```
agelas, line 36: @nathanielsimard I know that the gradients coming in are going to be made up of f32's, but it seems like this is not a very burn-ish way of constraining the inputs. Do you have any recommendations?

burn-core/src/grad_clipper/base.rs
```diff
@@ -0,0 +1,69 @@
...
+        let squared = tensor.clone().powf(2.0);
+        let sum = squared.sum();
+        let norm = sum.sqrt();
+        norm
+    }
+
+    fn clip_by_norm<B: Backend, const D: usize>(&self, grad: Tensor<B, D>, threshold: f32) -> Tensor<B, D> {
+        let norm = Self::l2_norm(&grad);
+        let norm_float = norm.into_scalar();
+        if norm_float > threshold { // FloatElem can't be compared to f32?
```
agelas, line 61: @nathanielsimard Also is there a way to compare FloatElem's to f32? Or does PartialOrd need to be implemented somewhere to allow for this.

burn-core/src/grad_clipper/base.rs
```diff
@@ -0,0 +1,69 @@
+use burn_tensor::{backend::Backend};
+use crate::tensor::Tensor;
+/// Gradient Clipper provides a way to mitigate the exploding gradients
+/// problem by clipping every component of the gradient to a value during 
+/// backpropagation.
+pub struct GradientClipper {
+    clip_value: Option<f32>,
+    clip_norm: Option<f32>,
+}
```
nathanielsimard, line 9: I think an enum is better in this case.

burn-core/src/grad_clipper/base.rs
nathanielsimard: You have the type `B::FloatElem` in the backend. You can also convert any numeric types into `B::FloatElem` using `elem()` function. You have to import `ElementConversion` trait I think. I would probably keep f32 in the struct and convert is when clip is called.

burn-core/src/grad_clipper/base.rs
nathanielsimard: You need to use element conversion to have the same type.

burn-core/src/optim/sgd.rs
```diff
@@ -57,6 +58,7 @@ impl<B: Backend> SimpleOptimizer<B> for Sgd<B> {
         tensor: Tensor<B, D>,
         mut grad: Tensor<B, D>,
         state: Option<Self::State<D>>,
+        gradient_clip: Option<GradientClipper>,
```
nathanielsimard, line 61: I'm not sure if you should put gradient clipper in the optimizer or in the training loop. What is your opinion on this?

agelas: That's a good point. With TensorFlow I know that you generally pass in your gradient clipping values to your optimizer with something like `optimizer = SGD(lr=0.01, momentum=0.9, clipvalue=1.0)`, at least if you're using Keras. With PyTorch though I think it's in the training loop. 

Attaching it the optimizer seems to be more simple and allows users to do a little less setup in the training loop. Also making it part of the optimizer directly makes things a bit more consistent during optimization. 

On the other hand, making it a separate function called during the training loop makes it more flexible and allows you to control when it's applied. So that could be more useful when working with highly customized or complicated training loops that need very specific tuning. 

Not sure what you're preference is. We can also just do both lol

nathanielsimard: If we do it in the optimizer, I think it should be declared in the configuration of each optimizer, a bit like weights decay is. The gradient clipper can be used standalone if someone wants to implement their own training loop, nobody is forcing gradient clipper to be used exclusively by optimizers, it's just a convenient configuration that can be turned on.

So I think we can keep the gradient clipper in the optimizer, it's convenient and doesn't really block any complex use case.

burn-core/src/optim/simple/adaptor.rs
```diff
@@ -45,8 +60,22 @@ where
...
-        let mut mapper =
-            SimpleOptimizerMapper::<M, B, O>::new(&self.optim, &mut self.records, &mut grads, lr);
+        if let Some(ref g_clipper) = self.gradient_clipper {
+            grads.iter_mut().for_each(|(id, grad)| {
+                let grad_tensor: Tensor<B, D> =
+                    grad.downcast().expect("Failed to downcast gradient");
+                let clipped_grad = g_clipper.clip_gradient(grad_tensor);
+                grads.register(id.clone(), clipped_grad);
+            })
+        }
```
agelas, line 70: @nathanielsimard Can I get your opinion on the best way to do this/if this is necessary for SimpleOptimizer? I think that if there's a gradient clipper being used with a simple optimizer, it's best and easiest to apply it to all the gradients. But I'm not sure what the best way to iterate over and subsequently clip all those gradients is. I've added iter_mut implementations to TensorContainer and GradientsParams because it doesn't look like there's a way to get all the gradients.

nathanielsimard: I don't think it's a good idea to iterate over all tensor to apply the clipper, since tensor are of different types (different dimensions) and you have to use the parameter id linked to the module to determine witch tensor type it is.

I think you can just call the clipper in the `map` function of the `SimpleOptimizerMapper` struct, just before `optimizer.step` is called, or directly in the `step` function of each optimizer implementation.

agelas:
@nathanielsimard Still need to add in some unit tests and clear out the `iter_mut`'s that I threw in, but other than that I think it's close.

burn-core/src/grad_clipper/base.rs
nathanielsimard: Is there a reason you kept the option?

burn-core/src/optim/grads.rs
```diff
@@ -81,6 +85,11 @@ impl GradientsParams {
         module.visit(&mut visitor);
         grads_params
     }
+
+    /// Provide an iterator over the gradient tensors.
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = (&ParamId, &mut TensorValue)> {
+        self.container.iter_mut()
+    }
```
nathanielsimard, line 92: Is this still necessary? I would like to avoid exposing the `Box<dyn Any>` since we may change it later.

burn-core/src/optim/sgd.rs
```diff
@@ -57,6 +61,7 @@ impl<B: Backend> SimpleOptimizer<B> for Sgd<B> {
         tensor: Tensor<B, D>,
         mut grad: Tensor<B, D>,
         state: Option<Self::State<D>>,
+        gradient_clip: Option<&GradientClipper>,
```
nathanielsimard, line 64: Same comment as adam

burn-core/src/optim/simple/adaptor.rs
```diff
@@ -88,11 +109,18 @@ where
...
+                grad
+            };
+
             let (tensor, state) = self.optimizer.step(
                 self.lr,
                 tensor.inner(),
-                grad,
+                clipped_grad,
                 record.map(|record| O::to_device(record.into_state(), &device)),
+                self.gradient_clipper,
```
nathanielsimard, line 123: If the adaptor is doing the clip value, I don't think we need to pass the clipper to the optimizer.

burn-core/src/optim/grads.rs
agelas: Nope I'm getting rid of all that stuff in grad.rs and container.rs/

burn-core/src/optim/simple/adaptor.rs
agelas: That's a good catch I'll adjust the adaptor and adam/sgd.

burn-core/src/grad_clipper/base.rs
nathanielsimard: @agelas

agelas: @nathanielsimard My bad, had to go back to using a  `struct` when I thought about making it a config to the optimizer. Should be fixed now.

agelas:
@nathanielsimard Should be good to go unless there's something else you'd like changed!

burn-core/src/grad_clipper/base.rs
```diff
@@ -0,0 +1,96 @@
...
+            let f_val: f32 = val.elem();
+            if f_val > threshold {
+                *val = threshold.elem();
+            } else if f_val < -threshold {
+                *val = (-threshold).elem();
+            }
+        });
+
+        Tensor::from_data_device(grad_data, &grad.device())
+    }
```
nathanielsimard, line 36: We should use tensor operations to do that instead of a for loop. So using `greater`, `lower` and `mask_fill`. This way we can leverage GPUs.

burn-core/src/grad_clipper/base.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    fn l2_norm<B: Backend, const D: usize>(tensor: &Tensor<B, D>) -> Tensor<B, 1> {
```
nathanielsimard, line 38: I would use a owned tensor here instead and call `clone()` in the `clip_by_norm` function.

burn-core/src/optim/sgd.rs
```diff
@@ -124,6 +126,13 @@ mod tests {
         assert!(record.is_empty());
     }
 
+    #[test]
+    fn can_attach_gradient_clipper() {
+        let optim = sgd_with_all().with_gradient_clipper(GradientClipper::ClipByValue(0.5));
+        let record = optim.to_record();
+        assert!(record.is_empty());
```
nathanielsimard, line 133: Does the test fail if you remove `with_gradient_clipper` ?

nathanielsimard (changes requested):
There are still some minor things to change, but it's indeed close to be merged.

burn-core/src/optim/sgd.rs
agelas: Now it does 😁

burn-core/src/grad_clipper/base.rs
agelas: Will creating masks from `greater_elem` and` lower_elem` do the same?

nathanielsimard:
There are some small things to be refactored, but it's easier if I do it, this is sufficiently good to be merged. I'll asked you @agelas for review with the small refactor to see what are your thoughts.

PR #306, opened 2023-04-22
Feat/group conv

nathanielsimard:
* Updated conv tests to use ranges instead of ones (more robust)
* Use conv and conv transpose options instead of function parameters. (Clippy warning too many arguments)


burn-tensor/src/tensor/ops/modules/conv.rs
```diff
@@ -76,36 +59,30 @@ pub(crate) fn conv1d_backward<B: Backend>(
...
-    }
+    let weight_grad = match options.groups == 1 {
+        true => conv1d_weight_grad_no_groups::<B>(x, output_grad.clone(), weight_shape, options),
+        false => conv1d_weight_grad_groups::<B>(
+            x,
+            B::zeros(weight_shape, &weight_device),
+            output_grad.clone(),
+            options,
+        ),
+    };
```
antimora, line 79: if else would have sufficed =D

antimora (approved):

antimora:
Good job. It is many changes just for a group. BTW, group is really useful to reduce number of ops and params with Depthwise Convolutions. So thank you.

burn-tensor/src/tensor/ops/modules/conv.rs
nathanielsimard: Yup, I just prefer the see the two functions called with no line between the two =D

PR #309, opened 2023-04-24
doc: add tensor + backend architecture doc

nathanielsimard:
Link: https://github.com/burn-rs/burn/blob/doc/tensor-architecture/ARCHITECTURE.md


antimora (approved):
Looks good. Please add a link from the root's Readme file.

PR #326, opened 2023-05-04
refactor: grad clipping

nathanielsimard:
@agelas I have made some changes to the gradient clipper:

* I have renamed it from `gradient clipper` to `grad clipping`.
* I have added a new configuration for gradient clipping, which is now included in both Adam and SGD configurations for seamless integration.
* I have also updated the code to use the Tensor API instead of the backend API in the `clip_by_value` function.


burn-core/src/optim/adam.rs
```diff
@@ -88,7 +93,12 @@ impl AdamConfig {
             },
             weight_decay: self.weight_decay.as_ref().map(WeightDecay::new),
         };
-        OptimizerAdaptor::from(optim)
+
+        let mut optim = OptimizerAdaptor::from(optim);
+        if let Some(config) = &self.grad_clipping {
```
agelas, line 98: @nathanielsimard Can you explain how config works? I tried playing with it previously but only got it to work with a weird struct setup. Is all that's required that the config have a default?

agelas (approved):

burn-core/src/optim/adam.rs
nathanielsimard: Not really, config can be enum (no default yet) or struct. Each member of the config needs to be (de)serializable, that's it. It's normaly a good idea to separate the config type from the actual type. In my case the type are pretty similar (two enums) but they may evolve differently and you don't have to handle a None case.

Issue #315, opened 2023-04-26
LogSoftmax and Softmax functions are numerically unstable


antimora:
**Describe the bug**
`log_softmax` and `softmax` functions produce `-inf` if the magnitudes are somewhat big.

**To Reproduce**
1. Input: `Input to log_softmax: Tensor { primitive: NdArrayTensor { array: [[-781.1664, -1043.7303, 890.2518, -1234.9999, -1622.5167, -2174.6855, -1304.8042, -1730.8314, -1072.6201, -1899.7662]], shape=[1, 10], strides=[10, 1], layout=CFcf (0xf), dynamic ndim=2 } }`
2. The output is: `Output from log_softmax: Tensor { primitive: NdArrayTensor { array: [[-inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf]], shape=[1, 10], strides=[10, 1], layout=CFcf (0xf), dynamic ndim=2 } }`

**Expected behavior**
The same input in PyTorch:

```python
>>> data = [[-781.1664, -1043.7303, 890.2518, -1234.9999, -1622.5167, -2174.6855, -1304.8042, -1730.8314, -1072.6201, -1899.7662]]
>>> x_data = torch.tensor(data)
>>> m = torch.nn.LogSoftmax(dim=1)
>>> m(x_data)
tensor([[-1671.4182, -1933.9822,     0.0000, -2125.2517, -2512.7686, -3064.9373,
         -2195.0559, -2621.0833, -1962.8718, -2790.0181]])
>>> 
```

**Additional context**
The softmax functions assume simple inputs and not the full range. Please see this blog post that shows a numerically stable solution: https://ogunlao.github.io/2020/04/26/you_dont_really_know_softmax.html

PR #346, opened 2023-05-11
perf: softmax num stability

nathanielsimard:


antimora (changes requested):
We need some tests. You can find values here: https://github.com/burn-rs/burn/issues/315

nathanielsimard:
> We need some tests. You can find values here: #315

Softmax and log softmax are tested in burn-tensor and burn-autodiff, if you remove the max it produces nan and the tests don't pass.
Cross entropy is also tested, so I'm confident in the code.

PR #347, opened 2023-05-11
Remove build panic from burn-tch

antimora:


burn-tch/src/backend.rs
```diff
@@ -9,7 +9,7 @@ use burn_tensor::backend::Backend;
 ///
 /// # Example
 ///
-/// ```rust
+/// ```no_run
```
nathanielsimard, line 12: What is no_run?

nathanielsimard (approved):

PR #348, opened 2023-05-11
Move dataset.py into its own file and add CustomEncoder

antimora:
1. Moved the python code out from rs file
2. added CustomEncoder to encode ndarray type


nathanielsimard (approved):
Way cleaner, I didn't know about `include_str!` when I first implemented this.

burn-dataset/src/source/huggingface/dataset.py
```diff
@@ -0,0 +1,152 @@
+import abc
+import json
+import argparse
+import numpy as np
+
+from datasets import load_dataset
```
antimora, line 6: What's this?

PR #351, opened 2023-05-12
feat: improve module recorder deserialization

nathanielsimard:
Added `ConstantRecord` so that deserialization can be made even when the `constant` names are not present in the file.


antimora (approved):
LGTM

PR #352, opened 2023-05-15
Improve softmax doc

nathanielsimard:


burn-tensor/src/tensor/activation/base.rs
```diff
@@ -12,17 +13,25 @@ pub fn gelu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
     Tensor::from_primitive(B::gelu(tensor.primitive))
 }
 
-/// Applies the softmax function.
+/// Applies the softmax function on the input tensor along the given dimension.
+///
+/// The `dim` argument must be smaller or equal than `D`, the number of dimensions the tensor has.
```
antimora, line 18: One of the asks from the https://github.com/burn-rs/burn/issues/350 is to explain `dim` param (explicitly probably to say it's a dimension). To me it's not clear if `dim` starts with zero or one. 

Here is how pytorch doc explains, maybe we could do something similar:

**dim** (int) – A dimension along which Softmax will be computed (so every slice along dim will sum to 1).

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -12,17 +13,25 @@ pub fn gelu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
...
     let tensor = tensor.exp();
     let tensor_tmp = tensor.clone().sum_dim(dim);
 
     tensor.div(tensor_tmp)
 }
 
-/// Applies the log softmax function.
+/// Applies the log softmax function on the input tensor along the given dimension.
+///
+/// The `dim` argument must be smaller or equal than `D`, the number of dimensions the tensor has.
```
antimora, line 31: The same comment as above.

antimora (changes requested):

Issue #334, opened 2023-05-05
ONNX: Streamline OP Node Conversion Implementation Process


antimora:
### Issue Description

Implement a trait for node conversion to streamline the process of implementing OP Node conversions.

### Motivation

At present, implementing an OP Node conversion requires making multiple changes across different parts of the codebase. A trait can simplify this process and improve code maintainability.

### Proposed Solution

Introduce the following trait for node conversion:

```rust
trait NodeConverter {
 
   /// Returns the field declaration type for the module.
   fn declaration_field_type() -> syn::Type;
   ...
}
```
This trait will make it easier to implement OP Node conversions by centralizing the required changes.

PR #355, opened 2023-05-19
Refactor: Burn Import use BurnGraph

### LINKED ISSUE #336: ONNX: Develop Intermediate Representation for Burn Code Generation
### Feature Description

Design and implement an intermediate representation (IR) specifically for Burn code generation in the ONNX framework.

### Motivation for Feature

Currently, the burn-import process converts ONNX protobufs to an intermediate representation (IR) closely resembling the ONNX data. This IR is subsequently converted to Burn source code. To streamline this process, it would be beneficial to introduce an additional IR layer, as follows: ONNX -> ONNX IR -> Burn IR -> Burn Source Code.

The data structure might look like this:

```rust
enum BurnNode {
    MatMul(MatMulArgs),
    ...
}

struct MatMulArgs {
    lhs: TensorNode,
    rhs: TensorNode,
}
```

nathanielsimard:
Use an internal representation for burn modules and tensor operations.

fix #336


burn-import/src/onnx/to_burn.rs
```diff
@@ -96,558 +95,213 @@ impl ModelGen {
...
-
-        }
+        graph
     }
 
-    /// Model implementation code
-    fn implementation(&self, imports: &mut HashSet<String>) -> TokenStream {
-        let forward_method = self.forward_method(imports);
+    fn matmul_conversion(node: Node) -> MatmulNode {
+        println!("Matmul Conversion");
```
antimora, line 159: The same comment about print statement.

burn-import/src/onnx/to_burn.rs
```diff
@@ -96,558 +95,213 @@ impl ModelGen {
...
-            NodeType::LogSoftmax => {
-                imports.insert("burn::tensor::activation::log_softmax".to_string());
-                let dim = log_softmax_config(node);
-
-                quote! { log_softmax(#input1, #dim) }
-            }
-            NodeType::Flatten => {
-                let (start_dim, end_dim) = flatten_config(node);
+    fn linear_conversion<PS: PrecisionSettings>(mut node: Node) -> LinearNode<PS> {
+        println!("Linear Conversion");
```
antimora, line 191: The same comment about print statement.

burn-import/src/onnx/to_burn.rs
```diff
@@ -96,558 +95,213 @@ impl ModelGen {
...
-                 )*)
-            }
-        }
-    }
-
-    /// Generates source code for the initialization method
-    fn new_method(&self) -> TokenStream {
-        let initialization_fields = self.initialization_fields();
+    fn conv2d_conversion<PS: PrecisionSettings>(mut node: Node) -> Conv2dNode<PS> {
+        println!("Conv2d conversion");
```
antimora, line 250: The same comment about print statement.

burn-import/tests/data/model1/model1.rs
```diff
@@ -1,67 +1,59 @@
...
+use burn::nn::LinearConfig;
 use burn::{
     module::Module,
-    nn,
-    record::{DefaultRecorder, Recorder},
     tensor::{backend::Backend, Tensor},
 };
 
-pub const INPUT1_SHAPE: [usize; 4usize] = [1usize, 1usize, 8usize, 8usize];
-pub const OUTPUT1_SHAPE: [usize; 2usize] = [1usize, 288usize];
```
antimora, line 14: Do we still want to provide some facility to know the input shape? The user may have to guess or know in advance but it's part of the model.

antimora (approved):
Looks great! As I was reviewing I became excited about your simplification. I'd be so much easier now to add more nodes. I guess your work tackles both #334 and #336.

I approved it but please see my comments with questions.

antimora (approved):
Looks great! As I was reviewing I became excited about your simplification. I'd be so much easier now to add more nodes. I guess your work tackles both #334 and #336.

I approved it but please see my comments with questions.

burn-import/tests/data/model1/model1.rs
nathanielsimard: The input and output shapes may vary, if you want to use a batch size of 1 or 32 it shouldn't matter, so I prefer not exporting it, making its usage similar to other burn modules. The best solution would be to export comments explaining each dimension (like batch_size, seq_length, channel, width, etc.) but I'm not sure this information is present in ONNX.

antimora: We could just document it in the source somewhere. I am not sure really. Or maybe we could provide some tooling that summarizes the model like this: http://www.xavierdupre.fr/app/onnxruntime/helpsphinx/auto_examples/plot_load_and_predict.html

We could think of a solution and do it later. I am okay with this.

antimora (approved):

PR #364, opened 2023-05-26
Add SqliteDatasetWriter to store items in sqlite db

antimora:


burn-dataset/src/dataset/sqlite.rs
```diff
@@ -57,16 +63,48 @@ pub struct SqliteDataset<I> {
...
+    /// The chosen table must share the same name as `split`, and must contain a primary key column named `row_id`.
+    ///
+    /// # Arguments
+    ///
+    /// * `db_file`: &str - The path to the SQLite database file.
+    ///
+    /// * `split`: &str - The name of the table in the database that the dataset is supposed to represent.
+    ///
+    /// * `row_serialized`: bool - Indicates the structure of the table. If `true`, each row in the table is expected
+    /// to have its fields MessagePack serialized into a single `item` column. If `false`, each field of the item
```
nathanielsimard, line 83: ... each row in the table is expected to be  __serialized with the MessagePack format__ in a single `item` column.

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -106,18 +145,28 @@ where
...
-            })
-            .unwrap();
-
-        // Return the first row if found else None (error)
-        rows.next().and_then(|res| match res {
-            Ok(val) => Some(val),
-            Err(_) => None,
-        })
+        if self.row_serialized {
+            // Fetch a with a single column `item` and deserialize it with MessagePack
```
nathanielsimard, line 149: Fetch a with a => Fetch with a

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -151,19 +200,190 @@ fn fetch_columns_and_len(
...
+
+    /// Return the SqliteDataset for the given split.
+    pub fn dataset(&self, split: &str) -> SqliteDataset<I> {
+        if !self.exists() {
+            panic!("Dataset does not exist: {}", self.name);
+        }
+
+        SqliteDataset::from_db_file(self.db_file().as_str(), split, true)
+    }
+
```
nathanielsimard, line 265: What is the reason for the saver struct to expose the dataset function? I think the saver should only expose save and the options (with_splits, with_base_dif, etc.), maybe the name is wrong?

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -218,4 +439,217 @@ mod tests {
...
+            .with_splits(&["train", "test"])
+            .init();
+
+        // Again create a file in the temp dir and it should fail
+        let _ds_saver2 = SqlDsSaver::new("preprocessed")
+            .with_base_dir(temp_dir_str.as_str())
+            .with_splits(&["train", "test"])
+            .with_overwrite(true)
+            .init();
+    }
```
nathanielsimard, line 551: It seems like testing privates 🤔. Those functions seem like utilities more than a public APIs that are going to survive through time.

nathanielsimard (changes requested):
I think there is still some work to be done to cleanup the APIs, but I like where this is going.

burn-dataset/src/dataset/sqlite.rs
antimora: After a sqlite db is created, we would to get a dataset access. It's a convenience instead of passing a db location and getting a dataset. 

I can rename the method to `get_dataset`, maybe?

burn-dataset/src/dataset/sqlite.rs
antimora: The only correct way to verify if something is initialized is to check it's state. We are checking the private fields but this is unit testing so we are allowed in this particular case. These fields are used in subsequent calls and it's important they're set correctly. Alternatively I could make these fields public or create public methods, also no harm but I didn't see use cases.


The line 545 should say:         // Again create a file in the temp dir and it should NOT fail

burn-dataset/src/dataset/sqlite.rs
antimora: Per our offline conversation. This method should go away because we do not wish to get a reader while the data is being written.

antimora:
OK. I have finally finished the big refactor. I have addressed all the things we talked about and more. 

Here is the top summary:

1. The methods now return result
2. Errors are mapped to local error struct
3. Separate into Write and Storage
4. Dataset reader now automatically figures out how data is stored.
5. Improved thread safety.
6. Use PathBuf/Path instead of strings (also makes more portable by not using "/" for directory designation). 
7. Writes to a temporary file and renames to db file after successful write. 
8. Disables journaling and sync for fast writes (at least 3X times faster writes). It's okay for DB is a new and can be recreated if computer crashes.

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -131,39 +228,380 @@ fn fetch_columns_and_len(
...
+            .ok_or("Unable to get parent directory")?;
+
+        if !db_file_dir.exists() {
+            fs::create_dir_all(db_file_dir)?;
+        }
+
+        // Create a temporary database file (will be persisted if writes complete or deleted it otherwise)
+        self.db_file_tmp = Some(NamedTempFile::new_in(db_file_dir)?);
+
+        // Create a connection pool
```
nathanielsimard, line 467: Wrong comment?

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -131,39 +228,380 @@ fn fetch_columns_and_len(
...
+    ///
+    /// * A `Result` which is `Ok` if the writer could be created, `Err` otherwise.
+    pub fn new<P: AsRef<Path>>(db_file: P, overwrite: bool) -> Result<Self> {
+        let writer = Self {
+            db_file: db_file.as_ref().to_path_buf(),
+            db_file_tmp: None,
+            splits: Arc::new(RwLock::new(HashSet::new())),
+            overwrite,
+            conn_pool: None,
+            is_completed: Arc::new(RwLock::new(false)),
```
nathanielsimard, line 432: Could be an atomic bool instead.

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -218,4 +657,172 @@ mod tests {
...
+    }
+
+    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+    pub struct Complex {
+        column_str: String,
+        column_bytes: Vec<u8>,
+        column_int: i64,
+        column_bool: bool,
+        column_float: f64,
+        colomn_complex: Vec<Vec<Vec<[u8; 3]>>>,
```
nathanielsimard, line 694: 😆

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -218,4 +657,172 @@ mod tests {
...
+            column_bytes: vec![1_u8, 2, 3],
+            column_int: 0,
+            column_bool: true,
+            column_float: 1.0,
+            colomn_complex: vec![vec![vec![[1, 23_u8, 3]]]],
+        };
+
+        let result = writer.write("train", &new_item);
+
+        assert!(result.is_ok());
```
nathanielsimard, line 761: I think you can remove those asserts in general, the `unwrap` will make the test crash. If we want more clarify, we could use expect with error messages.

nathanielsimard (changes requested):
Very nice refactoring, there are some very minor changes, but it looks pretty good!

burn-dataset/src/dataset/sqlite.rs
```diff
@@ -106,18 +193,28 @@ where
...
+                .unwrap()
+        } else {
+            // Fetch a row with multiple columns and deserialize it serde_rusqlite
+            statement
+                .query_row([row_id], |row| {
+                    // Deserialize the row with serde_rusqlite
+                    Ok(from_row_with_columns::<I>(row, &self.columns).unwrap())
+                })
+                .optional() //Converts Error (not found) to None
+                .unwrap()
```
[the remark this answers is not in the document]
antimora: In this case, we are returning `Option<I>`. By default, query will return an error if not found, but calling .optional() will return None instead, which we want. `unwrap()` unwraps Result<Option<I>, Error>. This means if there is an actual error (other than not-found), there will be a panic, which is only way with this API.

burn-dataset/src/dataset/sqlite.rs
antimora: Yes, I have looked into using AtomicBool. Though AtomicBool is safe to use for (read bool and immediate update bool), it's non-blocking. I need to block writer's write if the "complete" procedure is in the process of IO changes. I only update is_completed to true at the end if all IO finishes successfully. Setting is_completed at the beginning will not be technically correct. As far as I understood, AtomicBool is mainly for race conditions at the instruction level pipeline, so that the read and updating is in one go. In our case, we are doing IO.

PR #367, opened 2023-05-27
Feat/gather scatter

nathanielsimard:
Make indexing functions more similar to other frameworks.


burn-tensor/src/tensor/api/check.rs
```diff
@@ -374,6 +374,98 @@ impl TensorCheck {
         check
     }
 
+    pub fn gather<const D: usize>(dim: usize, shape: &Shape<D>, shape_indexes: &Shape<D>) -> Self {
```
antimora, line 377: I think we should start putting some minimal doc strings for all public APIs. Good pattern to establish.

burn-tensor/src/tensor/api/check.rs
```diff
@@ -374,6 +374,98 @@ impl TensorCheck {
...
+        Self::check_gather_scatter_indexes(Self::Ok, "Gather", dim, shape, shape_indexes)
+    }
+
+    pub fn scatter<const D: usize>(
```
antimora, line 381: same

burn-tensor/src/tensor/api/check.rs
```diff
@@ -374,6 +374,98 @@ impl TensorCheck {
...
+                    "The shape differs: {:?} != {:?}",
+                    shape_indexes.dims, shape_value.dims
+                )),
+            );
+        }
+
+        check
+    }
+
+    pub fn index_select<const D: usize>(dim: usize) -> Self {
```
antimora, line 408: same as above

burn-tensor/src/tensor/api/check.rs
```diff
@@ -374,6 +374,98 @@ impl TensorCheck {
...
+        Self::check_index_select_basic::<D>(Self::Ok, "index_select", dim)
+    }
+
+    pub fn index_select_assign<const D: usize>(dim: usize) -> Self {
```
antimora, line 412: Doc

antimora (changes requested):
Looks good. Requesting to add missing doc strings for the public API.

burn-tensor/src/tensor/api/check.rs
nathanielsimard: I don't think this should be a public API, just pub(crate) I'll do the fix

antimora (approved):

PR #368, opened 2023-05-27
refactor: onnx parsing

nathanielsimard:
1. Remove module state as it is implemented by the Burn IR now.
2. Simplify shape inference by removing duplicated logic.


nathanielsimard:
@antimora Also we don't need the full shape, just the dimentionality (D) for code generation. It might really speed up compilation if we don't need to run through the model for that information.

antimora (approved):
Looks cleaner. Thank you for the refactor.

antimora:
> @antimora Also we don't need the full shape, just the dimentionality (D) for code generation. It might really speed up compilation if we don't need to run through the model for that information.

Sounds good. I guess the input dimensions for Linear could be figured out from the weights, thus not needed.

Issue #313, opened 2023-04-25
LSTM: Long Short-Term Memory


nathanielsimard:
### Feature description

LSMT Module

### Feature motivation

Still useful in some applications.

PR #370, opened 2023-05-30
Feat/lstm

agelas:
@nathanielsimard I don't know if this is anything near what you were imagining for #313, but I was playing around with a potential implementation for LSTM cells the past week and figured it wouldn't hurt to throw it into a pr. This lstm implementation is stateful (so a cell can remember its state for future forward passes if needed) and unidirectional. 

The initializer default should probably be Xavier/Glorot, but it doesn't seem like that's available yet.

Also note that this doesn't support batched inputs, so if that's something you want that'll take a bit of reworking I think.


antimora:
Thank you! This feature will be a great addition.

I have created a feature request ticket for Xavier: https://github.com/burn-rs/burn/issues/371

Also, the PR is not passing clippy build. When you've got a chance, could you please run the fix?

agelas:
> Also, the PR is not passing clippy build. When you've got a chance, could you please run the fix?

@antimora looks like the dead_code warning was being a little over-zealous, should be fixed now with some updated module scoping.

nathanielsimard:
@agelas Thansk for the PR. I think it's best for network modules to be stateless, the only exception at the moment would be the batch norm which collects batch statistique during the forward pass. 

So the API could look like this:

```rust
let (output, state) = lstm.forward(input, state);
```

Like you mentionned, I would prefer having the batch size included in each tensor.

```
input: (batch_size, seq_length, d_in)
output: (batch_size, seq_length, d_out)
```

As a last note, I think we should avoid having getters in the code (`get_weights`). I think you can directly use `Param<Tensor>` for most parameters and not rely on the `Linear` module. Xavier is actually used by Linear,

agelas:
@nathanielsimard @antimora Thanks for the feedback, I'll start making the fixes.

agelas:
@nathanielsimard The forward function has now been updated to work with an input tensor of shape `(batch_size, seq_length, d_in)`. I also added a `squeeze()` function to the tensor api- it didn't look there was an easy way to do something like [this in PyTorch](https://pytorch.org/docs/stable/generated/torch.squeeze.html), but please correct me if I'm wrong. 

> As a last note, I think we should avoid having getters in the code (get_weights). I think you can directly use Param<Tensor> for most parameters and not rely on the Linear module.

How would you recommend I do this? I made the gate_controllers two linear layers, and I need access to the weights and biases of those linear layers. Just because `weight` and `bias` are `Param`'s doesn't make those private members visible right?

nathanielsimard:
@agelas  The squeeze is a nice addition, there wasn't a simpler way to do this before (reshape). 

Instead of making getters, you can make some fields `pub(crate)`, but note that you can also just store `Tensor` in `Param` and call `val` to have access to the tensor. I think both ways are fine.

As a last note, is it possible to return each cell and hidden state for each seq_length (maybe using `Tensor::cat`), otherwise it may be hard to stack multiple LSMT layers on top of each other.

I'll do a proper code review soon when those changes are made, thanks a lot for contributing :)

burn-core/src/nn/lstm/gate_controller.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn get_input_bias(&self) -> Option<Tensor<B, 1>> {
+        self.input_transform.get_bias()
+    }
+
+    pub fn get_hidden_bias(&self) -> Option<Tensor<B, 1>> {
+        self.hidden_transform.get_bias()
+    }
+
+    /// Only used for testing in lstm
+    pub fn create_with_weights(
```
antimora, line 73: If it's only used by tests, you can move it under the mod tests {..} as a helper. If you decide to keep it here, you can make it private, which should still be accessible by tests. Or you can put it behind `#[cfg(test)]`. 

Also I would change the doc comment to describe what this function does and you can mention it's used by testing.

burn-core/src/nn/lstm/lstm.rs
```diff
@@ -0,0 +1,295 @@
...
+
+#[derive(Config)]
+pub struct LSTMConfig {
+    /// The size of the input features.
+    pub d_input: usize,
+    /// The size of the hidden state.
+    pub d_hidden: usize,
+    /// If a bias should be applied during the Lstm transformation.
+    pub bias: bool,
+    /// Lstm initializer, should probably be Xavier or small random numbers
```
antimora, line 23: Please add a TODO and link this issue: https://github.com/burn-rs/burn/issues/371 so we can come back and update it. Also please update the issue ticket to mention about addition work to update NN lstm.

burn-core/src/nn/lstm/lstm.rs
```diff
@@ -0,0 +1,295 @@
...
+        Lstm {
+            input_gate,
+            forget_gate,
+            output_gate,
+            cell_gate,
+            batch_size: self.batch_size,
+            d_hidden: self.d_hidden,
+        }
+    }
+
```
antimora, line 80: Missing doc comment.

burn-tensor/src/tensor/api/base.rs
```diff
@@ -110,6 +110,50 @@ where
...
+    /// ```rust
+    ///
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::{Tensor, Shape};
+    ///
+    /// fn example<B: Backend>() {
+    ///     let tensor = Tensor::<B, 3>::ones(Shape::new([2, 1, 4]));
+    ///
+    ///     // Given a 3D tensor with dimensions (2, 1, 4), squeeze the dimension 1
+    ///     let squeezed_tensor: Tensor::<B, 2> = tensor.squeeze::<2>(1);
```
antimora, line 139: In this example, I think it's ok to leave out `<2>` from the method call. This should work too:

```rust
let squeezed_tensor: Tensor::<B, 2> = tensor.squeeze(1);
```

Or this

```rust
let squeezed_tensor = tensor.squeeze::<2>(1);
```

I prefer the first version.

burn-tensor/src/tensor/api/base.rs
```diff
@@ -110,6 +110,50 @@ where
...
+    ///
+    ///     // Resulting tensor will have dimensions (2, 4)
+    ///     println!("{:?}", squeezed_tensor.shape());
+    /// }
+    /// ```
+    pub fn squeeze<const D2: usize>(self, dim: usize) -> Tensor<B, D2, K> {
```
antimora, line 145: Missing unit tests. Please see how they are added in this PR for `flatten` functionality. It was not straightforward for me, so I am listing the steps here.

https://github.com/burn-rs/burn/pull/260/files

1. add test under `burn-tensor/src/tests/ops`
2. add module under `burn-tensor/src/tests/ops/mod.rs`
3. modify burn-tensor/src/tests/mod.rs and add macro call

antimora (changes requested):
This is a great addition indeed! The set of ops are getting rich. 

I did a first pass of review. Relative minor and straightforward to fix. Let me know if you need more clarification from me.

antimora (changes requested):
This is a great addition indeed! The set of ops are getting rich. 

I did a first pass of review. Relative minor and straightforward to fix. Let me know if you need more clarification from me.

burn-tensor/src/tensor/api/base.rs
agelas: @antimora Thanks, this was super helpful! As a side note it seems some ops are missing unit tests (like unsqueeze). Just though I'd bring it up.

agelas:
> @agelas The squeeze is a nice addition, there wasn't a simpler way to do this before (reshape).
> 
> Instead of making getters, you can make some fields `pub(crate)`, but note that you can also just store `Tensor` in `Param` and call `val` to have access to the tensor. I think both ways are fine.
> 
> As a last note, is it possible to return each cell and hidden state for each seq_length (maybe using `Tensor::cat`), otherwise it may be hard to stack multiple LSMT layers on top of each other.
> 
> I'll do a proper code review soon when those changes are made, thanks a lot for contributing :)

@nathanielsimard I'm fixing up @antimora 's comments, but with respect to your second comment, I don't think there should be too much difficulty stacking multiple LSTM layers. The input accepts tensors of `[batch_size, seq_length, hidden_dim]`, and both the `batched_cell_state` and `batched_hidden_state` have that shape as well. So in the training loop you should be able to just grab the `batched_hidden_state` and pass it along.

nathanielsimard:
Looking at the comment : 

```In this implementation, the Lstm returns only the last time step's output```

I though that the outout size would be `[batch_size, 1, d_model]`, but looking at the code.

```rust
// store the state for this timestep
batched_cell_state = batched_cell_state.index_assign(
    [0..self.batch_size, t..(t + 1), 0..self.d_hidden],
    cell_state.clone().unsqueeze(),
);
batched_hidden_state = batched_hidden_state.index_assign(
    [0..self.batch_size, t..(t + 1), 0..self.d_hidden],
    hidden_state.clone().unsqueeze(),
);
```

It seems like you actually return the cell and hidden state for each timestep. So in the end, I just think you should update the comment 😅.

burn-tensor/src/tensor/api/base.rs
antimora: Thanks, @agelas. I have filed a task ticket to add the missing unit tests: https://github.com/burn-rs/burn/issues/384

agelas:
> It seems like you actually return the cell and hidden state for each timestep. So in the end, I just think you should update the comment 😅.

@nathanielsimard Ah yes sorry I'll update the comment sorry for the confusion.

burn-core/src/nn/linear.rs
```diff
@@ -90,6 +90,14 @@ impl<B: Backend> Linear<B> {
...
         }
     }
+
+    pub fn get_weight(&self) -> Tensor<B, 2> {
+        self.weight.val()
+    }
+
+    pub fn get_bias(&self) -> Option<Tensor<B, 1>> {
+        self.bias.as_ref().map(|bias| bias.val())
+    }
```
nathanielsimard, line 100: I would remove the getters before merging.

burn-core/src/nn/lstm/gate_controller.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.hidden_transform.get_weight()
+    }
+
+    pub fn get_input_bias(&self) -> Option<Tensor<B, 1>> {
+        self.input_transform.get_bias()
+    }
+
+    pub fn get_hidden_bias(&self) -> Option<Tensor<B, 1>> {
+        self.hidden_transform.get_bias()
+    }
```
nathanielsimard, line 69: Same

burn-core/src/nn/lstm/lstm.rs
```diff
@@ -0,0 +1,303 @@
...
+    fn gate_product(
+        &self,
+        input: &Tensor<B, 2>,
+        hidden: &Tensor<B, 2>,
+        gate: &GateController<B>,
+    ) -> Tensor<B, 2> {
+        let input_product = input.clone().matmul(gate.get_input_weight());
+        let hidden_product = hidden.clone().matmul(gate.get_hidden_weight());
+
+        match (gate.get_input_bias(), gate.get_hidden_bias()) {
```
nathanielsimard, line 198: Using `pub(crate)` to access `input_transforms`(GateController) and `bias`(Linear)

```rust
let input_bias = gate.input_transforms.bias.map(Param::val);
```

nathanielsimard (changes requested):
Some changes to use `pub(crate)` instead of getters, but other than that, I think it's pretty good :)

burn-core/src/nn/lstm/gate_controller.rs
agelas: Getters all removed!

burn-core/src/nn/lstm/lstm.rs
agelas: @nathanielsimard I understand the case against getters, and I'm not going to push the issue, but I think that one could argue for readability/understandability here. To make the above work, I needed to do this:

```rust
let input_bias = gate
            .input_transform
            .bias
            .as_ref()
            .map(|bias_param| bias_param.val());
```
Direct field access is certainly more concise and cuts down on verboseness, but it's also less clear what is happening unless you are familiar with the structure of the `GateController` and `Linear` structs, and how they work with `Param`. Just food for thought.

agelas:
@nathanielsimard @antimora I think I got all the requested changes. When someone gets the chance, can they rerun the CI testing? It looks like there was an issue installing rust. Thanks!

burn-core/src/nn/lstm/lstm.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// Returns:
+    ///     A tuple of tensors, where the first tensor represents the cell states and
+    ///     the second tensor represents the hidden states for each sequence element.
+    ///     Both output tensors have the shape [batch_size, seq_length, hidden_size].
+    pub fn forward(
+        &mut self,
+        batched_input: Tensor<B, 3>,
+        state: Option<(Tensor<B, 2>, Tensor<B, 2>)>,
+    ) -> (Tensor<B, 3>, Tensor<B, 3>) {
+        let seq_length = batched_input.shape().dims.to_vec()[1];
```
antimora, line 132: Just noticed the dims  (array) are converted to vec(). Probably more efficient without `to_vec()`.

burn-core/src/nn/lstm/lstm.rs
```diff
@@ -0,0 +1,322 @@
...
+pub struct Lstm<B: Backend> {
+    input_gate: GateController<B>,
+    forget_gate: GateController<B>,
+    output_gate: GateController<B>,
+    cell_gate: GateController<B>,
+    batch_size: usize,
+    d_hidden: usize,
+}
+
+impl LSTMConfig {
```
antimora, line 43: Can you we also make the name the Camel, like LstmConfig? Someone pointed out naming is not up to convention: https://rust-lang.github.io/api-guidelines/naming.html

burn-core/src/nn/lstm/lstm.rs
```diff
@@ -0,0 +1,322 @@
...
+            .assert_in_range(0.0, 1.0);
+        lstm.cell_gate
+            .input_transform
+            .weight
+            .val()
+            .to_data()
+            .assert_in_range(0.0, 1.0);
+    }
+
+    #[test]
```
antimora, line 260: `#[test]` should be right above the function name and below the document.

burn-core/src/nn/lstm/gate_controller.rs
```diff
@@ -0,0 +1,85 @@
...
+        let l1 = LinearConfig::init_with(linear_config, record.input_transform);
+        let l2 = LinearConfig::init_with(linear_config, record.hidden_transform);
+
+        Self {
+            input_transform: l1,
+            hidden_transform: l2,
+        }
+    }
+
+    #[cfg(test)]
```
antimora, line 54: `#[cfg(test)]` should be right above the function name.

burn-core/src/nn/lstm/gate_controller.rs
```diff
@@ -0,0 +1,85 @@
...
+#[derive(Module, Debug)]
+pub struct GateController<B: Backend> {
+    /// Represents the affine transformation applied to input vector
+    pub(crate) input_transform: Linear<B>,
+    /// Represents the affine transformation applied to the hidden state
+    pub(crate) hidden_transform: Linear<B>,
+}
+
+impl<B: Backend> GateController<B> {
+    pub fn new(d_input: usize, d_output: usize, bias: bool, initializer: Initializer) -> Self {
```
antimora, line 25: Missing the docs.

burn-core/src/nn/lstm/gate_controller.rs
```diff
@@ -0,0 +1,85 @@
...
+            hidden_transform: LinearConfig {
+                d_input: d_output,
+                d_output,
+                bias,
+                initializer,
+            }
+            .init(),
+        }
+    }
+
```
antimora, line 43: Missing the docs.

antimora (changes requested):
Only minor cleanup, and it's my fault not catching right away. I apologize for asking you to make more changes.

Other than that I think it's ready to be merge. 

You can wait what @nathanielsimard has to say, so you don't have to go back and worth.

This PR is a great addition as I mentioned already, and your coding has improved greatly. Looking forward to seeing more submission.

antimora (approved):
Good to go from my end!

nathanielsimard (approved):

nathanielsimard:
@agelas Thanks a lot for the contribution, it's now merged 🎉

PR #376, opened 2023-06-01
Feat/wgpu backend setup

nathanielsimard:
Setup a backend using wgpu(https://wgpu.rs/)

The basic architecture is there with a few shaders implemented.


burn-wgpu/src/context.rs
```diff
@@ -0,0 +1,252 @@
...
+        let device_wgpu = device.clone();
+        let (device, queue) = pollster::block_on(adapter.request_device(
+            &DeviceDescriptor {
+                label: None,
+                features: wgpu::Features::empty(),
+                limits: wgpu::Limits::downlevel_defaults(),
+            },
+            None,
+        ))
+        .unwrap();
```
antimora, line 82: Probably we should make it clearer with expect message. I am not sure what would failing here.

burn-wgpu/src/kernel/base.rs
```diff
@@ -0,0 +1,87 @@
...
+    type Source = String;
+
+    fn generate() -> String {
+        let mut source = K::generate().as_ref().to_string();
+
+        source = source.replace("WORKGROUP_SIZE_X", &WORKGROUP_X_SIZE.to_string());
+        source = source.replace("WORKGROUP_SIZE_Y", &WORKGROUP_Y_SIZE.to_string());
+        source = source.replace("WORKGROUP_SIZE_Z", &WORKGROUP_Y_SIZE.to_string());
+        source = source.replace("elem", E::type_name());
+        source = source.replace("int", I::type_name());
```
antimora, line 65: I would recommend this super simple template library (text-placeholder): https://github.com/bernardoamc/text-placeholder. I have used it when I wanted to pre-generate constants from a template. It has an ability to parse template, which could be stored in advance. The placeholder replacement is linear afterwards.

No need to fix it now but to think about, which you probably have done so. WONNX also uses a template engine but it's heavy and geared toward HTML pages.

burn-wgpu/src/kernel/binary_elemwise.rs
```diff
@@ -0,0 +1,159 @@
...
+    let shape_out = Shape::new(shape_out);
+
+    let buffer = lhs
+        .context
+        .create_buffer(shape_out.num_elements() * core::mem::size_of::<E>());
+    let output = WGPUTensor::new(lhs.context.clone(), shape_out, Arc::new(buffer));
+
+    let kernel = lhs
+        .context
+        .compile::<KernelSettings<K, E, i32, 256, 1, 1>>();
```
antimora, line 84: Probably we want to document this 256 magic number into a constant. Not sure if this will need to change again.

burn-wgpu/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    burn_tensor::testgen_add!();
+    burn_tensor::testgen_sub!();
+    burn_tensor::testgen_div!();
+    burn_tensor::testgen_mul!();
+    burn_tensor::testgen_neg!();
+    burn_tensor::testgen_powf!();
+    burn_tensor::testgen_exp!();
+    burn_tensor::testgen_log!();
+    burn_tensor::testgen_relu!();
```
antimora, line 33: So cool to see a lot of ops are implemented.

burn-wgpu/src/template/binary_elemwise.wgsl
```diff
@@ -0,0 +1,35 @@
...
+        let stride_lhs = info[i + 1u];
+        let stride_rhs = info[i + 1u * dim + 1u];
+        let shape_lhs = info[i + 2u * dim + 1u];
+        let shape_rhs = info[i + 3u * dim + 1u];
+
+        index_lhs += global_id.x / stride_lhs % shape_lhs * stride_lhs;
+        index_rhs += global_id.x / stride_rhs % shape_rhs * stride_rhs;
+    }
+
+    BODY
```
antimora, line 34: Another small reason to use [text-placeholder](https://github.com/bernardoamc/text-placeholder), it's easier to see the placeholders - it may not be obvious WORKGROUP_SIZE_X is a placeholder.

antimora (changes requested):
Looks great! Can't wait to try out. 

Requesting one small change to add expect message. Another is my recommendation to use [text-placeholder](https://github.com/bernardoamc/text-placeholder) lib for templating

burn-wgpu/src/template/binary_elemwise.wgsl
nathanielsimard: This is a feature and a bug xD. I like it when linting and color highlighting work with .wgsl files, even if it makes it harder to  know what things will be changed. The trait `KernelGenerator` only has static methods, so the source code is generated and compiled just once for each kernel.

burn-wgpu/src/kernel/binary_elemwise.rs
nathanielsimard: This is the workgroup size. For now it is hardcoded, but I would like execute multiple kernels with different parameters to see which one performs the best (at runtime). This depends on the tensor shape and the actual GPU.

antimora (approved):

PR #378, opened 2023-06-02
Feat: mse loss

bioinformatist:
Hi my bro, the new mse loss was implemented with 3 reduction modes:

1. For the common `forward` (like `CrossEntropyLoss`) and the `forward_sum` method, the output is in shape `Tensor<B, 1>`, as the `forward` stands for applying mean reduction, while the `forward_sum` means the output values will be summed
2. For the `forward_none` method, the shape is `Tensor<B, 2>`

And to validate in Pytorch, by `docker run --rm -ti --ipc=host pytorch/pytorch:latest python`:

```python
import torch
t1 = torch.tensor([[0.0, 1.0], [3.0, 4.0]], dtype = torch.float32, requires_grad = True)
t2 = torch.tensor([[6.0, 7.0], [9.0, 10.0]], dtype = torch.float32, requires_grad = True)
t3 = torch.matmul(t1, t2)

t_targets = torch.tensor([[0.8, 0.2], [0.9, 0.1]], dtype = torch.float32, requires_grad = True)

# With default parameter (no reduction)
loss = torch.nn.MSELoss()
t4 = loss(t3, t_targets)
sum_res = t4.sum()
sum_res.backward()
print(t1.grad)
print(t2.grad)
```

And the result is (consistents with our `test_cross_entropy_loss_grad()`):

image


burn-core/src/nn/loss/mse.rs
```diff
@@ -0,0 +1,81 @@
...
+            backend: PhantomData::default(),
+        }
+    }
+
+    /// Compute the criterion on the input tensor.
+    ///
+    /// # Shapes
+    ///
+    /// - logits: [batch_size, num_targets]
+    /// - targets: [batch_size, num_targets]
```
nathanielsimard, line 30: I think it would be better to have the reduction handled by an enum here, so there would be only one forward function.

```rust
pub fn forward(&self, logits: Tensor<B, 2>, targets: Tensor<B, 2>, reduction: Reduction) -> Tensor<B, 1> {
    match reduction {
        Reduction::Meam => ...
        Reduction::Sum => ...
    }
}

pub enum Reduction {
   Mean,
   Sum,
}
```

Since the return type is different when no reduction is applied, I would keep the `forward_none`, but rename it to `forward_no_reduction`.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #379, opened 2023-06-03
feat: add basic matmul for wgpu backend

nathanielsimard:


burn-wgpu/src/lib.rs
```diff
@@ -20,7 +20,11 @@ pub use graphics::*;
 
 #[cfg(test)]
 mod tests {
-    type TestBackend = crate::WGPUBackend<crate::Vulkan, f32, i64>;
+    use super::*;
+
+    type TestBackend = WGPUBackend<Vulkan, f32, i64>;
```
antimora, line 25: Would it possible to use CPU or virtual device to test the WGPU backened? It'd be great if we could run the tests locally or on CI.

antimora (approved):
Looks like we are getting close to run a transformer on WGPU with this addition =D

Overall PR looks great. 

I had a question regarding the test backend. Please see my comment below.

burn-wgpu/src/lib.rs
nathanielsimard: It is running with the CPU device, but using the Vulkan Graphics API :)

PR #381, opened 2023-06-04
Feat/wgpu/swap dims

nathanielsimard:
* Add swap dims
* Add reshape
* Support autodiff for all operations except relu (masking not implemented yet)

### DIFF
--- a/burn-wgpu/src/ops/base.rs
+++ b/burn-wgpu/src/ops/base.rs
@@ -1,10 +1,10 @@
 use crate::{
     context::WorkGroup,
     element::WGPUElement,
-    kernel::{build_binary_info, KernelSettings},
+    kernel::{build_info, KernelSettings},
     kernel_wgsl,
     pool::get_context,
-    tensor::WGPUTensor,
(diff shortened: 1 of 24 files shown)


burn-wgpu/src/ops/base.rs
```diff
@@ -1,8 +1,14 @@
-use std::{marker::PhantomData, sync::Arc};
-
+use crate::{
+    context::WorkGroup,
+    element::WGPUElement,
+    kernel::{build_info, KernelSettings},
+    kernel_wgsl,
+    pool::get_context,
+    tensor::WgpuTensor,
+    GraphicsAPI, WGPUDevice,
```
antimora, line 8: Do you want to also change these to CamelCase so that we are consistent across the board?

Here is the list: 
 - WGPUElement -> WgpuElement
 - GraphicsAPI -> GraphicsApi
 - WGPUDevice -> WgpuDevice

antimora (approved):
Looks good to me. There is one change request for the naming consistency. 

You may merge after the change.

PR #385, opened 2023-06-05
build(Dockerfile): add Dockerfile for burn

bioinformatist:
According to https://github.com/burn-rs/burn/issues/175, it's better to have a Dockerfile for our project.

Below is a tutorial, maybe we could merge it into `README.md` or place it in a separate file.

## How to use

To build the image:

```bash
docker build --network host -t burn .
```

To run a container:

```bash
docker run --name burn -it burn sh
```

For developers who use VS Code, one can click on the button in the left corner of the status bar, then choose **Attach to Running Container...** to attach a new window to container `burn` which was just created.

image

Now you can develop as usual. To make things easier, you may need [rust-analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer) extension.

You can start a new project with `cargo new`, and specify Burn as dependency with absolute/relative path in `Cargo.toml`, just like:

```toml
burn = {  path = "/workspace/burn/burn" }
```

You can also mount your local file system to the container, and below is an example:

```bash
docker run --name burn -v $PWD:/workspace -it burn sh
```

## Known issue

The image has ~1.55 GB in size though it is based on Alpine Linux. For now it has two version of toolchains: `1.70.0-x86_64-unknown-linux-musl` and `nightly-x86_64-unknown-linux-musl`. The former may be used less frequently but was kept in case of `rustup override` for user's project.

## TODO

1. We could provide a **GPU** supported image as well as an [evcxr-jupyter](https://github.com/evcxr/evcxr/blob/main/evcxr_jupyter/README.md) supported image by Docker's [multi-stage builds](https://docs.docker.com/build/building/multi-stage/).
2. Shall we maintain a repository in DockerHub? If so, users can get different version of images by specifying tag and without need of building locally for convenience.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,37 @@
-FROM rust:1.70-alpine3.18
+FROM rust:1.70-slim-bullseye
 
-RUN apk add --no-cache \
-    # To get source code
+RUN apt update \
+    && apt install -y \
+    # To clone our repo
     git \
-    # To get musl header files
-    musl-dev 
+    # Needs for compiling openssl
+    pkg-config \
+    libssl-dev \
+    # Needs for compiling torch-sys
+    g++ \
+    # Needs for downloading libtorch
+    axel \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
 
-RUN rustup toolchain install nightly \
-    && rustup component add clippy rustfmt --toolchain nightly-x86_64-unknown-linux-musl \
-    && rustup default nightly
+RUN axel -qo /opt/libtorch.zip https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-2.0.1%2Bcpu.zip \
+    && cd /opt \
+    && unzip -qq libtorch.zip \
+    && rm /opt/libtorch.zip
 
+ENV LIBTORCH=/opt/libtorch
+ENV LD_LIBRARY_PATH=/opt/libtorch/lib
+
+RUN rustup component add \
+    # Used in `burn-import/src/formater.rs`
+    rustfmt \
+    clippy
+    
 WORKDIR /workspace
 
 RUN git clone https://github.com/burn-rs/burn.git
 
-RUN apk del git
+RUN apt autoremove -y \
+    git \
+    unzip \
+    axel
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM rust:1.70-alpine3.18
+
+RUN apk add --no-cache \
+    # To get source code
+    git \
+    # To get musl header files
+    musl-dev 
+
+RUN rustup toolchain install nightly \
+    && rustup component add clippy rustfmt --toolchain nightly-x86_64-unknown-linux-musl \
+    && rustup default nightly
+
+WORKDIR /workspace
+
+RUN git clone https://github.com/burn-rs/burn.git
+
+RUN apk del git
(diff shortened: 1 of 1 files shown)


antimora:
@bioinformatist, thank you so much for taking initiative and working on this. I saw this PR yesterday and I am thinking over what we want to achieve in the end. Definitely, I think it's easier for people to start working on the project right away - vscode also makes it easy just to open a project with Dockerfile and everything gets installed and cleaned up afterwards. 

At the moment I am still trying to scope the work. @nathanielsimard has suggested to make it work for GPU, which is probably valid. There might be already images with GPU support installed but it might NVIDIA specific.

Lets keep this draft open and collect some ideas and constraints. Your draft is a good start. I'll try testing it on my end.

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM rust:1.70-slim-bullseye
```
antimora, line 1: This version seems to have a high vulnerability. Probably this is very recent. 

https://hub.docker.com/layers/library/rust/1.70-slim-bullseye/images/sha256-53f49ce140a3d19c775da59c3d9b124c1830425ef5568ca45d1f29e76cb15f96?context=repo&tab=vulnerabilities

We should check if there is an updated version.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN axel -qo /opt/libtorch.zip https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-2.0.1%2Bcpu.zip \
+    && cd /opt \
+    && unzip -qq libtorch.zip \
+    && rm /opt/libtorch.zip
+
+ENV LIBTORCH=/opt/libtorch
+ENV LD_LIBRARY_PATH=/opt/libtorch/lib
+
+RUN rustup component add \
+    # Used in `burn-import/src/formater.rs`
```
antimora, line 26: Just a slight clarification - rustfmt and clippy operate independently from burn-import/src/formater.rs. That particular file has its own set of cargo dependencies.

antimora (changes requested):

Dockerfile
bioinformatist: Ahhh, I had never paid attention to the image vulnerability before. Thank you for pointing it out. It seems `bookworm` is the better one, I'll replace the base image with it.

Dockerfile
bioinformatist: Got it. I've removed the clippy. Users can install it only when needed.

Dockerfile
antimora: We hit the mine again =D 

1.70-slim-bookworm has the same vulnerability too. But it seems for different OS/ARCH, it's different.

https://hub.docker.com/layers/library/rust/1.70-slim-bookworm/images/sha256-162271b91daf42c769cb537ff6ec14bfd35a53e8392c8a30686c9b2368d25acb?context=repo&tab=vulnerabilities

This is what I see when I open it. Probably it defaults for my Mac M1 differently: 
image

bioinformatist: Emmm...It seems almost all Debian release have been suffering the poor vulnerability problem.

Since we also need GPU support, maybe we should turn to PyTorch image as the base.

The difference is to solve rust toolchain (use `1.70-slim-xxx`) or CUDA/CuDNN dependencies (PyTorch tags).

I think finally an elegant way will come with more experiments:thinking:

antimora:
Closing this for now to determine the goal of the dockerfile (which platforms, what goes inside, etc).

I put a comment in this ticket: https://github.com/burn-rs/burn/issues/175#issuecomment-1652554115

PR #387, opened 2023-06-05
Add more ops

nathanielsimard:


burn-wgpu/src/template/erf.wgsl
```diff
@@ -0,0 +1,22 @@
+// An approximation of the error function: https://en.wikipedia.org/wiki/Error_function#Numerical_approximations
```
antimora, line 1: Should mention this probably:

> 
> (maximum error: 1.5×10−7)
> 
> All of these approximations are valid for x ≥ 0. To use these approximations for negative x, use the fact that erf x is an odd function, so erf x = −erf(−x).

Are we safe for cases when x < 0? Can someone pass a negative x?

antimora (approved):
Asking to add more info to the doc to erf_positive function.

Approving ahead after the change.

burn-wgpu/src/template/erf.wgsl
nathanielsimard: The function below make sure it never happens, and that is why I called it erf_positive (only positive numbers).

```rust
fn erf(x: elem) -> elem {
    if (x < 0.0) {
        return -1.0 * erf_positive(-1.0 * x);
    }

    return erf_positive(x);
}
```

PR #389, opened 2023-06-06
ONNX: Add `Bool` element type

maekawatoshiki:
### Summary

- While parsing a model (hubert) mentioned at https://github.com/burn-rs/burn/issues/331), an error occurred due to the lack of support for boolean data types.
- I have added `ElementType::Bool` and the necessary utility code for it.

### DIFF
--- a/burn-import/src/onnx/ir.rs
+++ b/burn-import/src/onnx/ir.rs
@@ -67,7 +67,7 @@ pub enum TensorData {
     Int32(Vec<i32>),
     Int64(Vec<i64>),
     String(Vec<String>),
-    Bool(Vec<u8>),
+    Bool(Vec<bool>),
 }
 
 #[derive(Debug, Clone)]
--- a/burn-import/src/onnx/ir.rs
+++ b/burn-import/src/onnx/ir.rs
@@ -48,6 +48,7 @@ pub enum ElementType {
     Int64,
     String,
     Float16,
+    Bool,
 }
 
 #[derive(Debug, Clone)]
@@ -66,6 +67,7 @@ pub enum TensorData {
     Int32(Vec<i32>),
     Int64(Vec<i64>),
     String(Vec<String>),
+    Bool(Vec<u8>),
 }
 
 #[derive(Debug, Clone)]
@@ -351,6 +353,7 @@ impl fmt::Debug for TensorData {
             TensorData::Int32(v) => write!(f, "Int32({})", trunc(v)),
             TensorData::Int64(v) => write!(f, "Int64({})", trunc(v)),
             TensorData::String(v) => write!(f, "String({})", trunc(v)),
+            TensorData::Bool(v) => write!(f, "Bool({})", trunc(v)),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


burn-import/src/onnx/ir.rs
```diff
@@ -66,6 +67,7 @@ pub enum TensorData {
     Int32(Vec<i32>),
     Int64(Vec<i64>),
     String(Vec<String>),
+    Bool(Vec<u8>),
```
maekawatoshiki, line 70: I'm not sure whether to use `u8` or `bool`.

burn-import/src/onnx/to_burn.rs
```diff
@@ -287,6 +287,10 @@ impl Tensor {
             TensorData::Int32(val) => DataSerialize::new(val, self.shape.unwrap()).convert(),
             TensorData::Int64(val) => DataSerialize::new(val, self.shape.unwrap()).convert(),
             TensorData::String(_) => panic!("String tensor unsuported"),
+            TensorData::Bool(val) => {
+                DataSerialize::new(val.into_iter().map(u8::from).collect(), self.shape.unwrap())
+                    .convert()
```
maekawatoshiki, line 292: Because `ToPrimitive` is not implemented for `bool`, `Element` looks like not able to be implemented for `bool`.
What do you think?

nathanielsimard: Yep, `bool` doesn't implement `Element`. It causes problem sometimes, but `Element` really represent numeric types and bool isn't. 

Maybe look here https://stackoverflow.com/questions/72572544/how-to-convert-a-bool-array-into-a-byte-array for help on how to decode `Vec<u8>` into `Vec<bool>`. There are some manipulations to do. Not sure if there is a crate for that, but you could create two functions that convert `Vec<bool>` <=> `Vec<u8>` and test the flow properly. Those functions could be in `burn-tensor` since it can be used anywhere.

antimora: I have used bytemuck utility (crate) successfully in casting bytes into native types. Here is one example in ONNX: https://github.com/burn-rs/burn/blob/cb4f049eae4772b13d8fdcbb85ca0108e9edc3a3/burn-import/src/onnx/from_onnx.rs#L188

ONNX stores some information in bytes (u8s) to preserve the precision, so it should be straight forward to convert back with right tools.

maekawatoshiki: @antimora Yes, I know I can use bytemuck in a way like `TensorData::Bool(cast_slice(&tensor.raw_data[..]).to_vec())` if `TensorData::Bool(Vec<u8>)` is its definition, but now it is defined as `TensorData::Bool(Vec<bool>)` and `cast_slice` from bytemuck looks like it does not support `bool` here (raises compile error). This is what I was doing in previous commit https://github.com/burn-rs/burn/pull/389/commits/f926df282afbc96285cd10b4e132d328bf6e7323.

The problem I have in mind is:

> I realized that by converting bools into u8s for DataSerialize::new, we lose information on whether u8 was just an integer value or originally a bool value.

Aren't there any problem that we lose type information on original data?

nathanielsimard: I don't think `bytemuck` supports `bool`, you have to do it manually. I also think that we should not support `into_data_serialize` for bool tensors, similar to how we don't support it for string tensors.

maekawatoshiki:
I've modified the code so that it panics when `DataSerialize`ing bool tensors.

antimora (approved):

Issue #314, opened 2023-04-25
GRU: Gated Recurrent Unit


nathanielsimard:
### Feature description

GRU Module

### Feature motivation

Still useful in some applications.

agelas:
@nathanielsimard @antimora You can assign this issue to me- this should be relatively straightforward since GRUs are a simplification of LSTMs and I can pound out this issue when I have time.

Do you want this in a new folder or can I just put this inside the lstm folder? The GateController can be re-used for GRUs, and like I said before GRUs are technically a simplified variation of an LSTM, so I don't think users would be too surprised to find this under the lstm module.

antimora:
@agelas I assigned it to you per your request. @nathanielsimard can comment on the design.

nathanielsimard:
@agelas I think you can rename the folder `rnn` and put the file inside.  We might also change the visibility of the module:

```rust
pub mod rnn;
```
instead of:
```rust
mod rnn;
pub use rnn::*;
```

So that it's similar to how transformer and convolution layers are exported.

agelas:
@nathanielsimard I renamed the folder `rnn`. As for `mod.rs`, it looks like the transformer and convolution layers are exported using the latter method, i.e. transformer's `mod.rs` looks like:

```rust
mod decoder;
mod encoder;
mod pwff;

pub use decoder::*;
pub use encoder::*;
pub use pwff::*;
```

nathanielsimard:
I think we can re-export all RNN implementations to avoid long imports. Similar to conv and transformer modules.

antimora:
Closed via https://github.com/burn-rs/burn/pull/393

PR #393, opened 2023-06-09
Feat/gru

agelas:
Completes #314 .


antimora (approved):
Looks good from my end. 

@nathanielsimard should also review for correctness. I am less familiar with the details of GRU implementation.

nathanielsimard (approved):

PR #394, opened 2023-06-09
feat/Xavier glorot

louisfd:
- Added Xavier Glorot initialization. 
- Two versions: uniform or normal. 
- Changed LSTM default initialization to Xavier normal


burn-core/src/nn/lstm/lstm.rs
```diff
@@ -219,16 +217,16 @@ impl<B: Backend> Lstm<B> {
 
 #[cfg(test)]
 mod tests {
-
     use super::*;
     use crate::{module::Param, nn::LinearRecord, TestBackend};
     use burn_tensor::Data;
 
     #[test]
-    fn initializer_default() {
```
antimora, line 228: We should probably note with a TODO to test with default initializer (Xavier). @agelas is currently working in this area, perhaps he could add a test. At least we know we don't have the tests for the default initializer.

burn-core/src/nn/initializer.rs
```diff
@@ -14,7 +15,8 @@ pub enum Initializer {
     Constant(f64),
     Ones,
     Zeros,
-    // TODO: add Xavier initialization
+    XavierUniform(f64),
```
antimora, line 18: Since you're working on this. Could we add public documentation to the Initializer and its variants? This is for completeness. 

(sorry, it's actually my fault for not adding in the first place when I implemented it)

antimora (changes requested):
Looks great. Today I learned something new.

Requesting adding some documentation and TODO comment.

burn-core/src/nn/initializer.rs
louisfd: Of course!
In fact I shall refactor further, because I notice a strong similarity between the new Xavier init and the default initialization of linear and conv

antimora: > Of course! In fact I shall refactor further, because I notice a strong similarity between the new Xavier init and the default initialization of linear and conv

Yeah, you also probably noticed a peculiar initializer called UniformDefault (I made it up and @nathanielsimard could not come up with a better alternative name at the time). If it's the Xavier, then we should refactor as well. That would be great.

louisfd: Yes, UniformDefault is exactly what I'm changing. It's not exactly Xavier but it is similar. I will migrate this initialization in the initializer so the caller won't have to do it itself

antimora: Maybe it's one of those: https://pytorch.org/docs/stable/nn.init.html but I couldn't figure out.

burn-core/src/nn/initializer.rs
```diff
@@ -31,13 +33,45 @@ impl Initializer {
...
+    gain: &f64,
+    shape: S,
+) -> Tensor<B, D> {
+    let shape = shape.into();
+    let std = xavier_std(gain, &shape);
+    Tensor::<B, D>::random(shape, Distribution::Normal(0.0, std))
+}
+
+fn xavier_std<const D: usize>(gain: &f64, shape: &Shape<D>) -> f64 {
+    assert!(D >= 2);
```
nathanielsimard, line 67: I wouldn't add assert without adding an expalation:

```rust
assert!(D >= 2, "Can't compute standard deviation on shapes smaller than 2");
```

agelas:
@louisfd Just a heads up there's a merge conflict heading your way because both lstm's and gru's are under a new `rnn` module. Can you change the default initializer for gru to XavierNormal as well?

burn-core/src/nn/initializer.rs
```diff
@@ -110,4 +144,59 @@ mod tests {
...
+
+    #[test]
+    fn initializer_xavier_uniform_init() {
+        TB::seed(0);
+        let gain = 2.;
+        let (fan_in, fan_out) = (5, 6);
+        let bound = gain * sqrt(6. / (fan_in + fan_out) as f64);
+        let xavier_uniform: Tensor<TB, 2> =
+            Initializer::XavierUniform(gain).init([fan_in, fan_out]);
+        for item in xavier_uniform.to_data().value.iter() {
```
nathanielsimard, line 156: I would add a spacing here to separate the assertion part. You can also take ownsership of the data.

```rust
for item in xavier_uniform.into_data().values.into_iter() {
}
```

nathanielsimard (approved):

antimora (approved):
Clean and nice. Good to go. Thanks.

PR #395, opened 2023-06-11
Perf/wgpu/context

nathanielsimard:


burn-wgpu/src/ops/base.rs
```diff
@@ -181,6 +177,7 @@ impl<G: GraphicsApi> BaseOps<G> {
             "../template/index_assign_inplace.wgsl"
         );
 
+        println!("Can mut {:?}", tensor.can_mut());
```
antimora, line 180: print statement is left behind.

burn-wgpu/src/context/client.rs
```diff
@@ -0,0 +1,162 @@
...
+
+            // Wait for the buffer to be correctly registered so that inplace operations can be
+            // prioritize.
+            loop {
+                std::thread::sleep(std::time::Duration::from_micros(1));
+
+                if Arc::strong_count(&buffer_dest) == 1 {
+                    return buffer_dest;
+                }
+            }
```
antimora, line 81: The exit condition of this loop seems indeterministic. What are the possibilities this exit condition will never hit? Do we want to put a counter if it's a 1,000 (one second), then it panics?

antimora (changes requested):
Looks good. There are a couple minor changes.

burn-wgpu/src/context/client.rs
nathanielsimard: I added a assertion before, which will fail before going into the loop if it's called with wrong arguments.

```rust
            if wait_for_registered {
                assert_eq!(Arc::strong_count(&buffer_dest), 1, "You can't wait for the buffer to be registered when multiple references already exist.");
            }
```

It validates that the destination buffer has only one reference, so that it's safe to wait for it to returns to one reference after the cloning to send it to the context server.

antimora: What if the other party does not drop a reference? We do not have guarantees it will, especially in places like device to device communication (CPU-GPU). Theoretically it'll go infinite and rocket 🚀 will crash =D 

So my advice we program defensively in this case: we count ticks and exist if it takes too long.

antimora (approved):
I still think we should approach defensively in that loop. Consider my comment below. 

Other than that I think your looks good. Approving ahead to speed up turn around.

burn-wgpu/src/context/client.rs
nathanielsimard: If the reference isn't dropped it's a bug and it should crash, but I'm not sure if the state is even possible, the only thing that can crash is the channel, and since it's a sync channel, the execution will no go into the while loop. It's really a low level synchonisation between two threads using the atomic int frond the Arc instead of a mutex or another atomic primitive. The goal is to have the lowest performance hit possible while ensuring that inplace operations can be registered on the returned buffer. It should be considered unsafe for performance reason 😅. Otherwise I would have used a channel as callback (similar to how we read a buffer to bytes and sent them back). 

I'll add more documentation about that, but won't implement more assertion, since it may create more bugs or make the loop slower (so the best action would have been to use a callback channel instead).

PR #397, opened 2023-06-13
Add run-before-pr.sh script and fix clippy errors

### LINKED ISSUE #284: Add a before-commit.sh script to run local tests similar to CI
### Issue description

It has been observed that often developers miss running tests and checking formatting before committing their changes. This leads to issues being caught during CI testing, which could have been avoided with a simple pre-commit check. To streamline the process and ensure that developers can easily run the necessary tests, we propose creating a before-commit.sh script that performs the same build tasks as the CI pipeline.

### Proposed solution
We suggest adding a before-commit.sh script that developers can run before committing their changes. This script should include the following tasks:

1. Building the project with no_std configuration, if applicable
2. Running all tests
3. Checking code formatting

The script should return a non-zero exit code if any of the tasks fail, indicating that the developer should fix the issues before committing their changes.

Once the script is created, we should update the documentation to recommend developers to run the script before committing their changes. Additionally, we could consider adding a pre-commit hook that automatically runs the script to further streamline the process.

### Benefits

- Reduces the likelihood of issues being caught during CI testing, which can save time and resources.
- Ensures that the code is consistent with the project's formatting guidelines.
- Encourages developers to run tests and verify their changes before committing, leading to a more stable codebase.

antimora:
Fixes #284

1. Added the script
2. Added contributing guide and linked the script
3. Updated Readme document.

### DIFF
--- a/run-before-pr.sh
+++ b/run-before-pr.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# This script is run before a PR is created.
+# It is used to check that the code compiles and passes all tests.
+# It is also used to check that the code is formatted correctly and passes clippy.
+
+# Exit immediately if a command exits with a non-zero status.
+set -euo pipefail
+
+# Function to handle errors
+error_handler() {
+    local exit_status=$?
+    local line_number=$1
+    local command=$2
+
+    echo "Error on line $line_number"
+    echo "Command '$command' exited with status $exit_status"
+}
+
+# Signal trap to call error_handler when a command fails
+trap 'error_handler $LINENO $BASH_COMMAND' ERR
+
+# Function to build and test no_std
+build_and_test_no_std() {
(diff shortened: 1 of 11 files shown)


run-before-pr.sh
```diff
@@ -0,0 +1,88 @@
...
+
+    cd .. || exit
+}
+
+# Save the script start time
+start_time=$(date +%s)
+
+# Add wasm32 target for compiler.
+rustup target add wasm32-unknown-unknown
+rustup target add thumbv7m-none-eabi
```
nathanielsimard, line 66: I'm not sure I understand how those two lines add some verifications?

antimora: They will install missing targets. If they're installed, nothing happens. We can't expect the users to install these. These targets are used to test no_std.

PR #398, opened 2023-06-13
refactor/initializer

louisfd:
- Swap linear weight order from [input, output] to [output, input] to fit with the convention
- Refactor Initializer enum to output weights and (optionally) bias, which greatly simplifies model initialization
- Refactor Initializer::UniformDefault (which asked the model to implement its own initialization method) into Initializer::NormalizedUniform
- Put fan_in / fan_out computations in Shape


burn-core/src/nn/initializer.rs
```diff
@@ -10,74 +10,116 @@ use crate as burn;
...
     /// Fills tensor with specified value everywhere
     Constant(f64),
     /// Fills tensor with 1s everywhere
     Ones,
     /// Fills tensor with 0s everywhere
     Zeros,
+    /// Fills tensor with values drawn uniformly between specified values
+    Uniform(f64, f64),
+    /// Fills tensor with values drawn uniformly between -sqrt(1/fan_in) and sqrt(1/fan_in).
+    NormalizedUniform,
```
antimora, line 22: Thanks for the refactor! I was unsure about the origin of this uniform distribution during my own refactoring. However, I believe I now understand it. I've been trying to trace the origins of this initialization technique, and it appears to be based on Kaiming uniform initialization for linear layers, not XavierUniform as Nathaniel initially thought:

https://discuss.pytorch.org/t/clarity-on-default-initialization-in-pytorch/84696

> 1. In the Kaiming uniform initialization, `negative_slope=sqrt(5)` so the `gain=sqrt(2/6)=1/sqrt(3)`. If we substitute this into the `bound` formula, we get `bound = [1/sqrt(3)] * [sqrt(3/ fan_in)]` which simplifies to `bound = 1/sqrt(fan_in)`, which can be represented as `bound^2 = 1 / fan_in`.
> 2. In the linear implementation, `k= 1/in_features`, which in the case of Kaiming can be represented as `k=1/fan_in`. We want a boundary of `[-sqrt(k), sqrt(k)]` where `k = bound^2 = 1 / fan_in` from step 1.

Given this information, it may be worthwhile to introduce a proper Kaiming initialization and replace our current method:

1. Uniform: https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.kaiming_uniform_
2. Normal: https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.kaiming_normal_

I understand that this might significantly disrupt your current PR, so feel free to address this in a separate PR or file an issue ticket to handle it later. I'll leave that decision to you.

burn-tensor/src/tensor/shape.rs
```diff
@@ -14,6 +14,24 @@ impl<const D: usize> Shape<D> {
 
         num_elements
     }
+
```
antimora, line 17: We need documentation for this public API. If you could also mention why we need this (a typical use), it'd be great. Also having some demonstrative formula for calculation is helpful.

burn-tensor/src/tensor/shape.rs
```diff
@@ -14,6 +14,24 @@ impl<const D: usize> Shape<D> {
...
+    pub fn fan_in(&self) -> usize {
+        let receptive_field_size: usize = self.dims.iter().skip(2).product();
+        *self
+            .dims
+            .get(1)
+            .expect("Cannot get fan in of vector with dim < 2")
+            * receptive_field_size
+    }
+
+    pub fn fan_out(&self) -> usize {
```
antimora, line 27: Same

antimora (changes requested):
Looks good. Thanks for simplifying code!

There are minor requests. Please see comments. 

There is one question related to my recent discover. Please see my comment as well.

burn-core/src/nn/initializer.rs
antimora: @nathanielsimard Let us know of your thoughts as well.

burn-core/src/nn/initializer.rs
```diff
@@ -10,74 +10,116 @@ use crate as burn;
...
-    /// Fills tensor with values drawn from normal distribution with specified mean and std
-    Normal(f64, f64),
     /// Fills tensor with specified value everywhere
     Constant(f64),
     /// Fills tensor with 1s everywhere
     Ones,
     /// Fills tensor with 0s everywhere
     Zeros,
+    /// Fills tensor with values drawn uniformly between specified values
+    Uniform(f64, f64),
```
nathanielsimard, line 20: I think it would be valuable to add names to each argument when applicable.

```rust
pub enum Initializer {
    Uniform {
        min: f64,
        max: f64,
     },
     Normal {
         mean: f64,
         std: f64,
     } ,
     ...
}
```

I think it would help knowing what represent the value of each argument.

burn-core/src/nn/initializer.rs
```diff
@@ -10,74 +10,116 @@ use crate as burn;
...
     /// Fills tensor with values according to the normal version of Xavier Glorot initialization described in [Understanding the difficulty of training deep feedforward neural networks](https://proceedings.mlr.press/v9/glorot10a/glorot10a.pdf)
     XavierNormal(f64),
 }
 
 impl Initializer {
-    pub fn init<B: Backend, const D: usize, S: Into<Shape<D>>>(&self, shape: S) -> Tensor<B, D> {
+    pub fn init<B: Backend, const D: usize, S: Into<Shape<D>>>(
+        &self,
+        shape: S,
+        with_bias: bool,
```
nathanielsimard, line 35: I don't think it's a good idea to initialize weight and bias with a single function calls, though I admit the code before was pretty ugly. Often, the bias and the weight don't share the same initializer, so I don't think it sould be the default behavior. 

In fact, it's a bit weird to have an initializer config for a module's weight but none for the bias. Maybe we should replace the `bias` flag with an optional initializer.

```rust
/// Configuration to create a [Linear](Linear) layer.
#[derive(Config, Debug)]
pub struct LinearConfig {
    /// The size of the input features.
    pub d_input: usize,
    /// The size of the output features.
    pub d_output: usize,
    /// The type of function used to initialize the weight matrix
    #[config(default = "Initializer::NormalizedUniform")]
    pub initializer_weight: Initializer,
    /// (Optional) The type of function used to initialize the bias vector.
    /// If no initializer is passed, no bias will be used for the linear transformation.
    #[config(default = "Some(Initializer::NormalizedUniform")]
    pub initializer_bias: Option<Initializer>,
}
```

Then the code becomes prettier:

```rust
let weight = self.initializer_weight.init(shape_weight);
let bias = self.initializer_bias.map(|initializer| initializer.init(shape_bias));
```

If we want to go further, we could replace the `Option` with our own enum and implement init on that enum, but I'm not sure if it's a good idea yet 😅.

burn-core/src/nn/initializer.rs
```diff
@@ -10,74 +10,116 @@ use crate as burn;
...
+    ) -> (Tensor<B, D>, Option<Tensor<B, 1>>) {
+        let shape = shape.into();
         match self {
-            Self::Uniform(a, b) => Tensor::<B, D>::random(
-                shape,
+            Self::Constant(value) => constant_weight_and_bias(shape, with_bias, *value),
+            Self::Ones => constant_weight_and_bias(shape, with_bias, 1.),
+            Self::Zeros => constant_weight_and_bias(shape, with_bias, 0.),
```
nathanielsimard, line 41: Even if it makes sense to reuse the constant function, it's less performant and require more operations to be executed, so I would use the tensor functions `Tensor::ones` and `Tensor::zeros` for that instead.

burn-core/src/nn/initializer.rs
```diff
@@ -10,74 +10,116 @@ use crate as burn;
...
 }
 
-fn xavier_std<const D: usize>(gain: &f64, shape: &Shape<D>) -> f64 {
-    assert!(
-        D >= 2,
-        "Can't compute Xavier standard deviation on shapes smaller than 2"
-    );
+fn normalized_uniform<B: Backend, const D: usize>(
+    shape: &Shape<D>,
+) -> Distribution<<B as Backend>::FloatElem> {
```
nathanielsimard, line 102: Can it be replaced with `B::FloatElem` ?

nathanielsimard (changes requested):

burn-core/src/nn/initializer.rs
louisfd: I was also unsure about having one function call for both weight and bias. The reason I did it is that distribution such as the normalized uniform that was previously used or Xavier depend on their fan in, but if we only have the bias shape it cannot know the fan in because it only has the output size.

burn-core/src/nn/initializer.rs
antimora: I personally like tuples for the initialization. It's cleaner for the configuration like this: 
```rust
#[derive(Config)]
pub struct LstmConfig {
    /// Lstm initializer
    #[config(default = "Initializer::XavierNormal(1.0)")]
    pub initializer: Initializer,
}
```
With the named struct, you need extra braces and param name + `=`.

I vote we leave it.

louisfd:
I pushed some new code, where I changed uniformdefault/normalizeduniform for Kaiming initialization. Also there has been a major refactor since my last version so that initialization returns a tensor and not both weight and biases, all while considering fan_in/fan_out. 

It panics on two of the newer tests that passed in the main branch. @antimora I suspect the saved model from the new MNIST test to have its linear weights recorded as [d_input, d_output] instead of the now [d_output, d_input]

antimora (changes requested):
Looks good overall. Thanks for adding Kaiming* initializers - definitely the initializer has become richer. 

I'd like to request a couple changes: cleaning up the comments/docs/println. and replacing InitializerOptions with a simpler solution. Please see my inlined comment for details/explanation.

burn-core/src/nn/conv/conv2d.rs
```diff
@@ -74,24 +75,21 @@ pub struct Conv2d<B: Backend> {
...
             self.channels[1],
             self.channels[0],
             self.kernel_size[0],
             self.kernel_size[1],
         ]);
-
+        let fan_in = shape.fan_in();
+        let weight = self
+            .initializer
+            .init(shape, InitializerOptions::default().with_fan_in(fan_in));
```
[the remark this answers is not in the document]
louisfd: I was really trying to come up with something that won't force us to have None arguments on every simple call like
Initializer::Zeros.init(None, None)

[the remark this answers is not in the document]
louisfd: Yeah I like that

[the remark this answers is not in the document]
nathanielsimard: I think it's better to have an option struct here, since we may add more arguments in the future and adding an optional field to a struct isn't a breaking change, but adding a new argument to each `init` call is.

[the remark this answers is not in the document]
louisfd: What about init and init_with, where the latter takes an option struct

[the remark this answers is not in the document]
antimora: Looking at the PyTorch init implementation and thinking realistically we will not be adding more options. The options from the struct are passed from the caller to init and they aren't any besides fan in and out. Any other options are part of the enum struct. The true name for the option is actually InitParams, which is small in its scope. We could also refactor if needed but currently we do not have a good use case for the extra flexibility. Lets keep it simple =)

burn-core/src/nn/rnn/gru.rs
```diff
@@ -21,7 +21,7 @@ pub struct GruConfig {
     pub bias: bool,
     /// Gru initializer
     /// TODO: Make default Xavier initialization. https://github.com/burn-rs/burn/issues/371
-    #[config(default = "Initializer::Uniform(0.0, 1.0)")]
+    #[config(default = "Initializer::Uniform{min:0.0, max:1.0}")]
```
antimora, line 24: Can we also change this to Xavier? This probably got in during your PR after the merge and that's why we missed it.

antimora (approved):
Looks good. Thanks for the fixes. 

There is one little change still required please see my comment. 

I am approving in advance never the less. Please free to merge after the fix.

louisfd:
With the new way of specifying fan_in and fan_out using init_with, it had become overkill to unify their computation in the implementation of Shape.  Instead each model will need (or be able to, depends on the point of view) to specify its fan_in/fan_out when necessary. 
Thus we are no more obligated to assume that the outer dim is always first and inner dim second. I could then revert to the original linear shape, which saves us from transposing the weights in the forward function.

nathanielsimard (approved):

PR #404, opened 2023-06-15
refactor: better templating for wgsl shaders

nathanielsimard:


antimora (approved):
Looks good. 

Probably you're already taking advantage the templates parsing ability (parsed after initialization when calling `::new`). This way you cache the parsed templates so that your merges with values become quick.

PR #405, opened 2023-06-15
Fix wgpu context creation

nathanielsimard:


antimora (approved):
Looks good and works on my Mac M1.

PR #409, opened 2023-06-18
feat: mask_where

nathanielsimard:
Replace mask_scatter with mask_where.

The mask_scatter was implemented such that it was mask_where. Since mask_where is more common than mask_scatter, I just replaced the function. I added better tests and updated the doc. The is similar to https://pytorch.org/docs/stable/generated/torch.where.html?highlight=where#torch.where however the mask is reversed to be more inline with mask_fill, otherwise it would be the inverse.


antimora (approved):
Looks good to me.

PR #423, opened 2023-06-21
fix(loss): Declare all contents of reduction as public

bioinformatist:
The enum Reduction should also be public as it can be used when calling the forward process of MSE.


burn-tensor/src/tests/ops/sqrt.rs
```diff
@@ -2,6 +2,7 @@
 mod tests {
     use super::*;
     use burn_tensor::{Data, Tensor};
+    use std::f32::consts::SQRT_2;
```
antimora, line 5: This should be `use core::f32::consts::SQRT_2;` for no_std compatibility. 

Recently we have added a new script `./run-before-pr.sh` to simulate CI tests.  (`./run-before-pr.sh` will be renamed to `run-checks.sh` soon fyi). You run CI problems ahead.

antimora (changes requested):

burn-tensor/src/tests/ops/sqrt.rs
bioinformatist: Thanks for quick reply!:smile: I've fixed it.

antimora (approved):

PR #443, opened 2023-06-27
Fix: wgpu powf

nathanielsimard:
### Checklist

### Changes

Can raise negative number with powf. Also can run manually burn-core tests with wgpu backend. Not everything passes, so w'll wait before adding the check to the CI.

### Testing

Added two more generic tests.


antimora (approved):
LGTM

louisfd:
@nathanielsimard Before this goes into main I'd like to talk with you about it, as I'm still unsure the absolute value is a valid solution

PR #446, opened 2023-06-28
Perf/unary binary

nathanielsimard:
### Checklist

### Changes

How the unary and binary operations work (2D instead of 1D). The problem being 1D is a limitation on the size of tensor (max 65536 * 1024), now it's (65536 * 65536 * 1024)

### Testing

Added some specific tests to check the number of invocation is correct.

### DIFF
--- a/burn-wgpu/src/kernel/binary_elemwise.rs
+++ b/burn-wgpu/src/kernel/binary_elemwise.rs
@@ -61,7 +61,7 @@ pub fn binary_elemwise<K: StaticKernel, E: WgpuElement, const D: usize, const WO
     lhs: WgpuTensor<E, D>,
     rhs: WgpuTensor<E, D>,
 ) -> WgpuTensor<E, D> {
-    lhs.assert_is_on_save_device(&rhs);
+    lhs.assert_is_on_same_device(&rhs);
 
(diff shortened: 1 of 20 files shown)


louisfd (changes requested):
I have one simplification to suggest. 
The rest makes a lot of sense to me.

burn-wgpu/src/kernel/binary_elemwise.rs
```diff
@@ -72,65 +74,109 @@ pub fn binary_elemwise<K: StaticKernel, E: WgpuElement, const D: usize>(
...
+pub fn binary_elemwise_inplace<
+    K: StaticKernel,
+    E: WgpuElement,
+    const D: usize,
+    const WORKGROUP: usize,
+>(
+    lhs: WgpuTensor<E, D>,
+    rhs: WgpuTensor<E, D>,
+) -> WgpuTensor<E, D> {
+    lhs.assert_is_on_save_device(&rhs);
```
antimora, line 119: Should this be `assert_is_on_same_device` instead of `assert_is_on_save_device`? Maybe we should rename the function name.

antimora (commented):
Looks good. I had one small naming suggestion.

PR #448, opened 2023-06-29
Perf/wgpu/benches

nathanielsimard:
Add benchmarks tools for wgpu.


burn-wgpu/src/benchmark.rs
louisfd, line 1: Is it intended that benchmark.rs is in src, but benches files are outside src

louisfd (approved):

burn-wgpu/src/benchmark.rs
nathanielsimard: Yes, it's the way it's done with cargo bench.

PR #453, opened 2023-07-01
Refactor/wgpu/mask

nathanielsimard:
### Checklist

### DIFF
--- a/burn-wgpu/src/template/mask/where_inplace.wgsl
+++ b/burn-wgpu/src/template/mask/where_inplace.wgsl
@@ -14,9 +14,15 @@ var<storage, read> mask: array<u32>;
 @binding(3)
 var<storage, read> info: array<u32>;
 
+const WORKGROUP_SIZE_X = {{ workgroup_size_x }}u;
+
 @compute
-@workgroup_size({{ workgroup_size_x }}, 1, 1)
-fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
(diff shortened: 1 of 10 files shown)


burn-wgpu/src/template/mask/where_inplace.wgsl
```diff
@@ -33,12 +39,20 @@ fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
...
 
-        index_input += global_id.x / stride_input % shape_input * stride_input;
-        index_value += global_id.x / stride_input % shape_value * stride_value;
-        index_mask += global_id.x / stride_input % shape_mask * stride_mask;
+        index_input += id / stride_input % shape_input * stride_input;
+        index_value += id / stride_input % shape_value * stride_value;
+        index_mask += id / stride_input % shape_mask * stride_mask;
+    }
+
+    var condition = mask[index_mask] != 0u;
```
louisfd, line 47: technically, lines 47-51 could be simplified to 
`let condition = mask[index] != reverse;`
But I'm not sure the performance gain would outweigh the readability loss

louisfd (approved):

burn-wgpu/src/template/mask/where_inplace.wgsl
nathanielsimard: Nop, since you can only compare to 0s, not 1s, since the buffer cast from float to u32 can transform any non zero value to anything else, you can only do comparison with zero.

PR #456, opened 2023-07-02
fix: wgpu cat can support bigger tensors

nathanielsimard:
### Checklist

### Changes

Tensors can have 65536 * 65536 * 1024 elements vs 65536 * 1024 before.

### Testing

Add two tests


louisfd (approved):
Looks good

PR #457, opened 2023-07-02
Implement TryFrom for Data for nested slices and Vecs

Liburia:
### Checklist

### Changes
This PR allows Data to be constructed from 2D, 3D, and 4D slices and Vecs

### Testing

Tested only via the tests module

### DIFF
--- a/burn-tensor/src/tensor/data.rs
+++ b/burn-tensor/src/tensor/data.rs
@@ -380,8 +380,8 @@ impl<E: core::fmt::Debug + Copy> From<&[E]> for Data<E, 1> {
 /// Error for invalid nested data shape
 #[derive(Debug, Clone, Copy)]
 pub struct JaggedCollectionError(usize);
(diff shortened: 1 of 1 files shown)


antimora:
I am not sure if this solution scales and not sure if we should support deep nested Vec/arrays/slices in the burn API. We have some convenient tools for converting arrays/vec/slice and they're convince only and not required. 

I would recommend for any deep nested arrays/vec/slice to flatten first (row major), then call `from_floats` Tensor method (or Data) and reshape, like this in this example: https://github.com/burn-rs/burn/blob/main/examples/mnist-inference-web/src/web.rs#L43

```rust
// input: &[f32]

let input: Tensor<Backend, 3> = Tensor::from_floats(input).reshape([1, 28, 28]);
```
 
Currently V

antimora:
Closing this PR to come up with a better way to handle nested array conversion. This ticket will track this work: https://github.com/burn-rs/burn/issues/551

PR #459, opened 2023-07-02
Add Tensor as a module constant

antimora:
### Checklist

### Changes

Previously it was not possible to add a tensor attribute to a module struct. Now this is possible:

```rust
// Define the model structure
#[derive(Module, Debug)]
pub struct ClassificationModel<B: Backend> {
    transformer: TransformerEncoder<B>,
    embedding_pos: Embedding<B>,
    layer_norm: LayerNorm<B>,
    output: Linear<B>,
    n_classes: usize,
    max_seq_length: usize,
    sinusoids: Tensor<B, 2>, // <-- Tensor works now
}

```

### Testing

Unit testing and manually verifying exported module like the following:

```rust
fn main() {
    let trans_encoder = TransformerEncoderConfig::new(64, 512, 8, 4)
        .with_norm_first(false)
        .with_dropout(0.25);

    type TestBackend = NdArrayBackend<f32>;
    let model = ClassificationModelConfig::new(trans_encoder, 10, 64, 87).init::<TestBackend>();

    let recorder = PrettyJsonFileRecorder::<FullPrecisionSettings>::default();

    recorder
        .record(model.clone().into_record(), "model.json".into())
        .unwrap();
}
```

### DIFF
--- a/burn-core/src/module/param/constant.rs
+++ b/burn-core/src/module/param/constant.rs
@@ -1,5 +1,15 @@
-use crate::{self as burn, record::Record};
+use crate::{
+    self as burn,
+    module::{ADModule, Module, ModuleMapper, ModuleVisitor},
+    record::Record,
+};
 use burn::record::PrecisionSettings;
+use burn_tensor::{
+    backend::{ADBackend, Backend},
+    Tensor,
+};
+
+use super::ParamId;
 
 /// Record used for constant type implementing the [module](crate::module::Module) trait.
 #[derive(Debug, Clone, Copy, new)]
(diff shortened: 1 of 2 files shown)


burn-core/src/module/param/constant.rs
```diff
@@ -100,3 +110,79 @@ constant!(i64);
...
+
+    fn map<M: ModuleMapper<B>>(self, mapper: &mut M) -> Self {
+        // Important:
+        // We need to implement visit method for Tensor Module because
+        // to_device will be called during the visit method of the ModuleVisitor
+
+        // We are using a dummy param id because the visit method requires a param id
+        let dummy_param_id = ParamId::new();
+        mapper.map(&dummy_param_id, self)
+    }
```
nathanielsimard, line 135: Eventhough it's simpler, I think I would not implement `map` and `visit`, but all other methods manually. This way, it would avoid an allocation for the Param and it's more explicit. For example, I would not add the parameter count to the `num_params`, but I would handle `devices`,  `to_device` and `fork` correctly.

nathanielsimard (changes requested):

burn-core/src/module/param/constant.rs
antimora: We discussed offline and we concluded Module little more work, which will be tracked here: https://github.com/burn-rs/burn/issues/461.

PR #466, opened 2023-07-04
Refactor index => slice

### LINKED ISSUE #449: Rename tensor index to slice
### Feature description

Apologies if this exists already (and the short novel that follows), I surely couldn't find documentation of it.

I want the ability to create a slice of an existing tensor. 
This could either be by copy or by reference, I don't care. Ideally this would be by reference, even with the ability to mutate.

### Feature motivation

I need to be able to iterate over rows of a matrix, or matrices in a stack, etc.  These operations are simple to do in all the popular python ML packages, but not in Burn as far as I'm aware.  I propose a more general purpose API for tensor slicing.

One cute way to do this would be to implement the traits
- `Index<[Range<usize>; 1]>`
- `Index<[Range<usize>; 2]>`
- `Index<[Range<usize>; 3]>`
- `Index<[Range<usize>; 4]>`

Depending on the const dimensionality `D` of the tensor.  The API should probably provide an opaque `Tensor<...>` back, which internally remembers the data which backs it and the range of elements it addresses along the axes.

So in my perfect world you could write something like:

```rust
let view_a = tensor[[1, .., 0..3]];
let view_b = tensor[[1, .., 3..index]];

let a: f32 = important_op(view_a);
let b: f32 = important_op(view_b);

view_b *= (a - b);
```

I don't know how Tensor is implemented, so creating im/mutable references to it could be nasty and require reference counting and maybe locking, but this overhead is *probably* no big deal from a perf / memory perspective if the correct design is chosen.  Instead of forcing an Arc/Mutex onto every user of `Tensor`, perhaps a better approach would be to make the lifetimes and mutability visible to the compiler, although this comes along with the mutable-immutable borrowing collision issue.

Using Arc/Mutex, a `Tensor` could be backed by a `Data`, with a reasonable compromise:

```rust
struct Tensor<B, const D: usize> {
  backing: Data<B, D>,
  /// ...
}

enum Data<B, const D: usize> { 
  Owned(/* B-dependent value */),
  Shared {
    tensor: Arc<Mutex</* B-dependent value */>>,
    addressed_range: [ Range<usize>; D ]
  }
}
```

Locking could be done more intelligently by locking based on the addressed elements, not the entire tensor. E.g. you can read from one half just fine while an operation to mutate another half is occurring.  I don't know exactly how this would actually work efficiently.

Being able to slice into a tensor and read or mutate only a subset of its elements is very valuable and a fundamental operation needed for any ML / tensor library, so looking into solidifying an approach to it now would be good.

Dealing with different backend API semantics will likely not be as pretty.

nathanielsimard:
### Checklist

### Related Issues/PRs

Fix #449

Also refactor wgpu kernels.


burn-wgpu/src/kernel/index/slice.rs
```diff
@@ -0,0 +1,139 @@
+use crate::{
+    element::WgpuElement,
+    kernel::{build_info, elemwise_workgroup, KernelSettings},
+    kernel_wgsl,
+    tensor::WgpuTensor,
+};
+use burn_tensor::Shape;
+use std::ops::Range;
+
+kernel_wgsl!(IndexRaw, "../../template/index/index.wgsl");
```
louisfd, line 10: The kernel structs and files should be renamed

burn-wgpu/src/kernel/index/select.rs
```diff
@@ -0,0 +1,186 @@
+use crate::{
+    element::WgpuElement,
+    kernel::{build_info, elemwise_workgroup, KernelSettings},
+    kernel_wgsl,
+    tensor::WgpuTensor,
+};
+
+kernel_wgsl!(IndexSelect, "../../template/index/index_select.wgsl");
```
louisfd, line 8: The kernel structs and files should be renamed

louisfd (changes requested):
Good to go

antimora (approved):
Looks good from my end.

PR #481, opened 2023-07-07
Add sinusoidal positional embedding module

antimora:
### Checklist

NOTE: Currently the build on Mac M1 is broken, see https://github.com/burn-rs/burn/issues/480
so WGPU tests were skipped

### Changes

Add a brand new sinusoidal positional embedding module.

### Testing

Unit testing using values generated from a reference implementation.


burn-core/src/nn/sin_embedding.rs
```diff
@@ -0,0 +1,198 @@
...
+    // concatenate these two arrays along the second dimension
+    // (i.e., it adds the second array as new columns to the first array)
+    let scaled_time_sin_cos = scaled_time_sin
+        .into_iter()
+        .zip(scaled_time_cos.into_iter())
+        .map(|(mut x, mut y)| {
+            x.append(&mut y);
+            x
+        })
+        .collect::<Vec<_>>();
```
nathanielsimard, line 134: Would it be better to use tensor operations here instead? I think the arange function might be missing for the float tensor, but other than that it should be fine.

burn-core/src/nn/sin_embedding.rs
```diff
@@ -0,0 +1,198 @@
...
+    d_model: usize,
+
+    /// Max time scale to use.
+    #[config(default = "10_000")]
+    max_timescale: usize,
+}
+
+/// Lookup table to store sinusoidal vectors for positional embedding.
+#[derive(Module, Debug)]
+pub struct SinEmbedding<B: Backend> {
```
nathanielsimard, line 28: I would prefer `PositionalEncoding` instead of `SinEmbedding`.

burn-core/src/nn/sin_embedding.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::tensor::Tensor;
+use burn_tensor::Data;
+
+use libm::{cosf, expf, logf, sinf};
+
+/// Configuration to create an [SinEmbedding](SinEmbedding) layer.
+#[derive(Config)]
+pub struct SinEmbeddingConfig {
+    /// The number of embedding vectors.
+    n_embedding: usize,
```
nathanielsimard, line 16: I would rename it to `max_sequence_size` since it's easier to know what value should be used. A default of 5000 is often common.

burn-core/src/nn/sin_embedding.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl<B: Backend> SinEmbedding<B> {
+    /// Applies the forward pass on the input tensor by adding the sinusoids to the input.
+    ///
+    /// # Shapes
+    ///
+    /// - input: [batch_size, seq_length, d_model]
+    /// - output: [batch_size, seq_length, d_model]
+    pub fn forward(&self, input: Tensor<B, 3>) -> Tensor<B, 3> {
+        input.add(self.sinusoids.clone())
```
nathanielsimard, line 50: I think a slice is normally needed here. See https://pytorch.org/tutorials/beginner/transformer_tutorial.html for reference.

nathanielsimard: I often see dropout as well

nathanielsimard (changes requested):

burn-core/src/nn/sin_embedding.rs
antimora: It's easy to understand if the logic is in Rust and there is no advantage to use tensor ops for one time initialization. I'd suggest we'd leave it as is.

antimora:
I'll add the slice for varying length, and I will not add dropout per our discussiong. If someone wants it, they can add easily in their model.

antimora:
> It's easy to understand if the logic is in Rust and there is no advantage to use tensor ops for one time initialization. I'd suggest we'd leave it as is. 

I have refactored it to match positional encoding logic and simplified the logic.

antimora:
@nathanielsimard Fixed all the issues previously discussed. Ready for review again.

nathanielsimard (approved):

PR #485, opened 2023-07-09
Update plot rendering from textplots-rs to termplot

Poxxy:
### Checklist

### Related Issues/PRs

This solves issue here: https://github.com/burn-rs/burn/issues/211

### Changes

Rendering of plots relied on `textplot-rs` which has a couple of issues:

1. The dots were not rendering correctly, causing a confusion in what the graph is showing.
2. `textplot-rts` has a dependency with `nom` which causes a compilation warning when building `burn`

The changes made reflect the requirements for using `termplot`. The main two changes made:

1. Coloring is not an option for `termplot` so all references to color have been removed.
2. `termplot` uses `f64` values instead of `f32` values which were previously used. Conversions have been made where necessary.
3. `width` variable was of type `u32` but has been changed to `usize` to match `height`. Use of `usize` is better practice in this case as well.
4. Syntax has been modified as needed to work with `termplot`.

### Testing

Testing has been done by rendering plots for training data such as `mnist` example.

### DIFF
--- a/burn-train/src/metric/dashboard/plot.rs
+++ b/burn-train/src/metric/dashboard/plot.rs
@@ -172,10 +172,13 @@ impl TextPlot {
         };
 
         plot.set_domain(Domain(x_min..x_max))
-            .set_codomain(Domain(0.0..100.0))
+            .set_codomain(Domain(0.0..200.0))
             .set_size(Size::new(width, height))
             .add_plot(Box::new(plot::Graph::new(train_plot)))
             .add_plot(Box::new(plot::Graph::new(valid_plot)))
+            .set_title("Training Metrics")
+            .set_x_label("X-Axis: Iterations")
+            .set_y_label("Y-Axis: Accuracy")
             .to_string()
     }
 }
--- a/burn-train/src/metric/dashboard/plot.rs
+++ b/burn-train/src/metric/dashboard/plot.rs
@@ -1,6 +1,6 @@
 use rgb::RGB8;
 use terminal_size::{terminal_size, Height, Width};
-use textplots::{Chart, ColorPlot, Shape};
+use termplot::{Plot, Domain, Size, plot};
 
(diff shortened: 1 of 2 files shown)


Poxxy:
Attached is an example rendering on my system which uses VSCode dark mode on Linux Mint.

improved_rendering

antimora:
I can confirm no compilation warning. 

I have run your changes locally on Mac M1 via vscode's term, the dots still appear as before. So your term must have a nicer fonts that render the dots differently. Maybe there is a setting in the library to adjust this somehow so an average user can see the plots clearly.

Here is what I am seeing now: image

Also there are RED/BLUE labels in `cli.rs` that probably need to be updated since we don't have colors:

["  - {} RED: train | BLUE: valid \n{}",](https://github.com/Poxxy/burn/blob/0d17026a2d61ecd5f59bff74197ba0d466dccf07/burn-train/src/metric/dashboard/cli.rs#L140)

burn-train/src/metric/dashboard/plot.rs
```diff
@@ -137,13 +133,52 @@ impl TextPlot {
...
+        }));
+
+        plot.set_domain(Domain(x_min..x_max))
+            .set_codomain(Domain(0.0..100.0))
+            .set_size(Size::new(width, height))
+            .add_plot(Box::new(train_graph))
+            .add_plot(Box::new(valid_graph))
+            .set_title("Training Metrics")
+            .set_x_label("X-Axis: Iterations")
+            .set_y_label("Y-Axis: Accuracy")
```
nathanielsimard, line 181: The Y-Axis isn't always accuracy, it depends on the metric.

nathanielsimard (changes requested):
It doesn't work with the text-accuracy example from my side, so maybe there is more work to be done before being merged.

Issue #413, opened 2023-06-19
Add static `full` method for tensor initialization with custom values


antimora:
**Description:**

Currently, we have methods `Tensor::<B, D>::zeros()` and `Tensor::<B, D>::ones()` for initializing tensors with zeros and ones, respectively. However, there is no efficient way to initialize a tensor with a custom fill value. 

In the case of `Initializer::Constant` in `initializer.rs`, we currently use a workaround: `Tensor::<B, D>::zeros(shape) + *value`. This requires two steps for initialization and is inefficient.

**Proposal:**

1. Implement a new static method (`Tensor::<B, D>::full(shape, fill_value)) for `Tensor::<B, D>` that allows for the creation of a tensor with a custom fill value in a single step. The method should follow the syntax similar to: `Tensor::<B, D>::from_value(shape, value)`.
   
2. After implementing this, we need to update the `init_with` method for `Initializer::Constant` in `initializer.rs` to use the newly created method.

**Benefits:**

This will simplify the code, improve readability, and also increase the efficiency of tensor initialization with custom fill values by eliminating the need for an extra step. 

**Tasks:**

- [ ] Implement `Tensor::<B, D>::from_value(shape, value)` method
- [ ] Update `init_with` method in `Initializer::Constant` to use the new method
- [ ] Test to ensure everything works as expected
- [ ] Update relevant documentation

antimora:
Completed via https://github.com/burn-rs/burn/pull/486

PR #486, opened 2023-07-09
Add static full method for tensor intialization with custom values

paguseva:
### Checklist 

- [x]  Confirm that `run-checks.sh` has been executed.

### Related Issues/PRs

* #413 

### Changes

Added a new method for tensor initialization and implemented it for all of the backends. Used this in `init_with`.

### Testing

Unit tests

### DIFF
--- a/burn-core/src/nn/initializer.rs
+++ b/burn-core/src/nn/initializer.rs
@@ -95,7 +95,7 @@ impl Initializer {
         match self {
             // TODO replace with full() method when implemented
             // https://github.com/burn-rs/burn/issues/413 is the issue tracking this
-            Initializer::Constant { value } => Tensor::<B, D>::zeros(shape) + *value,
+            Initializer::Constant { value } => Tensor::<B, D>::full(shape, *value),
             Initializer::Ones => Tensor::<B, D>::ones(shape),
             Initializer::Zeros => Tensor::<B, D>::zeros(shape),
             Initializer::Uniform { min, max } => uniform_draw(shape, *min, *max),
--- a/burn-tensor/src/tensor/api/numeric.rs
+++ b/burn-tensor/src/tensor/api/numeric.rs
@@ -106,9 +106,23 @@ where
         Self::ones_device(shape, &B::Device::default())
     }
 
-    /// Create a tensor of the given shape where each element is zero.
+    /// Create a tensor of the given shape where each element is one.
     pub fn ones_device<S: Into<Shape<D>>>(shape: S, device: &B::Device) -> Self {
         Self::new(K::ones(shape.into(), device))
+    }
+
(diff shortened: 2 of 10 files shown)


burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -65,6 +65,25 @@ pub trait TensorOps<B: Backend> {
...
+    /// 
+    /// # Returns
+    /// 
+    /// The tensor filled with given value
+    fn full<const D: usize>(
+        shape: Shape<D>, 
+        fill_value: B::FloatElem, 
+        device: &B::Device
+    ) -> B::TensorPrimitive<D> {
+        Self::from_data(Data::full(shape, fill_value), device)
```
nathanielsimard, line 84: I think I would use `Self::zeros(shape, device) + fill_value` for the default implementation, it's more performant on most backends. We can also remove the `burn-wgpu` implementation since it's pretty much what it does. The implementation can be the same for `int_full`.

nathanielsimard (changes requested):

burn-core/src/nn/initializer.rs
```diff
@@ -95,7 +95,7 @@ impl Initializer {
         match self {
             // TODO replace with full() method when implemented
             // https://github.com/burn-rs/burn/issues/413 is the issue tracking this
```
antimora, line 97: Left over TODO comment should be removed as well.

burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -106,11 +106,25 @@ where
         Self::ones_device(shape, &B::Device::default())
     }
 
-    /// Create a tensor of the given shape where each element is zero.
+    /// Create a tensor of the given shape where each element is one.
```
antimora, line 109: Thanks for catching and remove the doc error.

antimora (changes requested):
Thank you for adding this new functionality. 

I have a minor dead todo comment removal request and ask to add missing unit tests. You mentioned the change was unit tested but I couldn't find it. 

If you didn't have a chance to add them and not sure how (because it's kinda obfuscated for tensor ops), here is a simple guide. Burn's tensor testing is performed by the backends and the tests are exposed via macros. Once you understand the flow, it's easy. 

1. add test under `burn-tensor/src/tests/ops`
2. add module under `burn-tensor/src/tests/ops/mod.rs`
3. modify burn-tensor/src/tests/mod.rs and add macro call
4. run tests (`cargo test`) under a backend (burn-ndarray or burn-tch) or run all tests via `run-checks`. 

You can also find PR example for `flatten`: https://github.com/burn-rs/burn/pull/260/files.

nathanielsimard (approved):
It looks good to me. After the comments from @antimora are resolved, I'll merge.

PR #488, opened 2023-07-11
Fix: can't use init or load with a record with correct gradient requi…

nathanielsimard:
### Changes

Implement a more specific recorder for param which creates a tensor with the same behavior as when creating a normal param.

### Testing

Add a basic test with the linear module to validate the behavior.


antimora (approved):
lgtm

antimora:
Looks good to me. If I understood the problem correctly, one cannot load param with require_grad enabled?

PR #490, opened 2023-07-12
Add arange with steps op for int tensor

antimora:
### Checklist

### Changes

Add arange with steps op for int tensor

### Testing

Added new unit tests

### DIFF
--- a/burn-tensor/src/tensor/api/int.rs
+++ b/burn-tensor/src/tensor/api/int.rs
@@ -5,20 +5,43 @@ impl<B> Tensor<B, 1, Int>
 where
     B: Backend,
 {
-    /// Returns a new integer tensor on the default device which values are generated from the given range.
+    /// Returns a new integer tensor on the default device.
+    ///
+    /// # Arguments
+    ///
+    /// * `range` - The range of values to generate.
     pub fn arange(range: Range<usize>) -> Self {
         Tensor::new(B::arange(range, &B::Device::default()))
     }
 
-    /// Returns a new integer tensor on the default device which values are generated
-    /// from the given range with the specified step.
+    /// Returns a new integer tensor on the default device.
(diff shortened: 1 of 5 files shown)


burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -137,11 +137,36 @@ pub trait TensorOps<B: Backend> {
...
     fn arange(range: Range<usize>, device: &B::Device) -> B::IntTensorPrimitive<1> {
-        let shape = Shape::new([range.end - range.start]);
         let value = range
             .into_iter()
             .map(|i| (i as i64).elem())
             .collect::<Vec<B::IntElem>>();
+        let shape = Shape::new([value.len()]);
+        let data = Data::new(value, shape);
+        B::int_from_data(data, device)
+    }
```
nathanielsimard, line 147: The default implementation should call `arange_step` instead. Most backends could only implement `arange_step` for performance and totaly ignore `arange`.

burn-tensor/src/tensor/api/int.rs
```diff
@@ -9,6 +9,13 @@ where
     pub fn arange(range: Range<usize>) -> Self {
         Tensor::new(B::arange(range, &B::Device::default()))
     }
+
+    /// Returns a new integer tensor on the default device which values are generated
+    /// from the given range with the specified step.
+    pub fn arange_step(range: Range<usize>, step: usize) -> Self {
+        Tensor::new(B::arange_step(range, step, &B::Device::default()))
+    }
```
nathanielsimard, line 17: I would add `arange_step_device` as well.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #496, opened 2023-07-13
Install dataset downloader python deps in venv

### LINKED ISSUE #495: Dataset importer fails with externally managed python installation
**Describe the bug**
When running the MNIST example on Debian, the dataset importer fails, because it tries to install python packages user-wide using pip, which is no longer possible on Debian, as the python installation is declared as `EXTERNALLY-MANAGED`. (This isn't specific to the MNIST example, it's just what I tried first)

**To Reproduce**
 1. Run the MNIST example on a Debian machine with python3 (>= 3.11.1-3) like so: `cargo run --example mnist --release --features ndarray`

**Output**
```
USER@HOST:~/workspace/burn$ cargo run --example mnist --release --features ndarray 
    Finished release [optimized] target(s) in 0.15s
warning: the following packages contain code that will be rejected by a future version of Rust: nom v1.2.4
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `target/release/examples/mnist`
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
  File "/home/USER/.cache/burn-dataset/importer.py", line 3, in <module>
    import pyarrow as pa
ModuleNotFoundError: No module named 'pyarrow'
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SqliteDataset(ConnectionPool(Error(Some("unable to open database file: /home/USER/.cache/burn-dataset/mnist.db"))))', burn-dataset/src/source/huggingface/mnist.rs:86:14
```

**Desktop:**
 - OS: Debian
 - Version: unstable

**Additional Information:**
- https://salsa.debian.org/cpython-team/python3/-/blob/master/debian/README.venv
- https://peps.python.org/pep-0668/

dengelt:
### Checklist

### Related Issues/PRs

fixes #495 

### Changes

Install the python dependencies for the dataset importer in a venv in the same cache directory as the script, and run the python3 executable from there.

### Testing

I tested this on my Debian unstable machine. I have not tested it on other machines / OSs.

### DIFF
--- a/burn-dataset/src/source/huggingface/downloader.rs
+++ b/burn-dataset/src/source/huggingface/downloader.rs
@@ -158,9 +158,9 @@ fn import(
     huggingface_token: Option<String>,
     huggingface_cache_dir: Option<String>,
 ) -> Result<(), ImporterError> {
-    install_python_deps()?;
+    let venv_python_path = install_python_deps(&base_dir)?;
 
-    let mut command = Command::new(PYTHON);
+    let mut command = Command::new(venv_python_path);
 
     command.arg(importer_script_path(&base_dir));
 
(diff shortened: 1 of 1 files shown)


nathanielsimard (approved):
@antimora can you validate if it works on mac? Otherwise everything seems good!

antimora (approved):
I have tested on MacOS on M1 and it works as expected. 

Definitely a cleaner solution because the installed packages will not populate the system packages and it will go into venv directory. 

I didn't know about this venv, so here is a link for others to learn from https://docs.python.org/3/library/venv.html

So basically, as long as we reference `~/cache/burn-dataset/venv/bin/python3`, the site environment will be assumed to be there and no need to activate the environment from the shell.

PR #498, opened 2023-07-14
Move main file into the bin directory for onnx_inference example

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR moves the `main.rs` binary into the `bin` directory. Reasons:
- In this way `lib.rs` is not only a placeholder, but instead we show how library APIs are imported and used in a binary
- It is possible to add more models, for different neural networks, adding a new file into `src/bin` directory. Thus we improve the example modularity

### Testing

Run `cargo build` on the example


nathanielsimard:
I didn't know about the `bin` directory and how it's processed by `cargo`, very nice.

PR #504, opened 2023-07-18
Bugfix/matmul/asymetric shapes

louisfd:
### Checklist

### Changes

There were some cases in which matmul tiling 2d failed. Those are addressed by the two new matmul tests. 
The bug was in fact in padding, where, when only one of row and col dimensions was already "round", it would be rounded yet again to the higher multiple of the divisor. 

### Testing

New tests in padding and matmul


antimora (approved):
LGTM

PR #507, opened 2023-07-19
Add MaxPool2d to burn-import

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR adds `MaxPool2d` to `burn-import` crate. 

Just two questions:
- Is it safe to discriminate among `MaxPool` operators using the `node.name` attribute?
- It seems I cannot generate code with `padding` in the test because an error occurs. Could you help me in solving that problem?

Thanks! 

### Testing


burn-import/src/onnx/op_configuration.rs
```diff
@@ -81,6 +82,41 @@ pub fn conv2d_config(curr: &Node) -> Conv2dConfig {
...
+            "strides" => attr_value_vec_i64(value, &mut strides),
+            "pads" => attr_value_vec_i64(value, &mut pads),
+            _ => {}
+        }
+    }
+
+    let padding = if pads.iter().all(|&x| x == 0) {
+        MaxPool2dPaddingConfig::Valid
+    } else {
+        todo!("max_pool2d: padding({pads:?}) is not fully supported");
```
antimora, line 105: It'd be great if we could implement padding as well. The same logic could be used by the Conv2d, since they're the same.

burn-import/src/onnx/to_burn.rs
```diff
@@ -152,6 +154,13 @@ impl ONNXGraph {
         for node in self.nodes {
             match node.node_type {
                 NodeType::Conv2d => graph.register(Self::conv2d_conversion::<PS>(node)),
+                NodeType::MaxPool => {
+                    if node.name == "maxpool2" {
+                        graph.register(Self::max_pool2d_conversion(node))
+                    } else {
+                        panic!("MaxPool: {} operator not supported", node.name);
+                    }
+                }
```
antimora, line 163: We should not rely node.name. Instead we should introduce MaxPool2d node type, like it was done for Conv2d (ONNX does not have Conv2d, only Conv). There is somewhere logic that determines if it's 2d.

antimora (changes requested):
Awesome. We have our first ONNX contributor =D 

There are few changes I am requesting to make. Please see my comments. They're straightforward to fix. Padding is a bit more involved but doable, and it'd be a double win (we could use for Conv2d).

burn-import/src/onnx/op_configuration.rs
```diff
@@ -81,6 +82,41 @@ pub fn conv2d_config(curr: &Node) -> Conv2dConfig {
...
+    };
+
+    if strides.iter().all(|&x| x != 1) {
+        todo!("max_pool2d: strides({strides:?}) are not fully supported");
```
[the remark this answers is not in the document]
Luni-4: Sure! Should I simply remove this condition? Because it only checks that strides values are **all** equals to 1.

burn-import/src/onnx/op_configuration.rs
Luni-4: The main problem with `padding` is that we can have something like this as node: `[2, 2, 2, 2]` and looking at `Conv2dPaddingConfig` I do not know which enum field I should choose for this case. [Here](https://github.com/burn-rs/burn/blob/main/burn-core/src/nn/conv/conv2d.rs#L43) I should choose `Explicit` for this case, but I have only two values and not four. And whens should I choose the `Same` field?

burn-import/src/onnx/op_configuration.rs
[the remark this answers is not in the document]
nathanielsimard: What you need to do is use the attributes `strides`, `padding`, `groups` and `dilations` and add them to the `Conv2dConfig`. I don't think you need to change the codegeneration part, since it's based on the config.

burn-import/src/onnx/op_configuration.rs
Luni-4: I put a `panic!("{:?}", node)` instruction for the `MaxPool` type and I got this:

```
Node { node_type: MaxPool, name: "/model.9/m/MaxPool", inputs: [Argument { name: "/model.9/cv1/act/Mul_output_0", ty: Tensor(TensorArg { dim: 0 }) }], outputs: [Argument { name: "/model.9/m/MaxPool_output_0", ty: Tensor(TensorArg { dim: 0 }) }], states: [], attrs: {"ceil_mode": Int64(0), "pads": Int64s([2, 2, 2, 2]), "strides": Int64s([1, 1]), "kernel_shape": Int64s([5, 5])} }', burn-import/src/onnx/from_onnx.rs:384:13
```

As you can see `"pads": Int64s([2, 2, 2, 2])` has four values

antimora:
CI test is failing: 

```
   Compiling mnist-inference-web v0.8.0 (/home/runner/work/burn/burn/examples/mnist-inference-web)
error[E0432]: unresolved import `crate::burn::node::pool2d`
  --> burn-import/src/burn/node/max_pool2d.rs:92:16
   |
92 |         node::{pool2d::MaxPool2dNode, test::assert_tokens},
   |                ^^^^^^ could not find `pool2d` in `node`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `burn-import` (lib test) due to previous error
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.

```

I am not sure why `run-checks.sh` script did not catch it. I ran it locally and it seems to be working fine.

Luni-4:
> CI test is failing:
> 
> ```
>    Compiling mnist-inference-web v0.8.0 (/home/runner/work/burn/burn/examples/mnist-inference-web)
> error[E0432]: unresolved import `crate::burn::node::pool2d`
>   --> burn-import/src/burn/node/max_pool2d.rs:92:16
>    |
> 92 |         node::{pool2d::MaxPool2dNode, test::assert_tokens},
>    |                ^^^^^^ could not find `pool2d` in `node`
> 
> For more information about this error, try `rustc --explain E0432`.
> error: could not compile `burn-import` (lib test) due to previous error
> warning: build failed, waiting for other jobs to finish...
> Error: Process completed with exit code 101.
> ```
> 
> I am not sure why `run-checks.sh` script did not catch it. I ran it locally and it seems to be working fine.

Sorry, my bad. It should be fixed now

burn-import/src/onnx/op_configuration.rs
antimora: > Yeah you can use explicit, but it's strange that you have more than 2 values for a 2d convolution 😅.

Basically they're left, up, right, and bottom (need to verify which is which). Here is the full document: https://onnx.ai/onnx/operators/onnx__Pad.html

Burn and [Pytorch's padding ](https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html) does support the full padding like in ONNX. Most likely all exported ONNX padding can be supported by Burn's current implementation.

Therefore, we should only extract values we are interested for and implement `ToTokens` for MaxPool2dPaddingConfig in `burn-import/src/burn/codegen.rs`. If we encounter, asymmetric top/bottom or left/right, then we will same we do not support it. Basically, we will support examples, such as `[2, 2, 2, 2]` or [2, 2, 3, 3], which will be equivalent as `Conv2dPaddingConfig::Explicit(2,2)` and `Conv2dPaddingConfig::Explicit(2, 3). 
`
@Luni-4, I hope this makes sense. I am standby if you need further help. If you could, tackle this conversion, the same can be used for Conv padding in next PRs easily - win/win =D

antimora (changes requested):

nathanielsimard (approved):

nathanielsimard:
@Luni-4 Awesome stuff, thanks a lot for contributing! 🎉

antimora (approved):
Looks great. Thank you for your contribution.

PR #509, opened 2023-07-19
Feat/wgpu/matmul transpose

louisfd:
### Checklist

### Changes

Allow efficient matmul for transposed tensors in wgpu. 
In tiling kernels, use provided strides instead of hardcoded ones
Call into_continuous only when no padding was made and batch has been swapped with row or col. 
Done for all 

### Testing

Three new tests for matmul: 
swap two batches
swap row and col
swap batch with row or col
All in a situation with no padding.


burn-wgpu/src/kernel/matmul/utils.rs
```diff
@@ -45,4 +45,35 @@ pub(crate) mod tests {
...
+            .matmul(y.swap_dims(swap_rhs[0], swap_rhs[1]));
+
+        let z = func(
+            x_wgpu.swap_dims(swap_lhs[0], swap_lhs[1]).into_primitive(),
+            y_wgpu.swap_dims(swap_rhs[0], swap_rhs[1]).into_primitive(),
+        );
+        let z = TestTensor::from_primitive(z);
+
+        std::println!("{z}");
+        std::println!("{z_reference}");
```
antimora, line 76: println statements are not removed.

burn-wgpu/src/template/matmul/blocktiling_2d/continuous.wgsl
```diff
@@ -79,18 +87,21 @@ fn main(
             let lhs_sm_position = thread_offset + load_index;
             let block_row = lhs_sm_position / B_K;
             let block_col = lhs_sm_position % B_K;
-            let lhs_position = offset_lhs + k + block_row * K + block_col;
+            let lhs_position = offset_lhs + block_row * lhs_stride_row + (k + block_col) * lhs_stride_col;
 
             if block_row < B_M {
                 shared_lhs[lhs_sm_position] = lhs[lhs_position];
+            } else {
+                // Patch for mac os bugfix
```
antimora, line 95: It'd be good to mention the name of the bug or behavior for the future maintainers to revert the fix if bug is no longer relevant. Do we have an external reference or internal?

antimora (changes requested):

burn-wgpu/src/ops/float_ops.rs
```diff
@@ -136,9 +136,6 @@ where
         lhs: FloatTensor<Self, D>,
         rhs: FloatTensor<Self, D>,
     ) -> FloatTensor<Self, D> {
-        let lhs = kernel::into_continuous(lhs);
-        let rhs = kernel::into_continuous(rhs);
-
```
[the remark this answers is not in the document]
louisfd: Nope! I will fix it

nathanielsimard (approved):

PR #513, opened 2023-07-20
Feat/wgpu/default device

nathanielsimard:


burn-wgpu/src/device.rs
```diff
@@ -26,14 +26,25 @@ pub enum WgpuDevice {
...
+    ///   1. DiscreteGpu
+    ///   2. IntegratedGpu
+    ///   3. VirtualGpu
+    ///   4. Cpu
+    ///
+    /// # Notes
+    ///
+    /// A device might be identified as [Other](wgpu::DeviceType::Other) by [wgpu](wgpu), in this case, we chose this device over
+    /// `IntegratedGpu` since it's often a discrete GPU.
+    MostPerformant,
```
antimora, line 43: We could also name it as `BestAvailable` to convey a couple of notions: Best (subjectively best) and Available (currently available on the system). I also was thinking to name it `Auto` but 'Auto` is too enigmatic and it sounds like it was automatically/dynamically selected during runtime.

antimora (commented):
See my one comment about the naming.

burn-wgpu/src/device.rs
nathanielsimard: Yes I prefer `BestAvailable` as well!

antimora (approved):

PR #514, opened 2023-07-20
Example/wgpu/mnist

louisfd:
MNIST example with wgpu


antimora (changes requested):
Could we also update `README.md`?

nathanielsimard (approved):

antimora (approved):

PR #517, opened 2023-07-22
Allow for more flexible Data creation via From by adding traits that have the first dimension as non-const

miestrode:
### Checklist

### Related Issues/PRs

The only relevant PR I can think of is #457, which generalizes on the work done here. However because the new traits added here aren't fallible I believe this would come out as a more welcome change.

### Changes

Sometimes one would like to for example, construct tensors from multidimensional data structures that don't have a const first dimension. Having a const first dimension isn't technically a requirement but yet only providing traits for non-const dimensions means losing the convenience of operating on `[[1, 2], [3, 4]` for example, as this isn't a borrow.

This PR seeks to add the support to do this for such use-cases without compromising others - this should be a SemVer minor change.

### Testing

Due to this PR only adding new traits not much testing was done besides using `run-checks.sh`.

### DIFF
--- a/burn-tensor/src/tensor/data.rs
+++ b/burn-tensor/src/tensor/data.rs
@@ -385,59 +385,57 @@ impl<E: core::fmt::Debug + Copy> From<&[E]> for Data<E, 1> {
     }
 }
 
-impl<E: core::fmt::Debug + Copy, const A: usize, const B: usize> From<[[E; B]; A]> for Data<E, 2> {
-    fn from(elems: [[E; B]; A]) -> Self {
-        let mut data = Vec::with_capacity(A * B);
-        for elem in elems.into_iter().take(A) {
-            for elem in elem.into_iter().take(B) {
-                data.push(elem);
+impl<E: core::fmt::Debug + Copy, const A: usize> From<&[[E; A]]> for Data<E, 2> {
+    fn from(elems: &[[E; A]]) -> Self {
+        let first_dim = elems.len();
+        let mut data = Vec::with_capacity(first_dim * A);
+        for elem in elems {
+            for elem in elem.iter().take(A) {
+                data.push(*elem);
             }
         }
 
-        Data::new(data, Shape::new([A, B]))
+        Data::new(data, Shape::new([first_dim, A]))
     }
 }
 
-impl<E: core::fmt::Debug + Copy, const A: usize, const B: usize, const C: usize>
-    From<[[[E; C]; B]; A]> for Data<E, 3>
+impl<E: core::fmt::Debug + Copy, const A: usize, const B: usize> From<&[[[E; B]; A]]>
+    for Data<E, 3>
 {
-    fn from(elems: [[[E; C]; B]; A]) -> Self {
-        let mut data = Vec::with_capacity(A * B * C);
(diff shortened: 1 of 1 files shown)


antimora:
I am not clear on the problem this PR is solving. It looks the existing code was duplicated with a small change. If you're attempting to avoid cloning the nested array, then it should be noted that the individual elements are cloned anyway. So I don't see a reason not to clone the nested array and use the existing the conversation From logic. Data will have to be cloned when creating a tensor.

Can you please elaborate what you meant by  "first dimension as non-const"?  

Also for the future, for any new logic we will need to see if it's correctly implemented by verifying with unit tests. This PR is missing unit tests and the description has stated it only adding Trait but actually they're implications. The tests should go here: 
https://github.com/burn-rs/burn/blob/9aca1837c205feff85b0df457fb8e2f7aa507134/burn-tensor/src/tensor/data.rs#L451

miestrode:
Yes, the exist code had a very small change added to it, kind of by design.
In summary I seek to add new implementations for `From` that allow you to pass a data structure whose first dimension is not known at compile time. From what I could discern from the API the only `impl` allowing you to do this is for `&[E]`. This is however not all that useful for constructing for example multidimensional tensors.

This is because all of the implementations of `From` working with multidimensional arrays have the first dimension of the array specified (because they are all of the form `[OtherMultidimensionalArray; N]`). I want to allow to do this for `&[OtherMultidimensionalArray`, because from what I can tell there isn't a convenient way to do this with the current API. I don't really see what this has to do with cloning.

Also, sorry for not writing any unit tests, I totally forgot to do that and will get to doing so soon™.

antimora:
I understand what you're getting at with the first dimension. It's essentially a slice of nested arrays.

Sorry for the confusion about the cloning comment. I overlooked the missing array count. Initially, I thought it was a reference to an array, but it's indeed a slice of arrays.

While I agree this utility can be helpful, the types seem quite specific, for instance, &[[[E; B]; A]] corresponds to a 3D array where the first dimension is determined at runtime. I believe it would be more beneficial to introduce a more generic solution within the framework, rather than attempting to cover all distinct combinations (up to 4D). As a case in point, recently PR #457 was submitted, which offers fully nested slices.

Since this is not the first time it has come up, I would like to invite @nathanielsimard and @louisfd to provide their input. We should decide how much conversion from different types we want to support.

miestrode:
I see. If my comment on this helps, I specifically opened up this because the conversions suggested by #457 could fail. The conversions are really a `TryFrom`, instead of a `From`. The nice thing about making these special cases is precisely that they shouldn't really fail.

Also, I should probably say I am also kind of more hoping for a general solution in comparison to the specific `impl`s the project currently has.

nathanielsimard:
@mestrode @antimora The `From` implementations are really for convenience, users can use the `Data::new` function to convert their data structure to Burn's `Data`. With that in mind, I think adding more `From` is good since we can see it as integrating with more data structures, which helps people convert their stuff to Burn. I don't think we need another level of abstraction for that, but we may use macros internally to reduce code duplication.

antimora:
Closing this PR to come up with a better way to handle nested array conversion. This ticket will track this work: https://github.com/burn-rs/burn/issues/551

PR #520, opened 2023-07-23
patch tanh bug on mac os

louisfd:
### Checklist

### Changes

Training mnist on wgpu always ended up with NaNs (on Mac OS only).  
The culprit is the tanh function, which on metal seems to have numerical errors with large values. 
With inputs within 43.67..44.36 it gives 0 which is wrong, and with inputs above 44.36 it gives NaN. 
I patched it by ceiling the input of tanh to 40. 

### Testing

New test for tanh to make sure it does not give NaNs or zeros for aforementioned values.


burn-wgpu/src/ops/float_ops.rs
```diff
@@ -392,6 +393,12 @@ where
     }
 
     fn tanh<const D: usize>(tensor: FloatTensor<Self, D>) -> FloatTensor<Self, D> {
+        let tensor = mask_fill(
+            tensor.clone(),
+            greater_elem(tensor, F::from_elem(40.0)),
+            F::from_elem(40.0),
+        );
+
```
antimora, line 401: Funny, I was about to work on a new [`clamp` operator](https://pytorch.org/docs/stable/generated/torch.clamp.html) that I needed to use for MobileNet. 

Do you mind adding a todo that instructs to replace with `clamp`?  And we need a note why we are doing it (basically the Metal has problems with over 40.0. BTW, why 40 if it fails at 43?

antimora (changes requested):

burn-wgpu/src/ops/float_ops.rs
louisfd: Done and done. 
I had just put 40 because it's a more round number. Anyway tanh outputs 1.0 for everything above ~12. 
But 43 makes more sense

nathanielsimard (approved):

antimora:
I ran the `run-checks.sh` locally and getting the same CI error:

```
   Doc-tests text-generation

running 0 tests

Diff in /Users/dilshod/Projects/burn/burn-wgpu/src/ops/float_ops.rs at line 1:
 use super::{numeric, BoolTensor, Device, FloatElem, FloatTensor, FullPrecisionBackend, IntTensor};
 use crate::kernel::{
-    self, unary_default, unary_inplace_default, unary_scalar_default,
-    unary_scalar_inplace_default,
+    self, unary_default, unary_inplace_default, unary_scalar_default, unary_scalar_inplace_default,
 };
 
 use crate::unary_scalar_inplace;
[burn]$ 
```

antimora (approved):
Looks good.

PR #524, opened 2023-07-24
Implement `Add` operator for Onnx

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR implements the `Add` operator for onnx

### Testing

### DIFF
--- a/burn-import/src/burn/node/add.rs
+++ b/burn-import/src/burn/node/add.rs
@@ -1,0 +1,83 @@
+use super::{Node, NodeCodegen};
+use crate::burn::{Scope, TensorType, Type};
+use burn::record::PrecisionSettings;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+#[derive(Debug, Clone, new)]
+pub struct AddNode {
+    pub lhs: TensorType,
+    pub rhs: TensorType,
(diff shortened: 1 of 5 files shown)


burn-import/src/burn/node/add.rs
```diff
@@ -0,0 +1,83 @@
...
+use burn::record::PrecisionSettings;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+#[derive(Debug, Clone, new)]
+pub struct AddNode {
+    pub lhs: TensorType,
+    pub rhs: TensorType,
+    pub output: TensorType,
+}
```
nathanielsimard, line 12: This will be very similar for all unary operations. I think it would be interesting to have a way to avoid code duplication with a trait that only defines the forward method. It may be done in another PR.

Luni-4:
@nathanielsimard 

My concern is that my code does not consider adding a scalar to a tensor at all. I do not know how to discriminate that case

burn-import/src/burn/node/add.rs
Luni-4: Yes, that could be an idea. I agree

antimora (approved):
Looks good.

PR #525, opened 2023-07-24
Add disk clean up step for CI tests

antimora:
### Changes

Currently CI tests are failing because of this error: `System.IO.IOException: No space left on device : `

### Testing

Not tested locally except to check for yaml formatting.


Nikaidou-Shinku:
I think I've run into this problem again: <https://github.com/tracel-ai/burn/actions/runs/8001779613>

PR #528, opened 2023-07-24
Add a `transpose` sugar: `tensor^T`

wbrickner:
### Checklist

`run-checks.rs` fails but with `burn-import` errors (I did not have anything to do with this).

### Related Issues/PRs
- https://github.com/burn-rs/burn/pull/509

### Changes

- zero-cost syntax sugar so we can write beautiful things like `tensor^T` to transpose a tensor.  This imitates the math notation for transposes and is a common pattern in tensor / linear algebra libraries.
- This defines a public zero-size-type `T`, for which we implement `core::ops::BitXor` to use the `^` caret operator.
- Does not interfere with any other present or future XOR ops or caret operator implementations.

### Testing

- it compiles, has no implementation to test, it defers to `self.transpose()`.

### DIFF
--- a/burn-tensor/src/tensor/api/base.rs
+++ b/burn-tensor/src/tensor/api/base.rs
@@ -405,6 +405,23 @@ where
     }
 }
 
+/// Transpose marker (zero-size type). Used to sugar the transpose of a tensor, e.g.
+/// ```rust
+/// use burn_tensor::backend::Backend;
+/// use burn_tensor::{Tensor, T};
+///
+/// fn example<B: Backend>() {
+///     let tensor = Tensor::<B, 2>::from_floats([[1.0, 2.0], [3.0, 4.0]]);
+///     let transposed = tensor^T;
+/// }
+/// ```
+pub struct T;
+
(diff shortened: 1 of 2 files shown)


antimora:
I ran `run-checks.sh` locally and I am getting this doc error which is also present on CI: 

```
   Doc-tests text-generation

running 0 tests

Diff in /Users/dilshod/Projects/burn/burn-tensor/src/tensor/api/base.rs at line 419:
 
 impl<B: Backend, const D: usize> core::ops::BitXor<T> for Tensor<B, D> {
     type Output = Self;
-    fn bitxor(self, _: T) -> Self::Output { self.transpose() }
+    fn bitxor(self, _: T) -> Self::Output {
+        self.transpose()
+    }
 }
 
 /// Trait that list all operations that can be applied on all tensors.
```

If it's due to burn-import, can you please update from the main (since then new commits were merged)?

antimora (approved):
Looks good, although I am not sure how useful this is =P

I am tagging @nathanielsimard

wbrickner:
not useful, beautiful! 😀

nathanielsimard (approved):

PR #529, opened 2023-07-24
burn-core: Use a specific structure for 1d padding

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

Use a new `PaddingConfig1d` struct shared with many operators. In this way we have a single implementation in a specific place without defining more types. It is possible to add new functions to this struct in order to support different padding implementation for new types of operator

### Testing


antimora (approved):
Looks good. Thanks for improving the core.

nathanielsimard (approved):

Issue #487, opened 2023-07-11
 Simplify tensor casting between different kinds


antimora:
**Feature Description**
We propose that we simplify the process of casting tensors between different kinds. Currently, converting a tensor of integer kind to a tensor of float kind is not intuitive and feels clunky. 

**Current Behavior**
Here's an example of what needs to be done currently to perform this casting:

```rust
use burn_ndarray::NdArrayBackend;
use burn::tensor::{Tensor, Data, Int};

type B = NdArrayBackend<f32>;

// create an int tensor with [0, 1, 2, 3, 4, 5]
let tensor_int = Tensor::<B, 1, Int>::arange(0..6);

// convert int tensor to float tensor
let tensor_float:Tensor::<B, 1> = Tensor::from_data(
    tensor_int
        .into_data()
        .convert()
);

println!("{}", tensor_float);
```

**Desired Behavior**
Ideally, there should be a simpler, more intuitive way to perform this kind of casting. This would improve the usability of the library and potentially reduce errors or confusion. Perhaps, the following API example could be used:

```rust
use burn_ndarray::NdArrayBackend;
use burn::tensor::{Tensor, Data, Int};

type B = NdArrayBackend<f32>;

// create an int tensor with [0, 1, 2, 3, 4, 5]
let tensor_int = Tensor::<B, 1, Int>::arange(0..6);

// convert int tensor to float tensor
let tensor_float = tensor_int.to_float();

println!("{}", tensor_float);
```

nathanielsimard:
I'm not sure about `to_float` since it can be ambiguous with `from_floats`, `from_bools` and `from_ints` that create tensors from data. An idea API would look like this for me.

```rust
use burn_ndarray::NdArrayBackend;
use burn::tensor::{Tensor, Data, Int};

type B = NdArrayBackend<f32>;

// create an int tensor with [0, 1, 2, 3, 4, 5]
let tensor_int = Tensor::<B, 1, Int>::arange(0..6);

// convert int tensor to float tensor
let tensor_float = tensor_int.cast::<Float>();

println!("{}", tensor_float);
``` 

I think this is doable without having to import any trait using associative types. (Already present in the `TensorKind` trait).
@antimora what are your thoughts?

antimora:
> I'm not sure about `to_float` since it can be ambiguous with `from_floats`, `from_bools` and `from_ints` that create tensors from data. An idea API would look like this for me.
> 
> ```rust
> use burn_ndarray::NdArrayBackend;
[...]

PR #530, opened 2023-07-25
Add functions to cast tensor to different kinds

trfdeer:
### Checklist

### Related Issues/PRs

Related Issue: #487 

### Changes

1. Added `.int()` method to convert float tensors to integer kind.
2. Added `.float()` method to convert int tensors to float kind.

### Testing

Tested by manually creating and converting tensors.

### DIFF
--- a/burn-tensor/src/tensor/api/float.rs
+++ b/burn-tensor/src/tensor/api/float.rs
@@ -118,8 +118,8 @@ where
     ///     let int_tensor = float_tensor.int();
     /// }
     /// ```
-    pub fn int(&self) -> Tensor<B, D, Int> {
-        Tensor::<B, D, Int>::from_data(self.to_data().convert())
+    pub fn int(self) -> Tensor<B, D, Int> {
+        Tensor::<B, D, Int>::from_data(self.into_data().convert())
     }
 
     /// Returns a new tensor with the same shape and device as the current tensor filled with zeros.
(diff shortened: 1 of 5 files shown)


antimora:
Great. Thank you. We will need some unit tests because we can accept the PR. 

Please see this recent PR for how to add a test: https://github.com/burn-rs/burn/pull/486/files

There are 3 parts: 

1. add test under tests/ops. Make sure the file name is `cast` and the macro on the top file includes the name: like this `#[burn_tensor_testgen::testgen(cast)]`
2. add the auto generated test here: burn-tensor/src/tests/mod.rs
3. expose the module here: burn-tensor/src/tests/ops/mod.rs

To test, you'll need to cd to burn-ndarray and run `cargo test`. The test will use burn-ndarray backend. `run-checks.sh` will test on all backends.

burn-tensor/src/tensor/api/float.rs
```diff
@@ -103,6 +104,24 @@ where
...
+    /// use burn_tensor::Tensor;
+    ///
+    /// fn example<B: Backend>() {
+    ///     let float_tensor = Tensor::<B, 1>::from_floats([1.0, 2.0]);
+    ///     let int_tensor = float_tensor.int();
+    /// }
+    /// ```
+    pub fn int(&self) -> Tensor<B, D, Int> {
+        Tensor::<B, D, Int>::from_data(self.to_data().convert())
+    }
```
nathanielsimard, line 123: I would take ownership here and call `into_data`. In a following PR we should call a backend function where each backend can decide how they do the conversion. For instance, in this current implementation, the wgpu backend would need to allocate at least 3 new buffers, where 0 or 1 could have been allocated with proper casting.

nathanielsimard (changes requested):
A little change to the API before we can merge it. Thanks!

antimora (approved):
Looks good. Thank you for your contribution.

nathanielsimard (approved):

PR #545, opened 2023-07-26
Add `Reshape` operator to `burn-import`

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR implements the `Reshape` operator and refactors the `burn-import` code a bit

### Testing

### DIFF
--- a/burn-import/src/burn/codegen.rs
+++ b/burn-import/src/burn/codegen.rs
@@ -3,32 +3,55 @@ use quote::quote;
 
 use burn::nn::PaddingConfig2d;
 
+fn convert_primitive<T: ToString>(primitive: T) -> TokenStream {
+    let value = primitive.to_string();
+    value.parse().unwrap()
+}
+
+fn convert_lists<'a, I, T: ToTokens>(list: I) -> TokenStream
+where
+    I: Iterator<Item = &'a T>,
+    T: 'a,
+{
+    let mut body = quote! {};
+
+    list.for_each(|item| {
+        let elem = item.to_tokens();
+        body.extend(quote! {#elem,});
+    });
+
+    quote! {
+        [#body]
+    }
+}
+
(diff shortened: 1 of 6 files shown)


burn-import/src/burn/codegen.rs
```diff
@@ -3,32 +3,55 @@ use quote::quote;
 
 use burn::nn::PaddingConfig2d;
 
+fn convert_primitive<T: ToString>(primitive: T) -> TokenStream {
+    let value = primitive.to_string();
+    value.parse().unwrap()
+}
+
+fn convert_lists<'a, I, T: ToTokens>(list: I) -> TokenStream
```
nathanielsimard, line 11: `convert_to_array` might be a better name ?

burn-import/src/burn/node/reshape.rs
```diff
@@ -0,0 +1,85 @@
...
+use burn::record::PrecisionSettings;
+use burn::tensor::DataSerialize;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+#[derive(Debug, Clone, new)]
+pub struct ReshapeNode {
+    pub input: TensorType,
+    pub output: TensorType,
+    pub shape: DataSerialize<i64>,
```
nathanielsimard, line 12: You should use `Vec<usize>` here, `DataSerialize` is most appropriate for tensors, not shapes. Maybe you'll have to use the `convert_to_array` function.

nathanielsimard (changes requested):

burn-import/src/burn/node/reshape.rs
Luni-4: Since `Onnx` format uses `i64` values, wouldn't `Vec<i64>` be a better choice? `usize` depends too much on the architecture

nathanielsimard (approved):

PR #547, opened 2023-07-26
Refactor test code and some missing operators test in `burn-import`

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR deduplicates some test code for `burn-import` crate 

### Testing

Run `cargo test`

### DIFF
--- a/burn-import/src/burn/node/base.rs
+++ b/burn-import/src/burn/node/base.rs
@@ -180,6 +180,7 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for Node<PS> {
 #[cfg(test)]
 pub(crate) mod tests {
     use crate::burn::{
+        codegen::ToTokens,
         graph::BurnGraph,
         node::{conv2d::Conv2dNode, matmul::MatmulNode, test::assert_tokens, NodeCodegen},
         TensorType,
@@ -200,7 +201,8 @@ pub(crate) mod tests {
         graph
     }
 
-    fn unary_operator_expected(function: TokenStream) -> TokenStream {
+    fn unary_operator_expected<const N: usize>(function: TokenStream) -> TokenStream {
+        let tensor_dim = N.to_tokens();
         quote! {
             use burn::{
                 module::Module,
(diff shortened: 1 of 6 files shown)


burn-import/src/burn/node/base.rs
```diff
@@ -178,17 +178,60 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for Node<PS> {
...
+
+            #[derive(Module, Debug)]
+            pub struct Model <B: Backend>{}
+
+            impl<B: Backend> Model <B> {
+                pub fn new_with(record: ModelRecord<B>) -> Self {
+                    Self { }
+                }
+                #[allow(clippy::let_and_return)]
+                pub fn forward(&self, tensor1: Tensor<B, 4>) -> Tensor<B, 4> {
```
nathanielsimard, line 218: I would use a const generic instead of 4 here. `unary_operator_expected<const D: usize>`

nathanielsimard (changes requested):
Just a small change before being ready :)

burn-import/src/burn/node/base.rs
Luni-4: Unfortunately if we use const generics, tests do not pass:

```
---- burn::node::sigmoid::tests::test_codegen_node stdout ----
thread 'burn::node::sigmoid::tests::test_codegen_node' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
 use burn::{
     module::Module,
     tensor::{backend::Backend, Tensor},
 };
 #[derive(Module, Debug)]
 pub struct Model<B: Backend> {}
 impl<B: Backend> Model<B> {
     pub fn new_with(record: ModelRecord<B>) -> Self {
         Self {}
     }
     #[allow(clippy::let_and_return)]
<    pub fn forward(&self, tensor1: Tensor<B, 4>) -> Tensor<B, 4> {
>    pub fn forward(&self, tensor1: Tensor<B, 4usize>) -> Tensor<B, 4usize> {
         let tensor2 = tensor1.sigmoid();
         tensor2
     }
 }
```

nathanielsimard (approved):

PR #548, opened 2023-07-26
refactor: burn-import unary operators

nathanielsimard:
Change how the unary operations are coded in `burn-import` to reduce code duplication.


burn-import/src/burn/node/base.rs
```diff
@@ -114,20 +108,17 @@ impl<PS: PrecisionSettings> Serialize for Node<PS> {
...
-            Node::Sigmoid(_) => "sigmoid",
+            Node::Add(_) => "add".to_string(),
+            Node::Matmul(_) => "matmul".to_string(),
+            Node::Constant(_) => "constant".to_string(),
+            Node::Conv2d(_) => "conv2d".to_string(),
+            Node::MaxPool2d(_) => "max_pool2d".to_string(),
+            Node::Linear(_) => "linear".to_string(),
+            Node::BatchNorm(_) => "batch_norm".to_string(),
+            Node::Equal(_) => "equal".to_string(),
+            Node::Unary(unary) => unary.name.to_string(),
```
Luni-4, line 121: Instead of `String`, can we define an `enum` for all `unary` operators? In this specific case we would have a double `match`, the second one after `Node::unary(node)`

burn-import/src/burn/node/unary.rs
```diff
@@ -0,0 +1,85 @@
...
+
+// Simple fn pointer that receive input as a token stream and return function call.
+type FnPointer = Arc<dyn Fn(TokenStream) -> TokenStream>;
+
+/// Node for all unary operators.
+#[derive(Clone, new)]
+pub struct UnaryNode {
+    pub input: TensorType,
+    pub output: TensorType,
+    pub name: String,
```
Luni-4, line 16: Can we define an enum instead of using a String?

Luni-4 (changes requested):
Looks good to me! Thank you! 

Just two comments

burn-import/src/burn/node/base.rs
nathanielsimard: This is only used in logging, so the important part is having something clean, we already have an enum, but the `Default` formating isn't pretty enough for printing. Maybe this should be refactoring into a `Display` function instead of a "name".

Luni-4: Yeah, that could be an idea

antimora: @nathanielsimard @Luni-4 

There is a [`strum_macros::Display`
](https://docs.rs/strum_macros/0.25.1/strum_macros/derive.Display.html) derive that you can use. You can also specify serialization types, see https://docs.rs/strum/latest/strum/additional_attributes/index.html. We will basically use this:

```rust
#[derive(Debug, strum_macros::Display)]
#[strum(serialize_all = "snake_case")]
```

We are using strum in burn-import.

burn-import/src/burn/node/unary.rs
antimora: Sounds like a good idea.

burn-import/src/burn/node/base.rs
Luni-4: @antimora  @nathanielsimard 

For now we can use this approach since it is already there, but I think we should try to use the less amount of dependencies as possible in my opinion. (when it's possible obviously)

antimora: I agree but we are already using this dependency in burn-import.

Luni-4 (approved):

PR #550, opened 2023-07-26
Add clamp, clamp_min, clamp_max tensor ops

antimora:
Note: this commit still missing int tensor implementation/tests and backward pass. This submitted for @nathanielsimard review/help

### Checklist

### Related Issues/PRs

### Changes

To be written here ...

### Testing

Added unit tests but it's missing int test tests

### DIFF
--- a/burn-autodiff/src/ops/tensor.rs
+++ b/burn-autodiff/src/ops/tensor.rs
@@ -1372,6 +1372,28 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
             }
         }
     }
+
+    fn clamp_min<const D: usize>(
+        _tensor: <ADBackendDecorator<B> as Backend>::TensorPrimitive<D>,
+        _min: <ADBackendDecorator<B> as Backend>::FloatElem,
+    ) -> <ADBackendDecorator<B> as Backend>::TensorPrimitive<D> {
+        todo!()
+    }
+
+    fn clamp_max<const D: usize>(
+        _tensor: <ADBackendDecorator<B> as Backend>::TensorPrimitive<D>,
+        _max: <ADBackendDecorator<B> as Backend>::FloatElem,
+    ) -> <ADBackendDecorator<B> as Backend>::TensorPrimitive<D> {
+        todo!()
+    }
+
+    fn clamp<const D: usize>(
+        _tensor: <ADBackendDecorator<B> as Backend>::TensorPrimitive<D>,
+        _min: <ADBackendDecorator<B> as Backend>::FloatElem,
+        _max: <ADBackendDecorator<B> as Backend>::FloatElem,
+    ) -> <ADBackendDecorator<B> as Backend>::TensorPrimitive<D> {
+        todo!()
+    }
 }
 
 /// Make sure the grad tensor has the given shape.
--- a/burn-autodiff/src/ops/tensor.rs
+++ b/burn-autodiff/src/ops/tensor.rs
@@ -1372,28 +1372,6 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
             }
         }
     }
-
-    fn clamp_min<const D: usize>(
-        _tensor: <ADBackendDecorator<B> as Backend>::TensorPrimitive<D>,
-        _min: <ADBackendDecorator<B> as Backend>::FloatElem,
-    ) -> <ADBackendDecorator<B> as Backend>::TensorPrimitive<D> {
-        todo!()
-    }
-
-    fn clamp_max<const D: usize>(
-        _tensor: <ADBackendDecorator<B> as Backend>::TensorPrimitive<D>,
-        _max: <ADBackendDecorator<B> as Backend>::FloatElem,
-    ) -> <ADBackendDecorator<B> as Backend>::TensorPrimitive<D> {
-        todo!()
-    }
-
-    fn clamp<const D: usize>(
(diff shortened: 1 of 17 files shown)


burn-tch/src/ops/tensor.rs
```diff
@@ -409,4 +409,42 @@ impl<E: TchElement> TensorOps<TchBackend<E>> for TchBackend<E> {
...
+        max: <TchBackend<E> as Backend>::FloatElem,
+    ) -> <TchBackend<E> as Backend>::TensorPrimitive<D> {
+        let min = min.elem::<f64>();
+        let max = max.elem::<f64>();
+
+        tensor.unary_ops(
+            |mut tensor| tensor.clamp_(min, max),
+            |tensor| tensor.clamp(min, max),
+        )
+    }
```
nathanielsimard, line 449: You should use `TchOps` here as well, but modify the `TchOps` clamp methods with the inplace operations that you have here (Better implementation)!

burn-tensor/src/tensor/api/float.rs
```diff
@@ -250,6 +250,49 @@ where
...
+    /// * `tensor` - The tensor to clamp.
+    /// * `max` - The maximum value.
+    ///
+    /// # Returns
+    ///
+    /// A new tensor with the values clamped over the given max value.
+    ///
+    pub fn clamp_max(self, max: B::FloatElem) -> Self {
+        Self::new(B::clamp_max(self.primitive, max))
+    }
```
nathanielsimard, line 295: This should go into numeric, it's not a float specific operation! Instead of using `B::clamp` you should use the Numeric trait on the Kind: `K::clamp`.

burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -1336,6 +1336,49 @@ where
...
+    ///
+    /// * `tensor` - The tensor to clamp.
+    /// * `max` - The maximum value.
+    ///
+    /// # Returns
+    ///
+    /// A new tensor with the values clamped over the given max value.
+    ///
+    fn clamp_max<const D: usize>(tensor: Self::Primitive<D>, max: Self::Elem)
+        -> Self::Primitive<D>;
```
nathanielsimard, line 1381: Update the doc to indicate that it's an internal trait.

burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -472,6 +472,61 @@ pub trait IntTensorOps<B: Backend> {
...
+    /// # Returns
+    ///
+    /// The clamped tensor.
+    fn int_clamp<const D: usize>(
+        tensor: B::IntTensorPrimitive<D>,
+        min: B::IntElem,
+        max: B::IntElem,
+    ) -> B::IntTensorPrimitive<D> {
+        Self::int_clamp_max(Self::int_clamp_min(tensor, max), min)
+    }
```
nathanielsimard, line 528: This should be removed.

burn-autodiff/src/ops/tensor.rs
```diff
@@ -1372,6 +1372,28 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
...
+        todo!()
+    }
+
+    fn clamp<const D: usize>(
+        _tensor: <ADBackendDecorator<B> as Backend>::TensorPrimitive<D>,
+        _min: <ADBackendDecorator<B> as Backend>::FloatElem,
+        _max: <ADBackendDecorator<B> as Backend>::FloatElem,
+    ) -> <ADBackendDecorator<B> as Backend>::TensorPrimitive<D> {
+        todo!()
+    }
```
nathanielsimard, line 1396: If you look at the default implementation, the gradient should be zero when its value was clamped during the forward pass. So there are two possible implementations:

1. Don't override the operation in the burn-autodiff backend, so the backward pass will be calculated with the `mask_fill` operation.
2. Override it, and keep the input tensor as state to compute the mask during the backward pass and then use the mask_fill.

I prefer the first method, since in both cases the memory requirement is the same, and even protetially lower (bool tensor < float tensor) and the implementation is way simpler (nothing to write). In terme of computation, there are 2 (potentially inplace) operations during the forward pass, and 1 (potentially inplace) operation  during the backward pass for the first solution. The second solution has 1 operation (never inplace) during the forward pass, and two (potentially inplace) operations during the backward pass.

nathanielsimard (changes requested):

antimora:
@nathanielsimard I finished changes per your comments but I could not accomplish two items:

1. I cannot remove int implementation. The compiler is complaining and telling I should have them.
2. I could not move Tch implementation to Ops level. The call fails because I am forced to use f64 for i64.

nathanielsimard (approved):
Everything seems fine now! See the last change to be made before merging.

burn-tch/src/ops/base.rs
```diff
@@ -369,4 +369,29 @@ impl<E: tch::kind::Element + Copy + Default> TchOps<E> {
...
+        TchTensor::new(tensor)
+    }
+
+    pub fn clamp<const D: usize, S: Into<tch::Scalar> + Clone>(
+        tensor: TchTensor<E, D>,
+        min: S,
+        max: S,
+    ) -> TchTensor<E, D> {
+        let tensor = tensor.tensor.clamp(min.into(), max.into());
+        TchTensor::new(tensor)
```
nathanielsimard, line 395: Before merging, can you use the `unary_ops` before executing the tensot ops? Just to make sure we call the inplace operation when possible.

nathanielsimard (changes requested):

nathanielsimard (approved):

Issue #506, opened 2023-07-18
Addition of `abs` tensor opperator


antimora:
**Is your feature request related to a problem? Please describe.**

I'm frequently working with numerical computations involving tensors, where it often becomes necessary to compute the absolute values of all elements within a tensor. Currently, we do not have a straightforward method for this, leading to inconvenient workarounds.

**Describe the solution you'd like**

It would be beneficial to add a new `abs` operator in the tensor that works on tensor objects. This operator would compute and return a new tensor with the absolute values of each element from the original tensor.

**Describe alternatives you've considered**

```rust
let tensor = tensor.powf(2.0).sqrt();
```

**Additional context**

Many popular machine learning and numerical computation libraries (like TensorFlow, PyTorch, NumPy) support this operation out-of-the-box. Therefore, including this operator will bring our library more in line with established standards, improving the experience for users transitioning from other libraries.

antimora:
> Can I work on this?

Go for it but be aware there is a serious learning curve mainly to understand the tensor and backend architecture. I am currently implementing `clamp` operator. I should be done by tomorrow and I'll have a PR that you can use as a guide. We do not have a documentation on this unfortunately. I might summed it up after I am done.

If you'd like to dive in and learn different parts of the system, it's a great way to do it.

antimora:
@trfdeer, there is a simpler task for a start: https://github.com/burn-rs/burn/issues/487 if you want to give it a go.

trfdeer:
Sure, I'll start with that

mmalczak:
Hi
I have ready implementation, but still need a few checks. I will submit PR tomorrow probably.

antimora:
@mmalczak , I have submitted PR for `clamp` operator if you need help/info: https://github.com/burn-rs/burn/pull/550 

I haven't written the details but I will.

antimora:
@mmalczak assigned this ticket to you since you already have something written.

antimora:
I believe this is done.

PR #553, opened 2023-07-27
 Addition of abs tensor opperator #506

mmalczak:
`run-checks.sh` has been executed.

Related issue:
https://github.com/burn-rs/burn/issues/506

Description:
Added abs implementation for various backends and for the tensor.

### Testing

Tests for backends were straighforward.
Autodiff implementation was verified with PyTorch using the following script:
```
import torch
tensor_1 = torch.tensor([[0., -1.], [3., 4.]], requires_grad=True)
tensor_2 = torch.tensor([[6., 7.], [9., -10.]], requires_grad=True)
tensor_3 = torch.matmul(tensor_1,tensor_2.abs())
tensor_4 = torch.matmul(tensor_3,tensor_2)
tensor_4.sum().backward()
print(tensor_1.grad)
print(tensor_2.grad)

```

### DIFF
--- a/burn-autodiff/src/ops/tensor.rs
+++ b/burn-autodiff/src/ops/tensor.rs
@@ -1155,19 +1155,15 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
             type State = B::TensorPrimitive<D>;
 
             fn backward(self, ops: Ops<Self::State, 1>, grads: &mut Gradients) {
-                unary::<B, D, D, _>(ops.parents, ops.node, grads, |grad| {
-                    let input = ops.state;
-                    let abs_input = B::abs(input.clone());
-                    let value = B::div(input, abs_input);
-
-                    B::mul(grad, value)
-                });
+                unary::<B, D, D, _>(ops.parents, ops.node, grads, |grad| B::mul(grad, ops.state));
             }
         }
 
         match Abs.prepare([tensor.node], [tensor.graph]).statefull() {
             OpsKind::Tracked(prep) => {
-                prep.finish(tensor.primitive.clone(), B::abs(tensor.primitive))
+                let output = B::abs(tensor.primitive.clone());
+                let state = B::div(tensor.primitive, output.clone());
+                prep.finish(state, output)
             }
             OpsKind::UnTracked(prep) => prep.finish(B::abs(tensor.primitive)),
         }
--- a/burn-autodiff/src/ops/tensor.rs
+++ b/burn-autodiff/src/ops/tensor.rs
@@ -1147,6 +1147,32 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
         }
     }
 
+    fn abs<const D: usize>(tensor: ADTensor<B, D>) -> ADTensor<B, D> {
+        #[derive(Debug)]
+        struct Abs;
+
+        impl<B: Backend, const D: usize> Backward<B, D, 1> for Abs {
+            type State = B::TensorPrimitive<D>;
+
+            fn backward(self, ops: Ops<Self::State, 1>, grads: &mut Gradients) {
+                unary::<B, D, D, _>(ops.parents, ops.node, grads, |grad| {
+                    let input = ops.state;
+                    let abs_input = B::abs(input.clone());
+                    let value = B::div(input, abs_input);
+
+                    B::mul(grad, value)
+                });
+            }
+        }
+
+        match Abs.prepare([tensor.node], [tensor.graph]).statefull() {
+            OpsKind::Tracked(prep) => {
+                prep.finish(tensor.primitive.clone(), B::abs(tensor.primitive))
+            }
+            OpsKind::UnTracked(prep) => prep.finish(B::abs(tensor.primitive)),
+        }
+    }
+
     fn cos<const D: usize>(tensor: ADTensor<B, D>) -> ADTensor<B, D> {
         #[derive(Debug)]
         struct Cos;
--- a/burn-tensor/src/tensor/api/float.rs
+++ b/burn-tensor/src/tensor/api/float.rs
@@ -72,6 +72,11 @@ where
         Self::new(B::sqrt(self.primitive))
     }
 
+    /// Applies element wise absolute value operation.
+    pub fn abs(self) -> Self {
+        Self::new(B::abs(self.primitive))
+    }
+
     /// Applies element wise cosine operation.
     pub fn cos(self) -> Self {
         Self::new(B::cos(self.primitive))
--- a/burn-tensor/src/tensor/api/float.rs
+++ b/burn-tensor/src/tensor/api/float.rs
@@ -72,11 +72,6 @@ where
         Self::new(B::sqrt(self.primitive))
     }
 
-    /// Applies element wise absolute value operation.
-    pub fn abs(self) -> Self {
-        Self::new(B::abs(self.primitive))
-    }
-
     /// Applies element wise cosine operation.
     pub fn cos(self) -> Self {
         Self::new(B::cos(self.primitive))
--- a/burn-tensor/src/tensor/ops/tensor.rs
+++ b/burn-tensor/src/tensor/ops/tensor.rs
@@ -920,7 +920,7 @@ pub trait TensorOps<B: Backend> {
     ///
     /// # Arguments
     ///
-    /// * `tensor` - The tensor to exponentiate.
+    /// * `tensor` - The tensor to take absolute value of.
     ///
     /// # Returns
     ///
--- a/burn-tensor/src/tensor/ops/tensor.rs
+++ b/burn-tensor/src/tensor/ops/tensor.rs
@@ -916,6 +916,17 @@ pub trait TensorOps<B: Backend> {
     /// A tensor with the same shape as `tensor` with square root values.
     fn sqrt<const D: usize>(tensor: B::TensorPrimitive<D>) -> B::TensorPrimitive<D>;
 
+    /// Returns a new tensor with absolute values.
+    ///
+    /// # Arguments
+    ///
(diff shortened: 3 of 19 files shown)


burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -858,6 +858,17 @@ pub trait TensorOps<B: Backend> {
     /// A tensor with the same shape as `tensor` with square root values.
     fn sqrt<const D: usize>(tensor: B::TensorPrimitive<D>) -> B::TensorPrimitive<D>;
 
+    /// Returns a new tensor with absolute values.
+    ///
+    /// # Arguments
+    ///
+    /// * `tensor` - The tensor to exponentiate.
```
antimora, line 865: Probably you meant to replace the word "exponentiate"

burn-tensor/src/tests/ops/abs.rs
```diff
@@ -0,0 +1,16 @@
...
+    #[test]
+    fn should_support_abs_ops() {
+        let data = Data::from([[0.0, -1.0, 2.0], [3.0, 4.0, -5.0]]);
+        let tensor = Tensor::<TestBackend, 2>::from_data(data);
+
+        let data_actual = tensor.abs().into_data();
+
+        let data_expected = Data::from([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]);
+        data_expected.assert_approx_eq(&data_actual, 3);
+    }
```
antimora, line 15: We also need tests for Int tensor. Please see this example: https://github.com/burn-rs/burn/pull/550/files#diff-0479313033166bd3d20e80b633111dada0400a368a82e66c96754edade7268dcR17-R22

antimora (changes requested):
Overall it looks good. Thank you for your contribution. 

Requesting one word change and testing for Int tensor (you might also be missing some int implementation). 

Please see my recent PR for how to add Int implementation. We are missing a good guide, so it's kinda confusing. For now, please refer to this PR file changes:

https://github.com/burn-rs/burn/pull/550/files

burn-autodiff/src/ops/tensor.rs
```diff
@@ -1147,6 +1147,32 @@ impl<B: Backend> TensorOps<ADBackendDecorator<B>> for ADBackendDecorator<B> {
...
+        struct Abs;
+
+        impl<B: Backend, const D: usize> Backward<B, D, 1> for Abs {
+            type State = B::TensorPrimitive<D>;
+
+            fn backward(self, ops: Ops<Self::State, 1>, grads: &mut Gradients) {
+                unary::<B, D, D, _>(ops.parents, ops.node, grads, |grad| {
+                    let input = ops.state;
+                    let abs_input = B::abs(input.clone());
+                    let value = B::div(input, abs_input);
```
nathanielsimard, line 1161: We could calculate the division directly during the forward pass and use it as the state. It would avoid recalculating the `abs` during the backward pass and doesn't increase the size of the state.

nathanielsimard (commented):
Maybe just look if we can increase the performance of the backward pass, but still very good job!

burn-tensor/src/tensor/ops/tensor.rs
mmalczak: Yes, sorry, I used different methods as templates and I forgot to change that

burn-autodiff/src/ops/tensor.rs
mmalczak: I am new to these kinds of libraries, but wouldn't it be inconsistent?

burn-tensor/src/tensor/ops/tensor.rs
mmalczak: Fixed

burn-autodiff/src/ops/tensor.rs
nathanielsimard: The API is made such that we can specilize each backward pass to have the best performance possible, so yes it is inconsistent in _implementation_ but not in _behavior_.

burn-tensor/src/tests/ops/abs.rs
mmalczak: In the example you have submitted, why is there no tensor.rs implementation for burn-autodiff?

mmalczak: Also, there is one more thing I do not quite understand.
In burn-tensor/src/tensor/ops/int_tensor.rs you provide implementation for int_clamp_min and int_clamp_max.
However, in the test (burn-tensor/src/tests/ops/clamp.rs) for integer implementation you use clamp_min and clamp_max methods. Why is that? 
I imagine that there is something going on in between burn-tensor/src/tensor/api/ and  burn-tensor/src/tensor/ops/, but I cannot really see that.

antimora: > In the example you have submitted, why is there no tensor.rs implementation for burn-autodiff?

Good point. Initially I had it. But in my offline discussion with @nathanielsimard during the review, he said since I have implemented the default method using the existing ops, my backward pass implementation resembled what autodif would have derived, it's better not implement it for Auto diff backend. He said you only need it if there is no default and/or you want a more optimized version. 

In your case, you do not have the default implementation ( ^2.sqrt() would be default but it's slow), you'll need to implement the backward pass.

antimora: > Also, there is one more thing I do not quite understand. In burn-tensor/src/tensor/ops/int_tensor.rs you provide implementation for int_clamp_min and int_clamp_max. However, in the test (burn-tensor/src/tests/ops/clamp.rs) for integer implementation you use clamp_min and clamp_max methods. Why is that? I imagine that there is something going on in between burn-tensor/src/tensor/api/ and burn-tensor/src/tensor/ops/, but I cannot really see that.

Yes, it's confusing and that's another reason we need a guide/documentation. 

Method names with `int_` come from the backend implementation. I think the prefix is to avoid collision. The test that I added actually tests the official Tensor API using a backend.

mmalczak: I have added integer implementation, but I have some doubts. I would be grateful if you could have a look

antimora:
OK. I'll review it. Meanwhile, can you rebase and resolve the existing conflicts? Thank you.

burn-tensor/src/tensor/api/float.rs
```diff
@@ -72,6 +72,11 @@ where
         Self::new(B::sqrt(self.primitive))
     }
 
+    /// Applies element wise absolute value operation.
+    pub fn abs(self) -> Self {
+        Self::new(B::abs(self.primitive))
+    }
```
nathanielsimard, line 78: You should use the `numeric` API instead, you can look at how `sum` is implemented there. https://github.com/burn-rs/burn/blob/f453f47efb1887ed736faf03157530931b2d3d18/burn-tensor/src/tensor/api/numeric.rs#L134C14-L134C14

nathanielsimard (changes requested):
Just a small change in the tensor/api, but otherwise it's pretty good.

burn-tensor/src/tensor/api/float.rs
antimora: @nathanielsimard what about the backward pass implementation? Was your suggestion is for exact optimization?

nathanielsimard: Yes, it's mostly to reduce the number of kernels executed when using `abs` with autodiff. The other review is still valid, this is just my new comments about the `int` and `float` implementations.

mmalczak: I have committed changes for numeric API. 
However, I would be grateful of explanation, why it should be there

mmalczak: If you want to work on documentation, maybe I could help you a little bit, by asking questions :D

burn-autodiff/src/ops/tensor.rs
mmalczak: To be honest I am not really sure how to do it right now. I need to dive more into the library.
I hope to have it ready in a few days

mmalczak: Could you please point me to an example where this kind of optimization is done?

burn-tensor/src/tensor/api/float.rs
antimora: > I have committed changes for numeric API. However, I would be grateful for explanation, why it should be there

@nathanielsimard knows the full answer but as I understood numeric if for int and float at the same time. If you have a specific behavior only for float than it goes into float. 

@nathanielsimard at some we'll need more explanation from you otherwise I'll be telling people legends =P

antimora:
Thanks, again, @mmalczak. It looks good from my end. Once @nathanielsimard okays, it's good to go!

Your help with documentation/question would be appreciated. The process to add new ops should be made clear because we expect to fill the gap by user usage/contribution.

burn-autodiff/src/ops/tensor.rs
mmalczak: Ok, now it's clear:) Thank you
I have committed suggested changes

mmalczak: Do you think it would make sense to have some benchmarks run automatically, to be able to track performance changes easily?

antimora: Yes, it makes sense to have benchmarks run automatically. We have some benches that measure ( see https://github.com/burn-rs/burn-benches) but they're run manually. One main challenge is having right machines with the devices to test different backends. Our CI uses GitHub provided machines, which is limited for full benchmarks. Perhaps, we could only test with CPU devices on provided backends.

antimora: > Do you think it would make sense to have some benchmarks run automatically, to be able to track performance changes easily?

Filed a ticket to track it: https://github.com/burn-rs/burn/issues/573

nathanielsimard (approved):
Thanks a lot for the contribution!

PR #556, opened 2023-07-27
Rewrite `publish.sh` script in Rust language

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR rewrites `publish.sh` shell script in `Rust` in order to be less error-prone and at the same time usable from different operating systems.

### Testing

Build `publish.rs` and run it locally


scripts/publish.rs
```diff
@@ -0,0 +1,127 @@
...
+    // Convert cargo search output into a str
+    let remote_version_str = str::from_utf8(&cargo_search_output.stdout)
+        .expect("Failed to convert cargo search output into a str");
+
+    // Extract only the remote crate version from str
+    let (remote_version, _) = remote_version_str
+        .split_once('=')
+        .map(|(_, second)| second.trim_start().split_once(' '))
+        .flatten()
+        .expect("Failed to get remote crate version");
```
nathanielsimard, line 46: The function shouldn't crash if there this is the first time being deployed, maybe returning `Option<String>` is preferable here and more explicit?

scripts/publish.rs
```diff
@@ -0,0 +1,127 @@
...
+        "Failed to write cargo publish on stdout",
+        "Failed to write cargo publish on stderr",
+    );
+}
+
+fn main() {
+    // Get crate name
+    let crate_name = env::args()
+        .nth(1)
+        .expect("You need to pass the crate name as first argument!");
```
nathanielsimard, line 104: What is the 0 th argument?

nathanielsimard (changes requested):
Just make sure the script can publish new crates (don't crash when a remote version doesn't exist). But very good job!

scripts/publish.rs
antimora: The executable path, so typically you'd skip the 0th arg.

scripts/publish.rs
```diff
@@ -0,0 +1,127 @@
+use std::env;
```
antimora, line 1: Could we please add documentation at the top that explains the script purpose and how to run it. It's kinda unconventional way to execute the script, so people won't know it: 

```
rustc scripts/publish.rs --crate-type bin --out-dir scripts
```

antimora (changes requested):
Good job putting all these comments and explain what each line/block does. 

I have a small request to make. Please see my inlined comment.

scripts/publish.rs
Luni-4: Yes, you're right. I'm going to add a check

Luni-4: I think it's better to check the results produced by `cargo search` in case of a non-existent crate and in that case return `None`

nathanielsimard (approved):

PR #566, opened 2023-07-31
AdamW implementation

wdoppenberg:
## AdamW implementation

I started with an easy one to get familiar. Please, if you have time, provide feedback where necessary.

### Checklist

- [x] Ensure PyTorch forward pass yields equal updated state with a precision of 6

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/543

### Changes

Added `adamw.rs` under `burn-core/src/optim`. Fixed some spelling mistakes ('Momemtum' -> 'Momentum').

### Testing

I kept the tests the same as they were for `Adam`, albeit with an updated set of expected weights + biases and a higher precision requirement. I will post a more detailed comparison to a PyTorch script under here.


wdoppenberg:
Might I also suggest renaming the `grad` parameter of `WeighDecay`'s `transform` method to `tensor`? So:

```rust
/// State of [WeightDecay](WeightDecay).
#[derive(Record, Clone, new)]
pub struct WeightDecayState<B: Backend, const D: usize> {
    last_step: Tensor<B, D>,
}

/// Weight decay implementation that transforms gradients.
pub struct WeightDecay<B: Backend> {
    penalty: B::FloatElem,
}

impl<B: Backend> WeightDecay<B> {
    /// Creates a new [WeightDecay](WeightDecay) from a [WeightDecayConfig](WeightDecayConfig).
    pub fn new(config: &WeightDecayConfig) -> Self {
        Self {
            penalty: config.penalty.elem(),
        }
    }

    /// Transforms a tensor by adding the penalty to the tensor and keeps track of the previous
    /// state of the tensor.
    ///
    /// # Arguments
    ///
    /// * `tensor` - Tensor to transform.
    /// * `state` - State of the optimizer.
    ///
    /// # Returns
    ///
    /// * `tensor` - Transformed tensor.
    /// * `state` - State of the optimizer.
    pub fn transform<const D: usize>(
        &self,
        tensor: Tensor<B, D>,
        state: Option<WeightDecayState<B, D>>,
    ) -> (Tensor<B, D>, WeightDecayState<B, D>) {
        let last_step = tensor.clone();

        let tensor = match state {
            Some(state) => state.last_step.mul_scalar(self.penalty).add(tensor),
            None => tensor,
        };

        (tensor, WeightDecayState::new(last_step))
    }
}```

**Review by nathanielsimard — CHANGES_REQUESTED** [review]
I would check if weight decay is applied correctly for both Adam and AdamW. Maybe this explains why the numbers are a bit different than PyTorch. I think we could use bigger values in the tests to make it fail (like weight decay of 0.5).

File: burn-core/src/optim/adamw.rs [hunk 1]
```diff
@@ -0,0 +1,324 @@
...
+            state_momentum = Some(state.momentum);
+        }
+
+        if let Some(weight_decay) = &self.weight_decay {
+            let (tensor_transformed, state) = weight_decay.transform(tensor.clone(), state_weight_decay);
+            state_weight_decay = Some(state);
+            tensor = tensor - tensor_transformed.mul_scalar(lr);
+        }
+
+        let (grad, state_momentum) = self.momentum.transform(grad, state_momentum);
```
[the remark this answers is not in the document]
wdoppenberg: Yeah I used the algorithm description on the PyTorch docs as reference. Basically held their `Adam` and `AdamW` pages next to each other to find the exact differences. I will resume work on this tonight.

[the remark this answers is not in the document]
nathanielsimard: Thanks :)

antimora:
We have fixed our CI. Please merge the latest from the main.

wdoppenberg:
I've pushed some of my work once again. It is not a done deal yet, I'm still getting significant variation from the output that PyTorch provides. Also, I wasn't exactly sure why weight decay needed state, so in the current implementation it just uses the decay factor $\lambda$ for $\theta_t = \theta_{t-1} - \gamma\lambda\theta_{t-1}$.

wdoppenberg:
Hi @antimora @nathanielsimard,

I recently had some time to work on this again, and I seem to have found the error, it appeared that the `AdaptiveMomentumState` wasn't initialized with $t=0$ but with $t=1$, which caused a discrepancy between this and the reference PyTorch implementation. Please let me know if you want to have the reference script saved somehow. 

Let me know what you think.

nathanielsimard:
@antimora @wdoppenberg I'm not sure if preserving the script in the codebase is necessary. I'm just worried about maintainability with tons of PyTorch scripts that never actually run. I think the case for ONNX is a bit different, since we never modify by hand the generated ONNX file. But we do write the tests with the generated numbers. Keeping two parts of the codebase in sync is something I would like to avoid. However, adding the script to the PR description can be great to keep track of what was used to generate the numbers.

I think it's a good idea to create a PR that fix the current adam implementation, with the potential refactor where you move the code into its own module.

antimora:
> @antimora @wdoppenberg I'm not sure if preserving the script in the codebase is necessary. I'm just worried about maintainability with tons of PyTorch scripts that never actually run. I think the case for ONNX is a bit different, since we never modify by hand the generated ONNX file. But we do write the tests with the generated numbers. Keeping two parts of the codebase in sync is something I would like to avoid. However, adding the script to the PR description can be great to keep track of what was used to generate the numbers.
> 
> I think it's a good idea to create a PR that fix the current adam implementation, with the potential refactor where you move the code into its own module.

Yeah I agree. We can keep the script in the PR for record keeping and code review. It will definitely help reviewers to peak how numbers are generated (not required but it'd be helpful). Also it gives the author to archive the script somewhere instead of discarding it.

antimora:
@wdoppenberg CI is failing due to formatting/clippy.

wdoppenberg:
> @wdoppenberg CI is failing due to formatting/clippy.

So that was it, I couldn't figure out what was making the pipeline fail. It should be fixed now.

burn-core/src/optim/adamw.rs
```diff
@@ -0,0 +1,369 @@
...
+    #[test]
+    fn test_adamw_optimizer_save_load_state() {
+        let linear = nn::LinearConfig::new(6, 6).init();
+        let x = Tensor::<TestADBackend, 2>::random([2, 6], Distribution::Default);
+        let mut optimizer = create_adamw();
+        let grads = linear.forward(x).backward();
+        let grads = GradientsParams::from_grads(grads, &linear);
+        let _linear = optimizer.step(LEARNING_RATE, linear, grads);
+        BinFileRecorder::<FullPrecisionSettings>::default()
+            .record(optimizer.to_record(), "/tmp/test_optim".into())
```
antimora, line 218: I think it's a good opportunity to improve the usage of temporary directories. Could we please use the `TempDir` from the `tempfile` crate? I realize this is how the previous tests were done, and I wish to refactor them later.

Here's what you'll need to do:
Add the following to burn-core's `Cargo.toml`:
```
[dev-dependencies]
tempfile = {workspace = true}
```
You can follow this usage example from [TempDir::new()](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.new):

```rust
use tempfile::TempDir;
let temp_dir = TempDir::new().unwrap();
```

The advantage of this approach is that it is cross-platform, and temporary directories are removed after the `temp_dir` variable goes out of scope.

antimora (changes requested):
Looks good. I am recommending we use `tempfile::TempDir` to create temp directories for testing. I added steps how to.

antimora (approved):
Looks good from my end. @nathanielsimard would still need to review it since he understand the optimization module.

burn-core/src/optim/adamw.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    fn given_linear_record(
+        weight: Data<f32, 2>,
+        bias: Data<f32, 1>,
+    ) -> nn::LinearRecord<TestADBackend> {
+        nn::LinearRecord {
+            weight: Param::from(Tensor::from_data(weight)),
+            bias: Some(Param::from(Tensor::from_data(bias))),
+        }
+    }
```
nathanielsimard, line 355: I feel like this could be shorten.

```rust
fn given_linear_layer(weight: Data<f32, 2>, bias: Data<f32, 1>) -> nn::Linear<TestADBackend> {
    let record = nn::LinearRecord {
        weight: Param::from(Tensor::from_data(weight)),
        bias: Some(Param::from(Tensor::from_data(bias))),
    };

    nn::LinearConfig::new(6, 6).init_with(record)
}
```

nathanielsimard (commented):
Just a small refactor proposition, @wdoppenberg let me know what you think and we can merge afterward.

burn-core/src/optim/adamw.rs
wdoppenberg: Agree, will push changes now

nathanielsimard (approved):

PR #568, opened 2023-07-31
Fix CI

nathanielsimard:
Use the stable version of graphics drivers PPA instead of the bleeding edge one. See this [article](https://itsfoss.com/install-mesa-ubuntu/) for reference.


antimora (approved):
Looks good to me. 

Judging from the commits I could see you tried many things. Lets hope this will fix it for good.

PR #570, opened 2023-08-01
Feat/wgpu/prng normal

louisfd:
### Checklist

### Changes

New kernel for normal distribution. 
Uses the same underlying pseudo-random generator as uniform (lcg-tausworthe)
But added Box-Muller transform to map it to normal distribution. 

### Testing

Tested that mean of data is indeed the expected mean
Used bin stats to check this property of normal distributions:  https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule


antimora (approved):
LGTM

PR #574, opened 2023-08-02
Feat/conv transpose2d

### LINKED ISSUE #540: Fully support Convolution transposed backward
For now, the kernels are implemented but the backward pass is not supported. The module is not exposed in burn-core because of that.

nathanielsimard:
## Related Issues/PRs

Partial Fix #540  (ConvTranspose1D to be done in a following PR)

## Changes

Change the init method for all convolution modules to be similar to PyTorch.

Conv1D: https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html
Conv2D: https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html
ConvTranspose2D: https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html

Fix init methods, shapes and documentation when using `groups > 1`


burn-core/src/nn/conv/conv1d.rs
```diff
@@ -62,18 +67,33 @@ pub struct Conv1d<B: Backend> {
 impl Conv1dConfig {
     /// Initialize a new [conv1d](Conv1d) module.
     pub fn init<B: Backend>(&self) -> Conv1d<B> {
-        let shape = [self.channels_out, self.channels_in, self.kernel_size];
+        let shape = [
+            self.channels_out,
+            self.channels_in / self.groups,
+            self.kernel_size,
+        ];
         let fan_in: usize = self.channels_in * self.kernel_size;
```
louisfd, line 75: TL;DR: divide fan_in by self.groups and revert initialization to what we had before. 

The former uniform Kaiming initialization
Was equivalent to
gain * Uniform {-k, k} with k = sqrt( 3 / fan_in) and gain = sqrt( 1 / 3 )
Which is therefore equivalent to
Uniform {-k, k} with k = sqrt( 1 / fan_in ) 

You have replaced it with
Uniform {-k, k} with k = sqrt( groups / fan_in )
With fan_in unchanged (i.e. still equal to channels_in * kernel_size). 

If you divide fan_in by self.groups, then you will get 
sqrt( 1 / new_fan_in ) = sqrt( 1 / (old_fan_in / groups) ) = sqrt( groups / old_fan_in ), leading to exactly the same thing you had. 
Therefore you can revert to the Kaiming initialization we had before. It will simply use a fan_in that is divided by the number of groups instead

louisfd (changes requested):
See comment.

burn-core/src/nn/conv/conv2d.rs
```diff
@@ -63,21 +68,32 @@ impl Conv2dConfig {
     pub fn init<B: Backend>(&self) -> Conv2d<B> {
         let shape = [
             self.channels[1],
-            self.channels[0],
+            self.channels[0] / self.groups,
```
antimora, line 71: Does it mean we had a bug that the shape was not calculated correctly for groups > 1?

BTW, in PyTorch's implementation, the function panics if the first channel is not divisible, which we should do as well. Otherwise, the value floored, e.g. 11/4.

burn-core/src/nn/conv/conv_transpose2d.rs
```diff
@@ -0,0 +1,183 @@
...
+    padding: [usize; 2],
+    padding_out: [usize; 2],
+}
+
+impl ConvTranspose2dConfig {
+    /// Initialize a new [conv transpose 2d](ConvTranspose2d) module.
+    pub fn init<B: Backend>(&self) -> ConvTranspose2d<B> {
+        let shape = [
+            self.channels[0],
+            self.channels[1] / self.groups,
```
antimora, line 74: Same comment about divisibility panic.

antimora (changes requested):
I agree with @louisfd if we can use the existing initializer then we should it it.

I also had a comment about group divisibility.

antimora (approved):
Looks good. Thanks for fixing the issues.

Zheby:
你好，我有一个请求，就是能够实现一下上采样吗？(Upsample),类似https://pytorch.org/docs/stable/generated/torch.nn.Upsample.html

antimora:
> 你好，我有一个请求，就是能够实现一下上采样吗？(Upsample),类似https://pytorch.org/docs/stable/generated/torch.nn.Upsample.html

Please file a separate ticket. There is a related ticket: https://github.com/burn-rs/burn/issues/455

Issue #564, opened 2023-07-30
Windows: Downloading Python dependencies Error


mashirooooo:
**Describe the bug**
When running examples and downloading Python dependencies in the virtual environment (venv) on Windows, the directory structure of the venv is inconsistent with that on Unix systems. This discrepancy leads to download errors

antimora:
Can you please what kind of errors and inconsistency do you see? It would be helpful to troubleshoot. Perhaps the paths were assumed.

mashirooooo:
When using venv on Unix systems, the Python executable `python3` is located in the `.bin` directory, whereas on Windows, the Python executable `python.exe` is located in the `Scripts` directory；
If possible, could you please make the script `burn-dataset/src/source/huggingface/downloader.rs` compatible with Windows? Currently, I have continued the process by using `Revert "Install dataset downloader python deps in venv (#496)"` because I'm not familiar with Python, and thus, I am unable to submit a pull request.

nathanielsimard:
Burn should be compatible with Windows, so we definitively need to fix it if there is a problem.

Macil:
The MNIST example doesn't work on Windows (assuming ~/.cache/burn-dataset/mnist.db doesn't exist yet and needs to be downloaded) and fails with this output:

```
PS E:\chris\Coding\burn> cargo run --example mnist --release --features ndarray
    Finished release [optimized] target(s) in 0.74s
warning: the following packages contain code that will be rejected by a future version of Rust: nom v1.2.4
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `target\release\examples\mnist.exe`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }', burn-dataset\src\source\huggingface\downloader.rs:237:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\release\examples\mnist.exe` (exit code: 101)
```

Changing the code to print the command before running it shows that it's trying to run ~/.cache/burn-dataset/venv/bin/python3, but there's no such executable. It works if it instead runs ~/.cache/burn-dataset/venv/Scripts/python. ([VirtualEnv arbitrarily uses the "Scripts" directory instead of "bin" on Windows.](https://stackoverflow.com/questions/43826134/why-is-the-bin-directory-named-differently-scripts-on-windows))

Issue #552, opened 2023-07-27
[RFC]: Continuous Integration Design 


Luni-4:
This issue is an extension of #428 and it is an RFC to verify if there is a consensus with this new design

### Feature description

A new design for continuous integration might be the following one:
- [x] One job for code formatting and linting issues
- [x]  One job that runs all `std` tests on `Linux`
- [x] One job that runs all `std` tests on `MacOs` (@dcvz)
- [x] One job that runs all `std` tests on `Windows` (@syl20bnr)
- [x] One job that runs all `no-std` tests

Let's try to tackle each point one-by-one

- (1) Some `GitHub` actions allow to visualize `clippy` lints and bad formatting as messages in a PR or in a report. This approach reduces the time spent in detecting these problems because they are immediately viewable. A contributor, especially a new one, does not have to scrape through CI logs in order to find them. This job could run **only** on a Linux operating system. Running this job on more operating system could be a waste of resources.

- (2, 3, 4) Run all current checks and tests on the specific operating system using a script. The `run-checks.sh` script could be rewritten in `Rust` language in order to be cross-platform and less error-prone. Collect code coverage data through Rust source code coverage and then aggregate the results using [grcov](https://github.com/mozilla/grcov), finally publish this data on [codecov](https://about.codecov.io/) in order to immediately visualize which lines of code are **not** covered by tests. (thanks to @gonzalesMK for the help)

- (5) Maintain the same checks which are already performed. Even this job will benefit from `run-checks.sh` rewriting in `Rust`.

### Feature motivation

Having a Continuous Integration that implements the points described above, in my opinion, reduces the effort in code maintainability, making the development more interesting, also decreases the possibility to have problems in the future.

Luni-4:
What we still have to implement :

- [ ] Use GitHub Actions [filters](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-filters) to avoid running CI for nothing
- [ ] Implement `Grcov` checks **ONLY** on CI
- [ ] Try another strategy to run `clippy` on CI
- [ ] Add MacOS and Windows configurations through matrix

gonzalesMK:
@Luni-4  I have 2 points to consider:

1. Should we run Rust Doc just once on the same job as Rust Clippy? 
2. Can we minimize running coverage? Maybe run coverage in Linux jobs and then run coverage for specific packages on other OS?

gonzalesMK:
[...]

PR #575, opened 2023-08-02
Fix huggingface downloader Windows support (#564)

Macil:
### Checklist

- [ ] Confirm that `run-checks.sh` has been executed. *(I'm not sure if this can be run on Windows. In the burn-dataset directory, I have manually tried "cargo test --all-features" which has the unrelated tests "dataset::sqlite::tests::sqlite_writer_write" and "dataset::sqlite::tests::sqlite_writer_write_multi_thread" fail, and tried "cargo doc --all-features" which works fine.)*

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/564

### Changes

The path that the huggingface dataset downloader runs Python from has been changed to support how Python VirtualEnv arranges its files on Windows ([which is arbitrarily different](https://github.com/pypa/virtualenv/commit/993ba1316a83b760370f5a3872b3f5ef4dd904c1)).

### Testing

The commands `cargo run --example mnist --release --features ndarray` and `cargo run --example mnist --release --features wgpu` have been tested. They both now get past the download step and successfully start training with this PR's changes.

### DIFF
--- a/burn-dataset/src/source/huggingface/downloader.rs
+++ b/burn-dataset/src/source/huggingface/downloader.rs
@@ -10,6 +10,10 @@ use thiserror::Error;
 
 const PYTHON: &str = "python3";
 const PYTHON_SOURCE: &str = include_str!("importer.py");
+#[cfg(not(target_os = "windows"))]
+const VENV_BIN_PYTHON: &str = "bin/python3";
+#[cfg(target_os = "windows")]
+const VENV_BIN_PYTHON: &str = "Scripts\\python";
 
 /// Error type for [HuggingfaceDatasetLoader](HuggingfaceDatasetLoader).
 #[derive(Error, Debug)]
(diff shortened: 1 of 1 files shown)


nathanielsimard:
We have an issue to rewrite the CI in Rust so that it can be run on all devices #552.

burn-dataset/src/source/huggingface/downloader.rs
```diff
@@ -10,6 +10,10 @@ use thiserror::Error;
 
 const PYTHON: &str = "python3";
```
nathanielsimard, line 11: I don't think this is used anymore, if this is the case, we should delete this line.

Macil: Yes, the "python3" constant is still successfully used to create the virtual env to begin with, but then that virtual env contains a "Scripts/python.exe" file.

mashirooooo:
If possible, please add code to initialize a "venv" environment and check if it is properly initialized to ensure the code runs smoothly.

PR #578, opened 2023-08-02
Full support for ONNX scalar operators and Constants

antimora:
### Checklist

### Related Issues/PRs

### Changes

There many fixes related to supporting ONNX scalar ops and Constants: 

1. Refactor Binary and Unary nodes to support scalar types.
2. Fix output and input of the graph. Previously it was assumed the first node contained the input. 
3. Introduce Scalar Burn type for source generation.
4. Move ONNX conversion tests into a separate crate that tests end to end (ONNX to source code, compile and run models).
5. Add/Sub/Mul/Div support scalar values.
6. Full Constant node support (scalar and tensor). The tensor weights are loaded from exported weights. 
7. Support code generation when Model module does not have any fields.
8. Added Cast operator (was needed for scalar type testing).
9. Some other small fixes.

### Testing

Added new tests and updated the existing.


burn-import/src/burn/node/constant.rs
```diff
@@ -1,72 +1,153 @@
...
-    Bool(bool),
+pub enum ConstantValue<PS: PrecisionSettings> {
+    /// Float constant.
+    Float32(f32),
+
+    /// Integer constant.
+    Int64(i64),
+
+    /// Tensor constant (does not hold a value).
+    Tensor(TensorType, TensorValue<PS>),
```
nathanielsimard, line 30: This is a bit confusing, the comment says does not hold a value but there is `TensorValue<PS>`.

burn-import/src/burn/node/constant.rs
```diff
@@ -1,72 +1,153 @@
...
     }
     pub fn val_tokens(&self) -> TokenStream {
         match self {
-            ConstantValue::Int(val) => quote! { #val },
-            ConstantValue::Float(val) => quote! { #val },
-            ConstantValue::Bool(val) => quote! { #val },
+            ConstantValue::Float32(val) => quote! { #val },
+            ConstantValue::Int64(val) => quote! { #val },
+            ConstantValue::Tensor(_, _) => {
+                panic!("Tensor constant is not assignable.")
```
nathanielsimard, line 49: Why not return `TensorValue<PS>` to tokens?

burn-import/src/burn/node/constant.rs
```diff
@@ -1,72 +1,153 @@
...
+                let shape = tensor_type.clone().shape.unwrap().to_tokens();
+                let dim = tensor_type.clone().dim.to_tokens();
+
+                if with_record {
+                    Some(quote! {
+                        let #name = record.#name;
+                    })
+                } else {
+                    Some(quote! {
+                        let #name: #ty = Tensor::<B, #dim>::zeros(#shape);
```
nathanielsimard, line 109: I think tensor fields should be `Param<Tensor<B, D>>`  even if constant. When constant, we should just call `set_require_grad(false)` during the init. This would avoid having to change the record type for `Tensor<B, D>`.

nathanielsimard: I think this is appropriate when we need the tensor to be included in the final model record but behaves as a constant (Constant model parameter VS just a constant tensor).

burn-import/src/burn/node/constant.rs
```diff
@@ -1,72 +1,153 @@
...
     fn into_node(self) -> Node<PS> {
         Node::Constant(self)
     }
+
+    fn field_serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        match &self.value {
+            ConstantValue::Tensor(_tensor_type, ds) => match ds {
+                TensorValue::Float(ds) => ds.serialize(serializer),
+                TensorValue::Int(ds) => ds.serialize(serializer),
+            },
```
nathanielsimard, line 147: I guess if we update the tensor type to `Param<Tensor>` we can use the `Record` API like in the `linear node`.

burn-import/src/burn/node/conv2d.rs
```diff
@@ -154,6 +154,8 @@ mod tests {
             Conv2dConfig::new([3, 3], [3, 3]).with_padding(PaddingConfig2d::Valid),
         ));
 
+        graph.register_input_output(vec!["input".to_string()], vec!["output".to_string()]);
```
nathanielsimard, line 157: I'm not sure to understand why we need to add this part to all nodes. Why don't we use the actual name of each input and output instead of hardcoding the names ?

burn-import/src/burn/node/unary.rs
```diff
@@ -78,122 +90,188 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for UnaryNode {
...
+    pub(crate) fn transpose(input: Type, output: Type) -> Self {
         let function = move |input| quote! { #input.transpose() };
         Self::new(input, output, UnaryNodeKind::Transpose, Arc::new(function))
     }
+
+    pub(crate) fn cast(input: Type, output: Type) -> Self {
+        let function = match output.clone() {
+            Type::Scalar(scalar) => {
+                let ty = scalar.ty();
+                move |input| quote! { #input as #ty }
```
nathanielsimard, line 126: If the output type is a tensor, we should generate another code block.

```rust
Tensor::from_data(Data::from([#input]).convert()).unsqueeze();
```

nathanielsimard (commented):
I have some questions, but overall, it looks good. The tests are a great addition, they are really essential!

burn-import/src/burn/graph.rs
```diff
@@ -246,10 +253,21 @@ impl<PS: PrecisionSettings> BurnGraph<PS> {
...
-            #[derive(Module, Debug)]
-            pub struct Model<B: Backend> {
-                #body
+        // Add dummy field if no field is present to avoid empty struct
+        // and make sure we can derive Module trait and use it in a model.
+        if body.is_empty() {
+            quote! {
+                #[derive(Module, Debug)]
+                pub struct Model<B: Backend> {
+                    _dummy: Tensor<B, 1>,
```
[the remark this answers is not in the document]
antimora: I tried but module was not implemented. Should I implement module?

burn-import/src/burn/node/batch_norm.rs
```diff
@@ -102,13 +102,13 @@ macro_rules! batch_norm_serialize {
 
 impl<PS: PrecisionSettings> NodeCodegen<PS> for BatchNormNode<PS> {
     fn input_types(&self) -> Vec<Type> {
-        vec![Type::Tensor(&self.input)]
+        vec![Type::Tensor(self.input.clone())]
```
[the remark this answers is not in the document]
antimora: Tensor type does not have data and it's small. It was difficult to work with lifetimes, often required changing the top structs to declare lifetimes.

burn-import/src/burn/node/constant.rs
antimora: Outdated comment. I'll remove it.

burn-import/src/burn/node/constant.rs
antimora: I think we can make it work without changing Tensor record. I'll try it.

burn-core/src/module/param/constant.rs
```diff
@@ -112,7 +112,7 @@ constant!(i16);
 constant!(i8);
 
 impl<const D: usize, B: Backend> Module<B> for Tensor<B, D> {
-    type Record = ConstantRecord;
+    type Record = Tensor<B, D>;
```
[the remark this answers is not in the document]
antimora: I will see if I can make it work without it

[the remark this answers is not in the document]
antimora: With your help, we removed and now using Param<Tensor> with grad disabled in place of tensor module field. This allows to save/load weights after a finetuning if needed.

burn-import/src/burn/node/constant.rs
antimora: With your help, we removed and now using Param with grad disabled in place of tensor module field. This allows to save/load weights after a finetuning if needed.

burn-import/src/onnx/to_burn.rs
```diff
@@ -419,8 +486,54 @@ impl Argument {
...
             ArgType::Tensor(tensor) => TensorType::new_float(self.name.clone(), tensor.dim),
+            _ => panic!("Can't transform to tensor."),
+        }
+    }
+
+    pub fn to_type(&self) -> Type {
+        match &self.ty {
+            ArgType::Tensor(tensor) => {
+                // Treat tensor with dim 0 as scalar
+                if tensor.dim == 0 {
```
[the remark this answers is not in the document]
antimora: Probably not because PyTorch/ONNX treats zero dim as scalar but burn treats tensor with 1 din and shape 1 as scalar. Unless we see a real example of ONNX with tensor.dim == 1 and shape = [1] , we should leave it as as it.

burn-import/src/onnx/to_burn.rs
```diff
@@ -186,64 +202,108 @@ impl ONNXGraph {
...
+            AttributeValue::Float32(val) => ConstantValue::Float32(val),
+            AttributeValue::Int64(val) => ConstantValue::Int64(val),
+            AttributeValue::Tensor(tensor) => {
+                // Treat zero dim tensor as scalar
+                if tensor.dim == 0 {
+                    match tensor.data.unwrap() {
+                        TensorData::Float32(val) => ConstantValue::Float32(val[0]),
+                        TensorData::Float64(val) => ConstantValue::Float32(val[0] as f32),
+                        TensorData::Int64(val) => ConstantValue::Int64(val[0]),
+                        TensorData::Int32(val) => ConstantValue::Int64(val[0] as i64),
```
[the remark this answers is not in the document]
antimora: OK. I have introduced Int32 and Float64. ONNX does not have these for scalars but it does for zero dim tensors. So I went ahead and introduced them.

burn-import/src/onnx/to_burn.rs
```diff
@@ -186,64 +202,108 @@ impl ONNXGraph {
...
-            AttributeValue::Int64s(val) => ConstantValue::Int(val[0] as i32),
-            _ => panic!("Unsupported constant node: {:?}", node),
+            AttributeValue::Float32(val) => ConstantValue::Float32(val),
+            AttributeValue::Int64(val) => ConstantValue::Int64(val),
+            AttributeValue::Tensor(tensor) => {
+                // Treat zero dim tensor as scalar
+                if tensor.dim == 0 {
+                    match tensor.data.unwrap() {
+                        TensorData::Float32(val) => ConstantValue::Float32(val[0]),
+                        TensorData::Float64(val) => ConstantValue::Float32(val[0] as f32),
```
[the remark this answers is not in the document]
antimora: OK. I have introduced Int32 and Float64. ONNX does not have these for scalars but it does for zero dim tensors. So I went ahead and introduced them.

burn-import/src/burn/node/conv2d.rs
antimora: Now the graph requires input and output information. Previously we were using the first node's inputs and last node's outputs as graph inputs and outputs but this is not true. Some middle node could also start needed inputs. I had to fix each test to add the input and output names.

burn-import/src/burn/node/unary.rs
antimora: Got it. I have added a todo for supporting to cast other types.

burn-import/src/onnx/to_burn.rs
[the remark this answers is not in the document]
nathanielsimard: Maybe we can add a comment on this to keep track on why we have casting here.

nathanielsimard (approved):

burn-import/src/onnx/to_burn.rs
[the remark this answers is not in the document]
antimora: Added the comment. Ready to merge =)

nathanielsimard (approved):

Issue #581, opened 2023-08-03
Improve error messages from problems when creating a virtual environment in Python for Burn.


nathanielsimard:
> If possible, please add code to initialize a "venv" environment and check if it is properly initialized to ensure the code runs smoothly. @mashirooooo

antimora:
I think related PR solves this issue.

PR #588, opened 2023-08-05
feat: check python venv environment initialized

mashirooooo:
### Checklist

- [ ] Confirm that `run-checks.sh` has been executed. (I'm not sure if this can be run on Windows. In the burn-dataset directory, I have manually tried "cargo test --all-features" which has the unrelated tests "dataset::sqlite::tests::sqlite_writer_write" and "dataset::sqlite::tests::sqlite_writer_write_multi_thread" fail, and tried "cargo doc --all-features" which works fine.)

### Related Issues/PRs

#575 

#581

### Changes

Check if both Python and the venv environment can be used successfully.

### Testing

The commands cargo run --example mnist --release --features ndarray and cargo run --example mnist --release --features wgpu have been tested

### DIFF
--- a/burn-dataset/src/source/huggingface/downloader.rs
+++ b/burn-dataset/src/source/huggingface/downloader.rs
@@ -227,10 +227,10 @@ fn check_python_version_is_3(python: &str) -> bool {
 
 /// get python3 name `python` `python3` or `py`
 fn get_python_name() -> Result<&'static str, ImporterError> {
-    let python_str_list = ["python", "python3", "py"];
-    for py in python_str_list.iter() {
-        if check_python_version_is_3(py) {
-            return Ok(py);
+    let python_name_list = ["python3", "python", "py"];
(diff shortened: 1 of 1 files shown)


burn-dataset/src/source/huggingface/downloader.rs
```diff
@@ -205,24 +244,32 @@ fn importer_script_path(base_dir: &Path) -> PathBuf {
 }
 
 fn install_python_deps(base_dir: &Path) -> Result<PathBuf, ImporterError> {
+    let python_name = get_python_name()?;
```
nathanielsimard, line 247: I don't think we need to call `get_python_name` if the venv is already initialized. Can we call this function only when creating a new venv?

mashirooooo: What you said is correct, and it has been changed now.

nathanielsimard (approved):

PR #593, opened 2023-08-06
Use buffered io for massive performance gains when loading and saving…

Gadersd:
… models

### Checklist

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/554

### Changes

The original model IO code did not use buffered IO which resulted in extremely slow model saving and loading for large files. I added buffered IO and found the saving and loading speed to be significantly improved. A once 8 minute load became a matter of seconds.

### Testing

I ran the run-tests.sh script to verify that there were no breaking changes and my own models loaded and ran without issue.

### DIFF
--- a/burn-core/src/record/file.rs
+++ b/burn-core/src/record/file.rs
@@ -2,8 +2,8 @@ use super::{bin_config, PrecisionSettings, Recorder, RecorderError};
 use core::marker::PhantomData;
 use flate2::{read::GzDecoder, write::GzEncoder, Compression};
 use serde::{de::DeserializeOwned, Serialize};
-use std::{fs::File, path::PathBuf};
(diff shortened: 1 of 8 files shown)


nathanielsimard (approved):
It's awesome that it took only a few lines of code to speed up that much file reading/loading. I though we were already using something similar to that with the streaming std API, but I guess the trait is implemented on different structs and the most performant ones to use are probably `BufReader` and `BufWriter`

nathanielsimard:
@Gadersd I think you have to run `cargo clippy --fix` followed by `cargo fmt --all` to fix the CI. Then we can merge.

Gadersd:
I just ran the formatting commands and commited.

PR #601, opened 2023-08-07
WGPU use best limits for the adaptor

nathanielsimard:


antimora (approved):
Looks good to me but I haven't tested on M1.

PR #603, opened 2023-08-07
Can configure wgpu max tasks

nathanielsimard:


antimora (changes requested):
Can we make this param passible instead of using environment param? We have a few options if you still want to use it as an env param. We could make it static and use it by builder through env param. 

The main reason I am bringing this up is to make no_std compatible and easy build for wasm.

antimora:
> We could use an `AtomicUsize` to accomplish the same thing. I just think that not having to recompile the project and play with an environment variable is more advantageous in this case. We could offer both ways of modifying this setting. @antimora, what do you think? Do you have any other suggestions?

I'm okay with not requiring compilation; it's simply the manner in which the value is passed that concerns me. It's analogous to passing a global variable but worse. Using global variables in the middle of code will inherit all associated problems (testing, debugging, etc).

It's acceptable if we use an environment variable from the place where we construct the server. It provides an opportunity to override the value from other sources. Perhaps we can even add a conditional feature flag for no_std compilation.

However, I acknowledge that your function interface will change for this parameter (and any downstream interfaces).

Maybe this is a good opportunity to think about some config structure for wgpu, as I am seeing many hardcoded values that might go through similar adjustments. If your interface already takes a config struct, you won't have to modify the function interfaces. What do you think?

PR #604, opened 2023-08-07
Feat/tensor casting

nathanielsimard:
Improve performance on tensor casting without having to convert tensors to data.


antimora (approved):
Looks good to me.

Issue #543, opened 2023-07-25
Implement missing Optimizers


louisfd:
Read the section on optimization before trying to implement an optimizer: https://github.com/burn-rs/burn/blob/main/ARCHITECTURE.md#optimization

- [X] SGD
- [X] Adam
- [x] AdamW
- [x] RMSProp
- [x] AdaGrad

AuruTus:
Hi, I'm interested in this and working on `RMSProp`. is there any other guy who has already started on it?

antimora:
@AuruTus go ahead and claim it. @wdoppenberg currently has an open PR https://github.com/burn-rs/burn/pull/566 for AdamW. You can see how it's been done.

wdoppenberg:
> @AuruTus go ahead and claim it. @wdoppenberg currently has an open PR #566 for AdamW. You can see how it's been done.

It's a WIP still, but I'd be happy to collaborate in some way if you're interested. Feel free to message me on discord.

CohenAriel:
Hello. I'd like to work on AdaGrad if there's no one already working on it.

antimora:
> Hello. I'd like to work on AdaGrad if there's no one already working on it.

go ahead ! happy coding.

AuruTus:
hi @CohenAriel , I noticed that the common `weight_decay` 's `transform` method has some bugs, so almost all optimizers will be affected. Can use temporary fix to make a workaround now. I think I can fix it later but there are still some bugs in my RMSProp now 😅.

The bug code and discussion can be refered here on [discord](https://discord.com/channels/1038839012602941528/1091796857996451942/1139895797958332426)

and this is my temp impl for ref:
```rust
    /// temp fix for Transform.
    pub fn transform_temp_fix<const D: usize>(
        &self,
        grad: Tensor<B, D>,
        tensor: Tensor<B, D>,
        state: Option<WeightDecayState<B, D>>,
    ) -> (Tensor<B, D>, Tensor<B, D>, WeightDecayState<B, D>) {
        let grad = match state {
            Some(state) => state.grad_last_step.mul_scalar(self.penalty).add(grad),
            None => tensor.clone().mul_scalar(self.penalty).add(grad),
        };

        (grad, tensor.clone(), WeightDecayState::new(tensor))
    }
```

CohenAriel:
Wow, thanks! I guess I can just not use weight decay in the test until it's fixed

AuruTus:
[...]

PR #607, opened 2023-08-08
Draft: Feat/RMSProp-optimizer

AuruTus:
### Checklist

### Related Issues/PRs

Solve `RMSProp` task of  #543

### Changes

implement the RMSProp optimizer referring to pyTorch

### Testing

TODO (still in draft)


AuruTus:
Just finish Rust code now. The numeric test (and other test like loading state) part is in draft and I'm going to write some PyTorch script against it.

wdoppenberg:
> Just finish Rust code now. The numeric test (and other test like loading state) part is in draft and I'm going to write some PyTorch script against it.

I modified the script I used for AdamW, should you want it:

```python
import torch
from torch.nn import Linear
from torch.optim import RMSprop

# Define a Linear layer with given weights and biases
def given_linear_layer(weights: torch.Tensor, biases: torch.Tensor) -> Linear:
	out_features, in_features = weights.shape
	layer = Linear(in_features, out_features, bias=True)

	layer.weight.data = torch.FloatTensor(weights)
	layer.bias.data = torch.FloatTensor(biases)

	return layer


def test_adamw():
	torch.set_printoptions(precision=6, sci_mode=True)

	LEARNING_RATE = 0.01

	prior_weights = torch.tensor([
			[-0.3206, 0.1374, 0.4043, 0.3200, 0.0859, 0.0671],
			[0.0777, -0.0185, -0.3667, 0.2550, 0.1955, -0.2922],
			[-0.0190, 0.0346, -0.2962, 0.2484, -0.2780, 0.3130],
			[-0.2980, -0.2214, -0.3715, -0.2981, -0.0761, 0.1626],
			[0.3300, -0.2182, 0.3717, -0.1729, 0.3796, -0.0304],
			[-0.0159, -0.0120, 0.1258, 0.1921, 0.0293, 0.3833],
		])
	prior_bias = torch.tensor([-0.3905, 0.0884, -0.0970, 0.1176, 0.1366, 0.0130])

	linear = given_linear_layer(
		torch.clone(prior_weights),
		torch.clone(prior_bias)
	)

	x_1 = torch.tensor([
		[0.6294, 0.0940, 0.8176, 0.8824, 0.5228, 0.4310],
		[0.7152, 0.9559, 0.7893, 0.5684, 0.5939, 0.8883],
	], requires_grad=True)

	x_2 = torch.tensor([
		[0.8491, 0.2108, 0.8939, 0.4433, 0.5527, 0.2528],
		[0.3270, 0.0412, 0.5538, 0.9605, 0.3195, 0.9085],
	], requires_grad=True)

	RMSPROP_KWARGS = {
		"lr": LEARNING_RATE,
		"alpha": 0.99,
		"eps": 1e-8,
		"weight_decay": 0.01,
		"momentum": 0.0,
		"centered": False,
	}

	optimizer = RMSprop(
		linear.parameters(),
		**RMSPROP_KWARGS
	)

	# First forward-backward pass and optimization step
	optimizer.zero_grad()
	y = linear(x_1)
	y.backward(torch.ones_like(y))
	optimizer.step()

	# Second forward-backward pass and optimization step
	optimizer.zero_grad()
	y = linear(x_2)
	y.backward(torch.ones_like(y))
	optimizer.step()

	# Print updated state
	print(f"\nUpdated layer weights: \n{linear.weight.data}")
	print(f"\nUpdated layer bias: \n{linear.bias.data}")

	# Print difference between prior and updated weights
	print(f"\nWeight difference: \n{linear.weight.data - prior_weights}")
	print(f"\nBias difference: \n{linear.bias.data - prior_bias}")
```

AuruTus:
The numeric test could be refactored with param arrays or the other same thing, but I keep it for later debug, for there are still some little parts to fix or refactor.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,528 @@
...
+    weight_decay: Option<WeightDecay<B>>,
+}
+
+/// State of [RMSProp](RMSProp)
+#[derive(Record, Clone, new)]
+pub struct RMSPropState<B: Backend, const D: usize> {
+    weight_decay: Option<WeightDecayState<B, D>>,
+    square_avg: Option<SquareAvgState<B, D>>,
+    centered: Option<CenteredState<B, D>>,
+    momentum: Option<RMSPropMomentumState<B, D>>,
```
nathanielsimard, line 82: Is it necessary to have those states wrapped into options?

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,528 @@
...
+                _ => grad.clone().mul_scalar(1. - alpha),
+            };
+            let (avg, square_avg) = match square_avg_state {
+                Some(state) => (
+                    state.square_avg.clone().sub(grad_avg.clone().powf(2.)),
+                    state.square_avg,
+                ),
+                _ => {
+                    panic!("uninitialized square_avg state")
+                }
```
nathanielsimard, line 148: We should avoid this by making `square_avg_state` not optinal.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,528 @@
...
+            };
+            (grad, Self { grad_avg, avg }, SquareAvgState { square_avg })
+        } else {
+            let grad_avg = Tensor::zeros(grad.shape());
+            let (avg, square_avg) = match square_avg_state {
+                Some(state) => (state.square_avg.clone(), state.square_avg),
+                _ => {
+                    panic!("uninitialized square_avg state")
+                }
```
nathanielsimard, line 157: We should avoid this by making `square_avg_state` not optinal.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,528 @@
...
+        RMSPropMomentumState<B, D>,
+    ) {
+        let (grad, centered_state) = match centered_state {
+            Some(state) => (
+                grad.clone()
+                    .div(state.avg.clone().sqrt().add_scalar(self.epsilon)),
+                state,
+            ),
+            _ => {
+                panic!("uninitialized centered state")
```
nathanielsimard, line 196: We should avoid this by making `centered_state` not optinal.

nathanielsimard (changes requested):
I think we can simplify things by removing options. I didn't review the Math, so I'll have to review it later.

burn-core/src/optim/rmsprop.rs
AuruTus: Oh, I didn't realize that I can shadow these intermediate variables in the `step` method before. I changed it now. And also add a match sentence to avoid always assigning zero tensors to `grad_avg` every time it calls the centered transform with `centered` is false.

burn-core/src/optim/rmsprop.rs
AuruTus: Done.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,500 @@
...
+    fn to_device(mut self, device: &B::Device) -> Self {
+        self.buf = self.buf.to_device(device);
+        self
+    }
+}
+
+impl<B: Backend> SimpleOptimizer<B> for RMSProp<B> {
+    type State<const D: usize> = RMSPropState<B, D>;
+
+    fn step<const D: usize>(
```
nathanielsimard, line 226: Very basic comment, but I would prefer to see this implementation first in the file bellow the struct definitions. Also a lot of functions don't need to be public, since they are only going to be used from the `step` method.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,500 @@
...
+            state_square_avg = Some(state.square_avg);
+            state_centered = Some(state.centered);
+            state_momentum = Some(state.momentum);
+        }
+
+        // weight_decay transform
+        if let Some(weight_decay) = &self.weight_decay {
+            let (grad_out, tensor_out) = weight_decay.transform_temp_fix(grad, tensor);
+            grad = grad_out;
+            tensor = tensor_out;
```
nathanielsimard, line 247: Looking at `transform_temp_fix` I think `tensor_out` is actually just `tensor`, so we don't need to set it here?

burn-core/src/optim/decay.rs
```diff
@@ -57,6 +57,16 @@ impl<B: Backend> WeightDecay<B> {
...
     }
+
+    /// temp fix for Transform.
+    pub fn transform_temp_fix<const D: usize>(
+        &self,
+        grad: Tensor<B, D>,
+        tensor: Tensor<B, D>,
+    ) -> (Tensor<B, D>, Tensor<B, D>) {
+        let grad = tensor.clone().mul_scalar(self.penalty).add(grad);
+        (grad, tensor.clone())
```
nathanielsimard, line 68: I think I miss read the PyTorch documentation when implementing the first version. `g_{t}` is actually calculated using `theta_{t-1}` not `theta_{t}`.

This is confusing since I expected the gradients of the forward pass to share the same timestep as the parameters that are used, but I don't make the rules 😅

So weight decay doesn't actually have a state, since the current tensor is available and that's why your new function works.

burn-core/src/optim/rmsprop.rs
nathanielsimard: If it's the case, I would only return the updated gradient from the function `transform_temp_fix`, not a tuple.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,500 @@
...
+        }
+
+        // square_avg transform
+        let (mut grad, state_square_avg) =
+            SquareAvgState::transform(self.alpha, grad, state_square_avg);
+
+        // centered trnsform
```
nathanielsimard, line 254: type `trnsform` => `transform`.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,500 @@
...
+        let (grad_out, state_square_avg, state_centered) = CenteredState::transform(
```
nathanielsimard, line 255: I think we should extract the `if centered` here instead of inside the `CenteredState::transform` function, since it's not its job to also calculate the other transform that doesn't have a state.

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,500 @@
...
+            self.momentum
+                .transform(grad, state_centered, state_momentum);
+
+        // transition state
+        let state = RMSPropState::new(state_square_avg, state_centered, state_momentum);
+
+        // tensor param transform
+        let delta = grad.mul_scalar(lr);
+        // (tensor - delta, Some(state))
+        let tmp = tensor - delta;
```
nathanielsimard, line 275: `tmp` can be `update` ?

burn-core/src/optim/rmsprop.rs
```diff
@@ -0,0 +1,500 @@
...
+                .square_avg
+                .clone()
+                .sub(grad_avg.clone().powf(2.));
+            (grad, square_avg_state, Self { grad_avg, avg })
+        } else {
+            let grad_avg = match centered_state {
+                Some(state) => state.grad_avg,
+                _ => Tensor::zeros(grad.shape()),
+            };
+            let avg = square_avg_state.square_avg.clone();
```
nathanielsimard, line 150: Having a `centered state` when `centered` is false doesn't seem correct. Maybe we should keep this state as optional 😅.

nathanielsimard (changes requested):
I looked at the math a bit more, let me know what you think!

burn-core/src/optim/rmsprop.rs
AuruTus: Got it. It looks better now. Will push it to the branch later.

burn-core/src/optim/rmsprop.rs
AuruTus: Emmm, but I need to pass a tensor's reference to this func then. I would thought to keep it looks like the other transform, just move in and move out.

But how about this style?

```rust
if let Some(weight_decay) = &self.weight_decay {
            (grad, tensor) = weight_decay.transform_temp_fix(grad, tensor);
        }
```
still move in, and move out values, but a bit simpler.

burn-core/src/optim/decay.rs
AuruTus: yeah, that's the point here. Actually there's no state for `weight_decay`, for the `tensor` is just the state. 😄

burn-core/src/optim/rmsprop.rs
AuruTus: sorry forget to remove the var for debugging 😅

burn-core/src/optim/rmsprop.rs
AuruTus: But actually the `avg` is always needed no matter what value the `centered` is, so there should be at least a `centered state`, IMO, to store the `avg`. And that's the question I want to get an advice if we should keep `grad_avg` or `centered state` in `Option` 🤔

burn-core/src/optim/rmsprop.rs
AuruTus: Can see the comment below. I would prefer keeping the `step` more simple and straightforward, because the `avg` <- `square_avg` can be placed within `centered transform`, which resembles the form of PyTorch.  (But they optimize it with a further step with square root value)

burn-core/src/optim/rmsprop.rs
AuruTus: Oh and the RMSProp's momentum also has this problem. When the `momentum` is less than or equals to 0, the state is useless too. I think this can be wrapped in option 🤔

AuruTus: emmm, I'm wondering if it's enough just to wrap these states with Option, because it will occupy the mem space too even it's None with a specefic Tensor type.

nathanielsimard: Buffers are already stored with a pointer, so you don't have to worry, the size of the Tensor struct is minimal.

burn-core/src/optim/rmsprop.rs
AuruTus: Done, only grad returned makes it simpler! 😄

burn-core/src/optim/rmsprop.rs
AuruTus: Make it now! I try to only wrap those fields in need.

AuruTus:
> I think it's ready to be merged 🎉 We will have to fix weights decay for all other optimizers, but this can be done in a following PR. Let me know @AuruTus if there is something else you want to do before merging.

Oh there are some static error in the test. Lemme fix it and then this part can be done. And we can fix all optimizers in a next PR.

nathanielsimard:
> Still cannot run `run-checks.sh` locally, for there are some tests failure in other crates like `dataset::sqlite::tests`? Just copy the style the other file uses. Hope it works. 👀

The CI script passes locally; the problem comes from a missing space on the device on the CI server. I'll merge.

PR #611, opened 2023-08-09
Global avg pool

antimora:
### Checklist

### Changes

Global avg pool

### Testing

Added new unit tests.


burn-import/src/onnx/from_onnx.rs
```diff
@@ -566,21 +566,16 @@ fn rename_inputs(
...
 
-    let mut counter: HashMap<String, usize> = HashMap::new();
-
     for node in nodes.iter_mut() {
-        // keep track of the number of nodes of each type
-        counter
-            .entry(node.name.clone())
-            .and_modify(|e| *e += 1)
-            .or_insert(1);
+        let mut counter = 1;
```
nathanielsimard, line 570: So the counter is now global instead of per name? Is there a reason for that other than simplicity?

antimora: No, the count is reset per each node. We had a bug for multiple outputs - the output names were not unique. I didn't come across this bug because we didn't have nodes with multiple outputs. Dropout can have multiple outputs. The second output is optional.

nathanielsimard (approved):

PR #613, opened 2023-08-09
Fix/wgpu/max pool2d backward

nathanielsimard:


nathanielsimard:
This should help with getting max_pool1d to work on wgpu: https://github.com/burn-rs/burn/pull/602

burn-wgpu/src/template/pool/max_pool2d_with_indices_backward.wgsl
```diff
@@ -72,10 +76,12 @@ fn main(
     for (var oh = oh_start; oh <= oh_end; oh++) {
         for (var ow = ow_start; ow <= ow_end; ow++) {
             let index = b * grad_stride_0 + c * grad_stride_1 + oh * grad_stride_2 + ow * grad_stride_3;
-            let index_max = u32(indices[index]);
+            if oh < grad_shape_2 && ow < grad_shape_3 {
```
louisfd, line 79: I found the following to be sliiightly faster 

Before the for loops:
`
let oh_max = min(oh_end, grad_shape_2 - 1u);
let ow_max = min(ow_end, grad_shape_3 - 1u);
`
For loops:
`for (var oh = oh_start; oh <= oh_max; oh++) {
            for (var ow = ow_start; ow <= ow_max; ow++) {
`
Then the if can be removed. 
The gain is just a few microseconds, but any gain is a good gain :)

louisfd (approved):
It's good. 
I have an improvement but it's for such a minor gain that I don't request it as mandatory

PR #625, opened 2023-08-10
Feature/burn book

wdoppenberg:
### Checklist

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/200

### Changes

There's (part of) a book now.


Luni-4:
@wdoppenberg 

Thanks a lot for this feature! 

We could add a `GithHub Actions` to automatically commits the `burn-book` into the blog repository, in this way it will be immediately accessible for anyone. What do you think?

wdoppenberg:
> @wdoppenberg
> 
> Thanks a lot for this feature!
> 
> We could add a `GithHub Actions` that automatically commits the `burn-book` into the blog repository, in this way it will be immediately accessible for anyone. What do you think?

Sounds good! Although maybe we'd need everyone to pitch in today and write something on each page, most important are the Quickstart and Installation pages. This was only a few minutes work so far, so having a bit more shouldn't take too much effort.

scripts/test_adamw.py
```diff
@@ -0,0 +1,72 @@
+import torch
```
nathanielsimard, line 1: I would remove that script, since it's not related to the current PR 😅

burn-book/src/zero-to-llama2.md
```diff
@@ -0,0 +1 @@
+# Zero to LLaMA2
```
nathanielsimard, line 1: I would remove that section for now, but I keep the idea, I will work next week on the book!

nathanielsimard (changes requested):
Thanks for laying down the book structure. It's mostly the readme reframe as a book, but it's a start. I will work on the book next week with a lot of additional content.

scripts/test_adamw.py
wdoppenberg: oops, that one slipped through

wdoppenberg:
> Thanks for laying down the book structure. It's mostly the readme reframe as a book, but it's a start. I will work on the book next week with a lot of additional content.

Yeah it was mostly to get the ball rolling. Everything that's in the book is copy-pasted from the readme atm. 

As far as examples go, maybe we could get @Gadersd to add some links to his excellent examples

PR #627, opened 2023-08-10
Add adagrad

CohenAriel:
### Checklist

### Related Issues/PRs

#543

### Changes

Add the [AdaGrad optimizer](https://pytorch.org/docs/stable/generated/torch.optim.Adagrad.html)

### Testing

I tested against the PyTorch implementation, however I couldn't get a better accuracy than 2. when testing the Adam implementation against PyTorch, I also couldn't get a better accuracy than 2, so I think the problem is with the test and not the AdaGrad implementation.


nathanielsimard:
@wdoppenberg  fixed the adam implementation, maybe you are facing the same problem, you can look at the fix https://github.com/burn-rs/burn/pull/621 (pretty simple).

CohenAriel:
I don't see anything that applies here.

I modified the script posted here #607 to get the numbers.

```python
import torch
from torch.nn import Linear
from torch.optim import Adagrad

# Define a Linear layer with given weights and biases
def given_linear_layer(weights: torch.Tensor, biases: torch.Tensor) -> Linear:
	out_features, in_features = weights.shape
	layer = Linear(in_features, out_features, bias=True)

	layer.weight.data = torch.FloatTensor(weights)
	layer.bias.data = torch.FloatTensor(biases)

	return layer


def test_adamw():
	torch.set_printoptions(precision=6)

	LEARNING_RATE = 0.01

	prior_weights = torch.tensor([
			[-0.3206, 0.1374, 0.4043, 0.3200, 0.0859, 0.0671],
			[0.0777, -0.0185, -0.3667, 0.2550, 0.1955, -0.2922],
			[-0.0190, 0.0346, -0.2962, 0.2484, -0.2780, 0.3130],
			[-0.2980, -0.2214, -0.3715, -0.2981, -0.0761, 0.1626],
			[0.3300, -0.2182, 0.3717, -0.1729, 0.3796, -0.0304],
			[-0.0159, -0.0120, 0.1258, 0.1921, 0.0293, 0.3833],
		])
	prior_bias = torch.tensor([-0.3905, 0.0884, -0.0970, 0.1176, 0.1366, 0.0130])

	linear = given_linear_layer(
		torch.clone(prior_weights),
		torch.clone(prior_bias)
	)

	x_1 = torch.tensor([
		[0.6294, 0.0940, 0.8176, 0.8824, 0.5228, 0.4310],
		[0.7152, 0.9559, 0.7893, 0.5684, 0.5939, 0.8883],
	], requires_grad=True)

	x_2 = torch.tensor([
		[0.8491, 0.2108, 0.8939, 0.4433, 0.5527, 0.2528],
		[0.3270, 0.0412, 0.5538, 0.9605, 0.3195, 0.9085],
	], requires_grad=True)

	Adagrad_KWARGS = {
		"lr": LEARNING_RATE,
		"eps": 1e-8,
		"lr_decay": 0.5,
		"weight_decay": 0.5,
	}

	optimizer = Adagrad(
		linear.parameters(),
		**Adagrad_KWARGS
	)

	# First forward-backward pass and optimization step
	optimizer.zero_grad()
	y = linear(x_1)
	y.backward(torch.ones_like(y))
	optimizer.step()

	# Second forward-backward pass and optimization step
	optimizer.zero_grad()
	y = linear(x_2)
	y.backward(torch.ones_like(y))
	optimizer.step()

	# Print updated state
	print(f"\nUpdated layer weights: \n{linear.weight.data}")
	print(f"\nUpdated layer bias: \n{linear.bias.data}")
```

Perhaps it has a problem?

CohenAriel:
Looks like without the weight decay the bias and the diagonal of the weights are correct, so there's probably a problem with the implementation

nathanielsimard (commented):
Everything looks good, however there maybe is a problem with how weight-decay is implemented, but we can resolve this in a following PR, since the problem touches almost all optimizers.

The only remaining part is to fix the CI. You should make sure that you are on the latest Rust stable, then run `cargo clippy --fix` and `cargo fmt --all`. I think it should solve the problem.

CohenAriel:
I have rustc 1.71.1, I ran both `cargo clippy --fix` and `cargo fmt --all` but it didn't change anything. Also, it compiles locally with no problems.

nathanielsimard:
@CohenAriel it seems we have a problem with the CI. I'll look into it.

CohenAriel:
:tada:

nathanielsimard (approved):

PR #616, opened 2023-08-09
Fixed Adam initialisation

wdoppenberg:
### Checklist

- [X] Confirm that `run-checks.sh` has been executed.

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/614
https://github.com/burn-rs/burn/pull/566 <-- would suggest to merge this first

### Changes

Changed the Adam test to have higher precision. See the related issue for more details. Updated `AdaptiveMomentumState` initialisation  (inside `transform`) to have $t_0 = 0$.

### Testing

See related issue

PR #630, opened 2023-08-11
bugfix: Revert last Adam bugfix

louisfd:
I'm reverting some changes done in this PR:
#616 

@wdoppenberg 
I thank you for enhancing the precision in Adam, but by changing the time of AdaptiveMomentumState from 1 to 0 some NaN appear (not in your test, strangely, but I made another one in which they appear). This made it impossible to learn with Adam. 
I'm changing it back to 1, which in my opinion makes sense if I look at the theory of Adam.  We never actually want the 0th time to be used in the moment correction phase. 
That being said, I could not figure why we do not exactly obtain the same values as torch. If you could look into that by finding another solution, that would be great.

### DIFF
--- a/burn-core/src/optim/adam.rs
+++ b/burn-core/src/optim/adam.rs
@@ -290,6 +290,45 @@ mod tests {
         weight_updated.assert_approx_eq(&weights_expected, ASSERT_PRECISION);
     }
 
+    #[test]
+    fn test_adam_optimizer_no_nan() {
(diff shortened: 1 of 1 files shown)


wdoppenberg:
I see. Apologies for not being more rigorous on the theory.

nathanielsimard (approved):
Merging this to fix main, but we should look into this more deeply.

PR #639, opened 2023-08-14
Changes and fixes required for squeezenet1.1-opset7 onnx

antimora:
### Checklist

### Changes

1. Add support for Dropout for opset 7.
2. Added AvgPool2d ONNX support.
3. Enhance Reshape to support dimension inference.
4. Other minor fixes.

### Testing

Created new unit tests and tested on `squeezenet1.1-opset7.onnx` file.


burn-import/src/onnx/op_configuration.rs
```diff
@@ -357,3 +392,31 @@ fn padding_config(pads: &[i64]) -> PaddingConfig2d {
...
+        match key.as_str() {
+            "allowzero" => attr_value_i64(value, &mut allowzero),
+            _ => {}
+        }
+    }
+
+    // Burn does not support zero size shape
+    if allowzero != 0 {
+        panic!("Zero shape size is not supported");
+    }
```
nathanielsimard, line 409: Not sure why allowzero is an int and not a bool. What am I missing here?

burn-import/src/onnx/op_configuration.rs
```diff
@@ -357,3 +392,31 @@ fn padding_config(pads: &[i64]) -> PaddingConfig2d {
...
+
+    let shape = match node.states.first() {
+        Some(state) => match &state.ty {
+            StateType::Tensor(tensor) => match tensor.data.as_ref() {
+                Some(TensorData::Int64(data)) => data.clone(),
```
nathanielsimard, line 414: Shape are stored as Int Tensor in ONNX right?

burn-tensor/src/tensor/api/base.rs
```diff
@@ -67,6 +67,60 @@ where
...
+    ///
+    ///    // Given a 3D tensor with dimensions (2, 3, 4), reshape it to (2, 12)
+    ///    let reshaped_tensor: Tensor::<B, 2> = tensor.reshape_infer([2, -1]);
+    ///
+    ///    // The resulting tensor will have dimensions (2, 12).
+    ///    println!("{:?}", reshaped_tensor.shape());
+    ///
+    /// }
+    ///
+    pub fn reshape_infer<const D2: usize>(self, shape: [i64; D2]) -> Tensor<B, D2, K> {
```
nathanielsimard, line 101: I don't think we should have another reshape method, but I agree that it would be convenient to support negative numbers. So, I propose to implement `From<[i64; D]> for Shape<D>` instead of adding a new function. This will have the benefit of a more uniform API.

Maybe it will clash with the current `From<[usize; D]> for Shape<D>` implementation, where users are forced to specify the array type (bad API). In that case, I propose to only keep the `From<i64>`.

nathanielsimard (changes requested):

burn-import/src/onnx/op_configuration.rs
antimora: This is from ONNX. In the spec, they use 0 or 1. I'll update with a documentation link. 

https://onnx.ai/onnx/operators/onnx__Reshape.html#attributes

burn-import/src/onnx/op_configuration.rs
antimora: Correct. See the ONNX spec (`shape` input): https://onnx.ai/onnx/operators/onnx__Reshape.html#inputs

burn-import/src/onnx/op_configuration.rs
nathanielsimard: Ho well 😅

burn-tensor/src/tensor/api/base.rs
antimora: I agree your suggestion is better. I had introduced reshape_infer to avoid compatibility with the existing usage but we found a way to accomplish without introducing any disruption. 

The interface is combined now into one method `reshape`. With your help, I introduced ReshapeArgs that serves as intermediate type that helps conversion from `[i32;D]`, `[usize;D]`, and Shape.

burn-tensor/src/tensor/api/check.rs
```diff
@@ -99,7 +99,7 @@ impl TensorCheck {
         check
     }
 
-    pub(crate) fn reshape_infer<const D: usize>(target: &[i64; D]) -> Self {
+    pub(crate) fn reshape_args<const D: usize>(target: &[i32; D]) -> Self {
```
nathanielsimard, line 102: I think I would rename to `reshape_args_i32` to highlight the fact that the check is for one variant of reshape arguments.

burn-tensor/src/tensor/api/check.rs
```diff
@@ -99,6 +99,30 @@ impl TensorCheck {
         check
     }
 
+    pub(crate) fn reshape_args<const D: usize>(target: &[i32; D]) -> Self {
+        let mut check = Self::Ok;
+
+        if target.iter().any(|&dim| dim < -1 || dim == 0) {
+            check = check.register(
+                "Reshape_infer",
```
nathanielsimard, line 107: Would still just put "Reshape" here for better error messages.

burn-tensor/src/tensor/api/check.rs
```diff
@@ -99,6 +99,30 @@ impl TensorCheck {
...
+                TensorError::new(
+                    "The given shape cannot contain negative dimensions (other than -1) and zero.",
+                )
+                .details(format!("Target shape: {:?}.", target)),
+            );
+        }
+
+        if target.iter().filter(|&x| x == &-1).count() > 1 {
+            check = check.register(
+                "Reshape_infer",
```
nathanielsimard, line 117: Same

nathanielsimard (approved):

PR #644, opened 2023-08-15
Change all APIs to accept `Into<Shape<D>>` + `$ cargo fmt --all`

wbrickner:
### Checklist

### Related Issues/PRs

Resolves https://github.com/burn-rs/burn/issues/634

### Changes

- Removes the need to do things like  `[1, 2, 3].into()` when providing tensor shapes (for users and contributors).
- Adds new implementations for making `Shape<D>`:

  ```
  &[i8],    &[i8;    D], Vec<i8>
  &[u8],    &[u8;    D], Vec<u8>
  &[i16],   &[i16;   D], Vec<i16>
  &[u16],   &[u16;   D], Vec<u16>
  &[i32],   &[i32;   D], Vec<i32>
  &[u32],   &[u32;   D], Vec<u32>
  &[i64],   &[i64;   D], Vec<i64>
  &[u64],   &[u64;   D], Vec<u64>
  &[isize], &[isize; D], Vec<isize>
  &[usize], &[usize; D], Vec<usize>
  ```
- Makes a couple of slight changes to make tests pass (removing `.into()` that is no longer needed, etc.)

The changes ended up being more extensive than I'd imagined, and I added some style in the choice for the new `From<T>`  implementations, so let me know if you disagree with them.

### Testing

All automated tests pass. There is no new logic to test.

### DIFF
--- a/burn-import/onnx-tests/tests/onnx_tests.rs
+++ b/burn-import/onnx-tests/tests/onnx_tests.rs
@@ -145,7 +145,7 @@ mod tests {
         // because the output tensor is too large to compare with the expected tensor.
         let output_sum = output.sum().into_scalar();
 
-        let expected_sum = -113.869_987; // from pytorch
+        let expected_sum = -113.869_99; // from pytorch
 
         assert!(expected_sum.approx_eq(output_sum, (1.0e-4, 2)));
     }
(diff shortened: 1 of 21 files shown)


burn-derive/src/config/analyzer_enum.rs
```diff
@@ -72,11 +72,11 @@ impl ConfigEnumAnalyzer {
...
         let variants = self.data.variants.iter().map(|variant| {
-            let variant_name = &variant.ident;
-            let (variant_input, variant_output) = self.gen_variant_field(variant);
+      let variant_name = &variant.ident;
+      let (variant_input, variant_output) = self.gen_variant_field(variant);
 
-            quote! { Self::#variant_name #variant_input => #enum_name::#variant_name #variant_output }
-        });
+      quote! { Self::#variant_name #variant_input => #enum_name::#variant_name #variant_output }
+    });
```
nathanielsimard, line 79: Formating problem?

nathanielsimard (changes requested):
I think there is a misconception here. Most traits in `burn-tensor` are targeted at backend developers and are called by concrete functions in `burn-tensor/src/api/*` using some intermediary traits to use the same struct `Tensor` for all tensor types, using static dispatch with the third generic argument (Kind).

This has the benefit of not having to import any traits to use any tensor operations. This is on those concrete functions that `impl Into<Shape<D>>` can be beneficial, and any concrete function in `burn-core`.

However, traits are designed to be easily implemented, not used. So with this in mind, having `impl Into<Shape<D>>` as input and being forced to call `into` isn't friendlier than directly receiving the shape.

So I think it's best to revert the changes on traits but keep them on concrete functions. Let me know your thoughts.

Issue #544, opened 2023-07-25
Training metrics


louisfd:
In burn-train, several metrics can be used during training. It would be great to have more!

- [X] Accuracy
- [X] Loss (the one in use)
- [X] CUDA utilization (memory&compute)
- [ ] Top-k accuracy
- [ ] CPU utilization
- [ ] CPU memory usage
- [ ] General GPU utilization
- [ ] General GPU memory usage
- [ ] Precision Recall
- [ ] AUC - ROC
- [ ] BLEU score
- [ ] ROUGE score

Elazrod56:
Hello, can I be assigned ? I'll deal with it

louisfd:
Of course! 😄

Elazrod56:
If I understood correctly, all Rust files that get metrics are located in burn-train/src/metric/

Just to check, is my `top_k_acc.rs` file in the right place ?
![issue_544_1](https://github.com/burn-rs/burn/assets/111651813/b03d3b38-5437-43dd-ae16-6d3aab4d3fb2)

louisfd:
Yes, right there :)

Elazrod56:
Just a question, by "CPU Memory usage", do you mean the RAM or something else ?

nathanielsimard:
Just the RAM, nothing fancy :)

Elazrod56:
I can't get myself to figure out how non-hardware metrics work, what they mean and how to implement them. Is that okay if I only implement the hardware metrics (CPU use, memory use, GPU use, GPU memory) ? I'll add temperature metrics for the CPU and the GPU to compensate my incapability to understand deep learning-specific metrics.

nathanielsimard:
Yes no problem, you can submit a PR with only one metric if you want!

oleksii-shyman:
Is this issue still relevant? I am interested to continue unfinished metrics.

I have the following questions:

1. I have AMD Radeon RX 5500 GPU on my machine. Is it good enough?
2. I am not really familiar with deep learning metrics, but I am willing to understand them. Would experienced dev be able to guide me in case I have questions?

nathanielsimard:
You can add metrics with any machine. To understand them, you can probably read some articles first like https://neptune.ai/blog/performance-metrics-in-machine-learning-complete-guide and then maybe use wikipedia for each individual metric you want to add.

duydl:
Hi, I’m just starting with Burn (and Rust) and I’m interested in this issue. I have a few questions I’d like to ask:

[...]

PR #647, opened 2023-08-16
Training metrics

Elazrod56:
### Checklist

### Related Issues/PRs

This PR partially solves issue #544, which asked for more training metrics.  

#### Metrics requested and successfully implemented
- General CPU utilization metric (use percentage)
- General memory (RAM + swap) utilization metric (use percentage)

#### Metrics requested but not implemented, and the reason why I couldn't implement it
- Top-k accuracy: I am not familiar with deep learning, and couldn't understand this metric properly
- General GPU utilization: My computer only has an AMD APU, no GPU.
- General GPU memory usage: My computer only has an AMD APU, no GPU.
- Precision Recall: I am not familiar with deep learning, and couldn't understand this metric properly
- AUC - ROC: I am not familiar with deep learning, and couldn't understand this metric properly
- BLEU score: I am not familiar with deep learning, and couldn't understand this metric properly
- ROUGE score: I am not familiar with deep learning, and couldn't understand this metric properly

#### Metrics not requested yet implemented
- CPU Temperature
- GPU Temperature

### Changes

In issue #544, it has been said that it would be great to have more training metrics. So I have added the metrics that I was skilled enough to add and that I could properly test on my hardware.

### Testing

I actually started developing metrics in separate cargo projects, and printed the data to the console.
To check the accuracy of this data, I have used a system tray that I have on my system (see attachment) which shows me hardware info (CPU use percentage, CPU temp, RAM and swap use percentage). Then I simply compared the data, and saw it was accurate.  
  
This is the system tray. From left to right: CPU use, CPU temp, RAM use (swap is displayed once I click it, date/time and battery)
system_tray_pr645  
  
Note: since my "GPU" is actually an APU, my GPU temp is roughly the same as my CPU temp, this is how I tested this metric.

Once I knew that my metric was accurate, I copied the code to BURN and adapted the Metric trait for my data type. And voilà, a new metric.  
I have mainly "copied" what I have seen in the CUDA metric, because I didn't know much about the BURN framework.

### DIFF
--- a/burn-train/src/metric/cpu_temp.rs
+++ b/burn-train/src/metric/cpu_temp.rs
@@ -28,16 +28,15 @@ impl Metric for CpuTemp {
 
     fn update(&mut self, _item: &Self::Input, _metadata: &MetricMetadata) -> MetricEntry {
         let sys = System::new();
-        let mut cpu_temp = CpuTemp::new();
 
         match sys.cpu_temp() {
-            Ok(temp) => cpu_temp.temp_celsius = temp,
-            Err(_) => cpu_temp.temp_celsius = f32::NAN,
+            Ok(temp) => self.temp_celsius = temp,
+            Err(_) => self.temp_celsius = f32::NAN,
         }
 
-        let formatted = format!("CPU Temp: {:.2}°C", cpu_temp.temp_celsius);
+        let formatted = format!("CPU Temp: {:.2}°C", self.temp_celsius);
 
-        let raw = format!("{:.2}", cpu_temp.temp_celsius);
+        let raw = format!("{:.2}", self.temp_celsius);
 
         MetricEntry::new(NAME.to_string(), formatted, raw)
     }
--- a/burn-train/src/metric/cpu_temp.rs
+++ b/burn-train/src/metric/cpu_temp.rs
@@ -1,0 +1,61 @@
+/// CPU Temperature metric
+
+use super::MetricMetadata;
+use crate::metric::{Metric, MetricEntry};
+use sysinfo::{ComponentExt, System, SystemExt, Component};
+
+static NAME: &str = "CPU_TEMP";
+
+/// CPU Temperature in celsius degrees
+pub struct CpuTemp {
+    temp_celsius: f32,
+}
+
+impl CpuTemp {
+    /// Creates a new CPU temp metric
+    pub fn new() -> Self {
+        Self {
+            temp_celsius: 0.
+        }
+    }
+}
+
+impl Default for CpuTemp {
+    fn default() -> Self {
+        CpuTemp::new()
+    }
+}
+
+impl Metric for CpuTemp {
+    type Input = ();
+
+    fn update(&mut self, _item: &Self::Input, _metadata: &MetricMetadata) -> MetricEntry {
+        let mut sys = System::new();
+
+        let mut cpu_temp = CpuTemp::new();
+
+        // refreshing components' info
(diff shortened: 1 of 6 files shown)


nathanielsimard (changes requested):
I think there are few possible improvements. Thanks for implementing those metrics, hardware usage is pretty useful information when training deep learning models!

Elazrod56:
Hi, I've addressed the changes requested. 
In cpu_temp.rs, I've decided to change crates because there was no way to accurately and reliably get the cpu temperature with sysinfo, so I went on with systemstat. I've kept sysinfo for the other metrics.

Let me know if there are any changes that are worth doing

antimora:
Formatting step in CI is failing.

Elazrod56:
What should I do ? `cargo fmt` ?

antimora:
> What should I do ? `cargo fmt` ?

To fix it, yes. If you have clippy errors, you can `cargo clippy --fix` as well.

Elazrod56:
Okay, I'll do it when I get back from work

burn-train/src/metric/cpu_temp.rs
```diff
@@ -0,0 +1,46 @@
...
+        CpuTemp::new()
+    }
+}
+
+impl Metric for CpuTemp {
+    type Input = ();
+
+    fn update(&mut self, _item: &Self::Input, _metadata: &MetricMetadata) -> MetricEntry {
+        let sys = System::new();
+        let mut cpu_temp = CpuTemp::new();
```
nathanielsimard, line 31: Instead of creating a new struct, we should just mutate the current one.

burn-train/src/metric/cpu_use.rs
```diff
@@ -0,0 +1,62 @@
...
+        CpuUse::new()
+    }
+}
+
+impl Metric for CpuUse {
+    type Input = ();
+
+    fn update(&mut self, _item: &Self::Input, _metadata: &MetricMetadata) -> MetricEntry {
+        let mut sys = System::new();
+        let mut cpu_use = CpuUse::new();
```
nathanielsimard, line 31: Instead of creating a new struct, we should just mutate the current one.

burn-train/src/metric/gpu_temp.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+impl Metric for GpuTemp {
+    type Input = ();
+
+    fn update(&mut self, _item: &Self::Input, _metadata: &MetricMetadata) -> MetricEntry {
+        let mut sys = System::new();
+
+        let mut gpu_temp = GpuTemp::new();
```
nathanielsimard, line 32: Instead of creating a new struct, we should just mutate the current one.

burn-train/src/metric/memory_use.rs
```diff
@@ -0,0 +1,73 @@
...
+        MemoryUse::new()
+    }
+}
+
+impl Metric for MemoryUse {
+    type Input = ();
+
+    fn update(&mut self, _item: &Self::Input, _metadata: &MetricMetadata) -> MetricEntry {
+        let mut sys = System::new();
+        let mut memory_use = MemoryUse::new();
```
nathanielsimard, line 39: Instead of creating a new struct, we should just mutate the current one.

nathanielsimard (changes requested):
I have the same comment on all metrics, when it's fixed I think we can merge.

nathanielsimard (approved):

PR #650, opened 2023-08-16
Feat/candle/initialize

louisfd:
I made the skeleton for the new [Candle](https://github.com/huggingface/candle) backend. 
_Add_ operation, along with basics such as from_data, is done. 

Next I will fill every mandatory operation 😄


burn-candle/src/backend.rs
```diff
@@ -0,0 +1,81 @@
...
+pub struct CandleBackend<F: FloatCandleElement, I: IntCandleElement> {
+    _float: PhantomData<F>,
+    _int: PhantomData<I>,
+}
+
+impl<F: FloatCandleElement, I: IntCandleElement> Backend for CandleBackend<F, I> {
+    type Device = CandleDevice;
+
+    type FullPrecisionBackend = CandleBackend<Self::FullPrecisionElem, Self::IntElem>;
+    type FullPrecisionElem = f64;
```
nathanielsimard, line 60: Full precision is normally f32, f64 is double precision.

nathanielsimard (approved):
A tiny change, but approve in advance.

nathanielsimard (approved):

LaurentMazare:
Wow, super impressive to see this going up that quickly! Thanks for putting this up! Definitely let us know if there is anything that we can do on the candle side to help or if you have any sort of feedback. Fwiw candle has a lot to be optimized yet, e.g. quantized tensors got a speedup of ~5x recently, accelerate on mac was just added - also a speedup of up to 5x, and there is a ton more to come in the next few weeks. It definetly helps if you report things being slow or even broken as we could prioritize these more heavily!

nathanielsimard:
@LaurentMazare, we have a backend-agnostic set of tests that we intend to execute using Candle (https://github.com/burn-rs/burn/tree/main/burn-tensor/src/tests). Furthermore, we will be comparing the performance of the Candle backend with other backends, including `tch`, using the benchmarking tool we've developed (https://github.com/burn-rs/burn-benches).

If we face any issues or bugs, we'll let you know :)

LaurentMazare:
Great, just to mention that for performance you may want to use the following features:
- `mkl` on intel platforms.
- `accelerate` on macOS.
- `cudnn` on nvidia GPUs - and `flash-attn` when using transformers architectures.
Even with that, I would expect candle to be quite crushed by tch at the moment but certainly looking forward to seeing some numbers (we also have to set up some benchmarking monitoring on our side). Hopefully this will get better over time (and the easiness of deployment of candle compared to tch/libtorch will still make it a good option).

PR #653, opened 2023-08-16
Feat/avg pool/include pad config

nathanielsimard:
Fix https://github.com/burn-rs/burn/issues/636

Support all avg pooling operations on all backends with autodiff.


louisfd (approved):
Wow that was a lot of count_include_pad
Looks fine to me!

PR #661, opened 2023-08-17
Onnx tests and bug fixes - BatchNorm, Identity, Relu, Sigmoid and Transpose

antimora:
### Checklist

### Changes

1. Add batch normalization ONNX test
2. Add support for Identity (this particular batch norm produced identities to be used to init input values for batch norm params)
3. Add Relu test
4. Add Sigmoid test
5. Add Transpose test
6. Add tests for graph int tensor inputs and outputs

### Testing

Ran the newly added tests.

### Related Issue

1. Fixes https://github.com/burn-rs/burn/issues/557
2. Fixes https://github.com/burn-rs/burn/issues/331


burn-import/onnx-tests/tests/transpose/transpose.py
```diff
@@ -0,0 +1,46 @@
...
+                      verbose=False, opset_version=16)
+
+    print("Finished exporting model to {}".format(file_name))
+
+    # Output some test data for use in the test
+    print("Test input data: {}".format(test_input))
+    print("Test input data shape: {}".format(test_input.shape))
+    output = model.forward(test_input)
+    print("Test output data shape: {}".format(output.shape))
+    print("Test output data: {}".format(output))
```
nathanielsimard, line 42: This is starting to become repetitive. Probably not in this PR, but I think it would be a good idea to add a utility Python module that we could import into each script.

burn-import/src/onnx/dim_inference.rs
```diff
@@ -247,13 +252,18 @@ fn unsqueeze_update_outputs(node: &mut Node) {
     }
 
     let node_input = &mut node.inputs[0];
-    let dim = match node_input.clone().ty {
-        ArgType::Tensor(tensor) => tensor.dim,
-        ArgType::Shape(dim) => dim,
+    let (dim, elem_type) = match node_input.clone().ty {
+        ArgType::Tensor(tensor) => (tensor.dim, tensor.elem_type),
+        ArgType::Shape(dim) => (dim, ElementType::Int64), // TODO: check if this is correct
```
nathanielsimard, line 257: Can we remove the TODO before merging (can we validate that?)

burn-import/src/onnx/dim_inference.rs
```diff
@@ -270,13 +280,31 @@ fn slice_update_outputs(node: &mut Node) {
...
+            // if the input is a tensor, the output is a tensor of bool
+            node.outputs[0].ty = ArgType::Tensor(Tensor {
+                elem_type: ElementType::Bool,
+                ..tensor
+            });
+        }
+        ArgType::Scalar(_) => {
+            node.outputs[0].ty = ArgType::Scalar(ElementType::Bool);
+        }
+        _ => panic!("Only tensor input is valid"),
```
nathanielsimard, line 304: Looking at the code, it seems like Scalar is also accepted?

burn-import/src/onnx/from_onnx.rs
```diff
@@ -534,6 +558,58 @@ fn lift_constants(nodes: &mut Vec<Node>) {
...
+        } else {
+            // Support pass through identity node
+            identity_nodes_to_remove.push(node.clone());
+        }
+    }
+
+    // Remove the identity nodes that are only used to connect two nodes
+    for identity_node in identity_nodes_to_remove.iter() {
+        let input = identity_node.inputs.first().unwrap();
+        let output = identity_node.outputs.first().unwrap();
```
nathanielsimard, line 596: I think we should replace the unwrap with an expect here: `expect("Identity node should have one input.")` and `expect("Identity node should have one output.")`

burn-import/src/onnx/from_onnx.rs
```diff
@@ -534,6 +558,58 @@ fn lift_constants(nodes: &mut Vec<Node>) {
...
+
+        // find the node that uses the identity node's output
+        for node in nodes.iter_mut() {
+            // let matched_input = node.inputs.iter().find(|x| x.name == output.name);
```
nathanielsimard, line 600: dead code

burn-import/onnx-tests/tests/transpose/transpose.py
antimora: Sure. It seems like a good idea.

nathanielsimard: Yeah I agree with code duplication. I think we could isolate setting the seed, setting the device, setting the printing precision and exporting the model to ONNX into a Python module, so that we can update those based on PyTorch releases. The remaining isn't generic enough to be shared.

burn-import/src/onnx/dim_inference.rs
antimora: yes, it's incorrect message. It should say only scalar is accepted for now.

nathanielsimard (approved):

PR #666, opened 2023-08-20
Remove channel field from maxpool1d config struct

bayedieng:
### Checklist

### Changes
This PR removes the channels field from the ```MaxPool1dConfig``` struct as it is not necessary to perform the operation

### Testing

Ran the checks which includes ```max_pool1d``` testing.


antimora (approved):
Thank you for catching and submitting this PR.

nathanielsimard (approved):

PR #668, opened 2023-08-20
Dilation maxpool

caiopiccirillo:
### Checklist

### Related Issues/PRs

All subtasks of https://github.com/burn-rs/burn/issues/622, which includes:

* Update MaxPool1d
* Update MaxPool2d
* Update burn-import

### Changes

Add dilation feature for maxpool.

### Testing

Unit tests added and double checked with Pytorch scripts.


burn-import/onnx-tests/tests/maxpool2d/maxpool2d.py
```diff
@@ -10,8 +10,7 @@ class Model(nn.Module):
     def __init__(self):
         super(Model, self).__init__()
 
-        # TODO support dilation=(3, 1) (see https://github.com/burn-rs/burn/issues/622)
-        self.maxpool2d1 = nn.MaxPool2d((4, 2), stride=(2, 1), padding=(2, 1))
+        self.maxpool2d1 = nn.MaxPool2d((4, 2), stride=(2, 1), padding=(2, 1), dilation=(3, 1))
```
antimora, line 13: Thanks for fixing burn-import as well. Could you please run this script, which will update the onnx file in the same directory? You also need to copy the test output after running the script and use the values in https://github.com/burn-rs/burn/blob/6b5ba770847ae0b4a2dda6bc584415c6c8008270/burn-import/onnx-tests/tests/onnx_tests.rs#L273-L274 . You can test by running the following in the same directory:

```shell
cargo test maxpool2d
```

caiopiccirillo: yeah sure, will do 😄

caiopiccirillo:
Following the PyTorch documentation I saw some GIFs, which they're pretty cool: [dilation](https://github.com/vdumoulin/conv_arithmetic/blob/master/gif/dilation.gif)

Can we add this kind of graphical example somewhere?

burn-import/onnx-tests/tests/maxpool2d/maxpool2d.py
caiopiccirillo: I had to use `dilation=(1, 3)` otherwise an error was thrown. Then I updated all files as you asked.

antimora: Yes, as long as we use something non-default, that's good. Thank you.

nathanielsimard (approved):

PR #669, opened 2023-08-21
support broadcast matmul

louisfd:
Very small PR in which candle dependancy is updated to support broadcast matmul


antimora (approved):
LGTM

nathanielsimard (approved):

PR #676, opened 2023-08-22
Perf/Empty Context Cache

agelas:
### Checklist

-  Currently on Windows and ran into this issue: https://github.com/rust-lang/cargo/issues/11544, but ran `clippy` and `cargo fmt`

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/619

### Changes

@louisfd This is a pretty brute-force method that relies on a potentially flawed assumption that I'd love to get your thoughts on. What I did is add an `is_tuning` flag to the `Context` that gets turned on by the `Tuner`'s `tune()` method, which tells the `Context` to explicitly tag the kernel that's being executed with a count each execution. After the benchmarks are run, the `Tuner` turns on another flag, `optimize_cache`, that signals to the `Context` that the next kernel that is run is the "winning" kernel. The latter part is where I'm a bit concerned. I assume that the **only** ComputePipeline used post-tuning is the winner, so the next execution will serve as confirmation of it being the fastest and all others can be cleared. 

I also considered cache timestamping or more direct Tuner-Context communication, but I wanted to refrain from additional cache management overhead/tighter coupling between the Tuner and Context. 

### Testing

TBD

### DIFF
--- a/burn-wgpu/src/context/base.rs
+++ b/burn-wgpu/src/context/base.rs
@@ -32,7 +32,6 @@ pub struct Context {
     device_wgpu: Arc<wgpu::Device>,
     cache: Mutex<HashMap<TemplateKey, (Arc<ComputePipeline>, usize)>>,
     is_tuning: Mutex<bool>,
-    pipeline_counter: Mutex<HashMap<TemplateKey, usize>>,
     client: ContextClientImpl,
     pub(crate) tuner: Tuner,
     pub(crate) device: WgpuDevice,
@@ -74,7 +73,6 @@ impl Context {
             client,
             cache: Mutex::new(HashMap::new()),
             is_tuning: false.into(),
-            pipeline_counter: Mutex::new(HashMap::new()),
             tuner: Tuner::new(),
             info,
         }
@@ -258,7 +256,6 @@ impl Context {
 
     fn retain_best_kernel(&self) {
         let mut cache = self.cache.lock();
-        let counter = self.pipeline_counter.lock();
         if cache.len() == 1 {
             return;
         }
@@ -268,7 +265,7 @@ impl Context {
             .max_by(|a, b| {
                 cache
                     .get(a)
-                    .map(|(_, count_a)| count_a) 
+                    .map(|(_, count_a)| count_a)
                     .unwrap_or(&0)
                     .cmp(&cache.get(b).map(|(_, count_b)| count_b).unwrap_or(&0))
             })
--- a/burn-wgpu/src/context/base.rs
+++ b/burn-wgpu/src/context/base.rs
@@ -107,16 +107,6 @@ impl Context {
         pipeline: Arc<ComputePipeline>,
         buffers: &[&Buffer],
     ) {
-        if !self.is_tuning.load(Ordering::Relaxed) && !self.tuning_template_ids.lock().is_empty() {
-            // clean cache of pipelines accumulated during tuning
-            let mut cache = self.cache.lock();
-            let mut tuning_template_ids = self.tuning_template_ids.lock();
-            for template_id in tuning_template_ids.iter() {
-                cache.remove(template_id);
-            }
-            tuning_template_ids.clear();
-        }
-
         let group_layout = pipeline.get_bind_group_layout(0);
 
         let entries = buffers
@@ -257,6 +247,15 @@ impl Context {
 
     pub fn stop_tuning(&self) {
         self.is_tuning.store(false, Ordering::Relaxed);
+
+        // clean cache of pipelines accumulated during tuning
+        let mut cache = self.cache.lock();
+        let mut tuning_template_ids = self.tuning_template_ids.lock();
+        for template_id in tuning_template_ids.iter() {
+            cache.remove(template_id);
+        }
+
+        tuning_template_ids.clear();
     }
 }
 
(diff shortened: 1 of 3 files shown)


louisfd (changes requested):
Hi @agelas, 
The major issue is that the context is shared for any type of operation running on the same device, therefore if you run let's say operation1, then operation2, then operation1 again, operation2 will have deleted the operation1 kernel from the cache. Clearing the whole cache is not an option. 
I think your count system interesting but it would have to be operation-wise. Keep in mind also that we may have several instances of the same operation at once, for different settings. For instance the best matmul kernel on 4x4 matrices is usually not the same kernel as the best one on 4096x4096 matrices. 
Timestamping also makes sense at first glance, but I think it would lead to concurrency problems. 
While I appreciate your effort of decoupling the tuner and the context, I'm starting to reconsider if it is really the way to go. Because in some sense, both cache should store the same kernels (with also non-autotuned ones in the context cache) as the winning kernels of autotune are precisely the ones that will be re-executed. I'd like to get @nathanielsimard 's input on that.

agelas:
@louisfd Thanks for the feedback and clarification- I think I wasn't understanding that the context acts on a per-device basis.

> Because in some sense, both cache should store the same kernels (with also non-autotuned ones in the context cache) as the winning kernels of autotune are precisely the ones that will be re-executed.

If I'm interpreting this correctly, do you mean that the Tuner should (in an ideal world) be operating on the Context's cache instead of having it's own, separate, cache which stores the same kernels but hashed differently? That makes sense to me because it seems like there's extra work being done if we have both `AutotuneKey` and `TemplateKey` describing essentially the same thing albeit at different levels of abstraction.

louisfd:
You're right, if the tuner cache caches a function, this function should also be cached by the context. But as you said it is on different levels of abstraction, and I think we want to keep it this way, especially in the perspective of using the tuner with other backends in the future.

agelas:
Just thinking out loud, but is there any reason there can't be a unified cache system that both the `Tuner` and `Context` can access? This could be indexed by some combination of `AutotuneKey` and `TemplateKey` attributes. There could be some cache manager handling all this that abstracts away the complexity of needing to deal with two caching systems, and also provides a consistent view of cached functions for both the `Tuner` and `Context`.

louisfd:
To tell the truth, the context and tuner cache systems were made at different points in time, and for different purposes. The context cache works under the assumptions that if a kernel is used once, it is probably used even more times afterwards. That was true until the addition of autotune, in which we specifically run many different kernels only once for benchmarking. This addition had the collateral effect of polluting the first cache. 
Also, like I said earlier, the context is specific to the wgpu backend, while we wanted autotune to be easily ported to other backends in some near future. 
So the first reason they are not unified is historical, and the second is about future plans... I do not think there is a fundamental reason not to at least try to have them merged.

nathanielsimard:
@louisfd @agelas I think the cache should be deleted just after selecting the candidate.

```rust
let candidates_to_try: ...;
let (candidate_best, candidates_other) = find_best(...);

for candidate in candidates_other {
    cache.clean(candidate.template_id());
}
```

We probably need to make available the template_key for each candidates, but it's probably very simple.

agelas:
@nathanielsimard What I have currently theoretically accomplishes what you outlined, but as to @louisfd's point this won't preserve all of the best pipelines per operation across the Context's cache.

nathanielsimard:
@agelas @louisfd  What I had in mind was to modify `KernelFunction` and `tuner.tune` to cleanup after the benchmarks. However it's not that simple. I think your approach with `Mutex<bool>` can be simpler. In fact, we can just not cache compute pipeline when `is_tuning` is true, simple as that! And I would probably use an atomic bool instead of a mutex.

nathanielsimard:
@agelas I just thought it might cause a problem, since benchmarks run multiple times for each kernel (excluding the first invocation) to accurately measure the execution time. If we don't cache kernels, then benchmarks will also measure the compile time, which is a big problem.

So following the current implementation, I think we should always cache compute pipelines, but when tuning, add template ids into a vector.

When tuning is false, we should clean up the list of template ids collected during tuning. We may recompile the chosen kernels multiple times, but it doesn't matter much at this point. The simplicity of the solution is more important than reducing the compilation overhead of chosen kernels one time during tuning. I plan to cache benchmark results, so it won't be impactful in this case.

agelas:
@nathanielsimard Gotcha, I've made the changes.

burn-wgpu/src/context/base.rs
```diff
@@ -97,6 +107,16 @@ impl Context {
...
+        if !self.is_tuning.load(Ordering::Relaxed) && !self.tuning_template_ids.lock().is_empty() {
+            // clean cache of pipelines accumulated during tuning
+            let mut cache = self.cache.lock();
+            let mut tuning_template_ids = self.tuning_template_ids.lock();
+            for template_id in tuning_template_ids.iter() {
+                cache.remove(template_id);
+            }
+            tuning_template_ids.clear();
+        }
+
```
nathanielsimard, line 119: Looking at the code, I think we can avoid locking when we aren't tuning; we can move this code in the `stop_tuning` function, just cleaning the cache right after tuning!

nathanielsimard (changes requested):
A small change, but otherwise pretty good :) thanks

burn-wgpu/src/context/base.rs
agelas: Good point, I'll make the fix.

nathanielsimard (approved):
Thanks for this, we finally converge to a solution 😅

louisfd (approved):
That started as a complicated issue, but we ended with something rather elegant, thanks a lot @agelas !

PR #682, opened 2023-08-23
Feat/candle/more operations

louisfd:
Added more operations to candle backend.


burn-candle/src/ops/base.rs
```diff
@@ -11,3 +20,84 @@ pub type IntElem<B> = <B as Backend>::IntElem;
 pub type IntTensor<B, const D: usize> = <B as Backend>::IntTensorPrimitive<D>;
 
 pub type BoolTensor<B, const D: usize> = <B as Backend>::BoolTensorPrimitive<D>;
+
+pub fn cat<E: CandleElement, const D: usize>(
+    tensors: Vec<CandleTensor<E, D>>,
+    dim: usize,
+) -> CandleTensor<E, D> {
+    let tensors: Vec<candle_core::Tensor> = tensors.iter().map(|t| t.clone().tensor).collect();
```
nathanielsimard, line 28: Use `into_iter` instead of `iter` here to avoid cloning.

```rust
tensors.into_iter().map(|t| t.tensor).collect();
```

burn-candle/src/ops/tensor.rs
```diff
@@ -22,39 +22,60 @@ impl<F: FloatCandleElement, I: IntCandleElement> TensorOps<CandleBackend<F, I>>
...
+                    .unwrap()
+                    .to_dtype(F::DTYPE)
+                    .unwrap(),
+            ),
+            Distribution::Bernoulli(prob) => CandleTensor::new(
+                candle_core::Tensor::rand(0., 1., shape, device)
+                    .unwrap()
+                    .gt(
+                        &(candle_core::Tensor::ones(shape, F::DTYPE, device).unwrap() * prob)
+                            .unwrap(),
```
nathanielsimard, line 39: We can use broadcasting to make it more efficient?

nathanielsimard (changes requested):
Just check where we can avoid creating tensors filled with a single value vs use broadcasting. Still, very good job 🎉

nathanielsimard (approved):

Issue #600, opened 2023-08-07
Support missing ONNX ops to import `optimum/all-MiniLM-L6-v2`


antimora:
[llogiq](https://www.reddit.com/user/llogiq/) on Reddit [requested support of missing ONNX Ops](https://www.reddit.com/r/rust/comments/15jc09o/comment/jv4ol0h/?utm_source=share&utm_medium=web2x&context=3). We are filing this issue ticket to prioritize these ops

The model: https://huggingface.co/optimum/all-MiniLM-L6-v2/blob/main/model.onnx

The ops used in this model (checked if supported): 

- [x] Add
- [x] Cast
- [x] Concat
- [x] Constant
- [x] Div
- [x] Erf
- [x] Gather
- [x] Gemm
- [x] MatMul
- [x] Mul
- [x] Pow
- [x] ReduceMean
- [x] Reshape
- [x] Shape
- [x] Slice
- [x] Softmax
- [x] Sqrt
- [x] Sub
- [x] Tanh
- [x] Transpose
- [x] Unsqueeze

All these ops are implemented in Burn and they now have to be supported by burn-import's op.

ShivangRawat30:
Hey I am currently learning rust, as it is a good first issue can you please assign this to me

antimora:
@ShivangRawat30, I would recommend you start with a unary operator Sqrt. Work on this and submit a PR. If you agree, just comment claim Sqrt. 

If you have questions and faster response, you can join Discord (you can find the link on Readme). 

Anyone else wants to work on any of the ops, you can just comment claim (some operator).

ShivangRawat30:
claim Sqrt

AuruTus:
👀  Hi I'm interested in this too. But I'm a bit confused if the new import file should be put under `burn-import/src/burn/node/`, and the burn impl in `burn-tensor` crate is what we should refer to, right? CMIIW 😃

antimora:
@AuruTus, since all these ops already available in Burn, they have to be implemented in burn-import. I would recommend to start with something simple. Maybe with Tanh?

AuruTus:
Thank you for the advice! Will look at it later.

claim Tanh

AuruTus:
I have free time now and can finish more. Claim erf.

antimora:
@AuruTus we added onnx file base testing to verify end to end conversion. Let me know if you have questions.

jmintb:
Claim Pow

CohenAriel:
[...]

PR #683, opened 2023-08-24
Feat/onnx import tanh

AuruTus:
### Checklist

### Related Issues/PRs

Solve tanh task #600 

### Changes

add new unary node `tanh` to import ONNX model.

### Testing

add new test entry under `tests/onnx_tests.rs` and pass.


burn-import/onnx-tests/tests/tanh/tanh.py
```diff
@@ -0,0 +1,40 @@
...
+class Model(nn.Module):
+    def __init__(self):
+        super(Model, self).__init__()
+
+    def forward(self, x):
+        return torch.tanh(x)
+
+
+def main():
+
```
antimora, line 18: Can we add rand.seed for reproducibility like this:

```python
    # Set random seed for reproducibility
    torch.manual_seed(0)
```

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -69,3 +69,8 @@ pub fn log_sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B
 pub fn silu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
     tensor.clone().mul(sigmoid(tensor))
 }
+
+/// Applies the tanh function
+pub fn tanh<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
+    tensor.tanh()
+}
```
antimora, line 76: Since this is brand new activation function, we will need a unit test similar to silu: https://github.com/antimora/burn/blob/828ef56cb381b57f0e6581117ba39cff101223a0/burn-tensor/src/tests/activation/silu.rs#L2 

To add a test: 
1. add a file under tests/activation
2. make sure to put a macro with a correct name: #[burn_tensor_testgen::testgen(tanh)] 
3. add to mod.rs like this: https://github.com/antimora/burn/blob/828ef56cb381b57f0e6581117ba39cff101223a0/burn-tensor/src/tests/activation/mod.rs#L4
4. add a module call like this: https://github.com/antimora/burn/blob/828ef56cb381b57f0e6581117ba39cff101223a0/burn-tensor/src/tests/mod.rs#L15


To test the activation function, run `cargo test` under `burn-ndarray` (or other backends). Tensor tests run with a backend.

antimora (changes requested):
Thank you so much for adding this op. There are a couple straight forward additions I am requesting. Please see my inline comments.

burn-tensor/src/tensor/activation/base.rs
AuruTus: Really thank you for the guides. Will complete this part later.

burn-import/onnx-tests/tests/tanh/tanh.py
AuruTus: Done with it.

burn-tensor/src/tensor/activation/base.rs
AuruTus: Oh I just noticed that there's already a unit test for `tanh` but under `ops` dir. Should we keep this? Or just move it under `activation`, for they actually have the same test data and logic.

nathanielsimard: I don't think the test is necessary, since it is already tested as a tensor operation. However adding this alias can be more intuitive.

nathanielsimard (changes requested):
I would remove the duplicated test, but other than that LGTM

burn-tensor/src/tensor/activation/base.rs
AuruTus: I think we can keep the new activation test to stay consistent with other activation funcs. It's indeed more intuitive for new maintainers to find calling chain and understand the project struct. And these test snippet are not huge burdens for `tensor` code base. Though the long name and redundancy are really ugly. 😅

antimora: I would leave it. For any public API, we should always have tests. It just happens that the underlying code/function is already tested but there is no guaranty unless we have tests. Also have tests will ensure confidence if we need to do a refactor.

AuruTus: Ok it should exist there. We can keep both.

nathanielsimard (approved):

antimora (approved):
Looks good. If you have a chance to update the readme file by checking off Tanh, it'd be great. 

Approving ahead.

PR #684, opened 2023-08-24
Do not use default device when running kernel::matmul::tune

jeromerobert:
### Checklist

`run-checks.sh` not runned because of error downloading pytorch (sorry).

### Changes

Do not use default device when running kernel::matmul::tune. Use the device of the involved Tensor instead of Device::default.

Without this change the bellow program fail on my hardware because my discrete card is too old and not fully Vulkan compatible.

### Testing
Cargo.toml:

```toml
burn-wgpu = { path = "/path/to/burn/burn-wgpu", features=["autotune"] }
```

Code sniped
```
type Backend = WgpuBackend<Vulkan, f32, i32>;
let device = WgpuDevice::Cpu;
let ta: Tensor<Backend, 2, Float> = Tensor::from_data_device(burn_tensor::Data {
    value: vec![0.],
    shape: burn_tensor::Shape::new([1,1]),
}, &device);
let tb = Tensor::from_data_device(burn_tensor::Data {
    value: vec![0.],
    shape: burn_tensor::Shape::new([1,1]),
}, &device);
ta.matmul(tb);
```

### DIFF
--- a/burn-wgpu/src/kernel/matmul/tune.rs
+++ b/burn-wgpu/src/kernel/matmul/tune.rs
@@ -216,14 +216,16 @@ where
     }
 
     fn prepare(&self, device: &WgpuDevice) -> Self::Args {
-        let lhs = Tensor::<WgpuBackend<G, E, i32>, D>::random(
+        let lhs = Tensor::<WgpuBackend<G, E, i32>, D>::random_device(
             self.shape_lhs.clone(),
             Distribution::Default,
+            device,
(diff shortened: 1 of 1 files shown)


louisfd:
That makes sense!

nathanielsimard:
I would remove the `to_device` at the end of each declaration.

louisfd:
Indeed, i'll fix it in a new pr

PR #692, opened 2023-08-24
Implement Quiet Softmax (`Attention Is Off By One`)

wbrickner:
### Checklist

### Related Issues/PRs

- https://github.com/burn-rs/burn/issues/691

### Changes

- I propose a `quiet_softmax` activation function (motivated by [Attention Is Off By One](https://www.evanmiller.org/attention-is-off-by-one.html)).
- I propose a `quiet_softmax` configuration field for `MultiHeadAttention` (on `MultiHeadAttentionConfig`), and to provide  this field as well for all the layers that use `MultiHeadAttention` internally (like `TransformerEncoder`).
- There is a case to be made that this should be the default softmax implementation (It is **not** default enabled in this PR)

### Testing

Run checks.


nathanielsimard (approved):

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -31,6 +31,27 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) ->
...
+/// It must in the range of `0` and `D-1`.
+pub fn quiet_softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
+    check!(TensorCheck::dim_ops::<D>("softmax", dim));
+
+    let tensor = tensor.clone() - tensor.detach().max_dim(dim);
+    let tensor = tensor.exp();
+    let tensor_tmp = tensor.clone().sum_dim(dim);
+    let one = Tensor::<B, D>::ones([1; D]).to_device(&tensor_tmp.device());
+
+    tensor.div(tensor_tmp + one)
```
nathanielsimard, line 52: I would use a scalar for the addition instead of a tensor:

```rust
tensor.div(tensor_tmp.add_elem(1))
```

nathanielsimard (changes requested):
Just a small change, but thanks a lot for implementing this.

The paper is indeed very interesting, and it's nice to provide that option as well. Similarly to using normalization first, it's not the default, but it's probably the better option.

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -31,6 +31,27 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) ->
...
+///
+/// The dimension argument `dim` specifies the dimension along which the function will be computed.
+/// It must in the range of `0` and `D-1`.
+pub fn quiet_softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
+    check!(TensorCheck::dim_ops::<D>("softmax", dim));
+
+    let tensor = tensor.clone() - tensor.detach().max_dim(dim);
+    let tensor = tensor.exp();
+    let tensor_tmp = tensor.clone().sum_dim(dim);
+    let one = Tensor::<B, D>::ones([1; D]).to_device(&tensor_tmp.device());
```
louisfd, line 50: You can use ones_device instead of ones and to_device

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -31,6 +31,27 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) ->
...
+/// This function is similar to the softmax function, but it allows for "no selection", e.g.,
+/// all outputs can tend to zero.
+///
+/// `softmax(x_i) = exp(x_i) / [ 1 + sum_j(exp(x_j)) ]`
+///
+/// # Notes
+///
+/// The dimension argument `dim` specifies the dimension along which the function will be computed.
+/// It must in the range of `0` and `D-1`.
+pub fn quiet_softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
```
louisfd, line 44: This would deserve a test, see burn-tensor/src/tests/activation/softmax.rs for the softmax test and make one similar for quiet softmax

louisfd (changes requested):
Hi @wbrickner 
I'm happy to see how as a community we can respond quickly to new things in our field. Thanks a lot! 
I have only one request, that the new function be tested, see my comment

burn-tensor/src/tensor/activation/base.rs
louisfd: Never mind, if you use the scalar addition suggested by @nathanielsimard

nathanielsimard:
@wbrickner Can you fix the comments and the merge conflicts? I think it would be ready to be merged :)

antimora:
@wbrickner, let us know if you need help with this. We do not wish to lose such important addition to Burn.

wbrickner:
got busy, I will implement this again today.

wbrickner:
Hey, so turns out I do need help! Sorry for screwing up the issue format by accidentally closing. I am having a lot of trouble getting burn checks to pass locally after pulling a fresh copy of burn. The implementation has been rewritten in the new copy (since a few days ago). It's fairly simple. My other road block is computing the derivatives by hand to get correct values to test against, the expressions get quite complicated. I suppose I should just write it out by hand and refresh on the differentiation rules haha. Sorry this was left open for so many months! I would like to move forward with it!

antimora:
Sounds good! We will look into this.

nathanielsimard (commented):
I think you need to run the correct formating: `cargo fmt --all`

wbrickner:
> I think you need to run the correct formating: `cargo fmt --all`

To clarify, I should format my changes and update the remote repo attached to this PR?
The checks pass in github actions, is this blocked from merging?

antimora:
> > I think you need to run the correct formating: `cargo fmt --all`
> 
> To clarify, I should format my changes and update the remote repo attached to this PR? The checks pass in github actions, is this blocked from merging?

OK, I have rebased it and ran the formatting. Now it's passing the checks. We probably need an approval to be merged.

burn-derive/src/config/analyzer_enum.rs
```diff
@@ -72,11 +72,11 @@ impl ConfigEnumAnalyzer {
...
         let variants = self.data.variants.iter().map(|variant| {
-            let variant_name = &variant.ident;
-            let (variant_input, variant_output) = self.gen_variant_field(variant);
+      let variant_name = &variant.ident;
+      let (variant_input, variant_output) = self.gen_variant_field(variant);
 
-            quote! { Self::#variant_name #variant_input => #enum_name::#variant_name #variant_output }
-        });
+      quote! { Self::#variant_name #variant_input => #enum_name::#variant_name #variant_output }
+    });
```
nathanielsimard, line 79: That formatting seems odd to me and somehow fmt can't update it 🤔

nathanielsimard (changes requested):
There are formatting issues in `burn-derive` cargo fmt has a hard time fixing them because of macros, but no changes should actually be made in `burn-derive`.

burn-derive/src/config/analyzer_enum.rs
wbrickner: What would you like me to do to resolve this issue and merge the PR? `cargo fmt --all` results in a bit identical repository. Is the issue these two lines of whitespace? Should I modify them manually? You mention no changes should be made to `burn-derive`.  Would like to get this closed out, apologies on the silliness of these problems.

AlexErrant: It's due to the `quote!` macro. https://github.com/rust-lang/rustfmt/issues/8

If you comment out `quote!`, format, and uncomment, it'll do the right thing.

AlexErrant: aaaaaactshully it's due to the default `max_width = 100`. If you add a file called `rustfmt.toml` and in it is `max_width = 110` then format, it works. I'll open another PR about this since it causes changes elsewhere and deserves its own discussion.

louisfd (approved):
This has been pending long enough 😅 I'll merge it, I think the formatting is fine

PR #699, opened 2023-08-25
Path type

asukaminato0721:
### Checklist

### Related Issues/PRs

https://github.com/burn-rs/burn/issues/698

### Changes

`&str` to `AsRef<Path>`

### Testing


antimora:
Thank you for your PR.

Build is failing. Please, rerun `run-checks` from the root. There is an example crate is failing to build.

nathanielsimard (changes requested):
I would also receive `P: AsRef<std::path::Path>` for the `load` function. Maybe you need to change the `to_string` to something like this: https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy

antimora (approved):
LGTM

nathanielsimard (approved):

PR #700, opened 2023-08-26
Expose element traits

MOZGIII:
### Checklist

### Related Issues/PRs

None, went straight to making a PR.

### Changes

Export the wgpu elements traits to the public as they are required for restricting generic implementations.

### Testing

I have my code that uses it via `patch`-ing burn to this branch.

### DIFF
--- a/burn-ndarray/src/element.rs
+++ b/burn-ndarray/src/element.rs
@@ -3,13 +3,15 @@ use libm::{exp, fabs, log, log1p, pow, sqrt};
 use libm::{expf, fabsf, log1pf, logf, powf, sqrtf};
 use ndarray::LinalgScalar;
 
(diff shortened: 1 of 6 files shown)


nathanielsimard:
@MOZGIII Yes this is required indeed. We will need to make more backend traits/functions public to make them more extensible.

MOZGIII:
@nathanielsimard I have tried to do some further work and realized that there is no proper support for `dyn Backend`s.
I think this would be super cool ti support, alongside with a runtime backend selector (and this is what I was trying to build actually).

I'll create an issue from my experiments to outline the required changes to burn to make it happen and, possibly, track the implementation progress.

nathanielsimard (approved):

PR #703, opened 2023-08-26
fix: dataloader reshuffle dataset each iteration

### LINKED ISSUE #701: DataLoader.shuffle() should reshuffle data at each epoch
### Feature description

DataLoader.shuffle() should reshuffle data at every epoch at each epoch.

### Feature motivation

By shuffling the data at each epoch, the model is exposed to different patterns in each epoch, which helps prevent it from overfitting to specific patterns in the training,

In PyTorch, it's implemented by random sampler: https://github.com/pytorch/pytorch/blob/0b868b19063645afed59d6d49aff1e43d1665b88/torch/utils/data/sampler.py#L40-L82

nathanielsimard:
Fix #701

### DIFF
--- a/burn-core/src/data/dataloader/batch.rs
+++ b/burn-core/src/data/dataloader/batch.rs
@@ -2,7 +2,11 @@ use super::{
     batcher::Batcher, BatchStrategy, DataLoader, DataLoaderIterator, MultiThreadDataLoader,
     Progress,
 };
-use burn_dataset::{transform::PartialDataset, Dataset};
+use burn_dataset::{
+    transform::{PartialDataset, ShuffledDataset},
+    Dataset,
+};
+use rand::{distributions::Standard, prelude::Distribution, rngs::StdRng, Rng, SeedableRng};
 use std::sync::Arc;
 
 /// A data loader that can be used to iterate over a dataset in batches.
@@ -10,14 +14,7 @@ pub struct BatchDataLoader<I, O> {
     strategy: Box<dyn BatchStrategy<I>>,
     dataset: Arc<dyn Dataset<I>>,
     batcher: Arc<dyn Batcher<I, O>>,
-}
-
-/// A data loader iterator that can be used to iterate over a data loader.
-struct BatchDataloaderIterator<I, O> {
-    current_index: usize,
-    strategy: Box<dyn BatchStrategy<I>>,
-    dataset: Arc<dyn Dataset<I>>,
-    batcher: Arc<dyn Batcher<I, O>>,
+    rng: Option<spin::Mutex<rand::rngs::StdRng>>,
 }
 
 impl<I, O> BatchDataLoader<I, O> {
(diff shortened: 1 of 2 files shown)


burn-core/src/data/dataloader/batch.rs
```diff
@@ -67,24 +76,49 @@ where
...
+
+        let mut dataloaders: Vec<Arc<dyn DataLoader<_> + Send + Sync>> =
+            Vec::with_capacity(num_threads);
+
+        // Create more rngs from the first one, one for each new dataloader.
+        let rngs = (0..num_threads).map(|_| {
+            rng.as_mut()
+                .map(|rng| StdRng::seed_from_u64(Distribution::sample(&Standard, rng)))
+        });
+
```
antimora, line 91: From what I could gather, here we are attempting to recreate [ThreadRng](https://rust-random.github.io/rand/rand/rngs/struct.ThreadRng.html), which probably should be used instead of StdRng. If so, please consider using ThreadRng, which most likely be beneficial for this use case.

Generally speaking I also have noticed that we use Std rgn with locks but it could be avoided with ThreadRng in other parts of our Burn code. We should review and see if we could replace rng with locks.

antimora (changes requested):

burn-core/src/data/dataloader/batch.rs
nathanielsimard: Using `ThreadRng` can lead to nondeterministic code execution. If someone wants to reproduce an experiment, disabling multithreading with the DataLoader should ensure that each batch is created in the same order and with the same values. A random number generator is inherently a sequential algorithm with a state. That's why we need to wrap it inside a Mutex, but I don't expect this to be problematic.

When using `ThreadRng`, we create a new instance with a different seed each time. I'm not sure if this is more performant, but I don't think performance should be a concern in this scenario. So, either way, it isn't problematic.

@louisfd  What are you thoughs on this, should we prefer using `ThreadRng`.

antimora: I see and it's a good point. I wonder if it's possible to make ThreadRng to seed its source from one place. There is this interface that the documentation mentions: https://rust-random.github.io/rand/rand/rngs/adapter/struct.ReseedingRng.html

louisfd: ThreadRng cannot be used for deterministic applications, we have no control over it as it takes its seed directly form the OS hardware rng. 
Therefore using StdRng is the way to go. 
The bottom example of this page is about deterministic multi-threaded: https://rust-random.github.io/book/guide-parallel.html
It does something similar to what you do, except their different seeds are contiguous indices, while yours are uniformly distributed. Both ways are fine.

louisfd (approved):
Very cool. I've done a little research to see if your multi thread deterministic strategy is adequate, and it turns out it is! So I'm approving 😄

nathanielsimard:
@antimora I merged after the review from @louisfd, the random number generator master 😎

PR #704, opened 2023-08-26
Feat: add dataset sampler with and without replacement

nathanielsimard:
### Changes

Add the ability to create dataset without replacement.

### Testing

Added more unit tests

### DIFF
--- a/burn-dataset/src/transform/sampler.rs
+++ b/burn-dataset/src/transform/sampler.rs
@@ -1,6 +1,6 @@
 use crate::Dataset;
-use rand::{distributions::Uniform, rngs::StdRng, Rng, SeedableRng};
-use std::{marker::PhantomData, sync::Mutex};
+use rand::{distributions::Uniform, rngs::StdRng, seq::IteratorRandom, Rng, SeedableRng};
+use std::{marker::PhantomData, ops::DerefMut, sync::Mutex};
 
 /// Sample items from a dataset.
 ///
@@ -8,7 +8,7 @@ use std::{marker::PhantomData, sync::Mutex};
 /// You have multiple options to instantiate the dataset sampler.
 ///
 /// * With replacement: This is the most efficient way of using the sampler because no state is
-///   required to keep indexes that have been selected.
+///   required to keep indices that have been selected.
 ///
 /// * Without replacement: This has a similar effect to using a
 ///   [shuffled dataset](crate::transform::ShuffledDataset), but with more flexibility since you can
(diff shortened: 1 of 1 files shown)


burn-dataset/src/transform/sampler.rs
```diff
@@ -17,22 +26,58 @@ where
...
+            Some(state) => {
+                let mut state = state.lock().unwrap();
+
+                if state.len() == 0 {
+                    // Refill the state using the same vector.
+                    (0..self.dataset.len()).for_each(|i| state.push(i));
+                }
+
+                let index = self.index_size(state.len());
+                state.remove(index)
```
antimora, line 72: `remove` operation can be costly `O(N)`. See https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove

I recommend we change the algorithm to shuffle the list of `state` once (using something like https://docs.rs/shuffle/latest/shuffle/)  and use a concurrent data structure to pop a value, perhaps something like: https://github.com/smol-rs/concurrent-queue.  This way you amortize randomization and reduce locking. 

Also, several locks can lead to deadlocks. I didn't work out if it could actually happen in this instance, but it's possible.

antimora (changes requested):
Please see my inline comment.

burn-dataset/src/transform/sampler.rs
nathanielsimard: We should shuffle the array once and pop until there is not item and just refill it. I don't think we need an external crate to do it, but I agree we should change the current implementation to avoid costly operations.

antimora (approved):
Thanks. It looks good. There is one clippy warning that you need to handle. But I am approving it ahead.

burn-dataset/src/transform/sampler.rs
```diff
@@ -1,38 +1,77 @@
...
+
+        match state.deref_mut() {
+            SamplerState::WithReplacement(rng) => rng.sample(Uniform::new(0, self.dataset.len())),
+            SamplerState::WithoutReplacement(rng, indices) => {
+                if indices.is_empty() {
+                    // Refill the state.
+                    *indices = (0..self.dataset.len()).choose_multiple(rng, self.dataset.len());
+                }
+
+                indices.pop().expect("State is never empty")
```
louisfd, line 72: Again, we should document the fact that WithoutReplacement actually starts over when all indices have been picked once. 
Also, "State is never empty" is a bit cryptic to me

PR #706, opened 2023-08-27
Feat: Some tweaks to make it more practical to integrate in a GUI app

dae:
### Checklist

Those wgpu tests sure are heavy! A few of them failed for me, but I presume that's some issue with my gpu/driver.

```
---- tests::aggregation::tests::test_should_mean stdout ----
thread 'tests::aggregation::tests::test_should_mean' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [2.4999998], shape: Shape { dims: [1] } }`,
   right: `Data { value: [2.5], shape: Shape { dims: [1] } }`', burn-wgpu/src/lib.rs:47:5

---- tests::aggregation::tests::test_should_mean_last_dim stdout ----
thread 'tests::aggregation::tests::test_should_mean_last_dim' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [0.99999994, 3.9999998], shape: Shape { dims: [2, 1] } }`,
   right: `Data { value: [1.0, 4.0], shape: Shape { dims: [2, 1] } }`', burn-wgpu/src/lib.rs:47:5

---- tests::div::tests::should_support_div_ops stdout ----
thread 'tests::div::tests::should_support_div_ops' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [0.0, 1.0, 1.0, 1.0, 1.0, 1.0], shape: Shape { dims: [2, 3] } }`,
   right: `Data { value: [0.0, 1.0, 1.0, 0.99999994, 1.0, 1.0], shape: Shape { dims: [2, 3] } }`', burn-wgpu/src/lib.rs:47:5


failures:
    tests::aggregation::tests::test_should_mean
    tests::aggregation::tests::test_should_mean_last_dim
    tests::div::tests::should_support_div_ops
```

### Related Issues/PRs

https://github.com/open-spaced-repetition/fsrs-optimizer-burn/issues/27

### Changes

Both commits are aimed at making it easier to use burn in a GUI app. The first allows opting out of the default CLI renderer, and the second allows us to avoid creating large files on disk when the code happens to have some info!() calls in the training path.

### Testing

The standard tests pass, and I tried a branch of our code that uses these options.

### DIFF
--- a/burn-train/src/learner/builder.rs
+++ b/burn-train/src/learner/builder.rs
@@ -3,7 +3,7 @@ use super::Learner;
 use crate::checkpoint::{AsyncCheckpointer, Checkpointer, FileCheckpointer};
 use crate::logger::{FileMetricLogger, MetricLogger};
 use crate::metric::dashboard::cli::CLIDashboardRenderer;
-use crate::metric::dashboard::Dashboard;
+use crate::metric::dashboard::{Dashboard, DashboardRenderer, MetricWrapper, Metrics};
 use crate::metric::{Adaptor, Metric, Numeric};
 use crate::AsyncTrainerCallback;
 use burn_core::lr_scheduler::LRScheduler;
@@ -24,7 +24,6 @@ where
     O: Optimizer<M, B>,
     S: LRScheduler,
 {
-    dashboard: Dashboard<T, V>,
     checkpointer_model: Option<Arc<dyn Checkpointer<M::Record> + Send + Sync>>,
     checkpointer_optimizer: Option<Arc<dyn Checkpointer<O::Record> + Send + Sync>>,
     checkpointer_scheduler: Option<Arc<dyn Checkpointer<S::Record> + Send + Sync>>,
@@ -33,6 +32,10 @@ where
     directory: String,
     grad_accumulation: Option<usize>,
     devices: Vec<B::Device>,
+    metric_logger_train: Option<Box<dyn MetricLogger + 'static>>,
+    metric_logger_valid: Option<Box<dyn MetricLogger + 'static>>,
+    renderer: Option<Box<dyn DashboardRenderer + 'static>>,
+    metrics: Metrics<T, V>,
 }
 
 impl<B, T, V, Model, Optim, LR> LearnerBuilder<B, T, V, Model, Optim, LR>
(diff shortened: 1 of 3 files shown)


nathanielsimard (approved):
Very good! The abstractions were already there, but not exposed. Thanks for this 🙏

dae:
I've discovered an issue with this - the CLI progress bars are getting drawn to the screen at the CLIDashboardRenderer::new() call. Will investigate a solution.

dae:
I had to rewrite the first commit, and fixing the above issue required a fair few extra changes I'm afraid.

Regarding exposing some of the metric internals to the builder: I wasn't sure of another way to store the metrics until the dashboard is created, and if you can see a better way, I'm open to suggestions.

burn-train/src/learner/builder.rs
```diff
@@ -210,8 +247,26 @@ where
...
+            Box::new(FileMetricLogger::new(format!("{directory}/train").as_str()))
+        });
+        let logger_valid = self.metric_logger_valid.unwrap_or_else(|| {
+            Box::new(FileMetricLogger::new(format!("{directory}/valid").as_str()))
+        });
+        let mut dashboard = Dashboard::new(renderer, logger_train, logger_valid);
+        dashboard.metrics_train = self.metrics_train;
+        dashboard.metrics_valid = self.metrics_valid;
+        dashboard.metrics_train_numeric = self.metrics_train_numeric;
+        dashboard.metrics_valid_numeric = self.metrics_valid_numeric;
```
nathanielsimard, line 267: If we go for a metric struct, we could pass it as parameter in the `new` instead. It would also remove all methods on the dashboard to add metrics, since they should be passed in the `new` function from now on. Those functions could be moved to the newly created metric struct if we see fit, but making each field `pub(crate)` may also be the better option.

burn-train/src/learner/builder.rs
```diff
@@ -33,6 +34,14 @@ where
     directory: String,
     grad_accumulation: Option<usize>,
     devices: Vec<B::Device>,
+    metric_logger_train: Option<Box<dyn MetricLogger + 'static>>,
+    metric_logger_valid: Option<Box<dyn MetricLogger + 'static>>,
+    renderer: Option<Box<dyn DashboardRenderer + 'static>>,
+    metrics_train: Vec<Box<dyn DashboardMetric<T>>>,
+    metrics_valid: Vec<Box<dyn DashboardMetric<V>>>,
+    metrics_train_numeric: Vec<Box<dyn DashboardNumericMetric<T>>>,
+    metrics_valid_numeric: Vec<Box<dyn DashboardNumericMetric<V>>>,
```
nathanielsimard, line 43: I think we should put all metrics in a struct. It wasn't flexible enough in the dashboard, but we should have a struct that just group those types.

nathanielsimard (changes requested):

dae:
Is the changed version what you had in mind?

nathanielsimard (approved):
Yes this is the kind of pattern that I was expecting, thanks 🙏

PR #708, opened 2023-08-28
Fix: processed count was one greater than total items at completion

dae:
### Checklist

### Changes

At the end of training, the processed count was one greater than the total count. By doing the increment
after the loop exit check, this is avoided.

### Testing

Standard tests, and manual observation that it performs correctly now.


louisfd (approved):
Indeed, that makes sense!

PR #709, opened 2023-08-28
Account for additional cases in Linear node conversion

antimora:
### Checklist

### Changes

Account for two additional cases in Linear node conversion. Previously we only converted Gemm node to a Linear module. Now, we are converting MatMul and Add optionally. See ONNX graph of 3 cases for illustration. 

### Testing

Add more test cases and ONNX example.

image


nathanielsimard (changes requested):
A small refactor for cleaner code, but otherwise great !

nathanielsimard (approved):

Issue #697, opened 2023-08-25
Some tensor ops lack tests


louisfd:
As it is a common task to fill every operation in a backend, I think all of them should be protected by a test. 
Some are missing, mostly for int_tensors: 

**Float Tensor**

- [x] random (other than default)
- [x] into_int
- [x] empty

**Bool Tensor**
- [x] bool_empty
- [x] bool_shape
- [x] bool_reshape
- [x] bool_cat
- [x] bool_equal
- [x] bool_equal_elem

**Int tensor**
- [x] int_empty
- [x] int_mask_where
- [x] int_gather
- [x] int_scatter
- [x] int_select
- [x] int_select_assign
- [x] int_cat
- [x] int_greater_equal_elem
- [x] int_add
- [x] int_add_scalar
- [x] int_sub
- [x] int_sub_scalar
- [x] int_mul
- [x] int_mul_scalar
- [x] int_div
- [x] int_div_scalar
- [x] int_zeros
- [x] int_ones
- [x] int_sum
- [x] int_sum_dim
- [x] int_mean_dim
- [x] int_argmax
- [x] int_argmin

antimora:
It's great you have identified the list. I have filed a similar ticket for missing tests. It includes instruction on how to add the tests. We can combine the tickets.

https://github.com/burn-rs/burn/issues/384

louisfd:
Thanks @antimora 
Just to be clear, the above list only concerns operations that have no default implementations and must necessarily be implemented for a new backend. 
There may be missing tests for operations that have default implementations, such as `unsqueeze`, even if they are not in that list. See #384

PR #710, opened 2023-08-28
Perf/tensor ops/tests

louisfd:
Some more tensor ops tests, see #697 
Fix ndarray backend bug
Bool_not instead of bool_equal_elem


nathanielsimard (approved):
✅🙏

PR #711, opened 2023-08-28
Allow opting out of SQLite support

dae:
burn-dataset includes a bundled copy of SQLite by default, which proves problematic - it doesn't allow us to use a system libsqlite if we'd like, and our usage of 'bundled' conflicts with burn-dataset's usage of 'bundled-full', resulting in databases that can't be opened.

Rusqlite's docs have the following to say:

 `bundled` causes us to automatically compile and link in an up to date
 version of SQLite for you. This avoids many common build issues, and
 avoids depending on the version of SQLite on the users system (or your
 system), which may be old or missing. It's the right choice for most
 programs that control their own SQLite databases.

 That said, it's not ideal for all scenarios and in particular, generic
 libraries built around `rusqlite` should probably not enable it, which
 is why it is not a default feature -- it could become hard to disable.

I've assumed you'd like to keep this functionality on by default, so I've just added a feature flag so that we can opt out of it.

### Checklist

### Testing

Standard tests, and confirmed this fixes the problem for us.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,11 +47,11 @@ proc-macro2 = "1.0.60"
 protobuf-codegen = "3.2"
 quote = "1.0.28"
 r2d2 = "0.8.10"
-r2d2_sqlite = {version = "0.22.0", features = ["bundled"]}
+r2d2_sqlite = {version = "0.22.0"}
 rayon = "1.7.0"
 rmp-serde = "1.1.1"
 rstest = "0.18.1"
-rusqlite = {version = "0.29", features = ["bundled-full"]}
+rusqlite = {version = "0.29"}
 sanitize-filename = "0.5.0"
 serde_rusqlite = "0.33.1"
 spin = {version = "0.9.8", features = ["mutex", "spin_mutex"]}
--- a/burn-core/Cargo.toml
+++ b/burn-core/Cargo.toml
@@ -43,7 +43,7 @@ burn-autodiff = { path = "../burn-autodiff", version = "0.9.0", optional = true,
   "export_tests",
 ] }
 burn-common = { path = "../burn-common", version = "0.9.0", default-features = false }
-burn-dataset = { path = "../burn-dataset", version = "0.9.0", default-features = false, optional = true }
+burn-dataset = { path = "../burn-dataset", version = "0.9.0", optional = true }
 burn-derive = { path = "../burn-derive", version = "0.9.0" }
 burn-tensor = { path = "../burn-tensor", version = "0.9.0", default-features = false }
 
--- a/burn-dataset/Cargo.toml
+++ b/burn-dataset/Cargo.toml
@@ -13,6 +13,7 @@ version = "0.9.0"
 [features]
 default = [
   "fake",
+  "sqlite-bundled"
 ]
 
 audio = [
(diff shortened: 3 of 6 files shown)


antimora:
Before we make this change, we should attempt to fix the actual problem. The solution opts out completely from using SQLite. Some might actually want a system installed SQLite but use SQLite dataset functionality. It sounds like this will work for you as well.

I was actually who introduced this functionality and familiar with its history. Initially we used the system based SQLite but someone had trouble ( it could have been a windows user). So we switched to bundle, easy way to fix.

dae:
Would you be open to both approaches? We don't need the functionality, and disabling it removes the risk that a future burn update will add an incompatible feature flag to its rusqlite, and removes any risk of one of our crates being a version ahead or behind the other's rusqlite.

If so, I can remove the bundled/bundled-full flags in this PR as well if you'd like. If you'd prefer not to add this feature flag, I can put through a new PR that just removes the bundled features instead.

nathanielsimard:
@dae @antimora I think both of you are right 😅. We should have a way to disable the dependency altogether, especially if you don't plan to use anything else than the in-memory dataset. So I propose multiple feature flags:

`sqlite`: Enable the SQLite dataset source.
`sqlite-system`: Enable the SQLite dataset with the system SQLite.
`huggingface`: Assume either `sqlite` or `sqlite-system` is activated.

I think those feature flags can help have smaller binaries and add more flexibility. What are your thoughts?

dae:
Did you mean a separate feature for huggingface? I can think of how to require either feature using #[cfg], but didn't know there's a way to declare that one feature requires either of two other features.

Which of the two sqlite features do you think should be the default? If @antimora is on board too, I can update this PR to do so.

nathanielsimard:
@dae, I don't think we can enforce two features, but it's often the case where one feature needs another one, and the project will not compile if you don't activate both features. There are a lot of examples with the `std` features in many crates that support `no_std`.

I think I would use the `sqlite` feature by default, since it's probably the one that will cause fewer problems when trying out Burn.

antimora:
> Did you mean a separate feature for huggingface? I can think of how to require either feature using #[cfg], but didn't know there's a way to declare that one feature requires either of two other features.
> 
> Which of the two sqlite features do you think should be the default? If @antimora is on board too, I can update this PR to do so.

Yes, I am totally fine with introducing the new feature flags: `sqlite` (or `sqlite-bundled`) and `sqlite-system`. I was going to mention this, in my initial response, as an option. I agree we should enable `sqlite-bundled` (or `sqlite`) by default. And add `#[cfg]` on modules if either features are enabled, then enable the modules.

Since, these flags are mutually exclusive (and the feature flags are usually additive), we cannot enforce it for dependency (maybe with `compile_error!(` but not sure if this work for dependencies). Perhaps, that's why we should just rename `sqlite` to `sqlite-bundle`, so that there is a little conceptual overlap with `sqlite-system`. 

Hope this is not a lot of work/change! And maybe you also might benefit from the system installed option too =) We all will benefit from your PR.

dae:
Does the updated version work for you guys? I updated the description too.

antimora:
Looks good, but there is one issue, which had to do how dataset was reexported from burn-core, which was reexported in burn package. I have discovered that the default features were off when I was testing mnist example: 

```
[burn/examples/mnist]$ cargo run --example mnist --release --features tch-gpu
```

Here is the diff on fixing it in `burn/burn-core/Cargo.toml`:

```diff
-burn-dataset = { path = "../burn-dataset", version = "0.9.0", default-features = false, optional = true }
+burn-dataset = { path = "../burn-dataset", version = "0.9.0", optional = true }
```

But one consequence of this is that `fake` feature is enable in the core now. Is this okay, @nathanielsimard ? 

Once this is fixed, it's approved. It works on Mac M1.

nathanielsimard (approved):

PR #716, opened 2023-08-29
Fix: ensure final CLI update happens

dae:
The merge of #708 unearthed a bug in the CLI code: if at completion time the update is within the throttling period, you can end up with a final output that appears as if the process didn't fully complete.

More info: https://github.com/open-spaced-repetition/fsrs-optimizer-burn/pull/36#issuecomment-1696736807

### Checklist


louisfd (approved):
LGTM
Thanks

PR #725, opened 2023-08-29
Feat/candle/module ops

louisfd:
### Checklist

### Changes

Add Candle supported module and activation ops. 
With this I think we are on par with current Candle advancements. 
I have submitted an issue at Candle to support slice_assign, which should allow us to enable many of the remaining tests. For module ops such as conv_transpose, etc. I think the Candle team has plans to work on them soon, we should check it periodically.


burn-candle/src/ops/module.rs
```diff
@@ -1,24 +1,90 @@
...
         bias: Option<FloatTensor<Self, 1>>,
         options: ConvOptions<2>,
     ) -> FloatTensor<Self, 4> {
-        todo!()
+        assert!(
+            options.dilation[0] == options.dilation[1]
+                && options.padding[0] == options.padding[1]
+                && options.stride[0] == options.stride[1],
+            "Candle does not support per dimension options in convolutions"
+        );
```
nathanielsimard, line 53: This probably breaks a lot of tests.

burn-candle/src/ops/module.rs
```diff
@@ -1,24 +1,90 @@
...
+    }
+
+    fn conv_transpose1d(
+        x: FloatTensor<Self, 3>,
+        weight: FloatTensor<Self, 3>,
+        bias: Option<FloatTensor<Self, 1>>,
+        options: ConvTransposeOptions<1>,
+    ) -> FloatTensor<Self, 3> {
+        panic!("Candle does not support conv_transpose1d")
     }
```
nathanielsimard, line 88: There is a default implementation that uses conv_transpose_2d, so just don't implement this function and it will work!

burn-tch/src/ops/module.rs
```diff
@@ -182,7 +185,20 @@ impl<E: TchElement> ModuleOps<TchBackend<E>> for TchBackend<E> {
             dilation as i64,
             false,
         );
-
+        println!(
+            "{:?}",
+            Tensor::<TchBackend<E>, 3>::from_primitive(TchTensor::new(tensor))
+                .to_data()
+                .value
+        );
```
nathanielsimard, line 193: Remove println!

burn-tensor/src/tests/module/maxpool1d.rs
```diff
@@ -9,7 +9,7 @@ mod tests {
     #[test]
```
nathanielsimard, line 9: Is there a reason to update the tests for max pool 1d? Maybe updating one test with no padding is good.

nathanielsimard (changes requested):

burn-candle/src/ops/module.rs
louisfd: Yes, that's why we can't run our convolution test suite 😢

burn-candle/src/ops/module.rs
louisfd: I implemented it on purpose. Because of the asserts I put in conv_transpose2d, we can't call conv_transpose1d from conv_transpose2d as it puts a different second option. So I was thinking about either rewrite conv_transpose1d from conv_transpose2d specifically for candle, or remove the assertions... it was getting confusing, so I chose to wait to see if they are going to write conv_transpose1d (they did write both conv_1d and conv_2d)

burn-tch/src/ops/module.rs
louisfd: 😮

burn-tch/src/ops/module.rs
```diff
@@ -182,7 +185,20 @@ impl<E: TchElement> ModuleOps<TchBackend<E>> for TchBackend<E> {
...
+        let tensor = tch::Tensor::max_pool1d(
+            &x.tensor,
+            kernel_size as i64,
+            stride as i64,
+            padding as i64,
+            dilation as i64,
+            false,
+        );
```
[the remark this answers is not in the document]
louisfd: There should not even be any change here, I was just using tch to print the result of a changed test and forgot to undo the change

burn-tensor/src/tests/module/maxpool1d.rs
louisfd: The idea is to have at least one test for the most basic scenario. With what candle offers as of now it's not possible to put padding in max pooling, so if every test has padding I cannot check that I've done the basic binding correctly.

nathanielsimard (approved):

PR #727, opened 2023-08-29
Update burn-import README

antimora:
### Changes

1. Moved supported op into its own document.
2. Moved dev guide into its own document.
3. Updated the main read to be a simple user guide.

### Testing

Viewed rendered documents.


nathanielsimard (changes requested):
Just update the Burn supported ops, but otherwise very good job, it's awesome to improve our documentation 🙏

nathanielsimard (approved):

PR #730, opened 2023-08-30
Book: Expanded the planned sections and added built-in module section

antimora:
Expanded the planned sections and added built-in module section


burn-book/src/SUMMARY.md
```diff
@@ -12,9 +12,23 @@
...
   - [Custom Training Loops]()
   - [Custom Metric]()
   - [Custom Kernels]()
     - [WGPU]()
+  - [WebAssembly]()
+  - [No-Std]()
+- [Contributing]()
+  - [Tenets]()
+  - [Design Philosophy]()
+  - [Architecture]()
```
nathanielsimard, line 33: I think this should be elsewhere, maybe in another book.

burn-book/src/building-blocks/module.md
```diff
@@ -99,3 +97,33 @@ pub trait ModuleMapper<B: Backend> {
...
+| `Gru`                | `nn.GRU`                                |
+| `LayerNorm`          | `nn.LayerNorm`                          |
+| `Linear`             | `nn.Linear`                             |
+| `Lstm`               | `nn.LSTM`                               |
+| `MaxPool1d`          | `nn.MaxPool1d`                          |
+| `MaxPool2d`          | `nn.MaxPool2d`                          |
+| `MultiHeadAttention` | `nn.MultiheadAttention`                 |
+| `PositionalEncoding` | _No direct equivalent_                  |
+| `TransformerDecoder` | `nn.TransformerDecoder`                 |
+| `TransformerEncoder` | `nn.TransformerEncoder`                 |
```
nathanielsimard, line 129: I would remove "a few" since it isn't that small and it will only grow, potentially making this line look odd in the future.

Also, the list isn't complete:

- CrossEntropyLoss
- MSELoss
- LayerNorm
- Embedding
- PositionalEncoding
- Relu

I would also separate them into sub-sections:

- Pooling
- Convolutions
- RNNs
- Transformer
- General
- Loss

At some point, we might create issues that document which modules we want to support in each section.

Finally, the `MSELoss` and `CrossEntropyLoss` don't implement `Module` yet, but we should annotate them even though they don't have state. They were always used inline, but we could put them into a module as is often done with PyTorch.

burn-book/src/SUMMARY.md
```diff
@@ -12,9 +12,23 @@
   - [Tensor](./building-blocks/tensor.md)
   - [Autodiff](./building-blocks/autodiff.md)
   - [Module](./building-blocks/module.md)
+  - [Config]()
+  - [Learner]()
+  - [Metric]()
+  - [Record]()
+  - [Dataset]()
+  - [Optimizer]()
```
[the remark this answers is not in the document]
antimora: OK. Moved to advanced section.

burn-book/src/SUMMARY.md
antimora: OK. Maybe we should add this to CONTRIBUTING.md. I'll remove them.

burn-book/src/SUMMARY.md
```diff
@@ -12,9 +12,23 @@
...
+- [Installation and Setup]()
```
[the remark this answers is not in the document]
antimora: OK. I'll move it over there. Since it was empty, I had put it there.

burn-book/src/building-blocks/module.md
antimora: > Finally, the `MSELoss` and `CrossEntropyLoss` don't implement `Module` yet, but we should annotate them even though they don't have state. They were always used inline, but we could put them into a module as is often done with PyTorch.

Filed a ticket for this: https://github.com/burn-rs/burn/issues/736

Rest is updated.

nathanielsimard (approved):

Issue #582, opened 2023-08-03
Performance WGPU: Managed Memory


nathanielsimard:
When creating buffers with `pub fn create_buffer(&self, size: usize) -> Arc<Buffer>`, we should try to reuse existing buffers that should be managed with a memory pool. This would reduce the amount of allocation and deallocation and increase performance while potentially reducing memory usage.

DrChat:
Found this library from the `wgpu` folk that could help: https://github.com/gfx-rs/range-alloc

Some working notes:
In `burn_wgpu::context::base::Context::create_buffer_with_data_options`, it would be best to avoid calling `Device::create_buffer_init` or `Device::create_buffer` directly and instead maintain our own pool of memory buffers.
The return value should be altered so that instead of returning a `wgpu::Buffer`, it returns an `Arc`-like struct that either describes a range inside of a `wgpu::Buffer` or a `wgpu::Buffer` (if the allocation is large enough).

nathanielsimard:
@DrChat Thanks for the suggestion. Once the system is in place, we should update `create_buffer_with_data_options` to reuse the pool. Also, I think there are potential problems with large buffers. Contrary to CUDA, I don't think we can allocate a very large buffer (as large as the physical memory), so it might reduce the kind of memory structure we could adopt. Though I'm unsure if the memory option is on shared storage buffers or buffers loaded into a kernel. Also, if we work with slice buffers, we should ensure that we can load them partially into a compute module, but maybe that is what `wgpu::BufferBinding` is for!

Creating our own buffer enum is interesting. I think it's probably a good refactor regardless of the implementation details!

DrChat:
Alright, so I put a little WIP branch together that implements this functionality.
Instead of loading buffers directly onto the GPU, the code now loads them into main memory and then into the GPU as they are demanded by compute tasks.

This is either running very nicely or very poorly on my system, depends on if the results are correct :)
Before this refactor though, I had no chance of running the model I'm running now.

nathanielsimard:
@DrChat you can submit a PR. Eventhough it is in WIP, we can better collaborate on this.

DrChat:
Yeah totally fair - submitted #734.

louisfd:
Closing as this was solved in #826

PR #734, opened 2023-08-30
Feat/wgpu: Refactor buffer management to allow for data buffers that are not resident on-GPU

DrChat:
### Checklist

### Related Issues/PRs

#582

### Changes

This PR refactors the memory management code so that buffers can be located in one of three places:
1) System memory (RAM)
2) GPU memory (dedicated buffer)
3) GPU memory (slice of shared buffer) (unimplemented)

Buffers are made resident in GPU memory on-demand when a compute call is scheduled.
If an allocation for a buffer fails for whatever reason, previously allocated buffers are evicted from GPU memory into system memory to make room for the new allocations.

### Testing

Locally with `stable-diffusion-xl-burn`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ tracing-core = "0.1.31"
 tracing-appender = "0.2.2"
 
 # WGPU stuff
+futures = "0.3.28"
 futures-intrusive = "0.5"
 pollster = "0.3"
 text_placeholder = {version = "0.5.0", features = ["struct_context"]}
--- a/burn-dataset/Cargo.toml
+++ b/burn-dataset/Cargo.toml
@@ -23,7 +23,7 @@ sqlite = ["__sqlite-shared", "dep:rusqlite"]
 sqlite-bundled = ["__sqlite-shared", "rusqlite/bundled"]
 
 # internal
-__sqlite-shared = ["dep:r2d2", "dep:r2d2_sqlite", "dep:serde_rusqlite"]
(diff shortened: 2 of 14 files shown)


nathanielsimard:
@DrChat Thanks for the PR. It includes quite a lot of changes, and I'm unsure how it will impact performance. Therefore, I suggest splitting this work into multiple PRs. Let's start with a refactoring of how buffers work.

I appreciate the work you've done with the buffer enum. Decoupling the actual buffer from the binding resource and replacing `buffer_init_with_data` were good moves. I propose creating a PR with only those components. This way, I can review it more effectively, and we can merge faster.

Regarding the memory pool, I believe we should address mutability using something more sophisticated than just the Arc strong count. I'm considering a more complex approach that would allow a buffer to be mutated even if partially shared with multiple tensors. This idea requires further consideration, but it should be built upon your buffer refactoring efforts!

DrChat:
I've pulled out the changes with the `BufferCache` since that code was indeed pretty messy and less than ideal.
I do need a good solution though for allocating buffers on-demand and evicting old ones as the GPU runs out of memory, since my GPU does not have enough memory to run the model I'm trying to run.

nathanielsimard:
I'm closing this, since we have refactored the wgpu backend with `burn-compute`. Thanks a lot for your PR, even if it didn't merged, it helped a lot in designing the `burn-compute` create 🙏!

DrChat:
Awesome, glad it helped! I took a peek at your implementation, and I can definitely say it looks much more polished than my draft :)

Issue #729, opened 2023-08-29
Add `training` and `inference` feature flags for selective re-export in `burn` meta crate


antimora:
The `burn` meta crate serves as an aggregator for various sub-crates, including `burn-core`, `burn-train`, and `burn-dataset`. It's often the only crate directly referenced in projects. To optimize for two main scenarios, we propose introducing the following feature flags:

1. **`training`**: Enables re-export of crates and modules specifically for training workflows.
2. **`inference`**: Enables only the minimal set of crates needed for inference tasks.

By implementing these feature flags, we can make the `burn` crate more lightweight and tailored for specific use-cases.

By default the two flags are enabled.

antimora:
@nathanielsimard , @louisfd , @dae, what are your thoughts regarding this proposal?

nathanielsimard:
Yes, I think it's a good idea. Of course, people will be able to create their own training pipeline, datasets, and use their own implementation without being forced to compile everything (looking at you, LibTorch 😅), but having them nicely packaged by default can't hurt.

Also, we might consider exporting the backends with none enabled by default.

```toml
burn = { version = "0.9.0", features = ["torch", "ndarray", "wgpu"] }
```

```rust
use burn::backend::{NdArrayBackend, TchBackend, WgpuBackend}
```

What are your thoughts, @antimora?

dae:
I suspect our app is a bit of an outlier here, as we'd like to use Burn for on-device training, but won't be taking advantage of the inference features (we'll just be feeding the output of our model directly into some simple equations).

For the training case, you currently have to import the backend (and things like autodiff too?) separately. I think it would a bit easier to follow if you were all or nothing here - either making all functionality available with feature flags and re-exports and letting users depend on a single crate, or consistently requiring them to import what they need from each crate.

antimora:
> What are your thoughts, @antimora?

We might as well do the whole thing as @dae has suggested, since the packages released in a lock step fashion anyway, and `burn` is just a meta package. 

We will just come up with a good feature sets and names. Lets this idea brew some more - it will definitely make Burn user friendly.

nathanielsimard:
@dae @antimora I don't think we have specific packages for general inference. `Burn-import` is a different beast, and you know when you need it. We could add a feature flag instead of adding each crate individually, but the complexity will grow if each sub-crate also has feature flags, like `ndarray blas`.

dae:
[...]

PR #740, opened 2023-08-31
Add ui/metrics feature flags

dae:
Partially implements #729:

- ~The burn crate no longer enables the training module unless the train feature is enabled.~
- Added new flags so the UI and/or system metrics can be excluded.

textplots has two security warnings that can be avoided when the CLI
is excluded:

33 │ atty 0.2.14 registry+https://github.com/rust-lang/crates.io-index
   │ ----------------------------------------------------------------- unsound advisory detected
   │
   = ID: RUSTSEC-2021-0145
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2021-0145
   = On windows, `atty` dereferences a potentially unaligned pointer.

16 │ ansi_term 0.12.1 registry+https://github.com/rust-lang/crates.io-index
   │ ---------------------------------------------------------------------- unmaintained advisory detected
   │
   = ID: RUSTSEC-2021-0139
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2021-0139
   = The maintainer has advised that this crate is deprecated and will not receive any maintenance.

### Checklist

### Related Issues/PRs

#729

### DIFF
--- a/burn-train/Cargo.toml
+++ b/burn-train/Cargo.toml
@@ -10,23 +10,33 @@ readme = "README.md"
 repository = "https://github.com/burn-rs/burn/tree/main/burn-train"
 version = "0.9.0"
 
+[features]
+cli = [
+    "indicatif",
+    "rgb",
+    "terminal_size",
+    "textplots",
+    "nvml-wrapper",
+    "sysinfo",
+    "systemstat"
+]
+
 [dependencies]
 burn-core = {path = "../burn-core", version = "0.9.0" }
 
-# Console
-indicatif = "0.17.5"
 log = {workspace = true}
 tracing-subscriber.workspace = true
 tracing-appender.workspace = true
 tracing-core.workspace = true
 
-# Metrics
-nvml-wrapper = "0.9.0"
-rgb = "0.8.36"
-terminal_size = "0.2.6"
-textplots = "0.8.0"
-sysinfo = "0.29.8"
-systemstat = "0.2.3"
+# CLI
+indicatif = { version = "0.17.5", optional = true }
+nvml-wrapper = { version = "0.9.0", optional = true }
+rgb = { version = "0.8.36", optional = true }
+terminal_size = { version = "0.2.6", optional = true }
+textplots = { version = "0.8.0", optional = true }
+sysinfo = { version = "0.29.8", optional = true }
+systemstat = { version = "0.2.3", optional = true }
 
 # Utilities
 derive-new = {workspace = true}
--- a/burn-train/Cargo.toml
+++ b/burn-train/Cargo.toml
@@ -11,14 +11,17 @@ repository = "https://github.com/burn-rs/burn/tree/main/burn-train"
 version = "0.9.0"
 
 [features]
+default = ["metrics", "ui"]
+metrics = [
+    "nvml-wrapper",
+    "sysinfo",
+    "systemstat"
+]
 ui = [
     "indicatif",
     "rgb",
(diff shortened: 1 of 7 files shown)


burn-train/Cargo.toml
```diff
@@ -10,21 +10,31 @@ readme = "README.md"
...
 
+[features]
+cli = [
+    "indicatif",
+    "rgb",
+    "terminal_size",
+    "textplots",
+    "nvml-wrapper",
+    "sysinfo",
+    "systemstat"
```
nathanielsimard, line 21: Those are used to compute metrics as well, you may still want them to be computed without the CLI to show them with a plot.

Probably not `nvml-wrapper`, since it's just text, but the other two can be useful regardless.

burn-train/src/metric/dashboard/plot.rs
```diff
@@ -146,4 +147,14 @@ impl TextPlot {
...
     }
+
+    /// Renders the text plot.
+    ///
+    /// # Returns
+    ///
+    /// The rendered text plot.
+    #[cfg(not(feature = "cli"))]
+    pub fn render(&self) -> String {
+        panic!("metrics feature not enabled on burn-train")
```
nathanielsimard, line 158: I think it should be the `cli` feature, there is no `metric` feature.

burn-train/src/metric/mod.rs
```diff
@@ -6,20 +6,30 @@ pub mod state;
...
+#[cfg(feature = "cli")]
 pub use cpu_use::*;
+#[cfg(feature = "cli")]
 pub use cuda::*;
+#[cfg(feature = "cli")]
 pub use gpu_temp::*;
 pub use learning_rate::*;
 pub use loss::*;
+#[cfg(feature = "cli")]
 pub use memory_use::*;
```
nathanielsimard, line 35: Except for the `cuda` metric, they should be enabled even without the cli feature.

burn/Cargo.toml
```diff
@@ -11,12 +11,14 @@ repository = "https://github.com/burn-rs/burn"
...
-default = ["std", "train"]
+default = ["std"]
 experimental-named-tensor = ["burn-core/experimental-named-tensor"]
 std = [
   "burn-core/std",
 ]
-train = ["std", "burn-train"] # Training requires std
+# Training requires std
+train = ["std", "burn-train"]
+train-cli = ["train", "burn-train/cli"]
```
nathanielsimard, line 21: I think `train-cli` should be enabled by default without having to specify `cli`. It is the most common usecase.

nathanielsimard (changes requested):
See my comments bellow. Having a cli feature flag is fine, but I would enable it by default. Also we should have a better implementation without using text-plot, probably using something like ratatui: https://github.com/ratatui-org/ratatui.

burn/Cargo.toml
antimora: Can we also document (inline is fine) that explains the difference between train and train-cli?

burn-train/Cargo.toml
dae: Ok, I've split these off into a separate flag - does that work?

nathanielsimard (approved):
Thanks @dae I'm still unsure about the feature flag names `metrics` and `ui`, but at least we offer the option now!

dae:
Conflicts fixed . No preferences on names - some other ideas: train-tui, train-sys-metrics.

nathanielsimard (approved):

PR #743, opened 2023-08-31
Apply markdown formatting using prettier

antimora:
### Changes

Applied markdown formatting using prettier. 

Moving forward I recommend applying formatting to keep the text on within the margins, and also keep tables tidy. 

Here is VSCode plug-in: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode 

### Testing

`mdbook serve`


louisfd (approved):
That's a good idea

PR #746, opened 2023-09-01
Convert some flaky tests to use approx comparisons

### LINKED ISSUE #741: Failed tests, and initial checks
On my first PR, the test execution of ./run-check.sh took around 15 minutes to run, and was rather daunting - I found myself thinking "will this ever end?". Subsequent runs are much faster. If the initial run is expected to be slow, it might be worth mentioning something in the PR template to re-assure users that it's not going to take forever.

Another issue is that when I run ./run-check.sh, I consistently get the following errors. I presume it's because of my system (Debian OldStable/RX 6600). Any ideas on how I could avoid these errors? I missed some docstring errors in a recent PR, and am not sure if that's because run-check was terminating early, or because the docstring tests are not covered by that script.

```
---- tests::aggregation::tests::test_should_mean stdout ----
thread 'tests::aggregation::tests::test_should_mean' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [2.4999998], shape: Shape { dims: [1] } }`,
 right: `Data { value: [2.5], shape: Shape { dims: [1] } }`', burn-wgpu/src/lib.rs:49:5

---- tests::aggregation::tests::test_should_mean_last_dim stdout ----
thread 'tests::aggregation::tests::test_should_mean_last_dim' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [0.99999994, 3.9999998], shape: Shape { dims: [2, 1] } }`,
 right: `Data { value: [1.0, 4.0], shape: Shape { dims: [2, 1] } }`', burn-wgpu/src/lib.rs:49:5

---- tests::div::tests::should_support_div_ops stdout ----
thread 'tests::div::tests::should_support_div_ops' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [0.0, 1.0, 1.0, 1.0, 1.0, 1.0], shape: Shape { dims: [2, 3] } }`,
 right: `Data { value: [0.0, 1.0, 1.0, 0.99999994, 1.0, 1.0], shape: Shape { dims: [2, 3] } }`', burn-wgpu/src/lib.rs:49:5


failures:
    tests::aggregation::tests::test_should_mean
    tests::aggregation::tests::test_should_mean_last_dim
    tests::div::tests::should_support_div_ops
```

dae:
Closes #741

### Checklist

### DIFF
--- a/burn-tensor/src/tests/ops/aggregation.rs
+++ b/burn-tensor/src/tests/ops/aggregation.rs
@@ -9,7 +9,7 @@ mod tests {
 
         let data_actual = tensor.mean().to_data();
 
-        assert_eq!(data_actual, Data::from([15.0 / 6.0]));
+        data_actual.assert_approx_eq(&Data::from([15.0 / 6.0]), 3);
     }
 
     #[test]
@@ -45,7 +45,7 @@ mod tests {
 
         let data_actual = tensor.mean_dim(1).to_data();
 
-        assert_eq!(data_actual, Data::from([[3.0 / 3.0], [12.0 / 3.0]]));
+        data_actual.assert_approx_eq(&Data::from([[3.0 / 3.0], [12.0 / 3.0]]), 3);
     }
 
     #[test]
--- a/burn-tensor/src/tests/ops/div.rs
+++ b/burn-tensor/src/tests/ops/div.rs
@@ -14,7 +14,7 @@ mod tests {
 
         let data_actual = output.into_data();
         let data_expected = Data::from([[0.0, 1.0, 1.0], [1.0, 1.0, 1.0]]);
-        assert_eq!(data_expected, data_actual);
+        data_expected.assert_approx_eq(&data_actual, 3);
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


antimora (approved):
LGTM

dae:
Sorry, just pushed through a follow-up commit - is that change ok?

antimora:
> Sorry, just pushed through a follow-up commit - is that change ok?

Yes, I can recheck it.

.github/pull_request_template.md
```diff
@@ -2,7 +2,8 @@
 
 ### Checklist
 
-- [ ] Confirm that `run-checks` script has been executed.
+- [ ] Confirm that `run-checks` script has been executed. The first run
+may take around 15 minutes.
```
antimora, line 6: I think this warning should be from the script its self. Besides, by the time, someone reaches the PR, they might have already ran it (we mention it also in CONTRIBUTING.md file, probably worth mentioning it there as well).

antimora (changes requested):
Please see my comment below.

.github/pull_request_template.md
dae: Ok, I've just added a comment at the top of the shell scripts instead. Considered adding it in a print statement at the top of run-checks.rs, but fear that it will get missed as the text scrolls down. WDYT?

nathanielsimard (approved):

antimora (approved):
Yeah, that's fine. We will improve run checks in the future.

PR #747, opened 2023-09-01
Perf/ndarray: Optimize `conv2d` operation

DrChat:
### Checklist

### Changes

1) Eliminate use of unsafe sharing by iterating through axis 0 of the output array.
2) Optimize indexing by converting `IxDyn` arrays into `Ix4` arrays
3) Try to use `&[E]` slices as much as possible instead of indexing operations (note: requires rows to be contiguous)
4) Add a special case for strides and dilations of 1 (seemingly very common) so that the auto-vectorizer can run

### Notes
Apparently this is efficient enough now that it's causing competition for vector resources between CPU hyperthreads (and thus is _faster_ if `rayon` is configured with a number of threads equivalent to physical cores on the system).

### Testing

Locally and against the test suite

### DIFF
--- a/burn-ndarray/src/ops/base.rs
+++ b/burn-ndarray/src/ops/base.rs
@@ -392,30 +392,36 @@ where
         arg(tensor, dim, CmpType::Min)
     }
 
-    pub fn clamp_min<const D: usize>(tensor: NdArrayTensor<E, D>, min: E) -> NdArrayTensor<E, D> {
-        let array = tensor.array.mapv(|x| match x < min {
+    pub fn clamp_min<const D: usize>(
+        mut tensor: NdArrayTensor<E, D>,
+        min: E,
+    ) -> NdArrayTensor<E, D> {
+        tensor.array.mapv_inplace(|x| match x < min {
             true => min,
             false => x,
         });
 
-        NdArrayTensor::new(array.into_shared())
+        tensor
     }
 
-    pub fn clamp_max<const D: usize>(tensor: NdArrayTensor<E, D>, max: E) -> NdArrayTensor<E, D> {
-        let array = tensor.array.mapv(|x| match x > max {
+    pub fn clamp_max<const D: usize>(
+        mut tensor: NdArrayTensor<E, D>,
+        max: E,
+    ) -> NdArrayTensor<E, D> {
+        tensor.array.mapv_inplace(|x| match x > max {
             true => max,
             false => x,
         });
 
-        NdArrayTensor::new(array.into_shared())
+        tensor
     }
 
     pub fn clamp<const D: usize>(
-        tensor: NdArrayTensor<E, D>,
+        mut tensor: NdArrayTensor<E, D>,
         min: E,
         max: E,
     ) -> NdArrayTensor<E, D> {
-        let array = tensor.array.mapv(|x| match x < min {
+        tensor.array.mapv_inplace(|x| match x < min {
             true => min,
             false => match x > max {
                 true => max,
@@ -423,7 +429,7 @@ where
             },
         });
 
-        NdArrayTensor::new(array.into_shared())
+        tensor
     }
 }
 
--- a/burn-ndarray/src/ops/conv.rs
+++ b/burn-ndarray/src/ops/conv.rs
@@ -2,12 +2,50 @@ use burn_tensor::{
     ops::{conv::calculate_conv_output_size, ConvOptions, ConvTransposeOptions},
     ElementConversion,
 };
-use ndarray::{s, Array3, Array4, Axis, Dim};
+use ndarray::{s, Array3, Array4, ArrayView2, ArrayViewMut2, Axis, Dim};
 
 use crate::{
     element::FloatNdArrayElement, iter_par, iter_range_par, ops::padding::apply_padding_4d,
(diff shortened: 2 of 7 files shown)


burn-ndarray/src/ops/conv.rs
```diff
@@ -25,59 +25,119 @@ pub(crate) fn conv2d<E: FloatNdArrayElement>(
...
-                                let ih = oh * stride_height + kh * dilatation_height;
-                                let iw = ow * stride_width + kw * dilatation_width;
+                        let x = x.slice(s![b, ic, .., ..]);
+                        let k = weights.slice(s![oc, weight_ic, .., ..]);
 
-                                let weight_ic = ic - (g * in_channels);
-                                output[[b, oc, oh, ow]] +=
-                                    x[[b, ic, ih, iw]] * weight.array[[oc, weight_ic, kh, kw]];
+                        for kh in 0..kernel_height {
+                            // let kr = &k.row(kh).to_slice().unwrap()[0..kernel_width];
```
nathanielsimard, line 64: Dead code

burn-ndarray/src/ops/conv.rs
```diff
@@ -25,59 +25,119 @@ pub(crate) fn conv2d<E: FloatNdArrayElement>(
...
+                                        for ow in 0..out_width {
+                                            let iw = ow + kw;
+                                            or[ow] += ir[iw] * k;
+                                        }
+                                    }
+                                } else {
+                                    for oh in 0..out_height {
+                                        // Construct a sub-slice view of the input row.
+                                        // This is done upfront so that rustc does not have to emit bounds checks
+                                        // in the hot loop below.
```
nathanielsimard, line 97: I would put those comments above the two duplicated loops explaining that we are building input row (ir) and output row (or) using slices to avoid bounds checks in the hot loops.

burn-ndarray/src/ops/conv.rs
```diff
@@ -25,59 +25,119 @@ pub(crate) fn conv2d<E: FloatNdArrayElement>(
...
-            if let Some(bias) = &bias {
-                for oh in 0..out_height {
-                    for ow in 0..out_width {
-                        output[[b, oc, oh, ow]] += bias.array[oc];
+                    if let Some(bias) = &bias {
+                        let bias = bias.array[oc];
+
+                        for oh in 0..out_height {
+                            let mut or = output.row_mut(oh);
+                            let or = &mut or.as_slice_mut().unwrap()[0..out_width];
```
nathanielsimard, line 122: The code that extract the output row and the input row are duplicated at multiple places, and are quite hard to read. I would extract them into function, maybe with `#[inline(always)]`:

```rust
#[inline(always)]
fn output_row<'a>(output: &'a mut Array3<E>, oh: usize) -> &'a mut [E] {
    let mut or = output.row_mut(oh);
    &mut or.as_slice_mut().unwrap()[0..out_width]
}
```

We could do the same for input row. What are your thoughts?

nathanielsimard (commented):
Thanks a lot @DrChat to improve performance of the conv operation! If you have a chance, can you compare performance with the main branch on your fork using https://github.com/burn-rs/burn-benches

I have a few comments, but it seems good!

louisfd (approved):
This is awesome. 
I did a quick speed check by running the tests 100,000 times the old and new implementations (on my Apple M2), and I got those speedups:
"simple" test (strides & dilations are 1) : from 51s to 12s
"complex" test (strides & dilations vary between 1 and 3) : from 32s to 14s
So, good job!
There was a comment left out in your code, see my other comment, but I'm approving in advance

DrChat:
> Thanks a lot @DrChat to improve performance of the conv operation! If you have a chance, can you compare performance with the main branch on your fork using https://github.com/burn-rs/burn-benches
> 
> I have a few comments, but it seems good!

Huh - I didn't realize that repository existed! I was looking around for benchmarks in this one :P
Will play around with it :)

burn-ndarray/src/ops/base.rs
```diff
@@ -392,38 +392,44 @@ where
         arg(tensor, dim, CmpType::Min)
     }
 
-    pub fn clamp_min<const D: usize>(tensor: NdArrayTensor<E, D>, min: E) -> NdArrayTensor<E, D> {
-        let array = tensor.array.mapv(|x| match x < min {
+    pub fn clamp_min<const D: usize>(
```
DrChat, line 395: This stuff was technically unrelated and could be removed - but I was experimenting with parallel maps in `ndarray`.

burn-ndarray/src/ops/conv.rs
DrChat: Unfortunately, I had to decompose these statements into two because `row_mut` creates an object that `as_slice_mut` borrows from.
In other words, such a function would have to return the `or` variable _and_ the slice created in order to make the borrow checker happy.

I agree that it's a bit dense - but I don't think a helper function would work in this case. The best I can do is add comments explaining everything that is happening here :)

nathanielsimard (approved):

nathanielsimard:
@DrChat Thanks for the PR. I think there is a problem with `burn-benches` due to dependencies clashing, so I will merge it as is and run benches later. It is obviously a significant performance improvement.

PR #753, opened 2023-09-01
Add ability to interrupt training loop

### LINKED ISSUE #714: Ability to interrupt training loop
When integrating burn into a GUI app, it would be nice to give the user the ability to interrupt a long training process if they decide they don't want to wait for it to complete.

The way I thought this might be implemented is allowing the user to pass in a `should_stop: Arc<AtomicBool>` to LearnerBuilder, which would pass it on to the learner. The learner would pass it on to run(_multi_device), and at the end of each loop iteration, if the flag is set, it would terminate processing. To minimize API changes, I thought it might be best to return the partially-changed model, rather than have it return an error flag.

Some possible downsides: AtomicBool is not [universally available](https://doc.rust-lang.org/std/sync/atomic/#portability); there may be a minor performance hit.

Is this something you guys would be amenable to if I write up a PR? Do you have alternative thoughts on how it might be implemented?

dae:
Closes #714

### Checklist

### Changes

@nathanielsimard I have a mild preference for the API in the first commit, as builders typically have data passed into them instead of out from them. If we wanted to share the underlying AtomicBool with other items, couldn't we also accomplish that in the future by adding a pub(crate) fn to the interrupter that returns its inner value? But not a big deal; either way works for me.

### Testing

Manual testing to confirm .stop() interrupts training.

### DIFF
--- a/burn-train/src/learner/base.rs
+++ b/burn-train/src/learner/base.rs
@@ -4,6 +4,8 @@ use burn_core::lr_scheduler::LRScheduler;
 use burn_core::module::{ADModule, Module};
 use burn_core::optim::Optimizer;
 use burn_core::tensor::backend::ADBackend;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
 
 /// Learner struct encapsulating all components necessary to train a Neural Network model.
 ///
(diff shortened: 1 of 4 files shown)


nathanielsimard (approved):

nathanielsimard:
@dae I think this version is less error-prone, but it's true that it is a bit weird to have the data going out of the builder instead of in. I think we might put the function `interceptor` on the built learner instead at some point to be more consistent, but I will merge this as is for now since it fixes the problem.

PR #739, opened 2023-08-31
Migrate from log4rs to tracing

dae:
update_log_file() is no longer public, as tracing expects things to only ever be initialized once. If the caller wants their own custom logging, they can set up their own tracing subscriber prior to training. This means we no longer need the log_to_file option in LearnerBuilder.

The logging output format is similar but not identical (eg module name is shown instead of file:line). This can be configured if the old format is preferable.

Closes #732

This removes a security warning:

error[vulnerability]: Potential segfault in the time crate
    │
487 │ time 0.1.45 registry+https://github.com/rust-lang/crates.io-index
    │ ----------------------------------------------------------------- security vulnerability detected

### Checklist

- [x] Confirm that `run-checks` script has been executed.

### Related Issues/PRs

#732

PR #754, opened 2023-09-02
Fix: log file creation could not be avoided

dae:
In #739 I removed log_to_file(), thinking that the logging would be avoided when a global log handler is already installed. But it appears that the file is created eagerly by the previous statement, so an option to opt out of its creation is still required.

https://github.com/burn-rs/burn/blob/3669d2a6d46f99d81296de8d9fb327b3e3b8da2d/burn-train/src/learner/log.rs#L11

### Checklist


nathanielsimard (approved):
Well, let's put the option back in 😅, being explicit isn't much of a problem here.

dae:
Thanks, sorry about the churn.

PR #755, opened 2023-09-02
Cross Entropy with label smoothing and weights.

ArvidHammarlund:
### Checklist

- [ X] Confirm that `run-checks` script has been executed.

### Changes

Added label smoothing and weights to cross entropy.

### Testing

Label smoothing requires a different implementation of the loss calculation, so functionality regarding that are tested against the default implementation with label smoothing = 0. Others are tested against manual implementations.


nathanielsimard (changes requested):
Thanks @ArvidHammarlund for the PR; it's a nice improvement of the current implementation.

Since the loss is starting to get more complicated with more options, I would make it a module by adding the `Module` derive. I would also add config, similar to other modules, which will avoid having to manually implement builder methods.

We could keep a `default` function to create the Loss:

```rust
impl Default for CrossEntropyLoss {
    fn default() -> Self {
        CrossEntropyLossConfig::new().init()
    }
}
```

But when using weights tensors, it's probably better to initialize the loss function within our module. By implementing the module trait, we will ensure that the weights are always on the same device as the other weights, which isn't checked with the current implementation.

ArvidHammarlund:
Good points! I will try to make the changes later today.

ArvidHammarlund:
I have now made the changes. I also added a binary version, and made it possible to use multiple pad tokens as well.

burn-core/src/nn/loss/cross_entropy.rs
```diff
@@ -25,36 +111,128 @@ impl<B: Backend> CrossEntropyLoss<B> {
...
+    }
+
+    fn forward_binary(&self, logits: Tensor<B, 2>, mut targets: Tensor<B, 1, Int>) -> Tensor<B, 1> {
+        let [batch_size, nr_classes] = logits.dims();
+
+        if let Some(alpha) = self.smoothing {
+            targets = targets * (1. - alpha) + alpha / nr_classes as f32;
+        }
+
+        let logits = logits.slice([0..batch_size, 0..1]).reshape([batch_size]);
```
nathanielsimard, line 128: I'm not sure we need the `slice`, just the reshape should do the job; if the logits have more than just one element per batch, I think it's better to panic than to train with bugs.

burn-core/src/nn/loss/cross_entropy.rs
```diff
@@ -1,21 +1,107 @@
...
+    smoothing: Option<f32>,
+
+    /// Create binary cross-entrop.
+    ///
+    /// Will reduce computation by interpolating secondary label.  
+    /// Should only be given labels Tensor with {0, 1}.
+    /// If weights are also used, only two should be given.     
+    /// Cannot use pad_token at the same time.
+    #[config(default = false)]
+    binary: bool,
```
nathanielsimard, line 37: I think we should support binary cross entropy with another module. The forward pass should have a different signature (only tensors of rank 1), the config options won't support padding and the weights config could be `[usize; 2]`. What do your think?

burn-core/src/nn/loss/cross_entropy.rs
```diff
@@ -1,21 +1,107 @@
...
+}
 
 /// Calculate the cross entropy loss from the input logits and the targets.
-#[derive(Clone, Debug)]
+#[derive(Module, Debug)]
 pub struct CrossEntropyLoss<B: Backend> {
-    pad_index: Option<usize>,
-    backend: PhantomData<B>,
+    pad_tokens: Option<Vec<usize>>,
+    weights: Option<Tensor<B, 1>>,
```
nathanielsimard, line 88: I would put the weights public, so we could print them, or use a tensor from someplace else.

nathanielsimard (commented):
Very good job 🙏! I'm unsure if we should mix binary cross entropy with the normal cross entropy, curious to hear your thoughts!

ArvidHammarlund:
I believe moving out the binary version would be an easy compile time check win (logits of dimension = 1), which is good; the only downside I could think of is that docs.rs can grow quite daunting to discern when codegen generate three additional items for every struct - maybe I could move record and recordItem to a subdirectory like nn::loss::records? Btw, should I make run time assertions for every forward pass, checking if e.g. no label are assigned an int greater than weights tensor length, or would that cause to high overhead?

nathanielsimard:
> I believe moving out the binary version would be an easy compile time check win (logits of dimension = 1), which is good; the only downside I could think of is that docs.rs can grow quite daunting to discern when codegen generate three additional items for every struct - maybe I could move record and recordItem to a subdirectory like nn::loss::records? Btw, should I make run time assertions for every forward pass, checking if e.g. no label are assigned an int greater than weights tensor length, or would that cause too high overhead?

1. About the docs, since the code is generated, the solution would be in `burn-derive`, so behond the scope of this PR.
2. Checks are very cheap at that level, so it won't be a performance problem.

ArvidHammarlund:
Ok. I am doing the revisions now!

ArvidHammarlund:
The smoothing parameter makes some buffers time out on WPGU backend, so this PR shouldn't be approved yet!

burn-core/src/nn/loss/cross_entropy.rs
```diff
@@ -25,85 +97,451 @@ impl<B: Backend> CrossEntropyLoss<B> {
...
+        assert!(
+            max_target
+                .clone()
+                .lower_elem(logits.dims()[1] as i32)
+                .into_data()
+                .value[0],
+            "Encounter target ({}) greater than feature dimension ({}).",
+            max_target,
+            logits_width
+        );
```
nathanielsimard, line 228: I would remove that assertion, asserting shapes is fine, but when it comes to the underlying tensor data, it's normally too slow.

burn-core/src/nn/loss/cross_entropy.rs
```diff
@@ -25,85 +97,451 @@ impl<B: Backend> CrossEntropyLoss<B> {
...
+#[derive(Config, Debug)]
+pub struct BinaryCrossEntropyLossConfig {
+    /// Create weighted binary cross-entropy.
+    ///
+    /// The loss of a specific sample will simply be given by: weight[y] * log(p(x)) * 1,
+    ///
+    /// # Pre-conditions
+    ///   - The order of the weight vector should correspond to the label integer assignment.
+    ///   - Targets assigned negative Int's will not be allowed.
+    pub weights: Option<Vec<f32>>,
```
nathanielsimard, line 242: Can it be `[f32; 2]` instead?

nathanielsimard (changes requested):
Some checks are too expensive to call during the forward pass. I would also move the binary cross entropy loss into another file `binary_cross_entropy.rs`. 

I think you commited the file `script/run-checks` is it a mistake?

ArvidHammarlund:
I made the changes now. 

> I think you commited the file `script/run-checks` is it a mistake?

Yes, I thought that was due to me adding units tests?  But I probably shouldn't use git add -A in such large code bases, anyways.

ArvidHammarlund:
Cargo says that the slice method with an array as index is not valid syntax for me. Is there something I am missing?

nathanielsimard:
> Cargo says that the slice method with an array as index is not valid syntax for me. Is there something I am missing?

I think you have to allow that warning in the project: `#![allow(clippy::single-range-in-vec-init)]`

nathanielsimard (approved):

nathanielsimard:
@ArvidHammarlund I'll fix the CI 😅

Issue #756, opened 2023-09-02
Replace `run-checks.[ps1/sh]` with `cargo-xtask` pattern instead


DrChat:
Using the [`cargo-xtask`](https://github.com/matklad/cargo-xtask) pattern for implementing this script would be a better natural fit instead of the current setup.
This also gains us the benefit that this can be expanded in the future.

For developers, the invocation of these checks would be changed from:
```
> run-checks.ps1
```

to:
```
> cargo xtask run-checks
```

This should be pretty straightforward to port over, as the script is already written in Rust :)

antimora:
Cc'ing @Luni-4

antimora:
Closing as complete via #757

PR #757, opened 2023-09-02
Add `cargo-xtask` helper and move scripts into it

DrChat:
### Checklist

### Related Issues/PRs

#756

### Changes

Move all helper scripts into a new `cargo-xtask` binary, following the [`cargo-xtask`](https://github.com/matklad/cargo-xtask) pattern.

The new way to run these helper scripts is via (e.g.):
```
> cargo xtask run-checks
```

### DIFF
--- a/.github/workflows/publish-template.yml
+++ b/.github/workflows/publish-template.yml
@@ -18,10 +18,7 @@ jobs:
       - name: install rust
         uses: dtolnay/rust-toolchain@stable
 
-      - name: compile publish binary
-        run: rustc scripts/publish.rs --crate-type bin --out-dir scripts
-
       - name: publish to crates.io
-        run: ./scripts/publish ${{ inputs.crate }}
+        run: cargo xtask publish ${{ inputs.crate }}
         env:
           CRATES_IO_API_TOKEN: ${{ secrets.CRATES_IO_API_TOKEN }}
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.75"
+clap = { version = "4.4.2", features = ["derive"] }
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,0 +1,34 @@
+use clap::{Parser, Subcommand};
+
+mod publish;
+mod runchecks;
+
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand)]
+enum Command {
+    /// Publish a crate to crates.io
+    Publish {
+        /// The name of the crate to publish on crates.io
+        name: String,
+    },
+    /// Run the specified `burn` tests and checks locally.
+    RunChecks {
+        /// The environment to run checks against
+        env: runchecks::CheckType,
+    },
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+
+    match args.command {
+        Command::RunChecks { env } => runchecks::run(env),
+        Command::Publish { name } => publish::run(name),
+    }
+}
--- a/xtask/src/runchecks.rs
+++ b/xtask/src/runchecks.rs
@@ -302,8 +302,22 @@ fn check_examples() {
     });
 }
 
+#[derive(clap::ValueEnum, Default, Copy, Clone, PartialEq, Eq)]
+pub enum CheckType {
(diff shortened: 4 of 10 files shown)


.github/workflows/publish-template.yml
```diff
@@ -18,10 +18,7 @@ jobs:
       - name: install rust
         uses: dtolnay/rust-toolchain@stable
 
-      - name: compile publish binary
-        run: rustc scripts/publish.rs --crate-type bin --out-dir scripts
-
       - name: publish to crates.io
-        run: ./scripts/publish ${{ inputs.crate }}
+        run: cargo-xtask publish ${{ inputs.crate }}
```
nathanielsimard, line 22: is `cargo-xtask` supposed to be `cargo xtask` ?

nathanielsimard (commented):
Very cool indeed thanks for the improvement! With the work from @Luni-4, the CI is starting to get pretty neat! 

I just have one comment, maybe a little mistake, see bellow!

.github/workflows/publish-template.yml
DrChat: Oops - yes it is :)

xtask/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.75"
```
Luni-4, line 9: Can we use `thiserror` instead of `anyhow`? In this way we can set our own kinds of errors

xtask/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+        env: runchecks::CheckType,
+    },
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+
+    match args.command {
+        Command::RunChecks { env } => runchecks::run(env),
+        Command::Publish { name } => publish::run(name),
```
Luni-4, line 32: Can we unwrap the `run` functions here instead of returning an error?

Luni-4 (changes requested):
Thanks a lot for your PR! 

Can we just write the advantages and disadvantages in using the `xtask` approach in a file called `INFRASTRUCTURE.md`? Just to take note why we have chosen this approach

xtask/Cargo.toml
DrChat: I would strongly recommend against using `thiserror` for a binary crate, as there is not much reason here to explicitly define error variants for programmatic scrutiny.
It's more important that the errors get propagated up to users with appropriate context, so they understand fully what caused the error and what we were trying to do when the error occurred.

`thiserror` is more appropriate for use in a library crate, where the consumer may want to handle the errors and/or return them up the stack.

xtask/src/main.rs
DrChat: I would recommend against doing this, as unwrapping here makes it appear to users that `cargo-xtask` had encountered an internal error.
It's best to return the result so that the error gets printed and the program exits with a normal error status code.

Luni-4: I agree with this analysis

xtask/Cargo.toml
Luni-4: Yes, seems legit. [Here](https://github.com/dtolnay/anyhow#comparison-to-thiserror) an official reference of your comments

Luni-4 (approved):

nathanielsimard (approved):
The only remaining part is fixing the merge conflicts!

xtask/src/runchecks.rs
```diff
@@ -313,18 +327,12 @@ fn main() {
...
-        Some("typos") => check_typos(),
-        Some("examples") => check_examples(),
-        Some(other) => panic!("Unexpected test type: {}", other),
-        None => {
+    match env {
+        CheckType::Std => std_checks(),
+        CheckType::NoStd => no_std_checks(),
+        CheckType::Typos => check_typos(),
+        CheckType::Examples => check_examples(),
+        CheckType::All => {
```
dae, line 335: I fear this has reintroduced the problem I tried to solve with https://github.com/burn-rs/burn/pull/762/commits/824cd6ae384adbc29e587b559ab37b6334ebfcd1, where any invalid input is treated as "all"

Luni-4: I think we can return an error if another kind of command has been inserted, what do you think @DrChat and @dae?

nathanielsimard: Already cover, since it uses clap, it's validated when parsing the input:

```bash
$ cargo xtask run-checks tt
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/xtask run-checks tt`
error: invalid value 'tt' for '<ENV>'
  [possible values: all, std, no-std, typos, examples]

For more information, try '--help'.
```

nathanielsimard (approved):

PR #758, opened 2023-09-03
Exclude image crate when sqlite feature disabled

dae:
Not needed for anything but huggingface, which is unavailable when sqlite is unavailable.

### Checklist


nathanielsimard (approved):
For now it's fine. I think we may have multiple feature flags for different modalities at some point, so it won't be linked with sqlite then.

PR #759, opened 2023-09-03
Re-export some submodules and add feature flags for them

dae:
## Pull Request Template

Implements some of the suggestions from #729. I think the type aliases will be useful for people new to Burn, but happy to cut them or move them if you prefer.

Is this the direction you were hoping to go with the meta crate?

### Checklist

### DIFF
--- a/examples/guide/examples/guide.rs
+++ b/examples/guide/examples/guide.rs
@@ -1,11 +1,12 @@
+use burn::backend::wgpu::AutoGraphicsApi;
 use burn::backend::{WgpuAutodiffBackend, WgpuBackend};
 use burn::data::dataset::Dataset;
 use burn::optim::AdamConfig;
 use guide::{model::ModelConfig, training::TrainingConfig};
 
 fn main() {
-    type MyBackend = WgpuBackend;
-    type MyAutodiffBackend = WgpuAutodiffBackend;
+    type MyBackend = WgpuBackend<AutoGraphicsApi, f32, i32>;
+    type MyAutodiffBackend = WgpuAutodiffBackend<AutoGraphicsApi, f32, i32>;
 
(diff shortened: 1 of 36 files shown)


examples/guide/examples/guide.rs
```diff
@@ -1,12 +1,13 @@
...
+use burn::data::dataset::Dataset;
 use burn::optim::AdamConfig;
-use burn_dataset::Dataset;
 use guide::{model::ModelConfig, training::TrainingConfig};
 
 fn main() {
-    type MyBackend = burn_wgpu::WgpuBackend<burn_wgpu::AutoGraphicsApi, f32, i32>;
-    type MyAutodiffBackend = burn_autodiff::ADBackendDecorator<MyBackend>;
+    type MyBackend = WgpuBackend;
+    type MyAutodiffBackend = WgpuAutodiffBackend;
```
nathanielsimard, line 8: For this example, let's keep the generics. It's documented this way in the guide and shows how to set the element types! But having defaults is indeed convenient.

nathanielsimard (commented):
This is pretty much what I had in mind! I'm still unsure if the `backend` module and the `autodiff` re-export should be implemented in `burn-core` vs in `burn`. It might reduce the number of feature flags duplication, but won't change the end-user API.

@dae Let me know what you think is the better way, I don't have a strong preference.

dae:
Because of the fact that burn-core is already using autodiff and already has the test-{backend} flags, it seems easier to keep the current setup despite the extra boilerplate routing the settings - at least it's consistent with the other exports like dataset that way. In the longer term it might be worth revisiting the crate layout - if burn-core focused on only stuff that needed to be shared between the different subcrates (or perhaps were merged into burn-common), that would make things a bit simpler when navigating around the packages.

nathanielsimard (approved):

PR #760, opened 2023-09-03
Doc: improve config builder docs

nathanielsimard:
Small enhancement where the docs from config fields are copied to their corresponding builder methods for better DX.

image


louisfd (approved):
Very cool, works well

PR #762, opened 2023-09-03
MSRV bumps

dae:
## Pull Request Template

I found burn was failing to compile on an older Rust. Quite a big jump was required I'm afraid.

### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: install rust
-        uses: dtolnay/rust-toolchain@1.65
+        uses: dtolnay/rust-toolchain@1.67
         with:
           components: rustfmt, clippy
 
@@ -58,7 +58,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: install rust
-        uses: dtolnay/rust-toolchain@1.65
+        uses: dtolnay/rust-toolchain@1.67
         with:
           components: rustfmt, clippy
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,9 +7,17 @@ on:
   pull_request:
     types: [opened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  test-burn-std:
+  tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust: [stable, 1.71.0]
+        test: [std, no-std]
     steps:
       - name: free disk space
         run: |
@@ -20,18 +28,19 @@ jobs:
           df -h
 
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
 
       - name: install rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@master
         with:
           components: rustfmt, clippy
+          toolchain: ${{ matrix.rust }}
 
       - name: caching
         uses: Swatinem/rust-cache@v2
         with:
           key: ${{ runner.os }}-rust-${{ hashFiles('**/Cargo.lock') }}
-          prefix-key: "v1-rust"
+          prefix-key: "v2-rust"
 
       - name: install llvmpipe and lavapipe
         run: |
@@ -41,40 +50,13 @@ jobs:
           sudo apt install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
 
       - name: run checks & tests
-        run: ./run-checks.sh std
-
-  test-burn-no-std:
-    runs-on: ubuntu-latest
-    steps:
-      - name: free disk space
-        run: |
-          df -h
-          sudo swapoff -a
-          sudo rm -f /swapfile
-          sudo apt clean
-          df -h
-
-      - name: checkout
-        uses: actions/checkout@v3
-
-      - name: install rust
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          components: rustfmt, clippy
-
-      - name: caching
-        uses: Swatinem/rust-cache@v2
-        with:
-          key: ${{ runner.os }}-rust-${{ hashFiles('**/Cargo.lock') }}
-
-      - name: run checks & tests
-        run: ./run-checks.sh no_std
+        run: ./run-checks.sh ${{ matrix.test }}
 
   check-typos:
     runs-on: ubuntu-latest
     steps:
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
 
       - name: run spelling checks using typos
         run: ./run-checks.sh typos
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         rust: [stable, 1.71.0]
-        test: [std, no-std]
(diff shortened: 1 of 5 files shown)


dae:
Re the last commit, an alternative approach would be to pin clippy to a specific version. That would ensure CI doesn't break when a new Rust release comes out, but will mean someone needs to bump the version in run-checks.rs after each release.

Luni-4:
I'm not convinced about pinning a MSRV version in CI, using the latest stable version would allow us to use the newest features stabilized by the Rust language. I propose instead to add a new CI script that runs to a specific stable version chosen by the community as MSRV. This leads to an additional CI overhead though. Perhaps we can update https://github.com/burn-rs/burn/issues/552 first in order to have a common decision and then integrate that decision in this PR. What do you think @dae?

dae:
(personally speaking, especially for a pre 1.0 crate, I think a rather small MSRV range makes sense, so I'm definitely not advocating that burn support ancient versions. I just think that if you've decided to support x versions back, CI should warn you if you're not doing that)

Luni-4:
> > using the latest stable version would allow us to use the newest features stabilized by the Rust language
> 
> If you're using the newest features, doesn't that mean you're breaking the MSRV? The logic behind using MSRV for the CI tests is to catch accidental usages of things that the MSRV doesn't support.

Yes, I agree with that, but in this PR we are replacing the latest stable version with a specific pinned version. In my opinion we can obtain what you want with an additional CI job called `MSRV-check` or something like. For what concerns the support of x versions back part, I think we should talk about that matter in the issue linked above

dae:
How do you enforce MSRV without pinning to a specific version? The intention is that the pin is updated over time as it makes sense to bump the MSRV version. 

Newer Rust releases don't typically break code targeting an older Rust version, but a separate CI run that tests against the latest version instead of the MSRV could certainly be added if it's felt that's needed.

Luni-4:
I was thinking something like this:

```yaml
  test-burn-std-1.72.0:
   test-burn-std:
    runs-on: ubuntu-latest
    steps:
      - name: free disk space
        run: |
          df -h
          sudo swapoff -a
          sudo rm -f /swapfile
          sudo apt clean
          df -h

      - name: checkout
        uses: actions/checkout@v3

      - name: install rust 1.72
        uses: dtolnay/rust-toolchain@1.72
        with:
          components: clippy, rustfmt

      - name: install rust 1.71
        uses: dtolnay/rust-toolchain@1.71
        with:
          components: rustfmt

      - name: caching
        uses: Swatinem/rust-cache@v2
        with:
          key: ${{ runner.os }}-rust-${{ hashFiles('**/Cargo.lock') }}

      - name: install llvmpipe and lavapipe
        run: |
          sudo apt-get update -y -qq
          sudo add-apt-repository ppa:kisak/kisak-mesa -y
          sudo apt-get update
          sudo apt install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers

      - name: run checks & tests
        run: ./run-checks.sh std

  test-burn-std:
    runs-on: ubuntu-latest
    steps:
      - name: free disk space
        run: |
          df -h
          sudo swapoff -a
          sudo rm -f /swapfile
          sudo apt clean
          df -h

      - name: checkout
        uses: actions/checkout@v3

      - name: install rust
        uses: dtolnay/rust-toolchain@stable
        with:
          components: rustfmt, clippy

      - name: caching
        uses: Swatinem/rust-cache@v2
        with:
          key: ${{ runner.os }}-rust-${{ hashFiles('**/Cargo.lock') }}

      - name: install llvmpipe and lavapipe
        run: |
          sudo apt-get update -y -qq
          sudo add-apt-repository ppa:kisak/kisak-mesa -y
          sudo apt-get update
          sudo apt install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers

      - name: run checks & tests
        run: ./run-checks.sh std
```

Just used `std` tests. But we can reduce duplication code using [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)

Luni-4:
I think we can have both of them in CI. This will lead to the following 4 jobs, assuming `1.72` as chosen MSRV:

- `test-std-1.72`
- `test-std`
- `test-no-std-1.72`
- `test-no-std`

For what concerns `test-std-*`s  jobs, we need to decide if we want them for `Windows` and `Macos` systems too.

I propose to run `MSRV` checks only for Linux to avoid stressing out our CI, so having the 4 jobs described above

dae:
Ok, I've switched it to a matrix and added separate tests for 1.71. GHA were failing with a mysterious error (including on the main branch: https://github.com/burn-rs/burn/actions/runs/6074535860); based on some comments on https://github.com/actions/runner/issues/1923 I tried switching to checkout@v2, and that seems to have worked around it.

nathanielsimard (approved):
LGTM thanks a lot for improving the CI. Waiting from final feedback from @Luni-4 before merging

.github/workflows/test.yml
```diff
@@ -20,18 +28,19 @@ jobs:
           df -h
 
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
```
Luni-4, line 31: We can use the new `v4` action

.github/workflows/test.yml
```diff
@@ -20,18 +28,19 @@ jobs:
...
         with:
           components: rustfmt, clippy
+          toolchain: ${{ matrix.rust }}
 
       - name: caching
         uses: Swatinem/rust-cache@v2
         with:
           key: ${{ runner.os }}-rust-${{ hashFiles('**/Cargo.lock') }}
-          prefix-key: "v1-rust"
+          prefix-key: "v2-rust"
```
Luni-4, line 43: Let's use `v4` here too

.github/workflows/test.yml
```diff
@@ -41,40 +50,13 @@ jobs:
...
-      - name: run checks & tests
-        run: ./run-checks.sh no_std
+        run: RUST_MATRIX=${{ matrix.rust }} ./run-checks.sh ${{ matrix.test }}
 
   check-typos:
     runs-on: ubuntu-latest
     steps:
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
```
Luni-4, line 59: Let's use `v4` here too

scripts/run-checks.rs
```diff
@@ -130,6 +130,11 @@ fn cargo_fmt() {
 
 // Run cargo clippy command
 fn cargo_clippy() {
+    if let Ok(matrix) = std::env::var("RUST_MATRIX") {
+        if matrix != "stable" {
+            return;
+        }
+    }
     // Run cargo clippy
```
Luni-4, line 138: We can use 

```rust
if std::env::var("RUST_MATRIX").map_or(false, |matrix| matrix != "stable") {
    return;
}
```

scripts/run-checks.rs
```diff
@@ -320,7 +325,8 @@ fn main() {
         Some("no_std") => no_std_checks(),
         Some("typos") => check_typos(),
         Some("examples") => check_examples(),
-        Some(_) | None => {
+        Some(other) => panic!("Unexpected test type: {}", other),
```
Luni-4, line 328: Thank you for catching this!

Luni-4 (changes requested):
Thanks a lot for your changes @dae! We are nearly there!

I agree with the `1.71` version as `MSRV`

Luni-4 (approved):
Thank you, it's fine for me now!

nathanielsimard (approved):

PR #771, opened 2023-09-05
Added Dim-iter

ArvidHammarlund:
### Checklist

### Changes

Added a dimension iterator method for easy and quick implementations of RNNs.

### Testing

Changed the LSTM and GRU implementations to use the method.

### DIFF
--- a/burn-core/src/nn/rnn/gru.rs
+++ b/burn-core/src/nn/rnn/gru.rs
@@ -118,11 +118,13 @@ impl<B: Backend> Gru<B> {
             None => Tensor::zeros([self.batch_size, seq_length, self.d_hidden]),
         };
 
-        for t in 0..seq_length {
-            let indices = Tensor::arange(t..t + 1);
-            let input_t = batched_input.clone().select(1, indices.clone()).squeeze(1);
-            let hidden_t = hidden_state.clone().select(1, indices.clone()).squeeze(1);
-
+        let [batch, _, d] = batched_input.dims();
+        for (t, (input_t, hidden_t)) in batched_input
+            .iter_dim(1)
+            .zip(hidden_state.clone().iter_dim(1))
+            .map(|(a, b)| (a.reshape([batch, d]), b.reshape([batch, d])))
+            .enumerate()
+        {
             // u(pdate)g(ate) tensors
             let biased_ug_input_sum = self.gate_product(&input_t, &hidden_t, &self.update_gate);
             let update_values = activation::sigmoid(biased_ug_input_sum); // Colloquially referred to as z(t)
--- a/burn-core/src/nn/rnn/lstm.rs
+++ b/burn-core/src/nn/rnn/lstm.rs
@@ -139,9 +139,12 @@ impl<B: Backend> Lstm<B> {
             ),
         };
 
-        for t in 0..seq_length {
-            let indices = Tensor::arange(t..t + 1);
-            let input_t = batched_input.clone().select(1, indices).squeeze(1);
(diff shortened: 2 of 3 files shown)


louisfd (changes requested):
Like for your other PR, I think some tests would be great for your iter_dim addition. In particular, I would test it on non-contiguous tensors. 
It's particularly important since you're adding to the public API in tensor/api/base.rs

ArvidHammarlund:
I added some tests now. How do you run the tests of burn-tensor/src/test directly, btw? Did only manage to run it through the script that runs the tests of the whole project.

nathanielsimard:
@ArvidHammarlund you run them with any backend, so you could cd in `burn-tch` then call `cargo test` and they will run.

burn-core/src/nn/rnn/gru.rs
```diff
@@ -118,11 +118,13 @@ impl<B: Backend> Gru<B> {
...
-        for t in 0..seq_length {
-            let indices = Tensor::arange(t..t + 1);
-            let input_t = batched_input.clone().select(1, indices.clone()).squeeze(1);
-            let hidden_t = hidden_state.clone().select(1, indices.clone()).squeeze(1);
-
+        let [batch, _, d] = batched_input.dims();
+        for (t, (input_t, hidden_t)) in batched_input
+            .iter_dim(1)
+            .zip(hidden_state.clone().iter_dim(1))
+            .map(|(a, b)| (a.reshape([batch, d]), b.reshape([batch, d])))
```
nathanielsimard, line 125: Could squeeze works here?

burn-core/src/nn/rnn/lstm.rs
```diff
@@ -139,9 +139,12 @@ impl<B: Backend> Lstm<B> {
...
 
-        for t in 0..seq_length {
-            let indices = Tensor::arange(t..t + 1);
-            let input_t = batched_input.clone().select(1, indices).squeeze(1);
+        let [batch, _, d] = batched_input.dims();
+        for (t, input_t) in batched_input
+            .iter_dim(1)
+            .map(|e| e.reshape([batch, d]))
+            .enumerate()
+        {
```
nathanielsimard, line 147: I'm not sure it's an improvement here, working with indices in that context in more similar with the math annotations of RNN, so it may be easier to read then when using iterators. I'm not againts providing an iterator in our tensor API, but I'm not sure this is the best place to use it.

nathanielsimard, line 147: I'm not sure it's an improvement here, working with indices in that context in more similar with the math annotations of RNN, so it may be easier to read then when using iterators. I'm not againts providing an iterator in our tensor API, but I'm not sure this is the best place to use it.

nathanielsimard (commented):
I'm not sure it's an improvement over using for loops in RNNs. Working with indices in that context is more similar to the math annotations, so it may be easier to read than when using iterators. I'm not against providing an iterator in our tensor API, but I'm not sure this is the best place to use it.

I'm sure it can be very useful in batch preprocessing and even in some modeling code, but not convinced in the case of RNNs. What are your thoughts, @louisfd and @agelas?

agelas:
@nathanielsimard @ArvidHammarlund This approach certainly looks a lot more Rusty! I agree that incorporating iterators into the tensor API is valuable. Unless there's a compelling performance reason, and if readability is a priority, then the original might be more suitable for this specific context due to its direct alignment with the mathematical representation of RNNs. No hard preference for either approach from me though.

nathanielsimard (approved):

PR #772, opened 2023-09-05
Fixed bug in binary cross entropy

ArvidHammarlund:
Fixed a bug in the implementation of weights for binary cross-entropy.


nathanielsimard (changes requested):
Can you add a test that validates the bugfix? Always a good idea ;)

ArvidHammarlund:
There was another bug as well😅! I didn't think of doing another test of the weights when I changed the implementation of the binary version from the general one, my bad! Will be more generous with tests going forward! Added a test for the binary version of smoothing as well.

nathanielsimard (approved):

PR #784, opened 2023-09-08
Avoid building `xtask` when invoking `cargo build`

DrChat:
Follow-up to #757

We can't run `cargo build` as-is anymore since it would attempt to rebuild the crate that is invoking the command.

### DIFF
--- a/xtask/src/runchecks.rs
+++ b/xtask/src/runchecks.rs
@@ -234,7 +234,7 @@ fn std_checks() {
     println!("Running std checks");
 
     // Build each workspace
-    cargo_build(&["--workspace"]);
+    cargo_build(&["--workspace", "--exclude=xtask"]);
 
     // Test each workspace
     cargo_test(&["--workspace"]);


xtask/src/runchecks.rs
```diff
@@ -234,7 +234,7 @@ fn std_checks() {
     println!("Running std checks");
 
     // Build each workspace
-    cargo_build(&["--workspace"]);
+    cargo_build(&["--workspace", "--exclude=xtask"]);
```
antimora, line 237: Maybe we should exclude from the workspace (in root's Cargo.toml), like the following?

```toml
exclude = [
    "examples/notebook",
    "xtask",
]
```

Was there a reason we made xtask as a member of workspace?

DrChat: It's part of the original `cargo-xtask` template - in order to invoke `cargo xtask` from anywhere within the workspace, it needs to be a workspace member

antimora: Reading through [their documentation](https://github.com/matklad/cargo-xtask) it appears, they allow option not to be a member as well. It seems like it might be needed if there are some related dependencies. 

I am okay either way. I am just bringing this up as an option.

DrChat: Relevant portion:
> The xtask crate may or may not be a part of the main workspace. Usually, but not always, the workspace setup is better. If xtask is a part of the workspace, you can share dependencies between xtask and main crates, and dependencies update process is easier. Additionally, you will be able to use xtask = "run --package xtask --" as an alias, which works regardless of Cargo's working directory. If xtask is not a part of the workspace, you can use different feature sets for shared dependencies, and you can cache xtask/target more easily on CI. It is advisable to commit xtask lockfile to the repository.

I have no strong opinions either way, but it's seemingly preferable to keep it as a member of the workspace unless there is a good reason to do otherwise :)

antimora (approved):

xtask/src/runchecks.rs
Luni-4: I agree with @DrChat, let's exclude it from compilation using the `exclude` option

nathanielsimard (approved):

PR #789, opened 2023-09-08
Made docs show features not included in std.

ArvidHammarlund:
title

### DIFF
--- a/burn-core/src/backend.rs
+++ b/burn-core/src/backend.rs
@@ -1,30 +1,41 @@
 /// Ndarray module.
+#[cfg_attr(docsrs, doc(cfg(feature = "ndarray")))]
 #[cfg(feature = "ndarray")]
 pub use burn_ndarray as ndarray;
+
+#[cfg_attr(docsrs, doc(cfg(feature = "ndarray")))]
 #[cfg(feature = "ndarray")]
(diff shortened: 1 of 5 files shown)


antimora:
Great! Can we also do the same for audio feature flag of burn-dataset?

ArvidHammarlund:
Done!

nathanielsimard (commented):
Would it be easier to add a compilation flag `doc`, which could be activated when building the docs before a release? I feel like we won't need to go and add `cfg_attr` over each compilation flag in the repo.

```toml
[features]
doc = ["ndarray", "tch", "autodiff", "wgpu", ...]
```

ArvidHammarlund:
In that case I think we only need to keep
 ``` 
[package.metadata.docs.rs]
all-features = true
```
since that would, according to my understanding, show all features on doc.rs which is the thing of most importance imo, and would do the same thing as you are suggesting. The cfg tags, on the other hand, just auto includes the features you selected when building the docs from the command line.

I just looked how Tokio was doing it, and they included both, but that might be unnecessary when I think about it.

Another way to get the second feature would be to use the nightly feature doc_cfg_auto.

nathanielsimard (approved):

PR #793, opened 2023-09-09
Added covariance and diagonal ops.

ArvidHammarlund:
### Checklist

### Changes

Needed covariance and diagonal matrices for a project which might be useful for others as well.

### Testing

Done similar as for existing operations.

### DIFF
--- a/burn-tensor/src/tensor/api/int.rs
+++ b/burn-tensor/src/tensor/api/int.rs
@@ -56,13 +56,9 @@ where
     ///
     /// * `size` - The size of the square matrix.
     pub fn diagonal(size: usize) -> Tensor<B, 2, Int> {
-        let indices = (0..size)
-            .map(|e| Tensor::<B, 1, Int>::from_ints([e as i32]))
-            .map(|e| e.unsqueeze())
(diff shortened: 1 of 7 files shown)


burn-tensor/src/tensor/api/int.rs
```diff
@@ -45,6 +45,22 @@ where
...
+    /// Create diagonal matrix.
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - The size of the square matrix.
+    pub fn diagonal(size: usize) -> Tensor<B, 2, Int> {
+        let indices = Tensor::arange(0..size).unsqueeze();
+        let ones = Tensor::ones([size]).unsqueeze();
+        Tensor::<B, 2, Int>::zeros([size, size]).scatter(0, indices, ones)
+    }
```
nathanielsimard, line 61: Move this function to the `numeric.rs` file, it will be available for both `Int` and `Float` tensors, avoiding casting.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #802, opened 2023-09-13
Fix train-minimal feature and ensure it gets tested

dae:
### Checklist

### DIFF
--- a/burn-train/src/metric/dashboard/cli_stub.rs
+++ b/burn-train/src/metric/dashboard/cli_stub.rs
@@ -1,14 +1,11 @@
-use crate::{
-    metric::dashboard::{DashboardMetricState, DashboardRenderer, TrainingProgress},
-    TrainingInterrupter,
-};
+use crate::metric::dashboard::{DashboardMetricState, DashboardRenderer, TrainingProgress};
 
 /// A simple renderer for when the cli feature is not enabled.
 pub struct CLIDashboardRenderer;
 
 impl CLIDashboardRenderer {
     /// Create a new instance.
-    pub fn new(_interuptor: TrainingInterrupter) -> Self {
+    pub fn new() -> Self {
         Self {}
     }
 }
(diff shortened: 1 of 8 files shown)


burn-train/src/metric/dashboard/cli_stub.rs
```diff
@@ -1,11 +1,14 @@
...
 
 /// A simple renderer for when the cli feature is not enabled.
 pub struct CLIDashboardRenderer;
 
 impl CLIDashboardRenderer {
     /// Create a new instance.
-    pub fn new() -> Self {
+    pub fn new(_interuptor: TrainingInterrupter) -> Self {
         Self {}
     }
```
nathanielsimard, line 13: I think the problem is that the renderer type is selected with feature flag over the type, where I think it should be when creating. I think we should have a function like the following:

```rust
pub fn rendered_default(interuptor: TrainingInterrupter) -> Rendered {
}
```

examples/guide/Cargo.toml
```diff
@@ -6,8 +6,13 @@ name = "guide"
 publish = false
 version = "0.10.0"
 
+[features]
+default = ["train"]
+train = ["burn/train"]
+train-minimal = ["burn/train-minimal"]
+
 [dependencies]
-burn = {path = "../../burn", features = ["train", "wgpu"]}
```
nathanielsimard, line 10: I think the guide needs the `train` feature flag!

nathanielsimard (changes requested):
I think we could improve where the renderer is created to avoid mistakes like this!

examples/guide/Cargo.toml
dae: It's included by the default features above. If you'd prefer not to have the extra complexity in the guide, can you suggest a simple other model this example could depend on? I based the example off of your custom-training-loop one.

nathanielsimard (approved):
Thanks for the fix! I think it's cleaner now!

PR #804, opened 2023-09-14
Improve cache key to better discriminate among caches

Luni-4:
### Checklist

### Changes

This PR improves the cache key taking into account `Rust` version and test environment

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
       - name: caching
         uses: Swatinem/rust-cache@v2
         with:
-          key: ${{ runner.os }}-rust-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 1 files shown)


dae:
I think the Rust version is already factored in: https://github.com/marketplace/actions/rust-cache#cache-details

Separate caches for std/non-std should be an improvement, and the total cache size across the 4 seems to be about 6GB, which is under the 10GB limit.

Luni-4:
@dae 

Yep, I agree with your analysis. This size could become a problem when we will add `Windows` and `MacOS` actions, but we can evaluate at that time in my opinion. My goal was to solve this error

```
Failed to save: Unable to reserve cache with key v4-rust-Linux-rust--tests-23d4eae1-894e3aac, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/main, Key: v4-rust-Linux-rust--tests-23d4eae1-894e3aac, Version: 95cda81848c40b56b71b455f4462d3fda43719d53b2cb3305caf419a71f4d12c
```

nathanielsimard (approved):

Issue #787, opened 2023-09-08
Unfold tensor operation


nathanielsimard:
### Feature description

Support the unfold operation. See the PyTorch documentation for full description: https://pytorch.org/docs/stable/generated/torch.nn.Unfold.html#torch.nn.Unfold

agelas:
@nathanielsimard I can try doing this, and I'm assuming you'd want to follow PyTorch's constraint of only allowing 4-dimensional input tensors? If that's the case, how would you get around the constraint of working with dimensionality of D? If I were to do this, I'd start by defining the output tensor with something like `let mut output = Tensor::zeros([N, C * kernel_size[0] * kernel_size[1], H_out * W_out]);`, and then basically it seems like you have to extract "patches" of the current tensor to project into the output, something like this:

```Rust
let patch_tensor: Tensor<B, D, K> = self.select(n, indices);

output = output.select_assign(2, some_other_indices, patch_tensor); // will error because expects a Tensor<_, _, 3> but patch_tensor is Tensor<_, K, D>
```

Is there a way to ensure `self.select(n, indices) returns a Tensor<_, _, 3>` or to cast/convert it appropriately?

nathanielsimard:
@agelas I'm unsure if we can implement this operations efficiently using existing tensor operations. It is similar to a convolution or a pooling operation, which are all implemented by backends with custom kernels.

Maybe we can implement this efficiently using conv2d with a custom weight matrix of ones and zeros, but probably not a `select_assign` or anything index related.

PR #819, opened 2023-09-19
Feat/tensor unfold

agelas:
## Pull Request Template
@nathanielsimard This is still a work-in-progress, but I used your idea for co-opting conv2d to perform most of the grunt work for unfold in the ndarray backend. Torch seems to have an unfold operation already, but it seems to only be on a per-dimension basis. I still need to see whether they are equivalent. For Wgpu, I'm not sure whether it's worth writing a whole shader file for this or if we should try tricking conv2d into doing it for us again.

Let me now what you think!

Update 10/9: Unfold4d has an implementation for `NdArray`, `TchBackend`, `WgpuBackend`, and `ADBackend`. The `Candle` backend panics as it's not supported. 

### Checklist

### Related Issues/PRs

#787 

### Changes

Add an unfold4d operation to ModuleOps. Like PyTorch, this is only for unfolding 4D, image-like tensors into 3D batched tensors.

### Testing

See `burn-tensor/src/tests/module/unfold4d.rs` and `burn-wgpu/src/kernel/unfold/unfold4d.rs`. I tested more complicated shapes like unfolding a [4, 64, 174, 120] tensor with kernel_size=[3,3], dilation=[15,15], padding=[1,1], and stride=[1,1], which yields an unfolded tensor with dimensions [4, 576, 13432]  [as described here](https://stackoverflow.com/questions/73276139/why-is-nn-unfold-not-giving-me-the-same-results-as-unfold/73276874#73276874). I opted to not include this in the burn-tensor module tests because it would take over a minute to run.


burn-ndarray/src/ops/unfold.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let inner_array = unfolded.array;
+    let reshaped_array = inner_array
+        .into_shape([batch_size, channels_out, l])
+        .unwrap()
+        .into_dyn()
+        .into_shared();
+
+    NdArrayTensor::new(reshaped_array)
+}
```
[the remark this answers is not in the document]
agelas: @nathanielsimard So this actually doesn't work :(

I think there actually needs to be another step where the values, belonging to the tensor that had conv2d applied to it (and is still 4D), need to be plugged into the right index in the unfolded 3-dimensional tensor. I'll leave a comment in the updated ndarray backend implementation.

burn-ndarray/src/ops/unfold.rs
```diff
@@ -0,0 +1,91 @@
...
+
+                    l_index += 1;
+                }
+            }
+        }
+    }
+
+    // Convert the reshaped_array into the format that your NdArrayTensor needs
+    let reshaped_array = reshaped_array.into_dyn().into_shared();
+    NdArrayTensor::new(reshaped_array)
```
agelas, line 90: @nathanielsimard I don't know if this is the best way to do this yet, but this ugly cherry-picking of values from the tensor that had cov2d applied to it to the return tensor with 3 dimensions does get us the right shape at the end of the day. In order to do something like this in burn_core, we might need more flexibility than `slice` and `slice_assign` provide when working with the base tensor. What do you think?

agelas:
@nathanielsimard I think this is ready for another look-over. I still need to incorporate dilation and padding into the shader for the wgpu backend, but the tensor indexing gymnastics needed to get the right values in the right places should mostly be sorted out now. Still need to check with actual values though.

burn-ndarray/src/ops/unfold.rs
```diff
@@ -0,0 +1,91 @@
...
+                for w in 0..l_dim_2 {
+                    // Copy the value from unfolded to reshaped
+                    let value = inner_array[[b, c, h, w]];
+                    reshaped_array[[b, c, l_index]] = value;
+
+                    l_index += 1;
+                }
+            }
+        }
+    }
```
nathanielsimard, line 86: Does a reshape does the same thing? `unfolded.reshape([batch_size, channels, dim1 * dim2])` Maybe we need to swap the two axes before, but it's probably way more efficient to do it that way.

agelas: @nathanielsimard Unfortunately no. The short explanation is that the number of elements before and after a `reshape` needs to be the same, but an `unfold` doesn't do that. If we look at PyTorch's example, a `[2, 5, 3, 4]`(120 elements) tensor unfolded with a kernel size of `[2, 3]` and default options leads to an unfolded tensor with dimensions `[2, 30, 4]` (240 elements).

This [Stack Overflow thread](https://stackoverflow.com/questions/53972159/how-does-pytorchs-fold-and-unfold-work) can probably give a better intuition as to why. This nice graphic from the top answer really nails why (although note that this is for _tensor unfolds_, not _nn unfolds_, which is what we're doing here. A nn unfold extracts spatial patches not "patches" along a dimension):

image

That being said, I agree that this way of extracting spatial patches is not particularly efficient, so I'm very open to other ways of doing this.

burn-tch/src/ops/module.rs
```diff
@@ -107,6 +108,96 @@ impl<E: TchElement> ModuleOps<TchBackend<E>> for TchBackend<E> {
...
+                            .copy_(&value);
+
+                        l_index += 1;
+                    }
+                }
+            }
+        }
+
+        TchTensor::new(reshaped)
+    }
```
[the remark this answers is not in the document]
agelas: @nathanielsimard This is sort of an extension of my last comment, but since `reshape` isn't applicable, and the extraction of spatial tensors requires some heavy-duty tensor indexing that `slice` and `slice_assign` can't do (at least with all my attempts so far), I'm not sure we can compose tensor operations the way you want to create a default implementation.

agelas:
@nathanielsimard Also, I'm sure it's something I did in the shader, but do you know how the runner can have an invalid memory access after passing a test? In the previous CI run a similar thing happened, except that time it got through an additional test as well. Locally all tests pass and I'm running on an ubuntu image as well like the runner. 
Screenshot from 2023-10-04 22-59-39

nathanielsimard:
> @nathanielsimard Also, I'm sure it's something I did in the shader, but do you know how the runner can have an invalid memory access after passing a test? In the previous CI run a similar thing happened, except that time it got through an additional test as well. Locally all tests pass and I'm running on an ubuntu image as well like the runner. Screenshot from 2023-10-04 22-59-39

I would merge main I think, there was a bug related to that and some flaky tests with burn-wgpu.

burn-ndarray/src/ops/unfold.rs
nathanielsimard: I think the size of the convolution matches the size of the output from the unfold operation:

From Pytorch:
Conv2d:

image

Unfold:
image


I tested using a simple reshape at the end (it's just a flatten) and the tests were passing, so I think it should always work!

nathanielsimard: What I tested with:

```rust
pub(crate) fn unfold4d<E: FloatNdArrayElement>(
    x: NdArrayTensor<E, 4>,
    kernel_size: [usize; 2],
    options: UnfoldOptions,
) -> NdArrayTensor<E, 3> {
    let [_batch_size, in_channels, _in_height, _in_width] = x.shape().dims;
    let stride = options.stride.unwrap_or([1, 1]);
    let padding = options.padding.unwrap_or([0, 0]);
    let dilation = options.dilation.unwrap_or([1, 1]);

    let weight = create_unfolding_weight(in_channels, kernel_size);
    let unfolded = conv2d(
        x,
        weight,
        None,
        ConvOptions {
            stride,
            padding,
            dilation,
            groups: 1,
        },
    );

    let [batch_size, channels_out, out_height, out_width] = unfolded.shape().dims;

    NdArrayOps::reshape(
        unfolded,
        Shape::new([batch_size, channels_out, out_height * out_width]),
    )
}
```

agelas: Yes you're right, I'll switch to using a `reshape` instead!

agelas:
@nathanielsimard Ok, I think I'm inching closer to something that works- the shapes seem to be coming out alright but there's some tweaking that needs to be done on the webgpu backend, locally I can see the values aren't coming out the same as the reference. Also the CI is complaining about something new now, looks like it's also breaking main as well.

agelas:
> @agelas we have a new Rust release with some new clippy changes probably, we will fix that! Since we are using almost only one kernel, I think we could use the same strategy with WebGPU than with the other backends reusing the conv2d kernel.

Yeah, I haven't renamed it yet but unfold.wgsl is actually creating the weight matrix now and not doing any unfolding.

agelas:
@nathanielsimard All the backends (besides Candle) look like they're producing the right thing. I don't think Candle supports unfold natively, so I'm not sure if I should just put a `panic!` there like `avg_pool2d_backward` or all the `max_pool`/`adaptive_avg_pool` operations, or try the weight tensor + conv2d trick. 

For the wgpu backend, I was having an abysmal time getting reliable results on Vulkan- I kept running into some sort of (possible) thread synchronization issue which screws up the weight matrix buffer. So I opted to calculate the indices of the weight matrix that need to be set to 1 on the Rust side and have the shader just set those. This approach should be much more consistent insofar as the conv2d kernel is consistent across different wgpu API's. Assuming you're ok with this, I still need to figure out how to dynamically set the size of the indices buffer so it doesn't constrain users.

nathanielsimard:
@agelas You can put panic! for now, focusing on making this PR smaller. I can probably go after and try to provide a default implementation, so if a backend implements the reshape and conv2d operation, unfold will just work.

nathanielsimard (approved):

PR #822, opened 2023-09-21
Support no-std for models generated by burn-import

antimora:
### Checklist

### Changes

1. Support no-std for models generated by burn-import.
2. Fix bug to exclude inputs with weights (discovered while testing SqueezeNet burn model).
3. Removed feature flag over HalfPrecision to support it under `no_std`.

### Testing

1. Converted ONNX tests to build under `no_std`
2. Tested SqueezeNet


nathanielsimard (approved):
Well it was easy !

PR #826, opened 2023-09-24
Refactor/burn compute wgpu

nathanielsimard:
I removed the experimental autotune capability from wgpu, we are going to reimplementing a better version within the burn-compute project, which will ensure reusability and simplicity across backends and compute servers.

This is a big refactor with improved performance and much better memory management. This was a mandatory refactor to support buffer reusability, with chunks and slices, and will simplify a lot the process of correctly supporting WebGPU on wasm.


louisfd (approved):
Very nice
I made a few minor local refactors by myself, which I pushed. 
Good to go

PR #831, opened 2023-09-25
Improve cache key for std checks

Luni-4:
### Checklist

### Changes

This PR tries to fix cache miss for `1.71.0` target removing dots from `key` option and replaces `Cargo.lock`, not present in the `main` branch, with `Cargo.toml`


antimora (approved):
Looks good to me.

nathanielsimard (approved):

PR #832, opened 2023-09-27
Added kl_divergence.rs

ArvidHammarlund:
### Checklist

### Changes

Added kl-divergence loss for VAE

### Testing

Checked against a Numpy implementation.

### DIFF
--- a/burn-core/src/nn/loss/kl_divergence.rs
+++ b/burn-core/src/nn/loss/kl_divergence.rs
@@ -6,8 +6,7 @@ use crate::{config::Config, module::Module};
 use burn_tensor::{backend::Backend, Tensor};
 
 /// Implementation of the KL-divergence between a given normal distribution to the standard normal distribution.  
-///
-/// It return the ratio of the standard normal distribution that is covered by the other normal distribution.      
+///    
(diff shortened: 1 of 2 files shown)


nathanielsimard (commented):
Thanks for contributing, It would be valuable to provide a PyTorch script to compare with, I'm not sure I can reproduce the results myself.

ArvidHammarlund:
Pytorch only has the point-wise kl-divergence, which basically considers the difference in likelihood of two distributions seeing the same sample. This kl_divergence on the other hand considering the difference between the whole modelled probability distibutions (if both are normal).

The point wise kl_divergence has such as simple formula that I don't think it is necessary to make a module for it (as seen here: https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html). Additionally, the distribution kl_divergence is more applicable to VAE's which I believe is the main area of usage for KL_divegence? 

A reference a PyTorch code would be: https://github.com/AntixK/PyTorch-VAE/blob/master/models/vanilla_vae.py

nathanielsimard:
@ArvidHammarlund This is really informative! I think it would be very valuable to have this as documentation over the module with the math notations. I'm going to work to make the latex rendering works pretty soon!

PR #833, opened 2023-09-28
Feat/async read

nathanielsimard:
This is another quite big refactor to bring async to wasm. This is unfortunately required to use WebGPU on the web, since the APIs are all async, and it's impossible to block on async functions. To avoid using `async-trait` on all traits, I created a `Reader` struct that encapsulates the async logic.

I then updated the Tensor API to have async functions when compiling to wasm. So it doesn't change the Tensor API but gives the opportunity for backends to support async reads of tensor data. Of course, the changes impact the compatibility of burn-core on wasm, and some features will not work (todo!). Those are not related to inference, so it's a non-blocker for now, since training isn't fully supported on wasm anyways.

The overall goal is to avoid as much as possible to decorate functions with async. This would result in a worse API without any performance benefit. For example, you wouldn't be able to print tensor values with just a print statement; everything would probably need to implement Send + Sync, which will create hard-to-read error messages, etc. Async environments are good for concurrency but not really appropriate for crunching numbers.

I also updated the `mnist-inference-wasm` to support wgpu as well with a simple feature flag 🎉


antimora (approved):
LGTM 

One thing though. We should update our book's wasm section and tell about async nature and why we did this.

examples/mnist-inference-web/src/web.rs
AlexErrant, line 1: Sorry for being dense, but what is the purpose of `#[cfg(not(target_family = "wasm"))]` and `#[cfg(target_family = "wasm")]` in this file? Since it targets the web, isn't `target_family = "wasm"` required?

antimora: @AlexErrant regular build does not work because `output` is used down the line.

AlexErrant: Why does the regular build need to work? `build-for-web.sh` calls `wasm-pack`, and I'm not sure why e.g. `cargo build` should ever be called.

antimora: Because we build from the root and also rust-analyzer needs to work. For MNIST example, we have test as well but for this one, I omitted but we probably should it.

antimora: @nathanielsimard and I went back and worth on this. I figured it was actually bad that the WGPU API changes depending the target, which is the real issue here.

nathanielsimard: Agreed, @AlexErrant given that you suceeded using web workers for `burn-train`, do you think it would be possible to have a similar strategy to make reading data have a sync api even on `wasm` ? If we run the WGPU server on a web worker thread, could we call `await` inside of it and use channels to communicate with the `main` thread? 

If that works, we could have a consistent API on all targets and it would reduce the complexity a ton!!

AlexErrant: Gonna prefix this with _"I'm **very** unfamiliar with `std::thread` or any async programming that isn't Javascript/C#"_.

If we can run WGPU in a spawned `std::thread`, then I suppose that could be used to block the spread of `async/await`. However `std::sync::mpsc` is, AFAIK, one-way communication, while the `async/await` model is more request-eventual-response. Is there another way to communicate with a spawned thread? Otherwise, I wonder if we will end up building some kind of event loop/polling in the main thread. This would seem fine if the async APIs are only called during initialization... perhaps less so elsewhere.

burn-wgpu/src/compute/server.rs
```diff
@@ -209,29 +201,70 @@ where
...
+    fn read(&mut self, handle: &server::Handle<Self>) -> Reader<Vec<u8>> {
+        #[cfg(target_family = "wasm")]
+        {
+            let future = self.buffer_reader(handle).read(self.device.clone());
+            return Reader::Future(Box::pin(future));
+        }
+
+        #[cfg(not(target_family = "wasm"))]
+        Reader::Concrete(self.buffer_reader(handle).read(&self.device))
+    }
```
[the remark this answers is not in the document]
antimora: It might be worth taking this discussion outside of the closed PR because I think we have something important to resolve in regards to async/sync data read and people might want to see it. I am not sure if we have an issue or thread regarding this.

examples/mnist-inference-web/src/web.rs
nathanielsimard: Our infrastructure allows us to use any communication channel between the compute server and the client, we already have an implementation using mpsc using `thread` to run the server.

burn-wgpu/src/compute/server.rs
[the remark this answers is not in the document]
nathanielsimard: @AlexErrant reading data is by nature a blocking operation, so no there isn't any problem doing busy waiting on a thread where its only job is to run kernels. We could update the file `burn-compute/src/channel/mpsc.rs` to support running the server on a web worker thread, and change the signature to returns `Vec<u8>` instead of `Reader<Vev<u8>>` from the client side.

```rust
/// Create a new mpsc compute channel.
    pub fn new(mut server: Server, bound: usize) -> Self {
        let (sender, receiver) = mpsc::sync_channel(bound);

        let _handle = web_workers::spawn(move || async {
            while let Ok(message) = receiver.recv() {
                match message {
                    Message::Read(handle, callback) => {
                        let data = server.read(&handle);
                        core::mem::drop(handle);
                        let data = data.await; // or something similar
                        callback.send(data).unwrap();
                    }
                   _ => todo!("same as before"),
                };
            }
        });
```

PR #835, opened 2023-09-28
Speedup CI

nathanielsimard:
Some settings to speedup CI


louisfd (approved):
Sounds legit to remove incremental compilation in CI

PR #836, opened 2023-09-29
wgpu: Yet another (faster) matmul

louisfd:
New matmul kernel that leverages WGSL's vec4 data struct. 
It's considerably faster on my laptop. You can check its performance by running `cargo bench` within the burn-wgpu crate. 
It's the one called "Tiling2DMatmulVec4Primitive"


burn-wgpu/src/kernel/matmul/utils.rs
```diff
@@ -41,6 +41,8 @@ pub(crate) mod tests {
         let z = func(x_wgpu.into_primitive(), y_wgpu.into_primitive());
         let z = TestTensor::from_primitive(z);
 
+        println!("{:?}", z_reference.to_data());
+        println!("{:?}", z.to_data());
```
nathanielsimard, line 45: Remaining  println!

nathanielsimard (changes requested):

burn-wgpu/src/kernel/matmul/utils.rs
louisfd: Classic

nathanielsimard (approved):

PR #838, opened 2023-09-30
Bugfix: epoch and iteration were swapped.

daniel-vainsencher:
### Checklist

### Related Issues/PRs

None, just saw a tiny problem and fixed it.

### Changes

The example was displaying iteration as epoch and vice versa, I swapped them.

### Testing

Reran the example and saw iteration updating more often than epoch as it should be.


louisfd (approved):
Thanks!

PR #839, opened 2023-10-02
Use thread-local RNG to generate IDs

dae:
We've been exploring dividing our data set up into multiple batches, and training those batches in parallel. I noticed that performance did not scale with core count, and after some digging, found that this was mainly due to the Mutex being used to generate IDs. With the following change, training across 16 cores went from 21s to 4.2s.

thread_rng was previously discussed on #703, but I don't believe that applies here, as this is just used for UUID creation.

Using string UUIDs here is also not the most efficient - would there be any interest in a follow-up PR that uses the bytes directly instead of converting them to a string?

### Checklist

I'm getting intermittent flakes with burn_wgpu, that happened even before this change. Maybe my GPU/driver?

```
---- tests::module_backward::tests::test_embedding_backward stdout ----
thread 'tests::module_backward::tests::test_embedding_backward' panicked at 'assertion failed: `(left == right)`
  left: `Data { value: [NaN, NaN, NaN, NaN, NaN, NaN], shape: Shape { dims: [2, 3] } }`,
 right: `Data { value: [3.0, 9.0, 7.0, 21.0, 35.0, 27.0], shape: Shape { dims: [2, 3] } }`', burn-wgpu/src/lib.rs:48:5
```

### DIFF
--- a/burn-common/src/id.rs
+++ b/burn-common/src/id.rs
@@ -1,4 +1,4 @@
-use crate::rand::{get_seeded_rng, Rng, SEED};
+use crate::rand::gen_random;
 use alloc::string::{String, ToString};
 use uuid::{Builder, Bytes};
 
@@ -8,15 +8,7 @@ pub struct IdGenerator {}
 impl IdGenerator {
     /// Generates a new ID in the form of a UUID.
     pub fn generate() -> String {
-        let mut seed = SEED.lock().unwrap();
-        let mut rng = if let Some(rng_seeded) = seed.as_ref() {
-            rng_seeded.clone()
-        } else {
-            get_seeded_rng()
-        };
-
-        let random_bytes: Bytes = rng.gen();
-        *seed = Some(rng);
+        let random_bytes: Bytes = gen_random();
 
         let uuid = Builder::from_random_bytes(random_bytes).into_uuid();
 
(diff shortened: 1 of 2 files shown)


louisfd (commented):
Hi @dae 
Thank you for this contribution. If there is a lot of ID generation, then I can well see how the mutex may affect performance. 

I think I can summarize the discussion in #703 as:
-use StdRng if you need to sometimes set the seed deterministically (like for testing)
-use ThreadRng for better performance (no shared mutex), but it will be impossible to set the seed

There are no tests within Burn that rely on the generation of ID being deterministic, but I wonder if it would impact users in the future to not be able to set their seed. 

That said, your code is fine, I'm not explicitly approving simply because I'll wait for @nathanielsimard 's input.

antimora (commented):
Looks good from my end. I had worked on statically generated seed for no-std use cases. I think your changes still work for this use case. 

I also was [going to change to thread random](https://github.com/burn-rs/burn/pull/374/files#diff-90681e71fa7908e299340bf8eab27e6cfa613a674a9f4acac9d5627776d57436R13-R22) but I deferred it for later. Glad you took care of it.

antimora:
@dae 

> Using string UUIDs here is also not the most efficient - would there be any interest in a follow-up PR that uses the bytes directly instead of converting them to a string?

@nathanielsimard and I were discussing to shorten the IDs if possible. One thing to keep in mind, these IDs exported and visible in JSON file.

nathanielsimard:
@dae  I have the same NaN problem with embedding backward, I'm looking into it!

PR #843, opened 2023-10-02
Use a u64 counter for autodiff NodeIDs

dae:
I measure an approximate 13.6% performance boost in our crate from this change.

### Checklist

### Related Issues/PRs

#839

### Changes

### Testing

### DIFF
--- a/burn-autodiff/src/graph/node.rs
+++ b/burn-autodiff/src/graph/node.rs
@@ -33,9 +33,11 @@ impl NodeID {
     /// Create a unique [node id](NodeID).
     pub fn new() -> Self {
         static COUNTER: AtomicU64 = AtomicU64::new(0);
-        Self {
-            value: COUNTER.fetch_add(1, Ordering::Relaxed),
+        let value = COUNTER.fetch_add(1, Ordering::Relaxed);
+        if value == u64::MAX {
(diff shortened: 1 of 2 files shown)


burn-autodiff/src/graph/node.rs
```diff
@@ -27,14 +26,15 @@ impl Node {
...
 #[derive(Clone, Hash, PartialEq, Eq, Debug)]
 pub struct NodeID {
-    pub value: String,
+    pub value: u64,
 }
 
 impl NodeID {
     /// Create a unique [node id](NodeID).
     pub fn new() -> Self {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
```
nathanielsimard, line 35: Yeah, I thought about using u64 for autodiff IDs; it makes more sense for that use case. However, for extremely long training, it might become a problem (overflow). Do we know what behavior occurs in that case? I think it would be better if it were to trigger a panic. In that case, the training would crash and could be restarted from a checkpoint, instead of potentially corrupting the weights without understanding why. Of course, it would have to be an exceptionally long training period, and if it becomes a problem, we could consider using u128 (two u64) in that scenario.

nathanielsimard (approved):

PR #844, opened 2023-10-03
Improve metrics

nathanielsimard:
The goal of this PR is to pave the way towards exposing metric aggregates to training components that need it to make decisions, like checkpointing and early stopping.

Some details:

  * Removed the GPU Temperature metric since sysinfo doesn't expose GPU components.
  * Added refresh logic for sysinfo metrics.
  * Did a bit of renaming and exposed CPU memory in gigabytes and removed the swap information, since swapping should almost never be used while training models.
  * Added the ability to read numeric metrics from a file. Still unsure how it's going to be used in future PRs, but it's important to retrieve that information.


burn-train/src/metric/cpu_use.rs
```diff
@@ -1,19 +1,38 @@
...
+    refresh_frequency: Duration,
+    sys: System,
+    current: f64,
 }
 
 impl CpuUse {
     /// Creates a new CPU metric
     pub fn new() -> Self {
-        Self { use_percentage: 0. }
+        let mut sys = System::new();
```
louisfd, line 17: The new refresh rate is a game changer haha

burn-train/src/metric/loss.rs
```diff
@@ -28,6 +28,8 @@ impl<B: Backend> LossMetric<B> {
 }
 
 impl<B: Backend> Metric for LossMetric<B> {
+    const NAME: &'static str = "Loss";
```
louisfd, line 31: Use that const on line 39

PR #845, opened 2023-10-03
removed batch_size field from GRU and LSTM

EddieMataEwy:
### Checklist

- I've run all **std** checks and all tests pass except for 8 tests in record that require the static file path `/tmp/burn_test_file_recorder`, which is not valid on Windows.

### Related Issues/PRs

I haven't seen any related issues so I decided to do this pull request directly, since there are barely any changes.

### Changes

I have removed the `batch_size ` field from both GRU and LSTM, because they limit flexibility in exchange for nothing really.
This change increases compatibility with the Pytorch API (not necessary but nice).
If you run a RL workload and algorithm like PPO, during the rollout phase (data generation) you run with batch 1 but then during training phase you switch to whatever batch size you deem correct. Eliminating the field increases the flexibility for these kinds of situations and it doesn't impact saving and loading weights or performance (from what I have observed so far).

### Testing

I have ran the tests provided in the library and have not seen any performance differences for my workloads. May require further testing but the changes are very minor so I expect everything to work fine, once people adjust to the breaking changes.

### DIFF
--- a/burn-core/src/nn/rnn/gru.rs
+++ b/burn-core/src/nn/rnn/gru.rs
@@ -23,8 +23,6 @@ pub struct GruConfig {
     /// Gru initializer
     #[config(default = "Initializer::XavierNormal{gain:1.0}")]
     pub initializer: Initializer,
-    /// The batch size.
-    pub batch_size: usize,
 }
 
 /// The Gru module. This implementation is for a unidirectional, stateless, Gru.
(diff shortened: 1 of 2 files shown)


agelas:
This is more concise, but I'm not sure I understand how including the `batch_size` in the config limits flexibility- it doesn't impose any constraints that aren't already present. Both LSTM's and GRU's are stateless, so a single config was never really intended to be re-used since you'd need to extract hidden/cell states regardless and pass them to another layer, and I find it helpful to be explicit with all the parameters tying a model together in those cases. If this is a matter of preference, maybe make it an option?

EddieMataEwy:
Also, instead of assigning `batch_size` and `seq_length` separately maybe it's better to do array unpacking as follows:
`let [batch_size, seq_length, _] = batched_input.shape().dims;`
Just a minor detail.

PR #847, opened 2023-10-05
Refactor burn-train

nathanielsimard:
Heavy refactor of `burn-train`.

## Changes

  * Added `LearnerComponents` to specify the abstraction boundaries necessary for the learner. It avoids having to specify most of the trait constraints in many places and simplifies the code.
  * Created `LearnerCheckpointer` to centralize checkpoint logic.
  * Renamed `LRScheduler` -> `LrScheduler` to follow the Rust standard.
  * Removed the `dashboard` within `burn-train`, which was a bit confusing. Replaced it with more appropriate names (`MetricsCallback`, `MetricsRenderer`, `MetricState`, `MetricUpdater`, `NumericMetricUpdater`, `TuiMetricsRenderer`, `CliMetricsRenderer`).
  * Removed `dyn` in Async Checkpointer.
  * Removed `dyn` and `mutex` in Async Callback.
  
## Breaking Changes

There are breaking changes, but most examples are untouched since type inference works the same way if used within the same function.


burn-train/src/learner/base.rs
```diff
@@ -1,91 +1,67 @@
...
-    LR: LRScheduler,
-{
-    pub(super) fn checkpoint(
-        model: &Model,
-        optim: &Optim,
-        scheduler: &LR,
-        checkpointer_model: &CheckpointModel<Model, B>,
-        checkpointer_optimizer: &CheckpointOptim<Optim, Model, B>,
-        checkpointer_scheduler: &CheckpointScheduler<LR>,
+impl<C: LearnerComponents> LearnerCheckpointer<C> {
```
louisfd, line 33: I'm starting to think there is too many times the letter C: callback, checkpoint, components

burn-train/src/learner/builder.rs
```diff
@@ -98,16 +98,16 @@ where
     /// * `renderer` - The custom renderer.
     pub fn renderer<DR>(mut self, renderer: DR) -> Self
     where
-        DR: DashboardRenderer + 'static,
+        DR: MetricsRenderer + 'static,
```
louisfd, line 104: if there is no more dashboard, should not use D letter

louisfd (changes requested):
Very nice. My only comments are about generics names

burn-train/src/learner/base.rs
nathanielsimard: We could use LC ?

Issue #841, opened 2023-10-02
Early stopping


nathanielsimard:
Add the option to keep the best weighs ranked by validation loss.
Stop the training when the X following epochs don't improve the metric.

## Optional

Being able to set the metric to be used for ranking.

PR #857, opened 2023-10-10
Feat training events

nathanielsimard:
Another big refactor of `burn-train` to start collecting training events and expose useful functions on top of it such as `find_epoch`. 

Useful for #841


burn-train/src/info/aggregates.rs
```diff
@@ -0,0 +1,163 @@
...
+            self.logger.epoch(self.epoch);
+        }
+    }
+
+    #[test]
+    fn should_find_epoch() {
+        let mut logger = TestLogger::new();
+        let mut aggregate = NumericMetricsAggregate::default();
+
+        logger.log(500.);
```
louisfd, line 144: I would add comments like
`// Epoch 1` and so on, so it is clear why we select epoch 2, cause I thought for a while it started at 0

examples/mnist/src/training.rs
```diff
@@ -58,16 +58,16 @@ pub fn run<B: ADBackend>(device: B::Device) {
...
-        .metric_valid_plot(AccuracyMetric::new())
-        .metric_train_plot(CpuUse::new())
-        .metric_valid_plot(CpuUse::new())
-        .metric_train_plot(CpuMemory::new())
-        .metric_valid_plot(CpuMemory::new())
-        .metric_train_plot(CpuTemperature::new())
-        .metric_valid_plot(CpuTemperature::new())
-        .metric_train_plot(LossMetric::new())
-        .metric_valid_plot(LossMetric::new())
+        .metric_train_numeric(AccuracyMetric::new())
```
louisfd, line 61: Breaking change here, we will have to update the book

louisfd (approved):
Approved. Just minor comments

PR #858, opened 2023-10-10
Update book

nathanielsimard:


louisfd (approved):
Nice
There will be changes also for the learner

PR #861, opened 2023-10-13
Feat/compute/autotune

louisfd:
I made a new autotune architecture in autotune. Works with dummy tests. 

Also, moved benchmarks back to burn-tensor from burn common because it is not used anywhere else and burn-common needs to work in no-std


burn-compute/src/tune/operation.rs
```diff
@@ -0,0 +1,13 @@
+use core::hash::Hash;
+
+/// Format of the inputs and outputs of a kernel
+pub trait HashableResources: PartialEq + Eq + Hash {
+    /// Description used as an autotune key
+    fn key(&self) -> String;
+}
+
+/// Type of operation for the kernel
+pub trait Operation: PartialEq + Eq + Hash {
```
nathanielsimard, line 10: Does this operation also needs to implement `PartialEq` `Eq` and `Hash`?

burn-compute/src/tune/base.rs
```diff
@@ -0,0 +1,71 @@
...
+use super::TuneBenchmark;
+
+/// Use to find and reuse the best kernel for some input
+pub struct Tuner<TB, O, S>
+where
+    TB: TuneBenchmark<O, S>,
+    O: Operation,
+    S: ComputeServer,
+{
+    benchmark_pool: Mutex<BenchmarkPool<TB, O, S>>,
```
nathanielsimard, line 19: I think the pool should be included in the compute server, in a mutable environment, so we should probably remove the mutex and make the functions mutable!

burn-compute/src/tune/benchmark_pool.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Returns the median duration among all samples
+    pub fn median_duration(&self) -> Duration {
+        let mut sorted = self.durations.clone();
+        sorted.sort();
+        *sorted.get(sorted.len() / 2).unwrap()
+    }
+}
+
+/// A benchmark that runs on server handles
+pub trait TuneBenchmark<O: Operation, S: ComputeServer> {
```
nathanielsimard, line 25: I think this trait should extend the Benchmark trait.

```rust
pub trait TuneBenchmark: Benchmark<Args = &[&Handle<S>]>{
    fn make_kernel(&self) -> S::Kernel;   
}
```

burn-compute/tests/dummy/tune/benchmark.rs
```diff
@@ -0,0 +1,62 @@
...
+        DummyElementwiseMultiplicationType,
+    },
+    DummyClient, DummyKernel, DummyServer,
+};
+
+use super::{AdditionOp, CacheTestOp, MultiplicationOp};
+
+#[derive(new)]
+pub struct DummyBenchmark<'a, O> {
+    client: &'a DummyClient,
```
nathanielsimard, line 22: The client is made to be cloned, so I would remove the lifetime here as it adds complexity without much benefit.

nathanielsimard (changes requested):
The main problem I see is that it's going to be hard to specify the `key` for different kernels. E.g we may want to only benchmark matmul with batch size of one, but with different M,N,K, while the oposite for binary operations. Also being forced to use the same handle as the current computation may slow down drastically the benchmarking phase. I would probably have a different trait:

```rust
pub trait TunableKernel {
    type Kernel;
    type BenchmarkArgs;
    
    fn key(&self) -> String;
    fn tunables(&self) -> (Vec<Self::Kernel>, Self::BenchmarkArgs);
    fn kernel(&self) -> Self::Kernel;
}
```

Someting like this, but will probably look very different. The goal is to give the freedom to each kernel to set what are the parameters used for the benchmarks as well as the caching strategy with the key function.

louisfd:
@nathanielsimard 
New version ready

burn-compute/src/tune/operation.rs
```diff
@@ -0,0 +1,67 @@
...
+            self.input_key()
+        )
+    }
+
+    /// Partial key referring to the operation type,
+    /// generally hardcoded for an autotune operation
+    fn operation_key(&self) -> String;
+
+    /// Partial key referring to the input shape
+    fn input_key(&self) -> String;
```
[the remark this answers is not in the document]
nathanielsimard: I think the key should be a struct!

burn-compute/src/tune/operation.rs
```diff
@@ -0,0 +1,67 @@
...
+impl<S: ComputeServer> Operation<S> {
+    /// Executes the operation on given handles and server, with the additional parameters
+    pub fn execute(&self, inputs: &[&Handle<S>], server: &mut S) {
+        let mut handles = inputs.to_vec();
+
+        let p = match self.parameters.clone() {
+            Some(parameter_handles) => parameter_handles.into_iter().collect::<Vec<Handle<S>>>(),
+            None => Vec::new(),
+        };
+        handles.extend(p.iter().collect::<Vec<&Handle<S>>>());
```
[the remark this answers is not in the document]
louisfd: I wish but it always ends up in errors such as:
`param does not live long enough
borrowed value does not live long enough`

burn-compute/tests/dummy/tune/benchmark.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    fn execute(&self, kernel: Box<dyn DummyKernel>, handles: &[&Handle<DummyServer>]) {
+        self.client.execute(kernel, handles);
+    }
+
+    fn sync(&self) {
+        self.client.sync()
+    }
+}
+
```
nathanielsimard, line 28: Is it still used?

louisfd: No, must have been brought back from the dead somehow when merging
Also, @nathanielsimard can you help out with the CI, I don't understand the where the no-std error comes from

PR #867, opened 2023-10-15
Fix formatting issues in the Burn Book's chapter 6: "Custom Training Loop"

miestrode:
In some of the code in this chapter, I believe there is a mistake in formatting the output for the training and validation loops. The epoch counter and iteration counter were accidentally supplied to the wrong locations in the format string. I've fixed that and also made formatting a tad bit more idiomatic, by using the names of variables directly in the format string where possible.

*As I am still not an expert in this framework and AI terminology, it could be that this edit was mistaken, but I find that unlikely to be the case. If it is, I am sorry.*

### DIFF
--- a/burn-book/src/custom-training-loop.md
+++ b/burn-book/src/custom-training-loop.md
@@ -76,11 +76,8 @@ pub fn run<B: ADBackend>(device: B::Device) {
             let accuracy = accuracy(output, batch.targets);
 
             println!(
-                "[Train - Epoch {} - Iteration {}] Loss {:.3} | Accuracy {:.3} %",
-                iteration,
-                epoch,
+                "[Train - Epoch {epoch} - Iteration {iteration}] Loss {:.3} | Accuracy {accuracy:.3} %",
                 loss.clone().into_scalar(),
-                accuracy,
             );
 
             // Gradients for the current backward pass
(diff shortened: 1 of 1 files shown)


burn-book/src/custom-training-loop.md
```diff
@@ -101,11 +98,8 @@ pub fn run<B: ADBackend>(device: B::Device) {
             let accuracy = accuracy(output, batch.targets);
 
             println!(
-                "[Valid - Epoch {} - Iteration {}] Loss {} | Accuracy {}",
-                iteration,
-                epoch,
+                "[Valid - Epoch {epoch} - Iteration {iteration}] Loss {} | Accuracy {accuracy}",
```
nathanielsimard, line 101: I think we should have the same formatting for both training and validation prints. There is no precision settings in the validation output `{:.3}`.

nathanielsimard (changes requested):

burn-book/src/custom-training-loop.md
miestrode: I mainly made my changes in an attempt to still keep the original styling of the code section, but IMO it could be improved. I assume you want to use the same style from the training prints, right? (I.E, precision of 3 digits and the % sign)

Personally I am not really a fan of just printing like this in the code, and I think the training TUI would make a lot more sense here, but I am also not sure how difficult it would be to actually add this to the code.

Another thing is that perhaps you should consider moving this chapter to another part of the book, as it's really less about a custom training loop and more about how a training loop in Burn works in general (which is still something that is very useful to know, it should perhaps just be presented in a different way).

nathanielsimard: Yes the goal is to have the same print statement for training and validation. Do you have some propositions in presenting this book section a bit differently?

Issue #872, opened 2023-10-17
Lstm panics when batch size is greater than 1


mosheduminer:
**Describe the bug**
The LSTM module provided by burn seems to always fail (panic) when batch size is not equal to 1.

**To Reproduce**
To reproduce, you can run this code:
```rust
use burn::{
    backend::TchBackend,
    nn::LstmConfig,
    tensor::{Data, Tensor},
};

type B = TchBackend<f32>;

fn main() {
    // This part works
    let lstm = LstmConfig::new(1, 1, false, 1).init::<B>();

    let a = Tensor::<B, 3>::from_data(Data::from([[[0.0]]]));
    println!("shape: {:?}", a.shape());
    lstm.forward(a, None);
    println!("success");

    // This part panics
    let lstm = LstmConfig::new(1, 1, false, 2).init::<B>();

    let a = Tensor::<B, 3>::from_data(Data::from([[[0.0]], [[0.0]]]));
    println!("shape: {:?}", a.shape());
    lstm.forward(a, None); // Panic
}
```

**Expected behavior**
I expect that the lstm should handle various batch sizes correctly, without panicking.

**Desktop (please complete the following information):**
 - OS: Windows 11

**Burn version**
- 0.9.0 (using `burn = { version = "0.9.0", features = ["train", "tch"] }`)

**Additional context**
I see that recently the batch size parameter was removed from the `LstmConfig` (though this change wasn't released yet), and is instead inferred dynamically at runtime. I suppose it's possible that this bug wasfixed as part of that change, but I have not checked to see whether that is the case.

agelas:
@mosheduminer Based on the description this bug is almost definitely still present in development/main. 

@nathanielsimard Based on the discord snippets if I had to guess it's probably got to do with the squeezing/unsqueezing, so either this:

```rust
for (t, input_t) in batched_input.iter_dim(1).enumerate() {
            let input_t = input_t.squeeze(1);
```
or this
```rust
// store the state for this timestep
batched_cell_state = batched_cell_state.slice_assign(
  [0..batch_size, t..(t + 1), 0..self.d_hidden],
  cell_state.clone().unsqueeze(),
);
batched_hidden_state = batched_hidden_state.slice_assign(
  [0..batch_size, t..(t + 1), 0..self.d_hidden],
  hidden_state.clone().unsqueeze(),
);
```

agelas:
@nathanielsimard I think this issue can be closed!

mosheduminer:
[...]

PR #873, opened 2023-10-17
Bug/lstm unsqueeze

agelas:
### Checklist

### Related Issues/PRs

#872 

### Changes

It looks like the `unsqueeze()` function prepends the singleton dimension to the tensor, when it needs to be inserted so that `[batch_size, hidden_size]` can be unsqueezed to `[batch_size, 1, hidden_size]`.

### Testing
It'll take awhile to run, but doing something similar to the use case mentioned on discord will successfully complete the forward pass now. 
``` rust 
#[test]
    fn test_repro() {
        let lstm = LstmConfig::new(512, 1024, false).init::<TestBackend>();
        let a = Tensor::<TestBackend, 3>::random([128, 256, 512], Distribution::Default);
        lstm.forward(a, None);
        println!("success");
    }
```

### DIFF
--- a/burn-core/src/nn/rnn/gru.rs
+++ b/burn-core/src/nn/rnn/gru.rs
@@ -140,9 +140,12 @@ impl<B: Backend> Gru<B> {
                 .mul(update_values.clone().sub_scalar(1).mul_scalar(-1)) // (1 - z(t)) = -(z(t) - 1)
                 + update_values.clone().mul(hidden_t);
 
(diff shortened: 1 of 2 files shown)


nathanielsimard (commented):
I think it would be valuable to add tests to validate the fix. Something that fails with the previous code, but passes with the new one!

nathanielsimard (approved):

PR #875, opened 2023-10-17
Also consider devices of type Other when trying to find the best device.

christophebiocca:
### Checklist

`./run-checks.sh all`, if that makes any difference. It gets errors due to bad tolerances, but they're identical to those on `main`.

### Related Issues/PRs

Meant to fix issue reported in discord: https://discord.com/channels/1038839012602941528/1059209073784008835/1163541031086723162

### Changes

WgpuDevice::BestAvailable fails if there is a device of type Other available and nothing else, as it doesn't consider them.

### Testing

Not tested beyond `run-checks`. Would probably need some kind of mocking? Would like guidance on how to test this change.


christophebiocca:
Fixed formatting.

nathanielsimard (approved):

PR #878, opened 2023-10-18
Feat/early stopping + burn train refactor

### LINKED ISSUE #841: Early stopping

nathanielsimard:
Another big refactor, removing the messy event collector and introduce an event store and an event processor.

Fix #841


burn-train/src/learner/early_stopping.rs
```diff
@@ -0,0 +1,185 @@
+use crate::metric::{
+    store::{Aggregate, Direction, EventStoreClient, Split},
+    Metric,
+};
+
+/// The condition that [early stopping strategies](EarlyStoppingStrategy) should follow.
+pub enum StoppingCondition {
+    /// When no improvement has happened since the given number of epochs.
+    /// In other words, when no best epoch has been found.
```
louisfd, line 9: I feel like this second sentence is just confusing. The condition is already clear from the first sentence, and there's always a "best" epoch even if it stagnates

burn-train/src/learner/early_stopping.rs
```diff
@@ -0,0 +1,185 @@
...
+    NoImprovementSince {
+        /// The number of epochs allowed to get worsen before it get better.
```
louisfd, line 11: The number of epochs allowed to worsen before it gets better.

burn-train/src/learner/early_stopping.rs
```diff
@@ -0,0 +1,185 @@
...
+            match store.find_metric(&self.metric_name, epoch, self.aggregate, self.split) {
+                Some(value) => value,
+                None => {
+                    log::warn!("Can't find metric for early stopping.");
+                    return false;
+                }
+            };
+
+        let is_best = match self.direction {
+            Direction::Lowest => current_value <= self.best_value,
```
louisfd, line 46: Shouldn't those be _strictly_ less than / greater than?
Because now even if there is no improvement and stays the same, it will be considered as improvement

burn-train/src/learner/early_stopping.rs
```diff
@@ -0,0 +1,185 @@
...
+        // Two points for the last epoch. Mean 1.5
+        epoch += 1;
+        process_train(&mut processor, 1.0, epoch);
+        process_train(&mut processor, 2.0, epoch);
+        end_epoch(&mut processor, epoch);
+
+        assert!(
+            early_stopping.should_stop(epoch, &store),
+            "Should stop since two following epochs are worse then the best."
+        );
```
louisfd, line 183: Add a test for the equality case

louisfd (changes requested):
This refactor is really becoming legendary! I feel we gain in both cleanness and flexibility. 
I have some comments concerning the StoppingCriterion part, the rest LGTM

PR #880, opened 2023-10-19
doc(burn-tensor): Add examples to slice operation to help noobs like me understand what it does

jggc:
### Checklist

- 535 tests failed but I'm pretty sure it's got nothing to do with my doc only changes

### Changes

Add examples of how `Tensor::slice` works

### Testing

burn-tensor tests are passing


burn-tensor/src/tensor/api/base.rs
```diff
@@ -272,8 +272,32 @@ where
     /// fn example<B: Backend>() {
     ///     let tensor = Tensor::<B, 3>::ones(Shape::new([2, 3, 3]));
     ///     let tensor_slices = tensor.slice([0..1, 0..3, 1..2]);
-    ///     println!("{:?}", tensor_slices.dims()); // [1, 3, 2]
-    ///     
+    ///     println!("expecting [1, 3, 1] : {:?}", tensor_slices.dims());
```
louisfd, line 275: Thanks, I realize the previous example was false, which must have been confusing!

louisfd (changes requested):
Funny PR name!
I just think that's a bit too many examples for the doc. Maybe select the three most insightful. You could convert the others to unit tests in `burn-tensor/src/tests/ops/slice.rs`. Tests are somewhat like documentation examples, with the advantage that you can't make a mistake like there was with [1, 3, 2]

jggc:
What about I make them doc tests so that they execute on cargo test? There doesn't seem to be any in the project so far but I think it't the proper solution to the current issue. If it really is too many some lines can also be hidden from documentation and still run as doc tests.

Also I know it is "a lot" but it is the amount that I personally needed to be confident that I understood properly what it does... For experienced people it shouldn't hurt much as it is just a 25 lines comment block that is easy to skip and IMHO noobs will be thankful to have comprehensive exemples right in their LSP.

Your project though so I won't insist any more if you insist :grin: 

And yeah the wrong example obviously made me extremely doubtful of my own understanding until I had validated the behavior with other frameworks and some youtube videos.

burn-tensor/src/tensor/api/base.rs
```diff
@@ -272,8 +272,32 @@ where
...
+    ///
+    ///     let tensor = Tensor::<TestBackend, 1, Int>::arange(0..12);
```
nathanielsimard, line 277: Should use `B` instead of `TestBackend`.

burn-tensor/src/tensor/api/base.rs
```diff
@@ -272,8 +272,32 @@ where
...
+    ///     println!("\nexpecting [1, 3, 1] : {:?}", tensor_slices.dims());
+    ///     println!("expecting [[[1],[5],[9]]] : {:?}", tensor_slices);
+    ///
+    ///     let tensor_slices = tensor.clone().slice([0..1, 1..3]);
+    ///     println!("\nexpecting dim [1, 2, 4] : {:?}", tensor_slices.dims());
+    ///     println!("expecting [[[4,5,6,7],[8,9,10,11]]] : {:?}", tensor_slices);
+    ///
+    ///     let tensor_slices = tensor.slice([0..1, 1..3, 1..2]);
+    ///     println!("\nexpecting [1, 2, 1] : {:?}", tensor_slices.dims());
+    ///     println!("expecting [[[5],[9]]] : {:?}", tensor_slices);
```
nathanielsimard, line 300: Instead of having all reshapes then all slices I would have them grouped with their corresponding tensor. Also a comment on top of each block could be really helpful:

```rust
/// Slice of rank 3 tensor
let tensor = tensor.reshape([1, 3, 4]);
println!("\nexpecting [[[0,1,2,3],[4,5,6,7],[8,9,10,11]]] : {:?}", tensor);
println!("expecting [1, 3, 1] : {:?}", tensor.dims());

let tensor_slices = tensor.clone().slice([0..1, 0..3, 1..2]);
println!("\nexpecting [1, 3, 1] : {:?}", tensor_slices.dims());
println!("expecting [[[1],[5],[9]]] : {:?}", tensor_slices);
```

nathanielsimard (changes requested):

jggc:
I did improve according to your comments :
- I reduced the number of examples to 3
- I added comments to explain the goal of each example

Let me know what still needs improvement

louisfd (approved):
Thanks, I think this will be very clear now, especially the third one

nathanielsimard (approved):

antimora:
I rebased it. Lets see if the CI is passing now..

PR #885, opened 2023-10-23
[Burn-train] Improve panic messages

nathanielsimard:
Replace a lot of `unwrap` with `expect` or `unwrap_or_else`.

### DIFF
--- a/burn-train/src/checkpoint/strategy/composed.rs
+++ b/burn-train/src/checkpoint/strategy/composed.rs
@@ -62,13 +62,19 @@ impl CheckpointingStrategy for ComposedCheckpointingStrategy {
             // We assume that the strategy would not want the current epoch to be saved.
             // So we flag it as deleted.
             if actions.is_empty() {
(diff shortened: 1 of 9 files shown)


burn-train/src/checkpoint/strategy/composed.rs
```diff
@@ -83,7 +89,12 @@ impl CheckpointingStrategy for ComposedCheckpointingStrategy {
         for epoch in epochs_to_check.into_iter() {
             let mut num_true = 0;
             for i in 0..self.strategies.len() {
-                if self.deleted.get(i).unwrap().contains(&epoch) {
+                if self
+                    .deleted
+                    .get(i)
+                    .expect("Ad many 'deleted' as 'strategies'.")
```
louisfd, line 95: As

louisfd (approved):
That's good. Weird to see positive error message but it appears to be the idiomatic way: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

PR #887, opened 2023-10-23
Enable candle cuda

nathanielsimard:


louisfd (approved):
Thanks, I hadn't done it because I'm on mac os.

PR #893, opened 2023-10-23
Add a `sync` feature to common, core, and tensor

AlexErrant:
### Checklist

- [?] Confirm that `run-checks` script has been executed.

Failed with the same error [as last time](https://github.com/burn-rs/burn/pull/830), more or less.

### Related Issues/PRs

Resolves https://github.com/burn-rs/burn/issues/865.

Continuing the discussion from the above issue: After a truly embarrassing amount of spinning in circles, I don't think I can get Asyncify to work with wgpu. WGPU is, at its foundation, async, [and critical apis don't have a synchronous version](https://github.com/gfx-rs/wgpu/issues/3509). Asyncify only works on wasm files. WGPU is async _Rust_, so asyncify can't "de-async" it for our consumption.

### Changes

So instead I pull the same trick as last time and feature flag stuff. I add a `sync` feature that I can use to get sync behavior when targeting WASM.

### Testing

Add

```
burn = {path = "../../burn", default-features = false, features = ["train-minimal", "ndarray-no-std", "sync"]}
getrandom = { version = "0.2", features = ["js"] }
```

to https://github.com/burn-rs/burn/blob/main/examples/mnist-inference-web/Cargo.toml, delete [this await](https://github.com/burn-rs/burn/blob/38e88a79bda0f5654b100e712539e7a8081ac3f5/examples/mnist-inference-web/src/web.rs#L68), and observe that ` ./build-for-web.sh ndarray` builds.

### DIFF
--- a/burn-common/Cargo.toml
+++ b/burn-common/Cargo.toml
@@ -15,6 +15,7 @@ default = ["std"]
 
 std = ["rand/std"]
 
+sync = []
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 async-trait = { workspace = true }
--- a/burn-common/Cargo.toml
+++ b/burn-common/Cargo.toml
@@ -15,7 +15,7 @@ default = ["std"]
 
 std = ["rand/std"]
 
-sync = []
+wasm-sync = []
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 async-trait = { workspace = true }
--- a/burn-common/src/reader.rs
+++ b/burn-common/src/reader.rs
@@ -1,7 +1,7 @@
 use alloc::boxed::Box;
 use core::marker::PhantomData;
 
-#[cfg(target_family = "wasm")]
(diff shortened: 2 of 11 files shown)


AlexErrant:
Maybe [this'll](https://github.com/tkaitchuck/constrandom/pull/30) fix CI?

nathanielsimard (changes requested):
I'm considering renaming the feature flags to something like `wasm-sync` since it exclusively modifies the behavior when targeting wasm. Another option is to introduce a feature flag called `async`, but it could be misleading, as most of the execution relies on asynchronous operations using channels.

Just the name `sync` is misleading, since even when not enabled, the APIs are sync when not targeting wasm.

nathanielsimard (approved):

AlexErrant:
Oh hah I did not expect that to fix CI.

nathanielsimard:
@AlexErrant It was a setting problem, so just a trigger would fix it!

PR #896, opened 2023-10-24
Add codecov token

nathanielsimard:


.github/workflows/test.yml
```diff
@@ -112,14 +112,15 @@ jobs:
       - name: Get coverage data for codecov
         if: matrix.rust == 'stable' && matrix.test == 'std'
         run: |
-          grcov . --binary-path ./target/debug/ -s . -t lcov --branch \
+          grcov . --binary-path ./target/debug/ -s . -t ${CODECOV_TOKEN} --branch \
```
Luni-4, line 115: I think we can leave `lcov` here, it is not necessary to set up a token

Luni-4 (changes requested):

PR #902, opened 2023-10-25
ci/Add coverage as xtask task

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR allows to run `coverage` locally, but first it is necessary to install `grcov`.

### Testing


xtask/src/runchecks.rs
```diff
@@ -209,28 +252,41 @@ fn std_checks() {
...
+    // Build each workspace
+    cargo_build(["--workspace", "--exclude=xtask"].into());
+
     // Produce documentation for each workspace
     cargo_doc(["--workspace"].into());
 
+    // Setup code coverage
+    if is_coverage {
+        setup_coverage();
+    }
```
nathanielsimard, line 275: The coverage got -11%, and looking at the report it seems like the reason is the lines with an `std` compilation flag are not tested (obviously), did  we setup the coverage before the `no-std` tests as well before this change?

Luni-4: The problem is related to the fact that I cannot exclude `xtask` and `examples` directories anymore. So we have more files which are not tested. I cannot understand why though, perhaps `xtask` is not run on the main repository?

nathanielsimard (approved):

PR #904, opened 2023-10-26
[Breaking] Refactor Backend Names

nathanielsimard:
This refactor aims to improve the clarity of Burn when selecting a backend. Also, it aims to make it easier to create new backends without having to deal with unreadable associative types through the use of type aliases (`FloatTensor`, `IntTensor`, `BoolTensor`, `FloatElem`, `IntElem`, and `Device`).

There are breaking changes, but only for the naming. The type aliases don't break anything but simplify the process of creating new backends!

Also, some other changes are made to be more in line with the Rust way of naming things, namely having `UpperCamelCase` for acronyms and contractions, as well as for types.
  - `ADBackend` => `AutodiffBackend`
  - `ADModule` => `AutodiffModule`
  - `ADTensor` => `AutodiffTensor`
  - `TestADBackend` => `TestAutodiffBackend`
  - `TestADTensor` => `TestAutodiffTensor`
  - `ADBackendDecorator` => `AutodiffDecorator`
  
One of the main changes for backend names is to remove the `Backend` suffix, since the types are now available through `burn::backend`. Also, I think it's way cleaner to have `Autodiff<LibTorch>` than to have `ADBackendDecorator<TchBackend>`. The `LibTorch` change is also motivated by better clarity. Yes, we use `tch` for the bindings, but the most interesting part of the backend isn't the binding, but the code that is used!

I removed the re-exported backend type alias such as `WgpuAutodiffBackend` since now the name of the autodiff backend isn't as obscure as before. The biggest benefit is that users will have access to the actual documentation of the backend they use instead of the small doc on top of each alias.

The autodiff backend name `ADBackendDecorator` was bothering me for some time. Including the pattern in the name isn't something really bad in some cases, but we don't add a `macro` suffix to all macros or `derive` to all `Derive`, so I think it's better to emphasize the actual behavior that is added instead of the pattern used. I think it's more appropriate to describe the pattern in the documentation.


antimora:
The description lacks mention of the fundamental rules or principles being followed. Please summarize all changes in terms of rules or principles, so others can easily follow and review your changes. This request stems from the fact that many people are accustomed to the previous convention, so adapting to the new changes will require a shift in both mindset and the code/APIs.

nathanielsimard:
> The description lacks mention of the fundamental rules or principles being followed. Please summarize all changes in terms of rules or principles, so others can easily follow and review your changes. This request stems from the fact that many people are accustomed to the previous convention, so adapting to the new changes will require a shift in both mindset and the code/APIs.

Agreed! I'm still working on it after an earlier self-review. More descriptions and changes are expected to better follow the Rust naming conventions!

burn-core/src/lib.rs
```diff
@@ -36,30 +36,26 @@ pub mod record;
...
+pub type TestBackend = burn_wgpu::Wgpu<burn_wgpu::Vulkan, f32, i32>;
 
 #[cfg(all(test, feature = "test-wgpu", target_os = "macos"))]
-pub type TestBackend = burn_wgpu::WgpuBackend<burn_wgpu::Metal, f32, i32>;
+pub type TestBackend = burn_wgpu::Wgpu<burn_wgpu::Metal, f32, i32>;
 
 #[cfg(feature = "std")]
 #[cfg(test)]
-pub type TestADBackend = burn_autodiff::ADBackendDecorator<TestBackend>;
+pub type TestADBackend = burn_autodiff::Autodiff<TestBackend>;
```
louisfd, line 58: I think we aim at changing every mention of 'AD'?

louisfd (approved):
It's especially great when long lines like `ADBackendDecorator<WgpuBackend<...` become `Autodiff<Wgpu<...`
I also like the fact that the backend word is now reserved for traits only. It's very wordy to say that NdArrayBackend implements Backend, when NdArray implements Backend is enough

PR #905, opened 2023-10-26
Fixing Docs.rs

ArvidHammarlund:
At the moment docs.rs fails to compile due to
not being able to link to Cuda or other backend
specific configurations, which it seemingly has
to do to compile the docs of those crates.

I have now removed the backends from being part
of the included features to be displayed on
docs.rs, so that only the crates that pertains
to the upper-level api remains.

As it was me who added the all-features flag in
the first place, I must apologize for any
inconvenience caused!

The resulting docs.rs output can be seen by running: cargo doc -F dataset -F default -F std -F train -F train-tui -F train-metrics -F dataset-sqlite --open

### DIFF
--- a/burn/Cargo.toml
+++ b/burn/Cargo.toml
@@ -62,4 +62,4 @@ burn-core = { path = "../burn-core", version = "0.11.0", default-features = fals
 burn-train = { path = "../burn-train", version = "0.11.0", optional = true, default-features = false }
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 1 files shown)


nathanielsimard:
@ArvidHammarlund No worries, it's hard to actually test the flags without actually pushing on docs.rs. Thanks for fixing it :)

nathanielsimard (approved):

PR #907, opened 2023-10-27
ci/Speed up typos checks

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR speeds up `typos` checks using the binary provided by a release and caches `xtask` and `typos` binaries

### Testing

### DIFF
--- a/xtask/src/runchecks.rs
+++ b/xtask/src/runchecks.rs
@@ -289,9 +289,6 @@ fn std_checks() {
 }
 
 fn check_typos() {
-    // Install typos-cli
-    cargo_install(["typos-cli", "--version", "1.16.5"].into());
-
     println!("Running typos check \n\n");
 
     // Run typos command as child process
--- a/xtask/src/runchecks.rs
+++ b/xtask/src/runchecks.rs
@@ -3,6 +3,7 @@
 //! It is used to check that the code compiles and passes all tests.
 //!
 //! It is also used to check that the code is formatted correctly and passes clippy.
+
 use std::env;
 use std::process::{Child, Command, Stdio};
 use std::str;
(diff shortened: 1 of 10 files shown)


xtask/src/runchecks.rs
```diff
@@ -289,9 +280,6 @@ fn std_checks() {
 }
 
 fn check_typos() {
-    // Install typos-cli
-    cargo_install(["typos-cli", "--version", "1.16.5"].into());
```
nathanielsimard, line 293: Does that mean that we can't check the typos with local checks?

Luni-4: I would suggest, as we have done for coverage, to users to personally install the software that they will run locally, in this way users can choose the desidered version and the correct operating system. In this way `cargo xtask` tasks remain cross-platform.

nathanielsimard: But we are already using Cargo to install typos-cli, so it is cross-platform right now! Maybe we can only install it locally but use the system version when the CI flag is activated?

Luni-4: Yes, but there are many options to be considered when installing a binary on different operating systems, but no problem, it should be fixed now

nathanielsimard (approved):

xtask/src/runchecks.rs
```diff
@@ -289,8 +290,18 @@ fn std_checks() {
...
+    // This path defines where typos-cl is installed on different
```
kurnevsky, line 293: `typos-cl` is a typo :)

Luni-4: Ugh, you're right! eh eh
Can you open a new PR to fix the typo, please?

PR #915, opened 2023-10-31
[burn-import] Add other DataType by conversion to i32

QuantumEntangledAndy:
### Checklist

### Related Issues/PRs

- None

### Changes

The current burn-import main dosent' implement any other data types than int32. 

Instead it marks it as TODO with a note abour converting to i32

https://github.com/burn-rs/burn/blob/96524d40a1cffc8c13394f70775ce53d5a1abd3b/burn-import/src/onnx/proto_conversion.rs#L37-L40

This fills out that todo by converting to i32.

### Testing

- Not yet, just discussing the changes first to see what is the scope of the PR is and where other changes should be made

Should I also for example make any adjustments to the cast nodes

### DIFF
--- a/burn-import/src/onnx/proto_conversion.rs
+++ b/burn-import/src/onnx/proto_conversion.rs
@@ -34,10 +34,62 @@ impl TryFrom<TensorProto> for Tensor {
                     Data::Float32s(tensor.float_data)
                 },
             ),
-            DataType::INT16 => {
-                // TODO : Add support for int16 by converting to int32
-                todo!("Add support for int16");
-            }
+            DataType::UINT8 => (
+                ElementType::Int32,
+                // Convert the raw data to a vector of ints
+                if !tensor.raw_data.is_empty() {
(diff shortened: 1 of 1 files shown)


antimora:
I was traveling and I'll comment on this in coming days.

antimora:
It looks good so far. Let me know if you want to continue working on this. Sorry this PR dropped from my radar since I was traveling and I was working on the tasks.

QuantumEntangledAndy:
I'm involved in some other major projects at the moment so don't have time to really work it up right now. 

Also worried about the larger implications of converting to 32. Usually a model uses 8bit to reduce memory foot prints. Ideally it would be nice to do u8 etc properly but I think that would mean changing all the back ends too.

antimora:
Ok. Let's close it for now. It would be easier if we have a real use case example with an onnx. Otherwise we will be working abstract. 

Feel free to reopen if you decide to work on it.

PR #923, opened 2023-11-01
Add warmup logic when calculating eta

### LINKED ISSUE #912: Autotune causes inaccurate estimated training time
The autotune mechanism runs many benchmarks at the beginning of a training execution, then the best operations are stored and used directly throughout the rest of the training. 

However, the estimated training time shown in the UI is computed with the time spent on the past iterations, times the number of remaining iterations. This is now inaccurate especially in the beginning (the estimation will be pessimistic), since the first iterations will likely take longer than the later ones because of autotune. 

Simple as it sounds, I think this will be non trivial to fix. Maybe at least it will be less blatant when #911 is implemented.

nathanielsimard:
Fix #912

### DIFF
--- a/burn-train/src/renderer/tui/progress.rs
+++ b/burn-train/src/renderer/tui/progress.rs
@@ -1,22 +1,20 @@
-use crate::renderer::TrainingProgress;
-
 use super::TerminalFrame;
+use crate::renderer::TrainingProgress;
 use ratatui::{
     prelude::{Alignment, Constraint, Direction, Layout, Rect},
     style::{Color, Style, Stylize},
     text::{Line, Span},
     widgets::{Block, Borders, Gauge, Paragraph},
 };
-use std::time::Instant;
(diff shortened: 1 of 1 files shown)


louisfd (changes requested):
The estimation has a hard time stabilizing in the beginning (on ag-news-train I had 20 to 40 iterations where the estimation was changing all the time). 
I don't think it's a matter of how much time you wait (I tried at 10 secs then deactivated this condition to wait 30 secs, only to get similar results), but rather the fact that when we start showing the time the iterations window on which it is based is very small because the warmup items are ignored. However I think the last part of the warmup could be used to stabilize the first iterations. 

Also, not directly related to your PR, but there is no singular in the estimate (it writes "1 hours")

nathanielsimard:
@louisfd Merging the current PR, since adding a minimum and maximum data points before showing the estimate is out of scope for this PR.

PR #925, opened 2023-11-02
ci/Add Windows pipeline

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR adds `Windows` CI to `burn`

### Testing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,6 +118,13 @@ jobs:
           curl -L "$GRCOV_LINK/$GRCOV_VERSION/grcov-x86_64-unknown-linux-musl.tar.bz2" |
           tar xj -C $HOME/.cargo/bin
 
+      - name: Install Vulkan
+        if: matrix.os == 'windows-latest'
+        uses: humbletim/install-vulkan-sdk@v1.1.1
+        with:
+          version: latest
+          cache: true
+
       - name: run checks & tests
         run: ${{ matrix.coverage-flags }} ${{ matrix.wgpu-flags }} cargo xtask run-checks ${{ matrix.test }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,18 +96,18 @@ jobs:
           sudo apt-get update
           sudo apt install -y libegl1-mesa libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
 
-      - name: Run cargo clippy for stable version
-        if: runner.os == 'Linux' && matrix.rust == 'stable' && matrix.test == 'std'
-        uses: giraffate/clippy-action@v1
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          # Run clippy for each workspace, targets, and featrues, considering
-          # warnings as errors
-          clippy_flags: --all-targets -- -Dwarnings
-          # Do not filter results
-          filter_mode: nofilter
-          # Report clippy annotations as snippets
-          reporter: github-pr-check
+      # - name: Run cargo clippy for stable version
+      #   if: runner.os == 'Linux' && matrix.rust == 'stable' && matrix.test == 'std'
+      #   uses: giraffate/clippy-action@v1
+      #   with:
+      #     github_token: ${{ secrets.GITHUB_TOKEN }}
+      #     # Run clippy for each workspace, targets, and featrues, considering
+      #     # warnings as errors
+      #     clippy_flags: --all-targets -- -Dwarnings
+      #     # Do not filter results
+      #     filter_mode: nofilter
+      #     # Report clippy annotations as snippets
+      #     reporter: github-pr-check
 
       - name: Install grcov
         if: runner.os == 'Linux' && matrix.rust == 'stable' && matrix.test == 'std'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,7 +135,7 @@ jobs:
           files: lcov.info
 
   check-typos:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: checkout
         uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -127,8 +127,17 @@ jobs:
         if: matrix.os == 'windows-2022'
         uses: napokue/setup-dxc@v1.1.0
 
-      - name: run checks & tests
-        run: ${{ matrix.coverage-flags }} ${{ matrix.wgpu-flags }} cargo xtask run-checks ${{ matrix.test }}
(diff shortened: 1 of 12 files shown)


Luni-4:
@nathanielsimard 

The Windows CI error is related to a part of `burn`'s code I do not know. It seems there is a missing files problem https://github.com/burn-rs/burn/actions/runs/6735375454/job/18308404823#step:9:2067

nathanielsimard:
> @nathanielsimard
> 
> The Windows CI error is related to a part of `burn`'s code I do not know. It seems there is a `missing files` problem https://github.com/burn-rs/burn/actions/runs/6735375454/job/18308404823#step:9:2067
> 
> If I'm not wrong we can fix the problem using https://doc.rust-lang.org/std/env/fn.temp_dir.html

Yeah, we should probably update the tests before merging this PR, using `temp_dir` is better than actually targeting the directory `/tmp` which may not exist on Windows.

antimora:
Saw the hack for windows. Please add the context and when this hack should be removed. Preferably we should have an issue or tondo fix the hack. The commit messages are invisible after a merge so we can't rely on them.

Luni-4:
> Saw the hack for windows. Please add the context and when this hack should be removed. Preferably we should have an issue or tondo fix the hack. The commit messages are invisible after a merge so we can't rely on them.

Thanks @antimora for your review! Yep, I will open an issue as soon as the CI for Windows will be finished highlighting the problematic functions. For now I'm just investigating which issues will be arising from the exection of this PR

Luni-4:
I cannot detect where the problem is in the `burn-wgpu` crate, as you can see from the commits histroy I have tried different approaches, but I always obtain the same error: `No adapter found for graphics API AutoGraphicsApi`

My attempts:

1. Using the same CI for Windows implemented in `wgpu` project, but it failed
2. Update `wgpu` version to `0.18`
3. Tried to only set up **only** the `WGPU_BACKEND: dx12` environment variable

Perhaps the main problem could be in the `burn` code, or in the `wgpu-hal` crate which comunicates directly with the device. I do not know. I do not have a Windows machine ready and usable to test `burn` code, so I can only use this CI machine. If someone is more expert than me with Windows and is willing to help me, I'm saying thanks in advance!

syl20bnr:
Going to look at this one.

syl20bnr:
Here is the current status of my investigations regarding burn-wgpu tests. I manually forced the wgpu backend to Dx12 and played with the priorities to select the GPU or CPU device.

| Backend | Device | Status |
|:--------|:-------|:------:|
| Vulkan  | GPU    | ✅     |
| Dx12    | GPU    | ✅*    |
| Dx12    | CPU    | ❌     |

(*) Passes minus 3 tests. It requires to tweak the [WORKGROUP_DEFAULT](https://github.com/Tracel-AI/burn/blob/main/burn-wgpu/src/kernel/base.rs#L13) value from 32 to 16 (my GPU is an RTX 4080).

Disabling parallel execution of tests with the Dx12 backend results in faster performance than the default parallelism setting, achieving speeds up to twice as fast.

Dx12 with CPU device is crashing during matmul operations.

I'll perform some cleanup in the branch. I believe we need a more flexible method to select the wgpu backend and device for test execution. We could implement this either through a specific configuration or possibly through code generation. Opting for a configuration parameter might be more intuitive and could simplify integration, making it easier and more adaptable for CI.

Luni-4:
@syl20bnr 

Wow, this is fantastic! Thanks a lot for your work and your analysis! 

I really like your configuration parameter idea. Just one thing, would it be possible to move the `temp_dir` function into `burn-common` crate? In this way, we can use that for other future tests

syl20bnr:
PR now ready for review. During the investigation phase, several temporary debug changes were introduced. I would appreciate a careful examination to ensure all such changes have been appropriately addressed or reverted.

**Key Points for Review:**

1. **Enabling wgpu Tests on macOS:** I've re-enabled wgpu tests for macOS as they appear to be passing without issues. Do we want to keep them enabled ?

2. **Disabling wgpu Tests on Windows:** Due to current issues, wgpu tests on Windows have been disabled. I created an issue for tracking: [Issue #1062](https://github.com/tracel-ai/burn/issues/1062). Please review and provide any additional insights or suggestions.

**Request for Actions:**

- @Luni-4: Could you please assist by marking this draft PR as ready for review? Your insights will also be valuable in the review process.

.github/workflows/test.yml
```diff
@@ -33,30 +33,39 @@ jobs:
   tests:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
Luni-4, line 36: Can we remove this line?

.github/workflows/test.yml
```diff
@@ -33,30 +33,39 @@ jobs:
...
             test: std
-            os: ubuntu-22.04
-          - wgpu-flags: DISABLE_WGPU=1
+          - os: macos-13
             rust: stable
             test: std
-            os: macos-13
+          - os: windows-2022
+            # powershell format
+            wgpu-flags: '$env:DISABLE_WGPU = "1";'
```
Luni-4, line 55: Since we use `shell: bash` for some Windows steps, do you think we should use the `bash` syntax here or better to leave the `Powershell` one?

.github/workflows/test.yml
```diff
@@ -107,25 +116,70 @@ jobs:
...
+
+      - name: (windows) run wgpu-info
+        if: runner.os == 'Windows'
+        shell: bash
+        run: |
+          echo "$PATH"
+          cargo install wgpu-info
+          which wgpu-info
+          export RUST_LOG=trace
+          wgpu-info
```
Luni-4, line 170: Do we want to leave this step as check or can we remove it?

Luni-4 (commented):
Thanks a lot for your hard word @syl20bnr!

In my opinion we can leave `wgpu` tests enabled for `MacOS`, better to have a check more than one less. 

As final step, can we move the filepath and `temp_path` mechanism and functions in `burn-common`?

syl20bnr:
> As final step, can we move the filepath and temp_path mechanism and functions in burn-common?

Sure, I forgot about that.

.github/workflows/test.yml
syl20bnr: I'll remove it, it was more for debugging purpose.

.github/workflows/test.yml
syl20bnr: Nice catch, I'll remove it.

syl20bnr: Done.

.github/workflows/test.yml
syl20bnr: I switched to bash shell.

.github/workflows/test.yml
syl20bnr: Removed.

nathanielsimard (approved):
LGTM 🙏

PR #926, opened 2023-11-02
Autotune: fix inputs

louisfd:
**Batch sizes**
At first we thought batch sizes would have no incidence on kernels, i.e. the time taken for many batches would just be proportional to the time for one batch. So for autotune, to run faster, we had reduced the number of batches to a constant. However, our tests have shown that the matmul algorithms have different behaviours on different batch sizes, therefore it is important to keep them into the computations. 

**Random inputs**
Before, we had only tensors of ones in autotune. It is possible that there will someday exist algorithms that have a biased behaviour on such trivial inputs, so it has to be replaced by random inputs. I chose uniform in [-10., 10.] but that's arbitrary.


burn-wgpu/src/kernel/prng/uniform.rs
```diff
@@ -51,6 +51,39 @@ pub fn random_uniform<G: GraphicsApi, E: WgpuElement, const D: usize>(
...
+    let kernel = StaticKernel::<
+        KernelSettings<UniformPrng, E, i32, WORKGROUP_DEFAULT, WORKGROUP_DEFAULT, 1>,
+    >::new(workgroup);
+
+    tensor.client.execute(
+        Box::new(kernel),
+        &[&output.handle, &info_handle, &args_handle],
+    );
+
+    output
```
nathanielsimard, line 84: It should call `random_uniform` instead of duplicating the same logic.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #930, opened 2023-11-03
Fixed the wrong order of the attributes

Aisuko:
### Checklist

### Related Issues/PRs

### Changes

The construct follow the attributes order in the struct.

### Testing


nathanielsimard (approved):
When using names it doesn't really matter, but for consistency it's not a bad idea!

Issue #935, opened 2023-11-07
Burn-Fusion: Caching relative graphs


louisfd:
When doing kernel fusion in Burn-Fusion, we want to be able to recognize graph patterns we have already seen so we do not need to go through the list of every Optimizations (this will save time especially when there are many Optimizations to check). 

The idea we have is to cache hashed relative graphs incrementally. Let me explain:
- Relative computation graph: we are going to see patterns such as `let C = A + B`. If we see later `let D = B + E` then it may be the same pattern. Although the tensors have changed, we should take the same optimization decisions.
- Incrementally: When we see a line like `let C = A + B`, followed by, let's say, `let D = C - 4`, then the cache should first hold what to do with `A`, then what to do with `A + B`, then what to do with `let C = A + B`, then what to do with `let C = A + B; let D = C - 4`. These should be four different entries in the cache. The first three ones should tell something like "continue", while the fourth may say "use this optimization", then the optimization is applied to the whole pattern of two lines. 
- Hashed: to be stored in the cache, these patterns (`A`, `A + B`, `let C = A + B`, `let C = A + B; let D = C - 4`) should be hashed. Note that they will not be string literals like shown in here for simplicity, but rather complex structs. Since we are always adding new stuff to already existing hashes, it would be a waste to recompute hashes from the beginning all the time. We should leverage the fact that hash functions evolve as we append stuff to the input, so we can progressively build only one hash and use several versions of it at different stages.

PR #939, opened 2023-11-09
Feat/op fusion decorator

nathanielsimard:
## Burn Fusion

A new backend decorator to perform graph optimizations on a stream of operations. Backends that want to support this functionality can implement the new `FusionBackend` trait in addition to the `Backend` trait.

This allows the backend to provide a list of potential optimizations to be performed instead of executing a list of normal backend operations.

For now, only the Wgpu backend implements the new trait, but it allows to actually test that the new backend works even without optimizations.

Following PRs will bring new optimizations to the Wgpu backend 🔥

## Overhead ?

There is some framework overhead introduced by this decorator, but since most backends are async, it shouldn't impact the speed of execution. In fact, I enabled the new backend decorator on the text classification example and the training time isn't longer than before.

The actual overhead introduced is very minimal and is constant per operation. We will introduce a cache to reduce the overhead of fusion operation implementations #935


louisfd (approved):
I've committed my (superficial) change requests, so it's good to go!

PR #943, opened 2023-11-10
Perf/wgpu/reduce dim

louisfd:
I wrote a new algorithm for reduction in wgpu. 
Contrary to the former, it can use many threads in the computation of a single output element, leveraging shared memory. 
Therefore, it is better on unbalanced tensors with shapes like [50, 10000, 50] (when reduce dim is the large one, here 1)
However it is slower on balanced tensors like [512, 512, 512], because the former was already very well parallelized in those cases. 
It makes it a very good case for autotune, which I have implemented too.
For now, sum_dim and mean_dim are done with the new algorithm, with autotune. ArgsMax, ArgsMin and Sum haven't been done yet.


nathanielsimard (approved):
🏎🏎🏎

PR #958, opened 2023-11-15
Add new backend comparison benchmark

nathanielsimard:
Benchmark how well a backend executes a custom activation function with a lot of basic tensor operations.


louisfd (approved):
That's all right :)

PR #962, opened 2023-11-16
Update readme

nathanielsimard:
Rendered link: https://github.com/burn-rs/burn/tree/update-readme


louisfd:
❤️‍🔥

antimora (changes requested):
Love the new updates. This README file is doing exactly what it's supposed to do: serve as a landing page for new users!

However, I highly recommend considering the following changes:

1. Consider shortening the "Automatic kernel fusion 💥" subsection by removing the large chunk of code. When I first opened it, I felt overwhelmed and worried that subsequent sections might be equally extensive and complex. This information could be relocated elsewhere for users interested in learning more.

2. It may not be immediately clear that subsections are expandable. Perhaps add a note at the top indicating that users should click to expand. Otherwise, they might appear to be just bullet points.

3. I suggest replacing the "🧨" (explosive) emoji after "Asynchronous Execution" with a "🐇" (rabbit) emoji. The "🧨" feels a bit intense, especially considering current global events involving violence.

4. Consider prefacing "plugins" with "swappable" in the phrase "backends as plugins," and bolding both words. This emphasizes the ease of swapping backends. Recently, someone mentioned on Discord that this was a decisive factor in choosing Burn, so highlighting this feature is important.

5. After WGPU, add "WebGPU API" in parentheses. Not everyone knows that WGPU is an implementation of WebGPU APIs. Keep in mind that people landing on this README might be new to Rust and unfamiliar with WGPU.

6. Replace the "🥵" emoji after "any backend" with "🔄."

7. I think "Training & Inference" should be moved to the **top** and expanded upon. This section doesn't fully emphasize how central these two components are. Burn was built from the ground up with training in mind. It's also worth noting how Burn, in comparison to frameworks like PyTorch, simplifies the transition from training to deployment, eliminating the need for code changes.

8. Add a "no-std" subsection (under training/inference) to clarify that the core parts of Burn are "no-std" compatible and enjoy first-class support.

9. Add links from each Backend to a backends Readme files. 

10. Also add back the removed platform support matrix to these backend readme files, which is an important information and should not be lost. 

@nathanielsimard @louisfd

alxpettit:
I highly recommend keeping the hot_face emoji. It's funny. xD

louisfd:
I thought so too 😆 I'll try to think of a more relevant spot for it

PR #963, opened 2023-11-16
feature(norm): Add GroupNorm

### LINKED ISSUE #685: Group Normalization
### Feature description

Group Normalization as described in https://pytorch.org/docs/stable/generated/torch.nn.GroupNorm.html

### Feature motivation
Group Normalization is quite present in lots of the more recent SOTA deep learning models and performs quite well, would be a nice addition to have.
 
### Suggest A Solution

Here's a rough example from a project I'm working assuming a 4d tensor. Not sure how I'd make it with ```n``` dimensions agnostic though:

```rust
#[derive(Config)]
pub struct GroupNormConfig {
    pub num_groups: usize,
    
    pub num_channels: usize,
    
    #[config(default = 1e-5)]
    pub epsilon: f32,
    
    #[config(default = true)]
    pub affine: bool,
}

impl GroupNormConfig {
    pub fn init<B: Backend>(&self) -> GroupNorm<B> {
        match self.affine {
            true => {
                if self.num_channels % self.num_groups == 0 {
                    panic!("Groups and channels have to be divisible by each other")
                }
                
                let gamma = Tensor::ones([1, self.num_channels, 1, 1]);
                let beta = Tensor::zeros([1, self.num_channels, 1, 1]);
                 GroupNorm {
                    num_groups: self.num_groups,
                    num_channels: self.num_channels,
                    epsilon: self.epsilon,
                    affine: self.affine,
                    gamma: Some(Param::from(gamma)),
                    beta: Some(Param::from(beta))
                }
            },
            false => {
                if self.num_channels % self.num_groups == 0 {
                    panic!("Groups and channels have to be divisible by each other")
                }
                
                  GroupNorm {
                    num_groups: self.num_groups,
                    num_channels: self.num_channels,
                    epsilon: self.epsilon,
                    affine: self.affine,
                    gamma: None,
                    beta: None
                }
            }
        }


    }
}
#[derive(Debug, Module)]
pub struct GroupNorm<B: Backend> {
    num_groups: usize,
    num_channels: usize,
    epsilon: f32,
    affine: bool, 
    gamma: Option<Param<Tensor<B, 4>>>,
    beta: Option<Param<Tensor<B, 4>>>
}

impl<B: Backend>GroupNorm<B> {
    pub fn forward(&self, x: Tensor<B, 4>) -> Tensor<B, 4> {
        let x_shape = x.dims();
        let x = x.reshape([x_shape[0], self.num_groups, (self.num_channels / self.num_groups) as usize, x_shape[2], x_shape[3]]);
        
        let x_mean = x.clone()
        .mean_dim(2)
        .mean_dim(3)
        .mean_dim(4);
        
        let x_var = x.clone()
        .var(2)
        .var(3)
        .var(4);
        
        let x = (x - x_mean) / (x_var + self.epsilon).sqrt();
        let x = x.reshape(x_shape);
        
        match self.affine {
            true => {
                let x = self.gamma.clone().unwrap().val() * x + self.beta.clone().unwrap().val();
                x
            },
            false => x
        }
    }
}
```

dcvz:
### Checklist

### Related Issues/PRs

Resolves #685


antimora:
@nathanielsimard and @louisfd would know more, but backward pass of GroupNorm also needs to be implemented.

burn-core/src/nn/norm/group.rs
```diff
@@ -0,0 +1,179 @@
...
+
+/// Applies Group Normalization over a mini-batch of inputs.
+///
+/// `Y = groupnorm(X) * γ + β`
+#[derive(Module, Debug)]
+pub struct GroupNorm<B: Backend> {
+    num_groups: usize,
+    num_channels: usize,
+    gamma: Param<Tensor<B, 1>>,
+    beta: Param<Tensor<B, 1>>,
```
nathanielsimard, line 34: I would wrap those in `Option` so when group norm is configured with `afine=false` no weights are added!

burn-core/src/nn/norm/group.rs
```diff
@@ -0,0 +1,179 @@
...
+    use burn_tensor::Data;
+
+    #[cfg(feature = "std")]
+    use crate::TestBackend;
+
+    #[cfg(not(feature = "std"))]
+    use crate::TestBackend;
+
+    #[test]
+    fn group_norm_forward() {
```
nathanielsimard, line 134: I think it would be valuable to have two tests, one with `affine=true` and one with `affine=false`.

nathanielsimard (changes requested):
Thanks a lot for this implementation! I only have minor comments, they should be easy to fix!

nathanielsimard:
> @nathanielsimard and @louisfd would know more, but backward pass of GroupNorm also needs to be implemented.

Not at this level, the autodiff system will take care of that!

antimora:
> > @nathanielsimard and @louisfd would know more, but backward pass of GroupNorm also needs to be implemented.
> 
> Not at this level, the autodiff system will take care of that!

Is it because GN uses Burn op primitives, which Autodiff already implements?

burn-core/src/nn/norm/group.rs
dcvz: What would you like to see tested in affine true? that the gamma and beta get set in the struct? Or that gamma and beta are getting updated during learning?

nathanielsimard: Just testing the forward pass with another set of numbers

dcvz:
@nathanielsimard I think I've addressed everything now!

antimora:
> @nathanielsimard I think I've addressed everything now!

Failing due to formatting

dcvz:
> > @nathanielsimard I think I've addressed everything now!
> 
> Failing due to formatting

Should be resolved now

antimora:
Can we also add an entry to the book? 

Here is the section: https://burn.dev/book/building-blocks/module.html#general

dcvz:
> Can we also add an entry to the book?
> 
> Here is the section: https://burn.dev/book/building-blocks/module.html#general

Added!

dcvz:
Looks like CI is failing but not due to the PR but in the install lavapipe. Might need a retrigger

antimora (approved):
LGTM. Thank you!

louisfd:
@nathanielsimard your requests have been fulfilled, so i merge

PR #966, opened 2023-11-18
feature(tensor): Add unsqueeze_dim helper

### LINKED ISSUE #876: Unsqueeze at Any Dimension
### Feature description

Implement a more flexible unsqueeze() function that matches the behavior of PyTorch's unsqueeze(). The new function should allow inserting a singleton dimension anywhere in the tensor, as opposed to only prepending.

### Feature motivation

The current implementation of *unsqueeze()* only prepends singleton dimensions. This limitation imposes constraints when manipulating tensor shapes. Having a more flexible unsqueeze operation will increase the usability and compatibility of the library.

* Modify the unsqueeze() function to take an additional argument, dim, which indicates the position where the singleton dimension should be inserted. Something like this:
```rust
// Before `dim`
dims[0..dim].copy_from_slice(&shape.dims[0..dim]);
// Insert singleton dimension at `dim`
dims[dim] = 1;
// After `dim`
dims[(dim+1)..].copy_from_slice(&shape.dims[dim..]);
```

dcvz:
### Checklist

### Changes

Adds a `unsqueeze_dim` method allowing you to target a specific dimension.

### Testing

I've added tests for the added method

### Related Issue

fixes #876

### DIFF
--- a/burn-tensor/src/tensor/api/base.rs
+++ b/burn-tensor/src/tensor/api/base.rs
@@ -257,6 +257,40 @@ where
         self.reshape(shape)
     }
 
+    /// Creates a new tensor with a dimension of size one inserted at the specified position.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::{Tensor, Shape};
+    ///
+    /// fn example<B: Backend>() {
+    ///     let tensor = Tensor::<B, 2>::ones(Shape::new([3, 3]));
+    ///     let tensor = tensor.unsqueeze_dim::<3>(1);
+    ///     println!("{:?}", tensor.shape());
+    ///     // Shape { dims: [3, 1, 3] }
+    /// }
+    /// ```
+    pub fn unsqueeze_dim<const D2: usize>(self, dim: usize) -> Tensor<B, D2, K> {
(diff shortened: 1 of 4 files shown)


burn-tensor/src/tensor/api/base.rs
```diff
@@ -257,6 +257,40 @@ where
...
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::{Tensor, Shape};
+    ///
+    /// fn example<B: Backend>() {
+    ///     let tensor = Tensor::<B, 2>::ones(Shape::new([3, 3]));
+    ///     let tensor = tensor.unsqueeze_dim::<3>(1);
```
antimora, line 270: Could we use this style instead?

```rust
let tensor:Tensor<B, 3> = tensor.unsqueeze_dim(1);
```

`<3>` stands out compared to other Burn examples.

antimora (changes requested):
Looks good. A couple changes requested:

1. Update the example with a preferred style. 
2. Update the Book tensor section with the new op: https://burn.dev/book/building-blocks/tensor.html#basic-operations

dcvz:
@antimora I think I've addressed all comments!

antimora (approved):
Thank you. Looks good.

PR #967, opened 2023-11-19
Fix image classifier build failure by upgrading versions

antimora:
### Checklist

- [x] Confirm that `run-checks` script has been executed.

### Changes

The image classifier web example was broken because of auto-tune was enabled by default. This fixes by 

### Testing

1. Ran the run-checks all
2. Ran the server and tested on the browser.

PR #969, opened 2023-11-19
Disable udeps checking till all unused deps are resolved

antimora:
### Changes

udeps is failing the current main branch since there are many false positives are unresolved.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -40,14 +40,14 @@ jobs:
           toolchain: nightly
           components: rustfmt
 
-      - name: Install cargo-udeps
-        env:
-          UDEPS_LINK: https://github.com/est31/cargo-udeps/releases/download
-          UDEPS_VERSION: v0.1.43
-        run: |
-          curl -L "$UDEPS_LINK/$UDEPS_VERSION/cargo-udeps-$UDEPS_VERSION-x86_64-unknown-linux-gnu.tar.gz" |
-          tar xz -C $HOME/.cargo/bin --strip-components 2
+      # - name: Install cargo-udeps
+      #   env:
+      #     UDEPS_LINK: https://github.com/est31/cargo-udeps/releases/download
+      #     UDEPS_VERSION: v0.1.43
+      #   run: |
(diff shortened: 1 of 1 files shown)


antimora:
CCing @Luni-4 

I noticed that my PR build is failing: https://github.com/Tracel-AI/burn/pull/967

I attempted to fix it but there were more. I think we should disable till we whitelist the false positives.

Luni-4:
@antimora 

Instead of disabling the check, we can mark the false-positivies as we have done here https://github.com/Tracel-AI/burn/blob/e5c6044062050bcee74d1e15369c9763ad7dd736/burn-dataset/Cargo.toml#L55-L56 in the specific crate. I guess the problem is that https://github.com/Tracel-AI/burn/pull/967/commits/7154755b0613ed4d505994a408c207b2e0ac6f32#diff-b1776cd66202f5cfa4b0878c00a6cc9fe41d82b86e303b2c7f485c5877b5e998R22 is not used in the code, that's why it is marked as unused deps. If we remove that dep, we cannot build the example, right?

antimora:
@nathanielsimard @Luni-4 this is a blocker. It takes too long to resolve all the issues. Lets disable it first and enable in another PR. I am not even sure how it was possible to enable it in the first place if there were unused dependencies in the first place. Pretty much most are false positive. 

My personal take is that we should not enforce upon PR check but rather audit the code time to time because the tooling of udep is huge to install on users computer and it finds too many false positive. Having an unused dependency for a short period is not super urgent.

Luni-4:
@antimora 

All right. Can we move this check before releasing a new `burn` version instead? In this way, we can remove `unused` deps from the current CI. Once the `dependencies` checker will be implemented within `xtask`, it will be simpler to detect dependencies locally. What do you think?

antimora:
@Luni-4 it is a great idea! I will update the PR.

antimora:
> @Luni-4 it is a great idea! I will update the PR.

I think it might cause a headache if we do this right before the release. Actions support scheduled scan. I think we should schedule it similar to other security scans.

antimora:
I filed a ticket (https://github.com/Tracel-AI/burn/issues/975) to take up and fix this in other issue.

nathanielsimard (approved):

PR #978, opened 2023-11-21
Support count_include_pad attr in avg_pool2d ONNX

### LINKED ISSUE #636: Support for `count_include_pad=False` for AvgPool1d and AvgPool2d
#### Description

The current implementation of the `AvgPool2d` layer in the Burn framework assumes that `count_include_pad=True`, meaning that zero-padding is included in the average pooling calculation. While this approach is valid and useful in some contexts, it can lead to limitations in flexibility and compatibility with other frameworks and standards.

#### Proposal

I propose adding a `count_include_pad` option to `AvgPool2dConfig`, with support for both `True` and `False` values. This change will allow users to configure whether or not padding is included in the average calculation, aligning more closely with common practices and standards in deep learning.

#### Why is this Important?

1. **Alignment with ONNX Specification**: The ONNX specification defaults to `count_include_pad=False` for average pooling. Aligning with this standard will enhance compatibility and make it easier for users to transfer models between Burn and other frameworks that adhere to the ONNX specification.

2. **Increased Flexibility**: By offering both options, users can choose the behavior that best fits their specific use case and dataset. This can lead to better performance and more accurate results, depending on the context.

3. **Avoiding Potential Issues**: Including padding in the average calculation (when not desired) might lead to down-weighting of actual feature values near the edges of the input. This can potentially result in a loss of information and a degradation in performance.

#### Action Items

1. [x] AvgPool1d in `core`
2. [x] AvgPool2d in `core`
3. [ ] AvgPool1d in `burn-import`
4. [x] AvgPool2d in `burn-import`

#### References
1. [ONNX AveragePool](https://onnx.ai/onnx/operators/onnx__AveragePool.html#attributes)
5. [AvgPool2d Pytorch](https://pytorch.org/docs/stable/generated/torch.nn.AvgPool2d.html)

antimora:
### Checklist

### Related Issues/PRs

Fixes #636

### Changes

1. `count_include_pad` attribute is supported now in avg_pool2d ONNX. Previously PyTorch was exporting an extra padding node which they stopped doing in the new version of PyTorch. 

### Testing

1. Updated ONNX file testing with new test cases. 
2. Ran checks.


nathanielsimard (approved):
LGTM

Luni-4 (commented):
LGTM for me too!

PR #979, opened 2023-11-21
Feat/backend comparison/persistence

louisfd:
In backend-comparison, the benchmarks are now persisted locally in ~/.cache as a json, the idea being that we may want to draw graphs from this data. 

Tests:
Tested locally. 
No unit tests because it's IO-bound


nathanielsimard (approved):
LGTM

Issue #982, opened 2023-11-21
Module Does Not Support Tensor<B, D, Int>


dcvz:
**Describe the bug**
For some reason Module structs do not allow having a Tensor of Kind = Int.

**Screenshots**

![Nov 21 Screenshot](https://github.com/Tracel-AI/burn/assets/2475932/5da1eea8-c0de-48e8-b219-be00cc8b3e55)

![Screenshot_2023-11-21_at_13 53 30](https://github.com/Tracel-AI/burn/assets/2475932/737e536e-85ac-430d-8112-18276f5a0336)

antimora:
This should be fixed now.

dcvz:
Can confirmed as fixed!

Issue #665, opened 2023-08-20
Implement Module<B> for Param<Tensor<B, D, Int>> and Param<Tensor<B, D, Bool>>


antimora:
**Current Behavior:**
The `Module<B>` trait is currently implemented only for `Param<Tensor<B, D, Float>>`. This leaves out support for both Boolean and Integer types.

**Error Message:**
When attempting to use an integer tensor, the following error is displayed:

```plaintext
#[derive(Module, Debug)]
   |          ^^^^^^ the trait `Module<B>` is not implemented for `Param<Tensor<B, 4, Int>>`
   |
   = help: the trait `Module<B>` is implemented for `Param<Tensor<B, D>>`
   = note: this error originates in the derive macro `serde::Serialize` 
(in Nightly builds, run with -Z macro-backtrace for more info)
```

**Code Sample:**
Here is an example of Rust code that illustrates the issue:

```rust
#[derive(Module, Debug)]
pub struct Model<B: Backend> {
    constant1: burn::module::Param<Tensor<B, 4, Int>>,
    phantom: core::marker::PhantomData<B>,
}
```

**Proposed Enhancement:**
Extend the implementation of `Module<B>` to include `Bool` and `Int`, accommodating use cases where integer and boolean tensors are stored as module fields.

antimora:
This should be fixed now.

PR #984, opened 2023-11-21
Fix/constant tensors

### LINKED ISSUE #461: Refactor and Simplify Device Handling
### Description
In the process of developing a Tensor module, it became clear that the current handling of device migration for modules can be improved and simplified. Specifically, the interaction between `to_device`, `map`, and `visit` methods is not intuitive and leads to confusion when trying to change the device of a module. 

Furthermore, we also found a problem with how the constant tensors are handled. It was observed that when the parent module is moved to a different device, the tensor field does not move to the new device. This needs to be addressed.

The code has evolved over time and seems to have been adapted to new requirements without a significant refactor. The team has recognized that there is a need for this code to be reviewed and restructured in order to simplify the mental model and minimize surprises for future developers.

### Steps

1. Refactor the device handling in the modules to improve and simplify the interaction between `to_device`, `map`, and `visit` methods.
2. Resolve the issue where the constant tensors do not move along with their parent module when changing devices.
3. Review and adjust the usage of the "B: Backend" reference in order to reduce confusion and align with the Rust standard lib behavior as closely as possible.
4. Perform an overall code review and refactor to simplify and clarify the codebase, keeping in mind the principle of adhering as closely as possible to the behaviors of "native" traits and derives in the Rust standard lib.

### Related Tickets

* https://github.com/burn-rs/burn/pull/459
* https://github.com/burn-rs/burn/issues/452

nathanielsimard:
In this effort to support constant tensors of any kind (`Int` and `Bool`, not just `Float`), I had to fix how we handle devices within modules.

Fix #461 #665 #982

### DIFF
--- a/burn-core/src/module/param/tensor.rs
+++ b/burn-core/src/module/param/tensor.rs
@@ -5,6 +5,7 @@ use crate::tensor::{
     Tensor,
 };
 use alloc::vec::Vec;
+use burn_tensor::{Bool, Int};
 
 impl<B: Backend, const D: usize> From<Tensor<B, D>> for Param<Tensor<B, D>> {
     fn from(value: Tensor<B, D>) -> Self {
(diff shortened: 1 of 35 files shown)


burn-core/src/module/param/tensor.rs
```diff
@@ -4,22 +4,36 @@ use crate::tensor::{
...
     Tensor,
 };
+use alloc::vec::Vec;
+use burn_tensor::{Bool, Int};
 
 impl<B: Backend, const D: usize> From<Tensor<B, D>> for Param<Tensor<B, D>> {
     fn from(value: Tensor<B, D>) -> Self {
         Param::new(ParamId::new(), value.require_grad())
     }
 }
```
antimora, line 15: Do we need to explain why we do not require grad?

antimora (approved):
Looks good. I had one comment. Please see below.

burn-core/src/module/param/tensor.rs
nathanielsimard: I added why we add require grad, other tensor (bool and int) can't have gradients

dcvz (approved):
Nice work! 🔥 🚀 

The `devices()` needing devices looks a bit odd to me. Especially when it sometimes leads being used as `self.devices(Vec::new())` but I think I'm maybe missing something.

antimora:
OK. We got a couple approvals and I need this PR for my onnx changes, hence I am merging.

nathanielsimard:
> Nice work! 🔥 🚀
> 
> The `devices()` needing devices looks a bit odd to me. Especially when it sometimes leads being used as `self.devices(Vec::new())` but I think I'm maybe missing something.

Maybe we could rename the method, but we collect all the devices in the module tree in the provided vector. The alternative would imply a lot more allocations. Maybe we could rename the method to `collect_devices` and add a default implementation `devices` that uses `collect_devices` underneath.

PR #988, opened 2023-11-21
Refactor/fusion/graph

nathanielsimard:
The goal of this PR is to remove the operation from the graph description, it's going to be needed when we will need to compute an ID for a graph description with a global and local graph version. It also simplifies a bit the code.

I had to modify the Distribution struct to remove the element generic, now it's more consistent.


louisfd (approved):
Seems legit, it does look like some places are much cleaner, in particular you leave out many `box`es

PR #990, opened 2023-11-22
Fix/devices api

nathanielsimard:
Linked to #984


dcvz (approved):
I like this a lot more! It keeps the improvement but also doesn’t muddy the base API until you care about it.

PR #993, opened 2023-11-23
feat: add nix flake for local dev

Kodylow:
### Checklist

- [ ] Made sure the book is up to date with changes in this PR. (Local Developer Environment changes, does not apply)

### Changes

Feature: adds nix flake for reproducible, deterministic builds and developer environments.
Feature: uses crane for incremental cargo dependency builds 
Feature: adds [flakebox](https://github.com/rustshop/flakebox) developer pipeline with typos, semgrep, and git hooks (commented out rustfmting, linting and githooks for now because those should be in a separate PR)
Feature: adds justfile for `just` command running
Fix: adds shebangs to shell scripts

### Testing

All developer environment changes. run-checks works in and out of nix developer shell


antimora:
I am hesitant to accept this PR because it contains so much generated code and I am not sure who would maintain all this. I am not even sure there are enough users for this.

Is there any way to maintain the port outside the code base? 

@nathanielsimard @syl20bnr @Luni-4 , please make a call on this.

Luni-4:
I do not know `nix` at all, but I agree about the great amount of generated code and the relative difficulty to maintain it.

@Kodylow Would it be possible to convert all of these scripts into a `xtask` file and then produce all the necessary artifacts from there? In this way everything is concentrated in a single point and, if the code is well documented, other developer can help in maintaining it

Kodylow:
Yup I can simplify it down to just the nix flake and remove the flakebox stuff that affects the git actions and workflow. 

Nix just lets you do the deterministic, reproducible builds so it'll download and build with the correct versioning for everything by just running `nix develop`

antimora:
Closing for now. The comments/feedback can be address in another PR.

PR #995, opened 2023-11-24
Test for max_dim/min_dim

gzsombor:
To prove, that if the dimension is set to 0, the calculation fails. The testcases are disabled

### Checklist

### Changes

I've tried to use tensor.max_dim(0) and noticed, that it doesn't work - as opposed to tensor.sum_dim(0) or tensor.mean_dim(0).
This doesn't change any code, just adds more tests - currently the broken tests are disabled.

### Testing

### DIFF
--- a/burn-tensor/src/tests/ops/maxmin.rs
+++ b/burn-tensor/src/tests/ops/maxmin.rs
@@ -14,6 +14,19 @@ mod tests {
     }
 
     #[test]
+    fn test_max_dim_with_indices_2d_with_dim_0th() {
+        let tensor = TestTensor::from_floats([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]);
+
+        let (output_actual, index_actual) = tensor.max_dim_with_indices(0);
+
+        let output_expected = Data::from([[3., 4., 5.]]);
+        let index_expected = Data::from([[1, 1, 1]]);
+
+        assert_eq!(output_expected, output_actual.into_data());
+        assert_eq!(index_expected, index_actual.into_data());
+    }
+
+    #[test]
     fn test_max_dim_with_indices_2d() {
         let tensor = TestTensor::from_floats([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]);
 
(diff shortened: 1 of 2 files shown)


burn-tensor/src/tests/ops/maxmin.rs
```diff
@@ -48,4 +48,72 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_min_dim_with_indices_2d_with_0th_dim() {
+        let tensor = TestTensor::from_floats([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]);
+
+        let (output_actual, index_actual) = tensor.min_dim_with_indices(0);
+
+        let output_expected = Data::from([[0., 1., 2.]]);
+        let index_expected = Data::from([[0], [0], [0]]);
```
louisfd, line 114: Should be `[[0, 0, 0]]`

burn-tensor/src/tests/ops/maxmin.rs
```diff
@@ -48,4 +48,72 @@ mod tests {
...
+        assert_eq!(output_expected, output_actual.into_data());
+    }
+
+    #[test]
+    fn test_max_dim_2d_with_0th_dim() {
+        let tensor = TestTensor::from_floats([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]);
+
+        let output_actual = tensor.max_dim(0);
+
+        let output_expected = Data::from([[3.], [4.], [5.]]);
```
louisfd, line 103: Should be `[[3., 4., 5.]]`

louisfd (changes requested):
Thanks for pointing out this bug!! 

I've investigated the cause and I'm pretty sure it's because in 
`burn-tensor/src/tensor/ops/tensor.rs`, in `max_dim`, `min_dim`, `max_dim_with_indices` and `min_dim_with_indices` 
we do 
`B::gather(D-1, tensor, index)`
while it should be 
`B::gather(dim, tensor, index)`

Could you fix it in that same PR? If you do please address my two other comments or your tests won't pass. Then if everything passes you can merge your test module with the one above. Thanks again!

gzsombor:
Yes, you are right, in all of your comments :)

louisfd (approved):
Thanks! Looks good now, I'll merge it 😄

PR #996, opened 2023-11-24
feature(tensor): Add narrow op

dcvz:
### Checklist

### Changes

Adds a narrow op to tensors

### Testing

Tests are there for happy and error cases


burn-tensor/src/tests/ops/narrow.rs
```diff
@@ -0,0 +1,41 @@
...
+    use super::*;
+    use burn_tensor::{Data, Shape, Tensor};
+
+    #[test]
+    fn test_narrow() {
+        let tensor: Tensor<TestBackend, 2> =
+            Tensor::from_data(Data::from([[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]]));
+
+        let output = tensor.narrow(0, 0, 2);
+        assert_eq!(output.shape(), Shape::from([2, 3]));
```
louisfd, line 12: I think a test comparing tensor data rather than just the shape would be more reassuring and self-explaining. 

Something like
```rust
let data_expected = Data::from([[1., 2., 3.], [4., 5., 6.]]);
assert_eq!(data_expected, data_actual);
```

Also you could add a test for the other dim if you want

louisfd (changes requested):
That's a nice API addition which will often be more straightforward that using slice 😄 
I have just a little request regarding tests

burn-tensor/src/tests/ops/narrow.rs
dcvz: Yup, forgot that, already adding it!

dcvz:
@louisfd addressed the comment!

louisfd (approved):
That's awesome, thank you!

PR #997, opened 2023-11-24
chore(tensors): fix narrow error messages

dcvz:
### Checklist

### Related Issues/PRs

Set out to fix a typo in the error messaged returned by `narrow` - also improves the message for start and dim to be more clear about what's out of bounds.


louisfd (approved):
Thanks, it's better indeed!

Issue #970, opened 2023-11-19
Add `chunk` Tensor Method


dcvz:
### Feature description

Add a `chunk` method on tensor to faciliate splitting a tensor into a given number of chunks

### Feature motivation

Its a fairly common operation supported by other ML frameworks.

### Actions

- [x] Default method implementation (@dcvz)
- [x] Backend implementation (@Kelvinyu1117)

dcvz:
> I would like to work on this issue, can you assign it to me?

Nice! Let me know if you need any guidance. You can also find me in the Discord as `@dcvz`.

antimora:
> I would like to work on this issue, can you assign it to me?

Done

Kelvinyu1117:
I just had a brief look at the source code, the file I should modify is `src/tensor/api/base.rs` right, and the interface is as follows:

```
pub fn chunk<const D2: usize>(Tensor<B, D2, K>& tensor , chunks: usize, dim: usize = 0) -> Vec<Tensor<B, D2, K>>
{
}
```

dcvz:
> I just had a brief look at the source code, the file I should modify is `src/tensor/api/base.rs` right? And the interface is as follows:

Correct! I think interface can likely be the following, if you do the chunking in base space:

```rust
pub fn chunk(self, chunks: usize, dim: usize) -> Vec<Self>
```

Although given that some of the backends already have their own implementations, it might be better to do this backend specific instead of general (depends on what optimizations they have). What do you say @nathanielsimard ?

Kelvinyu1117:
I think leveraging the implementation of the backend is a good idea!

antimora:
@dcvz can this issue be closed as the main PR is merged?

Kelvinyu1117:
I'm working on the backend implementation, we can use this issue or open another issue as a tracker.

PR #998, opened 2023-11-24
feature(tensor): Add chunk op

dcvz:
### Checklist

### Related Issues/PRs

Solves #970 (at least at a base implementation). I believe @Kelvinyu1117 might be working on a backend optimized version. I needed to implement this for my project, so thought I'd PR it.

### Changes

Adds a base tensor implementation for chunking.

### Testing

Tests have been made for the different cases for chunk.

### DIFF
--- a/burn-tensor/src/tests/ops/chunk.rs
+++ b/burn-tensor/src/tests/ops/chunk.rs
@@ -1,0 +1,76 @@
+#[burn_tensor_testgen::testgen(chunk)]
+mod tests {
+    use super::*;
+    use burn_tensor::{Data, Int, Shape, Tensor};
+
+    fn test_chunk_evenly_divisible() {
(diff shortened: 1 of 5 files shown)


Kelvinyu1117:
yup, probably I can do it on top of this PR.
I'm still looking at how can we fall back to our own implementation if the backend doesn't implement.

burn-tensor/src/tests/ops/chunk.rs
```diff
@@ -0,0 +1,79 @@
...
+            Data::from([8, 9]),
+            Data::from([10]),
+        ];
+
+        for (index, tensor) in tensors.iter().enumerate() {
+            assert_eq!(tensor.to_data(), expected[index]);
+        }
+    }
+
+    fn test_chunk_not_divisible() {
```
nathanielsimard, line 44: You are missing the `#[test]` attribute.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #1001, opened 2023-11-26
slice assign in candle

### LINKED ISSUE #976: Candle: support for slice assign
The operation `slice_assign` is very important for us in backward and we still don't have it with the Candle backend. 
There is an [issue open on the Candle GitHub](https://github.com/huggingface/candle/issues/1351). If they implement it then we can probably solve this issue easily. Otherwise it is not trivial, i have spent a few hours trying to get around this, unsuccessfully.

louisfd:
Candle finally made slice_assign, so I've updated the candle backend. 
I put this PR as draft because slice_assign is only on Candle main, not an official version, and we won't be able to publish if we rely on github rather than crates.io. 
So this is pending until candle releases its next version: https://crates.io/crates/candle-core (v0.3.1 at the moment)

fix #976

### DIFF
--- a/burn-candle/Cargo.toml
+++ b/burn-candle/Cargo.toml
@@ -18,7 +18,7 @@ derive-new = { workspace = true }
 burn-tensor = { path = "../burn-tensor", version = "0.11.0", default-features = false }
 half = { workspace = true }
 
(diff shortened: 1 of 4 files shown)


louisfd:
Will make a new one

Issue #986, opened 2023-11-21
Add `triu` and `tril` Tensor Method


dcvz:
### Feature description

Add a `triu` and `tril` methods to return the upper/lower of a matrix (2-D tensor) or batch of matrices, while setting the rest to 0.

### Feature motivation

Operation supported by other ML frameworks

antimora:
Added `tril` as well

dcvz:
@antimora can you assign me here?

antimora:
> @antimora can you assign me here?

Done!

dcvz:
PR for this now submitted #1002

antimora:
Closing as the PR is merged now.

PR #1002, opened 2023-11-27
feature(tensor): triu, tril & tri_compare

dcvz:
### Checklist

### Related Issues/PRs

Solves #986 

### Changes

Adds `triu` and `tril` tensor ops. Additionally adds a `tri_compare` where you can define the method to be used for comparison. 

### Testing

Tests have been added.


dcvz:
The only thing that stops this from being able to be be also applied to `Bool` tensors is that we don't have mask_fill there, a logical not operator would also work.

dcvz:
@nathanielsimard @louisfd this makes public a `tri_compare` method which may or may not be useful. Let me know if you think it should be just internal instead.

burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -462,6 +462,86 @@ where
...
+    /// the other elements of the result tensor out are set to 0.
+    ///
+    /// # Example
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::{Int, Tensor};
+    ///
+    /// fn example<B: Backend>() {
+    ///    let x = Tensor::<B, 2, Int>::from_ints([[1, 2, 3], [4, 5, 6], [7, 8, 9]]);
+    ///   let y = x.triu(1);
```
nathanielsimard, line 522: Small indent problem.

nathanielsimard (changes requested):
Thanks a lot for the PR, it's pretty well done! I think we should add a bit more detail to the examples, like printing the tensor and showing what the actual tensor would look like depending on the diagonal input.

Also, I think we should mark `tri_compare` as `pub(crate)` for now, since I'm not sure we want to support that API just yet.

dcvz:
@nathanielsimard thanks for the feedback! I believe everything is sorted now

burn-book/src/building-blocks/tensor.md
```diff
@@ -114,6 +114,8 @@ Those operations are available for numeric tensor kinds: `Float` and `Int`.
 | `tensor.clamp_min(min)`                          | `torch.clamp(tensor, min=min)`                 |
 | `tensor.clamp_max(max)`                          | `torch.clamp(tensor, max=max)`                 |
 | `tensor.abs()`                                   | `torch.abs(tensor)`                            |
+| `tensor.triu(diagonal)                           | `torch.triu(tensor, diagonal)`                 |
+| `tensor.tril(diagonal)                           | `torch.triu(tensor, diagonal)`                 |
```
antimora, line 118: `torch.triu` -> `torch.tril` for the second row.

antimora (changes requested):
Looks great. I found one typo in the book text. Also can we add a checkbox for the second column of the ONNX supported ops? We just want to indicate that triu op has been added in Burn (but not ONNX op yet).



https://github.com/Tracel-AI/burn/blob/main/burn-import/SUPPORTED-ONNX-OPS.md?plain=1#L194

burn-book/src/building-blocks/tensor.md
dcvz: nice, catch! updated

antimora (approved):
LGTM

Thanks for the addition!

nathanielsimard (approved):

PR #1003, opened 2023-11-27
Ci/Update dependabot

Luni-4:
### Checklist

### Related Issues/PRs

### Changes

This PR updates dependabot:

- Dependenciens checks run weekly
- Some dependencies have been grouped in specific PRs depending on their contexts

### Testing

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,26 @@
 version: 2
 
 updates:
-  - package-ecosystem: cargo
-    directory: /
+  - package-ecosystem: "github-actions"
+    directory: "/"
     schedule:
-      interval: daily
+      interval: "weekly"
 
-  - package-ecosystem: github-actions
-    directory: /
+  - package-ecosystem: "cargo"
+    directory: "/"
     schedule:
-      interval: daily
+      interval: "weekly"
+    groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,12 +1,26 @@
...
+        - "pollster"
+        - "text_placeholder"
+        - "wgpu"
+      wasm-bindgen-dependencies:
+        patterns:
+        - "js-sys"
+        - "wasm-bindgen"
+        - "wasm-bindgen-futures"
+        - "wasm-logger"
+        - "wasm-timer"
```
nathanielsimard, line 26: I would add a comment on top of each group to describe why they are required.

nathanielsimard (changes requested):

.github/dependabot.yml
Luni-4: Yep, I agree. I haven't considered other groups because I do not know which could be. If you someone has other dependencies groups in mind, I can add them to this PR, otherwise we can add further ones later

Luni-4: Ok, added comments. We can add more groups later, so we can merge this PR in my opinion

nathanielsimard (approved):

PR #1006, opened 2023-11-27
fix/docs/chunk

louisfd:
@dcvz 
I was a bit late to review your chunk operation. It's very nice but I found something was wrong with the doc. I made a test that explains why I'm changing it. Is this the behaviour you want?


louisfd:
@dcvz now it should have the right behaviour

dcvz (approved):
LGTM! 👍 Thanks for fixing it! Separating the happy path and not also makes it easier to understand than calculating current chunk size in the same loop.

nathanielsimard (approved):

PR #1009, opened 2023-11-28
chore(candle): Allow enabling accelerate

### LINKED ISSUE #968: Enable accelerate feature flag in candle for Macs
Enable accelerate feature flag in candle for Macs.

from Laurent Mazaré: 

> to mention that you may want to enable the accelerate feature on macos (or provide a way for your users to do so), on macos native builds it makes quite a large differences on matmuls.

dcvz:
### Checklist

### Related Issues/PRs

Fixes #968 

### Changes

Enables using accelerate for macOS

### Testing

Tests have been run


burn-autodiff/src/tests/transpose.rs
```diff
@@ -23,6 +23,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(feature = "accelerate"))]
     fn should_diff_swap_dims() {
         let tensor_1 =
```
nathanielsimard, line 28: I think we should fix this before merging.

nathanielsimard (commented):
@louisfd  can you check why the test fails on accelerate?

dcvz:
@louisfd @nathanielsimard I fixed the failing test on `accelerate`. Error previously was:
```bash
---- tests::ad_transpose::tests::should_diff_swap_dims stdout ----
thread 'tests::ad_transpose::tests::should_diff_swap_dims' panicked at burn-candle/src/ops/tensor.rs:141:68:
called `Result::unwrap()` on an `Err` value: MatMulUnexpectedStriding(MatMulUnexpectedStriding { lhs_l: Layout { shape: [2, 2, 2], stride: [4, 2, 1], start_offset: 0 }, rhs_l: Layout { shape: [2, 2, 2], stride: [1, 2, 4], start_offset: 0 }, bmnk: (2, 2, 2, 2), msg: "non-contiguous rhs" })
```

I resolve this by making sure the `rhs` is contiguous. This is a noop if it was already contiguous so I think it should be ok, but take a look.

xtask/src/runchecks.rs
```diff
@@ -249,6 +249,12 @@ fn burn_dataset_features_std() {
     cargo_doc(["-p", "burn-dataset", "--all-features"].into());
 }
 
+// Test burn-candle with accelerate (macOS)
+#[cfg(target_os = "macos")]
```
Luni-4, line 253: Just one question: Can not run on `Linux`?

dcvz: Nope. This is specifically leveraging the [Accelerate](https://developer.apple.com/documentation/accelerate) framework on macOS.

Luni-4: Thank you! Can you add a comment with a link in `runchecks.rs`? In this way we do not forget why there is that `cfg`.

antimora (approved):
LGTM. 

I have tested on M3 Max and it's working.

PR #1011, opened 2023-11-28
Use correct name of wgpu backend

mks-h:
### Checklist

### Related Issues/PRs

### Changes

Wgpu backend is named 'WgpuBackend', not 'Wgpu'.

### Testing

I'm following the guide, and the original code didn't work, while this one does.


dcvz:
The backend names were refactored in #904 but a version with that hasn't been published yet. On 0.10.0 it was still called `WgpuBackend` as you've discovered. However starting from the next version it will be `Wgpu`.

PR #1013, opened 2023-11-28
Add Cos, Exp, Gelu, Log unary ops onnx

antimora:
### Checklist

### Changes

Add Cos, Exp, Gelu, Log unary ops onnx.

### Testing

Added ONNX file tests.


burn-import/onnx-tests/tests/onnx_tests.rs
```diff
@@ -641,4 +646,69 @@ mod tests {
...
+    #[test]
+    fn cos() {
+        let model: cos::Model<Backend> = cos::Model::new();
+
+        let input = Tensor::<Backend, 4>::from_floats([[[[1.0, 4.0, 9.0, 25.0]]]]);
+
+        let output = model.forward(input);
+        let expected = Data::from([[[[0.5403, -0.6536, -0.9111, 0.9912]]]]);
+
+        output.to_data().assert_approx_eq(&expected, 4);
```
Luni-4, line 659: Not related to this PR: Is it possible to move `assert_approx_eq` into `burn-common` and define our own approximation comparator there? In this way we can use that in every `burn` crate and change its implementation in a single point.

burn-import/src/burn/node/unary.rs
```diff
@@ -400,4 +445,118 @@ mod tests {
...
+                Type::Tensor(TensorType::new_float("tensor2", 4)),
+            ),
+            quote! {
+                pub fn forward(&self, tensor1: Tensor<B, 4>) -> Tensor<B, 4> {
+                    let tensor2 = tensor1.cos();
+
+                    tensor2
+                }
+            },
+            vec!["tensor1".to_string()],
```
Luni-4, line 463: Not related to this PR: Can we return a &'static str instead of `String`?

Luni-4 (approved):
Fine for me, thanks! Just a series of nits not related to this PR but they might improve the `burn` code a bit

burn-import/onnx-tests/tests/onnx_tests.rs
nathanielsimard: For now it's a method used with the `Data` struct, do you have an idea how you want to generalise the method?

nathanielsimard (approved):
LGTM !

nathanielsimard:
@antimora There are failing tests, but I think after they are fixed we can merge.

PR #1014, opened 2023-11-29
Ci/Schedule dependencies checks

### LINKED ISSUE #975: Implement scheduled action for scanning unused dependencies
**Description**:

This issue focuses on setting up a scheduled action to regularly scan for unused dependencies (`udeps`) in our project. The aim is to ensure efficient dependency management without hindering the continuous integration process.

**Background**:

- Our current setup with `udeps` has been causing build failures due to unresolved false positives in the CI pipeline.
- The proposal to disable `udeps` temporarily raised concerns about the long-term management of dependencies.
- An alternative approach has been suggested to handle this through scheduled scans, similar to our security scans.

**Proposal**:

Instead of disabling `udeps` or moving its checks to the `publish.yml` script, we propose setting up a scheduled action. This action will regularly scan the repository for unused dependencies. This approach aims to maintain a clean dependency list without impacting the regular development flow.

**Advantages**:
- **Non-Blocking Development**: Developers can continue working without being blocked by `udeps` failures in CI.
- **Regular Maintenance**: Regular scans ensure that dependency management is an ongoing process rather than a last-minute rush before releases.
- **Early Detection**: Helps in identifying and resolving dependency issues early in the development cycle.

**Implementation Considerations**:

1. **Schedule Frequency**:
   - Determine how frequently the scan should run (e.g., weekly, bi-weekly).
   
2. **Scope of Scans**:
   - Decide if scans should cover the entire repository or specific modules.

3. **Handling Scan Results**:
   - Establish a protocol for addressing issues found during scans, including assigning responsibilities for resolving them.

4. **Integration with CI/CD**:
   - Ensure the scheduled action is seamlessly integrated with our existing CI/CD pipeline.

**Action Items**:
- [ ] Determine the optimal frequency for the scheduled `udeps` scans.
- [ ] Define the scope and parameters of the `udeps` scans.
- [ ] Develop a workflow for handling and addressing issues identified in the scans.
- [ ] Integrate the scheduled scan action into the existing CI/CD system.

**Additional References**:
- Related Discussions: [Disable udeps checking till all unused deps are resolved #969](https://github.com/Tracel-AI/Tracel-AI/pull/969), [Fix image classifier build failure by upgrading versions #967](https://github.com/Tracel-AI/Tracel-AI/pull/967)
- Example of False Positive Handling: `burn/burn-dataset/Cargo.toml` (lines 55-56).

Luni-4:
### Checklist

### Changes

This PR adds a scheduled event for dependencies checks and also improve the `audit` action through the addition of an `audit.toml` file.

This PR fixes #975, below the list of items:
- [x] Determine the optimal frequency for the scheduled udeps scans 
    - **weekly**: in this way we can immediately find unused deps, but at the same time we do not block development
- [x] Define the scope and parameters of the udeps scans.
    - **codebase**: Unused deps should be detected in each crate, that is why the `--all-targets` feature is enabled. If some false-positives are found, they can be ignored setting up the respective field in the `Cargo.toml` file 
- [x] Develop a workflow for handling and addressing issues identified in the scans.
    - **CI run logs**: We can use logs to identify the unused deps without creating new issues
    - **Issues for Audit vulnerabilities**:  Audit finds vulnerabilities in dependencies, we can open issues for dependencies vulnerabilities. It already does that on `main` branch as stated [here](https://github.com/actions-rust-lang/audit#inputs)
- [x] Integrate the scheduled scan action into the existing CI/CD system.
    - **Done** in this PR

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -3,6 +3,9 @@ name: dependencies
 on:
   schedule:
     - cron: '0 21 * * TUE' # Run every Tuesday at 21:00 (UTC)
+  push:
+    tags:
+      - 'v*.*.*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -1,17 +1,8 @@
 name: dependencies
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - '**/Cargo.lock'
-      - '**/Cargo.toml'
-  pull_request:
-    types: [opened, synchronize]
-    paths:
-      - '**/Cargo.lock'
-      - '**/Cargo.toml'
+  schedule:
+    - cron: '0 21 * * TUE' # Run every Tuesday at 21:00 (UTC)
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/dependencies.yml
```diff
@@ -1,17 +1,11 @@
...
-    paths:
-      - '**/Cargo.lock'
-      - '**/Cargo.toml'
-  pull_request:
-    types: [opened, synchronize]
-    paths:
-      - '**/Cargo.lock'
-      - '**/Cargo.toml'
+    tags:
+      - 'v*.*.*'
```
nathanielsimard, line 8: Not sure how this will impact the dependency checks? Is it going to run everytime we push a new release? Ultimately, it would be cool to check de dependencies _before_ doing a release.

audit.toml
```diff
@@ -0,0 +1,27 @@
+# Example audit config file
```
nathanielsimard, line 1: I don't think it's an example anymore 🙃

nathanielsimard (changes requested):

.github/workflows/dependencies.yml
Luni-4: Exactly, it will **also** run when we push a new release. Thus, it will run every Tuesday at 21:00 (UTC) **and** when we push a new release.

nathanielsimard (approved):

PR #938, opened 2023-11-09
WIP: burn-train in the browser

AlexErrant:
### Checklist

- [ ] Confirm that `run-checks` script has been executed.

This has never worked on my machine. Gonna lean on CICD for this.

### Related Issues/PRs

Closes https://github.com/burn-rs/burn/issues/921

### Changes

Here's part 1, where I replace a `spawn` in `EventStoreClient` with a webworker implementation pretty much copied straight from [here](https://www.tweag.io/blog/2022-11-24-wasm-threads-and-messages/). The total files changed is pretty large, but most of it is just a new example project I made called `examples/train-web`. That project uses PNPM and Vite, which stylistically diverges from the minimalistic `examples/mnist-inference-web`. I do so to more accurately reflect the usage in a real-world project that uses Typescript and NPM packages.

To make this PR easier to review, I'm following the atomic commit style and will refrain from force-pushing this branch.

### Testing

I have a branch [here](https://github.com/AlexErrant/burn/compare/train-browser..spawnWebWorkerTest) that demos `sender.send` working. Console output as follows:

```
train.js:325 INFO examples/train-web/train/src/lib.rs:17 
Hello from Rust
train.js:325 INFO burn-train/src/metric/store/client.rs:110 
Got msg: OnEventTrain(EndEpoch(1337))
train.js:325 INFO burn-train/src/metric/store/client.rs:110 
Got msg: OnEventValid(EndEpoch(314159))
train.js:325 INFO burn-train/src/metric/store/client.rs:110 
Got msg: End
```

### Next step

Either getting data loading of mnist working or exploring what it'll take to write to the filesystem.

PR #1017, opened 2023-11-29
cargo +nightly fmt

AlexErrant:
### Checklist

    nnnnope still broken on my box

### Related Issues/PRs

[Instigating comment.](https://github.com/Tracel-AI/burn/pull/692#pullrequestreview-1746838522) 

### Changes

[Format silently breaks on lines longer than the default, 100.](https://github.com/rust-lang/rustfmt/issues/3863#issuecomment-700435806) I enable two nightly features, [`error_on_line_overflow`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#error_on_line_overflow) and [`format_strings`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#format_strings). Despite being nightly, we can run `cargo +nightly fmt` [while remaining on the stable toolchain](https://www.reddit.com/r/rust/comments/lppk9l/why_most_rustfmt_options_are_still_unstable/gohwbtr/?context=1). 

* `format_strings` is useful for breaking up long lines as mentioned [here](https://github.com/rust-lang/rustfmt/issues/3863#issuecomment-678632593).
* `error_on_line_overflow` is useful for making long lines explicitly fail. You may still have to manually format and annotate with `#[rustfmt::skip]`, otherwise you get errors like:

```
error[internal]: line formatted, but exceeded maximum width (maximum: 100 (see `max_width` option), found: 105)
    --> /home/alex/burn/burn-autodiff/src/ops/tensor.rs:1088:1088:101
     |
1088 |         impl<B: Backend, const D: usize> Backward<B::FullPrecisionBackend, D, 1> for ToFullPrecision<B> {
     |                                                                                                     ^^^^^
     |
```

Grep this PR for `#[rustfmt::skip]` for example usage.

Interestingly setting `max_width = 200`, formatting, then setting `max_width = 100` and formatting seemed to shake loose [some changes](https://github.com/Tracel-AI/burn/commit/d864317f90932b0840670d9c41ae25f652e5ceac).

### Testing

I made some changes to `runchecks`, lessee what CI thinks.


Luni-4:
Personally I would not add `nighlty` formatting to CI because, along the time, it can cause more breaks than anything else. A solution could be to pin a specific `nightly` version, but a better solution would be to fix format problems using the `stable` version. But I'm not the one who should decide to the fate of this PR, just an advice

AlexErrant:
> a better solution would be to fix format problems using the stable version

Not possible, since `error_on_line_overflow` and `format_strings` are both not stable. Good point on pinning nightly though, so I pinned it to the same version as [here](https://github.com/Tracel-AI/burn/pull/938/files#diff-c1f462aaffc7c3f664c2ec217e79e51015006e70c6cd6f40b917be47873e6c6bR2). Should I use a newer version? IDK.

Note that the formatting changes done here are more or less the same as formatting without `error_on_line_overflow` or `format_strings`. Nightly will just make it so formatting will holler at you when you violate said rules.

louisfd:
@nathanielsimard would depending on nightly affect the release to crates.io?

nathanielsimard:
@AlexErrant Since most of the development is done on stable, I feel like it's better to stick with Rust fmt stable instead of nightly. Once of the main reason to use a formatter is to eliminate conflicts generated by using different formatters by different people. However, I'm not against accepting a pull request that reformats the code using fmt nightly and then re-reformats it using the stable version. This could improve the formatting without creating conflicts with the stable formatter—just improving the code style where the stable branch falls short.

AlexErrant:
Whoops I went a little too delete-happy in `runchecks.rs`.

nathanielsimard (approved):

PR #1020, opened 2023-11-30
Feat/fusion/cache

### LINKED ISSUE #935: Burn-Fusion: Caching relative graphs

nathanielsimard:
First part, relative and global graph conversion and graph optimization cache!

Another level of caching is possible in the wgpu crate, but it will be done in a following PR.

### Related Issues/PRs

Fix #935


burn-fusion/src/graph/execution.rs
```diff
@@ -1,59 +1,185 @@
...
-            if graph.is_empty() {
-                // No more ops to fuse.
+            if let ExecutionMode::NewOps = mode {
                 break;
             }
         }
     }
+
+    fn build(&mut self, graph: &mut Graph<B>, mode: ExecutionMode) -> BuildAction<'_, B> {
+        let offset = match mode {
```
louisfd, line 86: Very obscure, needs a comment

burn-fusion/src/graph/base.rs
```diff
@@ -1,100 +1,95 @@
...
+            global: Vec::new(),
+            relative: Vec::new(),
+            converter: RelativeGraphConverter::default(),
             ops: Vec::new(),
         }
     }
-    pub(crate) fn add(&mut self, description: Arc<TensorOpsDescription>, ops: Box<dyn Ops<B>>) {
-        self.operations.push(description);
+
+    pub(crate) fn split_graph(&self) -> (&[TensorOpsDescription], Option<&TensorOpsDescription>) {
```
louisfd, line 26: Should be named split_relative_graph

burn-fusion/src/graph/path/base.rs
```diff
@@ -0,0 +1,535 @@
...
+        } else {
+            CacheResult::OnPath
+        }
+    }
+
+    /// Signal the completion of a graph path that reached a new optimization.
+    ///
+    /// # Notes
+    ///
+    /// The optimization factory will only be call if the optimization is on a new graph.
```
louisfd, line 129: called

burn-fusion/src/graph/execution.rs
```diff
@@ -1,59 +1,185 @@
...
+        // Can only be lazy when not sync.
+        if let ExecutionMode::NewOps = mode {
+            if still_optimizing(&self.optimizations) {
+                return BuildAction::ContinueBuilding;
+            }
+        }
+
+        match find_best_optimization_index(&self.optimizations) {
+            Some(index) => {
+                let (relative, next_ops) = match mode {
```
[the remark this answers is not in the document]
nathanielsimard: The execution mode doesn't really belong to the graph, but I'll create a function in the graph execution that does that.

PR #1021, opened 2023-11-30
Update getting-started.md

shanmo:
This change solves the error `no Wgpu in backend` when running the `Getting Started` example on Mac OS with M1 

### Checklist

### Related Issues/PRs

No related Issues/PRs

### Changes

Change is made in the Book's markdown file

### Testing

Tested on my laptop with Mac OS M1 and there is no error

### DIFF
--- a/burn-book/src/getting-started.md
+++ b/burn-book/src/getting-started.md
@@ -29,10 +29,10 @@ cd my_burn_app
 then add the dependency with
 
 ```console
-cargo add burn --features wgpu
+cargo add burn burn-wgpu
 ```
 
-add `burn-wgpu = "0.10.0"` in Cargo.toml, and compile it by executing
+and compile it by executing
 
 ```console
 cargo build


dcvz:
Hi @shanmo -- this has been a source of confusion (see: https://github.com/Tracel-AI/burn/pull/1011). Essentially the backends have been renamed but those changes have not made it a new release yet. However the docs already reflect that change. The next version should be out soon, until then as you've discovered you should be using `WgpuBackend`.

burn-book/src/getting-started.md
```diff
@@ -32,7 +32,7 @@ then add the dependency with
 cargo add burn --features wgpu
 ```
 
-and compile it by executing
+add `burn-wgpu = "0.10.0"` in Cargo.toml, and compile it by executing
```
louisfd, line 35: If we change line 32 for 
`cargo add burn burn-wgpu` it will add both dependancies at the same time and ensure the latest version is fetched

louisfd (changes requested):
Thanks a lot,  
Indeed there had been some breaking changes and this had not been updated!
I think one thing could be written differently, see my comment

PR #1023, opened 2023-11-30
Progressive update -> Update Wgpu to WgpuBackend

projectfinalaudio:
## Proposed changes
- [ ] Updated Wgpu(type) to WgpuBackend.


louisfd (commented):
Thanks for pointing this out, 
Wgpu will be the new name for WgpuBackend from the next release onward (coming very soon). 
The book is up to date with the main branch, not the last release, so it can indeed get confusing when there are breaking changes like this one. However once the next version is released it will be all right. 
Closing your PR as we are preparing the next version now

PR #1024, opened 2023-11-30
chore(ci): Test more backends

dcvz:
> [!WARNING]
> This is still a WIP

### Checklist

### Related Issues/PRs

### Changes

### Testing


antimora (approved):
LGTM!

xtask/src/main.rs
```diff
@@ -2,6 +2,7 @@ use clap::{Parser, Subcommand};
 
 mod publish;
 mod runchecks;
+pub(crate) mod utils;
```
Luni-4, line 5: Can we write `mod utils` and import things as `use crate::crate_name::function`?

xtask/src/utils.rs
```diff
@@ -0,0 +1,99 @@
...
+                _ => None,
+            }
+        })
+        .collect();
+
+    workspaces
+}
+
+/// Start log group
+pub(crate) fn start_log_group(title: String) {
```
Luni-4, line 70: Instead of `String` we can use `Cow` in this way we can write:
`start_log_group("Something")` and `start_log_group(format!("something"))`. I guess we just need to write a generic that implements `Into<Cow>`

xtask/src/utils.rs
```diff
@@ -0,0 +1,99 @@
...
+
+pub(crate) enum WorkspaceMemberType {
+    Crate,
+    Example,
+}
+
+#[derive(Debug)]
+pub(crate) struct WorkspaceMember {
+    pub name: String,
+    pub path: String,
```
Luni-4, line 12: We can use `pub(crate)` or remove the visibility attribute if they are used in the same file

Luni-4 (changes requested):
Thanks a lot! 

I propose to change `println!` with `debug!` or `info!` macros, which can be enabled through filters.

Here an example of how to do that in `main`

```rust
   // Enable filter to log the information contained in the lib.
    let filter_layer = EnvFilter::try_from_default_env()
        .or_else(|_| {
            // We can add this through clap
            if opts.verbose {
                EnvFilter::try_new("debug")
            } else {
                EnvFilter::try_new("info")
            }
        })
        .unwrap();

    // Run tracer.
    tracing_subscriber::fmt()
        .without_time()
        .with_env_filter(filter_layer)
        .with_writer(std::io::stderr)
        .init();
```

and in `Cargo.toml`

```
tracing = "^0.1"
tracing-subscriber = { version = "^0.3", features = ["env-filter"] }
```

Luni-4:
@dcvz 

This new logging seems a bit complicated, I was proposing to replace `println!` with `info!` macro from `tracing`. Adding this new macros and series of `match`es complicates the code in my opinion

dcvz:
@Luni-4 if we want a logger, then I'd say let's do it right. This make sure it outputs correctly also to the Github Action logs, meaning we'll be able to see highlighted warnings and errors easier. It's also just about the same amount of boilerplate for the changes you were asking for logger, tracer and using Cow.

IMO - I think this is fine.. but I guess its up to you all to decide.

Luni-4:
I'm not against this new logger, just saying that is more complete, perhaps this is the right term, of what I have thought, but I'm fine with it

Luni-4 (approved):

dcvz:
@louisfd @nathanielsimard this one is ready to go!

xtask/src/logging.rs
```diff
@@ -0,0 +1,67 @@
...
+#[macro_export]
+macro_rules! group {
+    // group!()
+    ($($arg:tt)*) => {
+        let title = format!($($arg)*);
+        if std::env::var("CI").is_ok() {
+            log!(log::Level::Info, "::group::{}", title)
+        } else {
+            log!(log::Level::Info, "{}", title)
+        }
```
nathanielsimard, line 54: Why use a different rendering depending of the CI? can `::group::{}` works everywhere? Same question for all other occurences.

PR #1025, opened 2023-11-30
Refactor feature flags

nathanielsimard:
I refactored the burn feature flags to make them more automatic and easier to understand, as well as more flexible. Since I didn't know we could use the syntax `optional-dependency?/feature-flag`, we had to specialize each sub-dependency feature flag with all its variants (no-std, minimal, blas, accelerate, cuda, etc.). Now we can have global feature flags that will activate the proper feature flags of optional dependencies if they are activated.

I used the same logic to activate the default feature flags of sub-dependencies. So, when we deactivate the default features of burn, we will also deactivate the default features of added features like `ndarray` for no-std environments, which is probably the intended behavior.


nathanielsimard:
@dae It removes the `train-minimal` feature flag as well. If you deactivate the default features, the `tui` and `metrics` features aren't activated.

dcvz (approved):
I also didn't know about `optional?/dep`. This makes dep setup much clear! LGTM 👍

antimora (approved):
LGTM!

However! We need to double check if all examples run. Last time we refactored the feature flags, we broke many examples. So lets test examples with different flags well.

louisfd:
> LGTM!
> 
> However! We need to double check if all examples run. Last time we refactored the feature flags, we broke many examples. So lets test examples with different flags well.

I tried a few combinations, although not exhaustively, and did not find any problem. 
However, this makes me realize we really need to have clearer doc about all the available features somewhere.

dae:
I've refactored our code to use the new flags, and I've noticed one unfortunate downside, though there's probably nothing that can be done about it. With the new approach, the optional dependencies still get listed in Cargo.lock even when not in use (eg tch, wgpu, candle are all listed even if only using ndarray), and this also extends to the output of cargo-license. They don't appear to be making it into the build output thankfully, and cargo tree does not show them, but it makes it a bit harder to track which dependencies are being used now.

antimora:
@dae have you tried udeps cargo tool?

@Luni-4 has added dependency check in CI, that includes license check after the lgpl ordeal.

Btw, SQLite serde is Apache2 and MIT compliant now (latest release).

nathanielsimard:
@dae I think we may open an issue on Cargo https://github.com/rust-lang/cargo. I think is make sense that unused optional dependencies are not included in `Cargo.lock`.

PR #1026, opened 2023-11-30
Fix double broadcast with tch

nathanielsimard:
@dcvz Can you validate it fixes the problem you have with diffuser-rs?

### DIFF
--- a/burn-tch/src/ops/int_tensor.rs
+++ b/burn-tch/src/ops/int_tensor.rs
@@ -190,17 +190,30 @@ impl<E: TchElement> IntTensorOps<Self> for LibTorch<E> {
         lhs: TchTensor<i64, D>,
         rhs: TchTensor<i64, D>,
     ) -> TchTensor<i64, D> {
-        TchOps::div(lhs, rhs)
+        let copy = false;
+        let non_blocking = true;
+        let lhs: TchTensor<f64, D> =
+            TchTensor::new(lhs.tensor.to_dtype(tch::Kind::Float, non_blocking, copy));
+        let rhs: TchTensor<f64, D> =
+            TchTensor::new(rhs.tensor.to_dtype(tch::Kind::Float, non_blocking, copy));
+
+        let out = TchOps::div(lhs, rhs);
+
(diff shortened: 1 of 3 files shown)


dcvz:
@nathanielsimard I can confirm that the tests that were failing in my project, pass now 🚀

burn-tch/src/ops/int_tensor.rs
```diff
@@ -190,17 +190,30 @@ impl<E: TchElement> IntTensorOps<Self> for LibTorch<E> {
         lhs: TchTensor<i64, D>,
         rhs: TchTensor<i64, D>,
     ) -> TchTensor<i64, D> {
-        TchOps::div(lhs, rhs)
+        let copy = false;
```
dcvz, line 193: What does this actually do? seems like you're just forcing the float div to be used?
I took out all the changes in this file, and the failing tests on my repo and here still pass

burn-tensor/src/tests/ops/mul.rs
```diff
@@ -30,6 +30,17 @@ mod tests {
         assert_eq!(data_expected, data_actual);
     }
 
+    #[test]
+    fn test_mul_broadcast_2_dims() {
```
dcvz, line 34: do we care about adding a test for `Int`?

louisfd: I don't think it's necessary, even if tch supports it, in Burn we do not expose int matmul in our API

burn-tch/src/ops/int_tensor.rs
nathanielsimard: Yeah somehow I had to update this for the CI, I cast the i64 tensor to a f64 tensor without copying the data since it seems there are problems with integer division with tch.

burn-tensor/src/tests/ops/mul.rs
nathanielsimard: This is a normal mul that is available to the int tensor, but theorically this test validates the broadcasting logic that is present for all binary operations.

louisfd (approved):
LGTM

PR #1039, opened 2023-12-01
chore(infra): Share some properties across workspace

dcvz:
I noticed version bumping touches many places, this is an attempt to alleviate that a bit.


nathanielsimard:
@dcvz This is a great idea! I guess you should update with `main` and use `0.12.0` instead of `0.11.0`, the flaky tests are supposed to be gone.

dcvz:
@nathanielsimard updated!

nathanielsimard (approved):

PR #1041, opened 2023-12-02
Allow raw tensors to be serialized/deserialized directly with serde

jmacglashan:
### Checklist

I don't think the book needs updating for this.

### Related Issues/PRs

None.

### Changes

It is often useful in research to just dump some tensors to disk and to load them back up later. This can be for debugging purposes or otherwise. This PR implements serde traits for the Tensor struct so that this can be performed with any normal serde type and the tensors can also be put into structs with serde and serialized if some structure around them desired.

To do this, the impls go through the existing `DataSerialize` type so that this PR does not introduce any new concepts, only the impls.

Because `to_data` is an async function for wasm (without wasm-sync), and because serde is not for async, I gated this implementation to only non-wasm targets. (Deserialize interestingly seems to compile file without gating. So perhaps this would allow async wasm to deserialze serde written data, jut not serialize).

From a discussion on discord, it was suggested that when tensor quantization support is added, the `DataSerialize` may need to get an extra `Vec<u8>` field for any additional meta data. Since this feature is not finalized I decided not to add it here. Future compatibility should still be possible by adding serde default field values if it is ever added.

### Testing

I tested this feature by making a new crate and adding the following code, which worked correctly.
```rust
use burn::backend::NdArray;
use burn::tensor::Tensor;

fn main() {
    let tensor: Tensor<NdArray, 1> = Tensor::ones([1]);
    let encoded: Vec<u8> = bincode::serialize(&tensor).unwrap();
    let decoded: Tensor<NdArray, 1> = bincode::deserialize(&encoded).unwrap();
    println!("{}", decoded);
    assert_eq!(tensor.into_data(), decoded.into_data());
}
```

It might be nice to make this a unit test in Burn, but it wasn't clear to me what the right pattern is. I was going to add a test in `burn_tensor` but doing so required `bincode`. I tried added that to the dev dependencies (using the workspace dep), but it still wasn't found in the test. So I may be missing something about how the registered tests in Burn work. (I'm not very practiced with them in rust in general outside some minor stuff)

If turning this into a unit test would be preferred, let me know the right way to do it and in which crate it would be preferred.


nathanielsimard (changes requested):
We could add a unit test in `burn-core` since this is where the dependencies for serialization are.

louisfd (approved):
LGTM, it's a very important feature to have. 
I agree with @nathanielsimard that we should add the test

jmacglashan:
Sounds good! I'll look into adding the test to `burn_core`

jmacglashan:
Okay, I think this is ready for re-review. I added the test to `burn_core` in `record_resilience.rs` because it that seemed the most closely related and it didn't seem worth adding a whole new test file just for this one test. But if you'd prefer it in its own file or some other location let me know.

I also used json instead of bincode because compiler seemed to think bincode in `burn_core` didn't have serde (which is surprising since the workspace has it's on. Not sure what's happening there), but the test with json works just as well.

Finally, I'll note that after merging master, `run-checks all` no longer passes from issues in `burn_fusion`, but that seems unrelated to this PR and I think is just the state of `burn_fusion` on master.

nathanielsimard (approved):

nathanielsimard:
@jmacglashan I think you should update your branch with main for the CI to work correctly.

jmacglashan:
Thanks @nathanielsimard I just merged main and pushed. Hopefully that goes through now!

I will note that the `./run-checks.sh all` still produces errors, but there's in the `burn-fusion` package which may just be due to active development on master and look unrelated to this PR. E.g., I see:
```
error: this expression creates a reference which is immediately dereferenced by the compiler
    --> burn-fusion/src/stream/ops.rs:1112:49
     |
1112 |                     vec![&desc.x, &desc.weight, &bias, &desc.out]
     |                                                 ^^^^^ help: change this to: `bias`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
     = note: `-D clippy::needless-borrow` implied by `-D warnings`

error: this expression creates a reference which is immediately dereferenced by the compiler
    --> burn-fusion/src/stream/ops.rs:1119:49
     |
1119 |                     vec![&desc.x, &desc.weight, &bias, &desc.out]
     |                                                 ^^^^^ help: change this to: `bias`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

error: this expression creates a reference which is immediately dereferenced by the compiler
    --> burn-fusion/src/stream/ops.rs:1126:49
     |
1126 |                     vec![&desc.x, &desc.weight, &bias, &desc.out]
     |                                                 ^^^^^ help: change this to: `bias`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

error: this expression creates a reference which is immediately dereferenced by the compiler
    --> burn-fusion/src/stream/ops.rs:1133:49
     |
1133 |                     vec![&desc.x, &desc.weight, &bias, &desc.out]
     |                                                 ^^^^^ help: change this to: `bias`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
```

PR #1042, opened 2023-12-02
Fix some issues in Burn book

shanmo:
### Checklist

### Related Issues/PRs

No related issues 

### Changes

Minor update in the example provided by Burn book, it's recommended to use expect-as-precondition style error messages in https://doc.rust-lang.org/std/error/index.html#common-message-styles


nathanielsimard (approved):
Yup, makes sense. We should be more consistent with our unwrap/expect messages. It really helps in explaining why it's okay to call expect here.

Issue #1034, opened 2023-12-01
Candle backend crashes web app in image-classification-web


antimora:
**Describe the bug**
Candle backend crashes web app in image-classification-web. 

**To Reproduce**
1. cd `examples/image-classification-web`
2. run `./build-for-web.sh`
3. run `./run-server.sh`
4. open the browser with the served link
5. change backed to Candle
6. select image. 
7. open debug console and you'll notice error

**Expected behavior**
The image is classified without a crash.

**Screenshots**
<img width="1496" alt="image" src="https://github.com/tracel-ai/burn/assets/939125/3665bd6b-721f-4424-bead-f56c07968c43">

dcvz:
The issue is `Candle does not support excluding pad count in pooling`

dcvz:
Looks like the average pool config that is being used in the squeezenet is using `.with_count_include_pad(false)` which is not supported in Candle.

antimora:
> The average pool config that is being used in the squeezenet is using `.with_count_include_pad(false)` which is not supported in Candle.

We can fix this by:

1. Modify Candle backend with Zero padding the input before average pool (this is what PyTorch did).
2. File a ticket with Candle team to enhance the avr pool.

nathanielsimard:
> > The average pool config that is being used in the squeezenet is using `.with_count_include_pad(false)` which is not supported in Candle.
> 
> We can fix this by:
> 
> 1. Modify Candle backend with Zero padding the input before average pool (this is what PyTorch did).
> 2. File a ticket with Candle team to enhance the avr pool.

@louisfd do you think it's possible?

antimora:
Outdated and not planned.

PR #1051, opened 2023-12-04
chore(web): better panic 

dcvz:
### Checklist

### Changes

Provides better panic support on web to help debug issues like #1034

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ wasm-bindgen = "0.2.88"
 wasm-bindgen-futures = "0.4.38"
 wasm-logger = "0.2.0"
 wasm-timer = "0.2.5"
+console_error_panic_hook = "0.1.7"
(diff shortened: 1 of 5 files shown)


antimora (approved):
Ahhh super! that's what we were missing!

AlexErrant:
You might consider using [start](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-rust-exports/start.html). I do so [here](https://github.com/tracel-ai/burn/pull/938/files#diff-481b91772c3f0745e1e4ea6c9ab49705889638cb8f56832207dec6ad13154538R13-R17).

dcvz:
> here

Done! I considered it, but we already had a start point so thought that'd be fine. Updated to be more proper.

dcvz:
@antimora @nathanielsimard should be good to go now!

PR #1054, opened 2023-12-07
Move transformer prenorm on the residual path

Philonoist:
### Checklist

### Changes

The transformer module with the prenorm setting, normalized before the residual path.
According to https://arxiv.org/pdf/2002.04745.pdf, it should be done after diverging to the residual path.
Instead of `x=norm(x); x += FF(x)`, we should do this: `x += FF(norm(x))`
This is the recommended way to do normalizaiton in transformers today AFAIU.

### Testing

I ran the tests.


louisfd:
Hi @Philonoist 
The changes look good to me, although I'm not an expert in transformers. It's good that you've documented the steps!
However the test_autoregressive_norm_first seem to fail in the CI. Is it expected that values change with your new implementation?

spapinistarkware:
Hey. It only fails on torch backed for some reason. It passes ndarray backend. Have any clue why that might happeb?

antimora:
> Hey. It only fails on torch backed for some reason. It passes ndarray backend. Have any clue why that might happeb?

It's possible the implementation of some op for Torch backend has a bug.

louisfd:
Indeed, see @nathanielsimard 's comment: https://github.com/tracel-ai/burn/issues/1056

nathanielsimard:
@spapinistarkware I [fixed](https://github.com/tracel-ai/burn/pull/1064) a bug in tch, make sure you are updated with the new changes.

Philonoist:
Merged.

louisfd (approved):

PR #1058, opened 2023-12-11
fix/ndarray: remove reversed axes check

### LINKED ISSUE #1053: Wrong result when call reshape() after Tensor::stack
Here is the code.
```Rust
type Backend = NdArray;

let tensor = Tensor::<Backend, 1, Int>::arange(1..25).reshape([4, 6]);
let zeros: Tensor<Backend, 2, Int> = Tensor::zeros([4, 6]);
let intersperse =
    Tensor::stack::<3>([tensor.clone(), zeros.clone()].to_vec(), 2).reshape([4, 12]);

println!("{}", intersperse);
```

The result is: 
```
Tensor {
  data:
[[1, 2, 3, 4, 5, 6, 0, 0, 0, 0, 0, 0],
 [7, 8, 9, 10, 11, 12, 0, 0, 0, 0, 0, 0],
 [13, 14, 15, 16, 17, 18, 0, 0, 0, 0, 0, 0],
 [19, 20, 21, 22, 23, 24, 0, 0, 0, 0, 0, 0]],
  shape:  [4, 12],
  device:  Cpu,
  backend:  "ndarray",
  kind:  "Int",
  dtype:  "i64",
}
```
But the expected result should be: 
```
[[1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0],
 [7, 0, 8, 0, 9, 0, 10, 0, 11, 0, 12, 0],
 [13, 0, 14, 0, 15, 0, 16, 0, 17, 0, 18, 0],
 [19, 0, 20, 0, 21, 0, 22, 0, 23, 0, 24, 0]],
```
I can get correct results on LibTorch Backend and Wgpu Backend.

AuruTus:
### Checklist

### Related Issues/PRs

fix #1053

### Changes

Remove reversed axes layout check to allow column major tensor to be reshaped into row major.

### Testing

Ran `cargo test` under `burn-ndarray` and `run-checks all`

__NOTE__:  Although the shape() bug can be fixed by this, there is still some weird logic in `ndarry::concatenate(...)` function to compute new strides for empty tensors where they prefer to producing column major layout. Maybe we should call `reshape(...)`  in `NdArrayOps::cat(...)` too and add some tests against the `cat` result.

### DIFF
--- a/burn-ndarray/src/ops/base.rs
+++ b/burn-ndarray/src/ops/base.rs
@@ -72,7 +72,9 @@ where
             .unwrap()
             .into_shared();
 
-        NdArrayTensor { array }
+        // Transform column-major layout into row-major (standard) layout. (fix #1053)
+        let array = NdArrayTensor { array };
+        Self::reshape(array.clone(), array.shape())
     }
 
     fn to_slice_args<const D1: usize, const D2: usize>(
--- a/burn-ndarray/src/ops/base.rs
+++ b/burn-ndarray/src/ops/base.rs
@@ -493,3 +493,42 @@ fn arg<E: NdArrayElement, const D: usize>(
         array: output.into_shared(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn should_generate_row_major_layout_for_cat() {
+        let expected_shape: &[usize] = &[4, 6, 2];
+        let expected_strides: &[isize] = &[12, 2, 1];
+        let expected_array = NdArrayTensor::from_data(Data::<i32, 3>::from([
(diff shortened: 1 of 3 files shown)


nathanielsimard:
@AuruTus I think it would be valuable to add the code snippet from this issue https://github.com/tracel-ai/burn/issues/1053 as a test before merging. The test should be generic over the backend in `burn-tensor/src/tests`

AuruTus:
> @AuruTus I think it would be valuable to add the code snippet from this issue #1053 as a test before merging. The test should be generic over the backend in `burn-tensor/src/tests`

Sure, I add it under `stack` 's unit tests. That reshaping after stacking test can cover all backends now.

nathanielsimard (approved):

PR #1059, opened 2023-12-12
Adding Shape to the operations supported for import

edmondop:
### Checklist

### Related Issues/PRs

### Changes

### Testing


burn-import/onnx-tests/tests/onnx_tests.rs
```diff
@@ -333,6 +334,15 @@ mod tests {
         assert_eq!(output.to_data(), expected);
     }
 
+
+    #[test]
+    fn shape() {
+        let model: shape::Model<Backend> = shape::Model::new();
+        let output = model.forward();
+        assert_eq!(output, (1,1,1,4));
+    }
```
nathanielsimard, line 343: Shapes in burn are not defined by tuples, but by the `Shape<D>` struct. I think to properly support shapes with `burn-import` we need to output the correct shape type, but also implement `Tensor<B, D, Int>` from shape, since shapes are sometimes used as int tensor by the onnx graph. @antimora

nathanielsimard (changes requested):

burn-import/onnx-tests/tests/onnx_tests.rs
edmondop: @antimora can you provide your view? I am not sure I understood the recommandation

antimora: The tuple result is erroneous because the ONNX returns four outputs because of constants. See my comment below. 

tensor.shape() would return `Shape<D>` already so we should use this type. This result can be re-fed to `reshape` operation, which is a good test. Please see my comment below.

burn-import/src/burn/node/unary.rs
```diff
@@ -138,6 +140,11 @@ impl UnaryNode {
         Self::new(input, output, UnaryNodeKind::Relu, Rc::new(function))
     }
 
+    pub(crate) fn shape(input: Type, output: Type) -> Self {
+        let function = move |input| quote! { #input.shape()};
+        Self::new(input, output, UnaryNodeKind::Shape, Rc::new(function))
+    }
+
```
antimora, line 147: This needs a unit test. Please see the tests in the same module.

antimora (changes requested):
Thank you for keep enhancing ONNX import. I think you're close. Please see my inlined comments. And don't hesitate to ping us if you're stuck on something.

antimora:
Let me know if you need more help with this. I can take over if you'd like.

edmondop:
> Let me know if you need more help with this. I can take over if you'd like.

Let me try to do it this week

antimora:
For this one it's a bit hard to come up with ONNX file that exports Shape op. I have updated shape.py to export but this requires dynamic axes to be enabled. In other cases, pytorch onnx exporter converts the shape values to a constant. 

I have filed https://github.com/tracel-ai/burn/issues/1545 to complete the task.

PR #1069, opened 2023-12-14
Fusion wgpu compilation cache

nathanielsimard:
### Changes

Add a cache step where the compilation of fused kernels are only executed once, then the kernel ID is used without any overhead.

Refactor the way we create kernels with fusion into a DSL so that we can use it to create WGSL shaders easily, not just when using fusion. I migrated the unary kernels to use the new way, but following PRs will migrate more kernels and remove a lot of code.

_Drop_

I removed the clamp max and min functions in Fusion as well as in WGPU, since we need to reduce the number of operations that can be fused to their basic operations. Long story short, they will be fused, so we don't need specific kernels, and having those functions will actually break the fusing stream, so they actually hurt performance as well as add complexity.


burn-wgpu/src/template/unary_scalar.wgsl
```diff
@@ -1,21 +0,0 @@
-@group(0)
```
louisfd, line 1: Don't forget to delete safe tanh file

louisfd (changes requested):
Awesome, please look at my comments

louisfd (approved):
Perfect

PR #1071, opened 2023-12-17
Add mish and softplus activation function

pacowong:
### Checklist

### Related Issues/PRs
NA

### Changes
Mish demonstrated better results than both Swish and ReLU in [1]. It has been included in PyTorch as well.

[1]: https://arxiv.org/abs/1908.08681

### Testing
Please refer to the new test cases.

### DIFF
--- a/burn-dataset/Cargo.toml
+++ b/burn-dataset/Cargo.toml
@@ -18,7 +18,7 @@ audio = ["hound"]
 fake = ["dep:fake"]
 
 sqlite = ["__sqlite-shared", "dep:rusqlite"]
-sqlite-bundled = ["__sqlite-shared", "rusqlite?/bundled"]
+sqlite-bundled = ["__sqlite-shared", "rusqlite/bundled"]
 
 # internal
 __sqlite-shared = [
--- a/burn-dataset/Cargo.toml
+++ b/burn-dataset/Cargo.toml
@@ -18,7 +18,7 @@ audio = ["hound"]
 fake = ["dep:fake"]
 
 sqlite = ["__sqlite-shared", "dep:rusqlite"]
-sqlite-bundled = ["__sqlite-shared", "rusqlite/bundled"]
+sqlite-bundled = ["__sqlite-shared", "rusqlite?/bundled"]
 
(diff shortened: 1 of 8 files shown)


antimora (commented):
Thanks for addition. Can you also please add new operators in the book too?

louisfd (changes requested):
I verified the equations, everything looks good to me. 
It will be good to go once the book is updated and CI is fixed (you seem to have a formatting problem, run `cargo fmt --all`)

pacowong:
Thanks @antimora and @louisfd for the comments. The new functions are added to the book. [0aeaa71](https://github.com/tracel-ai/burn/pull/1071/commits/0aeaa7181887cd79469307c4d2b5a2046c5c1be3) and the formatting issue is fixed.

burn-dataset/Cargo.toml
```diff
@@ -18,7 +18,7 @@ audio = ["hound"]
 fake = ["dep:fake"]
 
 sqlite = ["__sqlite-shared", "dep:rusqlite"]
-sqlite-bundled = ["__sqlite-shared", "rusqlite/bundled"]
+sqlite-bundled = ["__sqlite-shared", "rusqlite?/bundled"]
```
nathanielsimard, line 21: I think the `?` is making the tests fail, and it's a bit unnecessary in this case.

nathanielsimard (approved):
Thanks for the addition; the only remaining work is to fix the CI. You can run it locally to debug things to help you, but I think it's related to my comment below.

burn-dataset/Cargo.toml
pacowong: Thanks. Not sure if my Rust is too new.

According to the [documentation] (https://doc.rust-lang.org/cargo/reference/features.html), 
```
You can add a ? as in "package-name?/feature-name" which will only enable the given feature if something else enables the optional dependency.
```

nathanielsimard: No Rust isn't too new, it's just that `rusqlite` is necessary for the feature flag to work I think.

PR #1078, opened 2023-12-18
Refactor/wgpu/binary

nathanielsimard:
Use the new data structure to generate binary operations and comparison ops.


louisfd (approved):
The naming is better now

Issue #518, opened 2023-07-23
Device in Tensor Type


Gadersd:
### Feature Description
There is a call for a feature that bundles the device a tensor is on in its type. This is to ensure that device compatibilities become compile-time errors instead of runtime errors, which would significantly ease development. This feature would be useful since it helps to detect and prevent critical errors during the development stage thereby saving a great deal of time during the debugging process.

### Feature motivation
The primary motivation behind this feature is to simplify the development process and make it less error-prone. By bundling the device a tensor is on with its type, developers can know in at compile time whether there are any device compatibility issues. This effectively shifts the error detection from runtime to compile time, thereby saving lots of time and efforts which would otherwise be spent on debugging post compilation. 

### Proposed Solution
A potential solution would be to implement a mechanism that includes device information as a category within the tensor type. This would require a shift in how tensors are defined in burn. For instance, instead of the current definition that may look like this:

```
Tensor<Backend, Dims, Float>
```
We could have something like:

```
Tensor<Backend, Dims, Backend::Device<CPU>, Float>
```

Here, the tensor's type now includes information regarding the device it's on, which makes tensor operation compatibility a compile-time concern.

#### Trade-offs 
One potential tradeoff is that this may result in more verbose code with longer tensor definitions. Additionally, this change might break backward compatibility with existing code that assumes the former style of tensor definition.

antimora:
The backend type is already carrying the Device type information [`burn-tensor/src/tensor/backend/base.rs` ](https://github.com/burn-rs/burn/blob/9aca1837c205feff85b0df457fb8e2f7aa507134/burn-tensor/src/tensor/backend/base.rs#L4). 

What's the problem did you come across? Can you give a working example? It'd be easier to understand the problem with an example, maybe.

Gadersd:
For example running the following code yields an error that the two tensors are on different devices when trying to add them together. It would be great if this was caught at compile time instead of runtime.
```
type Backend = burn_tch::TchBackend<f32>;
let device = burn_tch::TchDevice::Cuda(0);
let a: Tensor<Backend, 1, Int> = Tensor::arange_device(0..10, &device);
let b: Tensor<Backend, 1, Int> = Tensor::arange(0..10);
let c = a + b;
```

antimora:
> For example running the following code yields an error that the two tensors are on different devices when trying to add them together. It would be great if this was caught at compile time instead of runtime.
> 
> ```
> type Backend = burn_tch::TchBackend<f32>;
[...]

PR #1081, opened 2023-12-19
Explicit device tensors

kpot:
### Checklist

### Related Issues/PRs

Covers renaming and related refactoring as it was agreed and described in #518

### Changes

Allows for explicit creation of tensors on specific devices. All public interfaces (of layers, models, etc.) have been refactored accordingly, and now allow to specify a `Backend::Device` during initialization. Onnx macroses support generation of device-aware models too.

### Testing

All the tests I could find and run are passing, including

- `cargo test`
- `./run-checks.sh std`
- `./run-checks.sh no-std`
- `./run-checks.sh typos`
- `./run-checks.sh examples`


burn-core/src/optim/grads.rs
```diff
@@ -115,11 +116,11 @@ mod tests {
...
-    fn layer() -> Linear<TestAutodiffBackend> {
-        LinearConfig::new(20, 20).with_bias(true).init()
+    fn layer<B: Backend>(device: &B::Device) -> Linear<B> {
+        LinearConfig::new(20, 20).with_bias(true).init(device)
     }
 
-    fn random_tensor() -> Tensor<TestAutodiffBackend, 2> {
-        Tensor::<TestAutodiffBackend, 2>::random([2, 20], Distribution::Default)
+    fn random_tensor<B: Backend>(device: &B::Device) -> Tensor<B, 2> {
+        Tensor::<B, 2>::random([2, 20], Distribution::Default, &device)
```
nathanielsimard, line 124: I think you can remove the extra `&` before `device` here.

nathanielsimard (approved):
Thanks a lot for this work! I think it's going to be much less likely to have different device errors.

There is some place in `burn-import` where we use `from_data_default` and not `from_data`; not sure what should be the proper usage there, @antimora.

Approved from my side, but I would appreciate one more set of eyes on this, @antimora.

antimora (changes requested):
Oooh, that's a big change in the APIs.

Although I'm reviewing this a bit late, I understand and respect the design decisions made by the team. However, to mitigate potential disruption for our existing users, I'd like to request the following adjustments:

1. All `.init` module methods (in core; the examples can remain as is) should be accompanied by corresponding `.init_default` methods. These should call `.init` with the default device. This approach will allow for quick and easy refactoring of existing model code, essentially requiring just the addition of a `_default` suffix to the method calls. This suggestion aligns well with the changes you've made to other Tensor methods, which are undergoing similar refactoring.

2. In a similar fashion, the model generated from ONNX in the `burn-import/src/burn/graph.rs` file should include a `new_default()` method. This method would call `new` with the default device. If you could adjust the tests, it'd be great too but this more cleanliness and less verbosity.

kpot:
Hi!
@antimora @nathanielsimard I generally agree that something like `init_default` can be useful, but `new_default`... It is not very idiomatic to Rust naming conventions. In Rust, whenever you want to construct something with default settings, you simply implement `Default` trait, and create an instance by calling the `default()` method.

So, I propose to alter the naming. Again. Let's replace the vague `_default` suffix by `_autodev`. It hints at its purpose: automatic selection of device. And it has the same number of characters as `_default`, thus making it possible to do the refactoring more or less using a simple "find and replace" tool, without breaking code formatting.

Another decent names could be be 

- `_autoloc` (automatic location, to me sounds as neutral as `_autodev`), 
- `_adevice` (any device, hints at randomness)
- `_bestdev` (the best device we can choose, implies the choice is not random)
- `_somedev` (hits at an arbitrary choice of the device too)

but you can come up with something even better. What do you think?

antimora:
Sounds good. My top preferences:

- '_defaultdev'
- '_devauto'
- '_devdefault'
- '_autodev'
- '_devault'

kpot:
@antimora Hi! All the changes you requested were implemented. Specifically:

1. I chose `_devauto` suffix as a replacement for the `_default`. 
2. All modules that have their own weights now also have a `init_devauto()` method.
3. ONNX models now include `new_devauto()` method as well
4. All introduced `*_devauto` methods have at least one test covering them.

All the tests and checks I know of are passing.

nathanielsimard:
@kpot, I don't think the new naming is correct. First, "automatic" is associated with something intelligent and not error-prone, so "auto device" isn't the same as "default device." Second, I don't think using `dev` for "device" is actually a good idea; it's normally used for development, and it will be very confusing!

However, I agree that using `_default` isn't the best naming convention, so let me propose something a bit radical: remove all functions with the default device! It will make the API smaller, and someone can always use `Default::default()` explicitly where it applies. I think the only place where it makes sense to use the default device is in tests, not in production code where it can cause device errors.

Enforcing users to specify the device where it's required is not a bad idea, even if it's more verbose.

The only functions where the default device could be used is when implementing the `From` trait. So someone could still use the following:

```rust
let tensor = Tensor::from([5.0, 10.0]);
```

I believe this could be used for most tests. What is your opinion, @kpot, @antimora, @louisfd?

kpot:
@nathanielsimard Well, this was my [original proposal](https://github.com/tracel-ai/burn/issues/518#issuecomment-1806015115) too, to get rid of all the twin `*_device` methods and make them a new norm. Any constructor that internally creates some device-bound parameters must be given a device explicitly. Any half-measures like `_devauto` will inevitably lead to more bugs. So yeah, I support your suggestion. This will require a lot of refactoring within tests (yay, more refactoring!) but it's worth it.

@antimora Do you approve?

antimora:
> @nathanielsimard Well, this was my [original proposal](https://github.com/tracel-ai/burn/issues/518#issuecomment-1806015115) too, to get rid of all the twin `*_device` methods and make them a new norm. Any constructor that internally creates some device-bound parameters must be given a device explicitly. Any half-measures like `_devauto` will inevitably lead to more bugs. So yeah, I support your suggestion. This will require a lot of refactoring within tests (yay, more refactoring!) but it's worth it.
> 
> @antimora Do you approve?

Yes, I guess in the end it would be the right solution. So I approve it. 

I appreciate your willingness to go through big changes and stand by the right decision. I myself had to scrap a big chunk of code during a Burn PR review a few times.

nathanielsimard:
@kpot I think it would be easier to merge the PR right now as is and refactor the tests in other PRs. We only have to remove all occurrences of `autodev` in the codebase before deleting the functions. This way, it minimizes git conflicts as well as making the upcoming PR reviewable! What do you think, @antimora?

antimora (approved):

antimora:
> @kpot I think it would be easier to merge the PR right now as is and refactor the tests in other PRs. We only have to remove all occurrences of `autodev` in the codebase before deleting the functions. This way, it minimizes git conflicts as well as making the upcoming PR reviewable! What do you think, @antimora?

Yep. Lets go ahead and merge it!

PR #1087, opened 2023-12-20
Add persistent cache for autotune

### LINKED ISSUE #911: Autotune: benchmarks cached on disk
The autotune mechanism uses a cache to check if, for a specific operation on an input size within a specific range, the benchmarks have already been computed. If so, the fastest operation in that setting on your machine is stored in the cache and we just use this one. That way, it is likely only in the beginning of training/inference that we need to run benchmarks, and then the cache allows for reuse. 

However, the cache is stored in RAM, which means the benchmarks will be recomputed for every new execution, although they will likely reproduce the same results when run on the same machine. 

By storing the cache and/or all benchmark results in a file on disk, we will avoid the computation of benchmarks at every execution. 

We should invalidate the cache when the burn version is different.
The on disk key should be the autotune key + the device from the compute client.

syl20bnr:
### Checklist

### Related Issues/PRs

This PR fixes #911 

### Changes

- This PR adds a new persistent hash map to the TuneCache type.
- It serializes the persistent hash map as a vector to disk in location `~/.cache/burn/autotune/autotune-cache.json`.
- A checksum is computed to detect changes in operation sets such as key change or operations order changes
- the original in-memory cache is kept and it is initialized from the persistent cache at load time
- the first cache key lookup from the in-memory cache verifies that the checksums from the persistent cache and the current operation set are still the same

### Testing

Tested manually then added new integration tests to validate the checksum behavior and persistent cache invalidation.

### DIFF
--- a/burn-compute/Cargo.toml
+++ b/burn-compute/Cargo.toml
@@ -43,10 +43,10 @@ dirs = { workspace = true , optional = true}
 serde = { workspace = true, optional = true}
 serde_json = { workspace = true, features=["std"], optional = true}
 md5 = { version = "0.7.0", optional = true }
-rand.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 web-time = { version = "0.2.3" }
 
 [dev-dependencies]
 serial_test = "2.0.0"
+rand = { workspace = true }
--- a/burn-compute/Cargo.toml
+++ b/burn-compute/Cargo.toml
@@ -34,6 +34,7 @@ dirs = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true, features = ["std"] }
 md5 = "0.7.0"
+rand.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 web-time = { version = "0.2.3" }
--- a/burn-compute/Cargo.toml
+++ b/burn-compute/Cargo.toml
@@ -17,12 +17,21 @@ default = [
     "channel-mpsc",
     "channel-cell",
     "storage-bytes",
+    "autotune-persistent-cache",
 ]
-std = ["burn-common/std"]
+std = [
+    "burn-common/std",
+]
 channel-mutex = []
 channel-cell = []
 channel-mpsc = [] # Assume std
 storage-bytes = []
+autotune-persistent-cache = [
+    "dirs",
+    "md5",
+    "serde",
+    "serde_json",
+] # Assume std
 
 [dependencies]
 burn-common = { path = "../burn-common", version = "0.12.0", default-features = false }
@@ -30,10 +39,10 @@ derive-new = { workspace = true }
 spin = { workspace = true }
 log = { workspace = true }
 hashbrown = { workspace = true }
-dirs = { workspace = true }
-serde = { workspace = true }
-serde_json = { workspace = true, features = ["std"] }
-md5 = "0.7.0"
+dirs = { workspace = true , optional = true}
+serde = { workspace = true, optional = true}
+serde_json = { workspace = true, features=["std"], optional = true}
+md5 = { version = "0.7.0", optional = true }
 rand.workspace = true
 
 [target.'cfg(target_family = "wasm")'.dependencies]
--- a/burn-compute/Cargo.toml
+++ b/burn-compute/Cargo.toml
@@ -30,6 +30,10 @@ derive-new = { workspace = true }
 spin = { workspace = true }
 log = { workspace = true }
 hashbrown = { workspace = true }
+dirs = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true, features = ["std"] }
+md5 = "0.7.0"
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 web-time = { version = "0.2.3" }
--- a/burn-compute/src/tune/operation.rs
+++ b/burn-compute/src/tune/operation.rs
@@ -3,10 +3,9 @@ use alloc::string::String;
 use alloc::vec::Vec;
 use core::fmt::{Debug, Display};
 use core::hash::Hash;
-use serde::de::DeserializeOwned;
-use serde::Serialize;
 
 /// Default checksum for an operation set
+#[cfg(feature = "autotune-persistent-cache")]
 pub fn compute_checksum(autotunables: &[Box<dyn AutotuneOperation>]) -> String {
     let mut checksum = String::new();
     autotunables.iter().for_each(|op| {
@@ -29,6 +28,7 @@ pub trait AutotuneOperationSet<K>: Send {
     fn fastest(self: Box<Self>, fastest_index: usize) -> Box<dyn AutotuneOperation>;
 
     /// Compute checksum for the set by concatenating the op names together
+    #[cfg(feature = "autotune-persistent-cache")]
     fn compute_checksum(&self) -> String {
         compute_checksum(&self.autotunables())
     }
@@ -48,9 +48,13 @@ pub trait AutotuneOperation {
     fn clone(&self) -> Box<dyn AutotuneOperation>;
 }
 
-/// Trait alias
+#[cfg(feature = "autotune-persistent-cache")]
+/// Trait alias with support for persistent caching
 pub trait AutotuneKey:
-    Clone + Debug + PartialEq + Eq + Hash + Display + Serialize + DeserializeOwned
+    Clone + Debug + PartialEq + Eq + Hash + Display + serde::Serialize + serde::de::DeserializeOwned
 {
 }
+#[cfg(not(feature = "autotune-persistent-cache"))]
+/// Trait alias
+pub trait AutotuneKey: Clone + Debug + PartialEq + Eq + Hash + Display {}
 impl AutotuneKey for String {}
--- a/burn-compute/src/tune/operation.rs
+++ b/burn-compute/src/tune/operation.rs
@@ -1,7 +1,8 @@
 use alloc::boxed::Box;
 use alloc::string::String;
 use alloc::vec::Vec;
-use serde::{Serialize, Deserialize};
+use serde::Serialize;
+use serde::de::DeserializeOwned;
 use core::fmt::{Debug, Display};
 use core::hash::Hash;
 
@@ -43,5 +44,5 @@ pub trait AutotuneOperation {
 }
 
 /// Trait alias
-pub trait AutotuneKey: Clone + Debug + PartialEq + Eq + Hash + Display + Serialize + for<'de> Deserialize<'de> {}
+pub trait AutotuneKey: Clone + Debug + PartialEq + Eq + Hash + Display + Serialize + DeserializeOwned {}
 impl AutotuneKey for String {}
--- a/burn-compute/src/tune/operation.rs
+++ b/burn-compute/src/tune/operation.rs
@@ -1,6 +1,7 @@
 use alloc::boxed::Box;
 use alloc::string::String;
 use alloc::vec::Vec;
+use serde::{Serialize, Deserialize};
 use core::fmt::{Debug, Display};
 use core::hash::Hash;
 
@@ -16,6 +17,15 @@ pub trait AutotuneOperationSet<K>: Send {
     /// Returns the operation for the given index, matching the order
     /// returned by autotunables. Operation obtained here runs on original tensors
     fn fastest(self: Box<Self>, fastest_index: usize) -> Box<dyn AutotuneOperation>;
+
+    /// Compute checksum for the set by concatenating the op names together
+    fn compute_checksum (&self) -> String {
+        let mut checksum = String::new();
+        self.autotunables().iter().for_each(|op| {
+            checksum += op.name();
+        });
+        format!("{:x}", md5::compute(checksum))
+    }
 }
 
 /// Contains operation to run and inputs on which to run it
@@ -33,5 +43,5 @@ pub trait AutotuneOperation {
 }
 
 /// Trait alias
-pub trait AutotuneKey: Clone + Debug + PartialEq + Eq + Hash + Display {}
+pub trait AutotuneKey: Clone + Debug + PartialEq + Eq + Hash + Display + Serialize + for<'de> Deserialize<'de> {}
 impl AutotuneKey for String {}
--- a/burn-compute/src/tune/operation.rs
+++ b/burn-compute/src/tune/operation.rs
@@ -27,7 +27,7 @@ pub trait AutotuneOperationSet<K>: Send {
     /// returned by autotunables. Operation obtained here runs on original tensors
     fn fastest(self: Box<Self>, fastest_index: usize) -> Box<dyn AutotuneOperation>;
 
-    /// Compute checksum for the set by concatenating the op names together
+    /// Compute a checksum that can invalidate outdated cached auto-tune results.
     #[cfg(feature = "autotune-persistent-cache")]
     fn compute_checksum(&self) -> String {
         compute_checksum(&self.autotunables())
--- a/burn-compute/src/tune/operation.rs
+++ b/burn-compute/src/tune/operation.rs
@@ -1,10 +1,19 @@
 use alloc::boxed::Box;
 use alloc::string::String;
 use alloc::vec::Vec;
-use serde::Serialize;
-use serde::de::DeserializeOwned;
 use core::fmt::{Debug, Display};
 use core::hash::Hash;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+/// Default checksum for an operation set
+pub fn compute_checksum(autotunables: &[Box<dyn AutotuneOperation>]) -> String {
+    let mut checksum = String::new();
+    autotunables.iter().for_each(|op| {
+        checksum += op.name();
+    });
+    format!("{:x}", md5::compute(checksum))
+}
 
 /// Groups operations of the same type for autotune
 pub trait AutotuneOperationSet<K>: Send {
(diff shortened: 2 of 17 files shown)


syl20bnr:
The persistent cache is not working in `no-std` env. 
Maybe I can create a feature for it and we ignore it in `no-std` ? What would be the best way to handle this ?

nathanielsimard:
> The persistent cache is not working in `no-std` env. Maybe I can create a feature for it and we ignore it in `no-std` ? What would be the best way to handle this ?

Feature flag for `no-std`!

burn-compute/src/tune/operation.rs
```diff
@@ -3,6 +3,18 @@ use alloc::string::String;
...
+use serde::Serialize;
+
+/// Default checksum for an operation set
+#[cfg(feature = "autotune-persistent-cache")]
+pub fn compute_checksum(autotunables: &[Box<dyn AutotuneOperation>]) -> String {
+    let mut checksum = String::new();
+    autotunables.iter().for_each(|op| {
+        checksum += op.name();
+    });
+    format!("{:x}", md5::compute(checksum))
```
nathanielsimard, line 16: I wonder if the checksum type can be an integer (u64)?

burn-compute/src/tune/operation.rs
```diff
@@ -16,6 +28,12 @@ pub trait AutotuneOperationSet<K>: Send {
     /// Returns the operation for the given index, matching the order
     /// returned by autotunables. Operation obtained here runs on original tensors
     fn fastest(self: Box<Self>, fastest_index: usize) -> Box<dyn AutotuneOperation>;
+
+    /// Compute checksum for the set by concatenating the op names together
```
nathanielsimard, line 32: The doc is wrong, the implementations are free to implement this function with another strategy. => `Compute a checksum that can invalidate outdated cached auto-tune results.`

burn-compute/src/tune/tune_cache.rs
```diff
@@ -1,13 +1,36 @@
...
+    pub use std::fs;
+    pub use std::fs::File;
+    pub use std::io;
+    pub use std::path::Path;
+    pub use std::path::PathBuf;
+}
+
+#[cfg(feature = "autotune-persistent-cache")]
+use std_imports::*;
+
```
nathanielsimard, line 12: Nice trick ! 👌

burn-compute/src/tune/tune_cache.rs
```diff
@@ -1,13 +1,36 @@
...
 use super::AutotuneKey;
 use super::AutotuneOperation;
 use super::AutotuneOperationSet;
 use alloc::boxed::Box;
 use hashbrown::HashMap;
 
+/// Return the file path for the persistent cache on disk
+#[cfg(feature = "autotune-persistent-cache")]
+pub fn get_persistent_cache_file_path() -> PathBuf {
+    let home_dir = dirs::home_dir().expect("Could not get home directory");
```
nathanielsimard, line 22: The expect message should not be an error message, but an expectation:

```rust
let home_dir = dirs::home_dir().expect("Should have an home directory.");
```

burn-compute/src/tune/tune_cache.rs
```diff
@@ -20,24 +43,116 @@ pub enum TuneCacheResult<K> {
...
-        TuneCache {
-            cache: HashMap::new(),
+        #[cfg(feature = "autotune-persistent-cache")]
+        {
+            let mut cache = TuneCache {
+                cache: HashMap::new(),
+                persistent_cache: HashMap::new(),
+            };
+            if let Err(e) = cache.load() {
+                eprintln!(
```
nathanielsimard, line 53: use `log::warn!` instead of `eprintln!` here.

burn-compute/src/tune/tune_cache.rs
```diff
@@ -20,24 +43,116 @@ pub enum TuneCacheResult<K> {
...
+
+        #[cfg(not(feature = "autotune-persistent-cache"))]
+        {
+            TuneCache {
+                cache: HashMap::new(),
+            }
         }
     }
 
     #[allow(clippy::borrowed_box)]
```
nathanielsimard, line 69: Is the clippy warning still an issue? I don't see any borrowed box here (only an owned box).

burn-compute/src/tune/tune_cache.rs
```diff
@@ -20,24 +43,116 @@ pub enum TuneCacheResult<K> {
...
+        fastest_index: usize,
+    ) {
+        self.persistent_cache.insert(key, (checksum, fastest_index));
+    }
+
+    /// Load the persistent cache data from disk
+    #[cfg(feature = "autotune-persistent-cache")]
+    pub(crate) fn load(&mut self) -> Result<(), io::Error> {
+        let file_path = get_persistent_cache_file_path();
+        // note: reading file ro memory is faster than using
```
nathanielsimard, line 123: Typo: `ro` => `from`?

burn-compute/src/tune/tune_cache.rs
```diff
@@ -20,24 +43,116 @@ pub enum TuneCacheResult<K> {
...
+            self.cache.insert(key.clone(), (false, *index));
+        }
+        Ok(())
+    }
+
+    /// Save the persistent cache on disk
+    #[cfg(feature = "autotune-persistent-cache")]
+    pub(crate) fn save(&self) {
+        let file_path = get_persistent_cache_file_path();
+        let file = File::create(file_path).expect("Unable to open autotune persistent cache file");
```
nathanielsimard, line 153: `Unable to open autotune persistent cache file` => `Should be able to create the autotune persistent cache file`.

burn-compute/src/tune/tune_cache.rs
```diff
@@ -20,24 +43,116 @@ pub enum TuneCacheResult<K> {
...
+        let data = self.persistent_cache.iter().collect::<Vec<_>>();
+        serde_json::to_writer_pretty(file, &data)
+            .expect("Unable to write to autotune persistent cache");
```
nathanielsimard, line 156: `Unable to write to autotune persistent cache` => `Should be able to write to autotune persistent cache`

burn-compute/src/tune/tuner.rs
```diff
@@ -15,15 +15,16 @@ use crate::client::ComputeClient;
...
 pub struct Tuner<S: ComputeServer, C> {
     tune_cache: TuneCache<S::AutotuneKey>,
     _channel: PhantomData<C>,
 }
 
+#[allow(clippy::new_without_default)]
 impl<S: ComputeServer, C: ComputeChannel<S>> Tuner<S, C> {
-    /// Returns a tuner with empty cache
+    /// Returns a tuner with cache initialized from persistent cache
     pub fn new() -> Self {
```
nathanielsimard, line 28: I think you will need to implement `Default` to satisfy clippy:

```rust
impl<S: ComputeServer, C: ComputeChannel<S>> Default for Tuner<S, C> {
    fn default() -> Self {
        Self::new()
    }
}
```

When providing an empty public `new` method, it's also a good idea to implement `Default`.

burn-compute/src/tune/tuner.rs
```diff
@@ -62,16 +62,20 @@ impl<S: ComputeServer, C: ComputeChannel<S>> Tuner<S, C> {
...
         log::info!("Fastest result {fastest_name}-{key}");
 
-        self.tune_cache.cache_insert(key, fastest_index);
+        self.tune_cache.cache_insert(key.clone(), fastest_index);
+        #[cfg(feature = "autotune-persistent-cache")]
+        {
+            let checksum = autotune_operation_set.compute_checksum();
+            self.tune_cache
+                .persistent_cache_insert(key, checksum, fastest_index);
+            self.tune_cache.save();
```
nathanielsimard, line 76: I think it would be better to avoid writing to disk each time we auto-tune with persistent cache enabled. It's not a priority in this PR, but if you can implement the `Drop` trait for `Tuner` and call `self.tune_cache.save()`, it would save the cache when the tuner is deconstructed.

burn-compute/tests/dummy/compute.rs
```diff
@@ -30,3 +31,16 @@ pub fn client(device: &DummyDevice) -> DummyClient {
...
+        let storage = BytesStorage::default();
+        let memory_management =
+            SimpleMemoryManagement::new(storage, DeallocStrategy::Never, SliceStrategy::Never);
+        let server = DummyServer::new(memory_management);
+        let channel = MutexComputeChannel::new(server);
+        let tuner = Arc::new(Mutex::new(Tuner::new()));
+
+        ComputeClient::new(channel, tuner)
+    })
+}
```
nathanielsimard, line 46: If you need your own instance of `COMPUTE` for the tests, I would suggest to create this in the test file you need it with a better name: `persistent_cache_test_client`.

burn-compute/tests/integration_test.rs
```diff
@@ -146,3 +157,74 @@ fn autotune_cache_miss_test() {
...
+    assert_eq!(obtained_resource.read(), Vec::from([5, 6, 7, 8, 9]));
+}
+
+#[test]
+#[serial]
+#[cfg(feature = "std")]
+fn autotune_cache_different_checksums_return_a_cache_miss() {
+    // we use two clients in order to have freshly initialized autotune caches
+    let client1 = client(&DummyDevice);
+    let client2 = client2(&DummyDevice);
```
nathanielsimard, line 200: I think the second client should be generated after the cache is saved on disk. For that, I think we need to create a new client (not just a client to a static compute).

burn-train/Cargo.toml
```diff
@@ -11,7 +11,8 @@ repository = "https://github.com/tracel-ai/burn/tree/main/burn-train"
 version.workspace = true
 
 [features]
-default = ["metrics", "tui"]
+default = ["metrics"]
+# default = ["metrics", "tui"]
```
nathanielsimard, line 15: Keep the old defaults.

nathanielsimard (changes requested):
Good job, it's overall very clean, just a lot of small comments here and there!

burn-compute/src/tune/tune_cache.rs
```diff
@@ -1,13 +1,36 @@
...
+    let path = Path::new(&path_dir);
+    path.join("autotune-cache.json")
+}
+
 /// Use to find and reuse the best kernel for some input
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub(crate) struct TuneCache<K> {
-    cache: HashMap<K, usize>,
+    cache: HashMap<K, (bool, usize)>,
```
louisfd, line 31: Maybe make a struct CacheResult that has is_checked and index fields? It would be clearer than a tuple

louisfd: Also maybe give cache a name that distinguishes it better from persistent_cache, like in_memory_cache

burn-compute/src/tune/tune_cache.rs
```diff
@@ -1,13 +1,36 @@
...
+    #[cfg(feature = "autotune-persistent-cache")]
+    persistent_cache: HashMap<K, (String, usize)>,
```
louisfd, line 33: Again, a struct rather than a tuple?

burn-compute/src/tune/tune_cache.rs
```diff
@@ -20,24 +43,116 @@ pub enum TuneCacheResult<K> {
...
     ) -> TuneCacheResult<K> {
-        let index = self.cache.get(&autotune_operation_set.key());
-        if let Some(&i) = index {
-            return TuneCacheResult::Hit(autotune_operation_set.fastest(i));
+        let key = autotune_operation_set.key();
+        let result = self.cache.get_mut(&key);
+
+        #[cfg(feature = "autotune-persistent-cache")]
+        {
+            if let Some((is_checked, index)) = result {
```
louisfd, line 79: What is actually going on here? I think it would deserve a comment

burn-compute/tests/integration_test.rs
```diff
@@ -88,9 +88,14 @@ fn autotune_basic_multiplication_execution() {
...
 #[cfg(feature = "std")]
-fn autotune_cache_hit_test() {
+fn autotune_cache_same_key_return_a_cache_hit() {
     let client = client(&DummyDevice);
 
+    // note: the key name depends on the shapes of the operation set
+    // see log_shape_input_key for more info.
+
+    // in this test both shapes [1,3] and [1,4] end up with the same key name
+    // which is 'cache_test-1,4'
```
louisfd, line 98: It's good that you've added those comments! I guess the tests were confusing at first

louisfd (changes requested):
Very good PR, it's awesome that we won't need the autotune overhead at each new execution. I have some minor comments but globally it's very good

burn-compute/src/tune/operation.rs
syl20bnr: We can do it if we are OK with loosing half of the hash value.

burn-compute/src/tune/tuner.rs
syl20bnr: I added the attribute `#[allow(clippy::new_without_default)]` 
Is that OK or is it a strong convention to implement a default as well ?

burn-compute/src/tune/tuner.rs
syl20bnr: For reference, we tried to do it but COMPUTE is static and thus it is never dropped. We decided to keep saving the cache each time an operation set is autotuned.

burn-compute/src/tune/operation.rs
syl20bnr: We decided to keep it as a string.

burn-compute/src/tune/tune_cache.rs
syl20bnr: Done, that's way better 👍🏻

burn-compute/src/tune/tune_cache.rs
syl20bnr: With the last changes I believe the code is now understandable without a comment. Tell me what you think.

syl20bnr:
Ready for review.

I added a prefix to the cache filename with some device info passed to the tuner.
This is the responsibility of the backend to compute a prefix, for now we have only wgpu with the format `wgpu-<device number>-<wgpu backend>`. Example of resulting filename:
- `wgpu-9988-vulkan-autotune-cache.json`
- `wgpu-9988-dx12-autotune-cache.json`
- `wgpu-1514-dx12-autotune-cache.json`

Maybe we should add a function to a trait somewhere to force users to implement the prefix ?

burn-compute/src/tune/tune_cache.rs
louisfd: Yes it's clearer with the struct attribute names.

louisfd:
> * `wgpu-9988-vulkan-autotune-cache.json`
> * `wgpu-9988-dx12-autotune-cache.json`
> * `wgpu-1514-dx12-autotune-cache.json`
> 
> Maybe we should add a function to a trait somewhere to force users to implement the prefix ?

Hi @syl20bnr, 
I'm not too sure about how wgpu selects devices with different IDs often, and how much different ids affect the kernel speeds, but my feeling is that using the id in the cache name is too precise. If wgpu devices like `wgpu-9988-dx12` and `wgpu-1514-dx12` have similar behaviours we'll be running autotune far too often. I agree though that we must distinguish between vulkan and dx12 and for that we need to get the adapter info.

syl20bnr:
@louisfd 9988 is the GPU on my system and 1514 is the CPU. I can also add the device type to the prefix or even the device name but it can be long.

louisfd:
@syl20bnr ok I was not sure if that id changed within the same machine. If it's always the same number then it's fine

burn-compute/Cargo.toml
```diff
@@ -17,19 +17,33 @@ default = [
...
 burn-common = { path = "../burn-common", version = "0.12.0", default-features = false }
 derive-new = { workspace = true }
 spin = { workspace = true }
 log = { workspace = true }
 hashbrown = { workspace = true }
+dirs = { workspace = true , optional = true}
+serde = { workspace = true, optional = true}
+serde_json = { workspace = true, features=["std"], optional = true}
+md5 = { version = "0.7.0", optional = true }
+rand.workspace = true
```
nathanielsimard, line 46: Should it be optional or in the dev-dependencies section (I think you are only using it in tests)?

burn-compute/src/tune/tune_cache.rs
```diff
@@ -19,25 +63,159 @@ pub enum TuneCacheResult<K> {
...
+    /// Save the persistent cache on disk
+    #[cfg(feature = "autotune-persistent-cache")]
+    pub(crate) fn save(&self) {
+        let file_path = self.get_persistent_cache_file_path();
+        if let Some(parent_dir) = file_path.parent() {
+            if !parent_dir.exists() {
+                let msg = format!(
+                    "Should be able to create directory '{}' for autotune persistent cache file",
+                    parent_dir.to_str().unwrap()
+                );
```
nathanielsimard, line 202: Instead of always formatting the error msg, you can use the following pattern:

```rust
my_function().unwrap_or_else(|| panic!("custom message with format option: {somevariable}"));
```

burn-compute/src/tune/tune_cache.rs
```diff
@@ -19,25 +63,159 @@ pub enum TuneCacheResult<K> {
...
+                fs::create_dir_all(parent_dir).expect(&msg);
+            }
+        }
+        let msg = format!(
+            "Should be able to open autotune persistent cache file: {:?}",
+            &file_path.to_str().unwrap()
+        );
```
nathanielsimard, line 209: Same here

nathanielsimard (changes requested):

burn-compute/Cargo.toml
syl20bnr: Moved to dev deps.

nathanielsimard (approved):

syl20bnr:
@louisfd I will need your approval on this PR to be able to merge it.

louisfd (approved):
LGTM thanks for your efforts on this issue!!

PR #1093, opened 2023-12-22
Fix/wgpu/fusion

nathanielsimard:
Add support for fused element-wise integer operations.

But most importantly, actually enable fusion without the need for sync. That's right, we didn't have fused operations in real settings, and there was a bug when fusion was triggered by an operation instead of a sync, and this PR fixes this.

### DIFF
--- a/burn-fusion/src/handle.rs
+++ b/burn-fusion/src/handle.rs
@@ -30,11 +30,13 @@ impl<B: FusionBackend> HandleContainer<B> {
 
     /// Register a handle for the given [tensor id](TensorId).
     pub fn register_handle(&mut self, id: TensorId, handle: B::Handle) {
+        // log::info!("Register handle {:?}", id);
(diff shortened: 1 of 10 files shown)


burn-fusion/src/handle.rs
```diff
@@ -33,15 +33,21 @@ impl<B: FusionBackend> HandleContainer<B> {
...
     }
 
-    /// Get the handle for the given [tensor id](TensorId).
-    pub fn get_handle(&mut self, tensor: &TensorDescription) -> B::Handle {
+    /// Get the handle for the given [tensor id](TensorId). The status is used to know if the
+    /// tensor should be pop out of the current tensor map, necessary for inplace operations.
+    ///
+    /// # Warnings
+    ///
+    /// Make sure the status correspond to the operation you want ot execute the handles one,
```
louisfd: ot, handle, on

louisfd (approved):
😄

PR #1101, opened 2023-12-30
Exit event handler if KeyEvent is of release kind

amfaber:
Hello, I just ran through the MNIST example in the book, and I am really impressed with the project so far! Great balance between explicitness and easy of use. 

A fantastic touch to have metrics shown in a tui by default as well, though I noticed that on my Windows computer, each arrow key press changed the metric being displayed twice - once on pressing and once on release. Due to the example having only two metrics, this meant that I needed to hold down the key and hope that I landed on the correct metric when I let go.

I just had a quick go at fixing this issue by early-returning from the event handler if the KeyEvent is of the Release variant.

All checks passed with `run-checks.ps1 all`


burn-train/src/renderer/tui/metric_numeric.rs
```diff
@@ -107,6 +107,10 @@ impl NumericMetricsState {
     /// Handle the current event.
     pub(crate) fn on_event(&mut self, event: &Event) {
         if let Event::Key(key) = event {
+            match key.kind {
+                KeyEventKind::Press | KeyEventKind::Repeat => (),
+                KeyEventKind::Release => return,
+            }
```
nathanielsimard, line 113: I think the opposite would be preferred, we should perform the action on key release. But thanks for the fix :)

nathanielsimard (changes requested):

burn-train/src/renderer/tui/metric_numeric.rs
syl20bnr: I tested both ways on Windows and I prefer the behavior when the Press and Repeat events are enabled as @amfaber proposed, feels more natural and the terminal stays responsive when holding down the keys, at least on my system configuration. Would be great to have additional feedback on this.

amfaber: 👍
That is probably more "idiomatic"

amfaber: > 👍 That is probably more "idiomatic"

Oops, replied without refreshing my browser

From general gui experience, I greatly prefer actions to happen on release if they are "cancelable" with something like esc or by performing another action, such that one can change ones mind at the last moment. For something as simple as this that is instantly undoable (by pressing the opposing button) and that isn't cancelable, I have no strong feelings for either Press or Release

syl20bnr: On Press has the advantage of being instantaneous as soon as the key is pressed, it can also feels more snappy. With On Release nothing happens while the key is pressed which may not be ideal.

I guess we have three possible ergonomics here:
- 1) original version: Press and Repeat On / Release Off
- 2) opposite version: Press and Repeat Off / Release On
- 3) alternate version: Press On / Release and Repeat Off

 I like better 2) and 3) with Press On. But I have also no strong feeling about it so whatever works if fine.

amfaber: A quick test showed that holding down the button actually sends repeated "Press" events, not "Repeated" (on my windows 11 atleast). Using "Release" also felt more sluggish, as  @syl20bnr pointed out. With this in mind I suggest we keep the original setup of 
```rust
match key.kind {
    KeyEventKind::Press | KeyEventKind::Repeat => (),
    KeyEventKind::Release => return,
}
```

syl20bnr: I tested on MacOS with iTerm (which is widely used) and `Release` does not work. Only `Press` works. Furthermore `Repeat` does not appear to fire [1] and `Press` acts also as a `Repeat` which means that holding an arrow key down cycles through metrics. Unfortunately iTerm is quite slow and the experience is not as smooth as on Windows, keeping the right arrow key pressed select a random metric every second or so.

[1] tried with a different arms order like:

```rust
            match key.kind {
                KeyEventKind::Release | KeyEventKind::Repeat => return,
                KeyEventKind::Press => (),
            }
```

syl20bnr: It's a bit more snappy in Alacritty on MacOS but not so much, might be my old Intel Mac that is slow 😝

nathanielsimard (approved):

syl20bnr:
@amfaber can you rebase your PR on top of main ? It should fixes the CI checks.

antimora:
@syl20bnr rebased it myself. Lets push this PR through before the upcoming release.

syl20bnr:
Thank you @antimora ! I merged although the tests are not passing as the scope of the change is trivial and we know that CI is broken for another reason.

PR #1104, opened 2024-01-03
Refactor/fusion optimization

nathanielsimard:
Refactor the fusion operation to support saving the state of the JIT cache so that we can reduce cold start. This is also necessary if we want to combine fusion with autotune, which is planned, since the autotune cache should share the same generated ID from the dynamically created kernel.

This PR doesn't actually save anything, but it makes everything serializable and deserializable with the correct ID. The refactor consists mostly of removing `dyn Trait` so that we can actually serialize things. The dynamic trait is replaced with an enum, which makes sense in this context.

Also found a bug with burn-fusion when changing the device, so corrected that as well.


burn-fusion/Cargo.toml
```diff
@@ -15,9 +15,11 @@ default = ["std"]
...
+burn-common = { path = "../burn-common", version = "0.12.0" }
 hashbrown = { workspace = true }
-derive-new = {workspace = true}
-spin = {workspace = true}
-log = {workspace = true}
+derive-new = {workspace = true }
+spin = { workspace = true }
+log = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
```
syl20bnr, line 25:
```suggestion
serde = {workspace = true, features = ["std", "derive"]}
serde_json = {workspace = true, features = ["std"]}
```

syl20bnr (commented):
LGTM, great work !
I suppose we will add test when saving and loading are implemented ?

PR #1108, opened 2024-01-04
Remove annoying cargo warning

syl20bnr:
### Checklist

### Changes

Package level profiles are ignored when workspace is enabled so these lines are useless.

This commit moves the profile to the workspace level but only commented for reference as panic, lto and rpath cannot be overridden

Source: https://doc.rust-lang.org/cargo/reference/profiles.html#overrides

```
The precedence for which value is used is done in the following order (first match wins):

[profile.dev.package.name] — A named package.
[profile.dev.package."*"] — For any non-workspace member.
[profile.dev.build-override] — Only for build scripts, proc macros, and their dependencies.
[profile.dev] — Settings in Cargo.toml.
Default values built-in to Cargo.
Overrides cannot specify the panic, lto, or rpath settings.
```

Fixes these annoying warning on each workspace package:

image


nathanielsimard (commented):
Maybe we should keep those profiles in the Cargo.toml of each example, but commented? @antimora

antimora (approved):
Yeah, we can remove them since they're causing root level warnings. I am not sure how much this might affect the performance but since this is for demonstration purposes, it might be okay.

syl20bnr:
They actually do nothing already since they are ignored.
@nathanielsimard do we remove they completely ? In the end I don't think that's useful to move them at the root level.

nathanielsimard:
@syl20bnr yeah we could remove them altogether

syl20bnr:
OK, I removed them completely.

PR #1110, opened 2024-01-04
Remove `*_devauto` functions

kpot:
### Checklist

### Related Issues/PRs

This work is done as a part of #518, and implements [this decision](https://github.com/tracel-ai/burn/pull/1081#issuecomment-1864818261) to remove all `*_devauto` functions that construct tensors without the need to explicitly specify a device.

### Changes

All the tests, documentation examples etc. were updated.
I've also spotted and fixed a few remaining places that would still use a default device implicitly: `one_hot`, parameters in `BatchNorm` etc.

Sadly, one aspect I left untouched: de-serialization. It relies on this trait

```rust
/// Trait to define a family of types which can be recorded using any [settings](PrecisionSettings).
pub trait Record: Send + Sync {
    /// Type of the item that can be serialized and deserialized.
    type Item<S: PrecisionSettings>: Serialize + DeserializeOwned;

    /// Convert the current record into the corresponding item that follows the given [settings](PrecisionSettings).
    fn into_item<S: PrecisionSettings>(self) -> Self::Item<S>;

    /// Convert the given item into a record.
    fn from_item<S: PrecisionSettings>(item: Self::Item<S>) -> Self;
}
```

And currently I don't have a clear idea how to change it in a way that would make a de-serialization to an explicit device possible, while still remaining convenient for the `#[derive(Record)]` macro and non-tensor types like `Vec`.

So for now de-serialization is still going to be done on a default device first.

### Testing

On my machine it passed all the tests I know of.


nathanielsimard:
@kpot Thanks a lot! I'll do a proper review in the coming days. As for the serialization system, I think the best way will be to introduce a device somewhere in the trait, but this can be done in a following PR as well.

nathanielsimard (approved):
Looks good to me!

Just doing this refactor fixes some problems that we have, in mask generation, for instance, where the default device needed to be the same as the one passed as an argument. I'm pretty happy about the change, way harder to make a mistake and pretty clear where you need to pay attention to the device. Thanks a lot for this work.

We will wait for the CI to be fixed before merging, @syl20bnr.

syl20bnr:
@kpot CI is fixed in main, you can rebase this PR.

kpot:
> @kpot CI is fixed in main, you can rebase this PR.

Done

PR #1123, opened 2024-01-08
Wgpu fusion auto-vectorized operations

### LINKED ISSUE #1111: Support auto-vectorization of fused compute shaders
Support executing shaders with vectorized operations instead of scalar operations (`vec4`, `vec3`, and `vec2`). Dispatch to the right shader when the shapes are compatible with vector sizes.

nathanielsimard:
Fix #1111


louisfd (approved):
Looks very good to me, vec4 all the way!!

Issue #1072, opened 2023-12-18
Being able to collect backend-comparison benchmarks


nathanielsimard:
The goal of this issue is to allow users to share their benchmark results with the community so that we can report how fast burn is running on a variety of hardwares.

- [x] Create a web service where users can upload their results. We should have a github auth to ensure to a minimum the validity of the data.
- [x] Upload the data from a CLI (and run the benchmark from the CLI as well). We could code that CLI directly in `backend-comparison`.

syl20bnr:
Done 🎉

PR #1131, opened 2024-01-11
Refactor serialization of benchmarks

syl20bnr:
See changes below for a list of changes.

I  will add an "upload" function to  the persistence module in a further PR.

The current naming of the saved files need a bit of work, currently I just do `benchmarks_<timestamp>.json` but we could  have some race  condition and get some files overridden. --> We updated it to `bench_<name>_<timestamp>`.

The computed  values are saved in seconds, maybe we should  save them in milliseconds ? --> We  changed to microseconds.

You can see an example of produced file here: https://gist.github.com/syl20bnr/b0fdbbf0c3877b9e4ddf262a9cc388ac

### Checklist

### Related Issues/PRs

- https://github.com/tracel-ai/burn/issues/1072

### Changes

* flatten benchmarks data to make it easier to save documents to a database and query them
* split some information into their own fields like backend and device
* add new serialized info:
  - computed values (mean, median, variance, min, max)
  - number of samples
  - operation name
  - tensor shapes if any
* serialize to separate files, one file per benchmark run
* simplify persistence module to only a save method
* remove remaining `num_repeats` in benches execute functions. 

### Testing

Tested serialization to disk with all the available benchmarks.

### DIFF
--- a/backend-comparison/benches/binary.rs
+++ b/backend-comparison/benches/binary.rs
@@ -12,15 +12,11 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
     type Args = (Tensor<B, D>, Tensor<B, D>);
 
     fn name(&self) -> String {
-        "Binary Ops".into()
+        "binary".into()
     }
 
-    fn operation(&self) -> Option<String> {
-        Some("binary".to_string())
-    }
-
-    fn shapes(&self) -> Option<Vec<String>> {
-        Some(vec![format!("{:?}", self.shape.dims)])
+    fn shapes(&self) -> Vec<Vec<usize>> {
+        vec!(self.shape.dims.into())
     }
 
     fn execute(&self, (lhs, rhs): Self::Args) {
@@ -50,7 +46,7 @@ fn bench<B: Backend>(device: &B::Device) {
         device: device.clone(),
     };
 
-    save::<B>("binary", vec![run_benchmark(benchmark)], device).unwrap();
+    save::<B>(vec![run_benchmark(benchmark)], device).unwrap();
 }
 
 fn main() {
--- a/backend-comparison/benches/binary.rs
+++ b/backend-comparison/benches/binary.rs
@@ -16,7 +16,7 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
     }
 
     fn shapes(&self) -> Vec<Vec<usize>> {
-        vec!(self.shape.dims.into())
+        vec![self.shape.dims.into()]
     }
 
     fn num_repeats(&self) -> usize {
--- a/backend-comparison/benches/binary.rs
+++ b/backend-comparison/benches/binary.rs
@@ -1,4 +1,4 @@
-use backend_comparison::persistence::Persistence;
+use backend_comparison::persistence::save;
 use burn::tensor::{backend::Backend, Distribution, Shape, Tensor};
 use burn_common::benchmark::{run_benchmark, Benchmark};
 
@@ -13,6 +13,14 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
 
     fn name(&self) -> String {
         "Binary Ops".into()
+    }
+
+    fn operation(&self) -> Option<String> {
+        Some("binary".to_string())
+    }
+
+    fn shapes(&self) -> Option<Vec<String>> {
+        Some(vec![format!("{:?}", self.shape.dims)])
     }
 
     fn execute(&self, (lhs, rhs): Self::Args) {
@@ -42,7 +50,7 @@ fn bench<B: Backend>(device: &B::Device) {
         device: device.clone(),
     };
 
-    Persistence::persist::<B>(vec![run_benchmark(benchmark)], device)
+    save::<B>(vec![run_benchmark(benchmark)], device).unwrap();
 }
 
 fn main() {
--- a/backend-comparison/benches/binary.rs
+++ b/backend-comparison/benches/binary.rs
@@ -19,8 +19,12 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
         vec!(self.shape.dims.into())
     }
 
+    fn num_repeats(&self) -> usize {
+        self.num_repeats
+    }
+
     fn execute(&self, (lhs, rhs): Self::Args) {
-        for _ in 0..self.num_repeats {
+        for _ in 0..self.num_repeats() {
             // Choice of add is arbitrary
             B::add(lhs.clone().into_primitive(), rhs.clone().into_primitive());
         }
--- a/backend-comparison/benches/binary.rs
+++ b/backend-comparison/benches/binary.rs
@@ -4,7 +4,6 @@ use burn_common::benchmark::{run_benchmark, Benchmark};
 
 pub struct BinaryBenchmark<B: Backend, const D: usize> {
     shape: Shape<D>,
-    num_repeats: usize,
     device: B::Device,
 }
 
@@ -19,15 +18,9 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
         vec![self.shape.dims.into()]
     }
 
-    fn num_repeats(&self) -> usize {
-        self.num_repeats
-    }
-
     fn execute(&self, (lhs, rhs): Self::Args) {
-        for _ in 0..self.num_repeats() {
-            // Choice of add is arbitrary
-            B::add(lhs.clone().into_primitive(), rhs.clone().into_primitive());
-        }
+        // Choice of add is arbitrary
+        B::add(lhs.clone().into_primitive(), rhs.clone().into_primitive());
     }
 
     fn prepare(&self) -> Self::Args {
@@ -46,7 +39,6 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
 fn bench<B: Backend>(device: &B::Device) {
     let benchmark = BinaryBenchmark::<B, 3> {
         shape: [32, 512, 1024].into(),
-        num_repeats: 10,
         device: device.clone(),
     };
 
--- a/backend-comparison/benches/binary.rs
+++ b/backend-comparison/benches/binary.rs
@@ -50,7 +50,7 @@ fn bench<B: Backend>(device: &B::Device) {
         device: device.clone(),
     };
 
-    save::<B>(vec![run_benchmark(benchmark)], device).unwrap();
+    save::<B>("binary", vec![run_benchmark(benchmark)], device).unwrap();
 }
 
 fn main() {
--- a/backend-comparison/benches/custom_gelu.rs
+++ b/backend-comparison/benches/custom_gelu.rs
@@ -24,8 +24,7 @@ impl<B: Backend, const D: usize> Benchmark for CustomGeluBenchmark<B, D> {
     type Args = Tensor<B, D>;
 
     fn name(&self) -> String {
-        let kindstr = format!("{:?}", self.kind);
-        format!("gelu_{}", kindstr.to_lowercase())
+        "gelu".into()
     }
 
     fn options(&self) -> Option<String> {
--- a/backend-comparison/benches/custom_gelu.rs
+++ b/backend-comparison/benches/custom_gelu.rs
@@ -33,7 +33,7 @@ impl<B: Backend, const D: usize> Benchmark for CustomGeluBenchmark<B, D> {
     }
 
     fn shapes(&self) -> Vec<Vec<usize>> {
-        vec!(self.shape.dims.into())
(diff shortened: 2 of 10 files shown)


syl20bnr:
For the `.expect()` messages, I am not sure what to put in them. I see a lot of people putting an explanation of the error there and also sometimes people putting what is expected to  work. The word "expect" suggests the latter but that's not what I see online most of the time 🤷🏻‍♂️  What's our code guideline on this ?

backend-comparison/benches/binary.rs
```diff
@@ -15,6 +15,14 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
         "Binary Ops".into()
     }
 
+    fn operation(&self) -> Option<String> {
+        Some("binary".to_string())
+    }
```
nathanielsimard, line 20: I don't think we should have the name and the operation, the name is sufficient and more general than the operation.

backend-comparison/benches/binary.rs
```diff
@@ -15,6 +15,14 @@ impl<B: Backend, const D: usize> Benchmark for BinaryBenchmark<B, D> {
...
+
+    fn shapes(&self) -> Option<Vec<String>> {
+        Some(vec![format!("{:?}", self.shape.dims)])
+    }
```
nathanielsimard, line 24: Shapes are supposed to be numbers, so maybe `Vec<Vec<usize>>` is better. Also, I don't think we need an option of vec; an empty vec will do the job if no shapes are provided.

syl20bnr:
Update the file name with a bench name and the first 8 chars of a random uuid, examples:
- `benchmarks_gelu_1b2bb391.json`
- `benchmarks_matmul_f79409f7.json`

The name is passed as an argument to the `save` function because each ran bench in a benchmark can have a different name and operation used. For instance with `gelu` where we have 3 batches of benches with different gelus.

louisfd:
> For the `.expect()` messages, I am not sure what to put in them. I see a lot of people putting an explanation of the error there and also sometimes people putting what is expected to work. The word "expect" suggests the latter but that's not what I see online most of the time 🤷🏻‍♂️ What's our code guideline on this ?

From [here](https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style): _If you’re having trouble remembering how to phrase expect error messages remember to focus on the word “should” as in “env variable should be set by blah” or “the given binary should be available and executable by the current user”._

So I just stick to the word should and it works out

backend-comparison/src/persistence/base.rs
```diff
@@ -1,100 +1,105 @@
...
+///  [
+///    {
+///      "backend": "backend name",
+///      "device": "device name",
+///      "git_hash": "hash",
+///      "name": "benchmark name",
+///      "operation": "operation name",
+///      "shapes": ["shape dimension", "shape dimension", ...],
+///      "timestamp": "timestamp",
+///      "numSamples": "number of samples",
```
nathanielsimard, line 34: `numSamples` => `num_samples`

backend-comparison/src/persistence/base.rs
```diff
@@ -1,100 +1,105 @@
...
+///      "min": "duration in seconds",
+///      "max": "duration in seconds",
+///      "median": "duration in seconds",
+///      "mean": "duration in seconds",
+///      "variance": "duration in seconds"
+///      "rawDurations": ["duration 1", "duration 2", ...],
```
nathanielsimard, line 40: `rawDurations` => `raw_durations`

backend-comparison/src/persistence/base.rs
```diff
@@ -1,100 +1,105 @@
...
-    fn load(path: &PathBuf) -> Self {
-        let results = match File::open(path) {
-            Ok(file) => serde_json::from_reader(file)
-                .expect("Should have parsed to BenchmarkCommitResults struct"),
-            Err(_) => HashMap::default(),
-        };
+    let uuid = Uuid::new_v4().simple().to_string();
+    let short_uuid = &uuid[..8];
+    let file_name = format!("benchmarks_{}_{}.json", name, short_uuid);
+    let file_path = cache_dir.join(file_name);
```
nathanielsimard, line 62: I think using a timestamp is enough and will be better sorted on disk.

backend-comparison/src/persistence/base.rs
```diff
@@ -1,100 +1,105 @@
...
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("BenchmarkRecord", 11)?;
+        state.serialize_field("backend", &self.backend)?;
+        state.serialize_field("device", &self.device)?;
+        // Serialize fields of BenchmarkResult
+        state.serialize_field("rawDurations", &self.results.raw.durations)?;
+        state.serialize_field("numSamples", &self.results.raw.durations.len())?;
+        state.serialize_field("mean", &self.results.computed.mean.as_secs_f64())?;
```
nathanielsimard, line 93: Serialize in the smallest time unit possible, since we can benchmark things that take microseconds.

backend-comparison/benches/unary.rs
```diff
@@ -17,6 +17,14 @@ impl<B: Backend, const D: usize> Benchmark for UnaryBenchmark<B, D> {
         "Unary Ops".into()
```
louisfd, line 17: Name here

backend-comparison/benches/unary.rs
```diff
@@ -17,6 +17,14 @@ impl<B: Backend, const D: usize> Benchmark for UnaryBenchmark<B, D> {
...
     }
 
+    fn operation(&self) -> Option<String> {
+        Some("unary".to_string())
```
louisfd, line 21: Name again here

backend-comparison/benches/unary.rs
```diff
@@ -41,7 +49,7 @@ fn bench<B: Backend>(device: &B::Device) {
 
     let benchmark = UnaryBenchmark::<B, D>::new(shape, num_repeats, device.clone());
 
-    Persistence::persist::<B>(vec![run_benchmark(benchmark)], device)
+    save::<B>("unary", vec![run_benchmark(benchmark)], device).unwrap();
```
louisfd, line 52: Name again here

louisfd (changes requested):
It's very good in general. The only issue I see is how often we must tell the name of the benchmark (see my comments for example on unary). Could it not be unified in one place?

nathanielsimard (changes requested):
Instead of adding the new field "operation," we could add a field "ID" instead. "Name" is more generic, and we may have some benchmarks that aren't about single operations. Or maybe we need another structure:

- name (Could have a default implementation `core::any::type_name::<Self>().to_string()`)
- shapes
- options  (num_repeat, kind, etc.)

The ID could be generated from the concatenation of all of the above.

backend-comparison/src/persistence/base.rs
syl20bnr: This is on purpose, I have a comment on it in the `seralize` function. This is the convention for MongoDB field names.

backend-comparison/src/persistence/base.rs
syl20bnr: I can add it back, my reasoning was that we can already sort by creation date but that's not as simple indeed.

backend-comparison/benches/binary.rs
syl20bnr: I was  thinking about how we will query the database, I would rather have an operator field than using string search in a "free form" name (with 5 bench we have already inconsistency).

We could let the name be the free form field for nice naming and add more precise fields.

backend-comparison/src/persistence/base.rs
syl20bnr: See comment: https://github.com/tracel-ai/burn/pull/1131#discussion_r1448991987

backend-comparison/src/persistence/base.rs
syl20bnr: Oh and we can theoretically overwrite a file as I wrote initially.

Also it's handy to have the name  of the benchmark.

syl20bnr:
> It's very good in general. The only issue I see is how often we must tell the name of the benchmark (see my comments for example on unary). Could it not be unified in one place?

You are right, I thought about it but I cannot fetch the binary name at runtime, the only references are in Cargo,  in the command line, and the source file name. I'll try to find something.

backend-comparison/src/persistence/base.rs
nathanielsimard: yup keep the name, juste replacing the uuid with a timestamp, users should never run multiple benchmarks at the same time, the benchmarks won't be valid in this case anyways.

syl20bnr:
Ready for another  round of review.
Updated gist: https://gist.github.com/syl20bnr/b0fdbbf0c3877b9e4ddf262a9cc388ac

Files on disk naming examples:

```
  bench_binary_1705007825641.json
  bench_from_data_1705007836909.json
  bench_gelu_reference_1705007872731.json
  bench_gelu_withcustomerf_1705007872987.json
  bench_gelu_withreferenceerf_1705007872929.json
  bench_matmul_1705007855627.json
  bench_to_data_1705007832378.json
  bench_unary_1705007818349.json
```

I added `num_repeats` also  to the documents,  I believe they can be useful.

backend-comparison/benches/custom_gelu.rs
```diff
@@ -24,7 +24,16 @@ impl<B: Backend, const D: usize> Benchmark for CustomGeluBenchmark<B, D> {
     type Args = Tensor<B, D>;
 
     fn name(&self) -> String {
-        format!("Gelu {:?}", self.kind)
+        let kindstr = format!("{:?}", self.kind);
+        format!("gelu_{}", kindstr.to_lowercase())
```
nathanielsimard, line 28: This should be "gelu", the kind is included in the options

nathanielsimard (changes requested):
I would not save the `num_repeat` in the database for now. I plan on eventually removing it since it is often confused with `num_samples`. It was useful when we didn't have a reliable way to sync a backend without actually reading the data. Repeating an execution meant that the data reading was less impactful.

nathanielsimard:
> I would not save the `num_repeat` in the database for now. I plan on eventually removing it since it is often confused with `num_samples`. It was useful when we didn't have a reliable way to sync a backend without actually reading the data. Repeating an execution meant that the data reading was less impactful.

@syl20bnr Though it might be a good idea to save the number of samples.

backend-comparison/benches/custom_gelu.rs
syl20bnr: Ahah, ok I was expecting this one, I hesitated to do it. I did this to get a better filename but that's more a cosmetic tweak than a useful one.

syl20bnr:
@nathanielsimard I already save the number of samples. See the gist.

I believe the repeat can still be useful, check the docstring I added to it, I think it is a good concept to have:

```
    /// Number of executions per sample
```

Say we have a very quick operation to bench, repeating it to get a meaningful sample can be useful.

syl20bnr:
Ideally the notion of repeat could be integrated to the main loop but I tried it and it was not as trivial to do.

syl20bnr:
We  decided to remove the num_repeat completely from the benchmakrs.

nathanielsimard (approved):
🎉

burn-common/src/benchmark.rs
```diff
@@ -12,27 +13,30 @@ use std::time::Instant;
...
-    /// Returns the median duration among all durations
-    pub fn median_duration(&self) -> Duration {
+    /// Returns a tuple of durations: (min, max, median)
+    fn min_max_median_durations(&self) -> (Duration, Duration, Duration) {
         let mut sorted = self.durations.clone();
         sorted.sort();
-        *sorted.get(sorted.len() / 2).unwrap()
+        let min = *sorted.first().unwrap();
+        let max = *sorted.last().unwrap();
+        let var = *sorted.get(sorted.len() / 2).unwrap();
```
louisfd, line 29: wrong variable name?

burn-common/src/benchmark.rs
```diff
@@ -167,10 +224,84 @@ where
...
+            durations: vec![
+                Duration::new(10, 0),
+                Duration::new(20, 0),
+                Duration::new(30, 0),
+                Duration::new(40, 0),
+                Duration::new(50, 0),
+            ],
+        };
+        let mean = durations.mean_duration();
+        assert_eq!(mean, Duration::from_secs(30));
```
louisfd, line 289: That's a bit picky, but with these durations the mean is equal to the median. Maybe change them a bit so we can be sure we're finding the mean and not the median. You could remove the duration 50 and the mean should be 25.

louisfd (changes requested):
That's very nice, I think we reached the beautiful code stage. 
I have only two minor requests and then it's good to go

burn-common/src/benchmark.rs
syl20bnr: Nice catch ! Thank you !

louisfd (approved):

PR #1133, opened 2024-01-11
Elementwise pow op

skewballfox:
This is still in a draft state, I'll fill out the checklist once it's further along

### Checklist

### Related Issues/PRs

### Changes

In progress addition of [pow function](https://github.com/onnx/onnx/blob/main/docs/Operators.md#pow). Assumes that the type conversion when rhs is differently typed than lhs/output, happens to rhs prior to the operation

### Testing

I've created the python files, but I haven't yet generated the ONNX models. If someone beats me to it, their is a slight difference in spec between pytorch and onnx regarding pow. for pytorch, it seems it always outputs a [float tensor](https://pytorch.org/docs/stable/generated/torch.pow.html), so the pow_int needs to convert the output post op


burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -458,6 +458,64 @@ pub trait IntTensorOps<B: Backend> {
...
+    ///
+    /// # Returns
+    ///
+    /// The elements of `lhs` raised to the value of `rhs`. Result is an IntTensor.
+    fn int_pow_float<const D: usize>(
+        lhs: IntTensor<B, D>,
+        rhs: FloatTensor<B, D>,
+    ) -> IntTensor<B, D> {
+        Self::int_pow(lhs, B::into_int(rhs))
+    }
```
nathanielsimard, line 500: It can cause precision issues. I think we should only support `pow` and `powf` where the `rhs` is always `f32`. So in the case of an int, we would have `int_pow` and `int_powf`, and in the case of a float, we would have `pow` and `powf`. `pow` is always an elementwise operation, and `powf` is always a scalar operation. What are your thoughts?

nathanielsimard (changes requested):
See my comment bellow

burn-tensor/src/tensor/ops/int_tensor.rs
skewballfox: sounds great, it actually makes it a bit easier to implement the backends, and that brings down the number of functions to implement from 4 to 3 per backend(powf already exist)

nathanielsimard: Hmm, maybe to be cleaner we might have all 4 variants:

- `powi` (Rhs = Int Tensor)
- `powf` (Rhs = Float Tensor)
- `powi_scalar` (Rhs = Int Scalar)
- `powf_scalar` (Rhs = Float Scalar)

We could provide default implementations for `powi` and `powi_scalar` where `powi => powf` and `powi_scalar => powf_scalar`. Using the float implementation by default is necessary if we want to keep the precision.

skewballfox: > Hmm, maybe to be cleaner we might have all 4 variants:
> 
> * `powi` (Rhs = Int Tensor)
> * `powf` (Rhs = Float Tensor)
> * `powi_scalar` (Rhs = Int Scalar)
> * `powf_scalar` (Rhs = Float Scalar)
> 
> We could provide default implementations for `powi` and `powi_scalar` where `powi => powf` and `powi_scalar => powf_scalar`. Using the float implementation by default is necessary if we want to keep the precision.

Sorry just saw this, the main problem is `powf` predates this PR, it already exist in burn. I can change the naming scheme, but that will break any downstream projects that are calling that function

skewballfox: > Hmm, maybe to be cleaner we might have all 4 variants:
> 
> * `powi` (Rhs = Int Tensor)
> * `powf` (Rhs = Float Tensor)
> * `powi_scalar` (Rhs = Int Scalar)
> * `powf_scalar` (Rhs = Float Scalar)
> 
> We could provide default implementations for `powi` and `powi_scalar` where `powi => powf` and `powi_scalar => powf_scalar`. Using the float implementation by default is necessary if we want to keep the precision.

A) do you want me to change the name (and maybe the signature) of the existing `powf` function to make it fit the convention
B) am I correct in understanding that in total we'll have 8 functions (`[int_]pow{i,f}[_scalar]`) but all but 2 will have default implementations like the below:
```rust
fn int_powi<const D: usize>(lhs: IntTensor<B, D>, rhs: IntTensor<B, D>) -> IntTensor<B, D> {
        B::into_int(B::powf(B::int_into_float(lhs), B::int_into_float(rhs)))
    }
```

nathanielsimard: A) Yes I prefer having a consistent API than avoid a simple breaking change at this stage.
B) Almost, I think we should only provide a default implementation for 4 variants:

```rust
fn int_powi<const D: usize>(lhs: IntTensor<B, D>, rhs: IntTensor<B, D>) -> IntTensor<B, D> {
        B::int_powf(lhs, B::int_into_float(rhs)))
}
fn int_powi_scalar<const D: usize>(lhs: IntTensor<B, D>, rhs: IntElem<B>) -> IntTensor<B, D> {
        B::int_powf_scalar(lhs,  rhs.elem()))
}
fn powi<const D: usize>(lhs: FloatTensor<B, D>, rhs: IntTensor<B, D>) -> FloatTensor<B, D> {
        B::powf(lhs, B::int_into_float(rhs)))
}
fn powi_scalar<const D: usize>(lhs: FloatTensor<B, D>, rhs: IntElemr<B, D>) -> FloatTensor<B, D> {
        B::int_powf_scalar(lhs,  rhs.elem()))
}
 ```

Comment (skewballfox) [reply burn-tensor/src/tensor/ops/int_tensor.rs]: just saw this, I'll implement these changes in the next commit, might be tomorrow though

Comment (skewballfox) [reply burn-tensor/src/tensor/ops/int_tensor.rs]: > A) Yes I prefer having a consistent API than avoid a simple breaking change at this stage. B) Almost, I think we should only provide a default implementation for 4 variants:
> 
> ```rust
> fn int_powi<const D: usize>(lhs: IntTensor<B, D>, rhs: IntTensor<B, D>) -> IntTensor<B, D> {
>         B::int_powf(lhs, B::int_into_float(rhs)))
> }
> fn int_powi_scalar<const D: usize>(lhs: IntTensor<B, D>, rhs: IntElem<B>) -> IntTensor<B, D> {
>         B::int_powf_scalar(lhs,  rhs.elem()))
> }
> fn powi<const D: usize>(lhs: FloatTensor<B, D>, rhs: IntTensor<B, D>) -> FloatTensor<B, D> {
>         B::powf(lhs, B::int_into_float(rhs)))
> }
> fn powi_scalar<const D: usize>(lhs: FloatTensor<B, D>, rhs: IntElemr<B, D>) -> FloatTensor<B, D> {
>         B::int_powf_scalar(lhs,  rhs.elem()))
> }
> ```

won't the int_powf still need to call the regular `tensor::powf`? It does involve two conversions RHS to tensor, and output to Int, but if `powi` is just calling `powf`, then those conversions will be necessary unless the backend has a `mapv` iterator like ndarray

Comment (skewballfox) [reply burn-tensor/src/tensor/ops/int_tensor.rs]: the original request has been resolved.

File: burn-tch/src/ops/base.rs [hunk 2]
```diff
@@ -439,4 +439,16 @@ impl<E: tch::kind::Element + Copy + Default> TchOps<E> {
...
     }
+    pub fn pow<const D: usize>(
+        tensor: TchTensor<E, D>,
+        exponent: TchTensor<E, D>,
+    ) -> TchTensor<E, D> {
+        TchTensor::binary_ops_tensor(
+            tensor,
+            exponent,
+            |lhs, rhs| lhs.f_pow_tensor_(rhs).unwrap(),
+            |lhs, rhs| rhs.f_pow_tensor_(lhs).unwrap(),
```
nathanielsimard, line 450: I don't think it's equivalent, we should use the readonly version in this case.

burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -1589,6 +1589,50 @@ where
     /// For calculating abs of the elements of a tensor, users should prefer the [Tensor::abs](Tensor::abs) function,
     /// which is more high-level and designed for public use.
     fn abs<const D: usize>(tensor: Self::Primitive<D>) -> Self::Primitive<D>;
+
```
nathanielsimard, line 1592: In this file, you should actually add the user API operation implementations where the static dispatch is done. This isn't inside a trait, but the implementation uses the Kind to dispatch to the right implementation.

Example with powf_scalar:

```rust
    /// Applies element wise powf operation with a scalar.
    ///
    /// `y = x + s`
    pub fn powf_scalar<E: ElementConversion>(self, other: E) -> Self {
        Self::new(K::powf_scalar(self.primitive, other))
    }
```

burn-tch/src/ops/base.rs
skewballfox: I think I resolved it, but just to clarify, you mean switching to `f_pow` (which takes `&self` rather than `&mut self`) right?

burn-autodiff/src/utils.rs
```diff
@@ -22,3 +22,24 @@ pub fn duplicate<T: Clone + std::fmt::Debug, const N: usize>(
...
+//     nodes
+//         .iter()
+//         .flat_map(|node| match node {
+//             Some(_) => [obj.clone(), obj.clone()],
+//             None => None,
+//         })
+//         .collect::<Vec<_>>()
+//         .try_into()
+//         .unwrap()
+// }
```
nathanielsimard, line 45: Dead code, we should remove it.

burn-import/src/onnx/to_burn.rs
```diff
@@ -663,6 +664,25 @@ impl ONNXGraph {
         let output = node.outputs.first().unwrap().to_type();
         UnaryNode::neg(input, output)
     }
+    fn pow_conversion(node: Node) -> BinaryNode {
+        let lhs = node.inputs.first().unwrap().to_type();
+        let rhs = node.inputs.get(1).unwrap().to_type();
+        //let ty = lhs.ty();
```
nathanielsimard, line 670: I guess it's dead code?

burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -1589,6 +1609,46 @@ where
...
+    ///
+    /// # Arguments
+    /// * `tensor` - The tensor to apply power to.
+    /// * `power` - The power to apply to the tensor.
+    fn powi_scalar<const D: usize, E: ElementConversion>(
+        lhs: Self::Primitive<D>,
+        rhs: E,
+    ) -> Self::Primitive<D>; //{
+                             //     Self::powf_scalar(lhs, rhs)
+                             // }
```
nathanielsimard, line 1651: Dead code

burn-autodiff/src/tests/pow.rs
```diff
@@ -26,4 +26,35 @@ mod tests {
...
+        let tensor_3 = tensor_1.clone().powf(tensor_2.clone());
+        let grads = tensor_3.backward();
+
+        let grad_1 = tensor_1.grad(&grads).unwrap();
+        let grad_2 = tensor_2.grad(&grads).unwrap();
+        //[4 * 2**3, 2 * 7**1]
+        grad_1
+            .to_data()
+            .assert_approx_eq(&Data::from([32.0, 14.0]), 3);
+        //assert_eq!(grad_1.to_data(), data_2);
```
nathanielsimard, line 48: dead code.

nathanielsimard (changes requested):
Thanks a lot for this PR. Just some minor comments, but should be good to go pretty soon. Tagging @louisfd  to review as well.

louisfd (approved):
I double-checked the backward pass and it's well done!

nathanielsimard (approved):

Issue #1117, opened 2024-01-04
Support multiple optimization streams per device


nathanielsimard:
For now, Burn-fusion supports only one optimization stream for a device. However, when a device is used by multiple concurrent threads, the stream becomes interrupted all the time, and a lot of optimizations become impossible. Also, a lot more optimizations are created, but never reused because new situations happen nondeterministically.

PR #1135, opened 2024-01-12
Refactor/fusion stream

nathanielsimard:
This is a major refactor but doesn't change the behavior. It's quite hard to add more unit tests, but we will have to come up with a better testing strategy in the future. For now, the whole tensor and autodiff test suites are running on the fused backend, so everything is still pretty much covered by tests, but when one of many tests fail, it's hard to know why.

## Changes

- I had to change the way we handle errors for `one_hot`, since relying on panics _inside_ a backend operation is very bad, since it may corrupt the backend server, which might make all following tests fail. (Creating flaky tests)
- I created the `MultiStream` type, but it still does nothing interesting. This is where we will have to handle multiple concurrent streams. #1117
- I renamed `graph` to `stream` since the whole technique is about optimizing streams and not graphs.
- Instead of having a cache with everything, we now have an optimization store, where we just store the optimizations. The added policy handles what optimization can be reused based on a stream. It's way less confusing, separating the global optimization state that can be serialized from the policy's state, which tracks each new operation added to a stream.
- I also splitted the `GraphExecution` into two different types: `Processor` and `Explorer`. Since the `GraphExecution` didn't even execute the streams, I think it was quite a bad name. What it does is simply following a policy, which is either exploring/building new optimizations, deferring any execution, or actually executing an optimization. The execution is handled by the `Stream`, and the exploration by the `Explorer`, making the processor only responsible for handling the recurring logic and keeping the state up to date.


burn-fusion/src/stream/execution/processor.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        // Check if an optimization is available for this stream before creating a new opitmization.
+        //
+        // Specify a sync execution mode signaling that we want to know if an optimization is
+        // available right now even if it isn't the best one.
+        match policy.action(store, &stream_relative, ExecutionMode::Sync) {
+            Action::Execute(id) => {
+                // When we are in lazy mode, a next operation will be available.
+                //
+                // Since we are adding new opitmization only when the policy action is explore, we
```
louisfd, line 154: opitmization will fail typos check

louisfd (approved):
Wow it's starting to be really digestible. I like the store concept. 
You'll find my comments boring I think, it's mostly remaining occurrences of the word `graph`. I think we should remove them all to avoid confusion

Issue #944, opened 2023-11-10
Autotune: support int/bool ops


louisfd:
In autotune, we create random tensors as inputs. Since random only works with floats it makes it impossible to use autotune with int and bool.

Lazy solution: revert to creating ones tensor instead of random
Hacky solution: Run the random kernel on f32 instead of E then cast to int or bool using some
Long solution: Add support at large for bool random (bernoulli) and int random (uniform)

Once this is done we can add a call to autotune for sum_dim and mean_dim (int versions)

agelas:
@antimora I think I'll have some more free time coming up, I can tentatively take this.

antimora:
Int op autotuning landed in #1136, and autotune now lives in CubeCL. Closing.

PR #1136, opened 2024-01-12
Feat/autotune int ops

agelas:
### Checklist
WIP

### Related Issues/PRs

#944 
### Changes

Added int_random() to int tensor ops, new autotuning calls for sum_dim and mean_dim.
### Testing
Unit tests for int_random(), sum_dim() and mean_dim() on WgpuTensors with IntElements .


agelas:
@louisfd Finally got the chance to start on #944. I'm going down the long route and trying to add support at large for `random()` on int tensors. When you say "add a call to autotune for sum_dim and mean_dim (int versions)", I'm assuming that means creating an `impl<E: IntElement, const D: usize> SumDimAutotuneOperationSet<E, D>` and associated `AutotuneOperationSet` impl (and similar story for mean_dim)?

louisfd:
> @louisfd Finally got the chance to start on #944. I'm going down the long route and trying to add support at large for `random()` on int tensors. When you say "add a call to autotune for sum_dim and mean_dim (int versions)", I'm assuming that means creating an `impl<E: IntElement, const D: usize> SumDimAutotuneOperationSet<E, D>` and associated `AutotuneOperationSet` impl (and similar story for mean_dim)?

Hi @agelas, thanks for tackling this issue! 
The only place that was problematic was these lines in `burn-wgpu/src/kernel/reduce/tune/sum_dim.rs`

```rust
fn autotunables(&self) -> Vec<Box<dyn AutotuneOperation>> {
    let random_bounds: (E, E) = ((-10.0).elem::<E>(), (10.0).elem::<E>());
    let input = random_like_uniform(&self.input, random_bounds.0, random_bounds.1);
```

because random_like_uniform only worked with floats even if E was an int. I'm not sure if you could just rewrite those lines to make it call your new int ops instead of the lower level random_like_uniform or if you'll have trouble determining if you need to call the float or int version. If that's the case then yes you can duplicate the AutotuneOperationSet, but it should keep the same key. 

In the end you will only have to change sum_dim / mean_dim implementations in `burn-wgpu/src/ops/int_ops.rs` to something very similar to the float ones:
```rust
    fn sum_dim<const D: usize>(tensor: FloatTensor<Self, D>, dim: usize) -> FloatTensor<Self, D> {
        #[cfg(feature = "autotune")]
        {
            reduce::sum_dim_autotune(tensor, dim)
        }

        {
            let output = init_reduce_output(&tensor, dim);
            reduce::sum_dim(tensor, output, dim)
        }
    }
```

agelas:
@louisfd I'm gonna take a breather and ask for feedback here because this PR is starting to get on the large size. I think everything is working just about the way it should. I had to unfortunately basically copy the whole `AutotuneOperationSet` and associated functions/impls since the WgpuElement vs IntElement trait bound can't be determined at runtime. 

Also I'm not sure what's up with the `fusion::base::tests::aggregation::tests::test_should_mean_last_dim_int`, it worked on the prior commit and locally and the only thing I've changed is the book docs.

burn-fusion/src/stream/operation.rs
```diff
@@ -378,6 +378,11 @@ pub enum NumericOperationDescription<E> {
     /// Float => [clamp](burn_tensor::ops::TensorOps::clamp).
     /// Int => [clamp](burn_tensor::ops::IntTensorOps::int_clamp).
     Clamp(ClampOperationDescription<E>),
+    /// Operation corresponding to:
+    ///
+    /// Float => [random](burn_tensor::ops::TensorOps::random).
```
louisfd, line 383: It only corresponds to int_random, the float version is handled elsewhere

burn-ndarray/src/ops/int_tensor.rs
```diff
@@ -377,4 +378,20 @@ impl<E: FloatNdArrayElement> IntTensorOps<Self> for NdArray<E> {
...
+        distribution: Distribution,
+        device: &NdArrayDevice,
+    ) -> NdArrayTensor<i64, D> {
+        let mut seed = SEED.lock().unwrap();
+        let mut rng = if let Some(rng_seeded) = seed.as_ref() {
+            rng_seeded.clone()
+        } else {
+            get_seeded_rng()
+        };
+        let tensor = Self::int_from_data(Data::random(shape, distribution, &mut rng), device);
```
louisfd, line 393: Not sure what default distribution will do here? I suppose it will give floats in [0,1[ then floor them all to zero?

burn-candle/src/ops/int_tensor.rs
```diff
@@ -376,4 +376,38 @@ impl<F: FloatCandleElement, I: IntCandleElement> IntTensorOps<Self> for Candle<F
...
+    fn int_random<const D: usize>(
+        shape: Shape<D>,
+        distribution: Distribution,
+        device: &Device<Self>,
+    ) -> IntTensor<Self, D> {
+        let shape = &shape.dims;
+        let device = &(*device).into();
+        match distribution {
+            Distribution::Default => CandleTensor::new(
+                candle_core::Tensor::rand(0.elem::<F>(), 100.elem::<F>(), shape, device)
```
louisfd, line 389: [0, 100] is a bit arbitrary

burn-tch/src/ops/int_tensor.rs
```diff
@@ -418,4 +418,33 @@ impl<E: TchElement> IntTensorOps<Self> for LibTorch<E> {
...
+    fn int_random<const D: usize>(
+        shape: Shape<D>,
+        distribution: Distribution,
+        device: &LibTorchDevice,
+    ) -> TchTensor<i64, D> {
+        match distribution {
+            Distribution::Default => {
+                let mut tensor = TchTensor::<i64, D>::empty(shape, *device);
+                tensor
+                    .mut_ops(|tensor| tensor.rand_like_out(tensor))
```
louisfd, line 431: Same comment as for ndarray

burn-wgpu/src/ops/int_ops.rs
```diff
@@ -321,4 +333,25 @@ where
...
         kernel::repeat(tensor, dim, times)
     }
+
+    fn int_random<const D: usize>(
+        shape: Shape<D>,
+        distribution: Distribution,
+        device: &Device<Self>,
+    ) -> IntTensor<Self, D> {
+        let float_tensor = match distribution {
+            Distribution::Default => random_uniform::<G, F, D>(shape, device, 0.elem(), 10.elem()),
```
louisfd, line 343: Arbitrary again

louisfd (changes requested):
Nice work @agelas! 
The autotune part looks good, it will need refactor eventually but this will be a different issue (#1175)
The main concern I find is the Default Distribution. First, it must absolutely be the same configuration in all backends. At the moment it differs in every place. Second, we're not even sure what should be a default distribution in an integer setting. I think if we're going to output something arbitrary we might as well reject default distributions altogether (with a panic) and force the user to decide its bounds with Uniform. Or, we mimic the [0,1[ behaviour of floats but with just 0s and 1s which would be equivalent to a Bernoulli(0.5). What do you think?

agelas:
@louisfd Yes sorry, I kept changing the distributions around on the backends for testing purposes and forgot to make them all the same. I think it's nice to have a default just in case, and the flexibility to choose your own distribution is provided anyways without much extra effort . What the bounds should be though I'm not sure. If we mimic the [0,1] behavior of floats like you mentioned it would be pretty easy to port this over to bool tensors so I sort of lean towards that approach.

agelas:
@louisfd actually on second thought, maybe it would be best for the default distribution to be [0, 255]- I think that's the standard range for 8-bit quantization and would also be good for most image data as well, so that covers a lot of use cases right there.

louisfd (changes requested):
There are some small problems remaining, but overall looks good!

burn-wgpu/src/kernel/reduce/tune/mean_dim.rs
```diff
@@ -109,8 +113,110 @@ pub fn mean_dim_autotune<E: WgpuElement + Element, const D: usize>(
...
+        let input = random_like_uniform_int(&self.input, random_bounds.0, random_bounds.1);
+
+        let output = empty_device(
+            self.output.client.clone(),
+            self.output.device.clone(),
+            self.output.shape.clone(),
+        );
+
+        vec![
+            Box::new(MeanDimAutotune::<I, D>::new(
```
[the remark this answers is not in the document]
agelas: Whoops- good catch thanks!

burn-wgpu/src/template/prng/uniform_int_inner_loop.wgsl
```diff
@@ -0,0 +1,18 @@
...
+    state[1u] = taus_step_1(state[1u]);
+    state[2u] = taus_step_2(state[2u]);
+    state[3u] = lcg_step(state[3u]);
+    let random_u32 = state[0u] ^ state[1u] ^ state[2u] ^ state[3u];
+
+    // Modulus operation to fit within the range
+    let random_in_range = (random_u32 % range) + low;
+
+    output[write_index] = random_in_range;
+}
```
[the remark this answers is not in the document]
agelas: @louisfd is this shader being invoked for the test that's failing? The problem on CI was a fusion test `fusion::base::tests::aggregation::tests::test_should_mean_last_dim_int`, that used to work at commit [ae60d25](https://github.com/tracel-ai/burn/pull/1136/commits/ae60d257f63163e1cefc48cd0e67e8b81309c199) but then stopped working at [dca5b26](https://github.com/tracel-ai/burn/pull/1136/commits/dca5b26ff6889996ed846058cf134c4933352887) for some reason. It works locally for me though.

Now it seems the CI is complaining about something else though.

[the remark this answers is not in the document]
agelas: @louisfd Ok I'm back to failing on `test_should_mean_last_dim_int()` in the aggregation tests for burn-tensor. For both that and `test_should_sum_last_dim_int()`, shouldn't we be calling the `int_mean_dim` and `int_sum_dim()` though? Still not sure why `cast_float()` is causing issues though especially since I never use it, and if these tests use the base tensor functions, neither should even be calling anything from this PR.

```rust
#[test]
    fn test_should_mean_last_dim_int() {
        let tensor = TestTensorInt::from([[0, 1, 2], [3, 4, 5]]);

        let data_actual = tensor.mean_dim(1).to_data();

        assert_eq!(data_actual, Data::from([[1], [4]]));
    }

    #[test]
    fn test_should_sum_last_dim_int() {
        let tensor = TestTensorInt::from([[0, 1, 2], [3, 4, 5]]);

        let data_actual = tensor.sum_dim(1).to_data();

        assert_eq!(data_actual, Data::from([[3], [12]]));
    }
```

nathanielsimard:
The casting problem is quite obvious in `cast_float`.
This function is used in the prng shader.

```wgsl
fn cast_float(number: u32) -> {{ elem }} {
   return 2.3283064365387e-10 * {{ elem }}(number);
}
```

I think it should be :

```wgsl
fn cast_elem(number: u32) -> {{ elem }} {
   let tmp = 2.3283064365387e-10 * f32(number);
   return {{ elem }}(tmp);
}
```

It was assume that `{{ elem }}` was a float, but with this PR, it becomes an int `i32`. @agelas @louisfd

agelas:
@nathanielsimard In `burn-wgpu/src/kernel/prng/bernoulli.rs`, it looks like there's already a `cast_elem` function, so maybe `convert_elem` or something similar?

@louisfd Also, for the test that's failing, is there a way to switch to use `int_mean_dim(1)` instead of `mean_dim(1)`? I think that and the next test in the `aggregation.rs` file (`test_should_sim_last_dim_int()`) now have int versions that should probably be used.

```rust
#[test]
    fn test_should_mean_last_dim_int() {
        let tensor = TestTensorInt::from([[0, 1, 2], [3, 4, 5]]);

        let data_actual = tensor.mean_dim(1).to_data(); // <- use int_mean_dim(1) instead?

        assert_eq!(data_actual, Data::from([[1], [4]]));
    }
```

louisfd:
> @nathanielsimard In `burn-wgpu/src/kernel/prng/bernoulli.rs`, it looks like there's already a `cast_elem` function, so maybe `convert_elem` or something similar?
> 
> @louisfd Also, for the test that's failing, is there a way to switch to use `int_mean_dim(1)` instead of `mean_dim(1)`? I think that and the next test in the `aggregation.rs` file (`test_should_sim_last_dim_int()`) now have int versions that should probably be used.
> 
> ```rust
> #[test]
>     fn test_should_mean_last_dim_int() {
>         let tensor = TestTensorInt::from([[0, 1, 2], [3, 4, 5]]);
> 
>         let data_actual = tensor.mean_dim(1).to_data(); // <- use int_mean_dim(1) instead?
> 
>         assert_eq!(data_actual, Data::from([[1], [4]]));
>     }
> ```

Hi @agelas Sorry for the late reply. 

Regarding the cast functions, since it's getting a bit confusing I think we should be more explicit with the names:
`cast_X_to_Y` with X and Y changed to bool, float or elem when it depends. 

For your second question, maybe you're confusing tensor methods and backend operations? Because it's an int tensor the mean_dim method will call int_mean_dim underneath. 

```rust
impl<B: Backend> Numeric<B> for Int {
    // ...

    fn mean_dim<const D: usize>(tensor: Self::Primitive<D>, dim: usize) -> Self::Primitive<D> {
        B::int_mean_dim(tensor, dim)
    }

    // ...
}
```

For the CI error, it seems to be the modulo. Are you positive that both random_u32 and range are the same type?
```
  48 │     let random_in_range = (random_u32 % range) + low;
     │                                              ^^^^^^^^^^^^^^^^^^ naga::Expression [93]
  ```

agelas:
Hi @antimora @louisfd , sorry about the delay, life got in the way.

@louisfd  w/regards to the naming issue, I changed `cast_float` to `cast_u32_to_float` like you suggested. For the CI error, it seemed to be with the `KernelSettings`, but that's resolved now.

The CI is failing on codecov/path now at 63455c4 (the latest commit), I'm not sure what to do here since a lot of the code here gets used via a somewhat complicated mixture of macros and regular rust,

antimora:
@agelas, thank you for the update. Do not worry, the life is a priority. 

Regarding the coverage - it's not a hard requirement to hit 80%. We can override manually.

louisfd (approved):
LGTM
Many thanks 😄

PR #1137, opened 2024-01-12
Fix: Records no longer require Debug and Clone

nathanielsimard:
`Clone` and `Debug` are never used on `Record::Item` but were generated within the derive macro. It made it impossible to create a Record generic over another Record.


antimora:
Can you please give more context here? Just want to make sure my current deserialization won't break.

syl20bnr (approved):

antimora:
This broke my tests:

image


image

Clone is needed if item is behind vector.

PR #1138, opened 2024-01-12
Add fan_out to `Conv2dConfig::init`

laggui:
When using a custom `Initializer` with `fan_out_only: true`, the `Conv2dConfig::init()` panics because we only provide a value for `fan_in`.

```rust
let init = Initializer::KaimingUniform {
    gain: 1.0 / sqrt(3.0),
    fan_out_only: true,
};
let config = Conv2dConfig::new([5, 1], [5, 5]).with_initializer(init.clone());
let _ = config.init::<TestBackend>(&device);
```

This example results in:
```
thread 'main' panicked at burn-core/src/nn/initializer.rs:132:23:
Can't use Kaiming initialization without specifying fan. Use init_with method.
```

### Checklist

### Changes

- Added `fan_out` calculation to `Conv2dConfig::init()` to handle custom `Initializer` w/ `fan_out_only` set to `true`

### Testing

- Added simple `initializer_fan_out()` test to make sure `fan_out` is passed to `Initializer::init_with()` (previously panicked)

### DIFF
--- a/burn-core/src/nn/conv/conv2d.rs
+++ b/burn-core/src/nn/conv/conv2d.rs
@@ -70,17 +70,22 @@ impl Conv2dConfig {
             self.kernel_size[1],
         ];
 
-        let fan_in = self.channels[0] / self.groups * self.kernel_size.iter().product::<usize>();
+        let k = self.kernel_size.iter().product::<usize>() / self.groups;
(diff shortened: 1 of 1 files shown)


louisfd (approved):
Thanks a lot for the fix 😄

PR #1140, opened 2024-01-13
fix the problem of sigmoid gradient generating NaN

### LINKED ISSUE #1139: Small negative values cause the gradient of `sigmoid` to become `NaN`
Here is the code:
```rust
use burn::{
    backend::{Autodiff, NdArray},
    tensor::{activation, Data, Tensor},
};

fn main() {
    let data = Data::<f32, 1>::from([-90.0]);

    let tensor_1 = Tensor::<Autodiff<NdArray>, 1>::from_data(data, &device).require_grad();

    let tensor_2 = activation::sigmoid(tensor_1.clone());
    let grads = tensor_2.backward();

    println!("{}", grad_1);
}
```
The result is `NaN`.

wcshds:
### Checklist

### Related Issues/PRs

fix #1139 

### Changes

Use sigmoid's derivative formula directly to avoid differentiating `log` and `exp` in autodiff.

### DIFF
--- a/burn-tensor/src/tensor/activation/base.rs
+++ b/burn-tensor/src/tensor/activation/base.rs
@@ -78,7 +78,14 @@ pub fn log_softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize)
 
 /// Applies the sigmoid function.
 pub fn sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
-    log_sigmoid(tensor).exp()
+    match B::FloatElem::precision() {
+        Precision::Half => {
+            let tensor_full = tensor.to_full_precision();
+            let tensor_tmp = tensor_full.sigmoid();
+            Tensor::from_full_precision(tensor_tmp)
+        }
+        _ => tensor.sigmoid(),
+    }
 }
 
 /// Applies the log sigmoid function.
--- a/burn-tensor/src/tensor/activation/base.rs
+++ b/burn-tensor/src/tensor/activation/base.rs
@@ -78,14 +78,7 @@ pub fn log_softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize)
 
 /// Applies the sigmoid function.
 pub fn sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
-    match B::FloatElem::precision() {
-        Precision::Half => {
-            let tensor_full = tensor.to_full_precision();
-            let tensor_tmp = tensor_full.sigmoid();
-            Tensor::from_full_precision(tensor_tmp)
-        }
-        _ => tensor.sigmoid(),
-    }
+    Tensor::from_primitive(B::sigmoid(tensor.primitive))
 }
 
 /// Applies the log sigmoid function.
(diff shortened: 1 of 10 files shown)


burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -926,6 +926,22 @@ pub trait TensorOps<B: Backend> {
...
+    ///
+    /// # Returns
+    ///
+    /// A tensor with the same shape as `tensor` with sigmoid values.
+    fn sigmoid<const D: usize>(tensor: FloatTensor<B, D>) -> FloatTensor<B, D> {
+        B::exp(B::neg(B::log(B::add_scalar(
+            B::exp(B::neg(tensor)),
+            1.0_f32.elem(),
+        ))))
+    }
```
nathanielsimard, line 943: We can add this function in `burn-tensor/src/tensor/ops/activation.rs` instead.

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -78,7 +78,14 @@ pub fn log_softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize)
...
 pub fn sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
-    log_sigmoid(tensor).exp()
+    match B::FloatElem::precision() {
+        Precision::Half => {
+            let tensor_full = tensor.to_full_precision();
+            let tensor_tmp = tensor_full.sigmoid();
+            Tensor::from_full_precision(tensor_tmp)
+        }
+        _ => tensor.sigmoid(),
+    }
```
nathanielsimard, line 88: I don't think we need full precision here, as it is now declared as a method in the backend. The backend implementations can choose to use full precision regardless of the circumstances. Perhaps we can consider incorporating full precision into the default implementation.

nathanielsimard (changes requested):

burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -102,4 +103,43 @@ pub trait ActivationOps<B: Backend> {
...
+    fn sigmoid<const D: usize>(tensor: FloatTensor<B, D>) -> FloatTensor<B, D> {
+        let tensor_full = B::to_full_precision(&tensor);
+        let tensor_tmp = B::FullPrecisionBackend::exp(B::FullPrecisionBackend::neg(
+            B::FullPrecisionBackend::log(B::FullPrecisionBackend::add_scalar(
+                B::FullPrecisionBackend::exp(B::FullPrecisionBackend::neg(tensor_full)),
+                1.0.elem(),
+            )),
+        ));
+
+        B::from_full_precision(tensor_tmp)
```
nathanielsimard, line 125: Not sure @louisfd if there is a more numerically stable implementation possible here.

louisfd (changes requested):
Hi, can you check my comment about the naming of the argument to backward, afterwards it will be ready

burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -102,4 +103,43 @@ pub trait ActivationOps<B: Backend> {
...
+    /// # Arguments
+    ///
+    /// * `x` - The tensor.
+    /// * `grad` - The gradient.
+    ///
+    /// # Returns
+    ///
+    /// The output tensor.
+    fn sigmoid_backward<const D: usize>(
+        x: FloatTensor<B, D>,
```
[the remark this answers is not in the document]
wcshds: Done.

louisfd (approved):
Thanks a lot

PR #1143, opened 2024-01-15
[burn-fusion] save all execution plans for any trigger

nathanielsimard:
This is another significant refactor of burn-fusion, marking the second one in the series 😅.

Finally, I have discovered a method to effectively write tests that validate the correct functionalities. While it is essentially one large test, it primarily aims to ensure that all cases can be queued sequentially, maintaining a consistently proper state—a genuine way to thoroughly test the code. It's a beginning, but testing is straightforward, given that most "algorithm" structures are generic over the optimization type, not the fusion backend type. Consequently, each component can be tested with ease. More tests will be added in the future, but at least there's a manageable starting point for now.

I have also opted to replace many abbreviations with their full names, as I believe it enhances overall readability.

It's not a "pure" refactor like the last PR #1135; instead, some bugs/edge cases have been addressed and simplified:
- Previously, we consistently executed operations one step behind, even when it was clear that the optimal approach was to execute an action. This behavior was a result of complex code rather than an intentional design choice.
- There are now fewer edge cases associated with Async and Lazy executions.
- We no longer solely save optimizations in the store; instead, we save all  _execution plans_, which also include executing a series of operations. The format in which this information is saved is very different, but also very intuitive.


louisfd (approved):
Awesome. Not much to say but maybe check my comments in the test for my understanding

PR #1146, opened 2024-01-17
Add tests: Slice assign vs Cat in LSTM backward

### LINKED ISSUE #1098: After using `slice_assign`, gradient descent cannot track the model parameters
I found that after using `slice_assign` in the loss function, gradient descent cannot track the model parameters. I believe this is the main reason why the loss becomes NaN after the first iteration when I apply my [implementation of CTC loss](https://github.com/tracel-ai/burn/pull/1049) to the [CRNN model](https://github.com/wcshds/crnn-cjk).

```rust
pub fn forward(&self, input: Tensor<B, 3>) -> Tensor<B, 1> {
    let device = input.device();
    let [d1, d2, d3] = input.dims();

    let input2 = Tensor::empty(input.shape(), &device);
    input2
        .clone()
        .slice_assign([0..d1, 0..d2, 0..d3], input.clone());

    input2.mean()
}
```

louisfd:
In order to fix #1098, where it is mentioned that LSTM backward does not work in its current state but would if slice_assign were replaced by cat in its computations, I wrote some tests: one for slice_assign backward, and one specifically for LSTM backward, both comparing the result to the one we get with cat. 
For both tests it worked fine, making me unsure if there is really a bug with slice_assign.


wcshds:
I don't know if this is relevant to the problem. If use `let fake_loss = hidden_state;` instead of `let fake_loss = cell_state + hidden_state;`, the lstm which uses `slice_assign` will panic.

wcshds:
I think slice_assign does have some bugs. If you change the return value of the current lstm implementation to `(batched_cell_state.clone()), batched_hidden_state + batched_cell_state.clone() - batched_cell_state.clone ())`, the parameters of lstm can be updated.

nathanielsimard:
> I don't know if this is relevant to the problem. If use `let fake_loss = hidden_state;` instead of `let fake_loss = cell_state + hidden_state;`, the lstm which uses `slice_assign` will panic.

I think this is expected, since you only track gradients for tensors that require them or are connected to some. If you only use the hidden state, then maybe there is no connection to the actual parameters of the model.

burn-core/src/nn/rnn/lstm.rs
```diff
@@ -349,4 +349,103 @@ mod tests {
         assert_eq!(cell_state.shape().dims, [8, 10, 1024]);
         assert_eq!(hidden_state.shape().dims, [8, 10, 1024]);
     }
+
+    #[test]
+    fn test_batched_backward_pass() {
```
nathanielsimard, line 354: I think we should cleanup that test, like a pretty standard backward pass test, without having to use the cat function. Just checking if the gradients are calculated can be enough here.

nathanielsimard (changes requested):
See my comments bellow

wcshds:
> > I don't know if this is relevant to the problem. If use `let fake_loss = hidden_state;` instead of `let fake_loss = cell_state + hidden_state;`, the lstm which uses `slice_assign` will panic.
> 
> I think this is expected, since you only track gradients for tensors that require them or are connected to some. If you only use the hidden state, then maybe there is no connection to the actual parameters of the model.

But the final output in lstm is a tensor that contains the hidden state of each time step, which is exactly what I need, and also the output of pytorch. Cell state only plays a role in the lstm calculation process.

louisfd:
@nathanielsimard twice already I thought I had pushed my change but I just realized this PR is from my personal branch 😢 now it's there

nathanielsimard (approved):

nathanielsimard:
@louisfd I think you need to merge main and run clippy & fmt.

PR #1147, opened 2024-01-18
Support generic module

nathanielsimard:
Now we support generic modules:

```rust
#[derive(Module, Debug)]
struct MyModule<B: Backend, F: ForwardTrait> {
   module: F,
   fc1: nn::Linear<B>,
}
```

I also refactored a bit of code to make it a bit clearer, paving the way to support enum modules.


laggui:
Maybe just double check for the clippy suggestions, otherwise LGTM!
<sub>(granted, I have no experience with the intricacies of `derive` for codegen)</sub>

I validated my example for the custom resnet18 implementation with a generic residual block and it works. Great job 😄

laggui (approved):

PR #1149, opened 2024-01-18
[burn-fusion] support multiple concurrent optimization streams

### LINKED ISSUE #1117: Support multiple optimization streams per device

nathanielsimard:
Fix #1117


louisfd (approved):
LGTM

Issue #1148, opened 2024-01-18
Add an Example demonstrating a Regression problem


ashdtu:
### Feature description

Add an Example demonstrating a Regression problem

### Feature motivation

Burn examples contain classification problems but no regression example. It would be helpful to have an example demonstrating regression on a simple toy dataset.

onurcanbektas:
+1

ashdtu:
Addressed in PR #1150 :) Closing.

PR #1150, opened 2024-01-18
Example/regression

ashdtu:
### Checklist

### Related Issues/PRs

#1148 

### Changes

Create an example to demonstrate how to do Regression problem using Burn framework. The example uses Toy-diabetes dataset present as a standard dataset in sklearn [datasets](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_diabetes.html). The dataset is also available for download from Hugging Face hub [dataset](https://huggingface.co/datasets/Jayabalambika/toy-diabetes). 

### Testing
* The dataset was split into train/test split and the MSE Loss for both splits were logged. No overfitting in the given model.
* ./run-checks.sh all was executed to check for compilation validity in example. No separate unit tests written.


examples/simple-regression/src/dataset.rs
```diff
@@ -0,0 +1,162 @@
...
+        Self::new("train")
+    }
+
+    pub fn test() -> Self {
+        Self::new("test")
+    }
+
+    // pub fn get_full_data() -> SqliteDataset<DiabetesItem> {
+    //
+    // }
```
nathanielsimard, line 85: Dead code => should be removed.

examples/simple-regression/src/dataset.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl<B: Backend> DiabetesBatcher<B> {
+    pub fn new(device: B::Device) -> Self {
+        Self { device }
+    }
+}
+
+impl<B: Backend> Batcher<DiabetesItem, DiabetesBatch<B>> for DiabetesBatcher<B> {
+    fn batch(&self, items: Vec<DiabetesItem>) -> DiabetesBatch<B> {
+
```
nathanielsimard, line 133: Just a preference to remove empty lines at the beginning of functions :).

examples/simple-regression/src/dataset.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        let targets = items
+            .iter()
+            .map(|item| Tensor::<B, 1>::from_floats([(item.response as f32)], &self.device))
+            .collect();
+
+        let targets  = Tensor::cat(targets, 0);
+        let targets = (targets.clone() - targets.clone().min_dim(0))
+            .div(targets.clone().max_dim(0) - targets.clone().min_dim(0));
```
nathanielsimard, line 169: Same

examples/simple-regression/src/model.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+use burn::nn::loss::Reduction::Mean;
+use burn::nn::ReLU;
+
+#[derive(Module, Debug)]
+pub struct LinearModel<B: Backend> {
+    input_layer: Linear<B>,
+    output_layer: Linear<B>,
+    activation: ReLU
+}
```
nathanielsimard, line 19: I would rename it to `RegressionModel` since it's not technically linear (there is one activation function used).

examples/simple-regression/src/model.rs
```diff
@@ -0,0 +1,80 @@
...
+            .with_bias(true).init(device);
+        let output_layer = LinearConfig::new(64, 1)
+            .with_bias(true).init(device);
+
+        Self {
+            input_layer,
+            output_layer,
+            activation: ReLU::new()
+        }
+    }
```
nathanielsimard: Instead of a new function, I would create a config:

```rust
#[derive(Config)]
pub struct RegressionModelConfig {
    pub num_features: usize;
    #[config(default = 64)]
    pub hidden_size: usize;
}
```
Then I would implement init

```rust
impl RegressionModelConfig {
    pub fn init<B: Backend>(&self, device: &B::Device) -> RegressionModel<B> {
            let input_layer = LinearConfig::new(self.num_features, self.hidden_size)
            .with_bias(true).init(device);
        let output_layer = LinearConfig::new(self.hidden_size, 1)
            .with_bias(true).init(device);

        Self {
            input_layer,
            output_layer,
            activation: ReLU::new()
        }
    }
}
```

Something like that.

examples/simple-regression/src/training.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[config(default = 2)]
+    pub num_workers: usize,
+
+    #[config(default = 42)]
+    pub seed: u64,
+
+    pub optimizer: SgdConfig,
+
+    #[config(default = 10)]
+    pub input_feature_len: usize,
```
nathanielsimard, line 35: This should be replaced by the model config.

nathanielsimard (changes requested):
I've got some minor comments, but good job 👏

ashdtu:
Hi, made all changes suggested in latest commit.

nathanielsimard (approved):
Very clean, good job! 👏

PR #1159, opened 2024-01-20
Support more elemwise kernel fusion

nathanielsimard:
Added support for `Ones`, `Zeroes` and `Full`. Also had to support vectorization when no input and constant variables.

### DIFF
--- a/burn-wgpu/src/fusion/elemwise/kernel.rs
+++ b/burn-wgpu/src/fusion/elemwise/kernel.rs
@@ -137,8 +137,6 @@ impl ElementWiseSource {
             false => {
                 let reference_tensor = outputs[0];
                 let num_elems = calculate_num_elems_dyn_rank(&reference_tensor.shape);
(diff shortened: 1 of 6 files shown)


burn-wgpu/src/fusion/elemwise/kernel.rs
```diff
@@ -70,13 +70,32 @@ impl FusionKernel for VecElementWise {
 
             false
         };
+        let is_unavailable_output = |desc: &TensorDescription| {
+            let rank = desc.shape.len();
+
+            // The last dimension should be a multiple of the vector size.
+            if desc.shape[rank - 1] % self.source.factor != 0 {
```
louisfd, line 77: Simplifiable à
`[return] desc.shape[rank - 1] % self.source.factor != 0`

louisfd (approved):
Looks good. I have one optional thing to change but approved in advance.

PR #1161, opened 2024-01-22
New default recorder

### LINKED ISSUE #1151: Improve the record type performance
The current NamedMpkGz is too slow in term of serialization and deserialization, so we should investiguate alternatives or improve it.

laggui:
### Checklist

### Related Issues

Closes #1151

### Changes

After a bit of exploration and tests it was decided that we should revert to the `NamedMpkFileRecorder` (no compression) as the default. Any type of compression adds to much overhead and detracts the user experience.

- Changed the `DefaultFileRecorder`
- Updated the Burn book

### DIFF
--- a/burn-book/src/building-blocks/record.md
+++ b/burn-book/src/building-blocks/record.md
@@ -14,13 +14,13 @@ the flexibility possible to include any non-serializable field within your modul
 
 **Why not use safetensors?**
 
-`Safetensors` uses serde with the JSON file format and only supports serializing and deserializing
+[`safetensors`](https://github.com/huggingface/safetensors) uses serde with the JSON file format and only supports serializing and deserializing
 tensors. The record system in Burn gives you the possibility to serialize any type, which is very
 useful for optimizers that save their state, but also for any non-standard, cutting-edge modeling
 needs you may have. Additionally, the record system performs automatic precision conversion by using
 Rust types, making it more reliable with fewer manual manipulations. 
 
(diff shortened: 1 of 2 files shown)


burn-book/src/building-blocks/record.md
```diff
@@ -14,13 +14,13 @@ the flexibility possible to include any non-serializable field within your modul
...
 useful for optimizers that save their state, but also for any non-standard, cutting-edge modeling
 needs you may have. Additionally, the record system performs automatic precision conversion by using
 Rust types, making it more reliable with fewer manual manipulations. 
 
-It is important to note that the `safetensors` format uses the word *safe* to distinguish itself from Pickle, which is vulnerable to Python code injection. The safety comes from a checksum mechanism that guarantees that the data suffered no corruption. On our end, the simple fact that we use Rust already ensures that no code injection is possible. To prevent any data corruption, using a recorder with Gzip compression is recommended, as it includes a checksum mechanism.  
+It is important to note that the `safetensors` format uses the word *safe* to distinguish itself from Pickle, which is vulnerable to Python code injection. On our end, the simple fact that we use Rust already ensures that no code injection is possible. To prevent any data corruption, using a recorder with Gzip compression is recommended, as it includes a checksum mechanism.
```
nathanielsimard, line 23: Not sure if we should include this anymore: "To prevent any data corruption, using a recorder with Gzip compression is recommended, as it includes a checksum mechanism." Maybe updating it to: "If your storage mechanism doesn't handle data corruption, you may look into recorders that include a checksum, such as Gzip."

nathanielsimard (changes requested):

burn-book/src/building-blocks/record.md
laggui: Ahhh good point! Will fix.

nathanielsimard (approved):

PR #1165, opened 2024-01-22
Feat/recorder/custom device

### LINKED ISSUE #1160: Record items are loaded back into a default device
Now that a device is required when initializing NN modules (see  https://github.com/tracel-ai/burn/pull/1081), we should do the same when initializing with a record. However, this is not done (forgotten to be refactored), see: https://github.com/tracel-ai/burn/blob/main/burn-core/src/record/tensor.rs#L88-L134

nathanielsimard:
Fix #1160

I had the option to add a generic backend into the function `from_item` and directly into the trait `Record`. Setting the backend trait on the function would have created a trait conflict when the type that implements the Record trait is also generic over the backend trait, which is the case for module and tensor types.

After that, I just fixed the compilation errors. I had to update a bit the optimizer record type, but nothing really special and still is compatible with the previous record. It's just changing where the change from `AutodiffBackend` to `Backend` is done.

### DIFF
--- a/burn-core/src/record/tensor.rs
+++ b/burn-core/src/record/tensor.rs
@@ -85,7 +85,7 @@ impl<'de> Deserialize<'de> for BoolTensorSerde {
 
 // --- RECORD IMPLEMENTATIONS --- //
 
-impl<B: Backend, const D: usize> Record for Tensor<B, D> {
+impl<B: Backend, const D: usize> Record<B> for Tensor<B, D> {
     type Item<S: PrecisionSettings> = FloatTensorSerde<S>;
 
     fn into_item<S: PrecisionSettings>(self) -> Self::Item<S> {
@@ -96,12 +96,12 @@ impl<B: Backend, const D: usize> Record for Tensor<B, D> {
         FloatTensorSerde::new(self.into_data().convert().serialize())
     }
 
-    fn from_item<S: PrecisionSettings>(item: Self::Item<S>) -> Self {
-        Tensor::from_data(item.data.convert::<B::FloatElem>(), &B::Device::default())
+    fn from_item<S: PrecisionSettings>(item: Self::Item<S>, device: &B::Device) -> Self {
+        Tensor::from_data(item.data.convert::<B::FloatElem>(), device)
     }
 }
 
-impl<B: Backend, const D: usize> Record for Tensor<B, D, Int> {
+impl<B: Backend, const D: usize> Record<B> for Tensor<B, D, Int> {
     type Item<S: PrecisionSettings> = IntTensorSerde<S>;
 
     fn into_item<S: PrecisionSettings>(self) -> Self::Item<S> {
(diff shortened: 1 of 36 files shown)


burn-core/src/record/tensor.rs
```diff
@@ -112,12 +112,12 @@ impl<B: Backend, const D: usize> Record for Tensor<B, D, Int> {
         IntTensorSerde::new(self.into_data().convert().serialize())
     }
 
-    fn from_item<S: PrecisionSettings>(item: Self::Item<S>) -> Self {
+    fn from_item<S: PrecisionSettings>(item: Self::Item<S>, _device: &B::Device) -> Self {
         Tensor::from_data(item.data.convert(), &B::Device::default())
```
antimora, line 116: Shouldn't device be passed instead of `&B::Device::default()`?

burn-core/src/record/tensor.rs
```diff
@@ -128,7 +128,7 @@ impl<B: Backend, const D: usize> Record for Tensor<B, D, Bool> {
         BoolTensorSerde::new(self.into_data().serialize())
     }
 
-    fn from_item<S: PrecisionSettings>(item: Self::Item<S>) -> Self {
+    fn from_item<S: PrecisionSettings>(item: Self::Item<S>, _device: &B::Device) -> Self {
         Tensor::from_data(item.data, &B::Device::default())
```
antimora, line 132: Shouldn't device be passed instead of `&B::Device::default()`?

burn-derive/src/record/base.rs
```diff
@@ -22,12 +22,19 @@ struct RecordDeriveCodegen {
     name_item: Ident,
     gen: StructRecordItemCodegen,
     generics: Generics,
+    has_backend: bool,
```
antimora, line 25: Is this done for legacy? I am not sure when Backend would not be present

antimora (changes requested):
Wow this is massive as well. 

I found a couple issues. Please review my comments and tell me if they're valid:

1. Int and Bool tensors are still initialized from default device.
2. Needs better explanation/comment for has backend flag in the derive.

laggui (changes requested):
Overall LGTM. Some minor changes requested (mainly in agreement with @antimora).

burn-core/src/record/tensor.rs
nathanielsimard: Yes, thanks for spotting that mistake 🙏

burn-core/src/record/tensor.rs
nathanielsimard: Yes, thanks for spotting that mistake 🙏

burn-derive/src/record/base.rs
nathanielsimard: This is intended for record with and without the backend trait defined as generic. Since the `Record` trait didn't have any generic before, we didn't have to handle that case.

```rust
#[derive(Record)]
struct NoBackend;

// We need to generate the following

impl<B: Backend> Record<B> for NoBackend {
   ...
}

#[derive(Record)]
struct MyModule<B: Backend> {
  layer: nn::Linear<B>
}

// We need to generate the following, but we don't need to manually add the backend trait, since it's already defined.

impl<B: Backend> Record<B> for MyModule<B> {
   ...
}
```

antimora (approved):
Looks good from my end. There is one bit by @laggui you should respond/address.

laggui (approved):

PR #1167, opened 2024-01-23
fix: Batchnorm does not work properly when training on multiple devices

wcshds:
It seems that `running_var` and `runninng_mean` are shared across devices when training on multiple devices. Therefore, ensuring that `running_var`/`var` and `running_mean`/`mean` are on the same device will allow Batchnorm to work properly during multiple devices training.


nathanielsimard (approved):
LGTM

Issue #1141, opened 2024-01-15
Build error message in doc.rs burn page


syl20bnr:
**Describe the bug**

We have a build error banner on the documentation reference page hosted on `doc.rs`.

**To Reproduce**

 1. Click on the link: https://docs.rs/crate/burn/latest
 2. See the banner at the top

**Expected behavior**

No error banner is displayed.

**Screenshots**

![image](https://github.com/tracel-ai/burn/assets/1243537/124d1358-c409-4ce7-b041-75c90cf12f89)

PR #1168, opened 2024-01-23
Fix doc build on docs.rs

syl20bnr:
Should fix https://github.com/tracel-ai/burn/issues/1141

Unfortunately I was not able to test it properly because the docs.rs documentation is very sparse and I encountered several obstacles (even in an Ubuntu VM). I'll try to make it work before the release if I can get good support on zulip.

While I cannot guarantee that it will fix the issue, a local `cargo doc` seems to skip the `torch-sys` build with these added metadata (even if the name is related to `docs.rs`). So high chance we will be good 🙂 

### Checklist


nathanielsimard (approved):
I remember that it was needed when including this backend in the doc. So I think this will work.

PR #1173, opened 2024-01-25
Adding prod & prod_dim

unrenormalizable:
## Pull Request Template

> NOTE:
> 1. Following things still need to be done, creating this PR to get guidance on those. Go to the comments section of this PR and I point these out in their respective places.
> 1. tch, ndarray & candle dont support prod_axis required for this implementation. **QUESTION** should I implement it in the burn layer with some gymnastics or leave it as panic!(<platform does not support>) like e.g. burn-candle's int_div_scalar
> 1. there is large copy-pasta from sum implementation 
> 1. was expecting to write some more directed tests esp. in burn-wgpu but couldn't find tests for other ops. Am I missing something or do we expect the ops to be tested by the higher layers?

### Checklist

### Related Issues/PRs

n/a

### Changes

Adds prod & prod_dim in preparation for adding ReduceProd ONNX op. Really a most useless checkin, doing it for getting familiar with the codebase + fun.

### Testing

run-checks all


burn-candle/src/ops/int_tensor.rs
```diff
@@ -311,6 +311,14 @@ impl<F: FloatCandleElement, I: IntCandleElement> IntTensorOps<Self> for Candle<F
         CandleTensor::new(tensor.tensor.sum_keepdim(dim).unwrap())
     }
 
+    fn int_prod<const D: usize>(tensor: IntTensor<Self, D>) -> IntTensor<Self, 1> {
+        todo!();
```
unrenormalizable, line 315: tch, ndarray & candle dont support prod_axis required for this implementation. 

QUESTION should I implement it in the burn layer with some gymnastics or leave it as panic!() like e.g. burn-candle's int_div_scalar

burn-wgpu/src/kernel/reduce/tune/prod_dim.rs
```diff
@@ -0,0 +1,116 @@
+use burn_compute::tune::{AutotuneOperation, AutotuneOperationSet};
```
unrenormalizable, line 1: there is large copy-pasta from sum implementation

what is the recommendation here? parametrize the core implementation and call them with sum / prod arguments?

burn-wgpu/src/kernel/reduce/reduction_shared_memory.rs
```diff
@@ -27,6 +28,19 @@ impl StaticKernelSource for SumDimSharedMemory {
...
 
+impl StaticKernelSource for ProdDimSharedMemory {
+    fn source() -> SourceTemplate {
+        ReductionDimSharedMemoryRaw::source()
+            .register(
+                "shared_size",
+                (WORKGROUP_DEFAULT * WORKGROUP_DEFAULT).to_string(),
+            )
+            .register("initial", 0.0.to_string())
+            .register("update", "shared_memory[local_id] += value; ")
```
louisfd, line 39: I guess you'll want initial to be 1.0 and the update to be *=

burn-wgpu/src/kernel/reduce/tune/prod_dim.rs
louisfd: I guess we'll want to refactor this so that all reduce operations share some core. The list of reduction operations is growing (there's also this PR #1136 coming up with more autotuned reduce) so we can't afford to have this many duplicates. For now you can keep it as is but I'll open an issue for refactoring this.

burn-wgpu/src/template/reduction/recursive_prod.wgsl
```diff
@@ -0,0 +1,38 @@
...
+) {
+    let id_global = global_id.y * (num_workgroups.x * WORKGROUP_SIZE_X) + global_id.x;
+    let id_local = local_id.y * WORKGROUP_SIZE_X + local_id.x;
+
+    data[id_local] = input[id_global];
+
+    workgroupBarrier();
+
+    if id_local == 0u {
+        var prod = {{ elem }}(0);
```
louisfd, line 30: You should start at 1.0 or it won't compute much ;)

burn-candle/src/ops/int_tensor.rs
louisfd: I think it's rather bad to offer an operation in the API if it's gonna fail on 75% of our backends. I think if you can work out some default implementation in the burn layer it would be cool, otherwise we should not even offer the operation. 
Candle's int_div_scalar is an exception that we should fix eventually.

burn-autodiff/src/ops/tensor.rs
```diff
@@ -1077,6 +1104,34 @@ impl<B: Backend> TensorOps<Self> for Autodiff<B> {
...
+            type State = (Shape<D>, usize);
+
+            fn backward(self, ops: Ops<Self::State, 1>, grads: &mut Gradients) {
+                let (shape, dim) = ops.state;
+
+                unary::<B, D, D, _>(ops.parents, ops.node, grads, |grad| {
+                    let ones = B::ones(shape, &B::device(&grad));
+                    let grad = B::prod_dim(grad, dim);
+
+                    B::mul(ones, grad)
```
louisfd, line 1121: Why multiply by one?

burn-autodiff/src/ops/tensor.rs
```diff
@@ -1077,6 +1104,34 @@ impl<B: Backend> TensorOps<Self> for Autodiff<B> {
...
+
+        impl<B: Backend, const D: usize> Backward<B, D, 1> for ProdDim {
+            type State = (Shape<D>, usize);
+
+            fn backward(self, ops: Ops<Self::State, 1>, grads: &mut Gradients) {
+                let (shape, dim) = ops.state;
+
+                unary::<B, D, D, _>(ops.parents, ops.node, grads, |grad| {
+                    let ones = B::ones(shape, &B::device(&grad));
+                    let grad = B::prod_dim(grad, dim);
```
louisfd, line 1119: As an example, I think the derivative of prod_dim([a, b, c, d], 0) should be 
[bcd, acd, abd, bcd], then multiply by the unchanged grad. So you'll need to register the original input as state. Don't hesitate to call me out if you think I'm wrong I haven't thought of it too long

louisfd (changes requested):
Hi @unrenormalizable 
Thanks for the PR draft, it's looking good. See my comments

burn-candle/src/ops/int_tensor.rs
unrenormalizable: i agree. i will file issues on tch, ndarray and candle and implement a default.

burn-autodiff/src/ops/tensor.rs
unrenormalizable: yeah i am stupid. i will add unit tests for this.

unrenormalizable:
> Great! Let me know when you start working on ONNX part.

my apologies, taking way longer than expected. overestimated my learning abilities 🤣

antimora:
No worries. We are here to help.

unrenormalizable:
Folks, I am going to pause the prod/prod_dim work till required dependencies are available.

The main blocker is cumprod is required to implement the autodiff/grad component for prod. This is as per the [pytorch implementation](https://github.com/pytorch/pytorch/blob/main/torch/csrc/autograd/FunctionsManual.cpp#L813) which is a reasonable approach.

In addition I have filed the prod_axis requirements on [huggingface/candle/1620](https://github.com/huggingface/candle/issues/1620) & [rust-ndarray/ndarray/1351](https://github.com/rust-ndarray/ndarray/issues/1351) 

Following items from this PR will be taken forward in a separate PR.
1. reduce_dim_sum.wgsl edge case fix
2. reduce_dim_shared_memory.wgsl initialization issue

Issue #965, opened 2023-11-18
Add `dependencies` and `vulnerabilities` checks to `xtask`


Luni-4:
### Feature description

The `vulnerabilities` and `dependencies` checks could be added to `xtask` in order to be executed locally as well.

### Feature motivation

Adding more `xtask` tasks helps a developer in reducing the effort during the development process.

Luni-4:
Feel free to assign this issue to me

PR #1181, opened 2024-01-26
Add cargo xtask commands for dependencies and vulnerabilities checks

syl20bnr:
### Checklist

### Related Issues/PRs

Fixes issue https://github.com/tracel-ai/burn/issues/965

### Changes

This PR adds two new commands to `cargo xtask`:
- cargo xtask dependencies
- cargo xtask vulnerabilities

All the checks defined in our github workflows are added. I also implemented [all the sanitizers](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#sanitizer). I did not thoroughly test all of them, some additional tweaks might be necessary.

I tried to be as clear as possible in the command line documentation so that users can just explore what is available and then the help should guide them in order to correctly execute the checks (for instance some checks need to be executed using a nightly toolchain). The cargo third-party commands should be automatically installed as well.

Dependencies documentation generation is also skipped with `--no-deps`.


xtask/src/vulnerabilities.rs
```diff
@@ -0,0 +1,293 @@
...
+use crate::utils::{
+    ensure_cargo_command_is_installed, is_current_toolchain_nightly, run_cargo,
+    rustup_add_component, rustup_get_installed_targets, Params,
+};
+use crate::{endgroup, group};
+use crate::{logging::init_logger, utils::format_duration};
+use std::fmt;
+
+#[derive(clap::ValueEnum, Default, Copy, Clone, PartialEq, Eq)]
+pub enum VulnerabilityCheckType {
```
Luni-4, line 13: Can we use `pub(crate)` here, so we reduce the symbols present in the binary?

xtask/src/vulnerabilities.rs
```diff
@@ -0,0 +1,293 @@
...
+            endgroup!();
+        } else {
+            error!(
+                "You must use 'cargo +nightly' to run this check.
+ Install a nightly toolchain with 'rustup toolchain install nightly'."
+            )
+        }
+    }
+
+    fn flags(&self) -> String {
```
Luni-4, line 184: Can we return them as `&'static str` here?

xtask/src/vulnerabilities.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    // Returns true if the sanitizer is supported by the currently installed targets
+    fn is_target_supported(&self) -> bool {
+        let installed_targets = rustup_get_installed_targets();
+        let supported = self.supported_targets();
+        installed_targets.lines().any(|installed| {
+            supported
+                .iter()
+                .any(|target| target.to_string() == installed.trim())
```
Luni-4, line 260: Is it necessary to convert `target` into a `String` or we can compare `&str`?

xtask/src/runchecks.rs
```diff
@@ -5,136 +5,123 @@
...
-    group!("Rustup: {} add {}", command, target);
-    run_command(
-        "rustup",
-        &[command, "add", target],
-        "Failed to run rustup",
-        "Failed to wait for rustup child process",
-    );
-    endgroup!();
+#[derive(clap::ValueEnum, Default, Copy, Clone, PartialEq, Eq)]
+pub enum CheckType {
```
Luni-4, line 24: Can we use `pub(crate)` here?

xtask/src/dependencies.rs
```diff
@@ -0,0 +1,102 @@
...
+    endgroup, group,
+    logging::init_logger,
+    utils::{
+        ensure_cargo_command_is_installed, format_duration, is_current_toolchain_nightly,
+        run_cargo, Params,
+    },
+};
+
+#[derive(clap::ValueEnum, Default, Copy, Clone, PartialEq, Eq)]
+pub enum DependencyCheckType {
```
Luni-4, line 13: Can we use `pub(crate)` here?

xtask/src/dependencies.rs
```diff
@@ -0,0 +1,102 @@
...
+    All,
+    /// Perform an audit of all dependencies using the cargo-audit crate `<https://crates.io/crates/cargo-audit>`
+    Audit,
+    /// Run cargo-deny check `<https://crates.io/crates/cargo-deny>`
+    Deny,
+    /// Run cargo-udeps to find unused dependencies `<https://crates.io/crates/cargo-udeps>`
+    Unused,
+}
+
+pub fn run(dependency_check: DependencyCheckType) -> anyhow::Result<()> {
```
Luni-4, line 25: Can we use `pub(crate)` here?

Luni-4 (changes requested):
Thanks a lot for your work! 

Just simple changes and a question:

Instead of having patterns such as `"hi".to_string()", can we reduce allocations using `std::borrow::Cow<'static, str>` in the code?

xtask/src/vulnerabilities.rs
syl20bnr: I believe so, I use the Display trait implemented for Target variants. Any improvement over this is welcome, do you think of another way to do the comparison ?

syl20bnr:
Thank you for the review, I pushed some fixes.

> Instead of having patterns such as "hi".to_string()", can we reduce allocations using std::borrow::Cow<'static, str>` in the code?

Do you have a concrete example you can share ? 

 > Can we add these new commands to CI?
 
 I was planning to do that but in the end I prefer to open a different PR for this.

xtask/src/vulnerabilities.rs
Luni-4: I do not know if I understood the problem correctly: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1425862775d764bf7a531a8a756b01b4

xtask/src/dependencies.rs
```diff
@@ -0,0 +1,102 @@
...
+    Audit,
+    /// Run cargo-deny check `<https://crates.io/crates/cargo-deny>`
+    Deny,
+    /// Run cargo-udeps to find unused dependencies `<https://crates.io/crates/cargo-udeps>`
+    Unused,
+}
+
+pub(crate) fn run(dependency_check: DependencyCheckType) -> anyhow::Result<()> {
+    // Setup logger
+    init_logger().init();
```
nathanielsimard, line 27: You can do:

```rust
impl DependencyCheckType {
    pub(crate) fn run(self) -> anyhow::Result<()> {
        ...
    }
}
```

To allow the dot notation:

```rust
DependencyCheckType::All.run();
```

Very very optional, just if you think it's cleaner.

xtask/src/dependencies.rs
```diff
@@ -0,0 +1,102 @@
...
+        Params::from(["check"]),
+        HashMap::new(),
+        "Cargo deny should be installed and it should correctly run",
+    );
+    endgroup!();
+}
+
+/// Run cargo-udeps
+fn cargo_udeps() {
+    if is_current_toolchain_nightly() {
```
nathanielsimard, line 85: Is there a way to install automatically the toolchain? Maybe we can use `cargo +nightly` instead of `cargo`? Not sure if it works @Luni-4 ?

xtask/src/dependencies.rs
```diff
@@ -0,0 +1,102 @@
...
+    endgroup, group,
+    logging::init_logger,
+    utils::{
+        ensure_cargo_command_is_installed, format_duration, is_current_toolchain_nightly,
+        run_cargo, Params,
+    },
+};
+
+#[derive(clap::ValueEnum, Default, Copy, Clone, PartialEq, Eq)]
+pub(crate) enum DependencyCheckType {
```
nathanielsimard, line 13: Not sure if we actually need `Type` at the end, just `DependencyCheck` might be a better name. I often use `Type`, or rather `Kind`, when there is an enum and a struct of the same name that interact.

xtask/src/utils.rs
```diff
@@ -1,5 +1,182 @@
...
+        .args(&params.params)
+        .stdout(Stdio::inherit()) // Send stdout directly to terminal
+        .stderr(Stdio::inherit()); // Send stderr directly to terminal
+
+    if let Some(path) = path {
+        cargo.current_dir(path);
+    }
+
+    // Handle cargo child process
+    let rustup_process = cargo.spawn().expect(error);
```
nathanielsimard, line 41: I don't think this is a rustup process all the time, `cargo_process` would be a more accurate name I think.

xtask/src/utils.rs
```diff
@@ -1,5 +1,182 @@
...
+    handle_child_process(rustup_process, "Cargo process should run flawlessly");
+}
+
+/// Ensure that a cargo command is installed
+pub(crate) fn ensure_cargo_command_is_installed(command: &str) {
```
nathanielsimard, line 46: Just in term of naming, we ensure a _crate_ is installed. Since it's a _crate_ that is a binary, we can use it in a command. So I would rename the function to: `ensure_cargo_crate_is_installed`. We might have other commands that are installed differently in the future.

xtask/src/utils.rs
```diff
@@ -1,5 +1,182 @@
...
+use std::{
+    collections::HashMap,
+    path::Path,
+    process::{Child, Command, Stdio},
+    time::Duration,
+};
+
+use crate::{endgroup, group};
+
+// Cargo utils -----------------------------------------------------------
```
nathanielsimard, line 11: I would actually refactor the file structure:

```
- src/utils/cargo.rs
- src/utils/workspace.rs
- src/utils/rustup.rs
- src/utils/process.rs
- src/utils/time.rs
```

xtask/src/vulnerabilities.rs
```diff
@@ -0,0 +1,291 @@
...
+use crate::utils::{
+    ensure_cargo_command_is_installed, is_current_toolchain_nightly, run_cargo,
+    rustup_add_component, rustup_get_installed_targets, Params,
+};
+use crate::{endgroup, group};
+use crate::{logging::init_logger, utils::format_duration};
+use std::fmt;
+
+#[derive(clap::ValueEnum, Default, Copy, Clone, PartialEq, Eq)]
+pub(crate) enum VulnerabilityCheckType {
```
nathanielsimard, line 13: Same comment regarding `Type`.

xtask/src/vulnerabilities.rs
```diff
@@ -0,0 +1,291 @@
...
+
+                let features = self.cargo_features();
+                let mut args = vec!["--", "--color=always", "--no-capture"];
+                args.extend(features);
+
+                run_cargo(
+                    "test",
+                    args.into(),
+                    envs.into_iter().collect(),
+                    "Failed to run cargo fmt",
```
nathanielsimard, line 170: Is it `test` or `fmt` ?

xtask/src/vulnerabilities.rs
syl20bnr: Thank you for the example. Well I can do `.any(|target| target.to_string().eq(installed.trim()))` but that's still allocating a `String`. Since I use the idiomatic Display trait and since this code is not performance critical I believe we can let it like that. What do you think ?

nathanielsimard: If you want to be cleaner, we should parse the string into the enum instead. You would have to add `PartialEq` on the `Target` type, but it would be more robust.

nathanielsimard (changes requested):
Good job, not sure if all those checks are important, but at least we have a way to test them easily.

xtask/src/dependencies.rs
syl20bnr: My original solution installed the toolchain and even updated it but I then encountered consistency issue where the wrong toolchain was selected for nested cargo or even rustup process calls. So I switched to a design where the user explicitly set the toolchain on the command line. For commands that requires a nightly toolchain, an error is triggered if the user defines a non-nightly toolchain and a message explains what to do to fix the error.

xtask/src/utils.rs
syl20bnr: ahah good catch this is a leftover of my previous solution to circumvent the toolchain issue using `rustup run`.

xtask/src/vulnerabilities.rs
syl20bnr: Good catch. The expect messages are not conventional also. This parameter ends up being an expect message. I'll fix this next time I work in xtasks.

xtask/src/dependencies.rs
syl20bnr: Done

xtask/src/dependencies.rs
syl20bnr: I removed the `Type` suffix.

xtask/src/utils.rs
syl20bnr: Done

xtask/src/utils.rs
syl20bnr: Done !

syl20bnr:
Ready for another round of review.

I parsed the strings returned by rustup in the end as @nathanielsimard proposed. I also added some tests and converted some strings to constants. Utils module is now a directory.

nathanielsimard (approved):
LGTM

syl20bnr:
@Luni-4 Is that OK with you too ?

Luni-4 (approved):
Fine for me too! Thanks a lot for your work!

Issue #1154, opened 2024-01-19
Burn Book: Data manipulation guide


louisfd:
- Introduce every dataset transform and their uses
- Introduce the batcher

ashdtu:
Addressed in PR #1183 . Closing

Issue #1153, opened 2024-01-19
Burn Book: Improve Tensor type introduction


louisfd:
- Describe the rank
- Add a section about how to create new tensors
   - Describe Data type
   - Describe Element conversion
   - Describe Tensor conversion
- Insist on no stateful operations
- Describe ownership rules and cloning requirements

ashdtu:
Addressed in PR #1183  . Closing

Issue #1103, opened 2024-01-02
Not clear what 'guide' is


CurdledMilk:
I copied and pasted all the code in "The Burn Book" from sections 4.1 "model" -> 4.4 "Backend"
I then fixed all the imports so everything was imported only once and there was no import errors.... except for one.
I tried running it but it doesn't know what guide crate is. I tried adding the guide crate but it doesn't work. 

Without 'cargo add guide': 
failed to resolve: use of undeclared crate or module `guide`

With guide crate installed:
unresolved import `guide::model`

rMazeiks:
I believe `guide::model::ModelConfig` and `guide::training::train` refer to items that are not part of `burn` itself but were implemented as part of the tutorial – it's the struct and function you copy-pasted, just update the path to refer to it correctly.

You can likely replace `guide::model::ModelConfig` with just `ModelConfig` and let the IDE complete the import for you.

nathanielsimard:
I guess we should improve the process of following the book with the examples a bit.

ashdtu:
Addressed in PR #1183  . Closing

PR #1183, opened 2024-01-26
Docs/guide

ashdtu:
### Checklist

### Related Issues/PRs

#1103 : Better documentation of Guide example.
#1153: Improve Tensor introduction
#1154: Improve Dataset transform guide

### Changes
Corresponding to each issue, here are the changes in this PR:
#1103 : The changes suggested bridge the gap between the example code provided in examples/guide vs what is written in guide. The documentation currently lacked the partitioning of code into model, data, training files similar to the example code. This sometimes made it hard to follow if for eg someone is copy-pasting code from there. Fixed these issues and added helpful comments with reference to original example.

#1153
* Added introduction of  Tensor with rank, similarity/diff for people coming from PyTorch.
* Initialization from data, floats api. .convert() , elem etc.
*  Underline importance of ownership and cloning with example.
* No stateful operations : **TODO** (discuss)

#1154 
* Added introduction of dataset transform with example usages.
* Batcher omitted as already present in current guide in mnist tutorial and other concepts covered by Rust introduction commit.

### DIFF
--- a/burn-book/src/building-blocks/dataset.md
+++ b/burn-book/src/building-blocks/dataset.md
@@ -32,6 +32,50 @@ distributions.
 | `MapperDataset`   | Computes a transformation lazily on the input dataset.                                                                   |
 | `ComposedDataset` | Composes multiple datasets together to create a larger one without copying any data.                                     |
 
+Let us look at the basic usages of each dataset transform and how they can be composed together. The full documentation
+of each transform can be found at the [API reference](https://burn.dev/docs/burn/data/dataset/transform/index.html).
+
+* **SamplerDataset**: This transform can be used to sample items from a dataset with(default) and without replacement.
+Transform is initialized with a sampling size which can be bigger or smaller than the input dataset size. This is
+particularly useful in cases where we want to checkpoint larger datasets more often during training
+and smaller datasets less often as the size of an epoch is now controlled by the sampling size. Sample usage:
+```rust, ignore
+type DbPedia = SqliteDataset<DbPediaItem>;
+let dataset: DbPedia = HuggingfaceDatasetLoader::new("dbpedia_14")
+        .dataset("train").
+        .unwrap();
+                
+let dataset = SamplerDataset<DbPedia, DbPediaItem>::new(dataset, 10000);
+```
+
+* **ShuffledDataset**: This transform can be used to shuffle the items of a dataset. Particularly useful before splitting
+the raw dataset into train/test splits. Can be initialized with a seed to ensure reproducibility. 
+```rust, ignore
+let dataset = ShuffledDataset<DbPedia, DbPediaItem>::with_seed(dataset, 42);
+```
+
+* **PartialDataset**: This transform is useful to return a view of the dataset with a specified start, end index. Used
+to create train/val/test splits. Below example we show how to chain ShuffledDataset and PartialDataset to create splits.
+```rust, ignore
+// define chained dataset type here for brevity
+let PartialData = PartialDataset<ShuffledDataset<DbPedia, DbPediaItem>>;
+let dataset_len = dataset.len();
+let split == "train"; // or "val"/"test"
+
+let data_split = match split {
+            "train" => PartialData::new(dataset, 0, len * 8 / 10), // Get first 80% dataset
+            "test" => PartialData::new(dataset, len * 8 / 10, len), // Take remaining 20%
+            _ => panic!("Invalid split type"),                     // Handle unexpected split types
+        };
+```
+
+* **MapperDataset**: This transform is useful to apply a transformation on each of the items of a dataset. Particularly
+useful for normalization of image data when channel means are known. 
+
+* **ComposedDataset**: This transform is useful to compose multiple datasets downloaded from multiple sources(say 
+different HuggingfaceDatasetLoader sources) into a single bigger dataset which can be sampled from one source. 
+
+
 ## Storage
 
 There are multiple dataset storage options available for you to choose from. The choice of the
--- a/burn-book/src/building-blocks/dataset.md
+++ b/burn-book/src/building-blocks/dataset.md
@@ -1,6 +1,7 @@
 # Dataset
 
-Most deep learning training being done on datasets –with perhaps the exception of reinforcement learning–, it is essential to provide a convenient and performant API.
+Most deep learning training being done on datasets –with perhaps the exception of reinforcement learning–, it is
+essential to provide a convenient and performant API.
 The dataset trait is quite similar to the dataset abstract class in PyTorch:
 
 ```rust, ignore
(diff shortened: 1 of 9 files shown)


burn-book/src/building-blocks/dataset.md
```diff
@@ -32,6 +32,50 @@ distributions.
...
+the raw dataset into train/test splits. Can be initialized with a seed to ensure reproducibility. 
+```rust, ignore
+let dataset = ShuffledDataset<DbPedia, DbPediaItem>::with_seed(dataset, 42);
+```
+
+* **PartialDataset**: This transform is useful to return a view of the dataset with a specified start, end index. Used
+to create train/val/test splits. Below example we show how to chain ShuffledDataset and PartialDataset to create splits.
+```rust, ignore
+// define chained dataset type here for brevity
+let PartialData = PartialDataset<ShuffledDataset<DbPedia, DbPediaItem>>;
```
nathanielsimard, line 61: Prefer using `type` for type alias

```rust
type PartialData = PartialDataset<ShuffledDataset<DbPedia, DbPediaItem>>;
```

burn-book/src/building-blocks/dataset.md
```diff
@@ -32,6 +32,50 @@ distributions.
 | `MapperDataset`   | Computes a transformation lazily on the input dataset.                                                                   |
 | `ComposedDataset` | Composes multiple datasets together to create a larger one without copying any data.                                     |
 
+Let us look at the basic usages of each dataset transform and how they can be composed together. The full documentation
+of each transform can be found at the [API reference](https://burn.dev/docs/burn/data/dataset/transform/index.html).
```
nathanielsimard, line 36: I would add a sentence or two specifying that the dataset transformation are lazy by default except when specified otherwise.

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
 Note that the specific element types used for `Float`, `Int`, and `Bool` tensors are defined by
 backend implementations.
 
-## Operations
+In contrast to PyTorch where a Tensor is characterised by the size of it's each dimension, a Burn Tensor is
+characterized by the number of dimensions D in its declaration. The actual shape of the tensor is inferred from
```
nathanielsimard, line 17: I don't think PyTorch is characterized by the size of its each dimension. There is nothing specified for the element type, the shape, or the rank. All of those are runtime properties. In contrast, in Burn, we have those specified as types!

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
...
+let tensor_1 = Tensor::<Backend, 1>::from_floats([1.0, 2.0, 3.0, 4.0, 5.0]);
+
+// incorrect: let tensor_1 = Tensor::<Backend, 5>::from_floats([1.0, 2.0, 3.0, 4.0, 5.0]);
+// this will lead to an error and is for creating a 5-D tensor
+```
+
+### Initialization
+
+Burn Tensors are primarily initialized using the `from_data()` method which takes the `Data` struct as input. 
+The `Data` struct has two fields: value & shape. To access the data of any tensor, we use the method `.to_data()`.
```
nathanielsimard, line 30: `to_data()` if we want to reuse the tensor afterward, or `into_data()` otherwise.

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
...
+// correct: Tensor is 1-Dimensional with 5 elements
+let tensor_1 = Tensor::<Backend, 1>::from_floats([1.0, 2.0, 3.0, 4.0, 5.0]);
+
+// incorrect: let tensor_1 = Tensor::<Backend, 5>::from_floats([1.0, 2.0, 3.0, 4.0, 5.0]);
+// this will lead to an error and is for creating a 5-D tensor
+```
+
+### Initialization
+
+Burn Tensors are primarily initialized using the `from_data()` method which takes the `Data` struct as input.
```
nathanielsimard, line 29: Also add the fact that `Data` can be built from `slices`, `vectors`, etc. See https://github.com/tracel-ai/burn/blob/main/burn-tensor/src/tensor/data.rs

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
...
+Burn Tensors are primarily initialized using the `from_data()` method which takes the `Data` struct as input. 
+The `Data` struct has two fields: value & shape. To access the data of any tensor, we use the method `.to_data()`.
+Let's look at a couple of examples for initializing a tensor from different inputs. 
+
+```rust, ignore
+// Initialization from a float array
+let tensor_1 = Tensor::<Backend, 1>::from_data([1.0, 2.0, 3.0]);
+
+// equivalent to (recommended)
+let tensor_2 = Tensor::<Backend, 1>::from_floats([1.0, 2.0, 3.0]);
```
nathanielsimard, line 38: The different with from `floats` is that the element type is f32, where from data the element type is defined by the backend. When the backend is generic, you have to use some sort of convertion.

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
...
+        age: 10,
+        height: 100,
+        weight: 50.0
+    };
+let tensor_3 = Tensor::<Backend, 1>::from_data(Data::from([bmi.age as f32, bmi.height as f32, bmi.weight]).convert());
+
+```
+The `.convert()` method for Data struct is called in the last example to ensure that the data's primitive type is 
+consistent across all backends. The convert operation can also be done at element wise level as: 
+`let tensor_4 = Tensor::<B, 1, Int>::from_data(Data::from([(item.age as i64).elem()])`.
```
nathanielsimard, line 58: Specify that using `elem` is only possible when the trait `ElementConversion` is imported.

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
...
+    println!("{:?}", input.to_data());      // Success: [0.0, 0.33333334, 0.6666667, 1.0]
+    
+    // Notice that max, min have been moved in last operation so the below print will give an error.
+    // If we want to utlize them as well for further operations, they will need to be cloned in similar fashion.
+    // println!("{:?}", min.to_data());    
+```
+
+We don't need to be worried about memory overhead because with cloning, the tensor's buffer isn't copied, 
+and only a reference to it is increased. This makes it possible to determine exactly how many times a tensor is used, 
+which is very convenient for reusing tensor buffers and improving performance. For that reason, we don't provide
```
nathanielsimard, line 91: for reusing tensor buffers or even fusing operations into a single kernel (link to burn fusion maybe).

nathanielsimard (changes requested):
Some small corrections, see my comments bellow.

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,14 +13,85 @@ Tensor<B, D, Bool>     // Bool tensor
...
+
+```rust, ignore
+    let input = Tensor::<Wgpu, 1>::from_floats([1.0, 2.0, 3.0, 4.0]);
+    let min = input.clone().min();
+    let max = input.clone().max();
+    let input = (input.clone() - min.clone()).div(max - min);
+    println!("{:?}", input.to_data());      // Success: [0.0, 0.33333334, 0.6666667, 1.0]
+    
+    // Notice that max, min have been moved in last operation so the below print will give an error.
+    // If we want to utlize them as well for further operations, they will need to be cloned in similar fashion.
```
louisfd, line 85: There's a typo in utilize, i prefer `use` to utilize in general

antimora:
Quick feedback re formatting (haven't read the content in details yet). 

I recommend you format the text using something like prettier if you have a VS/Neovim plug-in. We have a configuration for prettier: https://github.com/tracel-ai/burn/blob/main/burn-book/.prettierrc.json (or you can define yours)

This helps formatting tables and wrapping markdown text.

ashdtu:
Many thanks @nathanielsimard, @louisfd, and @antimora for the suggestions. Incorporated in the latest commit. Let me know if it's good to go.

nathanielsimard (approved):

ashdtu:
Thank again for the review. Made the minor fixes suggested. Merging now!

Issue #1046, opened 2023-12-04
Module: Implement for Tuple


dcvz:
### Feature description

Implementing Module support for Tuple

### Feature motivation

Would allow syntax such as:

```rust
#[derive(Module, Debug)]
pub struct UNetMidBlock2D<B: Backend> {
    resnet: ResnetBlock2D<B>,
    attn_resnets: Vec<(AttentionBlock<B>, ResnetBlock2D<B>)>,
}
```

### Workaround

Use struct instead of tuple.

antimora:
Would a workaround of using struct work in this case? Suggesting in case if someone is stuck on this issue.

dcvz:
> Would a workaround of using struct work in this case? Suggesting in case if someone is stuck on this issue.

Yup! That’s the workaround I’ve gone with but raised the issue after a conversation with Nathaniel.

antimora:
Done via #1186

PR #1186, opened 2024-01-28
Added tuple modules

varonroy:
### Checklist

  - This change is not significant enough to warrant an addition to the book.

### Related Issues/PRs

Related to issue: #1046.

### Changes

This PR provides an implementation for tuple modules which reduce the amount of code that needs to written in some cases. Tuple modules also allow another way to group modules besides structs.
```rust
#[derive(Module, Debug)]
pub struct MyModule<B: Backend> {
    linear: Linear<B>,
    block: (
        Linear<B>,
        ReLU,
        Linear<B>,
        ReLU,
    ),
}
```
Previously, creating a structure similar to the `block` field above would require creating a separate struct.
```rust
#[derive(Module, Debug)]
pub struct Block<B: Backend> {
    l1: Linear<B>,
    act1: ReLU,
    l2: Linear<B>,
    act2: ReLU,
}
```

### Testing

`ModuleComposed` in `burn-core/tests/derive-module.rs` has been modified with an additional tuple field and all tests have passed.

### DIFF
--- a/burn-core/src/module/param/primitive.rs
+++ b/burn-core/src/module/param/primitive.rs
@@ -197,3 +197,66 @@ where
         self.map(|module| module.valid())
     }
 }
+
+macro_rules! impl_module_tuple {
+    ([$($l:ident),*][$($i:tt),*]) => {
+        impl<B, $($l,)*> Module<B> for ($($l,)*)
+        where
+            B: Backend,
+            $($l: Module<B> + Debug + Send + Sync + Clone,)*
+        {
+            type Record = ($($l::Record),*);
+
+            fn collect_devices(&self, mut devices: Vec<B::Device>) -> Vec<B::Device> {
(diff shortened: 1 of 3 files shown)


burn-core/src/module/param/primitive.rs
```diff
@@ -197,3 +197,66 @@ where
         self.map(|module| module.valid())
     }
 }
+
+macro_rules! impl_module_tuple {
+    ([$($l:ident),*][$($i:tt),*]) => {
```
nathanielsimard, line 202: I would add a simple comment explaining this macro: like `$l` is generic modules and `$i` is the positions.

nathanielsimard (commented):
Only a single minor comment, thanks a lot for the contribution!

burn-core/src/module/param/primitive.rs
varonroy: Done!

PR #1188, opened 2024-01-29
Feat/fusion/wgpu autotune

### LINKED ISSUE #1112: Support autotune of fused compute shaders
When creating new shaders, we should be able to execute them with different launch options based on benchmarks performed by the autotune system.

nathanielsimard:
Fix #1112


louisfd (approved):
Awesome, everything looks good to me

PR #1189, opened 2024-01-30
[xtask] Update xtask publish task

Luni-4:
### Checklist

### Changes

This PR updates `xtask` publish task and fixes some format inconsistencies and typos


Luni-4:
@syl20bnr 

I think we should also refactor the `publish.rs` file using the new structure, there is code duplication in there

syl20bnr (approved):
👍

Yes you are right, the file can be refactored to use the functions in utils.

Issue #1130, opened 2024-01-11
ONNX


wanglong001:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

antimora:
I will look into this soon. I have been busy with PyTorch import implementation.

polarathene:
## Reproduction

For reference, using the current git version of Burn (_has some improvements since the latest release tag [`v0.17.1`](https://github.com/tracel-ai/burn/releases/tag/v0.17.1)_).

Reproduction uses [this `compose.yaml`](https://github.com/tracel-ai/burn/issues/1714#issuecomment-2952373124).

No specific ONNX model linked, assuming YOLOv5 (v7.0 release): https://github.com/ultralytics/yolov5/releases/tag/v7.0

```bash
ONNX_MODEL_URL='https://github.com/ultralytics/yolov5/releases/download/v7.0/yolov5n.onnx' \
docker compose run --rm -it burn-onnx-import cargo build
```

```rust
   INFO onnx_ir::from_onnx: Parsing ONNX file: src/model/example.onnx
  DEBUG onnx_ir::from_onnx: Number of nodes: 263
  DEBUG onnx_ir::from_onnx: Number of inputs: 1
  DEBUG onnx_ir::from_onnx: Number of initializers: 120
  DEBUG onnx_ir::from_onnx: Number of outputs: 1
  DEBUG onnx_ir::from_onnx: Opset domain: "<default>", version: 17
  ERROR burn_import::logger: PANIC => panicked at /root/.cargo/git/checkouts/burn-6c277d792b0d5d7a/1f5ccd4/crates/onnx-ir/src/ir.rs:38:33:
  invalid tensor model.0.conv.weight
```

antimora:
I checked, it's failing on a brand new issue:

```
DEBUG onnx_ir::from_onnx: Adding node "reshape6"
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "Concat"
DEBUG onnx_ir::from_onnx: Renaming node "/model.24/Concat_3" to "concat17"
DEBUG onnx_ir::rank_inference: Inferring rank for node: concat17
DEBUG onnx_ir::node::concat: Concat rank inference for node concat17
DEBUG onnx_ir::node::concat: Concat using tensor input rank for concat17: 3
[...]

PR #1208, opened 2024-01-31
Add LeakyReLu implementation

KarstenB:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed. All complained about a typo and a non-working unit test (that is already failing on main for me). But std and no-std both worked.

### Related Issues/PRs

I implemented this because I received the error message in #1130. I then figured out that for my Model LeakyReLu was missing.

### Changes

I implemented the LeakyReLu activation function. I have no idea if the way it is done is the most efficient, but it gets the job done. 

### Testing

I wrote a small unit test in my code for the new LeakyRelu to ensure that it works as intended. I couldn't find a good place within the burn framework, so I would be happy to put it there if someone points me to it.

Also my model that is using the ReLu produces the exact same values as the python version.


antimora (commented):
Thanks for the addition. Could you also add the following tests?

1. ONNX file testing. Please see this crate: https://github.com/tracel-ai/burn/tree/main/burn-import/onnx-tests
2. Tensor test: https://github.com/tracel-ai/burn/tree/main/burn-tensor/src/tests/activation (use this PR on how to add test: https://github.com/tracel-ai/burn/pull/260/files). To test tensor crate, you'll to cd one of the backends, e.g. burn-ndarray and run test (cargo test). It will import the tensor tests.

antimora:
Regarding the typo test fail. We fixed it recently, please update from the main.

antimora:
@nathanielsimard do we add a module for LeakyRelu as well just like other activation functions. I was looking at our book on tensor and I couldn't see anything on activation functions for the tensor but only for nn (https://burn.dev/book/building-blocks/module.html#built-in-modules). 

If so, we also need to update our book here: https://github.com/tracel-ai/burn/blob/main/burn-book/src/building-blocks/module.md#built-in-modules

antimora:
Code coverage can guide what should be tested.

burn-tensor/src/tensor/api/float.rs
```diff
@@ -251,6 +251,15 @@ where
...
 
+    /// Applies the leaky relu function to the tensor.
+    /// The leaky relu function is defined as `x` if `x > 0` and `alpha * x` otherwise.
+    pub fn leaky_relu(self, alpha: f64) -> Self {
+        Self::new(B::leaky_relu(
+            self.primitive,
+            crate::ElementConversion::elem(alpha),
+        ))
+    }
+
```
nathanielsimard, line 262: Activation functions aren't callable on float tensors, so I would remove that function.

burn-tensor/src/tensor/activation/base.rs
```diff
@@ -8,6 +8,13 @@ pub fn relu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
     tensor.relu()
 }
 
+/// Applies the leaky rectified linear unit function.
+///
+/// f(x) = alpha * x for x < 0, f(x) = x for x >= 0
+pub fn leaky_relu<const D: usize, B: Backend>(tensor: Tensor<B, D>, alpha: f64) -> Tensor<B, D> {
+    tensor.leaky_relu(alpha)
```
nathanielsimard, line 15: The backend function should be called instead:

```rust
B::leaky_relu(tensor, alpha)
```

burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -8,6 +9,48 @@ use super::FloatTensor;
...
+    /// The output tensor.
+    fn leaky_relu<const D: usize>(
+        tensor: FloatTensor<B, D>,
+        alpha: super::FloatElem<B>,
+    ) -> FloatTensor<B, D> {
+        let mask = B::float_lower_elem(tensor.clone(), 0.elem());
+        let ones = B::float_ones(Shape { dims: [1; D] }, &B::float_device(&tensor));
+        let alpha_tensor = B::float_full(Shape { dims: [1; D] }, alpha, &B::float_device(&tensor));
+        let mul_mask = B::float_mask_where(ones, mask, alpha_tensor);
+        B::float_mul(tensor, mul_mask)
```
nathanielsimard, line 30: This seems wrong, the shapes of `ones` and `alpha` should be the same as the inputs.

```rust
        let mask = B::float_lower_elem(tensor.clone(), 0.elem());
        let alpha_tensor = B::float_mul(tensor.clone(), alpha.elem());
        
        // Update the tensor where the values are `> 0` by `tensor * alpha`.
        B::float_mask_where(tensor, mask, alpha_tensor)
```

nathanielsimard (changes requested):

burn-tensor/src/tensor/activation/base.rs
KarstenB: Done, except for the fact that I had to do some more conversions now.

burn-tensor/src/tensor/ops/activation.rs
KarstenB: Done, but used float_mul_scalar

KarstenB:
Thanks for the feedback. I applied it as good as I can, but I am not sure I understand the comment about the Module. Let me know if there is anything more that I need to do.

burn-tensor/src/tensor/api/float.rs
KarstenB: Done

burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -8,6 +8,49 @@ use super::FloatTensor;
...
+    /// # Returns
+    ///
+    /// The gradient.
+    fn leaky_relu_backward<const D: usize>(
+        output: FloatTensor<B, D>,
+        grad: FloatTensor<B, D>,
+        alpha: super::FloatElem<B>,
+    ) -> FloatTensor<B, D> {
+        let mask = B::float_lower_elem(output.clone(), 0.elem());
+        let scaled_tensor = B::float_mul_scalar(output.clone(), alpha.elem());
```
KarstenB, line 48: I am not quite sure whether I should take grad here instead of output...

KarstenB: That most definitely should be grad.

burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -8,6 +8,49 @@ use super::FloatTensor;
...
+    ///
+    /// # Returns
+    ///
+    /// The gradient.
+    fn leaky_relu_backward<const D: usize>(
+        output: FloatTensor<B, D>,
+        grad: FloatTensor<B, D>,
+        alpha: super::FloatElem<B>,
+    ) -> FloatTensor<B, D> {
+        let mask = B::float_lower_elem(output.clone(), 0.elem());
```
nathanielsimard, line 47: I feel like this is maybe wrong. This is supposed to be the same mask as during the forward pass, but it's calculated on the output? Maybe I'm wrong though. @louisfd 

I think this will work if the alpha is a positive float, but otherwise we lose the sign information.

antimora: tagging @louisfd

louisfd: You're right Nathaniel, if alpha<0 this won't work. However I don't think LeakyRelu is ever used with a negative alpha. We _could_ maybe assert that alpha is positive. 
But, since it would be easy to save the input as state instead of the output to cover the general case, I say we do that.

louisfd (changes requested):

burn-tensor/src/tensor/ops/activation.rs
KarstenB: I know that other libraries are checking for a negative Alpha and raising an exception if they encounter one. The reasoning for that is that the autodiff could not handle then negative case. I am fine with implementing either one. 

Just to be sure, you mean to implement it like this:
```rust
let mask = B::float_lower_elem(grad.clone(), 0.elem());
```
correct?

louisfd: Hi @KarstenB 
Not exactly, grad doesn't carry that information. For this function it's only a matter of renaming the argument `output` to `input`. Then it's where we call `leaky_relu_backward` that we must pass the input instead of the output. 

I think you haven't done that part yet but it should be easy since you can copy relu. In `crates/burn-autodiff/src/ops/activation.rs` you'll find the relu function, you can copy it and straightforwardly rename `relu` occurrences to `leaky_relu`. 

Then what matters is the `state` variable. At the bottom you'll find `let state = prep.checkpoint(&tensor);` which means the state is actually the input. Above you'll find `B::relu_backward(state, grad)`, so `state` is the input. I realize there's actually a misnaming for relu as well, because the input is given but `relu_backward`'s first arg is named output. It matters less for relu because the input is <0 and the output is ==0, and in relu_backward we use lower or equal. Because relu_backward already takes the input then you won't have to change anything else than the names in the autodiff function.

antimora:
LeakyRelu module was added. This RP will be about ONNX OP addition. I can update the outstanding changes and merge them.

KarstenB:
> LeakyRelu module was added. This RP will be about ONNX OP addition. I can update the outstanding changes and merge them.

Feel free to use my code in whatever way you want to! I currently don't have the bandwidth to continue working on this, but will be extremely happy when it exists as a feature :)

antimora:
OK. I have kept the new activation function minus `backward` impl. I updated the prior LeakyRelu module with the new activation function.

@nathanielsimard , @louisfd, please review it again.

antimora (approved):
I made the necessary fixes and now we need more approval from others.

crates/burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -8,6 +8,27 @@ use super::FloatTensor;
...
+    ///
+    /// The output tensor.
+    fn leaky_relu<const D: usize>(
+        tensor: FloatTensor<B, D>,
+        negative_slope: super::FloatElem<B>,
+    ) -> FloatTensor<B, D> {
+        let mask = B::float_lower_elem(tensor.clone(), 0.elem());
+        let scaled_tensor = B::float_mul_scalar(tensor.clone(), negative_slope.elem());
+
+        // Update the tensor where the values are `> 0` by `tensor * negative_slope`.
```
laggui, line 28: The comment seems incorrect but the actual implementation looks fine.

I think it should be: "where the values are `< 0`"

laggui (approved):
Implementation LGTM! Just a minor change request related to a comment but I'm approving in advance.

laggui (approved):

nathanielsimard (approved):

KarstenB:
@antimora Thanks a lot for finishing this up! Really grateful to see this being merged!

antimora:
@KarstenB Thank you for your contribution! Ended up using all of your code - just had to align with the existing code.

PR #1209, opened 2024-01-31
`NaN` and `Inf` detection in `assert_approx_eq`

skewballfox:
### Checklist

### Changes

small bug fix, so that now `assert_approx_eq` will detect `NaN` and `Inf` values and pass them if `a` and `b` are the same, and fail if they are not. 

right now, if one or either side is `NaN` or `Inf`, the err will be `NaN`. Any comparison with `NaN` is false, so some test pass when they shouldn't. 

I need to run `./run_checks` all, to confirm, but as of last night the only new test to fail were the candle powf functions, which is a bug on their end (issue [here](https://github.com/huggingface/candle/issues/1640)). 

### Testing

ran `./run_checks all` to confirm that the test now fails when it should: when there is a difference between the two sides that currently isn't caught (because the difference is incalculable)


louisfd (approved):
The code is perfect, so I approved. Unfortunately we will need to wait for Candle issue for the CI to pass.

nathanielsimard:
We should remove the powf macro in the candle backend instead. (comment line 79 of `burn-candle/src/lib.rs`  and add an issue. @skewballfox

skewballfox:
> We should remove the powf macro in the candle backend instead. (comment line 79 of `burn-candle/src/lib.rs` and add an issue. @skewballfox
> 
> This PR should be merged way before we wait for a fix from Candle.

I removed the powf test and created #1237

louisfd:
> > We should remove the powf macro in the candle backend instead. (comment line 79 of `burn-candle/src/lib.rs` and add an issue. @skewballfox
> > This PR should be merged way before we wait for a fix from Candle.
> 
> I removed the powf test and created #1237

You'll have to deactivate `burn_tensor::testgen_powf_scalar!();` too because it fails as well in the CI

PR #1212, opened 2024-01-31
[doc] Fix burn-tch doc for burn crate

syl20bnr:
### Checklist

### Changes

Set the correct feature for docs build.

### DIFF
--- a/burn-core/Cargo.toml
+++ b/burn-core/Cargo.toml
@@ -13,31 +13,41 @@ version.workspace = true
 [features]
 default = [
     "std",
-    "burn-dataset?/default",
-    "burn-ndarray?/default",
     "burn-candle?/default",
-    "burn-wgpu?/default",
-    "burn-fusion?/default",
-    "burn-tch?/default",
-    "burn-dataset?/default",
     "burn-common/default",
+    "burn-dataset?/default",
+    "burn-fusion?/default",
+    "burn-ndarray?/default",
+    "burn-tch?/default",
     "burn-tensor/default",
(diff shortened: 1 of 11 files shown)


burn-core/Cargo.toml
```diff
@@ -60,6 +60,7 @@ blas-netlib = ["burn-ndarray?/blas-netlib"]
 
 ndarray = ["burn-ndarray"]
 tch = ["burn-tch"]
+tch-doc-only = ["burn-tch/doc"]
```
nathanielsimard, line 63: I would rename the feature flag to `doc-only` for all crates (burn-tch, burn-core and burn)

nathanielsimard (changes requested):

syl20bnr:
Switching to draft.
This requires more thoughts than expected.

syl20bnr:
Not completely sure about this but this is an attempt to add the basis to make enabled features during docs.rs builds more explicit.

burn-tensor/Cargo.toml
```diff
@@ -12,6 +12,10 @@ version.workspace = true
 
 [features]
 default = ["std"]
+doc = [
+  "default",
+  "experimental-named-tensor",
```
nathanielsimard, line 17: Not sure I would include this one.

nathanielsimard (changes requested):
I don't think we should include experimental features for now.

syl20bnr:
Removed

nathanielsimard (approved):

PR #1219, opened 2024-02-01
Add a semantic versioning checker

Luni-4:
### Checklist

### Related Issues/PRs

A semantic versioning checker lints API changes for semver violations. Even this check will run periodically on CI not to stress a lot the entire infrastracture

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -1,0 +1,25 @@
+name: semver-checks
+
+on:
+  schedule:
+    - cron: '0 21 * * THU' # Run every Thursday at 21:00 (UTC)
+  push:
+    tags:
+      - 'v*.*.*' # Run when a new version is being published
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  dependencies:
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: checkout
+        uses: actions/checkout@v4
+
+      - name: Check semantic versioning violations
+        uses: obi1kenobi/cargo-semver-checks-action@v2
+
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -12,14 +12,18 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


Luni-4:
I tried to cross-compile `metal' on `Linux` but it is not possible as stated [here](https://github.com/gfx-rs/metal-rs?tab=readme-ov-file#documentation). How can I disable that compilation? Is there a way to automatically **not** enable the compilation on a system different from `MacOs` or I have to use `DISABLE_WGPU` environment variable?

Luni-4:
Ok, fixed everything

syl20bnr:
Thank you, that's a valuable check to have. Do we want to activate it before we hit the `1.0` mark ? Or is this action smart enough to understand projects in `0.x` ?

Luni-4:
It should work, it uses the same `vulnerabilities.yml` semantic and that actions run https://github.com/tracel-ai/burn/actions/runs/7731579136 when we published `0.12` version.

syl20bnr (approved):
Let's try this one, thank you 👍🏻

nathanielsimard (commented):
I'm not sure if this is a valuable check for us, since every release will be a breaking change until v1.0. Adding a check for patches seems a bit too much in term of process.

What do you think @antimora @syl20bnr ?

antimora:
> I'm not sure if this is a valuable check for us, since every release will be a breaking change until v1.0. Adding a check for patches seems a bit too much in term of process.
> 
> What do you think @antimora @syl20bnr ?

Probably the value to know what APIs are breaking. From [the docs](https://github.com/obi1kenobi/cargo-semver-checks-action), it appears it's use before a publish operation. 

I think it's also useful once we go 1.0. So it might be useful to get accustomed.

antimora (approved):

Luni-4:
> I'm not sure if this is a valuable check for us, since every release will be a breaking change until v1.0. Adding a check for patches seems a bit too much in term of process.
> 
> What do you think @antimora @syl20bnr ?

Actually it runs perfectly comparing `0.12.` and `0.11.1`, so we can use that before `1.0`. I agree with @antimora about getting accustomed, and yep, it runs **before** a publish operation

nathanielsimard:
Ok then, let's try it!

Issue #1198, opened 2024-01-31
Add common LR schedulers


laggui:
### Feature description

Right now we support LR schedulers but only provide a constant schedule or the Noam schedule.

Would be nice if we added a couple of common/popular LR schedulers:
- [x] Step LR
- [x] Cosine LR
- [x] Exponential LR
- [x] Linear LR
- [ ] Sequential LR schedulers (useful for combining things like warmup + cosine or linear decrease schedule)

miknyko:
Hi, just happen to know those scheduler a liitle bit. I could have this.

rubenjr0:
I could have a go at the step, exponential and cosine (maybe some variants, like with annealing) ones, since the linear one is already merged.

antimora:
@rubenjr0 Yes, please go ahead.

towerpark:
Hi,

I have a question about the Step LR scheduler, which I'm making a PR for. My question is, say the initial learning rate and gamma used to create the config instance is 0.5 and 0.1 respectively, should the first value `LrScheduler::step` returns be 0.5 or 0.05?

The reason why I'm asking it is the following schedulers never return the `initial_lr` value in the config class:
　[Exponential LR](https://github.com/tracel-ai/burn/blob/a9f941d403686abdb5ff2f3bea80219ecd6228a9/crates/burn-core/src/lr_scheduler/exponential.rs#L54)
　[Linear LR](https://github.com/tracel-ai/burn/blob/a9f941d403686abdb5ff2f3bea80219ecd6228a9/crates/burn-core/src/lr_scheduler/linear.rs#L60)
　[Cosine LR](https://github.com/tracel-ai/burn/blob/a9f941d403686abdb5ff2f3bea80219ecd6228a9/crates/burn-core/src/lr_scheduler/cosine.rs#L66)

I read the code, and it seems that the only way to obtain a learning rate is to call the `step` method. The first time `step` gets called, it uses the initial value to compute and return the next learning rate (I am new to Burn, so please correct me if I am missing something obvious).

Given the name `initial_lr` of the field, I think it's natural to expect that it will be used as the first learning rate value, so I am inclined to choose 0.5 as the answer to my question, but I am not sure about not following what existing code does.

Any advice would be greatly appreciated.

rubenjr0:
@towerpark After taking a look at the code it looks like the first step call would return 0.05.

In the tests for `ExponentialLR` we can see how the first step returns a learning rate smaller than the initial one:
```rust
[...]

PR #1220, opened 2024-02-01
Added StepLR scheduler

miknyko:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/1198

### Changes

Hi, I added the step lr according to the pytorch implementation.  I will add more LR schdulers later this week .
:D

### Testing

Testing unit added


nathanielsimard (approved):

burn-core/src/lr_scheduler/step.rs
```diff
@@ -0,0 +1,87 @@
...
+    gamma: f64,
+}
+
+/// Decays the learning rate of each parameter group by gamma every step_size epochs.
+#[derive(Clone, Debug)]
+pub struct StepLrScheduler {
+    init_lr: LearningRate,
+    step_size: i64,
+    gamma: f64,
+    step: f64,
```
laggui, line 27: Any reason why the step is `f64` instead of unsigned int like `usize`?

laggui (changes requested):
Overall the implementation looks good to me! Good job 👍 

Just have a few minor comments below.

burn-core/src/lr_scheduler/step.rs
nathanielsimard: I think in the state, it's just to avoid casting.

laggui: But in `to_record` the step is actually cast as `usize`, and then in `load_record` cast back as `f64`. Maybe I'm missing something, but it just seems like we could keep it `usize` all the way and avoid casting, no?

nathanielsimard: But it's not casted in the `step` function 😅

nathanielsimard: We could use `usize` but I would also update it in the Noam scheduler for consistency.

laggui: > But it's not casted in the `step` function 😅

Not when incremented, but when it is actually used it seems that it is also casted:

```rust
self.gamma.powf((self.step as i64 / self.step_size) as f64)
```

> We could use `usize` but I would also update it in the Noam scheduler for consistency.

Ahhhhh so that's where it comes from! I didn't realize that was the case.

miknyko:
Sorry guys, quite busy past two weeks, I will continue working on the schedulers

laggui:
> Sorry guys, quite busy past two weeks, I will continue working on the schedulers

No worries 🙂

PR #1223, opened 2024-02-01
Enable burn fusion by default

nathanielsimard:
To simplify the utilization of the new fusion backend and enhance performance by default, we will enable the fusion backend decorator using a feature flag instead of requiring users to add it, which may be forgotten.

```rust
type MyBackend = Fusion<Wgpu>; // Old
type MyBackend = Wgpu; // New
```

To achieve this, we will rename the current backend to `WgpuBackend` and export the type alias `Wgpu`, pointing to the correct backend type. We will also update the documentation for each with the appropriate modifications.


burn-wgpu/src/lib.rs
```diff
@@ -33,11 +33,49 @@ pub use graphics::*;
...
+/// This backend can target multiple graphics APIs, including:
+///   - [Vulkan] on Linux, Windows, and Android.
+///   - [OpenGL](crate::OpenGl) on Linux, Windows, and Android.
+///   - [DirectX 12](crate::Dx12) on Windows.
+///   - [Metal] on Apple hardware.
+///   - [WebGPU](crate::WebGpu) on supported browsers and `wasm` runtimes.
+///
+/// # Notes
+///
+/// This version of the [wgpu] backend doesn't uses [burn_fusion] to compile and optimize streams of tensor
```
louisfd, line 67: use

louisfd (approved):
LGTM

PR #1226, opened 2024-02-01
Fix problems with batch norm on LibTorch backend

nathanielsimard:
We have to manually check if the device is the same, since when it's the case, we need to keep the same storage reference and not create a new one.


louisfd (approved):
LGTM

PR #1228, opened 2024-02-01
Fix training dashboard metrics switch

nathanielsimard:
On Linux & Mac OS switching metrics were not working.


louisfd (approved):
Tested on Mac OS, it works

PR #1233, opened 2024-02-02
[CI] Add workflow to group dependabot PRs

syl20bnr:
This workflow uses the following GitHub action: https://github.com/github/combine-prs

It groups all the pending dependabot PRs into a single one if possible. It is configured to run every Monday at 1am and it can also be triggered on demand. 

We might need additional setup for authorizations.

### DIFF
--- a/.github/workflows/combine-dependabot-prs.yml
+++ b/.github/workflows/combine-dependabot-prs.yml
@@ -1,0 +1,23 @@
+name: Combine Dependabot PRs
+
+on:
+  schedule:
+    - cron: '0 6 * * MON' # Monday at 6:00am UTC
+  workflow_dispatch: # allows to manually trigger the workflow as well
+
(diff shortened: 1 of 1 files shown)


.github/workflows/combine-dependabot-prs.yml
```diff
@@ -0,0 +1,23 @@
+name: Combine Dependabot PRs
+
+on:
+  schedule:
+    - cron: '0 1 * * 1' # Monday at 01:00
```
Luni-4, line 5: Can we use `Mon` instead of `1`? More understandable

Luni-4 (changes requested):
Wow, seems really useful, thanks! 

Just a simple change

Luni-4 (approved):

Luni-4 (approved):

nathanielsimard (approved):

PR #1235, opened 2024-02-02
[CI] Simplify CI binaries updating 

Luni-4:
### Checklist

### Changes

- Use `latest` keyword instead of a specific tag for the binaries that allow this trick
- Move the version variable at the top of the `yml` files, so it's simpler to update their versions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,6 +48,12 @@ env:
   # Corresponds to https://github.com/gfx-rs/ci-build/releases
   CI_BINARY_BUILD: "build18"
 
+  # Typos version
+  TYPOS_VERSION: "1.16.20"
+
+  # Grcov version
+  GRCOV_VERSION: "0.8.18"
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -139,10 +143,9 @@ jobs:
       - name: Install grcov
         if: runner.os == 'Linux' && matrix.rust == 'stable' && matrix.test == 'std'
         env:
-          GRCOV_LINK: https://github.com/mozilla/grcov/releases/download
-          GRCOV_VERSION: v0.8.18
+          GRCOV_LINK: https://github.com/mozilla/grcov/releases/latest/download
```
syl20bnr, line 146: I would prefer that we stick with a fixed grcov version at the top of the file as well to be consistent. This way we can decide when we upgrade and not having the tool to update itself behind our back which can lead to potential compatibility breakage happening at any time without us being in control.

syl20bnr (changes requested):
Let's always pin a version when we can.

.github/workflows/test.yml
Luni-4: Yep, I had pinned binaries versions for your exact reasons, but I've seen that over the time no updates have been done at all. We have been using `grcov` for three/four months and it has never been updated. `dependabot` cannot do that either. I tried to look for an action that updates environment variables but without any success.

I was thinking to create an action that opens an issue every now and then to remember a CI update. It should be feasible and the noise would be very low. This is an approach.

Otherwise we can accept this PR as-is. If a breakage were to happen, that would be in the binary step, so we can pin the binary to the previous version fixing the problem in no time

syl20bnr: It's better to pin the version. Then we need to make sure we review the versions of our tools regularly. I like your proposal to create an issue on a schedule for this purpose. Then to make the process less tedious we can add an xtask command to check for new versions when it is trivial to code it.

Luni-4: An xtask command is another fantastic idea. Parsing `yml` file in `Rust` is not complex. We can extract all versions and check on `crates.io` whether new versions are available. For now I can add the remember action

syl20bnr (approved):

nathanielsimard (approved):

PR #1236, opened 2024-02-02
Unsqueeze op

skewballfox:
### Checklist

### Related Issues/PRs

None, just adding `unsqueeze_dims` to get us closer to full `ONNX` support

### Changes

adds a new function `unsqueeze_dims` to `burn-tensor/src/tensor/api/base`, which takes a slice of isizes as the second argument (to make it compatible with the [onnx unsqeeze op](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Unsqueeze). 

There are issues I'm encountering on burn import. Opening the models generated by both Torch and Onnx, neither are storing the values of second argument `axes`, so the only way to properly test it is for those to be a constant.

This might pose an issue with dimension inference in the case where the axis argument isn't constant, as the only way to know what that argument was supposed to be is to infer it from the output dimensions. 

### Testing

I adapted the examples from onnx operation spec to use as test for unsqueeze_dims, and I've generated onnx models using both torch and onnx. The `OnnxOpData` dataclass I'm hoping might be useful in the future for other models, It's using onnx helper and runtime to generate the models directly without using torch as an intermediary. 

This is useful when there is a discrepancy between a torch operation and an onnx one, such as the case with unsqueeze where torch only supports a single axis argument, and onnx supports an array of dims.  

Outside of that, it might be useful for scripting the creation of additional operators a bit more, potentially making it possible to generate models for new operators in bulk.


burn-import/out/reshape.rs
antimora, line 1: This file is probably checked in accidentally.

burn-import/src/burn/node/base.rs
```diff
@@ -146,6 +149,7 @@ impl<PS: PrecisionSettings> Node<PS> {
             Node::MaxPool2d(_) => "max_pool2d",
             Node::Reshape(_) => "reshape",
             Node::Unary(unary) => unary.kind.as_str(),
+            Node::Unsqueeze(_) => "unsqueeze_dims",
```
antimora, line 152: We should just name `unsqueeze` instead. These are string representation of enums.

burn-import/src/burn/node/reshape.rs
```diff
@@ -21,6 +21,7 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for ReshapeNode {
     }
 
     fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
+        println!("ReshapeNode::forward");
```
antimora, line 24: `println` is left.

burn-import/src/burn/node/unsqueeze.rs
```diff
@@ -0,0 +1,96 @@
...
+//         graph::BurnGraph,
+//         node::{test::assert_tokens, unsqueeze::UnsqueezeNode},
+//         TensorType,
+//     };
+
+//     #[test]
+//     fn test_codegen_nodes() {
+//         let mut graph = BurnGraph::<FullPrecisionSettings>::default();
+
+//         graph.register(UnsqueezeNode::new(
```
antimora, line 54: Tests are commented out.

burn-import/src/onnx/dim_inference.rs
```diff
@@ -284,31 +285,125 @@ fn mean_update_outputs(node: &mut Node) {
...
+        ),
+        _ => panic!("Unsqueeze: invalid input types"),
+    };
+    //need output way up here to avoid borrowing issues
+    let (mut tensor, output_shape) = match &node.outputs[0].ty {
+        ArgType::Tensor(tensor) => (tensor.clone(), tensor.shape.clone()),
+        _ => panic!("Unsqueeze: invalid output types"),
+    };
+    let mut remap_node = false;
+    println!("{:?}", node);
```
antimora, line 316: `println!` is left.

burn-import/src/onnx/from_onnx.rs
```diff
@@ -59,8 +60,16 @@ pub fn parse_onnx(onnx_path: &Path) -> ONNXGraph {
 
     // Convert the nodes
     let mut nodes: Vec<Node> = vec![];
+    println!("onnx_model.graph.node: {:#?}", onnx_model.graph.node);
```
antimora, line 63: println! is left.

burn-import/src/onnx/from_onnx.rs
```diff
@@ -173,6 +182,49 @@ fn move_initializer_data(initializer: &TensorProto, input: &mut Argument) {
     }
 }
 
+//this is an extremely hacky temporary solution while I figure out how to properly handle this
+//situation
```
antimora, line 186: Will this be handled in this PR or later? We should replace with a TODO with instructions if not handled in this PR.

Also please explain the context why this is needed and if any potential issues it might cause down the road.

burn-import/src/onnx/from_onnx.rs
```diff
@@ -335,7 +387,7 @@ fn rename_inputs(
 
     for node in nodes.iter_mut() {
         let mut counter = 1;
-
+        println!("node: {:#?}", node);
```
antimora, line 390: println! is left.

burn-import/src/onnx/op_configuration.rs
```diff
@@ -538,6 +538,29 @@ pub fn reshape_config(node: &Node) -> Vec<i64> {
...
 }
 
+//Note this function should only execute if the second input is a constant
+//if it wasn't and the output shape was known, unsqueeze has been remapped to reshape
+pub fn unsqueeze_config(node: &Node) -> Vec<i64> {
+    // if node.inputs.len() != 2 || node.inputs[1].value.is_none() {
+    //     println!("{:?}", node.inputs);
+    //     panic!("Reshape: shape tensor must be present");
+    // }
+    println!("{:?}", node.inputs);
```
antimora, line 548: Dead code and println!

burn-import/src/onnx/to_burn.rs
```diff
@@ -454,6 +456,14 @@ impl ONNXGraph {
 
         ReshapeNode::new(input, output, shape)
     }
+    fn unsqueeze_conversion(node: Node) -> UnsqueezeNode {
+        println!("YEEET");
```
antimora, line 460: println! is left

antimora (changes requested):
Thank for adding this new OP!

There are minor issues that need to be cleaned up because merging. I haven't verified if all cases are covered - it seems the implementation was not trivial.

antimora:
Added @nathanielsimard as a reviewer the new tensor OP.

burn-import/src/burn/node/unsqueeze.rs
skewballfox: Is there any reason to include test here, rather than strictly testing via the model?

burn-import/out/reshape.rs
skewballfox: `git rm --cached -r burn-import/out`

burn-tensor/src/tensor/api/base.rs
```diff
@@ -304,6 +304,80 @@ where
...
+    /// fn example<B: Backend>() {
+    ///     let device = Default::default();
+    ///     let tensor = Tensor::<B, 3>::ones(Shape::new([3, 4,5]), &device);
+    ///     let tensor: Tensor<B, 5> = tensor.unsqueeze_dims(&[0, 4]);
+    ///     println!("{:?}", tensor.shape());
+    ///     // Shape { dims: [1, 3, 4, 5, 1] }
+    /// }
+    /// ```
+    pub fn unsqueeze_dims<const D2: usize>(self, dims: &[isize]) -> Tensor<B, D2, K> {
+        let mut new_dims = [1; D2];
```
nathanielsimard, line 324: I feel like this can be implemented in a simpler way:

```rust
let new_dims = [1; D2];
let mut counter = 0;
for i in 0..D2 {
    if !dims.contains(i) {
        new_dims = old_dims[counter];
        counter += 1;
    }
}
tensor.reshape(new_dims);
```

No allocation, and `contains` is extremely fast on small slices (less than 10 elems).

nathanielsimard (changes requested):
Just commented on the added op, which should be simplified.

burn-tensor/src/tensor/api/base.rs
skewballfox: I could avoid the allocation by making the second argument `&'op mut [isize]`, and then mutate the new dims in place.

I just realized that the dims can also contain duplicates. The Onnx spec never specified that the values had to be unique, so I wrote it so that there could be duplicates. 

essentially the thinking was unsqueeze dims was equivalent to doing a series of single dim unsqueezes, and if you executed those unsqueezes, it wouldn't matter what order the operations happened: 2 values of `-1` would just result in two axes in the resulting tensor at the end

burn-import/src/burn/node/unsqueeze.rs
antimora: Yes, the plan is that code generation is independent from ONNX graph. The ONNX file testing is end-to-end.  We can add other computation graphs.

skewballfox: I uncommented the test and fixed it. Confirmed it's working locally.

antimora (approved):
From my end it looks good. 

Thank you for the improvements and OP addition. Burn ONNX is getting better!

nathanielsimard (approved):

Issue #1234, opened 2024-02-02
Support signed value for `Tensor::arange`


Nikaidou-Shinku:
### Feature description

Currently `Tensor::arange` accepts a `Range<usize>`, which can only generate unsigned value.
It will be great if we also support signed value.

### Suggest a Solution

Since the `usize` value is converted into `i64` by `(i as i64).elem()` in `FloatTensorOps::float_arange_step`, maybe we can make `Tensor::arange` directly accepts a `Range<i64>`.
If that's okay, I can submit a PR for it.

louisfd:
It's a good idea! Please go ahead

PR #1238, opened 2024-02-02
Support signed value for `Tensor::arange`

Nikaidou-Shinku:
### Checklist

### Related Issues/PRs

#1234

### Changes

Make `Tensor::arange` and all internal methods accept a `Range<i64>` instead of `Range<usize>` to support ranges of signed value.

### Testing

`bash ./run-checks.sh all` was passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1921,7 +1921,7 @@ dependencies = [
  "futures-sink",
  "futures-util",
  "http",
- "indexmap 2.2.1",
+ "indexmap 2.2.2",
  "slab",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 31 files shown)


antimora (changes requested):
1. A new unit test should be added to test all new edge cases (negatives and 0, overflow range). 

2. Extra caution when casting usize to i64 because it overflow.

nathanielsimard:
> 1. A new unit test should be added to test all new edge cases (negatives and 0, overflow range).
> 2. Extra caution when casting usize to i64 because it overflow.

I think we should check if we can replace the usize by i64 instead of casting in most places.

Nikaidou-Shinku:
I moved the functions about arange to `int_tensor.rs`, and add some test cases for 0 and negative numbers.
Regarding overflow, I don't have any good ideas. I noticed that for most cases where arange is used in the test cases, the range is taken from the shape of tensors. In this case, when an overflow occurs, I don't know what else to do but crash the program.

nathanielsimard (approved):

antimora (approved):
LGTM

Issue #936, opened 2023-11-07
Autodiff: checkpointing strategy


louisfd:
In autodiff, we should have a checkpointing strategy for better memory consumption (see for instance [https://www-sop.inria.fr/tropics/papers/DauvergneHascoet06.pdf](https://www-sop.inria.fr/tropics/papers/DauvergneHascoet06.pdf)) . 

Currently, for most operations run in the forward pass, a state will be saved for the backward pass. The state often consists of a few tensors, so it is needless to say that they accumulate and use a lot of memory. 

A way to use less memory for the backward pass would be to, instead of having kept the state in memory, recompute the forward pass of the operation to re-obtain the state, just before computing its backward pass. This will lead to more computations, but less memory consumption. 

This leads to a tradeoff between compute and memory. Some operations, like matrix multiplication, are "compute-bound", meaning the bottleneck is generally the actual computations, while some, such as element-wise multiplication, are "memory-bound", meaning the computation is actually so simple that the moving of data is the bottleneck. 

For compute-bound operations, it is better to keep the state than to recompute. But for memory-bound operations, we would benefit from recomputing. 

Also, if many operations are tagged as memory-bound, this will greatly help fusing kernels with Burn-Fusion, which will be able to fuse kernels transparently during the backward pass. 

The current strategy, where every state is saved, would simply become a specific case of the new strategy, where everything is considered compute-bound.

AuruTus:
Hi, I'm wondering how the toggle of this strategy should be added into burn's ad graph? The AD tool `tapenade` behind that paper seems to have an IR and a pair of directives to control which snippets should be treated with checkpoints.

louisfd:
Hi @AuruTus 
To be honest, haven't read the paper. I just thought the figures seem to explain well the concept of checkpointing, I'm not sure if we should follow what they did or conceptualize our own checkpointing strategy.  
We plan on tackling that issue early 2024; for now I haven't given it more thought than what is written above!

benjamin-macadam:
If you are discussing check-pointing strategies, it may be worth considering Jax's approach to AD, explained in [You Only Linearize Once](https://arxiv.org/pdf/2204.10923.pdf), since that can shed some light on what is going on with checkpointing. The idea is to break the the vector-jacobian product into two pieces - I'm going to use Haskell type signatures where -o is linear implication and ! means a variable may be reused (e.g. it is a smooth argument).

[...]

PR #1239, opened 2024-02-02
Feat/autodiff/checkpoint

louisfd:
### Checklist

### Related Issues/PRs

WIP for #936 

### Changes

Most of the logic for the autodiff checkpointing strategy. 
At the moment, the code is all independant from the rest of Burn, I suspect it won't pass the CI because everything is unused; thus why I put my PR as a draft. 
It has already undergone several refactorings, so I'm pretty satisfied with the cleanness. 
Next I will of course plug it in all autodiff operations. 

### Testing

Heavily tested in the tests.rs file


burn-autodiff/src/checkpoint/base.rs
```diff
@@ -0,0 +1,172 @@
...
+        // Fetch the state and decrement its number of required
+        let state = self.map.remove(node_id).unwrap();
+        let remaining_n_required = state.n_required() - 1;
+
+        // Downcast the state to whatever it is supposed to be
+        let downcasted = state
+            .get_state_content()
+            .downcast_ref::<T>()
+            .unwrap()
+            .clone();
```
nathanielsimard, line 58: You don't want to clone the state content here, you want to take ownweship of it. 

```rust

   // Fetch the state and decrement its number of required
        let state = self.map.remove(node_id).unwrap();
        let remaining_n_required = state.n_required() - 1;

        // Downcast the state to whatever it is supposed to be
        // If still needed after giving ownership, we copy it back to the hashmap
        if remaining_n_required > 0 {
            let new_stored_state = match state {
                State::Recompute { n_required: _ } => State::Recompute {
                    n_required: remaining_n_required,
                },
                State::Computed {
                    state_content,
                    n_required: _,
                } => State::Computed {
                    state_content,
                    n_required: remaining_n_required,
                },
            };

            let downcasted = new_stored_state
                .to_state_content()
                .downcast_ref::<T>()
                .unwrap()
                .clone();

            self.insert(node_id.clone(), new_stored_state);

            downcasted
        } else {
            let downcasted = state.into_state_content().downcast::<T>().unwrap();

            *downcasted
        }
 ```

You need `to` and `into` variant of `state_content()`

File: burn-autodiff/src/checkpoint/base.rs [hunk 2]
```diff
@@ -0,0 +1,172 @@
...
+
+impl Checkpoint {
+    /// Gives the output of the given node, by recursively asking parents to compute themselves
+    /// or give their pre-computed tensors.
+    pub fn get<T>(&mut self, node_id: NodeID) -> T
+    where
+        T: Clone + Send + Sync + 'static,
+    {
+        self.topological_sort(node_id.clone())
+            .iter()
```
nathanielsimard, line 127: `iter` => `into_iter`

burn-autodiff/src/checkpoint/base.rs
```diff
@@ -0,0 +1,172 @@
...
+pub(crate) struct RetroForwards {
+    map: HashMap<NodeID, Box<dyn RetroForward>>,
+}
+
+impl RetroForwards {
+    /// Executes the [RetroForward] for a given [NodeID] if the node's
+    /// [State] is [State::Recompute], otherwise does nothing.
+    pub fn forward(&self, node_id: &NodeID, inner_states: &mut InnerStates) {
+        if let State::Recompute { n_required: _ } = inner_states.get_ref(node_id).unwrap() {
+            self.map.get(&node_id).unwrap().forward(inner_states);
```
nathanielsimard, line 25: I didn't see where we remove the `RetroForward` from the `map`.

burn-autodiff/src/checkpoint/tests.rs
```diff
@@ -0,0 +1,405 @@
...
+        graph::{Node, NodeID, NodeRef, Requirement},
+    };
+
+    use super::*;
+
+    #[cfg(not(target_os = "macos"))]
+    pub type TestBackend = burn_wgpu::Wgpu<burn_wgpu::Vulkan, f32, i32>;
+
+    #[cfg(target_os = "macos")]
+    pub type TestBackend = burn_wgpu::Wgpu<burn_wgpu::Metal, f32, i32>;
```
nathanielsimard, line 66: Why not just using AutoGraphicsApi?

burn-autodiff/src/checkpoint/tests.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// Number of times required is defined here for tests; normally it should be incremented
+    /// during the forward pass while building the autodiff graph
+    fn insert_recompute(id: NodeID, inner_states: &mut InnerStates, n_required: Option<usize>) {
+        inner_states.insert(
+            id.clone(),
+            State::Recompute {
+                n_required: match n_required {
+                    Some(n) => n,
+                    None => 1,
+                },
```
nathanielsimard, line 81: `n_required.unwrap_or(1)`

burn-autodiff/src/checkpoint/tests.rs
```diff
@@ -0,0 +1,405 @@
...
+        [
+            NodeID::new(),
+            NodeID::new(),
+            NodeID::new(),
+            NodeID::new(),
+            NodeID::new(),
+            NodeID::new(),
+            NodeID::new(),
+        ]
+    }
```
nathanielsimard, line 116: That method isn't helpful, we can inline each `NodeId` where they are needed.

burn-autodiff/src/checkpoint/tests.rs
```diff
@@ -0,0 +1,405 @@
...
+
+    /// Make the leaves for a div tree
+    fn make_leaves<B: Backend>(device: &B::Device, ids: [NodeID; 4]) -> (InnerStates, NodeTree) {
```
nathanielsimard, line 119: Same, the NodeIds should be inlined.

burn-autodiff/src/checkpoint/tests.rs
```diff
@@ -0,0 +1,405 @@
...
+                device,
+            )),
+        );
+        insert_recompute(ids[6].clone(), &mut inner_states, None);
+
+        Checkpoint::new(inner_states, retro_forwards, nodes)
+    }
+    #[test]
+    fn div_lazy_tree_has_expected_leaves() {
+        let device = Default::default();
```
nathanielsimard, line 286: The tests should come _before_ the `helpers`

burn-autodiff/src/checkpoint/tests.rs
```diff
@@ -0,0 +1,405 @@
...
+                state_content: Box::new(out),
+                n_required: states.get_ref(&self.self_id).unwrap().n_required(),
+            },
+        );
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
```
nathanielsimard, line 47: Since you are already in a test module this is unecessary.

burn-autodiff/src/checkpoint/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod base;
+mod state;
+mod tests;
```
nathanielsimard, line 3: You should add the `#[cfg(test)]` here

nathanielsimard (commented):
I find the tests quite confusing, they normaly reflect `how` the code is going to be used, but it seems like the "scenario" is defined in many methods and assumed to be static. 

What I think should improve the code:

1. The "normal" forward pass that generates the Checkpoint struct with maybe some assertions on the `Computed` and `Recompute` tensors.
2. Simulate the backward pass that calculates the "retro forwards" and asserts the results.

The more the forward pass is similar to normal tensor operations, the better.

louisfd:
You should be happier with the new state of the production code. Tests are still convoluted but they were never really meant to outlive the development phase. I will migrate them to clean ones once I can use the autodiff api

PR #1260, opened 2024-02-05
[backend-comparison] Burnbench CLI

syl20bnr:
### Checklist

### Related Issues/PRs

Issue https://github.com/tracel-ai/burn/issues/1072

### Changes

Add a new binary to `backend-comparison` crate called `burnbench`.

There are two mods for this CLI: `term` and `tui`. The UI is a work in progress and must be enabled explicitly with the `tui` feature.

See the README for details about the CLI user experience. I tried to made it auto-discoverable and auto-correct as much as possible. For instance typos show the list of  supported benchmarks or backends with a proposal to fix it:

image

I chose to repeat the arguments `--backend` and `--bench` instead of using comma separated list because the code is simpler and it better leverage `clap`  and `strum`.

I chose l`-b` shorthand for a bench and `-B` for backend because the tool is about benchmarks so I believe that the natural lower case is for a bench. 

I'll  open a different  PR for the upload.

### Testing

Here is two commands to quickly test:

```sh
cargo run --bin burnbench -- list
```

```sh
cargo run --bin burnbench -- run --bench unary --bench binary --backend wgpu-fusion --backend wgpu
```


nathanielsimard (changes requested):
I don't have much comments, seems good in general!

nathanielsimard (approved):

PR #1266, opened 2024-02-05
Removed duplicate files that are causing problems on Mac

antimora:
Removed duplicate readme file that were causing problems on Mac's case in-sensitive file system


syl20bnr (approved):

louisfd (approved):
Thanks, I struggled with those this morning!

PR #1273, opened 2024-02-07
Fix chain pattern matching when multiple patterns are provided

laggui:
### Checklist

### Related Issues/PRs

Follow-up to PR #1269

### Changes

Chaining patterns as described in the other PR need to be applied to the *modified* name (i.e., `new_name`).


antimora (approved):
LGTM

laggui:
As discussed I've decided to add some tests to illustrate an example use case and avoid possible regressions

nathanielsimard (approved):

PR #1274, opened 2024-02-07
[Refactor] Create an Intermediate representation of compute shaders on the GPU

nathanielsimard:
Decouple de data structures used for generating gpu shaders from the ones compiling the shaders to wgsl. I added 2 dialects: `gpu` and `wgsl`, more will come.

There are some places where the compiler is hardcoded, but it's going to be updated and removed in following PRs as we migrate all operations to the `gpu` dialect and support compiling to `wgsl`.

### Todos
- [ ] Update the documentation


louisfd (approved):
Scanned it kinda fast but it's pretty awesome how ops are generic over the compiler

PR #1280, opened 2024-02-08
[Refactor] Just-In-Time Backend

nathanielsimard:
Refactor the `burn-wgpu` crate by introducing a `Runtime` trait. The new trait is composed of all the necessary types to be implemented so that a backend can be made out of it. The `WgpuBackend` has been renamed to the `JitBackend` where the runtime can be passed as an argument as such `JitBackend<WgpuRuntime<GraphicsApi, Float, Int>>`.

For now, some operations don't use the `wgsl::Compiler` but are already defined as a String (with WGSL syntax), so the backend isn't ready to be split into a new crate `burn-jit`. Following PRs will extract the remaining operations into the `gpu::Operation` dialect, and the `wgsl` tokens are going to be defined in the `Display` implementation of the `wgsl` representation. After that, we could start to introduce new Runtimes with their own compute server and compiler.

### DIFF
--- a/burn-wgpu/src/tensor/base.rs
+++ b/burn-wgpu/src/tensor/base.rs
@@ -1,40 +1,40 @@
 use crate::codegen::dialect::gpu::{Elem, Item, Operation, UnaryOperation, Variable};
 use crate::element::WgpuElement;
-use crate::{unary, JitRuntime};
+use crate::{unary, Runtime};
 use burn_compute::client::ComputeClient;
 use burn_compute::server::Handle;
 use burn_tensor::Shape;
 use std::marker::PhantomData;
 
 /// The basic tensor primitive struct.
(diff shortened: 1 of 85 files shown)


burn-wgpu/src/tensor/base.rs
```diff
@@ -1,36 +1,75 @@
...
+    R: Runtime,
+    E: JitElement,
+{
+    /// Compute client for the [runtime](Runtime).
+    pub client: ComputeClient<R::Server, R::Channel>,
     /// The buffer where the data are stored.
-    pub handle: WgpuHandle,
-    /// The shape of the current tensor.
+    pub handle: Handle<R::Server>,
+    /// The shape of the ctensor.
```
louisfd, line 19: you left a c

louisfd (approved):
Looks very good to me

PR #1285, opened 2024-02-09
[backend-comparison] Add GitHub authentication to burnbench CLI

syl20bnr:
### Checklist

### Related Issues/PRs

Issue https://github.com/tracel-ai/burn/issues/1072

### Changes

Add a new `auth` command to `burnbench` in order to create an access token that will be used to authenticate the user when sharing the benchmark results.

image

Add a `--share` argument to the `run`  command to upload the benchmark results. This is a WIP and the actual upload will be added once the server side used the access token to authenticate the user.

Found a crate called [github-device-flow](https://crates.io/crates/github-device-flow) to help with implementing the device flow in order to create a user access token.

For now the token is not refreshed when it expires. This can be implemented later.

Currently the GitHub application is a temporary one called `Burnbench POC` created under my account. Once this is approved and merged we can create the official one in the Tracel org and update the client ID.

### Testing

Would be great if you can test the auth flow with the following commands in this order:

First, verify that the CLI errors out if you are not authenticated and the `--share` argument is passed:

```sh
cargo run --bin burnbench -- run --share --benches unary --backends ndarray
```

Then follow the authentication flow with:

```sh
cargo run --bin burnbench -- auth
```

At last rerun the benchmark with `--share` argument:

```sh
cargo run --bin burnbench -- run --share --benches unary --backends ndarray
```

Note that the actual upload is not yet implemented.

### DIFF
--- a/backend-comparison/src/burnbenchapp/auth.rs
+++ b/backend-comparison/src/burnbenchapp/auth.rs
@@ -1,5 +1,4 @@
 use reqwest;
-use std::error::Error;
 use std::io::Write;
 #[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
@@ -13,7 +12,7 @@ static GITHUB_API_VERSION_HEADER: &str = "X-GitHub-Api-Version";
 static GITHUB_API_VERSION: &str = "2022-11-28";
 
 /// Return the file path for the auth cache on disk
-pub fn get_auth_cache_file_path() -> PathBuf {
+pub(crate) fn get_auth_cache_file_path() -> PathBuf {
     let home_dir = dirs::home_dir().expect("an home directory should exist");
     let path_dir = home_dir.join(".cache").join("burn").join("burnbench");
     #[cfg(test)]
@@ -22,33 +21,20 @@ pub fn get_auth_cache_file_path() -> PathBuf {
     path.join("token.txt")
 }
 
-/// Return true if the token is still valid
-#[inline]
-pub(crate) fn is_token_valid(token: &str) -> bool {
-    get_username_from_token(token).is_ok()
-}
-
-/// Retrieve the user name from the access token
-fn get_username_from_token(token: &str) -> Result<String, Box<dyn Error>> {
+/// Returns true if the token is still valid
+pub(crate) fn verify_token(token: &str) -> bool {
     let client = reqwest::blocking::Client::new();
-    // User-Agent is important otherwise GitHub rejects the request with a 403
-    // See: https://github.com/seanmonstar/reqwest/issues/918#issuecomment-632906966
     let response = client
         .get("https://api.github.com/user")
         .header(reqwest::header::USER_AGENT, "burnbench")
         .header(reqwest::header::ACCEPT, "application/vnd.github+json")
         .header(reqwest::header::AUTHORIZATION, format!("Bearer {}", token))
         .header(GITHUB_API_VERSION_HEADER, GITHUB_API_VERSION)
-        .send()?;
-    let response = response.json::<serde_json::Value>()?;
-    let username = response["login"].as_str().map(|s| s.to_string());
-    // Return an error if the login field is not found
-    username.ok_or_else(|| {
-        From::from(std::io::Error::new(
-            std::io::ErrorKind::NotFound,
-            "Username not found in the response",
-        ))
-    })
+        .send();
+    match response {
+        Ok(resp) => resp.status().is_success(),
+        Err(_) => false,
+    }
 }
 
 /// Save token in Burn cache directory and adjust file permissions
--- a/backend-comparison/src/burnbenchapp/auth.rs
+++ b/backend-comparison/src/burnbenchapp/auth.rs
@@ -23,6 +23,7 @@ pub fn get_auth_cache_file_path() -> PathBuf {
 }
 
 /// Return true if the token is still valid
+#[inline]
 pub(crate) fn is_token_valid(token: &str) -> bool {
     get_username_from_token(token).is_ok()
 }
@@ -66,12 +67,12 @@ pub(crate) fn save_token(token: &str) {
 }
 
 /// Return the token saved in the cache file
+#[inline]
 pub(crate) fn get_token_from_cache() -> Option<String> {
     let path = get_auth_cache_file_path();
-    match fs::read_to_string(path) {
-        Ok(contents) => contents.lines().next().map(str::to_string),
-        _ => None,
-    }
+    fs::read_to_string(path)
+        .ok()
+        .and_then(|contents| contents.lines().next().map(str::to_string))
 }
 
 #[cfg(test)]
--- a/backend-comparison/src/burnbenchapp/auth.rs
+++ b/backend-comparison/src/burnbenchapp/auth.rs
@@ -8,9 +8,9 @@ use std::{
     path::{Path, PathBuf},
 };
 
-pub(crate) static CLIENT_ID: &'static str = "Iv1.84002254a02791f3";
-static GITHUB_API_VERSION_HEADER: &'static str = "X-GitHub-Api-Version";
-static GITHUB_API_VERSION: &'static str = "2022-11-28";
+pub(crate) static CLIENT_ID: &str = "Iv1.84002254a02791f3";
+static GITHUB_API_VERSION_HEADER: &str = "X-GitHub-Api-Version";
+static GITHUB_API_VERSION: &str = "2022-11-28";
 
 /// Return the file path for the auth cache on disk
 pub fn get_auth_cache_file_path() -> PathBuf {
@@ -68,7 +68,7 @@ pub(crate) fn save_token(token: &str) {
 /// Return the token saved in the cache file
 pub(crate) fn get_token_from_cache() -> Option<String> {
     let path = get_auth_cache_file_path();
-    match fs::read_to_string(&path) {
+    match fs::read_to_string(path) {
         Ok(contents) => contents.lines().next().map(str::to_string),
         _ => None,
     }
--- a/backend-comparison/src/burnbenchapp/auth.rs
+++ b/backend-comparison/src/burnbenchapp/auth.rs
@@ -1,0 +1,158 @@
+use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
+use std::{
+    fs::{self, File},
+    path::{Path, PathBuf},
+};
+
+pub(crate) static CLIENT_ID: &'static str = "Iv1.84002254a02791f3";
+
+/// Return the file path for the auth cache on disk
+pub fn get_auth_cache_file_path() -> PathBuf {
+    let home_dir = dirs::home_dir().expect("an home directory should exist");
+    let path_dir = home_dir.join(".cache").join("burn").join("burnbench");
+    #[cfg(test)]
+    let path_dir = path_dir.join("test");
+    let path = Path::new(&path_dir);
+    path.join("token.txt")
+}
+
+/// Save token in Burn cache directory and adjust file permissions
+pub(crate) fn save_token(token: &str) {
+    let path = get_auth_cache_file_path();
+    fs::create_dir_all(path.parent().expect("path should have a parent directory"))
+        .expect("directory should be created");
+    let mut file = File::create(&path).expect("file should be created");
+    write!(file, "{}", token).expect("token should be written to file");
+    // On unix systems we lower the permissions on the cache file to be readable
+    // just by the current user
+    #[cfg(unix)]
+    fs::set_permissions(&path, fs::Permissions::from_mode(0o600))
+        .expect("permissions should be set to 600");
+    println!("✅ Token saved at location: {}", path.to_str().unwrap());
+}
+
+/// Return the token saved in the cache file
+fn get_token_from_cache() -> Option<String> {
+    let path = get_auth_cache_file_path();
+    match fs::read_to_string(&path) {
+        Ok(contents) => contents.lines().next().map(str::to_string),
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+use serial_test::serial;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+
+    fn cleanup_test_environment() {
+        let path = get_auth_cache_file_path();
+        if path.exists() {
+            fs::remove_file(&path).expect("should be able to delete the token file");
+        }
+        let parent_dir = path
+            .parent()
+            .expect("token file should have a parent directory");
+        if parent_dir.exists() {
+            fs::remove_dir_all(parent_dir).expect("should be able to delete the cache directory");
+        }
+    }
+
+    #[test]
+    #[serial]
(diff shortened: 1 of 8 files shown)


antimora:
I followed the steps successfully, and here are the results:

```
     Running benches/unary.rs (/Users/dilshod/Projects/burn/target/release/deps/unary-5a7517d0271fa0ff)

        Timestamp: 1707511316062
        Git Hash: 202247d6e59890b97c2dc2f61fcc9b0665a3e85f
        Benchmarking - unary
―――――――― Result ―――――――――
  Samples     10
  Mean        197.687ms
  Variance    1.002µs
  Median      197.450ms
  Min         196.849ms
  Max         200.591ms
―――――――――――――――――――――――――

Cleanup completed. Benchmark run(s) finished.
Sharing results...
[backend-comparison]%
```

antimora:
Also confirming that the permissions are set correctly on my Mac:

```
[backend-comparison]% ll /Users/dilshod/.cache/burn/burnbench/token.txt
-rw-------  1 dilshod  staff  40 Feb  9 14:41 /Users/dilshod/.cache/burn/burnbench/token.txt
```

antimora (approved):
LGTM

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+pub(crate) static CLIENT_ID: &str = "Iv1.84002254a02791f3";
+static GITHUB_API_VERSION_HEADER: &str = "X-GitHub-Api-Version";
+static GITHUB_API_VERSION: &str = "2022-11-28";
+
+/// Return the file path for the auth cache on disk
+pub fn get_auth_cache_file_path() -> PathBuf {
+    let home_dir = dirs::home_dir().expect("an home directory should exist");
+    let path_dir = home_dir.join(".cache").join("burn").join("burnbench");
+    #[cfg(test)]
+    let path_dir = path_dir.join("test");
```
nathanielsimard, line 20: So convenient!

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+    #[test]
+    #[serial]
+    fn test_save_token_when_file_does_not_exist() {
+        cleanup_test_environment();
+        let token = "unique_test_token";
+        // Ensure the file does not exist
+        let path = get_auth_cache_file_path();
+        if path.exists() {
+            fs::remove_file(&path).unwrap();
+        }
```
nathanielsimard, line 107: Does the `cleanup_test_environement` do that?

backend-comparison/src/burnbenchapp/base.rs
```diff
@@ -81,45 +99,94 @@ pub(crate) enum BenchmarkValues {
...
+    println!("🌐 Please visit for following URL in your browser (CTRL+click if your terminal supports it):");
+    println!("\n    {}\n", flow.verification_uri.clone().unwrap());
+    let user_code = flow.user_code.clone().unwrap();
+    println!("👉 And enter code: {}", &user_code);
+    if let Ok(mut clipboard) = Clipboard::new() {
+        if clipboard.set_text(user_code).is_ok() {
+            println!("📋 Code has been successfully copied to clipboard.")
+        };
+    };
+    thread::sleep(FIVE_SECONDS);
```
nathanielsimard, line 130: Why is it necessary? Maybe add a comment here.

backend-comparison/src/burnbenchapp/base.rs
```diff
@@ -81,45 +99,94 @@ pub(crate) enum BenchmarkValues {
...
+    match flow.poll(20) {
+        Ok(creds) => {
+            save_token(&creds.token);
         }
+        Err(e) => eprint!("Authentication error: {}", e),
+    };
```
nathanielsimard, line 136: Can't this be done in a loop with a fixed amount of retry and an exponential sleep decay?

backend-comparison/src/burnbenchapp/base.rs
```diff
@@ -81,45 +99,94 @@ pub(crate) enum BenchmarkValues {
...
+        if !is_token_valid(&token.unwrap()) {
+            eprintln!("Your access token is no longer valid.");
+            eprintln!("Run the command 'burnbench auth' again to get a new token.");
+            return;
+        }
+    }
+    if run_args.backends.is_empty() || run_args.benches.is_empty() {
+        println!("No backends or benchmarks specified. Please select at least one backend and one benchmark.");
+        return;
+    }
```
nathanielsimard, line 170: I think we can set the mininum number of `backends` and `benches` with clap (`num_args(0..)` to `num_args(1..)`)

nathanielsimard (changes requested):

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+    let response = response.json::<serde_json::Value>()?;
+    let username = response["login"].as_str().map(|s| s.to_string());
+    // Return an error if the login field is not found
+    username.ok_or_else(|| {
+        From::from(std::io::Error::new(
+            std::io::ErrorKind::NotFound,
+            "Username not found in the response",
+        ))
+    })
+}
```
Luni-4, line 51: Can we avoid to allocate the `String` here and use lifetimes in order to retrieve the `bool`? 

`get_username_from_token` seems not used anywhere in the code

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+    fs::create_dir_all(path.parent().expect("path should have a parent directory"))
+        .expect("directory should be created");
+    let mut file = File::create(&path).expect("file should be created");
+    write!(file, "{}", token).expect("token should be written to file");
+    // On unix systems we lower the permissions on the cache file to be readable
+    // just by the current user
+    #[cfg(unix)]
+    fs::set_permissions(&path, fs::Permissions::from_mode(0o600))
+        .expect("permissions should be set to 600");
+    println!("✅ Token saved at location: {}", path.to_str().unwrap());
```
Luni-4, line 65: Can we use a `log` function here instead of `println`?

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// Return the token saved in the cache file
+pub(crate) fn get_token_from_cache() -> Option<String> {
+    let path = get_auth_cache_file_path();
+    match fs::read_to_string(path) {
+        Ok(contents) => contents.lines().next().map(str::to_string),
+        _ => None,
+    }
```
Luni-4, line 74: We can simplify this `match` with `ok().flatten()` or directly `and_then`

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+    // On unix systems we lower the permissions on the cache file to be readable
+    // just by the current user
+    #[cfg(unix)]
+    fs::set_permissions(&path, fs::Permissions::from_mode(0o600))
+        .expect("permissions should be set to 600");
+    println!("✅ Token saved at location: {}", path.to_str().unwrap());
+}
+
+/// Return the token saved in the cache file
+pub(crate) fn get_token_from_cache() -> Option<String> {
```
Luni-4, line 69: We can use the `#[inline]` attribute

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,191 @@
...
+    let home_dir = dirs::home_dir().expect("an home directory should exist");
+    let path_dir = home_dir.join(".cache").join("burn").join("burnbench");
+    #[cfg(test)]
+    let path_dir = path_dir.join("test");
+    let path = Path::new(&path_dir);
+    path.join("token.txt")
+}
+
+/// Return true if the token is still valid
+pub(crate) fn is_token_valid(token: &str) -> bool {
```
Luni-4, line 26: We can use the `#[inline]` attribute

backend-comparison/src/burnbenchapp/base.rs
```diff
@@ -81,45 +99,94 @@ pub(crate) enum BenchmarkValues {
...
+}
+
+/// Create an access token from GitHub Burnbench application and store it
+/// to be used with the user benchmark backend.
+fn command_auth() {
+    let mut flow = match DeviceFlow::start(CLIENT_ID, None) {
+        Ok(flow) => flow,
+        Err(e) => {
+            eprintln!("Error authenticating: {}", e);
```
Luni-4, line 117: We can use a `log_err` function

Luni-4 (changes requested):
Fantastic! Thanks a lot!

Two considerations (you can check it up if they are worth):
- [ ] Can we replace `println` and `eprintln` with `log` functions?
- [ ] Perhaps I'm wrong, but it seems there are some tests that could be deduplicated using intermediate functions. Is this wanted?

backend-comparison/src/burnbenchapp/auth.rs
nathanielsimard: I don't think using liketime here works, since the lifetime of the `str` is associated with the response in JSON from the request that will be cleaned up after the function, so String is all right.

nathanielsimard:
> * [ ]  Can we replace `println` and `eprintln` with `log` functions?

I think `println` is meant to display a command line message more than logging.

backend-comparison/src/burnbenchapp/base.rs
syl20bnr: No it is expected to be a retry at a regular interval. I added a link to the documentation in your previous comment for the FIVE_SECONDS.

backend-comparison/src/burnbenchapp/base.rs
syl20bnr: Good point !!

syl20bnr: Made the fix along with setting the --benches and --backends arguments as required. I also fixed the documentation for these arguments that was wrong. 🙂

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: The test is about saving the token save when there is no cache file yet so I ensure that there is none at the beginning. I prefer to not assume that the state has been cleaned up correctly.

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: > get_username_from_token seems not used anywhere in the code

Well it is used in `is_token_valid` 😄
We could make it inline in `is_token_valid` but we would loose some meaning.

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: Fixed with `and_then`, thank you for showing me more idiomatic way of doing things 👍

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: Done

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: Done

syl20bnr:
@Luni-4 Not sure about the addition of a log mechanism for now given that we already write the benchmarks to disk in the cache directory. Those prints are just part of the UI in a sense. But maybe it can be useful to add a logger for those later, feel free to open an issue about it.

backend-comparison/src/burnbenchapp/auth.rs
Luni-4: @syl20bnr 

Yep, your analysis is correct. You can solve this returning directly the `bool`  in `get_username_from_token` or insert `get_username_from_token` into `is_token_valid`.  You do not catch the `Result` at all in that way because you are using `ok` so the error is never unwrapped.

Luni-4:
@nathanielsimard  @syl20bnr 

My idea of logging in a library (in our case, a crate) consists of discriminating among the different events that might happen when an API is called. A `println!` just prints a message on shell, but it does not provide any additional information about the event. Through `log` APIs,  we can express our attitudes towards that message communicating a precise behaviour to a user. A con is that we have to add another dependency, but it should be a lightweight one. 

Anyway, I agree maintaining a `println` and `eprintln`, they are just more imprecise in my opinion, not wrong, so we can add logging functions later. Yep, I can open an issue.

syl20bnr:
> Perhaps I'm wrong, but it seems there are some tests that could be deduplicated using intermediate functions. Is this wanted?

Each test tests only one thing and each case is different, we cannot really do much more than the cleanup function that is already made to reduce redundancy. Maybe we could be more fancy with the setup and teardown but that does not seem to be easy in Rust.

antimora:
@syl20bnr i have used rstest in burn-import. It is pretty cool and easy to use. Check it out if you are interested: https://docs.rs/rstest/latest/rstest/

syl20bnr:
@antimora unfortunately `rstest` does not provide a mechanism for setup and teardown. I believe that any attempt in doing it by ourselves will increase the complexity of the tests for little gain.

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: Got it, thank you. And I see your point. I am not sure yet that I don't want the username at the CLI tool level so I prefer to keep it now as this even if it is not really used.

Luni-4: A possible tradeoff could be to convert `get_username_from_token` in a closure.  If I recall correctly the compiler treats them differently from a function when it produces the code, without bringing into play the calling stack.

syl20bnr: I replaced this solution with `verify_token` that just checks for the status code of the request.

backend-comparison/src/burnbenchapp/auth.rs
```diff
@@ -0,0 +1,178 @@
...
+        .get("https://api.github.com/user")
+        .header(reqwest::header::USER_AGENT, "burnbench")
+        .header(reqwest::header::ACCEPT, "application/vnd.github+json")
+        .header(reqwest::header::AUTHORIZATION, format!("Bearer {}", token))
+        .header(GITHUB_API_VERSION_HEADER, GITHUB_API_VERSION)
+        .send();
+    match response {
+        Ok(resp) => resp.status().is_success(),
+        Err(_) => false,
+    }
```
Luni-4, line 37: We can simplify this with `map_or` function

Luni-4 (changes requested):
Thanks a lot for your changes! The code has been improved a lot! 

Just the last change for me

backend-comparison/src/burnbenchapp/auth.rs
syl20bnr: Got it, leveling up my Rust fu ! 😁

nathanielsimard (approved):

Luni-4 (approved):

PR #1287, opened 2024-02-10
Fix/conv2d/fan

louisfd:
Fix a bug introduced in #1138 where arithmetic could fail on usize type


laggui (approved):
LGTM! Thanks for adding the test 😎

nathanielsimard (approved):

PR #1289, opened 2024-02-10
Trying add a simple Image Accuracy metric for tests

Apogeum12:
### Checklist

### Changes

I want to be able to somehow compare the accuracy of the generated graphics with the expected ones using a threshold, e.g. 0.001.
In the forward_classification function for the test network, I calculated the error in this simple way, i.e. all pixels from all channels in one dimension.
```rs
pub fn forward_classification(
         &self,
         batch: MNISTBatch<B>,
     ) -> RegressionOutput<B> {
         let output = self.forward(batch.images.clone());
        
         let [batch_img_out, channel_img_out, height_img_out, width_img_out] = output.dims();
         let logits = output.reshape([batch_img_out, channel_img_out * height_img_out * width_img_out]);
         let targets = batch.targets.reshape([batch_img_out, channel_img_out * height_img_out*width_img_out]);
         let loss = MSELoss::new().forward(logits.clone(), targets.clone(), burn::nn::loss::Reduction::Mean);
         // let loss = Tensor::log(Tensor::ones_like(&pre_loss).div(pre_loss.sqrt())).mul_scalar(20.0); // PSNR <-- impl
         RegressionOutput::new(loss, logits, targets)
     }
```

Despite passing the tests in burn-train, I tested it on the network with the output image and I know that the results are bad, i.e. ImageAccuracy is e.g. 72490, I don't know if the counting function itself is bad (it's definitely not the best) or if it wrong updates the state. Overall, at this stage I don't really know what to do next, so I need help here. I wrote as much as I could at this stage of understanding the burn library and RUST itself. I will be grateful for help and tips, maybe it will also be useful for someone else, that's why I decided to make a pull request.

### Testing

`cargo test` into burn-reain folder:
```sh
running 16 tests
test checkpoint::strategy::composed::tests::should_delete_when_both_deletes ... ok
test metric::acc::tests::test_accuracy_with_padding ... ok
test metric::acc::tests::test_accuracy_without_padding ... ok
test metric::acc::tests::test_image_accuracy ... ok
test checkpoint::strategy::lastn::tests::should_always_delete_lastn_epoch_if_higher_than_one ... ok
test checkpoint::strategy::metric::tests::always_keep_the_best_epoch ... ok
test renderer::tui::progress::tests::calculate_progress_for_eta ... ok
test renderer::tui::progress::tests::calculate_progress_for_eta_with_warmup ... ok
test renderer::tui::progress::tests::test_format_eta ... ok
test learner::early_stopping::tests::early_stop_when_stays_equal ... ok
test renderer::tui::recent_history::tests::test_push_update_bounds_max_y ... ok
test learner::early_stopping::tests::early_stop_when_no_improvement_since_two_epochs ... ok
test renderer::tui::recent_history::tests::test_push_update_bounds_min_y ... ok
test renderer::tui::full_history::tests::test_points ... ok
test learner::early_stopping::tests::never_early_stop_while_it_is_improving ... ok
test metric::store::aggregate::tests::should_find_epoch ... ok

test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests burn-train

running 0 tests

```

### DIFF
--- a/burn-train/src/metric/acc.rs
+++ b/burn-train/src/metric/acc.rs
@@ -119,8 +119,7 @@ impl<B: Backend> Metric for ImageAccuracyMetric<B> {
         let outputs = input.outputs.clone().to_device(&input.outputs.device());
         let targets = input.targets.clone().to_device(&input.outputs.device());
 
-        let difference = outputs.sub(targets).abs(); 
-        
+        let difference = outputs.sub(targets).abs();
         // --- Matches For True ---
         let mask_correct = difference.clone().lower_equal_elem(self.threshold);
         let matches_correct = difference.clone().int().mask_fill(mask_correct, 1);
@@ -129,7 +128,8 @@ impl<B: Backend> Metric for ImageAccuracyMetric<B> {
         let matches_wrong = difference.int().mask_fill(mask_wrong, 0);
 
         let correct = matches_correct.add(matches_wrong);
-        let accuracy = correct.sum().into_scalar().elem::<f64>() / (num_pixels*batch_size ) as f64;
+
+        let accuracy = correct.sum().into_scalar().elem::<f64>() / (num_pixels * batch_size) as f64;
 
         self.state.update(
             100.0 * accuracy,
@@ -197,7 +197,7 @@ mod tests {
         let _entry = metric.update(&input, &MetricMetadata::fake());
         assert_eq!(50.0, metric.value());
     }
-    
+
     //? Need tests for ImageAccuracyMetric
     #[test]
     fn test_image_accuracy() {
@@ -209,9 +209,9 @@ mod tests {
                     [0.1, 0.2, 0.0], // 1 on 3 ( |outputs - targets| <= threshold )
                     [0.1, 0.3, 0.1], // 2 on 3
                     [0.5, 0.2, 0.0], // 1 on 3
-                    [0.0, 0.0, 0.0], // 2 on 3 >> 6 correct on 12 total 
+                    [0.0, 0.0, 0.0], // 2 on 3 >> 6 correct on 12 total
                 ],
-                &device
+                &device,
             ),
             Tensor::from_data(
                 [
@@ -219,7 +219,8 @@ mod tests {
                     [0.1, 0.0, 0.1],
                     [0.4, 0.2, 0.6],
                     [0.0, 0.1, 0.0],
-                ], &device
+                ],
+                &device,
             ),
         );
         let _entry = metric.update(&input, &MetricMetadata::fake());
--- a/burn-train/src/metric/acc.rs
+++ b/burn-train/src/metric/acc.rs
@@ -85,6 +85,69 @@ impl<B: Backend> Numeric for AccuracyMetric<B> {
     }
 }
 
+/// The Image accuracy metric.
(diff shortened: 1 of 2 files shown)


antimora:
Comparing images on a per-pixel basis is challenging. Here is my recommendation for simplifying the process:

1. **Convert Color Images to Grayscale:** If the image is a color image (3-channel), convert it to a 1-channel image by taking the average of the 3 channels. This effectively converts the image to grayscale.

2. **Downsample the Image:** Use an average pooling (2D) operation with a kernel size of 3x3 and a stride of 3x3. This will slightly downsample the image.

3. **Measure Similarity with Cosine Distance:** Use cosine distance to compare the images. A cosine distance of 0 indicates that the images are orthogonal (not similar), while a value of 1 indicates a perfect match. Any value in between represents varying degrees of similarity.

Below is a code snippet I've used in the past for calculating cosine distance. Please note that this code may need adjustments to be compatible with the current version of the library you are using, as I have not updated it recently.

```rust
fn cosine_distance(m1: &Tensor<B, 4>, m2: &Tensor<B, 4>) -> f32 {
    // cosine = np.dot(output_librosa,output_rust)/(norm(output_librosa)*norm(output_rust))

    let v1 = m1.clone().reshape([1, m1.shape().num_elements()]); // Converts 1 D
    let v2 = m2.clone().reshape([1, m2.shape().num_elements()]); // Converts 1 D

    let dot_product = v1.clone().matmul(v2.clone().transpose()).to_data().value[0];

    let norm1 = v1.powf(2.0).sum().sqrt().to_data().value[0];
    let norm2 = v2.powf(2.0).sum().sqrt().to_data().value[0];

    dot_product / (norm1 * norm2)
}
 ```

File: burn-train/src/metric/acc.rs [hunk 1]
```diff
@@ -85,6 +85,69 @@ impl<B: Backend> Numeric for AccuracyMetric<B> {
     }
 }
 
+/// The Image accuracy metric.
+#[derive(Default)]
+pub struct ImageAccuracyMetric<B: Backend> {
+    state: NumericMetricState,
+    threshold: f64,
+    _b: B,
```
antimora, line 93: I recommend you use [`PhantomData`](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#unused-type-parameters) to wrap `B`.

Apogeum12:
@antimora 
Thanks for the suggestions and help. To calculate network error and print it in the training interface, I don't have an issue. I work on simple Residual Conv Blocks. My issue is the accuracy values in the training interface, because now it prints complete fake data (without sense).
In this simple implementation, ImageAccuracy. Yes, cosine Distance will be better than threshold, as will SSIM. But for now, it will be nice to have anything that gives real results, even a threshold check.

ImageAccuracy

nathanielsimard:
@Apogeum12 the accuracy metric is used for classification problem, if you model the pixel with floats and what to measure how the float are close to each other, the accuracy metric isn't it!

Apogeum12:
@nathanielsimard I was hoping it'd work in some way to compute accuracy. So in this situation, I'll base it only on loss compute after implementation SSIM. I think that this pull request can be removed. Thanks for clearing situation.

PR #1296, opened 2024-02-12
Parser rewrite

skewballfox:
### Checklist

### Related Issues/PRs
- this might be a step towards supporting #778, if I rework the builder to use a vector of functions for per-node processing and post processing like I mentioned on discord.

If there are any others I should link, please let me know. 

### Changes

Rewrote the parser to handle everything but node removal in a single pass.  Node inputs and arguments are updated on node construction, and  `node.passed` is used to keep track of whether an IO argument is used. This parser only works if the nodes are topologically sorted, so I think instead of checking (which is $\Theta(N^2))$, we should call the function in the event of a panic, testing the model validity only after something else failed. 

There is still some housecleaning to be done, I need to either make it an actual builder which returns an `OnnxGraph`, or change the functions to take the necessary state as arguments. oh, and there is enough dead code here to fill a mortuary. 

There is  still quite a bit that could be done to make this faster or use less memory, such as altering `BurnGraph` to use `OnnxGraphIO`, but that's a PR for another time.

### Testing

I ran all Onnx Test to ensure that building the existing models still works the same.

If you would like me to add more test somewhere, let me know. 

I noticed that certain situations with initializers isn't tested in the onnx-tests, but are captured by running the examples. For example there were a couple of problems that would arise when an node input was both in the initializers and graph inputs that only showed up on the `squeezenet1.onnx` example. 

It might be a good idea to either make building those examples part of the onnx test  or create a few small models for testing that situation


antimora (changes requested):
It's my first pass and I have minor comments. I dive in more to understand your changes better because there are a lot of it.

antimora (approved):
LGTM. I am relying on tests to verify the correctness and @skewballfox's ability for large refactors. Let push it through.

@skewballfox , let me know if it's ready to be merged.

PR #1297, opened 2024-02-12
[CI] Set minimum supported version of Rust to 1.75

syl20bnr:


Luni-4 (approved):

antimora (changes requested):
We need to update the front documentation as well.

syl20bnr:
@antimora Good catch! I updated it.

antimora (approved):

Issue #1299, opened 2024-02-12
Chore: Use Candle's published crate version instead of git source code


antimora:
There are several unpublished Candle core features needed to fix issues with PyTorchRecorder (top level key, latest pt file format, fortran contiguous layout). For now, we would like to use a fix git version and once Candle publishes, we switch the version. This issue to track before our release.

nathanielsimard:
@antimora if we use a git version, it's impossible to deploy burn to crates.io, that's why we are sticking with a fixed version.

antimora:
Spoke offline and decided to wait for a new Candle release. Closing it.

PR #1300, opened 2024-02-12
Enhance PyTorchRecorder to pass top level key to extract state_dict

antimora:
This is needed for Whisper weight pt files.

### Checklist

### Related Issues/PRs

1. https://github.com/huggingface/candle/pull/1671
2. https://github.com/tracel-ai/burn/issues/1299

### Changes

1. Enhanced PyTorchRecorder to accept top level key to tract state_dict 
2. Updated the book
3. Added new test
4. Using Candle core head version at a specific revision. Filed #1299 to revert once we are about to release.

### Testing

Added a new unit test.


burn-import/src/pytorch/recorder.rs
```diff
@@ -115,6 +124,17 @@ impl LoadArgs {
         self.key_remap.push((regex, replacement.into()));
         self
     }
+
+    /// Set top level key to load state_dict from the file.
+    /// Sometimes the state_dict is nested under a top level key in a dict.
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The top level key to load state_dict from the file.
```
laggui, line 133: For consistency: use *top-level* and not "top level"

burn-import/src/pytorch/recorder.rs
```diff
@@ -115,6 +124,17 @@ impl LoadArgs {
...
+    pub fn with_top_level_key(mut self, key: &str) -> Self {
```
laggui, line 134: Reading this I was wondering if something like `with_nested_dict(key)` or `from_nested_dict(key)` would be more clear? The name is open to discussion, just giving my 2 cents :)

laggui (changes requested):
Thank you for the fix! I'm sure this will be really helpful 🙂 

Overall LGTM! Just some minor typos and an open discussion regarding the name of the new method 👍

nathanielsimard (commented):
Agree with @laggui comments, otherwise LGTM.

burn-import/src/pytorch/recorder.rs
antimora: `with_` prefix is a builder naming convention and I have used top_level_key through already. Not my favorite but consistent.

laggui: I don't have any better suggestions anyway, so I'll have to agree with the choice 😄

laggui (approved):

antimora:
We are waiting till a new Candle release. The PR implementation uses Git's version from Candle's main.

nathanielsimard (approved):

PR #1301, opened 2024-02-13
fix(book): add missing second parameter to CrosEntropyLoss constructor

apertureless:
CrossEntropyLoss::new() expects two parameters, the pad_index and the device

### Checklist

### Related Issues/PRs

None

### Changes

I think the code example in the book is a bit outdated? 
`CrossEntropyLoss::new()` is expecting two parameters, the `pad_index` and `device`

### DIFF
--- a/burn-book/src/basic-workflow/training.md
+++ b/burn-book/src/basic-workflow/training.md
@@ -22,7 +22,7 @@ impl<B: Backend> Model<B> {
         targets: Tensor<B, 1, Int>,
     ) -> ClassificationOutput<B> {
         let output = self.forward(images);
-        let loss = CrossEntropyLoss::new(None, &output.device().forward(output.clone(), targets.clone());
+        let loss = CrossEntropyLoss::new(None, &output.device()).forward(output.clone(), targets.clone());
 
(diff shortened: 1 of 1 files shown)


burn-book/src/basic-workflow/training.md
```diff
@@ -22,7 +22,7 @@ impl<B: Backend> Model<B> {
         targets: Tensor<B, 1, Int>,
     ) -> ClassificationOutput<B> {
         let output = self.forward(images);
-        let loss = CrossEntropyLoss::new(None).forward(output.clone(), targets.clone());
+        let loss = CrossEntropyLoss::new(None, &output.device().forward(output.clone(), targets.clone());
```
nathanielsimard, line 25: I think you are missing a closing ) before the `forward`:

`CrossEntropyLoss::new(None, &output.device()).forward(output.clone(), targets.clone());`

nathanielsimard (changes requested):

burn-book/src/basic-workflow/training.md
apertureless: You're right.  Should be fixed now.

nathanielsimard (approved):

PR #1302, opened 2024-02-13
fix(book): add missing device parameter to mode.init()

apertureless:
Fix outdated code in book. 
Add missing device parameter to model.init() function.

### Checklist

### Changes

Fix outdated code in book. 
Add missing device parameter to model.init() function.


nathanielsimard (approved):

nathanielsimard:
@apertureless thanks for fixing some mistakes in the book 🙏

PR #1311, opened 2024-02-14
Rename GELU to Gelu

antimora:
This is to follow Rust naming convention. It has been inconsistent how other structs are named.

### Checklist

### Changes

1. Rename GELU to Gelu in code
2. Change in docs

### Testing

Ran all checks


README.md
```diff
@@ -44,7 +44,7 @@ Using Burn means having your models optimized on any backend. When possible, we
 automatically and dynamically create custom kernels that minimize data relocation between different
 memory spaces, extremely useful when moving memory is the bottleneck.
 
-As an example, you could write your own GELU activation function with the high level tensor api (see
+As an example, you could write your own Gelu activation function with the high level tensor api (see
```
louisfd, line 47: I would certainly not change it here because it's not the Rust struct but the deep learning concept

louisfd (changes requested):
I agree there is inconsistency, for instance we have `Lstm` and `Gru` instead of LSTM and GRU, but we also have `ReLU`, `MSELoss`, ... I think if we want to make that move we should rename them too.

nathanielsimard (approved):

antimora:
@louisfd, it's a good suggestion. I will rename others in a new PR.

louisfd (approved):
Perfect!

PR #1320, opened 2024-02-17
[CI] Fix codecov and update to weekly the `.dependabot` file for `cargo`

### LINKED ISSUE #1294: Codecov actions broken since last update
The codecov action has been updated to version 4 with this commit: https://github.com/tracel-ai/burn/commit/5786ff4f47eb648d8e7a2acb6909d791abfc0039

This new version requires a token to be defined, see https://github.com/codecov/example-rust/blob/main/.github/workflows/rust.yml#L35-L40

Luni-4:
### Checklist

### Changes

This PR:
- fixes #1294 udpating `codecov`
- adds all `burn` dependencies to `.dependabot` in order to have a single weekly PR


syl20bnr (approved):
Thank you !
Do we need to create the token or is it already created ?

Luni-4:
I do not know @nathanielsimard can know this info

nathanielsimard:
The token is set in the github settings

.github/dependabot.yml
```diff
@@ -13,18 +13,87 @@ updates:
...
+        - "dashmap"
+        - "derive-new"
+        - "dirs"
+        - "fake"
+        - "flate2"
+        - "float-cmp"
         - "futures-intrusive"
+        - "getrandom"
+        - "github-device-flow"
+        - "gix-tempfile"
```
nathanielsimard, line 38: I would very much like to not duplicate out dependencies outside of `Cargo.toml`, otherwise this list isn't going to be comprehensive and will for sure become outdated quite soon.

Luni-4: Yep, I was thinking the same, but let's list the possible actions we can have:

- If we remove a dep then it is a nop for `.dependabot` since it cannot find that dep anymore. The dependency name can stay there without affecting anything.
- If we change or add a new dependency to `Cargo.toml`, an independent PR is created for that dependency because it is not part of the group. We can later decide to add that dependency to the `all` group, thus removing the independent PR.

We can also simplify this process adding a `xtask` task which updates `.dependabot` file using `Cargo.toml` dependencies and replace the current "aggregation PR" action with this new action which runs occasionally

syl20bnr: Can we drop the groups and use the combined PR action ? Seems to be less hassle than maintaining a list of dependencies.

syl20bnr: Agree, that would be perfect. We can scheduled dependabot just before the combine PR.

Luni-4: Done

syl20bnr:
Thank you @nathanielsimard
Shipping the fix!

PR #1328, opened 2024-02-19
Feature Addition: PRelu Module

Arjun31415:
### Checklist

### Related Issues/PRs

### Changes

### Testing

- [x] Tested on 2D data 
- [x] tested for positive, negative and 0 as weight param
- [x] tested forward pass


Arjun31415:
Adding docs to the book is pending.

Arjun31415:
I believe I'm doing the reshape and broadcast wrongly also. Need to check that too

antimora (changes requested):
Just a quick note. Can we also update the burn-books in these sections with your new activation function and module?

1. https://burn.dev/book/building-blocks/tensor.html#activation-functions
2. https://burn.dev/book/building-blocks/module.html#built-in-modules

Arjun31415:
Sure I will add the book section If my implementation is correct

Arjun31415:
I think I resolved all the merge conflicts now

crates/burn-core/src/nn/prelu.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::module::Module;
+use crate::module::Param;
+use crate::nn::Initializer;
+use crate::tensor::backend::Backend;
+use crate::tensor::Tensor;
+/// Parametric Relu layer.
+#[derive(Module, Debug)]
+pub struct PRelu<B: Backend> {
+    alpha: Param<Tensor<B, 1>>,
+    num_parameters: usize,
```
nathanielsimard, line 12: I don't think we need the `num_parameters` field. Also I think we should put the `alpha` parameter public and document it, I think it's a better API to make the parameters of module public.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -13,6 +13,32 @@ pub fn gelu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
...
+    let weight = if alpha.dims()[0] == 1 {
+        alpha.reshape([1; D])
+    } else if D >= 2 {
+        let channels = tensor.dims()[1];
+        let num_weights = alpha.dims()[0];
+        assert_eq!(
+            num_weights, channels,
+            "Number of channels and weights must be equal. Got no. of channels: {}, no. of weights: {}",
+            channels, num_weights
+        );
```
nathanielsimard, line 31: I would use the TensorCheck API to do that, so that we can have a constant formatting for tensor errors. 
You can look at how it is done here: https://github.com/tracel-ai/burn/blob/4427768570503fcb3c437a2be32af640bb34a6fc/crates/burn-tensor/src/tensor/api/numeric.rs#L19 and here: https://github.com/tracel-ai/burn/blob/4427768570503fcb3c437a2be32af640bb34a6fc/crates/burn-tensor/src/tensor/api/check.rs#L52

But from what I understand, you need the tensor to have a dimension with the same number of items as the alpha tensor, so:

```
tensor: [batch_size, channel, dim]
alpha: [channel],
```

However, in your code you assume that the channel is always the second dimension, so:

```
tensor = [batch_size, channel, ..., dim];
```

I think I would add that documentation on the PRelU module, just specifying the shape expectations.

Also, just adding a comment that you are reshaping the alpha tensor to `[1, channel, ...]` for broadcasting can help.

crates/burn-tensor/src/tensor/ops/activation.rs
```diff
@@ -58,6 +58,36 @@ pub trait ActivationOps<B: Backend> {
...
+    ///
+    /// The gradient.
+    fn prelu_backward<const D: usize>(
+        output: FloatTensor<B, D>,
+        grad: FloatTensor<B, D>,
+    ) -> FloatTensor<B, D> {
+        // f'(y) = 0 if y >0 or y if y<=0
+        let mask = B::float_greater_elem(output.clone(), 0.elem());
+        B::float_mask_fill(grad, mask, 0.elem())
+    }
```
nathanielsimard, line 90: I don't think this is correct, you would need to have the gradients for `min(0, x)` to be `alpha` when `X < 0` and `1` when `X >= 0`. With the chain rule, if would be selecting the right gradients depending on the mask and scaling the ones < 0 with alpha.

BUT I don't think this is necessary in this PR, so I would remove the `prelu_backward` function and rely on the autodiff system to calculate it for us. With `burn-fusion` it's going to be very fast and we may update this for other backends if it's to slow for them.

crates/burn-tensor/src/tests/activation/prelu.rs
```diff
@@ -0,0 +1,45 @@
...
+        ];
+        let tensor = TestTensor::from(data);
+
+        let data_actual =
+            activation::prelu(tensor, TestTensor::from([0.5, 0.25, 0.0, -0.8, -0.4])).into_data();
+        let data_expected = Data::from([
+            [-0.5500, 0.0000, 1.2000, 0.2500, 2.1600],
+            [-2.2835, 0.5600, -0.0000, 99.9000, -0.0000],
+        ]);
+        data_expected.assert_approx_eq(&data_actual, 9);
```
nathanielsimard, line 43: I would split this function into multiple test cases.

nathanielsimard (changes requested):
Good job, that's a nice addition, I have some comments, but I don't think they are going to be hard to resolve.

crates/burn-tensor/src/tensor/ops/activation.rs
Arjun31415: Isn't alpha the learnable parameter here. You have to differentiate w.r.t alpha if im not wrong. 

However I'll remove the backward function as requested

crates/burn-core/src/nn/prelu.rs
Arjun31415: Why is the `num_parameters` field not required. The number of weight can be either 1 (same weight across all channels) or `num_channel` (different weight for each channel which needs to be learnt)

Arjun31415: Referring to the docs from pytorch [here](https://pytorch.org/docs/stable/generated/torch.nn.PReLU.html)
and Candle implementation [here](https://github.com/huggingface/candle/blob/45d5322d62d59a2c9be5fe8c642d0fa56fbb73b1/candle-nn/src/activation.rs#L46C1-L49C2)

crates/burn-tensor/src/tensor/activation/base.rs
Arjun31415: I don't see which api function can be used because, there is no function which has a check like the one I'm using. I am checking the number of channels with the shape of the weight to be learnt `alpha`

Arjun31415:
I don't understand how the coverage is decreasing for test function 
https://app.codecov.io/gh/tracel-ai/burn/pull/1328/blob/crates/burn-tensor/src/tests/activation/prelu.rs

antimora:
> I don't understand how the coverage is decreasing for test function https://app.codecov.io/gh/tracel-ai/burn/pull/1328/blob/crates/burn-tensor/src/tests/activation/prelu.rs

@Arjun31415 , Missing `#[test]` above the function?

Arjun31415:
> > I don't understand how the coverage is decreasing for test function https://app.codecov.io/gh/tracel-ai/burn/pull/1328/blob/crates/burn-tensor/src/tests/activation/prelu.rs
> 
> @Arjun31415 , Missing `#[test]` above the function?

yeah that was silly of me

Arjun31415:
> Just a quick note. Can we also update the burn-books in these sections with your new activation function and module?
> 
>     1. https://burn.dev/book/building-blocks/tensor.html#activation-functions
> 
>     2. https://burn.dev/book/building-blocks/module.html#built-in-modules

Added docs

crates/burn-core/src/nn/prelu.rs
nathanielsimard: It's never read, the information is inside the alpha shape, so no need to have another field.

crates/burn-tensor/src/tensor/ops/activation.rs
nathanielsimard: Ho true, but yeah we can remove the function anyways.

crates/burn-core/src/nn/prelu.rs
Arjun31415: ok

antimora (approved):
LGTM from my end. Just take care of Nathaniel's comments.

Arjun31415:
@nathanielsimard  please check. I have made the changes as requested

nathanielsimard (approved):

PR #1335, opened 2024-02-20
feat: add prelude module for convenience

bioinformatist:
## Checklist

- [x] Confirmed that run-checks all script has been executed.

### DIFF
--- a/crates/burn-core/src/lib.rs
+++ b/crates/burn-core/src/lib.rs
@@ -70,10 +70,7 @@ pub mod prelude {
     pub use crate::{
         config::Config,
         module::Module,
-        nn::{loss::*, *},
-        tensor::{
-            backend::{AutodiffBackend, Backend},
-            Data, Tensor,
-        },
+        nn,
+        tensor::{backend::Backend, Data, Device, ElementConversion, Tensor},
     };
 }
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/lib.rs
```diff
@@ -62,3 +62,18 @@ pub type TestAutodiffBackend = burn_autodiff::Autodiff<TestBackend>;
...
+    //! burn::prelude::*` to your code to quickly get started with
+    //! Burn.
+    pub use crate::{
+        config::Config,
+        module::Module,
+        nn::{loss::*, *},
+        tensor::{
+            backend::{AutodiffBackend, Backend},
+            Data, Tensor,
+        },
```
nathanielsimard, line 77: I think this would be better:

```rust
pub use crate::{
    config::Config,
    module::Module,
    nn,
    tensor::{
        backend::Backend,
        Data, Tensor, ElementConversion, Device
    },
};
```

I wouldn't include any autodiff-related items for now, and I believe the `nn` module is sufficient.

bioinformatist: Yeah, I agree with it! Users should also take into account the usage of autodiff-related items as the `train` feature can be enabled when necessary now.

Issue #1263, opened 2024-02-05
Refactor: move all crates to `crates` folder


nathanielsimard:
Move all crates to the `crates` folder.

- burn-autodiff
- burn-candle
- burn-common
- burn-compute
- burn-core
- burn-dataset
- burn-derive
- burn-fusion
- burn-import
- burn-ndarray
- burn-no-std-tests
- burn-tch
- burn-tensor
- burn-traing
- burn-tensor-testgen
- burn-wgpu
- burn

antimora:
Well, we are doing it. Can we remove `burn-` prefix as well? We can leave the names with `burn-` as well. It's often a pain to keep typing `burn-` to cd.

syl20bnr:
I understand that's a bit tedious but I believe that prefix is important and it keep the folder name  consistent with the crate name.

Alternative proposition: we name the folder `burn-crates` and then we leave the prefix in the crate folders.

nathanielsimard:
I prefer to keep the structure `crates/burn-{name}` since this is their crate name after all and it would be confusing to have something else.

antimora:
@nathanielsimard, I ask you to reconsider keeping `burn-` prefix. It adds zero value and I don't think there would be a confusion. Navigating and reading will be much easier. 

Take this project for example: https://github.com/BurntSushi/ripgrep/tree/master/crates 

Here is the root of the project: https://github.com/BurntSushi/ripgrep/tree/master

I think the examples should be remain at the top root folder and should not be under crates.

I would like opinion of others here: @louisfd @syl20bnr @ashdtu @laggui

syl20bnr:
Hmmm we need to discuss about this.

If we remove the prefix, I suggest moving all the crates to the `crates` folder, and they should all have the `burn-` prefix in `Cargo.toml` for consistency. The only crate impacted by this convention is `backend-comparison`, which could be renamed to `benches`, resulting in `burn-benches` in its Cargo file.

We would also move the books to `books` folder, so `burn-book` does not appear at the root level.

laggui:
> @nathanielsimard, I ask you to reconsider keeping `burn-` prefix. It adds zero value and I don't think there would be a confusion. Navigating and reading will be much easier.
> 
> Take this project for example: https://github.com/BurntSushi/ripgrep/tree/master/crates
> 
> Here is the root of the project: https://github.com/BurntSushi/ripgrep/tree/master
> 
> I think the examples should be remain at the top root folder and should not be under crates.
> 
[...]

PR #1336, opened 2024-02-20
[refactor] Move burn crates to their own crates directory

syl20bnr:
### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/1263

### Changes

Move all the burn crates to their dedicated directory called `crates`.
We chose to keep the crate folder names to be the same as the actual crate name, see discussed in https://github.com/tracel-ai/burn/issues/1263

I'll update the links in the contributor book once this is merged.

### Testing

`run-checks` and making sure our CI passes.


Luni-4:
Fantastic! Perhaps we can add

```toml
members = [
  "crates/*",
]
```

to the main `Cargo.toml`.

I was wondering if it would be clearer to move  `examples` crates, since they are independent, into a new `burn` repository such as the `models` repository.

syl20bnr:
I updated the Cargo file with `crates/*`. 👍

We have no mechanism to publish examples alongside versions and they are using local crates from the repo. So for now it is better to keep them in the monorepo.

antimora:
> Fantastic! Perhaps we can add
> 
> ```toml
> members = [
>   "crates/*",
> ]
> ```
> 
> to the main `Cargo.toml` instead of each single crate
> 
> I was wondering if it would be clearer to move `examples` crates, since they are independent, into a new `burn` repository such as the `models` repository.

I would add we should keep in the same repository. It would be difficult to synchronize the versions

Luni-4:
To synchronize the versions we just have to change from `path` to `git` since they do not have to be published it should not be problematic

Luni-4 (approved):

Issue #1341, opened 2024-02-21
Add support for Any,All operations on Tensor


ashdtu:
### Feature description

Implement any(), any_dim(), all(), all_dim() Tensor operations for all types(Float, Int, Bool) similar to torch.any(input, dim), torch.all(input, dim). 

### Feature motivation
Useful to support these operations for performing boolean checks(logical_or, logical_and) and masking operations with tensors. 
In PyTorch these operations are supported for all tensor type(float, int, bool),

ashdtu:
Addressed in #1342 . Closing!

PR #1342, opened 2024-02-21
Add support for Any, All operations to Tensor

ashdtu:
### Checklist

### Related Issues/PRs

#1341 

### Changes

* Add support for any() (logical_or), all() (logical_and) operations on Tensor struct for all types (Float, Int, Bool). 
* For numeric tensor type(Float, Int), the input tensors are casted to Bool by checking against non-zero values. 
* The implementation supports working with these operations along a dim using the any_dim(), all_dim() methods.  
*  Burn book updated with these Tensor ops. 

### Testing

Test cases written for the implemented methods (any(), any_dim(), all(), all_dim()) with all tensor types(float, int, bool) in burn-tensor/src/tests/ops/any.rs or all.rs.


ashdtu:
@louisfd The bug in WGPU about the boolean casting to float/int is no longer a blocker in the updated code. I noticed that I can get rid of that by just changing the logic a bit and adding another bool_not op which converts the weird large values of 1065353216 for (1.0 f / 1 ) to exact zeros. There's no overflow now since this op happens before the final sum, so it's all ok. 

TL;DR There's no blocker for this PR and it works for all backends.

crates/burn-tensor/src/tensor/ops/bool_tensor.rs
```diff
@@ -302,4 +302,71 @@ pub trait BoolTensorOps<B: Backend> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `tensor` - The tensor to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor with a single element, True if any element in the tensor is True, False otherwise.
+    fn bool_any<const D: usize>(tensor: BoolTensor<B, D>) -> BoolTensor<B, 1> {
+        let sum = B::int_sum(B::bool_into_int(tensor.clone()));
```
louisfd, line 316: You shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/bool_tensor.rs
```diff
@@ -302,4 +302,71 @@ pub trait BoolTensorOps<B: Backend> {
...
+    /// * `dim` - The axis along which to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, D, Bool>` with the same size as input `tensor`, except in the `dim` axis
+    /// where the size is 1. The elem in the `dim` axis is True if any element along this dim in the input
+    /// evaluates to True, False otherwise.
+
+    fn bool_any_dim<const D: usize>(tensor: BoolTensor<B, D>, dim: usize) -> BoolTensor<B, D> {
+        let sum = B::int_sum_dim(B::bool_into_int(tensor.clone()), dim);
```
louisfd, line 334: Yous shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/bool_tensor.rs
```diff
@@ -302,4 +302,71 @@ pub trait BoolTensorOps<B: Backend> {
...
+    ///
+    /// * `tensor` - The tensor to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, 1, Bool>` with a single element, True if all elements in the input tensor
+    /// evaluate to True, False otherwise.
+    fn bool_all<const D: usize>(tensor: BoolTensor<B, D>) -> BoolTensor<B, 1> {
+        let num_elems = B::bool_shape(&tensor).num_elements();
+        let sum = B::int_sum(B::bool_into_int(tensor.clone()));
```
louisfd, line 350: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/bool_tensor.rs
```diff
@@ -302,4 +302,71 @@ pub trait BoolTensorOps<B: Backend> {
...
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, D, Bool>` with the same size as input `tensor`, except in the `dim` axis
+    /// where the size is 1. The elem in the `dim` axis is True if all elements along this dim in the input
+    /// evaluates to True, False otherwise.
+
+    fn bool_all_dim<const D: usize>(tensor: BoolTensor<B, D>, dim: usize) -> BoolTensor<B, D> {
+        let num_elems = B::bool_shape(&tensor).dims[dim];
+        let sum = B::int_sum_dim(B::bool_into_int(tensor.clone()), dim);
```
louisfd, line 369: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -997,4 +997,80 @@ pub trait IntTensorOps<B: Backend> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `tensor` - The tensor to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor with a single element, True if any element in the tensor is True, False otherwise.
+    fn int_any<const D: usize>(tensor: IntTensor<B, D>) -> BoolTensor<B, 1> {
+        let bool_tensor = B::int_equal_elem(tensor.clone(), 0.elem());
```
louisfd, line 1011: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -997,4 +997,80 @@ pub trait IntTensorOps<B: Backend> {
...
+    /// * `dim` - The axis along which to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, D, Bool>` with the same size as input `tensor`, except in the `dim` axis
+    /// where the size is 1. The elem in the `dim` axis is True if any element along this dim in the input
+    /// evaluates to True, False otherwise.
+
+    fn int_any_dim<const D: usize>(tensor: IntTensor<B, D>, dim: usize) -> BoolTensor<B, D> {
+        let bool_tensor = B::int_equal_elem(tensor.clone(), 0.elem());
```
louisfd, line 1031: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -997,4 +997,80 @@ pub trait IntTensorOps<B: Backend> {
...
+    /// * `tensor` - The tensor to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, 1, Bool>` with a single element, True if all elements in the input tensor
+    /// evaluate to True, False otherwise.
+
+    fn int_all<const D: usize>(tensor: IntTensor<B, D>) -> BoolTensor<B, 1> {
+        let num_elems = B::int_shape(&tensor).num_elements();
+        let bool_tensor = B::int_equal_elem(tensor.clone(), 0.elem());
```
louisfd, line 1050: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -997,4 +997,80 @@ pub trait IntTensorOps<B: Backend> {
...
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, D, Bool>` with the same size as input `tensor`, except in the `dim` axis
+    /// where the size is 1. The elem in the `dim` axis is True if all elements along this dim in the input
+    /// evaluates to True, False otherwise.
+
+    fn int_all_dim<const D: usize>(tensor: IntTensor<B, D>, dim: usize) -> BoolTensor<B, D> {
+        let num_elems = B::int_shape(&tensor).dims[dim];
+        let bool_tensor = B::int_equal_elem(tensor.clone(), 0.0f32.elem());
```
louisfd, line 1071: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -1179,4 +1179,78 @@ pub trait FloatTensorOps<B: Backend> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `tensor` - The tensor to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor with a single element, True if any element in the tensor is True, False otherwise.
+    fn float_any<const D: usize>(tensor: FloatTensor<B, D>) -> BoolTensor<B, 1> {
+        let bool_tensor = B::float_equal_elem(tensor.clone(), 0.0f32.elem());
```
louisfd, line 1193: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -1179,4 +1179,78 @@ pub trait FloatTensorOps<B: Backend> {
...
+        let bool_tensor = B::bool_not(bool_tensor);
+        let sum = B::float_sum(B::bool_into_float(bool_tensor));
+        B::float_greater_elem(sum, 0.0f32.elem())
```
louisfd, line 1196: I think `0.elem()` suffices?

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -1179,4 +1179,78 @@ pub trait FloatTensorOps<B: Backend> {
...
+    /// * `dim` - The axis along which to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, D, Bool>` with the same size as input `tensor`, except in the `dim` axis
+    /// where the size is 1. The elem in the `dim` axis is True if any element along this dim in the
+    /// input evaluates to True, False otherwise.
+
+    fn float_any_dim<const D: usize>(tensor: FloatTensor<B, D>, dim: usize) -> BoolTensor<B, D> {
+        let bool_tensor = B::float_equal_elem(tensor.clone(), 0.0f32.elem());
```
louisfd, line 1213: you shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -1179,4 +1179,78 @@ pub trait FloatTensorOps<B: Backend> {
...
+    ///
+    /// * `tensor` - The tensor to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, 1, Bool>` with a single element, True if all elements in the input tensor
+    /// evaluate to True, False otherwise.
+    fn float_all<const D: usize>(tensor: FloatTensor<B, D>) -> BoolTensor<B, 1> {
+        let num_elems = B::float_shape(&tensor).num_elements();
+        let bool_tensor = B::float_equal_elem(tensor.clone(), 0.0f32.elem());
```
louisfd, line 1231: shouldn't need to clone tensor

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -1179,4 +1179,78 @@ pub trait FloatTensorOps<B: Backend> {
...
+    /// * `dim` - The axis along which to test.
+    ///
+    /// # Returns
+    ///
+    /// A boolean tensor `Tensor<B, D, Bool>` with the same size as input `tensor`, except in the `dim` axis
+    /// where the size is 1. The elem in the `dim` axis is True if all elements along this dim in the input
+    /// evaluates to True, False otherwise.
+    fn float_all_dim<const D: usize>(tensor: FloatTensor<B, D>, dim: usize) -> BoolTensor<B, D> {
+        let num_elems = B::float_shape(&tensor).dims[dim];
+        let bool_tensor = B::float_equal_elem(tensor.clone(), 0.0f32.elem());
```
louisfd, line 1251: you shouldn't need to clone tensor

crates/burn-tensor/src/tests/mod.rs
```diff
@@ -79,6 +79,8 @@ macro_rules! testgen_all {
         burn_tensor::testgen_transpose!();
         burn_tensor::testgen_tri!();
         burn_tensor::testgen_powf!();
+        burn_tensor::testgen_any!();
+        burn_tensor::testgen_all_op!();
```
louisfd, line 83: oh we had a name clash here! all_op is a good name

crates/burn-tensor/src/tests/ops/any.rs
```diff
@@ -0,0 +1,46 @@
+#[burn_tensor_testgen::testgen(any)]
+mod tests {
+    use super::*;
+    use burn_tensor::{Data, Tensor};
+
+    #[test]
+    fn test_any() {
```
louisfd, line 7: I think we should also have versions of this test that give false. Otherwise the op could just be `return true`

crates/burn-tensor/src/tests/ops/all.rs
```diff
@@ -0,0 +1,46 @@
+#[burn_tensor_testgen::testgen(all_op)]
+mod tests {
+    use super::*;
+    use burn_tensor::{Data, Tensor};
+
+    #[test]
+    fn test_all() {
```
louisfd, line 7: I think we should also have versions of this test that give true. Otherwise the op could just be `return false`

louisfd (changes requested):
Awesome! Thank you. 
The logic is good everywhere. I do have some comments though, most of them minor but one is important, you must not do tensor.clone() if you're not obligated, this will augment reference count and may prevent operations to happen in-place. To make sure no occurrence of the clone is forgotten i commented them all 😄

crates/burn-tensor/src/tests/mod.rs
antimora: For consistency, I would have renamed `testgen_all` to something else. Maybe `testgen_every_op`?

louisfd:
@ashdtu While we're here, can you change this LSTM test line in lstm.rs (which was the original reason I needed the any op):
```
 // Asserts the gradients exist and are non zero
 assert!(*some_gradient.abs().sum().into_data().value.first().unwrap() > 0.);
 ```

crates/burn-tensor/src/tests/mod.rs
louisfd: Not a bad idea, but goes beyond the scope of this PR. all_op is good for now

PR #1348, opened 2024-02-22
Add sub-crates as members of workspace

antimora:
Sub-crates need to be added separately under members, otherwise they're skip in the build/run-checks. 

### Checklist

### Changes

Added burn-import sub-crates.

### Testing

run-checks all


syl20bnr (approved):
Good catch! 👍
I think it is OK that we list both of them explicitly as done in this PR and don't rely too much on wildcards.

PR #1359, opened 2024-02-24
Refactor/elemwise/kernel selection + dynamic fused inplace operations and broadcasting

nathanielsimard:
Refactor elemwise fused kernels to use the new compilation settings to identify kernels instead of pre-compiling every possible combination and having a complex selection process. This allows support for dynamic partial in-place operations as well as avoiding the execution of broadcasting procedures in fused kernels when not necessary (a new kernel is going to be created). These are very finely grained optimizations; no real speed improvements are going to be noticeable at first, but with more aggressive fusion and a longer stream of operations fused, it will help in reducing allocated memory.


crates/burn-wgpu/src/codegen/compilation.rs
```diff
@@ -20,23 +28,74 @@ pub struct CompilationInfo {
...
     vectorization: Vectorization,
-    inplace_available: bool,
     workgroup_size: WorkgroupSize,
+    reading_strategy: Vec<(u16, ReadingStrategy)>,
+}
+
+impl core::fmt::Display for CompilationSettings {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // The goal of this implementation is to generate the shortest representation
+        // that won't clash with any other compilation settings. This is crusial since we rely on
```
louisfd, line 53: crucial

crates/burn-wgpu/src/codegen/dialect/gpu/scope.rs
```diff
@@ -154,6 +154,16 @@ impl Scope {
         }
     }
 
+    pub(crate) fn read_globals(&self) -> Vec<(u16, ReadingStrategy)> {
+        self.reads_global
+            .iter()
+            .map(|(var, strategy, _)| match var {
+                Variable::GlobalInputArray(id, _) => (*id, *strategy),
+                _ => panic!("Wow this is not supposed to happen"),
```
louisfd, line 162: Wow? 😏

louisfd (approved):
LGTM

PR #1375, opened 2024-02-27
Make backend names in JSON reports match burnbench CLI 

errordeveloper:
### Checklist

### Related Issues/PRs

### Changes

### Testing

### DIFF
--- a/backend-comparison/src/persistence/base.rs
+++ b/backend-comparison/src/persistence/base.rs
@@ -28,6 +28,7 @@ pub struct BenchmarkRecord {
 ///  [
 ///    {
 ///      "backend": "backend name",
+///      "backendConfigName": "backend config name as appers in burnbench flag",
 ///      "device": "device name",
 ///      "gitHash": "hash",
 ///      "max": "duration in microseconds",
@@ -68,7 +69,7 @@ pub fn save<B: Backend>(
 
     let records: Vec<BenchmarkRecord> = benches
         .into_iter()
-        .map(|bench| BenchmarkRecord {
+        .map(|bench: BenchmarkResult| BenchmarkRecord {
             backend: B::name().to_string(),
             device: format!("{:?}", device),
             system_info: BenchmarkSystemInfo::new(),
@@ -157,6 +158,7 @@ impl Serialize for BenchmarkRecord {
             self,
             ("backend", &self.backend),
             ("device", &self.device),
+            ("backendConfigName", &self.results.backend_config_name),
             ("gitHash", &self.results.git_hash),
             ("max", &self.results.computed.max.as_micros()),
             ("mean", &self.results.computed.mean.as_micros()),
@@ -190,6 +192,7 @@ impl<'de> Visitor<'de> for BenchmarkRecordVisitor {
             match key.as_str() {
                 "backend" => br.backend = map.next_value::<String>()?,
                 "device" => br.device = map.next_value::<String>()?,
+                "backendConfigName" => br.results.backend_config_name = Some(map.next_value::<String>()?),
                 "gitHash" => br.results.git_hash = map.next_value::<String>()?,
                 "name" => br.results.name = map.next_value::<String>()?,
                 "max" => {
@@ -282,6 +285,7 @@ mod tests {
         let sample_result = r#"{
             "backend": "candle",
             "device": "Cuda(0)",
+            "backendConfigName": "candle-cuda",
             "gitHash": "02d37011ab4dc773286e5983c09cde61f95ba4b5",
             "name": "unary",
             "max": 8858,
@@ -345,6 +349,7 @@ mod tests {
         let record = serde_json::from_str::<BenchmarkRecord>(sample_result).unwrap();
         assert!(record.backend == "candle");
         assert!(record.device == "Cuda(0)");
+        assert!(record.results.backend_config_name == "candle-cuda");
         assert!(record.results.git_hash == "02d37011ab4dc773286e5983c09cde61f95ba4b5");
         assert!(record.results.name == "unary");
         assert!(record.results.computed.max.as_micros() == 8858);
--- a/backend-comparison/src/persistence/base.rs
+++ b/backend-comparison/src/persistence/base.rs
@@ -15,6 +15,7 @@ use std::{fs, io::Write};
 pub struct BenchmarkRecord {
     backend: String,
     device: String,
+    feature: String,
     system_info: BenchmarkSystemInfo,
     pub results: BenchmarkResult,
 }
@@ -29,6 +30,7 @@ pub struct BenchmarkRecord {
 ///    {
 ///      "backend": "backend name",
 ///      "device": "device name",
+///      "feature": "feature name",
 ///      "gitHash": "hash",
 ///      "max": "duration in microseconds",
 ///      "mean": "duration in microseconds",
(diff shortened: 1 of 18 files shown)


errordeveloper:
I found this quite handy, and thought I should open it up for comments from maintainers.

I've used this to plot the data, which I can also contribute somewhere.

```
import json
import os
import seaborn as sns
import matplotlib.pyplot as plt
from pandas import DataFrame

data = {
    "benchmark": [],
    "backendName": [],
    "mean": [],
    "median": [],
    "min": [],
    "max": []
}

for file in os.listdir("./"):
    if file.endswith(".json"):
        with open(file, "r") as f:
            result = json.load(f)

            data["benchmark"].append(result["name"])
            data["backendName"].append(result["backendName"])
            data["mean"].append(result["mean"] / 1000.0)  # convert to miliseconds keeping decimal point
            data["median"].append(result["median"])
            data["min"].append(result["min"])
            data["max"].append(result["max"])

plt.figure(figsize=(28, 20))

benchmarks = DataFrame(data)

p = sns.swarmplot(
    data=benchmarks,
    x="backendName", y="mean", hue="benchmark",
    log_scale=False,
    legend=True,
    palette="Set2",
    s=10, marker="D", linewidth=1,
)

for i in p.get_ygridlines():
    i.set_linestyle(':')
    i.set_linewidth(0.5)
    i.set_visible(True)

plt.ylabel("Mean Duration (miliseconds)")
plt.xlabel("Backends")

plt.legend(title="Benchmark")

plt.xticks(rotation=45)

plt.show()
```

backend-comparison/src/persistence/base.rs
```diff
@@ -31,11 +31,11 @@ pub struct BenchmarkRecord {
...
-///      "min": "duration in seconds",
-///      "max": "duration in seconds",
-///      "median": "duration in seconds",
-///      "mean": "duration in seconds",
-///      "variance": "duration in seconds"
+///      "min": "duration in microseconds",
+///      "max": "duration in microseconds",
+///      "median": "duration in microseconds",
+///      "mean": "duration in microseconds",
+///      "variance": "duration in microseconds"
```
errordeveloper, line 38: I suppose this should be split into a separate commit. Ideally, it'd be great if this was in the benchmarking doc. It took me an attempt to change the formatting of results on stdout before I found out about the JSON files.

nathanielsimard (changes requested):
I don't think we should pass the backend name in all function calls, I would put it into the benchmark trait. In the future we might change that backend_name to any runtime information that we need to collect.

errordeveloper:
@nathanielsimard sure, I've not spent any time thinking about the design, was just getting something to work and figured I'd open a PR and see if this is deemed useful at all.

errordeveloper:
@nathanielsimard PTAL, I've moved the logic into `Backend` and `Benchmark` crates  :)

crates/burn-ndarray/src/backend.rs
```diff
@@ -21,6 +21,18 @@ impl Default for NdArrayDevice {
     }
 }
 
+fn config_name() -> String {
```
errordeveloper, line 24: I've moved this one to the top of the file since it didn't seem right to have features config macros inside the trait directly, so this way it seemed more readable to me, less hidden surprises ;)

nathanielsimard (commented):
There is something that bothers me with the naming/implementation. There is a lot of overlapping information.

I think it might be better to have somthing as such:

```rust
fn device_features(device: &B::Device) -> Vec<String> {
    let mut features = Vec::new();
    
    #[cfg(feature = "std")]
    features.push("std".into());

    #[cfg(feature = "blas-openblas")]
    features.push("openblas".into());
    
    match device {
        Cuda => features.push("cuda".into());
        Metal => features.push("metal".into());
    };

    features
}
```

@syl20bnr do you have an opinion on this?

syl20bnr:
I feel that we can do much simpler by sticking closer to what this PR is all about. We could just store the arguments passed on the command line. The serialized arguments would be an object `burnbenchArgs` with a field `backend` and a field `bench` (although this one is less valuable). If later we add some new options to the CLI it might be handy to add it to the benchmark results so that we can reconstruct the original command line for other people to try executing the same benchmark.

errordeveloper:
@syl20bnr I agree in principle, although I'm not sure if I have more time for this. I did say in the PR title that it's about matching the names of backends, which is one surface area, but there are other ways of looking at it. Namely, I just want to plot the results and need keys that I could relate to. If burnbench just produced a plot for me, I'd also be quite happy and wouldn't need to worry about all this. So I'm not necessarily sure preservation of CLI arguments is the right UX facet to lean on, perhaps producing plots could be more useful, the rest is implementation detail. Of course, it depends on how far you want to take burnbench.

syl20bnr:
I understand and the current interaction between burnbench and the actual benchmarks is a bit contrieved and it would need a bit of refactoring. Currently the benchmark itself is responsible of persistence and upload of results which should now be the responsibility of our burnbench runner. Once we moved these responsibilities to our runner it will be easier to serialize more contextual informations and more.

For the plots we are indeed working on it and it will soon be announced on our discord, so stay tune!

errordeveloper:
So actually right now preserving CLI args would require some additional tricks, because results are stored per benchmark (and, e.g. data benchmark produces multiple files). Any burnbench invocations with multiple backends and multiple benchmarks would require some special treatment, as backends and benchmarks pass in the CLI would need to be mapped to results actual results and translated to a single invocation. Maybe it's not too hard as there is `cargo bench` invocation, but that will add more usability friction too, as backends map to features ;)

syl20bnr:
We have been thinking about this and adding a new function to the backend trait just for this is too invasive. In the current situation the probably best option if to modify the macro in `lib.rs` to pass the feature name to the `bench` function and then to the `save` function for serialization. Then the name can be retrieved by the CLI to be displayed in the report.

antimora:
@syl20bnr what do we do with this PR? Keep it open?

syl20bnr:
Yes we can keep it open. @errordeveloper do you have time to change the implementation ? If not I can do it,

errordeveloper:
The suggestion makes sense, however I don’t think I will find the time to implement it any time soon.

antimora:
OK. I have assigned to @syl20bnr then. Thank you, @errordeveloper , for the update and PR.

syl20bnr:
I made the changes. Here is an example of a report:

| Benchmark  | Feature     | Backend           | Device        | Median         |
|------------|-------------|-------------------|---------------|----------------|
| unary      | wgpu        | jit<wgpu>         | BestAvailable | 393.000µs      |
| unary      | wgpu-fusion | fusion<jit<wgpu>> | BestAvailable | 272.000µs      |
| binary     | wgpu        | jit<wgpu>         | BestAvailable | 820.000µs      |
| binary     | wgpu-fusion | fusion<jit<wgpu>> | BestAvailable | 396.000µs      |
| to_data    | wgpu        | jit<wgpu>         | BestAvailable | 34.270ms       |
| from_data  | wgpu        | jit<wgpu>         | BestAvailable | 154.599ms      |
| to_data    | wgpu-fusion | fusion<jit<wgpu>> | BestAvailable | 32.761ms       |
| from_data  | wgpu-fusion | fusion<jit<wgpu>> | BestAvailable | 39.941ms       |
| matmul     | wgpu        | jit<wgpu>         | BestAvailable | 980.000µs      |
| matmul     | wgpu-fusion | fusion<jit<wgpu>> | BestAvailable | 1.715ms        |
| max_pool2d | wgpu        | jit<wgpu>         | BestAvailable | 4.855ms        |
| max_pool2d | wgpu-fusion | fusion<jit<wgpu>> | BestAvailable | 4.299ms        |

errordeveloper:
Great we came to a consensus, it is quite similar to my original change, albeit argument is passed to ‘save’ instead of ‘run_benchmark’.

syl20bnr:
Indeed we went a bit full-circle on this one 😁 
Thank you for the PR!

PR #1380, opened 2024-02-28
Migrate NaiveReduceDim + Instruction set for SharedReduceDim

louisfd:
Two related yet separated things in this PR
- NaiveReduceDim migrated to gpu representation, supporting sum, mean, argmin, argmax
- Instruction set needed for SharedReduceDim was added:
    - workgroup size
    - local invocation id
    - num workgroups
    - workgroup barrier
    - while loop
    - shared memory


crates/burn-wgpu/src/codegen/dialect/wgsl/compiler.rs
```diff
@@ -147,6 +155,12 @@ impl<F: FloatElement, I: IntElement> Compiler<F, I> {
             gpu::Variable::ConstantScalar(index, elem) => {
                 wgsl::Variable::ConstantScalar(index, Self::compile_elem(elem))
             }
+            gpu::Variable::SharedMemory(index, item, size) => {
+                let item = Self::compile_item(item);
+                self.shared_memories
+                    .push(SharedMemory::new(index, item, size));
```
louisfd, line 161: TODO: this declares the shared memory another time each time we use it:
```
var<workgroup> shared_memory_0: array<f32, 1024>;

var<workgroup> shared_memory_0: array<f32, 1024>;

var<workgroup> shared_memory_0: array<f32, 1024>;
```

crates/burn-wgpu/src/codegen/dialect/gpu/scope.rs
```diff
@@ -306,4 +314,13 @@ impl Scope {
         self.locals.push(local);
         local
     }
+
+    /// Create a local variable of the given [item type](Item).
```
nathanielsimard, line 318: Create a _shared_ variable

crates/burn-wgpu/src/codegen/dialect/wgsl/compiler.rs
nathanielsimard: Just a contained index should fix this, since it's supposed to be a very small vector, we shouldn't use a set or an hashmap.

louisfd:
Closing. I will make a more comprehensive one imminently.

PR #1381, opened 2024-02-28
[backend-comparison] Upload benchmarks results to server

syl20bnr:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/1072

### Changes

Uploading is enabled with already implemented `--share` argument of the `burnbench` command line tool.

The `burnbench` binary passes the URL of the server and the auth token to the cargo bench process using the additional arguments `--sharing-url` and `--sharing-token` respectively.

The persistence module then upload the results when a `--sharing-url` is provided.

The URL is for now hardcoded. The endpoint is production when compiling in release mode and it is localhost otherwise.

### Tests

```
cargo run --bin burnbench --release -- auth
cargo run --bin burnbench --release -- run --benches unary --backends ndarray --share
```

### DIFF
--- a/backend-comparison/src/burnbenchapp/base.rs
+++ b/backend-comparison/src/burnbenchapp/base.rs
@@ -16,13 +16,26 @@ use super::{
 };
 
 const FIVE_SECONDS: time::Duration = time::Duration::new(5, 0);
+const USER_BENCHMARK_SERVER_URL: &str = if cfg!(debug_assertions) {
+    // development
+    "http://localhost:8000/benchmarks"
+} else {
+    // production
+    "https://user-benchmark-server-gvtbw64teq-nn.a.run.app/benchmarks"
+};
 
 /// Base trait to define an application
 pub(crate) trait Application {
     fn init(&mut self) {}
 
     #[allow(unused)]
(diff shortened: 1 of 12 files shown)


backend-comparison/src/burnbenchapp/base.rs
```diff
@@ -16,13 +16,26 @@ use super::{
 };
 
 const FIVE_SECONDS: time::Duration = time::Duration::new(5, 0);
+const USER_BENCHMARK_SERVER_URL: &str = if cfg!(debug_assertions) {
+    // development
+    "http://localhost:8000/benchmarks"
+} else {
+    // production
+    "https://user-benchmark-server-gvtbw64teq-nn.a.run.app/benchmarks"
```
nathanielsimard, line 24: I guess this URL will change right?

nathanielsimard (approved):

backend-comparison/src/burnbenchapp/base.rs
syl20bnr: It will not change as long as we keep the same service in Cloud Run.
Anyway this is temporary we will change this to use our domain at some point.

PR #1383, opened 2024-02-29
[CI] Add stale action

Luni-4:
### Changes

This PR adds the `stale` action. In the `yml` file there are the configuration parameters

### DIFF
--- a/.github/workflows/stale-pr.yml
+++ b/.github/workflows/stale-pr.yml
@@ -1,0 +1,42 @@
+name: Stale Pull Requests
+
+on:
+  schedule:
+    - cron: '0 12 * * *' # Run every day at 12:00 (UTC)
+
+# The minimum permissions required to run this Action
+permissions:
+  contents: write # only for delete-branch option
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale-pr:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Stale pull requests
+      uses: actions/stale@v9
+      with:
+        # The idle number of days before marking issues stale.
+        #
+        #  With a negative number like -1, no issues
+        #  will be marked as stale automatically.
+        days-before-issue-stale: -1
+        # The idle number of days before marking pull requests stale
+        days-before-pr-stale: 30
+        # The idle number of days before closing
+        # the stale pull requests (due to the stale label).
+        #
+        # With a negative number like -1, the pull requests
+        # will never be closed automatically.
+        days-before-pr-close: -1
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Stale pull requests
+      uses: actions/stale@v9
+      with:
+        # The idle number of days before marking issues stale.
+        #
+        #  With a negative number like -1, no issues
+        #  will be marked as stale automatically.
+        days-before-issue-stale: -1
+        # The idle number of days before marking pull requests stale
+        days-before-pr-stale: 20
```
syl20bnr, line 30: Let's try one month.

.github/workflows/stale-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+        # The idle number of days before closing
+        # the stale pull requests (due to the stale label).
+        #
+        # With a negative number like -1, the pull requests
+        # will never be closed automatically.
+        days-before-pr-close: -1
+        # Label to apply on staled pull requests
+        stale-pr-label: 'stale'
+        # The message that will be added as a comment to the pull request
+        stale-pr-message: 'This PR is stale for 20 days'
```
syl20bnr, line 40:
```suggestion
        stale-pr-message: ''This PR has been marked as stale because it has not been updated for over a month."
```

.github/workflows/stale-pr.yml
```diff
@@ -0,0 +1,42 @@
+name: Stale Pull Requests
+
+on:
+  schedule:
+    - cron: '0 12 * * FRI' # Run every Friday at 12:00 (UTC)
```
syl20bnr, line 5: We can run it every day I believe.

syl20bnr (approved):
Awesome, thank you!
Minor tweaks and then it is good to go.

.github/workflows/stale-pr.yml
Luni-4: I have chosen this configuration to reduce the amount of jobs for CI

antimora: Yes, I think we should run daily because +7 would make an issue double stale =D

Also, since this will remove a stale label if there is a comment/activity, we should remove it quickly:

```
# Remove `stale` label from pull requests on updates/comments
        remove-pr-stale-when-updated: true
```

antimora (changes requested):
This is a great addition and help with lots of PRs. One small request to make it daily.

antimora (approved):
LGTM. Thank you for this addition!

syl20bnr (approved):

Issue #856, opened 2023-10-10
Add element-wise minimum and other element-wise operations for tensors


EddieMataEwy:
### Feature description

Element wise functions such as getting the minimum element-wise of 2 tensors. Like torch.minimum().

### Feature motivation

Right now to do element wise minimum of 2 tensors I believe I need to do:
`tensor_1.clone().mask_where(tensor_2.clone().lower(tensor_1.clone()), tensor_2.clone())`
It would be better to do something like:
`let minimum = Tensor::min(tensor_1, tensor_2)`
Another example would be the clamp function, which can clamp the tensor only by fixed numbers and you would have to apply a couple similar `mask_where` as in the previous example.

### (Optional) Suggest a Solution

I think we should find which of these functions are missing first, and then depending on the optimization complexity we can implement them directly or do a simple wrap like the example provided.
Do you agree?

louisfd:
Hi @EddieMataEwy 
Thanks for the suggestion. 

The clamp functions already exist, they have this default implementation (clamp_min shown here):
```
let mask = Self::lower_elem(tensor.clone(), min);
B::mask_fill(tensor, mask, min)
```
And then for every backend we are free to re-implement it, like it is done in the torch backend where the torch clamp functions are called, or not, like in WGPU for now, so it falls back to using lower_elem and mask_fill. 

We could do this strategy for binary min and max, using the line you suggested as a default implementation, then re-implementing it for every backend will be a fun optimization to do. 

I'll tag this feature as a good first issue as you can inspire yourself from clamp.

EddieMataEwy:
I think we should also consider the possibility of allowing broadcastable tensors on some of these functions.
Say we want to clamp tensor_1 with tensor_2 as the lower value and with the number 3 as the higher value.
We could create a tensor full of 3s with the same size or allow a tensor of the same type with a single element 3 that will be used in all comparisons.
Of course, broadcastable tensors could have more elements.

What do you think?

Maybe we can decide on that down the road and focus on same size tensors first.

louisfd:
I believe we could repeat the pattern used for some other ops like addition. For addition we have both add and add_scalar. The former sums two tensors of rank D element-wise (with support for broadcast) and the former adds a scalar to every element. An advantage of this is that, for the scalar version, we do not have to launch a computation kernel with two tensor buffers as inputs, we can use a unary operation. So the result would be:
`tensor_1.clamp_min(tensor_2).clamp_max_scalar(3.)`

[...]

PR #1385, opened 2024-02-29
Element wise min/max between a pair of tensors

boondocklabs:
### Checklist

```
[2024-02-29T21:16:10Z INFO  xtask::runchecks] Time elapsed for the current execution: 00:06:31
```

### Related Issues/PRs

#856 

### Changes

Adds methods `min_pair()` and `max_pair()` to numeric tensors to provide pairwise min/max between a pair of tensors.

### Testing

Unit tests have been added to `minmax.rs`

### DIFF
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -425,7 +425,7 @@ where
     ///
     /// # Returns
     ///
-    /// A tensor `Tensor<B, D, K>` with the same shape as the input tensors containing the minium value found
+    /// A tensor `Tensor<B, D, K>` with the same shape as the input tensors containing the minimum value found
     /// between each element of the two source tensors.
     pub fn min_pair(self, other: Self) -> Self {
         let mask = other.clone().lower(self.clone());
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -357,7 +357,7 @@ where
         (tensor, index)
     }
 
-    /// Find the maximum pair wise values with another Tensor
+    /// Finds the maximum pair wise values with another Tensor
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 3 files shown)


crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -357,6 +357,21 @@ where
         (tensor, index)
     }
 
+    /// Find the maximum pair wise values with another Tensor
```
antimora, line 360: Find => Finds

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -357,6 +357,21 @@ where
...
+    ///
+    /// # Arguments
+    ///
+    /// * `other` - Other tensor to find maximum elements with
+    ///
+    /// # Returns
+    ///
+    /// A tensor `Tensor<B, D, Float>` with the same shape as the input tensors containing the maximum value found
```
antimora, line 368: I would remove `Tensor<B, D, Float>` because I think it works for int too.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -402,6 +417,21 @@ where
         (tensor, index)
     }
 
+    /// Find the minimum pair wise values with another Tensor
```
antimora, line 420: Find => Finds

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -402,6 +417,21 @@ where
...
+    ///
+    /// # Arguments
+    ///
+    /// * `other` - Other tensor to find minimum elements with
+    ///
+    /// # Returns
+    ///
+    /// A tensor `Tensor<B, D, K>` with the same shape as the input tensors containing the minimum value found
```
antimora, line 428: I'd remove `Tensor<B, D, K>`

antimora (changes requested):
Great! Thanks for the addition. 

I have very minor text update request.

crates/burn-tensor/src/tensor/api/numeric.rs
boondockenergy: Singular `Find` was inline with other comments. Do you want existing ones changed to plural as well?

antimora: It's a description that should read as [This function] Finds ... 

Others should follow the same but you can leave other functions as is for now.

antimora (approved):
LGTM. Thank you for adding it.

Waiting to hear from others.

PR #1391, opened 2024-03-01
bool cast kernel

### LINKED ISSUE #1315: Casting to Int/Float of Boolean Tensors for WGPU backend
**Describe the bug**
When boolean tensors are casted to int/float type using the .int()/.float() methods, the casted tensor return the assembly representations for 1 as 1065353216 instead of the actual 1. This is not the behaviour with the other backends which correctly return 1 for Int Tensor or 1.0 for float tensor respectively.

**To Reproduce**
let device = WgpuDevice::default();
type Backend = Wgpu;

// For Int Casting
let tensor_1 = Tensor::<Backend, 2>::from_floats([[1.0, 0.0, 3.0],[0.0, 0.0, 900.0]], &device);
let tensor_2 = tensor_1.clone().greater_elem(0.0).int(); //
println!("{:?}", tensor_2.into_data());

// Output: Data { value: [1065353216, 0, 1065353216, 0, 0, 1065353216], shape: Shape { dims: [2, 3] } }

// For float casting
let tensor_2 = tensor_1.clone().greater_elem(0.0).float(); //
println!("{:?}", tensor_2.into_data());
// Output: Data { value: [1065353200.0, 0.0, 1065353200.0, 0.0, 0.0, 1065353200.0], shape: Shape { dims: [2, 3] } }

**Expected behavior**
With other backends (ndarray, tch-cpu, tch-gpu) that is not the case and it correctly prints:
Data { value: [1, 0, 1, 0, 0, 1], shape: Shape { dims: [2, 3] } }

### LINKED ISSUE #1388: Tensor `all` operator results is incorrect on burn-wgpu when using with `lower_equal`
**Describe the bug**
The results are incorrect for burn-wgpu backend if used in conjunction with `lower_equal`. It works for burn-ndarray. And oddly under fusion it passes.

**To Reproduce**
Add the following test into `crates/burn-tensor/src/tests/ops/all.rs` and run the test under burn-wgpu:

```
[burn-wgpu]% cargo test test_bug1
```

```rust
#[test]
fn test_bug1() {
    let tensor1 = TestTensor::from([[0.0, 1.0, 0.0], [1.0, -1.0, 1.0]]) + 1e-6;
    let tensor2 = TestTensor::from([[0.0, 1.0, 0.0], [1.0, -1.0, 1.0]]) + 1e-6;

    let ge = tensor1.lower_equal(tensor2);
    println!("ge: {:#?}", ge.clone().into_data());

    let all = ge.clone().all();
    println!("all {}", all.clone().into_data());

    assert_eq!(Data::from([true]), all.clone().into_data());
}
```

Results in:

```
[burn-wgpu]% cargo test test_bug1
    Finished test [unoptimized] target(s) in 0.08s
     Running unittests src/lib.rs (/Users/dilshod/Projects/burn/target/debug/deps/burn_wgpu-42c7d3086f1b764b)

running 2 tests
test tests::all_op::tests::test_bug1 ... FAILED
test fusion::base::tests::all_op::tests::test_bug1 ... ok

failures:

---- tests::all_op::tests::test_bug1 stdout ----
ge: Data {
    value: [
        true,
        true,
        true,
        true,
        true,
        true,
    ],
    shape: Shape {
        dims: [
            2,
            3,
        ],
    },
}
all [false]
thread 'tests::all_op::tests::test_bug1' panicked at crates/burn-wgpu/src/lib.rs:96:5:
  left: Data { value: [true], shape: Shape { dims: [1] } }
 right: Data { value: [false], shape: Shape { dims: [1] } }


failures:
    tests::all_op::tests::test_bug1

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 1306 filtered out; finished in 0.04s

error: test failed, to rerun pass `--lib`

```

But this passes with burn-ndarray:

```
[burn-ndarray]% ct test_bug1
   Compiling burn-tensor v0.13.0 (/Users/dilshod/Projects/burn/crates/burn-tensor)
   Compiling burn-autodiff v0.13.0 (/Users/dilshod/Projects/burn/crates/burn-autodiff)
   Compiling burn-ndarray v0.13.0 (/Users/dilshod/Projects/burn/crates/burn-ndarray)
    Finished test [unoptimized] target(s) in 1.40s
     Running unittests src/lib.rs (/Users/dilshod/Projects/burn/target/debug/deps/burn_ndarray-c1c0a5464ba1876e)

running 1 test
test tests::all_op::tests::test_bug1 ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 581 filtered out; finished in 0.00s
```

louisfd:
In wgpu, bools are u32, where false is zero and true is anything else. 
Sometimes it can be an optimization to not change the bits, for instance in
passing from 1_f32 which has bit representation 00111111 10000000 00000000 00000000 
to the bool true and using afterwards for inplace operations, it's practical to not have to change the bits to 00000000 00000000 00000000 00000001

However, when we explicitly casted bool to something else, we used a kernel that did not know we were using bool, it just knew it had a u32, so it was treated like an integer. 

Hence, 1_f32 became true, which in turn became 1065353216 or 1065353216.0 (which is 00111111 10000000 00000000 00000000 as a u32). 

The fix is to call a new cast kernel who explicitly knows it must consider its u32 inputs as bools. 

Fix #1315 
Fix #1388

### DIFF
--- a/crates/burn-tensor/src/tests/ops/all.rs
+++ b/crates/burn-tensor/src/tests/ops/all.rs
@@ -58,4 +58,15 @@ mod tests {
         let data_expected = Data::from([[false], [true]]);
         assert_eq!(data_expected, data_actual);
     }
+
+    #[test]
+    fn test_all_with_bool_from_lower_equal() {
+        let tensor1 = TestTensor::from([[0.0, 1.0, 0.0], [1.0, -1.0, 1.0]]) + 1e-6;
+        let tensor2 = TestTensor::from([[0.0, 1.0, 0.0], [1.0, -1.0, 1.0]]) + 1e-6;
+
+        let ge = tensor1.lower_equal(tensor2);
+        let all = ge.clone().all();
+
+        assert_eq!(Data::from([true]), all.clone().into_data());
+    }
 }
--- a/crates/burn-wgpu/src/kernel/cast.rs
+++ b/crates/burn-wgpu/src/kernel/cast.rs
@@ -1,7 +1,18 @@
-use super::{KernelSettings, SourceTemplate, StaticKernelSource, WORKGROUP_DEFAULT};
+use super::{
+    DynamicKernelSource, KernelSettings, SourceTemplate, StaticKernelSource, WORKGROUP_DEFAULT,
+};
 use crate::{
-    compute::StaticKernel, element::JitElement, kernel::elemwise_workgroup, kernel_wgsl,
-    tensor::JitTensor, Runtime,
+    codegen::{
+        dialect::gpu::{gpu, Elem, Item, Scope, Variable, Visibility},
+        execute_dynamic, Compilation, CompilationInfo, CompilationSettings, Compiler, EagerHandle,
+        InputInfo, OutputInfo, WorkgroupLaunch,
+    },
+    compute::StaticKernel,
+    element::JitElement,
+    kernel::elemwise_workgroup,
+    kernel_wgsl,
+    tensor::JitTensor,
+    Runtime,
 };
 use std::{any::TypeId, marker::PhantomData};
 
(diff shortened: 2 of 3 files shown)


antimora (approved):
Looks good to me! Thank you for fixing it.

nathanielsimard (approved):

PR #1417, opened 2024-03-05
Splitted the JIT stuff from the Wgpu stuff

nathanielsimard:
The JIT backend still have a dependency on wgsl, since some shaders are written in it and not yet ported to our new compiler, but this isn't a problem for now.

Deleted some matmul kernels that aren't used or not relevant anymore.
All tests in `burn-jit` now follow the same pattern as with `burn-tensor` and `burn-autodiff`; each tests in exported and run on the concrete implementation.


crates/burn-jit/Cargo.toml
```diff
@@ -0,0 +1,57 @@
+[package]
+authors = ["nathanielsimard <nathaniel.simard.42@gmail.com>"]
+categories = ["science"]
+description = "WGPU backend for the Burn framework"
```
antimora, line 4: Description is out of date

Issue #1422, opened 2024-03-06
Migrate all WGSL templates to JIT GPU representation


louisfd:
- [x] cast
- [x] cat
- [x] contiguous
- [x] prng (bernoulli)
- [x] prng (uniform)
- [x] prng (normal)
- [x] mask fill
- [x] mask where
- [x] matmul tiling 2d: padded
- [x] matmul tiling 2d: unpadded
- [x] prod/prod_dim
- [x] conv2d
- [x] conv_transpose2d
- [x] avg_pool2d
- [x] avg_pool2d backward
- [x] adaptive_avg_pool2d
- [x] adaptive_avg_pool2d backward
- [x] interpolate: bicubic
- [x] interpolate: bilinear
- [x] interpolate: nearest
- [x] interpolate backward: nearest

TomWyllie:
"very hard" indeed - I was going to add in that FFT module including WGSL shaders, shall I hold off until a few more of these are done and then write them in this new format?

antimora:
> "very hard" indeed - I was going to add in that FFT module including WGSL shaders, shall I hold off until a few more of these are done and then write them in this new format?

I would vote not to wait and add it anyway and we can migrate. Or @louisfd and @nathanielsimard might suggest to do it in a new format, maybe?

louisfd:
I think the best is you write the wgsl one and i can do the migration while i'm at it

PR #1424, opened 2024-03-06
JIT migration: contiguous kernel

louisfd:
Contiguous kernel for #1422


nathanielsimard (approved):
I think you forgot to delete the wgsl file, but otherwise LGTM.

PR #1432, opened 2024-03-07
Support for non-contiguous indexes in PyTorchFileRecorder keys

### LINKED ISSUE #1420: Support for non-contiguous indexes in PyTorchFileRecorder keys
### Feature description
In Pytorch's `Sequential`, on occasion an activation layer, such as one without parameters, gets added in a sequence like [nn.Sequential(conv, relu, conv, relu, conv)](https://github.com/modelscope/modelscope/blob/master/modelscope/models/cv/table_recognition/modules/lore_detector.py#L243). This process generates non-continuous keys, such as `0, 2, 4`. In the current design, each key must be mapped one-by-one during the deserialization process. This behavior can lead to the generation of empty maps. 

Generated keys:
```json
  "model.ax.0.weight",
  "model.ax.0.bias",
  "model.ax.2.weight",
  "model.ax.2.bias",
  "model.ax.4.weight",
  "model.ax.4.bias",
  "model.ax.6.weight",
  "model.ax.6.bias",
  "model.ax.8.weight",
  "model.ax.8.bias",
```
Occured issues: https://github.com/tracel-ai/burn/discussions/1413

Repair operation:
```rust
.with_key_remap("model.ax.2", "model.ax.1")
.with_key_remap("model.ax.4", "model.ax.2")
.with_key_remap("model.ax.6", "model.ax.3")
.with_key_remap("model.ax.8", "model.ax.4")
```

Is there a way we can enhance the system's capacity to disregard the continuity of keys and thus avoid the creation of empty maps? Alternatively, can we enable the addition of activation layers directly to the `Sequential`?

antimora:
### Checklist

### Related Issues/PRs

Fixes #1420

### Changes

1. Removes empty hashmap items from vector.
2. Update the document
3. Add new unit tests

### Testing

Added new unit tests and checked rendered book.


crates/burn-import/pytorch-tests/tests/non_contiguous_indexes/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        let mut fc = vec![];
+        for fc_record in record.fc.into_iter() {
+            fc.push(conv2d_config.init_with(fc_record));
+        }
+        Net { fc }
+    }
+
+    /// Forward pass of the model.
+    pub fn forward(&self, x: Tensor<B, 4>) -> Tensor<B, 4> {
+        let x = relu(self.fc[0].forward(x));
```
laggui, line 28: If this serves as an example, we could change the forward to work with any number of elements in the vec sequence.

```rust
self.fc.iter().fold(x, |x_i, conv| relu(conv.forward(x_i)))
```

laggui (approved):
Seems like a good solution for now since we don't have a 1-to-1 match with a flexible `nn.Sequential` module (and we know that people just love using `nn.Sequential` in PyTorch even if it's not always the cleanest approach). Will make it easier to import the weights.

One minor comment, otherwise LGTM! Approving in advance.

crates/burn-import/pytorch-tests/tests/non_contiguous_indexes/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        let x = relu(self.fc[1].forward(x));
+        let x = relu(self.fc[2].forward(x));
+        let x = relu(self.fc[3].forward(x));
+
+        relu(self.fc[4].forward(x))
```
laggui, line 33: If this serves as an example, we could change the forward to work with any number of elements in the vec sequence.

```rust
self.fc.iter().fold(x, |x_i, conv| relu(conv.forward(x_i)))
```

antimora: It's pretty solution!

PR #1433, opened 2024-03-08
JIT Migration: PRNG

louisfd:
PRNG kernels for #1422


crates/burn-jit/src/codegen/kernel.rs
```diff
@@ -152,11 +369,52 @@ fn execute_settings<'a, R: Runtime, E: JitElement>(
...
+            Elem::Int => 1,
+            Elem::UInt => 2,
+            Elem::Bool => panic!("Bool scalars are not supported"),
+        })
+        .collect();
+
+    let mut handles_scalars = Vec::new();
+    for i in 0..3 {
+        for (j, scalar_priority) in scalar_priorities.iter().enumerate() {
+            if scalar_priority == &i {
```
nathanielsimard, line 397: Create an array instead of a vec. We know it's 3 in length, so it's gonna be way more efficient.

nathanielsimard (changes requested):

PR #1434, opened 2024-03-08
Fix tch view data corruption

### LINKED ISSUE #1430: Odd overwriting of tensor values while using LibTorch backend
Hi, I have this rather odd problem where sometimes a Tensor's values are overwritten by the value of another calculation down the road. It only happens in the LibTorch backend, not with the NdArray backend (only one I've tested). 

It seems to be a result of some sort of combination of reshaping and taking the exponent. Sometimes, when I take a slice, the original value will be replaced by that of the exponent of the slice (which is cloned). It really was a gnarly bug to find in the code but I managed to make a very small working example.  

```rust
#[test]
fn bizarre() -> Result<()> {
    let zeros = Tensor::<NdArray, 1>::zeros([2], &NdArrayDevice::default());
    zeros.clone().slice([1..2]).reshape([1]).exp();
    //Works as expected
    assert_eq!(
        zeros.to_data(),
        Tensor::<NdArray, 1>::zeros([2], &NdArrayDevice::default()).to_data()
    );

    let zeros = Tensor::<LibTorch, 1>::zeros([2], &LibTorchDevice::default());
    zeros.clone().slice([1..2]).reshape([1]).clone().exp();
    //Works as expected thanks to the second clone after reshaping.
    assert_eq!(
        zeros.to_data(),
        Tensor::<LibTorch, 1>::zeros([2], &LibTorchDevice::default()).to_data()
    );

    let zeros = Tensor::<LibTorch, 1>::zeros([2], &LibTorchDevice::default());
    zeros.clone().slice([1..2]).reshape([1]).exp();
    //Doesn't work, leads to zeroes being equal to [0.0, 1.0]
    assert_eq!(
        zeros.to_data(),
        Tensor::<LibTorch, 1>::zeros([2], &LibTorchDevice::default()).to_data()
    );

    Ok(())
}
```

The first two asserts work fine whereas the last one fails, as the second value of zeros has been replaced by e^0=1. (If I try with different values the number is consistently equal to e^x). 

I have no idea what could possibly be causing this, but it happens with both the current release version (0.12.1) and with the current commit on the main branch.

Thanks!

nathanielsimard:
When we created a partial view of a tensor, we checked if the view reference was the same as the parent reference, which obviously created a bug. It's now fixed by being explicit about what kind of reference we are tracking.

fix #1430

### DIFF
--- a/crates/burn-tch/src/tensor.rs
+++ b/crates/burn-tch/src/tensor.rs
@@ -9,13 +9,58 @@ use std::{marker::PhantomData, sync::Arc};
 #[allow(clippy::arc_with_non_send_sync)]
 pub type StorageRef = Arc<*mut c_void>;
 
+/// A reference to a tensor storage.
+#[derive(PartialEq, Debug, Clone)]
+pub enum Storage {
+    /// When a tensor is a partial view of another tensor.
+    View {
+        /// Storage reference for the whole buffer.
+        buffer_ref: StorageRef,
+        /// Storage reference for the partial buffer.
+        view_ref: StorageRef,
+    },
+    /// When a tensor use all of its buffer.
+    Owned {
+        /// Storage reference for the whole buffer.
+        buffer_ref: StorageRef,
+    },
+}
+
+impl Storage {
+    /// Check if the storage can be used inplace.
+    pub fn can_mut(&self) -> bool {
+        match self {
+            Storage::View {
+                buffer_ref: start_ref,
+                view_ref,
+            } => Arc::strong_count(start_ref) == 1 && Arc::strong_count(view_ref) == 1,
+            Storage::Owned {
+                buffer_ref: start_ref,
+            } => Arc::strong_count(start_ref) == 1,
+        }
+    }
+
+    /// Get the whote buffer reference.
+    pub fn buffer_ref(&self) -> &StorageRef {
+        match self {
+            Storage::View {
+                buffer_ref: start_ref,
+                view_ref: _,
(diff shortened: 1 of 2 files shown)


crates/burn-tch/src/tensor.rs
```diff
@@ -9,13 +9,58 @@ use std::{marker::PhantomData, sync::Arc};
...
+                buffer_ref: start_ref,
+                view_ref,
+            } => Arc::strong_count(start_ref) == 1 && Arc::strong_count(view_ref) == 1,
+            Storage::Owned {
+                buffer_ref: start_ref,
+            } => Arc::strong_count(start_ref) == 1,
+        }
+    }
+
+    /// Get the whote buffer reference.
```
louisfd, line 43: whole

crates/burn-tensor/src/tests/ops/reshape.rs
```diff
@@ -70,6 +70,18 @@ mod tests {
         assert_eq!(reshaped.shape(), [4, 3].into());
     }
 
+    #[test]
+    fn should_not_corrupt_after_slice() {
+        let zeros = Tensor::<TestBackend, 1>::zeros([2], &Default::default());
+        zeros.clone().slice([1..2]).reshape([1]).exp();
+
+        // May leads to zeroes being equal to [0.0, 1.0]
```
louisfd, line 78: lead

louisfd (approved):
LGTM, modulo two typos

PR #1436, opened 2024-03-08
Add Enum module support in PyTorchFileRecorder

### LINKED ISSUE #1431: Record `Deserializer` not implemented for enum
**Describe the bug**
I stumbled upon this issue when importing a PyTorch model with `PyTorchFileRecorder` to load it into the Burn equivalent definition.

The PyTorch model defines a module that depends on a condition (returns a conv block in one case, and depthwise conv block in the other). For that, I implemented the block as an enum since #1337 landed. But it fails with this error:

```
thread 'main' panicked at burn/crates/burn-core/src/record/serde/de.rs:325:9:
not implemented: deserialize_enum is not implemented
```

Straight from `unimplemented!("deserialize_enum is not implemented")` in [de.rs](https://github.com/tracel-ai/burn/blob/main/crates/burn-core/src/record/serde/de.rs)

Would be nice to add support to deserialize with enums now that they are supported for modules.

**MWE**

```py
import torch
from torch import nn, Tensor

class DwsConv(nn.Module):
    """Depthwise separable convolution."""

    def __init__(self, in_channels: int, out_channels: int, kernel_size: int) -> None:
        super().__init__()
        # Depthwise conv
        self.dconv = nn.Conv2d(in_channels, in_channels, kernel_size, groups=in_channels)
        # Pointwise conv
        self.pconv = nn.Conv2d(in_channels, out_channels, kernel_size=1, groups=1)

    def forward(self, x: Tensor) -> Tensor:
        x = self.dconv(x)
        return self.pconv(x)


class Net(nn.Module):
    def __init__(self, depthwise: bool = False) -> None:
        super().__init__()
        self.conv = DwsConv(3, 64, 3) if depthwise else nn.Conv2d(3, 64, 3)


if __name__ == "__main__":
    torch.manual_seed(42)  # To make it reproducible
    model = Net(depthwise=True)
    model_weights = model.state_dict()
    torch.save(model_weights, "conv-dws.pt")
```

```rust
use burn::{
    backend::{ndarray::NdArrayDevice, NdArray},
    module::Module,
    nn::conv::{Conv2d, Conv2dConfig},
    record::{FullPrecisionSettings, Recorder},
    tensor::backend::Backend,
};
use burn_import::pytorch::PyTorchFileRecorder;

#[derive(Module, Debug)]
pub enum Conv<B: Backend> {
    DwsConv(DwsConv<B>),
    Conv(Conv2d<B>),
}

#[derive(Module, Debug)]
pub struct DwsConv<B: Backend> {
    dconv: Conv2d<B>,
    pconv: Conv2d<B>,
}

#[derive(Module, Debug)]
pub struct Net<B: Backend> {
    conv: Conv<B>,
}

fn main() {
    let device = NdArrayDevice::default();
    let model: Net<NdArray> = Net {
        conv: Conv::Conv(Conv2dConfig::new([3, 64], [3, 3]).init(&device)),
    };

    let record = PyTorchFileRecorder::<FullPrecisionSettings>::new()
        .load("conv-dws.pt".into(), &device)
        .unwrap();

    model.load_record(record);
}
```

antimora:
### Checklist

### Related Issues/PRs

Fixes #1431

### Changes

1. Updated Deserializer logic for deserialize_enum
2. Added tests
3. Updated the book section

### Testing

1. New unit tests
2. Rendered the book


burn-book/src/import/pytorch-model.md
```diff
@@ -347,6 +347,58 @@ let record = PyTorchFileRecorder::<FullPrecisionSettings>::default()
     .expect("Should decode state successfully")
 ```
 
+### Models containing enum modules
+
+Burn supports models containing enum modules with new-type variants (tuple with one item). Importing
+weights for such models is supported by using the `PyTorchFileRecorder` automatically. However, it
```
laggui, line 353: Minor syntax reword:

Importing weights for such models is automatically supported by the `PyTorchFileRecorder`. However, ...

burn-book/src/import/pytorch-model.md
```diff
@@ -347,6 +347,58 @@ let record = PyTorchFileRecorder::<FullPrecisionSettings>::default()
...
+Dtype: F32
+---
+Key: conv.pconv.weight
+Shape: [2, 2, 1, 1]
+Dtype: F32
+```
+
+The weights will be imported into the `DwsConv` variant of the `Conv` enum module.
+
+If the variant types are identica, then the first variant is picked. Generally, it won't be a
```
laggui, line 399: ~~identica~~ identical

laggui (approved):
LGTM! Just two minor typos/wording issue but approving in advance.

Thanks for working on this 🙂

PR #1437, opened 2024-03-09
[CI] Use GitHub's own action to generate GitHub App token

syl20bnr:
Attempt to fix the token creation error by switching to GitHub's own action.


Luni-4 (approved):
Some tests are failing but this PR does not seem to be the cause

nathanielsimard (approved):

Issue #522, opened 2023-07-24
`sign` tensor operator for copying sign of one tensor onto another


wbrickner:
I need to take tensor `a`, and apply its sign of all of its elements to the elements of tensor `b`.

There seems to be no autodiff compatible way to do this, although all backends support it AFAIK.

Thank you

antimora:
We are crucially missing two tensor operations for this: `sign()` and `abs()`. If we had them, then you could do this:

```rust
# Get the sign of tensor_a
sign_a = tensor_a.sign()

# Make a new tensor that is the absolute value of tensor_b, multiplied by the sign of tensor_a
output_tensor = tensor_b.abs() * sign_a
```

For missing ops there are workarounds with the existing ops, which I can provide tomorrow after verifying they work on my computer. Stay tuned. 

I'll leave this issue open and convert to a feature request for `abs` and `sign`

wbrickner:
For now I have done the (very silly):

```rust
let sign = t.clone() / (t.clone() * t.clone()).sqrt();
```

although this is not guaranteed to yield elements in `{-1, +1}`.
What is the blocker for adding these operations to `Tensor`?

Also, I only care about the `tch` backend, so is there some way to specialize my code and just use tensor ops from `tch`, while maintaining compatibility with the `burn` autodiff / ml system?

nathanielsimard:
There is no example on how to add code specialized for a backend, but it is possible, you have to create functions not generic over the backend and you should be fine. The best way to do it is to create another trait `MyBackend: Backend + MyAdditionalFunctions` where you can implement the trait `MyAdditionalFunctions` for the backend you want to use. To support autodiff, you would need to implement the function for two backends: `TchBackend` and `ADBackendDecorator<TchBackend>` for autodiff. For now I would not recommend to do that, since it would probably be faster to add the ops to burn, and the lack of documentation/example can slow you down.

nathanielsimard:
There is no example on how to add code specialized for a backend, but it is possible. The best way to do it is to create another trait `MyBackend: Backend + MyAdditionalFunctions` where you can implement the trait `MyAdditionalFunctions` for the backend you want to use. To support autodiff, you would need to implement the function for two backends: `TchBackend` and `ADBackendDecorator<TchBackend>`. For now, I would not recommend to do that, since it would probably be faster to add the ops to burn, and the lack of documentation/example can slow you down.

The sign function can simply be the following:

```rust
[...]

PR #1444, opened 2024-03-10
Implement Huber loss

### LINKED ISSUE #1441: Problematic derivative of Tensor::abs and Huber loss
Implementing the Huber loss requires comparing the absolute value against some small `kappa`, then behaving linearly with the absolute value outside this bound, but quadratic inside. An example implementation would be:

```rust
fn huber_loss<const D: usize, B: Backend>(kappa: f32, delta: Tensor<B, D>) -> Tensor<B, D> {
    let u = delta.clone().abs();
    let is_small = u.clone().lower_elem(kappa);
    let inside = delta.powf_scalar(2.).mul_scalar(0.5);
    let outside = u.mul_scalar(kappa).sub_scalar( (kappa * kappa) / 2. );
    outside.mask_where(is_small, inside)
}
```

Using this implementation leads to `NaN` values showing up after a few steps of training. This is probably connected to the gradient computation in `Tensor::abs` but as the problematic small values should be masked out and instead use the (perfectly fine) gradient computation from the quadratic term, I'm not sure how that is happening.

WorldSEnder:
### Checklist

### Related Issues/PRs

Closes #1441 as I think the remaining feature request for a sign function is already tracked in #522.

### Changes

Implements the [Huber Loss function](https://en.wikipedia.org/wiki/Huber_loss).

Instead of strictly following the definition of using a sign or abs function, the implementation uses clamping, which computes the same value outside the `delta` bounds but is better behaved on the autodiff backend and does not need any extra primitive ops. See also #1441 for my first attempt of implementing this.

### Testing

Test data should cover all relevant branches of the operation, and critical points on the autodiff backend, i.e. zero residuals and the point where the loss switches between the branches. Test assertions have been generated from executing the equivalent in scipy.

Note: the `test_downsample_interpolation` test in `nearest_interpolate.rs` is failing locally for me. Not caused by the patch, I've ignored it when running run-checks.

### DIFF
--- a/burn-book/src/building-blocks/module.md
+++ b/burn-book/src/building-blocks/module.md
@@ -162,3 +162,4 @@ Burn comes with built-in modules that you can use to build your own modules.
 | ------------------ | --------------------- |
 | `CrossEntropyLoss` | `nn.CrossEntropyLoss` |
 | `MseLoss`          | `nn.MSELoss`          |
+| `HuberLoss`        | `nn.HuberLoss`        |
--- a/crates/burn-core/src/nn/loss/huber.rs
+++ b/crates/burn-core/src/nn/loss/huber.rs
@@ -1,0 +1,179 @@
+use crate as burn;
+
+use crate::{config::Config, module::Module};
+use burn_tensor::backend::Backend;
+use burn_tensor::Tensor;
+use core::marker::PhantomData;
+
+use super::Reduction;
+
+/// Configuration to create a [Huber loss](HuberLoss).
+#[derive(Config, Debug)]
+pub struct HuberLossConfig {
+    /// The bound where the Huber loss function changes from quadratic to linear behaviour.
+    pub delta: f32,
+}
+
+impl HuberLossConfig {
+    /// Initialize [Huber loss](HuberLoss).
+    pub fn init<B: Backend>(&self, device: &B::Device) -> HuberLoss<B> {
+        // device is not needed as of now, but we might want to prepare some data on it
+        // and its consistent with other loss functions
+        let _ = device;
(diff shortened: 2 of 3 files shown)


antimora:
Submitted a PR for `sign` tensor operator: https://github.com/tracel-ai/burn/pull/1446

WorldSEnder:
Note: I think the method of `clamp`ing even works out better for Huber than using `sign`, since it avoids a `mul_scalar(delta)`. Though I'm still interested in `sign` for alternative losses such as SmoothL1.

antimora:
CI failed due to:

```
 test tests::jit_fusion::var::tests::test_var_mean_bias ... ok
  
  failures:
  
  ---- tests::jit::kernel::normal::tests::subsequent_calls_give_different_tensors stdout ----
  thread 'tests::jit::kernel::normal::tests::subsequent_calls_give_different_tensors' panicked at crates/burn-wgpu/src/lib.rs:73:5:
  assertion failed: tensor_1.to_data().value[i] != tensor_2.to_data().value[i]
  
  
  failures:
      tests::jit::kernel::normal::tests::subsequent_calls_give_different_tensors
  
  test result: FAILED. 1341 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 55.86s
```

Rerunning to see if it's a fluke. Tagging @nathanielsimard and @louisfd since they're currently working in this area.

antimora:
Sign tensor op PR is merged.

louisfd (approved):
LGTM, but can you clarify the math in comments? I think it works out but it's a bit confusing, in particular I think `r`, `err` and `res` are three names for the same thing?

PR #1457, opened 2024-03-11
Migrate/jit/cat

louisfd:
Cat kernel for #1422 
Note: it was chosen to rely on a default implementation based on slice assign instead of making a kernel which would have the same performance anyway because of the dynamic number of inputs.


crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -1046,17 +1046,49 @@ pub trait FloatTensorOps<B: Backend> {
...
+        for tensor in tensors {
+            let mut indices = indices_select_all.clone();
+            let tensor_dim_length = B::float_shape(&tensor).dims[dim];
+            indices[dim] = output_index..tensor_dim_length;
+            output_index += tensor_dim_length;
+
+            tensor_output = B::float_slice_assign(tensor_output, indices, tensor)
+        }
+
+        tensor_output
```
nathanielsimard, line 1090: I think it would be better to implement this function in a specific module that work with the `Tensor` struct instead of the backend API. Here we could simply:

```rust
fn float_cat<const D: usize>(tensors: Vec<FloatTensor<B, D>>, dim: usize) -> FloatTensor<B, D> {

nathanielsimard (commented):
See me comment

crates/burn-tensor/src/tensor/ops/tensor.rs
louisfd: I tried, see #1473

nathanielsimard (approved):

PR #1467, opened 2024-03-13
Feat: Add Leaky Relu Model

Arjun31415:
### Checklist

### Changes

Added Leaky Relu Activation Layer. 
It internally just calls the functional prelu because they are almost the same.


antimora (changes requested):
Great. Thanks for the addition. A couple of changes needed:

1. Tests
2. Updating the book section with a new entry: https://burn.dev/book/building-blocks/module.html#general

Arjun31415:
> Great. Thanks for the addition. A couple of changes needed:
> 
>     1. Tests
> 
>     2. Updating the book section with a new entry: https://burn.dev/book/building-blocks/module.html#general

How do I test this ?. I have not added any new functionality,just calling the existing function within a Module

antimora:
> > Great. Thanks for the addition. A couple of changes needed:
> > ```
> > 1. Tests
> > 
> > 2. Updating the book section with a new entry: https://burn.dev/book/building-blocks/module.html#general
> > ```
> 

> How do I test this ?

The test should go here: https://github.com/tracel-ai/burn/tree/main/crates/burn-core/tests

> . I have not added any new functionality,just calling the existing function within a Module. I have noticed that there are no tests for GELU and RELU layers too, maybe because their functional code is being tested and covered. I do not know what to test here which already does not have a test written for it.

You're right it's missing for GELU and RELU but I believe we should have had them. Maybe it was justified because they're just modules on existing functions. `LeakyRelu` contains enough logic that needs verification. You must have tested to be confident it works, right? So we need the same tests checked in. This also helps a reviewer to see if all edge cases are covered.

antimora:
Linking an existing open PR for the same functionality: https://github.com/tracel-ai/burn/pull/1208

I am not sure why it was stalled previously. BTW, it's okay not to do the ONNX part. 

It contains tests you might want to reuse. Also it appears an activation function was added. I am not sure if this needs to be added in this case. I am tagging @nathanielsimard and @ashdtu

crates/burn-core/src/nn/leaky_relu.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Parametric Relu layer.
+#[derive(Module, Debug)]
+pub struct LeakyRelu<B: Backend> {
+    /// The weight used in Leaky ReLu
+    pub negative_slope: Tensor<B, 1>,
+}
+/// Configuration to create a [Leaky Relu](LeakyRelu) layer.
+#[derive(Config, Debug)]
+pub struct LeakyReluConfig {
+    /// The learnable weight alpha. Default is 0.01
```
nathanielsimard, line 16: It isn't learnable right?

Arjun31415: Yeah non learnable. That was left by mistake

antimora (approved):
Thank you very much the addition. It looks good.

PR #1472, opened 2024-03-14
Migrate/jit/matmul tiling 2d

louisfd:
Matmul tiling 2d kernels for #1422 

Bonus stuff: 
- macro for configurable unrolling of loops
- tiling 2d (unpadded) will automatically use the kernel with no bound checks when shapes are divisible by block sizes


crates/burn-jit/src/codegen/dialect/gpu/operation.rs
```diff
@@ -58,6 +59,7 @@ pub enum Operator {
     BitwiseXor(BinaryOperator),
     ShiftLeft(BinaryOperator),
     ShiftRight(BinaryOperator),
+    AssignVec4(AssignVec4Operator),
```
nathanielsimard, line 62: Should be assign

crates/burn-jit/src/kernel/matmul/tiling2d_shader.rs
```diff
@@ -0,0 +1,636 @@
...
+                                    gpu!(scope, remain_n = remain == 1u32);
+                                    gpu!(scope, if(remain_n).then(|scope|{
+                                        gpu!(scope, val_0 = input[position_0]);
+                                    }));
+                                }));
+                            }));
+                        }));
+
+                        let val_vec4 = scope.create_local(shared_memory.item());
+                        gpu!(scope, val_vec4 = vec4(val_0, val_1, val_2, val_3));
```
nathanielsimard, line 368: We should:

```rust
val_vec4[0] = val_0;
val_vec4[1] = val_1;
val_vec4[2] = val_2;
val_vec4[3] = val_3;
```

But we could do that in the macro.

crates/burn-jit/src/kernel/matmul/tiling2d_shader.rs
```diff
@@ -0,0 +1,636 @@
...
+        // Batch offset for the lhs & rhs matrices.
+        gpu!(scope, batch_dims = rank - 2u32);
+        gpu!(
+            scope,
+            range(0u32, batch_dims).for_each(|b, scope| {
+                let stride_lhs = scope.create_local(Elem::UInt);
+                let stride_rhs = scope.create_local(Elem::UInt);
+                let stride_output = scope.create_local(Elem::UInt);
+                let shape_lhs = scope.create_local(Elem::UInt);
+                let shape_rhs = scope.create_local(Elem::UInt);
```
nathanielsimard, line 105: Put all variable declaration outside of the loop.

crates/burn-jit/src/kernel/matmul/tiling2d_shader.rs
```diff
@@ -0,0 +1,636 @@
...
+
+        // Calculate offset.
+        let offset_lhs = scope.create_local(Elem::UInt);
+        let offset_rhs = scope.create_local(Elem::UInt);
+        gpu!(scope, offset_lhs = skip_row * lhs_stride_row);
+        gpu!(scope, offset_rhs = skip_col * rhs_stride_col);
+
+        // Batch offset for the output.
+        let offset_output = scope.create_local(Elem::UInt);
+        let batch_dims = scope.create_local(Elem::UInt);
```
nathanielsimard, line 92: I would create a function with almost everything at the beginning of the shader that returns a struct with all matrix multiply information required.

1. Gather the kernel information (offset, k, m, n, etc.)
2. Read the lhs & rhs tensor and fill the shared memory, then return a struct with the result.
3. Perform the matmul calculation and fill the output shared memory, then return a struct with the result.
4. Write the output shared memory to the output.

crates/burn-jit/src/kernel/matmul/tiling2d_shader.rs
```diff
@@ -0,0 +1,636 @@
...
+
+        gpu!(
+            scope,
+            range(0u32, n_loops).for_each(|i, scope| {
+                // Equivalent of looping from 0 to K with steps block_size_k
+                let k = scope.create_local(Elem::UInt);
+                gpu!(scope, k = i * block_size_k);
+
+                match self.assumption {
+                    Tiling2DAssumption::Round => {
```
nathanielsimard, line 167: bound_checks_required

crates/burn-jit/src/kernel/matmul/tiling2d_shader.rs
```diff
@@ -0,0 +1,636 @@
...
+                            scope,
+                            k,
+                            thread_col,
+                            thread_row,
+                            lhs_stride_col,
+                            lhs_stride_row,
+                            lhs,
+                            offset_lhs,
+                            shared_lhs,
+                            true,
```
nathanielsimard, line 179: Could be an enum

crates/burn-jit/src/kernel/matmul/tiling2d_shader/load_shared_memory.rs
```diff
@@ -0,0 +1,271 @@
...
+                    let position_0 = scope.create_local(Elem::UInt);
+                    let position_1 = scope.create_local(Elem::UInt);
+                    let position_2 = scope.create_local(Elem::UInt);
+                    let position_3 = scope.create_local(Elem::UInt);
+                    let remain_n = scope.create_local(Elem::Bool);
+
+                    let val_0 = scope.zero(elem);
+                    let val_1 = scope.zero(elem);
+                    let val_2 = scope.zero(elem);
+                    let val_3 = scope.zero(elem);
```
nathanielsimard, line 127: Can we extract those else where, the declaration is probably done in a for loop. Also, not sure we need to initialize val_0, val_1, val_2 and val_3 to zero.

nathanielsimard (commented):
Only one comment, otherwise looks good

PR #1474, opened 2024-03-14
Migrate/jit/prod

### LINKED ISSUE #1461: Implement `prod` and `prod_dim` for WGPU backend
Currently default unoptimized implementation is used for float tensor. To use with int tensor, one should convert to float (`tensor.float()`) and use `prod` or `prod_dim` methods.

louisfd:
Prod and prod_dim kernels for #1422 
Fix #1461 

There was also a bug in jit uniform prng with ints, which is fixed here.


antimora (approved):
LGTM. Thank you.

nathanielsimard (approved):

PR #1475, opened 2024-03-15
[RFC] Support switching burn-wgpu between WebGPU implementations (wgpu <-> Dawn)

p1-0tr:
### Checklist

### Related Issues/PRs

-

### Changes

Currently Burn can only be built against the `wgpu` WebGPU implementation. I found the ability to switch to Dawn useful, e.g. to get float16 support, which helps when trying to run Llama2 on top of Burn.

I've:
* split out any uses of `wgpu` into a separate module,
* added Dawn as a submodule,
* modified the build system to be able to build, generate bindings for, and link against Dawn,
* added a module which wraps the usage of the Dawn bindings to make it look more like `wgpu`,
* added explicit bounds checks for i/o arrays (otherwise Dawn will happily do out of bounds accesses),
* added support for `f16`.

I'd be keen to get your thoughts on my changes. The part that worries me the most is how much adding Dawn as a dependency complicates the build.

### Testing

I've ran `burn-wgpu` unit tests, with `wgpu` and `dawn` features enabled. A couple of cases fail with the `dawn` backend because of limits on the amount of storage buffers.


nathanielsimard (commented):
Thanks a lot for bringing that up! I think we should split the PR into multiple smaller work items so that we can review and merge parts of the solution faster.

- [ ] Add half precision support to the jit backend (f16 & bf16).
- [ ] Add the Dawn backend, (so we could activate wgpu and dawn feature flag without conflict)

As for the bound checks, I'm also writing a CUDA backend right now and  they  arent' done automaticaly like with wgpu, however I don't think we should always do them, since it increase branching, especially when a kernel will do a ton of writes on the same buffer in a loop (reduces, matmul, etc.). I added early return code in the jit codegen, so I would skip that part (for now) when integrating Dawn.

p1-0tr:
@nathanielsimard - Thanks for your feedback. The PR is split into 4 patches already, happy to submit those as separate PRs :)

>     * [ ]  Add half precision support to the jit backend (f16 & bf16).

I didn't submit this separately to begin with, because you won't be able to test it without Dawn for now. Basically `naga` will come back with an error saying it does not know what `enable f16;` means. But, yeah, I can see how this will make review easier.
 
>    ... so we could activate wgpu and dawn feature flag without conflict)

Just to be sure - you mean that making `wgpu`/`dawn` switching possible at runtime, would be preferred over it being a purely compile time choice?

> As for the bound checks, I'm also writing a CUDA backend right now and they arent' done automaticaly like with wgpu, however I don't think we should always do them, since it increase branching, especially when a kernel will do a ton of writes on the same buffer in a loop (reduces, matmul, etc.). I added early return code in the jit codegen, so I would skip that part (for now) when integrating Dawn.

👍 cool, I'll make emitting those optional, and only happen when Dawn is chosen as the backend.

errordeveloper (commented):
You don't need to do this if you have

```
# ~/.gitconfig

[core]
        excludesfile = <homeDir>/.gitignore_global
```

And in `<homeDir>/.gitignore_global` you can have a list of files for your favourite editor/IDE.

nathanielsimard:
> 👍 cool, I'll make emitting those optional, and only happen when Dawn is chosen as the backend.

I would just not make the changes, even if the Dawn backend can't be used right now. We won't add bound checks because of Dawn or CUDA; we will early return shaders based on the grid and the tensor shapes involved, so not a simple bound checks in all shaders. Since we are working on migrating all WGSL shaders to our JIT IR, it's not wise to work on it at the same time (we want to take the time to do it correctly).

nathanielsimard:
> Just to be sure - you mean that making `wgpu`/`dawn` switching possible at runtime, would be preferred over it being a purely compile time choice?

I simply prefer following the recommendations from the Rust team, which is that all feature flags are additive, not exclusive. So when the Dawn feature flag is activated, we should also export a backend named Dawn:

```rust
type MyBackend = burn::backend::Dawn;
// But we can also import the other backend if activated.
type MyOtherBackennd = burn::backend::Wgpu;
```

nathanielsimard:
> I didn't submit this separately to begin with, because you won't be able to test it without Dawn for now. Basically `naga` will come back with an error saying it does not know what `enable f16;` means. But, yeah, I can see how this will make review easier.

Yeah I understand, which is why starting with Dawn makes a lot of sense, but the features doesn't have to be activated to be merged, and I would simply not modify any wgsl shaders for now, since we are migrating all of them.

So as a summary: it would be very much appreciated if you can make a PR that introduce the `Dawn` backend with a feature flag that can be used along the `Wgpu` backend. Of course it won't be usable right away because of the bound-checks, but it can be merged really fast. Then you can submit a PR that adds half precision element types to the backend, which will work with `Dawn` but not `Wgpu`. We will work on fixing the bound checks, and voilà we will have `Dawn` and `Cuda` as new available backends 🎉

p1-0tr:
Sounds like a plan. Thanks :)

antimora:
@p1-0tr do you want to keep this PR open taking @nathanielsimard suggestion to open new separate PRs?

p1-0tr:
> @p1-0tr do you want to keep this PR open taking @nathanielsimard suggestion to open new separate PRs?

@antimora Happy to close it :) I was going to close it when I open the new ones (should be doing that this week).

PR #1476, opened 2024-03-15
Disable default features of burn-wgpu for benchmarking

errordeveloper:
Before this change benchmarks with `--backend wgpu` as well as
`--backend wgpu-fusion` showed `"backend": "fusion<jit<wgpu>>"` in
 the JSON report, so fusion was always enabled.
    
Before:
```
    $ grep backend ~/.cache/burn/backend-comparison/*
    ~/.cache/burn/backend-comparison/bench_binary_1710330455200.json:  "backend": "fusion<jit<wgpu>>",
    ~/.cache/burn/backend-comparison/bench_binary_1710331696932.json:  "backend": "fusion<jit<wgpu>>",
 ```
 
  After:
 ```
    $ grep backend ~/.cache/burn/backend-comparison/*
    ~/.cache/burn/backend-comparison/bench_binary_1710499297627.json:  "backend": "jit<wgpu>",
    ~/.cache/burn/backend-comparison/bench_binary_1710499307181.json:  "backend": "fusion<jit<wgpu>>",
```


syl20bnr:
Thank you for this fix ! It seems that doing that breaks the compilation when not using the workspace though. Can you try to build the crate with `cargo build -p backend-comparison` ?

errordeveloper (commented):
> Can you try to build the crate with `cargo build -p backend-comparison` ?  

Yeah, the breakage is quite odd. I'll try to dig in.

syl20bnr:
@nathanielsimard what do we want to do here ? Should we enable benchmarking without fusion enabled or is it not worth it because fusion will always be activated in practical cases ?

nathanielsimard:
We need to deactivate fusion by default, otherwise it's very hard to compare the benefits of our optimizations.

errordeveloper:
Now that I'm looking at this, it seems the issue I was seeing earlier was already addressed with some recent changes. I can see the bug exists on 261e7eca1d2f36cf70495c08274f002898913ab8,  but it was fixed already before  b429cc39c19780e561b697f9b970c04c29969fa5. I tried bisecting it, but not sure if there is a point to finding out what fixed it, not for me anyway, the fix appears to be withing a9d741ed..d43a0b3f90ee0eb9ce7978e5d4ad9208a645a43a.

errordeveloper:
Also, I can see there is lots of vomits with improvements to burnbench and the output is much nicer now!

Issue #1421, opened 2024-03-06
Add `topk` operation.


zemelLeong:
https://pytorch.org/docs/stable/generated/torch.topk.html

louisfd:
Somewhat related: #1369

PR #1488, opened 2024-03-18
Add tensor sorting operations

### LINKED ISSUE #1369: Add `argsort` tensor op
Maybe also `argsort_descending` for reverse order?

**Example**
```rust
let x = Tensor::from_data([1, 0, 3, 4, -1], &device)
x.argsort(0) // argsort(dim)
// [4, 1, 0, 2, 3]
```

laggui:
### Checklist

### Related Issues/PRs

Closes #1369
Unblocks easy implementation for `topk` issue #1421

### Changes

Added `sort(dim)`, `sort_with_indices(dim)` and `argsort(dim)` numeric tensor ops.
Also added `sort_descending(dim)`, `sort_descending_with_indices(dim)` and `argsort_descending(dim)` for sorting in descending order.

- Added `ElementComparison` trait for `Element` so elements can return an ordering (as required by `sort` methods)
- Implemented default tensor ops for all backends
- Implemented for `tch` backend
- Added autodiff backward

**TODO**
- [ ] Parallelize default implementation loop with rayon
  - *Note: we have `run_par`, `iter_par` and `iter_range_par` macros defined in burn-ndarray*

### Testing

Forward and backward unit tests.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/float.rs
+++ b/crates/burn-tensor/src/tensor/api/float.rs
@@ -270,7 +270,7 @@ where
     /// This sort is unstable (i.e., may reorder equal elements).
     #[cfg(all(not(feature = "wasm-sync"), target_family = "wasm"))]
     pub async fn sort(self, dim: usize) -> Tensor<B, D> {
-        Tensor::new(sort::<B, D, Float>(self.primitive, dim).await)
+        Tensor::new(sort::<B, D, Float>(self.primitive, dim, /*descending*/ false).await)
     }
 
     /// Sort the elements by value in ascending order along a given dimension.
@@ -280,7 +280,8 @@ where
     #[cfg(all(not(feature = "wasm-sync"), target_family = "wasm"))]
     pub async fn sort_with_indices(self, dim: usize) -> (Tensor<B, D>, Tensor<B, D, Int>) {
         check!(TensorCheck::sort_dim::<D>("Sort_with_indices", dim));
-        let (values, indices) = sort_with_indices::<B, D, Float>(self.primitive, dim).await;
+        let (values, indices) =
+            sort_with_indices::<B, D, Float>(self.primitive, dim, /*descending*/ false).await;
         (Tensor::new(values), Tensor::new(indices))
     }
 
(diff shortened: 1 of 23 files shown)


crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -261,4 +264,33 @@ where
...
+        let (values, indices) =
+            sort_with_indices::<B, D, Float>(self.primitive, dim, /*descending*/ false).await;
+        (Tensor::new(values), Tensor::new(indices))
+    }
+
+    /// Returns the indices that sort the elements by value in ascending order along a given dimension.
+    ///
+    /// This sort is unstable (i.e., may reorder equal elements).
+    #[cfg(all(not(feature = "wasm-sync"), target_family = "wasm"))]
+    pub async fn argsort(self, dim: usize) -> Tensor<B, D, Int> {
```
nathanielsimard, line 292: Could we put those into `numeric` to avoid duplication? With newer version of Rust, we can add `async` in traits.

nathanielsimard: Also, what is the reason we need an async API here?

crates/burn-tensor/src/tensor/element.rs
```diff
@@ -55,6 +58,12 @@ pub trait ElementRandom {
         Self: Sized;
 }
 
+/// Element ordering trait.
+pub trait ElementComparison {
+    /// Returns and [Ordering] between `self` and `other`.
+    fn cmp(&self, other: &Self) -> Ordering;
+}
```
nathanielsimard, line 65: Perfect 🔥

crates/burn-tensor/src/tensor/api/float.rs
laggui: Oh I guess I should update my rust version, it wouldn't let me do it initially.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -1852,6 +1911,93 @@ where
         distribution: Distribution,
         device: &B::Device,
     ) -> Self::Primitive<D>;
+
+    /// Sort the elements of the input `tensor` by value in ascending order along a given dimension.
```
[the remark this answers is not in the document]
laggui: Forgot to clean up the comments after adding descending! Good catch :)

crates/burn-tensor/src/tensor/api/float.rs
laggui: Nevermind it was a warning not an error 🤦 

```
use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
```

laggui: Will leave it like this for now to avoid the warnings. This will be temporary anyways with #1490.

nathanielsimard (approved):

PR #1491, opened 2024-03-19
Fix jit reduce tests

laggui:
As discussed, changed assert_approx_eq precision from 3 to 2 for all reduce tests.


louisfd (approved):
Awesome, thanks

PR #1505, opened 2024-03-22
[Breaking] add runtime options in wgpu init methods

nathanielsimard:
Add the ability to control the Runtime options in both wasm and normal Rust.


crates/burn-wgpu/src/runtime.rs
```diff
@@ -51,16 +51,52 @@ impl<G: GraphicsApi, F: FloatElement, I: IntElement> Runtime for WgpuRuntime<G,
...
+    pub dealloc_strategy: DeallocStrategy,
+    /// Control the slicing strategy.
+    pub slice_strategy: SliceStrategy,
+    /// Control the amount of compute tasks to be aggregated into a single GPU command.
+    pub max_tasks: usize,
+}
+
+impl Default for RuntimeOptions {
+    fn default() -> Self {
+        let max_tasks = match std::env::var("BURN_WGPU_MAX_TASKS") {
```
antimora, line 66: Can we build conditionally initialize this environment variable depending it's wasm or not?

nathanielsimard: no_std isn't supported for wgpu, but this will work with wasm.

antimora: Got it. Feels quick and dirty though =D

I'll approve it for now.

antimora (approved):

crates/burn-wgpu/src/runtime.rs
nathanielsimard: This option might even go away in future releases

PR #1508, opened 2024-03-22
Tensor expand operator

### LINKED ISSUE #1412: Add `expand` tensor operation similar to PyTorch
### Summary

I am proposing the addition of an `expand` tensor operation to the Burn deep learning framework, mirroring the functionality available in PyTorch. This operation is invaluable for broadcasting operations on tensors of different shapes, making certain types of tensor manipulations more intuitive and efficient.

### Motivation

In PyTorch, the `expand` method allows users to expand the size of an existing tensor without copying data. This is particularly useful for performing operations that require tensors to be of the same shape, enabling more memory-efficient computations by avoiding the creation of large intermediate tensors. Its inclusion in Burn would greatly enhance the framework's tensor manipulation capabilities, aligning it more closely with PyTorch's feature set and making it more appealing to users transitioning between frameworks or leveraging both for different aspects of their work.

### Technical Details

The `expand` function in PyTorch allows a tensor to be "viewed" with a larger size, with the new dimensions being broadcastable according to the broadcasting rules of tensors. The operation does not allocate new memory for the expanded tensor but provides a new view on the existing tensor, which is a very efficient way to perform many element-wise operations.

I propose that the API for the `expand` operation in Burn should closely mirror that of PyTorch for ease of use and familiarity. 

### Use Cases

This feature is crucial for operations where matching tensor dimensions is necessary without the need to duplicate data. For example, when adding a vector to each row of a matrix, or applying a mask across multiple dimensions of a tensor where the mask is initially only defined in a subset of those dimensions.

### Example

In PyTorch, the usage of `expand` is as follows:

```python
import torch
x = torch.tensor([1, 2, 3])
x_expanded = x.expand(3, 3)
# x_expanded is now [[1, 2, 3], [1, 2, 3], [1, 2, 3]] without copying data
```

A similar functionality in Burn would enable users to perform such operations with the same ease and efficiency.

### Conclusion

The addition of an `expand` tensor operation to Burn could significantly improve its usability and efficiency for a wide range of tasks, making it a more competitive and versatile tool in the deep learning ecosystem. I believe this feature would be a valuable asset to the framework and its users.

antimora:
### Checklist

### Related Issues/PRs

fixes #1412

### Changes

1. All backends
2. Fix issue with WGPU when stride == 0 
3. Add new tests 
4. Update docs

### Testing

New unit tests


crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -2437,6 +2437,81 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+        impl<B: Backend, const D1: usize, const D2: usize> Backward<B, D2, 1> for BroadcastDim<D1, D2> {
+            type State = (Shape<D1>, Shape<D2>);
+
+            fn backward(
+                self,
+                ops: Ops<Self::State, 1>,
+                grads: &mut Gradients,
+                _checkpointer: &mut Checkpointer,
+            ) {
+                let (shape_original, _shape_new) = ops.state;
```
nathanielsimard, line 2471: If you don't need `_shape_new`, then I think we should remove it from the state.

crates/burn-fusion/src/ops/boolean.rs
```diff
@@ -467,6 +468,44 @@ impl<B: FusionBackend> BoolTensorOps<Self> for Fusion<B> {
...
+
+        let desc = ExpandOperationDescription {
+            input: tensor.into_description(),
+            shape: shape.dims.into(),
+            out: out.to_description_out(),
+        };
+
+        out.client.register(
+            vec![stream],
+            OperationDescription::BaseInt(BaseOperationDescription::Expand(desc.clone())),
```
nathanielsimard, line 502: `BaseInt` => `BaseBool` I think.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -2437,6 +2437,81 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
             .stateless(B::float_sign(tensor.primitive))
     }
 
+    fn float_expand<const D1: usize, const D2: usize>(
+        tensor: FloatTensor<Self, D1>,
+        shape: Shape<D2>,
+    ) -> FloatTensor<Self, D2> {
+        #[derive(Debug)]
+        struct BroadcastDim<const D1: usize, const D2: usize>;
```
louisfd, line 2445: Why does the name switch to broadcast here?

crates/burn-fusion/src/ops/float.rs
```diff
@@ -1847,6 +1848,44 @@ impl<B: FusionBackend> FloatTensorOps<Self> for Fusion<B> {
...
+
+        let desc = ExpandOperationDescription {
+            input: tensor.into_description(),
+            shape: shape.dims.into(),
+            out: out.to_description_out(),
+        };
+
+        out.client.register(
+            vec![stream],
+            OperationDescription::BaseInt(BaseOperationDescription::Expand(desc.clone())),
```
nathanielsimard, line 1882: Same comment

crates/burn-fusion/src/ops/boolean.rs
```diff
@@ -467,6 +468,44 @@ impl<B: FusionBackend> BoolTensorOps<Self> for Fusion<B> {
         out
     }
 
+    fn bool_expand<const D1: usize, const D2: usize>(
+        tensor: BoolTensor<Self, D1>,
+        shape: Shape<D2>,
+    ) -> BoolTensor<Self, D2> {
+        #[derive(new)]
+        struct BroadcastToOps<const D: usize, const D2: usize> {
```
nathanielsimard, line 476: Should be ExpandOps

crates/burn-jit/src/ops/base.rs
```diff
@@ -80,6 +82,54 @@ pub(crate) fn permute<R: Runtime, E: JitElement, const D: usize>(
...
+                    };
+                } else {
+                    // Error handling: Dimension mismatch for broadcasting
+                    panic!(
+                        "Dimension mismatch: cannot broadcast dimension {} of tensor to target shape",
+                        tensor_dim
+                    );
+                }
+            } else {
+                // If the input tensor has fewer dimensions, treat missing dimensions as 1
```
nathanielsimard, line 115: Wrong comment: `treat missing dimensions as 1` => `treat missing dimensions as 0`

crates/burn-autodiff/src/ops/tensor.rs
louisfd, line 2471: If you don't need _shape_new then the associated type State can be only Shape<D1> and you don't have to save it at line 2508

crates/burn-tensor/src/tensor/api/check.rs
```diff
@@ -889,6 +889,57 @@ impl TensorCheck {
             false => self,
         }
     }
+    /// Checks if shapes are compatible for element-wise operations supporting broadcasting.
+    ///
+    /// Also checks if there is -1 in the shape which
```
nathanielsimard, line 894: Is the sentence completed?

nathanielsimard (changes requested):

louisfd (changes requested):
Small things related to autodiff, otherwise looks good!

crates/burn-autodiff/src/ops/tensor.rs
antimora: I should change it. The OP was named broadcast_to

crates/burn-jit/src/ops/base.rs
antimora: It was partially correct. The stride should be set to zero but treat the missing dim as 1. I added the following:

```rust
  // If the input tensor has fewer dimensions, treat missing dimensions as 1
  // and set stride to 0 (broadcasting)
  new_strides[i] = 0;
```

crates/burn-tensor/src/tensor/api/check.rs
antimora: Nope. Fixed it.

laggui (commented):
Not much to add to the others' comments. I was also confused by the constant switch between expand <-> broadcast but it seems that it will be addressed.

To be fair, in pytorch [`tensor.expand(shape)`](https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html#torch.Tensor.expand) is equivalent to [`torch.broadcast_to(tensor, shape)`](https://pytorch.org/docs/stable/generated/torch.broadcast_to.html).

I noticed the docstrings still used "broadcast" but it makes sense since they're only meant to explain the function.

nathanielsimard (approved):

louisfd (approved):

PR #1509, opened 2024-03-22
Migrate/jit/pooling

louisfd:
Some pooling kernels for #1422


crates/burn-jit/src/kernel/pool/standard_pool2d_shader.rs
```diff
@@ -0,0 +1,243 @@
...
+use crate::{
+    codegen::{Compilation, CompilationInfo, CompilationSettings, InputInfo, OutputInfo},
+    gpu::{gpu, Elem, Item, Scope, Variable, Visibility},
+    kernel::{DynamicKernelSource, SourceTemplate},
+    Compiler, JitElement, Runtime,
+};
+
+use super::PoolStrategy;
+
+pub(crate) struct StandardPool2dComputeShader<P: PoolStrategy, R: Runtime, E: JitElement> {
```
nathanielsimard, line 12: Why Standard? Maybe add a comment listing other options (maybe future options).

crates/burn-jit/src/kernel/pool/standard_pool2d_shader.rs
```diff
@@ -0,0 +1,243 @@
...
+            gpu!(scope, ih = oh * pool_stride_0);
+            gpu!(scope, dilated = kh * dilation_0);
+            gpu!(scope, ih += dilated);
+
+            gpu!(scope, within_padding_h = ih >= padding_0);
+            gpu!(scope, tmp_padding = ih < border_bottom);
+            gpu!(scope, within_padding_h = within_padding_h && tmp_padding);
+
+            gpu!(scope, if (within_padding_h).then(|scope| {
+                (0..self.kernel_size[1]).for_each(|kw| {
```
nathanielsimard, line 131: A good optimization would be to have a propertie that checks if we need to perform those bound checks (similar to matmul) Can be in another PR.

crates/burn-jit/src/kernel/pool/standard_pool2d_shader.rs
```diff
@@ -0,0 +1,243 @@
+use std::marker::PhantomData;
+
+use crate::{
+    codegen::{Compilation, CompilationInfo, CompilationSettings, InputInfo, OutputInfo},
+    gpu::{gpu, Elem, Item, Scope, Variable, Visibility},
+    kernel::{DynamicKernelSource, SourceTemplate},
+    Compiler, JitElement, Runtime,
+};
+
+use super::PoolStrategy;
```
nathanielsimard, line 10: I guess we will have max pool and avg pool that implement this trait right?

crates/burn-jit/src/kernel/pool/standard_pool2d_shader.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        let settings = CompilationSettings::default();
+        let shader = Compilation::new(info).compile(settings);
+        let shader = <R::Compiler as Compiler>::compile(shader);
+        SourceTemplate::new(shader.to_string())
+    }
+
+    fn id(&self) -> String {
+        format!(
+            "{:?}k={:?}pool_strategy={:?}",
```
nathanielsimard, line 237: `pool_strategy => pl` we should try to minimize id length.

crates/burn-jit/src/kernel/pool/standard_pool2d_shader.rs
louisfd: What I envisaged is that there would be standard vs adaptive shaders which relate to the layout/algorithm, with a pool strategy (max/max_with_indices/avg) which relate to what the computation does. 
But since we don't have adaptive_max_pooling I did not generalize the adaptive shader to the different pool strategies. Then since we don't really have several shader strategies, I dropped the standard terminology

nathanielsimard (approved):

PR #1529, opened 2024-03-25
Feat/backend bridge

### LINKED ISSUE #1522: Refactoring mixed precision training and inference with backend bridge
- [x] Create backend bridge
- [ ] Create different element type levels that the users can modify with backend settings.

nathanielsimard:
partial fix of #1522


louisfd (changes requested):
Looks good, only some minor typos and mostly remove a file that was not supposed to be pushed. Otherwise awesome new concept

louisfd (approved):
Good to go

PR #1539, opened 2024-03-27
Feat/lazy init

### LINKED ISSUE #1329: Improve module initialization.
I think having to define both `init` and `init_with` isn't the best API, and we should try to improve it. Also, there, for now, we only have the option to set the device used for deserialization in the recorder instead of being set in the `init` function.

There are two possible options:

## New Module Initialization trait

This option is to create a trait where users only need to implement one function to initialize a module with an optional record.

```rust
pub trait ModuleInitialization<B: Backend> {
    type Module;
    
    fn create_module(&self, device: &Device<B>, record: Option<Self::Module::Record>) -> Self::Module;
    
    fn init(&self, device: &Device<B>) -> Self::Module {
        self.create_module(device, None)
    }
    
    fn init_with(&self, device: &Device<B>, record: Record) -> Self::Module {
        self.create_module(device, Some(record))
    }
}
```

Example:

```rust
pub struct Linear<B: Backend> {
    pub weight: Param<Tensor<B, 2>>,
    pub bias: Param<Tensor<B, 2>>,
}

impl<B: Backend> ModuleInitialization<B> for LinearConfig {
    type Module = Linear<B>;
    
    fn create_module(&self, device: &Device<B>, record: Option<Self::Module::Record>) -> Self::Module { 
        let initialize_new = |config: &Self| {
            nn::LinearConfig::new(config.d_model, config.d_model)
                .with_initializer(self.initializer.clone())
                .init(device)
        };

        Self {
            weight: record.weight.map(|t| t.deserialize(device)).or_else(initialize_new),
            bias: record.bias.map(|t| t.deserialize(device)).or_else(initialize_new),
        }
   }
}
```

However, this solution doesn't fix the fact that users need to call `init_with` and not sequentially call `init().load(record)` which is not ideal.

## Lazy tensor initialization

The other solution would be to have lazy tensor initialization. For this to work, no tensor would be created until a function has been called on the parameter. This only removes the need to implement `init_with` and could encapsulate the device in the parameter. For this to work we would probably need to use unsafe to implement `Send` and `Sync`, but this would be correct, since we would create a new `RefCell` or `Cell` each time a tensor is accessed from a parameter.

```rust
// Private trait to keep the same syntax as before.
trait Parameter: Clone {
    type Device;
    type Record;

    fn device(&self) -> Self::Device;
}

/// Only tensor for now implement parameter.
impl<B: Backend, K: BasicOps<B>, const D: usize> Parameter for Tensor<B, D, K> {
    type Device = B::Device;
    type Record = <Self as Module<B>>::Record;

    fn device(&self) -> Self::Device {
        self.device()
    }
}

#[derive(new)]
enum ParamState<P: Parameter> {
    Initialized(P),
    Uninitialized {
        init: Box<dyn Fn() -> P>,
        device: P::Device,
    },
}
/// Define a parameter.
pub struct Param<P: Parameter> {
    id: ParamId,
    state: RefCell<ParamState<P>>,
}

impl<P: Parameter + Clone> Param<P> {
    /// Gets the parameter value.
    ///
    /// # Returns
    ///
    /// The parameter value.
    pub fn val(&self) -> P {
        // Note: the custom clone implementation would simply call `val`.
        
        let tensor = match &*self.state.borrow() {
            ParamState::Initialized(val) => return val.clone(),
            ParamState::Uninitialized { init, device: _ } => init(),
        };

        let state_updated = RefCell::new(ParamState::Initialized(tensor.clone()));
        self.state.swap(&state_updated);

        tensor
    }

    /// This would actually be implemented in `Module::load`
    pub fn load_record(self, record: P::Record) -> Self {
        let device = match &*self.state.borrow() {
            ParamState::Initialized(val) => val.device(),
            ParamState::Uninitialized { init: _, device } => device.clone(),
        };

        let tensor = record.deserialize(device);
        let state_updated = RefCell::new(ParamState::Initialized(tensor));
        self.state.swap(&state_updated);

        self
    }
}
```

nathanielsimard:
Fix #1329


Cargo.toml
```diff
@@ -80,6 +80,9 @@ hound = "3.5.1"
 image = "0.24.9"
 zip = "0.6.6"
 
+# Once cell is useful when we need OnceCell and Lazy with no_std support.
+once_cell = { version = "1.19.0", default-features = false }
+
```
antimora, line 85: I thought OnceCell is merged into Rust now. It seems it has no_std support too: https://doc.rust-lang.org/core/cell/struct.OnceCell.html. Is there any difference?

crates/burn-core/src/module/param/base.rs
```diff
@@ -1,40 +1,265 @@
...
+    ///
+    /// Because of [OnceCell](OnceCell), we have a guarantee that the initialization will only be called once,
+    /// but it may be called at the same time as `lazy_device` and `lazy_is_require_grad`, which is
+    /// when the lock is actually useful, waiting for the initialization to be completed before
+    /// returning the value.
+    ///
+    /// To avoid creating locks on already initialized parameter, we wrap the lock inside an
+    /// Option, the inner option is required for resetting the state onced initialized.
+    /// TLDR: RwLock(None) only happens on the param reference that is lazy, but was initialized,
+    /// all other parameters
```
louisfd, line 45: I can't parse this sentence

louisfd (approved):
LGTM, just once sentence to fix

crates/burn-core/src/module/param/base.rs
laggui: I feel like the last sentence is incomplete 😅

all other parameters [blank].. what?

laggui (changes requested):
Implementation LGTM!

The examples have been updated but I think we're missing changes to the book, no?

crates/burn-core/src/nn/norm/batch.rs
```diff
@@ -35,36 +36,21 @@ pub struct BatchNorm<B: Backend, const D: usize> {
 impl BatchNormConfig {
     /// Initialize a new [batch norm](BatchNorm) module.
     pub fn init<B: Backend, const D: usize>(&self, device: &B::Device) -> BatchNorm<B, D> {
-        let gamma = Tensor::ones([self.num_features], device);
-        let beta = Tensor::zeros([self.num_features], device);
+        let gamma = Initializer::Ones.init([self.num_features], device);
+        let beta = Initializer::Zeros.init([self.num_features], device);
 
         let running_mean = Tensor::zeros([self.num_features], device);
         let running_var = Tensor::ones([self.num_features], device);
```
antimora, line 44: Any reason why it's different for running*?

antimora (approved):
LGTM

I had one question you should see.

antimora (commented):
Do we need to update the book?

laggui (approved):
LGTM! Minor phrasing comments but approving in advance because it's nothing critical.

PR #1548, opened 2024-03-28
Numerically stable `log_sigmoid`

laggui:
While making progress on a fine-tuning classification example I stumbled upon an issue with our `log_sigmoid` implementation which returned `-inf` for large negative values.

I first attempted to use this common log-sum-exp trick
```
log(sigmoid(x)) = log(1/(1 + exp(-x)))
                = log(1) - log(1 + exp(-x))
                = -log(1 + exp(-x))
                = x - log(1 + exp(x))
```

Which resulted in this implementation:
```rust
if x >= 0 {
    -log(1 + exp(-x))  // ok for positive values
} else {
    x - log(1 + exp(x)) // ok for negative values
}
```

That worked on wgpu but gave me NaNs for large values near the min and max on ndarray. That's when I stumbled upon the [pytorch implementation](https://github.com/pytorch/pytorch/pull/2211) that goes a step further, as implemented in this PR.

### Checklist

### Changes

Changed our `log_sigmoid` implementation to be numerically stable for large values.

### Testing

Added unit tests for `log_sigmoid`.

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -125,13 +125,38 @@ pub fn sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D>
 
 /// Applies the log sigmoid function.
 pub fn log_sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
+    /// To avoid overflow, we use the log-sum-exp trick.
+    ///
+    /// ```ignore
+    /// log(sigmoid(x)) = log(1/(1 + exp(-x)))
+    ///                 = log(1) - log(1 + exp(-x))
+    ///                 = -log(1 + exp(-x))
+    ///                 = -log(exp(0) + exp(-x))
+    /// ```
+    /// The `exp(t)` of even a moderate-magnitude positive number can be astronomically huge, so we
+    /// subtract the `max(t, 0)` of each value (where `t = -x` in this case). This results in the
+    /// following equivalence:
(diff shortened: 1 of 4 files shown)


louisfd (approved):
LGTM

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -125,13 +125,38 @@ pub fn sigmoid<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D>
...
         Precision::Half => {
             let tensor_full = tensor.into_full_precision();
-            let tensor_tmp = tensor_full.neg().exp().add_scalar(1.0_f32).log().neg();
+            let tensor_tmp = numerically_stable_log_sigmoid(tensor_full);
             Tensor::from_full_precision(tensor_tmp)
         }
-        _ => tensor.neg().exp().add_scalar(1.0_f32).log().neg(),
+        _ => numerically_stable_log_sigmoid(tensor),
     }
 }
```
nathanielsimard, line 161: I think a descent speedup for backends that don't implement fusion would be to move `log_sigmoid` and `sigmoid` into `burn_tensor::ops::activation` with the default implementation provided. We could then override those activations in backends that don't support fusion such as tch and candle.

laggui: Agreed, I just tackled the scope of the current `log_sigmoid` implementation but that definitely came to mind.

Btw `sigmoid` is already in `ActivationOps` just not `log_sigmoid` yet.

Should I tackle this in a new PR or expand this one?

antimora (approved):
LGTM

PR #1569, opened 2024-04-01
Refactor: split JitKernel and SourceKernel

louisfd:
Refactor Jit Kernel into an enum, many renaming, isolate template stuff behind a compile flag, expose workgroup size


crates/burn-jit/Cargo.toml
```diff
@@ -15,6 +15,7 @@ default = ["autotune", "std", "burn-compute/default", "fusion"]
 std = []
 doc = ["default"]
 autotune = []
+extension = []
```
nathanielsimard, line 18: I would put `template` instead of extension, since you can do a backend extension without it, but it would be with the gpu dialect.

crates/burn-jit/src/codegen/kernel.rs
```diff
@@ -44,7 +42,7 @@ pub fn execute_static<R, K, E>(
         handles.push(handle);
     }
 
-    let kernel = Box::new(StaticKernel::<K>::new(workgroup));
+    let kernel = Kernel::Jit(Box::new(StaticKernel::<K, R::Compiler>::new(workgroup)));
```
nathanielsimard, line 45: Can we drop the static kernel source?

crates/burn-jit/src/codegen/kernel.rs
```diff
@@ -250,7 +248,9 @@ fn execute_dynamic<R, K, E1, E2, E3>(
         handles.push(handle);
     }
 
-    let kernel: Box<dyn Kernel> = Box::new(DynamicKernel::new(kernel, workgroup));
+    let kernel = Kernel::Jit(Box::new(DynamicKernel::<K, R::Compiler>::new(
+        kernel, workgroup,
+    )));
```
nathanielsimard, line 253: This is quite hugly, can we create a function on `DynamicKernel` that returns `Kernel` instead of `Self` ? Or we could implement `From<DynamicKernel> for Kernel` and simply call `DyanamicKernel::new(...).into())` which would be nice.

crates/burn-jit/src/compute/kernel.rs
```diff
@@ -1,45 +1,93 @@
...
 
-/// Kernel trait with the [source](SourceTemplate) that will be compiled and cached based on the
+/// Kernel for JIT backends
+///
+/// Notes: by default, only Jit variant exists,
+/// but users can add more kernels from source by activating the
+/// extension feature flag.
+pub enum Kernel {
+    /// A JIT GPU compute shader
+    Jit(Box<dyn JitKernel>),
```
nathanielsimard, line 18: Could we simply have `Jit(Arc<gpu::ComputeShader>)` instead. I think we should remove as much as possible `Arc<dyn` and `Box<dyn` when not necessary. Maybe we need to create `GpuJitKernel`:

```rust
pub struct GpuJitKernel {
    pub id: String,
    pub shader: Arc<gpu::ComputeShapder>,
    pub info: ShaderInformation,
}
```
Therefore we would have a variant `JitGpu(JitGpuKernel)` and we can add more later instead of using dynamic traits.

crates/burn-jit/src/compute/kernel.rs
```diff
@@ -61,52 +109,85 @@ impl WorkGroup {
...
-    kernel: K,
+/// Wraps a [dynamic jit kernel](DynamicJitKernel) into a [Jit kernel](JitKernel) with launch
+/// information
+pub struct DynamicKernel<K, C> {
+    shader: ComputeShader,
+    id: String,
     workgroup: WorkGroup,
+    _kernel: PhantomData<K>,
+    _compiler: PhantomData<C>,
 }
```
nathanielsimard, line 120: We would not need that if we have `GpuJitKernel`, but in this case we don't want any generic, only function generic instead of struct.

crates/burn-wgpu/Cargo.toml
```diff
@@ -14,6 +14,7 @@ version.workspace = true
 default = ["fusion", "burn-jit/default"]
 fusion = ["burn-fusion", "burn-jit/fusion"]
 autotune = ["burn-jit/autotune"]
+extension = ["burn-jit/extension"]
```
nathanielsimard, line 17: should be template

nathanielsimard (changes requested):

burn-book/src/advanced/backend-extension/custom-wgpu-kernel.md
```diff
@@ -172,24 +173,20 @@ kernel_wgsl!(FusedMatmulAddReluRaw, "./kernel.wgsl");
...
-            .register("workgroup_size_y", self.workgroup_size_y.to_string())
+        FusedMatmulAddReluRaw::source()
+            .register("workgroup_size_x", self.workgroup_size.x.to_string())
+            .register("workgroup_size_y", self.workgroup_size.y.to_string())
             .register("elem", E::type_name())
-    }
-
-    fn id(&self) -> String {
-        format!("{:?}", self)
+            .register("int", "i32")
```
nathanielsimard, line 189: We need to update the book

crates/burn-jit/src/compute/kernel.rs
```diff
@@ -1,112 +1,191 @@
...
-impl WorkGroup {
-    /// Calculate the number of invocations of a compute shader.
-    pub fn num_invocations(&self) -> usize {
-        (self.x * self.y * self.z) as usize
-    }
+/// Kernel trait with the ComputeShader that will be compiled and cached based on the
+/// provided id.
+///
+/// The kernel will be launched with the given [shader information](ShaderInformation).
+pub trait JitGpuKernel: Send + Sync {
```
nathanielsimard, line 72: I think we can remove the Gpu here, only JitKernel. I though we would not abstract the gpu::ComputeShader, which would force the name of this trait  to have a Gpu.

nathanielsimard (changes requested):

nathanielsimard (approved):
LGTM

PR #1571, opened 2024-04-02
Support multilabel binary cross entropy

laggui:
The current implementation of `BinaryCrossEntropyLoss` only supports the binary case for 1-D inputs and targets. This PR extends the current definition to support `Tensor<B, D>`.

For multi-label classification the typical shape is `[batch_size, num_classes]` and targets are provided as multi-hot encoded targets.

### Checklist

### Changes

Added support for multilabel BCE loss for input `Tensor<B, D>`

- Changed implementation to use `log_sigmoid(x)` instead of `log(sigmoid(x))` with logits (numerical stability)
- Update label smoothing for multiclass (N > 2)
- Update loss class weighting to support multilabel case and `Tensor<B, D>`
- Fixed unit tests to use pre-computed values to compare against instead of comparing the function results with random values

**⚠️ Breaking**
Changed the default config to use `BinaryCrossEntropyLoss` *without* logits (expects probabilities, not raw predictions/logits). I think this makes more sense as a default and allows to use the config to build the loss with something more natural:

```rust
BinaryCrossEntropyLossConfig::new().with_logits(true)
```

This is a bit of an opiniated change, could be reverted if really desired.

### Testing

Added unit tests for new multi-label case and updated the binary 1-D tests.

### DIFF
--- a/crates/burn-core/src/nn/loss/binary_cross_entropy.rs
+++ b/crates/burn-core/src/nn/loss/binary_cross_entropy.rs
@@ -1,6 +1,7 @@
 use crate as burn;
 
 use crate::{config::Config, module::Module};
+use alloc::vec::Vec;
 use burn_tensor::activation::log_sigmoid;
 use burn_tensor::{backend::Backend, Int, Tensor};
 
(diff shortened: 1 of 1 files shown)


louisfd (approved):
LGTM

antimora (approved):
LGTM

Issue #788, opened 2023-09-08
FFT Module (like torch.fft)


discordance:
Hello burn team!
Thank you for all the hard work!

### Feature description

It would be incredibly valuable to add would be a Discret Fourier Transform (FFT) module, similar to the torch.fft in [PyTorch](https://pytorch.org/docs/stable/fft.html).

### Feature motivation

Having it as a part of the framework would add a fundational building block when working with differentiable Digital Signal Processing (e.g dDSP), Spectral losses, and Audio related tasks.

TheAndrewJackson:
If this feature is added this crate might be helpful: https://github.com/ejmahler/RustFFT


It has very fast SIMD accelerated FFT processing.

antimora:
Since Burn uses different backends, we should we take advantage of this. Doing purely for CPU will have data transfer overhead for GPU backends. I believe @Gadersd has implemented something already using Burn ops. https://github.com/Gadersd/whisper-burn

Perhaps, we can implement a default for CPU and let GPU backends override with more specialized implementation.

Gadersd:
Feel free to use my FFT implementation, but be aware that it isn’t as general as the one torch uses. There are some padding options and checks that should be added for generality.

TomWyllie:
I've started work on a PR for this feature, based on an NdArray / Wgpu op I wrote as a custom backend a few months back. There's a few design decisions to get right. We don't support complex numbers at the moment so the way I see it there's 3 ways to handle complex inputs and outputs for each op:

1. Add a complex type as an `Element` - probably the most idiomatic but I'm not sure if this is practical
2. Require separate real and imaginary tensors
3. Require a complex dimension of length 2 as the last dimension

Secondly, I'm guessing nobody will be too upset if we don't have a backward pass for FFT? The derivative of the fourier transform of a function is itself complex, and propagating imaginary gradients around the place seems like a bad idea. I'm assuming folks will want FFT to preprocess / extract features rather than as an op that needs to be in the graph.

Thoughts?

antimora:
Having it for the forward pass will cover the majority of use cases. As you mentioned many simply need to extract features and it is better to do using the backend device instead of CPU.

antimora:
The only thing I could suggest is to file issues for the remaining work so others can pick up or refer to the deficiencies.

TomWyllie:
[...]

PR #1574, opened 2024-04-03
Fft

TomWyllie:
## Pull Request Template
My branch should be fine to merge into this new Burn `feats/ops/fft` branch. Outstanding tasks before this is ready for a PR to main:

### Checklist

Required
- [ ] Confirmed that `run-checks all` script has been executed (@TomWyllie)
- [ ] Made sure the book is up to date with changes in this PR (@TomWyllie).
- [ ] Implement function for `burn-fusion` backend - I can do this is if someone gives me a few pointers, otherwise will need someone else to add it in. (unassigned)
- [ ] Convert WGSL shader in `burn-jit` to the new JIT representation (@louisfd)
- [ ] Add in 2D FFT - relatively simple given that 1D FFT is implemented (@TomWyllie)

Possible tasks, but can be left to future PRs if necessary;
- [ ] Implement function for `burn-tch` backend, libtorch does support FFT.
- [ ] Implement `rfft`

### Related Issues/PRs
#788

### Changes

The Fourier transform, implemented as the Fast Fourier Transform (FFT), is an extremely useful step in feature extraction for a variety of ML domains, most notably in 1D for speech / audio, but also in 2D for calculation of spatial frequencies, or even calculating large convolutions or correlations in 2D, which is often faster in the frequency domain. 

Have implemented the basic 1D forwards `fft` using pure NdArray ops for the `burn-ndarray` backend, and using a WGSL shader for the `burn-wgpu` backend. The shaders are really fast, e.g. ~15 ms to compute 1024 FFTs each of length 1024 on my GPU. Implemented the inverse transform `ifft` using a wrapper of `fft` with conjugation.

The input shape is required to be `[batch_size, length, 2]` where the transform is applied over `length` which must be a power of 2, as the FFT implementation is Cooley-Tukey. Zero padding is usually used for other input lengths. The last dimension corresponds to (real, imaginary), since there is no complex type for Rust / Burn. Felt like overkill to build a whole complex type just for FFT. So if you have a real input signal of length `512` you'll have to unsqueeze batch to 1, then pad some zeros to manually set a zero imaginary part. Other approaches, e.g. separate real and imaginary tensors, would be possible.

### Testing
Following cases with known correct outputs, forwards and backwards (from `numpy.fft`):
single element
simple delta function (2x FFT len 2)
simple 1d case (2x FFT len 2)
even power 2 case (1x FFT len 16)
odd power 2 case (1x FFT len 32)

Also
round trip precision (1x FFT len 1024, 1x iFFT len 1024)
Handle non-power 2 input
Handle invalid representation of complex input.


antimora:
no_std tests are failing. It should be easy to fix:

image

TomWyllie:
ah yes, missed that, thank you. Will take a look later.

antimora:
It's odd that the PR diff is showing other merged files. It must be because the branch this PR is attempting to merge into is `tracel-ai:feat/ops/fft` and not main. Was this by intention?

@nathanielsimard , @louisfd , @laggui please advise on this.

TomWyllie:
Yes, the `tracel-ai:feats/ops/fft` branch was made for this PR. This isn't ready to merge into main yet, @laggui is going to rewrite the WGSL into the new JIT representation - see description above and linked issue. Also need to finish off a couple other smaller issues. I had already merged in latest changed on main after the branch was made though so we might need to resync the `fft` branch.

louisfd:
@antimora Yes, it's not ready for main yet but we need to collaborate

louisfd (approved):
Did not really review yet since does not merge on main

louisfd:
@TomWyllie i ported the kernel to jit representation on feat/ops/fft

PR #1532, opened 2024-03-26
LSTM Timeseries prediction example

NicoZweifel:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.

### Changes

- Adds a timeseries forecasting example using the `LSTM` that was added in https://github.com/tracel-ai/burn/pull/370, using a Partial Dataset from [Huggingface](https://huggingface.co/datasets/edarchimbaud/timeseries-1d-stocks).
- The Dataset is limited to 10000 entries at the moment. Training on the full Dataset seems to be buggy still. I am not sure if the normalization is messed up or if there might be a memory limitation or bug with the `SqliteDataset`.

I have not narrowed it down yet as I am using custom Datasets on my other burn project and they work fine (`InMemory` with data from alphavantage). I might need to spend some more time on it to figure it out but since it doesn't block me in my other goals and the example seems to work with 10000 entries I though I could publish this as a draft for now.

### Testing

`cargo run --example lstm --features tch-cpu`

Issue #1552, opened 2024-03-28
[Autodiff] Client/server architecture to handle subgraphs


nathanielsimard:
This should fix #1532

PR #1575, opened 2024-04-03
[Breaking] Make Tensor, Module, Optimizer !Sync + Refactor Autodiff

nathanielsimard:
The goal of this PR is to set all tensor data structures as `!Sync`, meaning they cannot be accessed by multiple concurrent threads simultaneously. Why? Because implementing locks in numerous places for something that doesn't hold significance is unnecessary, especially considering that all tensor operations are owned. Initially, this might not significantly impact performance, but it lays the groundwork for future optimizations. Additionally, it promotes best practices, such as sending owned tensors, modules, and/or optimizers to different threads by cloning them or changing their device using `to_device`.

It's important to note that this change doesn't affect the actual tensor's data buffer, which should still implement `Sync`. While a tensor can still be cloned multiple times and sent to various threads, you won't be able to wrap your tensor in an `Arc` before sending a shared reference to multiple threads. This ensures tensor stream stability, simplify the compilation pipeline, and more. The API was designed when tensor operations were read-only, meaning they didn't take ownership, necessitating the sharing of references among multiple threads. However, with the owned API now strongly in place and leveraged for optimizations, this requirement becomes obsolete.

This PR doesn't eliminate all locks; it's part of an ongoing effort. However, it does remove several locks. Most locks can be replaced by `RefCell` or `mpsc::channel` instead of `mpsc::sync_channel`, but this needs to be done carefully as it opens the door to different techniques for inner mutability.

# Autodiff

The current autodiff implementation necessitates the `AutodiffTensor` to implement both `Send` and `Sync`,  so I had to make changes during this update. I implemented the client/server architecture we discussed before in issue #1552, which should resolve issue #1532.

The current implementation is more efficient than the previous one by avoiding a lot of locking and maintaining separate graph representations for the backward pass and for memory management. This doesn't significantly change the training time of the examples, but it may impact smaller models with lots of tensor operations, such as small LSTM models with many layers.

PS: Memory management is necessary since not all created graphs will be consumed by a backward call, which could lead to memory leaks. We initiate the cleanup right after the backward step is executed, which is expected to fill the execution queue with all the backward kernels, thus this shouldn't block the main thread longer than it takes for the backward pass to finish.


crates/burn-autodiff/src/tensor.rs
```diff
@@ -67,59 +80,74 @@ impl<B: Backend, const D: usize> AutodiffTensor<B, D> {
...
         }
     }
 
     /// Register a step into a graph for that tensor.
-    pub fn register_step<O: Step + 'static>(mut self, ops: O) -> Self {
-        self.graph = self.graph.register(&self.node.id, Box::new(ops));
+    ///
+    /// # Warning
+    ///
+    /// This should be called only onced per tensor.
```
louisfd, line 136: once

crates/burn-core/src/module/param/running.rs
```diff
@@ -42,20 +42,23 @@ use threading::*;
 pub struct RunningState<V> {
     id: ParamId,
     values: Arc<Mutex<HashMap<ThreadId, V>>>,
-    value: Arc<RwLock<V>>,
+    value: Arc<Mutex<V>>,
 }
 
+// TODO: Remove at some point.
+// unsafe impl<V: Send> Send for RunningState<V> {}
```
louisfd, line 49: What's that for?
It's commented so should it be removed already?

louisfd (approved):
Looks good, just minor comments

PR #1580, opened 2024-04-05
Refactor/burn compute

nathanielsimard:
Instead of a single handle we now have:

- ID => Buf id
- Tensor Buf Handle => keep track of the number of tensor pointing to a buffer.
- Buf Handle => keep track of all reference to a buffer (necessary for dealloc)


crates/burn-compute/src/id.rs
```diff
@@ -25,26 +33,134 @@ macro_rules! storage_id_type {
     };
 }
 
+/// Reference to buffer handles, to be holded by a tensor.
```
louisfd, line 36: held

crates/burn-compute/src/id.rs
```diff
@@ -25,26 +33,134 @@ macro_rules! storage_id_type {
...
+#[derive(Clone, Debug)]
+pub struct TensorBufRef<Id> {
+    id: Arc<Id>,
+    all: Arc<()>,
+}
+
+/// Reference to buffer handles, to be holded by anything else than tensors.
```
louisfd, line 43: held

crates/burn-compute/src/id.rs
```diff
@@ -25,26 +33,134 @@ macro_rules! storage_id_type {
...
+
+    /// If the buffer can be dealloc.
+    pub(crate) fn can_be_dealloc(&self) -> bool {
+        Arc::strong_count(&self.all) <= 1
+    }
+}
+
 #[macro_export(local_inner_macros)]
-/// Create a new memory ID type.
+/// Create a new memory ID types.
```
louisfd, line 104: not sure why you added plural, but you should at least remove the "a"

crates/burn-compute/src/memory_management/simple.rs
```diff
@@ -116,10 +110,28 @@ impl DeallocStrategy {
...
+    handle: ChunkTensorBufHandle,
+    slices: Vec<SliceId>,
+}
+
+#[derive(new)]
+struct Slice {
+    storage: StorageHandle,
+    handle: SliceTensorBufHandle,
+    // It is important to keep the chunk handle inside the slice, since it increases the ref count
+    // on the chunk id and make the `is_free` method returns false until the slice is freed.
```
louisfd, line 125: makes the `is_free` method return

louisfd (changes requested):
A few typos. Also as discussed let's rename server buf handles to bindings

PR #1582, opened 2024-04-07
Add f16 support in the wgpu backend

p1-0tr:
### Checklist

### Related Issues/PRs

Split from: https://github.com/tracel-ai/burn/pull/1475
Dawn support: https://github.com/tracel-ai/burn/pull/1583

### Changes

The burn-wgpu backend currently does not support computations on 16 bit floats. This. for example, limits the ability to run LLMs on top of Burn, on widely available hardware. So, add 16 bit float support in burn-wgpu.

### Testing

I used this change on top of my changes which add the ability to run with Dawn instead of wgpu, to run llama2-burn with f16.

### DIFF
--- a/crates/burn-jit/src/codegen/dialect/gpu/shader.rs
+++ b/crates/burn-jit/src/codegen/dialect/gpu/shader.rs
@@ -21,6 +21,7 @@ pub enum Visibility {
 #[allow(missing_docs)]
 pub enum Elem {
     Float,
+    Half,
     Int,
     UInt,
     Bool,
@@ -36,6 +37,7 @@ impl Display for Elem {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::Float => f.write_str("float"),
+            Self::Half => f.write_str("half"),
             Self::Int => f.write_str("int"),
             Self::UInt => f.write_str("uint"),
             Self::Bool => f.write_str("bool"),
--- a/crates/burn-jit/src/element.rs
+++ b/crates/burn-jit/src/element.rs
@@ -1,5 +1,6 @@
 use crate::codegen::dialect::gpu;
 use burn_tensor::Element;
+use half::f16;
 
 /// The base element trait for the jit backend.
 pub trait JitElement:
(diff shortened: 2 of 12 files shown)


crates/burn-jit/src/codegen/dialect/gpu/shader.rs
```diff
@@ -21,6 +21,7 @@ pub enum Visibility {
 #[allow(missing_docs)]
 pub enum Elem {
     Float,
+    Half,
```
nathanielsimard, line 24: There is no need to add `Half` here, `Float` should cover all float types of all precisions in this context.

crates/burn-jit/src/element.rs
```diff
@@ -90,5 +91,27 @@ impl JitElement for f32 {
...
+    }
+    fn as_bytes(slice: &[Self]) -> &[u8] {
+        bytemuck::cast_slice(slice)
+    }
+    fn from_bytes(bytes: &[u8]) -> &[Self] {
+        bytemuck::cast_slice(bytes)
+    }
+    fn gpu_elem() -> gpu::Elem {
+        gpu::Elem::Half
+    }
```
nathanielsimard, line 106: The gpu element would be `Float` here.

crates/burn-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -78,6 +78,11 @@ impl<F: FloatElement, I: IntElement> WgslCompiler<F, I> {
         self.num_inputs = value.inputs.len();
         self.num_outputs = value.outputs.len();
 
+        let features = match F::gpu_elem() {
+            gpu::Elem::Half => vec![wgsl::Feature::ShaderF16],
+            _ => vec![],
+        };
```
nathanielsimard, line 84: I would check using `F::wgpu_elem() == Elem::F16` instead.

crates/burn-wgpu/src/compiler/wgsl/compiler.rs
```diff
@@ -129,6 +135,7 @@ impl<F: FloatElement, I: IntElement> WgslCompiler<F, I> {
     fn compile_elem(value: gpu::Elem) -> wgsl::Elem {
         match value {
             gpu::Elem::Float => F::wgpu_elem(),
+            gpu::Elem::Half => F::wgpu_elem(),
```
nathanielsimard, line 138: This line pretty much explains why we don't need `Half` in the `gpu::Elem` enum :)

nathanielsimard (changes requested):

laggui:
This belongs in cubecl now 🙂 will close here due to the big changes we made with the migration.

PR #1591, opened 2024-04-08
Add learner training report summary

### LINKED ISSUE #1264: Generate a training report in burn-train
Maybe use the file loggers to generate a report and print it to stdout after the CLI dashboard is closed. We could also generate a json file for the summary.

laggui:
### Checklist

### Related Issues/PRs

Closes #1264 

### Changes

Added `LearnerSummary` struct to retrieve recorded metrics during training.

- Added `NumericEntry` enum to properly aggregate values
  - Previous implementation did not take into account different mini-batch sizes which resulted in incorrect averages
- Changed `NumericMetricState` to use `NumericEntry` with corresponding value and batch size for each metric entry that is serialized
- Renamed `register_metric_train` -> `register_train_metric` in `Metrics` to match other method names

### Testing

Added aggregate test for `NumericEntry` usage with a logger and ran the examples with the newly added summary.

### Example Usage

Since the training artifacts are persistent (still on disk after a training is completed), the `LearnerSummary` doesn't have to be used only at the end of your training script. It can be used at any time by pointing to the artifact directory, and you can select which metrics to look for.

```rust
let summary = LearnerSummary::new(
    artifact_dir,
    &[AccuracyMetric::<B>::NAME, LossMetric::<B>::NAME],
);
println!("{}", summary);
```

or

```rust
let summary = LearnerSummary::new(
    artifact_dir,
    &["Accuracy", "Loss"],
);
println!("{}", summary);
```

Sample output:
```
======================== Learner Summary ========================
Total Epochs: 6


| Split | Metric   | Min.     | Epoch    | Max.     | Epoch    |
|-------|----------|----------|----------|----------|----------|
| Train | Accuracy | 93.492   | 1        | 99.133   | 5        |
| Train | Loss     | 0.029    | 5        | 0.243    | 1        |
| Valid | Accuracy | 97.630   | 1        | 98.910   | 5        |
| Valid | Loss     | 0.031    | 5        | 0.076    | 1        |
```

Feel free to suggest improvements on the output formatting!

### DIFF
--- a/burn-book/src/basic-workflow/training.md
+++ b/burn-book/src/basic-workflow/training.md
@@ -1,11 +1,11 @@
 # Training
 
-We are now ready to write the necessary code to train our model on the MNIST dataset.
-We shall define the code for this training section in the file: `src/training.rs`.
+We are now ready to write the necessary code to train our model on the MNIST dataset. We shall
+define the code for this training section in the file: `src/training.rs`.
 
-Instead of a simple tensor, the model should output an item that can be understood by the learner, a struct whose
-responsibility is to apply an optimizer to the model. The output struct is used for all metrics
-calculated during the training. Therefore it should include all the necessary information to
+Instead of a simple tensor, the model should output an item that can be understood by the learner, a
+struct whose responsibility is to apply an optimizer to the model. The output struct is used for all
+metrics calculated during the training. Therefore it should include all the necessary information to
 calculate any metric that you want for a task.
 
 Burn provides two basic output types: `ClassificationOutput` and `RegressionOutput`. They implement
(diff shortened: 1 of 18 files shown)


burn-book/src/basic-workflow/training.md
```diff
@@ -181,8 +193,9 @@ Once the learner is created, we can simply call `fit` and provide the training a
 dataloaders. For the sake of simplicity in this example, we employ the test set as the validation
 set; however, we do not recommend this practice for actual usage.
 
-Finally, the trained model is returned by the `fit` method, and the only remaining task is saving
-the trained weights using the `CompactRecorder`. This recorder employs the `MessagePack` format with
-`gzip` compression, `f16` for floats and `i16` for integers. Other recorders are available, offering
-support for various formats, such as `BinCode` and `JSON`, with or without compression. Any backend,
-regardless of precision, can load recorded data of any kind.
+Finally, the trained model is returned by the `fit` method. The trained weights are then saved using
+the `CompactRecorder`. This recorder employs the `MessagePack` format with `gzip` compression, `f16`
```
antimora, line 197: Do we still support compression?

antimora (approved):
LGTM 

I had one comment

burn-book/src/basic-workflow/training.md
laggui: Hmmm we do, but not for the `CompactRecorder` 😅 it's the default recorder with half precision.. and the default does not use compression anymore.

Good catch even if unrelated to the PR hahaha.

Will fix.

laggui:
Fixed your comment! Should I add tests for the expected summary results?

laggui:
> Maybe the `learner.fit(train, valid)` should return the summary with the trained model as well?
> 
> Otherwise LGTM!

I thought about that initially!

Went over a couple of iterations before landing on this to retrieve the summary. Wasn't sure if we *always* wanted to return the summary, but it can always be added with the current implementation.

laggui:
Made the requested changes we discussed offline to automatically display the summary at the end of `learner.fit()`.

PR #1594, opened 2024-04-09
Fixes to code examples in section 5.2

hrishim:
### Checklist

### Related Issues/PRs

None

### Changes

These changes are to the Burn book only. No changes to the code in Burn.

I think The Burn Book may need to be updated . The Tensor [section](https://burn.dev/book/building-blocks/tensor.html)  gives these code examples to create Tensors:

```Rust
let tensor_1 = Tensor::<Backend, 1>::from_floats([1.0, 2.0, 3.0, 4.0, 5.0]);

let tensor_1 = Tensor::<Wgpu, 1>::from_data([1.0, 2.0, 3.0]);
```
```
this function takes 2 arguments but 1 argument was supplied 
main.rs(17, 51): an argument of type `&WgpuDevice` is missing
```

However, I get a compiler error because from_floats and from_data takes an additional argument - the device. One has to do something like below to create a Tensor:

```Rust
let device = burn::backend::wgpu::WgpuDevice::default();
let pos_samples = Tensor::<Wgpu, 1>::from_data([1.0, 2.0, 3.0], &device);
```

### Testing

Since this is a change only to documentation (The Burn Book) I am not sure if running run_check.sh is required. I tried it any way and run_check.sh fails with this error

```
> ./run_check.sh
(base) hrishi@resnet:burn$ ./run-checks.sh 
    Finished dev [unoptimized] target(s) in 0.09s
     Running `target/debug/xtask run-checks`
error: the following required arguments were not provided:
  <ENV>
```

I ran ./run_check.sh std and that passed. I have not tested all other ENV options

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -20,7 +20,7 @@ The actual shape of the tensor is inferred from its initialization. For example,
 ```rust, ignore
 let floats = [1.0, 2.0, 3.0, 4.0, 5.0];
 
-//Get the default device
+// Get the default device
 let device = Default::default();
 
 // correct: Tensor is 1-Dimensional with 5 elements
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -20,7 +20,8 @@ The actual shape of the tensor is inferred from its initialization. For example,
 ```rust, ignore
 let floats = [1.0, 2.0, 3.0, 4.0, 5.0];
 
-let device = burn::backend::wgpu::WgpuDevice::default();
+//Get the default device
+let device = Default::default();
 
 // correct: Tensor is 1-Dimensional with 5 elements
 let tensor_1 = Tensor::<Backend, 1>::from_floats(floats, &device);
@@ -38,8 +39,6 @@ method `.to_data()` should be employed when intending to reuse the tensor afterw
 a tensor from different inputs.
 
 ```rust, ignore
-
-let device = burn::backend::wgpu::WgpuDevice::default();
 
 // Initialization from a given Backend (Wgpu)
 let tensor_1 = Tensor::<Wgpu, 1>::from_data([1.0, 2.0, 3.0], &device);
@@ -88,7 +87,6 @@ times will necessitate cloning it. Let's look at an example to understand the ow
 cloning better. Suppose we want to do a simple min-max normalization of an input tensor.
 
 ```rust, ignore
-let device = burn::backend::wgpu::WgpuDevice::default();
 let input = Tensor::<Wgpu, 1>::from_floats([1.0, 2.0, 3.0, 4.0], &device);
 let min = input.min();
 let max = input.max();
@@ -103,7 +101,6 @@ available for further operations. Burn Tensors like most complex primitives do n
 doing min-max normalization with cloning.
 
 ```rust, ignore
-let device = burn::backend::wgpu::WgpuDevice::default();
 let input = Tensor::<Wgpu, 1>::from_floats([1.0, 2.0, 3.0, 4.0], &device);
 let min = input.clone().min();
 let max = input.clone().max();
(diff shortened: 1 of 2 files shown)


burn-book/src/building-blocks/tensor.md
```diff
@@ -20,10 +20,12 @@ The actual shape of the tensor is inferred from its initialization. For example,
 ```rust, ignore
 let floats = [1.0, 2.0, 3.0, 4.0, 5.0];
 
+let device = burn::backend::wgpu::WgpuDevice::default();
```
laggui, line 23: Instead of making the device specific to wgpu, we can make the example generic to the backend.

Use this instead:
```rust
```

It will automatically select the default device on the selected backend.

burn-book/src/building-blocks/tensor.md
```diff
@@ -37,20 +39,22 @@ a tensor from different inputs.
 
 ```rust, ignore
 
+let device = burn::backend::wgpu::WgpuDevice::default();
```
laggui, line 42: Since the examples below are supposed to be generic, I don't think we need to carry over the device declaration. Let's remove this line (but keep the fixes to the tensor creation methods using `&device`, obviously).

burn-book/src/building-blocks/tensor.md
```diff
@@ -84,7 +88,8 @@ times will necessitate cloning it. Let's look at an example to understand the ow
 cloning better. Suppose we want to do a simple min-max normalization of an input tensor.
 
 ```rust, ignore
-let input = Tensor::<Wgpu, 1>::from_floats([1.0, 2.0, 3.0, 4.0]);
+let device = burn::backend::wgpu::WgpuDevice::default();
```
laggui, line 91: Same comment. I think we can remove the device declaration.

burn-book/src/building-blocks/tensor.md
```diff
@@ -98,7 +103,8 @@ available for further operations. Burn Tensors like most complex primitives do n
 doing min-max normalization with cloning.
 
 ```rust, ignore
-let input = Tensor::<Wgpu, 1>::from_floats([1.0, 2.0, 3.0, 4.0]);
+let device = burn::backend::wgpu::WgpuDevice::default();
```
laggui, line 106: Same comment.

laggui (changes requested):
Good catch! Guess we missed these with the API changes.

Some minor comments, but should be good to go after that.

FYI, you're right that the run checks script didn't necessarily apply here. But sometimes modifying the book we can make some typos and we have a check for that. As to why your attempt to run the script fail, it's because you're missing the environment `all` to check every environment: `./run-checks.sh all`.

burn-book/src/building-blocks/tensor.md
hrishim: Wouldn't Default have to be imported using a use something? 

When I try it on example code I get an error 

```
no function or associated item named `device` found for trait `std::default::Default`
function or associated item not found in `Default`
```

laggui: Copying the answer from our discord convo:

Not required, no. The compiler is able to figure out which default trait implementation to call based on the rest of the code (type inference). `Default::default()` is just `<_ as Default>::default()`

Try it for yourself:
```rust
use burn::{backend::Wgpu, tensor::Tensor};

// Type alias for the backend to use.
type Backend = Wgpu;

fn main() {
    // correct: Tensor is 1-Dimensional with 5 elements
    let tensor_1 = Tensor::<Backend, 1>::from_floats([1.0, 2.0, 3.0, 4.0, 5.0], &device);

    println!("{}", tensor_1);
}
```

Not sure in what context/program you're trying to use this snippet, but the example usage in this markdown are not meant to be complete with all modules declaration anyway. If you're using it like the snippet I wrote above it should work :)

hrishim: Thanks I understand. I have made the fixes and pushed

laggui (approved):
Thanks for your fixes! 🙂

PR #1596, opened 2024-04-09
[CI] Migrate workflows to Blaze runners

dcvz:
### Changes

Updates the workflow to use Blaze runners instead of default Github runners! 🔥


syl20bnr (approved):
8 and 9 min for both stable and 1.75 😍

dcvz:
> 8 and 9 min for both stable and 1.75 😍

with no cache!

Issue #1510, opened 2024-03-25
Implement `modulo` or `remainder` operation.


zemelLeong:
https://pytorch.org/docs/stable/generated/torch.remainder.html

laggui:
Accidentally closed this thinking we had full remainder support with the linked PRs, but actually this was only implemented for `tensor % scalar`. We still need element-wise remainder support.

PR #1597, opened 2024-04-09
Feat/remainder

agelas:
## Pull Request Template
I was doing some playing around and noticed this wasn't included, also I believe there's an open issue about this in #1510.

### Checklist

- [x] Made sure the book is up to date with changes in this PR (the contributor-book needs to be updated though).

### Related Issues/PRs

#1510 

### Changes

Added the equivalent of PyTorch's [remainder](https://pytorch.org/docs/stable/generated/torch.remainder.html) operation

### Testing

Added new tests to `burn-tensor`, will add more later once wgpu is figured out.

For `burn-jit`, I redirected the remainder to the modulo operator because in WGSL, it *should* be the same as PyTorch's definition. Here, the [%](https://math.hws.edu/graphicsbook/c6/s3.html#webgl.3.4) for WGSL is defined as

>  The return value is computed as x − y*floor(x/y). As with min and max, y can be either a vector or a float. The mod function can be used as a substitute for the % operator

In PyTorch, the remainder is defined as
> torch.remainder(a, b) == a - a.div(b, rounding_mode="floor") * b

So they should be equivalent, but there's some errors in the wgpu tests that I'll have to look into. I probably set something up wrong in  `numeric.rs` in `burn-jit`. The other backends seem to work. If anyone has any ideas though here is the test output:

```bash
---- tests::jit_fusion::remainder::tests::should_support_remainder_basic stdout ----
thread 'tests::jit_fusion::remainder::tests::should_support_remainder_basic' panicked at crates/burn-wgpu/src/lib.rs:78:5:
Tensors are not approx eq:
  => Position 0: 1 != -1 | difference 2 > tolerance 0.0010000000000000002
  => Position 2: 1 != -1 | difference 2 > tolerance 0.0010000000000000002

---- tests::jit_fusion::remainder::tests::should_support_remainder_float stdout ----
thread 'tests::jit_fusion::remainder::tests::should_support_remainder_float' panicked at crates/burn-wgpu/src/lib.rs:78:5:
Tensors are not approx eq:
  => Position 0: -0.5 != 1 | difference 1.5 > tolerance 0.0010000000000000002
  => Position 1: -1 != 0.5 | difference 1.5 > tolerance 0.0010000000000000002
  => Position 3: -0.5 != 1 | difference 1.5 > tolerance 0.0010000000000000002
  => Position 4: -1 != 0.5 | difference 1.5 > tolerance 0.0010000000000000002

failures:
    tests::jit::remainder::tests::should_support_remainder_basic 
    tests::jit::remainder::tests::should_support_remainder_float // this and the one above looks exactly like the below two
    tests::jit_fusion::remainder::tests::should_support_remainder_basic
    tests::jit_fusion::remainder::tests::should_support_remainder_float
```

### DIFF
--- a/crates/burn-wgpu/src/compiler/wgsl/instructions.rs
+++ b/crates/burn-wgpu/src/compiler/wgsl/instructions.rs
@@ -248,7 +248,7 @@ impl Display for Instruction {
                 f.write_fmt(format_args!("{out} = {item}({lhs}[{rhs}]);\n"))
             }
             Instruction::Modulo { lhs, rhs, out } => {
-                f.write_fmt(format_args!("{out} = {lhs} % {rhs};\n"))
+                f.write_fmt(format_args!("{out} = (({lhs} % {rhs}) + {rhs}) % {rhs};\n"))
             }
             Instruction::Sub { lhs, rhs, out } => {
                 f.write_fmt(format_args!("{out} = {lhs} - {rhs};\n"))
--- a/crates/burn-wgpu/src/compiler/wgsl/instructions.rs
+++ b/crates/burn-wgpu/src/compiler/wgsl/instructions.rs
@@ -218,6 +218,11 @@ pub enum Instruction {
         input: Variable,
         out: Variable,
     },
+    Remainder {
+        lhs: Variable,
+        rhs: Variable,
+        out: Variable,
+    },
 }
 
 impl Display for Instruction {
@@ -248,6 +253,9 @@ impl Display for Instruction {
                 f.write_fmt(format_args!("{out} = {item}({lhs}[{rhs}]);\n"))
             }
             Instruction::Modulo { lhs, rhs, out } => {
+                f.write_fmt(format_args!("{out} = (({lhs} % {rhs}) + {rhs}) % {rhs};\n"))
+            }
+            Instruction::Remainder { lhs, rhs, out } => {
                 f.write_fmt(format_args!("{out} = (({lhs} % {rhs}) + {rhs}) % {rhs};\n"))
             }
             Instruction::Sub { lhs, rhs, out } => {
--- a/crates/burn-wgpu/src/compiler/wgsl/instructions.rs
+++ b/crates/burn-wgpu/src/compiler/wgsl/instructions.rs
@@ -253,7 +253,7 @@ impl Display for Instruction {
                 f.write_fmt(format_args!("{out} = {item}({lhs}[{rhs}]);\n"))
             }
             Instruction::Modulo { lhs, rhs, out } => {
-                f.write_fmt(format_args!("{out} = (({lhs} % {rhs}) + {rhs}) % {rhs};\n"))
+                f.write_fmt(format_args!("{out} = {lhs} % {rhs};\n"))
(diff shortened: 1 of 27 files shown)


louisfd:
Thanks @agelas 
Looks good in general. 
For debugging JIT, you can add `println!("{source}");` right after line 147 in `crates/burn-wgpu/src/compute/server.rs` to see the resulting wgsl kernel. It's going to look a bit cryptic but since it's a very small kernel you should be fine.

agelas:
@louisfd Thanks, that's super useful.

louisfd:
@agelas I see it passes CI, is it working now?

agelas:
@louisfd yep!

crates/burn-wgpu/src/compiler/wgsl/instructions.rs
```diff
@@ -248,7 +248,7 @@ impl Display for Instruction {
                 f.write_fmt(format_args!("{out} = {item}({lhs}[{rhs}]);\n"))
             }
             Instruction::Modulo { lhs, rhs, out } => {
-                f.write_fmt(format_args!("{out} = {lhs} % {rhs};\n"))
+                f.write_fmt(format_args!("{out} = (({lhs} % {rhs}) + {rhs}) % {rhs};\n"))
```
louisfd, line 251: Is this complicated formula there in order to support negatives and floats?
We have to keep in mind that this Modulo instruction is used in other kernels, mostly with simple unsigned ints, and often in the hot loop so we want to keep it as simple as possible. 
I think there is a difference between modulo which is the basic operator % and remainder which has to take negative and floats into account, and the remainder more complicated formula should be at the kernel level in burn-jit, not in the instruction.

louisfd (changes requested):
I have a comment regarding the modification of the modulo wgsl instruction

crates/burn-wgpu/src/compiler/wgsl/instructions.rs
agelas: @louisfd yes, the `((a % b) + b) % b` form is there to keep the sign, I may have been a little over-zealous in matching the behavior of PyTorch. When you say that formula should be at the kernel level in burn-jit, do you mean it should implemented in `numeric.rs`, or do you mean something like it's own dedicated Operator::Remainder so there isn't that redirect to the pre-existing modulo operation?

It terms of what % means, in [Rust](https://stackoverflow.com/questions/31210357/is-there-a-modulus-not-remainder-function-operation) (and in WGSL), the % calculates the remainder, not the modulo. [Things get a bit wonky when negative numbers are used](https://stackoverflow.com/questions/13683563/whats-the-difference-between-mod-and-remainder). It's different in Python though hence the difference, and the need for PyTorch's [remainder](https://pytorch.org/docs/stable/generated/torch.remainder.html) operation.

agelas: @louisfd I went ahead and created a custom instruction for the remainder operation so we avoid any clash with the original one.

louisfd (approved):
Thanks @agelas for this op

PR #1599, opened 2024-04-10
[CI] Add a retry on adding ppa for kisak

dcvz:
### Changes

The `install llvmpipe and lavapipe` has been failing on the CI in the last days. This PR:
-  adds a retry mechanism that will try up to 5 times to add the repository and will wait 5s in between.
- extracts install step into sharable composite action


syl20bnr (approved):
Thanks for removing duplication as well !

PR #1600, opened 2024-04-11
Repeat ops autodiff & fusion + fix autodiff ones & zeros

louisfd:
I implemented the backward pass for repeat, enabling it to use the jit kernel of repeat instead of the fallback implementation with slice assign. 
I also added fusion support for repeat

Also fixed ones and zeros in autodiff which were mistakingly cpu-bound

### DIFF
--- a/crates/burn-autodiff/src/ops/bool_tensor.rs
+++ b/crates/burn-autodiff/src/ops/bool_tensor.rs
@@ -135,4 +135,12 @@ impl<B: Backend, C: CheckpointStrategy> BoolTensorOps<Self> for Autodiff<B, C> {
     ) -> BoolTensor<B, D2> {
         B::bool_expand(tensor, shape)
     }
+
+    fn bool_repeat<const D: usize>(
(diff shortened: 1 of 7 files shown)


laggui (approved):
Was curious to try the difference with the default implementation..

```rust
use std::time::Instant;

use burn::{
    backend::{wgpu::AutoGraphicsApi, Autodiff, Wgpu},
    tensor::backend::Backend,
    tensor::Tensor,
};

fn main() {
    type B = Autodiff<Wgpu<AutoGraphicsApi, f32, i32>>;
    let now = Instant::now();
    let t = Tensor::<B, 1>::from_floats([1.0], &device);
    let _rep = t.repeat(0, 1024 * 1024);

    B::sync(&device);
    let elapsed = now.elapsed().as_millis();
    println!("Took {} ms", elapsed);
}
```

With the default impl on main: Took 36066 ms
With your fix: Took 387 ms

... a timely fix 😄

PR #1602, opened 2024-04-11
Fix autodiff memory management graph cleaning

nathanielsimard:


crates/burn-autodiff/Cargo.toml
```diff
@@ -22,6 +22,7 @@ burn-tensor-testgen = { path = "../burn-tensor-testgen", version = "0.13.0", opt
 
 derive-new = { workspace = true }
 spin = { workspace = true }
+log = { workspace = true }
```
laggui, line 25: Leftover from debugging?

laggui (approved):
Easy review for a not-so-easy to debug issue 😄 

Tested your fix and it works.

Only have one minor comment, I think you forgot the log dependency from debugging, But approving in advance.

Issue #1603, opened 2024-04-11
Rotary Positional Encodings


ashdtu:
### Feature description

Add module to support computing rotary positional embeddings.

### Feature motivation

Rotary positional embeddings are used in most modern LLMs as a counter to positional encodings due to their ability to incorporate explicit relative position dependency.

ashdtu:
Addressed in PR #1604 . Closing

PR #1604, opened 2024-04-11
support for rotary positional encoding to transformer modules.

ashdtu:
### Checklist

### Related Issues/PRs
#1603 

### Changes
Implementation of Rotary positional encodings as an alternative to sinusoidal positional embeddings due to their ability to incorporate explicit relative position dependency. They are being used in recent LLM models such as Llama, Mistral etc.

### Testing

Test cases to verify output against PyTorch implementations of Rotary embeddings (https://github.com/lucidrains/rotary-embedding-torch).

### DIFF
--- a/crates/burn-core/src/nn/rope_encoding.rs
+++ b/crates/burn-core/src/nn/rope_encoding.rs
@@ -1,0 +1,213 @@
+use crate as burn;
+use crate::config::Config;
+use crate::module::Module;
+use crate::tensor::backend::Backend;
+use crate::tensor::Tensor;
+use alloc::vec;
+use burn_tensor::Int;
+
+/// Configuration to create a [RotaryEncoding](RotaryEncoding) layer.
+#[derive(Config, Debug)]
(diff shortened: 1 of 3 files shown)


crates/burn-core/src/nn/rope_encoding.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    /// Scaling factor for frequency computation. Defaults to 10000.0
+    #[config(default = "10000.0")]
+    theta: f32,
+}
+
+impl RotaryEncodingConfig {
+    /// Initialize a new [RotaryEncoding](RotaryEncoding) module.
+    ///
+    /// Panics if the size of input embedding dimension is not even.
```
dcvz, line 29: Can also mention theta needing to be positive

ashdtu: Thanks @dcvz, added this as well.

louisfd (approved):
LGTM, very well documented thanks

PR #1606, opened 2024-04-12
Fix missing device in custom training loop book example

### LINKED ISSUE #1605: The Burn book using outdated api?
The current version of the burn book calls CrossEntropyLoss::new with only a single argument where that method expects a second, being the device.  Also the init method the book previously prescribes requires the device to be passed where none is.  Is this intentional to be covered later in the guide (I skimmed it first but not in any great detail), or is this an actual problem (I tried just building but it would be useless to run until complete so I don't know if anything is wrong)?

The documentation is awesome btw

laggui:
### Checklist

### Related Issues/PRs

Fixes #1605

### Changes

Updated the book according to the latest API, as used in the actual example code.


louisfd (approved):
LGTM

PR #1616, opened 2024-04-12
Remove unused assets

laggui:


louisfd (approved):
I think it was there for historical reasons

PR #1622, opened 2024-04-15
Docs/update contributor book

agelas:
## Pull Request Template
Some of the links in the contributor book were stale/didn't work. This should update the contributor section in `Adding a new operation to burn` to links using the latest commit (or close to it) off main. The section on adding an op to fusion/wgpu also needed to be updated to include the jit crate, which I don't think was around when the contributor book was first created.

### Checklist
N/A

### Related Issues/PRs
N/A

### Changes
Updates to various links, as well as instruction updates.


agelas:
Also, is there a difference in purpose between `contributor-book/src/guides/frequently-encountered-issues/` and `contributor-book/src/frequently-encountered-issues/`? They have the same name and the stuff inside is named the same, but one (the guides version) is basically empty.

louisfd (approved):
Thanks @agelas 
I think the auto formatter adds a break line at the wrong place which messes the rendering of footnotes, but I'll merge it anyway because I'll be working on this today and I'll fix them

Issue #1560, opened 2024-04-01
ONNX conversion: Only tensor input is valid Argument


AdrianEddy:
**Describe the bug**
When trying to convert onnx model to Burn, I'm encountering this panic:
```
  WARN burn_import::onnx::dim_inference: Must implement dimension inference for Slice
  WARN burn_import::onnx::dim_inference: Must implement dimension inference for Slice
  ERROR burn_import::logger: PANIC => panicked at burn-import-0.12.1\src\onnx\dim_inference.rs:311:9:
  Only tensor input is valid Argument

  --- stderr
  thread 'main' panicked at burn-import-0.12.1\src\onnx\dim_inference.rs:311:9:
  Only tensor input is valid Argument
```
The `node_input` on that line is `{ name: "cast2_out1", ty: Scalar(Int64), value: None, passed: true }`

The model I'm trying to convert is [gmflow-scale2-regrefine6-mixdata-train320x576-4e7b215d.onnx](https://drive.google.com/file/d/1raYbnf8nnxg9mb4W6bg66PfkB1woK8QB/view?usp=sharing)

I also have alternative onnx model exported by someone else, which errors with a different message, might also be worth looking into: [gmflow-scale1-mixdata-train320x576-4c3a6e9a_1x3x480x640_sim.onnx](https://drive.google.com/file/d/1ARg0gCvOouXyHEuUgDxCJ_lRbbezbl6A/view?usp=sharing)

**To Reproduce**
1. Create a build.rs with:
```rust
fn main() {
    ModelGen::new()
        .input("gmflow-scale2-regrefine6-mixdata-train320x576-4e7b215d.onnx")
        .out_dir("model/")
        .run_from_script();
}
```
2. Watch it fail

**Expected behavior**
I expected it to finish conversion successfully

**Desktop (please complete the following information):**
 - OS: Windows 11, Rust 1.77.1, Burn 0.12

**Additional context**
Model exported using `torch.onnx.export` from [this repo](https://github.com/autonomousvision/unimatch/)

laggui:
According to the line that panics it seems to be due to an `Unsqueeze` op.

Looks like this op is not full supported yet (#600).

Do you know what else is missing for the full support @antimora ?

antimora:
Thanks, @AdrianEddy, for filing this issue. This helps us prioritize the ONNX ops. 

OK. I have checked the ONNX file and it contains lots of nodes (> 10K) but with a few types:

1. Add
2. BatchNormalization
[...]

PR #1634, opened 2024-04-15
[ONNX] Add not op and extend cast support to tensors

laggui:
### Related Issues/PRs

Progress towards #1560 (should fix cast op error encountered by OP)

### Changes

- Added not ONNX op import support
- Added cast ONNX op import support

### Testing


laggui:
Clippy fails because the generated code for the test case has too many outputs.. not sure if we should add `#allow(clippy::type_complexity)]` to the forward 🤔

antimora:
> Clippy fails because the generated code for the test case has too many outputs.. not sure if we should add `#allow(clippy::type_complexity)]` to the forward 🤔

It seems too eager here. I think it's fine the way you have it. Lets allow.

nathanielsimard (approved):
LGTM

Issue #1544, opened 2024-03-27
Full compatibility with Segment Anything


astnmsn:
I have not found any equivalent request

### Feature description

Addition of support for necessary operators to utilize the `vit_b` SAM Model which can be found [here](https://github.com/facebookresearch/segment-anything?tab=readme-ov-file#model-checkpoints): 

I have inspected the model using Netron and compared the nodes to the support list [here](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/SUPPORTED-ONNX-OPS.md)

Below is the list of operators in the network that are partially or fully missing from the support list
Operator (missing support)

- Mul (Import)
- MatMul (Import)
- Sin (Import)
- Shape (Import)
- Expand (Full)
- Not (Import)
- ReduceMean (Import)
- ConstantOfShape (Full)
- Where (Import)
- Slice (Import)
- OneHot (Import)
- Tile (Full)
- LayerNormalization (Import)
- Gemm (Full)
- ReduceMax (Import)
- Floor (Full)

### Feature motivation

I am currently trying to load and run an onnx model [segment-anything](https://github.com/facebookresearch/segment-anything) inside an image editing app in an effort to provide a masking experience similar to the demo available [here](https://segment-anything.com/demo). I am integrating it into an existing rust codebase using wgpu that compiles to wasm and runs in the browser.

Before I can select `burn` as the ml library to support this workflow, I need to be sure that it supports the operators specified in the model.

antimora:
Thanks for filing this. This is helpful as we prioritize ONNX ops. If you have a direct link to the ONNX file, can you also link this?

antimora:
Updating Expand to Import, since we just added this op. I need to update the docs.

antimora:
Submitted a PR to fix the supported OPs document: https://github.com/tracel-ai/burn/pull/1547

astnmsn:
This is the model [download link](https://dl.fbaipublicfiles.com/segment_anything/sam_vit_b_01ec64.pth) provided by the SAM repo

antimora:
I think it might be faster to implement the model manually in Burn and load the pth weights file, which we now support.

You can check out an existing model to see how it's done: https://github.com/tracel-ai/models/tree/main/resnet-burn

We also have a YOLOX object detection PR in the works: https://github.com/tracel-ai/models/pull/24

@laggui has written a great tutorial on this subject: https://dev.to/laggui/transitioning-from-pytorch-to-burn-45m

[...]

PR #1635, opened 2024-04-15
Add not ONNX op support

laggui:
### Related Issues/PRs

Progress towards #1544 

### Changes

Added not ONNX op import support

### Testing


laggui:
`cast` is required by the tests for `not` since torch exports with a cast node. Closing in favor of #1634 which combines both

PR #1636, opened 2024-04-15
Add reduce max ONNX op support

laggui:
### Related Issues/PRs

Progress towards #1544 

### Changes

Added reduce max ONNX op import support

### Testing


antimora (approved):
LGTM. Thanks for adding the ops.

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -423,3 +424,34 @@ fn conv_transpose2d_update_outputs(node: &mut Node) {
...
+        None => false,
+    };
+
+    if dim_only {
+        node.outputs[0].ty = ArgType::Tensor(tensor);
+    } else {
+        // NOTE: ReduceMax w/o keepdims reduces to a scalar value, but Burn doesn't have
+        // 0-dim tensor so we can't track or perform other ops on that value
+        // node.outputs[0].ty = ArgType::Scalar(tensor.elem_type);
+        node.outputs[0].ty = ArgType::Tensor(TensorType { dim: 1, ..tensor });
```
nathanielsimard, line 455: We could call `flatten` ?

laggui: Not sure I understand where `flatten` could help here? The result would still have a shape of [1], we can't have a tensor without dimensions (i.e., scalar) 🤔 

This comment was mainly to follow the same as ReduceMean which was partially implemented (completed in #1637).

nathanielsimard: Okay, I thought we didn't support the `Tensor::max` operation, which returns a tensor of rank 1. Maybe I would rework the comment to say that we use a tensor of rank one for the scalar, avoiding a call to `into_scalar`. Am I mistaken?

laggui: Yes! That's more like it. Basically we had two options with the current state of Burn: either return a scalar by adding `into_scalar` (but we can't track the following operations if there are any) or we return a tensor of rank 1 with the result (as the `Tensor::max` op currently does).

Since reduce mean was already implemented to return a tensor of rank 1, I followed the same approach.

Will fix the comment.

nathanielsimard (approved):

Issue #1499, opened 2024-03-21
Document limitations of broadcasting for matmul in Burn


antimora:
The current documentation for the `matmul` operation in Burn does not clearly explain the limitations of broadcasting support. This has led to confusion among users who expect broadcasting behavior similar to that of PyTorch or NumPy.

The main limitation is that Burn requires the rank (number of dimensions) of the input tensors to be the same for broadcasting to work. This is due to constraints imposed by Rust's type system. In contrast, frameworks like PyTorch and NumPy allow broadcasting even when the input tensors have different ranks, as long as certain rules are followed.

To address this issue, we should:
1. Update the documentation for `matmul` to explicitly state that broadcasting requires the input tensors to have the same rank. 
2. Provide examples showing how to use `unsqueeze` to manually adjust the rank of tensors before performing `matmul`.
3. Explain the technical reasons behind this limitation (Rust's type system not supporting const expressions needed to return max rank).
4. Consider automatically performing `unsqueeze` in certain cases to improve API usability, while being mindful of potential unintended consequences.

Additionally, we should enhance the Burn Book by linking relevant API documentation pages to provide users with quick access to exact function signatures and usage examples. This could be automated using pre-processor plug-ins (see #1197).

By improving the documentation and clearly communicating the limitations and workarounds, we can prevent user confusion and create a more frictionless API experience, even if full PyTorch-style broadcasting is not currently feasible in Burn.

nathanielsimard:
This isn't just about matmul, but all element-wise operations can be broadcasted: mul, add, sub, div, you name it. But of course, both LHS and RHS should have the same rank.

antimora:
CCing @laggui , @ashdtu

laggui:
Right now the docs for tensors basically only mention this:

> With Burn, you can focus more on _what_ the model should do, rather than on _how_ to do it. We take the responsibility
of making your code run as fast as possible during training as well as inference. The same principles apply to broadcasting; all operations support broadcasting unless specified otherwise.

We should definitely improve that to be more explicit.

laggui:
I think we should start by tackling 1, 2 and 3.

> 4. Consider automatically performing unsqueeze in certain cases to improve API usability, while being mindful of potential unintended consequences.

Could be captured in another issue/PR. I'd start with properly documenting the current broadcasting semantics.

[...]

PR #1638, opened 2024-04-15
Add matmul ONNX op support

laggui:
### Related Issues/PRs

Progress towards #1544 

### Changes

Added matmul ONNX op import support and marked mul (already supported).

### Testing


crates/burn-import/onnx-tests/tests/matmul/matmul.py
```diff
@@ -0,0 +1,44 @@
...
+    # Export to onnx
+    model = Model()
+    model.eval()
+    device = torch.device("cpu")
+    onnx_name = "matmul.onnx"
+    a = torch.arange(24, dtype=torch.float, device=device).reshape(1,2,3,4)
+    b = torch.arange(16, dtype=torch.float, device=device).reshape(1, 2, 4, 2)
+    # b = torch.ones(1, 2, 4, 2, device=device)
+    # a = torch.randn(1, 2, 3, 4, device=device)
+    # b = torch.randn(1, 2, 4, 2, device=device)
```
nathanielsimard, line 30: Dead code?

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -423,3 +424,12 @@ fn conv_transpose2d_update_outputs(node: &mut Node) {
...
     }
 }
+
+fn matmul_update_outputs(node: &mut Node) {
+    // NOTE: matmul only supported for float tensors (no broadcasting support)
+    if let ArgType::Tensor(tensor) = node.inputs[0].clone().ty {
+        node.outputs[0].ty = ArgType::Tensor(tensor);
+    } else {
+        panic!("Only tensor input is valid");
+    }
```
nathanielsimard, line 434: We are supposed to support broadcasting, but we have to be explicit and call unsqueeze during codegen.

antimora (changes requested):
Thanks for adding this OP. 

I think we have opportunity here to fix the broadcasting issue as well. 

Burn supports a partial broadcasting (see my issue regarding this: https://github.com/tracel-ai/burn/issues/1499) where if one of the dimensions is 1, then it's broadcasted. But broadcasting won't work if dimensions do not match (2 vs 1) because Burn cannot dynamically expand static constant type (D). However, there is an easy workaround the user can apply (as I mentioned in https://github.com/tracel-ai/burn/issues/1499) - namely use `unsqueeze` operations. But the user needs to know which dimension is greater. So we can do the same here since we are just generating rust code. We just need to know if dimensions do not match, we add `unsqueeze`.

I hope it's easy change for you but with this we will be 100% ONNX compatible. The same logic can be used for other ops such as addition.

crates/burn-import/src/onnx/dim_inference.rs
antimora: Haha I was writing the same. See my comment below.

crates/burn-import/onnx-tests/tests/matmul/matmul.py
laggui: Dead code strikes again 😄

laggui:
Oh it wasn't clear even to me that this type of broadcasting was supported lol.

Will fix! Thanks for the comments.

antimora:
> Oh it wasn't clear even to me that this type of broadcasting was supported lol.
> 
> Will fix! Thanks for the comments.

Same! That's why I have added this ticket (https://github.com/tracel-ai/burn/issues/1499) to document it in details. I think many will be confused by it too.

nathanielsimard (approved):
LGTM

antimora (approved):
Thank you for completing this task and making matmul broadcastable in broadest sense. 

Also thanks for explaining difference case scenarios offline to me. We definitely need a documentation to explain Burn's broadcasting rules.

PR #1639, opened 2024-04-15
Add shape ONNX op support

laggui:
### Related Issues/PRs

Progress towards #1544 

### Changes

Added shape ONNX op import support

### Testing


laggui:
Didn't realize torch always exports the shape op with a cast.. that's causing the CI to fail.

image


PR #1634 will fix that, but will fix the included graph for this PR.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -660,3 +660,41 @@ fn padding_config_1d(pads: &[i64]) -> PaddingConfig1d {
         panic!("Padding configuration ({:?}) not supported", pads);
     }
 }
+
+pub fn shape_config(curr: &Node) -> (usize, usize) {
+    if curr.inputs.len() != 1 {
+        panic!(
+            "Flatten: multiple inputs are not supported (got {:?})",
```
antimora, line 667: Flatten => Shape

antimora (approved):
One left over comment change. Approving ahead. 

We might have to revisit the output type. I know in the standard arrays are treated as 1d tensors but it would worth if we use shape output type because when we will have to feed into a reshape operation, it would be convenient to receive in rust array type. 

For now, if your solution works for your use case, we should use it till we encounter other use cases, such as reshape. With ONNX, I am taking a conservative approach because ONNX spec is too complex to implement at once.

laggui:
> We might have to revisit the output type. I know in the standard arrays are treated as 1d tensors but it would worth if we use shape output type because when we will have to feed into a reshape operation, it would be convenient to receive in rust array type.
> 
> For now, if your solution works for your use case, we should use it till we encounter other use cases, such as reshape. With ONNX, I am taking a conservative approach because ONNX spec is too complex to implement at once.

I initially thought about returning a shape array but then when I started testing I noticed that most of the time in an ONNX model the shape for a dimension is retrieved with `gather` to use the resulting scalar on other operations. So by returning a 1d tensor as specified in the standard our `gather` import should work for this sequence too.

But I agree, importing ONNX is not straightforward.

PR #1644, opened 2024-04-16
Many superficial fixes to the contributor book

louisfd:
I did an extensive review of the contributor book. I think once this is merged it will be ready to be hosted


Cargo.toml
```diff
@@ -25,7 +25,7 @@ license = "MIT OR Apache-2.0"
 [workspace.dependencies]
 async-trait = "0.1.79"
 bytemuck = "1.14"
-candle-core = { version = "0.4.1" }
+candle-core = { git = "https://github.com/huggingface/candle.git", branch = "main" }
```
antimora, line 28: Any reason for this change?

crates/burn-candle/src/backend.rs
```diff
@@ -104,6 +105,7 @@ impl<F: FloatCandleElement, I: IntCandleElement> Backend for Candle<F, I> {
             candle_core::Device::Metal(device) => {
                 // For some reason, device.wait_until_completed() does not seem to work,
                 // and neither does writing and reading a value with into_data
+                // See this issue: https://github.com/huggingface/candle/issues/2050
```
antimora, line 108: Probably Candle related changes are not meant to be here.

antimora (changes requested):
The text looks good but there are candle related changes that were probably not meant here.

Cargo.toml
louisfd: No, my bad

louisfd:
@antimora Thanks for the review. Fixed the unrelated candle stuff and also the run-checks command which was wrong.

antimora (approved):
LGTM

PR #1647, opened 2024-04-16
Enable native sign operation for Candle backend

### LINKED ISSUE #1447: Implement sign operator for Candle backend
Candle does not have sign operator (see https://github.com/huggingface/candle/issues/1827) and we are using default implementation which is not optimized (see https://github.com/tracel-ai/burn/pull/1446). Once the Candle op is available, update the backend.

antimora:
### Checklist

### Related Issues/PRs

Fixes #1447

### Changes

1. Use candle's method
2. Use git from the head (will be revered when Candle officially releases)

### Testing

1. `run-checks.sh all`


louisfd (changes requested):
Thanks Dilshod, 
I think it's safe to pin Candle on an unofficial version because we ourselves are not gonna release for a while (and when we release it's crucial to rely on official versions). This may also allow our device sync to work as they are working on it. 
However I think we should depend on a precise git revision by adding {...rev = "abc..."} otherwise if their main is broken ours will break too

louisfd (approved):
LGTM

Issue #1523, opened 2024-03-25
Remote backend with client server


nathanielsimard:
We should create a backend that can connect to a remote backend to execute tensor operations. The remote backend can be generic over the protocol used, which the first one can be HTTP.

- [ ] Refactor burn-fusion to extract useful data structures.
  - [ ] Tensor operation descriptions
  - [ ] Tensor handles container
  - [ ] Split the `FusionBackend` trait into the `TosirBackend` (Tensor Operation Stream Internal Representation) to have the `float_tensor(handle)`, `int_tensor(handles)` and that kind of functions already present in `FusionBackend`.
- [ ] Create an http server that receive the tensor operations and execute them using an inner backend provided as generic.
- [ ]  Create the remote backend that receive the protocol (http) as generic argument, and implement the backend trait, where all operations are requests to the server.

PR #1654, opened 2024-04-17
Move HandleContainer and Tensor Ops descriptions from burn-fusion to burn-tensor

syl20bnr:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/1523

### Changes

Move `HandleContainer`  and Tensor operations descriptions to `burn-tensor` crate.
Removed the `FusionDevice` and replaced it with a `DeviceOps` trait bound to `Backend::Device`.

For now added modules to `burn-tensor` are excluded from `no-std` as they rely on `Arc`.

There are a few added docstring to check, I added the temporary prefix `--ADDED--` to them.


Cargo.toml
```diff
@@ -139,5 +139,5 @@ sysinfo = "0.30.10"
 systemstat = "0.2.3"
 
 
-[profile.dev]
-debug = 0 # Speed up compilation time and not necessary.
+# [profile.dev]
+# debug = 0 # Speed up compilation time and not necessary.
```
laggui, line 143: Are we keeping this profile or removing it? If removing, I'd tend to delete and not comment out personally.

laggui (commented):
Big refactor 😄! Overall LGTM, will wait for @nathanielsimard's final approval since he is much more familiar with the affected changes than I am.

I think the added docstrings are fine. Let's make sure we don't forget to remove the `--ADDED--` prefix from those once accepted.

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #1659, opened 2024-04-18
Fix fusion with inplace

### LINKED ISSUE #1656: If the batch size is 1, testing LSTM on the Wgpu backend will fail
**Describe the bug**
If the batch size is 1, testing LSTM on the Wgpu backend will fail.

**To Reproduce**
1. Change the batch size of the input in `test_batched_forward_pass` of LSTM from 8 to 1.
https://github.com/tracel-ai/burn/blob/2a721a9d0c73e36fabd9c9cf337bd3ca4d8d1acc/crates/burn-core/src/nn/rnn/lstm.rs#L319-L330

to
```rust
    #[test]
    fn test_batched_forward_pass() {
        let lstm = LstmConfig::new(64, 1024, true).init(&device);
        let batched_input =
            Tensor::<TestBackend, 3>::random([1, 10, 64], Distribution::Default, &device);

        let (cell_state, hidden_state) = lstm.forward(batched_input, None);

        assert_eq!(cell_state.shape().dims, [1, 10, 1024]);
        assert_eq!(hidden_state.shape().dims, [1, 10, 1024]);
    }
```

2. Run the command: `cargo test -p burn-core --features test-wgpu`. The test will fail.

louisfd:
The bug in #1656 seems to be caused by fusion when it works on several read-write tensors. 

What happened was rather subtle: if the output of the fused kernel was a mutable tensor, then it was "Readwrite", even if there were only writes to it  in this specific kernel. So while looking to match the outputs with potential inplace inputs it would not find itself and could wrongly assume that another tensor with the same shape could was it. So it did not write to the right tensor. 

Batches of one allowed to detect it, but I think the bug happened with other batch sizes as well. 

@nathanielsimard I'm not sure why this id comparison was not there before. Could adding it break something? 

Fix #1656

### DIFF
--- a/crates/burn-jit/src/codegen/compilation.rs
+++ b/crates/burn-jit/src/codegen/compilation.rs
@@ -193,7 +193,10 @@ impl CompilationSettings {
                     if chosen.is_some() {
                         break;
                     }
-                    if desc.shape == desc_input.shape && input.item() == output.item() {
+                    if desc.shape == desc_input.shape
+                        && input.item() == output.item()
+                        && desc.id == desc_input.id
+                    {
                         chosen = Some(index);
                     }
                 }
--- a/crates/burn-jit/src/codegen/compilation.rs
+++ b/crates/burn-jit/src/codegen/compilation.rs
@@ -190,7 +190,8 @@ impl CompilationSettings {
 
                 let mut chosen = None;
                 for (index, (_, desc_input, input)) in potential_inplace.iter().enumerate() {
-                    if chosen.is_some() {
+                    if desc.id == desc_input.id {
+                        chosen = Some(index);
                         break;
                     }
                     if desc.shape == desc_input.shape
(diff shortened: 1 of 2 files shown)


crates/burn-jit/src/codegen/compilation.rs
```diff
@@ -193,7 +193,10 @@ impl CompilationSettings {
                     if chosen.is_some() {
                         break;
                     }
-                    if desc.shape == desc_input.shape && input.item() == output.item() {
+                    if desc.shape == desc_input.shape
+                        && input.item() == output.item()
+                        && desc.id == desc_input.id
+                    {
```
nathanielsimard, line 199: I don't think that's right, it's not a requirement! But I think we should verify before chosing an input to replace the output if the output is also an input, in which case we have to use it.

PR #1661, opened 2024-04-18
Fix bug 1645 (Unsqueeze OpSet 11)

### LINKED ISSUE #1642: Incorrect implementation of Transpose ONNX operation
**Describe the bug**

This PR https://github.com/tracel-ai/burn/pull/546/files incorrectly implemented transpose operation. Instead of using permute operator, this uses transpose Burn operator which only flips last two dimensions.

### LINKED ISSUE #1645: Support for unsqueeze ONNX opset 1
**Describe the bug**

Current ONNX implementation does not support Unsqueeze OpSet 1 where only one input is expected and axes is passed via attributes:  https://onnx.ai/onnx/operators/onnx__Unsqueeze.html#unsqueeze-1

**To Reproduce**

1. Follow the steps to download ONNX file from this issue: https://github.com/tracel-ai/burn/issues/1593
2. Run: `[burn-import]% cr ./clip_image_model_vitb32_batchsize1.onnx ./out`

**Expected behavior**

**Screenshots**

image

image

antimora:
### Checklist

### Related Issues/PRs

Fixes #1645

### Changes

1. Add ONNX files with opset 11 & 16 for unsqueeze op
2. Add tests
3. Fix dim output dim and configuration for opset 11

### Testing

1. New tests
2. run-checks


crates/burn-import/onnx-tests/tests/unsqueeze/unsqueeze_opset16.onnx
laggui, line 1: Technically the opset is 13. If we want to support different version I would refer to the corresponding versions.

https://github.com/onnx/onnx/blob/main/docs/Operators.md#Unsqueeze

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -251,92 +251,43 @@ fn reduce_mean_update_outputs(node: &mut Node) {
...
-        _ => panic!("Unsqueeze: invalid input types"),
+            None => None,
+        }
+    } else {
+        node.attrs
+            .iter()
+            .find_map(|(key, value)| match key.as_str() {
+                "axes" => Some(value.clone().into_i64s()),
+                _ => None,
+            })
```
laggui, line 271: For now this is probably fine, but if we want to actually differentiate between opset versions for ops I think we will need to get the opset from the graph metadata.

crates/burn-import/onnx-tests/build.rs
```diff
@@ -54,6 +54,8 @@ fn main() {
         .input("tests/pow/pow.onnx")
         .input("tests/pow/pow_int.onnx")
         .input("tests/unsqueeze/unsqueeze.onnx")
+        .input("tests/unsqueeze/unsqueeze_opset16.onnx")
```
laggui, line 57: See next comment (opset should be 13 not 16)

laggui (approved):
Overall LGTM. Just a minor change regarding the actual opset for the op (should be 13 not 16). Approving in advance.

I do have a question though: do we want to start supporting multiple opset versions for importing ops? What was the reasoning behind this one? I think it can easily explode and we might just accept recent opsets. Up for discussion.

crates/burn-import/onnx-tests/tests/unsqueeze/unsqueeze_opset16.onnx
antimora: Well, it depends what we meant by OpSet 16. It corresponds to the opset version parameter passed to  `torch.onnx.export`. 16 was the latest supported by pytorch at the time (now it's 17). That's why you see 16 used by other tests and not so much what opset was targeted. And I prefer we use this scheme instead of specific op to account for multiple operators which can contain different minimal opset versions.

laggui: Hmm ok actually now that you explain the intention it makes sense. You can dismiss my initial comment regarding the opset :)

crates/burn-import/onnx-tests/build.rs
antimora: See my above answer.

crates/burn-import/src/onnx/dim_inference.rs
antimora: If we need to protect the future changes from accidentally misbehaving, we need lock into specific opset version. Open next version only after certifying with tests of next opset version. This is a full time job which we do not have resource for.

laggui: Yeah that's also what I was worried about. We don't have enough bandwidth for that 😅 but your PR adds support for an older opset so I was wondering what the motivation for that was since we might not be able to (nor want to) support all possible versions.

Thanks for addressing that offline too. Your approach makes sense.

PR #1664, opened 2024-04-19
make `find_owned_graph` iterative

AlexErrant:
### Checklist

^nnnnnope, ` ./run-checks.sh std` fails on wgpu, probably because I'm on WSL, let's see what CI thinks.

### Related Issues/PRs

https://github.com/open-spaced-repetition/fsrs-browser/issues/23

### Changes

While upgrading to 0.13.1, fsrs-browser discovered that the WASM call stack was being blown by `find_owned_graph` because it was recursive. This PR makes `find_owned_graph` iterative.

Me no write Rust good. Plz be not afraid to clean up bad style or fix incorrectness.

### Testing

None.

### DIFF
--- a/crates/burn-autodiff/src/runtime/memory_management.rs
+++ b/crates/burn-autodiff/src/runtime/memory_management.rs
@@ -145,18 +145,21 @@ impl GraphMemoryManagement {
         Some(updated_graph_id)
     }
 
-    fn find_owned_graph(&mut self, graph_id: GraphId) -> Option<GraphId> {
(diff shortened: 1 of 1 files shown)


louisfd (approved):
This is indeed equivalent yet iterative code. Thanks a lot

AlexErrant:
Some discussion here https://discord.com/channels/1038839012602941528/1059209073784008835/1231600209478422619

nathanielsimard (commented):
Not sure this actually fixes the problem with burn-autodiff, we will work on a better solution!

Issue #1593, opened 2024-04-09
`Only tensor input is valid` when converting ONNX model


BinaryWarlock:
**Describe the bug**
```
ERROR burn_import::logger: PANIC => panicked at .../burn-import-0.12.1/src/onnx/dim_inference.rs:403:9:
  Only tensor input is valid
  ```
  
Which is in `conv2d_update_outputs`.
  
Adding extra printing, it seems to be taking in a `Scalar(Float32)`. I don't know where that's from.

**To Reproduce**
Using onnx-clip model which you can grab at https://lakera-clip.s3.eu-west-1.amazonaws.com/clip_image_model_vitb32.onnx (about 336 MiB)

1. Fix the input with `python -m onnxruntime.tools.make_dynamic_shape_fixed --dim_param image_batch_size --dim_value 1 clip_image_model_vitb32.onnx clip_image_model_vitb32_batchsize1.onnx`
2. Try importing with burn-input build script
3. Receive error

antimora:
I will investigate next week.

antimora:
The immediate issue is related to casting op, which will be fixed by this PR https://github.com/tracel-ai/burn/pull/1634

Once we merge outstanding PRs related to ONNX, I'll recheck it again. 

I also noticed that our Transpose operator is not correctly implemented. Filed a ticket: https://github.com/tracel-ai/burn/issues/1642

antimora:
OK. We made progress but I found another bug that's blocking.

BinaryWarlock:
@antimora Do you happen to know which ops this model is blocked on currently?

antimora:
> @antimora Do you happen to know which ops this model is blocked on currently?

I fixed a bunch of limitations we had (mostly related to an older OpSet version) and I added a few Ops but we still lack the following.

We are missing these implementations (mentioned in #1714)

1. `ConstantOfShape`
2. `Gather`

We need to fix this 
1. Reshape (needs to accept input for the shape).

The whole list of Ops used by the ONNX file. 

1. Add
2. Cast
3. Concat
4. Constant
5. ConstantOfShape
6. Conv
7. Div
8. Gather
9. Gemm
10. MatMul
11. Mul
12. Pow
13. ReduceMean
14. Reshape 
15. Shape
16. Sigmoid
17. Softmax
18. Split
19. Sqrt
20. Sub
21. Transpose
22. Unsqueeze


Please note, we might still discover some issues once we implement the missing ops and fix Reshape.

[...]

PR #1667, opened 2024-04-19
Fix reshape bug (support for opset version 1)

antimora:
### Checklist

### Related Issues/PRs

#1593 

### Changes

1. Use shape attribute

### Testing

Pytorch supports opset 7+ so I could not recreate the ONNX file but I was able to use the onnx from #1593


crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -195,28 +195,36 @@ fn concat_update_outputs(node: &mut Node) {
...
+            None => None,
+        }
     } else {
-        panic!("Reshape: int64s shape is expected per ONNX spec");
+        node.attrs
+            .iter()
+            .find_map(|(key, value)| match key.as_str() {
+                "shape" => Some(value.clone().into_i64s()),
+                _ => None,
+            })
```
laggui, line 213: If we're just looking for a single specific attribute, couldn't we use `.get(attr)` instead?

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -195,28 +195,36 @@ fn concat_update_outputs(node: &mut Node) {
...
-        dim,
-        shape: Some(shape),
-    });
+    if shape.is_some() {
+        node.outputs[0].ty = ArgType::Tensor(TensorType {
+            dim: shape.unwrap().len(),
+            shape: None, // shape is calculated at runtime
+            ..output
+        });
+    }
```
laggui, line 227: If we unwrap the value that we're checking for, we can use `if let Some(shape) = shape`

```rust
if let Some(shape) = shape {
    node.outputs[0].ty = ArgType::Tensor(TensorType {
        dim: shape,
        shape: None, // shape is calculated at runtime
        ..output
    });
}
```

laggui (approved):
Minor comments on form, otherwise LGTM! Approving in advance.

crates/burn-import/src/onnx/dim_inference.rs
antimora: Done

crates/burn-import/src/onnx/dim_inference.rs
antimora: Done

PR #1668, opened 2024-04-20
`Arc<EventStoreClient>` to `Rc<EventStoreClient>`

AlexErrant:
### Checklist

^ nope; let's see what CI thinks

### Related Issues/PRs

None

### Changes

I saw an opportunity to convert an Arc to an Rc and took it. Is this a good idea? IDK I suck at Rust.

### Testing

None.

### DIFF
--- a/crates/burn-train/src/checkpoint/strategy/metric.rs
+++ b/crates/burn-train/src/checkpoint/strategy/metric.rs
@@ -76,7 +76,7 @@ mod tests {
         },
         TestBackend,
     };
-    use std::sync::Arc;
+    use std::rc::Rc;
 
     use super::*;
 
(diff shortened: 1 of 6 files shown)


antimora:
The change seems random. For a proper record we need a better explanation why we are making this change. Arc is used for multithreading. Are you suggesting with your changes, it's thread-safe?

AlexErrant:
I think it should be threadsafe; otherwise wouldn't Rust's type system tell us?

nathanielsimard:
We could in fact use `Rc` in the burn train crate, since most of the things is actually synchronous!

nathanielsimard (approved):

PR #1679, opened 2024-04-22
Handle ndarray matmul broadcasting

### LINKED ISSUE #1646: [ndarray] General matmul broadcasting bug
**Describe the bug**
Was double-checking our broadcasting support following #1499.

The general matmul implementation for the ndarray backend doesn't seem to compute the correct thing w.r.t. batch matrix broadcasting.

**To Reproduce**
```rust
use burn::{
    backend::{NdArray, Wgpu},
    tensor::Tensor,
};

fn main() {
    type B = NdArray;
    // type B = Wgpu;

    // (j x 1 x n x m) @ (1 x k x m x p) -> (j x k x n x p)
    let tensor_1 = Tensor::<B, 4>::ones([3, 1, 2, 4], &device);
    let tensor_2 = Tensor::<B, 4>::ones([1, 5, 4, 6], &device);
    let dims1 = tensor_1.dims();
    let dims2 = tensor_2.dims();
    let tensor_3 = tensor_1.matmul(tensor_2);
    println!("{:?} @ {:?} = {:?}", dims1, dims2, tensor_3.dims());
    // [3, 1, 2, 4] @ [1, 5, 4, 6] = [3, 5, 2, 6]
}
```
This code panics at `tensor_1.matmul(tensor_2)`:

```sh
thread '<unnamed>' panicked at /home/laggui/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ndarray-0.15.6/src/dimension/mod.rs:361:5:
collapse_axis: Index 3 must be less than axis length 3 for array with shape [3, 2, 4]
thread '<unnamed>' panicked at /home/laggui/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ndarray-0.15.6/src/dimension/mod.rs:361:5:
collapse_axis: Index 4 must be less than axis length 3 for array with shape [3, 2, 4]
```

**Expected behavior**
The matmul should result in an output with shape `[3, 5, 2, 6]`. Works on wgpu.

lancelet:
### Checklist

### Related Issues/PRs

Fixes #1646. Related to #1499

### Changes

- Use strides to map linear batch indices from the output back to the input arrays.

In `matmul`, we have arrays of shape `[l_0, l_1, ..., l_N, m, k] @ [r_0, r_1, ..., r_N, k, n]`. The dimensions `l_i` and `r_i` may be broadcast using the following (fairly standard) rules:
1. If `l_i == r_i` then the output dimension size is this size (no broadcasting).
2. If either `l_i` or `r_i` are 1, then the output dimension is `max(l_i, r_i)` and the array with the dimension equal to 1 is broadcast along axis `i`.

(The innermost two dimensions of the output are always `[m, n]`, following standard matrix multiplication rules.)

When performing the stacked matrix multiplies within the batched `matmul`, it is necessary to be able to look up which matrices should be multiplied. In this PR, that is done using a stride approach:
1. We iterate over a flattened index into the output array's batches (same as before).
2. We convert this flattened batch index into a component batch index.
3. We multiply the component batch index by batch strides for the left and right arrays to produce a flattened index into both.

This uses a standard stride lookup technique for broadcasting, where the stride for a broadcast dimension is zero.

### Testing

- Added test for a `[2, 1, 2, 2] @ [1, 2, 2, 2] -> [2, 2, 2, 2]` case which was previously returning incorrect dimensions.
- Added tests for broadcast shape and stride calculations.


crates/burn-ndarray/src/ops/matmul.rs
```diff
@@ -11,101 +12,270 @@ pub(crate) fn matmul<E, const D: usize>(
...
-
-        let mut out_array = ndarray::Array3::<E>::zeros((batch_size, m, n));
+    let out: NdArrayTensor<E, D> = run_par!(|| {
+        let mut out_array = ndarray::Array3::<E>::zeros((num_out_batches, m, n));
         let unsafe_shared_out_array = UnsafeSharedRef::new(&mut out_array);
 
-        let lhs_array = lhs.array.into_shape((batch_size_lhs, m, k)).unwrap();
-        let rhs_array = rhs.array.into_shape((batch_size_rhs, k, n)).unwrap();
+        let lhs_array = NdArray::<E>::float_reshape(lhs, Shape::new([num_l_batches, m, k])).array;
+        let rhs_array = NdArray::<E>::float_reshape(rhs, Shape::new([num_r_batches, k, n])).array;
```
lancelet, line 38: I originally tried this with just `into_shape`, but that fails in some cases with an error about an incompatible layout. The original code was also calling `float_reshape` (below), so I've re-used that here.

crates/burn-ndarray/src/ops/matmul.rs
```diff
@@ -11,101 +12,270 @@ pub(crate) fn matmul<E, const D: usize>(
...
 }
 
-fn reshape<E: FloatNdArrayElement, const D: usize>(
-    tensor: NdArrayTensor<E, D>,
-) -> NdArrayTensor<E, 3> {
-    let shape = tensor.shape();
+#[derive(Debug, PartialEq)]
+struct Strides {
+    strides: Vec<usize>,
+}
```
lancelet, line 76: The `Vec` here is a little frustrating. I gather that with nightly, it is possible to write `[usize; {D - 2}]` or similar. However, I've taken care to allocate these with a fixed capacity, so hopefully they're not too much slower than a sized array.

laggui (commented):
Thanks for your contribution! 🎉

At first glance the implementation looks good to me. Will request @nathanielsimard since he implemented the previous version.

dcvz (approved):

antimora:
It looks like we have two approvals. @nathanielsimard, when you have a chance, could you also review it?

laggui (approved):

nathanielsimard (approved):
image

No performance loss was detected for the matmul with somewhat large shapes: https://burn.dev/benchmarks/community-benchmarks/?version1=21a2c6553c7d4df0b3de830112888b39fab6a9d0&versionLabel1=Version+1&version2=-&versionLabel2=Version+2&backend=ndarray&device=All&name=All&os=Pop%21_OS+22.4.0+%28jammy%29+%5B64-bit%5D&sysHardware=Any&user=SharpPrecision&search=true

@louisfd would like your review on this before merging.

crates/burn-ndarray/src/ops/matmul.rs
```diff
@@ -11,101 +12,270 @@ pub(crate) fn matmul<E, const D: usize>(
...
+    let mut cur_r_stride: usize = 1;
+    let mut cur_o_stride: usize = 1;
+    let mut l_strides = Vec::with_capacity(D - 2);
+    let mut r_strides = Vec::with_capacity(D - 2);
+    let mut o_strides = Vec::with_capacity(D - 2);
+    for i in (0..D - 2).rev() {
+        let l_dim = lsh.dims[i];
+        let r_dim = rsh.dims[i];
+
+        // Compatible dmensions are:
```
louisfd, line 147: dimensions

crates/burn-ndarray/src/ops/matmul.rs
```diff
@@ -11,101 +12,270 @@ pub(crate) fn matmul<E, const D: usize>(
...
+        } else if l_dim == 1 {
+            o_dim = r_dim; // broadcast the left
+            l_strides.push(0);
+            r_strides.push(cur_r_stride);
+        } else if r_dim == 1 {
+            o_dim = l_dim; // broadcast the right
+            l_strides.push(cur_l_stride);
+            r_strides.push(0);
+        } else {
+            panic!("Dimensions differ and cannot be broadcast.");
```
louisfd, line 164: broadcasted

louisfd (approved):
Thanks for fixing this bug!
I spotted two typos but overall looks good to me

PR #1680, opened 2024-04-22
Add layer norm onnx op support

laggui:
### Related Issues/PRs

Progress towards #1544 

### Changes

Added layer norm ONNX op import support

### Testing


crates/burn-import/src/burn/node/layer_norm.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
+        let input = scope.tensor_use_owned(&self.input, node_position);
+        let output = &self.output.name;
+        let field = &self.field.name;
+
+        // TODO: handle self.full_precision
+        quote! {
+            let #output = self.#field.forward(#input);
+        }
```
laggui, line 106: Was thinking we could match the elem precision and call `into_full_precision()` but the backendbridge requires the backend to be specified and here this is a generic... so can't do that right now.

Issue #1689, opened 2024-04-24
 Support for 0-Dimensional Tensors in Burn


antimora:
## Motivation
Currently, the Burn deep learning framework in Rust lacks support for 0-dimensional tensors (scalars). Adding support for 0-dimensional tensors would enhance the framework's capabilities and provide several benefits:

1. **Completeness**: Supporting 0-dimensional tensors would make Burn more complete and consistent with other deep learning frameworks that already support scalars. Notably, ONNX (Open Neural Network Exchange) format, widely used for interoperability between frameworks, often deals with 0-dimensional tensors.

2. **Simplified Operations**: Many deep learning operations, such as loss functions and regularization terms, often involve scalar values. Loss functions, in particular, are crucial for training models and are typically represented as 0-dimensional tensors. Having native support for 0-dimensional tensors would simplify the implementation and usage of such operations, making it easier to compute and optimize losses during training.

3. **Interoperability**: Seamless integration with other libraries and frameworks that utilize 0-dimensional tensors would be improved, enabling smoother interoperability and data exchange. This is particularly important when working with ONNX models that frequently incorporate 0-dimensional tensors.

4. **Reduced Workarounds**: Without 0-dimensional tensor support, users may need to resort to workarounds like using 1-dimensional tensors with a single element, which can be less intuitive and efficient.

5. **Avoiding Unnecessary Data Copying**: By supporting 0-dimensional tensors directly, Burn can avoid unnecessary data copying from the device (e.g., GPU) to the host (CPU) and vice versa. This can lead to improved performance and reduced memory overhead, especially when dealing with large-scale models and datasets.

## Proposed Solution
To address this limitation, we propose the following:

1. Extend the `Tensor` struct in Burn to support 0-dimensional tensors.
2. Implement necessary methods and traits for creating, manipulating, and operating on 0-dimensional tensors.
3. Update relevant functions and operations to handle 0-dimensional tensors correctly, with a focus on loss computation and optimization.
4. Ensure proper broadcasting and type promotion rules are followed when mixing 0-dimensional tensors with higher-dimensional tensors.
5. Add comprehensive unit tests to verify the correctness and consistency of 0-dimensional tensor support, including tests specifically related to loss functions.
6. Update the documentation and examples to showcase the usage of 0-dimensional tensors.

## Benefits
By implementing support for 0-dimensional tensors, Burn will:

- Provide a more complete and consistent API for tensor operations, aligning with ONNX and other frameworks.
[...]

Issue #1688, opened 2024-04-24
Convert ONNX graph inputs of 0-dim tensors to scalars


antimora:
Found a bug for cases when the graph input is 0-dim tensors.

```rust
error[E0308]: mismatched types
    --> crates/burn-import/onnx-tests/tests/onnx_tests.rs:1071:43
     |
1071 |         let output = model.forward(input, 1.0);
     |                            -------        ^^^ expected `Tensor<NdArray, 0>`, found floating-point number
     |                            |
     |                            arguments to this method are incorrect
     |
     = note: expected struct `Tensor<NdArray, 0>`
                  found type `{float}`
note: method defined here
    --> /Users/dilshod/Projects/burn/target/debug/build/onnx-tests-9ba10161c51b05aa/out/model/unsqueeze_opset16.rs:40:12
     |
40   |     pub fn forward(
     |            ^^^^^^^
...
43   |         input2: Tensor<B, 0>,
     |         --------------------

For more information about this error, try `rustc --explain E0308`.
error: could not compile `onnx-tests` (test "onnx_tests") due to 1 previous error
```

```python
#!/usr/bin/env python3

# used to generate model: unsqueeze.onnx

import torch
import torch.nn as nn


class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()
        self.axis = 3

    def forward(self, x, scalar):
        x = torch.unsqueeze(x, self.axis)
        y = torch.unsqueeze(scalar, 0)
        return x, y


def main():
    # Set seed for reproducibility
    torch.manual_seed(42)

    torch.set_printoptions(precision=8)

    # Export to onnx
    model = Model()
    model.eval()
    device = torch.device("cpu")

    test_input = (torch.randn(3, 4, 5, device=device),torch.tensor(1.0, device=device))
    model = Model()

    output = model.forward(*test_input)

    torch.onnx.export(model, test_input, "unsqueeze_opset16.onnx", verbose=False, opset_version=16)
    torch.onnx.export(model, test_input, "unsqueeze_opset11.onnx", verbose=False, opset_version=11)

    print(f"Finished exporting model")

    # Output some test data for use in the test
    print(f"Test input data of ones: {test_input}")
    print(f"Test input data shape of ones: {test_input[0].shape}")
    # output = model.forward(test_input)
    print(f"Test output data shape: {output[0].shape}")

    print(f"Test output: {output}")


if __name__ == "__main__":
    main()

```

antimora:
[...]

PR #1690, opened 2024-04-25
ONNX support for scalar unsqueeze 

antimora:
There is case when we need to unsqueeze scalar. This came about in https://github.com/tracel-ai/burn/issues/1560. See ONNX graph screenshot down below.

### Checklist

### Related Issues/PRs

1. https://github.com/tracel-ai/burn/issues/1689
2. https://github.com/tracel-ai/burn/issues/1688
3. https://github.com/tracel-ai/burn/issues/1560

### Changes

1. Added `Ignored` new-type to ignore types in the module.
2. Updated Model module to store device
4. Updated unsqueeze to deal with scalar input type.
5. Updated tests 

### Testing

1. New unit tests
2. Tested user onnx file
3. run-checks.sh

telegram-cloud-photo-size-1-5557782741991401409-y


crates/burn-import/src/burn/node/unsqueeze.rs
```diff
@@ -17,23 +17,45 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for UnsqueezeNode {
...
+    fn register_imports(&self, imports: &mut BurnImports) {
+        match &self.input {
+            Type::Scalar(_) => {
+                imports.register("burn::tensor::ElementConversion");
+            }
+            _ => {}
+        }
+    }
+
+    //
```
laggui, line 58: Dead comment?

laggui (commented):
I'm somewhat ambivalent about these changes. Obviously, to support unsqueeze for scalars right now we have to do something like this.

But at the same time I feel like this is gonna come back to haunt us for other reasons 😅

This is just very isolated, and with the state of other ops that have different behavior for scalars (e.g., some just capture the value in a 1D tensor with a single element).

As an isolated change to support unsqueeze for scalars, this is fine. I'll leave my review as a comment for now, interested in hearing feedback from others (@nathanielsimard and yourself included)

nathanielsimard (commented):
Same comment as @laggui, I think this is fine in the short term, but supporting scalar tensors from onnx without having to sync the device is very valuable, we'll have to think about how we can do it easily.

nathanielsimard (approved):

PR #1692, opened 2024-04-25
Upgrade wgpu

### LINKED ISSUE #1546: Upgrade to wgpu 0.19
### Feature description

Upgrade to wgpu 0.19

### Feature motivation

I'm attempting to integrate into an existing rust project that uses `wgpu` and compiles to wasm.
The current version of `burn` (which depends on wgpu 0.18.0) is incompatible with my project which requires `wgpu` 0.19 because of a mismatch with `web_sys` feature `GpuComputePassTimestampWrite` because a newer version of web_sys has renamed this feature (GpuComputePassTimestampWrite -> GpuComputePassTimestampWrites). There is a similar issue in the `wgpu` project which can be found [here](https://github.com/gfx-rs/wgpu/issues/4776).

When performing a simple attempt to bump the `wgpu` dependency in `burn` to `0.19` and integrating it into my own build, I came across an issue with `WgpuServer` not being `Send`. The underlying issue seems to stem from the fact that `dyn wgpu::context::DynContext + 'static` is not send. 

I believe this is a change that occurred in wgpu in the 0.19 version. 

Note: With some very minor fixes and the `wgpu` bump to 0.19 - `burn` is able to build. But the integration into a project which targets wasm breaks.

nathanielsimard:
Fix #1546

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -796,6 +796,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
+
+[[package]]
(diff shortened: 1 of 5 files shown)


antimora (approved):
LGTM

syl20bnr (approved):

syl20bnr:
Great news! Can we include this into the next minor release ?

PR #1693, opened 2024-04-25
Refactor element type to be decoupled from runtime

laggui:
### Checklist

### Changes

Decoupled the element types from the runtime
`JitBackend<WgpuRuntime<G, F, I>>` -> `JitBackend<WgpuRuntime<G>, F, I>`

- Removed `Compiler` associative element types
  - Removed `WgslCompiler` element type generics 
- Added generics for int and float types to `JitBackend` (and the kernels where required)
- Added `FloatKind` and `IntKind` enums for different precisions
- Removed `FullPrecision` runtime and compiler types

### Testing

All unit tests have passed with run-checks

### DIFF
--- a/crates/burn-jit/src/bridge.rs
+++ b/crates/burn-jit/src/bridge.rs
@@ -1,4 +1,6 @@
-use crate::{kernel, ops::to_device, tensor::JitTensor, JitBackend, Runtime};
+use crate::{
+    kernel, ops::to_device, tensor::JitTensor, FloatElement, IntElement, JitBackend, Runtime,
+};
 use burn_tensor::{
     backend::BackendBridge,
     ops::{FloatElem, FloatTensor},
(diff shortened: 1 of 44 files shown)


crates/burn-jit/src/bridge.rs
```diff
@@ -7,26 +9,34 @@ use core::marker::PhantomData;
...
     ROrigin: Runtime,
+    FOrigin: FloatElement,
+    IOrigin: IntElement,
     RTarget:
         Runtime<Device = ROrigin::Device, Server = ROrigin::Server, Channel = ROrigin::Channel>,
+    FTarget: FloatElement,
+    ITarget: IntElement,
 {
-    type Target = JitBackend<RTarget>;
+    type Target = JitBackend<RTarget, FTarget, ITarget>;
```
nathanielsimard, line 30: It should be the same runtime, so in fact the `PrecisionBridge` should not be generic over the runtime trait.

nathanielsimard (changes requested):
Only one change, but LGTM

crates/burn-jit/src/bridge.rs
laggui: Whoops! That makes a lot of sense, the runtime generic is totally obsolete now that the element types are decoupled 😅

PR #1710, opened 2024-04-29
Autodiff Memory Management: BFS

### LINKED ISSUE #1702: loss.backward() hangs after burn update 0.12 -> 0.13
**Describe the bug**
`loss.backward()` started to hang in my tests for both ndarray and wgpu.

**To Reproduce**
1. `git clone https://github.com/pointsgame/oppai-rs.git -b burn && cd oppai-rs/zero-burn`
2. `cargo test --release -- train_ndarray`
3. It hangs in [this line](https://github.com/pointsgame/oppai-rs/blob/d967bd0cc881afec108c9c2878c35fb003098e4c/zero-burn/src/model.rs#L241).

**Desktop:**
 - OS: linux

**Additional context**
I haven't tried to extract it to a minimal example, but the test is pretty straightforward. Say if you need it.

louisfd:
Use breadth first search algorithm instead of pure recursivity in autodiff memory management, because nodes could be visited recursively way too many times in some settings. 
Fix #1702


crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +115,109 @@ impl GraphMemoryManagement {
...
+    fn useful_propagation(&mut self, leaf_id: NodeID) {
+        let mut visited = HashSet::new();
+        let mut to_visit = Vec::new();
 
-        match mode {
-            Mode::TagAsUseful => {
-                self.statuses.insert(node_id, NodeMemoryStatus::Useful);
-                for parent in parents {
-                    self.useful_propagation(parent, Mode::TagAsUseful)
+        to_visit.push((leaf_id, Mode::Explore));
```
nathanielsimard, line 122: `let mut to_visit = Vec::from([(leaf_id, Mode::Explore)]`; ?

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +115,109 @@ impl GraphMemoryManagement {
...
-                to_delete.push(node_id);
-            }
+                _ => {
+                    to_delete.push(node_id);
+
+                    for parent in self
+                        .nodes
+                        .get(&node_id)
+                        .cloned()
+                        .unwrap_or(vec![])
```
nathanielsimard, line 205: `unwrap_or_default()`

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +115,109 @@ impl GraphMemoryManagement {
...
+                    let node_status = self
+                        .statuses
+                        .get(&node_id)
+                        .expect("All nodes should have received a status at this point")
+                        .clone();
+
+                    match node_status {
+                        NodeMemoryStatus::Useful => {
+                            // Nothing to do, was already tagged through some other path
+                            continue;
```
nathanielsimard, line 142: Not obvious why we don't have to explore the parents in this case.

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +115,109 @@ impl GraphMemoryManagement {
         }
     }
 
-    fn useful_propagation(&mut self, node_id: NodeID, mode: Mode) {
-        let parents = self.nodes.get(&node_id).cloned().unwrap_or(vec![]);
+    fn useful_propagation(&mut self, leaf_id: NodeID) {
+        let mut visited = HashSet::new();
```
nathanielsimard, line 119: visited could be allocated with the size of the graph I think.

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +115,109 @@ impl GraphMemoryManagement {
...
+            for parent in self
+                .nodes
+                .get(&node_id)
+                .cloned()
+                .unwrap_or(vec![])
+                .into_iter()
+            {
+                if !visited.contains(&parent) {
+                    to_visit.push((parent, next_mode.clone()));
+                }
```
nathanielsimard, line 171: As I understand it, we can still visit multiple times the same parent with different modes, but once we visited the parent once, we can't register new modes. Wouldn't it make sense to register the modes as well with priority (If `TagAsUseful` > `Explore`)?

The `visited` could be an `HashMap` containing the mode used. You can't register a parent with `Explore` when the parent was already visited, but you can with the mode `TagAsUseful` if the parent was visited with the mode `Explore`?

louisfd: Unlike the NodeMemoryStatus for which there is one status per node (at first Unknown, then maybe Unavailable after the first propagation, then maybe Useful after this propagation), there is not one Mode per node. 

The Mode made more sense in the previous form of the algorithm: we started in exploration until we found a node to tag as useful, then for this node to be usable we had to tag all of its parents as useful as well, so the algorithm switched to TagAsUseful for all of this branch. But with the `visited` approach we have to remember to go back to this mode when we get to a parent of a useful node. So the mode becomes tied to nodes in the vec and it's not very elegant as it was supposed to be a status of the algorithm, not the node. I think I can remove the concept of Mode altogether, it will be clearer.

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +107,98 @@ impl GraphMemoryManagement {
...
+            for parent in self
+                .nodes
+                .get(&node_id)
+                .cloned()
+                .unwrap_or_default()
+                .into_iter()
+            {
+                if !visited_as_useful.contains(&parent)
+                    && (Some(&NodeMemoryStatus::Useful) == self.statuses.get(&node_id)
+                        || !visited_as_unknown.contains(&parent))
```
nathanielsimard, line 151: I would refactor that a bit, very hard to read:

```rust
if visited_as_useful.contains(&parents); {
    continue;
}

let is_useful = Some(&NodeMemoryStatus::Useful) == self.statuses.get(&node_id);

if is_useful || !visited_as_unknown.contains(&parents) {
   to_visit.push((parent, Some(node_id)))
}
```

And I actually think there is still a performance problem and even a bug. I think it should be:

```rust
if visited_as_useful.contains(&parents); {
    continue;
}

let is_useful = Some(&NodeMemoryStatus::Useful) == self.statuses.get(&node_id);

if is_useful {
   to_visit.push((parent, Some(node_id)))
}
```

Since the vector `to_visit` is already filled with all nodes, you don't need to keep track of nodes that were visited as unknown.

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -115,81 +107,98 @@ impl GraphMemoryManagement {
...
-
-        match current_status {
-            NodeMemoryStatus::Useful => {
-                new_leaves.insert(node_id);
-            }
-            _ => {
-                let parents = self.nodes.get(&node_id).cloned().unwrap_or(vec![]);
-                for parent in parents {
-                    self.identify_leaves_and_deletables(parent, new_leaves, to_delete)
+        let mut visited = HashSet::new();
```
nathanielsimard, line 165: Should be initialized with the right capacity, which is `new_leaves.len()`

nathanielsimard (changes requested):

crates/burn-autodiff/src/runtime/memory_management.rs
louisfd: I rewrote the algorithm, it's more efficient and elegant

crates/burn-autodiff/src/runtime/memory_management.rs
louisfd: Hmm `new_leaves` is dynamically inserted during this algorithm, it's length 0 at that moment. 
Also we can't really know `visited` length in advance because it depends if the nodes the algorithm sees are useful.

nathanielsimard (approved):

vampire3232 (approved):

PR #1712, opened 2024-04-30
ONNX debug improvements

antimora:
Minor ONNX debug improvements.

### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/1560

### Changes

1. Collect missing Op names and display at once.
2. Other minor improvements.

### Testing

Tested on the user's onnx file.

### DIFF
--- a/crates/burn-import/src/onnx/op_configuration.rs
+++ b/crates/burn-import/src/onnx/op_configuration.rs
@@ -569,8 +569,9 @@ pub fn reshape_config(node: &Node) -> Vec<i64> {
         panic!("Zero shape size is not supported");
     }
 
+    // TODO: check "shape" attribute
     if node.inputs.len() != 2 || node.inputs[1].value.is_none() {
-        panic!("Reshape: shape tensor must be present");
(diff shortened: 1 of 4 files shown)


crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -569,8 +569,9 @@ pub fn reshape_config(node: &Node) -> Vec<i64> {
         panic!("Zero shape size is not supported");
     }
 
+    // TODO: check "shape" attribute
```
laggui, line 572: Is this specific to an opset that's not supported yet, hence the "TODO"?

laggui (approved):
LGTM, just had a question related to one of the comments.

crates/burn-import/src/onnx/op_configuration.rs
antimora: Yes, correct. I do not have a specific use case but since we handle similar one and I noticed from the spec, we are missing it. I just added TODO as a debt, which we do not need to make a payment on.

PR #1718, opened 2024-05-01
[Fusion] Support multi-precision fusion

### LINKED ISSUE #1694: [Fusion] Support multi-precision fusion
With the end goal of supporting fusion with multi-precision:

- [ ] `FusionServer` should not depend on the `FusionBackend` (similar to `AutodiffServer`)
  - [ ] `HandleContainer` should be generic over `Handle` and `Device`
  - [ ] `MultiStream` should not be generic over the backend (similar to `HandleContainer`)
- [ ] Fusion `PrecisionBridge` should be generic over target `T`

CC @nathanielsimard

nathanielsimard:
### Checklist

### Related Issues/PRs

Fix #1694


crates/burn-fusion/src/backend.rs
```diff
@@ -101,32 +109,65 @@ pub trait OptimizationBuilder<O>: Send {
...
 
+/// Type alias for `<R as FusionRuntime>::FusionDevice`.
+pub type FusionDevice<R> = <R as FusionRuntime>::FusionDevice;
+/// Type alias for `<R as FusionRuntime>::FusionHandle`.
+pub type FusionHandle<R> = <R as FusionRuntime>::FusionHandle;
+/// Type alias for `<R as FusionRuntime>::FusionClient`.
+pub type Client<R> = <R as FusionRuntime>::FusionClient;
+
 /// Trait that allows an existing [backend](Backend) to specify graph optimizations using
 /// [operation builder](crate::OptimizationBuilder).
```
laggui, line 135: Same docstring carried over from `FusionBackend`, was this intended?

crates/burn-tensor/src/tensor/element.rs
```diff
@@ -136,56 +144,64 @@ make_element!(
     ty f64 Precision::Double,
     convert |elem: &dyn ToPrimitive| elem.to_f64().unwrap(),
     random |distribution: Distribution, rng: &mut R| distribution.sampler(rng).sample(),
-    cmp |a: &f64, b: &f64| a.total_cmp(b)
+    cmp |a: &f64, b: &f64| a.total_cmp(b),
+    dtype DType::F32
```
laggui, line 148: Shouldn't this be `DType::F64`?

laggui (commented):
Man, I need another coffee after looking at all this code ☕ 

Overall, LGTM! Seems to have really simplified things to move the types to the runtime.

Only have two minor comments below.

crates/burn-tensor/src/tensor/element.rs
nathanielsimard: Yess

laggui (approved):

Issue #1665, opened 2024-04-19
CubeCL: Compute Language Extension in Rust for Multi-target GPU kernels


louisfd:
The current JIT representation is very practical for targeting multiple GPU languages; however the kernels are very hard to read and maintain as it is composed of many declarative macros and is too close to assembly (for instance, there's no support for complex one-liners that need operation priority). 

Thus, we will be developing a new language using procedural macros, that feels like writing plain Rust but compiles to a GPU language (using a backend's compiler and runtime like we have for wgpu at the moment). We call it CubeCL (Cube Compute Language), as a reference to the topology of workgroups often present in GPU languages. 

I will be working for several weeks on this, migrating (again) and optimizing our jit kernels along the way. When it's done it must also be well-documented because it shall become the standard way of writing custom kernels in user space. 

Here's a sketch of the steps to achieve (despite lots of uncertainty still):
- [x] Static analysis of Cube basic code
- [x] Codegen of JIT representation
- [x] Support of all operators
- [x] Support of GPU-specific instructions
- [x] Migrate & optimize kernels
- [x] Write an official doc

louisfd:
Closing this, as it is now complete here: https://github.com/tracel-ai/cubecl

PR #1722, opened 2024-05-03
Add cumulative sum tensor operation

allenqm:
## Pull Request Template

Starting a draft PR to align on a few things with maintainers as I dive into this.

Context:  Per [this convo](https://discord.com/channels/1038839012602941528/1038839013735399547/1234609825619644477), I wanted to add a cumulative product operation to burn.

My plan is to start with a cumulative sum operation.  Then cumulative product can be developed using cumulative sum, log, and exp.

@nathanielsimard, Items to align on upfront:
-  The name of the function should be `cumsum_dim`.  `cumsum` aligns with the pytorch api.  In burn, operations that take an explicit dim argument seem to have a `_dim` suffix.  Alternatively we could remove the suffix.
- The function should be implemented for Float and Int tensorkinds, but not bool.
- Backends:  While the implementations for tch, candle, and ndarray seem straightforward, I have questions about jit.  For jit, I cannot find a existing WGSL implementation for cumulative sum.  Is the right approach in this situation to create a WGSL compute shader for it?  Cumulative sum is probably hard to do well GPUs given the dependencies between elements, but I'm open to trying.  I'm new to WGSL.

### Checklist

### Related Issues/PRs

### Changes

### Testing

### DIFF
--- a/crates/burn-import/src/burn/node/binary.rs
+++ b/crates/burn-import/src/burn/node/binary.rs
@@ -14,6 +14,8 @@ pub enum BinaryType {
     Equal,
     Powf,
     Powi,
+    FloatCumsum,
+    IntCumsum,
 }
 
 impl BinaryType {
@@ -26,6 +28,8 @@ impl BinaryType {
             BinaryType::Equal => "equal",
             BinaryType::Powi => "powi",
             BinaryType::Powf => "powf",
+            BinaryType::FloatCumsum => "float_cumsum",
+            BinaryType::IntCumsum => "int_cumsum",
         }
     }
 }
@@ -173,6 +177,28 @@ impl BinaryNode {
         };
         Self::new(lhs, rhs, output, BinaryType::Powi, Arc::new(function))
     }
+    pub(crate) fn float_cumsum(lhs: Type, rhs: Type, output: Type) -> Self {
+        let function = match (&lhs, &rhs) {
+            (Type::Tensor(_), Type::Scalar(_)) => {
+                move |lhs, rhs| quote! { #lhs.float_cumsum(#rhs) }
+            }
+            _ => panic!("cumsum is is supported for tensor-scalar input pairs  only"),
+        };
+        Self::new(
+            lhs,
+            rhs,
+            output,
+            BinaryType::FloatCumsum,
+            Arc::new(function),
+        )
+    }
+    pub(crate) fn int_cumsum(lhs: Type, rhs: Type, output: Type) -> Self {
+        let function = match (&lhs, &rhs) {
+            (Type::Tensor(_), Type::Scalar(_)) => move |lhs, rhs| quote! { #lhs.int_cumsum(#rhs) },
+            _ => panic!("cumsum is is supported for tensor-scalar input pairs  only"),
+        };
+        Self::new(lhs, rhs, output, BinaryType::IntCumsum, Arc::new(function))
+    }
 }
 
 #[cfg(test)]
@@ -311,6 +337,15 @@ mod tests {
     #[test]
     fn test_binary_codegen_powf_scalar() {
         test_binary_operator_on_tensor_and_scalar!(powf, powf_scalar);
+    }
+    #[test]
+    fn test_binary_codegen_int_cumsum() {
+        test_binary_operator_on_tensor_and_scalar!(int_cumsum, int_cumsum_scalar);
+    }
+
+    #[test]
+    fn test_binary_codegen_float_cumsum() {
+        test_binary_operator_on_tensor_and_scalar!(float_cumsum, float_cumsum_scalar);
     }
 
     #[test]
--- a/crates/burn-import/src/onnx/to_burn.rs
+++ b/crates/burn-import/src/onnx/to_burn.rs
@@ -278,6 +278,7 @@ impl OnnxGraph {
                     graph.register(Self::conv_transpose2d_conversion(node))
                 }
                 NodeType::Pow => graph.register(Self::pow_conversion(node)),
+                NodeType::CumSum => graph.register(Self::cumsum_conversion(node)),
                 NodeType::Unsqueeze => graph.register(Self::unsqueeze_conversion(node)),
                 NodeType::Where => graph.register(Self::where_conversion(node)),
                 NodeType::Sign => graph.register(Self::sign_conversion(node)),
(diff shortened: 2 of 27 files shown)


louisfd:
Hi @allenqm 
Although you tagged Nathaniel, I think I can answer in his stead: 
- It's great that you made a default implementation for cumprod using log and exp 👍 
- For the naming, I think we can leave the `_dim` out because there cannot be a version for all dims at once. That's what we did for instance with sort, which acts on a dimension but for which no "global sort" exists. 
- Indeed, it must be implemented for int and float only, I saw you put it in `numeric`, which is the way to go 👍 

- For JIT, it is going to be a pain at the moment. There is no WGSL code anymore, the WGSL is always auto-generated from the intermediate JIT representation (kernels using the `gpu!` macro). It's honestly a pain to work with, it's not designed for new contributors to learn. I'm working on a language to rewrite them in an accessible way, see #1665, but it's not ready. 
- For the GPU algorithm, you're right that the dependancy between elements will make it difficult to have an efficient kernel. The straightforward way would be to spawn one thread for the whole dim to sum, and this thread fills all the output spots while accumulating the inputs in a local sum variable. But for large dim to sum it can be slow. Not sure if there's better solutions, I haven't done any research. 

I'm willing to write that kernel in the JIT intermediate representation if you want, so the operation becomes available soon; then we can optimize it later and with the upcoming language.

allenqm:
@louisfd Thanks so much for the guidance.

I will remove _dim suffix.

Thanks for offering to step in and write the kernel in the JIT intermediate representation.  I'll take you up on that.

I'm going to try and get the tch, candle, ndarray, and autodiff implementations done by EoD tomorrow.

Just to be clear: I haven't written anything specific for cumprod yet.  I was proposing that if we implement cumsum, then cumprod will be more straightforward as it could be described without new backend implementations (with the exception of autodiff), using the existing implementations of cumsum, exp, and log.  Let me know if my assessment here seems off.

crates/burn-ndarray/src/ops/tensor.rs
```diff
@@ -338,6 +338,17 @@ impl<E: FloatNdArrayElement> FloatTensorOps<Self> for NdArray<E> {
         NdArrayMathOps::sum_dim(tensor, dim)
     }
 
+    fn float_cumsum_dim<const D: usize>(
+        tensor: NdArrayTensor<E, D>,
+        dim: usize,
+    ) -> NdArrayTensor<E, D> {
+        let mut array = tensor.array.clone().into_owned();
```
allenqm, line 345: I *believe* the underlying array struct of `tensor` needs to be cloned, since NdArray's method for accumulating elements along an axis modifies an array's data inplace.  Referring to [this method](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.accumulate_axis_inplace)

allenqm:
tch, candle, ndarray, autodiff + tests, and tensor tests have been added.  Going to work on the onnx section of the contributor book next.

no action needed, just fyi @louisfd

allenqm:
Sorry for not flipping this to "Ready for Review" @louisfd .  I think I've got the required onnx files in place.  Can you take a look?

crates/burn-import/src/burn/node/binary.rs
```diff
@@ -14,6 +14,8 @@ pub enum BinaryType {
     Equal,
     Powf,
     Powi,
+    FloatCumsum,
+    IntCumsum,
```
laggui, line 18: [`CumSum`](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Cumsum) is a single operator, the same for int and float.

So we should also only have one node type, `BinaryType::Cumsum`. See for example `BinaryType::Sub`, which does split the int and float tests for the generated onnx files but it's still a single operation/node.

crates/burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -770,6 +770,18 @@ pub trait IntTensorOps<B: Backend> {
     /// The sum of all elements in the tensor along the dimension.
     fn int_sum_dim<const D: usize>(tensor: IntTensor<B, D>, dim: usize) -> IntTensor<B, D>;
 
+    /// Cumulative Sum of all elements in a tensor along a dimension.
```
laggui, line 773: Let's keep the capitalization at "Cumulative sum"

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -842,6 +842,19 @@ pub trait FloatTensorOps<B: Backend> {
     /// A tensor with the sum of all elements in `tensor` along `dim`.
     fn float_sum_dim<const D: usize>(tensor: FloatTensor<B, D>, dim: usize) -> FloatTensor<B, D>;
 
+    /// Cumulative Sum of all elements in a tensor along a dimension.
```
laggui, line 845: Same thing regarding capitalization

crates/burn-tensor/src/tensor/api/check.rs
```diff
@@ -803,6 +803,22 @@ impl TensorCheck {
...
+                ops,
+                TensorError::new(format!(
+                    "Can't perform a running calculation on a tensor with ({D}) dimensions on axis ({dim})"
+                )),
+            );
+        }
+
+        check
+    }
+
```
laggui, line 821: You could use the existing `TensorCheck::dim_ops` instead

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -779,6 +780,19 @@ impl OnnxGraph {
...
+        let output = node.outputs.first().unwrap().to_type();
+        match &lhs {
+            Type::Tensor(x) => match x.kind {
+                TensorKind::Int => BinaryNode::int_cumsum(lhs, rhs, output),
+                TensorKind::Float => BinaryNode::float_cumsum(lhs, rhs, output),
+                _ => panic!("cumsum function requires LHS to be int or float type"),
+            },
+            _ => panic!("cumsum function only supports LHS tensor type"),
+        }
+    }
```
laggui, line 795: By making cumsum a single node it should simplify this block

crates/burn-ndarray/src/ops/tensor.rs
laggui: Well `float_cumsum` takes ownership of the `tensor`, so I don't think the clone is required here.

laggui (changes requested):
Just some minor comments, but overall great job 👏

antimora:
Labeled this PR as needs help completing it. It would be a great feature to have in Burn.

allenqm:
Sorry for the delay.  I am planning to devote time this Friday to make the
requested changes.

crutcher:
`cumsum` for `bool` should be equivalent to an ANY(xs) operation;
`cumprod` for `bool` should be equivalent to an ALL(xs) operation.

This is the well defined extension of the + and * monoids into boolean algebra.

laggui:
Closing in favor of #2664 (WIP), which is up to date with main.

PR #1724, opened 2024-05-05
Fix unstable tests when run concurrently

AntBlo:
### Checklist

### Related Issues/PRs

None

### Changes

`str2writer` is used to write record files, causing a race condition between optim tests. 

The following line removes the file:
https://github.com/tracel-ai/burn/blob/152509c3787dec1c4dd7d35b1b2bb8f177da0f64/crates/burn-core/src/record/file.rs#L119

And if that happens between another test creating the file, and then trying to access it, the following error occur:
```
---- optim::adam::tests::test_adam_optimizer_save_load_state stdout ----
thread 'optim::adam::tests::test_adam_optimizer_save_load_state' panicked at crates/burn-core/src/optim/adam.rs:210:14:
called `Result::unwrap()` on an `Err` value: Unknown("No such file or directory (os error 2)")
```

Changed to include the specific optimizer in the record path

### Testing

Reproduction on main can be tested using:
`burn/crates/burn-core$ while cargo test; do :; done`

### DIFF
--- a/crates/burn-core/src/optim/adagrad.rs
+++ b/crates/burn-core/src/optim/adagrad.rs
@@ -172,7 +172,7 @@ mod tests {
         BinFileRecorder::<FullPrecisionSettings>::default()
             .record(
                 optimizer.to_record(),
-                std::env::temp_dir().as_path().join("test_optim"),
+                std::env::temp_dir().as_path().join("test_optim_adagrad"),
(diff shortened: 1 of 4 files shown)


antimora (approved):
Thank you for reporting this issue and providing your fix. LGTM.

PR #1733, opened 2024-05-06
Deprecate the use of std::collections::HashMap

nakashima-hikaru:
### Checklist

### Related Issues/PRs

### Changes

Replace std::collections::HashMap and std::collections::HashSet to hashbrown.

### Testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,6 +379,7 @@ dependencies = [
  "burn-tensor",
  "burn-tensor-testgen",
  "derive-new",
+ "hashbrown 0.14.5",
  "log",
  "spin",
 ]
@@ -524,6 +525,7 @@ dependencies = [
  "candle-core",
  "derive-new",
  "half",
+ "hashbrown 0.14.5",
  "log",
  "pretty_assertions",
  "proc-macro2",
@@ -639,6 +641,7 @@ dependencies = [
  "burn-ndarray",
  "crossterm",
  "derive-new",
+ "hashbrown 0.14.5",
  "log",
  "nvml-wrapper",
  "ratatui",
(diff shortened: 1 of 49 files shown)


antimora:
What's a rational for this big change?

Arjun31415:
Hashmap uses the same algorithm as HashBrown since rust 1.36, I feel there is no need to "deprecate" it. 
Furthermore the README of hashbrown says so (the only incentive is hashbrown works in no_std environment)
https://github.com/rust-lang/hashbrown

Luni-4:
@Arjun31415 

Do you think we could use [heapless](https://docs.rs/heapless/latest/heapless/) in some places for `no_std` environment? In this way we would have a static allocation instead of a dynamic one. It would be a restrictive choices though, because with `hashbrown` you can choose the allocator

Arjun31415:
I do not understand the reason for it though ? Is it because using the current hashmap is causing some slowdown ?. 
Also, by just reading the introduction docs of heapless, we would have to handle the case where the container grows beyond its capacity ourselves right?

Luni-4:
Your analysis is correct, I would like to understand if it's possible to use `heapless` for some contexts

nakashima-hikaru:
> Hashmap uses the same algorithm as HashBrown since rust 1.36, I feel there is no need to "deprecate" it. Furthermore the README of hashbrown says so (the only incentive is hashbrown works in no_std environment) https://github.com/rust-lang/hashbrown

Sorry, my misunderstanding.
When I looked at https://github.com/pola-rs/polars, I found that std::collections::HashMap and std::collections::HashSet are prohibited.
So I misunderstood that the algorithm of hashbrown should be used instead of std::collections.

Arjun31415:
> Your analysis is correct, I would like to understand if it's possible to use `heapless` for some contexts

I'm not sure, better to ask others.

antimora:
We already use hashbrown for no-std. We conditionally swap during a build. 

Heapless is not necessary because we need heap alloc (string, vec, etc)

PR #1735, opened 2024-05-06
Indices Operator

McArthur-Alford:
### Checklist

- [ x ] Confirmed that `run-checks all` script has been executed.
- [ x ] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs
None

### Changes

Added a indices function for int tensors. This is similar to pytorches `meshgrid`, or numpys `indices` functions though with slightly different arrangement. For example, the output of `Tensor::<B, 2, Int>::indices::<3>(Shape { dims: [2, 3] }, &device);` would be:

```
[[[0, 0],  [0, 1],  [0, 2]],
 [[1, 0],  [1, 1],  [1, 2]]],
```

### Testing

Added a super basic but functional test to make sure indices produces some expected typical outputs.

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -285,6 +285,7 @@ Those operations are only available for `Int` tensors.
 | ------------------------------------------------ | ------------------------------------------------------- |
 | `tensor.arange(5..10, device)       `            | `tensor.arange(start=5, end=10, device=device)`         |
 | `tensor.arange_step(5..10, 2, device)`           | `tensor.arange(start=5, end=10, step=2, device=device)` |
+| `tensor.indices(shape, device)`                  | `torch.meshgrid(tensors)`                               |
 | `tensor.float()`                                 | `tensor.to(torch.float)`                                |
 | `tensor.from_ints(ints)`                         | N/A                                                     |
 | `tensor.int_random(shape, distribution, device)` | N/A                                                     |
--- a/crates/burn-tensor/src/tensor/api/int.rs
+++ b/crates/burn-tensor/src/tensor/api/int.rs
@@ -1,4 +1,5 @@
-use crate::{backend::Backend, Data, Float, Int, Tensor};
+use crate::{backend::Backend, Data, Float, Int, Shape, Tensor};
+use alloc::vec::Vec;
 use core::ops::Range;
 
 #[cfg(all(not(feature = "wasm-sync"), target_family = "wasm"))]
@@ -68,6 +69,57 @@ where
     /// ```
     pub fn float(self) -> Tensor<B, D, Float> {
         Tensor::new(B::int_into_float(self.primitive))
+    }
+
+    /// Produces an indices tensor for the given shape and device.
+    /// The resulting tensor contains coordinates corresponding to each element in the shape at dimension D.
+    ///
+    /// # Arguments
+    ///
+    /// * `shape` - The shape specifying the dimensions of the tensor.
+    /// * `device` - The device to create the tensor on.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `D2` is not equal to `D+1`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    ///    use burn_tensor::Int;
+    ///    use burn_tensor::{backend::Backend, Shape, Tensor};
+    ///    fn example<B: Backend>() {
+    ///        let device = Default::default();
+    ///        let result = Tensor::<B, 2, Int>::indices::<3>(Shape { dims: [2, 3] }, &device);
+    ///        println!("{}", result);
+    ///    }
+    /// ```
+    pub fn indices<const D2: usize>(shape: Shape<D>, device: &B::Device) -> Tensor<B, D2, Int> {
+        if D2 != D + 1 {
+            panic!("D2 must equal D + 1 for Tensor::indices")
+        }
+
+        let Shape { dims } = shape;
+        let mut indices: Vec<Tensor<B, D, Int>> = Vec::new();
+
+        for dim in 0..D {
+            let dim_range: Tensor<B, 1, Int> = Tensor::arange(0..dims[dim] as i64, device);
+
+            let mut shape = [1; D];
+            shape[dim] = dims[dim];
+            let mut dim_range = dim_range.reshape(shape);
+
+            for (i, &item) in dims.iter().enumerate() {
+                if i == dim {
+                    continue;
+                }
+                dim_range = dim_range.repeat(i, item);
+            }
+
+            indices.push(dim_range);
+        }
+
+        Tensor::stack::<D2>(indices, D)
     }
 
     /// Sort the elements by value in ascending order along a given dimension.
--- a/crates/burn-tensor/src/tests/mod.rs
+++ b/crates/burn-tensor/src/tests/mod.rs
@@ -97,6 +97,7 @@ macro_rules! testgen_all {
         burn_tensor::testgen_sort_argsort!();
         burn_tensor::testgen_topk!();
         burn_tensor::testgen_remainder!();
+        burn_tensor::testgen_indices!();
 
         // test stats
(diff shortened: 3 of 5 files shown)


burn-book/src/building-blocks/tensor.md
```diff
@@ -285,6 +285,7 @@ Those operations are only available for `Int` tensors.
 | ------------------------------------------------ | ------------------------------------------------------- |
 | `tensor.arange(5..10, device)       `            | `tensor.arange(start=5, end=10, device=device)`         |
 | `tensor.arange_step(5..10, 2, device)`           | `tensor.arange(start=5, end=10, step=2, device=device)` |
+| `tensor.indices(shape, device)`                  | `torch.meshgrid(tensors)`                               |
```
laggui, line 288: Two notes:

1. Usage should be `Tensor::indices(shape, device)` like `Tensor::cat` or `Tensor::empty`.
2. Given the implementation here, there is no 1-to-1 equivalent for torch since meshgrid takes tensors as input and not the shape of the desired grid. In this case this is much closer to `numpy.indices` but the indexing is in cartesian space. The actual torch equivalent for the 2D example in your unit tests would be:
```py
yv, xv = torch.meshgrid([torch.arange(2), torch.arange(2)], indexing='xy')
grid = torch.stack((xv, yv), 2)
```

So in this case, I'm not sure we would have to provide the comparison in the table.

crates/burn-tensor/src/tensor/api/int.rs
```diff
@@ -70,6 +71,57 @@ where
...
+    /// ```rust
+    ///    use burn_tensor::Int;
+    ///    use burn_tensor::{backend::Backend, Shape, Tensor};
+    ///    fn example<B: Backend>() {
+    ///        let device = Default::default();
+    ///        let result = Tensor::<B, 2, Int>::indices::<3>(Shape { dims: [2, 3] }, &device);
+    ///        println!("{}", result);
+    ///    }
+    /// ```
+    pub fn indices<const D2: usize>(shape: Shape<D>, device: &B::Device) -> Tensor<B, D2, Int> {
```
laggui, line 97: Usually indices will be produced to use with a matrix/tensor (i.e., indexing in `i, j` like [`np.indices`](https://numpy.org/doc/stable/reference/generated/numpy.indices.html)) but I think the intention with this method is to produce a grid in cartesian space. With that in mind, I think I would make it a bit more explicit in the method's doc and rename the method to something like `nd_grid`, `meshgrid` or even more explicit `cartesian_grid` (also opened to suggestions).

laggui, line 97: Instead of forcing users to pass a `Shape`, we could have the api like this:

```rust
pub fn indices<S: Into<Shape<D>>, const D2: usize,>(shape: S, device: &B::Device) -> Tensor<B, D2, Int>
```

So an array could be provided by the user and it would work.

laggui (changes requested):
Thanks for working on this 🙂 

Implementation looks good, some minor comments on form.

laggui:
@nathanielsimard what do you think about the naming for this method? (see my comments for possible suggestions)

crates/burn-tensor/src/tensor/api/int.rs
nathanielsimard: `cartesian_grid` seems to be a good name!

nathanielsimard, line 97: I think we should move this method to the backend API so that backends can optimize it, since calling a lot of arange and repeat can be very expansive for big matrices. We should keep default implementation in the backend definition.

laggui: @McArthur-Alford if you're not sure what that means, see for example the `narrow` op. It is defined by a default implementation but also overridden by some backends.

McArthur-Alford: `cartesian_grid` sounds good to me. Ill get started on moving it to a backend op.

McArthur-Alford: > @McArthur-Alford if you're not sure what that means, see for example the `narrow` op. It is defined by a default implementation but also overridden by some backends.

Just looking for some clarification on this. It seems like the backend narrow op (i'm specifically looking at `int_narrow`) isn't actually used. Instead the API has its own definition of narrow that doesnt utilize the backend at all? Obviously you can still call `B::int_narrow`,  but my assumption would be that the API should be making calls to the backend implementation, which then themselves have a default implementation?

McArthur-Alford:
Alright, everything has been renamed and moved around. Feel free to review, let me know if there are any other changes we would like. I *think* I moved it into the backend ops correctly (based on narrow).

crates/burn-tensor/src/tensor/api/int.rs
laggui: Sorry for the delayed response.

`int_narrow` has a default implementation that works across all backends

```rust
fn int_narrow<const D: usize>(
    tensor: IntTensor<B, D>,
    dim: usize,
    start: usize,
    length: usize,
) -> IntTensor<B, D> {
    narrow::<B, D, Int>(tensor, dim, start, length)
}
```

And for backends that have the op, we override the method (see `int_narrow` in `burn-candle` or `burn-tch` for example).

For this PR, you don't need to override the default implementation for any backend. This can be added later on if required.

McArthur-Alford: I was more so wondering about the tensor API narrow function. That also calls  the default narrow, rather than the backend int_narrow/bool_narrow etc. It doesn't seem like the tensor API gets overridden. In that case, overriding the backend op wouldn't actually change what the tensor API is doing right?

Regardless, I should have gotten it working much like narrow, with that last commit.

laggui: Crap, you're right! I didn't realize the default narrow was called at the tensor API level.. that means it probably never dispatches to the backend even if it's implemented 😅

burn-book/src/building-blocks/tensor.md
```diff
@@ -285,7 +285,6 @@ Those operations are only available for `Int` tensors.
 | ------------------------------------------------ | ------------------------------------------------------- |
 | `tensor.arange(5..10, device)       `            | `tensor.arange(start=5, end=10, device=device)`         |
 | `tensor.arange_step(5..10, 2, device)`           | `tensor.arange(start=5, end=10, step=2, device=device)` |
-| `tensor.indices(shape, device)`                  | `torch.meshgrid(tensors)`                               |
```
laggui: We'll still document the API for `tensor.cartesian_grid` but we can just say the pytorch equivalent is N/A.

crates/burn-tensor/src/tensor/api/int.rs
```diff
@@ -90,36 +90,15 @@ where
...
-
-            indices.push(dim_range);
-        }
-
-        Tensor::stack::<D2>(indices, D)
+    pub fn cartesian_grid<S: Into<Shape<D>>, const D2: usize>(
+        shape: S,
+        device: &B::Device,
+    ) -> Tensor<B, D2, Int> {
+        Tensor::new(B::int_cartesian_grid::<S, D, D2>(shape, device))
```
laggui, line 101: Yess 🔥

That should correctly dispatch to the backend implementations as we add them later on

laggui (commented):
Yep! Looks good to me. Just a minor comment regarding the API doc.

Otherwise, let's fix the clippy CI.

crates/burn-tensor/src/tensor/api/int.rs
McArthur-Alford: ah, ok. I thought I was going crazy. Glad I helped find a bug at least!

burn-book/src/building-blocks/tensor.md
McArthur-Alford: Fixed! 👍🏻

laggui (approved):

PR #1750, opened 2024-05-09
Refactor/jit fusion

nathanielsimard:
The goal is to make the reading and write position of each operation explicit so we can actually customize that part with further fusion.


louisfd (approved):
LGTM

Issue #1714, opened 2024-04-30
Contributing to Burn: Implementing Missing ONNX Ops


antimora:
The Burn community would greatly appreciate your help in completing the missing ONNX operations in the [`burn-import`](https://github.com/tracel-ai/burn/tree/main/crates/burn-import) crate. By contributing to this effort, you can help expand the functionality and usability of Burn for a wider range of machine learning models.

### Top Requested Ops

- [ ] Expand
- [ ] Tile
- [ ] Slice

### Relatively Easy Ops (Similar to Existing Implemented Ops)

- [ ] ArgMax
- [ ] AveragePool1d
- [ ] ConvTranspose1d
- [ ] Greater
- [ ] GreaterOrEqual
- [ ] Less
- [ ] LessOrEqual
- [ ] Max
- [ ] MaxPool1d
- [ ] Mean
- [ ] Min
- [ ] PRelu
- [ ] Pad
- [ ] Range
- [ ] ReduceMin
- [ ] ReduceProd
- [ ] ReduceSum
- [ ] Squeeze
- [ ] Sum
- [ ] TopK
- [ ] Trilu

### Harder Ops (Not Similar to Existing Implemented Ops)

- [ ] GRU
- [ ] GroupNormalization
- [ ] If
- [ ] InstanceNormalization
- [ ] LSTM
- [ ] MatMulInteger
- [ ] OneHot
- [ ] RNN
- [ ] RandomNormal
- [ ] RandomNormalLike
- [ ] RandomUniform
- [ ] RandomUniformLike
- [ ] Resize
- [ ] Scatter

For a comprehensive list of all supported ONNX Ops, please refer to the [SUPPORTED-ONNX-OPS.md](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/SUPPORTED-ONNX-OPS.md) file in the Burn repository.

## Getting Started

To begin contributing, please follow the instructions in the contributor book:

```sh
# Build the book and open in a browser
cargo xtask books contributor open
```

Once the book is built, navigate to [http://localhost:3011/guides/onnx-to-burn-conversion-tool.html](http://localhost:3011/guides/onnx-to-burn-conversion-tool.html) for detailed guidance on how to implement missing ONNX operations in the `burn-import` crate.

## Related Issues Submitted by Users

Several users have submitted issues related to missing ONNX operations. These issues can provide valuable context and help prioritize the implementation of specific ops:

[...]

PR #1753, opened 2024-05-11
Squeeze Onnx Import

agelas:
### Checklist

### Related Issues/PRs

#1714 

### Changes

Added the squeeze operation for to `burn-import`

### Testing

- onnx_tests
- codegen_nodes test


crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -971,3 +971,31 @@ pub fn transpose_config(curr: &Node) -> Vec<i64> {
 
     perm
 }
+
+pub fn squeeze_config(curr: &Node) -> Vec<i64> {
+    for (key, value) in curr.attrs.iter() {
+        match key.as_str() {
+            "axes" => return value.clone().into_i64s(),
+            _ => {}
```
antimora, line 979: Great! This will support [ONNX OPset 13 ](https://onnx.ai/onnx/operators/onnx__Squeeze.html#squeeze-13)

Can we, to be sure it works, add a unit test for OPset 13? This is similar to unsqueeze with opset 16 and 13.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -971,3 +971,31 @@ pub fn transpose_config(curr: &Node) -> Vec<i64> {
...
+        ArgType::Tensor(tensor) => {
+            assert_eq!(tensor.dim, 1, "Squeeze: axes tensor must be 1D");
+            if let Some(Data::Int64s(shape)) = input_value.value.as_ref() {
+                shape.clone()
+            } else {
+                panic!("Tensor data type must be int64")
+            }
+        }
+        _ => panic!("Arg for squeeze must be tensor or scalar"),
+    }
```
antimora, line 1000: `axes` can contain negative values, which means counting dimensions from the back (see [squeeze spec](https://onnx.ai/onnx/operators/onnx__Squeeze.html#inputs)), and Burn squeeze only supports positive values (see [doc](https://burn.dev/docs/burn/prelude/struct.Tensor.html#method.squeeze)). So we should account for this.

We are already doing this for one dimension (see [code](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/onnx/op_configuration.rs#L335-L338)) for `gather_config`, so you can see the logic. We need to do this for all items in `axes`.

antimora (changes requested):
Thank you so much for completing a missing ONNX op in Burn. One down!

You've got almost perfect. We just need a couple fixes:

1. Add test case to verify axes by adding ONNX OpSet 13 (please see my inlined comments)
2. Account for negative values in axes.

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -266,6 +267,33 @@ fn reduce_mean_update_outputs(node: &mut Node) {
     }
 }
 
+/// Update the output tensor dimension
+fn squeeze_update_output(node: &mut Node) {
+    if let Some(Data::Int64s(axes)) = &node.inputs[1].value {
+        if axes.len() != 1 {
+            panic!("Squeeze: Only one axis should be specified for squeezing.");
```
laggui, line 274: @antimora do we want full squeeze support? Multiple axes can be specified according to [the spec](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Squeeze-13), which would translate to multiple `.squeeze(dim)` calls in Burn code.

laggui (commented):
Thanks for your contribution! I have the same comments as already mentioned, and one question/comment regarding the squeeze op spec support.

This PR adds support for squeeze op on a single dim, so support is not complete but as long as it is explicit I have no issues adding full support if needed in another PR.

crates/burn-import/src/onnx/dim_inference.rs
antimora: Good point. It skipped me too.

agelas: The `unsqueeze` op has `unsqueeze`, `unsqueeze_dim`, and `unsqueeze_dims` in burn. If we want to follow the same strategy, we could add a `squeeze_dims` op in burn, although that would probably belong in another PR as @laggui said.

crates/burn-import/src/onnx/op_configuration.rs
agelas: Unsqueeze uses OPset 16 and 11. I think my original one used 16, so I added one for 13. Is that what you meant?

antimora: Yes, both should be good. OpSet 13 has `axes` attribute so it should work.

crates/burn-import/src/onnx/dim_inference.rs
laggui: Sounds good. We can start with partial support (only one squeeze axis) with this PR.

crates/burn-import/onnx-tests/tests/squeeze/squeeze.onnx
laggui, line 1: Any reason why this is still present if we split the test for opset 13 and 16?

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -266,6 +267,33 @@ fn reduce_mean_update_outputs(node: &mut Node) {
...
+        }
+    } else {
+        panic!("Squeeze: Axes input must be an integer list.");
+    };
```
laggui, line 278: If we're checking for axes in the attributes in `squeeze_config` to support another opset we need to check here too to make sure the output is properly adjusted.

See the `unsqueeze_update_output` function which is doing something similar (though it captures the axes to support unsqueeze on multiple axes).

laggui (commented):
Thanks for the changes! I have two minor comments and then it should be good to go! 🙂

crates/burn-import/onnx-tests/tests/squeeze/squeeze.onnx
agelas: The short answer is that it breaks if it's not there lol.

The longer answer, and I'm not 100% sure on this, is that in `to_burn.rs` it looks like we take the .onnx file of whatever name the Node points to, in this case "squeeze.onnx" to generate the model. The exact same thing happens with unsqueeze, there's an "unsqueeze.onnx", "unsqueeze_opset16.onnx", and an "unsqueeze_opset11.onnx". I think the _opset\<number\> files are just there to make sure it's compatible, but the default behavior is inside "unsqueeze.onnx". Otherwise, I think we'd have to rename the nodes to unsqueeze_opset\<whatever\> which will get confusing.

laggui: Oh, I didn't realize unsqueeze had the same thing.

Forget my comment!

crates/burn-import/src/onnx/dim_inference.rs
agelas: Sounds good, should be a little closer to what you had in mind now!

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -266,6 +267,41 @@ fn reduce_mean_update_outputs(node: &mut Node) {
...
+        match &node.inputs[1].value {
+            Some(value) => match value {
+                Data::Int64s(axes) => Some(axes.clone()),
+                _ => panic!("Squeeze: invalid input types"),
+            },
+            None => None,
+        }
+    } else {
+        node.attrs.get("axes").cloned().map(|v| v.into_i64s())
+    };
```
laggui, line 282: LGTM! 🙂 

We lost the check to make sure a single axis is specified (since we don't support multiple axes in this PR).

Could you add that back? Possible just below with the `axes.is_none()` condition, e.g. :

```rust
if axes.is_none() {
    panic!("Squeeze must specify an axis");
} else if axes.len() > 1 {
    // some panic message
}

laggui (commented):
Great! Thanks for making the changes. I only have one minor comment left (we lost a check with the last committed change and I think we should keep it).

Should be good to go after that!

crates/burn-import/src/onnx/dim_inference.rs
agelas: Done!

laggui (approved):
Thanks for going through the changes!

LGTM 🚀

laggui:
Hi @mtobin-tdab 👋 

The ONNX spec is quite large so we don't necessarily aim to cover all the edge cases when adding support for an operator. That's why this PR was merged - we just make sure to panic when there's is a case which is not supported yet (such as yours).

You can definitely open an issue to ask for support for multiple axes. That's what issues are for (bugs, feature requests, etc.).

In the meantime, if you can generate your model with multiple squeeze nodes instead then yes that should work!

agelas:
@mtobin-tdab I think `squeeze_dims` added here in https://github.com/tracel-ai/burn/pull/1779 is what you might be looking for.

agelas:
Ah wait, actually I think I see the issue. It looks like I didn't update the dimension inference, sorry about that!

laggui:
Oh I didn't even see the follow-up PR for `squeeze_dims` 😄 looks like you've correctly identified the issue

PR #1771, opened 2024-05-16
Contributor Book: Onnx to Burn Conversion

agelas:
### Checklist
n/a

### Changes

The previous instructions in onnx-to-burn-conversion.md for **Implement Missing Operators** was basically "implement it", which threw me for a loop when I realized too late that this was a tad more involved than originally anticipated. I think the contributor book could use some more documentation/lip service around this topic.

Please let me know if something doesn't look right/edits are in order/I got something completely wrong. I'll add links later after PR #1753 is merged.

### DIFF
--- a/contributor-book/src/guides/onnx-to-burn-conversion-tool.md
+++ b/contributor-book/src/guides/onnx-to-burn-conversion-tool.md
@@ -234,8 +234,7 @@ the graph can rely on correct tensor sizes, which is critical for both compiling
 runtime execution efficiency.
 
 If something is amiss (ie weird panics are happening), after doing this step and the dimensions of
-your output tensor differs from the dimensions of your input, see the warning at the end of this
-section.
+your output tensor differs from the dimensions of your input, see the warning at the very end.
 
 ### Step 6: Integrate into the Graph Building Process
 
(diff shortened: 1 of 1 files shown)


nathanielsimard (approved):
Wow, that is a very well-written and detailed documentation on how to add new operations to ONNX. Tagging @laggui to make sure I didn't miss anything while reviewing 😅

contributor-book/src/guides/onnx-to-burn-conversion-tool.md
```diff
@@ -73,6 +75,193 @@ To extend `burn-import` with support for new ONNX operators, follow these steps:
...
+- `output_types` and `input_types` return the tensor (or element) types for the output and inputs of
+  the node, respectively.
+- `forward` generates the Rust code that performs the operation during the execution phase. The
+  `quote!` macro is used to generate rust code. Ensure that this is syntactically correct using Burn
+  code.
+- `into_node` wraps the specific node in a general `Node` type, facilitating its inclusion in the
+  broader Burn graph structure.
+
+This file is also where you would put `test_codegen_nodes()`, to make sure that the generated code
+works within the Burn library.
```
laggui, line 115: Awesome!

Can we add a small asterisk for binary and unary ops? The codegen is mostly implemented for these nodes in `binary.rs` and `unary.rs`, and each new op only has to define a method to execute the function on the input(s).

No need to detail the process as much here, just a small note that it is slightly different for those but it works in the same way.

laggui (commented):
Wow that's really awesome! 🔥 🔥 

I had only a minor comment below, otherwise it looks great.

contributor-book/src/guides/onnx-to-burn-conversion-tool.md
agelas: @laggui Absolutely, will add!

laggui (approved):

PR #1779, opened 2024-05-19
Feat/squeeze dims

agelas:
### Checklist

### Related Issues/PRs

#1780 

### Changes

Adds a new `squeeze_dims` function that's more spec compliant with ONNX and PyTorch. Specifically, users can pass in multiple dimensions, negative indices, or no dimensions at all, in which case the function will squeeze all singleton dimensions.  The `burn-import` crate has also been updated to use this.

### Testing

Added unit tests to `burn-tensor`

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -300,8 +300,8 @@ where
 
     /// Removes specified dimensions of size 1 from a tensor's shape. This function takes a tensor and
     /// an array of dimensions (`dims`) to be squeezed. If `dims` is provided, only the dimensions
-    /// specified in this array will be removed. Each dimension in `dims` must correspond to a size of 1
-    /// in the tensor; otherwise, the function will panic. If `dims` is empty, all single-dimensional entries
+    /// specified in this array will be removed. Each dimension in `dims` should correspond to a size of 1
+    /// in the tensor; otherwise, the dimension will not be squeezed. If `dims` is empty, all single-dimensional entries
     /// in the tensor will be removed. If entries in `dims` are negative, then dimensions will be counted
     /// from the back.
     ///
@@ -315,7 +315,7 @@ where
     ///
     /// # Returns
     ///
-    /// A new `Tensor<B, D2, K>` instance with the specified dimension removed.
+    /// A new `Tensor<B, D2, K>` instance with the specified dimensions removed.
     ///
     /// # Example
     ///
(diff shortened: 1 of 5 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -298,6 +298,95 @@ where
...
+        ));
+
+        // Calculate new dimensions
+        let mut new_dims = Vec::new();
+        for (index, &dim_size) in current_dims.iter().enumerate() {
+            // Exclude the dimension if it's explicitly marked for squeezing and is of size 1
+            if dim_indices.contains(&index) && dim_size == 1 {
+                check!(TensorCheck::squeeze::<D2>(index, &current_dims));
+                continue;
+            }
```
agelas, line 380: @nathanielsimard this check doesn't do much of anything right now since the `==1` basically means there's no chance of it erroring out. The `squeeze()` function will panic if you ask it to squeeze a dimension that isn't 1. For this, I'm not sure if you want to keep the same behavior as `squeeze()` or not. It looks like the ONNX spec says it will raise an error if you try squeezing a non-1 dimension. But PyTorch would just leave that dimension unchanged.

louisfd: In my opinion it should behave like our `squeeze`, so panic on non-1 dimensions

crates/burn-tensor/src/tests/ops/squeeze.rs
```diff
@@ -35,6 +35,58 @@ mod tests {
...
+    #[test]
+    fn should_squeeze_dims_with_negative_indices() {
+        let tensor = Tensor::<TestBackend, 4>::ones(Shape::new([2, 1, 3, 1]), &Default::default());
+        let squeezed_tensor: Tensor<TestBackend, 2> = tensor.squeeze_dims(&[-3, -1]);
+        let expected_shape = Shape::new([2, 3]);
+        assert_eq!(squeezed_tensor.shape(), expected_shape);
+    }
+
+    /// Test to make sure the function doesn't do anything if a non-singleton dimension is squeezed
+    #[test]
```
louisfd, line 66: Following my other comment, this test should panic

louisfd (changes requested):
Hi @agelas 
Thanks a lot, it's very well documented and tested. I think we should panic on non-1 dimensions; that would be the only thing to change.

crates/burn-tensor/src/tests/ops/squeeze.rs
agelas: Sounds good, I'll make the change.

louisfd (approved):
LGTM

PR #1788, opened 2024-05-20
Add a feature to initialize from an existing wgpu adapter/device/queue

ArthurBrussee:
### Checklist

### Related Issues/PRs
Some Discord discussions: https://discord.com/channels/1038839012602941528/1229453567694209105/1230282427629441075

### Changes

This is useful when interacting with other wgpu applications (eg. displaying a burn tensor as a texture in egui). The existing devices are keyed by the wgpu Device ID. Alternatively they could be keyed per adapter which would be more inline with other burn WgpuDevice's (one per adapter), but also there's no real inherent reason to.

This also involves making Queue into an Arc. Alternatively, this could give up ownership of the queue, but it's helpful to be able to synchronize burn operations and custom wgpu operations.

Nb: Even with this feature it's still tricky to say, copy a burn buffer to a wgpu texture, as there's no access to wgpu bindings in the public Burn API. I've got a feature for that locally but not sure what the best way is, might open an issue for it.

### Testing

run-checks all, using this feature in some local development.

### DIFF
--- a/crates/burn-wgpu/src/compute/server.rs
+++ b/crates/burn-wgpu/src/compute/server.rs
@@ -18,7 +18,7 @@ use wgpu::{
 pub struct WgpuServer<MM: MemoryManagement<WgpuStorage>> {
     memory_management: MM,
     device: Arc<wgpu::Device>,
-    queue: wgpu::Queue,
+    queue: Arc<wgpu::Queue>,
     encoder: CommandEncoder,
     pipelines: HashMap<String, Arc<ComputePipeline>>,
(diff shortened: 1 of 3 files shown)


ArthurBrussee:
Just checking on this! If this doesn't seem like the way to go I'm very happy to consider other approaches to initializing from existing WGPU devices, or the larger goal of WGPU interop in general. Thanks again!

nathanielsimard (approved):
This looks really clean 👏 thanks

PR #1795, opened 2024-05-21
Completely separate Argument and Node handling in burn import

skewballfox:
### Checklist

### Related Issues/PRs

None, I was working on generalizing some of the burn-import code for luminal (and others), and I realized this would help. 

### Changes

One thing that was proving difficult during the initial refactor is having some argument data in Nodes, and some in graph IO.  a few functions such as lifting constants and dim inference were prone to break frequently during the process of refactoring burn-import.

This completely separates argument storage and operation from the nodes struct and moves it to graph_io. Now, Node only stores the original Onnx name for the argument which acts as a key. All changes to arguments happen within graph_io. 

This also various cleanup efforts to the graph builder: unsqueeze handling has been moved to dim inference, identity handling was removed, constant handling was reworked.  

I think graph_io should be moved either to ir.rs or into it's own file, but would rather wait on confirmation for that.

### Testing

I'm testing with `onnx-test` and the image-classification-web example


antimora:
I have some concerns that the new API may be more complex for users to add new ops, which was a key reason for introducing the IR. Could we explore ways to simplify the API while still achieving our other goals? I'm open to discussing this further to find the optimal solution.​​​​​​​​​​​​​​​​

P.S. I am still traveling so I will be slow to respond.

skewballfox:
I just did a push which made the changes we talked about yesterday: I made a second node type which stored arguments and converted the nodes at the final step of parsing when creating the onnx graph. This keeps the downstream api unchanged.

I'll merge in the changes from main next.

I can try another design where more work is done when during proto conversion, and I might be able to make it work so that storing argument separately can be avoided entirely. Not sure if that should be this PR or another one, given that making it work requires mutations to be strictly forward (node inputs can be added or replaced but not mutated), and honestly I don't know if that's the case.

skewballfox:
Closing in favor of #1857

PR #1797, opened 2024-05-22
Cube: first ported kernel + comptime support + variable reuse + cleanup

louisfd:
The convolution2d kernel has been ported to Cube


nathanielsimard (approved):
LGTM !

Issue #1805, opened 2024-05-23
Burn-WGPU tests fail on Windows with Radeon 6950


hexd0t:
**Describe the bug**
Running `run-checks.ps1 all` fails for burn-wgpu with 4 failed tests during the std tests.

**To Reproduce**
Checkout the current `main` branch.
Open a Powershell Terminal, cd to the repo root, execute `. .\run-checks.ps1 all`

**Expected behavior**
All tests should pass.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: Server 2025 Preview (build 26212.5000), equivalent to Win 11
 - GPU: Radeon 6950, Driver 31.0.24027.1012 / AMD Software 24.3.1
 - rustc 1.78.0

**Additional context**

```
failures:

---- tests::jit::module_nearest_interpolate::tests::test_downsample_interpolation stdout ----
thread 'tests::jit::module_nearest_interpolate::tests::test_downsample_interpolation' panicked at crates\burn-wgpu\src\lib.rs:75:5:
Tensors are not approx eq:
  => Position 3: 7 != 6 | difference 1 > tolerance 0.0010000000000000002
  => Position 9: 161 != 160 | difference 1 > tolerance 0.0010000000000000002
  => Position 15: 315 != 314 | difference 1 > tolerance 0.0010000000000000002
  => Position 21: 469 != 468 | difference 1 > tolerance 0.0010000000000000002

---- tests::jit::remainder::tests::should_be_zero stdout ----
thread 'tests::jit::remainder::tests::should_be_zero' panicked at crates\burn-wgpu\src\lib.rs:75:5:
Tensors are not approx eq:
  => Position 0: 0 != 3.5 | difference 3.5 > tolerance 0.0010000000000000002
  => Position 1: 0 != 3.5 | difference 3.5 > tolerance 0.0010000000000000002
  => Position 2: 0 != 3.5 | difference 3.5 > tolerance 0.0010000000000000002

---- tests::jit_fusion::module_nearest_interpolate::tests::test_downsample_interpolation stdout ----
thread 'tests::jit_fusion::module_nearest_interpolate::tests::test_downsample_interpolation' panicked at crates\burn-wgpu\src\lib.rs:75:5:
Tensors are not approx eq:
  => Position 3: 7 != 6 | difference 1 > tolerance 0.0010000000000000002
  => Position 9: 161 != 160 | difference 1 > tolerance 0.0010000000000000002
[...]

PR #1806, opened 2024-05-23
Feat: Implement ONNX RandomUniform + RandomNormal in burn-import

hexd0t:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed (unrelated tests fail for me, see #1805, but the ones for the import succeed)

### Related Issues/PRs

Implements a part of #1714

### Changes

Adds a Node for ONNX RandomUniform to burn-import.
It supports all of the operation's attributes except `seed`, since I've found no way of providing a seed value to the Tensor (but since the specific Pseudorandom Generator isn't given, models cannot really rely on using fixed seeds anyways, so the impact is hopefully negligible)

### Testing

Created a testcase along the instructions of the contributor book. Since I've found no way of easily fixing the random seed in the testcase, it only checks the resulting tensor dimensions, not actual golden values.


crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -71,6 +71,8 @@ pub fn dim_inference(node: &mut Node, graph_io: &mut OnnxGraphIO) {
         NodeType::PRelu => same_as_input(node),
         NodeType::Where => where_update_outputs(node),
         NodeType::Squeeze => squeeze_update_output(node),
+        NodeType::RandomUniform => explicit_shape_update_output(node),
+        NodeType::RandomNormal => explicit_shape_update_output(node),
```
laggui, line 75: This PR only tackles `RandomUniform`, mind removing the `RandomNormal`? Just so the appropriate warning is emitted when trying to go through the random normal conversion on import.

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -120,6 +122,47 @@ fn constant_update_outputs(node: &mut Node) {
...
+        .into_i64s();
+
+    let elem_type = match dtype {
+        DataType::FLOAT => ElementType::Float32,
+        DataType::INT32 => ElementType::Int32,
+        DataType::INT64 => ElementType::Int64,
+        DataType::DOUBLE => ElementType::Float64,
+        DataType::BOOL => ElementType::Bool,
+        _ => panic!("tensor with type {dtype:?} not supported"),
+    };
```
laggui, line 151: Might want to restrict to the actual dtypes supported for random distributions (i.e. floating points).

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -374,6 +377,33 @@ impl OnnxGraph {
...
+            .map(|val| val.clone().into_f32() as f64)
+            .unwrap_or(1.0f64);
+        let low = node
+            .attrs
+            .get("low")
+            .map(|val| val.clone().into_f32() as f64)
+            .unwrap_or(0.0f64);
+
+        if node.attrs.contains_key("seed") {
+            warn!("seed attribute is not supported!");
```
laggui, line 401: Nice touch with the warning 👍

crates/burn-import/src/burn/node/random_uniform.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn into_node(self) -> Node<PS> {
+        Node::RandomUniform(self)
+    }
+
+    fn register_imports(&self, imports: &mut crate::burn::BurnImports) {
+        imports.register("core::ops::Deref");
+        imports.register("burn::tensor::Distribution");
+        imports.register("burn::prelude::Shape");
+    }
+}
```
laggui, line 67: Looks like this is missing the codegen tests that are found for other nodes. See `reshape.rs` for example.

crates/burn-import/src/burn/node/random_uniform.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn output_types(&self) -> Vec<Type> {
+        vec![Type::Tensor(self.output_ty.clone())]
+    }
+
+    fn forward(&self, _scope: &mut Scope, _node_position: usize) -> TokenStream {
+        let output = &self.output_ty.name;
+        let shape = self.get_output_shape();
+        let dist = self.get_distribution();
+        quote! {
+            let #output = Tensor::random(#shape, #dist, self.device.deref());
```
laggui, line 54: `Tensor::random` takes a reference to the device iirc, so I think it should be `&self.device` no?

laggui (changes requested):
Nice work! Some minor comments/changes but overall looks good 👍

crates/burn-import/src/onnx/dim_inference.rs
hexd0t: Since `RandomNormal` has no input, `temporary_pass_through_stub` crashes trying to copy the input's dimensions.
But since `RandomNormal` and `RandomUniform` are essentially the same, I'll just also implement `RandomNormal` as well.

Would you rather have two almost exactly copy-paste nodes, or a single node with a `Distribution`-Parameter? If the latter, any input on naming the struct? Just `RandomNode`?

crates/burn-import/src/burn/node/random_uniform.rs
hexd0t: This stems from the fact that self.device is not directly a `Backend::Device`, but rather a `module::Ignored<Backend::Device>`:

```rust
45 |           let randomuniform1_out1 = Tensor::random(
   |  ___________________________________^
46 | |             Shape::new([2usize, 3usize]),
47 | |             Distribution::Uniform(0f64, 1f64),
48 | |             &self.device,
49 | |         );
   | |_________^ expected associated type, found `Ignored<<B as Backend>::Device>`
   |
   = note: expected associated type `<B as burn::prelude::Backend>::Device`
                       found struct `Ignored<<B as burn::prelude::Backend>::Device>`
```

Alternatively, `&*self.device` would work and also not require importing the Deref trait - personally, I find it looks a bit cleaner with the explicit deref, but have no hard feelings either way, so I'll change it.

laggui: Oh! Totally forgot about the newly added `Ignored` type wrapper. Never mind that comment then :)

crates/burn-import/src/onnx/dim_inference.rs
laggui: I don't mind having two almost exactly the same nodes, but `RandomDistributionNode` with a `Distribution` field sounds good if you want to take a stab at this.

laggui (approved):
Thanks for contributing! 🙏 

LGTM. One small typo to fix but I'll approve in advance.

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -143,11 +144,8 @@ fn explicit_shape_update_output(node: &mut Node) {
 
     let elem_type = match dtype {
         DataType::FLOAT => ElementType::Float32,
-        DataType::INT32 => ElementType::Int32,
-        DataType::INT64 => ElementType::Int64,
         DataType::DOUBLE => ElementType::Float64,
-        DataType::BOOL => ElementType::Bool,
-        _ => panic!("tensor with type {dtype:?} not supported"),
+        _ => panic!("tensor with type {dtype:?} not supported for RandomU"),
```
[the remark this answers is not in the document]
hexd0t: Fixed

laggui:
🚀 🚀

PR #1807, opened 2024-05-23
Feat/cube/remaining ops

louisfd:
Many ops for cube 
Still not satisfied with how tensor ops need to specify their inner type as generic, eg: `Tensor::<F>::shape(tensor, 1u32)` when in fact F has no impact at all here. But a next step is to allow method calls like `tensor.shape(1u32)`, maybe it can be fixed then. 
Contributes to #1665


crates/burn-cube/tests/language/tensor.rs
```diff
@@ -0,0 +1,59 @@
...
+                var: GlobalInputArray(0, Item { \
+                    elem: Float(F32), \
+                    vectorization: 1 \
+                }), \
+                out: Local(0, Item { \
+                    elem: UInt, \
+                    vectorization: 1 \
+                }, 0) \
+            }), \
+            Metadata(Stride { \
```
nathanielsimard, line 36: Do we really want to compare on the string representation?

louisfd: No, that was just laziness on my part 😅

nathanielsimard (approved):

PR #1813, opened 2024-05-26
feat: expand onnx import

JachymPutta:
### Changes

Include the expand operator to burn-import

### Testing

Manual
`cargo test expand`

### Related Issues

#1714


crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -470,6 +471,33 @@ fn equal_update_outputs(node: &mut Node) {
...
+    let shape = if node.inputs.len() == 2 {
+        match &node.inputs[1].value {
+            Some(value) => match value {
+                Data::Int64s(shape) => Some(shape.clone()),
+                _ => panic!("Expand: invalid input types"),
+            },
+            None => None,
+        }
+    } else {
+        node.attrs.get("shape").cloned().map(|v| v.into_i64s())
```
antimora, line 484: This is not applicable for Expand since there is no attribute `shape` the spec: https://onnx.ai/onnx/operators/onnx__Expand.html. So we should remove it.

antimora (changes requested):
Looks good! Thank you for your contribution. 

Minor. We should remove a case when we extract `shape` from Expand node attributes. Spec does not have it compared to other node types. Probably it was left from copying. 

Once this is done, I believe we can merge it.

crates/burn-import/src/onnx/dim_inference.rs
JachymPutta: Yes, you're right! I will get rid of it:)

antimora (approved):
Looks Good To Me

PR #1814, opened 2024-05-26
Feature/onnx argmax

will-maclean:
### Checklist

- [✅] Confirmed that `run-checks all` script has been executed.
- [✅] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

[_Original Issue_](https://github.com/tracel-ai/burn/issues/1714)

### Changes

- Created `crates/burn-import/onnx/tests/argmax/argmax.py` with a simple model that calls `.argmax(dim)` on an input tensor, and onnx code to convert the model into a ONNX file
- Created an `ArgMaxNode` in `crates/burn-import/src/burn/node/argmax.rs` to store ArgMax functionality. Also added necessary helper functions
- Created tests in `onnx_tests.rs`

### Testing

A test called `argmax` in `onnx_tests.rs` checks that the created model is able to generate the correct argmax outputs.

One thing I did notice was that the only params implemented for the existing tensor argmax function are the dim/axis - keepdims and select_last_index, which are also params for the ONNX argmax node, don't seem to exist in burn. I set them to their defauls, however this may cause issues if trying to import an ONNX model where e.g. keepdims=false. Happy to hear people's input on this.


crates/burn-import/onnx-tests/tests/argmax/argmax.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+# used to generate model: onnx-tests/tests/concat/concat.onnx
```
laggui, line 3: It's just a comment, but might as well point to the right file: onnx-tests/tests/argmax/argmax.onnx

crates/burn-import/onnx-tests/tests/argmax/argmax.py
```diff
@@ -0,0 +1,41 @@
...
+import torch
+import torch.nn as nn
+
+class Model(nn.Module):
+    def __init__(self, argmax_dim: int = 0):
+        super(Model, self).__init__()
+        self._argmax_dim = argmax_dim
+
+    def forward(self, x):
+        # Concatenate along the channel dimension
```
laggui, line 14: Left over comment from concat template

crates/burn-import/src/burn/node/argmax.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn forward(
+        &self,
+        scope: &mut crate::burn::Scope,
+        node_position: usize,
+    ) -> proc_macro2::TokenStream {
+        let axis = self.axis.to_tokens();
+
+        //NOTE: are select_last_index and keep_dims supported?
+        let _select_last_index = self.select_last_index.to_tokens();
+        let _keepdims = self.keepdims.to_tokens();
```
laggui, line 36: `select_last_index` is not supported in Burn, so we could add a warning when checking the node config if it is set.

Same for `keep_dims`, the argmax method in Burn returns a tensor of the same shape as the input. But if the tensor has more than 1 dimension, we could generate the code to get the same output as the ONNX node by doing `argmax` followed by a squeeze on the selected axis. For this PR, feel free to keep the support limited to `keep_dims` set to 1 (and panic when checking the node config if it is 0) or add the sequence of ops with a squeeze after the argmax.

And then you can remove the unused fields from the codegen here.

crates/burn-import/src/burn/node/argmax.rs
```diff
@@ -0,0 +1,108 @@
...
+#[cfg(test)]
+mod tests {
+
+    use burn::record::FullPrecisionSettings;
+
+    use super::*;
+    use crate::burn::{graph::BurnGraph, node::test::assert_tokens, TensorType};
+
+    #[test]
+    fn test_codegen_gather() {
```
laggui, line 59: Whoops, let's rename the codegen test to identify the correct op (argmax)

crates/burn-import/src/burn/node/argmax.rs
```diff
@@ -0,0 +1,108 @@
...
+        let mut graph = BurnGraph::<FullPrecisionSettings>::default();
+
+        graph.register(ArgMaxNode::new(
+            TensorType::new_float("tensor1", 2),
+            TensorType::new_float("tensor2", 2),
```
laggui, line 64: The output tensor (`tensor2`) should be int no?

crates/burn-import/src/burn/node/base.rs
```diff
@@ -145,6 +148,7 @@ impl<PS: PrecisionSettings> Serialize for Node<PS> {
 impl<PS: PrecisionSettings> Node<PS> {
     pub fn name(&self) -> &str {
         match self {
+            Node::ArgMax(_) => "argmax1",
```
laggui, line 151: "argmax1" -> "argmax"

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -14,6 +14,8 @@ use super::{
 pub fn dim_inference(node: &mut Node, graph_io: &mut OnnxGraphIO) {
     match node.node_type {
         NodeType::Add => same_as_input(node),
+        // NodeType::ArgMax => same_as_input(node),
```
laggui, line 17: Please remove dead code comments 🙂

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -321,6 +323,78 @@ fn reduce_mean_update_outputs(node: &mut Node) {
...
+    //             output_shape = None;
+    //         }
+    //     }
+
+    //     node.outputs[0].ty = ArgType::Tensor(TensorType {
+    //         dim: tensor.dim.clone() - 1,
+    //         shape: output_shape.clone(),
+    //         elem_type: ElementType::Int64,
+    //     });
+    // }
```
laggui, line 395: Same comment. Please remove dead code comments 🙂

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -468,6 +468,44 @@ pub fn softmax_config(node: &Node) -> usize {
...
+        ArgType::Tensor(tensor) => tensor,
+        _ => panic!("Only tensor input is valid"),
+    };
+
+    // extract the attributes
+    for (key, value) in node.attrs.iter() {
+        match key.as_str() {
+            "axis" => axis = value.clone().into_i64(),
+            "select_last_index" => select_last_index = value.clone().into_i64(),
+            "keepdims" => keepdims = value.clone().into_i64(),
```
laggui, line 496: Instead of capturing the values which are not used, we should check that the node config is supported for the matching op in burn.

In other words, if `select_last_index` != 1 we should warn that this config is ignored. And if this PR only adds support for `keepdims` set to 1 we should panic otherwise.

laggui (changes requested):
Thank you for tackling the ONNX track!

Overall, the implementation looks good 🙂 Just some minor comments/changes to complete the PR.

will-maclean:
Hi laggui, thanks for the feedback. I addressed those points (hopefully I didn't miss any) and set any `keepdims=false` to panic, with warnings for trying to set `select_last_index`.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -492,18 +491,29 @@ pub fn argmax_config(node: &Node) -> (usize, usize, usize) {
     for (key, value) in node.attrs.iter() {
         match key.as_str() {
             "axis" => axis = value.clone().into_i64(),
-            "select_last_index" => select_last_index = value.clone().into_i64(),
+            "select_last_index" => log::warn!(
+                "select_last_index param for argmax is ignored in burn (got {:?})",
+                value
+            ),
```
laggui, line 497: I think we can still capture the `select_last_index` value here, but only warn if it is 1 (because the default implementation pretty much everywhere including Burn is to return the first max value, not last).

laggui (approved):
I have a minor request just to limit the warning based on the value for `select_last_index`.

Otherwise, looks good! So I'll approve in advance and we can merge when addressed.

will-maclean:
Should be all good now :)

laggui (approved):
Thank you!!

PR #1815, opened 2024-05-26
Feat: burn-import implement ONNX ConstantOfShape

hexd0t:
### Checklist

### Related Issues/PRs

Although `ConstantOfShape` is not explicitly there, this still benefits  #1714

### Changes

Add `ConstantOfShape` support to onnx import.

### Testing

Codegen and overall import tests as described in the contributor book.


crates/burn-import/src/burn/node/constant_of_shape.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn output_types(&self) -> Vec<Type> {
+        vec![self.output.clone()]
+    }
+
+    fn forward(&self, _scope: &mut Scope, _node_position: usize) -> TokenStream {
+        let output = self.output.name();
+        let input = self.input.name();
+        let value = self.value.val_tokens();
+        quote! {
+            let #output = Tensor::full(#input.shape(), #value, &*self.device);
```
hexd0t, line 108: Should the value be de-/serialized into a field? The 'normal' Constant OP does this for `Tensor`s, but not `Scalar`s, but I'm not sure I have grasped why...

nathanielsimard, line 108: This should be `&self.device`, I don't think the `*` is necessary.

nathanielsimard: I don't think this is necessary.

hexd0t: `self.device` is a `module::Ignored<Device>`, so the asterisk is used to trigger the deref function without needing to import the deref trait and calling it explicitly

crates/burn-import/onnx-tests/tests/constant_of_shape/constant_of_shape.py
```diff
@@ -0,0 +1,53 @@
...
+                outputs=["output1"],
+                name="/ConstantOfShape",
+                value=onnx.helper.make_tensor("value", data_type=onnx.TensorProto.FLOAT, dims=[1], vals=[1.125])
+            ),
+        ],
+        inputs=[
+            onnx.helper.make_value_info(
+                name="input1",
+                type_proto=onnx.helper.make_tensor_type_proto(
+                    elem_type=onnx.TensorProto.INT64, shape=[2, 3, 2]
```
laggui, line 30: Is the onnx graph generated valid?

The input is supposed to be a 1D tensor with values `[2, 3, 2]`, not shape `[2, 3, 2]`.

You can check with `onnx.checker.check_model`

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -122,6 +123,24 @@ fn constant_update_outputs(node: &mut Node) {
...
+        .map(|v| v.clone().into_tensor().elem_type)
+        .unwrap_or(ElementType::Float32); // If not given, defaults to 0 as float32
+
+    if let ArgType::Tensor(input_type) = &node.inputs[0].ty {
+        node.outputs[0].ty = ArgType::Tensor(TensorType {
+            elem_type: value_type,
+            dim: input_type.dim,
+            shape: None,
+        });
+    } else {
```
laggui, line 139: Again, I think we should make sure that the input is a 1D tensor.

crates/burn-import/src/burn/node/constant_of_shape.rs
```diff
@@ -0,0 +1,185 @@
...
+        );
+    }
+
+    #[test]
+    fn test_codegen_nodes() {
+        let mut graph = BurnGraph::<FullPrecisionSettings>::default();
+
+        graph.register(ConstantOfShapeNode::new(
+            Type::Tensor(TensorType::new_float("tensor1", 4)),
+            Type::Tensor(TensorType::new_float("tensor2", 4)),
```
laggui, line 148: The input `tensor1` should have rank 1 (not 4).

crates/burn-import/src/burn/node/constant_of_shape.rs
```diff
@@ -0,0 +1,185 @@
...
+        vec![self.output.clone()]
+    }
+
+    fn forward(&self, _scope: &mut Scope, _node_position: usize) -> TokenStream {
+        let output = self.output.name();
+        let input = self.input.name();
+        let value = self.value.val_tokens();
+        // Note: in the generated code, self.device is a &module::Ignored<Device>, so to get a &Device, &* is needed:
+        quote! {
+            let #output = Tensor::full(#input.shape(), #value, &*self.device);
```
laggui, line 109: According to [the spec](https://github.com/onnx/onnx/blob/main/docs/Operators.md#ConstantOfShape), the input is a 1D tensor. So to convert to burn we probably need to do something like this instead:

```rust
let input = scope.tensor_use_owned(&input, node_position);
quote! {
    let #output = Tensor::full(#input.to_data().value, #value, &#input.device());
}
```

Also, the spec supports a constant scalar when no dimensions are given, so we would have to handle this particular use case. But in this PR we can always limit the scope to tensors for now.

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -433,6 +422,39 @@ impl OnnxGraph {
...
+                Data::Float32(val) => val.into(),
+                Data::Float32s(vals) => ConstantValue::from_vec(vals),
+                Data::Float64(val) => val.into(),
+                Data::Float64s(vals) => ConstantValue::from_vec(vals),
+                Data::Int32(val) => val.into(),
+                Data::Int32s(vals) => ConstantValue::from_vec(vals),
+                Data::Int64(val) => val.into(),
+                Data::Int64s(vals) => ConstantValue::from_vec(vals),
+                Data::Bool(val) => val.into(),
+                Data::Bools(vals) => ConstantValue::from_vec(vals),
```
laggui, line 450: We only need to handle scalar types, not vectors. The value is a single element, so no need for `Float32s`, `Float64s`, etc.

laggui (changes requested):
A couple of changes needed to follow the spec, but otherwise a great start! 💪

antimora:
I will take over this PR and fix issues.

antimora:
I am trying to resolve some conflicts. It looks like @laggui added ConstantOfShape implementation in "Refactor tensor data" (#1916). I know he had worked on this previously, so it might have been checked in accidentally.

crates/burn-import/src/burn/node/constant_of_shape.rs
antimora: I can confirm. It works.

crates/burn-import/onnx-tests/tests/constant_of_shape/constant_of_shape.py
antimora: Fixed it by introducing shape type.

crates/burn-import/src/onnx/dim_inference.rs
antimora: Fixed it by introducing shape type.

crates/burn-import/src/burn/node/constant_of_shape.rs
antimora: Fixed it by introducing shape type.

crates/burn-import/src/burn/node/constant_of_shape.rs
antimora: Fixed it by introducing shape type.

crates/burn-import/src/onnx/to_burn.rs
antimora: Fixed

antimora: Actually the data comes as 1d tensor (arrays) (see https://onnx.ai/onnx/operators/onnx__ConstantOfShape.html) so we need to handle arrays not scalars. I removed scalars.

antimora:
@skewballfox, if you have a chance, please take a look at this PR. I started using Shape arg type instead of using Tensor because shape operation is outside tensor operation but ONNX treats them as if they tensors.

antimora (approved):
I fixed unresolved issues and started using Shape ArgType which we had but didn't use it. I am approving since I looked at the code and fixed issues.

crates/onnx-ir/src/dim_inference.rs
```diff
@@ -128,6 +129,31 @@ fn constant_update_outputs(node: &mut Node) {
...
+    let dim = match &node.inputs[0].ty {
+        // Sometimes the input is a tensor, sometimes it's a shape
+        // because didn't have a chance to update the IR
+        ArgType::Tensor(input_type) => input_type.dim,
+        ArgType::Shape(dim) => *dim,
+        _ => panic!("ConstantOfShape node must have a Tensor or Shape type input"),
+    };
+
+    // Fix the input type to be a shape
+    node.inputs[0].ty = ArgType::Shape(dim);
```
skewballfox, line 151: This *might* cause an issue depending on the origin  of the input, and how the code generation is handled. 

If `node.inputs[0]` is the output of another node, and that node's output is a tensor, and the `ArgType::Shape(dim)` generates a vector, then passing that will lead to incorrect code generation.

I suspect this is also an issue with the transpose in `onnx-ir/src/coalesce.rs`, and there is two ways I can think of to handle it:
- The easier one in this case is to avoid mutating the input, and check for `ArgType::Tensor` in `burn-import/src/node/costant_of_shape.rs`, and if it is then call `.dim()`.
- The harder, though perhaps better long term, way of handling it is to either rework the parser to inject nodes to do transforms when inputs are mutated, or store/pass a list of arguments where transforms need to happen. I think we can avoid mutating the list of nodes (and thus avoid needing to do an extra pass), since we know the input name that needs to transform by this point in the parsing process

crates/burn-import/onnx-tests/tests/constant_of_shape/constant_of_shape.py
antimora: Thanks! I fixed it and actually it helped identify a bug in dim inference.

crates/onnx-ir/src/dim_inference.rs
antimora: @skewballfox yes, I agree the current the input/output type inference is not fully polished up. I will file an issue ticket to refactor for a long term usage. We can apply lessons learned. 

I have a test where an output of Shape node  is Shape arg type. We need to use Shape arg type to declare array types in Rust. So we need to know its size. 

BTW, I was looking at how Netron figures out shapes and it's pretty complicated. Fortunately, for our use case, only ranks and data type matter the most. I think most other DL frameworks track shapes dynamically (may except dfdx). So rank inference will be useful for other frame works.

laggui (approved):

hexd0t:
Thanks; sorry, I had other things preventing me from finishing this.

antimora:
> Thanks; sorry, I had other things preventing me from finishing this.

No worries! This happens. Thanks for starting it and contributing. Now it's done.

PR #1832, opened 2024-05-28
Fix `DataSerialize` conversion for elements of the same type

laggui:
### Checklist

### Related Issues/PRs

Improve record loading speed for [llama-burn](https://github.com/tracel-ai/models/pull/35)

### Changes

- Check origin and destination element type and skip conversion for same type element
- Reverted version to 0.14.0 (erroneously bumped twice since last release)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,7 +242,7 @@ dependencies = [
 
 [[package]]
 name = "backend-comparison"
-version = "0.15.0"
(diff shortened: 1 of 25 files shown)


laggui:
Guess it wasn't such a great idea to include the version downgrade.. CI fails because of that. Will try to fix.

nathanielsimard (approved):

PR #1834, opened 2024-05-28
feat: added range onnx import

JachymPutta:
### Checklist

### Changes

Add the range import to onnx

### Testing

`cargo test range`

### Related Issue

#1714


crates/burn-import/src/burn/node/range.rs
```diff
@@ -0,0 +1,118 @@
...
+            _ => panic!("End must be a scalar"),
+        };
+
+        let step = match &self.step {
+            Type::Scalar(s) => {
+                let name = s.name.clone();
+                quote! { #name }
+            }
+            _ => panic!("Step must be a scalar"),
+        };
```
laggui, line 49: If the three fields are all supposed to be scalar, I suggest we change the node fields to:

```rust
pub struct RangeNode {
    pub start: ScalarType,
    pub end: ScalarType,
    pub step: ScalarType,
    pub output: TensorType,
}
```

and the logic to check the extract the scalar types can go in the `range_conversion`.

laggui (changes requested):
Good job! 🙂 

I have one minor request over form when checking the types, otherwise looks good to me.

crates/burn-import/src/burn/node/range.rs
laggui: Well yes, scalars are simply represented as 0-dim tensors. Basically what I am suggesting is just to move the type handling outside of the forward node method to the `range_conversion` function instead, a bit like this:

```rust
  fn range_conversion(node: Node) -> RangeNode {
      let output = node.outputs.first().unwrap().to_tensor_type();
      let start = node.inputs.first().unwrap().to_type();
      let end = node.inputs.get(1).unwrap().to_type();
      let step = node.inputs.get(2).unwrap().to_type();

      // Check/convert start, end & step to ScalarType just as it was done in the forward method
     // ...

      RangeNode::new(start, end, step, output)
  }
```

laggui (approved):
LGTM! 🚀

antimora (approved):
LGTM

PR #1842, opened 2024-05-31
Feat/cube/struct support

louisfd:
Support for using structs in Cube. 
There are still some todos:
- [ ] cube annotation over struct to generate the expand version
- [ ] use it in the convolution, with launch args stuff


louisfd (commented):
Lgtm

PR #1856, opened 2024-06-05
feat: added slice onnx import

JachymPutta:
### Checklist

### Related Issues/PRs

Helping with #1714 

### Changes

Added the slice import to onnx

### Testing

`cargo test -p burn-import --color=always -- --color=always`
`cargo test -p onnx-tests --color=always`


crates/burn-import/onnx-tests/tests/onnx_tests.rs
```diff
@@ -459,6 +460,24 @@ mod tests {
...
 
+    #[test]
+    fn slice() {
+        let model: slice::Model<Backend> = slice::Model::default();
+        let device = Default::default();
+
+        let input = Tensor::<Backend, 2>::from_floats(
+            [
+                [1., 2., 3., 4., 5., 6., 7., 8., 9., 10.],
+                [1., 2., 3., 4., 5., 6., 7., 8., 9., 10.],
```
laggui, line 471: Maybe change the values of the second dim just to make it clear what data is sliced with this test 😅

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -0,0 +1,104 @@
...
+        // let steps = match &self.steps {
+        //     Some(steps) => steps.to_tokens(),
+        //     None => quote! { None },
+        // };
+
+        // compine starts and ends into ranges
+        // example: starts = [0, 0, 0, 0], ends = [1, 1, 1, 1]
+        // ranges = [0..1, 0..1, 0..1, 0..1]
+        let starts = &self.starts;
+        let ends = &self.ends;
```
laggui, line 40: I think we still need to check the axes since there is no guarantee that they always including the leading axes (e.g., if `axes = [2, 3]` for values of `starts` and `ends`), but it's not entirely clear even as described in the [spec](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Slice).

If we keep this solution without checking for non leading axes, I would at least parse the values of `axes` and error on invalid/unsupported values (i.e., skipping leading axes).

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -1015,6 +1015,53 @@ pub fn shape_config(curr: &Node) -> (usize, usize) {
...
+        ArgType::Tensor(tensor) => {
+            assert_eq!(tensor.dim, 1, "Slice: ends tensor must be 1D");
+            if let Some(Data::Int64s(shape)) = end_value.as_ref() {
+                shape.iter().map(|x| *x as usize).collect()
+            } else {
+                panic!("Tensor data type must be int64")
+            }
+        }
+        _ => panic!("Only tensor input is valid for shape"),
+    };
```
laggui, line 1049: According to the [spec](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Slice), we should handle negative values for `starts`, `end` and `axes`.

We should also be able to map the `axes` specified to the correct ranges for `tensor.slice(ranges)`.

Also, we should still parse steps and make sure that values != 1 are not accepted (panic) since Burn only supports regular ranges with step 1.

laggui (changes requested):
Thanks for tackling this op! 🙏 

A couple of minor requests, and perhaps removing the commented code once everything is completed :)

crates/burn-import/src/onnx/op_configuration.rs
JachymPutta: Makes sense, I will put those checks in place

crates/burn-import/onnx-tests/tests/onnx_tests.rs
JachymPutta: 😃 right, yeah

crates/burn-import/src/onnx/op_configuration.rs
JachymPutta: How would we handle negative values for `starts`, `ends` and `axes` since we would need to know the length of the input to calculate the new index and that's only known at runtime? I could always throw an error, but there might be a better way to do this.
`a[-1] == a[a.len() - 1]`

laggui: Oh right. I just checked the implementation for other ops which deal with negative axes and turns out that it's assumed the tensor dims are known when parsing the graph in these nodes... see for example the [shape node config](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/onnx/op_configuration.rs#L1007-L1013).

In this case, we have two options:
1) Assume the `tensor.dim` is valid at this point (as is done in the current ops)
2) Throw an error for negative axis.

I don't have any preference per se.

JachymPutta: Cool, thanks!

laggui (approved):
LGTM! 🚀 

Thanks for making the changes and handling the cases we don't support for now.

PR #1858, opened 2024-06-06
Speedup client.create for small allocations.

ArthurBrussee:
### Checklist

- [x ] Confirmed that `run-checks all` script has been executed.
- [ x] Made sure the book is up to date with changes in this PR.

### Changes

I've noticed client.create() is quite slow when churning through lots of small kernels. It seems allocating a buffer each time is causing some overhead. After poking at WGPU a bit it seems a solution is to use their StagingBelt - effectively a pool.

It seems queue.write_buffer_with() would be even faster - but I couldn't get that to work. As it doesn't live on the encoder, you can't get the order of scheduling right.

What really would be faster though, is a different approach. These small allocations seem to come from the "info" buffers for various kernels. These really might need some special considerations - they should be bound as uniform data, and maybe a kernel could manage it's own "argument buffer" instead of relying on the burn allocator.

I might pursue that in the future, but figured this might be neat anyway for discussion, and who knows maybe other small allocations appear.

### Testing
Here's a comparison of a forward pass before/after. Nb this is just measuring CPU overhead. GPU is likely slightly faster as well.

traceCompare

### DIFF
--- a/crates/burn-wgpu/src/compute/server.rs
+++ b/crates/burn-wgpu/src/compute/server.rs
@@ -1,3 +1,5 @@
+use std::num::NonZeroU64;
+
 use super::WgpuStorage;
 use alloc::{borrow::Cow, sync::Arc};
 use burn_compute::{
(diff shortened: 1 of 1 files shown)


nathanielsimard (approved):
LGTM Thanks for the improvement 🙏

PR #1861, opened 2024-06-06
Get resources from server

ArthurBrussee:
I'm working towards mixing custom WGPU commands & Burn operations.

Now that you can init from an[ existing WGPU device](https://github.com/tracel-ai/burn/commit/e0a1094f894e83f30e04a661d5513cdaa76dbb50), a next step is allowing to get handles to the underlying WGPU resource of a buffer. This *has* to be used with care, but could allow mixing any kind of custom wgpu commands.

One question is - is it better to have a method get_resources(Vec<Binding) -> Vec<Resource> ? That could prevent some synchronization traffic. On the other hand it's a bit more complicated to use and wastes a vector allocation when asking for a single resource, not sure!

Another question could be - read_resource vs get_resource, not sure which is better either :)

After this PR, the one remaining issue is how to synchronize burn - something like client.submit() (client.sync() works but is slower than need be).

### DIFF
--- a/crates/burn-compute/src/channel/base.rs
+++ b/crates/burn-compute/src/channel/base.rs
@@ -1,4 +1,7 @@
-use crate::server::{Binding, ComputeServer, Handle};
+use crate::{
+    server::{Binding, ComputeServer, Handle},
+    storage::ComputeStorage,
+};
 use alloc::vec::Vec;
 use burn_common::reader::Reader;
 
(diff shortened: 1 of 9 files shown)


crates/burn-compute/src/channel/base.rs
```diff
@@ -8,6 +11,12 @@ pub trait ComputeChannel<Server: ComputeServer>: Clone + core::fmt::Debug + Send
     /// Given a binding, returns owned resource as bytes
     fn read(&self, binding: Binding<Server>) -> Reader<Vec<u8>>;
 
+    /// Given a resource handle, return the storage resource.
+    fn get_resource(
+        &self,
+        binding: Binding<Server>,
+    ) -> <Server::Storage as ComputeStorage>::Resource;
```
nathanielsimard, line 18: Get resource is better than read. Reading kind of assumes we actually do something with the memory, where get is simply returning the resource.

nathanielsimard (approved):
I'm not sure how we could sync the server, but I think it would make sense to expose the encoder from the wgpu side, meaning one would be able to send its own encoder as well as the burn encoder to the wgpu queue, reducing the overhead.

PR #1863, opened 2024-06-07
feat: resize onnx import

mosure:
### checklist

- [x] confirmed that `run-checks all` script has been executed.
- [x] made sure the book is up to date with changes in this PR.

### related issues/PRs

- onnx op import sprint: https://github.com/tracel-ai/burn/issues/1714

### changes

support `Resize` onnx import

### testing

`cargo test -p burn-import --color=always -- --color=always`
`cargo test -p onnx-tests --color=always`


laggui:
In the codegen for the node you could get the values in the size tensor. Something like

```rust
#size.to_data().value
```

mosure:
> In the codegen for the node you could get the values in the size tensor. Something like
> 
> ```rust
> #size.to_data().value
> ```
> 
> /edit: shoot I didn't realize the `output_size` is a fixed size array, that won't work.

it is a good idea, something like:
```rust
let data: [i64; 2] = input.to_data()
    .value
    .iter()
    .map(|x| x.elem())
    .collect::<Vec<i64>>()
    .try_into()
    .unwrap();
```

laggui:
Yeah something like that could work to preserve the flexibility.

I think in most cases the sizes could be constant so we might be able to get the values directly when parsing the graph, but not entirely sure.

I see in the [spec](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Resize-13) that ONNX also supports providing a `scales` instead, but let's limit the scope of this PR. We can just parse the other unsupported fields/inputs and throw an error message for the cases we don't support.

Luni-4 (approved):
Thanks a lot @mosure! It's fine for me! 

I'm going to wait for a final review by @laggui though, just to be sure I haven't forgotten something

crates/burn-import/src/onnx/dim_inference.rs
```diff
@@ -287,6 +288,51 @@ fn reshape_update_outputs(node: &mut Node) {
...
+    //     Data::Int64s(output_size) => output_size.clone(),
+    //     _ => panic!("Resize: invalid output_size type"),
+    // };
+
+    // let input_shape = input.shape.as_ref().expect("Resize: Input shape must be known");
+
+    // let mut shape = input_shape.clone();
+    // shape[input_shape.len() - 2] = output_size[0] as usize;
+    // shape[input_shape.len() - 1] = output_size[1] as usize;
+    // let shape = shape.into();
```
laggui, line 328: Let's remove the commented dead code :)

laggui (commented):
LGTM! Just a minor request to remove the commented dead codeé

I think this is probably the operator with the most attributes and optional inputs I've seen thus far 😅

Glad you didn't try to cover every use case because we can't support them anyway hahah

laggui (approved):

PR #1864, opened 2024-06-07
Add option to flush queue instead of waiting for completion.

ArthurBrussee:
### Checklist

### Related Issues/PRs

This adds the last bit of infrastructure needed for custom burn<->wgpu interop, after initializing from an existing wgpu server (#1788), and getting WGPU resources (#1861).

### Changes
After this PR, the setup for custom interop roughly looks like:

```
init_existing_device(adapter, device, wgpu_queue);

// Do some burn operations.
let x = Tensor::from_floats();
let y = x * 2 + 5.0; // Some calculations we really need in our wgpu code.

// Or however you have access to the client.
let client = y.into_primitive().client;

// Flush burn operations - they now live on the wgpu queue.
client.submit();

// Get a reference to y binding.
let y_wgpu_binding= client.get_resource(y);

// Run custom commands on the queue. wgpu makes guarantees that any resources have been
// calculated up to this point.
run_my_custom_wgpu_code(wgpu_queue, y_wgpu_binding);
```

This does require a seperate encoder for your custom wgpu code. I think that's ok - a custom wgpu app might have it's own already anyway - but in the future we could expose some notion of the servers encoder to share resources.

### DIFF
--- a/crates/burn-compute/src/channel/base.rs
+++ b/crates/burn-compute/src/channel/base.rs
@@ -26,6 +26,9 @@ pub trait ComputeChannel<Server: ComputeServer>: Clone + core::fmt::Debug + Send
     /// Executes the `kernel` over the given `bindings`.
     fn execute(&self, kernel: Server::Kernel, bindings: Vec<Binding<Server>>);
 
+    /// Submit the current queued tasks to the server.
+    fn submit(&self);
+
     /// Wait for the completion of every task in the server.
     fn sync(&self);
 }
--- a/crates/burn-compute/src/channel/base.rs
+++ b/crates/burn-compute/src/channel/base.rs
@@ -3,7 +3,7 @@ use crate::{
     storage::ComputeStorage,
 };
 use alloc::vec::Vec;
-use burn_common::reader::Reader;
+use burn_common::{reader::Reader, sync_type::SyncType};
 
 /// The ComputeChannel trait links the ComputeClient to the ComputeServer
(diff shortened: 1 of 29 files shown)


crates/burn-compute/src/channel/base.rs
```diff
@@ -26,6 +26,9 @@ pub trait ComputeChannel<Server: ComputeServer>: Clone + core::fmt::Debug + Send
     /// Executes the `kernel` over the given `bindings`.
     fn execute(&self, kernel: Server::Kernel, bindings: Vec<Binding<Server>>);
 
+    /// Submit the current queued tasks to the server.
+    fn submit(&self);
```
nathanielsimard, line 30: Do we want to have a submit function here or pass an option to the `sync` method? I'm unsure which one is the most intuitive. It's a bit weird to have an exposed `submit` function along `execute` and `sync`. 

```rust
fn sync(&self, options: SyncOptions);

#[derive(Clone, Copy)]
enum SyncOptions {
    /// Sync the execution queue.
    Queue,
    //// Sync the execution queue and wait for all tasks to be completed.
    Full,
}
```

ArthurBrussee: Ah alright! Yeah that might be more consistent.

It does slightly complicate things as for better or worse it kind of forces this to be exposed on the backend trait. Lmk how this looks!

nathanielsimard (approved):
Awesome! 👏

PR #1876, opened 2024-06-10
Feat: Add `movedim` tensor operator

LilDojd:
### Checklist

### Related Issues/PRs

None

### Changes

Added `movedim` default implementation. Mirrors the [PyTorch one](https://pytorch.org/docs/stable/generated/torch.movedim.html).

This just wraps a call to permute with sanity checks, but moving one/few axes while keeping all others inbound is a common enough operation to warrant its specific convenient syntactic sugar.

### Testing

Added unit tests.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -9,6 +9,8 @@ use alloc::string::String;
 #[cfg(any(feature = "wasm-sync", not(target_family = "wasm")))]
 use alloc::vec;
 
+use hashbrown::HashMap;
+
 use burn_common::{reader::Reader, stub::Mutex};
 use core::{fmt::Debug, isize, ops::Range};
 use core::{iter::repeat, usize};
@@ -198,7 +200,7 @@ where
         ));
         let rank = self.dims().len();
 
-        let mut m = std::collections::HashMap::new();
+        let mut m = HashMap::new();
         for (d, s) in destination_dims.iter().zip(source_dims.iter()) {
             m.insert(*d, *s);
         }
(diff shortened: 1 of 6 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -171,6 +173,61 @@ where
...
+    /// - If the source and destination dimensions are not of the same length.
+    /// - If the source and destination vectors contain duplicate values.
+    /// - If the source and destination vectors contain values that are out of bounds.
+    ///
+    /// # Returns
+    ///
+    /// The tensor with the dimensions moved.
+    // This is a semantic sugar for `permute`. It is used widely enough, so we define a separate Op
+    // for it
+    pub fn movedim<S: MovedimArgs>(self, src: S, dst: S) -> Tensor<B, D, K> {
```
LilDojd, line 199: I've just realised the signature should be more general:

```diff
-    pub fn movedim<S: MovedimArgs>(self, src: S, dst: S) -> Tensor<B, D, K> {
+    pub fn movedim<S1: MovedimArgs, S2: MovedimArgs>(self, src: S1, dst: S2) -> Tensor<B, D, K> {
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -171,6 +173,61 @@ where
...
+    // for it
+    pub fn movedim<S1: MovedimArgs, S2: MovedimArgs>(self, src: S1, dst: S2) -> Tensor<B, D, K> {
+        let source_dims = src.into_dim_vec::<D>();
+        let destination_dims = dst.into_dim_vec::<D>();
+
+        check!(TensorCheck::movedim_args_length(
+            &source_dims,
+            &destination_dims
+        ));
+        let rank = self.dims().len();
```
nathanielsimard, line 207: I think just `D` would work.

crates/burn-tensor/src/tensor/api/check.rs
```diff
@@ -172,6 +173,94 @@ impl TensorCheck {
...
+                TensorError::new("The given dimensions are out of bounds.").details(format!(
+                    "Current tensor has {D} dimensions, but the given dimensions are {:?}.",
+                    dims
+                )),
+            );
+        }
+
+        // Check there are no duplicates
+        let mut uniq = HashSet::new();
+        let duplicates = dims.iter().find(|&&x| !uniq.insert(x));
```
nathanielsimard, line 228: I think we should avoid creating hashmap or hashset when possible, especially when working with sequences < 10 in size, just doing for loops would probably be the best here.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -171,6 +173,61 @@ where
...
+
+        let mut m = HashMap::new();
```
nathanielsimard, line 209: I would avoid using an hashmap here. maybe a BTreeMap would be lighter.

nathanielsimard (changes requested):
Thanks for the contribution!

crates/burn-tensor/src/tensor/api/base.rs
LilDojd: Right, my bad

LilDojd:
Thanks for the review! I have addressed your comments in recent commit

nathanielsimard (approved):

Issue #1535, opened 2024-03-26
Move `full` and `pad` functions to `BaseTensor` trait and make `bool` a proper tensor element type


antimora:
Currently, the `full` function is implemented separately for each tensor type. To make it work for `bool` tensors as well, it was moved to the `BaseTensor` trait. However, this introduced some issues:

1. The `bool` type does not implement the `Element` trait, which is required by functions in `BaseTensor`. 
2. The `Element` trait depends on numeric traits like `ToPrimitive`, `Zero`, and `One`, which `bool` does not implement.
3. Calling `elem()` to convert values to the element type is a temporary workaround and should be cleaned up.

To properly address this, we should:

1. Create our own versions of the `Zero`, `One`, and `ToPrimitive` traits that are not coupled to numeric types.
2. Remove the dependency on `num_traits` in the base `Element` trait.
3. Make the `Element` trait not require numeric traits.
4. Implement the new `Element` trait for `bool`.
5. Move the `full` function to `BaseTensor` once `bool` properly implements `Element`.

This will allow `bool` to be treated as a proper tensor element type, simplify the implementation of `full` and other functions in `BaseTensor`, and remove the need for workarounds like calling `elem()`.

For now, the `pad` function changes should be moved to the numeric tensor implementation to unblock the current PR. The `full` function move and `bool` element type cleanup can be handled separately in this ticket.

antimora:
int_full implementation: 

```rust
    fn bool_full<const D: usize>(
        shape: Shape<D>,
        value: bool,
        device: &Device<B>,
    ) -> BoolTensor<B, D> {
        B::int_equal_elem(
            B::int_zeros(shape, device),
            if value { 0.elem() } else { 1.elem() },
        )
    }
```

antimora:
We need full working for bool tensor because ConstantOfShape can initialize with bool values and we can't use full because of this error:

```
   Compiling onnx-tests v0.14.0 (/Users/dilshod/Projects/burn/crates/burn-import/onnx-tests)
error[E0277]: the trait bound `burn::tensor::Bool: Numeric<B>` is not satisfied
   --> /Users/dilshod/Projects/burn/target/debug/build/onnx-tests-50e78da95971b883/out/model/constant_of_shape_full_like.rs:53:37
    |
53  |         let constantofshape3_out1 = Tensor::full(shape3_out1, true, &*self.device);
[...]

PR #1878, opened 2024-06-11
Implement `Element` for `bool`

laggui:
### Checklist

### Related Issues/PRs

Required for #1535 and `Data` refactor

### Changes

Since the traits `Zero`, `One` and `ToPrimitive` from `num-traits` could not be implemented for `bool`:
- ~~Added new `Zero` and `One` element identities (without `Add` and `Mul` requirement)~~ Removed the traits (previously kept for historical reasons)
- Added new `ToElement` trait (adapted from `num_traits::ToPrimitive`)

`Element` is now implemented for `bool`, which should simplify a couple of things.


crates/burn-ndarray/src/element.rs
```diff
@@ -20,7 +19,6 @@ where
 /// A general element for ndarray backend.
 pub trait NdArrayElement:
     Element
-    + One
```
laggui, line 23: `Element` already requires `One`

crates/burn-ndarray/src/ops/base.rs
```diff
@@ -567,12 +567,14 @@ where
             tensor
                 .array
                 .mapv(|x| {
-                    if x > E::zero() {
-                        E::one()
-                    } else if x < E::zero() {
-                        -E::one()
+                    let zero = 0.elem();
+                    let one = 1.elem::<E>();
```
nathanielsimard, line 571: Do you think it would be better to declare `zero` and `one` outside the `mapv` ?

laggui:
Not sure why clippy failed on that commit specifically, seems unrelated.

antimora (approved):
LGTM

antimora:
Will this fix #1535?

laggui:
> Will this fix #1535?

The ops won't be implemented in this PR but with these changes they could be refactored properly.

nathanielsimard (approved):
LGTM, however I would add num_trait to the NOTICE.md file before merging.

PR #1887, opened 2024-06-13
add dependency management for python

skewballfox:
### Checklist

### Changes

added a pyproject.toml and a few files generated by rye. a consistent easy to setup virtual environment would at the very least make it easy to setup burn when bouncing between systems, and might help new contributors a bit. 

It might be beneficial to make a `.vscode/settings.json` file to point the python extension to the venv since (at least with rye), it will be in the same spot and vscode won't activate by default otherwise since it's off the default search path. plus we can add recommended extensions and configs but that is out of scope for this PR. 
 
### Testing

I set up the environment and created a few ONNX files?


antimora:
It would be helpful to explain how to install the dependencies in README under the onnx-tests project: https://github.com/tracel-ai/burn/blob/main/crates/burn-import/onnx-tests/README.md

skewballfox:
so slight hiccup. If we use rye like this, the `.venv` in the directory onnx-test, which is off the search path if opening the full burn project. So setting up for vscode would mean manually selecting the python interpreter the first time. 

I could try poetry, but IME for tensor libraries it's hit or miss, mainly because the common scientific computing libraries tend to break spec frequently

antimora (approved):
LGTM.

Thanks for project set up improvements.

nathanielsimard (approved):
LGTM

PR #1888, opened 2024-06-13
Remove GraphicsAPI generic for WgpuRuntime

ArthurBrussee:
### Checklist

### Changes

Slightly cheeky PR so bear with me :) 

Ran into some annoying trait bounds while trying to do some stuff and rather than just grumble, felt like I could remove this entirely! A WgpuRuntime is a WgpuRuntime - irregardless of what the graphics API is. This is particularly evident with `runtime::init_existing_device()` - it's possible to register a DX12 device, and yet run it on a BurnRuntime<Vulkan> which just seems wrong!

I understand it's used to manually pick a graphics API, but, you can already do this with

```
runtime::init_sync<Vulkan>();
```

This could also be in RuntimeOptions perhaps, for consistency.

All that said, this is a bit of miscellaneous design change in a PR so thanks for bearing with me :)

### DIFF
--- a/crates/burn-wgpu/src/runtime.rs
+++ b/crates/burn-wgpu/src/runtime.rs
@@ -1,7 +1,7 @@
 use crate::{
     compiler::wgsl,
     compute::{WgpuServer, WgpuStorage},
-    GraphicsApi, WgpuDevice,
+    AutoGraphicsApi, GraphicsApi, WgpuDevice,
 };
 use alloc::sync::Arc;
 use burn_common::stub::RwLock;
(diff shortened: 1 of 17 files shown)


crates/burn-wgpu/src/runtime.rs
```diff
@@ -51,7 +46,8 @@ impl<G: GraphicsApi> Runtime for WgpuRuntime<G> {
 
     fn client(device: &Self::Device) -> ComputeClient<Self::Server, Self::Channel> {
         RUNTIME.client(device, move || {
-            let (adapter, device_wgpu, queue) = pollster::block_on(create_wgpu_setup::<G>(device));
+            let (adapter, device_wgpu, queue) =
+                pollster::block_on(create_wgpu_setup::<AutoGraphicsApi>(device));
```
nathanielsimard, line 50: So if I understand it well, this is where the default runtime is actually created for a device type. It uses the `AutoGraphicsApi` to select the most appropriate adaptor for the device. If you want to specify manually the graphics API for a device type, you have to use `init_sync` or `init_async`!

I don't think this is a bad thing, now the `WgpuBackend` only has typical backend generics, and runtime options can be changed using the init functions! Before merging I would add a bit more documentation on the `WgpuRuntime` and `WgpuBackend` pointing to the right direction to specify backend and runtime options.

ArthurBrussee: Yep that's right! Seemed coherent to me to have one "advanced initialize" place, glad you agree :) 

Have added some docs now, and a random code cleanup, lmk how this looks!

nathanielsimard (approved):

PR #1889, opened 2024-06-13
Make autodiff compile on wasm

ArthurBrussee:
### Checklist

### Changes

Argwhere, nonzero, and various sorting options aren't available on wasm atm as they require sync readbacks.

The autodiff backend tries to use them unconditionally and fails to compile.

In the future, I do really need argwhere/nonzero so that'll be an interesting challenge!


nathanielsimard (approved):
I think the current default implementations are done on the CPU using the data struct, but it shouldn't be hard to write a CubeCL kernel that does those things, so shouldn't be a blocker in the future and should work everywhere.

ArthurBrussee:
Thank you! And ah for argwhere / nonzero I think it's a bit trickier as they are sparse operations - the returned shape depends on how many arguments were true, so it will require some form of GPU readback, which is a pain on wasm :)

Issue #1890, opened 2024-06-14
to_device not working


SupaMaggie70Incorporated:
**Describe the bug**
When a `Module` is created on a GPU device(e.g. Cuda on candle or tch, DiscreteGpu on wgpu), training is just fine. However, when a model is created on cpu and sent to the gpu using Module::to_device(), it is unable to train, though it can evaluate and give losses just fine. The "training" completes way faster, except for the first 1-3 iterations likely caused by first use of functions, but the model appears to not be updated at all.

**To Reproduce**
Create a `Module` on the cpu, send it to the gpu using `to_device()`, and attempt to train it.

**Expected behavior**
One would expect the model to train

**Desktop (please complete the following information):**
 - OS: Ubuntu 2022.04
 - Backends: Candle, tch, wgpu, probably all due to Autodiffbackend
 - Burn version 13.2, most recent at time of this bug report

**Smartphone (please complete the following information):**

**Additional context**
This is probably an issue with AutodiffBackend, as it is shared across backends, and prevents gradients/training but doesn't prevent evaluation and loss.

nathanielsimard:
Experimenting with this a bit, I don't think there is a bug, but maybe a misuse of `Module::to_device`. You can't initialize a module on a device, switch it to another one using `to_device`, and expect the gradients to be kept for the module on the new device. The new module, on the GPU, is considered an in-between node where the gradients are consumed during backprop, and are only useful to compute the gradients for the first module (on the CPU).

What you are probably searching for is the method `Module::fork` where there is no link between the parent module and the new moved module, which can be optimized. Just to illustrate:

```rust
#[test]
fn module_to_device() {
    type B = Autodiff<Wgpu>;
    let device_gpu = WgpuDevice::default();
    let device_cpu = WgpuDevice::Cpu;

    let initializer = Initializer::Ones;

    let tensor_cpu: Param<Tensor<B, 1>> = initializer.init_with([32], None, None, &device_cpu);
    let tensor_gpu = tensor_cpu.clone().to_device(&device_gpu);

    let loss = tensor_gpu.val().tanh().sum();

    let grads = loss.backward();

    let grad_gpu = grads.get(&tensor_gpu.val().into_primitive());
    let grad_cpu = grads.get(&tensor_cpu.val().into_primitive());

    assert!(grad_cpu.is_some(), "Passes");
    assert!(grad_gpu.is_some(), "Fails");
}
```

But it's the oposite when you use fork:

```rust
#[test]
[...]

PR #1901, opened 2024-06-17
Doc: Improve module to_device/fork docs

nathanielsimard:
Related #1890

### DIFF
--- a/crates/burn-core/src/module/base.rs
+++ b/crates/burn-core/src/module/base.rs
@@ -97,17 +97,18 @@ pub trait Module<B: Backend>: Clone + Send + core::fmt::Debug {
     ///
     /// # Notes
     ///
-    /// This is similar to [to_device](Module::to_device), but it ensures the module will
-    /// have its own autodiff graph.
+    /// This is similar to [to_device](Module::to_device), but it ensures the output module on the
+    /// new device will have its own autodiff graph.
     fn fork(self, device: &B::Device) -> Self;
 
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/module/base.rs
```diff
@@ -97,17 +97,18 @@ pub trait Module<B: Backend>: Clone + Send + core::fmt::Debug {
...
 
     /// Move the module and all of its sub-modules to the given device.
     ///
     /// # Warnings
     ///
-    /// The device operations will be registered in the autodiff graph. Therefore, be sure to call
-    /// backward only one time even if you have the same module on multiple devices. If you want to
-    /// call backward multiple times, look into using [fork](Module::fork) instead.
+    /// The operation supports autodiff and it will be registered when activated. However, this may
+    /// not be what you want. The output model will be an intermediary model, meaning that you
```
laggui, line 109: What's an "intermediary model"? We might want to clarify 🤔

nathanielsimard: I took the term to mirror `intermediary tensor` when computing backprop. Which is a node that has gradients, but only as a mean to compute other gradients, and therefore their value is flushed during the backward calculation. Not sure how can I make it more clear without defining how autodiff works 😅

laggui: Ahhh ok! I get it now.

Yeah we can keep it like that.

laggui (approved):

PR #1905, opened 2024-06-17
feat: Make RetroForward public

femshima:
Currently, RetroForward is private, making it impossible to implement a memory-bound operation.
Here, I made RetroForward and relevant struct/functions available to outside of burn-autodiff.

### Checklist

`tests::jit_fusion::remainder::tests::should_be_zero` test failed both on main and this branch, therefore I could not confirm the `run-checks all` success.

### Related Issues/PRs

#1358: Implemented RetroForward.

### Changes

- RetroForward and BackwardStates were made public.
- Some of the BackwardStates' functions were also made public.
- Removed reference to State in BackwardStates' doc comment.

### Testing

No unit test was added in this PR, though I confirmed that I now can use `retro_unary` in my project.

### DIFF
--- a/crates/burn-autodiff/src/checkpoint/mod.rs
+++ b/crates/burn-autodiff/src/checkpoint/mod.rs
@@ -1,7 +1,8 @@
 /// Checkpointer module
 pub mod base;
 pub(crate) mod builder;
-pub(crate) mod retro_forward;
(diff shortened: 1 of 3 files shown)


louisfd (approved):
LGTM, thanks!

PR #1908, opened 2024-06-18
Fix wasm error as it can't use autotune-persistent-cache

ArthurBrussee:
### Checklist

### Changes

Wasm can't use autotune-persistent-cache. However, it's enabled by default, and some burn crates pull in burn-compute with defaults.

Exlude these defaults in some places and instead let the default propogate.

My cargo-fu isn't very good so I'm not entirely sure if this is the preferred way to do this! One alternative for example is to change all cfgs to `and(cfg(feature=persistent), not(target(wasm))`, but I'm not sure.


nathanielsimard (approved):
I think it's fine, we set the right feature flags in `burn-jit` anyways.

PR #1909, opened 2024-06-19
Feat/cube/compile error

nathanielsimard:
Introduce basic error handling into CubeCL. Obviously not all errors are handled, but we should always improve the message when we encounter an error.

### DIFF
--- a/crates/burn-cube-macros/src/codegen_function/base.rs
+++ b/crates/burn-cube-macros/src/codegen_function/base.rs
@@ -125,7 +125,14 @@ pub(crate) fn codegen_expr_with_comptime(
                 syn::Expr::Unary(op) => codegen_unary(op, loop_level, variable_tracker),
                 syn::Expr::Field(field) => codegen_field(field, loop_level, variable_tracker),
                 syn::Expr::Struct(struct_) => codegen_struct(struct_, loop_level, variable_tracker),
-                _ => panic!("Codegen: Unsupported {:?}", expr),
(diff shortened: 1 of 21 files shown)


crates/burn-cube-macros/src/codegen_function/base.rs
```diff
@@ -125,7 +125,14 @@ pub(crate) fn codegen_expr_with_comptime(
...
                 syn::Expr::Field(field) => codegen_field(field, loop_level, variable_tracker),
                 syn::Expr::Struct(struct_) => codegen_struct(struct_, loop_level, variable_tracker),
-                _ => panic!("Codegen: Unsupported {:?}", expr),
+                syn::Expr::Range(range) => syn::Error::new_spanned(
+                    range,
+                    "Range is not supported, use [range](cubecl::prelude::range) instead.",
+                )
+                .to_compile_error(),
+                _ => {
+                    syn::Error::new_spanned(expr, "Expression is not supported").to_compile_error()
```
louisfd, line 134: I think losing the expr debug information will remove helpful info when developing Cube, but I guess it would be very ugly for the user to see.

louisfd (approved):
Looking good! Can't wait to write some CubeCL with that

crates/burn-cube-macros/src/codegen_function/base.rs
nathanielsimard: Well the expr is supposed to be highlighted, so I think it's better than before.

Issue #1357, opened 2024-02-23
Print model structure like with PyTorch


antimora:
### Feature description

Want to see a models structure at a glance like when you print a pytorch model:

```python
import whisper
model = whisper.load_model("tiny")
print(model)
```
Result:

```
Whisper(
  (encoder): AudioEncoder(
    (conv1): Conv1d(80, 384, kernel_size=(3,), stride=(1,), padding=(1,))
    (conv2): Conv1d(384, 384, kernel_size=(3,), stride=(2,), padding=(1,))
    (blocks): ModuleList(
      (0-3): 4 x ResidualAttentionBlock(
        (attn): MultiHeadAttention(
          (query): Linear(in_features=384, out_features=384, bias=True)
          (key): Linear(in_features=384, out_features=384, bias=False)
          (value): Linear(in_features=384, out_features=384, bias=True)
          (out): Linear(in_features=384, out_features=384, bias=True)
        )
        (attn_ln): LayerNorm((384,), eps=1e-05, elementwise_affine=True)
        (mlp): Sequential(
          (0): Linear(in_features=384, out_features=1536, bias=True)
          (1): GELU(approximate='none')
          (2): Linear(in_features=1536, out_features=384, bias=True)
        )
        (mlp_ln): LayerNorm((384,), eps=1e-05, elementwise_affine=True)
      )
    )
    (ln_post): LayerNorm((384,), eps=1e-05, elementwise_affine=True)
  )
  (decoder): TextDecoder(
    (token_embedding): Embedding(51865, 384)
    (blocks): ModuleList(
      (0-3): 4 x ResidualAttentionBlock(
        (attn): MultiHeadAttention(
          (query): Linear(in_features=384, out_features=384, bias=True)
          (key): Linear(in_features=384, out_features=384, bias=False)
          (value): Linear(in_features=384, out_features=384, bias=True)
          (out): Linear(in_features=384, out_features=384, bias=True)
        )
        (attn_ln): LayerNorm((384,), eps=1e-05, elementwise_affine=True)
        (cross_attn): MultiHeadAttention(
          (query): Linear(in_features=384, out_features=384, bias=True)
          (key): Linear(in_features=384, out_features=384, bias=False)
          (value): Linear(in_features=384, out_features=384, bias=True)
          (out): Linear(in_features=384, out_features=384, bias=True)
        )
        (cross_attn_ln): LayerNorm((384,), eps=1e-05, elementwise_affine=True)
        (mlp): Sequential(
[...]

PR #1912, opened 2024-06-20
Print model structure like with PyTorch - Part 1

antimora:
This is part 1 of PRs. This is initial version that works and I would like to merge before continuing with customizing remaining NN modules. I did for Conv1d, Linear and LayerNorm modules. I will do for the rest in other PRs.

### Checklist

### Related Issues/PRs

#1357

### Changes

Updated the burn-derive and burn-core to support module display with customization.

### Testing

1. Default display:
```
Whisper {
  encoder: Encoder {
    conv1: Conv1d {stride: 1, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    conv2: Conv1d {stride: 2, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    positional_embedding: ParamTensor {rank: 2, shape: [1500, 384]}
    blocks: Vec<0..4> {
      0: ResidualAttentionBlock {
        attn: MultiHeadAttention {
          n_head: 6
          n_state: 384
          query: Linear {d_input: 384, d_output: 384, bias: true}
          key: Linear {d_input: 384, d_output: 384, bias: false}
          value: Linear {d_input: 384, d_output: 384, bias: true}
          out: Linear {d_input: 384, d_output: 384, bias: true}
        }
        attn_ln: LayerNorm {d_model: 384, epsilon: 0.00001}
```

2. Show all attributes:
```
Whisper {
  encoder: Encoder {
    conv1: Conv1d {weight: ParamTensor {rank: 3, shape: [384, 80, 3]}, bias: ParamTensor {rank: 1, shape: [384]}, stride: 1, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    conv2: Conv1d {weight: ParamTensor {rank: 3, shape: [384, 384, 3]}, bias: ParamTensor {rank: 1, shape: [384]}, stride: 2, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    positional_embedding: ParamTensor {rank: 2, shape: [1500, 384]}
    blocks: Vec<0..4> {
      0: ResidualAttentionBlock {
        attn: MultiHeadAttention {
          n_head: 6
          n_state: 384
          query: Linear {weight: ParamTensor {rank: 2, shape: [384, 384]}, bias: ParamTensor {rank: 1, shape: [384]}}
          key: Linear {weight: ParamTensor {rank: 2, shape: [384, 384]}, bias: None}
          value: Linear {weight: ParamTensor {rank: 2, shape: [384, 384]}, bias: ParamTensor {rank: 1, shape: [384]}}
          out: Linear {weight: ParamTensor {rank: 2, shape: [384, 384]}, bias: ParamTensor {rank: 1, shape: [384]}}
        }
        attn_ln: LayerNorm {gamma: ParamTensor {rank: 1, shape: [384]}, beta: ParamTensor {rank: 1, shape: [384]}, epsilon: 0.00001}
```
3. Show number of params:
```
Whisper {
  encoder: Encoder {
    conv1: Conv1d {stride: 1, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1), params: 92544}
    conv2: Conv1d {stride: 2, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1), params: 442752}
    positional_embedding: ParamTensor {rank: 2, shape: [1500, 384]}
    blocks: Vec<0..4> {
      0: ResidualAttentionBlock {
        attn: MultiHeadAttention {
          n_head: 6
          n_state: 384
          query: Linear {d_input: 384, d_output: 384, bias: true, params: 147840}
          key: Linear {d_input: 384, d_output: 384, bias: false, params: 147456}
          value: Linear {d_input: 384, d_output: 384, bias: true, params: 147840}
          out: Linear {d_input: 384, d_output: 384, bias: true, params: 147840}
          params: 590976
        }
        attn_ln: LayerNorm {d_model: 384, epsilon: 0.00001, params: 768}
```
4. Show param ids:
```
Whisper {
  encoder: Encoder {
    conv1: Conv1d {stride: 1, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    conv2: Conv1d {stride: 2, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    positional_embedding: ParamTensor {rank: 2, shape: [1500, 384], id: apd49r4c00}
    blocks: Vec<0..4> {
      0: ResidualAttentionBlock {
        attn: MultiHeadAttention {
          n_head: 6
          n_state: 384
          query: Linear {d_input: 384, d_output: 384, bias: true}
          key: Linear {d_input: 384, d_output: 384, bias: false}
          value: Linear {d_input: 384, d_output: 384, bias: true}
          out: Linear {d_input: 384, d_output: 384, bias: true}
        }
        attn_ln: LayerNorm {d_model: 384, epsilon: 0.00001}
```

### DIFF
--- a/crates/burn-core/src/module/display.rs
+++ b/crates/burn-core/src/module/display.rs
@@ -436,7 +436,7 @@ impl Content {
     /// # Returns
     ///
     /// Updated `Content` instance with the single item added.
-    pub fn add_single<T: ModuleDisplay>(mut self, value: &T) -> Self {
+    pub fn add_single<T: ModuleDisplay + ?Sized>(mut self, value: &T) -> Self {
         if !self.attributes.is_empty() {
             panic!("Cannot add single item when attributes are set.");
         }
--- a/crates/burn-core/src/module/display.rs
+++ b/crates/burn-core/src/module/display.rs
@@ -10,14 +10,32 @@ use core::fmt::{Display, Write};
 /// Default display settings for a module.
 pub trait ModuleDisplayDefault {
     /// Attributes of the module used for display purposes.
+    ///
+    /// # Arguments
+    ///
+    /// * `_content` - The content object that contains display settings and attributes.
+    ///
+    /// # Returns
+    ///
+    /// An optional content object containing the display attributes.
     fn content(&self, _content: Content) -> Option<Content>;
 }
 
 /// Trait to implement custom display settings for a module.
+///
+/// In order to implement custom display settings for a module,
+/// 1. Add #[custom_display] attribute to the module struct after #[derive(Module)]
+/// 2. Implement ModuleDisplay trait for the module
 pub trait ModuleDisplay: ModuleDisplayDefault {
-    //TODO mention that attribute needs to be passed
-
     /// Formats the module with provided display settings.
+    ///
+    /// # Arguments
+    ///
+    /// * `passed_settings` - Display settings passed to the module.
+    ///
+    /// # Returns
+    ///
+    /// A string representation of the formatted module.
     fn format(&self, passed_settings: DisplaySettings) -> String {
         let settings = if let Some(custom_settings) = self.custom_settings() {
             custom_settings.inherit(passed_settings)
@@ -85,11 +103,23 @@ pub trait ModuleDisplay: ModuleDisplayDefault {
     }
 
     /// Custom display settings for the module.
+    ///
+    /// # Returns
+    ///
+    /// An optional display settings object.
     fn custom_settings(&self) -> Option<DisplaySettings> {
         None
     }
 
     /// Custom attributes for the module.
+    ///
+    /// # Arguments
+    ///
+    /// * `_content` - The content object that contains display settings and attributes.
+    ///
+    /// # Returns
+    ///
+    /// An optional content object containing the custom attributes.
     fn custom_content(&self, _content: Content) -> Option<Content> {
         None
     }
@@ -121,7 +151,7 @@ impl Default for DisplaySettings {
     fn default() -> Self {
         DisplaySettings {
             show_param_id: None,
-            show_all_attributes: None, // by default, show only the attributes that are not default
+            show_all_attributes: None,
             show_num_parameters: None,
             new_line_after_attribute: None,
             indentation_size: None,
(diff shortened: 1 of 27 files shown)


crates/burn-core/src/module/display.rs
```diff
@@ -0,0 +1,543 @@
...
+                write!(result, ", params: {}", self.num_params()).unwrap();
+            }
+        }
+        if settings.new_line_after_attribute() {
+            write!(result, "{indent_close_braces}}}").unwrap();
+        } else {
+            write!(result, "}}").unwrap();
+        }
+
+        result
```
nathanielsimard, line 118: I'm unsure, but maybe we should extract this into a function that we call inside the default implementation. This way, the function could be compiled only once and reused for all modules. In the current case, I think the compiler is generating that function for all types, creating a bigger binary with slower compile time. This is just an intuition and it is definitely hard to measure, but I think it's not a bad thing to extract this into one function.

crates/burn-core/src/module/display.rs
```diff
@@ -0,0 +1,543 @@
...
+            updated.show_num_parameters = Some(show_num_parameters);
+        }
+
+        if let Some(new_line_after_attribute) = top.new_line_after_attribute {
+            updated.new_line_after_attribute = Some(new_line_after_attribute);
+        }
+
+        if let Some(indentation_size) = top.indentation_size {
+            updated.indentation_size = Some(indentation_size);
+        }
```
nathanielsimard, line 289: We we want to only override values that are missing from the current settings?

crates/burn-core/src/module/param/primitive.rs
```diff
@@ -52,6 +57,20 @@ where
     }
 }
 
+impl<T: ModuleDisplay> ModuleDisplayDefault for Option<T> {
+    fn content(&self, content: Content) -> Option<Content> {
+        match self {
+            Some(module) => content.add_single(module).optional(),
+            None => {
+                let none_string = "None".to_string();
+                content.add_single(&none_string).optional()
```
nathanielsimard, line 66: could be `content.add_single("None").optional()` I think?

crates/burn-core/src/module/param/tensor.rs
```diff
@@ -147,6 +150,22 @@ impl<const D: usize, B: Backend> Module<B> for Param<Tensor<B, D>> {
     }
 }
 
+impl<const D: usize, B: Backend> ModuleDisplayDefault for Param<Tensor<B, D>> {
```
nathanielsimard, line 153: Could you have a single implementation generic over the kind?

nathanielsimard (commented):
Only a few comments, but it's looking good, thanks a lot !

crates/burn-core/src/module/display.rs
antimora: To clarify and confirm I asked these question on Rust forum:

In Rust, when implementing a default method as part of a trait that can be implemented multiple times, is it true that:

1. The compiler generates this function separately for all types implementing the trait?
2. This approach potentially leads to a larger binary size and slower compile times?
3. Extracting the common functionality into a separate function and calling it from the default trait implementation could optimize this?
4. Such an extraction would allow the function to be compiled only once and reused across all modules?

If these points are accurate, what are the trade-offs to consider when deciding whether to implement this optimization?

Reply (https://discord.com/channels/442252698964721669/443150878111694848/1253892142267498599):

1. is true if you dont provide another implementation of the function
2. is not really true, the compiler merges identical functions so it wont increase the binary size unless the function does different things depending on the type
3. wouldnt really help because of 2, unless you extract only the identical stuff into a function and left the different stuff in the default impl
4. yes, if the final code is identical for all types

So in the end, I do not think it makes sense to extract the method because the function does behave differently depending on the Self type. Even if I would extract, it would be pretty much the same: self needs to be passed.

crates/burn-core/src/module/display.rs
antimora: No, we completely want to override from the very top. So here is a use case:

I have my Whisper model and I implementing my custom settings:

```rust
impl<B: Backend> ModuleDisplay for Whisper<B> {
    fn custom_settings(&self) -> Option<burn::module::DisplaySettings> {
        DisplaySettings::new()
             .with_new_line_after_attribute(true)
            .optional()
    }
}
```

By default if there are no settings set, `new_line_after_attribute` will give true at one level. At `Conv1d` level there is a custom settings that sets `new_line_after_attribute` to false to make display render compact. But since I set new_line_after_attribute to true at Whisper level, it's passed down overriding anything else, include at `Conv1d` level.

crates/burn-core/src/module/param/primitive.rs
antimora: I get this error:

```rust
error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> /Users/dilshod/Projects/burn/crates/burn-core/src/module/param/primitive.rs:64:40
    |
64  |             None => content.add_single("None").optional(),
    |                             ---------- ^^^^^^ doesn't have a size known at compile-time
    |                             |
    |                             required by a bound introduced by this call
    |
```

But I fixed it with this new signature:

```rust
    pub fn add_single<T: ModuleDisplay + ?Sized>(mut self, value: &T) -> Self {
```

Now, it works.

crates/burn-core/src/module/param/tensor.rs
antimora: I get this error:

```
the trait bound `burn_tensor::Tensor<B, D, K>: param::base::Parameter` is not satisfied
the trait `param::base::Parameter` is not implemented for `burn_tensor::Tensor<B, D, K>`
```

So to take advantage of individual implementation, I have added tensor kind. This what it looks like now:

```
Whisper {
  encoder: Encoder {
    conv1: Conv1d {stride: 1, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    conv2: Conv1d {stride: 2, kernel_size: 3, dilation: 1, groups: 1, padding: Explicit(1)}
    positional_embedding: ParamTensor {rank: 2, shape: [1500, 384], kind: float, id: usor44svfc}
 ```

**Comment by antimora** [comment]
Implemented display for Conv2d, Dropout, BathcNorm modules. And made show params on by default. 

Here is how MNIST model print out looks like:

```
Model {
  conv1: ConvBlock {
    conv: Conv2d {stride: [1, 1], kernel_size: [3, 3], dilation: [1, 1], groups: 1, padding: Valid, params: 80}
    norm: BatchNorm {num_features: 8, momentum: 0.1, epsilon: 0.00001, params: 32}
    activation: Gelu
    params: 112
  }
  conv2: ConvBlock {
    conv: Conv2d {stride: [1, 1], kernel_size: [3, 3], dilation: [1, 1], groups: 1, padding: Valid, params: 1168}
    norm: BatchNorm {num_features: 16, momentum: 0.1, epsilon: 0.00001, params: 64}
    activation: Gelu
    params: 1232
  }
  conv3: ConvBlock {
    conv: Conv2d {stride: [1, 1], kernel_size: [3, 3], dilation: [1, 1], groups: 1, padding: Valid, params: 3480}
    norm: BatchNorm {num_features: 24, momentum: 0.1, epsilon: 0.00001, params: 96}
    activation: Gelu
    params: 3576
  }
  dropout: Dropout {prob: 0.5}
  fc1: Linear {d_input: 11616, d_output: 32, bias: false, params: 371712}
  fc2: Linear {d_input: 32, d_output: 10, bias: false, params: 320}
  activation: Gelu
  params: 376952
}
```

nathanielsimard (approved):

PR #1936, opened 2024-06-28
Consistent sync/async handling, allow more functions to be async for wasm.

ArthurBrussee:
### Checklist

### Changes

Currently, there is a split between wasm/non-wasm targets. Some functions are sync by default, and async just on wasm, which forces you to conditionally compile them. Other functions are always sync, but migth panic on wasm.

This change makes it so that API is consistently _not_ async, but allows some more functions (argwhere, nonzero) to be called as async if need be. All other functions that might need to be async, now instead panic if they really cannot synchronize. This means the wasm-sync flag can be removed - things will "just work" if you use a backend with sync data reads (ndarray), and might get panics on WebGPU, telling you you *really* have to switch to async. When going through the trouble of async-ifying your code you can now also just keep things async, rather than having to conditionally compile & have futures block within your async loop which is generally not recommend.

Engineering wise, I tried to have the async functions now use "normal" async-ness. Before, it used some mix of a futures, and a "Reader" trait which basically implemented a small Future with continuations etc. This should be a bit more consistent. [Performance wise, it all compiles away if the future resolves immediatly (even if the future is boxed).](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=73a907dc091c0c83efc86e0fae4d8bbc)

Some other misc notes:
- The async sorting functions are gone. We _could_ re-introduce all sorting with async versions, but really sorting isn't inherently async, it just happens that it doesn't have a GPU implementation yet, so it seems like a different kind of beast.
- MSPC now uses async channels. It's necesarry at least for reading, and mixing a normal and async channel didn't seem necesarry.
- There was already a futures_intrusive channel as a dep I saw later on. async-channel is more up to data afaik so maybe better to keep that.
- If in the future WASM can be properly threaded / block / ... it should hopefully now be easier to just deprecate / delete all _async methods as people can rely on the normal API being sync.

### Testing

Burn tests, internal project using the async methods on wasm for readbacks.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1448,7 +1448,7 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
     /// which is more high-level and designed for public use.
     fn into_data_async<const D: usize>(
         tensor: Self::Primitive<D>,
-    ) -> impl Future<Output = TensorData>;
+    ) -> impl Future<Output = TensorData> + Send;
 
     /// Creates a tensor from the given data.
     ///
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -876,7 +876,7 @@ where
     pub fn into_scalar(self) -> K::Elem {
         crate::try_read_sync(self.into_scalar_async()).expect(
             "Failed to read tensor data synchronously. This can happen on platforms 
-            that don't support blocking futures like WASM. Try into_scalar_async instead..",
+            that don't support blocking futures like WASM. Try into_scalar_async instead.",
         )
     }
 
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -6,7 +6,6 @@ use alloc::format;
 use alloc::string::String;
 use alloc::vec;
 
-use burn_common::reader::try_read_sync;
 use burn_common::stub::Mutex;
 use core::future::Future;
 use core::iter::repeat;
@@ -675,7 +674,7 @@ where
 
     /// Converts the data of the current tensor.
     pub fn into_data(self) -> TensorData {
-        burn_common::reader::try_read_sync(self.into_data_async()).expect(
+        crate::try_read_sync(self.into_data_async()).expect(
             "Failed to read tensor data synchronously. 
         This can happen on platforms that don't support blocking futures like WASM. 
         If possible, try using into_data_async instead.",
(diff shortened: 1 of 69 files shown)


antimora:
Thanks for the PR!

no-tests are failing currently.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -875,22 +872,22 @@ where
...
     /// If the tensor doesn't have one element.
-    #[cfg(any(feature = "wasm-sync", not(target_family = "wasm")))]
+    /// If the backend fails to read the tensor data synchronously.
     pub fn into_scalar(self) -> K::Elem {
-        check!(TensorCheck::into_scalar(&self.shape()));
-        let x = self.into_data().iter().next().unwrap();
-        x
+        crate::try_read_sync(self.into_scalar_async()).expect(
+            "Failed to read tensor data synchronously. This can happen on platforms 
+            that don't support blocking futures like WASM. Try into_scalar_async instead..",
```
nathanielsimard, line 879: One extra '.' at the end.

crates/burn-tensor/src/tensor/api/sort.rs
```diff
@@ -37,43 +37,7 @@ where
...
-    dim: usize,
-    descending: bool,
-) -> K::Primitive<D>
-where
-    <K as BasicOps<B>>::Elem: Element,
-{
-    let device = K::device(&tensor);
-    let data = K::into_data(tensor).read().await;
-
+    let data = try_read_sync(K::into_data_async(tensor)).expect("Failed to synchonously read tensor data. This operation is not supported ntil this backend has a GPU sorting implementation.");
```
nathanielsimard, line 40: typo: ntil => until.

crates/burn-tensor/src/tensor/api/sort.rs
```diff
@@ -129,44 +92,7 @@ where
...
-    dim: usize,
-    descending: bool,
-) -> (K::Primitive<D>, IntTensor<B, D>)
-where
-    <K as BasicOps<B>>::Elem: Element,
-{
-    let device = K::device(&tensor);
-    let data = K::into_data(tensor).read().await;
-
+    let data = try_read_sync(K::into_data_async(tensor)).expect("Failed to synchonously read tensor data. This operation is not supported ntil this backend has a GPU sorting implementation.");
```
nathanielsimard, line 95: Same typo

nathanielsimard (changes requested):
Just found a couple of typos, but otherwise I think it's going to be easier this way to handle asyncness on wasm. Normaly feature flags should be additive, not break an API, so this fixes that.

ArthurBrussee:
Thanks for the quick review, glad you think this can work until WASM sorts out the woeful threading situation :) Fixed the typos.

nathanielsimard (approved):

PR #1939, opened 2024-06-28
Perf: cube reuse shape and strides

nathanielsimard:
Small optimization to avoid reading the global array of metadata when accessing a constant shape.


louisfd (approved):
LGTM

PR #1958, opened 2024-07-02
Always derive Cube features from adapter

ArthurBrussee:
### Checklist

### Changes

The new cube features are derived from adapter.features(), but then also required to be specified in eg. init_existing_device. Besides being a bit annoying it can also be a problem - specifying a feature does _not_ mean it's now valid to use, the adapter from the existing device still defines what is/isn't possible.

So, instead in all flavours just derive what functions there really are.


nathanielsimard (approved):
LGTM thanks !

PR #1959, opened 2024-07-02
Feat/cube/matmul tiling

louisfd:
Fully vectorized Matmul tiling 2d in Cube

It should replace the older tiling 2d versions soon, but right now it still needs to put inputs contiguous.


crates/burn-jit/src/tests/matmul.rs
```diff
@@ -406,6 +406,151 @@ mod tests {
         }
     }
 
+    mod tiling2d_cube {
```
louisfd, line 409: Tests have been played with and don't necessarily fit their name

louisfd:
In #1994 instead

Issue #1752, opened 2024-05-09
Better memory management in Burn Compute


mepatrick73:
### Feature description
Implementing a memory management allocation strategy that is more aggressive in reusing old memory.
- Should reduce memory fragmentation
- Should reduce the memory allocation overhead by reusing unused memory

### Feature motivation
Currently the memory allocator just frees unused memory every n allocations. This leads to a very "zig-zaggy" memory usage pattern. Furthermore, if a chunk has some free space, it will not be used by another allocation leading to internal fragmentation.

### (Optional) Suggest a Solution
Instead of freeing memory every n allocations, wait until memory usage hits a certain threshold and then try to free unused memory.
Use the idea of offset to reuse free memory within a chunk if there is enough space.

PR #1960, opened 2024-07-02
Feat/fixed chunk alloc by class

mepatrick73:
### Checklist

### Related Issues/PRs

#1752
### Changes
Added sizes classes in memory management. If an allocation is asked for that is under the class threshold, that class threshold will be allocated. This is so that when we alloc bigger and bigger slices, we are stilll able to reuse. I.E if I ask 128mb, 150mb then 170mb, I'm still able to reuse allocations without having to realloc. Also recoded extend max memory to work. It's not currently being used due to computation being slow after being used. However it was tested for correctness.

### Testing

Tested against simple mm on Text-classification ag-news-training on release mode and saw 60% speedup improvement


crates/burn-compute/src/memory_management/memory_pool/base.rs
```diff
@@ -198,6 +218,11 @@ impl MemoryPool {
         size: usize,
         sync: Sync,
     ) -> MemoryPoolHandle {
+        //if self.recently_allocated_size > (0.8 * self.max_chunk_size as f64) as usize {
+        //    self.extend_max_memory(storage);
+        //    self.recently_added_chunks = Vec::new();
+        //    self.recently_allocated_size = 0;
+        //}
```
nathanielsimard, line 225: Dead code

crates/burn-compute/src/memory_management/memory_pool/base.rs
```diff
@@ -371,6 +397,7 @@ impl MemoryPool {
     ) -> ChunkHandle {
         let padding = calculate_padding(size);
         let effective_size = size + padding;
+        //let effective_size = self.max_chunk_size;
```
nathanielsimard, line 400: dead code

nathanielsimard (changes requested):

nathanielsimard (approved):

PR #1962, opened 2024-07-03
Dynamic memory management preset + updated wgpu buffer memory management

nathanielsimard:
It's now more stable and much faster. The maximum memory usage is probably higher than before, but this is because it shows the maximum amount of memory used by the GPU, not the current amount. There is no deallocation, and we allocate in a way that is fragmentation-friendly for better speed, possibly at the cost of slightly higher maximum memory usage.


mepatrick73 (approved):
Looks good to me! Maybe when we do chunk removal, we don't want to reset the cursor_chunk to 0, but for now I think it's fine.

Issue #464, opened 2023-07-04
4 bit / 8 bit model training / inference capabilities


okpatil4u:
### Feature description

Llama.cpp has gained traction as it enables inferring models at 2,3,4,5,6,8,16 and 32 bit precision. Would it be possible to enable inference level quantisation capabilities into burn ?

### Feature motivation

Faster inference empowers deployment at edge (either web or laptops)

nathanielsimard:
We are currently working on a GPU backend based on [wgpu](https://wgpu.rs), quantization is on our roadmap.

thedevleon:
Related to this, I would also love to see support for i8 , i16 , i32 , i64 quantization for inference, i.e. to run models on embedded MCUs without a dedicated FPU (i.e. esp32c3, esp32c6) with the NdArray backend.

thedevleon:
Coming back to this after a while, seems that [quantization support](https://github.com/tracel-ai/burn/pull/1963) has been added, but NdArray [still only supports float tensors](crates/burn-ndarray/src/tensor.rs) - is there any plans to add support for integer tensors in NdArray as well to make used of the quantized model?

laggui:
> Coming back to this after a while, seems that [quantization support](https://github.com/tracel-ai/burn/pull/1963) has been added, but NdArray [still only supports float tensors](crates/burn-ndarray/src/tensor.rs) - is there any plans to add support for integer tensors in NdArray as well to make used of the quantized model?

Actually, ndarray has `NdArrayQTensor` (currently only i8 quantization type supported):

https://github.com/tracel-ai/burn/blob/2338912dafb7f9f18c8d14b851f7c7fd6ec542c7/crates/burn-ndarray/src/backend.rs#L57

But currently all [`QTensorOps`](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/ops/qtensor.rs) just perform dequantize -> float op -> quantize by default.

Support for specific ops will come, but ndarray might not be the first targeted backend. Everything is in place though so adding new qtensor ops should be almost as straightforward as regular tensor ops 🙂

polarathene:
FWIW, the popular [`mistral.rs` is open towards switching from Candle to Burn](https://github.com/EricLBuehler/mistral.rs/issues/1345#issuecomment-2944419115) but a blocker raised was quantization support.

It seems like steady progress is being made, is there any roadmap / checklist that can be tracked or is this issue the primary source of information on where Burn is at for this support?

---

[...]

PR #1963, opened 2024-07-03
Add tensor quantization

laggui:
### Checklist

### Related Issues/PRs

Progress towards #464 

### Changes

Support for static per-tensor quantization.
- Added new `DType::QFloat`
- Added backend trait associative type `QuantizedTensorPrimitive`
- Added `QTensorOps` for `quantize`/`dequantize` ops
- Added `TensorPrimitive` enum for float tensors to include float and qfloat (quantized type)
  - All existing float ops now automatically retrieve the `.tensor()` (dequantized) before calling the backend implementation on the `FloatTensorPrimitive`
- Added `QuantizationStrategy` enum with per-tensor affine and symmetric int8 quantization

*Note: QAT support should be added in a future PR*

### Testing

Added unit tests for affine and symmetric per-tensor quantization


crates/burn-tensor/src/tensor/api/autodiff.rs
```diff
@@ -53,13 +87,13 @@ impl<B: AutodiffBackend> BasicAutodiffOps<B> for Float {
...
     ) -> <Self::InnerKind as TensorKind<<B as AutodiffBackend>::InnerBackend>>::Primitive<D> {
-        B::inner(tensor)
+        TensorPrimitive::Float(B::inner(tensor.tensor()))
     }
 
     fn from_inner<const D: usize>(
         inner: <Self::InnerKind as TensorKind<<B as AutodiffBackend>::InnerBackend>>::Primitive<D>,
     ) -> <Self as TensorKind<B>>::Primitive<D> {
-        B::from_inner(inner)
+        TensorPrimitive::Float(B::from_inner(inner.tensor()))
```
nathanielsimard, line 96: We will need to implement a q_inner and q_from_inner at some point.

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -119,23 +137,34 @@ where
     /// }
     /// ```
     pub fn int(self) -> Tensor<B, D, Int> {
-        Tensor::new(B::float_into_int(self.primitive))
+        // NOTE: should `tensor.int()` return the quantized values?
```
nathanielsimard, line 140: No since we don't support QInt.

nathanielsimard (approved):
Some comments, but LGTM great job!

crates/burn-tensor/src/tensor/api/float.rs
laggui: Whoops, that comment should have been removed 😄

crates/burn-tensor/src/tensor/api/autodiff.rs
laggui: Oh, this should be fairly straightforward. I'll add it.

laggui:
Not sure what's up with the wgpu tests, they seem to be failing intermittently even on main

```
error: test failed, to rerun pass `-p burn-wgpu --lib`
  
  Caused by:
    process didn't exit successfully: `/home/runner/work/burn/burn/target/debug/deps/burn_wgpu-c73d9b81b38abf9c --color=always` (signal: 11, SIGSEGV: invalid memory reference)
```

nathanielsimard:
> Not sure what's up with the wgpu tests, they seem to be failing intermittently even on main
> 
> ```
> error: test failed, to rerun pass `-p burn-wgpu --lib`
>   
>   Caused by:
>     process didn't exit successfully: `/home/runner/work/burn/burn/target/debug/deps/burn_wgpu-c73d9b81b38abf9c --color=always` (signal: 11, SIGSEGV: invalid memory reference)
> ```

Working on it

PR #1964, opened 2024-07-04
Add subtract tensor from scalar for ONNX sub op

johnhuichen:
### Checklist

### Related Issues/PRs

[Help Wanted: Implementing ONNX Ops](https://github.com/tracel-ai/burn/issues/1714)

### Changes

Previous sub and sub_int do not handle a scalar subtracting a tensor. This is a problem when implementing ONNX ops like pad. 

Added support for sub and sub_int to handle that scenario

### Testing

```bash
cargo xtask run-checks all
```

crates/burn-import/onnx-tests
```bash
cargo test 
```

crates/burn-import
```bash
cargo test
```


crates/burn-import/src/burn/node/binary.rs
```diff
@@ -131,6 +131,9 @@ impl BinaryNode {
             (Type::Tensor(_), Type::Tensor(_)) => move |lhs, rhs| quote! { #lhs.sub(#rhs) },
             (Type::Tensor(_), Type::Scalar(_)) => move |lhs, rhs| quote! { #lhs.sub_scalar(#rhs) },
             (Type::Scalar(_), Type::Scalar(_)) => move |lhs, rhs| quote! { #lhs - #rhs },
+            (Type::Scalar(_), Type::Tensor(_)) => {
+                move |lhs, rhs| quote! { #rhs.mul_scalar(-1).add_scalar(#lhs) }
```
antimora, line 135: I think it would be more efficient if we have one tensor op and rely on compiler to negate `#lhs`. We can rewrite as follows: `#rhs.add_scalar(-(#lhs)). So generated code might look like this: `#rhs.add_scalar(-(- 42)). And Rust compiler will precompute number literal correctly.

antimora (changes requested):
Thank you so much for fixing the bug. 

It looks good overall. I have minor suggestion to imporove.

laggui (commented):
Other than the already requested change, LGTM!

Thanks for fixing :)

crates/burn-import/src/burn/node/binary.rs
johnhuichen: Thanks or review @antimora!

Actually since this is lhs (scalar) - rhs (tensor), looks like -#rhs.sub_scalar(#lhs) produced the right result. I will make an update.

crates/burn-import/src/burn/node/binary.rs
```diff
@@ -131,6 +131,7 @@ impl BinaryNode {
             (Type::Tensor(_), Type::Tensor(_)) => move |lhs, rhs| quote! { #lhs.sub(#rhs) },
             (Type::Tensor(_), Type::Scalar(_)) => move |lhs, rhs| quote! { #lhs.sub_scalar(#rhs) },
             (Type::Scalar(_), Type::Scalar(_)) => move |lhs, rhs| quote! { #lhs - #rhs },
+            (Type::Scalar(_), Type::Tensor(_)) => move |lhs, rhs| quote! { -#rhs.sub_scalar(#lhs) },
```
antimora, line 134: Thank you! 

FYI @nathanielsimard , @laggui , @louisfd

Another reason to support native Scalar type in Burn. See our earlier design discussion: https://github.com/tracel-ai/burn/issues/1689#issuecomment-2078212769

`{ -#rhs.sub_scalar(#lhs) },` solution to `scalar - tensor` will result in two operations instead of one.

antimora (approved):
Thank you for your fix! LGTM

antimora:
I'll merge to the main once the CI passes again.

PR #1965, opened 2024-07-04
Refactor/jit/unary

nathanielsimard:
Refactor unary operations using Cube. Added check bounds, which is necessary for the cuda jit runtime.


crates/burn-cube/src/frontend/operation/clamp.rs
```diff
@@ -0,0 +1,38 @@
...
+    prelude::{CubeContext, CubePrimitive, UInt, BF16, F16, F32, F64, I32, I64},
+    unexpanded,
+};
+
+use super::unary_expand;
+
+pub trait Clamp: CubePrimitive + Sized {
+    /// Clamp the input value between the max and min values provided.
+    #[allow(unused_variables)]
+    fn clamp(input: Self, max_value: Self, min_value: Self) -> Self {
```
louisfd, line 12: Should be min_value before max_value

crates/burn-cube/src/lib.rs
```diff
@@ -54,6 +54,28 @@ pub fn calculate_cube_count_elemwise(num_elems: usize, cube_dim: usize) -> CubeC
     CubeCount::new(cube_count_x as u32, cube_count_y as u32, 1)
 }
 
+pub fn tensor_vectorization_factor(factors: &[u8], shape: &[usize], strides: &[usize]) -> u8 {
```
louisfd, line 57: It should be possible to vectorize over the dimension whose stride is 1, whatever dimension it is.

louisfd (changes requested):
Looking good in general, 
I think you should reverse clamp args, and we should discuss vectorization on other dims than last

PR #1975, opened 2024-07-05
Feat: Dynamic cube count dispatch

ArthurBrussee:
### Checklist

### Changes

I have some custom kernels where only the GPU knows how many valid items were processed. Further dispatches then only need to process these items. WebGPU and co habe a mechanism for this - execute_indirect. This PR makes CubeCount an enum where either a static amount of work is dispatched, or a dynamic amount based on a buffer.

In CUDA land, the more common approach is to have kernels launch other kernels, which is a more general method of achieving this effect. I'm not sure if CudaRC has a mechanism to do this from their API. Creating a dummy kernel that launches the real kernel with the right settings could work. For now, it just reads the data from the buffer which isn't optimall but at least correct.

The code does get a bit annoying as CubeCount is now generic over the server. I wisth rust had the option to make CubeCount::Fixed a full valid return type! To prevent these generics further leaking into everything I've split the CubeCount and CubeTask. 

Things would also be a bit easier if the ComputeServer knew about CubeCount, but I don't want to make cube a dependency of burn-compute. One option is to move the CubeCount to burn-compute and just name it DispatchCount. For now I've done the same associated type trick as is in place for CubeTask.

### Testing

Tests, ran my own project using dynamic dispatch.


crates/burn-cuda/src/compute/server.rs
```diff
@@ -101,12 +108,29 @@ impl<MM: MemoryManagement<CudaStorage>> ComputeServer for CudaServer<MM> {
...
+
+        let count = match count {
+            CubeCount::Static(x, y, z) => (x, y, z),
+            // TODO: CUDA doesn't have an exact equivalen of dynamic dispatch. Instead, kernels are free to launch other kernels.
+            // One option is to create a dummy kernel with 1 thread that launches the real kernel with the dynamic dispatch settings.
+            // For now, just read the dispatch settings from the buffer.
+            CubeCount::Dynamic(binding) => {
+                let data = self.read_sync(binding);
+                let data: Vec<u32> = bytemuck::cast_slice(&data).to_vec();
+                (data[0], data[1], data[2])
```
nathanielsimard, line 129: Not sure we are forced to call `to_vec`

```rust
let data: &[u32] = bytemuck::cast_slice(&data);
assert!(data.len() == 3, "Dynamic cube count should contain 3 values");
(data[0], data[1], data[2])
```

crates/burn-compute/src/server.rs
```diff
@@ -17,6 +17,8 @@ where
 {
     /// The kernel type defines the computation algorithms.
     type Kernel: Send;
+    /// The type for the number of workgroups to dispatch.
+    type DispatchCount: Send;
```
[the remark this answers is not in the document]
nathanielsimard: Or even more general, dispatch options, which can be none in some tests.

nathanielsimard (changes requested):
Thanks for the PR. I think there are only minor things to fix before we can merge. 

Sometime we use work groups and sometimes we use cube, we should be consistent and use cube most of the time (work group is WebGPU, block is CUDA)

crates/burn-cuda/src/compute/server.rs
ArthurBrussee: Oh right nice spot!

ArthurBrussee:
Tried to fix most of the workgroup -> cube conversions :) The CUDA compiler uses it quite a lot too, but leaving that for now.

nathanielsimard (approved):

PR #1966, opened 2024-07-04
Add models and examples reference

laggui:
Figured I'd tackle this before we eventually get to the release date and we still haven't done it.

### Checklist

- [x] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

Closes #1808 
Closes #1809

### Changes

- Added reference to the models repo in the book
- Added the list of examples in the book and README

When rendered, the examples section looks like this:

image

PR #1977, opened 2024-07-05
Fix warnings when using `record-backward-compat`

laggui:
Fix some small omissions in the tensor data refactor #1916 with the `record-backward-compat` feature flag

- Unused import
- Allow deprecated use of `DataSerialize`
- Leftover generic on dimension


laggui:
Looks like a typo was introduced in #1966 (CI didn't run cuz it only affected markdown files)..

Seems to have been fixed in other PRs so I'll leave it as is.

nathanielsimard (approved):

PR #1945, opened 2024-06-29
Implement 3D and transposed 3D convolutions.

booti386:
Hi!
Please review this PR carefully, I started working both on neural networks and this crate only 4 days ago.
I hope there are no major mistakes.

### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

Fixes #1278: Conv3d feature request

### Changes

Implement 3D and 3D transposed convolutions for all the supported backends.

PR #1978, opened 2024-07-05
Implement 3D and transposed 3D convolutions

nathanielsimard:
Since the Conv3d PR broke the CI for an unknown reason (yet) we reverted it to allow other work to continue.

This PR tries to add the commit back (#1945)  to main without breaking the CI.


syl20bnr:
The push history has disappeared from the PR thread but here are the results when tweaking the `tasks_max` field of `RuntimeOptions` for wgpu with 7 test runs:

| Max Tasks | ✅ | ❌ |
|:---------:|:--:|:--:|
| 1         | 7  | 0  |
| 32        | 3  | 4 |

I'll push a rebase with main and the modification of the default max tasks to 1, if the run passes I'll merge to main.

PR #1994, opened 2024-07-08
Cube: Matmul tiling

louisfd:
Matmul tiling again. Has undergone major refactoring and performance enhancing since my last draft.


crates/burn-jit/src/tensor/layout.rs
```diff
@@ -0,0 +1,121 @@
+#[derive(PartialEq, Eq, Debug)]
+pub(crate) enum MemoryLayout {
+    /// Memory is wholly contiguous, with row major layout
+    Contiguous,
+    /// Permutations happened, but may not impact some kernels
+    MildlyPermuted {
+        /// Last two dims are inverted
+        transposed: bool,
+        /// Some permutations exist in batch dimensions
+        batch_swap: bool,
```
nathanielsimard, line 10: We assume the the batch dimensions are all except the two lasts, which is not always the case. We should put a comment here

crates/burn-wgpu/src/compute/server.rs
```diff
@@ -104,6 +104,7 @@ where
         }
 
         let compile = kernel.compile();
+        println!("{}", compile.source);
```
nathanielsimard, line 107: Should remove before merge.

PR #2000, opened 2024-07-09
Module weight quantization

laggui:
### Checklist

### Changes

Static per-tensor module quantization support added with `quantize_weights` method
- Added `Quantizer`, `MinMaxCalibration` and `QuantizationScheme` to define the quantization
- Changed the tensor record implementation to keep `DType::QFloat` tensors as is (no conversion)
- Added methods to `QTensorOps` to support quantized float tensors
- Added quantization docs

### Testing

Added unit tests for `MinMaxCalibration`


nathanielsimard (approved):
LGTM very clean

PR #2003, opened 2024-07-10
Enable optimized handling of bytes

laggui:
### Checklist

### Changes

Added `serde_bytes` attribute to enable efficient de/serialization of tensor data.

### Testing

In practice, this reduces the `NamedMpkFileRecorder` file size and improves de/serialization speed 2-4x (tested mpk and bin).


nathanielsimard (approved):
🚀

PR #2007, opened 2024-07-11
Implement ONNX Pad Operator

johnhuichen:
### Checklist

### Related Issues/PRs

### Changes

Implement ONNX pad by calling existing tensor.pad

<strong>Input tensor</strong>: currently supporting only Float32 but will extend the support to other floats, int and boolean values

<strong>Pads</strong>: currently supporting only positive pads 

<strong>Constant value</strong>: type restricted to the input tensor element type, currently supporting only Float32

<strong>mode</strong>: currently supporting only "constant" 

TODO:

Support int, boolean tensors

Currently passing a negative constant value may cause an issue. Need to investigate further

### Testing

1. Test ONNX model produces expected outputs for f32 input tensor and positive pads before building ONNX model
1. Test pad codegen creates expected rust code
1. Test generated pad rust code produces expected outputs for f32 input tensor and positive pads


crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,49 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+        assert_eq!(pads.len(), input_dim * 2);
+
+        if input_dim == 1 {
+            vec![pads[0], pads[1], 0, 0]
+        } else {
+            let left = pads[input_dim - 1];
+            let top = pads[input_dim - 2];
+            let right = pads[pads.len() - 1];
+            let bottom = pads[pads.len() - 2];
+            vec![left, right, top, bottom]
```
antimora, line 778: This assumes the user wants to pad last 2 dimensions which works for the Burn's pad but it's silent. We should break if ONNX configuration does not fit Burn's API. Obviously ONNX version is more generic, which requires Burn's code be updated as well (not required for this PR).

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,49 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+        }
+    }
+
+    let pads = get_pads(node);
+    // TODO: support int, boolean
+    let constant_value: f32 = match &node.inputs[2].value {
+        Some(Data::Float32s(shape)) => shape.first().unwrap().to_owned(),
+        _ => 0.,
+    };
```
antimora, line 787: We need to panic and not be silent. Also, there could be only one input (data) and attributes for others see [OpSet 1](https://onnx.ai/onnx/operators/onnx__Pad.html#pad-1)

antimora (changes requested):
Thank you for your addition!

Overall looks good. However, we need to tighten the pad configuration. We need to panic aggressively if the user is passing more configuration that our current APIs support. It will be an opportunity to enhance Burn's pad function.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,49 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+                        // TODO: support negative pads
+                        panic!("Negative pad is not supported");
+                    }
+                    x as usize
+                })
+                .collect(),
+            _ => panic!("Pads data type must be int64"),
+        };
+
+        assert_eq!(pads.len(), input_dim * 2);
```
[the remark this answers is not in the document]
laggui: Yeah the `tensor.pad` works for 3d. But I think this assertion is coming straight from the ONNX spec:

> `pads` should be a 1D tensor of shape [2 * input_rank]

So this check is fine and generic based on the spec, but for Burn conversion we will need to make sure the configured padding can be specified as `(left, right, top, bottom)`.

laggui (commented):
Thanks for adding this implementation! I think overall it looks good.

I agree with the comments made by @antimora, we need to be more strict/clear about the unsupported configurations.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,49 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+
+        if input_dim == 1 {
+            vec![pads[0], pads[1], 0, 0]
```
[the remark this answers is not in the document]
laggui: Pretty sure it won't given that it operates on the last two dimensions.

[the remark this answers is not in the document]
johnhuichen: Burn's padding doesn't work for 1D because of
crates/burn-tensor/src/tensor/api/numeric.rs
```
        padded_dims[D - 2] += top + bottom;
```

I will add a panic check

johnhuichen:
@antimora @laggui 

I have made onnx conversion stricter by forcing three inputs. Also added a warning about pad applying to only the last two dimensions.  

can you guys take another look?

crates/burn-import/src/onnx/op_configuration.rs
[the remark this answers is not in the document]
johnhuichen: the ONNX specification requires the relation between pads length and input tensor rank as laggui said. 

The code below converts ONNX pads to the shape required for Burn's pad function

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,64 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+
+        if input_dim > 2 {
+            log::warn!("Pad: padding will only be applied to the last two dimensions");
+        }
+
+        let left = pads[input_dim - 1];
+        let top = pads[input_dim - 2];
+        let right = pads[pads.len() - 1];
+        let bottom = pads[pads.len() - 2];
+        vec![left, right, top, bottom]
```
laggui, line 790: I think the correct conversion when parsing `pads` should check the pad order `[x1_begin, x2_begin, ..., x1_end, x2_end, ...]`, and we should make sure that all dimensions outside of the last two are not padded otherwise the conversion is not valid (and thus, panic instead of warn).

Following the given [spec example](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Pad), a 4d input with padded on the last two dimensions should have `pads = [0, 0, x2_begin, x3_begin, 0, 0, x2_end, x3_end]` (i.e., `x1_begin`, `x1_end`, `x2_begin`, and `x2_end` are all zero)

laggui (commented):
That's better! I think there's still just one issue with the current way the padding is parsed.. we should not simply warn when the operation might be incorrect.

We should make sure that the provided values will work for the Burn conversion, otherwise panic.

crates/burn-import/src/onnx/op_configuration.rs
johnhuichen: ok that's a good idea

johnhuichen: I have pushed the update

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -748,43 +748,66 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+        // TODO: Burn's pad should support 1D tensor
+        if input_dim < 2 {
+            panic!("Pad: input tensor should be rank 2 or higher");
+        }
+
+        let left_index = input_dim - 1;
+        let top_index = input_dim - 2;
+        let right_index = pads.len() - 1;
+        let bottom_index = pads.len() - 2;
+        let index_list = [left_index, top_index, right_index, bottom_index];
```
antimora, line 787: This seems to accept pads with pads.len() > 4 but Burn API only accepts 4. Right?

johnhuichen: If pads.len() > 4, this code checks if the user wants to pad any dimension than the last two:

yes -> panic
no -> the pad operation is doable, and it calculates l, t, r, b

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,72 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+
+        for index in 0..pads.len() {
```
laggui, line 788: Not sure why the CI didn't trigger at the PR level, but in the review mode I can see a suggestion:

```
[clippy] reported by reviewdog 🐶
error: the loop variable `index` is used to index `pads`
   --> crates/burn-import/src/onnx/op_configuration.rs:788:22
    |
788 |         for index in 0..pads.len() {
    |                      ^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_range_loop
    = note: `-D clippy::needless-range-loop` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_range_loop)]`
help: consider using an iterator and enumerate()
    |
788 |         for (index, <item>) in pads.iter().enumerate() {
    |             ~~~~~~~~~~~~~~~    ~~~~~~~~~~~~~~~~~~~~~~~
```

Can you make the requested changes please? Otherwise CI will fail when we merge.

crates/burn-import/src/onnx/op_configuration.rs
laggui: The way the padding is specified for the ONNX Pad op is a bit different since it supports padding more than just the last two dimensions. So we have to parse the provided `pads` to make sure it is valid, and then return the Burn valid padding with only 4 values for (left, right, top, bottom).

I think the check is valid. For a 4d input where we should only accept padding the last two dimensions, it means that `pads` should be `[0, 0, x2_begin, x3_begin, 0, 0, x2_end, x3_end]`. For the last two dimensions expected to be `[H, W]`, this means that `x2_begin = top`, `x3_begin = left`, `x2_end = bottom` and `x3_end = right`.

crates/burn-import/src/onnx/op_configuration.rs
johnhuichen: Of course. Just pushed my changes.

laggui (approved):

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,72 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
     )
 }
 
+/// Create a PadConfig from the attributes of the node
+pub fn pad_config(node: &Node) -> PadConfig {
+    fn get_pads(node: &Node) -> Vec<usize> {
+        if node.inputs.len() != 3 {
+            panic!("Pad: must provide three inputs")
+        }
```
antimora, line 753: The says at least 2 required (so check should be  node.inputs.len() > 1) 

3rd and 4th inputs are optional.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -745,6 +745,72 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
...
+        let right = pads[right_index];
+        let bottom = pads[bottom_index];
+        vec![left, right, top, bottom]
+    }
+    fn get_constant_value(node: &Node) -> f32 {
+        // TODO: support int, boolean
+        match &node.inputs[2].value {
+            Some(Data::Float32s(shape)) => shape.first().unwrap().to_owned(),
+            _ => panic!("Pad: should provide a constant value input to pad with, for example 0.0"),
+        }
```
antimora, line 805: 3rd input is optional and should be default to 0 if absent.

antimora (changes requested):
One minor fix is required. Please see my inlined comment. We should be able to merge afterwards.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -748,8 +748,8 @@ pub fn layer_norm_config(node: &Node) -> (LayerNormConfig, bool) {
 /// Create a PadConfig from the attributes of the node
 pub fn pad_config(node: &Node) -> PadConfig {
     fn get_pads(node: &Node) -> Vec<usize> {
-        if node.inputs.len() != 3 {
-            panic!("Pad: must provide three inputs")
+        if node.inputs.len() < 2 {
+            panic!("Pad: must provide two inputs")
```
antimora, line 752: I would add "at least".

"must provide two inputs" => "must provide at least two inputs"

antimora (approved):
Looks good. Thanks for fixing. Minor wording update needed but I am approving in advance.

Issue #1915, opened 2024-06-21
Trouble importing FaceONNX detector model: `Only tensor indices is valid`


SimonBrandner:
**Describe the bug**

Error during importing the detector model: 

```
ERROR burn_import::logger: PANIC => panicked at /home/simon/Data1/GIT/Rust/burn/crates/burn-import/src/onnx/dim_inference.rs:820:14:
Only tensor indices is valid    
```

**To Reproduce**
1. Download detector model from https://drive.google.com/file/d/1tB7Y5l5Jf2270IisgSZ3rbCQs0-pkNIQ/view?usp=drive_link (also attached down below)
2. Move the file to `src/model/detector.onnx`
3. Add this to `build.rs`

```rs
	ModelGen::new()
		.input("src/model/detector.onnx")
		.out_dir("model/")
		.run_from_script();

```

4. Build and see error

**What I've figured out so far**

- There is something going on when updating the gather outputs: https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/onnx/dim_inference.rs#L812

Attachment:
[face_detector_640.onnx.zip](https://github.com/user-attachments/files/16045435/face_detector_640.onnx.zip)

laggui:
As you correctly pointed out the issue is with the gather node.

We check for the inputs to make sure they are in accordance with the spec, but when checking for the indices we expect a tensor but your model uses a scalar (rank 0 tensor) - which is still valid.

I'll open a PR to fix that.

laggui:
Also I just realized your model was generated with an older opset (v9) and we might not support all of the ops there. Right now the ops we have should support the latest ops at opset 13. Some also have support for older opset but not all (e.g., `Slice` which I see in your model is opset v1 and we don't parse the attributes).

The ONNX spec is large and our bandwidth for that is a bit limited.

antimora:
We should probably fix this to support older opset. 

As a work around, can you convert to a new opset? There are some tools to upgrade the onnx file.

SimonBrandner:
I tried [this](https://github.com/openppl-public/ppl.nn/blob/master/docs/en/onnx-model-opset-convert-guide.md) and it did not help... That said from the comments my understanding is that the `Only tensor indices is valid` issue isn't due to an old opset, is it?

SimonBrandner:
This dirty fix seems to be working:

```diff
[...]

Issue #1882, opened 2024-06-12
burn-import initializer tensor's not added to scope


mosure:
**Describe the bug**
nodes with input tensors of category `initializer` fail to import due to initializer tensor names not being in scope

**To Reproduce**
add a minimal onnx graph /w initializer tensor, e.g.

```python
class Model(nn.Module):
    def __init__(self):
        super(Model, self).__init__()

    def forward(self, x):
        x = F.interpolate(x, size=(2, 2), mode='bilinear', align_corners=True)
        return x
```

![image](https://github.com/tracel-ai/burn/assets/35904762/4a6cc603-99e9-4bfc-9033-68259210f02c)
![image](https://github.com/tracel-ai/burn/assets/35904762/77199871-95e9-48ff-b2b6-0f3848de71f1)

**Expected behavior**
burn-import generated model contains tensor initializers, e.g. `_Concat_output_0` is defined in the scope

**Actual behavior**
burn import fails with `_Concat_output_0` is not in scope, panic at this line: https://github.com/tracel-ai/burn/blob/71bd5efbfaf220b5058dbf4f963326e0b0d38527/crates/burn-import/src/burn/scope.rs#L53

**Desktop (please complete the following information):**
 - OS: [Windows]

**Additional context**
this occurs in `resize`, `gather`, `mul_scalar`, and `sub_scalar` node imports as well (with initializer tensors)

antimora:
@skewballfox, do you think this is related to #1857?

skewballfox:
was this encountered with the version of burn last published in crates.io or on main? 

also, I'm a bit confused. is `/Concat_output_0` both a node output and an initializer?

mosure:
this is encountered on main.

here is a more minimal reproducible example which clarifies the `/Concat_output_0` initializer/output ambiguity:

![image](https://github.com/tracel-ai/burn/assets/35904762/194004c5-1587-40a3-bbe5-9c60b3c0531b)

```python
#!/usr/bin/env python3

# used to generate model: onnx-tests/tests/resize/resize.onnx

import onnx
from onnx import helper, TensorProto, numpy_helper
import numpy as np

def main() -> None:
    input_tensor = helper.make_tensor_value_info("input_tensor", TensorProto.FLOAT, [1, 1, 4, 4])

    # Define the sizes tensor as an initializer
    sizes = np.array([1, 1, 2, 2], dtype=np.int64)
[...]

PR #2008, opened 2024-07-11
Constant tensor support in burn-import

skewballfox:
### Checklist

### Related Issues/PRs

- #1915 
- #1882
- #1560 
- #1593

### Changes

Fixes 2 issues:
- In older models, Initializers might be used in place of constants, so those inputs need to be renamed to avoid generating invalid code. This either renames all initializers that are not inputs (commented out) or ~~checks for initializers in nodes known to be a problem(currently `Add` and `Mul`)~~. EDIT: have to go with option 1 because initializers might be reused, which means tracking name changes 
- So far, all constants have been scalars or vectors, which aren't checked during scoping. This causes scope checking to panic during the backward pass since those inputs were never registered as a graph input or node output. Right now, the fix is to simply register those inputs as a constant during the pass over inputs

this makes a slight non-fix change to scopes, changing variables value from `Vec<TensorVariable>` to `TensorVariable` given there is only one variable pushed to each entry. It also registers outputs and checks inputs in the same iteration, which works given the graph is guaranteed to be a DAG.

### Unresolved questions

- Right now the fix for the scope panic is to register any unregistered tensors during the backward pass over inputs. This might be buggy. For example what if the model file is simply invalid and references a variable that simply doesn't exist (~~which would make it through onnx-ir~~)?  Should we try to detect and notify the user their model is faulty? EDIT: it would fail during TensorType conversion at the latest)
    - the alternative is to track and pass in constants/initializers separately from onnx-ir. This would be a bit more work but might be the option we want to go with if it turns out we can't track constants on the fly during scoping.
- Lastly, how should constants be generated? As module level global statics? Made an attribute of the model struct and declared during `Model::new()`? Should we rework constant node type to support Tensors or treat them as a separate beast altogether?

### Testing

we need to add a model to onnx test, but generating one for issue one might be hard. Could we just add a legacy folder to throw in older models we can't generate directly?


antimora:
I think we should make "an attribute of the model struct and declared during Model::new()". Record loading happens at the module level and we plan to support multi/sub graphs as separate standalone models.

antimora:
Regarding testing. Yeah I wish we could store legacy ONNX files but there are a few issues: 1) License and size (which might slow down testing). 

However, we can probably use a quick python script that deletes other nodes. I think this is supported by onnx package. So we could instruct to remove everything else except a portion and randomize weights (so no copying).

antimora:
I have reviewed the PR and believe a simpler and more scalable solution would be to extract inputs with initializers for node types such as `Sub`, `Add`, etc., as `Constants`. The current solution in the PR introduces a new concept of constants, which already exists in `burn-import` under the `Constant` node type (a working example can be found in the Sub ONNX test: [link](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/onnx-tests/tests/sub/sub.py)).

The solution should follow a similar approach to this script that I use to extract `Sub`/`Add` initializers to constants: [link](https://github.com/antimora/face-onnx/blob/main/onnx_initializers_to_consts.py). It's essentially the reverse process of constant lifting that we use for `Conv` and other nodes.

Having said that I think it'd be easier to start with a new PR than making changes.

PR #2009, opened 2024-07-11
Refactor/cube/expand & fix double imports

nathanielsimard:


louisfd (approved):
LGTM
I like the ::launch and __expand. Much better api and way to hide expand stuff

PR #2025, opened 2024-07-16
Refactor tensor quantization for q_* ops

laggui:
### Checklist

### Changes

Refactored the quantization methods to support quantization params on the backend device (no sync). Laying the ground work before adding all float ops as `q_*` ops in `QTensorOps`.

- Added `QTensorPrimitive` trait for backends to implement
  - The scheme should be linked to the primitive while the qparams can stay on the backend device and only synced when required
- Added `QuantElement` to backends that implement quantization (allows for selecting from valid quantization types, e.g. u8 or i8)
- Added `QuantizationParameters` (+ `QuantizationParametersPrimitive`)
- `tensor.quantize()` method now takes a `QuantizationScheme` and `QuantizationParameters`
  - In the future, dynamic quantization can be implemented with the addition of `tensor.quantize_dynamic()` which will only take a `QuantizationScheme`
- `Calibration` techniques should only compute the range of values, and `QuantizationScheme` can `compute_q_params`

### Testing

Added unit tests in burn-tch and burn-ndarray for quantized tensor primitive.


nathanielsimard (approved):
LGTM

PR #2032, opened 2024-07-17
Only poll device when reading data instead of blocking

ArthurBrussee:
### Changes
According to [wgpu docs](https://docs.rs/wgpu/latest/wgpu/struct.BufferSlice.html#method.map_async), the device only needs to be polled for the map_async to complete at some point. Waiting for the channel does the actual waiting (async or blocking).

This is not very different, but, Maintain::Wait also prevents other threads from submitting commands, leading to stalls in multithreaded scenarios. Here's a before and after of a multithreaded app:

Screenshot 2024-07-17 at 12 04 06
Screenshot 2024-07-17 at 12 04 51

(Send batch is waiting on a read).

### DIFF
--- a/crates/burn-wgpu/src/compute/server.rs
+++ b/crates/burn-wgpu/src/compute/server.rs
@@ -179,7 +179,7 @@ where
                     .expect("Unable to send buffer slice result to async channel.")
             });
 
-            device.poll(wgpu::Maintain::Wait);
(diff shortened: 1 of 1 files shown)


ArthurBrussee:
Hm, converting this to draft for a sec as it seems tests are timing out waiting for data :/ Maybe poll isn't enough after all, will investigate.

PR #2035, opened 2024-07-17
Bug/Remove Squeeze Panic for Multiple Dimensions

### LINKED ISSUE #2033: Onnx Squeeze to accept a list of axes
### Feature description

https://onnx.ai/onnx/operators/onnx__Squeeze.html
The Squeeze operation to accept a list of axes to squeeze. 

### Feature motivation

This is an option in ONNX which currently fails when converting:

```
  DEBUG onnx_ir::from_onnx: checking node squeeze1 for constants    
  DEBUG onnx_ir::from_onnx: checking input Argument { name: "const_axes__420", ty: Tensor(TensorType { elem_type: Int64, dim: 1, shape: Some([2]) }), value: Some(Int64s([2, 3])), passed: false } for const    
  ERROR burn_import::logger: PANIC => panicked at crates/onnx-ir/src/dim_inference.rs:463:9:
  Squeeze must specify only 1 axis, found 2
```

agelas:
### Checklist

### Related Issues/PRs

Fixes #2033 

### Changes

Removes panic if user passes more than 1 axes to a Squeeze node in their ONNX graph, should've been part of #1779.


laggui (commented):
Thanks for fixing the issue!

If providing multiple axes is supported we should probably add a test to cover the use case 🤔 What do you think?

agelas:
I think that would definitely be helpful for catching things like this. I know `burn-import` got split up a bit- is there a place now for handling a test like that? Right now, the only test for `squeeze_dims` is in `crates/burn-tensor/src/tests/ops/squeeze.rs`, so only on the burn side but not ONNX conversion.

laggui:
On the ONNX side, we generate the ONNX files to test with this: https://github.com/tracel-ai/burn/blob/main/crates/burn-import/onnx-tests/tests/squeeze/squeeze.py

Right now it only covers a single axis, but we could provide an input with multiple axes to squeeze. For example `[3, 4, 1, 5, 1]` and `torch.squeeze(x, (2, 4))`. So the included test would cover both use cases.

agelas:
@laggui Hmm, so that behavior doesn't look supported on the torch->ONNX conversion side. When I use your example with `torch.squeeze(x, (2, 4))`, I get this error:

```bash
  File "/burn/crates/burn-import/onnx-tests/tests/squeeze/squeeze.py", line 46, in <module>
    main()
  File "/burn/crates/burn-import/onnx-tests/tests/squeeze/squeeze.py", line 32, in main
    torch.onnx.export(model, test_input, "squeeze_opset16.onnx", verbose=False, opset_version=16)
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/utils.py", line 516, in export
    _export(
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/utils.py", line 1596, in _export
    graph, params_dict, torch_out = _model_to_graph(
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/utils.py", line 1139, in _model_to_graph
    graph = _optimize_graph(
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/utils.py", line 677, in _optimize_graph
    graph = _C._jit_pass_onnx(graph, operator_export_type)
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/utils.py", line 1940, in _run_symbolic_function
    return symbolic_fn(graph_context, *inputs, **attrs)
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/symbolic_opset11.py", line 930, in squeeze
    dim = symbolic_helper._get_const(dim, "i", "dim")
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/symbolic_helper.py", line 178, in _get_const
    return _parse_arg(value, desc)
  File "/burn/myenv/lib/python3.10/site-packages/torch/onnx/symbolic_helper.py", line 83, in _parse_arg
    return int(node_val)
ValueError: only one element tensors can be converted to Python scalars
```
I did change the file a bit though with no arguments passed in, which squeezes all singleton dimensions. I know that's not quite what you asked for, but at least we're squeezing multiple dimensions, and it might be the best we can do since it seems torch itself has difficulty with the conversion. I'm using torch 2.1.1 too.

laggui:
Oh, well I guess you're right then. If we want to add a test we'll have to construct the graph manually. I could provide a script for that later when I get some time.

agelas:
Thanks for the script. I added that to `squeeze.py` and got this, so it looks all good:

image

Since it's not using torch at all, do we need to add it to `onnx_tests.rs`?

agelas:
Ok, fingers crossed everything should be working now. Testing the model you constructed also lead me to finding another bug with how output dimensions were calculated, so adding that test was a huge plus.

laggui (approved):
LGTM! Thanks for adding the multiple dim support and test cases 🙏

antimora (approved):
LGTM

PR #2040, opened 2024-07-19
Upgrade to wgpu 22.0.0

ArthurBrussee:
### Changes

Upgrade wgpu to 22.0.0. There's two breaking things:

- Have to specifiy a MemoryHint, choosing between performance and lower memory usage. GIven wgpu has it's own block allocator now, MemoryUsage seems better. I did a quick test on the train sample and see no clear performance difference between either.

- Need to specify a pipeline cache. This allows to cache compiled shader outputs. As per the comment in the wgpu code this is usually done automatically though not on Android seemingly. Could be nice someday.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,11 +183,11 @@ checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
 name = "ash"
-version = "0.37.3+1.3.251"
+version = "0.38.0+1.3.281"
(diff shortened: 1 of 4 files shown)


nathanielsimard:
@ArthurBrussee Most of the codebase now lives on the [CubeCL repo](https://github.com/tracel-ai/cubecl), can you made the PR there instead? I'm actively working on the migration, should land very soon on the burn main branch.

ArthurBrussee:
Yeah makes sense, will create a PR there once the migration lands!

PR #2051, opened 2024-07-22
Fix checks_channels_div_groups condition and ONNX conv import with groups

### LINKED ISSUE #2043: Strange model initialization errors
There are a couple of strange errors that I am see when I load Face ONNX and what to just instantiate a model. 

These errors can be reproduced in this repo: https://github.com/antimora/face-onnx

By running:

```
cargo run --bin with_weights
```

Which will result in this error:

```
Uniform::new called with `low` non-finite.
```

and 

```
cargo run --bin without_weights
```

Which will result in this error:
```
cast_slice>PodCastError(TargetAlignmentGreaterAndInputNotAligned)
```

laggui:
### Checklist

### Related Issues/PRs

Closes #2043 

### Changes

The check condition was not actually checking for both values to be divisible, so I fixed the condition.

### Testing

Added a unit test.

### DIFF
--- a/crates/burn-core/src/nn/conv/checks.rs
+++ b/crates/burn-core/src/nn/conv/checks.rs
@@ -2,7 +2,7 @@ pub(crate) fn checks_channels_div_groups(channels_in: usize, channels_out: usize
     let channels_in_div_by_group = channels_in % groups == 0;
     let channels_out_div_by_group = channels_out % groups == 0;
 
-    if !channels_in_div_by_group && !channels_out_div_by_group {
+    if !channels_in_div_by_group || !channels_out_div_by_group {
         panic!(
             "Both channels must be divisible by the number of groups. Got \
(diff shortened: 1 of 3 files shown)


laggui:
Huh, conv transpose tests are failing in burn-import. But it looks like the values are incorrectly converted.

For example, the `conv_transpose2d` test fails with:
```
Both channels must be divisible by the number of groups. Got channels_in=3, channels_out=4, groups=2
```

but the node actually defines `channels_in=4`, `channels_out=6`. I think there might be a bug there.

laggui:
Turns out the groups weren't taken into account when importing the ONNX models 🙂everything should be fixed now.

laggui:
Still seems to be an issue when trying to import the model from the linked issue. Trying to fix it right now.

antimora (approved):
LGTM!

Thanks for fixing it!

PR #2055, opened 2024-07-23
Extend affine quantization [min, max] range to ensure zero-point

laggui:
### Checklist

### Changes

While adding unit q_* ops for the current float ops I realized affine quantization must ensure that the input range includes zero to enforce the zero-point representation. By definition, the zero-point value corresponds to the value 0 in the float32 realm.

### Testing

Added zero-point unit test and adjusted other unit tests values to reflect these changes.


nathanielsimard (approved):
LGTM

PR #2063, opened 2024-07-26
Refactor xtask to use tracel-xtask and refactor CI workflow

syl20bnr:
This PR refactor the xtask crate to usethe new crates `xtask-common` and `xtask-macros` from the [xtask repository](https://github.com/tracel-ai/xtask).

It also refactors the `test.yml` into the `ci.yml` files using composite actions to setup rust and OSes in the `actions` directory. The big matrix have been split into several jobs. Now each OS has its own jobs with smaller matrices.  Checks and documentation have their own jobs too.

It introduces a new convention which is that all the integration tests file names (the tests that are in a `tests` directory) must start with `test_` prefix.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+        shell: bash
+        run: cargo xtask --execution-environment no-std ci unit-tests
+
+  windows-std:
+    runs-on: windows-2022
+    env:
+      DISABLE_WGPU: '1'
+    strategy:
+      matrix:
+        rust: [stable, 1.79.0]
```
nathanielsimard, line 233: For window I would only test on stable.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+      # --------------------------------------------------------------------------------
+      - name: Integration Tests
+        shell: bash
+        run: cargo xtask ci integration-tests
+
+  macos-std:
+    runs-on: blaze/macos-14
+    strategy:
+      matrix:
+        rust: [stable, 1.79.0]
```
nathanielsimard, line 274: Same for mac

PR #2075, opened 2024-07-29
Refactor/jit cube/mask

louisfd:
Made mask kernels, along with bool_cast, with Cube. 
Also fixed cube count calculations needing a usize, which now needs the cube dim.


crates/burn-jit/src/kernel/cast/bool_cast.rs
```diff
@@ -1,12 +1,14 @@
...
+use crate::{tensor::JitTensor, JitElement, JitRuntime};
+use cubecl::{calculate_cube_count_elemwise, prelude::*, CubeDim};
+
+#[cube(launch)]
+fn bool_cast_kernel<T: Numeric>(input: &Tensor<UInt>, output: &mut Tensor<T>) {
+    if input[ABSOLUTE_POS] >= UInt::new(1) {
+        output[ABSOLUTE_POS] = T::from_int(1);
+    } else {
+        output[ABSOLUTE_POS] = T::from_int(0);
+    }
```
nathanielsimard, line 10: We need to add out-of-bound check and not write anything ot the output tensor when `input[ABSOLUTE_POS] >= input.len()`

PR #2107, opened 2024-08-05
Modify contributing md scripts to solve conflicts between doc and scripts

### LINKED ISSUE #2106: Conflicts bewteen Contributing.md  and scripts
I would like to suggest that some minor modifications on doc and implementations to conflicts between them. Hope it will improve new comers to join the project.

1.no_std is typo, but it should be no-std
The doc tell that there are only two environment of std and no-std, but it now includes four environments due to script.

```
./run-checks.sh no_std
    Finished `dev` profile [optimized] target(s) in 0.16s
     Running `target/debug/xtask run-checks no_std`
error: invalid value 'no_std' for '<ENV>'
  [possible values: all, std, no-std, typos, examples]

  tip: a similar value exists: 'no-std'

For more information, try '--help'.
```

2. “If no environment value has been passed, run both std and no_std checks”, this is incorrect because when I use the script without environment args, it causes an error as follow. So, some fixes are necessary. 

```
./run-checks.sh       
    Finished `dev` profile [optimized] target(s) in 0.15s
     Running `target/debug/xtask run-checks`
error: the following required arguments were not provided:
  <ENV>

Usage: xtask run-checks <ENV>
```

I will think and do the fix later as PR.

tiruka:
### Checklist

### Related Issues/PRs

Fixes #2106

### Changes

1.no_std is typo, but it should be no-std

no_std is typo, but it should be no-std

The doc tell that there are only two environment of std and no-std, but it now includes four environments due to script.

```rust
% ./run-checks.sh no_std
    Finished `dev` profile [optimized] target(s) in 0.16s
     Running `target/debug/xtask run-checks no_std`
error: invalid value 'no_std' for '<ENV>'
  [possible values: all, std, no-std, typos, examples]

  tip: a similar value exists: 'no-std'

For more information, try '--help'.
```

2.environment values

“If no `environment` value has been passed, run both `std` and `no_std` checks”, this is incorrect because when I use the script without environment args, it causes an error as follow. 

```rust
% ./run-checks.sh       
    Finished `dev` profile [optimized] target(s) in 0.15s
     Running `target/debug/xtask run-checks`
error: the following required arguments were not provided:
  <ENV>

Usage: xtask run-checks <ENV>
```

It is because the implementation and docs are different. As of now, I believe script comments are correct and modify the codes by modifying default value so as to keep `All` as it is and executes checks except examples. But if you prefer to others, I will follow it, so please let me know.

3.torch installation. 
I got in trouble when executing `./run-checks.sh all` because I have never installed torch on my pc. It is because `./run-checks.sh all` executes `crates/burn-tch` and it requires torch installation. Therefore, adding the link of `crates/burn-tch` will definitely help developers like me to solve the issue, so I added the link on the `CONTRIBUTING.md`.

### Testing

I run `./run-check.sh` and checked to `All` is executed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,7 @@ The `environment` argument can assume **ONLY** the following values:
 - `examples` to check the examples compile
 
 If no `environment` value has been passed, run all checks except examples.
+If you have an error related to `torch` installation, see [Burn Torch Backend Installation](./crates/burn-tch/README.md#Installation)
 
 ## Continuous Deployment
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,7 +94,7 @@ You may also want to enable debugging by creating a `.vscode/settings.json` file
 On Unix systems, run `run-checks.sh` using this command
 
 ```
-run-checks.sh environment
+./run-checks.sh environment
 ```
 
 On Windows systems, run `run-checks.ps1` using this command:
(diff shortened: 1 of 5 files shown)


antimora (approved):
LGTM!

Thanks for improving the script and documentation!

PR #2110, opened 2024-08-05
Fix aggregation results slice

### LINKED ISSUE #2109: [LibTorch] tensor slice error for aggregated tensors
Using `tensor.mean()`, `tensor.sum()` or `tensor.prod()` produces a 0-dim tch tensor internally (though Burn treats it as rank 1).

When you try to print this tensor, a narrow operation is performed and the program panics.

```rust
let t: Tensor<LibTorch, 1> = Tensor::from_floats([1.0], &Default::default());
let out = t.slice([(0..1)]); // fine
println!("{out}");

let t: Tensor<LibTorch, 1> = Tensor::from_floats([0.0, 1.0, 2.0], &Default::default());
let out = t.mean();
println!("mean");
let out = out.slice([(0..1)]); // panic!
```

As discovered by a discord user.

laggui:
### Checklist

### Related Issues/PRs

Fixes #2109 

### Changes

View aggregation results tch tensor primitive as 1d tensor.

### Testing

Added unit test for the edge case

### DIFF
--- a/crates/burn-tch/src/ops/base.rs
+++ b/crates/burn-tch/src/ops/base.rs
@@ -48,21 +48,16 @@ impl<E: tch::kind::Element + Copy + Default> TchOps<E> {
         tensor: TchTensor<E, D1>,
         ranges: [Range<usize>; D2],
     ) -> TchTensor<E, D1> {
-        let mut inner = tensor.tensor.shallow_clone();
-
-        // Handle 0-dim tch tensor slice for one element (treated as rank 1 w/ burn_tensor::Tensor)
-        if inner.size().is_empty() && D1 == 1 && ranges[0] == (Range { start: 0, end: 1 }) {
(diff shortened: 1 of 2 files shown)


laggui:
The alternative solution to this would be to unsqueeze the results of all aggregations (mean, sum, prod) so the tch primitive is not a 0-dim tensor. Maybe that's better actually 🤔

antimora (approved):
LGTM.

One concern I had about in-place operation on the view. I tested it and it seems to work fine on the view.

```python
>>> x
tensor([1, 2, 3])
>>> t = x.float().mean()
>>> t
tensor(2.)
>>> t.shape
torch.Size([])
>>> t = t.view(1)
>>> t
tensor([2.])
>>> t.shape
torch.Size([1])
>>> t.add_(1.0)
tensor([3.])
```

PR #2112, opened 2024-08-05
Add Hard sigmoid activation function

wingertge:
### Checklist

### Related Issues/PRs

Required to load the DbNet and PaddlePaddle OCR ONNX models used in RapidOCR.

### Changes

1. Add (configurable) hard sigmoid activation function
2. Add ONNX import support for the HardSigmoid node

### Testing

New unit tests and end-to-end ONNX tests.


crates/burn-import/onnx-tests/tests/hard_sigmoid/hard_sigmoid.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
+
+# used to generate model: relu.onnx
```
antimora, line 3: relu.onnx -> hard_sigmoid.onnx

antimora (approved):
LGTM from my end!

One part that I am not sure about is if we need hard_sigmoid_backward function as well?

@nathanielsimard can you please confirm?

wingertge:
> One part that I am not sure about is if we need hard_sigmoid_backward function as well?

The only other activation function with a scalar parameter is `leaky_relu` which also doesn't have a backwards function so I couldn't figure out how to implement it. If there is a way I'd be glad to add it.

laggui (commented):
Thanks for adding the implementation all the way through ONNX import!

For the backward, I think it's fine for now. We can let autodiff handle it since it's implemented as a combination of other ops.

PR #2114, opened 2024-08-05
Fix #2091 bug (in-place after expand)

### LINKED ISSUE #2091: Runtime error after expand op using Libtorch backend
```rust
use burn::{
    backend::{libtorch::LibTorchDevice, LibTorch},
    nn::EmbeddingConfig,
    tensor::{Int, Tensor},
};

fn main() {
    type Backend = LibTorch;
    let device = LibTorchDevice::Cpu;
    let positions: Tensor<Backend, 2, Int> = Tensor::arange(0..200, &device).expand([2, -1]);
    let embed = EmbeddingConfig::new(514, 256).init(&device);
    let res = embed.forward(positions + 2);
    println!("{res}");
}

```
QQ_1722642632521

This code works well on the `NdArray` backend.

antimora:
### Checklist

### Related Issues/PRs

Fixes #2091

### Changes

1. Introduce `should_clone_hint` flag to clone
2. Add a regression test

### Testing

Recreated the bug using a regression test

### DIFF
--- a/crates/burn-tch/src/ops/base.rs
+++ b/crates/burn-tch/src/ops/base.rs
@@ -493,8 +493,13 @@ impl<E: tch::kind::Element + Copy + Default> TchOps<E> {
         tensor: TchTensor<E, D>,
         shape: Shape<D2>,
     ) -> TchTensor<E, D2> {
-        let tensor = tensor.tensor.broadcast_to(shape.dims.map(|x| x as i64));
-        TchTensor::new(tensor)
+        let storage = tensor.storage.clone();
+        let broadcasted_tensor = tensor.tensor.broadcast_to(shape.dims.map(|x| x as i64));
+        let mut expanded_tensor = TchTensor::from_existing(broadcasted_tensor, storage);
+
+        // Hint that the tensor should be cloned for in-place operations
+        expanded_tensor.should_clone_hint = true;
+        expanded_tensor
     }
 
     pub fn sort<const D: usize>(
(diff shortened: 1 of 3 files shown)


laggui (approved):
Solution is simple enough, LGTM!

antimora:
@nathanielsimard Let me know if it's okay solution since you designed and implemented.

crates/burn-tch/src/ops/base.rs
```diff
@@ -493,8 +493,13 @@ impl<E: tch::kind::Element + Copy + Default> TchOps<E> {
...
     ) -> TchTensor<E, D2> {
-        let tensor = tensor.tensor.broadcast_to(shape.dims.map(|x| x as i64));
-        TchTensor::new(tensor)
+        let storage = tensor.storage.clone();
+        let broadcasted_tensor = tensor.tensor.broadcast_to(shape.dims.map(|x| x as i64));
+        let mut expanded_tensor = TchTensor::from_existing(broadcasted_tensor, storage);
+
+        // Hint that the tensor should be cloned for in-place operations
+        expanded_tensor.should_clone_hint = true;
+        expanded_tensor
```
nathanielsimard, line 502: This isn't the right fix. The storage should "know" that a tensor can't be reused. The issue likely arises because tch doesn't execute anything during the expand operation, leading to the buffer being smaller than the actual size of the tensor. The correct solution is to ensure that the buffer's size in storage matches the tensor's requirements and to assert that each element in the tensor has its own allocated memory space, without sharing.

nathanielsimard (changes requested):

crates/burn-tch/src/ops/base.rs
antimora: After we spoke offline, I agree with you on this. We need a more generic solution. 

I have removed `should_clone_hint` flag and replaced with can_mut tensor method that checks if dimensions were broadcasted. This will prevent other potential issues and perhaps issues if tch backend was extended.

nathanielsimard (approved):

PR #2117, opened 2024-08-06
Convert `reduce_dim_naive` kernel to use the `#[cube]` derive macro

cBournhonesque:
## Pull Request Template

Following a discussion with @nathanielsimard on discord, I wanted to try to convert the reduce kernel to use the cube derive macro.
(originally because I wanted to know how to set the `CubeCount` for a kernel).

The new macro is pretty ergonomic, it's actually crazy that it also works for `impl Trait`.
I have a couple remaining issues though:

### Checklist

- [x] Implement `CubeType` for a tuple of `CubeType`s: https://github.com/tracel-ai/cubecl/issues/48
- [ ] How do I get the max value/min value of a `Numeric`? Do I need to add a trait for it?
- [x] Need to add the computation of the `CubeCount`

### Testing

Unit tests


crates/burn-jit/src/kernel/reduce/naive/shader.rs
```diff
@@ -1,148 +1,42 @@
...
-        let item_output = EO::cube_elem().into();
+#[cube(launch)]
+pub(crate) fn naive_reduce_dim_compute_shader<RD: ReduceDimNaive<EI>, EI: Numeric, EO: Numeric>(
+    input: &Tensor<EI>,
+    output: &mut Tensor<EO>,
+    dim: UInt,
+) {
+    if ABSOLUTE_POS >= output.len() {
+        return;
+    }
```
nathanielsimard, line 15: If you are doing bound checks, you can change `#[cube(launch)]` to `#[cube(launch_unchecked)`] and call `launch_unchecked` for improved performance.

nathanielsimard (changes requested):
Some changes to be done, but that's crazy how flexible cube is, I never though we would be able to migrate the reduce trait to cube with that amount of ease.

crates/burn-jit/src/kernel/reduce/naive/base.rs
```diff
@@ -1,32 +1,39 @@
...
+        accumulator: <Self::Accumulator as cubecl::frontend::CubeType>::ExpandType,
+        current_value: <EI as cubecl::frontend::CubeType>::ExpandType,
+        _i: <UInt as cubecl::frontend::CubeType>::ExpandType,
+    );
+
+    fn __expand_assign_naive<EO: Numeric>(
+        context: &mut cubecl::frontend::CubeContext,
+        output: <Tensor<EO> as cubecl::frontend::CubeType>::ExpandType,
+        accumulator: <Self::Accumulator as cubecl::frontend::CubeType>::ExpandType,
+        _shape_reduce_dim: <UInt as cubecl::frontend::CubeType>::ExpandType,
```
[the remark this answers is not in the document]
cBournhonesque: oh nice!

crates/burn-jit/src/kernel/reduce/naive/shader.rs
cBournhonesque: done!

cBournhonesque:
> Some changes to be done, but that's crazy how flexible cube is, I never though we would be able to migrate the reduce trait to cube with that amount of ease.

Yes, it's crazy that the code just works like normal rust but is still handled correctly! Very nice that it works across traits as well

nathanielsimard (approved):

Issue #2111, opened 2024-08-05
ONNX import issue: `index out of bounds` in `pad_config()` due to `node.input` only having len of `2`


SimonBrandner:
```
ERROR burn_import::logger: PANIC => panicked at /home/simon/.cargo/git/checkouts/burn-178c6829f420dae1/f7639bd/crates/burn-import/src/onnx/op_configuration.rs:802:27:
index out of bounds: the len is 2 but the index is 2
```

Link to model: https://drive.google.com/file/d/1E6fNKzGE3_JhEokgqqHDqJxwqw3UUku9/view?usp=drive_link

antimora:
Yes, it seems all ops in the ONNX file should be supported.

antimora:
CC @mepatrick73 it's a good one to take to fix a bug.

mepatrick73:
I'll get on it!

mepatrick73:
This problem is now fixed, there are more issues with the model. Specifically the avgpool operation uses ceilling mode which we currently do not support. If you could make a new issue, that would be wonderful. 

Anyhow, I'll close this issue as it is now fixed.

SimonBrandner:
Thanks @mepatrick73!

PR #2120, opened 2024-08-06
bug fix: adding bounds checking to pad ONNX inputs

mepatrick73:
Constant value is an optional value.
Adding a bounds check to make sure we've gotten enough inputs

### Checklist

### Related Issues/PRs

#2111

### Changes
Added a bounds check to the inputs of the Pad operation.

### Testing
Tested using the ONNX model mentionned in #2111


mepatrick73:
While working on this code, I saw that the pad operation doesn't handle the optional axes input. Should we add a panic message saying we don't currently support the axes inputs? @laggui  @antimora

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -799,10 +799,13 @@ pub fn pad_config(node: &Node) -> PadConfig {
...
         // TODO: support int, boolean
-        match &node.inputs[2].value {
-            Some(Data::Float32s(shape)) => shape.first().unwrap().to_owned(),
-            _ => 0.0,
-        }
+        node.inputs
+            .get(2)
+            .and_then(|input| match &input.value {
+                Some(Data::Float32s(shape)) => shape.first().cloned(),
+                _ => None,
```
antimora, line 806: We need to improve the logic a bit. The implementation seems to have a bug. 

If get(2) is None, then default to 0.0.
If get(2) is Some with Float32s then return the first item from `constant_value` (btw `shape` should be renamed to `constant_value`. (I also recommend handling Float16s, Float64s with casting, just in case)
If get(2) is Some with other that floats, then panic (int or boolean probably). Currently I guess it's not fully supported.

antimora (changes requested):
There is a logical bug in the solution. Please see my inlined comment.

antimora:
> While working on this code, I saw that the pad operation doesn't handle the optional axes input. Should we add a panic message saying we don't currently support the axes inputs? @laggui @antimora

Yes, it's good idea to button down checks

laggui (commented):
Apart from the improvements already flagged, the implementation LGTM!

And yes, adding more checks for unsupported configurations (such as a panic message for axes) is a good idea.

Thanks fixing this 🙂

antimora (approved):
LGTM

Issue #2115, opened 2024-08-05
Error importing the FaceONNX model `landmarks_68_pfld`: `Only tensor input is valid but received: Shape(4)`


SimonBrandner:
```
ERROR burn_import::logger: PANIC => panicked at /home/simon/.cargo/git/checkouts/burn-178c6829f420dae1/f7639bd/crates/onnx-ir/src/dim_inference.rs:814:15:
Only tensor input is valid but received: Shape(4)
```

Link to model: https://drive.google.com/file/d/1qgM6ZqMyB60FYlzzxNDyUefifLS0lhag/view?usp=sharing

mepatrick73:
PR #2120 should fix the mentionned issue

antimora:
> [c2b6318](https://github.com/tracel-ai/burn/commit/c2b6318fc3232536c91910923e09ad0f9f30763a) switched to using `ArgType::Shape` for the output of the shapes-Operator instead of just using a 1D `Tensor`.
> 
> As far as I can tell, the ONNX Spec does not differentiate between Shapes and other 1D Tensors; each time an input or output argument contains a shape, the spec defines it as
> 
> > 1D tensor. The shape of the expected output tensor.
> 
> or similar.
> 
> Now, I can see 2 ways forward:
> 
> * Differentiating Args that are a shape vs. just a random 1D Tensor might still be useful, but then other functions like gather need to also work on Shapes.
> * Or the Shape ArgType is dropped in favor of just using 1D Tensors everywhere they were used, aligning with how the ONNX spec does it.
> 
> Is there a way to know if many operators besides `Gather` are usually fed the output from the `Shape` Op?

We started using Shape arg/input/output type. Please see my comment here in the PR: https://github.com/tracel-ai/burn/pull/1815#issuecomment-2212032793

Although ONNX spec does not differentiate types (scalar, shape, tensor are all the same), Burn and other frameworks do. Shape calculation will be a overkill if we use tensor ops because shape information is stored on CPU so it will require constant syncing from GPU device. 

Our reshape onnx node can accept Shape arg type.

SimonBrandner:
After https://github.com/tracel-ai/burn/pull/2128 was merge this is now:

```
ERROR burn_import::logger: PANIC => panicked at /home/simon/.cargo/git/checkouts/burn-178c6829f420dae1/27ca6ce/crates/onnx-ir/src/dim_inference.rs:819:9:
Gather: indices tensor rank above 1 not supported    
```

[...]

PR #1843, opened 2024-05-31
Feat/gather import

agelas:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

#1714 

### Changes

[Gather](https://onnx.ai/onnx/operators/onnx__Gather.html) and [GatherElements](https://onnx.ai/onnx/operators/onnx__GatherElements.html) were separated into their own folders on the ONNX side. Confusingly, ONNX's `GatherElements` is equivalent to PyTorch/burn's `gather` function, while ONNX's `Gather` seems to be equivalent to burn's `select`. In PyTorch, the ONNX `Gather` is an `index_select`, which burn's `select` more or less seems equivalent to when using 1D indexes.

### Testing

Added unit tests.

PR #2128, opened 2024-08-07
feat: adding shape support for gather ONNX operation

mepatrick73:
### Checklist

### Related Issues/PRs

#2115

### Changes

### Testing

### DIFF
--- a/crates/onnx-ir/src/dim_inference.rs
+++ b/crates/onnx-ir/src/dim_inference.rs
@@ -810,11 +810,6 @@ fn gather_update_outputs(node: &mut Node) {
         panic!("Gather requires two inputs: data and indices");
     }
 
-    let input_tensor = match &node.inputs[0].ty {
-        ArgType::Tensor(tensor) => tensor,
-        ty => panic!("Only tensor input is valid but received: {:?}", ty),
-    };
-
     let indices_tensor = match &node.inputs[1].ty {
         ArgType::Tensor(tensor) => tensor,
         _ => panic!("Only tensor indices is valid"),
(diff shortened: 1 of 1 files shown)


mepatrick73:
I've resolved the user's issue. There's a section of code that checks if the rank of the indices tensor is greater than 1, and if so, it triggers a panic. This check seems unnecessary since the underlying gather operation in Burn supports multi-dimensional tensors (though I might be mistaken). The user's model uses a multi-dimensional indices tensor, causing a crash before completing the parsing of the Gather node. If I remove this check, the operation graph progresses further. Should I remove the indices_dim check?

laggui:
> I've resolved the user's issue. There's a section of code that checks if the rank of the indices tensor is greater than 1, and if so, it triggers a panic. This check seems unnecessary since the underlying gather operation in Burn supports multi-dimensional tensors (though I might be mistaken). The user's model uses a multi-dimensional indices tensor, causing a crash before completing the parsing of the Gather node. If I remove this check, the operation graph progresses further. Should I remove the indices_dim check?

If you check the code, the ONNX Gather op actually translates to `tensor.select(dim, indices)`. There is another ONNX op called GatherElements which corresponds to burn's gather operation on a tensor.

So the rank 1 tensor limitation is because `select` only accepts 1d indices.

Linking the original gather PR #1843  for reference.

laggui:
@mepatrick73 this is ready for review right? The changes look good to me w.r.t supporting shape input.

antimora:
Looks good to me so far

hexd0t:
When Gathering from a `Shape`, shouldn't the output also be an `ArgType::Shape` instead of converting it to a `Tensor`?

mepatrick73:
The rank of the output tensor could be bigger than 1, therefore Shape is not appropriate.

mepatrick73:
> @mepatrick73 this is ready for review right? The changes look good to me w.r.t supporting shape input.

Yes it is ready for review :)

laggui (approved):
Thanks for fixing the shape support 🙏

PR #2138, opened 2024-08-08
Add comments for matmul kernel

cBournhonesque:
## Pull Request Template

I'm new to GPU programming, and I'm writing some comments that can hopefully help others understand the kernels.

### DIFF
--- a/crates/burn-jit/src/kernel/matmul/simple.rs
+++ b/crates/burn-jit/src/kernel/matmul/simple.rs
@@ -1,6 +1,6 @@
 //! Naive matmul kernel implementation
 //!
-//! Each local thread will compute a single element of the output matrix.
+//! Each local unit will compute a single element of the output matrix.
 use crate::{
     kernel::{into_contiguous, Kernel, SUBCUBE_DIM_APPROX},
     ops::swap_dims,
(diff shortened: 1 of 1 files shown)


crates/burn-jit/src/kernel/matmul/simple.rs
```diff
@@ -1,3 +1,6 @@
+//! Naive matmul kernel implementation
+//!
+//! Each local thread will compute a single element of the output matrix.
```
nathanielsimard, line 3: I would use unit instead of thread to be more consistent with the naming.

crates/burn-jit/src/kernel/matmul/simple.rs
```diff
@@ -66,6 +70,8 @@ fn matmul_kernel<F: Float>(
     let unroll_sum = Comptime::map(vectorization_factor, |w: UInt| w != UInt::new(1));
     if Comptime::get(unroll_sum) {
         let mut accum = F::new(0.);
+        // we unroll the loop to sum `vectorization_factor` elements at once, which lets us
```
nathanielsimard, line 73: This whole block is simply to handle vectoriation, since when vectorization is enabled, we have `vectorization_factor` different accumulators that need to be summed before stored in the output tensor.

cBournhonesque: yes that's what I was trying to convey by saying that we want so sum `vectorization_factor` at once; feel free to tweak the comment if you feel like it's not accurate

nathanielsimard (approved):

PR #2148, opened 2024-08-09
Fix ONNX Gather codegen for Shape input

hexd0t:
### Checklist

### Related Issues/PRs

This finishes some work still left after #2128 for ONNX Gather nodes to accept Shape inputs.

Without it, the actual codegen fails (Note that the error message is misleading, the Arg here is not a scalar but a Shape(2), but the message predates Shapes being used in ONNX imports):

```
  ERROR burn_import::logger: PANIC => panicked at burn\crates\burn-import\src\onnx\to_burn.rs:1245:18:
  Can't transform scalar to tensor.

  --- stderr
  thread 'main' panicked at burn\crates\burn-import\src\onnx\to_burn.rs:1245:18:
  Can't transform scalar to tensor.
  stack backtrace:
     0: std::panicking::begin_panic_handler
               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library\std\src\panicking.rs:652
     1: core::panicking::panic_fmt
               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library\core\src\panicking.rs:72
     2: burn_import::onnx::to_burn::<impl core::convert::From<&onnx_ir::ir::Argument> for burn_import::burn::ty::TensorType>::from
     3: burn_import::onnx::to_burn::ParsedOnnxGraph::gather_conversion
     4: burn_import::onnx::to_burn::ParsedOnnxGraph::into_burn
     5: burn_import::onnx::to_burn::ModelGen::run_from_script
     6: burn_import::onnx::to_burn::ModelGen::run_from_script
     7: burn_import::onnx::to_burn::ModelGen::run_from_script
     8: build_script_build::burn::import_models
     9: build_script_build::burn::import_models
    10: build_script_build::burn::import_models
    11: core::ops::function::FnOnce::call_once
```

It will probably also be needed for the model in issue #2115 to import successfully.

### Changes

Check if the input value is a Shape instead of a Tensor, and if so, convert the Shape to a tensor before doing the `select` operation.

### Testing

Added an ONNX-Test for this case, as well as a codegen unittest.


laggui (commented):
Just a minor comment for now.

For the new shape type introduced this LGTM.

I'd like to get @antimora's feedback though. I know you wanted to avoid simple shape operations to be done on GPU and retrieved to CPU, but with this typical sequence of tensor -> shape -> gather ops we have the shape results on CPU and the indices for gather as a tensor. Hence we have to either convert the shape back to a tensor (as per this PR) or get the indices as a vec... didn't we just come back full circle to your initial concern?

/edit: after chatting offline I think this case is unavoidable with this sequence of ops - so let's proceed 🙂

antimora:
> Just a minor comment for now.
> 
> For the new shape type introduced this LGTM.
> 
> I'd like to get @antimora's feedback though. I know you wanted to avoid simple shape operations to be done on GPU and retrieved to CPU, but with this typical sequence of tensor -> shape -> gather ops we have the shape results on CPU and the indices for gather as a tensor. Hence we have to either convert the shape back to a tensor (as per this PR) or get the indices as a vec... didn't we just come back full circle to your initial concern?

We spoke offline. It's true we want to use Shape as much as possible (only when dealing with shapes) but in this instance is unavoidable since `select` needs a multidimensional tensor and `shape` information comes from CPU (struct).

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -27,7 +27,15 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for GatherNode {
         node_position: usize,
     ) -> proc_macro2::TokenStream {
         let dim = self.dim.to_tokens();
-        let input = scope.tensor_use_owned(&self.input, node_position);
+        let input = match &self.input {
+            Type::Tensor(in_tensor) => scope.tensor_use_owned(in_tensor, node_position),
+            Type::Shape(in_shape) => {
+                // TODO: do we need a shape_use_owned for the scope, in case the Shape is used multiple times?
```
antimora, line 33: We could probably pass a reference (make it slice) to `Tensor::from_data` since there is an From implementation. This way we borrow and underlying elements are copied:

image

@laggui what do you think?

laggui: Agreed! 👍

hexd0t: Done as suggested

hexd0t:
>  It's true we want to use Shape as much as possible

I haven't implemented this here, but at least for the case where a Scalar index is selected from a Shape, a special path could be added to output another Shape with just the selected size (it would always be dim 1) completely on the CPU.
At least from the (admittedly few) ONNX I've seen where Shape and Gather are combined, the index was usually a Scalar anyways.

Of course, adding this will probably expose even more Ops that cannot handle Shape inputs correctly yet, so implementing something like that would likely introduce regressions in terms of model compatibility at first.

antimora (approved):
LGTM

laggui (approved):

PR #2141, opened 2024-08-08
Implement ONNX Gather for scalar indices

hexd0t:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

#2115

### Changes

Allow ONNX Gather with Scalar = Rank 0 Tensor as Index.

### Testing

An onnx-test has been added as well as a codegen test.

PR #2149, opened 2024-08-09
Fix indices dim check in gather_update_outputs

laggui:
### Checklist

### Related Issues/PRs

We missed a little something in PR #2141

### Changes

Updated `gather_update_outputs` to accept scalar argtype when checking for indices dim.

### DIFF
--- a/crates/onnx-ir/src/dim_inference.rs
+++ b/crates/onnx-ir/src/dim_inference.rs
@@ -810,19 +810,20 @@ fn gather_update_outputs(node: &mut Node) {
         panic!("Gather requires two inputs: data and indices");
     }
 
-    let indices_tensor = match &node.inputs[1].ty {
-        ArgType::Tensor(tensor) => tensor,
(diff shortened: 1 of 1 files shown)


laggui:
@hexd0t you can also review since you contributed the last PR

antimora (approved):
LGTM

hexd0t:
LGTM, although I'm not sure why the `gather_scalar`-Test I had added didn't fail?

Also, in the ONNX model I'm currently trying, I get `Tensor(TensorType { elem_type: Int64, dim: 0, shape: Some([]) })` instead of Scalars at this point (which I understand burn doesn't really want to have anyways)

laggui:
Seems like the model generated for the test actually is a `TensorType` with dim 0 and not a scalar type.

The error was reported to me by a user, and in his case the node received a `ScalarType`, hence the addition in this PR.

I didn't have time to check why in one case it remained a tensor type with dim 0 while the other was converted to a scalar type, but that might be a parsing issue 🤔

Issue #2118, opened 2024-08-06
Wgpu backend of image-classification-web example cannot work


wcshds:
When I choose the wgpu backend, I get errors in the console.

![image](https://github.com/user-attachments/assets/4b096904-72de-4536-8eca-b0cd56797ce9)

After I disable the `autotune` feature of [burn-wgpu](https://github.com/tracel-ai/burn/blob/a53f459f205889a22ecea3713bbae12d3de7eb0c/examples/image-classification-web/Cargo.toml#L21-L23), the wgpu backend still cannot work.

![image](https://github.com/user-attachments/assets/7b10cb01-98f2-4c71-b030-19999f6a4f9a)

wcshds:
I tried printing the tensor converted from the input f32 slice in the console. It seems that the tensor's data has been corrupted from the very beginning.
![image](https://github.com/user-attachments/assets/cddffe9b-a4b9-4327-9644-c50018603b32)

antimora:
@nathanielsimard @louisfd just be aware of this jit related bug on WASM/WebGPU

antimora:
I would be great if we had WebGPU tests on CI #810

wcshds:
This bug is caused by the limited precision of Rust's display of f32 values. It should be an easy fix.

I changed the code from [](https://github.com/tracel-ai/cubecl/blob/cfe0b0204380cbd0931f478194a053a6ac35d1cb/crates/cubecl-wgpu/src/compiler/wgsl/base.rs#L262-L263):

```rust
FloatKind::F32 => f.write_fmt(format_args!("{}f", *val as f32)),
FloatKind::F64 => f.write_fmt(format_args!("{}f", { *val })),
```
to:
```rust
FloatKind::F32 => f.write_fmt(format_args!("{:.32}f", *val as f32)),
FloatKind::F64 => f.write_fmt(format_args!("{:.64}f", { *val })),
```
It can fix the bug.

Jonarod:
I thought maybe it was due to incompatible versions between the main branch and the examples, but I have the same issue after downloading the [0.13.2 release](https://github.com/tracel-ai/burn/archive/refs/tags/v0.13.2.zip) out-of-the-box then `cd examples/image-classification-web` then `./build-for-web.sh`. In fact, it would not even compile without further adding the `features=["autotune"]` to the `burn-wgpu` crate dependency manually before compiling.

[...]

PR #2158, opened 2024-08-12
Make image-classification-web work with the wgpu backend

Jonarod:
### Related Issues/PRs

#2118 

### Changes

Update `cubecl` revision to include Rust precision fix.
Remove `autotune` feature for `burn-wgpu` from example's `Cargo.toml`.


ArthurBrussee:
Autotune should work after https://github.com/tracel-ai/cubecl/pull/69

Autotune still makes things a bit faster on the web so it's nice to leave it enabled - it's just the persistent caching which doesn't work.

PR #2166, opened 2024-08-15
Fix equal/not-equal infinity numbers for burn-ndarray

antimora:
This PR fixes a bug in burn-ndarray where tensor equal and not equal ops do not operate correctly for negative/positive infinities correctly. This was discovered when using contains_nan operator for a tensor that contained -inf numbers with burn-ndarray backend.

### Checklist

### Changes

1. Fixed the bug in float_equal
2. Updated logic int_equal to match float_equal
3. Fixed potential tensor mutation bug in bool_equal by making it consistent with float_equal
4. Added regression tests

### Testing

Added regression tests.


laggui (approved):
LGTM! Thanks for finding & fixing the bug 🙏

PR #2170, opened 2024-08-15
Update model.bin mnist inference web + add cuda-jit flag for ag-news-infer

laggui:
- MNIST model was incompatible with latest `TensorData` changes
- `cuda-jit` flag was only used for training


antimora (approved):
LGTM

PR #1699, opened 2024-04-26
Fix inverted epoch - iteration counts in valid progress

laggui:
The iteration and epoch counts are inverted in the custom training loop example.

```
[Valid - Epoch 0 - Iteration 3] Loss 0.2914832 | Accuracy 90.625
[Valid - Epoch 1 - Iteration 3] Loss 0.08417652 | Accuracy 95.3125
[Valid - Epoch 2 - Iteration 3] Loss 0.1853882 | Accuracy 93.75
[Valid - Epoch 3 - Iteration 3] Loss 0.027896263 | Accuracy 100
[Valid - Epoch 4 - Iteration 3] Loss 0.027823677 | Accuracy 100
[Valid - Epoch 5 - Iteration 3] Loss 0.11582196 | Accuracy 96.875
[Valid - Epoch 6 - Iteration 3] Loss 0.08146551 | Accuracy 98.4375
[Valid - Epoch 7 - Iteration 3] Loss 0.09941933 | Accuracy 96.875
```

Found this little mistake while running all the examples for the patch.

PR #2171, opened 2024-08-19
Switches epoch and iteration to be in the proper order for the custom training loop section of the book

jguhlin:
### Checklist

### Changes

In the custom training loop section of the burn book, iteration and epoch are in the wrong order. This just switches it around.


laggui (approved):
Whoops, looks like I forgot to fix the book in #1699 😅 

Thanks for catching that!

PR #2182, opened 2024-08-19
Fix/wgpu/matmul web

### LINKED ISSUE #2178: `mnist-inference-web` wgpu matmul returns all zeros
**Describe the bug**
The wgpu matmul implementation returns all zeros when running in the browser.

Also reported by a discord user: https://discord.com/channels/1038839012602941528/1059209073784008835/1273995743790628876

**To Reproduce**
Run the `mnist-inference-web` example built for wgpu.

The scores returned are always the same for all digits.

image

Upon further debugging, the `fc1` layer always returns zeros because of the matmul issue (tiling2d implementation w/ cube). Note that using the simple matrix multiply works (and cmma is not available in this case).

**Expected behavior**
The results from the matmul should not be zero, and thus the scores correct.

laggui:
### Checklist

### Related Issues/PRs

Fixes #2178

### Changes

Work around the cubecl/wgpu issue by forcing the simple matmul strategy until read write strategy is fixed.

### Testing

The [MWE](https://github.com/laggui/burn-2178-mwe) works and the mnist-inference-web results are coherent.

### DIFF
--- a/crates/burn-jit/src/kernel/matmul/base.rs
+++ b/crates/burn-jit/src/kernel/matmul/base.rs
@@ -23,6 +23,7 @@ pub enum MatmulStrategy {
 }
 
 impl Default for MatmulStrategy {
+    #[allow(unreachable_code)]
     fn default() -> Self {
         // if wasm, force simple kernel for now (regardless of autotune)
         // issue: https://github.com/tracel-ai/burn/issues/2178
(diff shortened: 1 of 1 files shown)


laggui:
After some offline discussion we decided to go another route to solve the issue closer to the source.

PR #2184, opened 2024-08-19
Add closeness tensor report

antimora:
This is a tensor utility to check closeness of two float tensors quick for various tolerance levels. This helps figure at what level two tensors are the same or similar.

### Checklist

### Changes

1. New function
2. New book section

### Testing

Ran an example in the document and verified visually.

Here is its output:

image

### DIFF
--- a/crates/burn-tensor/src/tensor/report.rs
+++ b/crates/burn-tensor/src/tensor/report.rs
@@ -34,35 +34,35 @@ use super::{backend::Backend, Tensor};
 /// ```text
 /// Tensor Closeness Check Results:
 /// ===============================
-/// Epsilon: 1.00e-1
+/// Epsilon: 1e-1
 ///   Close elements: 10/10 (100.00%)
 ///   [PASS] All elements are within tolerance
 ///
-/// Epsilon: 1.00e-2
+/// Epsilon: 1e-2
 ///   Close elements: 10/10 (100.00%)
 ///   [PASS] All elements are within tolerance
 ///
-/// Epsilon: 1.00e-3
+/// Epsilon: 1e-3
 ///   Close elements: 9/10 (90.00%)
 ///   [WARN] Most elements are within tolerance
 ///
-/// Epsilon: 1.00e-4
+/// Epsilon: 1e-4
 ///   Close elements: 6/10 (60.00%)
 ///   [FAIL] Significant differences detected
 ///
-/// Epsilon: 1.00e-5
+/// Epsilon: 1e-5
 ///   Close elements: 5/10 (50.00%)
 ///   [FAIL] Significant differences detected
 ///
-/// Epsilon: 1.00e-6
+/// Epsilon: 1e-6
 ///   Close elements: 5/10 (50.00%)
 ///   [FAIL] Significant differences detected
 ///
-/// Epsilon: 1.00e-7
+/// Epsilon: 1e-7
 ///   Close elements: 5/10 (50.00%)
 ///   [FAIL] Significant differences detected
 ///
-/// Epsilon: 1.00e-8
+/// Epsilon: 1e-8
 ///   Close elements: 5/10 (50.00%)
 ///   [FAIL] Significant differences detected
 ///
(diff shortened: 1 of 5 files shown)


ArthurBrussee:
I would love to have this in unit tests as well. Would it maybe be possible to pass a min_threshold (relative or absolute) and assert that checks at least up to that eps work? Then when the unit tests fail you'd have a great starting point to know what to do!

antimora:
> I would love to have this in unit tests as well. Would it maybe be possible to pass a min_threshold (relative or absolute) and assert that checks at least up to that eps work? Then when the unit tests fail you'd have a great starting point to know what to do!

Yeah I agree it would be useful for unit tests but it would probably too much if we leave it in. It's probably better to add it write before an assertion to find best level and then remove it.

crates/burn-tensor/src/tensor/report.rs
```diff
@@ -0,0 +1,111 @@
...
+            "  Close elements: {}/{} ({:.2}%)",
+            count, num_elements, percentage
+        );
+
+        if percentage == 100.0 {
+            println!("  \x1b[32m[PASS]\x1b[0m All elements are within tolerance");
+        } else if percentage >= 90.0 {
+            println!("  \x1b[33m[WARN]\x1b[0m Most elements are within tolerance");
+        } else {
+            println!("  \x1b[31m[FAIL]\x1b[0m Significant differences detected");
```
laggui, line 104: The color coding is pretty cool, but do we know if it works properly on different OS's and terminals?

crates/burn-tensor/src/tensor/report.rs
```diff
@@ -0,0 +1,111 @@
...
+            .clone()
+            .is_close(expected.clone(), Some(*epsilon), Some(*epsilon));
+        let data = close.clone().into_data();
+        let num_elements = data.num_elements();
+        let count = data
+            .to_vec::<bool>()
+            .unwrap()
+            .iter()
+            .filter(|&&x| x)
+            .count();
```
laggui, line 90: Instead of doing `data.to_vec::<bool>().unwrap().iter()` you could simplify to `data.iter::<bool>()` 🙂

crates/burn-tensor/src/tensor/report.rs
```diff
@@ -0,0 +1,111 @@
...
+///   [FAIL] Significant differences detected
+///
+/// Closeness check complete.
+/// ```
+
+pub fn check_closeness<B: Backend, const D: usize>(output: &Tensor<B, D>, expected: &Tensor<B, D>) {
+    println!("\x1b[1mTensor Closeness Check Results:\x1b[0m");
+    println!("===============================");
+
+    for epsilon in [1e-1, 1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8].iter() {
```
laggui, line 77: Would be nice to have this configurable instead, but for now I think it's totally fine as a utility function.

laggui (commented):
Thanks for adding this utility! Just a minor comment + question, otherwise LGTM

crates/burn-tensor/src/tensor/report.rs
antimora: Yeah probably you're right. So I updated to use `colored` crate that should account for OS portability.

crates/burn-tensor/src/tensor/report.rs
antimora: Fixed.

crates/burn-tensor/src/tensor/report.rs
antimora: For now we will just it to make the params easy.

crates/burn-tensor/src/tensor/report.rs
```diff
@@ -0,0 +1,110 @@
...
+        println!("{}", "Tensor Closeness Check Results:".bold());
```
laggui, line 80: Looks like you removed the epsilon display and forgot to change the actual header print 😄

laggui (approved):
LGTM!

PR #2186, opened 2024-08-20
Fix LayerNorm normalization.

### LINKED ISSUE #2185: LayerNorm does not normalize its input correctly
**Describe the bug**
Both from pytorch documentation...:
https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html#torch.nn.LayerNorm
```
y = (x − E[x]) / sqrt(Var[x] + ϵ)∗γ + β
```

... and from pytorch ATen implementation...:
https://github.com/pytorch/pytorch/blob/6564e746ed494505be5f65aa99f8cf1884bb6342/aten/src/ATen/native/cpu/layer_norm_kernel.cpp#L115
```python
  rstd_val = float(1) / std::sqrt(rstd_val + eps)
```

... they normalize by `sqrt(var + eps)`. Now if we take a look at burn...:
https://github.com/tracel-ai/burn/blob/c29ed43441fc5b3286749816c9069ed90f178a77/crates/burn-core/src/nn/norm/layer.rs#L70

... we normalize by `sqrt(var) + eps`, which leads to precision issues.

booti386:
### Checklist

### Related Issues/PRs

Fixes #2185 

### Testing

Add a new test with a large epsilon value.

### DIFF
--- a/crates/burn-core/src/nn/norm/layer.rs
+++ b/crates/burn-core/src/nn/norm/layer.rs
@@ -67,7 +67,7 @@ impl<B: Backend> LayerNorm<B> {
     pub fn forward<const D: usize>(&self, input: Tensor<B, D>) -> Tensor<B, D> {
         let (var, mean) = input.clone().var_mean_bias(D - 1);
 
-        let input_normalized = input.sub(mean).div(var.sqrt().add_scalar(self.epsilon));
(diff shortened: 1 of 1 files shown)


laggui (approved):
Thank you for fixing it so quickly! 🙏

PR #2189, opened 2024-08-21
Feat: Allow onnx-import expand op with non-const shapes

hexd0t:
### Checklist

### Related Issues/PRs

None

### Changes

In ONNX-Import, the expand Op previously expected to be able to extract a constant shape from the model. If this wasn't possible, conversion paniced.
With this change, the shape can now also be taken in at runtime, either from a Shape or a Tensor.

### Testing

Codegen and ONNX-Import tests have been added for both the shape and the tensor case.


crates/burn-import/src/burn/node/expand.rs
```diff
@@ -17,14 +23,40 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for ExpandNode {
...
+            ExpandShape::Static(static_shape) => static_shape.to_tokens(),
+            ExpandShape::Runtime(Type::Tensor(shape_tensor)) => {
+                // since we don't take ownership of the shape_tensor, we don't need `tensor_use_owned` here:
+                let tensor_name = &shape_tensor.name;
+                let dim = shape_tensor.shape.as_ref().unwrap()[0];
+                // the shape of the tensor is already validated statically to be rank one when parsing the input
+                // we'll need to download the Tensor from device to cpu for expand operation.
+                // Also, we'll need to convert it to an array for conversion into BroadcastArgs
+                quote! {
+                    TryInto::<[i64; #dim]>::try_into(#tensor_name.to_data().as_slice::<i64>().unwrap()).unwrap()
```
laggui, line 49: Instead of using `as_slice::<i64>()` we should use `as_slice::<B::IntElem>()` since the integer type can be different between backends (e.g., wgpu uses i32).

With this change the whole line would look like this:

```rust
TryInto::<[B::IntElem; #dim]>::try_into(#tensor_name.to_data().as_slice::<B::IntElem>().unwrap()).unwrap()
```

laggui (changes requested):
Thank you for adding this support 🙏 

Just a minor change required for the codegen to work on all backends, otherwise LGTM!

crates/burn-import/src/burn/node/expand.rs
hexd0t: Done, although I've had to generalize the `BroadcastArgs` impl to all `[Element; _]` - compared to previously where it was only implemented for `[i32, _]`, the Shape Array will now be cast to `i64` before the Shape conversion happens, so theoretically it got slightly worse for the i32 case. Since specialization is still unstable, I haven't found a good way to circumvent this.

laggui: I think this is fine for tensor shapes 🙂arrays are only ever going to have a small number of integers for that, plus it's just a nice-to-have API to accept inferred dimensions (-1)

laggui: Technically, you could use the `ElementComparison` and `ElementConversion` traits to do something like this:

```rust
// x < -1
x.cmp(&(-1).elem::<E>()) == Order::Less
```

but then you have to convert the valid values (> 0) to usize for the shape anyway and the element traits are not implemented for this primitive, which means you would have to do something like `x.to_i64() as usize` anyway.

So your current approach looks fine.

laggui (approved):
👏 👏

crates/burn-import/src/burn/node/expand.rs
hexd0t: Another drawback of going that route is that it wouldn't work for the unsigned types (cannot express -1 as an u32 to compare against for example), while with the current implementation, those should work as well (although you cannot make use of the inferred dimensions with unsigned types).

Anyway, thanks for looking into this

laggui: Right! Though we didn't have conversion for `u32` before these changes anyway 😄 but you're correct

PR #2194, opened 2024-08-22
Add RoPE `init_with_frequency_scaling`

laggui:
### Checklist

### Related Issues/PRs

Required for [Llama 3.1](https://github.com/tracel-ai/models/pull/38)

### Changes

There exists many extensions to the basic rotary position embedding (RoPE). By adding a new initialization method which accepts a custom function to transform the frequencies, we can apply custom frequency scaling as done in [Llama 3.1 implementation](https://github.com/meta-llama/llama-models/blob/main/models/llama3/reference_impl/model.py#L45).

### Testing

Added unit test.


nathanielsimard (approved):
LGTM

Issue #2151, opened 2024-08-11
ONNX import issue: `Pad: only float values are currently supported for constant value, submit an issue on github` in `pad_config()`


SimonBrandner:
```
ERROR burn_import::logger: PANIC => panicked at /home/simon/.cargo/git/checkouts/burn-178c6829f420d
Pad: only float values are currently supported for constant value, submit an issue on github
```
Link to model: https://drive.google.com/drive/folders/1zfzHNeGju1r1-5vishZ--uaQNSorA0SJ

mepatrick73:
I am under the assumption that this is for the landmarks_68_pfld.onnx model.
I am currently using main rev and using the following code 
```rust 
use burn_import::onnx::ModelGen;

fn main() {
    // Generate Rust code from the ONNX model file
    ModelGen::new()
        .input("src/model/landmarks_68_pfld.onnx")
        .out_dir("model/")
        .run_from_script();
}
```
When I run cargo build, I get the following error : 
```console
  ERROR burn_import::logger: PANIC => panicked at /home/patrick/work/burn/crates/burn-import/src/onnx/op_configuration.rs:1056:9:
  Reshape: shape tensor must be present for Node { node_type: Reshape, name: "reshape1", inputs: [Argument { name: "averagepool2d1_out1", ty: Tensor(TensorType { elem_type: Float32, dim: 4, shape: Some([0, 3, 112, 112]) }), value: None, passed: false }, Argument { name: "concat1_out1", ty: Tensor(TensorType { elem_type: Float32, dim: 1, shape: None }), value: None, passed: false }], outputs: [Argument { name: "reshape1_out1", ty: Tensor(TensorType { elem_type: Float32, dim: 0, shape: None }), value: None, passed: false }], attrs: {} }

  --- stderr
  thread 'main' panicked at /home/patrick/work/burn/crates/burn-import/src/onnx/op_configuration.rs:1056:9:
[...]

PR #2195, opened 2024-08-22
pad-input-fix: adding support for pads as attributes

mepatrick73:
### Checklist

### Related Issues/PRs
#2151 

### Changes
Added support to pad being passed as an attribute. Also adding a check for the mode 

### Testing

Ran user model mentionned in #2151


crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -781,19 +792,40 @@ pub fn pad_config(node: &Node) -> PadConfig {
...
+                        .clone()
+                        .into_i64s()
+                        .iter()
+                        .map(|&x| {
+                            if x < 0 {
+                                panic!("Pad: Negative pad is not supported");
+                            }
+                            x as usize
+                        })
+                        .collect()
```
laggui, line 814: If we are supporting the `pads` as attribute, which corresponds to [version 2](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Pad-2) of the operator, then we should also parse the `value` attribute to make sure we are correctly capturing all parameters of the node.

Make sure it doesn't come in conflict with the `constant_value` input parsing for later versions of the operator 🙂

laggui (changes requested):
Some small changes required to correctly support opset 2 of the `Pad` operator.

Should be good to go after this 👍

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -872,7 +873,15 @@ pub fn pad_config(node: &Node) -> PadConfig {
                     Some(Data::Float64(constant_value)) => Some(*constant_value as f32),
                      _ => panic!("Pad: only float values are currently supported for constant value, submit an issue on github"),
                 })
-                .unwrap_or(0.0)
+                .unwrap_or(0.0);
+
+        if node.attrs.contains_key("value") {
+            constant_value = node.attrs.get("value").and_then(|value| match value {
```
laggui: Quick clippy suggestion:

```
error: using `Option.and_then(|x| Some(y))`, which is more succinctly expressed as `map(|x| y)`
   --> crates/burn-import/src/onnx/op_configuration.rs:879:30
    |
879 |               constant_value = node.attrs.get("value").and_then(|value| match value {
    |  ______________________________^
880 | |                 AttributeValue::Float32(value) => Some(*value),
881 | |                 _ => panic!("Pad: only float32 values are currently supported for constant value as attribute, submit an issue on github"),
882 | |             }).unwrap_or(0.0);
    | |______________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#bind_instead_of_map
    = note: `-D clippy::bind-instead-of-map` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::bind_instead_of_map)]`
help: try
    |
879 ~             constant_value = node.attrs.get("value").map(|value| match value {
880 ~                 AttributeValue::Float32(value) => *value,
    |
```

So you can just apply the same match statement in `.map(...)` but you don't need to wrap the result in `Some(*value)` because map will only be applied when value is present.

laggui (approved):
Whoops, last comment didn't catch your latest commit for the clippy fix.

LGTM with the changes, thanks!

PR #2196, opened 2024-08-22
Tensor type indent fix

mepatrick73:
### Checklist

### Related Issues/PRs
#2151 
#2115
Dependant on #2195

### Changes
If a TensorType name is a number, adding _ in front to make it a valid Ident for proc_macro2

### Testing

Ran user model mentionned in #2151

### DIFF
--- a/crates/burn-import/src/onnx/op_configuration.rs
+++ b/crates/burn-import/src/onnx/op_configuration.rs
@@ -768,20 +768,9 @@ pub fn tile_config(node: &Node) -> TileConfig {
 
 /// Create a PadConfig from the attributes of the node
 pub fn pad_config(node: &Node) -> PadConfig {
-    fn get_input_values(node: &Node, index: usize) -> Vec<i64> {
-        // If the input is not provided, return an empty vector
-        if node.inputs.get(index).is_none() {
-            return Vec::new();
-        }
-
-        match &node.inputs[index].value {
-            Some(Data::Int64s(shape)) => shape.clone(),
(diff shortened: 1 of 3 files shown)


crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -768,9 +768,20 @@ pub fn tile_config(node: &Node) -> TileConfig {
...
+        // If the input is not provided, return an empty vector
+        if node.inputs.get(index).is_none() {
+            return Vec::new();
+        }
+
+        match &node.inputs[index].value {
+            Some(Data::Int64s(shape)) => shape.clone(),
+            _ => panic!("Tensor data type must be int64"),
+        }
+    }
```
antimora, line 781: Did you intend to include pad related changes? I see you have #2195 open. Perhaps, this is based on that branch and that's why these changes appear?

antimora (approved):
LGTM other than the pad related changes from #2195

crates/burn-import/src/onnx/op_configuration.rs
mepatrick73: Yeah, I had to make the pad related changes in order to test the user model.

laggui (commented):
Thanks for the fix 🙂

Changes for the type ident look good to me! But we should remove the pad changes from this PR before merging (especially since the pad PR needs some minor adjustments).

laggui (approved):

Issue #2192, opened 2024-08-22
burn-import getting error `Gather: indices tensor rank above 1 not supported`


catethos:
**Describe the bug**
When I am trying to import this onnx model: [onnx/model.onnx · Xenova/paraphrase-multilingual-mpnet-base-v2 at m...](https://huggingface.co/Xenova/paraphrase-multilingual-mpnet-base-v2/blob/main/onnx/model.onnx), I am getting this error `Gather: indices tensor rank above 1 not supported`
 
**To Reproduce**
1. download the onnx file from the hf hub
2. run the build script to convert

**Expected behavior**
the conversion should work without error

laggui:
@alteredoxide I know you were working on a PR to resolve that issue, just wanted to let you know we can track the progress here.

alteredoxide:
@laggui Thanks! My attention has been elsewhere for the past couple days, but the last round of tests after in-lining the gather code all passed. I just want to review everything once before my final commit+submit. I'll probably get around to that this Saturday.

laggui:
> @laggui Thanks! My attention has been elsewhere for the past couple days, but the last round of tests after in-lining the gather code all passed. I just want to review everything once before my final commit+submit. I'll probably get around to that this Saturday. 

No rush! Be sure to request my review so I don't miss your PR when you're ready :)

alteredoxide:
@laggui I have created the PR here: #2199 but I don't have the ability to specifically add/request your review.

laggui:
> @laggui I have created the PR here: #2199 but I don't have the ability to specifically add/request your review.

Ah whoops forgot about the contributor restrictions 😅 thank you, will take a look at this tomorrow!

alteredoxide:
Haha no worries

laggui:
Completed in #2199

Issue #2023, opened 2024-07-16
burn-import panic


doronnac:
**Describe the bug**
Hi there, trying to import some ONNX files causes the following panic:
burn-import-0.13.2/src/onnx/from_onnx.rs:363:64:
  index out of bounds: the len is 1 but the index is 1    

My guess is this has something to do with burn-import being wip.
If that's the case, I'll be happy to give it a go at fixing it, with some tips if possible.
I see this project as very valuable so I'd like to support it.

**To Reproduce**
[Repro](https://github.com/doronnac/burn-import-repr1)

antimora:
We will be happy to accept your contribution. You can launch contributor book to read up more on burn onnx.

You should try the main branch. We have over 20 new ops. All contributed recently.

alteredoxide:
@doronnac Have you made any progress on this, or at least come up with a temporary fix/workaround? I am also running into this issue.

antimora:
> @doronnac Have you made any progress on this, or at least come up with a temporary fix/workaround? I am also running into this issue.

Please try main branch if you haven't yet. We made many updates since last release

alteredoxide:
@doronnac Yeah, I have tried two different models:
- https://huggingface.co/mixedbread-ai/mxbai-embed-large-v1/tree/main/onnx/model.onnx
- https://huggingface.co/Alibaba-NLP/gte-large-en-v1.5/tree/main/onnx/model.onnx

I have also tried the quantized versions of both of these, but those result in entirely different errors.

alteredoxide:
@antimora Is there a particular reason for not supporting rank > 1 for the indices in `dim_inference` `Gather` nodes? It seems there is already code in place to compute the output rank for any input shape, but it's  coded to panic if `indices_tensor.dim > 1`? If I comment that block out, then parsing finishes for the model I'm trying to load.
However, there is a different downstream issue that I don't think is related, but perhaps it is?...
```
DEBUG burn_import::burn::graph: Registering node => 'shape'    
  DEBUG burn_import::burn::graph: Registering node => 'constant'    
  ERROR burn_import::logger: PANIC => panicked at /path/to/burn/crates/burn-import/src/onnx/to_burn.rs:1245:18:
  Can't transform scalar to tensor.
```

antimora:
[...]

PR #2199, opened 2024-08-25
Feature/codegen gather indices greater than rank 1

alteredoxide:
### Checklist

    - **NA:** there was nothing in the book regarding gather to begin with

### Related Issues/PRs

- #2023 (where convo began)
- #2192 (most relevant)

### Changes

- Adds support for index tensors of rank > 1 for `Gather` nodes in the burn-import crate.
- The relevant "gather" py models have been modified accordingly, and new test rust models generated.
- Added several tests to `crates/burn-import/src/burn/node/gather.rs` for sanity checks that are faster to than run-checks, but they aren't strictly necessary.

### Testing

- Ran `./run-checks all`
- Ran `cargo test -p burn-import -- burn::node::gather`, which ran the newly added tests for multiple cases:
    - **`test_codegen_gather_idx_1d`:** tests codegen on 1d index input
    - **`test_codegen_gather_idx_2d`:** tests codegen on 2d index input
    - **`test_codegen_gather_shape_input`:** tests codegen for case of a shape input instead of data tensor
    - **`test_codegen_gather_scalar_idx`:** tests codegen for case of the index being passed as scalar instead of tensor
    - **`test_gather_tensor`:** tests actual functionality of gather on input tensor using a 2D index tensor
    - **`test_gather_scalar_idx`:** like `test_gather_tensor` but with scalar index
    - **`test_gather_shape_input`:** tests functionality for shape input instead of data tensor


crates/burn-import/src/burn/node/gather.rs
```diff
@@ -27,33 +27,79 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for GatherNode {
...
+                let slice = Tensor::select(#input, #dim, indices);
+                let #output = slice.squeeze::<#output_rank>(#dim);
+            },
+            _ => match index_rank {
+                1 => quote! {
+                    let #output = Tensor::select(#input, #dim, indices);
+                },
+                _ => quote! {
+                    extern crate alloc;
+                    use alloc::vec::Vec;
```
laggui, line 67: This is not required for `burn-import` crate which doesn't support no_std

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -27,33 +27,79 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for GatherNode {
...
                 // borrow the array as a slice and use that to create the Tensor:
-                quote! { Tensor::from_data(&#in_shape_name as &[_], &*self.device) }
+                quote! { Tensor::<B, 1, Int>::from_data(&#in_shape_name as &[_], &*self.device) }
             }
             _ => panic!("Gather needs Scalar or Shape input, got {:?}!", self.input),
         };
+
         let output = &self.output.name;
 
+        let gather = match &self.index {
```
laggui, line 56: Instead of having two `match` statements on `self.index` one after the other, we should combine the logic for the match arms.

crates/burn-import/onnx-tests/tests/gather/gather.py
laggui, line 1: I think we should keep the gather 1d test case as before since they have different logic and add a new test for gather nd.

crates/burn-import/onnx-tests/tests/onnx_tests.rs
```diff
@@ -456,10 +456,10 @@ mod tests {
 
         let device = Default::default();
 
-        let input = Tensor::<Backend, 2>::from_floats([[1., 2., 3.], [4., 5., 6.]], &device);
-        let index = Tensor::<Backend, 1, Int>::from_ints([0, 2], &device);
+        let input = Tensor::<Backend, 2>::from_data([[1.0, 1.2], [2.3, 3.4], [4.5, 5.7]], &device);
+        let index = Tensor::<Backend, 2, Int>::from_data([[0, 1], [1, 2]], &device);
+        let expected = TensorData::from([[[1f32, 1.2], [2.3, 3.4]], [[2.3, 3.4], [4.5, 5.7]]]);
         let output = model.forward(input, index);
-        let expected = TensorData::from([[1f32, 3.], [4., 6.]]);
```
laggui, line 462: (repeating myself here) let's keep the existing test for gather_1d and add a new test for gather_nd 🙂

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -27,33 +27,79 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for GatherNode {
...
+                    let index_flat = match n_dims {
+                        1 => indices.reshape([1, -1]),
+                        n if n >= 2 => indices.flatten::<2>(0, n - 2),
+                        _ => panic!("Number of dimensions must be greater than 0"),
+                    };
+
+                    for idxs in index_flat.iter_dim(0) {
+                        let idxs = idxs.squeeze::<1>(0);
+                        let slice = Tensor::select(#input.clone(), #dim, idxs);
+                        out.push(slice);
```
laggui, line 80: Let's refactor the vector creation to a map + collect instead:

```rust
let out = index_flat.iter_dim(0).map(|idxs| Tensor::select(#input.clone(), #dim, idxs.squeeze::<1>(0))).collect();
```

laggui (changes requested):
Great job! 💪 The ONNX import with codegen is not necessarily the most straightforward part of the codebase to jump into without any prior knowledge 😅 The implementation looks good!

I just have a couple of changes (mostly regarding the tests removed/added, and one to simplify the codegen conditions for the node).

laggui:
Ahh I guess it would depend on the burn config in the project! If you have default features disabled (so, no std) then it might be required.

In this case, you're right. So we should probably use alloc, but instead of having it in the codegen I'm pretty sure we could add it conditionally in the `register_imports` (see for example [`Squeeze`](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/burn/node/unsqueeze.rs#L49)).

alteredoxide:
Cool, I'll check that out, thanks. I was wondering where/how imports were set for codegen.

alteredoxide:
@laggui A couple notes:
1. I can confirm that the build fails without using alloc:
```
error[E0433]: failed to resolve: use of undeclared type `Vec`
  --> /home/ao/projects/third-party/rust/burn/target/debug/build/onnx-tests-6ab92c211618b15e/out/model/gather.rs:45:23
   |
45 |         let mut out = Vec::new();
   |                       ^^^ use of undeclared type `Vec`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `onnx-tests` (test "onnx_tests") due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```
2. Registering imports as follows results in ``use of undeclared crate or module `alloc` ``:
```
    fn register_imports(&self, imports: &mut BurnImports) {
        match &self.index {
            Type::Tensor(_) => {
                imports.register("alloc::vec::Vec");
            },
            _ => {}
        }
    }
```

I also tried adding `imports.register("extern crate alloc")`, just in case, but it didn't like that.
Any suggestions?

laggui:
@alteredoxide Actually, if you use the suggested `iter_dim` + `map` + `collect` implementation instead, you don't need to import `Vec`. And the type hint (e.g., `collect::<Vec<_>>()`) shouldn't be required since the compiler will infer the type from the `Tensor::stack` signature.

Give it a shot and let me know how that goes 🙂

alteredoxide:
@laggui Funny, the moment I submitted my last message, I had a feeling the collect was going to come up. I hadn't forgotten about it; I was just testing each change as I made them. But you're right, that solved the issue!

I'm pretty sure I made all requested changes, and tests + run_checks are green. Lmk if you want any further tweaks.

crates/burn-import/onnx-tests/build.rs
```diff
@@ -32,7 +32,8 @@ fn main() {
         .input("tests/exp/exp.onnx")
         .input("tests/expand/expand.onnx")
         .input("tests/flatten/flatten.onnx")
-        .input("tests/gather/gather.onnx")
```
laggui, line 35: If we're not using this file anymore due to the rename, we should remove it from the workspace 🙂

laggui (approved):
Thanks for applying the changes! 👍 

Don't forget to remove the (now unused) `gather.onnx` file from the workspace now that the 1d and 2d cases have been renamed.

I'm approving in advance 🙂

alteredoxide:
@laggui `gather.onnx` has been deleted and change pushed... I don't have write access, so I can't merge. Thanks!

Issue #2180, opened 2024-08-19
Crash when calling Tensor::nonzero() on a tensor that all values are zero.


KhaledHosseini:
**Describe the bug**

Crash when calling Tensor::nonzero() on a tensor that all values are zero. Is this a bug or is the behavior intentional? (because of zero size tensor).
**To Reproduce**

 
```rust
use burn::{
    backend::wgpu::Wgpu,
    tensor::{backend::Backend, Tensor},
};

type MyBackend= Wgpu;

fn main() {
   let device = Default::default();
   let bools = Tensor::<MyBackend, 1>::zeros([1000], &device).bool();
   // let bools = Tensor::<MyBackend, 1>::full([1000],0.0, &device).bool();
    let non_zero_indices = Tensor::nonzero(bools); // panics ... we need to use Tensor::any() before calling this function.
}

```

**Expected behavior**
Should return a vector of zero size tensors or a vector of Option<Tensor>.

PR #2212, opened 2024-08-28
Nonzero should return an empty vec for zero tensors

### LINKED ISSUE #1946: bool nonzero panics when the entire tensor is false, on wgpu
**Describe the bug**
the nonzero operator for boolean tensors, if given a tensor that is entirely false, panics. It isnt a particularly nice panic either.

The error:
```
thread 'main' panicked at /home/mcarthur/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.20.1/src/backend/wgpu_core.rs:2996:5:
wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
    Buffer binding size 0 is less than minimum 4
      note: buffer = `<Buffer-(11, 2, Vulkan)>`
```

This was produced by the following:
```rust
type B = Autodiff<Wgpu>;
let device = burn::backend::wgpu::WgpuDevice::BestAvailable;
let tensor: Tensor<B, 1, Bool> = Tensor::from_ints([0, 0, 0], &device).equal_elem(1);
tensor.nonzero();
```

Because of the error message, I tested it with NdArray and it worked fine, so I suspect its a WGPU issue.

**Expected behavior**
I would expect it to either panic with a more helpful error.... or give me an empty tensor

- Nixos, unstable
- AMD gpu and intel CPU, in case it is relevant

laggui:
### Checklist

### Related Issues/PRs

Fixes #1946 #2180 

Blocked by this [wgpu fix in cubecl](https://github.com/tracel-ai/cubecl/pull/86)

### Changes

Tensors where all elements are zero should return an empty vector since no valid indices represent non-zero elements.

### Testing

Added unit test (currently fails with wgpu without the fix).


laggui:
Huh, I expected the added unit test to fail on wgpu without the cubecl fix merged 🤔 it fails locally when I test the same operation in a simple program

```rust
let tensor = Tensor::<Wgpu, 1, Bool>::from([false, false, false, false, false]);
let output = tensor.nonzero();
```

```
thread 'main' panicked at /home/laggui/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cubecl-runtime-0.2.0/src/memory_management/memory_pool/small.rs:109:9:
  left: 0
 right: 32
```

nathanielsimard (approved):

PR #2213, opened 2024-08-28
Improve ONNX Tensor-shape tracking

hexd0t:
### Checklist

### Related Issues/PRs

(Includes / based on #2189 - dividing them out just to rebase back onto it seems unnecessary busywork)

### Changes

- Calculate result of broadcasting in dim_inference
- keep Shape info when converting from Argument to TensorType
- Remove a few sources of Dim = 0 Tensors, create Scalars instead
- Clean up dim_inference a bit

### Testing

Added a few asserts that are checked by existing test cases.

### DIFF
--- a/crates/burn-import/src/onnx/to_burn.rs
+++ b/crates/burn-import/src/onnx/to_burn.rs
@@ -424,12 +424,7 @@ impl ParsedOnnxGraph {
 
     fn random_uniform_conversion(node: Node) -> RandomUniformNode {
         let output = node.outputs.first().unwrap();
-        // cannot use output.to_tensor_type() here, since it drops the shape info...
-        let output_type = if let Type::Tensor(t) = Type::from(output) {
-            t
-        } else {
-            panic!("RandomUniform output type is no Tensor.");
-        };
+        let output_type = TensorType::from(output);
 
         let high = node
             .attrs
@@ -451,12 +446,7 @@ impl ParsedOnnxGraph {
 
     fn random_normal_conversion(node: Node) -> RandomNormalNode {
         let output = node.outputs.first().unwrap();
-        // cannot use output.to_tensor_type() here, since it drops the shape info...
-        let output_type = if let Type::Tensor(t) = Type::from(output) {
-            t
-        } else {
-            panic!("RandomNormal output type is no Tensor.");
-        };
+        let output_type = TensorType::from(output);
 
         let mean = node
             .attrs
@@ -480,11 +470,12 @@ impl ParsedOnnxGraph {
         // Additional types needed for ConstantOfShape:
         use crate::burn::node::constant_of_shape::ConstantValue;
 
-        let input = node
-            .inputs
-            .first()
-            .expect("ConstantOfShape requires an input tensor");
-        let output = node.outputs.first().unwrap();
+        let input = Type::from(
+            node.inputs
+                .first()
+                .expect("ConstantOfShape requires an input tensor"),
+        );
+        let output = Type::from(node.outputs.first().unwrap());
 
         // The value of the output elements.Should be a one-element tensor.
         // If not specified, it defaults to a tensor of value 0 and datatype float32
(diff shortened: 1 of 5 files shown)


crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -1082,12 +1073,27 @@ impl ParsedOnnxGraph {
...
         let shape = expand_config(&node);
 
-        ExpandNode::new(input, output, shape)
+        // dim_interference left the dim at zero, so it needs to be filled before converting to TensorType:
+        assert_eq!(
+            node.outputs.len(),
+            1,
+            "ExpandNode must have exactly 1 output!"
+        );
+        let mut output_arg = node.outputs.pop().unwrap();
```
laggui, line 1086: Instead of taking `mut node: Node` and using `.pop()`, why not use `.first() as before?

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -1082,12 +1073,27 @@ impl ParsedOnnxGraph {
...
     }
 
-    fn expand_conversion(node: Node) -> ExpandNode {
+    fn expand_conversion(mut node: Node) -> ExpandNode {
         let input = TensorType::from(node.inputs.first().unwrap());
-        let output = TensorType::from(node.outputs.first().unwrap());
         let shape = expand_config(&node);
 
-        ExpandNode::new(input, output, shape)
+        // dim_interference left the dim at zero, so it needs to be filled before converting to TensorType:
```
laggui, line 1080: I know it's just a comment, but "dim interference" should be "dim inference" 😄

crates/onnx-ir/src/proto_conversion.rs
```diff
@@ -242,21 +242,27 @@ impl TryFrom<ValueInfoProto> for Argument {
...
-            elem_type,
-            shape: Some(
-                tensor_proto
-                    .shape
-                    .dim
-                    .iter()
-                    .map(|x| x.dim_value() as Dim)
-                    .collect(),
-            ),
+        let ty = if tensor_proto.shape.dim.is_empty() {
```
laggui, line 245: I could be wrong, but aren't scalars sometimes also represented as a tensor with `dim: 0`? 🤔

laggui (commented):
Thanks for the cleanup and improved support!

Overall, the implementation LGTM 💪 

Just some minor comments below before we can merge.

hexd0t:
I've updated #2189 and will rebase this once the other PR is merged

laggui:
Thank you! Just merged the other PR 🙏

crates/burn-import/src/onnx/to_burn.rs
hexd0t: The Argument in `node.outputs[0]` has `.dim = 0` from `dim_inference` (as the typo-ridden comment also explains 😅), but `TensorType::from` doesn't allow `dim = 0` anymore.
Hence, we need to modify the Arg some way. I saw 3 and thought this the best:
- Clone the Argument. This would work without `mut Node` and with the first, but would unnecessarily clone stuff, which goes against my microoptimizing tendencies
- get an &mut to the Arg within the `outputs`-Vec: still needs `mut Node`, could go with `iter_mut().first()` instead of `.first()`
- `pop()` the single output to a local `mut` Var and edit it there.

It's not a hill I'll die on, so if you prefer another option, I can change it

crates/onnx-ir/src/proto_conversion.rs
hexd0t: In ONNX? Yes, that's what this is checking for. In Burn, I think a Tensor with dim = 0 is unwanted.

Note that `tensor_proto.shape.dim` is a `Vec` of the shape dimensions (maybe another rename candidate, similar to `Tensor::dim` -> `rank` ), so `dim = tensor_proto.shape.dim.len()` (see also line 251 below), therefore the check  `dim == 0` becomes `tensor_proto.shape.dim.len() == 0`, and `clippy` ahs hammered into me to replace `Vec::len() == 0` with `Vec::is_empty()`

crates/burn-import/src/onnx/to_burn.rs
laggui: Ahhh you're right, that's my bad. I didn't realize that the `match` expression result was assigned to `output_arg_tensor`.

All good, we can leave it as is.

crates/burn-import/src/onnx/to_burn.rs
hexd0t: Fixed together with the rebase

laggui (approved):

PR #2218, opened 2024-08-29
Fix ONNX where op for scalar inputs

hexd0t:
### Checklist

### Related Issues/PRs

#2213 exposes Operators that cannot handle `Scalar` inputs, since they no longer are treated as dim-0 Tensors

### Changes

Update the `dim_inference` for Where, so it can actually accept Scalar inputs.
Also rework the Codegen to handle uploading of Scalars to Tensors where needed.

### Testing
Added codegen tests as well as onnx tests.


laggui (approved):
Hey thanks for going through all of this 😄 

No comments on the implementation, looks good! 🙏

PR #2220, opened 2024-08-30
Fixed raspberry pi pico example not compiling

BjornTheProgrammer:
The import for the model didn't get renamed properly when the name changed from onnx-inference-rp2040 to raspberry-pi-pico.

### Checklist

### Related Issues/PRs

N/a

### Changes

Just renamed import

### Testing

Ran the built example change on the pico.


laggui (approved):
Dang that slipped through the review 😅 

Thanks for catching that!

PR #2226, opened 2024-08-30
Fix panic messages being invisible in tui mode

### LINKED ISSUE #2062: To see the logs and panic need to compile release version.
**Describe the bug**
I'm trying to train the model. There is inconsistency between debug and release version or I'm doing it wrong. If I run it like `cargo run -- model train` I do not see any logs if something causing application to fail. To see the logs I have to do some kind of dirty hack, compile as release version (and painc="abort"). In this case I can see what's going on:

```
2024-07-25T01:57:26.754083Z  INFO wgpu_core::instance: Adapter Metal AdapterInfo { name: "Apple M1 Pro", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }
thread 'main' panicked at /Users/username/.cargo/registry/src/index.crates.io-6f17d22bba15001f/burn-tensor-0.13.2/src/tensor/api/float.rs:181:9:
                                                                                                                                                                  === Tensor Operation Error ===
           Operation: 'Matmul'
                                Reason:
                                           1. The inner dimension of matmul should be the same, but got 12 and 8. Lhs shape [128, 12], rhs shape [8, 128].

                                                                                                                                                           stack backtrace:
                                                                           fish: Job 1, 'RUST_BACKTRACE=1 ./app model …' terminated by signal SIGABRT (Abort)
```

I use latest version of Burn framework: 0.13 with ["wgpu", "train"]

**To Reproduce**
Write an burn app which will compile but cause panic.  Run with `cargo run` and run compiled into release mode app with `panic = "abort"`.

**Expected behavior**
I would like to see panics in debug version also.

 - OS: iOS
 - Browser: safari
 - Version: Sonoma 14.5

As far as I understand it can be that TUI switching into the Alternate Screen and logs go there.

PaulWagener:
### Checklist

### Related Issues/PRs

[Issue 2062: To see the logs and panic need to compile release version.](https://github.com/tracel-ai/burn/issues/2062)

### Changes

This PR fixes #2062 by installing a custom panic hook that resets the terminal to normal mode before running the active panic hook.

### Testing

The issue can be reproduced by calling panic!() while the TUI is active. With this change the panic message is visible to the user.

### DIFF
--- a/crates/burn-train/src/renderer/tui/renderer.rs
+++ b/crates/burn-train/src/renderer/tui/renderer.rs
@@ -7,6 +7,8 @@ use crossterm::{
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
 use ratatui::{prelude::*, Terminal};
+use std::panic::{set_hook, take_hook, PanicHookInfo};
+use std::sync::Arc;
 use std::{
     error::Error,
     io::{self, Stdout},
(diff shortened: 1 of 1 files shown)


laggui (approved):
Thanks for this fix! It'll be a lot more explicit now 😅

PR #2249, opened 2024-09-04
[burn-autodiff] Fix abs NaN when output is 0

AsherJingkongChen:
### Checklist

### Related Issues/PRs

#1441 

### Changes

The previous `float_abs` on `Autodiff` backend produces NaN when the output is zero.

I've made a small change on how the gradient is derived.

### Testing

I have added a test to cover the case. It works as how PyTorch does.

There is a `0.0` in the `tensor_2`, which produces a `0.0` because `abs(0.0) = 0.0`.


nathanielsimard (approved):

louisfd (approved):
Yup, the math works out. Thanks!

Issue #1061, opened 2023-12-12
CI Setup: Github Action Templates


nathanielsimard:
Create github action templates to avoid duplication regarding step setup such as installing dependencies.

syl20bnr:
This is done, we have now reusable composite actions and workflows in our dedicated [github-actions repository](https://github.com/tracel-ai/github-actions).

PR #2261, opened 2024-09-09
A github action example for burn with tch feature enabled

kingwingfly:
### Checklist

### Related Issues/PRs

- #1061 

### Changes

- Add a GitHub action example
- Make setup step of `burn-tch` for macOS more correctly.

### Testing

The GitHub action can work well in [my repo](https://github.com/kingwingfly/burnxp/actions) correctly.

### DIFF
--- a/examples/tch-ci/CHANGELOG.md
+++ b/examples/tch-ci/CHANGELOG.md
@@ -1,0 +1,18 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](https://semver.org).
+
+<!--
+Note: In this file, do not use the hard wrap in the middle of a sentence for compatibility with GitHub comment style markdown rendering.
+-->
+
+## [Unreleased]
+### [0.1.1-alpha1] - 2024-09-05
+
+- Example
+
+## [0.1.0] - 2024-09-03
+
+- Example
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ members = [
 
 exclude = [
     "examples/notebook",
+    "examples/tch-ci",
     "examples/raspberry-pi-pico", # will cause dependency building issues otherwise
     # "crates/burn-cuda",     # comment this line to work on burn-cuda
 ]
--- a/crates/burn-tch/README.md
+++ b/crates/burn-tch/README.md
@@ -211,6 +211,8 @@ export LIBTORCH=/absolute/path/to/libtorch/
 export LD_LIBRARY_PATH=/absolute/path/to/libtorch/lib:$LD_LIBRARY_PATH
 ```
 
+**Note:** if relative path is used, the compiler may not work correctly.
+
 **Note:** make sure your CUDA installation is in your `PATH` and `LD_LIBRARY_PATH`.
 
 </details><br>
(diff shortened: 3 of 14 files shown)


examples/tch-ci/CHANGELOG.md
```diff
@@ -0,0 +1,18 @@
+# Changelog
```
syl20bnr, line 1: Not sure we want to keep this file. This will likely become out of date as we are not used to do this. You can remove it.

syl20bnr (commented):
@kingwingfly Not sure what you tried to achieve here. Can you elaborate of what this PR does exactly and how it solves the linked issue ?

examples/tch-ci/CHANGELOG.md
kingwingfly: I may misread the issue. Sorry. If so, please just close this PR.

I guess [the issue](https://github.com/tracel-ai/burn/issues/1061) is asking for a `github action` yaml example to run build and release workflow, and I happened to do this in an `tch` feature enabled project, so I made this PR to provide an example. But now I found that it may not ask for an example...

As to the `CHANGELOG`, it's necessary to the release step in my github workflow example.

kingwingfly:
Changes this PR does:

- modified burn-tch's README:
  - make preparations about macOS MPS more correct
  - a link to new added GitHub workflow example
  - hint about `.cargo/config.toml` to avoid so many env vars
- tch-ci (an example of GitHub workflow to automatize build and release step for burn projects)
  - I may misunderstand the original issue, maybe it's not asking for such an example
  - `dist/setup_*`: scripts to setup torch env
  - `dist/train_*`: scripts to setup env vars and run train binary
  - the `release.yaml` will replace the suffix of `dist/*` and put them in `run/` to let end-user call
  - `scripts/*`: scripts to help developers setup the env, and they're used by `release.yaml`
  - `CHANGELOG`: this is needed by the `create-release` step in `release.yaml`
  - Things `release.yaml` does can be found in the added README

syl20bnr:
> I guess https://github.com/tracel-ai/burn/issues/1061 is asking for a github action yaml example to run build and release workflow, and I happened to do this in an tch feature enabled project, so I made this PR to provide an example. But now I found that it may not ask for an example...

Indeed the issue was more about simplifying our existing CI workflow by making reusable composite actions and workflows (which we put in out [github-actions repo](https://github.com/tracel-ai/github-actions)). So basically the issue https://github.com/tracel-ai/burn/issues/1061 is done, I am going to close it.

Now I find it interesting that you have a job to setup torch etc... because right now our CI workflow does not support the tests in `burn-tch` and `burn-cuda`  and both crates are disabled as you can see here: https://github.com/syl20bnr/burn/blob/main/xtask/src/commands/test.rs#L35-L39

I see that your actions only creates packages though, they actually don't run any tch code right ?

kingwingfly:
> I see that your actions only create packages though, they actually don't run any tch code right ?

It just compile the binary, so only nvcc in CUDA is prepared. And the size of CUDA + libtorch is too big to run on GitHub action, or a storage warning would be raised. Apart from that, github's machine may not have a GPU. However, for macOS, due to it only depends on python's lobtorch, the bibary can probably run on it.

kingwingfly:
> And the size of CUDA + libtorch is too big to run on GitHub action

Maybe  this is not right, I found this by installing CUDA from local, maybe things would change by installing from network.

PR #2271, opened 2024-09-11
enable doc_auto_cfg to show feature-req-hint in docs.rs

kingwingfly:
### Checklist

### Related Issues/PRs

None

### Changes

According to [this](https://github.com/rust-lang/rust/issues/43781), `cargo doc` has an cool feature `doc_auto_cfg`, which helps show the features-required hints like `tokio`.

```rust
# in lib.rs
#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```

```toml
# in Cargo.toml
[package.metadata.docs.rs]
features = ["doc"]
rustdoc-args = ["--cfg", "docsrs"]
```

Apart from that, I also add `documentation` keys in `[package]`.

### Testing

```sh
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --features doc --no-deps -p burn --open
```

I just checked that it works and show things like:

Screenshot 2024-09-11 at 17 24 06

But I don't know how to automize this with xtask, and it needs nightly Rust as the version used by docs.rs.

### DIFF
--- a/crates/burn/src/lib.rs
+++ b/crates/burn/src/lib.rs
@@ -1,5 +1,6 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![warn(missing_docs)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 //! # Burn
 //!
--- a/crates/burn/src/lib.rs
+++ b/crates/burn/src/lib.rs
@@ -93,6 +93,8 @@
 //!   - `std`: Activates the standard library (deactivate for no_std)
 //!   - `network`: Enables network utilities (currently, only a file downloader with progress bar)
 //!   - `experimental-named-tensor`: Enables named tensors (experimental)
+//!
+//! You can also check the details in sub-crates [`burn-core`](https://docs.rs/burn-core) and [`burn-train`](https://docs.rs/burn-train).
 
 pub use burn_core::*;
 
--- a/crates/burn/src/lib.rs
+++ b/crates/burn/src/lib.rs
@@ -96,8 +96,7 @@
 //!   - `experimental-named-tensor`: Enables named tensors (experimental)
 
 pub use burn_core::{
-    config, constant, grad_clipping, module, nn, prelude, record, serde,
-    tensor, LearningRate,
+    config, constant, grad_clipping, module, nn, prelude, record, serde, tensor, LearningRate,
 };
 
 #[cfg(feature = "std")]
--- a/crates/burn/src/lib.rs
+++ b/crates/burn/src/lib.rs
@@ -96,9 +96,12 @@
 //!   - `experimental-named-tensor`: Enables named tensors (experimental)
 
 pub use burn_core::{
-    config, constant, data, grad_clipping, lr_scheduler, module, nn, optim, prelude, record, serde,
+    config, constant, grad_clipping, module, nn, prelude, record, serde,
     tensor, LearningRate,
 };
+
+#[cfg(feature = "std")]
+pub use burn_core::{data, lr_scheduler, optim};
 
 /// Backend module.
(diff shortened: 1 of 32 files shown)


nathanielsimard (changes requested):
Somehow it doesn't work for everything:

image

I would have expected the backend modules to show which feature flag is necessary, do you have an idea why it doesn't appear to work for those?

kingwingfly:
Oh, it seems not so simple. However, it works well for `burn-core` and `burn-train`:

Screenshot 2024-09-17 at 00 11 01

But  it fails when coming to `burn`.

It's so buggy, and I may spend more time in finding the fixing method.

kingwingfly:
@nathanielsimard 

I found a way to fix this, and now it works for `burn` package. However, it is just to re-export all things in `burn_core` and `burn_train` instead of easily `pub use burn_core::*`, now it looks quite tedious.

laggui:
> @nathanielsimard
> 
> I found a way to fix this, and now it works for `burn` package. However, it is just to re-export all things in `burn_core` and `burn_train` instead of easily `pub use burn_core::*`, now it looks quite tedious.
> 
> I have compared the new doc page and the old, and I think nothing lost. I hope I didn't forget exporting something.

That seems a little error prone 😅 would be nice if there was a way to do this properly without re-exporting

kingwingfly:
> That seems a little error prone 😅 would be nice if there was a way to do this properly without re-exporting

I hope so. I found current less graceful solution by a minimal example. I'll try asking for help from the feature developper and the community.

kingwingfly:
I just found this is allowed:
```rust
use crate_a::*; // instead of `use crate_a::{B, C, D, ...}

#[cfg(feature = "a")]
use crate_a::A;
```

Now it will be less possible to make a mistake. 

By the way, as the features are given in the `Cargo.toml`, it is reasonable to use it in the package somewhere explicitly maybe... 

The features are written in `Cargo.toml` only for generating feature page, and user can still enable them with `--features tch,train`. So, using it in `lib.rs` can at least make them more meaningful. 

And I found there are also many problems about features in this repo. For example, my editor told me `tch` and `train` features are enabled even they are not included in `default` feature. And they are actually enabled by `example/text-generation`. This can be prevented by [required-features](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field), and it's also tedious to fix ( here's an example: 1. [set required features](https://github.com/kingwingfly/encrypt-config/blob/dev/tests/Cargo.toml) and 2. [set --features for each example](https://github.com/kingwingfly/encrypt-config/blob/dev/.github/scripts/test.sh) )... I'll make another issue as I become available.

nathanielsimard:
@kingwingfly When building an example from the root workspace directory, all dependencies listed in the workspace are built, not one the ones specified in the example. I though this was a cargo workspace thing, is there a way to remove that behavior?

kingwingfly:
@nathanielsimard 

#2289

Using required-feature can solve this problem. But it then needs to modify test and example shell scripts to enable features needed.

crates/burn/src/lib.rs
```diff
@@ -96,8 +97,45 @@
...
     pub use burn_train::*;
+
+    /// The metric module.
+    pub mod metric {
+        pub use burn_train::metric::*;
+
+        #[cfg(feature = "metrics")]
+        pub use burn_train::metric::{
+            CpuMemory, CpuTemperature, CpuUse, CudaMetric, TopKAccuracyInput, TopKAccuracyMetric,
+        };
```
nathanielsimard, line 131: I don't like the fact we need to know when adding a new metrics, that it should be manually re-exported in the burn root crate. Only for the feature flag to show on the docs :/

nathanielsimard (approved):

Issue #698, opened 2023-08-25
Use `AsRef<Path>` instead of `&str` for file operations


Colerar:
e.g.

https://github.com/burn-rs/burn/blob/fb2a71bb81e1a688b4cdae38729b24dd9361283f/burn-core/src/config.rs#L47

This allow caller pass `PathBuf`, `String`, `&str` and [all types implemented `AsRef<Path>`](https://doc.rust-lang.org/stable/std/convert/trait.AsRef.html#implementors), and it's API-compatible, it won't break any existing code.

antimora:
I agree.

Should we search all places where it's used, or do we know where it is used already?

nathanielsimard:
I think we should search for more usage, especially regarding saving configuration in JSON.

ivnsch:
I'll give this a go.

laggui:
Seems like this was completed in #699 and #1919

PR #2273, opened 2024-09-12
Use AsRef<Path> instead of &str for file operations

ivnsch:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/698

### Changes

Replaced all file related `&str` with `AsRef<Path>`

Exception: https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/burn/graph.rs#L347
This generates code, where `std::path::Path` can be added, but std is not available.
Also not sure whether e.g. [here](https://github.com/ivnsch/burn/blob/replace_str_with_as_ref/examples/guide/src/inference.rs#L10) the order of the type variables is correct, should it match the order of the parameters, or just be appended (current)?

Some calls got slightly noisier, but not necessarily an issue https://github.com/ivnsch/burn/blob/replace_str_with_as_ref/examples/guide/src/bin/infer.rs#L13

Anyway, just opening the PR and waiting for feedback.

### Testing

I ran `./run-checks.sh all` and `cargo test`. `cargo test` is currently not helping much, because of all the cuda errors (don't have nvidia and not sure how to exclude them).

These changes seem unlikely to affect the book, but can review it once this has received some feedback.

### DIFF
--- a/crates/burn-core/tests/test_derive_config.rs
+++ b/crates/burn-core/tests/test_derive_config.rs
@@ -1,5 +1,6 @@
 use burn::config::{config_to_json, Config};
 use burn_core as burn;
+use std::path::Path;
 
 #[derive(Config, Debug, PartialEq, Eq)]
 pub struct TestEmptyStructConfig {}
@@ -26,7 +27,7 @@ pub enum TestEnumConfig {
 
 #[cfg(feature = "std")]
 #[inline(always)]
-fn file_path(file_name: &str) -> std::path::PathBuf {
+fn file_path<P: AsRef<Path>>(file_name: P) -> std::path::PathBuf {
     std::env::temp_dir().join(file_name)
 }
 
--- a/crates/burn-import/src/onnx/to_burn.rs
+++ b/crates/burn-import/src/onnx/to_burn.rs
@@ -102,14 +102,14 @@ impl ModelGen {
     }
 
     /// Set output directory.
-    pub fn out_dir(&mut self, out_dir: &str) -> &mut Self {
-        self.out_dir = Some(Path::new(out_dir).into());
+    pub fn out_dir<P: AsRef<Path>>(&mut self, out_dir: P) -> &mut Self {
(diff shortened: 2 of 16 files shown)


laggui (commented):
Thanks for taking a look at the issues!

I think the intention from the linked issue was for the user facing APIs. IMO the usage for strings/slices doesn't need to be changed for *everything* in the repo (such as examples, or stuff that is not user facing).

>  Also not sure whether e.g. [here](https://github.com/ivnsch/burn/blob/replace_str_with_as_ref/examples/guide/src/inference.rs#L10) the order of the type variables is correct, should it match the order of the parameters, or just be appended (current)?

I don't think the order matters that much. Usually it follows the arguments but that's just because most people add the generics as they read from left to right, following the order of the args. So it wouldn't matter really in this case, but anyway I wouldn't necessarily change the code in the examples as it is.

ivnsch:
Ok, the majority of the changes are in examples, except a test ([burn-core/tests/test_derive_config.rs](https://github.com/ivnsch/burn/blob/replace_str_with_as_ref/crates/burn-core/tests/test_derive_config.rs#L30)) and `burn-fork/crates/burn-import/src/onnx/to_burn.rs` [change 1](https://github.com/ivnsch/burn/blob/replace_str_with_as_ref/crates/burn-import/src/onnx/to_burn.rs#L105), [change 2](https://github.com/ivnsch/burn/blob/replace_str_with_as_ref/crates/burn-import/src/onnx/to_burn.rs#L111). Would these be user facing? not very familiar with the project yet.

laggui:
These changes are also not user facing 🙂

Now that I went back and read the linked issue, it seems that the specific example provided (for the `Config` trait methods) has already been addressed in #699.

And even more changes have been made in #1919. I think the issue was mistakenly not closed since it was not directly linked as fixed in these PRs.

ivnsch:
Alright, closing this then!

Issue #805, opened 2023-09-15
Autotune Conv2d


agelas:
### Feature description
This is a more specific feature related to Issue #617. After playing around a bit with the shader for conv2d, I think it might be worth reworking the kernel to allow for autotuning to adjust the workgroup size and allow for loop unrolling of the innermost loop. More details below, but I think we can achieve greater utilization and saturation of host GPU's with some relatively straightforward adjustments.

### Feature motivation
This can help with speeding up models and more efficient resource utilization. As you can see with some rudimentary testing below, autotuning can usually find a (barely) faster pipeline*. What I'd like to draw attention to though is the variance. Autotuning seems able to identify much more _consistent_ compute pipelines. This can allow for better predictability and stable resource allocation.

\* The computer I did this on is a bit of a potato right now. I'd be curious to see if someone can see similar trends with larger batch sizes, higher resolution images, and/or more channels in the input tensor. Also using a graphics API besides Vulkan. 

#### Notes on the data
1. The only GPU I had readily available lying around is an old Nvidia GT 1030. All results below were run using Vulkan, the Dx12 drivers have some weird invocation limit that I'm having trouble getting around.
2. This data is very much cherry-picked. There were some runs where the variance of autotuning was higher/min time of the benchmark run was higher. But the overall trend aligns with what's shown below.
3. All benchmarks were run with no bias tensor and `ConvOptions {stride: [1,1], padding: [1,1], dilation: [1,1], groups: 1 };`. The bulk of this data is centered on input tensors of shape [32, 3, 224, 224], which should be a fairly reasonable input to something like ImageNet, VGG, ResNet-50.

| Input Tensor Shape| Type + Bench Run| Samples | Mean     | Variance | Median  | Min      | Max      |
|------------------------|-----------|---------|----------|----------|---------|----------|----------|
| [8, 3, 224, 224]        | Autotune 1 | 5       | 489.791ms| 80.587µs | 492.758ms| 478.120ms | 501.855ms |
| [8, 3, 224, 224]        | Default  1 | 5       | 504.855ms| 281.282µs| 511.095ms| 481.818ms | 522.776ms |
| [10, 3, 224, 224]      | Autotune 2 | 5       | 619.695ms| 192.013µs| 623.878ms| 593.247ms| 633.878ms|
| [10, 3, 224, 224]      | Default 2   | 5       | 614.064ms| 215.598µs| 618.749ms| 594.727ms| 630.743ms|
[...]

PR #2287, opened 2024-09-17
Introduce autotuning to `conv2d` and `conv_transpose2d` with a new `im2col`/`GEMM` algorithm

wingertge:
### Checklist

### Related Issues/PRs

This goes some way to resolving this old enhancement suggestion: #805 

### Changes

Adds the required infrastructure to autotune `conv2d` and `conv_transpose2d`, as well as adding a second algorithm based on `im2col` which provides significant speedups at the cost of memory usage. I'd like to add more algorithms when I have time, but this will already put in the infrastructure to make that much easier and less breaking.

#### Update; Now includes implicit GEMM when:
- CMMA is available
- `batch_size * out_h * out_w` is divisible by 16
- `out_channels` is divisible by 16
- `in_channels * kernel_h * kernel_w` is divisible by 16

### Testing
The autotuned, direct (current) and `im2col` implementations pass all existing tests (except matches_reference_backend, see below).

#### Direct
`batch_size = 4`
| Benchmark | Feature     | Backend             | Device                  | Median         |
|-----------|-------------|---------------------|-------------------------|----------------|
| conv2d    | wgpu        | `jit<wgpu>`         | BestAvailable           | 8.688ms        |
| conv2d    | wgpu-fusion | `fusion<jit<wgpu>>` | BestAvailable           | 8.351ms        |
| conv2d    | cuda-jit    | `jit<cuda>`         | CudaDevice { index: 0 } | 6.250ms        |

`batch_size = 16`
| Benchmark | Feature     | Backend             | Device                  | Median         |
|-----------|-------------|---------------------|-------------------------|----------------|
| conv2d    | wgpu        | `jit<wgpu>`         | BestAvailable           | 31.056ms       |
| conv2d    | wgpu-fusion | `fusion<jit<wgpu>>` | BestAvailable           | 32.943ms       |
| conv2d    | cuda-jit    | `jit<cuda>`         | CudaDevice { index: 0 } | 26.165ms       |

#### Im2col
`batch_size = 4`
| Benchmark | Feature     | Backend             | Device                  | Median         |
|-----------|-------------|---------------------|-------------------------|----------------|
| conv2d    | wgpu        | `jit<wgpu>`         | BestAvailable           | 4.831ms        |
| conv2d    | wgpu-fusion | `fusion<jit<wgpu>>` | BestAvailable           | 4.759ms        |
| conv2d    | cuda-jit    | `jit<cuda>`         | CudaDevice { index: 0 } | 2.617ms        |

`batch_size = 16 (split into 4 sub-batches)`
| Benchmark | Feature     | Backend             | Device                  | Median         |
|-----------|-------------|---------------------|-------------------------|----------------|
| conv2d    | wgpu        | `jit<wgpu>`         | BestAvailable           | 21.284ms       |
| conv2d    | wgpu-fusion | `fusion<jit<wgpu>>` | BestAvailable           | 21.207ms       |
| conv2d    | cuda-jit    | `jit<cuda>`         | CudaDevice { index: 0 } | 17.202ms       |

#### Implicit GEMM
`batch_size = 16`
| Benchmark | Feature  | Backend     | Device                  | Median         |
|-----------|----------|-------------|-------------------------|----------------|
| conv2d    | cuda-jit | `jit<cuda>` | CudaDevice { index: 0 } | 4.859ms        |


nathanielsimard (commented):
LGTM, waiting for @louisfd to review before merging, but thanks a lot 🙏

crates/burn-jit/src/kernel/conv/conv2d/direct.rs
```diff
@@ -50,7 +49,7 @@ fn conv2d_kernel<F: Float>(
     let oh = ABSOLUTE_POS / output.stride(2) % output.shape(2);
     let ow = ABSOLUTE_POS / output.stride(3) % output.shape(3);
 
-    let g = (weight.shape(0) + oc) % args.groups;
+    let g = oc / args.channels_per_group;
```
louisfd, line 52: Is this the fix for the bug you mentioned about not having the exact same result as torch?

crates/burn-jit/src/kernel/conv/conv2d/im2col.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    for kernel_y in 0..args.kernel_h {
+        #[unroll(unroll_w)]
+        for kernel_x in 0..kernel_w {
+            let kernel_pos = kernel_y * kernel_w + kernel_x;
+            let col_pos = col_idx + kernel_pos * args.col_size_1;
+
+            let y = (out_y * args.stride_h + kernel_y * args.dilation_h) as i32 - args.padding_h;
+            let x = (out_x * args.stride_w + kernel_x * args.dilation_w) as i32 - args.padding_w;
+            if y >= 0 && x >= 0 && y < height as i32 && x < width as i32 {
```
louisfd, line 75: Are there configurations where we are guaranteed to always be within bounds? Is so we could have a comptime strategy to remove the if when not necessary.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
```diff
@@ -0,0 +1,343 @@
...
+};
+use half::f16;
+
+use crate::{
+    ops::{numeric::empty_device, permute, reshape},
+    tensor::JitTensor,
+    FloatElement, IntElement, JitBackend, JitRuntime,
+};
+
+const WARP_SIZE: u32 = 32;
```
louisfd, line 20: I think it's fine to hardcode these values for now, but they should be passed to the cube kernel via a comptime struct, because the day they will change (for amd it's 64 for instance) we won't have to change the kernels, just the launch. Same for other consts below.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
```diff
@@ -0,0 +1,343 @@
...
+const CMMA_M: u32 = 16;
+const CMMA_N: u32 = 16;
+const CMMA_K: u32 = 16;
+const CMMA_INPUT_TILE_SIZE: u32 = CMMA_M * CMMA_K;
+const CMMA_FILTER_TILE_SIZE: u32 = CMMA_K * CMMA_N;
+const CMMA_OUT_TILE_SIZE: u32 = CMMA_M * CMMA_N;
+
+const WARPS_PER_CUBE: u32 = 8;
+
+const CUBE_DIM_X: u32 = 128;
```
louisfd, line 31: I'm genuinely curious, why did you choose (128, 2) and not (32, 8) for instance, is it better?

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
```diff
@@ -0,0 +1,343 @@
...
+    out: &mut Tensor<F>,
+    gemm_m: u32,
+    gemm_n: u32,
+    gemm_k: u32,
+    slice_size: u32,
+    args: &KernelArgs,
+    #[comptime] kernel_h: u32,
+    #[comptime] kernel_w: u32,
+) {
+    let kernel_size = comptime! { kernel_h * kernel_w };
```
louisfd, line 157: Doesn't arithmetic on two comptime values always give a comptime? Why the need for a comptime macro?

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
```diff
@@ -0,0 +1,343 @@
...
+            let out_x = ((abs_slice_row % batch_stride) % y_stride) / x_stride;
+
+            let channel = my_slice_idx / kernel_size;
+
+            let kernel_y = (my_slice_idx / kernel_w) % kernel_h;
+            let kernel_x = my_slice_idx % kernel_w;
+            let y = (out_y * args.stride_h + kernel_y * args.dilation_h) as i32 - args.pad_h;
+            let x = (out_x * args.stride_w + kernel_x * args.dilation_w) as i32 - args.pad_w;
+
+            if x >= 0 && x < w && y >= 0 && y < h {
```
louisfd, line 256: We will probably want a comptime decision to whether or not have to do this if.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
```diff
@@ -0,0 +1,343 @@
...
+            weight_tile[n] = f16::cast_from(
+                weight[out_channel * weight.stride(0)
+                    + in_channel * weight.stride(1)
+                    + kernel_y * weight.stride(2)
+                    + kernel_x * weight.stride(3)],
+            );
+        }
+
+        /**************************** Bounds Check + CMMA Op*********************************/
+        if a_row < gemm_m && k < gemm_k && b_col < gemm_n {
```
louisfd, line 291: We will probably want a comptime decision to whether or not have to do this if.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
```diff
@@ -0,0 +1,343 @@
...
+            cmma::load(&matrix_a, input_tile.as_slice(), CMMA_K);
```
louisfd, line 292: When doing the matmul, i found you can safely reuse a fragment for several executions. So this line could go in the outer loop.

crates/burn-jit/src/kernel/conv/conv2d/col2im.rs
```diff
@@ -0,0 +1,289 @@
...
+        let ones = ones_device(
+            client.clone(),
+            device.clone(),
+            Shape::new([1, batch_size * im_h * im_w]),
+        );
+        let bias = reshape(bias, Shape::new([im_channels, 1]));
+        let bias = JitBackend::<R, E, I>::float_matmul(bias, ones);
+        let bias = reshape(bias, Shape::new([im_channels, batch_size, im_h, im_w]));
+        let bias = swap_dims(bias, 0, 1);
+        image = JitBackend::<R, E, I>::float_add(image, bias);
```
louisfd, line 134: Why not start the convolution with bias as the input?

louisfd (changes requested):
I have a few questions, but overall it's very good, awesome!

crates/burn-jit/src/kernel/conv/conv2d/base.rs
```diff
@@ -0,0 +1,125 @@
...
+    conv2d_direct, conv2d_im2col, conv_transpose2d_autotune, conv_transpose2d_col2im,
+    conv_transpose2d_direct, implicit_gemm::conv2d_implicit_gemm,
+};
+
+/// The strategy to be used when launching a convolution kernel.
+pub enum Conv2dStrategy {
+    /// A simple direct convolution.
+    Direct,
+    #[cfg(feature = "autotune")]
+    /// Using autotune to chose the best kernel based on runtime information.
```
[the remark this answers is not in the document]
wingertge: Fun fact: that typo is also on `MatmulStrategy`, which is how it got into my code (I copy pasted). I'll fix it in both.

crates/burn-jit/src/kernel/conv/conv2d/direct.rs
wingertge: Exactly.

crates/burn-jit/src/kernel/conv/conv2d/im2col.rs
wingertge: Since we're iterating over out dimensions it should always be in bounds as long as padding == 0. I'll add a comptime path for that.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
wingertge: Yeah I wanted to read these constants from `cubecl` ideally. I guess I'll just put them in a struct for now then so it's easier to change later.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
wingertge: That was left over from a refactor, I'll just remove it.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
wingertge: It makes sense when you look at the dimensionality of the matrix. The `x` dimension is `batch_size * out_h * out_w` which is much larger than the `y` dimension, which is `in_channels * kernel_h * kernel_w`. `128 x 2` is a decent approximation of the relative sizes between `M` and `N`. This lopsidedness caused issues before with the `im2col` because the matmul kernel would pick something like `144k x 1` as its cube count because it uses a square `CubeDim` but the actual matrix is anything but square.

crates/burn-jit/src/kernel/conv/conv2d/implicit_gemm.rs
wingertge: I'm not sure what you mean? I am reusing the fragments, but I still need to load a tile and execute the matmul for each `k`.

crates/burn-jit/src/kernel/conv/conv2d/col2im.rs
wingertge: Because this is faster. The columns matrix contains a lot of duplicates, so you would end up adding the bias several times for each pixel. Adding it at the end takes far fewer operations.

louisfd (approved):
LGTM

PR #2288, opened 2024-09-19
doc: improve doc for burn-tch

kingwingfly:
### Checklist

### Related Issues/PRs

#2261
(That PR is a misunderstanding of a issue, however, its change to `burn-tch/README.md` may be adaptable. )

### Changes

- make the setup step for macOS MPS more correct
- add some hint and tips

### Testing

Only doc changes.

### DIFF
--- a/crates/burn-tch/README.md
+++ b/crates/burn-tch/README.md
@@ -261,14 +261,23 @@ For a more complete example using the `tch` backend, take a loot at the
 
 ## Too many environment variables?
 
-Try `.cargo/config.toml` ([cargo book](https://doc.rust-lang.org/cargo/reference/config.html#env)):
+Try `.cargo/config.toml` ([cargo book](https://doc.rust-lang.org/cargo/reference/config.html#env)).
 
-For example, Linux.
+Instead of setting the environments in your shell, you can manually add them to your `.cargo/config.toml`:
+
+```toml
+[env]
+LD_LIBRARY_PATH = "/absolute/path/to/libtorch/lib"
+LIBTORCH = "/absolute/path/to/libtorch/libtorch"
+```
+
+Or use bash commands below:
+
 ```bash
 mkdir .cargo
 cat <<EOF > .cargo/config.toml
 [env]
-LD_LIBRARY_PATH = "$(pwd)/libtorch/lib:$LD_LIBRARY_PATH"
-LIBTORCH = "$(pwd)/libtorch"
+LD_LIBRARY_PATH = "/absolute/path/to/libtorch/lib:$LD_LIBRARY_PATH"
+LIBTORCH = "/absolute/path/to/libtorch/libtorch"
 EOF
 ```
--- a/crates/burn-tch/README.md
+++ b/crates/burn-tch/README.md
@@ -281,3 +281,4 @@ LD_LIBRARY_PATH = "/absolute/path/to/libtorch/lib:$LD_LIBRARY_PATH"
 LIBTORCH = "/absolute/path/to/libtorch/libtorch"
 EOF
 ```
+This will automatically include the old `LD_LIBRARY_PATH` value in the new one.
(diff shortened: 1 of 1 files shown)


crates/burn-tch/README.md
```diff
@@ -243,15 +243,32 @@ is to use a PyTorch installation. This requires a Python installation.
 _Note: MPS acceleration is available on MacOS 12.3+._
 
 ```shell
-pip install torch==2.2.0
+pip install torch==2.2.0 numpy==1.26.4 setuptools
```
laggui, line 246: Why the added explicit dependencies for numpy and setuptools? 🤔

crates/burn-tch/README.md
```diff
@@ -243,15 +243,32 @@ is to use a PyTorch installation. This requires a Python installation.
...
+
+For example, Linux.
+```bash
+mkdir .cargo
+cat <<EOF > .cargo/config.toml
+[env]
+LD_LIBRARY_PATH = "$(pwd)/libtorch/lib:$LD_LIBRARY_PATH"
+LIBTORCH = "$(pwd)/libtorch"
+EOF
+```
```
laggui, line 274: Instead of giving commands to write this to the file, I think I would put a simple sentence to explain the steps (and use absolute path instead of `$(pwd)` which may not always be the case for everyone).

Something like this maybe?

## Too many environment variables?

Instead of setting the environments in your shell, you can add them to your `.cargo/config.toml`

```
[env]
LIBTORCH = /absolute/path/to/libtorch/
LD_LIBRARY_PATH = /absolute/path/to/libtorch/lib:$LD_LIBRARY_PATH
```

laggui (commented):
Thanks for adding a bit more context, the `config.toml` part should be useful to others 🙂 

Just have some comments otherwise LGTM!

crates/burn-tch/README.md
kingwingfly: When I preparing the environment, I just `pip install torch==2.2.0`. When `cargo run`, an error raised saying `need numpy`, then I installed `numpy`, it then said need `numpy<2`, then I reinstalled the newest 1.x numpy. It said need `setuptools`. After that, it compiled.

I work on macOS and `venv`, and `pip`(24.2) will not install the dependencies automatically.

crates/burn-tch/README.md
kingwingfly: > LD_LIBRARY_PATH = /absolute/path/to/libtorch/lib:$LD_LIBRARY_PATH

In `config.toml`, `..libtorch$LD_LIBRARY_PATH` may not work and the value should be wrapped with `"`. I'll test and modify.

kingwingfly: I tested, and that's true the `$LD_LIBRARY_PATH` will be keep as `$LD_LIBRARY_PATH` in env vars. 
```
[tagger/src/lib.rs:19:9] ld = "/run/media/louis/ZhiTai/rust/burnxp/libtorch:$PATH"
```
---

```toml
[env]
LD_LIBRARY_PATH = "/absolute/path/to/libtorch/lib"
LIBTORCH = "/absolute/path/to/libtorch/libtorch"
```

Maybe this is Okey? As on my machine, `LD_LIBRARY_PATH` is actually empty. 

Or recommending a bash script like `cat<<EOF > file ... EOF` is also okey. I see `kubernetes`' documentation uses such method too.

laggui: Ahhh yeah you're right.. I definitely didn't think this one through before 😅 

In that case, a combination of both might work best. Leave the instructions with the dummy example `/absolute/path/to/libtorch` and under we can give the instructions to prepend it to the variable while keeping your existing var!

crates/burn-tch/README.md
laggui: Huh, that's weird. Dependencies are usually automatically installed..

I don't have a mac to test so I believe you. Let's leave the additional dependencies in the instructions in this case 👍

laggui (approved):
Thanks for improving the docs 😄

PR #2290, opened 2024-09-20
Add candle `CudaDevice` and `MetalDevice` to avoid creating a new unique device each time

laggui:
### Checklist

### Related Issues/PRs

Flagged by a discord user: https://discord.com/channels/1038839012602941528/1091796857996451942/1286409544280576110

The candle device handling is clumsy. Turns out, every time you create a new device struct for the same device index, a new unique identifier is created - meaning that both structs (to represent the same device) are not equal.

The original issue flagged a problem for metal device due to this inequality, but in my tests the behavior leads to an insane training time with CUDA (ETA > 2hrs).

This is because the `tensor.to_device(device)` method in Candle actually transfers the data via the CPU for CUDA devices that are different. A terrible side effect of using different unique identifiers to represent the same device index.

### Changes

The `CandleDevice` enum now captures the underlying device structs to avoid creating a new one with a unique identifier each time.

To keep the API simple (and similar to the previous usage), I've added creation methods for `Cuda` and `Metal` variants:
```rust
// Create a Cuda device from its index
let device = CandleDevice::cuda(0);
// Create a Metal device from its index
let device = CandleDevice::metal(0);
```

Also had to remove the `Copy` implementation because the candle devices do not implement it.

### Testing

Tested (locally) the `mnist` example with a new candle feature flag to use the candle backend. The training now completes in approx. 3-4 minutes.


nathanielsimard (approved):

ivnsch:
Discord user here - this fixed https://github.com/huggingface/candle/issues/2484

PR #2293, opened 2024-09-21
Add precision classification metric

tsanona:
### Checklist

### Related Issues/PRs

#544

### Changes

Hello, so I wanted to add precision to the metrics and ended up realizing that implementing a confusion matrix would not only help with precision but also  other classification metrics. So I implemented it also along with some code to deal with classification averaging (micro, macro) and thresholds .  
I also sneaked in some code to generate some dummy data for testing the metrics. I'm not sure it is decent enough but open for suggestions.
I do realize that it is a long PR, if necessary I can split it into different ones.
I'm new to rust dev and come from python so lemme know if I need to adjust some patterns.

### Testing

Like said above I added some dummy classification data generator and tested confusion matrix and precision metric's methods with it.


crates/burn-train/src/metric/confusion_matrix.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub false_positive: Tensor<B, 2, Bool>,
+    pub true_negative: Tensor<B, 2, Bool>,
+    pub false_negative: Tensor<B, 2, Bool>,
+}
+
+impl<B: Backend> ConfusionMatrix<B> {
+    pub fn from(predictions: Tensor<B, 2, Bool>, targets: Tensor<B, 2, Bool>) -> Self {
+        let stats_tensor = predictions.int() + targets.int() * 2;
+        Self {
+            true_positive: stats_tensor.clone().equal_elem(3),
```
oojo12: hey there, I hope everyone doesn't mind me commenting. I am newish to Rust trying to learn and was also working on this so thought it'd be a good learning opportunity for me.

Wouldn't this only work in the binary classification and not multi classification case? I see that in `crates/burn-train/src/metric/precision.rs` line 76 you're using `.greater_elem` and from the code it looks like that just returns a boolean true/false -> 1/0. Given that it doesn't appear that it is possible for the prediction to ever be > 1. Or am I misunderstanding?

crates/burn-train/src/metric/precision.rs
```diff
@@ -0,0 +1,166 @@
...
+                    [0, 1, 0],
+                ],
+                &device),
+        );
+
+        let _entry = metric.update(&input, &MetricMetadata::fake());
+        assert_eq!(75.0, metric.value());
+    }
+
+    /*#[test]
```
oojo12, line 116: looks like you forgot to remove the commented code :)

tsanona: Ah, thank you :star2:

antimora:
+1 confusion

laggui:
Ha, I was just discussing this with another user on discord last week 😄 

Thanks for the PR, I'll try to look into it today!

crates/burn-train/src/lib.rs
```diff
@@ -25,3 +25,156 @@ pub use learner::*;
...
+        let predictions = changed_targets
+            .random_like(Distribution::Uniform(0.0, THRESHOLD - 0.1))
+            .sub(changed_targets.clone())
+            .abs();
+
+        (
+            ClassificationInput::new(predictions, targets.clone()),
+            targets.float().sub(changed_targets),
+        )
+    }
```
laggui, line 146: Reusing a mock input for multiple tests is a great idea, but 1) it should be split for different use cases tested (e.g., binary, multiclass and multilabel) and 2) it doesn't need to be so complicated to generated a mock input 😅 

I think using a dummy input with a smaller number of samples and hardcoded predictions / target errors makes it easier for everyone. That way, you also don't have to write tests for the functions you added to reuse come code in your tests 😁 (as with the added `test_predictions_targets_match` and `test_error_rate`).

crates/burn-train/src/lib.rs
```diff
@@ -25,3 +25,156 @@ pub use learner::*;
...
+
+    use burn_core::tensor::cast::ToElement;
+    use burn_core::tensor::TensorData;
+    use strum::IntoEnumIterator;
+
+    #[test]
+    fn test_predictions_targets_match() {
```
laggui, line 153: By simplifying the test inputs we should be able to remove this test (whose sole purpose is to test the reused code).

crates/burn-train/src/lib.rs
```diff
@@ -25,3 +25,156 @@ pub use learner::*;
...
+            TensorData::assert_eq(
+                &(target_diff.clone() + thresholded_prediction.float()).to_data(),
+                &input.targets.clone().float().to_data(),
+                true,
+            );
+        }
+    }
+
+    #[test]
+    fn test_error_rate() {
```
laggui, line 166: Same here.

By simplifying the test inputs we should be able to remove this test (whose sole purpose is to test the reused code)

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
 
+/// Classification Metric trait
+///
+/// Requires implementation [Metric](Metric)<Input = ClassificationInput<B>>
+pub trait ClassificationMetric<B: Backend>: Metric<Input = ClassificationInput<B>> {
+    /// Sets threshold. Default 0.5
+    fn with_threshold(self, threshold: f64) -> Self;
+    /// Sets average type. Default Micro
+    fn with_average(self, average: ClassificationAverage) -> Self;
+}
```
laggui, line 70: Hmmm I don't think a new trait is necessary for that. I'd remove it from the current implementation.

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
+
+/// The [classification metric](ClassificationMetric) input type.
+#[derive(new, Debug)]
+pub struct ClassificationInput<B: Backend> {
+    /// Sample x Class Non thresholded predictions
+    pub predictions: Tensor<B, 2>,
+    /// Sample x Class target mask
```
laggui, line 77: This is a one-hot encoded target, right? Maybe just change the description to make it clear.

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
+///Aggregation types for Classification metric average
+#[derive(EnumIter, Copy, Clone, Debug)]
+pub enum ClassificationAverage {
+    /// overall aggregation
+    Micro,
+    /// over class aggregation
+    Macro,
+    // /// over class aggregation, weighted average
+    //Weighted(Box<[f64]>), todo!()
+}
```
laggui, line 90: I would move the struct and enum to the `precision.rs` module for now. These shouldn't be in the base module as it doesn't apply to all metrics.

And since it's only used for precision at this time it can go there until more metrics are added with a similar pattern.

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
+    /// Sample x Class Non thresholded predictions
+    pub predictions: Tensor<B, 2>,
+    /// Sample x Class target mask
+    pub targets: Tensor<B, 2, Bool>,
+}
+
+///Aggregation types for Classification metric average
+#[derive(EnumIter, Copy, Clone, Debug)]
+pub enum ClassificationAverage {
+    /// overall aggregation
```
laggui, line 84: Let's be a bit more descriptive. The micro average should say something like `Aggregates the statistics of all classes before averaging.`

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
+    Micro,
+    /// over class aggregation
```
laggui, line 86: A bit more descriptive here too. Suggestion: `Computes the statistics independently for each class before averaging.`

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
+    Macro,
+    // /// over class aggregation, weighted average
+    //Weighted(Box<[f64]>), todo!()
```
laggui, line 89: We can remove the dead code/comment if this is not implemented.

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,84 @@ pub trait Metric: Send + Sync {
...
+        }
+    }
+
+    ///convert to averaged metric, returns float
+    pub fn to_averaged_metric<B: Backend>(self, aggregated_metric: Tensor<B, 1>) -> f64 {
+        self.to_averaged_tensor(aggregated_metric)
+            .into_scalar()
+            .to_f64()
+    }
+}
```
laggui, line 138: Speaking strictly of average types for classification metrics, I think these methods are specific to each metric. Instead of implementing methods for `ClassificationAverage`, the average reduction code should probably be in the metric computation itself.

With this change, `ClassificationAverage` will be a simple enum to represent the average type (without any methods).

crates/burn-train/src/metric/confusion_matrix.rs
```diff
@@ -0,0 +1,161 @@
+use super::{ClassificationAverage, ClassificationInput};
+use burn_core::prelude::{Backend, Int, Tensor};
+use std::fmt::{self, Debug};
+
+#[derive(Clone)]
+pub struct ConfusionMatrix<B: Backend> {
+    confusion_matrix_map: Tensor<B, 2, Int>,
+    classification_average: ClassificationAverage,
+}
```
laggui, line 9: This is not actually a confusion matrix?

Looking at the implementation, it looks like an interface to compute the true positives, false positives, true negatives and false negatives statistics. But it does not actually represent a confusion matrix with the predicted vs expected values.

For that reason, I would rename the struct (and its `confusion_matrix_map` field). Since it is related to type I and type II errors, maybe the name should be related to that (e.g., `StatsError`).

It might even be that it could be simply a function to compute what's required 🤔 but for now let's start with that.

crates/burn-train/src/metric/confusion_matrix.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+impl<B: Backend> ConfusionMatrix<B> {
+    pub fn new(
+        input: &ClassificationInput<B>,
+        threshold: f64,
+        classification_average: ClassificationAverage,
+    ) -> Self {
+        let thresholded_predictions = input.predictions.clone().greater_elem(threshold);
```
laggui, line 37: This assumes the input has been transformed to the `[0, 1]` range (e.g., post sigmoid activation for the binary case).

We should either add the option to apply a sigmoid before, or/and document the usage to state that outputs are expected to be in that range.

laggui (changes requested):
Ok, just took some time to review! Thanks again for contributing 🙏 

Overall, the numerical implementation looks correct! 🙂 

But I have some comments regarding the implementation.

Also, I would refactor the tests to split them into smaller individual units. I see that you wanted to reuse as much code, which is great, but in this case it also makes the tests complicated.

Each functionality should be isolated in the tests. It makes it easier to 1) understand what is going on (and what is being tested) and 2) pin point what test failed in the event of a regression.

With these changes, the `ClassificationType` enum iter can probably be eliminated so the binary, multiclass and multilabel cases are separated in different test cases.

crates/burn-train/src/lib.rs
tsanona: Yeah that's fair enough, I'll reduce this complicated mess to some static examples for binary, multi class and multi label examples. Thank you

laggui: You can use the function to generate an input if you want since it has been validated. Just copy the values of the tensor so it's fixed and then the function can go 😄 (you can probably reduce the number of samples)

crates/burn-train/src/lib.rs
tsanona: changed

crates/burn-train/src/lib.rs
tsanona: changed

crates/burn-train/src/metric/base.rs
tsanona: Removed!

crates/burn-train/src/metric/base.rs
tsanona: changed!

crates/burn-train/src/metric/base.rs
tsanona: changed

crates/burn-train/src/metric/base.rs
tsanona: changed, thank you

crates/burn-train/src/metric/base.rs
tsanona: Ouki, removed for now

crates/burn-train/src/metric/base.rs
tsanona: Hum, I'm not sure I understand, if I move these methods to the metric itself, won't it be necessary to implement it for every new metric that uses a confusion matrix in the background?

crates/burn-train/src/metric/confusion_matrix.rs
tsanona: I see your point, I called it confusion matrix because it was more intuitive of a name. I propose then ConfusionStats, I don't love Stats Error because it also includes statistics of correct values. What do you think?

crates/burn-train/src/metric/confusion_matrix.rs
tsanona: I added usage documentation.

tsanona:
Hey sorry it took me some time, I made some changes and continued the discussion on some of your comments. Lemme know what you think :)

laggui:
> Hey sorry it took me some time, I made some changes and continued the discussion on some of your comments. Lemme know what you think :)

No worries 🙂 Thanks for addressing my comments, I'll take a look at the changes today!

laggui (changes requested):
Great stuff! Especially with the tests, much easier to understand what is being tested now 😄 

I have some minor comments, but once these are addressed I think we should be good to merge!

crates/burn-train/src/lib.rs
tsanona: all done

crates/burn-train/src/metric/base.rs
tsanona: moved!

crates/burn-train/src/metric/base.rs
tsanona: Fair enough, I've moved the `aggregate` and `average` functions` to the implementation of `ConfusionStats`.

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,77 @@ pub trait Metric: Send + Sync {
...
+
+impl<B: Backend> From<ClassificationInput<B>> for (Tensor<B, 2>, Tensor<B, 2, Bool>) {
+    fn from(val: ClassificationInput<B>) -> Self {
+        (val.predictions, val.targets)
+    }
+}
+
+/// Class Averaging types for Classification metrics.
+#[derive(EnumIter, Copy, Clone, Debug)]
```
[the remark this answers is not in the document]
tsanona: removed

crates/burn-train/src/metric/base.rs
```diff
@@ -56,6 +59,77 @@ pub trait Metric: Send + Sync {
...
+pub enum ClassAverageType {
+    ///Computes the statistics over all classes before averaging
+    Micro,
+    ///Computes the statistics independently for each class before averaging
+    Macro,
+}
+
+impl ClassAverageType {
```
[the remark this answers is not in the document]
tsanona: moved

crates/burn-train/Cargo.toml
```diff
@@ -36,9 +36,13 @@ crossterm = { workspace = true, optional = true }
 # Utilities
 derive-new = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"] }
+strum = { workspace = true }
+yare = "3.0.0"
```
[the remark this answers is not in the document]
tsanona: changed to rstets

crates/burn-train/src/metric/confusion_stats.rs
```diff
@@ -0,0 +1,257 @@
...
+    use burn_core::prelude::{Tensor, TensorData};
+    use yare::parameterized;
+
+    #[parameterized(
+    binary_micro = {Binary, Micro, [1].into()},
+    binary_macro = {Binary, Macro, [1].into()},
+    multiclass_micro = {Multiclass, Micro, [3].into()},
+    multiclass_macro = {Multiclass, Macro, [1, 1, 1].into()},
+    multilabel_micro = {Multilabel, Micro, [5].into()},
+    multilabel_macro = {Multilabel, Macro, [2, 2, 1].into()})]
```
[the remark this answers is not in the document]
tsanona: ah thanks, chnaged

crates/burn-train/Cargo.toml
```diff
@@ -36,9 +36,13 @@ crossterm = { workspace = true, optional = true }
...
+
 
 [dev-dependencies]
 burn-ndarray = { path = "../burn-ndarray", version = "0.15.0" }
+approx = "0.5.1"
```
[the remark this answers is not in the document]
tsanona: removed

tsanona:
Hey sorry I've added an extra change that I realized was missing. With just thresholds one couldn't use the metric correctly for multiclass thus I added top_k as an mutually exclusive option so it makes sense.

laggui:
No problem, will try to make time to review this today or tomorrow!

crates/burn-train/src/metric/precision.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[case::multilabel_macro(Multilabel, Macro, (2.0/3.0 + 2.0/3.0 + 0.5)/3.0)]
+    fn test_precision(
+        #[case] class_type: ClassificationType,
+        #[case] avg_type: ClassAverageType,
+        #[case] expected: f64,
+    ) {
+        let input = dummy_classification_input(&class_type);
+        let mut metric = PrecisionMetric::<TestBackend>::default();
+        metric = match class_type {
+            Multiclass => metric.with_top_k(1),
```
laggui, line 122: We should probably add a test for `top_k` > 1 to make sure everything is covered 🙂

laggui (changes requested):
Thanks for addressing the changes!

Just a couple of comments regarding some stuff that has moved or been added in the latest commits, but we're almost there 😄

crates/burn-train/src/metric/confusion_stats.rs
```diff
@@ -0,0 +1,328 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let to_vec = |tensor_data: Tensor<B, 1>| {
+            tensor_data
+                .to_data()
+                .to_vec::<f32>()
+                .expect("A vector representation of the input Tensor is expected")
+        };
+        let ratio_of_support_vec =
+            |metric: Tensor<B, 1>| to_vec(self.clone().ratio_of_support(metric));
+        f.debug_struct("ConfusionMatrix")
```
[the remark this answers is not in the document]
tsanona: Hu nicely spotted, thanks!

crates/burn-train/src/metric/classification.rs
```diff
@@ -0,0 +1,26 @@
+use burn_core::prelude::{Backend, Bool, Tensor};
+
+/// Input for classification tasks.
+#[derive(new, Debug, Clone)]
+pub struct ClassificationInput<B: Backend> {
+    /// Sample x Class Non thresholded normalized predictions.
+    pub predictions: Tensor<B, 2>,
+    /// Sample x Class one-hot encoded target.
+    pub targets: Tensor<B, 2, Bool>,
```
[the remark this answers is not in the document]
tsanona: That's true, I've been looking at the existing code in burn-trai/src/learner/classification and I think it would be easier to just use the adaptor to convert between bin/multiclass/multilable outputs to the general one hot encoded metrics, non?

crates/burn-train/src/metric/classification.rs
```diff
@@ -0,0 +1,26 @@
...
+impl<B: Backend> From<ClassificationInput<B>> for (Tensor<B, 2>, Tensor<B, 2, Bool>) {
+    fn from(val: ClassificationInput<B>) -> Self {
+        (val.predictions, val.targets)
+    }
+}
+
+/// Class Averaging types for Classification metrics.
+#[derive(Copy, Clone)]
+#[allow(dead_code)]
+pub enum ClassAverageType {
```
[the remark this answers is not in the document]
tsanona: Oh I just think it fits better and is more descriptive of what it is

crates/burn-train/src/metric/confusion_stats.rs
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    pub fn precision(self) -> f64 {
+        let class_average = self.class_average;
+        Self::average(
+            self.clone().true_positive() / self.predicted_positive(),
+            class_average,
+        )
+    }
+}
```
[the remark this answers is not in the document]
tsanona: Yeah fair enough, I moved it here because it was easier, but I'll try to find a way to do it neatly within the `PrecisionMetric`

crates/burn-train/src/metric/precision.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Creates a new metric instance with default values.
+    fn default() -> Self {
+        Self {
+            state: NumericMetricState::default(),
+            _b: PhantomData,
+            threshold: Some(0.5),
+            class_average: ClassAverageType::Micro,
+            top_k: None,
+        }
+    }
```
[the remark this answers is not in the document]
tsanona: Yeah I was avoiding doing that because then if `PrecisionMetric` is initialized as `binary` for example and then `threshold is set if becomes a multiclass/multilable metric and can become confusing in my opinion. We could maybe create the enum `ClassificationType` which could be used to limit  whether we could set threshold or top_k, what do you think?

crates/burn-train/src/metric/precision.rs
tsanona: you're right, added

crates/burn-train/src/metric/classification.rs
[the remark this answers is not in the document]
laggui: Fair enough 🙂 this is a bit subjective and I'm not against the change, still descriptive enough

How do you feel about `ClassReduction`? Just throwing it out there as it came to mind.

crates/burn-train/src/metric/precision.rs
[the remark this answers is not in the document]
laggui: Hmmm yeah that could be annoying.

Maybe a more explicit change could be to have the user-facing metrics to be defined for each type, e.g. `BinaryPrecision`, `MulticlassPrecision` and `MultilabelPrecision`?

Internally it could call common code, but at least we know that the configurations should be valid.

Lmk what you think 🙂

crates/burn-train/src/metric/classification.rs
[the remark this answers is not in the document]
tsanona: I like it, changed it ~

crates/burn-train/src/metric/precision.rs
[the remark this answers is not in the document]
tsanona: Right, that's also how torchmetrics hadles it. I think I agree, I'll make some changes :rocket:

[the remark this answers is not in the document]
laggui: Ahh I wasn't aware of that 😄 

Thanks for addressing all of my comments btw 💪

crates/burn-train/src/metric/classification.rs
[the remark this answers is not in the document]
laggui: Yeah my comment was not meant to say that this is where is should be handled specifically 😄

laggui:
Fyi @antimora not all changes have been addressed so further review will be pending until then.

The metric should handle the different configurations for binary, multiclass and multi-label (previously suggested separate structs `BinaryPrecision`, `MulticlassPrecision` and `MultilabelPrecision`).

tsanona:
Hey hey, sorry it took me some time, I was trying to find a way to tie the metric with the type of Output chosen when implementing the TrainStep for the Model but I wasn't successful so I just went with what I pushed now. I also found an error on the one_hot function that we pushed before. Lemme know what you think :)

crates/burn-tensor/src/tensor/api/int.rs
```diff
@@ -52,10 +52,14 @@ where
     /// ```
     pub fn one_hot(self, num_classes: usize) -> Tensor<B, 2, Int> {
         check!(TensorCheck::one_hot_tensor(self.clone(), num_classes));
-        let [num_samples] = self.dims();
-        let indices = self.unsqueeze();
-        let values = indices.ones_like();
-        Tensor::zeros([num_samples, num_samples], &indices.device()).scatter(1, indices, values)
+        if num_classes == 1 {
+            self.unsqueeze_dim(1)
+        } else {
```
laggui, line 57: Having a single class doesn't really make sense does it? We can remove this condition, but now that you bring this up this could be verified in the `TensorCheck::one_hot_tensor`.

crates/burn-train/src/metric/confusion_stats.rs
```diff
@@ -34,7 +35,7 @@ impl<B: Backend> ConfusionStats<B> {
         predictions: Tensor<B, 2>,
         targets: Tensor<B, 2, Bool>,
         threshold: Option<f64>,
-        top_k: Option<usize>,
+        top_k: Option<NonZeroUsize>,
```
laggui, line 38: We should probably make the `top_k` for the precision metric `NonZeroUsize` as well then :)

crates/burn-train/src/metric/precision.rs
```diff
@@ -19,9 +21,29 @@ pub struct PrecisionMetric<B: Backend> {
     top_k: Option<usize>,
 }
 
-#[allow(dead_code)]
 impl<B: Backend> PrecisionMetric<B> {
-    ///convert to averaged metric, returns float
+    /// Initiate metric with threshold, used for Binary and Multilabel Classification.
```
laggui, line 25: Initiate -> Initialize

crates/burn-train/src/metric/precision.rs
```diff
@@ -19,9 +21,29 @@ pub struct PrecisionMetric<B: Backend> {
...
+    pub fn new_with_threshold(threshold: f64) -> Self {
+        Self {
+            threshold: Some(threshold),
+            ..Default::default()
+        }
+    }
+
+    /// Initiate metric with top_k, used for Multiclass Classification.
```
laggui, line 33: Same. Initiate -> Initialize

laggui (changes requested):
I think the changes are good overall!

Just come minor comments.

Also, what about splitting the metric to have `BinaryPrecision`, `MulticlassPrecision` and `MultilabelPrecision` instead of users having to fiddle with the right `top_k` and `threshold` parameters?

crates/burn-tensor/src/tensor/api/int.rs
tsanona: Yeah fair, changed

crates/burn-train/src/metric/confusion_stats.rs
tsanona: changed, but I think it might be easier for the user if `.with_top_k()` uses usize as a parameter no?

crates/burn-train/src/metric/precision.rs
tsanona: changed

crates/burn-train/src/metric/precision.rs
tsanona: changed

tsanona:
I've separated the Precision metric by classification type. There is a lot of repeated code but I couldn't find a better way to do it, got any advice? Lemme know what you think :rocket:

crates/burn-train/src/metric/classification.rs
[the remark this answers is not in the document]
tsanona: Yeah of course, I just wanted to get your thoughts on what I had thought to do. Now we have the `ClassificationOutput` and `MultiLabelClassificationOutput` adapted for `ClassificationInput`. It works but I'm still not super happy about it since then the user is able to use, for example `BinaryPrecisionMetric` with a `MultiLabelClassificationOutput`. My idea for the future would be to have separated Inputs and Outputs for each of the classification types such that this would not be possible and would complain at compile time. Thoughts?

crates/burn-train/src/metric/precision.rs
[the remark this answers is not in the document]
tsanona: Yeah no worries thank you for your patience and effort on this PR :pray:

crates/burn-train/src/metric/confusion_stats.rs
laggui: `with_top_k(...)` could still take a `usize` as input but internally convert it to `NonZeroUsize`, e.g.

```rust
let k = NonZeroUsize::new(k).expect("top_k must be non-zero");
```

tsanona: ay ay, shall do that! :star:

crates/burn-train/src/learner/classification.rs
```diff
@@ -27,6 +27,23 @@ impl<B: Backend> Adaptor<LossInput<B>> for ClassificationOutput<B> {
...
+            PrecisionInput::new(
+                self.output.clone(),
+                self.targets.clone().one_hot(num_classes).bool(),
+            )
+        } else {
+            PrecisionInput::new(
+                self.output.clone(),
+                self.targets.clone().unsqueeze_dim(1).bool(),
+            )
+        }
```
laggui, line 43: Shouldn't this also transform the targets with `.one_hot(...)` but force the num classes to 2 (assuming binary classification)?

laggui (commented):
@tsanona 

I made some changes to reduce the repeated code. The `PrecisionMetric` now has a config and can only be created from one of
- `PrecisionMetric::binary(threshold)`
- `PrecisionMetric::multiclass(top_k)`
- `PrecisionMetric::multilabel(threshold)`

I think this reduces the friction points we previously discussed.

See my comment also for the adaptor.

crates/burn-train/src/metric/classification.rs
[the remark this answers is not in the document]
laggui: I understand where you're coming from! But I don't see a straightforward way to do this.. we would have to have a different implementation for the binary, multiclass and multilabel precision metrics because the input is an associated type for the `Metric` trait.

For now I think it's fine.

crates/burn-train/src/learner/classification.rs
tsanona: Hum, I think that would be covered by the case above since then `num_classes == 2`. This, I think, is slightly different from binary classification, is more multi class with 2 classes. Finally, I don't think it would work as output and targets should have the same shape.

tsanona:
Yeah thank you for the changes it looks much better! I like the config solution. Just added pub to the PrecisionMetric, all else seems fine to me. Also, just out of curiosity, why isn't it possible to use the general ClassificationInput?

crates/burn-train/src/learner/classification.rs
laggui: Spam vs not spam (i.e., positive vs negative) is still considered binary classification. I think what you are talking about is just that one uses sigmoid to model the positive output (so there is only one score) vs softmax where you have a score for each. But the targets are still represented as 0 or 1.

I'm not sure what `self.targets.clone().unsqueeze_dim(1).bool()` is supposed to represent 🤔 I could be missing something though

laggui (commented):
Good catch on the public 😅

> Also, just out of curiosity, why isn't it possible to use the general ClassificationInput?

I simply renamed it since it only applies to the precision metric for now.

Should be good to merge, just one comment left 🙂

crates/burn-train/src/learner/classification.rs
tsanona: Sure, at the end of the day they all are the same thing. The point I was trying to make is that in binary classification tasks I expect the output of the model to be (batch_size x 1) while in multi class I would expect (batch_size x 2), thus the targets for the first would be transformed from (batch_size) -> (batch_size, 1) and the second (batch_size) -> (batch_size, 2) so they match the shapes of the outputs. Still this is a preference, if you think it doesn't make sense we can just assert that the second dim of output  is not less than 2.

laggui: Ahhh ok nvm I thought this would lead to some issues because the targets are expected to be one-hot encoded, but in reality this is not entirely true for the binary case for a single scalar output and target. The operations performed will still be valid.

laggui (approved):

PR #2305, opened 2024-09-25
Onnx op topk

oojo12:
### Checklist

   - some failed I think due to my computer. I.e `test_rotary_encoding_forward` failed but passed when I ran it via the ui

### Related Issues/PRs
#1714 

### Changes
- added support for onnx topk op. Relies on version 1 of topk node. Other versions take K as an input instead of an attribute, [ref](https://onnx.ai/onnx/operators/onnx__TopK.html)
- made ` fn tanh_should_not_have_numerical_bugs_on_macos()` only run on macos
- updated `IOEntry::Node` in the `input_names_map`. This was important as before it was never incremented with the number of outputs so you would always have _1 as the output name suffix even if there were x >= 2 outputs.

### Testing
ran the below:
`cargo test`
`./run-checks.sh all`

instructions listed [here](https://github.com/tracel-ai/burn/blob/6d96e8d8086d2309c425f2c8a43a8246f8c454d2/crates/burn-import/onnx-tests/README.md#adding-new-tests) and [here](https://burn.dev/contributor-book/guides/onnx-to-burn-conversion-tool.html#onnx-to-burn-conversion-tool-development-guide)


crates/onnx-ir/src/from_onnx.rs
```diff
@@ -151,7 +151,7 @@ impl GraphData {
         for output in node.outputs.iter_mut() {
             self.input_name_map.insert(
                 output.name.clone(),
-                IOEntry::Node(self.processed_nodes.len(), 0),
+                IOEntry::Node(self.processed_nodes.len(), out_count - 1),
```
laggui, line 154: Oh, I guess we didn't handle multiple outputs without this change right? 😅 Thanks for fixing!

laggui (changes requested):
Thanks for the addition! 🙏 

Overall, implementation looks good! I have a few comments below.

crates/onnx-ir/src/from_onnx.rs
```diff
@@ -370,6 +369,7 @@ pub fn parse_onnx(onnx_path: &Path) -> OnnxGraph {
     let builder = OnnxGraphBuilder::default();
     let graph = builder.build(&onnx_model);
 
+    log::info!("Onnx graph: {:#?}", graph);
```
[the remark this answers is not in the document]
oojo12: ah yeah it was. Forgot to remove it, done now

crates/burn-jit/src/tests/unary.rs
```diff
@@ -4,6 +4,7 @@ mod tests {
     use burn_tensor::{Distribution, Tensor};
 
     #[test]
+    #[cfg(target_os = "macos")]
```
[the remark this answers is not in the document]
oojo12: solid. Should I raise another pr for it then? Without `#[cfg(target_os = "macos")]` the test fails on my machine.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -795,6 +795,18 @@ pub fn tile_config(node: &Node) -> TileConfig {
...
+    let value = node.attrs.get(key).unwrap().clone().into_i64();
+    value
+}
+
+/// Create a TopKConfig from the attributes of the node.
+pub fn top_k_config(node: &Node) -> TopKConfig {
+    let axis: i64 = extract_attr_value_i64(node, "axis");
+    let k: i64 = extract_attr_value_i64(node, "k");
+    TopKConfig::new(axis, k)
+}
```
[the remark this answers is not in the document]
oojo12: I added support for the later versions however, it seems I can't test them because with the code as is. The tests require the input to be initialized with a value and doing so makes the IR not have input anymore. For example, I get the following error after initializing:

```
No variable with name input1
```

I'm incrementally pushing commits, so I'll just resolve all the comments when I finish everything.

crates/onnx-ir/src/from_onnx.rs
oojo12: Haha yeah it this is where all the debug lines came from 😅. And anytime it was kinda fun to thoroughly look through the code learned some things.

oojo12:
Finished making the changes.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -795,6 +802,38 @@ pub fn tile_config(node: &Node) -> TileConfig {
...
+
+    // if axis is negative, it is counted from the end
+    if axis < 0 {
+        axis += data_tensor.dim as i64;
+    }
+
+    let largest = match node.attrs.get("largest") {
+        Some(val) => val.clone().into_i64(),
+        _ => 1,
+    };
```
laggui, line 832: If you're not checking for "k" as the second input of the node (for opsets 10, 11) and just adding support for opset 1, then we don't need to check for the "largest" attribute here. It's only present in the later version 11 of the op.

So we can remove this from the config and node.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -726,20 +726,53 @@ where
     }
 
     /// Returns the `k` largest elements of the given input tensor along a given dimension.
-    pub fn topk(self, k: usize, dim: usize) -> Tensor<B, D, K> {
+    pub fn topk(self, k: usize, dim: usize, largest: Option<usize>) -> Tensor<B, D, K> {
```
laggui, line 729: Since we're only adding support for opset 1, the "largest" option is not necessary. We can remove the changes to the tensor API.

In any case, it would have been preferable to add it as another method, similar to `sort` & `sort_descending`. Something like `topk` and `topk_smallest`. But not required for this PR 🙂

laggui (changes requested):
Sorry for the delay of the follow-up review 😅 been busy the past couple of days.

The changes look good to me! But we can remove the tensor API changes and ONNX op config for topk largest (see my comments)

crates/burn-tensor/src/tensor/api/numeric.rs
oojo12: sounds good will remove the tensor api changes as well

crates/burn-import/src/onnx/op_configuration.rs
oojo12: sure thing will remove and resubmit tonight

crates/burn-import/onnx-tests/build.rs
```diff
@@ -107,6 +107,7 @@ fn main() {
         .input("tests/sum/sum_int.onnx")
         .input("tests/tanh/tanh.onnx")
         .input("tests/tile/tile.onnx")
+        .input("tests/top_k/top_k.onnx")
```
laggui, line 110: Looks the CI caught something I missed! This file doesn't exist anymore with your changes :)

Issue #2302, opened 2024-09-25
MetricsRenderer for non-terminals


lancelet:
### Feature description

When running in a non-terminal (when `!std::io::stdout().is_terminal()`), I think that the burn `Learner` should use a `MetricsRenderer` that is appropriate for an append-only log (eg. `CliMetricsRenderer`).

### Feature motivation

1. When logging training to a file, current output is difficult to use.  eg `cargo run --example mnist --release --features wgpu > log.txt` produces a log file that is unusable for most purposes.
2. When doing a dry-run of training in a GitHub CI agent, I observed a crash.

### (Optional) Suggest a Solution

The `default_renderer` function should check `!std::io::stdout().is_terminal()`, and use `CliMetricsRenderer` instead of `TuiMetricsRenderer`.

### Workaround

If you need this in the mean time, you can set `CliMetricsRenderer` manually:

```
let learner_builder = LearnerBuilder::new(ARTIFACT_DIR)
  // ...
  .renderer(CliMetricsRenderer)
  // ...
  .build(...);
```

PR #2307, opened 2024-09-26
Use CliMetricsRenderer when not in a terminal

lancelet:
### Checklist

### Related Issues/PRs

Fixes: https://github.com/tracel-ai/burn/issues/2302

### Changes

- When stdout is not a terminal, `TuiMetricsRenderer` can create junk output. This change checks that `stdout` is a terminal before using `TuiMetricsRenderer` and falls-back to `CliMetricsRenderer` when necessary.

- `default_renderer` now returns `Box<dyn Renderer>`, but this (hopefully) shouldn't be a problem because `LearnerBuilder` requires a boxed dyn anyway.

- `CliMetricsRenderer` has been changed from `dbg!`, which writes to `stderr`, to `println!("{:?}", ...)`, which writes to `stdout`. This is probably what users would expect when running a job piped to a file.

### Testing

- Tested with `tui` enabled that: a) normal training uses `TuiMetricsRenderer` and b) piping to a file uses `CliMetricsRenderer`.
- Tested with `tui` disabled that everything still works with `CliMetricsRenderer`.

### DIFF
--- a/crates/burn-train/src/renderer/mod.rs
+++ b/crates/burn-train/src/renderer/mod.rs
@@ -1,26 +1,31 @@
+#[cfg(feature = "tui")]
+use std::io::IsTerminal;
+
 mod base;
 pub use base::*;
 
-#[cfg(not(feature = "tui"))]
 mod cli;
-#[cfg(not(feature = "tui"))]
-pub use cli::CliMetricsRenderer as SelectedMetricsRenderer;
 
 #[cfg(feature = "tui")]
 mod tui;
 use crate::TrainingInterrupter;
-#[cfg(feature = "tui")]
-pub use tui::TuiMetricsRenderer as SelectedMetricsRenderer;
 
-/// The TUI renderer, or a simple stub if the tui feature is not enabled.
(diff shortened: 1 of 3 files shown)


crates/burn-train/src/renderer/mod.rs
```diff
@@ -1,26 +1,31 @@
+#[cfg(feature = "tui")]
+use std::io::IsTerminal;
+
 mod base;
 pub use base::*;
 
-#[cfg(not(feature = "tui"))]
 mod cli;
```
lancelet, line 7: Now `cli` may be required even if `tui` is enabled. This is because `cli::CliMetricsRenderer` is the fall-back when `stdout` is not a terminal.

lancelet:
Btw; I'm not sure what to do with the Codecov report. Would appreciate some guidance about where to put tests, etc. (for these side-effecting loggers) if action is required.

nathanielsimard (approved):
LGTM. Tests aren't crucial for that part, it's hard to write (and maintain) for that kind of logic based on global states anyways.

PR #2309, opened 2024-09-26
Move LrSchedule generic to make it easier to use

ArthurBrussee:
### Changes

Fixing a random papercut I ran into. When you manually create a scheduler object like `LinearLrSchedule`, you have to call the step function like

```rust
LrScheduler::<B>::step(&mut scheduler);
```

This is because the Scheduler impl requires a backend which is otherwise unconstrained. This backend is only needed for saving/loading, so just move the generic there.


laggui (approved):
Thanks for  the fix 🙏

```diff
- let lr = LrScheduler::<TestBackend>::step(&mut scheduler);
+ let lr = scheduler.step();
```

That does feel a lot nicer 😄

PR #2311, opened 2024-09-26
Add utility methods to split gradients to GradientParams

ArthurBrussee:
## Changes

Currently, working with multiple different learning rates in Burn can be a bit of a hassle. The easiest way is to create a different GradientParams for each set of variables, and then step the optimizer for each of these. Eg. in my codebase I had something like:

```rust
let mut grad_means = GradientsParams::new();
grad_means.register(
    splats.means.id.clone(),
    splats.means.grad_remove(&mut grads).unwrap(),
);
let mut grad_opac = GradientsParams::new();
grad_opac.register(
    splats.raw_opacity.id.clone(),
    splats.raw_opacity.grad_remove(&mut grads).unwrap(),
);
let mut grad_rest = GradientsParams::new();
grad_rest.register(
    splats.sh_coeffs.id.clone(),
    splats.sh_coeffs.grad_remove(&mut grads).unwrap(),
);
grad_rest.register(
    splats.rotation.id.clone(),
    splats.rotation.grad_remove(&mut grads).unwrap(),
);
grad_rest.register(
    splats.log_scales.id.clone(),
    splats.log_scales.grad_remove(&mut grads).unwrap(),
);
```

This PR just adds some utility methods to turn this into

```rust
let grad_mean = GradientsParams::from_params(&mut grads, &splats, &[splats.means.id]);
let grad_opac = GradientsParams::from_params(&mut grads, &splats, &[splats.raw_opacity.id]);
// Since the first call modifies the grads this will just be the remaining grads.
let grad_rest = GradientsParams::from_module(&mut grads, &splats);
```

Not entirely sure about the names/shape of the API! It feels like this could be a method on Gradients instead but that's currently not really public. 

To make the API slightly simpler I've made ParamId: Copy by making the ID a 64 bit random hash or something instead of a 6(?) byte random string encoded ID.

I've left from_grads in place as is to not break compatibility but also added a from_module, which allows people to extract a GradientParams for a specific module without consuming the gradients, so you could do something like

```rust
let linear_grads = GradientParams::from_module(&mut all_grads, &self.model.linear.a);
let rest_grads = GradientParams::from_grads(all_grads, &self.model); 
```

from_grads could also just be the same thing as from_module if it's ok to break compat.

Something that's a bit gnarly that if you were to extract a GradientsParam for each param individually, each optimizer step would map over all parameters and things become O(n^2). I doubt that matters much in practice.

### DIFF
--- a/crates/burn-common/src/id.rs
+++ b/crates/burn-common/src/id.rs
@@ -1,21 +1,14 @@
-use alloc::string::String;
-
 use crate::rand::gen_random;
-
-use data_encoding::BASE32_DNSSEC;
 
 /// Simple ID generator.
 pub struct IdGenerator {}
 
 impl IdGenerator {
     /// Generates a new ID.
-    pub fn generate() -> String {
-        // Generate 6 random bytes (281,474,976,710,656 combinations)
-        let random_bytes: [u8; 6] = gen_random();
-
-        // Encode the random bytes in base32 DNSSEC
-        // 6 bytes encodes to 10 lower case characters, e.g. "3uu5e6vv7c"
-        BASE32_DNSSEC.encode(&random_bytes)
+    pub fn generate() -> u64 {
+        // Generate a random u64 (18,446,744,073,709,551,615 combinations)
+        let random_bytes: [u8; 8] = gen_random();
+        u64::from_le_bytes(random_bytes)
     }
 }
 
@@ -23,7 +16,7 @@ impl IdGenerator {
 mod tests {
     use super::*;
 
-    use alloc::{collections::BTreeSet, string::String};
+    use alloc::collections::BTreeSet;
 
     #[cfg(feature = "std")]
     use dashmap::DashSet; //Concurrent HashMap
@@ -31,15 +24,10 @@ mod tests {
     use std::{sync::Arc, thread};
 
     #[test]
-    fn not_empty_test() {
-        assert!(!IdGenerator::generate().is_empty());
-    }
-
-    #[test]
     fn uniqueness_test() {
         const IDS_CNT: usize = 10_000;
 
-        let mut set: BTreeSet<String> = BTreeSet::new();
+        let mut set: BTreeSet<u64> = BTreeSet::new();
 
         for _i in 0..IDS_CNT {
             assert!(set.insert(IdGenerator::generate()));
@@ -55,7 +43,7 @@ mod tests {
         const NUM_REPEATS: usize = 1_000;
         const EXPECTED_TOTAL_IDS: usize = NUM_THREADS * NUM_REPEATS;
 
-        let set: Arc<DashSet<String>> = Arc::new(DashSet::new());
+        let set: Arc<DashSet<u64>> = Arc::new(DashSet::new());
 
         let mut handles = vec![];
 
--- a/crates/burn-core/src/optim/visitor.rs
+++ b/crates/burn-core/src/optim/visitor.rs
@@ -4,10 +4,11 @@ use burn_tensor::{backend::AutodiffBackend, Tensor};
 use core::marker::PhantomData;
 
 #[derive(new)]
-pub struct GradientsParamsConverter<'a, M: AutodiffModule<B>, B: AutodiffBackend> {
-    grads: B::Gradients,
-    grads_params: &'a mut GradientsParams,
+pub struct GradientsParamsConverter<'a, 'b, M: AutodiffModule<B>, B: AutodiffBackend> {
+    grads: &'a mut B::Gradients,
+    grads_params: &'b mut GradientsParams,
     phatom: PhantomData<M>,
+    filter: Option<Vec<ParamId>>,
 }
 
 #[derive(new)]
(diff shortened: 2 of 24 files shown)


AsherJingkongChen:
It is a minor but practical change. I think `ParamId` is fine to be `u64` instead of `String`.

ArthurBrussee:
Thanks! Yeah let me just actually include the ID -> u64 change here as well, just so this has a nicer API from the get-go.

crates/burn-core/src/optim/visitor.rs
```diff
@@ -4,10 +4,11 @@ use burn_tensor::{backend::AutodiffBackend, Tensor};
 use core::marker::PhantomData;
 
 #[derive(new)]
-pub struct GradientsParamsConverter<'a, M: AutodiffModule<B>, B: AutodiffBackend> {
-    grads: B::Gradients,
-    grads_params: &'a mut GradientsParams,
+pub struct GradientsParamsConverter<'a, 'b, M: AutodiffModule<B>, B: AutodiffBackend> {
```
nathanielsimard, line 7: I don't think we need 2 lifetimes, but I may be wrong.

crates/burn-common/src/id.rs
```diff
@@ -1,45 +1,33 @@
...
-        // Generate 6 random bytes (281,474,976,710,656 combinations)
-        let random_bytes: [u8; 6] = gen_random();
-
-        // Encode the random bytes in base32 DNSSEC
-        // 6 bytes encodes to 10 lower case characters, e.g. "3uu5e6vv7c"
-        BASE32_DNSSEC.encode(&random_bytes)
+    pub fn generate() -> u64 {
+        // Generate a random u64 (18,446,744,073,709,551,615 combinations)
+        let random_bytes: [u8; 8] = gen_random();
+        u64::from_le_bytes(random_bytes)
```
nathanielsimard, line 11: This might be a breaking change, making old models not able to load. Or do you take care of that in the deserializer?

crates/burn-core/src/optim/visitor.rs
ArthurBrussee: Yeah fair, don't need the flexibility.

crates/burn-common/src/id.rs
ArthurBrussee: Ah right yes.... What's the procedure here? Do I need to handle upgrading the data somewhere?

nathanielsimard: I'm actually unsure. Maybe we keep the `[u8; 8]` as the type carried by the `ParamId`, and I think `[u8; 8]` actually implements `Copy` so that would be great. Then we can use the right decoding/encoding algorithm during serialization/deserialization.

ArthurBrussee: Well it was 6 bytes before! But anyway couldn't find much where changing the type helped. In the end found that there's already a ParamSerde defined that really defines the payload. I've added a custom deserializer in there that can load the old format!

ArthurBrussee: Hm deserializing two potentially different formats is actually bit harder - switched to just always serializing / deserializing base32 strings (either 6 or 8 bytes in length).

ArthurBrussee:
Ah great! Thanks for checking the backwards compat. I've added a small test at least for decoding "legacy" 6 bit IDs, doesn't exactly guarantee compatibility but it's something.

I've added a small section to the custom training loop, lmk if that looks good, I'm really not familiar with the "non-custom" training loop so idk how things would work there haha.

I don't know much about how to update the old models so if you're happy to do that that'd be amazing!

laggui (approved):
LGTM!

The section in the custom training loop is perfect for now! We don't really have a section dedicated for optimizers yet, maybe with documentation changes it'll be moved but that's not for now 🙂 

The models to update are strictly for the examples that use the binary record format (not as portable). Just need to save them to another format and re-save them with the latest changes. I'll take care of that in a minute.

laggui:
Hmmm looks like it doesn't entirely work for backward compatibility.

Currently fails in `deserialize` when decoding a param id `30b82c23-788d-4d63-a743-ada258d5f13c` for the model I'm trying to convert in the examples.

Gonna investigate a bit further.

ArthurBrussee:
Oh hurr that's weird :/ 30b82c23-788d-4d63-a743-ada258d5f13c seems like way more than a 6 bytes id! I guess those might be coming from a different source?

laggui:
Yeah the param ids used to be 128-bit UUIDs before it was changed in #1912, and the `mnist-inference-web` model I was trying to convert was generated before the switch.

It didn't break before since these were also stored as a string, but this explains the error.

I still think the rest of the changes are good! I'll just have to check for this case since users might still have UUIDs in their records that would otherwise break when trying to load.

laggui:
Just added additional support for 16-byte uuid. Turns out the binary model doesn't have to be updated, just had to add backward compatibility for the parameter ids 🙂

laggui (approved):

PR #2323, opened 2024-09-30
Add onnx op trilu

tiruka:
## Pull Request

Add ONNX Operator for Trilu in burn.

### Checklist

### Related Issues/PRs

### Changes

Add trilu onnx operators in burn-import
Modify related files

### Testing

- cargo test
- ./run-checks.sh all


crates/burn-import/onnx-tests/tests/test_onnx.rs
```diff
@@ -1871,6 +1872,25 @@ mod tests {
...
+        let model: trilu::Model<Backend> = trilu::Model::new(&device);
+        let input = Tensor::<Backend, 3>::from_floats(
+            [[[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]]],
+            &device,
+        );
+        let expected = TensorData::from([[
+            [1.0_f32, 2.0_f32, 3.0_f32],
+            [0.0_f32, 5.0_f32, 6.0_f32],
+            [0.0_f32, 0.0_f32, 9.0_f32],
+        ]]);
```
laggui, line 1887: Since `Trilu` implements both the triangular upper and lower, we should probably add a test for both cases.

You've already implemented the export script to generate both, so we could simply have `trilu_upper.onnx` and `trilu_lower.onnx`.

laggui (changes requested):
Thanks for adding this op 🙂 

Implementation looks correct, just a minor comment regarding the tests.

Otherwise, great job!

crates/burn-import/onnx-tests/tests/test_onnx.rs
tiruka: @laggui I appreiciate your review :) I modified codes to add tests for both upper and lower. Hope it will be fine.

laggui (approved):
Thanks for addressing the changes! LGTM 🙂

PR #2336, opened 2024-10-05
Add functionality to ensure ONNX node graph is sorted

DrChat:
### Checklist

### Changes

Microsoft's [Phi-3 ONNX model](https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx) is not topographically sorted.
It's fairly easy to implement a topographical sort though both to ensure that `burn` can still tolerate non-compliant models, and to ensure that the model's graph is valid.

**Special note:** This also fixes a bug where `is_top_sorted` did not account for unnamed input/output edges.

### DIFF
--- a/crates/onnx-ir/src/from_onnx.rs
+++ b/crates/onnx-ir/src/from_onnx.rs
@@ -4,18 +4,20 @@ use std::{
     path::Path,
 };
 
-use crate::node_remap::remap_node_type;
+use crate::{node_remap::remap_node_type, protos::GraphProto};
 
 use super::{
     coalesce::coalesce,
     ir::{Data, OnnxGraph, TensorType},
     proto_conversion::convert_node_proto,
-    protos::{ModelProto, NodeProto, TensorProto, ValueInfoProto},
+    protos::{ModelProto, TensorProto, ValueInfoProto},
 };
 
 use super::dim_inference::dim_inference;
 use super::ir::{ArgType, Argument, Node, NodeType};
 
+use log::{debug, warn};
+use petgraph::{graph::NodeIndex, Graph};
 use protobuf::Message;
 
 const LIFT_CONSTANTS_FOR_NODE_TYPES: [NodeType; 12] = [
@@ -349,15 +351,16 @@ pub fn parse_onnx(onnx_path: &Path) -> OnnxGraph {
 
     // Open the file
     let mut file = File::open(onnx_path).expect("Unable to open file");
-    let onnx_model: ModelProto =
+    let mut onnx_model: ModelProto =
         Message::parse_from_reader(&mut file).expect("Unable to parse ONNX file");
 
     // ONNX nodes must be topologically sorted per spec:
     // https://github.com/onnx/onnx/blob/main/docs/IR.md#graphs
-    debug_assert!(
-        onnx_model.graph.node.is_top_sorted(),
-        "Nodes are not topologically sorted"
-    );
+    //
+    // Some ONNX models are still not topographically sorted, so we'll just do it here
+    // to ensure that the model can be loaded.
+    onnx_model.graph = onnx_model.graph.map(|graph| topo_sort(graph));
+
     log::debug!("Number of nodes: {:?}", onnx_model.graph.node.len());
     log::debug!("Number of inputs: {:?}", onnx_model.graph.input.len());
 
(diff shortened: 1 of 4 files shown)


crates/onnx-ir/src/from_onnx.rs
```diff
@@ -408,40 +411,96 @@ pub(crate) fn remap_unsqueeze_to_reshape(node: &mut Node, out_arg: &Argument) {
...
 
-impl TopologicalSortable for Vec<NodeProto> {
-    fn is_top_sorted(&self) -> bool {
-        // Create a hashmap to store the position of each node in the vector
-        let position: HashMap<String, usize> = self
-            .iter()
-            .enumerate()
-            .map(|(idx, node)| (node.name.clone(), idx))
-            .collect();
+/// Topographically sort an ONNX node graph.
```
laggui, line 415: Pretty minor, but I think you meant *topologically* and not *topographically*

crates/onnx-ir/src/from_onnx.rs
```diff
@@ -349,15 +351,16 @@ pub fn parse_onnx(onnx_path: &Path) -> OnnxGraph {
...
     // ONNX nodes must be topologically sorted per spec:
     // https://github.com/onnx/onnx/blob/main/docs/IR.md#graphs
-    debug_assert!(
-        onnx_model.graph.node.is_top_sorted(),
-        "Nodes are not topologically sorted"
-    );
+    //
+    // Some ONNX models are still not topographically sorted, so we'll just do it here
+    // to ensure that the model can be loaded.
+    onnx_model.graph = onnx_model.graph.map(|graph| topo_sort(graph));
```
laggui, line 362: Instead of always sorting, maybe we should keep the current behavior and only sort if it is not already sorted.

laggui (commented):
I know most model graphs are supposed to be topologically sorted, but that's a nice addition to support more models!

Overall, implementation looks good! Just some minor comments

crates/onnx-ir/src/from_onnx.rs
DrChat: Oops! Let me fix that, thanks :)

crates/onnx-ir/src/from_onnx.rs
DrChat: I'm thinking that it might be better to just always "sort" the graph, or really moreso to sanitize the input graph.
The `topo_sort` function (maybe a misnomer? I could rename it) does a lot of validation in the process of building the graph structure just to make sure it's valid overall before sending it through the rest of the codebase.

It'd probably save people some grief from seeing really odd errors downstream when the graph structure was wrong in the first place :)

laggui: I'm not sure I've encountered invalid graphs a lot, especially since libraries that export to ONNX usually take care to follow the specs. But it can't hurt to check 🙂 

You're probably right that in the case that a user provides an invalid graph as input, this will definitely make it more clear as to what is wrong.

crates/onnx-ir/src/from_onnx.rs
```diff
@@ -408,40 +411,96 @@ pub(crate) fn remap_unsqueeze_to_reshape(node: &mut Node, out_arg: &Argument) {
...
-impl TopologicalSortable for Vec<NodeProto> {
-    fn is_top_sorted(&self) -> bool {
-        // Create a hashmap to store the position of each node in the vector
-        let position: HashMap<String, usize> = self
-            .iter()
-            .enumerate()
-            .map(|(idx, node)| (node.name.clone(), idx))
-            .collect();
+/// Topologically sort an ONNX node graph.
+fn topo_sort(mut input_graph: GraphProto) -> GraphProto {
```
laggui, line 416: In this case, you're right that the function is a bit of a misnomer. Maybe something along the lines of `sanitize_graph_toposort` would be better.

laggui (changes requested):
Looks good!

I think I would just rename the function to be a bit more representative since it doesn't *just* toposort the graph, as you suggested.

/edit: ah, also please check out the clippy suggestions in the CI 🙂

antimora:
Sorry late to the review. 

We had topological sorting of nodes but we removed it: https://github.com/tracel-ai/burn/pull/589 because the nodes are expected to be sorted per specification. I know it's easy to sort them and I coded up myself. We should enforce the specification rather fixing them on the fly. Fixing badly formed ONNX files should be a separate tool/process. Also fixing problems should be done explicitly, i.e. the user should be aware.

@laggui I advocate we do not merge this PR.

laggui:
> Sorry late to the review. 
> 
> We had topological sorting of nodes but we removed it: https://github.com/tracel-ai/burn/pull/589 because the nodes are expected to be sorted per specification. I know it's easy to sort them and I coded up myself. We should enforce the specification rather fixing them on the fly. Fixing badly formed ONNX files should be a separate tool/process. Also fixing problems should be done explicitly, i.e. the user should be aware.
> 
> @laggui I advocate we do not merge this PR.   
> 
> 

Thanks for the feedback!

I was also under the impression that graphs should be topologically sorted according to the spec.

But how come popular models like Phi-3 are exported without sticking to the spec? I think I've also seen other libraries like onnx runtime add support for these cases by sorting the graph as a loading step.

DrChat:
I think it's more likely that the Phi-3 model actually _is_ topologically sorted and that there was a bug in the original `is_topo_sorted` (as mentioned with the empty input names).
However, FWIW, I added a warning message `ONNX graph was not topgraphically sorted`, and it does get printed for the Phi-3 model (perhaps spuriously).

IIRC, [phi3-mini-128k-instruct-cpu-int4-rtn-block-32.onnx](https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx/blob/main/cpu_and_mobile/cpu-int4-rtn-block-32/phi3-mini-128k-instruct-cpu-int4-rtn-block-32.onnx) is indeed the ONNX model in question.

laggui:
Thanks for linking to the ONNX model in question. It appears that the model is actually not topologically sorted, which is a bit unusual.

Note that this large model (>2GB) also requires a second external file ("phi3-mini-128k-instruct-cpu-int4-rtn-block-32-acc-level-4.onnx.data") which stores the weights.

```py
import onnx
onnx.checker.check_model("phi3-mini-128k-instruct-cpu-int4-rtn-block-32-acc-level-4.onnx")
```
```
  File "<stdin>", line 1, in <module>
  File "/home/laggui/miniconda3/envs/torch/lib/python3.10/site-packages/onnx/checker.py", line 163, in check_model
    C.check_model_path(
onnx.onnx_cpp2py_export.checker.ValidationError: Nodes in a graph must be topologically sorted, however input '/model/attn_mask_reformat/attn_mask_subgraph/Gather/output_0' of node:
name: /model/rotemb_caches_subgraph/Greater OpType: Greater
 is not output of any previous nodes.
 ```

sebhtml:
> Sorry late to the review.
> 
> We had topological sorting of nodes but we removed it: #589 because the nodes are expected to be sorted per specification. I know it's easy to sort them and I coded up myself. We should enforce the specification rather fixing them on the fly. Fixing badly formed ONNX files should be a separate tool/process. Also fixing problems should be done explicitly, i.e. the user should be aware.
> 
> @laggui I advocate we do not merge this PR.

+1 

Reference: https://github.com/onnx/onnx/blob/master/docs/IR.md#graphs

PR #2337, opened 2024-10-06
Make {Cosine,Exponential,Linear} LR schedulers return the initial LR at the first step

towerpark:
### Checklist

  (It was executed with `DISABLE_WGPU=1` because currently I have the development environment in a container with no access to the physical GPU)

### Related Issues/PRs

#1198

### Changes

The PR includes the following changes for Cosine, Exponential, and Linear LR schedulers:

* Make the first call to `.step()` return the initial LR instead of the second one.
  Reason:
  　The current behavior skips the initial LR, which is unintuitive and inconsistent with other schedulers.
  　This is a breaking change. To retain the previous behavior, users can run `.step()` once to discard the first value before using it to build a `Learner`.

* Include only fields that change across steps when creating a `Record`.
  Reason: Fields that remain unchanged during training should be kept as part of training configurations. Also, the current behavior is inconsistent with other schedulers.

* Improve tests:
  　* Add a test case for saving and loading records
  　* Test concrete LR values instead of changing trends
  　* Rename the test module to the conventional name `tests`
  　* Add a test case for the `num_iters` field (Linear LR scheduler only)
  Reason: Increase test coverage and align with the convention of module naming.

* Improve documentation comments:
  　* Clarify that the first call to `.step()` returns the initial LR
  　* Add some links
  　* Fix the name of restart of Cosine LR scheduler (Should be "warm restart")
   Reason: Describe the behavior of the first call to `.step()` clearly, and make navigation easier.

### Testing

* Added test cases that targeted the code changes.
* Executed `DISABLE_WGPU=1 bash ./run-checks.sh` and ensured that it exited with code 0.
* Checked changes of documentation comments in a browser manually.
* Confirmed that the changes didn't affect the Book by searching LR in it and reading the contents.

==========
Update: Add description of tests for the Book.

### DIFF
--- a/crates/burn-core/src/lr_scheduler/base.rs
+++ b/crates/burn-core/src/lr_scheduler/base.rs
@@ -17,3 +17,60 @@ pub trait LrScheduler: Send + Sync {
     /// Load the state of the scheduler as a [record](Record).
     fn load_record<B: Backend>(self, record: Self::Record<B>) -> Self;
 }
+
+#[cfg(test)]
+pub(super) mod test_utils {
+    use super::*;
+    use crate::TestBackend;
+
+    pub fn check_lr_sequence<I, S>(mut scheduler: S, expected_lrs: I)
+    where
+        I: IntoIterator<Item = LearningRate>,
+        S: LrScheduler,
+    {
+        // Depending on how learning rates are computed by the scheduler, floating-point arithmetic
+        // error might exceed f64::EPSILON, so we use a larger epsilon here.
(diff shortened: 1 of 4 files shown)


nathanielsimard (approved):
LGTM waiting for @laggui review before merging.

laggui (approved):
Thanks for taking this in charge! Also appreciate the changes summary in the OP 😄 

LGTM! 🚀

towerpark:
You're welcome :)

PR #1553, opened 2024-03-28
WindowDataset/windows function

NicoZweifel:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

- https://github.com/tracel-ai/burn/issues/1551

### Changes

- Adds `WindowDataset` and a `windows` implementation in `Dataset`.

### Testing

- `cd crates/burn-dataset && cargo test window`

PR #2338, opened 2024-10-06
Windows/WindowsIterator/WindowsDataset

NicoZweifel:
### Checklist

### Related Issues/PRs

- Continuation of https://github.com/tracel-ai/burn/pull/1553
- Part of/Originated from https://github.com/tracel-ai/burn/pull/1532

### Changes

- Adds a `Window` trait for all `Dataset<I>` Instances.
- Adds a `Windows` trait for all `Dataset<I>` Instances to keep `base.rs` clean.
- Adds a proper `WindowsDataset`.
- Changes `windows` function to return an Iterator because of the lifetime.

### Testing

`cd crates/burn-dataset && cargo test window`

### Notes

Technically the `windows` function functionality could also be achieved by doing sth like:

```rs
 let iter = WindowsDataset::new(dataset, 3).iter();
```
instead of 

```rs
 let iter = dataset.windows(3);
```

That being said, I could imagine use-cases where the `windows` function is used and it would also be kinda neat if you could collect a Dataset from a `WindowsIterator` again, but with the current lifetime architecture I was not able to achieve this yet and it is probably more a "nice to have" than anything else.


crates/burn-dataset/src/transform/window.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// - `size`: The size of the windows.
+    ///
+    /// # Returns
+    ///
+    /// A `WindowsDataset` instance.
+    pub fn new(dataset: D, size: usize) -> Self
+    where
+        D:,
+    {
+        let size = NonZeroUsize::new(size).expect("window size must be non-zero");
```
NicoZweifel, line 124: I considered not panicing here and passing a `NonZeroUsize` instead but I came to the conclusion that not having to declare `NonZeroUsize::new(x)` and `expect("")` or `unwrap()` every single time is worth panicking in the constructor here. 

If you want to avoid panicking in the constructor, I can update the PR accordingly.

NicoZweifel (commented):
Am happy with how it looks and works now.

crates/burn-dataset/src/transform/window.rs
nathanielsimard: Nan it's fine.

nathanielsimard (approved):
LGTM, waiting for @laggui before merging.

laggui (approved):

PR #2344, opened 2024-10-07
Refactor/elemwise fusion

nathanielsimard:
Refactor how fusion is implemented for element-wise operations in `burn-jit`.

The main difference is that we use the high-level API of CubeCL instead of modifying the IR directly. It's more flexible and will be use to create more dynamic fusion with the fuse on write pattern.


laggui (approved):
Pretty massive changes 🥵 

LGTM

PR #2346, opened 2024-10-08
Use core::error::Error for no-std

antimora:
### Checklist

### Related Issues/PRs

https://github.com/rust-lang/rust/issues/103765

### Changes

Removed feature gate for decorating error types with core::error.

### Testing

run all checks.


antimora:
> Awesome! Though it looks like the feature is not stabilized yet

I forgot we maintain two last versions. This works only with the latest stable Rust. So we'll have to wait.

antimora:
Reopened since MSRV is 1.81 now per  #2388

PR #2347, opened 2024-10-08
Fix autodiff memory leak

### LINKED ISSUE #2333: Heap memory leak from `AutodiffTensor`
**Describe the bug**

The memory leaks come from `AutodiffTensor`. They only appear in the case:
1. The backend is `Autodiff<B>` (AD-enabled).
2. The model/param/tensor requires grad (Tracked).

I profile a simple example using Apple's Instruments.app (v15.3), and I find memory leaks.

**To Reproduce**

I opened a [repository](https://github.com/AsherJingkongChen/burn-model-linear) as an example of memory leaks. It trains a simple linear model. Without Apple's Instruments.app, you can find memory leaks using [Valgrind](https://valgrind.org).

You can clone the example git repository here: https://github.com/AsherJingkongChen/burn-model-linear.git

**Expected behavior**

I expect no memory leak comes from any tracked `AutodiffTensor`.

**Screenshots**

* `leak.1.trace.png`: There are 5100 persistent allocations from `AutodiffTensor::new` in about 2000 iterations.

    

**Desktop**

 - OS: macOS
 - Arch: ARM64
 - Version: 14.6.1

**Additional context**

I think there are some circular references of `rc: Arc<NodeId>` after the step is tracked. Sadly, it is too hard to locate the bug. I need tools to show me where circular references are.

The leak problem would be severe when I frequently replace the autodiff tensors or train a large model in hours. 🥺

nathanielsimard:
Fix #2333

### DIFF
--- a/crates/burn-autodiff/src/runtime/memory_management.rs
+++ b/crates/burn-autodiff/src/runtime/memory_management.rs
@@ -7,7 +7,7 @@ use std::{
 
 #[derive(Default, Debug)]
 pub struct GraphMemoryManagement {
-    nodes: HashMap<NodeRefCount, Vec<NodeID>>,
+    pub nodes: HashMap<NodeRefCount, Vec<NodeID>>,
     leaves: HashSet<NodeID>,
     statuses: HashMap<NodeID, NodeMemoryStatus>,
 }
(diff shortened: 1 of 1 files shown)


AsherJingkongChen:
LGTM, no leak is detected in MWE of #2333 when using `Autodiff<NdArray>` as the backend.

AsherJingkongChen (approved):

laggui (approved):
On top of the tests we did yesterday, I just validated that the CIFAR example still converges and checked the map length + capacity over a training. Both look good! 😎

PR #2349, opened 2024-10-09
Add  ConvTranspose1d ONNX op

tiruka:
### Checklist

### Related Issues/PRs

### Changes

Add ConvTranspose1d ONNX op in accordance with other conv transposed d onnx.

### Testing

`run-checks all` has passed.


crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -209,6 +210,75 @@ pub fn max_pool2d_config(curr: &Node) -> MaxPool2dConfig {
...
+    let bias = curr.inputs.len() == 3;
+
+    // Extract channels from the weight tensor shape [out_channels, in_channels]
+    let shape = weight.shape.clone().unwrap();
+    let channels: [usize; 2] = [shape[1] * group, shape[0]];
+
+    // Create the ConvTranspose1d configuration
+    ConvTranspose1dConfig::new(channels, kernel_shape[0] as usize)
+        .with_stride(stride[0] as usize)
+        .with_padding(pads[0] as usize) // Only use the first padding value for 1D
```
laggui, line 275: Actually, your changes along with the commend here made me realize that the padding is not handled correctly for conv transpose ops with ONNX import.

ONNX defines `pads` as `[x1_begin, x2_begin...x1_end, x2_end,...]`, where `xi_begin` the number of pixels added at the beginning of axis `i` and `xi_end`, the number of pixels added at the end of axis `i`.

But we only support symmetric padding (`xi_begin` = `xi_end`), that's why the [padding for `ConvTranspose2d`](https://github.com/tracel-ai/burn/blob/main/crates/burn-core/src/nn/conv/conv_transpose2d.rs#L68) only takes two values (one for each dimension, symmetric on both sides) and `ConvTranspose1d` only takes one value.

We need to check that the values defined are symmetric, similar to how it's checked for convolution ops (e.g., [`padding_config_conv2d`](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/onnx/op_configuration.rs#L1000) - but only check that the values are not negative or asymmetric).

If you want to fix the other conv transpose ops in this PR feel free to 🙂 otherwise I can tackle this in a separate PR after this one.

laggui (changes requested):
Thanks for adding the op!

Implementation looks good, just a minor comment regarding padding.

Thanks to your comment it actually made me realize that the other conv transpose ops don't check the padding correctly 😅

crates/burn-import/src/onnx/op_configuration.rs
tiruka: @laggui I appreciate your feedback and review. I will take the padding issue, so give me some time to look into it.

tiruka:
@laggui 
Thank you for your merging it. I am happy I could make contributions. 😀
BTW, I am about to push the following codes with tests to make some refactorings. 😅
What's in my mind is to make a common check function for negativeness, and symmetry about pads.
Do you feel if it is good? If so, I will make another PR. I'd like to hear your opinion.

```
fn check_symmetric_pads(pads: &Vec<i64>) {
    // Check the length of pads
    if pads.len() % 2 != 0 {
        panic!("Invalid length: pads must have an even number of elements");
    }

    let half_len = pads.len() / 2;

    // Check for negative values
    if pads.iter().any(|&pad| pad < 0) {
        panic!("Negative pad values are not supported");
    }

    // Check for asymmetric padding
    for i in 0..half_len {
        if pads[i] != pads[half_len + i] {
            panic!("Asymmetric padding is not supported");
        }
    }
}
```
Involved functions will be padding_config_1d, padding_config_2d, padding_config_3d, and conv_transpose1d_config, conv_transpose2d_config, conv_transpose3d_config.

Thank you in advance.

laggui:
I actually was thinking about that while reading your changes, but I don't think it is worth it right now.

The number of lines reduced just to have one function do the check isn't much, and the current checks make it more explicit for each op.

Issue #2357, opened 2024-10-11
Implement softmin


NoahSchiro:
### Feature description

Implement `softmin` for tensors. Softmin is very similar to softmax, but it scales the _lowest numbers_ to be closer to 1, as opposed to the highest numbers, which is the case for softmax. 

### Feature motivation

We already have `softmax` and `softplus` implemented, so it seems apt to include `softmin` to round it out. The activation function is useful in some niche cases. I believe it has some uses in RL and energy based models according to a quick google search.

### (Optional) Suggest a Solution

I have a PR ready to go, just wanted to make an issue first

NoahSchiro:
This is resolved by the merging of #2358

PR #2358, opened 2024-10-11
Softmin

NoahSchiro:
### Checklist

### Related Issues/PRs

#2357 

### Changes

Just a very simple implementation of softmin that follows the implementation of softmax. I am mostly doing this to get more familiar with the crate and learn about how I can contribute :)

### Testing

New test added. Same as softmax, but the values of the expected output tensor is different of course.

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -87,13 +87,8 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> 
 /// The dimension argument `dim` specifies the dimension along which the function will be computed.
 /// It must in the range of `0` and `D-1`.
 pub fn softmin<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
-    check!(TensorCheck::dim_ops::<D>("softmax", dim));
-
-    let tensor = tensor.clone() - tensor.detach().max_dim(dim);
-    let tensor = (-tensor).exp();
-    let tensor_tmp = tensor.clone().sum_dim(dim);
-
-    tensor.div(tensor_tmp)
+    check!(TensorCheck::dim_ops::<D>("softmin", dim));
+    softmax(tensor.neg(), dim)
 }
 
 /// Applies the softplus function
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -89,7 +89,7 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> 
 pub fn softmin<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
     check!(TensorCheck::dim_ops::<D>("softmax", dim));
 
-    let tensor = tensor.clone() - tensor.detach().min_dim(dim);
+    let tensor = tensor.clone() - tensor.detach().max_dim(dim);
     let tensor = (-tensor).exp();
     let tensor_tmp = tensor.clone().sum_dim(dim);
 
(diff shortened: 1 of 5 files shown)


NoahSchiro:
By the way, I am still working on figuring out the architecture of this library, so if there is something I missed somewhere, please let me know and I'd be happy to add. 

What is the distinction between functionality that needs an implementation for all backends, and what can be implemented once and run on any backend? It seems like softmax (and therefore softmin) fall into this latter category, but I wasn't entirely sure.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -78,6 +78,24 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) ->
...
+/// `softmin(x_i) = exp(-x_i) / sum_j(exp(-x_j))`
+///
+/// # Notes
+///
+/// The dimension argument `dim` specifies the dimension along which the function will be computed.
+/// It must in the range of `0` and `D-1`.
+pub fn softmin<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
+    check!(TensorCheck::dim_ops::<D>("softmax", dim));
+
+    let tensor = tensor.clone() - tensor.detach().min_dim(dim);
```
nathanielsimard, line 92: Not sure if we want to subtract `min_dim`. It's for numerical stability, so maybe it should be `max_dim` like softmax ? @laggui

NoahSchiro: Thanks for this @nathanielsimard I wasn't entirely sure what was going on with this line... I did largely copy paste the softmax code for this. I will switch this back to max_dim here.

It appears that max_dim / min_dim has no impact on the test results.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -78,6 +78,24 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) ->
...
+/// Applies the softmin function on the input tensor along the given dimension.
+///
+/// `softmin(x_i) = exp(-x_i) / sum_j(exp(-x_j))`
+///
+/// # Notes
+///
+/// The dimension argument `dim` specifies the dimension along which the function will be computed.
+/// It must in the range of `0` and `D-1`.
+pub fn softmin<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
+    check!(TensorCheck::dim_ops::<D>("softmax", dim));
```
laggui, line 90: Should change the string to "softmin" so the reported error points to the correct origin

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -78,6 +78,24 @@ pub fn softmax<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) ->
...
+
+    let tensor = tensor.clone() - tensor.detach().max_dim(dim);
+    let tensor = (-tensor).exp();
+    let tensor_tmp = tensor.clone().sum_dim(dim);
+
+    tensor.div(tensor_tmp)
```
laggui, line 96: We could simplify this as `softmax(tensor.neg(), dim)` since softmin(x) is the equivalent of softmax(-x)

laggui (changes requested):
> What is the distinction between functionality that needs an implementation for all backends, and what can be implemented once and run on any backend? It seems like softmax (and therefore softmin) fall into this latter category, but I wasn't entirely sure.

As you probably noticed in this case, the activation functions are defined via tensor operations. So the base tensor ops can be defined for each backend, but the activation functions can be defined with the help of these ops 🙂 

Regarding the PR, I think we can simplify the implementation since `softmin(x) = softmax(-x)`.

NoahSchiro:
Thanks for the comments @laggui! I have resolved those issues.

laggui (approved):

PR #2362, opened 2024-10-13
Contributor Book: Fix the link of primitive types in the "Serialization" page

towerpark:
### Checklist

    (Executed with DISABLE_WGPU=1 because currently my development environment is in a container with no access to a physical GPU)

### Related Issues/PRs

None.

### Changes

* Problems:
The link of primitive types in the "Serialization" page is an intra-doc link pointing to a place that doesn't exist. Also its name `primitives types` doesn't sound very natural (English is not my first language, so please  correct me if I'm wrong).

* Changes:
Replace the intra-doc link with an external one pointing to the source code file in "burn-core", and change its name to `primitive types`.

### Testing

* Self-reviewed the change by `git diff`.
* Checked the affected part by reading it in a browser.

### DIFF
--- a/contributor-book/src/project-architecture/serialization.md
+++ b/contributor-book/src/project-architecture/serialization.md
@@ -40,8 +40,8 @@ which field is a parameter or a constant, or we assume that each field implement
 
 The second solution was chosen as it simplifies the code generation and reduces the size of the user
 API. This means that the `Module` trait should be implemented by
-[primitives types](./burn-core/src/module/param/primitive.rs). The following diagrams highlight the
-main types and traits used in the solution.
+[primitive types](https://github.com/tracel-ai/burn/blob/6d96e8d8086d2309c425f2c8a43a8246f8c454d2/crates/burn-core/src/module/param/primitive.rs).
+The following diagrams highlight the main types and traits used in the solution.
 
 <div align="center">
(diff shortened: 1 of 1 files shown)


contributor-book/src/project-architecture/serialization.md
```diff
@@ -40,8 +40,8 @@ which field is a parameter or a constant, or we assume that each field implement
 
 The second solution was chosen as it simplifies the code generation and reduces the size of the user
 API. This means that the `Module` trait should be implemented by
-[primitives types](./burn-core/src/module/param/primitive.rs). The following diagrams highlight the
-main types and traits used in the solution.
+[primitive types](https://github.com/tracel-ai/burn/blob/6d96e8d8086d2309c425f2c8a43a8246f8c454d2/crates/burn-core/src/module/param/primitive.rs).
```
nathanielsimard, line 43: I would use the main branch for the link, otherwise it could become outdated. And for an architecture overview, that's fine not keeping up with versions.

nathanielsimard (changes requested):

contributor-book/src/project-architecture/serialization.md
towerpark: Ok, I will fix it. Using "main" indeed makes more sense.

PR #2364, opened 2024-10-14
Fix xtask args which are unmodified when upgrading xtask commands

tiruka:
### Checklist

### Related Issues/PRs

None

### Changes

In the current contributor-book/src/getting-started/setting-up-the-environment.md, the cargo xtask args are incorrect.
This may be because this portion was overlooked.
This PR fixes it.


laggui (commented):
Actually, while your change is technically correct, the description doesn't match the command anymore.

The commands were customized in PR #2063 iirc and I think the changes were not reflected in the book. Thanks for spotting that!

I think we actually need to update the commands used in the book

@syl20bnr can you double check this please?

syl20bnr (approved):
All good, thanks for the correction!

PR #2377, opened 2024-10-16
Make param thread safe

ArthurBrussee:
### Changes

Currently, when holding a Module over an async await the compiler complains that Box<dyn Fn()> + Send isn't thread safe and needs to be Sync. The reasons for this are somewhat beyond me...  Same goes for state, again not entirely sure why this needs to be Sync for the future to be Send.

### DIFF
--- a/crates/burn-core/src/module/param/base.rs
+++ b/crates/burn-core/src/module/param/base.rs
@@ -1,8 +1,7 @@
 use super::ParamId;
 use alloc::boxed::Box;
 use alloc::format;
-use burn_common::stub::RwLock;
-use core::cell::OnceCell;
+use burn_common::stub::{RwLock, SyncOnceCell};
 use core::ops::Deref;
 
 /// Parameters are the fundamental building blocks of [modules](crate::module::Module) where they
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/module/param/base.rs
```diff
@@ -33,10 +32,10 @@ use core::ops::Deref;
 pub struct Param<T: Parameter> {
     /// The unique ID of this parameter. This is used by eg. optimizers to associate a gradient with a specific parameter.
     pub id: ParamId,
-    state: OnceCell<T>,
+    state: SyncOnceCell<T>,
```
nathanielsimard, line 35: The goal of using `OnceCell` is that we don't have to lock each time we access a parameter, only during lazy initialization. By definition, `Param` can't be `Sync`, and it's intended that way because almost all tensor operations are owned and consume the tensor. When you clone a parameter, we actually initialize it so that both versions have the same state. If you could send it to another thread without cloning it (i.e., implementing `Sync`), then the initialization could be triggered on both threads at the same time, and we would need a synchronization method.

nathanielsimard:
> Currently, when holding a Module over an async await the compiler complains that Box<dyn Fn()> + Send isn't thread safe and needs to be Sync. The reasons for this are somewhat beyond me... Same goes for state, again not entirely sure why this needs to be Sync for the future to be Send.

If you replace `Fn` with `FnOnce` does it work?

ArthurBrussee:
My bad, turns out this is mainly a bad error message. A smaller repro shows the actual problem:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

You just can't hold Send references to a module which makes sense. In a jumbled async error it just pointed to these places having to by Sync but was too quick :P 

(fwiw per your other comments - FnOnce would be nice but then moves out of self which becomes tricky).

PR #2380, opened 2024-10-17
Implement tester

ivnsch:
### Checklist

### Related Issues/PRs

I can open an issue

### Changes

Added a `Tester` (analogous to `Learner`) to burn-train, to perform the testing step.

The idea is that it wraps privately a `Learner`, which it instantiates such that it behaves as a tester (no checkpoints, one epoch, etc). Attempts to create a `Tester` from scratch, led to duplication of virtually the entire `burn-train` crate, given that the modules are strongly coupled (as well as to validation dataset via generics, etc.). Refactoring this would led to a rewrite of the entire crate, which seems overkill.

### Testing

The current state can be tested with this example repo:
https://github.com/ivnsch/tns_brn/tree/95e76cac0318d90b2010dcf37ae02fd6ea582643

### DIFF
--- a/crates/burn-train/src/tester/base.rs
+++ b/crates/burn-train/src/tester/base.rs
@@ -7,9 +7,9 @@ use burn_core::optim::Adam;
 use burn_core::record::{DefaultFileRecorder, FullPrecisionSettings};
 use burn_core::tensor::backend::AutodiffBackend;
 
-/// Learner struct encapsulating all components necessary to train a Neural Network model.
+/// Tester struct which tests a Neural Network model.
 ///
-/// To create a learner, use the [builder](crate::tester::LearnerBuilder) struct.
+/// To create a tester, use the [builder](crate::tester::TesterBuilder) struct.
 pub struct Tester<TC: TesterComponents> {
     pub(crate) learner: Learner<
         LearnerComponentsMarker<
--- a/crates/burn-train/src/tester/base.rs
+++ b/crates/burn-train/src/tester/base.rs
@@ -1,0 +1,31 @@
+use crate::checkpoint::{FileCheckpointer, MetricCheckpointingStrategy};
+use crate::components::LearnerComponentsMarker;
+use crate::components_test::TesterComponents;
+use crate::Learner;
+use burn_core::optim::adaptor::OptimizerAdaptor;
+use burn_core::optim::Adam;
+use burn_core::record::{DefaultFileRecorder, FullPrecisionSettings};
+use burn_core::tensor::backend::AutodiffBackend;
+
+/// Learner struct encapsulating all components necessary to train a Neural Network model.
+///
+/// To create a learner, use the [builder](crate::tester::LearnerBuilder) struct.
+pub struct Tester<TC: TesterComponents> {
+    pub(crate) learner: Learner<
+        LearnerComponentsMarker<
+            TC::Backend,
+            f64,
+            TC::Model,
+            OptimizerAdaptor<
+                Adam<<<TC as TesterComponents>::Backend as AutodiffBackend>::InnerBackend>,
+                TC::Model,
+                TC::Backend,
+            >,
+            FileCheckpointer<DefaultFileRecorder<FullPrecisionSettings>>,
+            FileCheckpointer<DefaultFileRecorder<FullPrecisionSettings>>,
+            FileCheckpointer<DefaultFileRecorder<FullPrecisionSettings>>,
+            TC::EventProcessor,
+            MetricCheckpointingStrategy,
+        >,
+    >,
+}
--- a/crates/burn-train/src/components_test.rs
+++ b/crates/burn-train/src/components_test.rs
@@ -1,0 +1,30 @@
+use crate::metric::processor::EventProcessor;
+use burn_core::{module::AutodiffModule, tensor::backend::AutodiffBackend};
+use std::marker::PhantomData;
+
+/// All components necessary to train a model grouped in one trait.
+pub trait TesterComponents {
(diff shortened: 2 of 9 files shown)


ivnsch:
Opening the PR already to see if there's feedback for this approach. Some things pending, like removing the validation data set parameter and improving naming.

crates/burn-train/src/tester/base.rs
```diff
@@ -0,0 +1,31 @@
...
+                Adam<<<TC as TesterComponents>::Backend as AutodiffBackend>::InnerBackend>,
+                TC::Model,
+                TC::Backend,
+            >,
+            FileCheckpointer<DefaultFileRecorder<FullPrecisionSettings>>,
+            FileCheckpointer<DefaultFileRecorder<FullPrecisionSettings>>,
+            FileCheckpointer<DefaultFileRecorder<FullPrecisionSettings>>,
+            TC::EventProcessor,
+            MetricCheckpointingStrategy,
+        >,
```
ivnsch, line 29: Dummy types for what's not needed - basically, the first/simplest matching types I found. There might be better approach?

crates/burn-train/src/tester/base.rs
```diff
@@ -0,0 +1,31 @@
...
+use burn_core::optim::Adam;
+use burn_core::record::{DefaultFileRecorder, FullPrecisionSettings};
+use burn_core::tensor::backend::AutodiffBackend;
+
+/// Learner struct encapsulating all components necessary to train a Neural Network model.
+///
+/// To create a learner, use the [builder](crate::tester::LearnerBuilder) struct.
+pub struct Tester<TC: TesterComponents> {
+    pub(crate) learner: Learner<
+        LearnerComponentsMarker<
```
ivnsch, line 15: Dummy types for what's not needed - for now the first/simplest matching types I found. Could be improved by implementing no-op types / phantom data.

ivnsch:
The testing flow is completing successfully. There are some problems, however:
- Console shows "training" labels. Changing this is not trivial, because the renderer as well as metrics are coupled to the training and validation flows, and adding support for "testing" labels would require a new variant for all the related functions / classes. Or possibly a complete refactoring. Maybe for now the labels can be leniently left like this.
- The `Learner` is always instantiated with 1 `num_epochs` and there have been no other related changes. It runs only one epoch, the renderer shows correctly 1/1 Epochs, but the summary says "total epochs: 10" and the metrics also appear to reference 10 epochs.  Any ideas? Otherwise I'll investigate.

ivnsch:
The code quality error is caused by what I [commented](https://github.com/tracel-ai/burn/pull/2380#discussion_r1804283663), waiting on feedback to see how to address that.

laggui (commented):
I think this is trending in the right direction, but there are a lot of things I would try to do differently. I understand that it can be intimidating to make bigger changes in someone else's code, but I think to properly add this feature to the crate it requires more than simply trying to fit the `Learner` to use for testing.

There's a ton of stuff which is not required from the learner in order to evaluate, as you probably noticed when you wrote the `TesterComponents` trait. As a matter of fact, the model doesn't even need to be on the autodiff backend.

We probably want to start by the learner components trait to require the tester trait.

```rust
pub trait TesterComponents {
    /// The backend in used for the testing.
    type EvalBackend: Backend;
    /// The model to test.
    type EvalModel: Module<Self::EvalBackend> + core::fmt::Display + 'static;
    type EventProcessor: EventProcessor + 'static;
}

// Alias just to make it more concise
type LearnerBackend<L> = <L as LearnerComponents>::Backend;
type LearnerModel<L> = <L as LearnerComponents>::Model;

pub trait LearnerComponents:
    TestComponents<
    // Force the eval model and backend from the inner types
    EvalModel = <LearnerModel<Self> as AutodiffModule<LearnerBackend<Self>>>::InnerModule,
    EvalBackend = <LearnerBackend<Self> as AutodiffBackend>::InnerBackend,
> {
    // ...
}
```

And we can have a different struct from the `Learner` which removes the requirements for the optimizer, scheduler, checkpointer, etc. Calling `learner.fit(...)` just to evaluate the model seems a bit overkill, we technically only need to do something similar to the `EvalEpoch.run(...)` once for evaluation. We can still reuse a bunch of the components.

But now I'm wondering if, at this point, we shouldn't differentiate between the valid and test metrics. So a trained model could all have train/valid/test metrics reside in the artifact dir and show a summary.

I think that's the correct route, though it requires many changes. I could probably start a draft of that next week once we have the next release ready 🤞

ivnsch:
I initially intended to create more dedicated components, but as you're suggesting (and I've mentioned), it requires extensive changes, which I'm currently not ideal to make (without a disproportionate time investment). I understood that you semi-suggested this wrapper idea (something with setting the epochs to 1, etc. it was vague however, Discord), but yeah, it's hacky, if possible, better to implement it correctly straight away.

My opinion anyway about the proposed idea: making `LearnerComponents` derive `TesterComponents` may fit structurally, but the semantics are weird, as in domain terms, the learner clearly doesn't have testing function. This might become a more concrete problem later. Decoupling some modules could also help, e.g. renderer IMO should be "out"-only and not reference back e.g. `metric`. Ideally also decoupled from specifically training/validation (and now testing). This kind of thing is what led me to have to copy everything in `burn-train` in my first attempt trying to find a more custom route.

laggui:
I think your PR has actually illustrated a possible path forward which should not be too complicated (though, that does not mean that it does not involve many changes 😅 ).

Thanks for the feedback as well! I'll take that into account and once I open a PR for that I'll tag you for an informal review (though you have no obligation, just feel like it might be good since it was your idea initially 🙂 ).

PR #2383, opened 2024-10-18
Initialize accumulator to bias for implicit GEMM to save an expensive `float_add`

wingertge:
### Checklist

### Changes

Loads the bias into the initial accumulator if present, to avoid an unnecessary extra binop.

### Testing

All existing tests pass and no behaviour was changed


nathanielsimard (approved):
LGTM

PR #2386, opened 2024-10-19
Add SPIR-V compiler backend to `burn-wgpu`

wingertge:
### Checklist

### Changes

Adds a feature to enable the new alternative SPIR-V compiler for the WGPU backend. When the feature is enabled the compiler defaults to SPIR-V, but can still be overriden with a generic.

### Testing

New testing option has been added to run SPIR-V tests in addition to the `wgsl` ones. All tests pass. `xtask validate` passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1420,10 +1420,11 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.2.0"
-source = "git+https://github.com/tracel-ai/cubecl?rev=7854b343975c990dd8bb1b4b68b3bc9bda488c1d#7854b343975c990dd8bb1b4b68b3bc9bda488c1d"
+source = "git+https://github.com/tracel-ai/cubecl?rev=547c3fe249deae7d0efe13a47d368c34ce9a736c#547c3fe249deae7d0efe13a47d368c34ce9a736c"
 dependencies = [
  "cubecl-core",
  "cubecl-cuda",
+ "cubecl-hip",
  "cubecl-linalg",
  "cubecl-runtime",
  "cubecl-wgpu",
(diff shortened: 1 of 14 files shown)


wingertge:
Prev Rust version is still set to 1.80 rn, it should be 1.81. That's why `prev` fails rn, lint reasons were introduced in 1.81. I don't think I should update GitHub workflows myself.

nathanielsimard (approved):
The new option is a killer feature, thank you so much 🔥

On another note, which could be solved in a following PR; I think we should revisit the name of our feature flags and backends. To make things simple we should have:

* `tch` => `libtorch`
* `wgpu-spirv` => `vulkan`
* `wgpu` => `webgpu`
* `cuda-jit` => `cuda`

Some feature flags are meant to customize a backend, they should be specific for that backend:

* `cuda` => `candle-cuda`, `libtorch-cuda`
* `metal` => `candle-metal`, `libtorch-metal`
* `openblas` => `ndarray-openblas`
* `accelerate` => `ndarray-accelerate`

Essentially, we're making backends made with `cubecl` first class, meaning we don't need to prefix things with `cubecl` like `cubecl-vulkan`, `cubecl-cuda` or `cubecl-fusion`, but for the third-party backends, I think it's more flexible, since we won't have feature flags that may clash with our own. In that sense, we may also rename `burn-jit` to `burn-cubecl`, since it's pretty much the backend built on top of `cubecl`.

wingertge:
That sounds like a good idea and something I think we can do along with writing a proper CUDA readme and adding `cubecl` to the custom kernels section in the book. For now I only updated the docs related to SPIR-V.

nathanielsimard:
> That sounds like a good idea and something I think we can do along with writing a proper CUDA readme and adding `cubecl` to the custom kernels section in the book. For now I only updated the docs related to SPIR-V.

Gonna create an issue with the comment above.

Issue #2391, opened 2024-10-20
Make Tensor Sync


kingwingfly:
### Feature motivation

Instead of storing image pixels in `Vec`, I switched to `Tensor`:
```rust
#[derive(Debug, Clone)]
pub(crate) struct ImageData<B: Backend> {
    data: Tensor<B, 1>,
    tags: Tensor<B, 1, Int>,
    path: PathBuf,
}
```

Then
```rust
// new a Batcher
impl<B: Backend> Batcher<ImageData<B>, ImageBatch<B>> for ImageBatcher { ... }
```

Then I got stuck here:
```rust
// error
DataLoaderBuilder::new(ImageBatcher::new())
```

`DataLoaderBuilder::new` hopes `ImageData` to be `Sync`:
```rust
impl<I, O> DataLoaderBuilder<I, O>
where
    I: Send + Sync + Clone + std::fmt::Debug + 'static, // Here
    O: Send + Clone + std::fmt::Debug + 'static,
{
    pub fn new<B>(batcher: B) -> Self
    where
        B: DynBatcher<I, O> + 'static,
    {
        ...
    }
}
```
but it's not, due to `Tensor` is not.

Then I check `Tensor` in [docs.rs](https://docs.rs/burn/0.14.0/burn/prelude/struct.Tensor.html#impl-Sync-for-Tensor%3CB,+D,+K%3E), and found it needing `TensorKind::Primitive` to be `Sync`. Again 

```rust
/// A type-level representation of the kind of a tensor.
pub trait TensorKind<B: Backend>: Clone + core::fmt::Debug {
    /// The primitive type of the tensor.
    type Primitive: Clone + core::fmt::Debug + Send; // hope to be Sync
    ...
}

impl<B: Backend> TensorKind<B> for Float {
    type Primitive = TensorPrimitive<B>;
    ...
}

impl<B: Backend> TensorKind<B> for Int {
    type Primitive = B::IntTensorPrimitive;
    ...
}

impl<B: Backend> TensorKind<B> for Bool {
    type Primitive = B::BoolTensorPrimitive;
    ...
}
```

So if `TensorPrimitive`, `B::IntTensorPrimitive` and `B::BoolTensorPrimitive` are all `sync`, Tensor will be `sync`. 

```rust
#[derive(Debug, Clone)]
/// A primitive tensor representation.
pub enum TensorPrimitive<B: Backend> {
    /// Float tensor primitive.
    Float(B::FloatTensorPrimitive),
[...]

PR #2392, opened 2024-10-20
make tensor sync

kingwingfly:
### Checklist

### Related Issues/PRs

#2391

### Changes

make `Tensor` sync

### Testing

I tried it in my ResNet learning project, and it works with feature `tch`... However, I'm not sure if it's really safe to do so.


laggui (approved):
Sorry for the delayed review 😅 

We also wanted to make sure it would not impact anything else with this restriction, but looks good 👍

PR #2403, opened 2024-10-22
Add should_run for convs instead of panicking

ArthurBrussee:
### Changes
Conv2D currently panic in im2col / col2im for images that are too large when autotuning. This works ok on std as tuning uses catch_unwind, but this crashes on wasm.

### Testing
Running some convs on wasm, doesn't crash anymore!

### DIFF
--- a/crates/burn-jit/src/kernel/conv/conv2d/tune/conv_transpose2d.rs
+++ b/crates/burn-jit/src/kernel/conv/conv2d/tune/conv_transpose2d.rs
@@ -113,28 +113,9 @@ fn should_run<R: JitRuntime, F: FloatElement, I: IntElement>(
         stride: [stride_h, stride_w],
         ..
     } = op.options.clone();
-
-    let out_h = calculate_conv_transpose_output_size(
-        kernel_h,
-        stride_h,
-        padding_h,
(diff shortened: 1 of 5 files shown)


crates/burn-jit/src/kernel/conv/conv2d/tune/conv_transpose2d.rs
```diff
@@ -93,3 +98,43 @@ fn create_key<R: JitRuntime, E: FloatElement>(
...
+        padding_w,
+        padding_out_w,
+        dilation_w,
+        input_w,
+    );
+    match index {
+        // im2col
+        1 => batches_per_run(batch_size, input_h, input_w).is_some(),
+        // Implicit gemm.
+        2 => can_do_implicit_gemm(&op.input, &op.weights, op.options.groups, out_h, out_w),
```
wingertge, line 137: There's no transposed implicit GEMM

ArthurBrussee: Ah right ok! Removed it

nathanielsimard (approved):
LGTM

PR #2413, opened 2024-10-23
Extend tensor operations

tsanona:
### Checklist

### Changes

Add some new tensor operations for convenience.

### Testing

New function `argtopk` is tested on doc, other functions use already existing functions.


burn-book/src/building-blocks/tensor.md
```diff
@@ -186,6 +186,7 @@ Those operations are available for numeric tensor kinds: `Float` and `Int`.
 | `tensor.argmin(dim)`                                            | `tensor.argmin(dim)`                           |
 | `tensor.argsort(dim)`                                           | `tensor.argsort(dim)`                          |
 | `tensor.argsort_descending(dim)`                                | `tensor.argsort(dim, descending=True)`         |
+| `tensor.argtopk(k, dim)`                                        | `tensor.topk(k, dim).indices`                  |
```
laggui, line 189: We have `tensor.topk_with_indices(k, dim)` where you can easily ignore the returned values and only keep the indices. I would not add `tensor.argtopk` as an operation.

burn-book/src/building-blocks/tensor.md
```diff
@@ -260,7 +264,6 @@ Those operations are only available for `Float` tensors.
 | `tensor.log1p()`                             | `tensor.log1p()`                   |
 | `tensor.matmul(other)`                       | `tensor.matmul(other)`             |
 | `tensor.one_hot(index, num_classes, device)` | N/A                                |
```
laggui, line 266: Not introduced by this PR, but I just realized that this should be `Tensor::one_hot(index, num_classes, device)`.

Would be good to fix the API doc especially since I think we should rename your current addition to `tensor.one_hot(num_classes)` (see comments below).

crates/burn-tensor/src/tensor/api/int.rs
```diff
@@ -25,6 +27,34 @@ where
...
+    ///
+    /// fn example<B: Backend>() {
+    ///     let device = B::Device::default();
+    ///     let indices: Tensor<B, 1, Int> = Tensor::from_ints([0, 1, 2, 3], &device);
+    ///     let one_hot = indices.one_hot_encode(4);
+    ///     println!("{}", one_hot.to_data());
+    ///     // [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]
+    /// }
+    /// ```
+    pub fn one_hot_encode(self, num_classes: usize) -> Tensor<B, 2, Int> {
```
laggui, line 51: As mentioned earlier, I would simply rename to `one_hot`. It's nicer 🙂

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -742,6 +757,47 @@ where
...
+    ///    let tensor = tensor.argtopk(2, 1);
+    ///    println!("{}", tensor);
+    ///    // Tensor { data: [
+    ///    //   [2, 1],
+    ///    //   [0, 1],
+    ///    //   [1, 2]
+    ///    // ], ... }
+    /// }
+    /// ```
+    pub fn argtopk(self, k: usize, dim: usize) -> Tensor<B, D, Int> {
```
laggui, line 796: This can be removed from the PR (see earlier comments).

laggui (changes requested):
Great addition for the `one_hot` method, but I would not add `argtopk` (unnecessary). 👍 

Thanks for fixing all the typos at the same time 😄

burn-book/src/building-blocks/tensor.md
tsanona: Yeah that's fair enough, I just added it because I had a use for just the indices and since sort, sort_with_indices and argsort, exited I thought of doing the same for topk. Does it not make a difference (in terms of resources) to just produce the top k values versus the values and it's indexes?

burn-book/src/building-blocks/tensor.md
tsanona: Oh yeah good point I'll change that~

burn-book/src/building-blocks/tensor.md
```diff
@@ -274,20 +277,20 @@ Those operations are only available for `Float` tensors.
...
 | Burn API                                         | PyTorch Equivalent                                      |
-| ------------------------------------------------ | ------------------------------------------------------- |
+|--------------------------------------------------|---------------------------------------------------------|
 | `tensor.arange(5..10, device)`                   | `tensor.arange(start=5, end=10, device=device)`         |
 | `tensor.arange_step(5..10, 2, device)`           | `tensor.arange(start=5, end=10, step=2, device=device)` |
 | `tensor.float()`                                 | `tensor.to(torch.float)`                                |
 | `tensor.from_ints(ints)`                         | N/A                                                     |
 | `tensor.int_random(shape, distribution, device)` | N/A                                                     |
 | `tensor.cartesian_grid(shape, device)`           | N/A                                                     |
+| `tensor.one_hot_encode(num_classes)`             | N/A                                                     |
```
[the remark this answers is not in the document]
tsanona: Sure, changed the name. Yeah I also agree that the other `one_hot ` function is a bit limited. Just out of curiosity, we're not removing it because it is a breaking change?

crates/burn-tensor/src/tensor/api/int.rs
tsanona: changed ~

crates/burn-tensor/src/tensor/api/numeric.rs
tsanona: see first comment

burn-book/src/building-blocks/tensor.md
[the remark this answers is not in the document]
laggui: Yeah I think I'd deprecate it first and then remove it since it is user-facing.

burn-book/src/building-blocks/tensor.md
laggui: I think it is not as common and easily achieved with a combination of two ops (as you implemented). That's why I don't think adding this to the tensor operations is not necessary, otherwise the number of operations could become very crowded.

Regarding resources, the difference is probably minimal, but that depends on the size of the tensor. So you can either use `top_k_with_indices` while discarding the values or `argsort_descending` + `select` as in your implementation.

tsanona: alrighty, removed

burn-book/src/building-blocks/tensor.md
```diff
@@ -243,51 +245,51 @@ Those operations are available for numeric tensor kinds: `Float` and `Int`.
...
```
laggui, line 292: Think you missed this in the rename 😉

laggui (commented):
LGTM! 🚀 

Just one minor thing you missed in the book API

burn-book/src/building-blocks/tensor.md
tsanona: ah damn it, thank you

laggui (approved):

laggui:
Awesome 🙏

PR #2414, opened 2024-10-23
Dont attempt naive reduction when reduce_dim is too high

ArthurBrussee:
### Changes
Currently, when autotuning a reduction burn might attempt a naive reduction even for something like the mean() of an 800x800 image. This is sufficiently slow that it's problematic - taking over ~2-3 seconds of GPU time on an M1. This seems to be especially problematic on wasm where tuning is asynchronous, so as other work is running it can take even longer, meaning the app has bad performance for up to minutes before tuning is complete.

This changes it so naive reduction isn't tried above 8000 elements. That's arbitrary - but I suspect it's well above the threshold where a naive reduction makes sense. Happy to bump it up even further if it's a concern.

### Testing
My app warms up much faster on WASM with this change!

### DIFF
--- a/crates/burn-jit/src/kernel/reduce/tune/base.rs
+++ b/crates/burn-jit/src/kernel/reduce/tune/base.rs
@@ -103,6 +103,19 @@ where
             _ => panic!("Fastest index is out of bound"),
         }
     }
+
+    fn should_run(&self, key: &JitAutotuneKey, index: usize) -> bool {
+        let JitAutotuneKey::ReduceDim(key) = key else {
+            unreachable!();
+        };
+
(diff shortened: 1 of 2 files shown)


ArthurBrussee:
I know less about this, but it seems running a shared kernel when the reduce_dim is very small can also be very slow and cause issues. Should similair logic apply there?

laggui:
> I know less about this, but it seems running a shared kernel when the reduce_dim is very small can also be very slow and cause issues. Should similair logic apply there?

I haven't worked on autotune much but it seems like this should also apply.

ArthurBrussee:
Is this looking ok? Would be nice to have all the autotuning changes in!

nathanielsimard (approved):

PR #2416, opened 2024-10-24
Fix alignment issue of TensorData bytes

### LINKED ISSUE #2375: Wrong usage of unsafe `Vec::from_raw_parts` in TensorData
https://github.com/tracel-ai/burn/blob/b29367c1b8763f0e7d81a32c574259d21b33a1f5/crates/burn-tensor/src/tensor/data.rs#L84

This unsafe line has an incorrect usage of `Vec::from_raw_parts` due to re-interpreting a pointer to some `E` as a pointer to `u8`. Note that the [docs state](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts) in the safety requirements:

> T needs to have the same alignment as what ptr was allocated with.

This is because some allocators care about the alignment of these types, hence the buffer that was initially allocated with alignment for the arbitrary (even possibly user-defined) type `E` will be deallocated with alignment of `1`. Some allocators do not care about alignment where this bug does not surface. In any case, miri will complain about this minimal test case:

```
#[cfg(test)]
#[test]
fn bug() {
    let _ = burn::tensor::TensorData::new(vec![0xdeadu32], [1]);
}
```
>  incorrect layout on deallocation: alloc80436 has size 4 and alignment 4, but gave size 4 and alignment 1

---

Comment: The asserts at the top of the function is pointless, since `size_of::<u8>()` is guaranteed to return `1`.
Comment2: The unconditional truncation of the input vector was a bit surprising - instead of an error due to length mismatch as in the case when the input vector is too short - and made me look closer into the function in the first place ;)

WorldSEnder:
### Checklist

  I have some issues with flaky tests, in particular `quantize_dynamic_int8` and `downsample_interpolation` are off-by-1 in a few cases. I've ignored them locally, they seem to be fine in CI.
  Further, I've stumbled over a test failure seemingly related to my AMD GPU driver
> ACO ERROR:
>   In file ../mesa-24.1.4/src/amd/compiler/aco_assembler.cpp:1056
>   Unsupported opcode:  v4: %370:v[8-11] = v_wmma_f32_16x16x16_f16 (latekill)%369:v[36-43].xx, (latekill)%359:v[48-55].xx, %368:v[8-11].xx
- [ ] Made sure the book is up-to-date with this PR. Not applicable, changes to internals only.

### Related Issues/PRs

Fixes #2375.

### Changes

Add a new opaque structure `Bytes` which holds on to tensor data. As explained in the linked issue, using a `Vec<u8>` is incorrect, as this will deallocate the memory with an incorrect alignment/memory layout given to the allocator, which violates the [allocator contract](https://doc.rust-lang.org/std/alloc/trait.Allocator.html#tymethod.deallocate). Instead, the structure remembers the alignment of the data and uses that when deallocating.

For serialization, only the bytes are written and no alignment is taken into account. Instead, when deserializing, the data is allocated with over-alignment (currently `align_of::<u128>`) which should be sufficient to interpret a slice of the data as a slice of some larger type, such as `f32` etc.

Note that this means that serialization is not strictly a round-trip, which can show in the difference between how slices and `Vec` works:
- to re-interpret a `[u8]` as a `[E]`, the slice needs to be aligned at least as strictly as the alignment of `E`.
- to re-interpet a `Vec<A>` as a `Vec<B>`, the alignments must be *exactly* equal. This means that a deserialized instance of `Bytes` possibly can not be converted via `Bytes::try_into_vec`. In case this is needed in the future, one would either need to get access to the needed alignment during deserialization or copy the data to a correctly aligned buffer. Such an API is possible but not included in this PR.

### Testing

Some limited tests of `TensorData` has been run through miri manually to confirm that it doesn't complain after the patch. In case you want tests in the repo, I would need more guidance how to set that up, since miri should probably only run a limited number of tests for CI performance reasons.


WorldSEnder:
Re Codecov report:

The missed lines are for the most part the Debug impl of Bytes. I think it makes sense to invest a bit more time into nice formatting there, any comments on what the output should be would be appreciated.

laggui:
Sorry for the delayed response! We discussed this offline but forgot to get back to you.

First, thanks for identifying the potential issue.

While these changes to handle alignment concerns more generically could certainly provide better safety, it also adds complexity and might not be immediately necessary. That's why we're a bit hesitant on making this change for now. As you noticed this is not an issue for most allocators.

Of course that doesn't mean it is *not* an issue, just that it does not affect the majority of use cases. So we should still keep this in mind, and we might circle back in the future thanks to your investigation.

WorldSEnder:
While I understand the added complexity makes it difficult to review, I am a bit disappointed. There is no way around the fact that the currently implementation is misusing `unsafe`. `miri` rightfully complains about it. Yes, on the major platforms I have on my hand (linux and windows) the allocator plays along, yet `burn-tensor` is `no_std`. I do not have a ras-pi on my hands to test this, but the [burn book](https://burn.dev/burn-book/advanced/no-std.html) mentions the `embedded-alloc` crate and that already seems much more fickle about the alloc requirements. After all, this is rust.

I only noticed this when I implemented the deserialization logic, but please note that the current impl will also happily allocate a `Vec<u8>` aligned to "just" a `u8` even for element types of `f32` or whatever you element type one has that will surprisingly and inconveniently fail to by cast to a slice of its own element type [for example here](https://github.com/tracel-ai/burn/blob/03e4df3ba82891a5e0e2172aba7143e55ff88dae/crates/burn-tensor/src/tensor/data.rs#L160).

In any case, I suppose there is no problem with just leaving this open for the time being.

Just one more question: Would you be less wary if the new unsafe code was in a tested outside crate such as bytemuck? Some parts such as the (de)serialization I think don't make sense to be outside burn, but most of the "byte container" could possibly make more sense upstream.

laggui:
> Don't read this as me trying to rush this in, I would also not be happy to see this merged without a proper review.

No worries, we're on the same page here 🙂

I think our main concerns were more in terms of trade-offs, mostly time and possible ramifications to this breaking change (though I think my position has changed on that.. as you'll see below).

> `burn-tensor` is `no_std`. I do not have a ras-pi on my hands to test this, but the [burn book](https://burn.dev/burn-book/advanced/no-std.html) mentions the `embedded-alloc` crate and that already seems much more fickle about the alloc requirements

I'm glad you brought this up! Honestly I think I kind of forgot to take that into consideration.. this should be especially important for Burn as we aim for portability. I think our previous position was that the current implementation was not an issue for most modern processors and platforms, but that's not good enough.

I should really get a ras-pi myself and start messing around with Burn on `no_std` so we can improve this.

> Just one more question: Would you be less wary if the new unsafe code was in a tested outside crate such as bytemuck? Some parts such as the (de)serialization I think don't make sense to be outside burn, but most of the "byte container" could possibly make more sense upstream.

I'm not too wary about this tbh.

I gotta run but wanted to make sure to get back to you with my current thoughts.

In short, in my previous comment I was opened to the change but did not think it was required for the short term. But now I think we should make sure that this is addressed (ideally, without performance impacts to de/serialization - so we should make sure to benchmark that for some practical use cases).

WorldSEnder:
> I gotta run but wanted to make sure to get back to you with my current thoughts.
> 
> In short, in my previous comment I was opened to the change but did not think it was required for the short term.

No worries, I too think we are on the same page :) While strictly speaking violating the allocator contract like atm could lead to memory unsafety, I don't think any allocator actually does use the violated guarantees in a meaningful way (but I'm open to surprises). On the other hand, misaligned allocations *would* be a real concern but at least as far as I can tell mostly lead to panics, not undefined behaviour or violations of memory safety, so a relaxed approach to fixing this is fine :+1: Sorry if I came off as a bit panicky

laggui:
@WorldSEnder I was looking at your PR yesterday and I think we'd like to stay away from a new opaque type for now.

I also benchmarked some solutions to preserve alignment and none of them come close to the current pointer re-interpret sadly.

I think we can keep both approaches instead based on some target architectures for now, since we know that heap allocations are overaligned on mainstream systems. And for others with stricter alignment rules, we fallback to something similar to what you implemented in the `Bytes` opaque type. We also add some documentation for anyone that might come across the code in the future 🙂 

```rust
    /// Initializes a new tensor data structure from the provided values.
    fn init<E: Element, S: Into<Vec<usize>>>(
        mut value: Vec<E>,
        shape: S,
        dtype: DType,
    ) -> Self {
        // Ensure shape is valid
        let shape = shape.into();
        let shape_numel = Self::numel(&shape);
        value.truncate(shape_numel);
        let numel = value.len();
        assert_eq!(
            shape_numel, numel,
            "Shape {:?} is invalid for input of size {:?}",
            shape, numel,
        );

        let bytes = Self::to_bytes(value);

        Self {
            bytes,
            shape,
            dtype,
        }
    }

    fn to_bytes<E: Element>(value: Vec<E>) -> Vec<u8> {
        // Fast path for architectures with heap allocation overalignment.
        #[cfg(all(
            not(miri),
            any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64")
        ))]
        {
            let mut value = value;
            let factor = std::mem::size_of::<E>() / std::mem::size_of::<u8>();
            let len = value.len() * factor;
            let capacity = value.capacity() * factor;
            let ptr = value.as_mut_ptr();

            std::mem::forget(value);

            // SAFETY: Practically, mainstream systems align heap allocations beyond
            // the minimum required alignment for the element type (e.g., 8 bytes on 64-bit
            // systems).
            //
            // This is not guaranteed by Rust's memory model, which follows strict alignment
            // requirements as specified by the type's layout.
            //
            // But we leverage the overalignment and efficient unaligned access to interpret
            // the same memory layout as individual bytes instead of `E` elements.
            // This simple trick makes this pointer cast acceptable (and requires no copy).
            unsafe { Vec::from_raw_parts(ptr as *mut u8, len, capacity) }
        }

        // Fallback for architectures with stricter alignment rules.
        #[cfg(any(
            miri,
            not(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))
        ))]
        {
            let num_bytes = value.len() * std::mem::size_of::<E>();
            let mut bytes = vec![0u8; num_bytes];
            unsafe {
                std::ptr::copy_nonoverlapping(
                    value.as_ptr() as *const u8,
                    bytes.as_mut_ptr(),
                    num_bytes,
                );
            }
            bytes
        }
    }
```

I also added `miri` to the conditions to make sure it doesn't complain 😅 let me know how you feel about this.

We probably need to be careful about `f64` elements for 32-bit systems too so the condition might need to be changed.

WorldSEnder:
@laggui I think in the long run, an opaque wrapper should happen (note the wrapper in this PR does no mem-copy so should not have the overhead you mention). A publicly exposed field of `Vec<u8>` lets the user `mem::swap` in any `Vec<u8>` that they may have constructed themselves. I'd want to double check if that entails that the storage has to be aligned if the system allocator always aligns it, my current assumption would be that it entails more weirdness that an opaque wrapper struct would cover.

Secondly, the target detection is perhaps 90% there. Sadly, I don't think there is a way to detect the specific *allocator* being used, and the check is of course only detecting the `target_*`, not whether [System](https://doc.rust-lang.org/std/alloc/struct.System.html) is being used as the [`#[global_allocator]`](https://doc.rust-lang.org/std/prelude/rust_2024/attr.global_allocator.html) and as far as I know there is no way to detect which allocator ends up being used in the end as the forwarding [`Global`](https://doc.rust-lang.org/std/alloc/struct.Global.html) does not expose this.
Note for future reference: It seems the choice of default system allocator is made [in the std in here](https://doc.rust-lang.org/src/std/sys/alloc/mod.rs.html#70-94) if we want to go the route of just using system information as a proxy.

What exactly is the concern with an opaque struct? It seems there would be a simple almost-zero-cost conversion from `Vec<u8>` to `TensorData`, a `Deref<Target=[u8]>` impl and reasonably efficient conversion back to `Vec<u8>` (though that might entail a mem-copy or the downside of possibly-undefined-behaviour-maybe-just-panicking.

PS: Did you encounter any slowdown with this PR in your benchmarking? If so, this would be a concern and I would put in more work to make sure I didn't miss any unreasonable overhead.

WorldSEnder:
> Yeah I think we could (should) remove the public field and users could only get the slice with the current available methods. So no direct access to the underlying bytes. Not sure why we left it public actually 😅

I was wondering if it was intentional that you could push further bytes after construction, cause that seems a bit weird since you could push e.g. a single `u8` even when the TensorData contains `u16` etc. Good to hear that this was never intended, thought I missed something :D

> Was trying to cover the most important cases with this condition where stricter alignment requirements are necessary, but I realize this is not 100%. 

After more investigation, and a bit longer look than I had wished, I found that the default allocator on windows [actually does use the passed align in a non-trivial way](https://github.com/rust-lang/rust/blob/5958f5e08fa88ee95ede8c00f1b89befe0372d54/library/std/src/sys/pal/windows/alloc.rs#L166). Specifically, data below `MIN_ALIGN` is just allocated as-is, but anything beyond that will allocate additional space and contain a header with a pointer to the actual allocation, pictogram:

```
 v --- aligned to MIN_ALIGN, this pointer is allocated from HeapAlloc
 | x x x x Header(ptr) | u u u u u u u ..... |
 ^----------------^-^
                       ^  aligned to requested alignment, this pointer is returned to user-code
```

The reason the current code gets away with it - supposedly - is because on 64-bit platforms [`MIN_ALIGN`](https://github.com/rust-lang/rust/blob/5958f5e08fa88ee95ede8c00f1b89befe0372d54/library/std/src/sys/pal/common/alloc.rs#L7-L40) is 16, which is larger than the data types that currently get put into a tensor. But it's awefully brittle and it just feels like corner-case after corner-case to find these kind of "exceptions" to the allocator playing nicely along to wrong alignment.

> it introduces a breaking change

Yeah, the timing close to the release of 0.15 is a bit unfortunate. Not sure about your planned release cycles, I wouldn't mind holding it until the breaking 0.16 comes along naturally.

> Also tiny note on the current Bytes implementation: what about sending between threads?

Thread-safety is indeed a good point. The current `Bytes` should be `Send + Sync` without a problem since `Element: Send + Sync` and `&Bytes` offering only read access, but that could be encoded in some static asserts and tests.

> The results seem to be on par with the current, which is great 🙏 

There seems to be a slight slowdown in deserialization, which might perform an extra copy to an aligned buffer. I think in that case we can get away with actually checking the runtime alignment instead (still storing and de-allocating with the alignment of the actual buffer) and not copying/only copying when someone wants to convert it an actual `Vec<E>`. I will take another look 👍

WorldSEnder:
Alright, have added both suggestions (`Sync` and `Send` + avoiding a copy if the deserialized vector ends up being aligned). I have added further tests and these can be run under miri. In case you want to do that yourself, try it with

> cargo +nightly miri test --package burn-tensor --lib -- tensor::bytes::tests --show-output

WorldSEnder:
@laggui I wanted to merge the changes from quantization and in the process came up with a few questions where I'm not sure what's intended and what is not, again about serialization.

There should be no difference in this PR and the state on `main`, but (de)serialization seems to assume that data in storage is in system byte order. I don't see any conversion from/to network byte order. So serializing a tensor on a little endian (LE) system such as x86, then deserializing on a BE system (well mostly SPARC; but fairly uncommon I think) would shuffle the bytes and not recover the data as intended. Further network order is big endian (which is lame, as most systems have to perform a byte shuffle for what are merely historic reasons). So the question is this: Since serialization is performed as a byte sequence where endianess matters, is the current state just assuming little endian everywhere and not correcting for it correct?

If the answer to the above is yes, and there is no external constraints on the byte order, I don't see why the byte packing (which *does* seem to pad with 0s, then reinterpret as a Vec of **big endian** `u32`) could not just use the system endianess and pad then reinterpret without endianess conversion and do so in place. The current implementation seems to create 2 extra Vecs. One in `pack_i8s_to_u32s`, then another in `to_vec()`.

In any case, I will implement a `fn extend_from_slice` for `Bytes` to accomodate storing the quantization strategy there.

A lot of methods seem to ignore the difference between `values_as_bytes` and `as_bytes` and try to make use of the additional quantization data, too (such as `iter`, `to_vec`). Is that intentional?

laggui:
> @laggui I wanted to merge the changes from quantization and in the process came up with a few questions where I'm not sure what's intended and what is not, again about serialization.
> 
> There should be no difference in this PR and the state on `main`, but (de)serialization seems to assume that data in storage is in system byte order. I don't see any conversion from/to network byte order. So serializing a tensor on a little endian (LE) system such as x86, then deserializing on a BE system (well mostly SPARC; but fairly uncommon I think) would shuffle the bytes and not recover the data as intended. Further network order is big endian (which is lame, as most systems have to perform a byte shuffle for what are merely historic reasons). So the question is this: Since serialization is performed as a byte sequence where endianess matters, is the current state just assuming little endian everywhere and not correcting for it correct?

Yep, we don't check for endianness atm so we assume everything is fine for now. Might have to check in the future.

> If the answer to the above is yes, and there is no external constraints on the byte order, I don't see why the byte packing (which *does* seem to pad with 0s, then reinterpret as a Vec of **big endian** `u32`) could not just use the system endianess and pad then reinterpret without endianess conversion and do so in place. The current implementation seems to create 2 extra Vecs. One in `pack_i8s_to_u32s`, then another in `to_vec()`.

Funny you bring this up, I realized the other day while making some changes that the "packing" order might not be as optimal. Quantization is still WIP but we were just about to refactor the quantization kernels in jit along with the representation. Learning some things as we go :)

> In any case, I will implement a `fn extend_from_slice` for `Bytes` to accomodate storing the quantization strategy there.
> 
> A lot of methods seem to ignore the difference between `values_as_bytes` and `as_bytes` and try to make use of the additional quantization data, too (such as `iter`, `to_vec`). Is that intentional?

In `iter` we chose not to dequantize the values so it is intentional for the current behavior. Doesn't make as much sense in `to_vec`, we should change that.

WorldSEnder:
> Yep, we don't check for endianness atm so we assume everything is fine for now. Might have to check in the future.

Ah okay! If you do, that could be a good spot to move the assumption of a `MAX_ALIGN` to something more `dtype` dependent, as I would assume that converting byte order also needs to know about the underlying datatype, which could in one swoop also fix the runtime alignment (`try_enforce_runtime_align` in this PR).

laggui:
> I was wondering if it was intentional that you could push further bytes after construction, cause that seems a bit weird since you could push e.g. a single `u8` even when the TensorData contains `u16` etc.

I think there might be some use cases for that actually.. I discussed with @nathanielsimard last week and iirc this was one of the reasons for keeping it as `Vec<u8>` at the moment. Correct me if I'm wrong, but this was for some work in `burn-remote`? Can't remember exactly what you mentioned so maybe you could chime in here with your points.

nathanielsimard:
I don't recall a reason to push `u8` into the vector.

WorldSEnder:
@syl20bnr I believe https://github.com/tracel-ai/github-actions/blob/b87057a7f68360d9d01a8da21de874c38e7acbab/setup-rust/action.yml#L23-L26 is accidentally run on windows machines as well, leading to some tests failing there.

syl20bnr:
Argh, thank you, fixing it.

syl20bnr:
@WorldSEnder should be fixed, I restarted the job.

WorldSEnder:
Okay, implemented `Bytes::extend_from_byte_slice` to support appending quantization data. This means `TensorData::quantized` should use at one less vec copy in most cases. Note that `pack_i8s_to_u32s` could probably be done in place, but that's left for another PR.

I realized that most methods that use the bytes, such as the aforementioned `as[_mut]_slice` and `to_vec/into_vec` refuse on the grounds of mismatching datatypes ("normal" vs quantized) before trying to touch the unrelated quantization data. Perhaps this could be a deliberate choice telling users to unquantize before being allowed to use these methods anyway.

As a side-effect of working `Bytes` into `quantized`, this now has a length check (unified with new), I suppose this was meant to be there in the first place. Let me know if I missed anything :)

juntyr:
TLDR This PR should be merged asap. Fixing a soundness issue that could turn into a security advisory should not be stalled for this long.

I recently started experimenting with burn, enjoyed it a lot, but now came across this PR and I'm concerned.

> As explained in the linked issue, using a `Vec<u8>` is incorrect, as this will deallocate the memory with an incorrect alignment/memory layout given to the allocator, which violates the [allocator contract](https://doc.rust-lang.org/std/alloc/trait.Allocator.html#tymethod.deallocate).

This is still an understatement. When violating a safety condition of an unsafe method, you need to assume that this causes undefined behaviour (and in this case it immediately does and is caught by Miri).

> While these changes to handle alignment concerns more generically could certainly provide better safety, it also adds complexity and might not be immediately necessary. That's why we're a bit hesitant on making this change for now. As you noticed this is not an issue for most allocators.

The Rust compiler is allowed to do anything with a program that exhibits undefined behaviour. So any program that uses burn *somewhere* could in theory be compiled into complete garbage because burn includes unsound code. The compiler is not as optimising/punishing that every soundness issue infects compilation this way, as can be seen by this code running just fine, but that can change with any new release of Rust. It would be a shame if some larger application upgrades their Rust version and suddenly crashes at runtime because the compiler is now smart enough to exploit this undefined behaviour for further optimisations.

While you are right that this does issue does not manifest on the most popular platforms, Rust has a high standard for writing sound code - "it hasn't segfaulted on my machine yet" is not sufficient in Rust. Safe Rust code must never exhibit undefined behaviour (and so any soundness issues should be fixed as soon as they're discovered) and should *never* segfault, and since you are exposing `TensorData` with a safe interface, you are required to make sure that *no* undefined behaviour can occur when calling it from safe code.

laggui:
> "it hasn't segfaulted on my machine yet" is not sufficient in Rust.

To be fair, it hasn't segfaulted on *any* machine for Burn users 😅 the current implementation is safe on the target platforms using Rust's default allocator. The discussion in this PR details why it works in practice with the supported tensor data types, and other allocators were out of scope. Now, as I mentioned in my previous comments, this doesn't mean that this is a non issue.

I actually would have liked to merge this two weeks ago following the planned release, but we had to work on other issues and improvements (here and in CubeCL) which have caused us to push our release schedule. And the plan was to make this breaking change for the following release.

juntyr:
I’m happy to hear the PR will be merged soon <3

I really like burn so far and want to use it for a longer term project.

The default Rust std allocator is unspecified, but currently delegates to the System allocator. After a few more layers of redirection, we end up with its implementation here (for Unix, in adjacent files for other operating systems): https://github.com/rust-lang/rust/blob/master/library/std/src/sys/alloc/unix.rs You can see that it boils down to libc being called on Unix.

However, different targets use different allocators to back the system and so you would have to check all of their implementations to see if your stronger requirements are met.

Furthermore, supplying a different global allocator is not a feature that a Rust crate can choose to not support. The global allocator is opaque to all libraries and the only assumptions you can make are in its safety contract.

So you could, with just safe code, change the allocator to one where freeing memory with a different (but same-sized) layout than it was allocated with (the effect of creating a Vec from the raw parts of a Vec of a type with a different size and alignment) does mess things up. In that case, the UB would still be burn’s fault.

All of this is to say that UB must be taken seriously.

Perhaps it might be a good idea to run your test suite in CI with Miri (probably only the pure Rust parts, so e.g. just the NdArray backend and only on smaller examples since Miri is quite slow) so that you can have more reassurance that your code is safe.

laggui:
@WorldSEnder thanks for keeping this up to date with the changes on main 🙏 

Before merging I wanted to be sure that there were no big performance impacts, but running a quick serialization / deserialization test showed ~30% slowdown for deserialization. In my previous benchmarks I only tested the module loading (without saving / loading on disk), so this discrepancy seems to be directly related to the `Deserialize` implementation.

**main**
| Benchmark   | Feature | Backend   | Device | Median         |
|-------------|---------|-----------|--------|----------------|
| serialize   | ndarray | `ndarray` | Cpu    | 835.821ms      |
| deserialize | ndarray | `ndarray` | Cpu    | 114.644ms      |

**Bytes struct**
| Benchmark   | Feature | Backend   | Device | Median         |
|-------------|---------|-----------|--------|----------------|
| serialize   | ndarray | `ndarray` | Cpu    | 799.097ms      |
| deserialize | ndarray | `ndarray` | Cpu    | 148.214ms      |

Once the deserialization is on par, this should be good to merge 🙂 

I'll look into this tomorrow but feel free to ping me otherwise.

<details>
<summary>Benchmark Code </summary>

```rust
use backend_comparison::persistence::save;
use burn::record::{FullPrecisionSettings, HalfPrecisionSettings, NamedMpkFileRecorder};
use burn::tensor::backend::Backend;
use burn::tensor::Device;
use burn::{config::Config, module::Module, nn};
use burn_common::benchmark::{run_benchmark, Benchmark};
use derive_new::new;

#[derive(Module, Debug)]
struct BenchmarkModule<B: Backend> {
    linears: Vec<nn::Linear<B>>,
}

#[derive(Config, Debug)]
struct BenchmarkConfig {
    linear: nn::LinearConfig,
    num_layers: usize,
    path: String,
}

impl BenchmarkConfig {
    pub fn init<B: Backend>(&self, device: &B::Device) -> BenchmarkModule<B> {
        BenchmarkModule {
            linears: (0..self.num_layers)
                .map(|_| self.linear.init(device))
                .collect(),
        }
    }
}

#[derive(new)]
struct SerializeBenchmark<B: Backend> {
    config: BenchmarkConfig,
    device: Device<B>,
}

impl<B: Backend> Benchmark for SerializeBenchmark<B> {
    type Args = (
        BenchmarkModule<B>,
        NamedMpkFileRecorder<HalfPrecisionSettings>,
    );

    fn name(&self) -> String {
        "serialize".into()
    }

    fn execute(&self, (module, recorder): Self::Args) {
        module
            .save_file(self.config.path.clone(), &recorder)
            .unwrap();
    }

    fn prepare(&self) -> Self::Args {
        let module = self.config.init(&self.device);

        (module.clone(), NamedMpkFileRecorder::new())
    }

    fn sync(&self) {
        B::sync(&self.device)
    }
}

#[derive(new)]
struct DeserializeBenchmark<B: Backend> {
    config: BenchmarkConfig,
    device: Device<B>,
}

impl<B: Backend> Benchmark for DeserializeBenchmark<B> {
    type Args = (
        BenchmarkModule<B>,
        NamedMpkFileRecorder<FullPrecisionSettings>,
    );

    fn name(&self) -> String {
        "deserialize".into()
    }

    fn num_samples(&self) -> usize {
        50
    }

    fn execute(&self, (module, recorder): Self::Args) {
        let _ = module.clone();
        module
            .load_file(self.config.path.clone(), &recorder, &self.device)
            .unwrap();
    }

    fn prepare(&self) -> Self::Args {
        let module = self.config.init(&self.device);

        (module.clone(), NamedMpkFileRecorder::new())
    }

    fn sync(&self) {
        B::sync(&self.device)
    }
}

#[allow(dead_code)]
fn bench<B: Backend>(
    device: &B::Device,
    feature_name: &str,
    url: Option<&str>,
    token: Option<&str>,
) {
    let config = BenchmarkConfig::new(
        nn::LinearConfig::new(2048, 2048),
        12,
        "/tmp/serialize.mpk".into(),
    );

    let benchmark_serialize = SerializeBenchmark::<B>::new(config.clone(), device.clone());
    let benchmark_deserialize = DeserializeBenchmark::<B>::new(config.clone(), device.clone());

    save::<B>(
        vec![run_benchmark(benchmark_serialize)],
        device,
        feature_name,
        url,
        token,
    )
    .unwrap();
    save::<B>(
        vec![run_benchmark(benchmark_deserialize)],
        device,
        feature_name,
        url,
        token,
    )
    .unwrap();
}

fn main() {
    backend_comparison::bench_on_backend!();
}
```
</details>

laggui (approved):

Issue #1477, opened 2024-03-15
Precision errors running tests


McArthur-Alford:
**Describe the bug**
When running the tests, I run into some precision errors. I have ran them a handful of times, and they seem to consistently be the following two tests:
```
tests::jit::module_nearest_interpolate::tests::test_downsample_interpolation
tests::jit_fusion::module_nearest_interpolate::tests::test_downsample_interpolation
```

The full log is:
```
failures:

---- tests::jit::module_nearest_interpolate::tests::test_downsample_interpolation stdout ----
thread 'tests::jit::module_nearest_interpolate::tests::test_downsample_interpolation' panicked at crates/burn-wgpu/src/lib.rs:73:5:
Tensors are not approx eq:
  => Position 3: 7 != 6 | difference 1 > tolerance 0.0010000000000000002
  => Position 9: 161 != 160 | difference 1 > tolerance 0.0010000000000000002
  => Position 15: 315 != 314 | difference 1 > tolerance 0.0010000000000000002
  => Position 21: 469 != 468 | difference 1 > tolerance 0.0010000000000000002

---- tests::jit_fusion::module_nearest_interpolate::tests::test_downsample_interpolation stdout ----
thread 'tests::jit_fusion::module_nearest_interpolate::tests::test_downsample_interpolation' panicked at crates/burn-wgpu/src/lib.rs:73:5:
Tensors are not approx eq:
  => Position 3: 7 != 6 | difference 1 > tolerance 0.0010000000000000002
  => Position 9: 161 != 160 | difference 1 > tolerance 0.0010000000000000002
  => Position 15: 315 != 314 | difference 1 > tolerance 0.0010000000000000002
  => Position 21: 469 != 468 | difference 1 > tolerance 0.0010000000000000002
```

**To Reproduce** 
 1. Clone the latest version of main
 2. Use run-checks.sh and wait for the errors

**Hardware:**
 - OS: Linux (NixOS 24.04 with kernel 6.7.6)
 - GPU: AMD ATI Radeon RX 6900 XT
 - CPU: Intel i9-12900k

WorldSEnder:
As I noted in #1444 I ignored these tests too on my local run. Hardware:

Arch Linux, kernel on 6.7.9
Intel i7-13700K
AMD Radeon RX 5700 XT

antimora:
@nathanielsimard since you guys are working on this

laggui:
Pretty sure these have been fixed since 😅 feel free to re-open otherwise

PR #2423, opened 2024-10-25
Add step learning rate scheduler

towerpark:
### Checklist

The check script was executed with `DISABLE_WGPU=1`, because otherwise I would get an error similar to #1477, which is unrelated to this PR.

### Related Issues/PRs

#1198

### Changes

This PR adds the step learning rate scheduler, which is on the todo list in the issue above.

Initial learning rate and gamma are not checked when initializing a scheduler, because I think atypical values such as zero and negative values might be useful in some cases like [debugging](https://datascience.stackexchange.com/q/89518). Printing a warning is probably better but I'm not sure (I would be glad to hear your thoughts).

### Testing

* Added test cases for the scheduler
* Run `DISABLE_WGPU=1 bash ./run-checks.sh` and ensured that it exited with code 0
* Checked documentation by reading it in a browser
* Confirmed that the changes didn't affect the Books by searching LR in them and reading the contents

==========
Edit: Key mis-press (nothing changed).

### DIFF
--- a/crates/burn-core/src/lr_scheduler/step.rs
+++ b/crates/burn-core/src/lr_scheduler/step.rs
@@ -1,0 +1,178 @@
+use burn_tensor::backend::Backend;
+
+use crate as burn;
+
+use super::LrScheduler;
+use crate::{config::Config, LearningRate};
+
+/// The configuration for create a [step learning rate scheduler](StepLrScheduler).
+///
+/// This scheduler returns the learning rate `initial_lr` from the start, and keeps doing so until
+/// the same value has been given for `step_size` times. Then it multiplies the learning rate by
+/// `gamma` before repeating the process.
+///
+/// ## Notes
+///
+/// The [step](StepLrScheduler::step) method of the scheduler panics if it is called more than
+/// `i32::MAX + 1` times.
+#[derive(Config)]
+pub struct StepLrSchedulerConfig {
+    // The learning rate at the initial step.
+    initial_lr: LearningRate,
+    // The number of iterations over which the learning rate remains unchanged before the next
+    // update.
+    step_size: usize,
+    /// The factor by which the learning rate is multiplied with each update. Default: 0.1.
+    #[config(default = 0.1)]
+    gamma: f64,
+}
+
+impl StepLrSchedulerConfig {
+    /// Initializes a [step learning rate scheduler](StepLrScheduler).
+    ///
+    /// # Errors
+    ///
+    /// An error will be returned if any field is out of the acceptable range.
+    pub fn init(&self) -> Result<StepLrScheduler, String> {
+        // `initial_lr` and `gamma` are not checked because atypical values such as zero and
+        // negative values might be useful in some cases like debugging (e.g.,
+        // https://datascience.stackexchange.com/q/89518).
+        if self.step_size == 0 {
+            return Err("Step size must be greater than 0".into());
+        }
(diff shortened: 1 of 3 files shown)


crates/burn-core/src/lr_scheduler/step.rs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        Ok(StepLrScheduler {
+            init_lr: self.initial_lr,
+            step_size: self.step_size,
+            gamma: self.gamma,
+            iter_idx: -1,
+        })
+    }
+}
```
laggui, line 51: Is there a particular reason why the initialization uses `Result` and returns and `Err` instead of assertions/panics for the value checks?

Just asking because the pattern is different from the other schedulers.

Also, regarding `gamma` and `init_lr`, I think a warnings would be a good idea as you suggested.

laggui (commented):
Thanks for the addition!

Implementation looks good 🙂 

Just some minor comments

crates/burn-core/src/lr_scheduler/step.rs
towerpark: Thank you for your feedback.

I do have 2 reasons for using `Result`. One is for leaving callers the choice of how to handle the error. The other one is that returning `Result` seems acceptable since I find there are some other structs (not LR schedulers) doing so too. But neither of the reasons is a strong one and it does break the consistency, so if it's better to stick with the assertion way, I will fix that part.

As for the warning part, it seems that warnings should be printed by `log::warn!`, but I find a little issue with it. This method is a noop unless users explicitly create an application logger and call `.with_application_logger(None)` before building the learner. This is because the built-in logger is created in `LearnerBuilder::build`, which is unavailable when the LR scheduler is initialized.

I tried the MNIST example, however, and found that actually there were some log (`AdapterInfo` from cubecl_wgpu and `New execution plan 0` from burn_fusion) being discarded due to being logged at a stage too early, so letting users create the logger may not be that bad.

laggui: I actually agree for `Result`, just that we would have to change the behavior of existing schedulers for consistency 😅 so for this PR addition I would keep it consistent with others, and possibly make the change to `Result` for all schedulers in a follow-up PR.

Regarding the warning, good catch! I'll think about what we should do, but the warning at the optimizer init level should be valid. If we're not logging because it hasn't been initialized, that's another issue to fix somewhere else I believe.

towerpark: Yeah, one thing at a time is better. I will replace `Result` with an assertion and add warnings by `log::warn!` for initial LR and gamma, and leave the issues with returning value of `.init()` and logging to future PRs.

laggui: Perfect. We should open issues for the follow-up works 👍

towerpark:
I am sorry for the late update.

A fixup commit has been added to change the behavior of `StepLrSchedulerConfig::init`:
　* Make the method assert instead of returning `Result`
　* Add warning logs for atypical values of initial LR and gamma

Warning logs for initial LR and gamma are not getting test code, and were tested manually instead because there seems no straightforward way to do it. I have tried but couldn't come up with a solution. This is explained a little more in the source file.

laggui (approved):
No worries! Thanks for the addition 🙂

Issue #2361, opened 2024-10-12
Implement SegmentationMask in image_folder.rs (dataset::vision)


anthonytorlucci:
No open issues related to SegmentationMask.

### Feature description

Complete the implementation of SegmentationMask in image_folder.rs.

### Feature motivation

I'm currently working on a semantic segmentation example (simply UNet) and feel that implementing this feature will be a better solution than writing a new custom dataset.

### (Optional) Suggest a Solution

It appears to have been started or bookmarked for future work.

```rust
/// Annotation type for different tasks.
#[derive(Debug, Clone, PartialEq)]
pub enum Annotation {
    /// Image-level label.
    Label(usize),
    /// Multiple image-level labels.
    MultiLabel(Vec<usize>),
    /// Object bounding boxes.
    BoundingBoxes(Vec<BoundingBox>),
    /// Segmentation mask.
    SegmentationMask(SegmentationMask),
}
```

Needed to complete:

```rust
/// Raw annotation types.
#[derive(Deserialize, Serialize, Debug, Clone)]
enum AnnotationRaw {
    Label(String),
    MultiLabel(Vec<String>),
    // TODO: bounding boxes and segmentation mask
}
```

```rust
/// Parse the image annotation to the corresponding type.
fn parse_image_annotation(
    annotation: &AnnotationRaw,
    classes: &HashMap<String, usize>,
) -> Annotation {
    // TODO: add support for other annotations
    // - [ ] Object bounding boxes
    // - [ ] Segmentation mask
    // For now, only image classification labels are supported.

    // Map class string to label id
    match annotation {
        AnnotationRaw::Label(name) => Annotation::Label(*classes.get(name).unwrap()),
        AnnotationRaw::MultiLabel(names) => Annotation::MultiLabel(
            names
                .iter()
                .map(|name| *classes.get(name).unwrap())
                .collect(),
        ),
    }
}
```

Need to test Mapper and Loader. Additional code may be needed as well.

csking101:
Hi, I am new to open source, could you suggest how to proceed further?

anthonytorlucci:
@csking101 , I've just about completed the task and plan to submit a pull request in the next few days. To learn more about open source contributing, you could watch this ticket and follow the pull request (once requested). I'm sure the maintainers will have comments; then watch how the iterative process continues and ultimately leads to either good code being added to the project or a failure to meet expectations. Either way, it will be a learning experience. Thank you for joining the conversation and stay tuned!

[...]

PR #2426, opened 2024-10-26
Feat/2361 segmentation mask

anthonytorlucci:
### Checklist

_TODO_: Should add a new example to [dataset](https://burn.dev/burn-book/building-blocks/dataset.html) illustrating `new_segmentation_with_items()` method for the `ImageFolderDataset`.

### Related Issues/PRs

- [Issue 2361](https://github.com/tracel-ai/burn/issues/2361)

### Changes

Implemented necessary components for `SegmentationMask` to be used with `ImageFolderDataset`.

### Testing

Tests mimic the tests of the multilabel classification. New images have been added to the `tests/data` directory. These images are small 8 x 8 pixel images created with a python script.


laggui:
Thanks for taking the lead on this 😄 

I've been busy with the latest release but I should be able to review the changes this week!

anthonytorlucci:
> Thanks for taking the lead on this 😄
> 
> I've been busy with the latest release but I should be able to review the changes this week!

I'm happy to get it started and become a regular contributor in the future, but I don't think this PR is ready to merge. I'd be happy to take any suggestions and keep working on it including:
- additional example in burn book
- as part of a larger UNet example (model is the easy part; data is notoriously difficult!)

laggui:
> I'm happy to get it started and become a regular contributor in the future, but I don't think this PR is ready to merge.

No worries! In this case, I converted this to a draft PR and when you're ready just mark it as ready for review and/or ping me 🙂 


> I'd be happy to take any suggestions and keep working on it including:
> - additional example in burn book
> - as part of a larger UNet example (model is the easy part; data is notoriously difficult!)

That would be awesome! If you want to keep it simple for the current PR, I would start with the dataset implementation. Then, the examples can be added in follow-up PRs 👍

anthonytorlucci:
For completeness, the python code used to generate the new images and masks is below:

```python
"""
Synthetic Image Segmentation Generation
"""
from typing import List, Tuple
from pathlib import Path
import numpy as np
import matplotlib.pyplot as plt
import cv2

def save_image_to_text_file(image: np.ndarray, filename: Path) -> None:
    """Saves an image as a text file with pixel values."""
    with open(filename, 'w') as file:
        for row in image:
            file.write(' '.join(str(pixel) for pixel in row) + '\n')

def rgb_to_grayscale(rgb: Tuple[int, int, int]) -> int:
  """Converts a list of RGB values to its grayscale equivalent.

  Args:
    rgb: A list of RGB values in the range 0 - 255; [R, G, B].

  Returns:
    An integer representing the grayscale intensity.
  """
  return int(np.round(0.299 * rgb[0] + 0.587 * rgb[1] + 0.114 * rgb[2]))

def checkerboard_pattern(height:int, width:int, color1:Tuple[int, int, int], color2:Tuple[int, int, int]) -> (np.ndarray, np.ndarray):
    # Initialize the source image with zeros
    source_image = np.zeros((height, width, 3), dtype=np.uint8)

    # Initialize the mask with zeros
    mask = np.zeros((height, width, 3), dtype=np.uint8)
    m1 = 1  #rgb_to_grayscale(color1)
    m2 = 2  #rgb_to_grayscale(color2)
    print(m1, m2)
    # Let's create a checkerboard pattern
    for i in range(height):
        for j in range(width):
            if (i + j) % 2 == 0:
                source_image[i, j] = color1
                mask[i, j] = (m1, m1, m1)
            else:
                source_image[i, j] = color2
                mask[i, j] = (m2, m2, m2)
    return source_image, mask

def horizontal_stripes(height:int, width:int, color1:Tuple[int, int, int], color2:Tuple[int, int, int]) -> (np.ndarray, np.ndarray):
    # Initialize the source image with zeros
    source_image = np.zeros((height, width, 3), dtype=np.uint8)

    # Initialize the mask with zeros
    mask = np.zeros((height, width, 3), dtype=np.uint8)
    m1 = 1  #rgb_to_grayscale(color1)
    m2 = 2  #rgb_to_grayscale(color2)
    # Let's create horizontal stripes pattern
    for i in range(height):
        for j in range(width):
            if i % 2 == 0:
                source_image[i, j] = color1
                mask[i, j] = (m1, m1, m1)
            else:
                source_image[i, j] = color2
                mask[i, j] = (m2, m2, m2)
    return source_image, mask

def vertical_stripes(height:int, width:int, color1:Tuple[int, int, int], color2:Tuple[int, int, int]) -> (np.ndarray, np.ndarray):
    # Initialize the source image with zeros
    source_image = np.zeros((height, width, 3), dtype=np.uint8)

    # Initialize the mask with zeros
    mask = np.zeros((height, width, 3), dtype=np.uint8)
    m1 = 1  #rgb_to_grayscale(color1)
    m2 = 2  #rgb_to_grayscale(color2)
    # Let's create vertical stripes pattern
    for i in range(height):
        for j in range(width):
            if j % 2 == 0:
                source_image[i, j] = color1
                mask[i, j] = (m1, m1, m1)
            else:
                source_image[i, j] = color2
                mask[i, j] = (m2, m2, m2)

    return source_image, mask

def random_distribution_2colors(height:int, width:int, color1:Tuple[int, int, int], color2:Tuple[int, int, int], random_seed:int = 42) -> (np.ndarray, np.ndarray):
    # Initialize the source image with zeros
    source_image = np.zeros((height, width, 3), dtype=np.uint8)

    m1 = 1  #rgb_to_grayscale(color1)
    m2 = 2  #rgb_to_grayscale(color2)
    np.random.seed(random_seed)  # For reproducibility
    random_mask = np.random.choice([m1, m2], size=(height, width))
    for i in range(height):
        for j in range(width):
            if random_mask[i, j] == m1:
                source_image[i, j] = color1
            else:
                source_image[i, j] = color2
    mask = np.empty_like(source_image)
    mask[:, :, 0] = random_mask
    mask[:, :, 1] = random_mask
    mask[:, :, 2] = random_mask

    return source_image, mask

def random_distribution_3colors(height:int, width:int, color1:Tuple[int, int, int], color2:Tuple[int, int, int], color3:Tuple[int, int, int], random_seed:int = 42) -> (np.ndarray, np.ndarray):
    # Initialize the source image with zeros
    source_image = np.zeros((height, width, 3), dtype=np.uint8)

    np.random.seed(random_seed)  # For reproducibility
    m1 = 1  #rgb_to_grayscale(color1)
    m2 = 2  #rgb_to_grayscale(color2)
    m3 = 3  #rgb_to_grayscale(color3)
    random_mask = np.random.choice([m1, m2, m3], size=(height, width))
    for i in range(height):
        for j in range(width):
            if random_mask[i, j] == m1:
                source_image[i, j] = color1
            elif random_mask[i, j] == m2:
                source_image[i, j] = color2
            else:
                source_image[i, j] = color3
    mask = np.empty_like(source_image)
    mask[:, :, 0] = random_mask
    mask[:, :, 1] = random_mask
    mask[:, :, 2] = random_mask

    return source_image, mask


if __name__ == "__main__":
    # Define image dimensions
    IMAGE_HEIGHT, IMAGE_WIDTH = 8, 8

    # Define colors in RGB
    MAROON = (128,0,0)
    DARK_RED = (139,0,0)
    BROWN = (165,42,42)
    FIREBRICK = (178,34,34)
    CRIMSON = (220,20,60)
    RED = (255,0,0)
    TOMATO = (255,99,71)
    CORAL = (255,127,80)
    LIGHT_CORAL = (240,128,128)
    SALMON = (250,128,114)
    ORANGE_RED = (255,69,0)
    ORANGE = (255,165,0)
    YELLOW = (255,255,0)
    DARK_GREEN = (0,100,0)
    GREEN = (0,128,0)
    LIME = (0,255,0)
    SPRING_GREEN = (0,255,127)
    SEA_GREEN = (46,139,87)
    TEAL = (0,128,128)
    AQUA = (0,255,255)
    CYAN = (0,255,255)
    TURQUOISE = (64,224,208)
    AQUA_MARINE = (127,255,212)
    STEEL_BLUE = (70,130,180)
    MIDNIGHT_BLUE = (25,25,112)
    NAVY = (0,0,128)
    BLUE = (0,0,255)
    ROYAL_BLUE = (65,105,225)
    BLUE_VIOLET = (138,43,226)
    INDIGO = (75,0,130)
    SLATE_BLUE = (106,90,205)
    PURPLE = (128,0,128)
    THISTLE = (216,191,216)
    PLUM = (221,160,221)
    VIOLET = (238,130,238)
    MAGENTA = (255,0,255)
    ORCHID = (218,112,214)
    MEDIUM_VIOLET_RED = (199,21,133)
    PALE_VIOLET_RED = (219,112,147)
    PINK = (255,192,203)
    LEMON_CHIFFON = (255,250,205)
    SANDY_BROWN = (244,164,96)
    BURLY_WOOD = (222,184,135)
    MOCCASIN = (255,228,181)
    MISTY_ROSE = (255,228,225)
    LIGHT_STEEL_BLUE = (176,196,222)
    LAVENDER = (230,230,250)
    ALICE_BLUE = (240,248,255)
    HONEYDEW = (240,255,240)

    # Generate checkerboard pattern
    image_chkr, mask_chkr = checkerboard_pattern(
        height=IMAGE_HEIGHT,
        width=IMAGE_WIDTH,
        color1=CRIMSON,
        color2=AQUA
    )
    image_horz, mask_horz = horizontal_stripes(
        height=IMAGE_HEIGHT,
        width=IMAGE_WIDTH,
        color1=INDIGO,
        color2=CORAL
    )
    image_vert, mask_vert = vertical_stripes(
        height=IMAGE_HEIGHT,
        width=IMAGE_WIDTH,
        color1=PINK,
        color2=NAVY
    )
    image_rnd2, mask_rnd2 = random_distribution_2colors(
        height=IMAGE_HEIGHT,
        width=IMAGE_WIDTH,
        color1=MAGENTA,
        color2=CYAN,
        random_seed=42  # For reproducibility
    )
    image_rnd3, mask_rnd3 = random_distribution_3colors(
        height=IMAGE_HEIGHT,
        width=IMAGE_WIDTH,
        color1=TURQUOISE,
        color2=ORCHID,
        color3=BURLY_WOOD,
        random_seed=42  # For reproducibility
    )

    # # Optional: Print the array representations
    # print("Source Image Array (RGB):")
    # print(image)
    # print("\nMask Image Array:")
    # print(segmentation_mask)

    # ----- Save the results to disk ---
    results_path = Path(__file__).parent.parent.joinpath('results8x8')
    assert results_path.exists(), "The results directory does not exist. Please create it."
    # Save the image and mask to disk as PNG files
    cv2.imwrite(results_path.joinpath("image_checkerboard.png"), image_chkr)
    cv2.imwrite(results_path.joinpath("mask_checkerboard.png"), mask_chkr)
    cv2.imwrite(results_path.joinpath("image_random_2colors.png"), image_rnd2)
    cv2.imwrite(results_path.joinpath("mask_random_2colors.png"), mask_rnd2)
    cv2.imwrite(results_path.joinpath("image_random_3colors.png"), image_rnd3)
    cv2.imwrite(results_path.joinpath("mask_random_3colors.png"), mask_rnd3)

    # Save the image and mask array data to a column-delimited text file
    # save_image_to_text_file(image_chkr, results_path.joinpath("image_checkerboard.txt"))
    mask_chkr = mask_chkr[..., 0]  # Convert 3D mask to 2D
    save_image_to_text_file(mask_chkr, results_path.joinpath("mask_checkerboard.txt"))

    # save_image_to_text_file(image_rnd2, results_path.joinpath("image_random_2colors.txt"))
    mask_rnd2 = mask_rnd2[..., 0]  # Convert 3D mask to 2D
    save_image_to_text_file(mask_rnd2, results_path.joinpath("mask_random_2colors.txt"))

    # save_image_to_text_file(image_rnd3, results_path.joinpath("image_random_3colors.txt"))
    mask_rnd3 = mask_rnd3[..., 0]  # Convert 3D mask to 2D
    save_image_to_text_file(mask_rnd3, results_path.joinpath("mask_random_3colors.txt"))
```

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -104,7 +104,8 @@ pub struct ImageDatasetItem {
 enum AnnotationRaw {
     Label(String),
     MultiLabel(Vec<String>),
-    // TODO: bounding boxes and segmentation mask
+    SegmentationMask(Vec<String>),
```
laggui, line 107: I think it would be preferable to have the raw form of a segmentation mask point to the mask image path on this. That way, the masks would only be loaded when an item is fetched from the dataset.

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -401,6 +408,42 @@ impl ImageFolderDataset {
...
+    ///
+    /// * `items` - List of dataset items, each item represented by a tuple `(image path, labels)`.
+    /// * `classes` - Dataset class names.
+    ///
+    /// # Returns
+    /// A new dataset instance.
+    pub fn new_segmentation_with_items<P: AsRef<Path>, S: AsRef<str>>(
+        items: Vec<(P, SegmentationMask)>,
+        classes: &[S],
+    ) -> Result<Self, ImageLoaderError> {
```
laggui, line 423: With the suggested changes from my previous comment, it would make more sense (and probably be more practical from a user standpoint) to have a `new_segmentation` method that instead takes a list of `(image path, mask path)` pairs (so `Vec<(P, P)>`).

We still need the method to accept the class names, which can be used to map to a class id similar to what is already done. These identifiers would be use to map a pixel value to a class.

For added flexibility we could provide a way to have specific pixel values map to a class.

laggui (changes requested):
Sorry for the delayed review!

On the right path 🙂  But I have a couple comments of the current implementation.

I don't think it is desirable to have to provide the segmentation masks directly when loading a dataset. For image segmentation datasets with a lot of items (and larger image sizes), this won't fit into memory. I think we should instead take paths to the masks and parse them when accessing an item.

anthonytorlucci:
Suggested changes made with a few comments and questions:
- For the current implementation, we will only consider segmentation masks that are images and assume all channels are the same, i.e. a greyscale image where each pixel value corresponds to a single class.
- A new function was created `fn image_path_to_pixel_depth` which was copied from the `PathToImageDatasetItem::map()` function for input images. This has lead to **duplicated code** which I haven't addressed (not wanting to modify anything outside the segmentation scope).
- The last line of the method `pub fn new_segmentation_with_items(_)` calls `Self::with_items(_)` which creates an `InMemoryDataset`. As @laggui pointed out, this could be problematic for large images or large datasets. I'm not sure what the solution is here.
-  Finally, the image values appear to be in the reverse order of what I expected, specifically, each grouping of 3 values in  `Vec<PixelDepth>` is ordered (blue, green, red) rather than (red, green, blue). Has anyone seen this before?

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -76,7 +76,7 @@ pub enum Annotation {
 #[derive(Debug, Clone, PartialEq)]
 pub struct SegmentationMask {
     /// Segmentation mask.
-    pub mask: Vec<usize>,
+    pub mask: Vec<PixelDepth>,
```
laggui, line 79: For segmentation masks, I don't think it makes as much sense to have them stored as `PixelDepth` since the image should have integer values to represent a class index. We could keep that as a vec of usize.

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -129,15 +130,77 @@ struct PathToImageDatasetItem {
...
+            .into_rgb32f()
+            .iter()
+            .map(|&x| PixelDepth::F32(x))
+            .collect(),
+        ColorType::Rgba32F => image
+            .into_rgba32f()
+            .iter()
+            .map(|&x| PixelDepth::F32(x))
+            .collect(),
+        _ => panic!("Unrecognized image color type"),
```
laggui, line 189: Following the previous comment, I don't think we should support floating point segmentation masks. Images with alpha channels should probably not be supported as mask input either, so no `La8`, `La16`, `Rbga8`, `Rgba16` on top of that 🤔 

With these changes the function can probably be renamed since it will be specifically parsing segmentation mask input images to return the vec of class identifiers (i.e., indices). Probably want to make sure the returned vec is represents a single channel too since a segmentation mask is just 2D with each (x,y) pixel representing a class.

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -148,6 +211,20 @@ fn parse_image_annotation(
...
+            // each pixel in the first channel corresponds to a class.
+            // multi-channel image segmentation is not supported at this time.
+            Annotation::SegmentationMask(SegmentationMask {
+                mask: mask_image
+                    .into_iter()
+                    .enumerate()
+                    .filter(|(i, _)| i % 3 == 0)
+                    .map(|(_, pixel)| pixel)
+                    .collect(),
+            })
```
laggui, line 226: The filtering here will probably not be required given the suggested changes in the previous comment.

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -160,6 +237,7 @@ impl Mapper<ImageDatasetItemRaw, ImageDatasetItem> for PathToImageDatasetItem {
         let image = image::open(&item.image_path).unwrap();
 
         // Image as Vec<PixelDepth>
+        // NOTE: the following logic has been copied to a separate function to be used for Segmentation Masks as well
```
laggui, line 240: We'll be able to remove this comment 🙂

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -401,6 +479,36 @@ impl ImageFolderDataset {
         Self::with_items(items, classes)
     }
 
+    /// Create an image segmentation dataset with the specified items.
+    ///
+    /// # Arguments
+    ///
+    /// * `items` - List of dataset items, each item represented by a tuple `(image path, labels)`.
```
laggui, line 486: > each item represented by a tuple `(image path, labels)`

To be a bit more precise, it should say "annotation path" (or "segmentation mask path"?) instead of "labels".

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -401,6 +479,36 @@ impl ImageFolderDataset {
...
+    /// * `classes` - Dataset class names.
+    ///
+    /// # Returns
+    /// A new dataset instance.
+    pub fn new_segmentation_with_items<P: AsRef<Path>, S: AsRef<str>>(
+        items: Vec<(P, P)>,
+        classes: &[S],
+    ) -> Result<Self, ImageLoaderError> {
```
laggui, line 494: Awesome! That's exactly what I meant, so there should not be any memory issues and users are not forced to pre-load all of their segmentation masks into memory to create a dataset 👍

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -611,4 +720,128 @@ mod tests {
...
+        assert_eq!(
+            TEST_CHECKERBOARD_MASK_PATTERN
+                .iter()
+                .map(|&x| PixelDepth::U8(x))
+                .collect::<Vec<PixelDepth>>(),
+            image_path_to_vec_pixel_depth(&root.join("annotations").join("mask_checkerboard.png")),
+        );
+
+        // checkerboard image
+        // TODO: investigate why the channels appear to be reversed, i.e (blue, green, red) rather than (red, green, blue)
```
laggui, line 746: Is this still the case? Anyway the segmentation masks should be a 2D image `[H, W]` (single channel) so the test cases might be reworked with the other changes.

laggui (changes requested):
That's great! Most of my comments have been addressed, what's left is basically changes that are relevant to the questions you asked.

Also, to answer
> The last line of the method pub fn new_segmentation_with_items(_) calls Self::with_items(_) which creates an InMemoryDataset. As @laggui pointed out, this could be problematic for large images or large datasets. I'm not sure what the solution is here.

I don't think this is problematic now since the raw annotations just point to a path. So your `InMemoryDataset` will hold (image, annotation) path pairs only, which should not be much of an issue.

crates/burn-dataset/src/vision/image_folder.rs
anthonytorlucci: modified function name to `segmentation_mask_to_vec_usize` and only support luma8, luma16, rgb8, and rgb16. fixed in next commit.

crates/burn-dataset/src/vision/image_folder.rs
anthonytorlucci: filtering was removed and put in the `segmentation_mask_to_vec_usize`. fixed in the next commit.

crates/burn-dataset/src/vision/image_folder.rs
anthonytorlucci: removed in next commit.

crates/burn-dataset/src/vision/image_folder.rs
anthonytorlucci: fixed in the next commit.

anthonytorlucci:
It's still unclear to me at which point in the training workflow the image and annotation will be transformed to burn::tensor::Tensor<B, 4, Float> with shape [batch_size, 3, height, width] and burn::tensor::Tensor<B, 4, Int> with shape [batch_size, 1, height, width], respectively.

laggui:
> It's still unclear to me at which point in the training workflow the image and annotation will be transformed to burn::tensor::Tensor<B, 4, Float> with shape [batch_size, 3, height, width] and burn::tensor::Tensor<B, 4, Int> with shape [batch_size, 1, height, width], respectively.

That is usually implemented in the `Batcher` (see [this example](https://github.com/tracel-ai/burn/blob/main/examples/custom-image-dataset/src/data.rs#L52-L105) as reference). Though now that tensors are `Sync` it could also be done when retrieving a single item in the dataset, and the batcher would only concatenate the tensors.

laggui (approved):
Thanks for addressing all my comments 🙏

PR #2427, opened 2024-10-27
Add binary `remainder` for tensor

### LINKED ISSUE #1510: Implement `modulo` or `remainder` operation.

med1844:
### Checklist

  - Note: The burn book already documents remainder without specifying it's only `remainder_scalar`. This PR implements the expected behavior, so no book changes needed.

### Related Issues/PRs

Closes #1510 

### Changes

Added binary `remainder` to existing backends, including tests & autodiff.

The operation follows PyTorch's behavior where the remainder has the same sign as the divisor (e.g., 2.0 % -1.5 = -1.0), rather than Rust's default behavior (2.0 % -1.5 = 0.5).

### Testing

Added tests for `burn-candle`, `burn-ndarray`, `burn-autodiff`, and `burn-tensor`.

### Questions

- How to display `Instruction`
  https://github.com/tracel-ai/burn/pull/1597/files#diff-f71eb223e26e9a2944748c2e095167cb4ed9e7fcb6ccfda110c6c3f3869eab99R258-R259
  Currently, `remainder_scalar` is implemented as `((lhs % rhs) + rhs) % rhs`, whereas `remainder` is implemented as `lhs - (lhs / rhs).floor() * rhs`. I'm not sure which to display.
- Quantization tests
  I tried to add tests for quantized tensors (also for rounding functions, see #2372) but couldn't figure out how to generate quantized tensors exactly the same as existing tests, please shed some light on this!
- Update on segfault when running `run-checks.sh std`
  Turns out it's caused by `bridge::byte::tests::should_support_dual_byte_bridge` when trying to create a tensor on device2, which panics: `cubecl-wgpu/src/runtime.rs:297:17:
No adapter found for graphics API AutoGraphicsApi`. Not sure how to solve this issue.


crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -70,6 +70,14 @@ where
         Self::new(K::remainder_scalar::<E>(self.primitive, other))
     }
 
+    /// Applies element wise the remainder operation with a scalar.
+    ///
+    /// `y = x2 % x1`
+    #[allow(clippy::should_implement_trait)]
```
laggui, line 76: Any reason for this attribute?

laggui (commented):
Thanks for adding the operation 🙂 

Implementation looks good overall, just a few comments below.

To answer your questions...

> How to display Instruction

I don't believe this is required for your PR. All of the compiler stuff mentioned from the linked PR has been moved to cubecl anyway.

> Quantization tests
> I tried to add tests for quantized tensors (also for rounding functions, see https://github.com/tracel-ai/burn/pull/2372) but couldn't figure out how to generate quantized tensors exactly the same as existing tests, please shed some light on this!

The values were manually calculated to represent the float tensors with the correct quantized values and parameters. In your tests, you can reuse the existing values for the inputs and, as with the other tests, check for equality on the *dequantized* values. That way the tests are easy to validate.

Take the `exp()` test for example:
```rust
#[test]
fn should_support_exp_ops() {
    // Quantized [[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]
    // NOTE: we use affine quantization to reduce quantization errors since `exp()` amplifies the error
    let data = TensorData::quantized(
        vec![-128i8, -77, -26, 25, 76, 127],
        [2, 3],
        QuantizationStrategy::PerTensorAffineInt8(AffineQuantization::init(0.019607844, -128)),
    );
    let tensor = TestTensor::<2>::from_data(data, &Default::default());

    let output = tensor.exp();
    let expected = TensorData::from([[1.0, 2.71830, 7.3891], [20.0855, 54.5981, 148.4132]]);

    // Precision 1 to approximate de/quantization errors
    output
        .dequantize()
        .into_data()
        .assert_approx_eq(&expected, 1);
}
```

> Update on segfault when running run-checks.sh std
Turns out it's caused by bridge::byte::tests::should_support_dual_byte_bridge when trying to create a tensor on device2, which panics: cubecl-wgpu/src/runtime.rs:297:17: No adapter found for graphics API AutoGraphicsApi. Not sure how to solve this issue.

Ahhh this might just be due to your setup. For the multi-backend we use ndarray and wgpu as test backends, but wgpu might not be able to detect any device (possibly missing drivers) for your machine. Could be fixed if you install vulkan drivers, but otherwise you could ignore the failure locally and let it run on CI.

crates/burn-tensor/src/tensor/api/numeric.rs
med1844: Other methods in this fn needs them to avoid clippy warning like 

> method `add` can be confused for the standard trait method `std::ops::Add::add`

But I think that warning would only occur if we are using name `rem` in our case. So you are correct! Will remove this attribute for both `remainder` and `remainder_scalar`

med1844:
Thank you for the extra info on the quantization part! I have been able to produce the exact same output using python:

```python
import numpy as np
import numpy.typing as npt
from dataclasses import dataclass


@dataclass
class QuantizedTensor:
    quantized_tensor: npt.NDArray[np.int8]
    scale: float
    zero_point: float

    @classmethod
    def from_tensor(
        cls, in_tensor: npt.NDArray[np.float32] | list[float | int]
    ) -> "QuantizedTensor":
        dtype = np.int8
        tensor = np.array(in_tensor, dtype=np.float32)
        qmin, qmax = np.iinfo(dtype).min, np.iinfo(dtype).max
        vmin, vmax = tensor.min(), tensor.max()

        scale = (vmax - vmin) / (qmax - qmin)
        v_zero_point = -((vmax - vmin) / 2 + vmin)
        q_zero_point = (qmax - qmin) / 2 + qmin

        quantized_values = np.round(
            (tensor + v_zero_point) / scale + q_zero_point
        ).astype(dtype)
        quantized_values = np.clip(quantized_values, qmin, qmax)

        return cls(quantized_values, scale, v_zero_point / scale + q_zero_point)

    def to_tensor(self) -> npt.NDArray[np.float32]:
        return (self.quantized_tensor.astype(np.float32) - self.zero_point) * self.scale


if __name__ == "__main__":
    quant = QuantizedTensor.from_tensor(list(map(float, range(6))))
    print(quant)
    print(quant.to_tensor())
```

Which prints:

```
QuantizedTensor(quantized_tensor=array([-128,  -77,  -26,   25,   76,  127], dtype=int8), scale=0.0196078431372549, zero_point=-128.0)
[0. 1. 2. 3. 4. 5.]
```

I will use this to generate tests. This might also help other contributors in case in the future they are also adding new operators.

laggui:
If you keep the same quantized input values as one that is already generated for an existing test, you can assume that the dequantized output values should be approximately equal to the floating point results for the operation. So the reference can be kept as float and you don't necessarily need to dig into quantization, but it is a good learning experience 😄

crates/burn-ndarray/src/ops/base.rs
```diff
@@ -205,6 +209,15 @@ where
         NdArrayTensor { array }
     }
 
+    pub fn remainder(lhs: NdArrayTensor<E>, rhs: NdArrayTensor<E>) -> NdArrayTensor<E> {
+        // let array = ((lhs.array % rhs.array.clone()) + rhs.array.clone()) % rhs.array;
```
laggui, line 213: We can remove the dead code.

crates/burn-tensor/src/tests/quantization/ops/ceil.rs
```diff
@@ -0,0 +1,24 @@
+#[burn_tensor_testgen::testgen(q_ceil)]
+mod tests {
+    use super::*;
+    use burn_tensor::quantization::{AffineQuantization, QuantizationStrategy};
+    use burn_tensor::{Tensor, TensorData};
+
+    #[test]
+    fn should_support_ceil_ops() {
+        let data = TensorData::quantized(
```
laggui, line 9: Could you add the original float values represented by the quantized tensor data in a comment? It's useful to get a quick glance at the test values.

crates/burn-tensor/src/tests/quantization/ops/floor.rs
```diff
@@ -0,0 +1,23 @@
+#[burn_tensor_testgen::testgen(q_floor)]
+mod tests {
+    use super::*;
+    use burn_tensor::{Tensor, TensorData};
+
+    #[test]
+    fn should_support_floor_ops() {
+        let data = TensorData::quantized(
```
laggui, line 8: Same here regarding the original float values.

crates/burn-tensor/src/tests/quantization/ops/remainder.rs
```diff
@@ -8,6 +8,43 @@ mod tests {
...
+
+        let output = lhs.remainder(rhs);
+        let expected = TensorData::from([
+            0.983006477355957,
+            0.9960784912109375,
+            0.010457515716552734,
+            0.9882352948188782,
+            0.0026143789291381836,
+            2.9882352352142334,
+        ]);
```
laggui, line 39: The expected value should be `[1., 1., -0., 1., 0., 0.]`, no? For `[-3.0, -2.0, -1.0, 1.0, 2.0, 3.0]` % `[2.0, 3.0, 1.0, 2.0, 1.0, 3.0]`.

The result should be close to the same operation in floating point representation, minus some precision errors due to de/quantization.

The first 5 values have de/quantization errors but why is the last one off? Is it a quantization error?

crates/burn-tensor/src/tests/quantization/ops/remainder.rs
```diff
@@ -28,6 +65,42 @@ mod tests {
...
+        );
+
+        let output = lhs.remainder(rhs);
+        let expected = TensorData::from([
+            1.0007842779159546,
+            2.001568555831909,
+            0.05992525815963745,
+            0.06563401222229004,
+            0.2960371971130371,
+        ]);
```
laggui, line 95: The expected result should be `[1.0, 2.0, 0.0949, 0.0698, 0.2824]`, minus the de/quantization errors. I think it makes it more clear if you declare the result of the actual operation as if it was computed in floating point representation, and then the precision can be adjusted for the `assert_approx_eq`.

This comment probably applies for other expected results in following tests, but lmk what you think otherwise.

crates/burn-tensor/src/tests/quantization/ops/remainder.rs
```diff
@@ -62,11 +166,50 @@ mod tests {
...
+        );
+        let rhs = TestTensor::<1>::from_data(
+            // [1.4843, 1.135, 2.1563, 1.0862, 0.5034, 3.6587]
+            TensorData::quantized(
+                vec![-25i8, -49, 22, -52, -93, 127],
+                [6],
+                QuantizationStrategy::PerTensorAffineInt8(AffineQuantization::init(0.014363, -128)),
+            ),
+            &device,
+        );
```
laggui, line 193: If you used symmetric quantization instead of affine, maybe the results would be much closer to no remainder? 🤔

crates/burn-tensor/src/tests/quantization/ops/remainder.rs
med1844: It should be a quantization error because lhs and rhs have different value ranges. After dequantizing, the last element of lhs is 2.9882352, and 2.9960785 for rhs. The result is close to 3 (or 0 if modulo 3), so I think it's not that off.

laggui: Ahhh ok thanks for the details! Makes sense in this case

med1844:
I have updated the code according to the comments. Please review again, thank you.
For the MacOS CI failure, maybe we should find someone who has Apple hardware to replicate and figure out what happens; or we could just add to the book that the remainder operator may not work properly on Mac.

crates/burn-candle/src/lib.rs
```diff
@@ -86,6 +86,7 @@ mod tests {
     burn_tensor::testgen_mul!();
     burn_tensor::testgen_neg!();
     burn_tensor::testgen_permute!();
+    burn_tensor::testgen_remainder!();
```
laggui, line 89: Comment this out for now and add a note explaining the observed failure on macos.

crates/burn-candle/src/lib.rs
```diff
@@ -159,6 +160,7 @@ mod tests {
     burn_autodiff::testgen_ad_mul!();
     burn_autodiff::testgen_ad_neg!();
     burn_autodiff::testgen_ad_recip!();
+    burn_autodiff::testgen_ad_remainder!();
```
laggui, line 163: If the autodiff remainder also fails on macos then we could also comment this test out with a note.

laggui (commented):
LGTM! Thanks for addressing 🙏 

Regarding the macos test failure, it appears to always fail with candle so perhaps we could simply disable the test there specifically.

med1844:
I have commented out both of them and CI is finally happy. Please review, thank you!

laggui (approved):
Thank you 🎉

PR #2443, opened 2024-10-31
DefaultDevice should be an alias of BestAvailable

ArthurBrussee:
DefaultDevice was added as a new name of BestAvailable which is marked deprecated. They should be aliases of each other, not distrinct devices.


laggui (approved):

laggui:
Linking for ref: https://github.com/tracel-ai/cubecl/pull/210

PR #2454, opened 2024-11-02
Add more type support for burn-jit

wingertge:
### Checklist

### Related Issues/PRs

Implements the new types and parameterized testing from https://github.com/tracel-ai/cubecl/pull/207

### Changes

Adds support for `flex32`, `i8`, `i16`, `i64`, `u8`, `u16`, `u64` support to burn, parameterize tests to ensure all types work correctly

### Testing

New parameterized tests


crates/burn-cuda/src/lib.rs
```diff
@@ -17,6 +17,7 @@ mod tests {
     use burn_jit::JitBackend;
 
     pub type TestRuntime = cubecl::cuda::CudaRuntime;
+    pub use half::{bf16, f16};
 
-    burn_jit::testgen_all!();
+    burn_jit::testgen_all!(f32: [f16, bf16, f32], i32: [i8, i16, i32, i64]);
```
nathanielsimard, line 22: Syntax question, why `f32: [f16, bf16, f32]` instead of `float: [f16, bf16, f32]` ? Do we assert the values in f32 that is why `f32` is used?

crates/burn-jit/src/fusion/on_write/trace.rs
```diff
@@ -383,6 +421,18 @@ impl FuseOnWriteTrace {
...
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
+            SequenceArg::new(),
```
nathanielsimard, line 435: Not sure if there is a better way to do this 😅

crates/burn-jit/src/tests/mod.rs
```diff
@@ -111,28 +123,37 @@ macro_rules! testgen_jit {
...
 macro_rules! testgen_jit_fusion {
     () => {
+        $crate::testgen_jit_fusion!(f32: [], i32: []);
+    };
+    ($f_def:ident: [$($float:ident),*], $i_def:ident: [$($int:ident),*]) => {
         use super::*;
         use burn_jit::tests::{burn_autodiff, burn_fusion, burn_ndarray, burn_tensor};
 
         pub type TestBackend = burn_fusion::Fusion<JitBackend<TestRuntime, f32, i32>>;
+        pub type TestBackend2<F, I> = burn_fusion::Fusion<JitBackend<TestRuntime, F, I>>;
```
nathanielsimard, line 139: TestBackend2 is the one that can be injected with the precision right? Not sure if we could come up with a better name.

wingertge: I couldn't come up with a better name and it's only used in the macro to instantiate and override `TestBackend` with the correct types, so I left it for now. It's not used in the actual tests, just the macro.

crates/burn-cuda/src/lib.rs
wingertge: `f32` is the "default" type, since I didn't want to do matrix testing (it would blow up test time to much). So we run all relevant tests with each float type and `i32` as the int type,  and then we run all the tests that use `Int` with  `f32` as the float type and then each int type. I think I can actually just use the type set on the default test backend though, so I think I'll remove this.

wingertge: It was mostly a carry over from `cubecl` where we don't have a backend that defines the default type

crates/burn-jit/src/fusion/on_write/trace.rs
nathanielsimard: It's simply that each precision has its state. It would be cleaner to have a map `Map<ElemenType, State>`, instead of listing all posibilities. I'll think how I can refactor it, but not in this PR.

nathanielsimard (approved):

PR #2466, opened 2024-11-08
Add ROC AUC metric

vincentmasse:
### Checklist

### Related Issues/PRs

#544 

### Changes

Add ROC AUC metric for binary classification

### Testing

Unit tests for the new metric.

### DIFF
--- a/crates/burn-train/src/metric/roc_auc.rs
+++ b/crates/burn-train/src/metric/roc_auc.rs
@@ -1,0 +1,208 @@
+use core::f64;
+use core::marker::PhantomData;
+
+use super::state::{FormatOptions, NumericMetricState};
+use super::{MetricEntry, MetricMetadata};
+use crate::metric::{Metric, Numeric};
+use burn_core::tensor::backend::Backend;
+use burn_core::tensor::{ElementConversion, Int, Tensor};
+
+/// The ROC AUC metric for binary classification.
+#[derive(Default)]
+pub struct ROCAUCMetric<B: Backend> {
+    state: NumericMetricState,
+    _b: PhantomData<B>,
+}
+
+/// The [ROC AUC metric](ROCAUCMetric) input type.
+#[derive(new)]
+pub struct ROCAUCInput<B: Backend> {
+    outputs: Tensor<B, 2>,
+    targets: Tensor<B, 1, Int>,
+}
+
+impl<B: Backend> ROCAUCMetric<B> {
+    /// Creates the metric.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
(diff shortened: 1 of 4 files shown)


crates/burn-train/src/metric/roc_auc.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    fn binary_roc_auc(&self, probabilities: &Tensor<B, 1>, targets: &Tensor<B, 1, Int>) -> f64 {
+        let n = targets.dims()[0];
+
+        let n_pos = targets.clone().sum().into_scalar().elem::<u64>() as usize;
+
+        // Early return if we don't have both positive and negative samples
+        if n_pos == 0 || n_pos == n {
+            return 0.0;
+        }
```
laggui, line 38: Although the metric is technically invalid in such cases, I understand that returning 0 in such cases at least helps for aggregation.

Though I wonder if returning NaN would be better to immediately indicate that the result is undefined? Or perhaps logging a warning in addition 🤔

laggui (changes requested):
Thanks for the addition!

I have some minor comments, otherwise great start 🙂

crates/burn-train/src/metric/roc_auc.rs
vincentmasse: I updated the code to return NaN, but I'm not sure if this is the correct approach, as it depends on the standard used for similar situations elsewhere in the code.

crates/burn-train/src/metric/roc_auc.rs
```diff
@@ -0,0 +1,208 @@
...
+pub struct ROCAUCMetric<B: Backend> {
+    state: NumericMetricState,
+    _b: PhantomData<B>,
+}
+
+/// The [ROC AUC metric](ROCAUCMetric) input type.
+#[derive(new)]
+pub struct ROCAUCInput<B: Backend> {
+    outputs: Tensor<B, 2>,
+    targets: Tensor<B, 1, Int>,
```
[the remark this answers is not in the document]
vincentmasse: I was planning implementing this metric for multiclass classification. For multi-label classification, I plan to use the Label Ranking Average Precision (LRAP) metric instead.

crates/burn-train/src/metric/roc_auc.rs
laggui: I don't think we have this case in any other metric right now, and both approaches could work. That's why I'm wondering how we should handle such instances.

If we return 0 in this condition, I would additionally `log::warn!(...)` a message to at least leave a trace that the metric could not be computed for the batch. This might be a better option than returning NaN.

vincentmasse: Updated

laggui (approved):

PR #2467, opened 2024-11-09
[Fix] Prevent various OOB accesses and discontiguous buffer bugs

wingertge:
### Checklist

### Changes

Fixes OOB bugs in the reduce kernels and `col2im`.
Fixes bugs when running scalar ops with discontiguous buffers (buffers with empty memory regions created by things like `slice`) by making them contiguous first (this is almost always faster than strided access).

### Testing

All tests now pass with no weird values or segfaults

### DIFF
--- a/crates/burn-fusion/src/ops/int.rs
+++ b/crates/burn-fusion/src/ops/int.rs
@@ -1694,9 +1694,9 @@ impl<B: FusionBackend> IntTensorOps<Self> for Fusion<B> {
 
         impl<B: FusionBackend> Operation<B::FusionRuntime> for ExpandOps<B> {
             fn execute(self: Box<Self>, handles: &mut HandleContainer<B::Handle>) {
-                let input = handles.get_bool_tensor::<B>(&self.desc.input);
-                let output = B::bool_expand(input, self.desc.shape.into());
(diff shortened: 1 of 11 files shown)


wingertge:
Windows are failing with a file permission error, seems like a CI bug

crates/burn-fusion/src/ops/int.rs
```diff
@@ -1694,9 +1694,9 @@ impl<B: FusionBackend> IntTensorOps<Self> for Fusion<B> {
 
         impl<B: FusionBackend> Operation<B::FusionRuntime> for ExpandOps<B> {
             fn execute(self: Box<Self>, handles: &mut HandleContainer<B::Handle>) {
-                let input = handles.get_bool_tensor::<B>(&self.desc.input);
-                let output = B::bool_expand(input, self.desc.shape.into());
-                handles.register_bool_tensor::<B>(&self.desc.out.id, output);
+                let input = handles.get_int_tensor::<B>(&self.desc.input);
+                let output = B::int_expand(input, self.desc.shape.into());
+                handles.register_int_tensor::<B>(&self.desc.out.id, output);
```
laggui, line 1699: Good catch 😅

laggui (approved):
Awesome, ty!🙏

PR #2477, opened 2024-11-11
Replace crates by dependi

vincentmasse:
### Checklist

### Related Issues/PRs

Crates is an archived project.

### Changes

Using dependi instead of crates

### Testing

N/A


vincentmasse:
> Thanks for flagging and addressing the required change!
> 
> Looks like we also point to crates in the [contributor book](https://github.com/tracel-ai/burn/blob/main/contributor-book/src/getting-started/configuring-your-editor.md?plain=1#L14). Mind changing that as well? 🙏

Done

laggui (approved):

PR #2479, opened 2024-11-11
Refactor quantization tensor data representation

laggui:
### Checklist

### Changes

`DType::QFloat` no longer contains the `QuantizationStrategy` with quantization parameters as it is not known at compile-time. Instead, use the existing `QuantizationScheme` enum which only indicates the method and dtype.

Also changed the representation to pack the quantized values (as was previously done in `burn-jit` exclusively) and include the quantization parameters as the bytes.

### Testing

Added unit tests for packed/unpacked quantization parameters.

### DIFF
--- a/crates/burn-jit/src/ops/qtensor.rs
+++ b/crates/burn-jit/src/ops/qtensor.rs
@@ -4,10 +4,9 @@ use alloc::vec::Vec;
 use burn_tensor::{
     ops::{FloatTensor, IntTensor, QTensorOps, QuantizedTensor},
     quantization::{
-        QTensorPrimitive, QuantizationParametersPrimitive, QuantizationScheme,
-        QuantizationStrategy, QuantizationType,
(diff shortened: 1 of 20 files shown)


crates/burn-jit/src/ops/qtensor.rs
```diff
@@ -51,27 +35,21 @@ where
...
-                        qtensor: packed_tensor(pack_i8s_to_u32s(&data), data.shape, device),
-                        scheme: strategy.scheme(),
+                        qtensor: packed_tensor(data.values_as_bytes(), data.shape.clone(), device),
+                        scheme,
                         qparams: JitQuantizationParameters::new(
-                            q.scale.elem(),
-                            Some(q.offset.elem()),
+                            qparams.scale,
+                            qparams.offset,
                             device,
```
nathanielsimard, line 49: Do we still need to have the scale and the offset in the representation?

PR #2483, opened 2024-11-12
Add float cast tensor op

laggui:
### Checklist

### Changes

Add `tensor.cast(dtype)` op for floating point casting to different precision types.

Currently implemented for `candle` and `tch`. Had to remove the element type generic on both tensor primitives for this change.

Will be implemented in a follow-up PR for: `burn-ndarray`, `burn-jit`, `burn-fusion` and `burn-router`.

**Important:** handling different floating point precision types in operations between multiple tensors might require automatic type promotion based on a selected strategy (e.g., cast to the higher precision type like [torch](https://pytorch.org/docs/stable/tensor_attributes.html#type-promotion-doc)). This is not in the current scope, instead backends can panic with a dtype mismatch.

### Testing

Added a unit test.


nathanielsimard (approved):
LGTM

Issue #1970, opened 2024-07-04
Too many bindings of type StorageBuffers in Stage ShaderStages(COMPUTE)


ianmarmour:
**Describe the bug**

When attempting to perform any training of my model that with a batch size > 1 there are too many bindings of type `StorageBuffer`. This leads to a panic in WGPU and causes the training process to crash. It appears that the correct limits for the adapter are being inferred but potentially not respected by fusion if fusion is disabled this error isn't encountered.

**To Reproduce**

1. Use WGPU as the device type on a Mac.
2. .... unsure... 
3. `cargo run` to perform training with batch size > 1
4. On the second batch of items the panic will occur.

**Expected behavior**

Proper allocation of the correct number of maximum available `StorageBuffers`.

**Screenshots**

```
2024-07-04T01:51:02.976015Z  INFO burn_fusion::stream::store::base: New execution plan 76 - Operations: 1 - Triggers 1    
2024-07-04T01:51:02.976063Z  INFO burn_fusion::stream::store::base: New execution plan 77 - Operations: 3 - Triggers 1    
2024-07-04T01:51:02.985188Z  INFO burn_jit::fusion::kernel: Compiling ... "mri0pvnx16y16z16g0vs7ubdfg"    
2024-07-04T01:51:03.194143Z  INFO burn_jit::fusion::kernel: Compiling ... "mri0pvnx16y16z1675o2aql5m4"    
2024-07-04T01:51:03.404701Z  INFO burn_compute::tune::tuner: Fastest result burn_jit::fusion::kernel::AutotunableKernel<burn_wgpu::runtime::WgpuRuntime>-Fusion ElemWise - num_operations: 4 shape: [32, 64, 2, 64]    
2024-07-04T01:51:03.428641Z  INFO burn_jit::fusion::kernel: Compiling ... "mi0o0ri0pvnx16y16z16g0vs7ubdfg"    
2024-07-04T01:51:03.515667Z  INFO burn_fusion::stream::store::base: New execution plan 78 - Operations: 99 - Triggers 1    
[...]

PR #2486, opened 2024-11-13
Custom fusion

ArthurBrussee:
### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/1970

### Changes
Expose enough information to allow custom operations to use fusing. Also expose a function to resolve a tensor from a fusion backend.

Draft until https://github.com/tracel-ai/cubecl/pull/256 lands & some other bits are cleaned up.

This also fixes https://github.com/tracel-ai/burn/issues/1970 by bailing on fusing when going over max bindings.

### DIFF
--- a/crates/burn-fusion/src/client/base.rs
+++ b/crates/burn-fusion/src/client/base.rs
@@ -67,6 +67,18 @@ where
     where
         B: FusionBackend<FusionRuntime = R>;
     /// Change the client of the given float tensor.
+    fn resolve_tensor_float<B>(&self, tensor: FusionTensor<R>) -> B::FloatTensorPrimitive
+    where
+        B: FusionBackend<FusionRuntime = R>;
+    /// Change the client of the given float tensor.
+    fn resolve_tensor_int<B>(&self, tensor: FusionTensor<R>) -> B::IntTensorPrimitive
+    where
+        B: FusionBackend<FusionRuntime = R>;
+    /// Change the client of the given float tensor.
+    fn resolve_tensor_bool<B>(&self, tensor: FusionTensor<R>) -> B::BoolTensorPrimitive
+    where
+        B: FusionBackend<FusionRuntime = R>;
+    /// Change the client of the given float tensor.
     fn change_client_float<B>(
         &self,
         tensor: TensorDescription,
--- a/crates/burn-fusion/src/client/base.rs
+++ b/crates/burn-fusion/src/client/base.rs
@@ -66,15 +66,15 @@ where
     ) -> impl Future<Output = TensorData> + Send
     where
         B: FusionBackend<FusionRuntime = R>;
-    /// Change the client of the given float tensor.
+    /// Resolve the given float tensor to a primitive tensor.
     fn resolve_tensor_float<B>(&self, tensor: FusionTensor<R>) -> B::FloatTensorPrimitive
     where
         B: FusionBackend<FusionRuntime = R>;
-    /// Change the client of the given float tensor.
+    /// Resolve the given int tensor to a primitive tensor.
     fn resolve_tensor_int<B>(&self, tensor: FusionTensor<R>) -> B::IntTensorPrimitive
     where
         B: FusionBackend<FusionRuntime = R>;
-    /// Change the client of the given float tensor.
+    /// Resolve the given bool tensor to a primitive tensor.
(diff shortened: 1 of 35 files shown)


crates/burn-fusion/src/client/base.rs
```diff
@@ -67,6 +67,18 @@ where
...
+    where
+        B: FusionBackend<FusionRuntime = R>;
+    /// Change the client of the given float tensor.
+    fn resolve_tensor_int<B>(&self, tensor: FusionTensor<R>) -> B::IntTensorPrimitive
+    where
+        B: FusionBackend<FusionRuntime = R>;
+    /// Change the client of the given float tensor.
+    fn resolve_tensor_bool<B>(&self, tensor: FusionTensor<R>) -> B::BoolTensorPrimitive
+    where
+        B: FusionBackend<FusionRuntime = R>;
```
nathanielsimard, line 80: Docs is not up to date.

crates/burn-jit/src/fusion/on_write/builder.rs
```diff
@@ -16,10 +16,40 @@ use cubecl::ir::Elem;
...
+    }
+
+    fn register(&mut self, build: impl FnOnce(&mut FuseOnWriteTraceBuilder)) -> bool {
+        let mut cloned = self.builder.clone();
+        build(&mut cloned);
+        if cloned.estimate_bindings() > self.max_bindings {
+            return false;
+        }
+        self.builder = cloned;
+        true
```
nathanielsimard, line 47: cloning every time is a bit wasteful. I would only do it if `builder_current_state.estimate_bindings() + new_op.nodes() > self.max_bindings`

crates/burn-jit/src/fusion/on_write/builder.rs
```diff
@@ -28,6 +58,13 @@ impl OptimizationBuilder<FuseOnWriteTrace> for FuseOnWriteBuilder {
             return;
         }
 
+        // TODO: If tensors are re-used, we should count them only once.
+        self.num_tensors += op.nodes().len();
+        if self.num_tensors >= 10 {
+            self.status = OptimizationStatus::Closed;
+            return;
+        }
```
nathanielsimard, line 66: It should use `estimate_bindings` I think, and 10 should be replaced by the value in the device's hardware properties.

crates/burn-tensor/src/repr/operation.rs
```diff
@@ -16,6 +16,21 @@ use crate::{
 
 use super::{QuantizationParametersDescription, QuantizedTensorDescription};
 
+/// Custom operation in fusion stream, declaring it's inputs and outputs.
+#[derive(Clone, Debug, Hash, PartialEq, Serialize, Deserialize)]
+pub struct CustomOpDescription {
+    /// Input tensors used in this the custom operation.
```
nathanielsimard, line 22: I would add an `id` field (a string for now). And maybe a `scalars` field too, but it's not important to have it now.

crates/burn-router/src/runner.rs
```diff
@@ -1202,6 +1202,9 @@ where
                     handles.register_float_tensor::<B>(&desc.out.id, output);
                 }
             },
+            OperationDescription::Custom(_) => {
+                panic!("Can't execute custom operation here")
```
nathanielsimard, line 1206: At some point we should resolve a function based on the custom description id and call it with the handles and the description as input.

crates/burn-jit/src/fusion/on_write/builder.rs
ArthurBrussee: That doesn't really work - we don't know how many nodes someone is adding in the build function. And we can't easily keep track of it as adding them mutates the builder.

estimate_bindings() also seems more expensive than a clone in the first place?

I've at least added a fast path for the first operation which should be common. If that's not enough might need to rethink how these operations are added so we can count them more easily.

crates/burn-jit/src/fusion/on_write/builder.rs
ArthurBrussee: Woops this was leftover dummy stuff

ArthurBrussee:
I've also added a fix for a deadlock in fusion on wasm in this PR now. It might or might not make things run a bit smoothly on other platforms too, as it was incorrectly holding the server lock while waiting for a read.

nathanielsimard (approved):

Issue #2487, opened 2024-11-14
Memory Leak and Performance Degradation in `GraphMemoryManagement` due to Unreleased Nodes


jnamika:
**Describe the bug**
There appears to be a node leak in the `GraphMemoryManagement`. By adding a debug line at the end of `GraphMemoryManagement::free_unavailable_nodes()` to display `self.nodes.len()`, we can observe that the size of self.nodes increases by one with each call.

**To Reproduce**
The issue can be reproduced regardless of the backend. For example, running `burn/examples/mnist` demonstrates the memory leak.

**Expected behavior**
`GraphMemoryManagement` should delete unneeded nodes as expected, preventing any memory leaks.

**Additional context**
The `GraphMemoryManagement::clear_unused_roots()` method performs a linear search over `self.nodes`, so as training progresses and leaked nodes accumulate, the overall performance gradually degrades.
<img width="389" alt="grafana" src="https://github.com/user-attachments/assets/b82db3b6-b1b9-4a16-80c1-a8061907cd48">

AsherJingkongChen:
Is it relevant to #2042?

jnamika:
Thank you for pointing this out! This issue may indeed be related to #2042, but it is definitely connected to #2333. In fact, the `clear_unused_roots()` function was introduced in pull request #2347 to address the memory leak reported in #2333. This function does introduce a linear scan over `self.nodes`, leading to a performance regression over time.

However, the root cause here seems to be that `self.nodes` continues to grow due to nodes not being released, which is an issue that predates the introduction of `clear_unused_roots()`. So ideally, the regression should not be addressed by reverting the changes from #2347. Instead, we should focus on fixing the underlying node management problem to ensure that `self.nodes` is properly pruned, preventing both memory leaks and performance degradation.

Let me know if I can provide additional details!

AsherJingkongChen:
Could you provide the minimal example? I'd like to track the issue, but I can't solve it since I am not familiar with GMM.

jnamika:
Before I could create a minimal example, I found a solution and have submitted a pull request with the fix. 
It’s a very small change; could you review it when convenient? 
Thank you!

PR #2488, opened 2024-11-14
verifying parent nodes' existence

jnamika:
Fix https://github.com/tracel-ai/burn/issues/2487

Since the loop over self.nodes.iter() in GraphMemoryManagement::clear_unused_roots() does not process nodes in child-to-parent order, if a node's parents are removed first, that node will not be deleted and will remain. To address this, I added a check to verify if all parents are absent from self.nodes. If none of the parents exist in self.nodes, the node is treated as having no parents and is marked for deletion.

### DIFF
--- a/crates/burn-autodiff/src/runtime/memory_management.rs
+++ b/crates/burn-autodiff/src/runtime/memory_management.rs
@@ -84,8 +84,8 @@ impl GraphMemoryManagement {
         for (id, parents) in self.nodes.iter() {
             let is_useful = matches!(self.statuses.get(id), Some(NodeMemoryStatus::Useful));
 
-            let parents_absent =
-                parents.is_empty() || parents.iter().all(|p| !self.nodes.contains_key(p));
+            // Check if parents are either empty or absent from self.nodes
+            let parents_absent = parents.iter().all(|p| !self.nodes.contains_key(p));
 
             if !is_useful && Arc::strong_count(id) == 1 && parents_absent {
                 to_delete.push(*id.as_ref())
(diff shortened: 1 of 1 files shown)


crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -84,7 +84,10 @@ impl GraphMemoryManagement {
         for (id, parents) in self.nodes.iter() {
             let is_useful = matches!(self.statuses.get(id), Some(NodeMemoryStatus::Useful));
 
-            if !is_useful && Arc::strong_count(id) == 1 && parents.is_empty() {
+            let parents_absent =
+                parents.is_empty() || parents.iter().all(|p| !self.nodes.contains_key(p));
```
AsherJingkongChen, line 88: If `parents.is_empty() == true`, `parents.iter().all(...) == true`. If the second expression works, the first expression i.e. `is_empty` has no effect.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all

AsherJingkongChen (changes requested):

AsherJingkongChen (approved):
Hi, I used my project as "my" example program to test your changes.

I used Apple's Instruments.app to profile heap memory every 10 seconds, below are the results:

- c7233bfbce53e24cef524b6be1c670839d8f1862: 9 of 11 snapshots has leaks
- 5c91918bcb8db70d54a4dd7c5f0cee9c58093233: 0 of 12 snapshots has leaks

It indicates that your changes **work** for me.

As for MNIST example, I don't know why I can't run it with:

```shell
./target/debug/examples/mnist
```

It looks like I messed up `$DYLD_LIBRARY_PATH` again.

laggui (approved):
Thanks for both flagging the issue and fixing it 🎉 

@AsherJingkongChen thank you also for validating

Issue #2233, opened 2024-09-01
Can the pretty cool TUI remain on my shell even after training finished?


kingwingfly:
### Feature description

I'd prefer the TUI remains on my shell after training finished. 

### Feature motivation

So that I can observe the scatter diagram better. 

### (Optional) Suggest a Solution
```rust
let learner = LearnerBuilder::new(artifact_dir)
        .metric_train_numeric(LossMetric::new())
        ...
        .auto_quit(false)    // <- allow this
        .build(
            ...
        );
```

laggui:
This was added with #2489.

Current usage requires the user to be explicit about the TUI renderer usage:

```rust
let learner = LearnerBuilder::new(ARTIFACT_DIR)
    // ...
    .devices(vec![device.clone()])
    .num_epochs(config.num_epochs)
    .summary();

let tui = TuiMetricsRenderer::new(learner.interrupter(), None).persistent();
let learner = learner
    .renderer(tui)
    .build(Model::new(&device), config.optimizer.init(), 1e-4);
```

While it "breaks" the builder pattern, I think it's perfectly fine for now. If the training interrupter is provided another way this could be improved to not "break" the builder pattern.

PR #2489, opened 2024-11-14
Add option to request manual quit on tui

vincentmasse:
### Checklist

### Related Issues/PRs

#2233 

### Changes

Add option to learner builder to wait for user to quit the renderer.

### Testing

Manual test, TUI waits for q to close.

### DIFF
--- a/crates/burn-train/src/renderer/tui/renderer.rs
+++ b/crates/burn-train/src/renderer/tui/renderer.rs
@@ -44,7 +44,7 @@ pub struct TuiMetricsRenderer {
     interuptor: TrainingInterrupter,
     popup: PopupState,
     previous_panic_hook: Option<Arc<PanicHook>>,
-    manual_quit: bool,
+    persistent: bool,
 }
 
 impl MetricsRenderer for TuiMetricsRenderer {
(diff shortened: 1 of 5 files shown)


vincentmasse:
I wasn't sure about the best way to implement this, particularly if adding it to the builder was the right approach. Let me know if you have any suggestions.

vincentmasse:
image
I got this error while running run_checks. I opened the PR since it's not linked with the changes I made. I suspect it's because the intel graphics used isn't implementing cooperative matrix multiplication.

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -116,9 +117,15 @@ impl TuiMetricsRenderer {
             interuptor,
             popup: PopupState::Empty,
             previous_panic_hook: Some(previous_panic_hook),
+            manual_quit: false,
         }
     }
 
+    /// Enable manual quit after training.
+    pub fn enable_manual_quit(&mut self) {
```
laggui, line 125: I think I would rename the method and take ownership to return the renderer. That way, we can keep the flow similar to the builder when declaring the renderer, e.g. `TuiMetricsRenderer::new(...).persistent()`

```rust
pub fn persistent(mut self) -> Self {
    self.manual_quit = true;
    self
}
```

the `manual_quit` field could also be renamed but not too critical.

laggui (commented):
Ah that's pretty good!

I think we could make one minimal change, after which the usage would look something like this:

```rust
let learner = LearnerBuilder::new(ARTIFACT_DIR)
    // ...
    .devices(vec![device.clone()])
    .num_epochs(config.num_epochs)
    .summary();

let tui = TuiMetricsRenderer::new(learner.interrupter(), None).persistent();
let learner = learner
    .renderer(tui)
    .build(Model::new(&device), config.optimizer.init(), 1e-4);
```

While it "breaks" the builder pattern, I think it's perfectly fine for now. We could eventually improve this but I don't think it's a critical issue atm.

laggui (approved):
Thanks 🎉

Issue #2440, opened 2024-10-30
Implement a Split operator


Luni-4:
Issues which make use of the `Split` operator:  https://github.com/tracel-ai/burn/issues/675 

### Feature description

Implement the [Split](https://onnx.ai/onnx/operators/onnx__Split.html) operator as:

- [x] Burn operator
- [ ] `burn-import` operator so that ONNX files containing it can be parsed 

### Feature motivation

This operator is used in some neural networks, so having it in `burn` would allow to access a wider set of neural networks.

Luni-4:
37822fd closes this issue

Thanks to @agelas, @laggui, and @jameshiew for their fantastic work! :clap: :fire:

PR #2490, opened 2024-11-14
Feat/Split Operator

agelas:
### Checklist

### Related Issues/PRs

#2440

### Changes

Adds support for the split operation. The ONNX -> Burn conversion will be part of a separate PR.

### Testing

Added tests under `burn-tensor`.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1271,7 +1271,7 @@ where
     /// ```
     pub fn split(self, split_size: usize, dim: usize) -> Vec<Self> {
         check!(TensorCheck::split::<D>(
-            self.shape().dims[dim],
+            self.shape().dims.as_ref(),
             split_size,
             dim
         ));
@@ -1314,7 +1314,7 @@ where
     /// ```
     pub fn split_with_sizes(self, split_sizes: Vec<usize>, dim: usize) -> Vec<Self> {
         check!(TensorCheck::split_with_sizes::<D>(
-            self.shape().dims[dim],
+            self.shape().dims.as_ref(),
             &split_sizes,
             dim
         ));
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1191,7 +1191,7 @@ where
         Self::new(narrow::<B, K>(self.primitive, dim, start, length))
     }
 
-    /// Attempts to split the tensor along the given dimension into chunks.
+    /// Attempts to split the tensor into a specified number of chunks along a given dimension.
     /// May return less chunks than requested if the tensor size is not divisible by the number of chunks.
     ///
     /// When the given dimension is evenly divisible by the number of chunks, the chunks will be of equal size.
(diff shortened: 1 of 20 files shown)


crates/burn-tensor/src/tests/ops/chunk.rs
```diff
@@ -4,6 +4,7 @@ mod tests {
     use alloc::vec::Vec;
     use burn_tensor::{Int, Shape, Tensor, TensorData};
 
+    #[test]
```
agelas, line 7: I think this was accidentally left out in #998

crates/burn-tensor/src/tests/mod.rs
```diff
@@ -107,6 +107,7 @@ macro_rules! testgen_quantization {
         burn_tensor::testgen_q_sin!();
         burn_tensor::testgen_q_slice!();
         burn_tensor::testgen_q_sort_argsort!();
+        // burn_tensor::testgen_q_split!();
```
agelas, line 110: @antimora @louisfd It looks like there's some sort of nuance here that I'm a bit out of the loop on- do I need to do something extra for the quantized set of tests/do I even need quantized tests here?

crates/burn-tensor/src/tensor/api/split.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    let mut start = 0;
+    while start < size {
+        let length = usize::min(split_size, size - start);
+        tensors.push(narrow::<B, K>(tensor.clone(), dim, start, length));
+        start += length;
+    }
+
+    tensors
+}
```
nathanielsimard, line 41: I think this function will likely by the fastest in most cases, since there is not kernel to execute, only metadata to update.

crates/burn-tensor/src/tests/mod.rs
louisfd: Adding quantized tests is not well documented at the moment, so you're not expected to do it. 
@laggui will work on making it more straightforward and he'll be able to adapt your non-quantized tests.

louisfd (approved):
LGTM

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2111,10 +2196,58 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
...
+    /// A vector of tensors.
+    ///
+    /// # Remarks
+    /// This is a low-level function used internally by the library to call different backend functions
+    /// with static dispatch. It is not designed for direct usage by users, and not recommended to import
+    /// or use this function directly.
+    ///
+    /// To split a tensor, users should prefer the [Tensor::split](Tensor::split) function,
+    /// which is more high-level and designed for public use.
+    fn split(tensor: Self::Primitive, split_size: usize, dim: usize) -> Vec<Self::Primitive>;
```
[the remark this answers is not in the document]
agelas: Correct- and then `split_with_sizes` goes a step further and lets you specify the size of each chunk.

nathanielsimard (approved):

Luni-4:
Thanks a lot @agelas for your implementation! And thank you, @louisfd and @nathanielsimard for your reviews!

PR #2492, opened 2024-11-15
Fix/burn remote deadlock + burn fusion draining

nathanielsimard:
Improved a lot of things related to streams and training a model on multiple threads.

### DIFF
--- a/crates/burn-hip/src/lib.rs
+++ b/crates/burn-hip/src/lib.rs
@@ -24,9 +24,9 @@ pub type Hip<F = f32, I = i32> = burn_fusion::Fusion<JitBackend<HipRuntime, F, I
 // #[cfg(test)]
 // mod tests {
 //     use burn_jit::JitBackend;
-// 
+//
 //     pub type TestRuntime = cubecl::hip::HipRuntime;
 //     pub use half::{bf16, f16};
-// 
+//
 //     burn_jit::testgen_all!();
 // }
--- a/crates/burn-hip/src/lib.rs
+++ b/crates/burn-hip/src/lib.rs
@@ -18,13 +18,15 @@ pub type Hip<F = f32, I = i32> = JitBackend<HipRuntime, F, I>;
 #[cfg(feature = "fusion")]
 pub type Hip<F = f32, I = i32> = burn_fusion::Fusion<JitBackend<HipRuntime, F, I>>;
 
-#[cfg(target_os = "linux")]
-#[cfg(test)]
-mod tests {
-    use burn_jit::JitBackend;
-
-    pub type TestRuntime = cubecl::hip::HipRuntime;
-    pub use half::{bf16, f16};
-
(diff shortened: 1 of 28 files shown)


crates/burn-hip/src/lib.rs
```diff
@@ -18,13 +18,15 @@ pub type Hip<F = f32, I = i32> = JitBackend<HipRuntime, F, I>;
...
+// #[cfg(target_os = "linux")]
+// #[cfg(test)]
+// mod tests {
+//     use burn_jit::JitBackend;
+//
+//     pub type TestRuntime = cubecl::hip::HipRuntime;
+//     pub use half::{bf16, f16};
+//
+//     burn_jit::testgen_all!();
+// }
```
laggui, line 32: Not sure if this is meant to be commented out just locally or actually committed to fix some HIP stuff?

crates/burn-remote/src/client/worker.rs
```diff
@@ -103,7 +103,9 @@ impl ClientWorker {
                     match msg {
                         Message::Binary(bytes) => {
                             let response: TaskResponse = rmp_serde::from_slice(&bytes).expect("Can deserialize messages from the websocket.");
+                            log::info!("On response lock");
                             let mut state = state_ws.lock().await;
+                            log::info!("On response");
```
laggui, line 108: Debugging left overs or actual logs?

crates/burn-remote/src/client/worker.rs
```diff
@@ -117,12 +119,14 @@ impl ClientWorker {
...
-                while let Some(req) = rec.recv().await {
+                while let Ok(req) = rec.recv().await {
                     let task = match req {
                         ClientRequest::WithSyncCallback(task, callback) => {
-                            let mut state = state.lock().await;
                             if let Task::Compute(_content, id) = &task {
+                            log::info!("With sync callback lock..");
+                                let mut state = state.lock().await;
                                 state.register_callback(*id, callback);
+                            log::info!("With sync callback done");
```
laggui, line 129: Debugging left overs or actual logs?

examples/text-classification/Cargo.toml
```diff
@@ -42,3 +42,4 @@ tokenizers = { version = "0.20.3", default-features = false, features = [
 # Utils
 derive-new = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"] }
+log = { workspace = true }
```
laggui, line 45: Left over? I don't see it being used in the example.

laggui (approved):
Approving in advance, just have some minor comments that might need to be addressed before merging.

crates/burn-hip/src/lib.rs
nathanielsimard: It breaks rust-analyzer, so let's comment them until it's fixed.

PR #2499, opened 2024-11-16
Implicit GEMM optimizations/bug fixes

wingertge:
### Checklist

### Changes

Adds minor optimizations and support for `tf32` to implicit GEMM, fixes some bugs I encountered while testing the changes

### Testing

CUDA tests for `f32` check the new `tf32` feature, aligned reference backend test checks the unchecked weight loader on SPIR-V, and CUDA if `tf32` is disabled.


crates/burn-jit/src/kernel/interpolate/bilinear.rs
```diff
@@ -2,7 +2,7 @@ use cubecl::{calculate_cube_count_elemwise, prelude::*};
 
 use crate::{tensor::JitTensor, FloatElement, JitRuntime};
 
-#[cube(launch_unchecked)]
+#[cube(launch)]
 fn interpolate_bilinear_kernel<F: Float>(input: &Tensor<F>, output: &mut Tensor<F>) {
```
nathanielsimard, line 6: If there a reason to put this checked now?

crates/burn-jit/src/kernel/reduce/shared/kernel.rs
```diff
@@ -4,7 +4,7 @@ use crate::{kernel::reduce::init_reduce_output, tensor::JitTensor, JitElement, J
 
 use super::base::ReduceDimShared;
 
-#[cube(launch_unchecked)]
+#[cube(launch)]
 pub fn reduce_dim_shared_kernel<
```
nathanielsimard, line 8: Same

Issue #2500, opened 2024-11-16
one_hot fails for Int Tensor


maun:
**Describe the bug**
one_hot fails for Int Tensor

**To Reproduce**

1. Run example from function docs
```
fn example<B: Backend>() {
    let device = B::Device::default();
    let indices: Tensor<B, 1, Int> = Tensor::from_ints([0, 1, 2, 3], &device);
    let one_hot = indices.one_hot(4);
    println!("{}", one_hot.to_data());
    // [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]
}
```

2. Error:
```
=== Tensor Operation Error ===
  Operation: 'Scatter'
  Reason:
    1. The tensor shape should be the same as the index tensor shape. The shape differs at dimension 0: 4 != 1

stack backtrace:
   0: rust_begin_unwind
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
   2: core::panicking::panic_display
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:264:5
   3: burn_tensor::tensor::api::numeric::<impl burn_tensor::tensor::api::base::Tensor<B,_,K>>::scatter::panic_cold_display
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panic.rs:100:13
   4: burn_tensor::tensor::api::numeric::<impl burn_tensor::tensor::api::base::Tensor<B,_,K>>::scatter
             at /Users/manuel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/burn-tensor-0.15.0/src/tensor/api/check.rs:1181:17
   5: burn_tensor::tensor::api::int::<impl burn_tensor::tensor::api::base::Tensor<B,1_usize,burn_tensor::tensor::api::kind::Int>>::one_hot
             at /Users/manuel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/burn-tensor-0.15.0/src/tensor/api/int.rs:58:9
   6: core::example
```

**Expected behavior**
no error

**Fix**

To fix this `.unsqueeze()` must be changed to `unsqueeze_dim(1)`.

[...]

PR #2501, opened 2024-11-16
Fix one_hot implementation for Int Tensors (#2500)

maun:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/2500
Introduced in https://github.com/tracel-ai/burn/pull/2413

### Changes

Fix

### Testing

Ran the example, it worked.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/int.rs
+++ b/crates/burn-tensor/src/tensor/api/int.rs
@@ -55,7 +55,7 @@ where
         let [num_samples] = self.dims();
         let indices = self.unsqueeze_dim(1);
         let values = indices.ones_like();
-        Tensor::zeros([num_samples, num_samples], &indices.device()).scatter(1, indices, values)
+        Tensor::zeros([num_samples, num_classes], &indices.device()).scatter(1, indices, values)
     }
 }
 
--- a/crates/burn-tensor/src/tensor/api/int.rs
+++ b/crates/burn-tensor/src/tensor/api/int.rs
@@ -53,7 +53,7 @@ where
     pub fn one_hot(self, num_classes: usize) -> Tensor<B, 2, Int> {
         check!(TensorCheck::one_hot_tensor(self.clone(), num_classes));
         let [num_samples] = self.dims();
-        let indices = self.unsqueeze();
+        let indices = self.unsqueeze_dim(1);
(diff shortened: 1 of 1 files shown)


crates/burn-tensor/src/tensor/api/int.rs
```diff
@@ -53,7 +53,7 @@ where
     pub fn one_hot(self, num_classes: usize) -> Tensor<B, 2, Int> {
         check!(TensorCheck::one_hot_tensor(self.clone(), num_classes));
         let [num_samples] = self.dims();
-        let indices = self.unsqueeze();
+        let indices = self.unsqueeze_dim(1);
         let values = indices.ones_like();
         Tensor::zeros([num_samples, num_samples], &indices.device()).scatter(1, indices, values)
```
tsanona, line 58: Hey hey, yes thank you for changing this. Only realized later and was trying push the fix on another PR.
Forget not to also change this :rocket: 
```suggestion
        Tensor::zeros([num_samples, num_classes], &indices.device()).scatter(1, indices, values)
```
Just so I know in the future, where are these tensor operations tested?
Thanks! :star2:

maun:
I don't know, first commit here ;)
Ah, thanks for the fix, had it in my example, but forgot it here.

crates/burn-tensor/src/tensor/api/int.rs
maun: I don't know, first commit here ;) Ah, thanks for the fix, had it in my example, but forgot it here.

tsanona:
Ah right, hopefully the reviewer will answer it. Thanks for changing it!

crates/burn-tensor/src/tensor/api/int.rs
tsanona: Ah right, hopefully the reviewer will answer it. Thanks for changing it!

agelas: @tsanona Tests for this are in `crates/burn-tensor/src/tests/ops/squeeze.rs` (both squeeze and unsqueeze are in here).

nathanielsimard (approved):
Thanks for the fix!

laggui (approved):

Issue #2450, opened 2024-11-01
Add more examples to `Tensor` type


InAnYan:
I would like to see examples for every method in `Tensor` types.

That would be useful from people that came from PyTorch. Because Python language is highly dynamic and Rust is not, it's not easy to map PyTorch code to Rust with Burn

InAnYan:
I would also be awesome if you could show how to use indexing, like there:

https://docs.rs/tch/latest/tch/index/index.html

antimora:
Outdated

PR #2517, opened 2024-11-20
Fix tri mask ops return docstring

laggui:
### Checklist

### Related Issues/PRs
As pointed out [on discord](https://discord.com/channels/1038839012602941528/1059209073784008835/1308818474491052052)
#2450

### Changes

- Fixed the docstring for the return type (condition was the logical inverse, the mask is filled with `false` along the specified 
region).
- Added examples to illustrate the usage (and expected output)


Jaghov (approved):
Looks great :)

PR #2519, opened 2024-11-21
Add test int one_hot and change ops docs in the book

tsanona:
### Checklist

### Related Issues/PRs

#2500

### Changes

Add missing tests for Tensor<_, _, Int>::one_hot.
Change some ops documentation in the book.

### Testing

see above


laggui (approved):
Thanks for the fixes 🙂

PR #2526, opened 2024-11-22
[Feat] 8-bit bool for JitBackend

wingertge:
### Checklist

### Changes

Adds generic param for bool with the `JitBackend`, so `u8` can be used where supported.

### Testing

Adds tests to JIT backends to make sure boolean ops work with both `u32` and `u8` (where supported).


crates/burn-jit/src/fusion/on_write/trace.rs
```diff
@@ -273,9 +274,9 @@ impl FuseOnWriteTrace {
                     }
                 }
 
-                // We encode bool tensors as u32.
+                // We encode bool tensors as `B`.
                 let dtype = match tensor_global.dtype {
-                    DType::Bool => DType::U32,
+                    DType::Bool => B::dtype(),
```
nathanielsimard, line 279: Would it be better to pass the `bool_dtype` in the constructor instead of keeping the generic argument?

crates/burn-jit/src/kernel/mask/mask_where.rs
```diff
@@ -88,22 +92,23 @@ fn mask_where_readonly<R: JitRuntime, EI: JitElement, EM: JitElement>(
...
         cube_count,
         cube_dim,
-        input.as_tensor_arg::<EI>(1),
-        mask.as_tensor_arg::<EM>(1),
-        value.as_tensor_arg::<EI>(1),
-        output.as_tensor_arg::<EI>(1),
+        input.as_tensor_arg::<EI>(vectorization),
+        mask.as_tensor_arg::<EM>(vectorization),
+        value.as_tensor_arg::<EI>(vectorization),
+        output.as_tensor_arg::<EI>(vectorization),
```
nathanielsimard, line 104: Does it work?

nathanielsimard (commented):
Only a few comments. Also I think I would prefer having the generic argument named `BT` instead of `B`, since `B` is often used in burn for `Backend`, so that might be confusing sometimes.

crates/burn-jit/src/kernel/mask/mask_where.rs
wingertge: Yes, CUDA splits it into `vectorization` ternaries and both wgsl and SPIR-V support passing a Boolean vector to mix the true and false vector based on each value in the bool vector. All mask tests were passing too.

crates/burn-jit/src/fusion/on_write/trace.rs
wingertge: I think the generic is simpler for now

nathanielsimard (approved):

PR #2527, opened 2024-11-23
Make .init() method of LR schedulers return Result

### LINKED ISSUE #2510: LR Scheduler init should return `Result<Scheduler, ...>` instead of panicking
As discussed in [#2423](https://github.com/tracel-ai/burn/pull/2423#discussion_r1818106985).

Invalid initialization should not panic. We should refactor all schedulers to follow the same pattern (especially since warnings won't show up with the built-in logger + learner).

towerpark:
### Checklist

`run-checks` script was executed with `DISABLE_WGPU=1`, because otherwise I would get an unrelated error  similar to #1477.

### Related Issues/PRs

Fix #2510

### Changes

* Make `.init()` method of LR schedulers return `Result` instead of panicking for invalid configuration.
Add check and test code for Noam LR scheduler accordingly.
`ConstantLr` is not changed because there is nothing to check.

* Update affected examples (text-generation & text-classification).

* Clean up related test code (eliminate unnecessary constants and helper function)

### Testing

* Ensured that `DISABLE_WGPU=1 bash ./run-checks.sh` ran with no error
* Checked documentation by reading it in a browser
* Confirmed by grep-ing "scheduler" in the source code that the Books were not affected

### DIFF
--- a/crates/burn-core/src/lr_scheduler/base.rs
+++ b/crates/burn-core/src/lr_scheduler/base.rs
@@ -1,3 +1,5 @@
+pub(super) use alloc::string::String;
+
 use burn_tensor::backend::Backend;
 
 use crate::{record::Record, LearningRate};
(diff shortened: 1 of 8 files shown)


laggui (approved):
LGTM! Thanks for making this change 🙂

PR #2528, opened 2024-11-23
 Remove dtype rewrite

ArthurBrussee:
### Changes

I did something like 

```
let a = create_tensor(123, u32);
let b = create_tensor(123, a.dtype())
```

and got weird errors about sizes not being a multiple of 4, which after digging made me find that b was a bool tensor now!

I think that can be a pretty bad footgun, dtype() should return the actual dtype(). Bool tensors might need some reword to properly fix this and not display as u32, but I think that's better than having this API.


nathanielsimard:
The display problem should be resolved simply by passing the info that the data is encoding booleans in `burn-jit` not impacting the burn_tensor apis

laggui (approved):

PR #2530, opened 2024-11-24
[Optimization] Add custom NCHW to NHWC kernel for implicit GEMM

wingertge:
### Checklist

### Changes

Adds a custom NCHW to NHWC transpose kernel for use in `implicit_gemm`. This is faster than normal `into_contiguous` by specializing on this specific transposition.

### Testing

All tests compatible with `implicit_gemm` pass with the new kernel, added a new test to ensure the kernel output is the same as `into_contiguous`.

### DIFF
--- a/crates/burn-jit/src/template/base.rs
+++ b/crates/burn-jit/src/template/base.rs
@@ -19,12 +19,13 @@ pub struct SourceKernel<K> {
 }
 
 impl<C: Compiler, K: KernelSource> CubeTask<C> for SourceKernel<K> {
-    fn compile(&self, _mode: ExecutionMode) -> CompiledKernel<C> {
+    fn compile(&self, _options: &C::CompilationOptions, _mode: ExecutionMode) -> CompiledKernel<C> {
         let source_template = self.kernel_source.source();
         let source = source_template.complete();
 
         CompiledKernel {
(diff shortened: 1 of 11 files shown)


crates/burn-jit/src/template/base.rs
```diff
@@ -19,12 +19,13 @@ pub struct SourceKernel<K> {
...
 
 impl<C: Compiler, K: KernelSource> CubeTask<C> for SourceKernel<K> {
-    fn compile(&self, _mode: ExecutionMode) -> CompiledKernel<C> {
+    fn compile(&self, _options: &C::CompilationOptions, _mode: ExecutionMode) -> CompiledKernel<C> {
         let source_template = self.kernel_source.source();
         let source = source_template.complete();
 
         CompiledKernel {
-            name: Some(core::any::type_name::<K>()),
+            entrypoint_name: "kernel".to_string(),
```
AsherJingkongChen, line 27: Why change entry point name?

wingertge: It's an update to cubecl, `name` is now `debug_name` and there's a new `entrypoint_name`. However, it should be "main" and not "kernel", I'm fixing that in the PR I'm currently opening.

PR #2538, opened 2024-11-25
Remove precision bridge and complete `float_cast` implementation for ndarray

laggui:
### Checklist

### Related Issues/PRs

#2483 #2511 

### Changes

- Removed backend bridge (redundant now that we have `tensor.cast(dtype)`
- Added public `tensor.dtype()` method (helpful to perform to/from a type, e.g. full precision usage)
- Implemented `float_cast` for ndarray backend
  - Changed primitive to `NdArrayTensorFloat` enum
  - Refactored float ops to use `execute_with_float_dtype` macro which handles ops w/ generics based on the float enum variant

### Testing

Cast unit test + float ops unit tests


crates/burn-ndarray/src/tensor.rs
```diff
@@ -180,6 +339,20 @@ where
...
+    //     NdArrayTensor::new(array)
+    // }
+
+    // pub(crate) fn to_float(self) -> NdArrayTensorFloat {
+    //     match E::dtype() {
+    //         DType::F64 => NdArrayTensorFloat::F64(self.cast_to()),
+    //         DType::F32 => NdArrayTensorFloat::F32(self.cast_to()),
+    //         _ => unimplemented!("Invalid float type"),
+    //     }
+    // }
```
nathanielsimard, line 355: Dead code

laggui: Dead code detector on point, as always 😄

nathanielsimard (approved):

PR #2540, opened 2024-11-26
Check nonzero stride, dilation and groups

### LINKED ISSUE #2481: `conv2d` should check if `dilation > 0`
### Feature description

`tensor::module::conv2d()` should check if `dilation` is greater than 0.

I haven't checked, but `conv1d` and `conv3d` also should have that check included.

### Feature motivation

Mathematically convolution with dilation (0,0) doesn't really make sense. The implementation in backends can handle dilation (0,0) without error, but the output is nonsensical.

Presumably a simple `if () {panic!()}` will work.

laggui:
### Checklist

### Related Issues/PRs

Fixes #2481 

### Changes

Added checks for non-zero values in conv, conv transpose, deform conv and unfold.

To keep it simple I left the fields as usize and simply added a custom `new` method with the checks, but we could eventually hold `NonZeroUsize` fields instead.

### Testing

Added unit tests.

### DIFF
--- a/crates/burn-jit/src/fusion/on_write/trace_builder.rs
+++ b/crates/burn-jit/src/fusion/on_write/trace_builder.rs
@@ -40,7 +40,7 @@ impl FuseOnWriteTraceBuilder {
         let meta = 1;
         let inputs = self.inputs.len() as u32;
         let outputs = self.output_tensors().len() as u32;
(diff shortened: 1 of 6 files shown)


antimora (approved):
LGTM!

PR #2545, opened 2024-11-26
[Optimization] Implicit gemm rewrite

wingertge:
### Checklist

### Related Issues/PRs

Requires https://github.com/tracel-ai/cubecl/pull/309 to land first

### Changes

Adds a brand new implicit GEMM implementation that uses the matmul primitives in `cubecl`. This is slower for small k sizes, but *much* faster for large ones, and more flexible. I'm keeping the current implementation because it's significantly faster for certain sizes, and uses a significantly different loader strategy (loading only within each warp, which skips cross warp syncs).

Adds a number of new convolution benchmarks to test performance with different sizes and characteristics.

### Testing

All non-group tests pass, and CRAFT has the expected output with all layers using the new implicit GEMM. This tests many different and relatively large layers. Adds two new regression tests for bugs discovered during implementation.


crates/burn-jit/src/kernel/conv/conv2d/base.rs
```diff
@@ -24,6 +21,9 @@ pub enum Conv2dStrategy {
     /// Implicit GEMM implementation of convolution. Lower memory usage but requires CMMA and
     /// has constraints on tensor shape.
     ImplicitGemm,
+    /// Implicit GEMM implementation of convolution. Lower memory usage but requires CMMA and
+    /// has constraints on tensor shape.
+    ImplicitGemmNew,
```
nathanielsimard, line 26: Instead of `New` postfix, it could be named `ImplicitGemmLarge` ?

crates/burn-jit/src/kernel/conv/conv2d/gemm/algorithm.rs
```diff
@@ -0,0 +1,153 @@
...
+    type EG: Numeric;
+    type ES: Numeric;
+    type EA: Numeric;
+
+    type TileMatmul: tile::Matmul<Self::ES, Self::EA> + MatmulKernel<Self::ES, Self::EA>;
+
+    type StageSize: StageSize;
+    type StageMatmul: stage::Matmul<Self::ES, Self::EG, Self::EA> + MatmulKernel<Self::ES, Self::EG>;
+
+    type GlobalMatmul: Convolution<Self::EG, Self::ES, Self::EA, Self::StageMatmul>
```
nathanielsimard, line 34: I would rename `GlobalMatmul` => `GlobalConvolution`.

crates/burn-jit/src/kernel/conv/conv2d/gemm/algorithm.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        CubeCount::Static(cubes_needed_m, cubes_needed_n, 1)
+    }
+}
+
+/// Cmma convolution accumulating in half precision
+pub struct CmmaHalf<EG: Numeric, Stage: StageSize> {
+    pub _eg: PhantomData<EG>,
+    pub _stage: PhantomData<Stage>,
+}
```
nathanielsimard, line 120: Could the `Cmma` struct be generic over the accumulation precision?

crates/burn-jit/src/kernel/conv/conv2d/gemm/homogeneous/base.rs
```diff
@@ -0,0 +1,450 @@
...
+                acc,
+                config.to_smm_config(),
+            );
+
+            sync_units();
+
+            Self::LhsLoader::advance_view(&mut lhs_loader, k_step);
+            Self::RhsLoader::advance_view(&mut rhs_loader, k_step);
+        }
+
```
nathanielsimard, line 115: Somehow adding a `sync_units` after the for loop improved performance for the matmul. I think it makes sure all units in a plane are sync following the loop which improve the execution of following operations.

crates/burn-jit/src/kernel/conv/conv2d/gemm/homogeneous/base.rs
```diff
@@ -0,0 +1,450 @@
...
+            bias,
+            out,
+            config,
+            config.has_bias,
+        );
+    }
+}
+
+#[cube(launch_unchecked)]
+pub(crate) fn launch<
```
nathanielsimard, line 251: I would rename that to `implicit_conv` instead. It improves the name displayed in `ncu` and with our profiler.

crates/burn-jit/src/kernel/conv/conv2d/gemm/launch.rs
```diff
@@ -0,0 +1,259 @@
...
+
+/// Perform a 2D convolution using the implicit GEMM (im2col) algorithm, using cubecl tiling matmul
+/// components. Uses [`CmmaLargeMAlgorithm`] for the stage size
+///
+/// * `input` - The input feature map
+/// * `weight` - The weights (filter) applied to each kernel
+/// * `bias` - The bias added to each channel
+/// * `options` - The options to use for the convolution
+///
+///
```
nathanielsimard, line 50: Empty lines in comment block 😅

crates/burn-jit/src/kernel/conv/conv2d/gemm/loader/base.rs
```diff
@@ -0,0 +1,50 @@
...
+            &this.tensor_view,
+            &mut this.stage.as_slice_mut(),
+            Ident::Rhs,
+            config,
+        );
+        RhsReader::new(this.stage)
+    }
+
+    fn advance_view(this: &mut Self, k_offset: u32) {
+        this.tensor_view.update_view(k_offset, Ident::Rhs);
```
nathanielsimard, line 48: Do you need to duplicate this? I don't believe we actually need to have the constraint `G: global::Config` in the trait, only `G` is good enough. The `Algorithm` trait can make the link between the two types.

nathanielsimard (commented):
It looks awesome! Feels great to reuse a lot of components. There are still some improvements that we can make in our "design paradigm", especially in how we pass around the config. But this is beyond the scope of this PR.

I have a few comments, but it would also be great for @louisfd to review.

crates/burn-jit/src/kernel/conv/conv2d/base.rs
wingertge: I think I want to call it complex instead, since it's not really about size it's about the relative shape and factors like k size, and I also want to experiment with other variations like k splitting, producer consumer, etc. So I think complex is a better descriptor.

crates/burn-jit/src/kernel/conv/conv2d/gemm/loader/base.rs
wingertge: You're right, this used to be required when the config was a generic on the function, but it's now superfluous.

crates/burn-jit/src/kernel/conv/conv2d/gemm/homogeneous/base.rs
wingertge: I'll benchmark it

wingertge: So the benchmark results are very odd. I tried 4 different ways of syncing, and the one I used initially was overall the fastest for CUDA, but adding a sync before the load and after the loop was significantly faster for SPIR-V. Only syncing where absolutely needed was the slowest by far. Very odd behaviour. I'll stick with lots of sync for now because it's only 10-15% slower than the current implementation on CUDA (and there's margin of error), but 30% faster on SPIR-V.

nathanielsimard (approved):
LGTM we can merge after the conflicts are resolved!

wingertge:
Done 👍

PR #2551, opened 2024-11-26
Matmul + CubeCL Update

nathanielsimard:


crates/burn-jit/src/kernel/matmul/base.rs
```diff
@@ -40,14 +31,11 @@ pub fn matmul<R: JitRuntime, E: FloatElement>(
...
     match strategy {
-        MatmulStrategy::Simple { grid_x, grid_y } => {
-            let out = init_matmul_output::<R, E>(&lhs, &rhs);
-            matmul_simple::<R, E>(lhs, rhs, out, grid_x, grid_y)
-        }
         MatmulStrategy::Cube => {
             let out = init_matmul_output::<R, E>(&lhs, &rhs);
             let client = &lhs.client;
             cubecl::linalg::matmul::launch_ref::<R, E>(
+                &Default::default(),
```
louisfd, line 38: Since default in cubecl is Accelerated, if you're on wgpu and have autotune disabled, it will fail?

louisfd: Since a very recent PR launch_ref of matmul in CubeCL returns a Result, so we get an error if Accelerated is not available. That way we can have a fallback here

PR #2557, opened 2024-11-27
Add `QTensorOps` docs + refactor tests to simplify inputs

laggui:
### Checklist

### Changes

- Refactored qtensor ops tests w/ `QTensor` helper struct
  - Previously defined inputs was clunky, error prone and difficult to understand. Now anyone can more easily add tests for operations using floating point values. `QTensor` methods will automatically quantize the inputs.
- Added missing `q_split` tests
- Added `QTensorOps` docs to burn contrib book
- Fixed zero scale edge cases everywhere


louisfd (approved):
LGTM! Much cleaner :)

PR #2561, opened 2024-11-28
Make optimizer state public

ArthurBrussee:
For some use cases, it helps to do surgery on the internal optimizer state (in my case - I dynamicall add parameters, and want to add zero's as the momenta for these).

This could be done by going from optimizer -> record -> state, doing the modifications, and going back to an optimizer. A dedicate API could also work but just exposing fields seems simplest.

Also remove the Backend bound on WeightDecay - it doesn't make much sense to have WeightDecay be a FloatElement, and it complicates all bounds. AdamW already used a f32 for decay, so matching that.

### DIFF
--- a/crates/burn-core/src/optim/decay.rs
+++ b/crates/burn-core/src/optim/decay.rs
@@ -4,13 +4,13 @@ use crate as burn;
 use crate::record::Record;
 
 use crate::config::Config;
-use crate::tensor::{ElementConversion, Tensor};
+use crate::tensor::Tensor;
 
 /// Configuration to create [weight decay](WeightDecay).
 #[derive(Config)]
 pub struct WeightDecayConfig {
     /// L2 penalty.
-    pub penalty: f64,
+    pub penalty: f32,
 }
 
 /// State of [weight decay](WeightDecay).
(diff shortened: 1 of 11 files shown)


crates/burn-core/src/optim/decay.rs
```diff
@@ -21,11 +21,11 @@ pub struct WeightDecayState<B: Backend, const D: usize> {
 
 /// Weight decay implementation that transforms gradients.
 #[derive(Clone)]
-pub struct WeightDecay<B: Backend> {
-    penalty: B::FloatElem,
+pub struct WeightDecay {
+    penalty: f32,
```
laggui, line 25: Makes sense! For uniformity, might as well change the `WeightDecayConfig` penalty to f32 as well.

And then in `WeightDecay::new(...)` we don't need to call `.elem()`.

laggui (approved):
LGTM!

One small note, but I'll approve in advance.

Also CI issues will be fixed with #2562

crates/burn-core/src/optim/decay.rs
ArthurBrussee: Oh yes nice catch :)

PR #2568, opened 2024-11-29
Feat/Split ONNX Import

agelas:
### Checklist

### Related Issues/PRs

#2440 

### Changes

Adds split to the list of supported ops that can be imported via ONNX.

### Testing

New onnx tests and test codegen.


agelas:
@antimora  For nodes that produce multiple outputs, is there a recommended pattern for how to assign and return these outputs in the forward method? I tried to do the reverse of `concat`, but still having a bit of trouble.

Also, in the generated ONNX graph, all three outputs of the graph are named "split1_out1", but the outputs of the node are named uniquely. How do we ensure that each output of a multi-output node is named uniquely (assuming they should be)? I keep getting this error when I try generating the IR: 
```
thread 'main' panicked at crates/burn-import/src/burn/graph.rs:566:40:
Output type not found for split1_out1
```
I have a feeling that the non-unique names might be tripping this up, because the overall structure of the outputs is consistent with other IRs I've generated.

I also copied some of the generated graph to show you what I mean.
```
ParsedOnnxGraph(
    // omitting the constant node for brevity
            Node {
                // omitting a bunch of stuff here too
                outputs: [
                    Argument {
                        name: "split1_out1",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 2,
                                shape: None,
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                    Argument {
                        name: "split1_out2",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 2,
                                shape: None,
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                    Argument {
                        name: "split1_out3",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 2,
                                shape: None,
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                ],
                attrs: {
                    "axis": Int64(
                        0,
                    ),
                },
            },
        ],
        inputs: [
           // ignoring this
        ],
        outputs: [
            Argument {
                name: "split1_out1", <--- this name is the same for everything in outputs[]
                ty: Tensor(
                    TensorType {
                        elem_type: Int64,
                        dim: 2,
                        shape: None,
                    },
                ),
                value: None,
                passed: false,
            },
            Argument {
                name: "split1_out1",
                ty: Tensor(
                    TensorType {
                        elem_type: Int64,
                        dim: 2,
                        shape: None,
                    },
                ),
                value: None,
                passed: false,
            },
            Argument {
                name: "split1_out1",
                ty: Tensor(
                    TensorType {
                        elem_type: Int64,
                        dim: 2,
                        shape: None,
                    },
                ),
                value: None,
                passed: false,
            },
        ],
    },
)
```

antimora:
Sorry I forgot to reply. I'll review it tomorrow or next week soon. I have been working on removing static shapes.

agelas:
@laggui Do you think you could take a look?

laggui:
> @laggui Do you think you could take a look?

Will do 🫡 should have time tomorrow.

Sorry I saw that someone was already taking care of this last time so totally checked it off my mind lol

crates/burn-import/src/burn/node/split.rs
```diff
@@ -0,0 +1,137 @@
...
+        let dims_literal = proc_macro2::Literal::usize_unsuffixed(dims);
+
+        let vec_tensor_type = quote! {Vec<Tensor<B, #dims_literal>>};
+
+        let other_type = OtherType {
+            name: syn::Ident::new("split_tensors", proc_macro2::Span::call_site()),
+            ty: vec_tensor_type,
+        };
+
+        vec![Type::Other(other_type)]
```
laggui, line 36: Fixed number of outputs based on the parsed node outputs

```suggestion
        self.outputs
            .iter()
            .map(|t| Type::Tensor(t.clone()))
            .collect()
```

crates/burn-import/src/burn/node/split.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    fn input_types(&self) -> Vec<Type> {
+        vec![Type::Tensor(self.input.clone())]
+    }
+
+    fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
+        let input = scope.tensor_use_owned(&self.input, node_position);
+        let axis = self.config.axis.to_tokens();
+        let split_tensors = syn::Ident::new("split_tensors", proc_macro2::Span::call_site());
```
laggui, line 46: With the above changes, we can safely unpack into a fixed number of outputs
```suggestion
        let outputs = self
            .outputs
            .iter()
            .map(|t| t.name.clone())
            .collect::<Vec<_>>();

        // We know the actual number of outputs, so we can safely unpack into a fixed number
        let unpack_outputs = quote! {
            let [#(#outputs),*] = split_tensors.try_into().unwrap();
        };
```

crates/burn-import/src/burn/node/split.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        let split_code = if let Some(split_sizes) = &self.config.split_sizes {
+            let split_sizes_tokens = split_sizes.to_tokens();
+            quote! {
+                let #split_tensors = #input.split_with_sizes(#split_sizes_tokens, #axis);
```
laggui, line 51: We can then unpack the outputs
```suggestion
                let mut split_tensors = #input.split_with_sizes(#split_sizes_tokens, #axis);
                #unpack_outputs
```

crates/burn-import/src/burn/node/split.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+        } else {
+            let num_outputs = self.config.num_outputs.unwrap();
+            let num_outputs_tokens = num_outputs.to_tokens();
+            quote! {
+                let #split_tensors = #input.split(#num_outputs_tokens, #axis);
```
laggui, line 57: Same here
```suggestion
                let mut split_tensors = #input.split(#num_outputs_tokens, #axis);
                #unpack_outputs
```

crates/burn-import/src/burn/node/split.rs
```diff
@@ -0,0 +1,137 @@
...
+            SplitConfig {
+                axis: 0,
+                num_outputs: Some(2),
+                split_sizes: None,
+            },
+        ));
+
+        graph.register_input_output(
+            vec!["tensor1".to_string()],
+            vec!["split_tensors".to_string()],
```
laggui, line 99: Which means the number of outputs will follow here
```suggestion
            vec!["tensor2".to_string(), "tensor3".to_string()]
```

crates/burn-import/src/burn/node/split.rs
```diff
@@ -0,0 +1,137 @@
...
+                }
+
+                #[allow(clippy::let_and_return, clippy::approx_constant)]
+                pub fn forward(
+                    &self,
+                    tensor1: Tensor<B, 2>,
+                ) -> Vec<Tensor<B, 2>> {
+                    let split_tensors = tensor1.split(2, 0);
+
+                    split_tensors
```
laggui, line 130: And so the generated function should look like this instead:
```suggestion
                ) -> (Tensor<B, 2>, Tensor<B, 2>) {
                    let split_tensors = tensor1.split(2, 0);
                    let [tensor2, tensor3] = split_tensors.try_into().unwrap();
                    (tensor2, tensor3)
```

laggui (commented):
> How do we ensure that each output of a multi-output node is named uniquely (assuming they should be)

Following the ONNX spec, each output has to be uniquely named.

> I have a feeling that the non-unique names might be tripping this up, because the overall structure of the outputs is consistent with other IRs I've generated.

I'm not sure what you mean, because in my case the node output names and the graph output names are all unique:

<details>
<summary>Output names</summary>

```
outputs: [
    Argument {
        name: "split1_out1",
        ty: Tensor(
            TensorType {
                elem_type: Int64,
                dim: 2,
                shape: None,
            },
        ),
        value: None,
        passed: false,
    },
    Argument {
        name: "split1_out2",
        ty: Tensor(
            TensorType {
                elem_type: Int64,
                dim: 2,
                shape: None,
            },
        ),
        value: None,
        passed: false,
    },
    Argument {
        name: "split1_out3",
        ty: Tensor(
            TensorType {
                elem_type: Int64,
                dim: 2,
                shape: None,
            },
        ),
        value: None,
        passed: false,
    },
],
```
</details>

Regarding the implementation, while the number of outputs can vary from one node to another, I think for one node instance the number of outputs is actually known. And anyway we don't support control-flow operations like loops, so this is a good starting point.

For that reason, we can simplify the split output type to simply use the collected names. And the codegen can unpack into a fixed number of outputs. See my comments below.

Just a note: I inspected the generated `split.onnx` test file and I think you'll have an issue with the constant int (see #2625). We either need to fix the constant int issue, or you can try to circumvent this for the split test with a different model.

agelas:
@laggui Something changed in the last ~3 months and now all the node output names are unique for me too now, it looks like time fixed the issue 🤣. I went ahead and took your recommendation to simplify the output types to the collected values- I think that part is all good to go.

You are correct that it looks like I'm running into the issue described in #2625. Is a fix forthcoming or should I go ahead and try to find a workaround for now?

agelas:
@laggui I took a peek around the discussion in #2625 and the solution in @jameshiew wip branch seems to do the trick. @jameshiew I hope you don't mind if I include it here.

crates/burn-import/src/burn/node/constant.rs
```diff
@@ -121,9 +121,17 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for ConstantNode {
...
+                    crate::burn::TensorKind::Int => Some(quote! {
+                        let #name: burn::module::Param<#ty> = burn::nn::zeros_int(#shape, device);
+                    }),
+                    crate::burn::TensorKind::Float => Some(quote! {
+                        let #name: burn::module::Param<#ty> = burn::nn::zeros_float(#shape, device);
+                    }),
+                    crate::burn::TensorKind::Bool => Some(quote! {
+                        let #name: burn::module::Param<#ty> = burn::nn::zeros_bool(#shape, device);
+                    }),
+                }
```
laggui, line 134: The match on tensor kind should fix this! But I don't think we want to have public methods for `zeros_int`, `zeros_float`, `zeros_bool` just for this reason.

The code is fairly straightforward, so we can simply add it here to handle the correct codgen. For example:

```rust
Param::uninitialized(
    ParamId::new(),
    move |device, _require_grad| Tensor::<B, D, burn_tensor::Int>::zeros(shape.clone(), device),
    device,
    false,
)
```

For bool,

jameshiew:
> @laggui I took a peek around the discussion in #2625 and the solution in @jameshiew wip branch seems to do the trick. @jameshiew I hope you don't mind if I include it here.

Definitely go ahead! Feel free to include anything from my fork

laggui (approved):

laggui:
Sorry for the initial delay, this one slipped through the cracks after another reviewer was tagged 😅 

I made sure to include @jameshiew as co-author for the constant initialization fix that was adapted!

PR #2584, opened 2024-12-03
Migrate matmul autotune to macro and fix accelerated

wingertge:
Fix matmul autotune by checking for availability in `should_run`.

### Checklist

### Related Issues/PRs

Should fix https://github.com/ArthurBrussee/brush/issues/53

### Changes

Migrates matmul autotune to the macro to clean up the code.
Implements a simple `should_run` to check for availability before running accelerated.

### Testing

Tests passing.


laggui (approved):
LGTM!

/edit: pinged @nathanielsimard in case I missed something

crates/burn-jit/src/kernel/matmul/tune/base.rs
```diff
@@ -15,73 +16,45 @@ use crate::{
...
+            a: Elem::Float(FloatKind::F16),
+            b: Elem::Float(FloatKind::F16),
+            c: Elem::Float(FloatKind::F32),
+            m: 16,
+            k: 16,
+            n: 16,
+        }),
+        _ => true,
     }
 }
```
nathanielsimard, line 59: We made all the checks in the matmul launch code. It's quite complex because it's not always the same instructions & precisions that are used! Normally, the function will panic on the checks, and therefore not be considered in the autotune. It will not run if the right features are not enabled on the device! `should_run` should not duplicate that logic.

nathanielsimard (changes requested):

crates/burn-jit/src/kernel/matmul/tune/base.rs
wingertge: Which is why this came up to begin with

ArthurBrussee:
Would it be possible to merge this? I'm still hitting the panic on wasm so currently can't upgrade Burn.

laggui:
@ArthurBrussee should be good to go now!

PR #2604, opened 2024-12-09
Refactor jit quantized tensor representation

laggui:
### Checklist

### Related Issues/PRs

On the way to burn-fusion quantize

Follow-up to #2479

### Changes

- Simplified quantized tensor handle repr (one handle per qtensor)
- Removed strategy from QTensorPrimitive trait
- Refactored burn-jit quantized tensor repr and kernel
  - Removed QJitTensor (use JitTensor with QFloat dtype)
  - Refactored cube kernels w/ stored quantization params in cube tensor and new

### Testing

Unit tests


crates/burn-fusion/src/tensor.rs
```diff
@@ -133,6 +128,21 @@ impl<R: FusionRuntime> FusionTensor<R> {
...
+        if let DType::QFloat(_scheme) = self.dtype {
+            let id = self.stream;
+            self.client
+                .clone()
+                .read_tensor_quantized::<B>(self.into_description(), id)
+                .await
+        } else {
+            panic!("Expected quantized float dtype, got {:?}", self.dtype)
+        }
+    }
```
nathanielsimard, line 144: We should not call await here. it creates deadlocks.

```rust
pub(crate) fn q_into_data<B>(self) -> impl Future<Output=TensorData>
    where
    {
        if let DType::QFloat(_scheme) = self.dtype {
            let id = self.stream;
            let fut = self.client
                .clone()
                .read_tensor_quantized::<B>(self.into_description(), id);
                async move {
                    fut.await
                }
        } else {
            panic!("Expected quantized float dtype, got {:?}", self.dtype)
        }
    }
 ```

File: crates/burn-jit/src/kernel/quantization/qtensor.rs [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+            QuantizationScheme::PerTensorAffine(_) => match comptime!(tensor.line_size()) {
+                // For line size of 1, the zero-point offset is the penultimate value in the buffer
+                1 => i32::cast_from(tensor[len - 2][tensor.line_size() - 1]),
+                // For any other line size > 1, scale and zero-point offset are the last two elements
+                _ => i32::cast_from(tensor[len - 1][tensor.line_size() - 2]),
+            },
+            // Symmetric quantization only contains the scaling factor, so we return 0 for the zero-point offset
+            QuantizationScheme::PerTensorSymmetric(_) => 0,
+        }
+    }
```
nathanielsimard, line 53: Is it possible to have only one function that returns a tuple? We can avoid 1 read in global memory when line size if > 1

crates/burn-wgpu/src/lib.rs
```diff
@@ -22,7 +22,7 @@ pub type SpirV = cubecl::wgpu::spirv::VkSpirvCompiler;
 #[cfg(feature = "spirv")]
 type Compiler = SpirV;
 #[cfg(feature = "spirv")]
-type Byte = u8;
+type Bool = u8;
```
nathanielsimard, line 25: Is it ok?

laggui: Unrelated to this PR, but if you run the checks it fails for spirv

```
error[E0412]: cannot find type `Bool` in this scope
  --> crates/burn-wgpu/src/lib.rs:63:37
   |
63 | pub type Wgpu<F = f32, I = i32, B = Bool, C = Compiler> =
   |                                     ^^^^
   |
help: a builtin type with a similar name exists
   |
63 | pub type Wgpu<F = f32, I = i32, B = bool, C = Compiler> =
   |                                     ~~~~
help: perhaps you intended to use this type
   |
63 | pub type Wgpu<F = f32, I = i32, B = bool, C = Compiler> =
   |                                     ~~~~
help: consider importing this struct
   |
12 + use burn_tensor::Bool;
   |

For more information about this error, try `rustc --explain E0412`.
```

so I fixed it here. I think it was a leftover from the byte type rename.

crates/burn-jit/src/kernel/quantization/qtensor.rs
laggui: I tried that at first but I hit a limitation on the cube expand.

Probably need to fix it at the cubecl level but wanted to get this PR going first.

iirc it would not handle the tuple correctly, but let me try again.

crates/burn-fusion/src/tensor.rs
laggui: Changed all other `into_data` methods for `FusionTensor` 👍

PR #2612, opened 2024-12-12
Extend ImageFolderDataset to support import of COCO detection data

jin-eld:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed*.
      * with wgpu errors not related to this PR

### Changes

COCO is a popular dataset which defines an own dataset format: https://cocodataset.org/#format-data

This commit introduces an `ImageFolderDataset::new_coco_detection()` function, which expects a path to the JSON annotations file in COCO format and a path where the actual images are stored.

Note, that while COCO also offers segmentation and pose estimation data, for now only the import of detection (bounding boxes) data is supported.

### Testing

Apart from testing with actual COCO data, a small dataset containing three images and annotations in COCO format has been added to `crates/burn-dataset/tests/data`. Each image has a different amount of annotations, the test checks if the imported dataset contains the correct number of images and the expected number of annotations per image.


crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -809,4 +1017,30 @@ mod tests {
...
+                Annotation::BoundingBoxes(v) => {
+                    if file_name == "two_dots_and_triangle.jpg" {
+                        assert_eq!(v.len(), 3);
+                    } else if file_name == "dot_triangle.jpg" {
+                        assert_eq!(v.len(), 2);
+                    } else if file_name == "one_dot.jpg" {
+                        assert_eq!(v.len(), 1);
+                    } else {
+                        panic!("{}", format!("unexpected image name: {}", item.image_path));
+                    }
```
laggui, line 1040: I would add a check to validate the expected bounding boxes, not just the number of bounding boxes.

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -468,6 +476,37 @@ impl ImageFolderDataset {
...
+    ) -> Result<Self, ImageLoaderError> {
+        let file = fs::File::open(annotations_json)
+            .map_err(|e| ImageLoaderError::IOError(format!("Failed to open annotations: {}", e)))?;
+        let json: Value = serde_json::from_reader(file).map_err(|e| {
+            ImageLoaderError::ParsingError(format!("Failed to parse annotations: {}", e))
+        })?;
+
+        let classes = Self::parse_coco_classes(&json)?;
+        let annotations = Self::parse_coco_bbox_annotations(&json)?;
+        let items = Self::parse_coco_images(&images_path, annotations, &json)?;
```
laggui, line 502: Tiny nitpick, but I'd remove the `parse_coco_*` methods from the `ImageFolderDataset` and make them standalone functions (in the same module).

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -82,7 +85,7 @@ pub struct SegmentationMask {
 /// Object detection bounding box annotation.
 #[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
 pub struct BoundingBox {
-    /// Coordinates.
+    /// Coordinates in [x,y,width,height] format.
```
laggui, line 88: Just double checked the format, and for clarity I think I would be even more precise by stating `[x_min, y_min, width, height]` since there also exists `[cx, cy, width, height]` which could be confusing.

laggui (changes requested):
That's awesome! The implementation is rather concise 😄 

Looks like it fit pretty well with the provisioned types for `ImageFolderDataset`.

I have some minor comments, but otherwise great job!

crates/burn-dataset/src/vision/image_folder.rs
jin-eld: I was not sure if having all those coordinates in the test wasn't too much, but sure, I can add the checks for the actual boxes too.

crates/burn-dataset/src/vision/image_folder.rs
jin-eld: Hmm, OK, it's a bit of a mix right now, i.e. `check_extensions()` and `with_items()` are struct members, which was my reasoning for adding the COCO stuff there too, but we can move the COCO helpers out.

crates/burn-dataset/src/vision/image_folder.rs
jin-eld: Oops, I actually missed that, was not aware of the `_min` thing at all, will update the comment.

jin-eld:
@laggui updated, I usually amend and force push to keep the history clean, hope that's OK for you.

crates/burn-dataset/src/vision/image_folder.rs
laggui: I think it's valuable to make sure we are parsing the correct values.

crates/burn-dataset/src/vision/image_folder.rs
laggui: `with_items(...)` returns an instance of the struct.

Though I agree, `check_extensions` could also be moved out. Feel free to make this change too if you feel like it would make more sense 🙂

crates/burn-dataset/src/vision/image_folder.rs
laggui: Looks like the comment wasn't updated with the last changes 🤔

laggui (commented):
> I usually amend and force push to keep the history clean, hope that's OK for you.

Might mess up the github review UI for stuff that was already reviewed and untouched, but I don't have any issues with that. We squash & merge anyway so it will be reduced to a single commit 👍 

Looks like the comment update was missed, otherwise it will be good to go!

jin-eld:
> > I usually amend and force push to keep the history clean, hope that's OK for you.
> 
> Might mess up the github review UI for stuff that was already reviewed and untouched, but I don't have any issues with that. We squash & merge anyway so it will be reduced to a single commit 👍
> 
> Looks like the comment update was missed, otherwise it will be good to go!

sorry, C&P fail 🤦🏻‍♂️

laggui (approved):

laggui:
Thank you 🙏

PR #2613, opened 2024-12-15
Feature add new one hot function meeting multi-dimensions (ranks)

tiruka:
## Target of This Pull Request

First, I attempted to implement a one-hot operation for ONNX. However, I realized that the existing one-hot function did not meet the requirements and, in fact, did not support multidimensional inputs at all. As I explored solutions, including the [ONNX specifications](https://onnx.ai/onnx/operators/onnx__OneHot.html), Pytorch, Tensorflow, I concluded that it was necessary to implement a new one-hot function. This led to the creation of this implementation, which I am now submitting as a pull request.
(Pytorch also does not implement complet one hot function, though.)
- https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html
- https://www.tensorflow.org/api_docs/python/tf/one_hot

Hope this will work for burn and community.

### Checklist

### Related Issues/PRs

Indirectly related to onnx issues https://github.com/tracel-ai/burn/issues/1714

### Changes

Newly implemented one hot method for numeric tensor. The reason it should belong to numeric is the return value should be defined by on_value and off_value, not tensor itself. So, the output can take either types of int and float. 
This function comprehensively covers all aspects defined by ONNX, including depth, on_value, off_value, and axis, and complies with the one-hot operator specifications introduced in ONNX version 11 and later. By developing this, I believe it becomes possible to handle multidimensional one-hot encoding while also providing a concise and efficient implementation of the ONNX operator. For these reasons, I deemed it essential to create this function.

I considered removing and updating the existing one-hot method, but I decided to take a more conservative approach by leaving the existing method as it is and adding a new one instead.

### Testing

Adding tests on `crates/burn-tensor/src/tests/ops/one_hot.rs` and passing `run-checks all`.


laggui (commented):
I'll take some time to look at this later, but just a couple comments before reviewing the actual code.

1. We don't *need* to have all of the ops comply with the ONNX spec.
2. Introducing this as a new operation means that we now have multiple definitions for one-hot. One definition should take over, otherwise it makes everything cluttered.
3. Regarding the motivation, do you actually need this one-hot definition? Or is it simply  for ONNX conversion? If only the latter, than it can probably just live in the ONNX import code.

tiruka:
@laggui Thank you for your comments.


The existing `one_hot` function only operates on rank-1 tensors, which limits its usability. 
```
impl<B> Tensor<B, 1, Int> {
  ...
  pub one_hot() {
  ...
  
  }
}  
```
For current float version one hot, I do not come up with any use case.

In PyTorch, for example, the function is minimally designed to support multiple dimensions, and this aspect is something that needs improvement in our framework as well. 
[Pytorch example](https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html)
```
F.one_hot(torch.arange(0, 5) % 3, num_classes=5)
tensor([[1, 0, 0, 0, 0],
        [0, 1, 0, 0, 0],
        [0, 0, 1, 0, 0],
        [1, 0, 0, 0, 0],
        [0, 1, 0, 0, 0]])
```

Furthermore, another major framework, TensorFlow, not only supports multiple dimensions but also provides flexibility with parameters such as `axis` and `values`. 
[Tensorflow example](https://www.tensorflow.org/api_docs/python/tf/one_hot)
```
indices = [[0, 2], [1, -1]]
depth = 3
tf.one_hot(indices, depth,
           on_value=1.0, off_value=0.0,
           axis=-1)  # output: [2 x 2 x 3]
# [[[1.0, 0.0, 0.0],   # one_hot(0)
#   [0.0, 0.0, 1.0]],  # one_hot(2)
#  [[0.0, 1.0, 0.0],   # one_hot(1)
#   [0.0, 0.0, 0.0]]]  # one_hot(-1)
```
Further usecases
- https://www.tensorflow.org/federated/tutorials/federated_learning_for_image_classification
- https://www.tensorflow.org/federated/tutorials/sparse_federated_learning


The ability to configure multiple dimensions, `axis`, and `values` is an expected feature in popular frameworks, and I believe this would greatly benefit Burn users, myself included, by helping the framework stay aligned with modern expectations. This one hot function is not closed only for ONNX.

Regarding the concern about having multiple definitions, I also have the same sentiment and agree that unification is necessary. My proposed new function is designed to support both `int` and `float` types, making it closer to the `one_hot` definitions found in other frameworks. As such, I would advocate deprecating the existing implementation and unifying it with this new version. If there is agreement on this approach, I would be happy to submit changes either as part of this PR or in a separate PR to address these points. 

I look forward to your feedback and hope for your support in making this improvement.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2030,6 +2030,95 @@ where
...
+    ///     // [0.0, 0.0, 5.0]]]
+    /// }
+    /// ```
+    pub fn one_hot_plus<K2: Numeric<B>, const D2: usize>(
+        self,
+        depth: usize,
+        on_value: K2::Elem,
+        off_value: K2::Elem,
+        axis: i64,
+    ) -> Tensor<B, D2, K2> {
```
laggui, line 2068: Technically speaking, one-hot encoding only has "hot" values encoded as 1 and the rest is 0. I think that's what the majority of use cases require.

For that reason, I think we could split this into `tensor.one_hot(num_classes)` which would have the one-hot dim added to the last dimension and `tensor.one_hot_fill(num_classes, on_value, off_value, axis)` for a more configurable output 🤔

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2030,6 +2030,95 @@ where
...
+        }
+
+        // Convert the input tensor to integer indices
+        let indices: Tensor<B, D, Int> =
+            Tensor::from_data(self.to_data().convert::<i64>(), &device);
+
+        // Insert the new dimension for the one-hot representation
+        shape.insert(axis as usize, depth);
+
+        // Create masks for valid index range [-depth, depth-1]
```
laggui, line 2093: I am not sure I've seen a lot of negative indexing for one-hot encoding, but at this point it will cover everything so I don't have any strict opposition to this support 😅

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2030,6 +2030,95 @@ where
         // Assign the original tensor data to the appropriate slice of the padded tensor
         padded_tensor.slice_assign(ranges, self)
     }
+    /// Create a one-hot encoded tensor with configurable `depth`, `on_value`, `off_value`, and `axis` including high-ranked tensors.
+    ///
+    /// # Arguments
+    ///
+    /// * `depth`: The number of classes for the one-hot encoding, which defines the size of the one-hot dimension.
```
laggui, line 2037: Not a big fan of `depth`, I think `num_classes` is more descriptive / representative.

laggui (commented):
Ok, makes sense! Thanks for the detailed response.

I think it is especially useful for the rank > 1 use cases, the rest of the configurable stuff seems less relevant to me. But I understand that there could be value in supporting the broad spec.

See my comments below 🙂 

Regarding the multiple definitions, I think I would deprecate the other definitions since this can do it all. Just make sure to adapt the existing tests.

crates/burn-tensor/src/tensor/api/numeric.rs
tiruka: I agree with your comment, I rename it.

crates/burn-tensor/src/tensor/api/numeric.rs
tiruka: Negative index expression arguably comes from python expression, so it will help users re-use python codes directly.

crates/burn-tensor/src/tensor/api/numeric.rs
tiruka: Thank you for your suggestions, let me try modify codes, replacing existing `one_hot` functions.

tiruka:
@laggui I modified codes, please review them again (maybe after your Christmas vacation, enjoy!).

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -182,25 +179,17 @@ where
...
-
-        let mut dims = [1; D];
-        dims[D - 1] = num_classes;
-        let shape = Shape::new(dims);
-        let ranges: Vec<_> = shape.dims.iter().map(|dim| 0..*dim).collect();
-        let tensor = Tensor::zeros(shape, device);
-        let mut ranges: [core::ops::Range<usize>; D] = ranges.try_into().unwrap();
-        ranges[D - 1] = index..index + 1;
-
-        tensor.slice_assign(ranges, Tensor::ones(Shape::new([1; D]), device))
```
laggui, line 203: This function definition doesn't clash with the new "multi-purpose" definition of `one_hot`. To correctly deprecate this I think we should simply mark it for deprecation and remove it at a later time.

```rust
#[deprecated(
    since = "0.16.0",
    note = "`Tensor::one_hot(...)` will be removed in the future, please use the new `tensor.one_hot(...)` method instead"
)]
```

We could also change the docstring to include the equivalent call for the example given.

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -182,25 +179,17 @@ where
...
+    pub fn one_hot<const D2: usize>(self, num_classes: usize) -> Tensor<B, D2> {
+        check!(TensorCheck::one_hot_tensor(self.clone(), num_classes));
+        self.one_hot_fill(num_classes, 1.0, 0.0, -1)
```
laggui, line 192: (as per my previous comment)

Instead of having `one_hot` defined for int and float, we should have only one definition (and doc) in numeric.

crates/burn-tensor/src/tensor/api/int.rs
```diff
@@ -129,4 +101,29 @@ where
...
+    ///     let indices: Tensor<B, 1, Int> = Tensor::from_ints([0, 1, 2, 3], &device);
+    ///     let one_hot: Tensor<B, 4, Int> = indices.one_hot(4);
+    ///     println!("{}", one_hot.to_data());
+    ///     // [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]]
+    /// }
+    /// ```
+    pub fn one_hot<const D2: usize>(self, num_classes: usize) -> Tensor<B, D2, Int> {
+        check!(TensorCheck::one_hot_tensor(self.clone(), num_classes));
+        self.one_hot_fill(num_classes, 1.0, 0.0, -1)
+    }
```
laggui, line 128: (same comment)

Instead of having `one_hot` defined for int and float, we should have only one definition (and doc) in numeric.

burn-book/src/building-blocks/tensor.md
```diff
@@ -157,6 +157,7 @@ Those operations are available for all tensor kinds: `Int`, `Float`, and `Bool`.
 | `tensor.narrow(dim, start, length)`   | `tensor.narrow(dim, start, length)`                                       |
 | `tensor.not_equal(other)`             | `x != y`                                                                  |
 | `tensor.permute(axes)`                | `tensor.permute(axes)`                                                    |
+| `tensor.one_hot_fill(depth, on_value, off_value, axis)` | N/A                                                     |
```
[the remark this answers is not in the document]
tiruka: Thank you for your comments, I moved it.

burn-book/src/building-blocks/tensor.md
```diff
@@ -258,7 +259,7 @@ Those operations are only available for `Float` tensors.
 
 | Burn API                                      | PyTorch Equivalent                 |
 | --------------------------------------------- | ---------------------------------- |
-| `Tensor::one_hot(index, num_classes, device)` | N/A                                |
+| `tensor.one_hot(num_classes)`                 | `torch.nn.functional.one_hot`      |
```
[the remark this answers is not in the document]
tiruka: Thank you for your review, I moved it.

crates/burn-tensor/src/tensor/api/float.rs
tiruka: Thank you for your review, I moved it.

crates/burn-tensor/src/tensor/api/int.rs
tiruka: Thank you for your review, I moved it.

crates/burn-tensor/src/tensor/api/float.rs
tiruka: Thank you for your review. Okay, I followed your advice. `one_hot` name is already used in numeric.rs, so I added `_` and renamed it to `_one_hot`, due to absence of overload in rust.

tiruka:
@laggui 
I am back now and hank you for your review. I modified them.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2030,6 +2030,106 @@ where
...
+    /// # Example
+    ///
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::Tensor;
+    ///
+    /// fn example<B: Backend>(){
+    ///     let device = Default::default();
+    ///     let indices: Tensor<B, 1> = Tensor::from_floats([0.0, 1.0, 2.0, 3.0], &device);
+    ///     let one_hot: Tensor<B, 4> = indices.one_hot(4);
```
laggui, line 2044: Shouldn't the output be a 2D tensor?

```suggestion
    ///     let one_hot: Tensor<B, 2> = indices.one_hot(4);
```

crates/burn-tensor/src/tests/ops/one_hot.rs
```diff
@@ -1,74 +1,115 @@
...
 
-        let tensor = TestTensor::<1>::one_hot(1, 2, &device);
-        let expected = TensorData::from([0., 1.]);
-        tensor.into_data().assert_eq(&expected, false);
+    #[test]
+    fn float_should_support_one_hot_index() {
+        let tensor = TestTensor::<1>::from([2.0]);
+        let one_hot_tensor: Tensor<TestBackend, 1, Float> =
+            tensor.one_hot::<2>(10).flatten::<1>(0, 1);
+        let expected = TensorData::from([0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]);
```
laggui, line 28: Instead of using flatten on the output, I suggest we give the expected output as is:
```suggestion
        let one_hot_tensor = tensor.one_hot::<2>(10);
        let expected = TensorData::from([[0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]]);
```

crates/burn-tensor/src/tests/ops/one_hot.rs
```diff
@@ -1,74 +1,115 @@
...
 
     #[test]
     fn int_should_support_one_hot() {
-        let device = Default::default();
-
-        let index_tensor = TestTensorInt::<1>::arange(0..5, &device);
-        let one_hot_tensor = index_tensor.one_hot(5);
-        let expected = TestTensorInt::eye(5, &device).into_data();
+        let tensor = TestTensorInt::<1>::from([0, 1, 4]);
+        let one_hot_tensor: Tensor<TestBackend, 2, Int> = tensor.one_hot(5).int();
```
laggui, line 49: For an int input, the output should already be int

```suggestion
        let one_hot_tensor: Tensor<TestBackend, 2, Int> = tensor.one_hot(5);
```

crates/burn-tensor/src/tests/ops/one_hot.rs
```diff
@@ -1,74 +1,115 @@
...
     }
 
     #[test]
     #[should_panic]
     fn int_one_hot_should_panic_when_index_exceeds_number_of_classes() {
-        let device = Default::default();
-        let index_tensor = TestTensorInt::<1>::arange(0..6, &device);
-        let one_hot_tensor = index_tensor.one_hot(5);
+        let tensor = TestTensorInt::<1>::from([5]);
+        let result: Tensor<TestBackend, 2, Int> = tensor.one_hot(5).int();
```
laggui, line 58: Same
```suggestion
        let result: Tensor<TestBackend, 2, Int> = tensor.one_hot(5);
```

crates/burn-tensor/src/tests/ops/one_hot.rs
```diff
@@ -1,74 +1,115 @@
...
+        ]));
+
+        let one_hot_tensor: Tensor<TestBackend, 3, Int> = tensor.one_hot_fill(10, 3.0, 1.0, 1);
+
+        one_hot_tensor.into_data().assert_eq(&expected, true);
+    }
+
+    #[test]
+    fn one_hot_fill_with_negative_axis_and_indices() {
+        let tensor = TestTensorInt::<2>::from([[0, 2], [1, -1]]);
```
laggui, line 83: If the output is expected to be float, the input should be float
```suggestion
        let tensor = TestTensor::<2>::from([[0, 2], [1, -1]]);
```

crates/burn-tensor/src/tensor/api/numeric.rs
laggui: For this reason, I think we need to add this condition to the `TensorCheck`:

> Tensor of rank one greater than input tensor 'indices', i.e. rank(output) = rank(indices) + 1

In other words/code:
```rust
if D + 1 != D2
```

laggui (changes requested):
Alright should be good to go after this round! 🙂

crates/burn-tensor/src/tensor/api/check.rs
```diff
@@ -461,8 +461,8 @@ impl TensorCheck {
         check
     }
 
-    pub(crate) fn one_hot_tensor<B: Backend>(
-        index_tensor: Tensor<B, 1, Int>,
+    pub(crate) fn one_hot_tensor<B: Backend, const D: usize, K: Numeric<B>>(
```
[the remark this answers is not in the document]
tiruka: As you mentioned, this part have been there before, so I would like to keep it as it is to separate responsibility, IMO.

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -184,12 +182,28 @@ where
...
+    ///     let indices: Tensor<B, 1> = Tensor::from_floats([2.0], &device);
+    ///     let one_hot: Tensor<B, 1> = indices.one_hot::<2>(10).flatten::<1>(0, 1);
+    ///     println!("{}", one_hot.to_data());
+    ///     // [0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+    /// }
+    #[deprecated(
+        since = "0.16.0",
+        note = "`Tensor::one_hot(...)` will be removed in the future, please use the new `tensor.one_hot(...)` method instead"
+    )]
+    pub fn _one_hot(index: usize, num_classes: usize, device: &B::Device) -> Self {
```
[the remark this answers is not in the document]
tiruka: Thank you for your comment, I deleted it.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2030,6 +2030,106 @@ where
...
+    ///     println!("{}", one_hot.to_data());
+    ///     // [[1.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, 0.0], [0.0, 0.0, 1.0, 0.0], [0.0, 0.0, 0.0, 1.0]]
+    /// }
+    /// ```
+    pub fn one_hot<const D2: usize>(self, num_classes: usize) -> Tensor<B, D2> {
```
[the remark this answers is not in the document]
tiruka: Thank you for your comment, I modified accordingly.

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2030,6 +2030,106 @@ where
...
+    ///     // [0.0, 0.0, 5.0]]]
+    /// }
+    /// ```
+    pub fn one_hot_fill<K2: Numeric<B>, const D2: usize>(
+        self,
+        num_classes: usize,
+        on_value: f32,
+        off_value: f32,
+        axis: i64,
+    ) -> Tensor<B, D2, K2> {
```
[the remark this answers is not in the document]
tiruka: Thank you for your comment, I modified accordingly.

crates/burn-tensor/src/tensor/api/numeric.rs
tiruka: Thank you for your comment, I modified and added check function accordingly.

crates/burn-tensor/src/tests/ops/one_hot.rs
tiruka: Thank you for your comment, I modified accordingly.

crates/burn-tensor/src/tests/ops/one_hot.rs
tiruka: Thank you for your comment, I modified accordingly.

crates/burn-tensor/src/tests/ops/one_hot.rs
tiruka: Thank you for your comment, I modified accordingly.

crates/burn-tensor/src/tests/ops/one_hot.rs
tiruka: Thank you for your comment, I modified accordingly.

laggui (approved):
LGTM! Made a minor change to the tensor check message

tiruka:
@laggui Thank you for your much help! I really appreciate it to complete this pr. 👍

PR #2620, opened 2024-12-16
Relax Fn requirements

ArthurBrussee:
### Changes

For param.map() I really want a FnOnce so I can move captured variables. Initialization also seems to better fit with a FnOnce.


laggui (approved):
Makes sense 🙂

PR #2622, opened 2024-12-17
Fused/matmul

nathanielsimard:
The only missing part is autotune, it needs to be done before making a new burn release.


syl20bnr:
I fixed a couple of things while debugging it but the main fix was just a `cargo update`.

backend-comparison/benches/matmul.rs
```diff
@@ -21,15 +21,17 @@ impl<B: Backend, const D: usize> Benchmark for MatmulBenchmark<B, D> {
         vec![self.shape_lhs.dims.clone(), self.shape_rhs.dims.clone()]
     }
 
-    fn execute(&self, (lhs, rhs): Self::Args) {
-        lhs.matmul(rhs);
+    fn execute(&self, (lhs, rhs, bias): Self::Args) {
+        let bias = bias.unsqueeze();
+        relu(lhs.matmul(rhs) + bias);
```
louisfd, line 26: Should be another benchmark "fused matmul" and leave matmul like it was.

louisfd (approved):
Looks good, except matmul benchmark should be split between solely matmul, and matmul fused

PR #2623, opened 2024-12-17
burn-import: add some tests for ConstantNode

jameshiew:
### Checklist

### Related Issues/PRs

I've been trying to implement OneHot ONNX op (https://github.com/tracel-ai/burn/issues/1714) in a [WIP draft branch](https://github.com/jameshiew/burn/pull/1). The ONNX model ends up containing a constant integer vector `values=[0, 1]` used by the OneHot op, this vector was causing issues when trying to test the model. I looked at `ConstantNode` and these are the tests so far I could get working while investigating.

Issues for `ConstantNode`
https://github.com/tracel-ai/burn/issues/2624 - constant tensors aren't populated with values
https://github.com/tracel-ai/burn/issues/2625 - generated code for const int tensors doesn't compile

### Changes

- added a helper method `ConstantNode::tensor_ty_tokens` for tests, but this PR otherwise shouldn't be changing how `ConstantNode` currently works
- add codegen tests for ConstantNode (i32/64 + f32/64 scalar, tensors)
- add ONNX model tests for i32/64 + f32/64 scalars - implicitly testing by adding the constant

### Testing

Ran added tests

```shell
cargo xtask check all
cargo nextest run --manifest-path crates/burn-import/Cargo.toml
cargo nextest run --manifest-path crates/burn-import/onnx-tests/Cargo.toml
```

I checked the .onnx models contain the expected scalar constants using Netron

<details>
<summary>Screenshots</summary>
f32
f64
i32
i64
</details>


crates/burn-import/onnx-tests/tests/test_onnx.rs
```diff
@@ -2183,6 +2187,46 @@ mod tests {
...
+        let output = model.forward(input.into());
+        assert_eq!(expected_output, output.to_data());
+    }
+
+    #[test]
+    fn add_constant_f64() {
+        let device = Default::default();
+        let model = constant_f64::Model::<Backend>::new(&device);
+        let input = TensorData::zeros::<f64, _>(Shape::from([2, 3, 4]));
+        let expected_output = TensorData::full(Shape::from([2, 3, 4]), 2f32);
```
jameshiew, line 2205: I'm not sure if the addition in the model is coercing f64 -> f32 (and i32 -> i64 below). I wasn't sure how to get PyTorch to just forward the constant so these tests use the add operation also

laggui: Maybe by having the output return the constant only? But a simple constant addition works too.

In case you're curious, you could also manually define the onnx graph like [the ConstOfShape script](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/onnx-tests/tests/constant_of_shape/constant_of_shape.py). PyTorch tends doesn't always have a 1-to-1 correspondence for ops, so in such cases it could be easier to define the graph manually.

> I'm not sure if the addition is coercing f64 -> f32 somewhere (and i32 -> i64 below)

The floating point and integer data types are defined by the backend used. A model is not as statically defined like an ONNX graph. If you look at the other tests, the input(s) and output(s) are created using the `Tensor` methods, not from `TensorData`.

jameshiew: Thanks for the extra context, sorry it's taken me a bit to get back to this PR! I've rebased on latest `main` and updated these tests to use `Tensor` instead of `TensorData`. If I understand correctly now - the Burn backend used in these tests (`type Backend = burn_ndarray::NdArray<f32>`) uses fixed element types f32 and i64 - so these types will be expected in the final output data even if the ONNX model is using e.g. f64 or i32.

laggui: > sorry it's taken me a bit to get back to this PR

No worries 🙂 

> the Burn backend used in these tests (type Backend = burn_ndarray::NdArray<f32>) uses fixed element types f32 and i64 - so these types will be expected in the final output data even if the ONNX model is using e.g. f64 or i32.

That's correct.

We actually support different floating point types now with `tensor.cast(dtype)` (so even if a backend uses `f32` as the default floating point type, tensors could be cast to another precision). But for ONNX import, we currently don't take the precision dtype into account.

jameshiew: Thanks, feel free to take anything from this PR or close if easier, I probably can't contribute to it in the near future but would like to come back to looking at Burn sometime!

laggui: Actually the tests look good! Guess I forgot to add a comment last time I checked, but my only comment would be to reduce the functions use in the tests (keep it as constant / static as possible).

Was there a particular reason for retrieving the types in the tests, e.g.

```rust
let con = const_tensor.to_token_stream();
let ty = constant.ty_tokens();
let tensor_ty = constant.tensor_ty_tokens();
let shp = shape_to_tokens(&shape);
```

and then using them to test the generated code?

I would keep the expected output definition constant like the other codegen tests.

laggui (approved):
Was going through pending PRs, figured I'd apply the suggested changes myself 😅 so we can finally close this

PR #2648, opened 2025-01-02
add fscore, reorganize mod and change error in doc.

tsanona:
### Checklist

### Related Issues/PRs
#544

### Changes

Add F<sub>β</sub>ScoreMetric
Reorganize `metric\mod` to mach existing order.

### Testing
Added tests for new metric

### DIFF
--- a/burn-book/src/building-blocks/metric.md
+++ b/burn-book/src/building-blocks/metric.md
@@ -9,7 +9,7 @@ throughout the training process. We currently offer a restricted range of metric
 | TopKAccuracy     | Calculate the top-k accuracy in percentage              |
 | Precision        | Calculate precision in percentage                       |
 | Recall           | Calculate recall in percentage                          |
-| FScore           | Calculate f<sub>β</sub>score in percentage                       |
+| FBetaScore           | Calculate F<sub>β </sub>score in percentage         |
 | AUROC            | Calculate the area under curve of ROC in percentage     |
 | Loss             | Output the loss used for the backward pass              |
 | CPU Temperature  | Fetch the temperature of CPUs                           |
--- a/burn-book/src/building-blocks/metric.md
+++ b/burn-book/src/building-blocks/metric.md
@@ -4,11 +4,12 @@ When working with the learner, you have the option to record metrics that will b
 throughout the training process. We currently offer a restricted range of metrics.
 
 | Metric           | Description                                             |
-|------------------|---------------------------------------------------------|
+| ---------------- | ------------------------------------------------------- |
 | Accuracy         | Calculate the accuracy in percentage                    |
 | TopKAccuracy     | Calculate the top-k accuracy in percentage              |
 | Precision        | Calculate precision in percentage                       |
 | Recall           | Calculate recall in percentage                          |
+| FScore           | Calculate f<sub>β</sub>score in percentage                       |
 | AUROC            | Calculate the area under curve of ROC in percentage     |
 | Loss             | Output the loss used for the backward pass              |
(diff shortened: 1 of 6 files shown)


burn-book/src/building-blocks/metric.md
```diff
@@ -4,11 +4,12 @@ When working with the learner, you have the option to record metrics that will b
 throughout the training process. We currently offer a restricted range of metrics.
 
 | Metric           | Description                                             |
-|------------------|---------------------------------------------------------|
+| ---------------- | ------------------------------------------------------- |
 | Accuracy         | Calculate the accuracy in percentage                    |
 | TopKAccuracy     | Calculate the top-k accuracy in percentage              |
 | Precision        | Calculate precision in percentage                       |
 | Recall           | Calculate recall in percentage                          |
+| FScore           | Calculate f<sub>β</sub>score in percentage                       |
```
laggui, line 12: Suggestion: rename to `FBetaScore`. It seems that many people refer to [F-score](https://en.wikipedia.org/wiki/F-score) as F-beta score (as it is parameterized by beta).

See sklearn for example: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.fbeta_score.html

laggui: Also, can you fix the trailing spaces that misalign the right table border 😅

laggui (commented):
Thanks for contributing yet another metric 😄 

Minor comment regarding the name, otherwise should be good to go!

burn-book/src/building-blocks/metric.md
tsanona: Yeah I was torn and chose the smallest name, but I also agree with this, will change!
Ah yes will format the table didn't realize it was misaligned.

tsanona: While I agree the F should be capitalized, it feels like it isn't consistent with the other metric names, should I also change the others or just capitalize the F?

tsanona:
All good, it's been fun to contribute! :star2: I have a few more ideas now to adapt the existing (accuracy+topkaccuracy and auroc) ones to the same format as the ones I've implemented. Is there someplace where we can better discuss on how to do this ? (since it will be braking code and I'm not sure how to handle that)

burn-book/src/building-blocks/metric.md
laggui: I think it's consistent. The F-score is always capitalized (based on its [origin](https://en.wikipedia.org/wiki/F-score#Etymology)), ROC is an acronym for Receiver Operating Characteristic. Other names are not capitalized because they're really just nouns (precision, recall, accuracy). I guess there could be a debate over top-k vs Top-k, but even [wikipedia](https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Average_precision) doesn't capitalize it 😅

laggui:
> I have now plans to adapt the existing metrics (accuracy+topkaccuracy, auroc and hamming) to the same format as the ones I've implemented. (this will be breaking code, which I'm not fully sure on how to handle)

Not sure what do yo have in mind?

We try to limit breaking changes, but sometimes that's not possible. Depends on the objective I guess!

> I also want to add a few more losses and, `or` and `and` ops to Bool tensors and `scatter` and `select*` ops to all tensors (not just numeric). If there's some priority/no necessity of any of these I'd be interested in discussing 🚀 Should I just create prs or is there a better way to start these discussions?

I think missing features are mostly captured in issues. If a feature request is not already filed, you can open an issue for the request and the discussion can start there 🙂 

For other general dev discussions and thoughts, this usually happens on discord.

laggui (approved):

PR #2099, opened 2024-08-04
Add cumsum

cBournhonesque:
## Pull Request Template

I'm opening the PR draft early just to see if i'm on the right track!

### Checklist

- [ ] Confirmed that `run-checks all` script has been executed.
- [x] Made sure the book is up to date with changes in this PR.
- [x] ndarray
- [x] candle
- [x] tch
- [x] autodiff
- [ ] jit
- [x] fusion
- [ ] import

### Related Issues/PRs

Issue: https://github.com/tracel-ai/burn/issues/1720

### Changes

Add cumsum for ndarray

### Testing

Added unit test

PR #2664, opened 2025-01-06
Add cumsum tensor op

laggui:
### Checklist

### Related Issues/PRs

#1722 #2099

### Changes

Added `tensor.cumsum(dim)` operation

TODO:
- [ ] JIT implementation w/ cubecl kernel
  - We have `cubecl-reduce` at this time but we might need to add `cubecl-scan` for this family of ops (some refs: [wgsl](https://github.com/b0nes164/GPUPrefixSums) [simd](https://en.algorithmica.org/hpc/algorithms/prefix/) [cuda](https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda))

### Testing

Unit test.


mush42:
This op is very important for many NAR TTS models, and it is blocking porting one of my open-source TTS models to burn.

antimora:
So much has changed, we should probably create a new PR by now.

PR #2667, opened 2025-01-07
[burnbench] Import code from github-device-flow crate

### LINKED ISSUE #2548: RUSTSEC-2024-0370: proc-macro-error is unmaintained
| Details |  |
| --- | --- |
| Package | `proc-macro-error` |
| Version | `1.0.4` |
| Warning | unmaintained |
| URL | https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/20 |
| Patched Versions | n/a |

proc-macro-error's maintainer seems to be unreachable, with no commits for 2 years, no releases pushed for 4 years, and no activity on the GitLab repo or response to email.

proc-macro-error also depends on `syn 1.x`, which may be bringing duplicate dependencies into dependant build trees.

## Possible Alternative(s)

- [manyhow](https://crates.io/crates/manyhow)
- [proc-macro-error2](https://crates.io/crates/proc-macro-error2)
- [proc-macro2-diagnostics](https://github.com/SergioBenitez/proc-macro2-diagnostics)

syl20bnr:
We only use some of the code of the crate and some of its code is redundant with our own implementation.

Moreover we don't need some of its dependencies like clap which is on a previous major version. So this allows to remove a whole dependency tree from the lock file.

Fixes #2548

### Testing

Re-authenticate with GitHub.

### DIFF
--- a/backend-comparison/src/burnbenchapp/auth/github_device_flow.rs
+++ b/backend-comparison/src/burnbenchapp/auth/github_device_flow.rs
@@ -1,0 +1,230 @@
+// Initially from: https://github.com/jakewilkins/gh-device-flow
+use std::collections::HashMap;
+use std::{fmt, result::Result, thread, time};
+
+use chrono::offset::Utc;
+use chrono::{DateTime, Duration};
+use serde::{Deserialize, Serialize};
+
+pub fn credential_error(msg: String) -> DeviceFlowError {
+    DeviceFlowError::GitHubError(msg.into())
+}
+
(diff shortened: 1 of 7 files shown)


backend-comparison/src/burnbenchapp/auth/github_device_flow.rs
```diff
@@ -0,0 +1,232 @@
+// Initialy from: https://github.com/jakewilkins/gh-device-flow
```
laggui, line 1: Typo: Initially

laggui (approved):
LGTM!

PR #2670, opened 2025-01-08
Fix/autotune error handling

nathanielsimard:


crates/burn-jit/src/kernel/conv/conv2d/tune/conv2d.rs
```diff
@@ -173,19 +145,39 @@ fn should_run<R: JitRuntime, F: FloatElement>(
         // GEMM balanced
         4 => check_algo!(Balanced, F, &op.input, &conv_problem),
         _ => true,
-    }
+    };
+
+    println!("{index} => {should_run}");
```
maxtremblay, line 150:
```suggestion
```

crates/burn-jit/src/kernel/conv/conv2d/tune/conv2d.rs
```diff
@@ -173,19 +145,39 @@ fn should_run<R: JitRuntime, F: FloatElement>(
...
     let cube_count = ImplicitCmmaConv::cube_count(&selection, conv_problem);
 
+    let max_cube_dim = 65535;
+
+    if let cubecl::CubeCount::Static(x, y, z) = cube_count {
+        println!("({x}, {y}, {z})");
+        if x > max_cube_dim || y > max_cube_dim || z > max_cube_dim {
+            return false;
+        }
+    }
```
maxtremblay, line 179:
```suggestion
    let max_cube_dim = 65535; // This is the maximum cube count on wgpu

    if let cubecl::CubeCount::Static(x, y, z) = cube_count {
        if x > max_cube_dim || y > max_cube_dim || z > max_cube_dim {
            return false;
        }
    }
```

maxtremblay: Also, it would be better if this was either a constant or query from the device.

maxtremblay (approved):

PR #2681, opened 2025-01-12
Use float intrinsics for deform_conv2d backward, fix `into_data` for padded tensors

wingertge:
### Checklist

### Changes

Uses float intrinsics instead of CAS where available - `f32` is currently disabled because of a bug in CUDA I need to investigate. `f16` and `bf16` work fine.
Fixes `into_data` for padded tensors by slicing to the expected length before creating the `TensorData`.

### Testing

All tests pass except for `remainder` on wgpu/spirv, which is still broken due to a driver bug.

### DIFF
--- a/crates/burn-jit/src/kernel/conv/deform_conv_transpose2d.rs
+++ b/crates/burn-jit/src/kernel/conv/deform_conv_transpose2d.rs
@@ -5,8 +5,10 @@ use burn_tensor::{
     Shape,
 };
 use cubecl::{
-    calculate_cube_count_elemwise, cube, ir::Elem, prelude::*, AtomicFeature, CubeDim, CubeLaunch,
-    Feature,
+    calculate_cube_count_elemwise, cube,
+    ir::{Elem, FloatKind},
+    prelude::*,
+    AtomicFeature, CubeDim, CubeLaunch, Feature,
 };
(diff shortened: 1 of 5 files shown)


crates/burn-jit/src/kernel/conv/deform_conv_transpose2d.rs
```diff
@@ -558,47 +596,78 @@ fn deform_col2img_kernel<F: Float>(
...
     let y =
-        f32::cast_from(out_y * args.stride_h + kernel_y * args.dilation_h) - args.pad_h + offset_y;
+        F::cast_from(out_y * args.stride_h + kernel_y * args.dilation_h) - args.pad_h + offset_y;
     let x =
-        f32::cast_from(out_x * args.stride_w + kernel_x * args.dilation_w) - args.pad_w + offset_x;
+        F::cast_from(out_x * args.stride_w + kernel_x * args.dilation_w) - args.pad_w + offset_x;
 
     for dy in -1..=1 {
-        #[unroll]
+        //#[unroll]
```
laggui, line 604: ☝️

laggui (commented):
Might have missed a comment from debugging

laggui (approved):

PR #2692, opened 2025-01-13
implement benchmark for reduce kernel

maxtremblay:
This implements a bench marking utilities for the reduce kernel.


louisfd (approved):
LGTM

Issue #2701, opened 2025-01-15
GRU implementation does not use hidden state in iteration, differs from PyTorch


nwhitehead:
**Describe the bug**

At a high level, the GRU implementation gives different results than PyTorch. I believe there are two parts to this.

(1) The formula used for the "new" gate, `g(t)`, is implemented following the original GRU paper. PyTorch uses a slightly different ordering which changes results but is supposedly more efficient. [1] (See 4th NOTE)

(2) The GRU implementation of `forward` iterates over the sequence length of the input and computes hidden state updates, but computations at time `t` do not see hidden state updates to time `t-1`. This means the output is wrong for all sequence positions after the first position.

[1]: https://pytorch.org/docs/stable/generated/torch.nn.GRU.html

**To Reproduce**

Create a trivial GRU module with `input_size=2` and `hidden_size=1` and random initialization, and apply to random tensor with shape `[1, 2, 2]` (batch, sequence length, input features).

Observe that both outputs `h_0` and `h_1` do not match PyTorch.

**Expected behavior**

I think the most useful thing is to match the behavior of PyTorch as much as possible.

**Exploratory Fixes**

Fixing issue (1) makes the first output match PyTorch but the second output still does not match. Fixing issue (1) and (2) makes both outputs match.

Fixing issue (1) involved extending `gate_product` in `gru.rs` to accept an optional `reset` argument, then:
```rust
match (input_bias, hidden_bias, reset) {
    (Some(input_bias), Some(hidden_bias), Some(r)) => {
        input_product + input_bias.unsqueeze() + r.clone().mul(hidden_product + hidden_bias.unsqueeze())
    }
...
```
And calling it:
```rust
            let biased_ng_input_sum = self.gate_product(&input_t, &hidden_t, Some(&reset_values), &self.new_gate);
```

I'm not sure of the proper fix for issue (2). My test code does this in `gru.rs`:
```rust
        for (t, (input_t, hidden_t)) in batched_input
            .iter_dim(1)
            .zip(hidden_state.clone().iter_dim(1))
            .enumerate()
        {
            let input_t = input_t.squeeze(1);
            let hidden_t = hidden_t.squeeze(1);
            let hidden_t = if t > 0 {
                hidden_state.clone().slice([0..batch_size, (t - 1)..t, 0..self.d_hidden]).squeeze(1)
            } else {
                hidden_t
            };
```
[...]

PR #2704, opened 2025-01-15
Fix GRU to match pytorch (#2701).

nwhitehead:
### Checklist

### Related Issues/PRs

This addresses issue #2701

### Changes

* Update GRU implementation of "new" gate to match pytorch implementation. This can change numerical output in some cases.

* Add GRU unit test with sequence length > 1.

* Fix GRU input state dimensions and hidden state handling. This is an API change since the dimensions of the optional hidden state input are being corrected to the right sizes.

These changes do affect numerical results and change the API slightly. I think just updating to the correct API dimensions seems like the best thing since the previous implementation was incorrect, not just different than pytorch.

### Testing

These changes were tested with a small unit test. For this test the correct values were computed manually using the equations for GRU.

I tested these changes against PyTorch. The weights and biases from PyTorch were saved then split into sections using a custom script (to split apart the weights for each gate). Input and output tensors were separately saved and then loaded into a test rust program. Everything was randomly initialized. With this PR the results from burn and torch were almost identical (within 6 decimal digits). I tried input sizes of 1, 2, and 8. I tried hidden sizes of 1, 2, and 8. I tried sequence lengths of 1, 2, and 3.

### DIFF
--- a/crates/burn-core/src/nn/rnn/gru.rs
+++ b/crates/burn-core/src/nn/rnn/gru.rs
@@ -24,10 +24,10 @@ pub struct GruConfig {
     ///
     /// This configuration option controls how the reset gate is applied to the hidden state.
     /// * `true` - (Default) Match the initial arXiv version of the paper [Learning Phrase Representations using RNN Encoder-Decoder for
-    /// Statistical Machine Translation (v1)](https://arxiv.org/abs/1406.1078v1) and apply the reset gate after multiplication by
-    /// the weights. This matches the behavior of [PyTorch GRU](https://pytorch.org/docs/stable/generated/torch.nn.GRU.html#torch.nn.GRU).
+    ///   Statistical Machine Translation (v1)](https://arxiv.org/abs/1406.1078v1) and apply the reset gate after multiplication by
+    ///   the weights. This matches the behavior of [PyTorch GRU](https://pytorch.org/docs/stable/generated/torch.nn.GRU.html#torch.nn.GRU).
     /// * `false` - Match the most recent revision of [Learning Phrase Representations using RNN Encoder-Decoder for Statistical Machine
-    /// Translation (v3)](https://arxiv.org/abs/1406.1078) and apply the reset gate before the weight multiplication.
+    ///   Translation (v3)](https://arxiv.org/abs/1406.1078) and apply the reset gate before the weight multiplication.
     ///
     /// The differing implementations can give slightly different numerical results and have different efficiencies. For more
     /// motivation for why the `true` can be more efficient see [Optimizing RNNs with Differentiable Graphs](https://svail.github.io/diff_graphs).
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/nn/rnn/gru.rs
```diff
@@ -100,80 +100,81 @@ impl GruConfig {
...
-            let biased_rg_input_sum = self.gate_product(&input_t, &hidden_t, &self.reset_gate);
+            let biased_rg_input_sum =
+                self.gate_product(&input_t, &hidden_t, None, &self.reset_gate);
             let reset_values = activation::sigmoid(biased_rg_input_sum); // Colloquially referred to as r(t)
-            let reset_t = hidden_t.clone().mul(reset_values); // Passed as input to new_gate
 
             // n(ew)g(ate) tensor
-            let biased_ng_input_sum = self.gate_product(&input_t, &reset_t, &self.new_gate);
+            let biased_ng_input_sum =
+                self.gate_product(&input_t, &hidden_t, Some(&reset_values), &self.new_gate);
```
laggui, line 142: Since both implementations are valid, what do you think about having a field like `reset_after` to preserve both. And we could default to `reset_after = true` 

Correct me if I am wrong, but with your changes this could be an easy condition:
```rust
let biased_ng_input_sum = if self.reset_after {
  self.gate_product(&input_t, &hidden_t, Some(&reset_values), &self.new_gate)
} else {
  let reset_t = hidden_t.clone().mul(reset_values); // Passed as input to new_gate
  self.gate_product(&input_t, &reset_t, None, &self.new_gate);
}
```

And document the differences between both in the docstring.

Lmk what you think.

laggui (commented):
Thanks for bringing this up and tackling it yourself 🙏 

Fun fact, the reset gate changes don't originate from pytorch 😄 

The implementation we had is based on the latest [v3](https://arxiv.org/abs/1406.1078v3) revisions (published at EMNLP) and has reset gate applied to hidden state before matrix multiplication.

The changes in your PR are based on the original [v1](https://arxiv.org/abs/1406.1078v1) and applies the reset gate after.

Curiously, pytorch notes efficiency for their differing implementation (without much explanation). If you're interested, check out this [awesome explanation](https://svail.github.io/diff_graphs/#sec-variant-3--move-the-reset-gate) behind the motivation to move the reset gate.

Your implementation LGTM! But what do you think about supporting both via a config? And we could provide the references I just linked in the doc.

crates/burn-core/src/nn/rnn/gru.rs
nwhitehead: Yes, that seems like a good idea to me. I added a commit to do this.

laggui (approved):
Thank you 🙏 

I just fixed the clippy doc issue and covered both versions in the tests.

PR #2705, opened 2025-01-15
Clean up -jit suffix in feature flags and modules

laggui:
### Checklist

### Changes

Feature flags:
- `"cuda-jit"` -> `"cuda"`
- `"hip-jit"` -> `"hip"`
- `"wgpu-spirv"` -> `"vulkan"` (for `burn` and `burn-core`, which enables `burn-wgpu/spirv`)

Modules:
- `burn::backend::cuda_jit` -> `burn::backend::cuda`
- `burn::backend::hip_jit` -> `burn::backend::hip`

Backends:
- `CudaJit` -> `Cuda`
- `HipJit` -> `Hip`


syl20bnr (approved):
Nice! 
Tested on the AMD runner.

You can update the commented tests in `burn-hip/src/lib.rs` to match the CUDA ones. I did not execute them all, I just verified that there was no compilation issues.

```rs
mod tests {
    use burn_jit::JitBackend;

    pub type TestRuntime = cubecl::hip::HipRuntime;
    pub use half::f16;

    burn_jit::testgen_all!([f16, f32], [i8, i16, i32, i64], [u8, u32]);
}
```

nathanielsimard (commented):
I think we should create a type alias for vulkan.

```rust
pub type Vulkan = Wgpu<Compiler = Spirv>;
```

Also, I would keep a feature flag `spirv` in `burn_wgpu`, so that users can customize the compiler used for `Wgpu` without using the type alias.

Also, the `Wgpu` backend should not have a default compiler, I think. We could create a `WebGPU` type alias like `Vulkan` with the default compiler being `Wgsl`.

So, we can have a hierarchy of types, where we don't compromise on flexibility but make it clearer for `Burn` users.

New feature flags:

```toml
vulkan = ["burn_wgpu", "spirv"]
webgpu = ["burn_wgpu", "burn_wgpu/wgsl"]
wgpu = ["burn_wgpu"]
```
We have to make it clear that the `Wgpu` backend is decoupled from the compiler; users can create their own feature flag to choose the compiler if they want it to change depending on the platform. We could (and probably should) create a composed compiler that will become the default `Wgpu` compiler, where each compiler feature flag activates a new compiler option to be determined at runtime. However, for now, we can't do that just yet, and users won't have the best performance by default because of that, and it's not clear how to do that right now.

Also, having the possibility to just have one compiler activated will always be a good option to reduce binary size and ease deployment.

laggui:
@nathanielsimard Just made the suggested changes, though I kept the default compiler to wgsl for now to keep compatibility with the current `Wgpu` usage.

nathanielsimard (approved):

PR #2707, opened 2025-01-16
Clean up train system metrics

laggui:
### Checklist

### Changes

Changed the `"metrics"` flag name to `"sys-metrics"` in `burn-train`, which only includes the *system* related metrics (as they require additional dependencies). This was the original intent for the feature flag (as documented at the burn root feature flag level). The other training metrics can automatically be included.


antimora (approved):
LGTM. Seems a straight forward change.

nathanielsimard (approved):

PR #2708, opened 2025-01-16
Fix for ONNX code generation bug #2702

### LINKED ISSUE #2702: ONNX import generates integer for Interpolate2DConfig when it should be f32
**Describe the bug**

The generated code for this onnx model: https://github.com/opencv/opencv_zoo/blob/main/models/face_detection_yunet/face_detection_yunet_2023mar.onnx

contains this code:

```rust
        let resize2 = Interpolate2dConfig::new()
            .with_output_size(None)
            .with_scale_factor(Some([2, 2]))
            .with_mode(InterpolateMode::Nearest)
            .init();
```

This doesn't compile since scale factor should be f32.

@laggui states:

> Ahhh the conversion to a token stream uses f32.to_string(), but if the value doesn't have a decimal point the compiler complains that it's not f32.

https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/burn/codegen.rs#L9

**Expected behavior**

The code should compile.

 - OS: Arch
 - Version: Burn 93cafc41b509e618ca46b680b3015659b97a0de3 0.17.0

**Additional context**

Discord chat context: https://discordapp.com/channels/1038839012602941528/1059209073784008835/1329098432580816936

antimora:
### Checklist

### Related Issues/PRs

Fixes #2702

### Changes

Instead of using `to_string()` substituted with `format!` macro.

### Testing

Added a new test case to render `2.0`

### DIFF
--- a/crates/burn-import/src/burn/codegen.rs
+++ b/crates/burn-import/src/burn/codegen.rs
@@ -5,8 +5,9 @@ use burn::nn::PaddingConfig1d;
 use burn::nn::PaddingConfig2d;
 use burn::nn::PaddingConfig3d;
 
-fn convert_primitive<T: ToString>(primitive: T) -> TokenStream {
-    let value = primitive.to_string();
+fn convert_primitive<T: core::fmt::Debug>(primitive: T) -> TokenStream {
+    let value = format!("{:?}", primitive);
+
     value.parse().unwrap()
 }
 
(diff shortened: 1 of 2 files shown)


laggui (approved):
LGTM!

torsteingrindvik:
Can confirm it fixed my issue, thanks!

Issue #1187, opened 2024-01-29
Support importing GGUF files


richardanaya:
I apologize if this seems too far fetched, but it seemed in line with how ONNX generation works.

antimora:
If gguf contains the model graph information, then we can use what burn-import ONNX facility. In our burn-import, we convert ONNX graph to IR (intermediate representation) (see [this doc](https://github.com/tracel-ai/burn/blob/main/burn-import/DEVELOPMENT.md)). So, it would possible to convert the model graph to IR and generate source code + weights. 

If gguf contains only weights, we can go burn-import pytorch route, where we only download weights.

antimora:
From my brief research,  GGUF format contains metadata + tensor weights. This aligns with burn-import pytorch route and not burn-import/ONNX

Here is one Rust lib to parse GGUF file: https://github.com/Jimexist/gguf

antimora:
GGUF spec: https://github.com/ggerganov/ggml/pull/302

antimora:
Parser in Rust: https://github.com/Jimexist/gguf

leflambeur:
Hi, it has been about a year since this was last updated, since then pre-existing models on HF often come in GGUF for quantised or Safetensor formats for non-quantised.

I think it would be useful to people new to the space to understand how burn can be leveraged with these formats, as they seem to be the most common formats available to start from scratch.

Specifically importing quantised GGUF models as I couldn't see much in the docs.

Candle is okay for this, but its support for models is a little spotty with quantised models which are more accessible to people with fewer resources.

I saw in #1323 some added pieces were available for reconstructing config files but I am wondering about simply ingesting a gguf model and using it with burn directly similar to the import options for onnx or pytorch without people needing to figure out how to reverse engineer what gguf is doing under the hood with little guidance.

GGUF's single file format seems like an ideal target for burn's use-case to me and the format is much more universally accessible, similar to ONNX on paper.

I am happy to contribute docs, I just need a bit of direction to start testing with the current capabilities or indicators that it is even possible.

antimora:
I'll be happy to assist if you decide to submit a pr. 

We can leverage Candle's reader similar PyTorch pt reader. We can use the existing burn-import infrastructure.  It should be somewhat easier now that PyTorch pt import works.

leflambeur:
[...]

PR #2721, opened 2025-01-20
Support importing safetensors format

### LINKED ISSUE #626: Support importing safetensors format
Safetensors (de)serialization

### Feature description

Load, at either tensor- or module-level, weights from a `.safetensors` file.

### Feature motivation

To be able to tap into the huggingface ecosystem it would be useful to have the ability to (de)serialize for [`safetensors`](https://github.com/huggingface/safetensors). This would obviously have the limitation of these files merely containing a key-value JSON for all weights, and no description of the model's topology/architecture just like ONNX does. Ideally this would help grow the [Model Zoo](https://github.com/burn-rs/models).

This would then mostly be a convenience feature, both for sharing and using existing models. 

Possibly we could start by deriving the `TryFrom` trait for their `TensorView` and our `Tensor` and take it from there. Huggingface's `candle` (Rust) framework [already has the capability to (de)serialize from `.safetensors`](https://github.com/huggingface/candle#how-to-use-).

wandbrandon:
### Checklist

### Related Issues/PRs

Fixes #626

### Changes

This pull request introduces support for the SafeTensors format in the Burn library, enabling secure and efficient model weight imports. It includes updates to documentation, new examples, and test cases to demonstrate the usage of SafeTensors. The changes also enhance the project's modularity by adding SafeTensors-specific features and dependencies.

### Support for SafeTensors Format:
* Added SafeTensors support to `burn-import` with a new feature flag and dependencies in `Cargo.toml`. This includes updates to the default features to enable SafeTensors by default. 

### Documentation Updates:
* Updated the Burn Book to include a new section on SafeTensors in the summary and import documentation. This includes a detailed guide for exporting and importing SafeTensors weights, troubleshooting, and advanced features like key remapping and framework-specific adapters. 
* Enhanced the README with information about SafeTensors support, including updated descriptions for importing models and links to relevant guides.

### Codebase Adjustments:
* Removed outdated examples and added a new example directory for importing model weights, aligning with the new SafeTensors functionality.

These changes collectively improve the Burn library's flexibility and usability, particularly for users looking to leverage the SafeTensors format for secure and efficient model weight handling.

### Testing

Added end-to-end safetensors-tests.


crates/burn-import/src/pytorch/reader.rs
```diff
@@ -42,11 +42,22 @@ where
     PS: PrecisionSettings,
     B: Backend,
 {
-    // Read the pickle file and return a vector of Candle tensors
-    let tensors: HashMap<String, CandleTensor> = pickle::read_all_with_key(path, top_level_key)?
-        .into_iter()
-        .map(|(key, tensor)| (key, CandleTensor(tensor)))
-        .collect();
+    //check if it's a safetensors file
+    let is_safetensors = path.extension().is_some_and(|ext| ext == "safetensors");
```
laggui, line 46: Not sure if that's a very robust way to differentiate between both 😅 pretty sure I've seen a lot of `.safetensor` files (without the plural form).

I think we should add a field to the `LoadArgs` instead, and users can then specify when they're loading a safetensor file:

```rust
LoadArgs::new(...).with_safetensors(true)
```

examples/pytorch-import/build.rs
```diff
@@ -17,7 +17,7 @@ fn main() {
 
     // Load PyTorch weights into a model record.
     let record: model::ModelRecord<B> = PyTorchFileRecorder::<FullPrecisionSettings>::default()
-        .load("pytorch/mnist.pt".into(), &device)
+        .load("pytorch/mnist.safetensors".into(), &device)
```
laggui, line 20: To check if we should load the pickle file or safetensor, we can add a `safetensors` feature flag to the example and check it here with something like:

```rust
let ext = if std::env::var("CARGO_FEATURE_SAFETENSORS").is_ok() {
  "safetensors"
} else {
  "pt"
};
```

and then load the correct file.

We should update the README with a small mention.

laggui (changes requested):
Thanks for the addition 🙏 

Looks pretty good overall, just some minor comments.

Nikaidou-Shinku:
IMO maybe we can have something like `pub mod safetensors;` under a new feature gate in crate `burn-import` so users could build exactly what they need, since safetensors does not seem to be a format strongly related to PyTorch.

wandbrandon:
> IMO maybe we can have something like `pub mod safetensors;` under a new feature gate in crate `burn-import` so users could build exactly what they need, since safetensors does not seem to be a format strongly related to PyTorch.

I think this is a good point, and it also builds the scaffolding for potentially rewriting it to remove the Candle dependency.

antimora (changes requested):
> IMO maybe we can have something like `pub mod safetensors;` under a new feature gate in crate `burn-import` so users could build exactly what they need, since safetensors does not seem to be a format strongly related to PyTorch.


Yes, I strongly agree with this opinion. My intention was that we have a separate support for different formats. We can reuse common code in burn-import but entry point should be different. We need a separate module and feature for GGUF files, see https://github.com/tracel-ai/burn/issues/1187, and it would follow the same pattern. 

PyTorch's `pt` and Safetensors are unrelated. We should not mix up in code or documention as such. It just happens we're using Candle's reader but it could be different.

antimora:
I suggest creating a dedicated `SafeTensorFileRecorder` to handle SafeTensor files independently from PyTorch's `.pt` files. This approach ensures a clear separation between different file formats and supports framework-specific transformations during the import process.

Additionally, I propose providing configurable options (via `LoadArgs`, [similar to PyTorch's recorder](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/pytorch/recorder.rs#L86)) within the recorder to specify the appropriate transformation adapter. By default, this could use the [PyTorchAdapter](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/pytorch/adapter.rs#L24) but allow customization for other frameworks, such as TensorFlow. This design enhances flexibility and decouples the handling of different tensor file formats. Moreover, it might be beneficial to support passing a user-defined implementation of `BurnModuleAdapter` when needed.

Lastly, we should replicate PyTorch import tests to ensure comprehensive coverage. Over time, we can expand these tests to include SafeTensor files exported from TensorFlow.

One more thing: we should introduce a new feature flag, `safetensors`.

laggui (changes requested):
Thanks for making the changes.

That is *a lot* of code duplication 😅 I don't think "replicate" was meant it this way hahah

We should re-use the existing `PyTorchAdapter` when both `pytorch` and `safetensors` features are enabled. Definitely don't want to copy its implementation over. And by default we can use the existing `DefaultAdapter` in `burn::record::serde::adapter` to simply load the safetensor file as is.

For the tests, we also don't need to copy all the python scripts. We can just have the existing scripts under pytorch-tests save both in pickle and safetensor formats. And since the current tests added come from pytorch, we can add the additional safetensor tests to the existing pytorch tests under the `safetensors` feature flag guard. We would need the safetensor recorder to use the pytorch adapter anyway, so these tests can live under the pytorch-tests (with the addition of the `safetensors` feature flag). If we want to add standalone tests for models saved in safetensors that don't require any other transformations we could expand that to have safetensors-tests.

antimora (changes requested):
Thank you for taking this up. This feature will help many that requested this.

This is in the right direction but we should try reducing code duplications. I would prioritize Rust code de-duplication first. We can leave the example and test duplicated for now (because it will take time). It's up to @laggui to allow it.

I also suggest creating a new section under the book specifically for SafeTensors. You can talk about the transformation for model modules.

The last thing, don't forget to update `LoadArgs` for SafeTensorsFileRecorder per my previous comment.

laggui:
> **I would prioritize Rust code de-duplication first**. We can leave the example and test duplicated for now (because it will take time).

Agreed for the bold part.

ivila:
Just a feature request, can ignore this if you think it is unnessesary😂: since safetensors now supports `no_std`, could we also have a `SafeTensorBytesRecorder`?

laggui:
> Just a feature request, can ignore this if you think it is unnessesary😂: since safetensors now supports `no_std`, could we also have a `SafeTensorBytesRecorder`?

A bit hesitant on adding new record formats tbh, not sure of the value. To import safetensor format (like the initial target of this PR), sure. But maybe you could provide a bit more info to justify? 🙂

antimora:
> > Just a feature request, can ignore this if you think it is unnessesary😂: since safetensors now supports `no_std`, could we also have a `SafeTensorBytesRecorder`?
> 
> A bit hesitant on adding new record formats tbh, not sure of the value. To import safetensor format (like the initial target of this PR), sure. But maybe you could provide a bit more info to justify? 🙂

We don't need a new type. We can provide with an arg option.

reneleonhardt:
@laggui This has been a lot of work. What can be done to complete the review?

antimora:
> @laggui This has been a lot of work. What can be done to complete the review?

Still the PR needs to be refactored. 

Do you use have a use case for this feature? We could complete this ourselves.

reneleonhardt:
Didn't want to rush anything, just saw the 2 year old feature request and your bot adding the stale label.
I don't fully understand the described limitations, but importing/integrating HuggingFace or Kaggle sounds like a very big feature to boost burn adoption 😅

antimora:
> Didn't want to rush anything, just saw the 2 year old feature request and your bot adding the stale label. I don't fully understand the described limitations, but importing/integrating HuggingFace or Kaggle sounds like a very big feature to boost burn adoption 😅

It's all about priorities. It's been requested but there wasn't real demand in use cases =)

Not difficult to implement because we can rely on PytorchRecorder's foundation.

wandbrandon:
Hi all apologies for the wait on this one, got tied up with other projects and work. I can finalize this soon as it's relatively simple.

wandbrandon:
pushed a new commit, I fixed the adapter as per requested, I wanted to remove the duplicate code, but frankly feel a little icky depending so largely on the pytorch feature implementation. Do we want to abstract away some of the logic in the pytorch feature into more generic part of burn import?

antimora:
> pushed a new commit, I fixed the adapter as per requested, I wanted to remove the duplicate code, but frankly feel a little icky depending so largely on the pytorch feature implementation. Do we want to abstract away some of the logic in the pytorch feature into more generic part of burn import?

If you're okay, I can make these changes directly in this PR. It might be faster turn around considering how many files in this PR. It's mainly about organizing files. I think you got the functionality working.

wandbrandon:
> > pushed a new commit, I fixed the adapter as per requested, I wanted to remove the duplicate code, but frankly feel a little icky depending so largely on the pytorch feature implementation. Do we want to abstract away some of the logic in the pytorch feature into more generic part of burn import?
> 
> If you're okay, I can make these changes directly in this PR. It might be faster turn around considering how many files in this PR. It's mainly about organizing files. I think you got the functionality working.

Yes! 100% okay with this. I know safetensors is a highly sought out feature.

antimora (approved):
I made the fixes and improvements directly. Therefor, I am approving it.

antimora:
All changes are in place. Waiting for @laggui. 

If you have a chance, @wandbrandon, please review as well.

wandbrandon:
> All changes are in place. Waiting for @laggui.
> 
> If you have a chance, @wandbrandon, please review as well.

Looks great, appreciate the fast turnaround!

README.md
```diff
@@ -426,13 +426,16 @@ Our ONNX support is further described in
 
 <details>
 <summary>
-Importing PyTorch Models 🚚
+Importing PyTorch or SafeTensors Models 🚚
```
laggui, line 429: I believe `SafeTensors` is incorrectly capitalized (across all this PR). It should be `Safetensors`

burn-book/src/import/safetensors-model.md
laggui, line 1: This is *very* similar to our pytorch import example. Even the content is almost the same, just re-hashed. So I'm not sure if there is actual value in the current state to have two sections, where the biggest difference is the file recorder used.

The content just seems to have been rephrased and formatting is not aligned with the existing section.

crates/burn-import/safetensors-tests/Cargo.toml
laggui, line 1: We can test the safetensors import with the current pytorch tests, or keep it even simpler and narrow it down to fewer use cases.

Anyway all the `safetensors-tests` seem to be copied over from the pytorch tests.

Right now there is just an enormous amount of duplication.

crates/burn-import/src/safetensors/recorder.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Adapters handle potential differences in tensor formats or naming conventions
+/// between the source framework (like PyTorch or TensorFlow) and Burn.
+#[derive(Debug, Clone, Default, Copy)]
+pub enum AdapterType {
+    /// Adapts tensors assuming they originated from PyTorch.
+    #[default]
+    PyTorch,
+
+    /// Adapts tensors assuming they originated from TensorFlow.
+    TensorFlow,
```
laggui, line 116: Don't need a tensorflow variant if we don't support it. The enum can be expanded later.

examples/import-model-weights/README.md
```diff
@@ -0,0 +1,123 @@
...
+- `src/bin/`: Contains all example binaries
+  - `pytorch.rs`: Example for importing PyTorch weights
+  - `safetensors.rs`: Example for importing SafeTensors weights
+  - `convert.rs`: Example for converting between weight formats
+  - `namedmpk.rs`: Example for using named MessagePack files
+- `src/model.rs`: Defines the model architecture
+- `src/inference.rs`: Contains the inference logic
+- `weights/`: Contains pre-trained model weights
+  - `mnist.pt`: PyTorch model weights
+  - `mnist.safetensors`: SafeTensors model weights
```
laggui, line 115: Not sure this is very relevant, especially given the detailed sections above.

examples/import-model-weights/src/bin/convert.rs
laggui, line 1: Can you explain the motivation behind having a standalone "convert" binary?

This is just importing the weights and saving them back, which can be done in the other scripts instead if we wish. But it's already [detailed in the book anyway](https://burn.dev/burn-book/import/pytorch-model.html#how-to-import-a-pytorch-model).

examples/import-model-weights/src/bin/namedmpk.rs
```diff
@@ -0,0 +1,60 @@
...
+//! - `<image_index>`: The index of the MNIST test image to classify (0-9999)
+//! - `<model_path>`: Path to the model file (without .mpk extension)
+//!
+//! ## Example
+//!
+//! ```shell
+//! cargo run --bin namedmpk -- 35 /tmp/burn-convert/mnist
+//! ```
+//!
+//! This example is typically used after converting a model with the `convert` binary.
```
laggui, line 26: Given the example usage in the README, this feels a bit too verbose 🤔 also, the code is fairly short and commented.. so not sure we need even more comments?

Maybe it's just me. But given that the example is relatively simple, I'd rather jump right into the code (which is clear).

examples/import-model-weights/src/bin/pytorch.rs
```diff
@@ -0,0 +1,49 @@
...
+//!   If omitted, a default index (42) will be used
+//!
+//! ## Example
+//!
+//! ```shell
+//! cargo run --bin pytorch -- 15
+//! ```
+//!
+//! This example uses the `PyTorchFileRecorder` from `burn_import` to seamlessly
+//! load PyTorch weights into a Burn model without requiring an intermediate conversion step.
```
laggui, line 26: Same comment

examples/import-model-weights/src/bin/safetensors.rs
```diff
@@ -0,0 +1,53 @@
...
+//!
+//! ## Example
+//!
+//! ```shell
+//! cargo run --bin safetensors -- 42
+//! ```
+//!
+//! This example uses the `SafeTensorsFileRecorder` from `burn_import` to seamlessly
+//! load SafeTensors weights into a Burn model without requiring an intermediate conversion step.
+//! SafeTensors is a safer and more efficient alternative to PyTorch's native format.
```
laggui, line 28: Same comment

examples/import-model-weights/src/inference.rs
```diff
@@ -1,22 +1,37 @@
...
+//!
+//! ## Features
+//!
+//! - Backend-agnostic implementation (`<B: Backend>`)
+//! - Command-line argument support for choosing test images
+//! - Proper input normalization for MNIST dataset
+//! - Helpful output with links to visualize the tested image
+//!
+//! This module is used by all examples in the `bin/` directory to standardize
+//! the inference process regardless of where the model weights originated.
```
laggui, line 21: Same comment

examples/import-model-weights/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+//! - `model`: MNIST CNN model architecture compatible with imported weights
+//!
+//! ## Example Binaries
+//!
+//! - `pytorch`: Demo for importing PyTorch weights
+//! - `safetensors`: Demo for importing SafeTensors weights
+//! - `convert`: Tool for converting between weight formats
+//! - `namedmpk`: Demo for using Burn's native format
+//!
+//! For detailed usage instructions, see the README.md or the documentation in each binary.
```
laggui, line 24: Same comment

examples/import-model-weights/src/model.rs
```diff
@@ -1,13 +1,28 @@
...
+//! - Log-softmax output for 10 classes (digits 0-9)
+//!
+//! ## Key Components
+//!
+//! - `Model<B>`: The main model structure with backend-agnostic implementation
+//! - `ModelRecord<B>`: Type alias for serialized model records
+//! - Methods for initializing models and loading weights from records
+//!
+//! The model architecture is compatible with weights imported from PyTorch
+//! and SafeTensors formats, which can be loaded using the appropriate recorder.
```
laggui, line 18: Same 😅

examples/import-model-weights/src/bin/namedmpk.rs
laggui: Also, just realized this is a different example that loads the saved mpk record. I don't think it is necessary.

laggui (changes requested):
The added support is nice 🙂 

But I have some comments, mostly re-iterating the earlier points in this discussion. We should avoid code duplication, and in some cases even bloated documentation.

examples/import-model-weights/src/bin/convert.rs
antimora: To demonstrate how to convert to our record types. Yeah it's redundant in terms of information but it completes the example. We used to have `build.rs` that would convert from pt but I removed it.

examples/import-model-weights/src/bin/namedmpk.rs
antimora: For completeness we should keep. It's good demo to convert and load back. This was implicit previously in import-pytorch example.

crates/burn-import/safetensors-tests/Cargo.toml
antimora: I tried coming up with one big test for safetensors but I think it's better to extract adapter specific conversion tests from PyTorch as well. So we will have 3 tests: Adapter conversion, PyTorch loading and Safetensors loading. 

I filed a separate ticket to do this refactor: #3138

burn-book/src/import/safetensors-model.md
antimora: We discussed offline to keep two sections separate but have the same format and language between the two. We will figure out how to remove redundant. For now it seems it's easier to have two.

crates/burn-import/safetensors-tests/Cargo.toml
laggui: The changes mentioned in the linked issue are not part of the current codebase, only part of this PR. So they haven't been introduced yet 😅 

This PR has over 3k additions and 74 files changed, but only a tiny fraction of that is actually related to the new feature. Most of that comes from the test duplications, which are not necessary. I am not in favor of introducing such a big amount of duplication just to eventually fix it (or even worse, remain unchanged for longer)...

Tbh I'd almost rather have a single model test for safetensors, even if it doesn't cover *all* possible modules. The adapter stuff is common with pytorch, so for now we can just have a small portion test correct loading of a safetensors file.

antimora: Remove the other files and replace with simple test. Closed #3138 as won't do.

burn-book/src/import/README.md
```diff
@@ -1,10 +1,13 @@
 # Importing Models
 
-The Burn project supports the import of models from various frameworks, emphasizing efficiency and
-compatibility. Currently, it handles two primary model formats:
+The Burn project provides robust support for importing models from various machine learning frameworks, allowing you to leverage pre-trained models and weights in your Burn applications. These import capabilities are designed with efficiency, security, and compatibility in mind.
```
laggui, line 3: Didn't really notice before, but "The Burn project" sounds a bit weird lol.

And I think I would again keep things simple, e.g.

```suggestion
Burn supports importing models from other frameworks and file formats, enabling you to use pre-trained weights in your Burn applications.
```

burn-book/src/import/README.md
```diff
@@ -1,10 +1,13 @@
...
 
-2. [PyTorch](./pytorch-model.md): Enables the loading of PyTorch model weights into Burn’s native model
-   architecture, ensuring seamless integration.
+Burn currently supports three primary model import formats:
+
+| Format | Description | Use Case |
+|--------|-------------|----------|
+| [**ONNX**](./onnx-model.md) | Open Neural Network Exchange format | Direct import of complete model architectures and weights from any framework that supports ONNX export |
+| [**PyTorch**](./pytorch-model.md) | PyTorch weights (.pt, .pth) | Loading weights from PyTorch models into a matching Burn architecture |
+| [**Safetensors**](./safetensors-model.md) | Hugging Face's secure tensor format | Secure, efficient, and language-agnostic weight loading without pickle dependencies |
```
laggui, line 13: I don't think "secure tensor format" is entirely accurate. It is safer than pickle. I'd argue it's probably generally faster though. Also, the use case for is pretty similar. We just swap the pickle dep for safetensors to load the weights into a matching Burn architecture.

```suggestion
| [**Safetensors**](./safetensors-model.md) | Hugging Face's model serialization format | Loading a model's tensor weights into a matching Burn architecture |
```

laggui (commented):
Thanks for removing the bloated tests 😅 

Much better now. I have a few comments mostly regarding docs

burn-book/src/import/pytorch-model.md
```diff
@@ -335,20 +316,20 @@ weights resulting in:
...
+### Partial Model Loading
+
+You can selectively load weights into a partial model, which is useful for:
 
-`PyTorchFileRecorder` enables selective weight loading into partial models. For instance, in a model
-with both an encoder and a decoder, it's possible to load only the encoder weights. This is done by
-defining the encoder in Burn, allowing the loading of its weights while excluding the decoder's.
+- Loading only the encoder from an encoder-decoder architecture
+- Fine-tuning specific layers while initializing others randomly
+- Creating hybrid models combining parts from different sources
```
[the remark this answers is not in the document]
antimora: I am doing this for my model but I can't share it =D.

Lets keep, it highlights capability. People can easily figure without explicit example.

burn-book/src/import/safetensors-model.md
```diff
@@ -0,0 +1,287 @@
...
+"model.layers.2.bias"
+```
+
+### Partial Model Loading
+
+You can selectively load weights into a partial model, which is useful for:
+
+- Loading only the encoder from an encoder-decoder architecture
+- Fine-tuning specific layers while initializing others randomly
+- Creating hybrid models combining parts from different sources
```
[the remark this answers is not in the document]
antimora: Let's keep it. It's informative.

laggui (commented):
Missed a small thing in the previous review! Otherwise, should be good to go after.

Thanks for taking this over btw!

laggui (approved):

laggui:
> Thanks for reviewing the long PR. Takes up lots of context switching.

Of course! Longer PRs take more time to review carefully so time-to-merge is usually longer as well 😅 

There has been an uptick in activity since the 0.17 release too so I have to balance it out. Sorry if your other PRs are not entirely reviewed yet, should come soon!

PR #2735, opened 2025-01-22
Add new FromData operation description

### LINKED ISSUE #2719: Memory leak
**Describe the bug**
The CudaJit Backend leaks memory when creating and dropping Tensors.

**To Reproduce**

    fn test_memory_leak() {
        unsafe { libloading::Library::new("libtorch.so").unwrap() };

        // number of loops to overflow GPU memory are dependend on your hardware
        let loops = 10_000;

        // this works and memory does not grow
        test_device_memory_leak::<burn::backend::Candle>(
            &burn::backend::candle::CandleDevice::cuda(0),
            loops,
        );

        // this works and memory does not grow
        test_device_memory_leak::<burn::backend::LibTorch>(
            &burn::backend::libtorch::LibTorchDevice::Cuda(0),
            loops,
        );

        // this overflows my 10 gig memory at aboout loop 2600
        test_device_memory_leak::<burn::backend::CudaJit>(
            &burn::backend::cuda_jit::CudaDevice::new(0),
            loops,
        );
    }

    fn test_device_memory_leak<B: Backend>(device: &<B as Backend>::Device, loops: usize) {
        const DIM: usize = 4;
        let shape = [32, 4, 84, 84];
        let l = shape.iter().product();
        let data_v = vec![0; l];

        for i in 0..loops {
            if i % 100 == 0 {
                println!("i: {}", i);
            }
            let td = TensorData::new(data_v.clone(), shape);
            let t: Tensor<B, DIM> = Tensor::from_data(td, device);
            drop(t);
        }
    }

**Expected behavior**
drop(t) should free the memory

**Screenshots**

Fedora 41, Nvidia cuda 12.6

laggui:
### Checklist

### Related Issues/PRs

Fixes #2719

### Changes

- Added a new operation description for `FromData`, similar to other tensor creation ops (zeros, random, etc.)
- Refactored fusion to use the `FromData` description (fixes linked issue, since an operation is registered the handles are correctly freed)
- Fixed empty tensor op in fusion in the process (same issue as from data)

### DIFF
--- a/crates/burn-tensor/src/repr/operation.rs
+++ b/crates/burn-tensor/src/repr/operation.rs
@@ -6,6 +6,7 @@ use alloc::borrow::ToOwned;
 use alloc::boxed::Box;
 use alloc::{string::String, vec, vec::Vec};
 
+use crate::TensorData;
 use crate::{
     ops::{
         ConvOptions, ConvTransposeOptions, DeformConvOptions, InterpolateMode, InterpolateOptions,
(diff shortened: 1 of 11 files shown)


crates/burn-tensor/src/repr/operation.rs
```diff
@@ -1677,6 +1692,14 @@ impl ModuleOperationDescription {
     }
 }
 
+impl core::hash::Hash for FromDataOperationDescription {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.out.hash(state);
+
+        self.data.bytes.hash(state);
+    }
+}
```
nathanielsimard, line 1701: I think it's a very very bad idea to do that, we should just skip the hash for the bytes. Or panic here even.

laggui: Good catch! Didn't realize it used the default implementation for slice hash since `Bytes` implements deref to return a slice 😬

nathanielsimard (approved):

PR #2744, opened 2025-01-27
burn-book typo - correct `smp_serde` to `rmp_serde` as per crate's name in url

cameronbraid:
Typo - self explanatory change


laggui (approved):
Whoops 😅 good catch!

PR #2753, opened 2025-01-28
[Feat] Add new burn-vision crate with one initial op

wingertge:
### Checklist

### Changes

Adds a new `burn-vision` crate and an initial implementation of `connected_components` and `connected_components_with_stats` (as in [opencv](https://docs.opencv.org/3.4/d3/dc0/group__imgproc__shape.html#ga107a78bf7cd25dec05fb4dfc5c9e765f)).

The algorithms are WIP and rely on https://github.com/tracel-ai/cubecl/pull/446, but I'd like some early feedback on the structure of the new crate and the backend implementations.

### Testing

Adds new connected components tests.


laggui (commented):
Since this is a backend extension, what do you think about keeping the implementation within the vision crate instead?

Similar to how you've added a default `cpu_impl`, the vision crate would look something like this:
```
burn-vison/src/
└── ops/
    ├── cpu_impl/
    ├── jit/
    ├── mod.rs
    └── base.rs
```

It would only move your current additions in `burn-jit/src/kernel/vision` to `burn-vision/src/ops/jit` (and change the imports to use `burn_jit`). The `jit` module can be under a feature flag.

That way, the backend implementations are kept to the required stuff only and the vision trait implementations are isolated within the crate.

laggui:
> That's how I originally had it, I wasn't sure which way was correct. I guess that's one vote for keeping it in the crate. I think if backends had more support for these ops it would make sense to implement it in each backend, but there's really no vision op support for any of the other backends. They all just rely on opencv. So I guess as long as we only have two implementations it makes more sense to keep it in `burn-vision`.

Yeah I was debating that as well when going over the code. But I think this is too specific to live in the backend implementations like you said, it will probably stay restricted to our own backends and not third-party backends. So it makes more sense in this state to have it as an extension in the vision crate. 

> Anyways I'm still working on CPU support, all the fast algorithms use decision forests and they're full of `goto` and `#define`. I'm trying to modify the `GRAPHGEN` library to generate the popular algorithms' decision trees in a Rust-compatible way, but it's rough 😅 Linking to opencv would be another option but has its own problems (C++ dependency, converting between burn tensors and opencv `Mat`, etc). Not sure what the correct solution there is.

Yeah I'm sure opencv would be a pain to use here 😬 but at the same time it offers so much for image processing lol. Must be a pain to link though.

A default CPU implementation is not a hard requirement either for the vision crate (that is, unless you also need it for your own application lol). This is an extension after all..

wingertge:
> A default CPU implementation is not a hard requirement either for the vision crate (that is, unless you also need it for your own application lol). This is an extension after all..

My vision (hah) for this crate is to provide easy access to image pre and postprocessing for burn, since the current solutions are very awkward (imageproc is pretty incomplete and doesn't support GPU acceleration, opencv is a pain to build and transferring data across FFI is hard). Part of that is always having a fallback implementation available, so you know it always at least works, and uses acceleration where possible.

Ideally, if someone is missing an operation, we could relatively easily port the default opencv implementation from C++ to Rust and set it as the default, then if someone wants to add GPU support it can be done later. Once I get GRAPHGEN up and running it should be fairly easy to do even with decision forest algorithms. That's why I'm attempting to modify the codegen instead of manually translating the generated code.

crates/burn-vision/Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[package]
+authors = ["nathanielsimard <nathaniel.simard.42@gmail.com>"]
```
laggui, line 2: You could at least add your GH handle to the authors list 😄 no obligations ofc

crates/burn-vision/src/backends/cpu/connected_components.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+pub fn connected_components_with_stats<B: Backend>(
+    img: BoolTensor<B>,
+    connectivity: Connectivity,
+    _options: ConnectedStatsOptions,
+) -> (IntTensor<B>, ConnectedStatsPrimitive<B>) {
+    let device = B::bool_device(&img);
+    let (labels, stats) = run::<B, ConnectedStatsOp>(img, connectivity, ConnectedStatsOp::default);
+    println!("{stats:?}");
```
laggui, line 28: Left over print
```suggestion
```

crates/burn-vision/src/backends/cpu/connected_components.rs
```diff
@@ -0,0 +1,229 @@
...
+    let img = Tensor::<B, 3, Bool>::from_primitive(img);
+    let [batches, height, width] = img.shape().dims();
+    let img = img.into_data().convert::<u8>().to_vec::<u8>().unwrap();
+    let img = Array3::from_shape_vec((batches, height, width), img).unwrap();
+    let mut stats_res = Vec::with_capacity(batches);
+
+    let process = match connectivity {
+        Connectivity::Four => spaghetti_4c::process::<UnionFind>,
+        Connectivity::Eight => spaghetti::process::<UnionFind>,
+    };
```
laggui, line 48: I'm not familiar with connected components labeling, but nice to see they have spaghetti hahahah

laggui (commented):
I'm not familiar with these CV algorithms so I can't really comment on that, but otherwise the vision crate as an extension looks good!

The codegen is not pretty but glad to see that you managed to get it working 😅 

Some small stuff below.

crates/burn-vision/Cargo.toml
wingertge: I always feel very self-conscious putting my name in other people's repos, but I guess I can do it for a new crate 😅

crates/burn-vision/src/backends/cpu/connected_components.rs
wingertge: it's quite literally called "Spaghetti" because the original goto-based tree is absolute spaghetti code 😅

crates/burn-vision/Cargo.toml
laggui: Ahh I can understand, there's no obligation 🙂 But you've done so much in Burn and CubeCL, figured you should at the very least add your name to the new vision crate you're working on!

laggui (approved):

laggui:
@nathanielsimard new `burn-vision` crate! Do we want to [add it to the published crates](https://github.com/tracel-ai/burn/blob/main/.github/workflows/publish.yml) already?

Lmk if you have any additional comments, otherwise will be good to merge!

crates/burn-vision/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+extern crate alloc;
+
+/// Backend implementations for JIT and CPU
+pub mod backends;
+mod ops;
+mod tensor;
+
+#[cfg(feature = "export-tests")]
+#[allow(missing_docs)]
```
nathanielsimard, line 21: I don't think we need to export the tests, since they run in this crate

nathanielsimard (commented):
Another point to keep all the kernels in `burn-vision`. Compiling kernels is quite expensive, and if they are all within the same crate, they can't be parallelized by Cargo. I plan to move the `fusion` portion of `burn-jit` into its own crate for that reason as well, so less stuff in `burn-jit` is better.

crates/burn-vision/src/backends/cpu/ops.rs
```diff
@@ -0,0 +1,52 @@
...
+        let stats = crate::ConnectedStatsPrimitive::<Self> {
+            area: stats.area,
+            left: stats.left,
+            top: stats.top,
+            right: stats.right,
+            bottom: stats.bottom,
+            max_label: stats.max_label,
+        };
+        (labels, stats)
+    }
```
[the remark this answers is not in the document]
wingertge: The reason I did that is because `Tensor` is generic over the backend, and it would be much more ergonomic to be able to use autodiff tensors for pre/post-processing. Not sure how big the actual use case is, but you might need some vision ops for at least evaluation so having it implemented makes things a lot nicer.

crates/burn-vision/src/backends/jit/connected_components/hardware_accelerated.rs
```diff
@@ -0,0 +1,610 @@
...
+
+            let mut d = start_distance(pixels_y_1, 32);
+            distance_y_1 = d + select(d == 32, distance_y_1, 0);
+            d = start_distance(pixels_y, 32);
+            distance_y = d + select(d == 32, distance_y, 0);
+        }
+    }
+}
+
+#[cube(launch)]
```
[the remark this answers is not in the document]
wingertge: Probably, I'm just a bit paranoid after that big OOB write issue we had a while ago. As long as we keep it in mind in case any future issues pop up, it should be fine to run unchecked.

crates/burn-vision/src/lib.rs
wingertge: Integration tests count as running in a separate crate, and we need to use integration tests because of annoying rust limitations around `macro_export` (which is generated by `burn-tensor-testgen`).

crates/burn-vision/src/backends/cpu/ops.rs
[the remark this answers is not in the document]
nathanielsimard: The batcher and evaluation is done on the inner backend when training, at least within `burn-train`.

crates/burn-vision/src/lib.rs
nathanielsimard: I don't think we're using `burn-tensor-testgen` in `burn-core`, and yeah we're probably not running the tests as integration tests for that reason.

crates/burn-vision/src/backends/cpu/ops.rs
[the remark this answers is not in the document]
wingertge: I'll remove it then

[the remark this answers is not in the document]
wingertge: I'll remove it then

crates/burn-vision/src/lib.rs
wingertge: We might want to parameterize some of the tests at some point, so I think using testgen right away is better. I just didn't add a parameterized test yet because connected components are almost never used with anything other than u32. I believe opencv also supports u16 but I don't think it's work adding parameterized tests until there's more ops. It's good to have the infrastructure for it regardless.

PR #2756, opened 2025-01-30
Fix BackendValues in backend-comparison after removal of jit suffix

syl20bnr:
This PR also sorts the variants alphabetically like in the `Cargo.toml` file.


laggui (commented):
Good catch, makes more sense to match the new backend flags.


The audit CI failure is due to `polars 0.44.2`, looks like they just released `0.46.0` a couple days ago that fixed the dependency causing this issue.

laggui (approved):

PR #2762, opened 2025-02-01
Add chapter to book for nightly use

bioinformatist:
One may meet this problem when use Burn and another crate with `hashbrown=^0.14` or earlier simultaneously.

8d8b6f70392272c36efdde952b4af3f

Similar to https://github.com/rust-lang/hashbrown/issues/564.

In our case, the older version of `hashbrown` is used by `dashmap`.

I've tried to add new feature flags to Burn, but it soon became a tedious business.

Another way is to add alias for older `hashbrown` like what the `polars` has done: https://github.com/pola-rs/polars/pull/19091/files, but I think it's not elegant enough as it can also be complicated.

So maybe the better way is to tell the users the solution in the book.

For more info: https://github.com/rust-lang/hashbrown/issues/564 and https://github.com/zakarumych/allocator-api2/issues/19.

### DIFF
--- a/burn-book/src/SUMMARY.md
+++ b/burn-book/src/SUMMARY.md
@@ -33,3 +33,4 @@
   - [Custom Optimizer]()
   - [WebAssembly]()
   - [No-Std](./advanced/no-std.md)
+  - [Nightly](./advanced/nightly.md)
--- a/burn-book/src/advanced/nightly.md
+++ b/burn-book/src/advanced/nightly.md
@@ -1,0 +1,17 @@
+# Nightly
+
+Burn has nightly support by default. However, one may meet this error during compilation:
+
+```
+error[E0658]: use of unstable library feature `allocator_api`
+```
+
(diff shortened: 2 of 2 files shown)


laggui (commented):
Thanks for providing the steps and links!

I agree that having something like this documented is helpful for future reference, but I'm not sure we should add a nightly section to the book just for this.

Perhaps it could be left as an outstanding issue so users can easily refer to it?

bioinformatist:
> I'm not sure we should add a nightly section to the book just for this
I also feel something strange here 😅 since this issue may be solved by stable Rust in the future. Or let's move it to a pinned issue?

laggui:
> I also feel something strange here 😅 since this issue may be solved by stable Rust in the future. Or let's move it to a pinned issue?

A pinned issue for nightly stuff already makes more sense 😄

laggui:
Closing as discussed. Feel free to open an issue with the description and solution.

Issue #2751, opened 2025-01-27
PoissonNLLLoss for regression


salvomcl:
### Feature description

[`PoissonNLLLoss`](https://pytorch.org/docs/stable/generated/torch.nn.PoissonNLLLoss.html)

### Feature motivation

Hi all, I’ve noticed that this loss function is not available, despite being popular across several industries. 
If you think it’s worth adding, I’d be happy to contribute by implementing it.

Thank you!

laggui:
I'm not super familiar with this loss function, but contributions are always welcomed!

salvomcl:
Great, I'll work on this!

PR #2765, opened 2025-02-02
Feat: Add PoissonNLL loss

salvomcl:
### Checklist

### Related Issues/PRs

This PR addresses issue #2751.

### Changes

This PR introduces the `PoissonNLLLoss` module, implementing the Poisson negative log-likelihood loss function. 
In particular I followed the`pytorch` implementation to make the transition from PyTorch as smooth as possible for users.

Key additions:  
- `PoissonNLLLossConfig`: Configuration structure with parameters for `log_input`, `full`, and `eps`.  
- `PoissonNLLLoss`: Implements forward and element-wise loss computation with optional reduction (`Mean`, `Sum`).  
- Validation checks to ensure appropriate input dimensions and non-negative target values.  

### Testing

- Added unit tests to verify:  
  - Correctness of the loss computation with different configurations.  
  - Behavior when `log_input` is `true` and `false`.  
  - Handling of the Stirling approximation term when `full` is enabled.  
  - Gradient computation correctness via backpropagation tests.  
  - Assertion failures for invalid inputs (negative targets, shape mismatches, and non-positive `eps`).

### DIFF
--- a/crates/burn-core/src/nn/loss/poisson.rs
+++ b/crates/burn-core/src/nn/loss/poisson.rs
@@ -209,3 +209,181 @@ impl PoissonNLLLoss {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::tensor::TensorData;
+    use crate::TestBackend;
+    type TestTensor<const D: usize> = Tensor<TestBackend, D>;
+
+    #[test]
+    fn test_poisson_nll_loss() {
+        let predictions = TensorData::from([0., 0., -40., 1., 2., 3.]);
+        let targets = TensorData::from([1., 4.5, 2.5, 0., 0., 2.]);
+
+        let device = Default::default();
+
+        let predictions = TestTensor::<1>::from_data(predictions, &device);
+        let targets = TestTensor::<1>::from_data(targets, &device);
+
+        let poisson = PoissonNLLLossConfig::new().init();
+
+        let loss_sum = poisson.forward(predictions.clone(), targets.clone(), Reduction::Sum);
+        let loss = poisson.forward(predictions.clone(), targets.clone(), Reduction::Auto);
+        let loss_no_reduction = poisson.forward_no_reduction(predictions, targets);
+
+        let expected = TensorData::from([1.0000, 1.0000, 100.0000, 2.7183, 7.3891, 14.0855]);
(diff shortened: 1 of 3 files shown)


crates/burn-core/src/nn/loss/poisson.rs
```diff
@@ -0,0 +1,389 @@
+use std::f32::consts::PI;
```
laggui, line 1: For no-std we should use this instead

```suggestion
use core::f32::consts::PI;
```

crates/burn-core/src/nn/loss/poisson.rs
```diff
@@ -0,0 +1,389 @@
...
+/// L(predictions, target) = predictions - target * log(predictions) + log(target!)
+/// ```
+/// The last term (`log(target!)`) can be omitted or approximated using Stirling's formula.
+/// The approximation is applied for `target > 1`, while for `target <= 1`, zeros are added to the loss.
+///
+/// For more details, see:
+/// <https://en.wikipedia.org/wiki/Poisson_regression#Maximum_likelihood-based_parameter_estimation>
+#[derive(Module, Debug, Clone)]
+#[module(custom_display)]
+pub struct PoissonNLLLoss {
```
laggui, line 87: To follow the naming convention with other structs, this should be renamed to `PoissonNllLoss` (and the config to `PoissonNllLossConfig`, etc.)

crates/burn-core/src/nn/loss/poisson.rs
```diff
@@ -0,0 +1,389 @@
...
+        loss_no_reduction.into_data().assert_approx_eq(&expected, 5);
+
+        let expected = TensorData::from([21.9920]);
+        loss.into_data().assert_approx_eq(&expected, 5);
+
+        let expected = TensorData::from([131.9518]);
+        loss_sum.into_data().assert_approx_eq(&expected, 5);
+    }
+
+    #[test]
```
laggui, line 289: The `TestAutodiffBackend` type is under the `std` feature flag, so we probably need to guard this test.

```suggestion
    #[cfg(feature = "std")]
    #[test]
```

laggui (changes requested):
LGTM! Thanks for the addition 🙂 

Just some minor changes left

salvomcl:
Thanks for the review @laggui! Just pushed the requested changes! :)

laggui (approved):

PR #2778, opened 2025-02-05
Fix/from data copy

nathanielsimard:


crates/burn-tensor/src/repr/operation.rs
```diff
@@ -81,6 +80,8 @@ pub enum OperationDescription {
     Float(DType, FloatOperationDescription),
     /// Module operation.
     Module(ModuleOperationDescription),
+    /// Initialize operation.
+    Init(InitOperationDescription),
```
laggui, line 84: Pretty similar to `BaseOperationDescription::Empty` (which is also a tensor creation op). Maybe we could merge them.

nathanielsimard: Empty can be fused, not Init, so I would keep them separate for now.

PR #2782, opened 2025-02-06
Fix constant creation in fusion to cast at compile time, not runtime

wingertge:
### Checklist

### Changes

Changes the fusion `read` kernel to cast literals at compile time, not runtime.

### Testing

All tests pass, and issues encountered in https://github.com/ArthurBrussee/brush/tree/spirv have been fixed.

### DIFF
--- a/crates/burn-wgpu/src/lib.rs
+++ b/crates/burn-wgpu/src/lib.rs
@@ -15,8 +15,8 @@ pub use cubecl::flex32;
 pub use cubecl::CubeDim;
 
 pub use cubecl::wgpu::{
-    init_device, init_setup, init_setup_async, MemoryConfiguration, RuntimeOptions, WgpuDevice,
-    WgpuResource, WgpuRuntime, WgpuSetup, WgpuStorage,
+    init_device, init_setup, init_setup_async, MemoryConfiguration, RuntimeOptions, WgpuCompiler,
+    WgpuDevice, WgpuResource, WgpuRuntime, WgpuSetup, WgpuStorage,
 };
 // Vulkan and WebGpu would have conflicting type names
 pub mod graphics {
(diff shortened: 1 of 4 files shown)


crates/burn-wgpu/src/lib.rs
```diff
@@ -14,10 +14,7 @@ pub use burn_jit::{BoolElement, FloatElement, IntElement};
 pub use cubecl::flex32;
 pub use cubecl::CubeDim;
 
-pub use cubecl::wgpu::{
-    init_device, init_setup, init_setup_async, MemoryConfiguration, RuntimeOptions, WgpuDevice,
-    WgpuResource, WgpuRuntime, WgpuSetup, WgpuStorage,
-};
+pub use cubecl::wgpu::*;
```
laggui, line 17: The types and functions were manually exported to avoid conflicts with the new `Vulkan` and `WebGpu` backend type aliases. Not great though since we have to manually export the useful types and might miss some 😅 

Perhaps a better solution is to keep the graphics API stuff under a public mod to access with `cubecl::wgpu::graphics` instead of re-exporting. But this would have to be changed in cubecl-wgpu.

laggui (commented):
Thanks for the quick fix!

Just a comment regarding the cubecl wgpu re-exports.

laggui (approved):

crates/burn-wgpu/src/lib.rs
wingertge: it was mostly because the cubecl dependency didn't have the reexport merged yet and I wanted to have it work on either version, but I just updated to the newer cubecl so it's redundant.

laggui: Yeah I saw in your cubecl PR! We could still fix this so we don't need to isolate the graphics API types re-export. But for now this is fine.

PR #2783, opened 2025-02-07
Remove from_data conversions in backends

laggui:
### Checklist

### Related Issues/PRs

Follow-up to #2778 

### Changes

With the last PR, the data conversion happens at the tensor API level. Backends should not have behave differently to handle conversions themselves.

### DIFF
--- a/crates/burn-jit/src/ops/base.rs
+++ b/crates/burn-jit/src/ops/base.rs
@@ -8,7 +8,7 @@ pub(crate) fn from_data<R: JitRuntime, E: JitElement>(
 ) -> JitTensor<R> {
     let shape: Shape = (&data.shape).into();
     let client = R::client(device);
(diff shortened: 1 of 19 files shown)


crates/burn-jit/src/ops/base.rs
```diff
@@ -8,7 +8,7 @@ pub(crate) fn from_data<R: JitRuntime, E: JitElement>(
 ) -> JitTensor<R> {
     let shape: Shape = (&data.shape).into();
     let client = R::client(device);
-    let buffer = client.create(data.convert::<E>().as_bytes());
+    let buffer = client.create(data.as_bytes());
 
     JitTensor::new_contiguous(client, device.clone(), shape, buffer, E::dtype())
```
nathanielsimard, line 13: The `dtype` should be the one from the data.

nathanielsimard (approved):

Issue #2624, opened 2024-12-17
Constant tensors in ONXX models don't get values populated


jameshiew:
**Describe the bug**
Loading an ONNX model, if there is a constant tensor it doesn't seem to get populated with values. Not 100% this is a bug or I'm doing something wrong. I've been trying with models generated for testing.

**To Reproduce**
I have a ONNX test in a branch on my fork - [constant_tensor_f32](https://github.com/jameshiew/burn/pull/3/files#diff-5843104da1330be193b7f5c5efd286fef9e5f1e1884cdd5e3d74830b34030b5aR2235-R2244). It's using an ONNX model generated from this [PyTorch script](https://github.com/jameshiew/burn/pull/3/files#diff-aeb5940d5e0d1c8e5d917d87e82f5bfa1e83870d5690352eec4d86ae3fdc9d24) - [constant_tensor_f32.onnx](https://github.com/jameshiew/burn/blob/1a27bfcb7fd80adf016a48416676e1c10b8a80c9/crates/burn-import/onnx-tests/tests/constant/constant_tensor_f32.onnx), and trying to add a constant tensor [[2,2],[2,2]] to input [[0,0],[0,0]].

Running the ONNX tests (`cargo nextest run --manifest-path crates/burn-import/onnx-tests/Cargo.toml`) gives this error:

```
--- STDERR:              onnx-tests::test_onnx tests::constant_tensor_f32 ---
thread 'tests::constant_tensor_f32' panicked at crates/burn-import/onnx-tests/tests/test_onnx.rs:2242:9:
assertion `left == right` failed
  left: TensorData { bytes: [0, 0, 0, 64, 0, 0, 0, 64, 0, 0, 0, 64, 0, 0, 0, 64], shape: [2, 2], dtype: F32 }
 right: TensorData { bytes: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], shape: [2, 2], dtype: F32 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Expected behavior**
The test should pass with expected output [[2,2],[2,2]]

**Screenshots**
In Netron the model looks like it contains the [[2,2],[2,2]] constant tensor.

<img width="1136" alt="Screenshot 2024-12-17 at 17 43 11" src="https://github.com/user-attachments/assets/a9ac5761-d532-44ed-97dc-0f0cc42c3b83" />

**Desktop (please complete the following information):**
[...]

PR #2784, opened 2025-02-07
One hot

akshitgaur2005:
### Checklist

### Related Issues/PRs

#1544
#1714

### Changes

Implement OneHot node for burn_import

### Testing
~~Not yet tested.~~

~~This is a draft PR as I cannot figure out how to parse depth field of OneHot, as this is my first time working on such a project.~~

~~Any guidance will be helpful.~~

All relevant test cases pass, for some reason, [should_have_no_remainder](https://github.com/tracel-ai/burn/blob/60158231a1d9a8a257c207d79e2d8992cd869d2a/crates/burn-tensor/src/tests/ops/remainder.rs#L91) test case is failing.


laggui:
Btw in case you missed it we have a `Tensor::one_hot_fill` method that was added in #2613 and closely follows the [ONNX spec](https://onnx.ai/onnx/operators/onnx__OneHot.html) 🙂

akshitgaur2005:
Oh, I missed it. I can't find its documentation though?

laggui:
It was added after the last release, so it's not on the hosted documentation yet. But it's [here](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/api/numeric.rs#L2087) (and in the [book](https://github.com/tracel-ai/burn/blob/main/burn-book/src/building-blocks/tensor.md?plain=1#L232)).

The up-to-date book on main can be browsed locally if you launch:
```
cargo xtask books burn open
```

akshitgaur2005:
Sorry if this is stupid, but my setup is showing no method one_hot_fill for tensor. Current burn version 0.16.0

akshitgaur2005:
One thing that I am stuck on  is the node constant1 has correct value in its attrs but its output has value None. This is throwing me off entirely and I am stuck on this for better part of the day now.


``` json
ParsedOnnxGraph(
    OnnxGraph {
        nodes: [
            Node {
                node_type: Constant,
                name: "constant1",
                inputs: [],
                outputs: [
                    Argument {
                        name: "constant1_out1",
                        ty: Scalar(
                            Int64,
                        ),
                        value: None,
                        passed: false,
                    },
                ],
                attrs: {
                    "value": Tensor(
                        Tensor {
                            elem_type: Int64,
                            dim: 0,
                            data: Some(
                                Int64s([3]),
                            ),
                            shape: Some(
                                [],
                            ),
                        },
                    ),
                },
            },
            Node {
                node_type: Constant,
                name: "constant2",
                inputs: [],
                outputs: [
                    Argument {
                        name: "constant2_out1",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 1,
                                shape: Some(
                                    [
                                        2,
                                    ],
                                ),
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                ],
                attrs: {
                    "value": Tensor(
                        Tensor {
                            elem_type: Int64,
                            dim: 1,
                            data: Some(
                                Int64s([0, 1]),
                            ),
                            shape: Some(
                                [
                                    2,
                                ],
                            ),
                        },
                    ),
                },
            },
            Node {
                node_type: OneHot,
                name: "onehot1",
                inputs: [
                    Argument {
                        name: "input1",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 1,
                                shape: Some(
                                    [
                                        3,
                                    ],
                                ),
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                    Argument {
                        name: "constant1_out1",
                        ty: Scalar(
                            Int64,
                        ),
                        value: None,
                        passed: false,
                    },
                    Argument {
                        name: "constant2_out1",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 1,
                                shape: Some(
                                    [
                                        2,
                                    ],
                                ),
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                ],
                outputs: [
                    Argument {
                        name: "onehot1_out1",
                        ty: Tensor(
                            TensorType {
                                elem_type: Int64,
                                dim: 1,
                                shape: Some(
                                    [
                                        3,
                                    ],
                                ),
                            },
                        ),
                        value: None,
                        passed: false,
                    },
                ],
                attrs: {
                    "axis": Int64(
                        -1,
                    ),
                },
            },
        ],
        inputs: [
            Argument {
                name: "input1",
                ty: Tensor(
                    TensorType {
                        elem_type: Int64,
                        dim: 1,
                        shape: Some(
                            [
                                3,
                            ],
                        ),
                    },
                ),
                value: None,
                passed: true,
            },
        ],
        outputs: [
            Argument {
                name: "onehot1_out1",
                ty: Tensor(
                    TensorType {
                        elem_type: Int64,
                        dim: 1,
                        shape: Some(
                            [
                                3,
                            ],
                        ),
                    },
                ),
                value: None,
                passed: false,
            },
        ],
    },
)
```

akshitgaur2005:
This PR is most probably blocked due to https://github.com/tracel-ai/burn/issues/2624

akshitgaur2005:
It seems to work correctly for reshape node atleast. But when trying to use OneHot with the command 
``` bash 
cargo r -- ./onnx-tests/tests/one_hot/one_hot.onnx ./out
```
It does no check for constant values in the node. I don't know if you have to enable it yourself as no such thing is given in the book.

akshitgaur2005:
Only test cases of burn-wgpu are failing

crates/burn-import/src/burn/node/one_hot.rs
```diff
@@ -0,0 +1,103 @@
...
+        let input = scope.tensor_use_owned(&self.input, node_position);
+        let output = &self.output.name;
+
+        let num_classes = &self.num_classes;
+        let on_value = &self.on_value;
+        let off_value = &self.off_value;
+        let axis = &self.axis;
+
+        quote! {
+            let #output = #input.one_hot_fill(#num_classes, #on_value.into(), #off_value.into(), #axis).float();
```
laggui, line 36: According to [the spec](https://onnx.ai/onnx/operators/onnx__OneHot.html#outputs), the output type should match the on/off values type.

In Burn, the output type for `one_hot_fill` is the same as the input type (which differs slightly from ONNX).

The current codegen might lead to unexpected behavior since it doesn't match the ONNX node. We should fix this.

If we decide to stick with the current implementation and always return a float tensor, we should check the tensor dtype in `one_hot_config` and panic if the value tensor is not a float.

If instead we want to correctly support the operation, we need to check the values tensor dtype and apply the output cast based on that dtype.

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -1817,3 +1817,10 @@ pub fn squeeze_config(curr: &Node) -> Vec<i64> {
 
     axes
 }
+
+pub fn one_hot_config(curr: &Node) -> (usize, f32, f32, i64) {
+    let depth = curr.inputs[1].value.clone().unwrap().into_i64();
+    let values = curr.inputs[2].value.clone().unwrap().into_i64s();
+    let axis = curr.attrs.get("axis").unwrap().clone().into_i64();
+    (depth as usize, values[1] as f32, values[0] as f32, axis)
```
laggui, line 1825: Technically the num classes (depth) and on/off values don't have to be constants, so the values couldn't be parsed straight from the graph as they are right now.

To be able to handle such cases, we would have to capture them as tensor inputs and change the codegen.

I don't think this is not mandatory for this PR, we can start with a simple implementation that only supports const values and expand in the future. But just wanted to make note.

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -1264,6 +1266,14 @@ impl ParsedOnnxGraph {
         let config = trilu_config(&node);
         TriluNode::new(input, output, config)
     }
+
+    fn one_hot_conversion(node: Node) -> OneHotNode {
+        let input = TensorType::from(node.inputs.first().unwrap());
+        let output = TensorType::from(node.outputs.first().unwrap());
+
+        let (_, on_value, off_value, axis) = one_hot_config(&node);
+        OneHotNode::new(input, output, 3, on_value, off_value, axis)
```
laggui, line 1275: The `num_classes`  should not be hardcoded to 3.

laggui (changes requested):
Great start! I have a couple of comments to make sure that the current implementation matches the ONNX spec.

Also, we can mark the op as supported once completed: https://github.com/tracel-ai/burn/blob/main/crates/burn-import/SUPPORTED-ONNX-OPS.md

crates/burn-import/src/burn/node/one_hot.rs
akshitgaur2005: What would be the best way to do that?

Sorry new to both rust and burn.

crates/burn-import/src/onnx/op_configuration.rs
laggui: Yes, but these values might not be constant so we cannot always read them directly from the graph. In this case, the tensor input represents on/off values for one-hot encoding, so in practice these are most likely constant. Same for the number of classes. But there is no guarantee 🙂

crates/burn-import/src/burn/node/one_hot.rs
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+            TensorKind::Float => {
+                quote! {
+                    let #output = #input.one_hot_fill(#num_classes, #on_value.into(), #off_value.into(), #axis).float();
+                }
+            }
+            _ => panic!("Values should be numeric"),
+        }
+    }
```
laggui, line 50: Technically, the output can also be bool! So if we want to fully cover the input and values/output dtype, I think the match arms should look like this:

```suggestion
        let input_type = &self.input.kind;
        let output_type = &self.values_type.kind; // output is tied to values type
        match (input_type, output_type) {
            (TensorKind::Int, TensorKind::Int) | (TensorKind::Float, TensorKind::Float) => {
                quote! {
                    let #output = #input.one_hot_fill(#num_classes, #on_value.into(), #off_value.into(), #axis);
                }
            }
            (TensorKind::Int, TensorKind::Float) => {
                quote! {
                    let #output = #input.one_hot_fill(#num_classes, #on_value.into(), #off_value.into(), #axis).float();
                }
            }
            (TensorKind::Float, TensorKind::Int) => {
                quote! {
                    let #output = #input.one_hot_fill(#num_classes, #on_value.into(), #off_value.into(), #axis).int();
                }
            }
            (TensorKind::Int, TensorKind::Bool) | (TensorKind::Float, TensorKind::Bool) => {
                quote! {
                    let #output = #input.one_hot_fill(#num_classes, #on_value.into(), #off_value.into(), #axis).bool();
                }
            }
            (TensorKind::Bool, _) => panic!("Input should be numeric"),
        }
```

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -1817,3 +1817,10 @@ pub fn squeeze_config(curr: &Node) -> Vec<i64> {
 
     axes
 }
+
+pub fn one_hot_config(curr: &Node) -> (usize, [f32; 2], i64) {
+    let depth = curr.inputs[1].value.clone().unwrap().into_i64();
```
laggui, line 1822: We should add a better error message if we cannot read the tensor values (i.e., not constant):
```suggestion
    let depth = curr.inputs[1]
        .value
        .clone()
        .expect("OneHot: Only constant depth is currently supported")
        .into_i64();

    let values = curr.inputs[2]
        .value
        .clone()
        .expect("OneHot: Only constant on/off values is currently supported")
        .into_f32s();
```

crates/burn-import/src/onnx/op_configuration.rs
```diff
@@ -1817,3 +1817,10 @@ pub fn squeeze_config(curr: &Node) -> Vec<i64> {
...
+    let values = curr.inputs[2].value.clone().unwrap().into_f32s();
+    let axis = curr.attrs.get("axis").unwrap().clone().into_i64();
```
laggui, line 1824: `axis` [is optional](https://onnx.ai/onnx/operators/onnx__OneHot.html#attributes), we should default to `-1` as per the spec if it's not specified.

laggui (changes requested):
Good job! We're almost there 😄 

This op has a bit more restrictions that the `Floor` op

crates/burn-import/src/burn/node/one_hot.rs
akshitgaur2005: Can I commit these suggestions directly through the option here on github?

laggui: Well, looks like I accidentally deleted a closing bracket in my code snippet 🫠 

The power (and curse) of using the GitHub suggested changes to directly commit the code lol

PR #2785, opened 2025-02-08
Remove empty format precision specifier

hkBst:
A format precision specifier consisting of a dot and no number actually does nothing and has no specified meaning. Currently this is silently ignored, but it may turn into a warning or error.

See https://github.com/rust-lang/rust/issues/131159 and https://github.com/rust-lang/rust/pull/136638

Please leave a comment there if you have any opinion about this. If you remember why this was written this way that could help improve the diagnostics rustc gives.


hkBst:
BTW Here's a log of the crater run that detected this, where you can see what the effect of the proposed stricter detection is: https://crater-reports.s3.amazonaws.com/pr-136638/try%2377e966048a5976133ba6b8f26d38d558f2a84355/reg/burn-tensor-0.16.0/log.txt

laggui (approved):

PR #2793, opened 2025-02-10
Fuse gather

nathanielsimard:
Added indexed tensor ids to keep track of tensors that can't be accessed using the same vectorization factor as other fused operations.


laggui (approved):
LGTM!

Just a minor comment

PR #2810, opened 2025-02-13
Optimize connected components

wingertge:
### Checklist

### Changes

Uses NdArray's inner `Vec` where possible in `into_data`.
Changes `bool_into_data` to not force the data to `bool`, and instead use `Backend::BoolElem`. Manually converting to `bool` no longer makes sense when we have an explicit type for boolean elements.
Uses raw pointers for generated code, since bounds checks are proven by nature of the generation algorithm and it's a free speedup.
Removes batch support for connected components since it didn't really make sense, especially for the `with_stats` version. This matches `opencv`.

### Testing

Passes all tests


laggui (approved):
Thanks for making the changes for the bool elem! Pretty much in line with other element types now since it can be backend-specific.

PR #2814, opened 2025-02-14
bump cubecl version with dummy implementations

maxtremblay:


crates/burn-cubecl/src/fusion/matmul/optimization.rs
```diff
@@ -373,7 +373,7 @@ impl FusedMatmul {
                 }
             }
             FusedMatmulSelector::Pipelined => {
-                match matmul_launch_kernel::<R, EG, PipelinedSelector<Accelerated>>(
+                match matmul_launch_kernel::<R, EG, SimplePipelinedSelector<Accelerated>>(
```
louisfd, line 376: I forgot to tell you, but I permuted some names, it's a bit confusing: 
- What was before _Standard_ is now _Simple_
- There is a new _SimplePipelined_ which should have its own entry 
- What was _Pipelined_ is now _DoubleBuffering_. 
- _Specialized_ did not change

louisfd (changes requested):
See my comment

PR #2815, opened 2025-02-14
Refactor/cubecl fusion

nathanielsimard:
**Improve Burn's compilation time significantly :)**  

- Moved backend re-exports to the root `burn` crate, so we don't need to recompile `burn-core` when working on a backend (better caching).  
- Created a new crate, `burn-cubecl-fusion`, that exports all optimizations used by `burn-cubecl` when fusion is activated.  
- Refactored all fusion optimizations to use a dynamic element type, avoiding Rust generics.  

Overall, the fusion part of `burn-cubecl` went from **41s** compilation time to **6.5s**! `burn-cubecl` now takes around **10s** and compiles in parallel with `burn-cubecl-fusion`.


laggui (approved):
Apart from what was discussed offline, just some minor comments.

Approving in advance.

Issue #2829, opened 2025-02-19
Training a model in no_std.


ivila:
### Feature description

I want to try using burn in a `no-std` environment for training a model inside ARM TrustZone with the [apache/incubator-teaclave-trustzone-sdk](https://github.com/tracel-ai/burn/issues/apache/incubator-teaclave-trustzone-sdk). However, I need to use the `optim` module in `burn-core` and the` burn-autodiff` crate, both of which **currently require std**.

### Feature motivation

1. make `optim` in `burn-core` support no-std.
2. make `burn-autodiff` support no-std.

### (Optional) Suggest a Solution

Actually, this can be easily achieved, as they already support no-std. The steps include:

1. Changing some imports from `std` to `core` or `alloc`.
2. Replacing `std::collections::{HashMap, HashSet}` with `hashbrown::{HashMap, HashSet}` when `no-std`.
3. Importing `num_traits::float::Float` when `no-std`.

PR #2830, opened 2025-02-19
Support training in no-std

ivila:
### Checklist

- [*] Confirmed that `run-checks all` script has been executed.
- [*] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

#2829

#### 1. burn-core
Make `optim` module support no-std:
1. Manually import `alloc::vec::Vec` and `alloc::boxed::Box` when `no-std`
2. Use `num_traits::float::Float` to support `sqrt` function in `no-std`
3. Mark all of the tests in `optim` `std only` as they require TestAutodiffBackend.

#### 2. burn-autodiff
Fix `no-std` feature and provide a facade `crate::libs` for clean codes:
1. Changing some imports from `std` to `core` or `alloc`.
2. Replacing `std::collections::{HashMap, HashSet}` with `hashbrown::{HashMap, HashSet}` when `no-std`
4. Use `num_traits::float::Float` to support some math functions in `no-std`
5. Mark `burn-autodiff` NO_STD in `xtask`

#### 3. burn
Fix dependency of `burn-autodiff`: add `default-features = false` on dependency of `burn-autodiff`

### Testing

Successfully compile a TA(trusted application) and run with it.

image

PS: I tried `run-checks.sh all` but it always failed by network reason(cannot reach huggingface.co).

### DIFF
--- a/crates/burn-autodiff/src/graph/node.rs
+++ b/crates/burn-autodiff/src/graph/node.rs
@@ -1,7 +1,7 @@
-use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::Arc;
+use core::sync::atomic::{AtomicU64, Ordering};
 
 use crate::checkpoint::retro_forward::RetroForward;
+use crate::libs::{Arc, Vec};
 use crate::runtime::AutodiffClientImpl;
 
 use super::Requirement;
--- a/crates/burn-autodiff/src/lib.rs
+++ b/crates/burn-autodiff/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(not(feature = "std"), no_std)]
 #![warn(missing_docs)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
@@ -34,3 +35,28 @@ pub use backend::*;
 
 #[cfg(feature = "export_tests")]
 mod tests;
+
+/// A facade around all the types we need from the `std`, `core`, and `alloc`
+/// crates. This avoids elaborate import wrangling having to happen in every
+/// module.
+mod libs {
+    cfg_block::cfg_block! {
+        if #[cfg(feature = "std")] {
+            pub use std::collections::{HashMap, HashSet};
+            pub use std::sync::Arc;
+            pub use std::{vec, vec::Vec};
+            pub use std::string::String;
+            pub use std::boxed::Box;
+            pub use std::format;
+            pub use std::borrow::ToOwned;
+        } else {
+            pub use hashbrown::{HashMap, HashSet};
+            pub use alloc::sync::Arc;
+            pub use alloc::{vec, vec::Vec};
+            pub use alloc::string::String;
+            pub use alloc::boxed::Box;
+            pub use alloc::format;
+            pub use alloc::borrow::ToOwned;
+        }
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -575,8 +575,11 @@ dependencies = [
  "burn-common",
  "burn-tensor",
  "burn-tensor-testgen",
+ "cfg_block",
  "derive-new 0.7.0",
+ "hashbrown 0.15.2",
  "log",
+ "num-traits",
  "spin",
 ]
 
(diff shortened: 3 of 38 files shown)


laggui (commented):
Will take a look at this tomorrow! But just a preliminary comment: I think we can just replace most types with their alloc types in all cases. That would probably just leave the collections stuff which is only used in a couple modules, so maybe we don't need to have a crate-level module to re-export compatible types.

ivila:
Updates:
1. Directly import from `alloc` and `cor`e, but retain a crate-level module named `collections` that contains `HashMap` and `HashSet` to avoid duplicate feature-checking code.
2. Add the `no-std` target `aarch64-unknown-none` and test the `no-std` build of `burn-autodiff` specifically for that target.

crates/burn-autodiff/src/lib.rs
```diff
@@ -34,3 +35,15 @@ pub use backend::*;
...
+/// A facade around for HashMap and HashSet.
+/// This avoids elaborate import wrangling having to happen in every module.
+mod collections {
+    cfg_block::cfg_block! {
+        if #[cfg(feature = "std")] {
+            pub use std::collections::{HashMap, HashSet};
+        } else {
+            pub use hashbrown::{HashMap, HashSet};
+        }
+    }
```
laggui, line 48: Since there is only one line now for collections, we can remove the cfg_block dependency and use

```suggestion
    #[cfg(feature = "std")]
    pub use std::collections::{HashMap, HashSet};
    #[cfg(not(feature = "std"))]
    pub use hashbrown::{HashMap, HashSet};
```

laggui (commented):
Thanks for extending the portability!

Looks good so far, just a some comments below.

> Mark all of the tests in optim std only as they require TestAutodiffBackend.

Now that you've made burn-autodiff no-std compatible this should be good to go (on 64-bit platforms, that is), no? Some tests save or load a state from a record file so these will need to be marked for std, but I think the rest should be fine 🤔 

Also, `NO_STD_64BIT_ONLY_CRATES` with burn-autodiff is only due to the atomics at his point, unless I am forgetting something else... If that's the case, we could use `portable-atomic` to add the support. But that can be left for the future, mostly thinking out loud here 😄

ivila:
Updates:

1. In `burn-autodiff`: Replace `core::sync::atomic::{AtomicU64, Ordering}` with `portable-atomic::{AtomicU64, Ordering}` to support 32-bit platforms.
2. In `burn-autodiff`: Remove the dependency on `cfg-block`.
3. In `xtask`: Revert changes in `xtask` and add `burn-autodiff` to `NO_STD_CRATES`.
4. In `burn-core`: Enable tests for the `optim` module in `no-std`, replace `BinFileRecorder` with `BinBytesRecorder` when `no-std`, and remove the `dev-dependency` on `tempfile`.

ivila:
checking on the failed build of target `thumbv6m-none-eabi`

laggui (commented):
Awesome! Thanks for the changes

> After consideration, I believe we should leave this as is and skip building burn-autodiff for the thumbv6m-none-eabi target.

Makes sense, this PR already makes it much closer to full no-std.

Will come back to this on Monday but should be good to go 🙂

ivila:
Updates: rebase to latest mainline

crates/burn-autodiff/src/graph/node.rs
```diff
@@ -1,5 +1,5 @@
-use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::Arc;
+use alloc::{sync::Arc, vec::Vec};
+use portable_atomic::{AtomicU64, Ordering};
```
laggui, line 2: I would keep this under a feature gate for `#[cfg(not(feature = "std"))]` and use the `std` atomics when available.

portable-atomic has an std feature but iirc it's not enabled by default.

laggui (commented):
One minor change left, missed it last time 😅 

After that LGTM!

crates/burn-autodiff/src/graph/node.rs
ivila: I think a feature gate of `#[cfg(not(target_has_atomic = "64"))]` is better?
image

laggui: Absolutely! Good catch 🙂

laggui (approved):
LGTM! Thanks for addressing the changes

xtask/src/commands/build.rs
```diff
@@ -31,7 +31,17 @@ pub(crate) fn handle_command(
                 if *build_target != "Default" {
                     build_args.extend(vec!["--target", *build_target]);
                 }
+
+                let mut crates = NO_STD_CRATES.to_vec();
+
                 if *build_target == ARM_NO_ATOMIC_PTR_TARGET {
+                    // Temperory remove `burn-autodiff` from building with the
```
laggui, line 38:
```suggestion
                    // Temporarily remove `burn-autodiff` from building with the
```

Issue #2478, opened 2024-11-11
Refactor `onnx-ir` to remove static shape inference and rely on rank inference only


antimora:
The goal of this refactor is to remove all static shape inference within the `onnx-ir` module and focus solely on rank inference. This shift aims to:

1. **Simplify the Shape Inference Process**:
   - Remove static shape inference logic and rely on rank-based inference to determine tensor ranks across operations. This approach will support the necessary runtime shapes without involving complex static shape calculations.
   
2. **Align with ONNX and Runtime Shapes**:
   - `onnx-ir` has distinctions between onnx-dynamic-shapes, runtime-shapes, and static-shapes, often causing confusion. By removing static shape inference, we emphasize runtime-shapes, which are central to burn-import, thereby reducing the cognitive load on developers.
   
3. **Enable Cleaner and More Focused Code**:
   - The current codebase has shape and rank inference intertwined, leading to redundant boilerplate and added complexity, particularly in ONNX operations. With this refactor, we aim to streamline code maintenance and make the inference logic easier for developers to follow and extend.

### Rationale

Static shape inference at build time should be handled separately, reducing the dependency on static shape information. This proposal emphasizes a cleaner, rank-inference-only approach that aligns with ONNX’s dynamic capabilities and reduces the overhead for burn-import contributors.

### Action Items

- **Remove Static Shape Inference Code**: Eliminate existing shape inference logic from `onnx-ir`.
- **Refactor for Rank-Only Inference**: Adjust the `onnx-ir` and tensor APIs to perform rank inference exclusively.
- **Testing**: Ensure the updated `onnx-ir` handles a variety of ONNX models with consistent rank inference across all operations.

### Benefits

This refactor will unburden developers from managing complex static shapes, streamline the review process, and make Burn more adaptable for dynamic ONNX models.

antimora:
CCing: @laggui, @skewballfox, tiruka, @hexd0t, @nathanielsimard

skewballfox:
While this is a bit short of runtime support, I've been thinking about a way to support compile time dynamic shapes with burn import. It essentially boils down to:
- change the type dim to be an enum where a value is either a usize or some kind of identifier ( I guess a string). 
- add a method to graph_builder for setting to pass in a Hashmap that provides the initial size of any named dynamic dimensions for the inputs
- each time a the size of a shape needs to be inferred (the enum is a named variant) we compute it's size from the information we have so far. 

[...]

PR #2831, opened 2025-02-19
Rename dim to rank for ONNX import

antimora:
## Pull Request Template

This PR is part of #2478 change. Initially I made tons of changes but later I decided to break into smaller PRs to be more manageable for a review.

### Checklist

### Related Issues/PRs

#2478

### Changes

Remained `dim` to `rank` in all places I could identify. 

### Testing

`run-checks all` succeeded.


crates/onnx-ir/src/rank_inference.rs
laggui, line 1: Duplicate of the `dim_inference` where the function was already renamed to `rank_inference`? We can keep one of the two modules 😄

crates/burn-import/src/burn/ty.rs
```diff
@@ -38,7 +38,7 @@ pub struct ScalarType {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ShapeType {
```
laggui, line 39: A bit confusing now that the type is for shape but it actually represents the rank 🤔

laggui (commented):
I have a couple of comments below.

Good idea to split the work into smaller chunks when possible, makes it more manageable and easier to review 😄

crates/onnx-ir/src/rank_inference.rs
antimora: Strange my previous code re-appeared.

crates/burn-import/src/burn/ty.rs
antimora: I am open to suggesting but rank might be okay here since it matches tensor rank.

laggui:
CI is failing for unrelated reasons (disk space issue on the runners). Gonna have to take a look at that 👀

antimora:
Tried to re-run the failed jobs but it's failing

laggui:
> Tried to re-run the failed jobs but it's failing

Cleared the cache for linux runners yesterday and that seemed to help but I have no provision to do this for the macos runners directly via GH. Something must be wrong though if we're hitting the disk space limit on our runners...

crates/burn-import/src/burn/ty.rs
laggui: Yeah nvm that's fine as is 👍

laggui (approved):
LGTM!

PR #2833, opened 2025-02-21
Changes how vectorization is applied during fusion

nathanielsimard:
Instead of limiting vectorization based on the most constraining factor, we read inputs multiple times when they can't match the highest vectorization factor. This way, we use the best vectorization factor for each tensor, no matter if others aren't able to.

### DIFF
--- a/crates/burn-cubecl-fusion/src/matmul/optimization.rs
+++ b/crates/burn-cubecl-fusion/src/matmul/optimization.rs
@@ -9,10 +9,15 @@ use burn_fusion::stream::Context;
 use burn_ir::{BinaryOpIr, TensorStatus};
 use cubecl::linalg::matmul::components;
 use cubecl::linalg::matmul::components::tile::accelerated::Accelerated;
+use cubecl::linalg::matmul::components::tile::TileMatmulFamily;
 use cubecl::linalg::matmul::components::MatmulProblem;
-use cubecl::linalg::matmul::kernels::matmul::{
-    DoubleBufferingSelector, MatmulSelector, SimpleSelector, SpecializedSelector,
-};
+use cubecl::linalg::matmul::kernels::matmul::double_buffering::DoubleBufferingAlgorithm;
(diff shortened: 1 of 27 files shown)


crates/burn-cubecl-fusion/src/on_write/io.rs
```diff
@@ -126,6 +173,113 @@ pub fn read_input<C: CubePrimitive>(
...
+                    &layout.tensor
+                }
+                Arg::Output(index, ..) => {
+                    let layout = outputs.tensors.index(index);
+                    &layout.tensor
+                }
+                _ => comptime![panic!("Invalid ref layout.")],
+            };
+
+            // Very brut force, not really efficient, but not easy to optimize and not a very
```
laggui, line 203: Tiny typo
```suggestion
            // Very brute force, not really efficient, but not easy to optimize and not a very
```

crates/burn-cubecl-fusion/src/matmul/optimization.rs
```diff
@@ -9,10 +9,15 @@ use burn_fusion::stream::Context;
...
-use cubecl::linalg::matmul::kernels::matmul::{
-    DoubleBufferingSelector, MatmulSelector, SimpleSelector, SpecializedSelector,
-};
+use cubecl::linalg::matmul::kernels::matmul::double_buffering::DoubleBufferingAlgorithm;
+use cubecl::linalg::matmul::kernels::matmul::simple::SimpleAlgorithm;
+use cubecl::linalg::matmul::kernels::matmul::specialized::SpecializedAlgorithm;
+use cubecl::linalg::matmul::kernels::matmul::{select_kernel, Algorithm};
+// use cubecl::linalg::matmul::kernels::matmul::{
+//     DoubleBufferingSelector, MatmulSelector, SimpleSelector, SpecializedSelector,
+// };
```
laggui, line 20: Dead
```suggestion
```

laggui (approved):

PR #2834, opened 2025-02-21
[Fix] `tch` upgrade

wingertge:
### Checklist

### Changes

Adds a hack to get newer `tch` versions to work with CUDA on Windows. This allows us to finally upgrade the version again.
It basically compiles a simple C++ program that uses the CUDA functions so the linker understands they're a dependency. If the CUDA library isn't found in the path, a fallback CPU dummy is used.

### Testing

Asked people in the Discord to test the build on various targets.


laggui (commented):
Great 😄 we can finally upgrade, plus we have a meaningful error message when something goes wrong with CUDA.

I'll make sure to run the tests locally for the auto download and manual setup on Monday for both linux & windows, but should be good to merge!

laggui (approved):

OmegaLambda1998:
Probably not helpful for the windows CUDA people but on linux with ROCM (RX7700S) this works perfectly

laggui:
It was also tested on Windows. The only issue is for automatic download builds, `tch` doesn't expose the downloaded libtorch dir like they do on Linux so we cannot detect the `torch_cuda.dll`. But we've added a helpful panic message when that happens.

The matching libtorch lib can still be automatically downloaded, but users must provide the path.

Issue #2049, opened 2024-07-22
Feature Request: Full Gemm operation / node


mtobin-tdab:
### Feature description

Introduction of ONNX Gemm operation conversion
https://onnx.ai/onnx/operators/onnx__Gemm.html

### Feature motivation

Useful for optimisations 

Currently get:

```
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "Gemm"    
DEBUG onnx_ir::from_onnx: renaming node "model/..../dense/MatMul_Gemm__6"    
ERROR burn_import::logger: PANIC => panicked at crates/onnx-ir/src/coalesce.rs:55:9:
Full Gemm node not supported yet. 
```

0rvar:
Got stuck on this now while trying to port wake word detection from tflite to Burn. Converted pretrained .tflite models to ONNX, but ModelGen doesn't accept any of them due to Gemm: `Full Gemm node not supported yet.`

antimora:
Closing this since GEMM is supported OP via #2841

PR #2841, opened 2025-02-24
Gemm

akshitgaur2005:
### Checklist

### Related Issues/PRs

#2049
#1544 

### Changes

Implement full Gemm Node for ONNX

### Testing

Not yet tested


akshitgaur2005:
Currently stuck on this. Any help will be appreciated!

```rust
 INFO onnx_ir::from_onnx: Parsing ONNX file: ./onnx-tests/tests/gemm/gemm.onnx
DEBUG onnx_ir::from_onnx: Number of nodes: 1
DEBUG onnx_ir::from_onnx: Number of inputs: 3
DEBUG onnx_ir::from_onnx: Number of initializers: 3
DEBUG onnx_ir::from_onnx: Number of outputs: 1
 WARN onnx_ir::from_onnx: Input A is also an initializer. Initializer as default values are currently not supported
 WARN onnx_ir::from_onnx: Input B is also an initializer. Initializer as default values are currently not supported
 WARN onnx_ir::from_onnx: Input C is also an initializer. Initializer as default values are currently not supported
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "Gemm"
DEBUG onnx_ir::from_onnx: renaming node "gemm_node"
DEBUG onnx_ir::from_onnx: adding node "gemm1"
 INFO onnx_ir::from_onnx: Finished parsing ONNX file: ./onnx-tests/tests/gemm/gemm.onnx
DEBUG burn_import::onnx::to_burn: Writing debug graph file: "./out/gemm.graph.txt"
DEBUG burn_import::burn::graph: Registering node => 'gemm'
 INFO burn_core::record::file: File exists, replacing
DEBUG burn_import::burn::graph: Building the scope nodes len => '1'
ERROR burn_import::logger: PANIC => panicked at crates/burn-import/src/burn/scope.rs:43:13:
No variable with name input1
thread 'main' panicked at crates/burn-import/src/burn/scope.rs:43:13:
No variable with name input1
```

crates/burn-import/onnx-tests/tests/gemm/gemm.py
laggui, line 1: The `gemm.onnx` file committed doesn't correspond to this script. As a matter of fact, I don't think this will actually generate a `Gemm` node.

laggui (commented):
The `gemm.onnx` model file is invalid, the graph doesn't even have any inputs.

image

You need to define the graph inputs to be used by the `Gemm` node.

crates/burn-import/onnx-tests/tests/gemm/gemm.py
akshitgaur2005: image

It shows inputs correctly for me?!

laggui: This corresponds to the `Gemm` *node* inputs, not the graph inputs

akshitgaur2005:
It works now with gemm.onnx file but not with the model in linear.onnx


```rust
 INFO onnx_ir::from_onnx: Parsing ONNX file: ./onnx-tests/tests/linear/linear.onnx
DEBUG onnx_ir::from_onnx: Number of nodes: 4
DEBUG onnx_ir::from_onnx: Number of inputs: 3
DEBUG onnx_ir::from_onnx: Number of initializers: 5
DEBUG onnx_ir::from_onnx: Number of outputs: 3
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "Gemm"
DEBUG onnx_ir::from_onnx: renaming node "/linear1_with_gemm/Gemm"
DEBUG onnx_ir::from_onnx: adding node "gemm1"
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "MatMul"
DEBUG onnx_ir::from_onnx: renaming node "/linear2_with_matmul/MatMul"
DEBUG onnx_ir::coalesce: peeking next node for bias conversion
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "MatMul"
DEBUG onnx_ir::from_onnx: adding node "matmul1"
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "MatMul"
DEBUG onnx_ir::from_onnx: renaming node "/linear3_with_matmul/MatMul"
DEBUG onnx_ir::coalesce: peeking next node for bias conversion
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "Add"
 WARN onnx_ir::from_onnx: Input /linear3_with_matmul/MatMul_output_0 not found, should only happen when peeking
DEBUG onnx_ir::from_onnx: adding node "matmul2"
 INFO onnx_ir::from_onnx: Finished parsing ONNX file: ./onnx-tests/tests/linear/linear.onnx
DEBUG burn_import::onnx::to_burn: Writing debug graph file: "./out/linear.graph.txt"
ERROR burn_import::logger: PANIC => panicked at crates/burn-import/src/burn/ty.rs:207:19:
"linear1_with_gemm.weight" is not a valid Ident
thread 'main' panicked at crates/burn-import/src/burn/ty.rs:207:19:
"linear1_with_gemm.weight" is not a valid Ident
```

laggui:
> It works now with gemm.onnx file but not with the model in linear.onnx

Ok will take a look at this in a bit

laggui (commented):
> "linear1_with_gemm.weight" is not a valid Ident

This is simply because the input name contains a period `.`, and when we convert the node input names to variables this is invalid. We could apply a filter to replace periods with underscores in the name.

But I see that you have removed the gemm_to_linear conversion, we could keep that when it matches the pattern!

akshitgaur2005:
> But I see that you have removed the gemm_to_linear conversion, we could keep that when it matches the pattern!

What would be the benefit of doing that though?

Handling Gemm nodes as GEMM would keep things straightforward, in my opinion, like if someday, there is some bug related to gemm node, we can debug that directly instead of going through cases and wondering if the bug is in gemm's implementation, linear's implementation or the conversion logic, etc....

akshitgaur2005:
> This is simply because the input name contains a period ., and when we convert the node input names to variables this is invalid. We could apply a filter to replace periods with underscores in the name.

So should that be handled in this PR or a separate one?

laggui:
> > But I see that you have removed the gemm_to_linear conversion, we could keep that when it matches the pattern!
> 
> What would be the benefit of doing that though?

Well the conversion could be handled by the `GemmNode` instead. If the inputs to the `Gemm` are initializers, then surely it was meant to be exported as linear layer. So importing it in Burn with a linear layer makes sense imo 🤔 

> So should that be handled in this PR or a separate one?

It might not be required if we perform the conversion to linear, as it was handled before. But otherwise it could be included in this PR if it's required to make things work!

akshitgaur2005:
Wait how do I actually do it though, any function trying to do so will have two output types, either linear and gemm or direct modification (no output) and gemm.

Sorry I am a beginner to rust.

laggui:
> Wait how do I actually do it though, any function trying to do so will have two output types, either linear and gemm or direct modification (no output) and gemm.

Maybe I lost the context through the replies, but I'm not sure exactly what you mean by that.

Generally speaking, you could return an enum and have two variants based on the input. But in this case specifically I am not sure exactly what function you are referring to (I'm guessing something to make a distinction between a regular Gemm node and a Linear gemm?)

So as I said earlier, keeping the current LinearNode conversion might just be easier since it is a distinct Gemm node pattern that we can already handle. Paths for debugging should still be pretty clear even if the input is a Gemm node since we output the graph info.

akshitgaur2005:
I was talking about a single function to delegate the responsibility of both cases, full gemm node and the one already implemented.

As the convert to linear function is executed first before the actual conversion to rust code, I guess just removing the panics should be enough. Let me just try that!

akshitgaur2005:
Could somebody help with this error-

```rust
  --> /home/akshit/Storage/projects/burn/target/debug/build/onnx-tests-41879a2725f93fcf/out/model/gemm.rs:49:9
   |
47 |     ) -> Tensor<B, 2> {
   |          ------------ expected `Tensor<B, 2>` because of return type
48 |         let gemm1_out1 = (input1.matmul(input2) * 1f32) + (input3 * 1f32);
   |                          ------------------------------------------------ here the type of `gemm1_out1` is inferred to be `Tensor<B, 3>`
49 |         gemm1_out1
   |         ^^^^^^^^^^ expected `2`, found `3`
   |
   = note: expected struct `Tensor<_, _, 2>`
              found struct `Tensor<_, _, 3>`

error[E0061]: this method takes 3 arguments but 2 arguments were supplied
    --> crates/burn-import/onnx-tests/tests/test_onnx.rs:2327:28
     |
2327 |         let output = model.forward(a, b);
     |                            ^^^^^^^------ argument #3 of type `Tensor<NdArray, 3>` is missing
     |
note: expected `3`, found `2`
    --> crates/burn-import/onnx-tests/tests/test_onnx.rs:2327:36
     |
2327 |         let output = model.forward(a, b);
     |                                    ^
     = note: expected struct `Tensor<_, _, 3>`
                found struct `Tensor<_, _, 2>`
note: expected `3`, found `2`
    --> crates/burn-import/onnx-tests/tests/test_onnx.rs:2327:39
     |
2327 |         let output = model.forward(a, b);
     |                                       ^
     = note: expected struct `Tensor<_, _, 3>`
                found struct `Tensor<_, _, 2>`
note: method defined here
    --> /home/akshit/Storage/projects/burn/target/debug/build/onnx-tests-41879a2725f93fcf/out/model/gemm.rs:42:12
     |
42   |     pub fn forward(
     |            ^^^^^^^
43   |         &self,
44   |         input1: Tensor<B, 3>,
     |         --------------------
45   |         input2: Tensor<B, 3>,
     |         --------------------
46   |         input3: Tensor<B, 3>,
     |         --------------------
help: provide the argument
     |
2327 |         let output = model.forward(/* Tensor<NdArray, 3> */, /* Tensor<NdArray, 3> */, /* Tensor<NdArray, 3> */);
```

I can't see how the rank is ending up at 3?!

crates/burn-import/onnx-tests/build.rs
```diff
@@ -43,6 +43,7 @@ fn main() {
         .input("tests/gather/gather_scalar_out.onnx")
         .input("tests/gather_elements/gather_elements.onnx")
         .input("tests/gelu/gelu.onnx")
+        .input("tests/gemm/gemm.onnx")
```
laggui, line 46: Included here

crates/burn-import/onnx-tests/tests/gemm/gemm.py
```diff
@@ -0,0 +1,67 @@
...
+        output_path (str): Path to save the ONNX model
+    """
+    # Define input and output shapes
+    batch_size = 2
+    m, k, n = 2, 3, 4  # A: (m, k), B: (k, n), C: (m, n)
+
+    # Define the graph inputs and outputs
+    A = helper.make_tensor_value_info('A', TensorProto.FLOAT, [batch_size, m, k])
+    B = helper.make_tensor_value_info('B', TensorProto.FLOAT, [batch_size, k, n])
+    C = helper.make_tensor_value_info('C', TensorProto.FLOAT, [batch_size, m, n])
```
laggui, line 21: From the graph/node definition here..

laggui (commented):
> Could somebody help with this error-
> [...]
> I can't see how the rank is ending up at 3?!

The test case you defined doesn't match the ONNX model in `onnx-tests/tests/gemm/gemm.onnx`, which is the one you're using given the defined input in the build.rs: `.input("tests/gemm/gemm.onnx")`.

That model expects three inputs of shape `[2, 2, 3]`, `[2, 3, 4]` and `[2, 2, 4]` respectively. Hence, the compiler correctly points out the error:

```
error[E0061]: this method takes 3 arguments but 2 arguments were supplied
    --> crates/burn-import/onnx-tests/tests/test_onnx.rs:2327:28
     |
2327 |         let output = model.forward(a, b);
[...]
note: method defined here
    --> /home/akshit/Storage/projects/burn/target/debug/build/onnx-tests-41879a2725f93fcf/out/model/gemm.rs:42:12
     |
42   |     pub fn forward(
     |            ^^^^^^^
43   |         &self,
44   |         input1: Tensor<B, 3>,
     |         --------------------
45   |         input2: Tensor<B, 3>,
     |         --------------------
46   |         input3: Tensor<B, 3>,
     |         --------------------
```

If you want to have tests for different node configurations, you might need to have additional ONNX models (to include accordingly in the `onnx_tests`).

akshitgaur2005:
The generated codecov report contains regression only on the files not even part of this PR, and I haven't removed any test??

New to these things btw, so sorry if it is something stupid

laggui:
> The generated codecov report contains regression only on the files not even part of this PR, and I haven't removed any test??
> 
> New to these things btw, so sorry if it is something stupid

You can safely ignore, the codecov bot probably detected those as changes since they were included from the merge main commit 😅

crates/burn-import/src/burn/node/gemm.rs
```diff
@@ -0,0 +1,142 @@
...
+use burn::record::PrecisionSettings;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+#[allow(clippy::too_many_arguments)]
+#[derive(Debug, Clone, new)]
+pub struct GemmNode {
+    pub a: TensorType,
+    pub b: TensorType,
+    pub c: Option<TensorType>,
```
laggui, line 12: Actually, `C` could also be a scalar. So we should probably accept `Option<Type>`.

crates/burn-import/src/burn/node/gemm.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub output: TensorType,
+    pub alpha: f32,
+    pub beta: f32,
```
laggui, line 15: For `alpha = 1.0` and `beta = 1.0` we should probably avoid the multiplication. In this case, it might be easier to take `Option<f32>` instead and check the attribute values during conversion.

crates/burn-import/src/burn/node/gemm.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        let product = quote! {#a.clone().matmul(#b.clone())};
+        let scaled_product = quote! {#product * #alpha};
+
+        if let Some(ref c) = self.c {
+            let c = scope.tensor_use_owned(c, node_position);
+
+            quote! {
+                let mut d = (#scaled_product).zeros_like();
+                d = #c.unsqueeze();
```
laggui, line 65: Why not simply `#c.unsqueeze()`? Creating a new tensor filled with zeros just to reassign the variable to the unsqueezed tensor is unnecessary.

crates/onnx-ir/src/rank_inference.rs
```diff
@@ -1074,3 +1075,53 @@ fn one_hot_output_shape(node: &mut Node) {
...
+    } else {
+        (b_shape[0], b_shape[1])
+    };
+
+    if k_a != k_b {
+        panic!(
+            "Gemm: inner dimensions do not match (got {} and {})",
+            k_a, k_b
+        );
+    }
```
laggui, line 1117: This check should not be performed here. The shapes are not always available, so we can let the execution crash during runtime if the provided inputs are not valid.

crates/onnx-ir/src/coalesce.rs
```diff
@@ -51,8 +52,8 @@ pub(crate) fn convert_gemm_to_linear(node: &mut Node) {
 
         // Transpose the weights
         transpose_linear_node_weights(node);
-    } else {
-        panic!("Full Gemm node not supported yet.");
+        //} else {
+        //    panic!("Full Gemm node not supported yet.");
```
laggui, line 56: We can delete the commented lines

laggui (changes requested):
Good job! Glad to see you managed to make it work.

The current implementation is a bit restrictive, so we should either cover the whole scope or panic with a clear message for unsupported configurations (e.g., `C` is a scalar).

We should probably add some other tests since the `Gemm` node is only tested for all inputs `A`, `B,` and `C` of the same shape. Not strictly required to cover all the possible combinations since there are quite a few optional attributes and conditions in this particular case. For example, we could add another model (or include another node in the current ONNX) with one of the inputs being transposed and no `C`. If we add support for scalar `C` input, we should also include it in a test.

See also my comments on the implementation below 🙂

crates/burn-import/src/burn/node/gemm.rs
akshitgaur2005: >C (optional, heterogeneous) - T:
>
>Optional input tensor C. If not specified, the computation is done as if C is a scalar 0. The shape of C should be unidirectional broadcastable to (M, N).


>Type Constraints
>
>    T in ( tensor(bfloat16), tensor(double), tensor(float), tensor(float16), tensor(int32), tensor(int64), tensor(uint32), tensor(uint64) ):
>
>    Constrain input and output types to float/int tensors.

C should not be a scalar in my opinion.
According to the spec, C would be scalar only when no input is given for it.

crates/burn-import/src/burn/node/gemm.rs
laggui: Like I said, pretty sure the compiler should be able to infer the const rank `D` from the other tensors used in the operation, e.g.

```rust
(input1.matmul(input2) * alpha) + input3.unsqueeze()
```
so you don't need to specify it. Are you getting errors otherwise?

Also, you just made me realize that I missed the `.clone()` on `#a` and `#b`. These are not necessary 🙂

akshitgaur2005: Yes that is what I was doing at first, but compiler complains for these.

laggui: Just took 2min to clone and make the change.. doesn't seem to be an issue for the compiler on my end 🤔 

This is what the forward looks like with my changes:
```rust
        let a = scope.tensor_use_owned(&self.a, node_position);
        let b = scope.tensor_use_owned(&self.b, node_position);

        let output = &self.output.name;
        let alpha = self.alpha;
        let beta = self.beta;
        let trans_a = self.trans_a;
        let trans_b = self.trans_b;

        let a = if trans_a != 0 {
            quote! {#a.transpose()}
        } else {
            quote! {#a}
        };

        let b = if trans_b != 0 {
            quote! {#b.transpose()}
        } else {
            quote! {#b}
        };

        let product = quote! {#a.matmul(#b)};
        let scaled_product = quote! {#product * #alpha};

        if let Some(ref c) = self.c {
            let c = scope.tensor_use_owned(c, node_position);

            quote! {
                let #output = (#scaled_product) + (#c.unsqueeze() * #beta);
            }
        } else {
            quote! {
                let #output = #scaled_product;
            }
        }
    }
```

Compiler is happy and the tests pass!

akshitgaur2005: Btw, is there a way to just run the test for this, as running each test takes a very loong time

akshitgaur2005: Oh Ok! It works now, moving this-
```rs
let #output = (#scaled_product) + (#c.unsqueeze() * #beta);
```
to a single line seems to have solved the issue! Otherwise it was giving the error that-


```rs
48 |         let mut d = (input1.matmul(input2) * 1f32).zeros_like();
   |                                    ------ value moved here
49 |         d = input3.unsqueeze();
50 |         let gemm1_out1 = (input1.matmul(input2) * 1f32) + (d * 1f32);
   |                                         ^^^^^^ value used here after move
```

crates/burn-import/src/burn/node/gemm.rs
laggui: It states that the tensor `C` should be unidirectional broadcastable to `(M, N)`. This statement is valid for a scalar. It's also noted in the [broadcasting specs](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md#unidirectional-broadcasting).

> the computation is done as if C is a scalar 0
This excerpt also supports the statement. *as if* does not mean that `C` cannot be a scalar. On the contrary, it indicates that the computation can be done for a scalar `C`, and exemplifies the special case of `C = 0`.

akshitgaur2005: Hey! What should I do here?-

```rust
       if let Some(ref c) = self.c {
            //let c = scope.tensor_use_owned(c, node_position);
            let c = match c {
                Type::Tensor(tensor) => scope.tensor_use_owned(tensor, node_position),
                Type::Scalar(scalar) => TensorType::from(scalar),
                _ => panic!("C should be Tensor or Scalar!"),
            };
```
```gemm.rs```

For context, C is now Option</Type/>-

```rs
let c = if node.inputs.len() > 2 {
            Some(Type::from(node.inputs.get(2).unwrap()))
        } else {
            None
        };
```
```to_burn.rs```

Comment (laggui) [reply crates/burn-import/src/burn/node/gemm.rs:12]: The match is fine but the results cannot be the same: one is a tensor and the other a scalar. So the codegen will be different for both arms. For the scalar you don't (and can't) call `unsqueeze()` anyway, unless you create a 1D tensor from the scalar data but that's not optimal. The implementations can simply branch out based on the type.

Comment (akshitgaur2005) [reply crates/burn-import/src/burn/node/gemm.rs:12]: No, what I was asking for is that what is the correct way to extract and use that scalar value. Currently I am doing this, which I understand is not the best way to do it-

```rs
if let Some(ref c) = self.c {
            //let c = scope.tensor_use_owned(c, node_position);

            match c {
                Type::Tensor(tensor) => {
                    let c_tensor = scope.tensor_use_owned(tensor, node_position);
                    quote! {
                        let #output = (#scaled_product) + (#c_tensor.unsqueeze() * #beta);
                    }
                }
                Type::Scalar(scalar) => {
                    //let c_scalar = scope.tensor_use_owned(scalar, node_position);
                    let c_scalar = scalar.to_full_tensor(&[1 as usize]);
                    println!("{c_scalar}");
                    quote! {
                        let #output = (#scaled_product) + (#c_scalar.unsqueeze() * #beta);
                    }
                }
                _ => panic!("C should be Tensor or a Scalar!"),
            }
```

crates/onnx-ir/src/rank_inference.rs
akshitgaur2005: The shape information is needed here though-

```rs
 node.outputs[0].ty = ArgType::Tensor(TensorType {
        rank: 2,
        shape: Some(vec![m, n]),
```

crates/burn-import/src/burn/node/gemm.rs
laggui: Ahh sorry I didn't catch that.

You can use the identifier for the scalar. See how it's done in other nodes, e.g. mask_where:
https://github.com/tracel-ai/burn/blob/a6b5210aee3c367cc58875e252efd9ab3cdf0b81/crates/burn-import/src/burn/node/mask_where.rs#L35-L45

crates/onnx-ir/src/rank_inference.rs
laggui: Not mandatory no, it can be left as `None` . The shape is not used at compile-time.

Your comment made me realize that you hardcoded the rank as `2`. That's incorrect, it should be the same as the input rank.

crates/burn-import/src/burn/node/gemm.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    use super::*;
+    use crate::burn::{
+        graph::BurnGraph,
+        node::{gemm::GemmNode, test::assert_tokens},
+        ScalarKind, ScalarType, TensorType,
+    };
+
+    #[test]
+    fn test_codegen_nodes() {
```
laggui, line 120: Like I mentioned in the last review, we should add tests for some of the other configurations! Especially for the `alpha` or `beta` != 1.0 and no `C`.

At the very least for the codegen, the test ONNX model itself is not as strict of a requirement.

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -1317,6 +1319,19 @@ impl ParsedOnnxGraph {
...
     }
+
+    fn gemm_conversion(node: Node) -> GemmNode {
+        let a = TensorType::from(node.inputs.first().unwrap());
+        let b = TensorType::from(node.inputs.get(1).unwrap());
+        let c = if node.inputs.len() > 2 {
+            Some(Type::from(node.inputs.get(2).unwrap()))
+        } else {
+            None
+        };
```
laggui, line 1330: Same as
```suggestion
        let c = node.inputs.get(2).map(Type::from);
```

crates/onnx-ir/src/rank_inference.rs
```diff
@@ -1074,3 +1075,19 @@ fn one_hot_output_shape(node: &mut Node) {
...
 }
+
+fn gemm_output_shape(node: &mut Node) {
+    let a_rank = match &node.inputs[0].ty {
+        ArgType::Tensor(tensor) => tensor.rank,
+        _ => panic!("Input A should be a tensor!"),
+    };
+
+    node.outputs[0].ty = ArgType::Tensor(TensorType {
+        rank: a_rank,
```
laggui, line 1086: Since broadcasting is supported, the output rank should be `max(a_rank, b_rank)`.

laggui (commented):
Some minor comments left, mostly need to correctly handle the output rank and ideally add some tests for other node configurations.

Should be good to go after that! 🙂

crates/burn-import/src/burn/node/gemm.rs
akshitgaur2005: How am I supposed to set the values of attrs in the test? And each test would have to have a different test_codegen?

Could you please tell how to do these things?

laggui: Sorry, guess I should have been more precise 😅 

> a different test_codegen?

Yep, precisely! Basically each instance can be defined as a separate codegen test. See for example the gather node: https://github.com/tracel-ai/burn/blob/main/crates/burn-import/src/burn/node/gather.rs#L150

akshitgaur2005: I have added two new test cases, one for alpha=beta=0.5 and one for no c. Although codegen test works fine, it does not care what the output is? Could you take a look at that

For example-
```rs
let expected =
            Tensor::<Backend, 2>::from_data(TensorData::from([[19.0, 22.], [43., 50.]]), &device);
```
changing the values in this makes no difference?

akshitgaur2005:
The failure does not have anything to do with this PR-


```rs
Audit Rust Dependencies
  Command line: cargo audit -q --color always
  Crate:     ring
  Version:   0.17.9
  Title:     Some AES functions may panic when overflow checking is enabled.
  Date:      2025-03-06
  ID:        RUSTSEC-2025-0009
  URL:       https://rustsec.org/advisories/RUSTSEC-2025-0009
  Solution:  Upgrade to >=0.17.12
  Dependency tree:
  ring 0.17.9
  ├── rustls-webpki 0.102.8
  │   └── rustls 0.23.23
  │       ├── ureq 2.12.1
  │       │   ├── torch-sys 0.19.0
  │       │   │   ├── tch 0.19.0
  │       │   │   │   └── burn-tch 0.17.0
  │       │   │   │       ├── burn-vision 0.17.0
  │       │   │   │       ├── burn-core 0.17.0
  │       │   │   │       │   ├── burn-train 0.17.0
  │       │   │   │       │   │   └── burn 0.17.0
  │       │   │   │       │   │       ├── wgan 0.1.0
  │       │   │   │       │   │       ├── text-generation 0.17.0
  │       │   │   │       │   │       ├── text-classification 0.17.0
  │       │   │   │       │   │       ├── simple-regression 0.17.0
  error: 1 vulnerability found!
  ```

laggui (approved):
LGTM! Thanks for addressing all the comments 🙂 

I updated the Gemm node support in the ONNX ops list

akshitgaur2005:
Ah shit sorry! Forget something everytime! :disappointed:

laggui:
> Ah shit sorry! Forget something everytime! 😞

No worries! I just updated the dependency causing the CI issue so that should also be resolved when the PR is merged.

PR #2842, opened 2025-02-24
[Feat] Basic Morphology ops

wingertge:
### Checklist

### Changes

Adds new basic morphology operations (dilate/erode) to `burn-vision`. The implementation is based on `opencv`'s vectorized implementation and currently relies on a fork of `pulp` and a new wrapper crate to provide type generics for it, [`macerator`](https://github.com/wingertge/macerator). Once the required changes to pulp are merged, I can publish a release for the crate.
Makes slight optimizations to `TensorData` functions so booleans don't have their bit patterns checked for every bytemuck cast. This overhead adds up for fast operations.

### Testing

Adds tests on sample images for all different morphology options.


crates/burn-vision/src/ops/base.rs
```diff
@@ -132,10 +185,110 @@ pub trait VisionOps<B: Backend> {
...
+    ) -> FloatTensor<Self> {
+        let input = Tensor::<Self, 3>::from_primitive(TensorPrimitive::Float(input));
+        morph(input, kernel, MorphOp::Dilate, opts)
+            .into_primitive()
+            .tensor()
+    }
+}
+
+/// Vision ops on quantized float tensors
+pub trait QVisionOps: Backend {
```
laggui, line 268: Is it practically relevant to offer vision ops for quantized inputs? Especially in this case where the input is expected to be an image tensor for morphological ops 🤔

crates/burn-vision/src/backends/cpu/morphology/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+        DType::F16 | DType::BF16 => morph_typed::<B, K, f32>(
+            data.convert::<f32>(),
+            shape,
+            kernel,
+            op,
+            iter,
+            btype,
+            bvalue,
+            &device,
+        ),
```
laggui, line 90: Shouldn't the output be converted back to the input dtype (f16 or bf16)?

laggui (commented):
Last time I had to mess with morphological ops it was in MATLAB 😅 been a while lol

I have some comments mostly regarding usage, the implementation seems to be thoroughly tested so I'll trust the results 😄

crates/burn-vision/src/ops/base.rs
wingertge: Probably not strictly necessary but it's barely any overhead since I already want to support u8/i8.

crates/burn-vision/src/backends/cpu/morphology/mod.rs
wingertge: `from_data` already converts to the native datatype of the backend.

crates/burn-vision/src/ops/base.rs
laggui: Yeah the implementation is basically free 😄 was just wondering

crates/burn-vision/src/backends/cpu/morphology/mod.rs
laggui: Ahh yes I missed that in `morph_typed`.

The only types currently affected are float tensors since we can cast tensors to a different dtype than native (hence my comment). If we want to preserve the input dtype we would need to call `from_data_dtype` to enforce that.

But that might not be a strict requirement.

PR #2843, opened 2025-02-25
examples: add mnist-no-std

ivila:
### Checklist

### Related Issues/PRs

#2830 

### Changes

Add example: mnist-no-std

### Testing

Successfully run the two examples.


ivila:
Test failed by `no space left on device`.

image

laggui:
> Test failed by `no space left on device`.

Yeah we're having issues on our macos runners, don't think this is related to your changes.

Will take a look at your example tomorrow! 🙂

examples/mnist-no-std/src/inference.rs
```diff
@@ -0,0 +1,62 @@
+use crate::{proto::*, util};
+use alloc::vec::Vec;
+use burn::{
+    prelude::*,
+    record::{FullPrecisionSettings, Recorder, RecorderError},
+    tensor::ElementComparison,
+};
+use burn_no_std_tests::{mlp::MlpConfig, model::MnistConfig};
```
laggui, line 8: This model definition comes from the mnist example anyway, so might as well define it in this example 😄

examples/mnist-no-std/src/inference.rs
```diff
@@ -0,0 +1,62 @@
...
+        let output = self.0.forward(tensor);
+        let output = burn::tensor::activation::softmax(output, 1).into_data();
+        assert_eq!(output.num_elements(), 10);
+        output
+            .iter::<f32>()
+            .enumerate()
+            .max_by(|a, b| a.1.cmp(&b.1))
+            .map(|(index, _)| index)
+            .unwrap() as u8
+    }
```
laggui, line 33: We can simply call `output.argmax(1)` and then `.into_scalar()`

examples/mnist-no-std/src/train.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        trainer.replace(NoStdTrainer::new(DEVICE, seed).with_learning_rate(lr));
+    }
+
+    pub fn train(images: &[u8], labels: &[u8]) -> Output {
+        assert!(images.len() % MNIST_IMAGE_SIZE == 0);
+        let images: &[MnistImage] = bytemuck::cast_slice(images);
+
+        let mut trainer = TRAINER.lock();
+        assert!(trainer.is_some());
```
laggui, line 129: Instead of performing assertions and then `.unwrap()` it would be nicer to `.expect(...)` with a clear message (e.g. "Trainer has not been initialized" for this particular case).

examples/mnist-no-std/examples/train.rs
```diff
@@ -0,0 +1,74 @@
...
+        for (iteration, data) in valid_datasets.chunks(args.batch_size).enumerate() {
+            let images: Vec<MnistImage> = data.iter().map(|v| v.0).collect();
+            let labels: Vec<u8> = data.iter().map(|v| v.1).collect();
+            let output = no_std_world::valid(bytemuck::cast_slice(images.as_slice()), &labels);
+            println!(
+                "[Valid - Epoch {} - Iteration {}] Loss {:.3} | Accuracy {:.3} %",
+                epoch, iteration, output.loss, output.accuracy,
+            );
+        }
+    }
```
laggui, line 68: Since you defined a nice interface for the trainer (and `no_std_world` functions), why not a `fit(...)` function for this?

examples/mnist-no-std/README.md
```diff
@@ -0,0 +1,86 @@
...
+    Trains a new model and exports it to the given path.
+
+    ``` shell
+    cargo run --release --example train
+    ```
+
+    This example downloads the MNIST dataset, trains a new model, and outputs
+    the model to the given path(default: `model.bin`).
+
+    You can run `cargo run --release --example train -- --help` for detail
```
laggui, line 21: Small typo
```suggestion
    You can run `cargo run --release --example train -- --help` for detailed
```

examples/mnist-no-std/README.md
```diff
@@ -0,0 +1,86 @@
...
+    # cargo run --release --example infer -- --binary-path=samples/8.bin
+    cargo run --release --example infer -- -i ${binary_path}
+    ```
+
+    This command loads the model the model from the given
+    path(default: `model.bin`) and tests it with the given binary, and prints
+    the inference result. For convenience, you can use the sample binaries in
+    the `samples` folder.
+
+    You can run `cargo run --release --example infer -- --help` for detail
```
laggui, line 39: Small typo
```suggestion
    You can run `cargo run --release --example infer -- --help` for detailed
```

laggui (commented):
Thanks for contributing a no-std example 🙂 

I have some minor comments below.

Not sure if MNIST training on (tens of thousands) images is the best example for no-std 🤔 It gets the point across to demonstrate the logic, components and mechanisms used, but other tasks might be more relevant. I don't really have any suggestions apart from other toy examples though, so it can remain like that for now.

examples/mnist-no-std/examples/train.rs
ivila: Because there is no `println` in `no_std`, the caller must invoke functions step by step and receive feedback from `no_std_world`. For example, in TrustZone, there are two systems running on the same machine: one is Linux, and the other is TEEOS (the `no_std` world, bare metal env). The caller from Linux invokes functions via SMC (Secure Monitor Call) repeatedly, receives output through shared memory, and prints it to the screen.

Should I add some description in `README.md`?

ivila:
Both of the tests failed by `no space left on device`

examples/mnist-no-std/examples/train.rs
laggui: Ahh I thought maybe the progress report via print could be abstracted away. But now I see what you mean given the two systems.

It doesn't necessarily have to be in the README, could also just be some appropriate inline comments in the training loop. I'll leave this up to you 🙂

laggui (commented):
> Both of the tests failed by `no space left on device`

MacOS runners should be fixed now. Not sure why your linux runner also failed, seems it had  bigger disk usage during setup than the other instances 🤔 deleted some cache to see if it fixes the issue.

ivila:
> > Both of the tests failed by `no space left on device`
> 
> MacOS runners should be fixed now. Not sure why your linux runner also failed, seems it had bigger disk usage during setup than the other instances 🤔 deleted some cache to see if it fixes the issue.

I think I may have found the reason. Please hold on while I am checking.

ivila:
The reason why `no space left on device`: I forgot that I should download the MNIST dataset to the temp folder instead of the current working directory😂

Updates:
1. For `train`: Download the MNIST dataset to the `temp` folder.
2. For `train`: Add comments to describe the training workflow.

laggui:
> The reason why `no space left on device`: I forgot that I should download the MNIST dataset to the temp folder instead of the current working directory😂

The example is not actually executed on the CI, so it shouldn't download MNIST. I see that it (understandably) still failed. But I still don't think this is related to your changes specifically.

examples/mnist-no-std/examples/train.rs
laggui: Just saw the comments. I think there's enough details for this use case thanks.

If I understand correctly, this example isn't specific to no-std but more so to a secure enclave environment where there is a host and target like TrustZone. I'm not super familiar with this space, so not sure about practical use cases. Is this to offload the processing to specific target hardware while the host is only used for logging / tracking? So it could not only be applied to a secure environment setup (i.e., your described use case with TrustZone), but also IoT or embedded device with host communication?

ivila: Yes, you are correct—this example is specifically designed for scenarios running in a bare-metal environment.

Unlike the `mnist` and `custom-training-loop` examples, which perform training in a std environment, no_std_world operates under significant limitations. Take OP-TEE (a TrustZone OS) as an example—though it's not the only case, as similar constraints apply to other bare-metal environments:

1. Lack of std features: There is no standard network support. OP-TEE, for instance, provides a non-standard TCPStream with limited functionality, while some IoT environments may only support Bluetooth without any TCP at all. As a result, fetching data must be handled by the caller (e.g., Linux). In this example, Linux fetches the MNIST dataset and provides the images as parameters (via shared memory) when communicating with OP-TEE.

2. No interruptions: Unlike Linux, where you can terminate training at any time (e.g., using Ctrl-C or kill), OP-TEE does not support interruptions. Once execution starts, it must run to completion. Even shutting down Linux will not interrupt the process, as the machine must wait for OP-TEE to shut down first—which won’t happen until all jobs have finished execution. That’s why, in the example, the caller (running in Linux) consumes parameters like `num_epochs` and `batch_size`. Each call to `no_std_world` is kept short to allow cancellation between calls.

This is the reasoning behind the statement in `README.md`:

> "It is the caller's responsibility to provide the data and control the workflow."

laggui: Thank you for understanding!

Please feel free to submit the example when the separate repository comes along 🙂

PR #2846, opened 2025-02-26
MatMulInteger

NewBornRustacean:
# Add MatMulInteger ONNX operation to Burn
Hello, Burn community! I'm trying to add `MatMulInteger` to `burn-import`

## Checklist

## Related Issues/PRs
This PR addresses part of issue #1714  which requests implementation of missing ONNX operations in the `burn-import` crate. Specifically, it implements the MatMulInteger operation which was listed under "Harder Ops (Not Similar to Existing Implemented Ops)".

## Changes
This PR implements the MatMulInteger (ONNX opset 10) operation in the burn-import crate, allowing users to import ONNX models that utilize quantized matrix multiplication. The implementation:

1. Creates a new `MatMulIntegerNode` structure to handle the operation
2. Implements proper dimension handling and broadcasting for tensors with different ranks
3. Supports optional zero points for quantized inputs (both scalar and tensor)
4. Follows the ONNX specification for integer matrix multiplication
5. Handles the conversion of quantized inputs with appropriate zero point subtraction

## Testing
- [x] Unit tests for code generation that verify the correct Rust code is produced
- [ ] Validation of tensor types to ensure compatibility with ONNX specifications (int8/uint8 inputs, int32 outputs)
- [ ] Proper handling of different tensor dimensions with broadcasting
- [ ] Tests for both scalar and tensor-based zero points

Additional integration tests with actual ONNX models containing the MatMulInteger op will be added.


crates/burn-import/src/burn/node/matmul_integer.rs
```diff
@@ -0,0 +1,213 @@
...
+                        let #output = (#lhs.unsqueeze_dims(&#axes) - #a_zero_point).matmul((#rhs - #b_zero_point)).squeeze(#squeeze_dim);
+                    }
+                } else {
+                    quote! {
+                        let #output = (#lhs.unsqueeze_dims(&#axes) - #a_zero_point).matmul((#rhs - #b_zero_point));
+                    }
+                }
+            }
+            Ordering::Equal => quote! {
+                let #output = (#lhs - #a_zero_point).matmul((#rhs - #b_zero_point));
```
laggui, line 132: Just a quick note: matmul is [only available for float tensors](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/api/float.rs#L181) right now.

We could probably move this to the numeric operations, but I think we would need to expose the cubecl implementation for numerics (iirc it's only exposed for floats, but actually implemented for numerics).

This is probably gonna be a bit of a blocker for the current PR 😅

laggui:
Also, ignore the macos CI faling.. we're having issues with the space on the runners since yesterday 🥲

crates/burn-import/src/burn/node/matmul_integer.rs
laggui: In case you haven't stumbled upon the contributor guide, there are a couple of steps detailed to add a new operator: https://burn.dev/contributor-book/guides/onnx-to-burn-conversion-tool.html#adding-new-operators

NewBornRustacean: Thanks to quick and kind review! that was my concern also(the matmul thing). 
I'll dig more and ping you when it's ready 👍 

(thanks to heads up for the ci failing, btw)

NewBornRustacean: wow, thanks for this superfast reply!
yeah, exactly it is "harder" jobs 😅 but interesting!

for now, I was trying to add `int_matmul` and implement `int_matmul` for `NdArrayTensor` to simple test.
(following the guide you shared. thanks!)

laggui: Sounds good! I think tch should also be easy since their matmul supports integers out of the box iirc.

antimora:
Implemented by https://github.com/tracel-ai/burn/pull/3672

PR #2851, opened 2025-02-28
[Feat] SIMD acceleration for ndarray backend

wingertge:
### Checklist

### Changes

Adds SIMD versions of binary, binary_elem, unary ops.
Adds SIMD versions of conv2d, max_pool2d, avg_pool2d.

Since the SIMD versions have tighter requirements than normal ops (not all operations supported on scalars are supported on SIMD, and other requirements apply for convolution and pooling), each operation that has a SIMD version will first attempt to launch it and return the value if successful, then fall back to a normal scalar version.
All SIMD versions are gated behind the `simd` feature which is enabled by default. This is just to be give users the option to disable it on targets like WASM that don't support it, and reduce compilation time.

### Testing

Most tests don't apply to SIMD due to the tighter requirements, but `craft-burn` has the expected output and uses a variety of convolution and pooling layers in many different configurations. I've used it before to reproduce hard to track down bugs, so it's pretty comprehensive.

### Benchmarking

Listing all numbers would be excessive, but the relative speedup ranges from `~1.0x` (for simple ops that are memory bound and were already auto-vectorized) to `~10x` (one of the convolution benches, pooling). The typical speedup is around `4-6x`.

#### Convolution benches:
No SIMD (except auto-vectorization):
```
| Benchmark                                       | Feature | Backend   | Device | Median         |
|-------------------------------------------------|---------|-----------|--------|----------------|
| conv2d-input_16x512x512_weight_16x3x3_stride_1  | ndarray | `ndarray` | Cpu    | 193.628ms      |
| conv2d-input_3x227x227_weight_96x11x11_stride_4 | ndarray | `ndarray` | Cpu    | 52.052ms       |
| conv2d-input_3x231x231_weight_96x11x11_stride_4 | ndarray | `ndarray` | Cpu    | 53.430ms       |
| conv2d-input_3x227x227_weight_64x7x7_stride_2   | ndarray | `ndarray` | Cpu    | 62.588ms       |
| conv2d-input_64x224x224_weight_64x7x7_stride_2  | ndarray | `ndarray` | Cpu    | 1.134s         |
| conv2d-input_96x24x24_weight_256x5x5_stride_1   | ndarray | `ndarray` | Cpu    | 54.090ms       |
| conv2d-input_256x12x12_weight_512x3x3_stride_1  | ndarray | `ndarray` | Cpu    | 41.314ms       |
| conv2d-input_3x224x224_weight_64x3x3_stride_1   | ndarray | `ndarray` | Cpu    | 47.512ms       |
| conv2d-input_64x112x112_weight_128x3x3_stride_1 | ndarray | `ndarray` | Cpu    | 133.503ms      |
| conv2d-input_64x56x56_weight_64x3x3_stride_1    | ndarray | `ndarray` | Cpu    | 20.241ms       |
| conv2d-input_128x28x28_weight_128x3x3_stride_1  | ndarray | `ndarray` | Cpu    | 17.958ms       |
| conv2d-input_256x14x14_weight_256x3x3_stride_1  | ndarray | `ndarray` | Cpu    | 29.423ms       |
| conv2d-input_512x7x7_weight_512x3x3_stride_1    | ndarray | `ndarray` | Cpu    | 60.309ms       |
| conv2d-input_96x224x224_weight_64x1x1_stride_1  | ndarray | `ndarray` | Cpu    | 374.665ms      |
```
New SIMD implementation:
```
| Benchmark                                       | Feature      | Backend   | Device | Median         |
|-------------------------------------------------|--------------|-----------|--------|----------------|
| conv2d-input_16x512x512_weight_16x3x3_stride_1  | ndarray-simd | `ndarray` | Cpu    | 43.854ms       |
| conv2d-input_3x227x227_weight_96x11x11_stride_4 | ndarray-simd | `ndarray` | Cpu    | 6.077ms        |
| conv2d-input_3x231x231_weight_96x11x11_stride_4 | ndarray-simd | `ndarray` | Cpu    | 3.771ms        |
| conv2d-input_3x227x227_weight_64x7x7_stride_2   | ndarray-simd | `ndarray` | Cpu    | 7.177ms        |
| conv2d-input_64x224x224_weight_64x7x7_stride_2  | ndarray-simd | `ndarray` | Cpu    | 120.966ms      |
| conv2d-input_96x24x24_weight_256x5x5_stride_1   | ndarray-simd | `ndarray` | Cpu    | 20.297ms       |
| conv2d-input_256x12x12_weight_512x3x3_stride_1  | ndarray-simd | `ndarray` | Cpu    | 12.093ms       |
| conv2d-input_3x224x224_weight_64x3x3_stride_1   | ndarray-simd | `ndarray` | Cpu    | 22.637ms       |
| conv2d-input_64x112x112_weight_128x3x3_stride_1 | ndarray-simd | `ndarray` | Cpu    | 46.648ms       |
| conv2d-input_64x56x56_weight_64x3x3_stride_1    | ndarray-simd | `ndarray` | Cpu    | 7.196ms        |
| conv2d-input_128x28x28_weight_128x3x3_stride_1  | ndarray-simd | `ndarray` | Cpu    | 5.179ms        |
| conv2d-input_256x14x14_weight_256x3x3_stride_1  | ndarray-simd | `ndarray` | Cpu    | 11.570ms       |
| conv2d-input_512x7x7_weight_512x3x3_stride_1    | ndarray-simd | `ndarray` | Cpu    | 23.494ms       |
| conv2d-input_96x224x224_weight_64x1x1_stride_1  | ndarray-simd | `ndarray` | Cpu    | 68.904ms       |

### DIFF
--- a/crates/burn-cubecl-fusion/src/elemwise/builder.rs
+++ b/crates/burn-cubecl-fusion/src/elemwise/builder.rs
@@ -27,7 +27,7 @@ impl<R: Runtime> ElementWiseBuilder<R> {
                 FuseSettings {
                     broadcast: true,
                     output_shape_updates: true,
-                    inplace: false,
+                    inplace: true,
                     vectorization: true,
                 },
             ),
--- a/crates/burn-cubecl-fusion/src/shared/kernel.rs
+++ b/crates/burn-cubecl-fusion/src/shared/kernel.rs
@@ -445,15 +445,12 @@ fn select_indices<C: Numeric>(
         //
         // Therefore the same indices are used to fetch multiple entries in the input tensor.
 
-        let write_pos_input = write_pos * line_size_ref;
-        let stride_input_line = global_stride(inputs, comptime![config.rank - 1], pos_input);
-
         if comptime![dim > 0] {
             let index_before = global_offset(
                 inputs,
                 outputs,
                 locals,
-                write_pos_input,
+                write_pos,
                 comment!(input.clone()),
                 comptime![Some((0u32, dim))],
                 config,
@@ -466,7 +463,7 @@ fn select_indices<C: Numeric>(
                 inputs,
                 outputs,
                 locals,
-                write_pos_input,
+                write_pos,
                 comment!(input.clone()),
                 comptime![Some((dim + 1, config.rank))],
                 config,
@@ -474,6 +471,8 @@ fn select_indices<C: Numeric>(
             index += index_after;
         }
 
+        let stride_input_line = global_stride(inputs, comptime![config.rank - 1], pos_input);
+        let write_pos_input = write_pos * line_size_ref;
         let coordinate_dim = write_pos_input / stride_dim_ref % shape_dim_ref;
         let offset_dim = read_input::<u32>(
             inputs,
@@ -485,7 +484,6 @@ fn select_indices<C: Numeric>(
             None,
         );
 
-        index *= line_size_ref;
         index += offset_dim[0] * stride_input_dim;
 
         #[unroll]
--- a/crates/burn-cubecl-fusion/src/shared/trace/base.rs
+++ b/crates/burn-cubecl-fusion/src/shared/trace/base.rs
@@ -77,8 +77,10 @@ impl FuseTrace {
         OutputPlanner::<R>::new(&self.inputs, &self.outputs, &self.views)
             .run::<BT>(client, device, context, &mut plan);
 
-        VectorizationPlanner::<R>::new(&self.views, &self.reads, &self.indexed)
-            .run(runner, context, &mut plan);
+        if self.settings.vectorization {
+            VectorizationPlanner::<R>::new(&self.views, &self.reads, &self.indexed)
+                .run(runner, context, &mut plan);
+        }
 
         match LaunchPlanExecutor::<R>::new(&self.scalars, &self.views, &self.ops)
             .execute::<_, BT>(client, runner, context, plan)
--- a/crates/burn-cubecl/src/kernel/pool/adaptive_avg_pool2d.rs
+++ b/crates/burn-cubecl/src/kernel/pool/adaptive_avg_pool2d.rs
@@ -1,60 +1,62 @@
-use crate::{element::CubeElement, ops::numeric::empty_device, tensor::CubeTensor, CubeRuntime};
+use crate::{
+    element::CubeElement,
+    kernel::conv::nchw_to_nhwc,
+    ops::{max_vectorization, numeric::empty_device, permute},
+    tensor::CubeTensor,
+    CubeRuntime,
+};
 use burn_tensor::Shape;
 use cubecl::{calculate_cube_count_elemwise, prelude::*};
 
 #[cube(launch)]
-fn adaptive_avg_pool2d_direct<E: Numeric>(input: &Tensor<E>, output: &mut Tensor<E>) {
-    let (output_stride_0, output_stride_1, output_stride_2, output_stride_3) = (
-        output.stride(0),
-        output.stride(1),
-        output.stride(2),
-        output.stride(3),
-    );
-    let (output_shape_0, output_shape_1, output_shape_2, output_shape_3) = (
-        output.shape(0),
-        output.shape(1),
-        output.shape(2),
-        output.shape(3),
-    );
-    let (input_stride_0, input_stride_1, input_stride_2, input_stride_3) = (
+fn adaptive_avg_pool2d_direct<E: Numeric>(input: &Tensor<Line<E>>, output: &mut Tensor<Line<E>>) {
+    if ABSOLUTE_POS >= output.len() {
+        terminate!();
+    }
+
+    let (out_h, out_w, channels) = (output.shape(1), output.shape(2), output.shape(3));
+    let channel_lines = channels / output.line_size();
+    let (in_stride_b, in_stride_h, in_stride_w, in_stride_c) = (
         input.stride(0),
         input.stride(1),
         input.stride(2),
         input.stride(3),
     );
-    let (input_shape_2, input_shape_3) = (input.shape(2), input.shape(3));
+    let (in_h, in_w) = (input.shape(1), input.shape(2));
 
-    let b = (ABSOLUTE_POS / output_stride_0) % output_shape_0;
-    let c = (ABSOLUTE_POS / output_stride_1) % output_shape_1;
-    let oh = (ABSOLUTE_POS / output_stride_2) % output_shape_2;
-    let ow = (ABSOLUTE_POS / output_stride_3) % output_shape_3;
+    let c = (ABSOLUTE_POS % channel_lines) * input.line_size();
+    let pos = ABSOLUTE_POS / channel_lines;
+    let ow = pos % out_w;
+    let pos = pos / out_w;
+    let oh = pos % out_h;
+    let b = pos / out_h;
 
-    let ih_start = start_index(oh, output_shape_2, input_shape_2);
-    let ih_end = end_index(oh, output_shape_2, input_shape_2);
+    let ih_start = start_index(oh, out_h, in_h);
+    let ih_end = end_index(oh, out_h, in_h);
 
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -131,12 +129,16 @@ half = { version = "2.4.1", features = [
...
+macerator = { git = "https://github.com/wingertge/macerator.git", rev = "5d124bbbf5c4262343f20e8d63c9071d35f2c064" }
+# macerator = { path = "../macerator" }
 matrixmultiply = { version = "0.3.9", default-features = false }
 ndarray = { version = "0.16.1", default-features = false }
 num-traits = { version = "0.2.19", default-features = false, features = [
     "libm",
 ] } # libm is for no_std
 openblas-src = "0.10.11"
+# pulp = { path = "../pulp/pulp" }
+pulp = { git = "https://github.com/wingertge/pulp.git", rev = "c550ba4eda98dcd964dc374fa4d508bfff9f03c8", default-features = false }
```
nathanielsimard, line 141: We'll have to wait until we there is a released version of those crates before merging. We can't release burn to crates.io when dependencies aren't released.

nathanielsimard (commented):
Everything looks fine, but I would add more documentation. Fast algorithmic code with a lot of unsafe is not easy to maintain. Defining the strategy for each algo, the cases to checks and why it's correct to use unsafe would be a huge help.

Cargo.toml
wingertge: yeah I'm working with the author of `pulp` to get my required changes merged and then I can publish to `crates.io`. Just wanted to get the review done so we can merge quickly later 👍

wingertge:
> Everything looks fine, but I would add more documentation. Fast algorithmic code with a lot of unsafe is not easy to maintain. 
> Defining the strategy for each algo, the cases to checks and why it's correct to use unsafe would be a huge help.

Unfortunately SIMD tends to involve unsafe a lot because the only way to ensure the compiler uses vectorized load/store is to use the intrinsics that operate on pointers. I will add comments that explain why it's safe and what conditions need to hold of course. A lot of the loads and stores are at least mostly bounds checked because they use the safe index function from ndarray. The only opportunity for unchecked OOB is from the fact that more than one element is loaded at once, and ndarray only checks that the first element is in bounds.

wingertge:
I added safety comments wherever needed, and refactored unary and binary a bit to simplify things. Let me know if anything else is unclear. The only real safety things now are in convolution, the rest goes from a checked index directly into a pointer function so it's only "unsafe" by technicality.

wingertge:
I decided to just add my own backend to `macerator` at least for the forseeable future, because there were a lot of ergonomics hangups with pulp anyways. The downside is it's a bit of extra maintenance (though most of it is auto-generated with macros), and we no longer have avx-512 support (but that was nightly only anwyays). The upside is I added WASM support (but it requires setting a RUSTLFAG since WASM doesn't support runtime feature detection), and the ergonomics are *much* nicer. Either way, this means I was able to publish the crate so we should be fine to merge as long as nothing else pops up. Tests all pass as before, `craft-burn` works fine, etc. So it's no longer blocked on the released crate requirement.

backend-comparison/benches/conv2d.rs
```diff
@@ -220,6 +220,7 @@ fn bench<B: Backend>(
     let mut results = Vec::new();
 
     for bench in benches {
+        println!("Running {}", bench.name());
```
laggui, line 223: Left over debug statement? Or useful given the amount of conv2d benches? 😅

crates/burn-ndarray/src/ops/module.rs
```diff
@@ -46,10 +50,14 @@ impl<E: FloatNdArrayElement, I: IntNdArrayElement, Q: QuantElement> ModuleOps<Se
...
-            x, weight, bias, options
-        )
-        .into())
+        module_op!(inp(x, weight), opt(bias), E, |x, weight, bias| {
+            #[cfg(feature = "simd")]
+            let (x, weight, bias) = match try_conv2d_simd(x, weight, bias, options.clone()) {
+                Ok(out) => return out.into(),
+                Err(args) => args,
+            };
+            conv2d::<E>(x, weight, bias, options).into()
```
laggui, line 59: Since the pattern is pretty similar for all simd ops, maybe we could have automatic fallbacks 🤔 

But totally fine for now, it's only 3 ops.

crates/burn-tensor/src/tensor/quantization/strategy.rs
```diff
@@ -519,8 +519,7 @@ mod tests {
     #[test]
     fn test_int8_symmetric_quantization_per_block_grid_3d() {
         let shape = &[2, 4, 4];
-        let x = vec![
-            // 2x2 blocks: [[-1.8, -1.0, 0.0, 0.5], [-0.8, 1.2, 0.25, 0.5]]
```
laggui, line 523: Probably didn't mean to delete the comment

backend-comparison/benches/conv2d.rs
wingertge: I think it's useful, especially on the slower backends where the full suite can easily take a few minutes.

crates/burn-tensor/src/tensor/quantization/strategy.rs
wingertge: Seems like the auto fix somehow deleted it. I restored it.

laggui:
I cleared the cache earlier today for the runners associated with your PR in case some useless artifacts had been carried over..

You can see the warning
```
Cache not found for keys: v11-rust-stable-linux-linux-std-tests-Linux-0d1eee6f-15f82ea3, v11-rust-stable-linux-linux-std-tests-Linux-0d1eee6f
```

And the disk space after rust was setup (21G avail instead of 12G in the previous failed job)
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        73G   [53](https://github.com/tracel-ai/burn/actions/runs/13795281381/job/38585332869?pr=2851#step:4:55)G   21G  72% /
tmpfs           7.9G  172K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda15      105M  6.1M   99M   6% /boot/efi
/dev/sdb1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```

Seems to have fixed the problem for now.. not sure how confident I am that it won't happen again 😅

laggui (approved):
Finally had time to poke around and check the differences in disk usage.

This should be good to go! With the added dependencies, the changes between this branch and main are just over 1GB.

main
```
12M     crates
23G     target
# find . -name '*.profraw' -print0 | du -ch --files0-from=- | grep total || echo "No .profraw files found"
2.7G    total
```

simd additions
```
11M     crates
22G     target
# find . -name '*.profraw' -print0 | du -ch --files0-from=- | grep total || echo "No .profraw files found"
2.7G    total
```

We're operating at the limit of the CI disk space but still fine. Pretty confident that deleting the old caches removed some artifacts from the other revisions not used anymore like the git versions of macerator and pulp (among others maybe). No disk space issue on the CI runs since.

I think we really need to restructure the workspace sooner than later.

wingertge:
Yeah there's unfortunately no way around extra disk space, the SIMD support requires adding a few dozen new ops to pulp anyways and that will unfortunately be duplicated for every SIMD version (Scalar, x86_64-v2, x86_64-v3 and eventually AVX-512) by virtue of how runtime detection works. But for an up to 10x speedup it's definitely a worthwhile tradeoff I think.

laggui:
> Yeah there's unfortunately no way around extra disk space, the SIMD support requires adding a few dozen new ops to pulp anyways and that will unfortunately be duplicated for every SIMD version (Scalar, x86_64-v2, x86_64-v3 and eventually AVX-512) by virtue of how runtime detection works. But for an up to 10x speedup it's definitely a worthwhile tradeoff I think.

Agreed! Was mostly to document my findings after making sure the changes would not break the CI once again 😅

PR #2854, opened 2025-03-02
Feat: Add tan trig function

Msa360:
### Checklist

### Related Issues/PRs
None

### Changes

Add the tangent operator in **FloatTensorOps** and **QTensorOps** and add it to the **tensor** api `tensor.tan()`.

### Testing

Tested using **PyTorch** and **Python** _tan_ operator as a reference. While adding the tests for tan(), I realized the file `crates/burn-import/src/onnx/dim_inference.rs` doesn't exists anymore, but it was still used in the docs file `contributor-book/src/guides/onnx-to-burn-conversion-tool.md`. Maybe it should be updated... Since I wasn't sure what to do I implemented the same tests as **cos** and **sin** because they shouldn't be very different.


laggui (approved):
LGTM! Thanks for also adding the ONNX import compatibility 🙂 

> While adding the tests for tan(), I realized the file crates/burn-import/src/onnx/dim_inference.rs doesn't exists anymore, but it was still used in the docs file contributor-book/src/guides/onnx-to-burn-conversion-tool.md. Maybe it should be updated...

Ahh good catch! Looks like this change was missed in PR #2831 when renaming to rank inference.

PR #2868, opened 2025-03-04
Add rank inference for tan

Msa360:
### Checklist

### Related Issues/PRs

#2854

### Changes

Tan doesn't change the shape, to the output shape is the same as input shape.

### Testing

No tests needed.


laggui (approved):
Whoops, missed that in the previous review 🤦

Thanks for the quick fix!

PR #2888, opened 2025-03-10
update cubecl

louisfd:


crates/burn-tensor/src/tensor/quantization/strategy.rs
```diff
@@ -519,8 +519,7 @@ mod tests {
     #[test]
     fn test_int8_symmetric_quantization_per_block_grid_3d() {
         let shape = &[2, 4, 4];
-        let x = vec![
-            // 2x2 blocks: [[-1.8, -1.0, 0.0, 0.5], [-0.8, 1.2, 0.25, 0.5]]
```
laggui, line 523: Probably didn't mean to remove the comment in the fix

crates/burn/Cargo.toml
```diff
@@ -124,3 +124,6 @@ burn-remote = { path = "../burn-remote", version = "0.17.0", default-features =
 burn-router = { path = "../burn-router", version = "0.17.0", default-features = false, optional = true }
 burn-tch = { path = "../burn-tch", version = "0.17.0", optional = true }
 burn-wgpu = { path = "../burn-wgpu", version = "0.17.0", optional = true, default-features = false }
+
+[patch.crates-io]
+cudarc = "0.13.9"
```
laggui, line 129: Why do we need to specify the same version as the one used in cubecl-cuda?

laggui (approved):
LGTM! Just some minor comments.

Btw audit issue was fixed in #2896

crates/burn/Cargo.toml
louisfd: Helps ensure all dependencies use cudarc 0.13.9 to avoid version conflicts.

laggui: Thought cargo dependency resolution would take care of that 😄 but I also thought the crates in `[patch]` could only be used to specify a different source (e.g., use a patched github rev instead of published crate) - not just a specific version.

You learn something new everyday!

Issue #2277, opened 2024-09-13
Improve learner multi-gpu ergonomics


laggui:
Especially w.r.t data loading, the data should automatically be transferred to the correct device.

Right now the batch is moved to the first GPU and users are moving the inputs to the correct device before inference (which works, but is less nice and might limit multi-gpu usage since the first GPU always gets filled first).

kingwingfly:
Hi, bro. How do you find the way to correctly work with multi-gpus on `burn`. 

> Right now the batch is moved to the first GPU and users are moving the inputs to the correct device

Could you please tell me how? 

I defined Learner with `learnerbuilder.devices(vec![Cuda(0), Cuda(1)])`, but init model on `Cuda(0)`.

- How to move mini_batch to correct GPU and in which part of the code to do this?
- The model is on GPU(0), but data may be on GPU(1). They are different devices, how can it work?

It's so kind of you if you could provide me any help!

laggui:
Take a look at the [text classification example](https://github.com/tracel-ai/burn/tree/main/examples/text-classification) 🙂 more specifically, how the inputs are [moved to the same device](https://github.com/tracel-ai/burn/blob/main/examples/text-classification/src/model.rs#L68-L71) as the model.

kingwingfly:
Thanks a looooot!  I managed to. The key point is that the model is forked onto each device and `model.devices()` will return a `Vec<Device>` whose first elem is the device the model currenctly on.

PR #2890, opened 2025-03-10
Improve multi-device data loading strategy

laggui:
### Checklist

### Related Issues/PRs

#2277 

### Changes

The data loader strategy currently requires the batcher implementation to load the data on a specific device. For a single GPU, this is fine. But for multi-gpu, the same device (in our examples, the first) is always used and the data then has to be moved to the same device as the model before the inference step.

This is not ideal since a single GPU might get overloaded when loading the data. The `to_device` implementation for our cubecl backends goes through a CPU buffer, which makes things even worse.

This PR changes the `Batcher` trait to provide the correct device for the batch

```rust
impl<B: Backend> Batcher<B, MyItem, MyBatch<B>> for MyBatcher {
    fn batch(&self, items: Vec<MyItem>, device: &B::Device) -> MyBatch<B> {
        // The correct `device` is already provided for the impl to use
    }
}
```

To enable the device selection for data loading, the `DataLoader` trait is now generic over a backend for which a device can be set.

```rust
pub trait DataLoader<B: Backend, O>: Send {
     // [...]

    /// Sets the device for the data loader, ensuring the batches are assigned to the correct device.
    fn set_device(&mut self, device: B::Device);
    /// Returns a new data loader containing a subset of the data.
    fn slice(&self, start: usize, end: usize) -> Box<dyn DataLoader<B, O>>;
}
```

Data loaders can also be sliced to return a subset of the data. For multi-device, this allows us to split a single data loader into multiple data loaders to get one data loader per worker (i.e., one per device), so the data is loaded on the correct device before being passed to the model.

To allow the `MultiThreadDataLoader` to be split, it now holds the required fields for the dataset. The data loaders are lazily initialized when `.iter()` is called.

### Testing

Unit tests. Don't have a multi-gpu machine at hand to check the improvements 😅 though changes are fairly straightforward.


examples/guide/src/training.rs
```diff
@@ -74,8 +74,8 @@ pub fn train<B: AutodiffBackend>(artifact_dir: &str, config: TrainingConfig, dev
 
     B::seed(config.seed);
 
-    let batcher_train = MnistBatcher::<B>::new(device.clone());
-    let batcher_valid = MnistBatcher::<B::InnerBackend>::new(device.clone());
+    let batcher_train = MnistBatcher::default();
+    let batcher_valid = MnistBatcher::default();
```
nathanielsimard, line 78: It could be the same batcher now.

laggui: Yupp! Even simpler 😄

crates/burn-core/src/data/dataloader/lazy.rs
```diff
@@ -0,0 +1,354 @@
...
+    ///
+    /// # Returns
+    /// A vector of lazy data loaders, each initialized with one of the provided distribution strategies.
+    /// Each returned data loader represents a portion of the original dataset and can be independently
+    /// processed. The number of returned data loaders matches the number of distribution strategies provided.
+    fn split(
+        &self,
+        distributors: Vec<Box<dyn DistributionStrategy<Resource = Self::Resource>>>,
+    ) -> Vec<Box<dyn LazyDataLoader<O, Resource = Self::Resource>>>;
+}
```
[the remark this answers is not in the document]
laggui: Hmmm not sure how this would help. This trait is used to assign different dispatchers to each data loader split. The dispatcher is linked to the data loader, not dataset.

Essentially this trait was added to split a data loader into multiple data loaders, each with a partial view of the dataset (so `PartialDataset` is actually used in the implementation) and a specific dispatcher.

So for multi-GPU training, the original data loader is created for the entire dataset. But when multiple devices are specified, the data loader is split into `num_devices` partial data loaders that dispatch a batch for a specific device. That way, each worker is responsible for a single data loader and device.

The trait could be removed, but it makes it more difficult since we're always working with a `dyn DataLoader` that is created by the user for the learner.

Lmk what you think.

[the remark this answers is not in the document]
laggui: This has been further simplified. Data loaders are now generic over the backend to allow device selection/assignment and can be sliced.

crates/burn-core/src/data/dataloader/builder.rs
```diff
@@ -91,27 +112,137 @@ where
     /// # Returns
     ///
     /// The data loader.
-    pub fn build<D>(self, dataset: D) -> Arc<dyn DataLoader<O>>
+    pub fn build<D>(self, dataset: D) -> Arc<dyn DataLoader<B, O>>
```
nathanielsimard, line 115: If we put the `DataLoader` trait inside an Arc, we should remove the `DynDataLoader` trait.

crates/burn-core/src/data/dataloader/multithread.rs
```diff
@@ -26,32 +46,94 @@ struct MultiThreadsDataloaderIterator<O> {
...
+            batcher,
+            num_threads,
+            device,
+            rng,
+            dataloaders: OnceCell::new(),
+        }
+    }
+
+    /// Force initialization if needed.
+    fn initialize(&self) -> &[Box<dyn DynDataLoader<B, O>>] {
```
nathanielsimard, line 89: Could we use `Arc<dyn DataLoader<B, O>` here instead?

crates/burn-train/src/learner/train_val.rs
```diff
@@ -144,14 +144,40 @@ impl<LC: LearnerComponents> Learner<LC> {
...
+                };
+                let mut dataloader = dataloader_train.slice(start, end);
+                dataloader.set_device(self.devices[i].clone());
+                dataloaders.push(Arc::from(dataloader));
+                start = end;
+            }
+            dataloaders
+        } else {
+            vec![dataloader_train]
+        };
```
nathanielsimard, line 171: I would extract that into a new function, maybe called: `split_dataloaders(dataloader, &devices)`

crates/burn-train/src/learner/train_val.rs
```diff
@@ -175,6 +201,7 @@ impl<LC: LearnerComponents> Learner<LC> {
                 break;
             }
 
+            // TODO: multi-device validation?
```
nathanielsimard, line 204: Still relavant?

laggui: Forgot that I added this comment 😅 was mostly for my own sake

The comment is still relevant I guess, cuz the validation loop still runs on a single device (same behavior as before). We just split the training across GPUs.

For most data parallel use cases this is fine I think, unless the validation set is huge splitting across multiple GPUs probably doesn't have as much of a benefit. But we could do it in the future.

crates/burn-core/src/data/dataloader/multithread.rs
laggui: Yeah probably, but in the implementation we share each data loader to a thread.

With a box and `clone_dyn()` this didn't require any further requirements for `Sync`. So if we switch to `Arc<dyn DataLoader<B, O>>` we probably need to have another bound (and propagate it).

crates/burn-core/src/data/dataloader/builder.rs
laggui: Yeah that makes sense. But I'm curious why this wasn't already the case?

crates/burn-core/src/data/dataloader/multithread.rs
laggui: Updated:

`MultiThreadDataLoader` simply keeps a vec of `BatchDataLoader`. The `Box<dyn DataLoader>` was superfluous, left over from the previous impl. `BatchDataLoader` is always used to create an instance.

```rust
pub struct MultiThreadDataLoader<B: Backend, I, O> {
    // Configuration parameters needed for initialization
    strategy: Box<dyn BatchStrategy<I>>,
    dataset: Arc<dyn Dataset<I>>,
    batcher: Box<dyn DynBatcher<B, I, O>>,
    device: B::Device,
    rng: Option<rand::rngs::StdRng>,
    num_threads: usize,

    // The lazily initialized data loaders
    dataloaders: OnceCell<Vec<BatchDataLoader<B, I, O>>>,
}
```

so `initialize()` now returns a slice of `BatchDataLoader`s, and we can remove the `DynDataLoader` trait with `clone_dyn()`.

PR #2923, opened 2025-03-18
Deterministic rng in tests

chlobes:
### Checklist

### Problem

1. Normal distributions are unbounded, so there is a small chance the variables or their properties will be outside the test range.
2. The RNG is a global static, this introduces a race condition when tests are run in parallel that allows seeded tests to be non-deterministic

Example: `thread 'nn::initializer::tests::initializer_normal_init' panicked at crates/burn-core/src/nn/initializer.rs:258:9:
Expected variance to be between 1.0 += 0.1, but got 1.1060424`

### Solution

If tests rely on specific rng, they should be tagged `#[serial]`.
If tests do not rely on specific rng (uniform or constant distribution), they should not be seeded.

### Testing

Ran the affected tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -666,6 +666,7 @@ dependencies = [
  "rmp-serde",
  "serde",
  "serde_json",
+ "serial_test",
  "spin",
  "thiserror 2.0.11",
  "uuid",
--- a/crates/burn-core/Cargo.toml
+++ b/crates/burn-core/Cargo.toml
@@ -121,6 +121,7 @@ burn-autodiff = { path = "../burn-autodiff", version = "0.17.0" }
 burn-dataset = { path = "../burn-dataset", version = "0.17.0", features = [
     "fake",
 ] }
+serial_test.workspace = true
 
 [package.metadata.docs.rs]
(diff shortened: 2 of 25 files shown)


chlobes:
I took a look and it seems like making serial no_std would be difficult. So just disabled the attribute in no_std.

laggui:
> 2. The RNG is a global static, this introduces a race condition when tests are run in parallel that allows seeded tests to be non-deterministic
> 
> Example: `thread 'nn::initializer::tests::initializer_normal_init' panicked at crates/burn-core/src/nn/initializer.rs:258:9: Expected variance to be between 1.0 += 0.1, but got 1.1060424`
> 

Hmm you raise a good point. Don't think I would force the tests to run sequentially though, they should not be dependent of other tests. In other words, I don't think the tests are the root issue, but rather the static RNG.

Instead, we might want to have a thread-local RNG, e.g.

```rust
thread_local! {
    pub(crate) static SEED: RefCell<Option<StdRng>> = RefCell::new(None);
}
```

CC'ing @nathanielsimard for his thoughts

PR #2931, opened 2025-03-20
Update to edition 2024

laggui:
### Checklist

### Changes

Upgraded to edition 2024 and MSRV 1.85

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ env:
   # Note: It is not possible to define env vars in composite actions.
   # To work around this issue we use inputs and define all the env vars here.
 
-  RUST_PREVIOUS_VERSION: 1.84.0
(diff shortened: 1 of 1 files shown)


crates/burn-fusion/src/client/mutex.rs
```diff
@@ -84,8 +84,10 @@ where
     where
         B: FusionBackend<FusionRuntime = R>,
     {
-        let mut server = self.server.lock();
-        server.read_float::<B>(tensor, stream)
+        // Clone the Arc to extend its lifetime
+        let server = self.server.clone();
+
+        async move { server.lock().read_float::<B>(tensor, stream).await }
```
laggui, line 90: There's a lot of noise due to the cargo fmt changes, but this is the most significant change across all files.

Lifetime elision rules are more strict.

reneleonhardt:
Wow, awesome work @laggui! 🚀

PR #2935, opened 2025-03-20
Added graph node pruning

Knight-Ops:
### Checklist

- [ ] Confirmed that `run-checks all` script has been executed. (Fails in Burn-Tensor)
- [ ] Made sure the book is up to date with changes in this PR. (N/A)

### Related Issues/PRs

N/A

### Changes

Currently certain ONNX files have extraneous empty output strings when parsed using burn-import. This causes issues with is_top_sorted because the empty strings match on the inputs and outputs even though they aren't actually connected in the graph.

This also causes warnings when attempting to guess the inputs for various node types since the input doesn't have a type associated with it in the graph.

An example ONNX file that triggers this issue : https://huggingface.co/HuggingFaceTB/SmolLM2-360M-Instruct/blob/main/onnx/model.onnx

The fix here is to go through the entire graph once it is loaded from the `onnx` file and prune every node's inputs and outputs `Vec<String>` to remove empty strings. Added bonus here is that it should improve the performance of topology checks, as well as any other iterating over all inputs or outputs since we can remove useless entries from the tree.

Example of empty output:
image

Example of empty input:
image1

Both of these types are in Microsoft's domain, so perhaps this isn't a true ONNX issue, but this pruning shouldn't affect any other implementations that are fully populated.

### Testing

Tested by reloading the problematic ONNX file which now fails on an unimplemented ONNX Op, progress.

### DIFF
--- a/crates/onnx-ir/src/from_onnx.rs
+++ b/crates/onnx-ir/src/from_onnx.rs
@@ -353,24 +353,6 @@ pub fn parse_onnx(onnx_path: &Path) -> OnnxGraph {
     let mut onnx_model: ModelProto =
         Message::parse_from_reader(&mut file).expect("Unable to parse ONNX file");
 
-    // Prune the node inputs and outputs of all empty strings. This can cause issues with
-    // checking the topology of the graph, as well as cause issues when guessing input and output
-    // types. Empty strings seem to be parsed from the ONNX file when an input or output is not populated.
-    // As an added bonus, this will also improve performance on model conversion.
-    onnx_model
-        .graph
-        .as_mut()
-        .expect("Graph not found")
-        .node
-        .iter_mut()
-        .for_each(|node| {
-            // Clear the input of empty string entries
-            node.input.retain(|s| !s.is_empty());
-
-            // Clear the output of empty string entries
-            node.output.retain(|s| !s.is_empty());
-        });
-
     // ONNX nodes must be topologically sorted per spec:
     // https://github.com/onnx/onnx/blob/main/docs/IR.md#graphs
     debug_assert!(
@@ -445,6 +427,11 @@ impl TopologicalSortable for Vec<NodeProto> {
         for node in self {
             // Iterate over each output of the node
             for output in &node.output {
+                // If the output is empty, we don't want to check the rest of the graph, inputs and outputs that are optional
+                // can end up as empty strings, so we can't use that as a reason to count the graph as not sorted
+                if output.is_empty() {
+                    continue;
+                }
                 // Iterate over each other node in the vector
                 for other_node in self {
                     // If the other node has an input that matches the current output
--- a/crates/onnx-ir/src/from_onnx.rs
+++ b/crates/onnx-ir/src/from_onnx.rs
@@ -343,14 +343,33 @@ impl OnnxGraphBuilder {
 ///
 /// * If the file cannot be opened
 /// * If the file cannot be parsed
+/// * If the graph is missing from the parsed ModelProto
 /// * If the nodes are not topologically sorted
 pub fn parse_onnx(onnx_path: &Path) -> OnnxGraph {
     log::info!("Parsing ONNX file: {}", onnx_path.display());
 
     // Open the file
     let mut file = File::open(onnx_path).expect("Unable to open file");
-    let onnx_model: ModelProto =
+    let mut onnx_model: ModelProto =
         Message::parse_from_reader(&mut file).expect("Unable to parse ONNX file");
+
+    // Prune the node inputs and outputs of all empty strings. This can cause issues with
+    // checking the topology of the graph, as well as cause issues when guessing input and output
+    // types. Empty strings seem to be parsed from the ONNX file when an input or output is not populated.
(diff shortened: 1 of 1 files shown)


Knight-Ops:
When investigating an issue, I have realized that we can't just nuke these empty strings since many of the inputs for various Ops are getting inputs by the offset. It may be a more major investment, but it could make more sense to map each node's inputs into a hashmap or something where we can put Options around inputs to both keep information on them, as well as not cause issues when we are doing the topo checking.

Knight-Ops:
I reverted back to my original Topo empty string skipping approach, fixes the same problem, just seems a bit more hacky.

laggui (commented):
Is this actually valid in ONNX?

The [spec states](<https://github.com/onnx/onnx/blob/main/docs/IR.md#nodes>) for node inputs and outputs (strings):

> It must refer to either a graph input, a graph initializer or a node output.

In this specific scenario, there are not *official* ONNX ops, but rather contributed ops for ONNX Runtime.

See [GroupQueryAttention](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.GroupQueryAttention) for example. According to the node definition, these inputs are indeed optional, and thus should simply not be provided. But in this case they are instantiated with empty strings names 😅 So it looks like incorrect usage of optional inputs..

(btw, even if you fix these issues the model still won't be able to be imported because of these externally defined operators which are not from the official ONNX domain).

Knight-Ops:
Great question, Microsoft has been known to take some liberties with specifications and what the requirements are. Yeah very true it won't load, but it also won't fail to load with "Graph nodes are not topologically sorted".

It does bring up a question on optional inputs, if they **ARE NOT** empty strings, how do you tell which optional argument you have? I don't know in the base ONNX spec if there are any that have multiple optional arguments.

Knight-Ops:
Here is an example from ONNX: https://onnx.ai/onnx/operators/onnx__Dropout.html. If I have a Arg vector of length 2, how do I tell if I have `ratio` or `training_mode` when we are deserializing a into a `pub input: ::std::vec::Vec<::std::string::String>,`? In this case you could tell from the Tensor type since they don't overlap, but I don't know that non-overlapping input types is a reliable option.

laggui:
> It does bring up a question on optional inputs, if they ARE NOT empty strings, how do you tell which optional argument you have? I don't know in the base ONNX spec if there are any that have multiple optional arguments.

Well, right now we only check the inputs by index (e.g., `node.inputs[0]`) but this actually contains an argument that we can check for the name to match the correct input.

https://github.com/tracel-ai/burn/blob/5674ff8ae904d33a41eb615f7659e475208fa4cf/crates/onnx-ir/src/ir.rs#L11-L26

This comes from parsing the protobuf definition. So for these ops we need to make more robust checks.

Knight-Ops:
I *believe* that `name` is not the name of the argument from the ONNX spec, but instead is the name of the edge on the graph. So it doesn't actually contain any context at all.

laggui:
> I _believe_ that `name` is not the name of the argument from the ONNX spec, but instead is the name of the edge on the graph. So it doesn't actually contain any context at all.

Oh, you're absolutely right! Been context switching too much between the PRs and my work lol.

Guess you found another flaw in the current graph parsing implementation

https://github.com/tracel-ai/burn/blob/09e9b0063b466d725fcda11292decd9aa5ecc2a1/crates/onnx-ir/src/ir.rs#L225

The inputs are captured as a `Vec<Argument>` during conversion

https://github.com/tracel-ai/burn/blob/main/crates/onnx-ir/src/proto_conversion.rs#L189

But in reality, we need to keep a reference to the input names. They're part of the protobuf and spec for a reason. But we just ignore them 😓

Knight-Ops:
Looking into the protobuf spec and the deserialization, I don't think the names exist in the specification at all. As such, I don't think there is another way to determine which input you are getting unless you explicitly fill each input with **something**. Based on the data that I have, it is just empty strings for Optional, but non-provided arguments. I am not sure how else you could track it with the info we have.

laggui:
Ok so apparently there is more than one way to represent optional inputs! I encountered trailing optional inputs, which can easily be omitted and validated. I also remembered dynamic optionals. But I just scrolled through [the spec](<https://github.com/onnx/onnx/blob/main/docs/IR.md#optional-inputs-and-outputs>) and looks like the empty string name is actually a valid way to represent static inputs

> There are two ways to leave an optional input or output unspecified: the first, available only for trailing inputs and outputs, is to simply not provide that input or output; the second method is to use an empty string in place of an input or output name.

In this case, we can effectively ignore empty strings as applied in your PR.

Just make sure to fix the clippy error 🙂

Knight-Ops:
I will do you one better, fix clippy and remove my extra comment that is no longer true. Thanks for all the feedback! Yeah this spec is a bear. Seems like some questionable decisions in some places to me, but what do I know!

laggui (approved):

laggui:
Canceled the macos job.. looks like it's hanging on some tests (possibly due to the new persistent autotune cache lock?) but totally unrelated to this PR.

PR #2936, opened 2025-03-20
Added output tensor size inference/copy

Knight-Ops:
### Checklist

- [ ] Confirmed that `run-checks all` script has been executed. (Fails in burn-tensor)
- [ ] Made sure the book is up to date with changes in this PR. (N/A)

### Related Issues/PRs

#2441

### Changes

Pretty minor changes here, according to the ONNX specification [RandomNormalLike](https://onnx.ai/onnx/operators/onnx__RandomNormalLike.html) and [RandomUniformLike](https://onnx.ai/onnx/operators/onnx__RandomUniformLike.html) the output shape is copied from the input tensor. I just added a helper function to just copy the parameters from the input TensorType and renaming it to the output.

### Testing

`cargo test` for all `burn-import` tests.

### DIFF
--- a/crates/burn-import/src/burn/node/random_normal_like.rs
+++ b/crates/burn-import/src/burn/node/random_normal_like.rs
@@ -30,9 +30,9 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for RandomNormalLikeNode {
         vec![Type::Tensor(self.output.clone())]
     }
 
-    fn forward(&self, _scope: &mut Scope, _node_position: usize) -> TokenStream {
+    fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
         let output = &self.output.name;
-        let input = &self.input.name;
+        let input = scope.tensor_use_owned(&self.input, node_position);
         let dist = self.get_distribution();
         quote! {
             let #output = #input.random_like(#dist);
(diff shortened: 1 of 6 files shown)


Knight-Ops:
This fixes one problem, but doesn't fix another. This fixes the issue where the output is incorrect for the Random*Like operator, but it cannot fix the issue where other Ops already have inputs set. I am looking a way that I could do this with the current setup, seems like we would need to do graph traversal, and fix all of the output edges.

Knight-Ops:
After staring at this some more I think I found the proper place to do the Node rewrite, this means I can revert the previous changes and just do the copied size at the Node itself by adding a custom handler for these functions. I am sure there are more Onnx functions that need this same treatment, so this seems like a good central place to do it. Let me know if there are better options.

Knight-Ops:
I guess this really should have been a draft PR since this is a bit of a disaster, but I think it finally is where it needs to be. The issue was an edge case during `dim_inference`/`rank_inference` where the RandomNormalLike Op had a dim/rank defined, but had a Shape of None.  Which can happen out of pytorch conversions. The fix is simply copying the input `TensorType` if there is no shape defined. This means that if a shape IS defined on the input that takes priority, otherwise we just copy the input.

laggui (approved):

laggui:
>The issue was an edge case during `dim_inference`/`rank_inference` where the RandomNormalLike Op had a dim/rank defined, but had a Shape of None.

Yeah we can't assume that the shape is always going to be present.. (as discussed on discord 😅)

Thanks for the fix!

PR #2938, opened 2025-03-20
Support multiple outputs being tracked in BurnGraph during ONNX conversion

Knight-Ops:
### Checklist

### Related Issues/PRs

N/A

### Changes

Currently outputs are not registered with the scope of the `BurnGraph` this is fine if the output is the last layer, but if you have multiple outputs throughout the model, they aren't tracked for cloning. By adding this registration, we fix the issue of outputs not being tracked.

### Testing

Ran through all `cargo test`s in `burn-import`


laggui (commented):
Ahhh good catch! Do you have an MWE graph that could be used to test this? Would be nice to capture this in a test.

Knight-Ops:
Yeah I will get a test added!

laggui:
Awesome, thanks! Will wait before merging

Knight-Ops:
@laggui I added a quick test for it.

laggui (approved):
🙏 🙏

Issue #2930, opened 2025-03-20
ONNX burn-import: Can't transform Shape(4) to tensor


ungerpeter:
I tried to transform an onnx model to burn code using burn-import crate, but the build script fails with the following error:

`ERROR burn_import::logger: PANIC => panicked at /Users/axpun/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-import-0.16.0/src/onnx/to_burn.rs:1334:18:
  Can't transform Shape(4) to tensor.`

This is the onnx model I tried to import:
[https://huggingface.co/Soykaf/RTMW3D-x/tree/main/onnx](https://huggingface.co/Soykaf/RTMW3D-x/tree/main/onnx)

According to the supported onnx operations table, shape should be supported:
[https://github.com/tracel-ai/burn/blob/main/crates/burn-import/SUPPORTED-ONNX-OPS.md](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/SUPPORTED-ONNX-OPS.md)

**To Reproduce**
1. Download onnx model from HF via link above
2. Follow the burn guide for onnx model import ([https://burn.dev/burn-book/import/onnx-model.html](https://burn.dev/burn-book/import/onnx-model.html))
3. run cargo build

Knight-Ops:
Good news : I just put in a PR to fix this issue specifically. Shape should not successfully be processed.

Bad news : The model you are referencing has a few unsupported Onnx Ops : `Unsupported ops: [Abs, DepthToSpace, Abs, Abs, Split, Split]`

ungerpeter:
> Good news : I just put in a PR to fix this issue specifically. Shape should not successfully be processed.
> 
> Bad news : The model you are referencing has a few unsupported Onnx Ops : `Unsupported ops: [Abs, DepthToSpace, Abs, Abs, Split, Split]`

Thank you!

AdrianEddy:
I have the same issue with a different model, and looks like that PR was closed. I think this issue should be opened again? 
I'm trying this model: https://huggingface.co/TheEeeeLin/HivisionIDPhotos_matting/blob/main/retinaface-resnet50.onnx

antimora:
Reopening it per AdrianEddy request.

polarathene:
## Reproduction

For reference, using the current git version of Burn (_has some improvements since the latest release tag [`v0.17.1`](https://github.com/tracel-ai/burn/releases/tag/v0.17.1)_).

[...]

PR #2939, opened 2025-03-21
Fix Shape, Reshape, and invalid rust variable names.

Knight-Ops:
### Checklist

### Related Issues/PRs

#2930

#### Shape 

- **shape_update_outputs** : Changed the output type to a Tensor as per the Onnx specification. It always outputs a `Tensor`, not a `Shape`.

#### Reshape

- Added improved error reporting. Its really hard to tell where errors are coming from when doing the build, so I made sure everything had `Reshape:` for some extra information
- **reshape_config** : Removed check on input length and input value. The input.value is None, because the input is a Tensor, you can't use value you have to check the Tensor.shape()
- **shape_update_outputs** : Fixed output Tensor elem_type and ensured that `shape` was a valid Vec, this is again done by not trying to get the `value`, but instead looking at the `shape` on the Tensor.

#### format_name
- Simple change, just replace characters that are invalid Rust variable names with underscores. This could possibly be done with a regex, but I don't know that it is entirely necessary right now. I just replaced the characters I ran into issues with while testing.

### Testing

Run `cargo test` on `burn-import` crate with all passing. Also using file referenced in #2930 as a test file.

### DIFF
--- a/crates/burn-import/src/burn/ty.rs
+++ b/crates/burn-import/src/burn/ty.rs
@@ -69,7 +69,9 @@ impl Type {
         if name_is_number {
             format!("_{}", name)
         } else {
-            name.to_string()
+            // Format the name by replacing invalid rust variable names, there is probably more to do here
+            // but these are the ones I ran into.
+            name.replace(".", "_").replace(":", "_")
         }
     }
     pub fn name(&self) -> &Ident {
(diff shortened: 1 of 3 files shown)


Knight-Ops:
Not sure what really to add here, I added back in some of the old code as a first approach specific on `Reshape` the model referenced in #2930 didn't like how `Reshape` was implemented (Value never existed).

My best guess here is that the original code worked for Constant shapes, where-as in theory you can have a Tensor calculated at runtime based on my understanding. We should check first for constants and then fall back to dynamic tensors.

laggui (commented):
Just a comment regarding the `ArgType::Shape`.

This is part of the intermediate representation between ONNX <> Burn. In burn shapes are not actually tensors, but arrays. I believe this was the initial motivation behind the variants for `ArgType`.

ONNX uses tensor types for pretty much all values, but the reality is a bit more complex. Even scalars are simply represented as 0-dim tensors.

So the output of shape should remain `ArgType::Shape` to maintain compatibility with the current codegen.

I haven't looked at the underlying cause in the linked issue, but from the error it looks like there is a node that should handle the `ArgType::Shape`.

PR #2944, opened 2025-03-22
Check channels_in matches in convolution layers

chlobes:

When channels_in to a convolution call mismatches the weights, backends break in various confusing ways.
This check avoids that with an early panic and informative error message.
I also added unit tests to ensure this check does panic.

### DIFF
--- a/crates/burn-core/src/nn/conv/conv1d.rs
+++ b/crates/burn-core/src/nn/conv/conv1d.rs
@@ -137,7 +137,13 @@ impl<B: Backend> Conv1d<B> {
     /// - input: `[batch_size, channels_in, length_in]`
     /// - output: `[batch_size, channels_out, length_out]`
     pub fn forward(&self, input: Tensor<B, 3>) -> Tensor<B, 3> {
-        let [_batch_size, _channels, length] = input.dims();
+        let [_batch_size, channels_in, length] = input.dims();
+        let expected = self.weight.dims()[1] * self.groups;
+        assert_eq!(
+            channels_in, expected,
+            "This conv layer requies a channels_in dimension of {expected}, but got {channels_in}"
+        );
+
         let padding = self
             .padding
             .calculate_padding_1d(length, self.kernel_size, self.stride);
@@ -199,5 +205,15 @@ mod tests {
             alloc::format!("{}", conv),
             "Conv1d {stride: 1, kernel_size: 5, dilation: 1, groups: 1, padding: Valid, params: 130}"
         );
+    }
+
+    #[test]
+    #[should_panic = "This conv layer requies a channels_in dimension of 5, but got 3"]
+    fn input_channels_mismatch() {
+        let config = Conv1dConfig::new(5, 5, 5);
+        let conv = config.init::<TestBackend>(&Default::default());
+
+        let input = Tensor::<TestBackend, 3>::zeros([1, 3, 10], &Default::default());
+        let _ = conv.forward(input);
     }
 }
--- a/crates/burn-core/src/nn/conv/conv1d.rs
+++ b/crates/burn-core/src/nn/conv/conv1d.rs
@@ -202,12 +202,12 @@ mod tests {
     }
 
     #[test]
-    #[should_panic = "Number of channels in input tensor and input channels of convolution must be equal. got: 3, expected: 5"]
+    #[should_panic = "Number of channels in input tensor and input channels of convolution must be equal. got: 4, expected: 5"]
     fn input_channels_mismatch() {
-        let config = Conv1dConfig::new(5, 5, 5);
+        let config = Conv1dConfig::new(5, 3, 3);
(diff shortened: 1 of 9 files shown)


crates/burn-core/src/nn/conv/conv1d.rs
```diff
@@ -137,7 +137,13 @@ impl<B: Backend> Conv1d<B> {
     /// - input: `[batch_size, channels_in, length_in]`
     /// - output: `[batch_size, channels_out, length_out]`
     pub fn forward(&self, input: Tensor<B, 3>) -> Tensor<B, 3> {
-        let [_batch_size, _channels, length] = input.dims();
+        let [_batch_size, channels_in, length] = input.dims();
+        let expected = self.weight.dims()[1] * self.groups;
+        assert_eq!(
+            channels_in, expected,
+            "This conv layer requies a channels_in dimension of {expected}, but got {channels_in}"
```
laggui, line 144: Typo: `requies` -> `requires`.

I would also simplify the message to: `"Expected input to have {expected} channels, but got {channels_in} instead"`

laggui (commented):
Thanks for the added validation!

If we are going to perform a check, I think it would be at the function level in `burn-tensor/src/tensor/modules.rs` 🤔

Also, what do you think about a debug assertion for such runtime checks? During development these can be enabled, and then disabled in release mode.

chlobes:
If we're doing the checks at the function level, then it seems sensible to use `check!`, I'll take a look at writing TensorCheck method(s) for that.

I'm not sure about runtime vs debug checks, I often build in release while debugging because debug gives me eg OOM or stack overflow errors.
 I think whatever solution is used could be put inside the `check!` macro so it gets applied uniformly.
 Could just feature flag the whole thing, `[cfg(not(feature = "elide_checks"))]` so that turning on `elide-checks`, would compile out all the checks.
 All this is obviously out of scope of course.

crates/burn-tensor/src/tensor/api/check.rs
```diff
@@ -1194,6 +1195,62 @@ impl TensorCheck {
...
+        let expected = weight[1] * options.groups;
+        if channels != expected {
+            check = check.register(
+                ops,
+                TensorError::new("Number of channels in input tensor and input channels of convolution must be equal.")
+                .details(format!("got: {channels}, expected: {expected}")),
+            );
+        }
+        check
+    }
```
laggui, line 1216: The checks are pretty much the same for all convolution types. I think we can keep a single check then, and use `TensorCheck::conv(...)` with different `ops` value.

```rust
pub fn conv(
    ops: &str,
    input_shape: Shape,
    weight_shape: Shape,
    groups: usize,
) -> Self {
    let mut check = TensorCheck::Ok;
    let input_channels = input_shape.dims[1];
    let expected = weight_shape.dims[1] * groups;
    if channels != expected {
        check = check.register(
            ops,
            TensorError::new("Number of channels in input tensor must match the convolution input channels.")
            .details(format!("Got {channels} channels, expected {expected}")),
        );
    }
    check
}
```

Btw transpose conv options also have `groups`.

laggui (changes requested):
Sounds good!

We can simplify this a bit, then should be good to go 🙂

crates/burn-tensor/src/tensor/api/check.rs
chlobes: They do have groups, but (I think?) those groups are not relevant to channels_in size

chlobes: Reasoning why I used `[usize; D]` over `Shape` is to make index out of bounds errors statically impossible.
In fact now I think about it, to really ensure that we should have a comptime assertion.

Reasoning for having seperate methods for each conv type was so that the conv function itself doesn't need to "know", but that's admittedly flismy reasoning and I'm happy to reduce codesize 🙂

chlobes:
Is this good?

crates/burn-tensor/src/tensor/api/check.rs
laggui: > They do have groups, but (I think?) those groups are not relevant to channels_in size

Ahh you're right, the groups affect the channels_out in conv transpose. Forgot about that for a moment

laggui: > Reasoning why I used [usize; D] over Shape is to make index out of bounds errors statically impossible.

The checks are all used within conv functions that have fixed const rank `D`. So if the tensors are invalid (weight or input) it would already fail to compile at that level.

I understand the motivation though!

laggui (approved):
> Is this good?

Based on our discussion it made more sense to split the conv vs transpose conv checks made sense. Made the changes myself since you had already split it initially 😅 Also removed an accidentally committed file.

LGTM now! But lmk if you have any other comments, will wait before merging.

chlobes:
Thanks for your help, LGTM!

PR #2951, opened 2025-03-24
Improve documentation page for "Tensor"

PtiLuky:
### Checklist

### Changes

* fixed some mistakes in the documentation (such as sometimes stating "element wise" when it should have been "with a scalar"). Or example code with wrong values shown (for tensor.min_dim_with_indices)
* changed some examples to make them more explicit about the function behavior.
* Developed a bit some function description when text was quite vague.

### Things to discuss

This is my first PR, I am new to Burn crate and struggled to understand some functions for which I needed to check directly the code. The goal of this PR is to avoid this for next newcomers!  
Also if this PR shall be split in several smaller PRs, please tell me. I didn't feel like I should have made one PR per function documented...

More general remarks about the documentation:

* There is no indications for contributing to documentation in the contribution guide, this is a bit missing, specially for the 2 following points
* The verbs are not consistent, sometimes functions are described with "*Does* this and that" and sometimes it's "Do this and that", is there any recommendation here? If so making it explicit in the guidelines would be great!
* The usage of the vocabulary `rank`, `dimension`, `dimensions`, `size` and `shape` is not always consistent, quite often I found "_tensor of dimensions [x, y, z]_", the idea is totally understandable here, but in this case the right word is "_tensor of shape [x, y, z]_", no? I am asking just to know if I should try to be accurate in the vocab for next PR (maybe even replace existing misuses) or if it's not important and should not be changed?

### Testing

Visual testing of the generated doc.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1583,9 +1583,9 @@ where
     ///
     /// For each dimension:
     /// - If current size is `1` and the requested size is `n` > 1, the tensor will be extended by
-    /// repeating itself in this dimension.
+    ///   repeating itself in this dimension.
     /// - If the current size is `n` > 1, then output size must be `n`. `-1` can be used to inferred this
-    /// automatically.
+    ///   automatically.
     /// - Otherwise tensor cannot be broadcasted.
     ///
     /// # Arguments
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -533,7 +533,7 @@ where
     /// # Panics
     ///
     /// If the size in the squeezed dimension is not 1.
-    ///  
+    ///
     /// # Returns
     ///
     /// A new `Tensor<B, D2, K>` instance with the specified dimension removed.
@@ -670,7 +670,7 @@ where
     /// # Panics
     ///
     /// If `D2` is lower than the actual number of dimensions.
-    ///  
+    ///
     /// # Returns
     ///
     /// A new `Tensor<B, D2, K>` instance with the specified dimensions added.
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -215,7 +215,7 @@ where
         Tensor::new(K::reshape(self.primitive, shape))
     }
 
-    /// Transpose the tensor. Permute the 2 last dimensions of the tensor.
+    /// Transpose the tensor.
     ///
     /// For a 2D tensor, this is the standard matrix transpose. For `D > 2`, the transpose is
     /// applied on the last two dimensions. For example, the transpose of a tensor with shape
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -173,7 +173,7 @@ where
         self.primitive.shape()
     }
 
-    /// Reshape the tensor to have the given shape.
+    /// Reshape the tensor to have the given shape, keeping the same values in the same order.
     ///
     /// A `-1` in the shape is used to infer the remaining dimensions, e.g.: `[2, -1]`
     /// will reshape the tensor with [2, 3, 4] dimensions to [2, 12].
@@ -213,7 +213,11 @@ where
         Tensor::new(K::reshape(self.primitive, shape))
     }
 
-    /// Transpose the tensor.
+    /// Transpose the tensor. Permute the 2 last dimensions of the tensor.
+    ///
+    /// For a tensor with a shape `[1, 2, 3, 4]` the result will be a tensor with a shape `[1, 2, 4, 3]`.
+    ///
+    /// See also [`permute`](Tensor::permute).
     ///
     /// # Arguments
     ///
@@ -368,8 +372,11 @@ where
     ///     println!("{moved}");
     /// }
     /// ```
-    // This is a syntactic sugar for `permute`. It is used widely enough, so we define a separate Op
-    // for it
+    /// 
+    /// # Note
+    /// 
+    /// This is a syntactic sugar for `permute`. It is used widely enough, so we define a separate Op
+    /// for it
     pub fn movedim<S1: MovedimArgs, S2: MovedimArgs>(self, src: S1, dst: S2) -> Tensor<B, D, K> {
         let source_dims = src.into_dim_vec::<D>();
         let destination_dims = dst.into_dim_vec::<D>();
@@ -521,8 +528,12 @@ where
     ///
     /// # Type Parameters
     ///
-    ///  - 'D2': The resulting number of dimensions in the squeezed tensor.
+    ///  - `D2`: The resulting number of dimensions in the squeezed tensor.
     ///
+    /// # Panics
+    /// 
+    /// If the size in the squeezed dimension is not 1.
+    ///  
     /// # Returns
     ///
     /// A new `Tensor<B, D2, K>` instance with the specified dimension removed.
@@ -573,7 +584,7 @@ where
     ///
     /// # Type Parameters
     ///
-    ///  - 'D2': The resulting number of dimensions in the squeezed tensor.
+    ///  - `D2`: The resulting number of dimensions in the squeezed tensor.
     ///
     /// # Returns
     ///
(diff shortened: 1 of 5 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -173,7 +173,7 @@ where
         self.primitive.shape()
     }
 
-    /// Reshape the tensor to have the given shape.
+    /// Reshape the tensor to have the given shape, keeping the same values in the same order.
```
laggui, line 176: I get what you're trying to describe, but not a fan of the wording 🤔 

Suggestion: 
```
Reshape the tensor to have the given shape.
The tensor has the same data and number of elements as the input.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -213,7 +213,11 @@ where
         Tensor::new(K::reshape(self.primitive, shape))
     }
 
-    /// Transpose the tensor.
+    /// Transpose the tensor. Permute the 2 last dimensions of the tensor.
+    ///
+    /// For a tensor with a shape `[1, 2, 3, 4]` the result will be a tensor with a shape `[1, 2, 4, 3]`.
```
laggui, line 218:
```suggestion
    /// Transpose the tensor. Permute the 2 last dimensions of the tensor.
    ///
    /// For a 2D tensor, this is the standard matrix transpose. For `D > 2`, the transpose is applied on the last two dimensions. For example, the transpose of a tensor with shape `[1, 2, 3, 4]` will have shape `[1, 2, 4, 3]`.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -650,8 +661,19 @@ where
     }
 
     /// Unsqueeze the current tensor. Create new dimensions to fit the given size.
+    /// New dimensions will be added in front of existing dimensions.
```
laggui, line 664: I think we could just be more specific with the current description:

```
Unsqueeze the current tensor. Create new leading dimensions to fit the given size.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -650,8 +661,19 @@ where
...
+    ///
+    /// # Type Parameters
+    ///
+    ///  - `D2`: The resulting number of dimensions in the unsqueezed tensor.
+    ///
+    /// # Panics
     ///
-    /// If the output size is higher than the current tensor.
+    /// If `D2` is lower than the actual number of dimensions.
```
laggui, line 672:
```suggestion
    /// If the output size `D2` is smaller than the current number of dimensions.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -905,7 +927,11 @@ where
         K::device(&self.primitive)
     }
 
-    /// Returns a new tensor on the given device.
+    /// Returns a new tensor on the given device filled with the values of the current tensor.
+    ///
+    /// # Note
+    ///
+    /// Unlike the name implies, this is a `into_` conversion that takes ownership of current tensor.
```
laggui, line 934: How about:

```
Move the tensor to the given device.
```

While I understand the note about `into_`, not sure it's required 🤔 `to_device(&device)` is more natural and the semantic intent is pretty clear I think. Even more so if we change the docstring to state that the method *moves* the tensor to the given device.

Lmk what you think!

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -981,7 +1007,8 @@ where
     ///
     /// # Returns
     ///
-    /// A new tensor with the given dimension repeated `times` times.
+    /// A new tensor with the given dimension repeated `times` times, the new tensor has the same rank but
+    /// a modified shape with a size multiplied by `times` in the given dimension.
```
laggui, line 1011: The added description feels redundant, no? Perhaps only a small clarification is needed at the beginning of the docstring

```
/// Repeat the tensor along the given dimension.
///
/// The output tensor has the same shape, except along the given dimension.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1041,7 +1072,7 @@ where
         tensor
     }
 
-    /// Applies element-wise equal comparison and returns a boolean tensor.
+    /// Applies element-wise equal comparison and returns a boolean tensor with the same shape.
```
laggui, line 1075: While we're at it.. this can probably be even more clear.

What do you think of changing the description and adding a return close
```
/// Applies element-wise equal comparison.
///
/// # Returns
/// A boolean tensor that is `true` where input is equal to `other` and `false` elsewhere.
```

Given that the operation is element-wise, I don't think we need to specify that the output has the same shape as the input.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1068,7 +1099,7 @@ where
         Tensor::new(K::equal(self.primitive, other.primitive))
     }
 
-    /// Applies element-wise non-equality comparison and returns a boolean tensor.
+    /// Applies element-wise non-equality comparison and returns a boolean tensor with the same shape.
```
laggui, line 1102: Same comment

```
/// Applies element-wise non-equality comparison.
///
/// # Returns
/// A boolean tensor that is `true` where input is not equal to `other` and `false` elsewhere.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1548,6 +1581,13 @@ where
 
     /// Broadcast the tensor to the given shape.
     ///
+    /// For each dimension:
+    /// - If current size is `1` and the requested size is `n` > 1, the tensor will be extended by
+    ///   repeating itself in this dimension.
```
laggui, line 1586: This is misleading, because expand does not actually repeat the values (which would imply allocating new memory for the repeated output tensor).

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1548,6 +1581,13 @@ where
...
+    /// - If the current size is `n` > 1, then output size must be `n`. `-1` can be used to inferred this
+    ///   automatically.
+    /// - Otherwise tensor cannot be broadcasted.
```
laggui, line 1589: I think it might be better if we instead clarify the description:

eturns a new view of the self tensor with singleton dimensions expanded to a larger size.

```
Broadcast the tensor to the given shape.

Only singleton dimensions can be expanded to a larger size. Other dimensions must have the same size (which can be inferred with `-1`).
````

**Review by laggui — CHANGES_REQUESTED** [review]
Thanks for improving the docs! This might be a common friction point for new users 🙏 

I have a couple of comments, lmk what you think.

**Comment by laggui** [comment]
Regarding your other remarks:

More general remarks about the documentation:

> - There is no indications for contributing to documentation in the contribution guide, this is a bit missing, specially for the 2 following points

That's a good point! We should probably add at least a small section.

> - The verbs are not consistent, sometimes functions are described with "Does this and that" and sometimes it's "Do this and that", is there any recommendation here? If so making it explicit in the guidelines would be great!

I agree, it would be nice to stick to the same pattern / tense for the operations. It's not a big deal, but standardizing the docs would be nice. I don't have any personal preference for the tense, as long as it is clear 😅 

> - The usage of the vocabulary rank, dimension, dimensions, size and shape is not always consistent, quite often I found "tensor of dimensions [x, y, z]", the idea is totally understandable here, but in this case the right word is "tensor of shape [x, y, z]", no? I am asking just to know if I should try to be accurate in the vocab for next PR (maybe even replace existing misuses) or if it's not important and should not be changed?

Hahah yeah there might be a slight abuse of language in some instances where dimensions/shape/size is used 😬  If you want to improve the docs to disambiguate the usage of these terms, you can do it in another PR 🙂

File: crates/burn-tensor/src/tensor/api/base.rs (hunk 5 above)
Comment (PtiLuky) [reply crates/burn-tensor/src/tensor/api/base.rs:934]: Yes, after a second read on this one, it does not actually follow the semantic of `to_` + `casted type`, and it's actually quite clear that the intent is not to cast the vector into a device...
I will remove this "Note" then.

File: crates/burn-tensor/src/tensor/api/base.rs (hunk 9 above)
Comment (PtiLuky) [reply crates/burn-tensor/src/tensor/api/base.rs:1586]: Uhm... then I probably did not fully understand this method myself. I will remove this comment to avoid leaving a misleading comment.

File: crates/burn-tensor/src/tensor/api/base.rs [hunk 11]
```diff
@@ -213,7 +215,13 @@ where
         Tensor::new(K::reshape(self.primitive, shape))
     }
 
-    /// Transpose the tensor.
+    /// Transpose the tensor. Permute the 2 last dimensions of the tensor.
```
laggui, line 218: Forgot this edit in my previous review comment. That's on me 😅 

```suggestion
    /// Transpose the tensor.
```

crates/burn-tensor/src/tensor/api/base.rs
laggui: Expanding a tensor does not allocate new memory. Instead, it modifies the strides of the tensor to *virtually* expand it along the specified dimensions.

This is done by setting the stride to 0 for the expanded dimensions, meaning every element along that dimension points to the same memory location. This effectively "repeats" values without duplication.

Hopefully this clears it up a bit 😅

PtiLuky: Yes clear (and clever)!

Issue #2924, opened 2025-03-19
Grad.get leads to panic due to unwrap() instead of returning the None option


VirtualNonsense:
Hey everyone!
I'm currently trying to visualize gradients and trying to access the inner workings of gradients.
During this I noticed that accessing an ID that is not present within the instance of GradientsParams leads to a panic. 

The issue for this seems to lie in this [line](https://github.com/tracel-ai/burn/blob/26269a69efa0e26a7deef44959464db0445f4ab7/crates/burn-tensor/src/tensor/container.rs#L48)

If i understand this section correctly something like this should do the trick
```rust
    /// Get a tensor with the given ID.
    pub fn get<B>(&self, id: &ID) -> Option<TensorPrimitive<B>>
    where
        B: Backend,
    {
        let grad = self.tensors.get(id)?;

        match grad.downcast_ref::<TensorPrimitive<B>>(){
            Some(tensor) => Some(tensor.clone())
            None => None
        }
    }
```

VirtualNonsense:
First of all: thanks for the response!
> Hmm I'm not sure this is the cause of your issue. If the ID is not present, then self.tensors.get(id)? would propagate None to the   calling function. So the downcast should not happen on None.

Yeah of course. 
Sorry maybe I was a bit quick to jump to a conclusion 😬
The issue seems to be the down casting operation bellow.

first of all here is the full stack backtrace:
```bash
cargo run --package sketchy_pix2pix --bin train_pix2pix --release
    Finished `release` profile [optimized] target(s) in 1.54s
     Running `target\release\train_pix2pix.exe`
[00:00:00] ###########-----------------------------       1/4       Epochs
[00:00:00] ----------------------------------------       0/61262   Training:                                                                          
thread 'main' panicked at C:\Users\anachtmann\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\burn-tensor-0.16.0\src\tensor\container.rs:48:14:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x7ff65b2cbd73 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h985b9d5f3de9f12f
   1:     0x7ff65b319b3a - core::fmt::write::h9a96aa6da4d5ad38
   2:     0x7ff65b2c1762 - std::io::Write::write_fmt::h74a921bbf668ad6e
[...]

PR #2965, opened 2025-03-27
Enhance gradient api

VirtualNonsense:
### Checklist

### Related Issues/PRs

#2924 

### Changes

I adjusted the `TensorContainer` according the discussion in #2924 and adjusted the usage accordingly. 
Question: Should the return type be propagated?


laggui (commented):
> Question: Should the return type be propagated?

We have two options:
1) Propagate the result to the `GradientsParams` methods (`get(...)` and `remove(...)`)
2) `.unwrap()` the result, which will panic with the error type (or `.expect(...)` with a descriptive message)

but I don't think we should not simply ignore the error with `.ok()`. This will be even more difficult to understand for user error 😅

VirtualNonsense:
Hey!
Please excuse my rather cautious approach. I'm a little bit burned by my corporate environment where even minor changes to lead to people scream at you 😅

If you don't mind I'll choose the first option after getting the check to run at my machine.

laggui:
Ha, no need to apologize! 😄 

Your feedback is important, especially since you were one of the first users to report this discrepancy. So if you feel that one option is more natural / helpful for an end-user, that's great!

VirtualNonsense:
Please excuse the delay. 

Here is my proposed solution for #2924.
I finally did not change the return type of the backend trait methods and chose to panic instead since it might break a lot of code and I don't feel like this change is important enough to justify such a change.
Let me know if I can enhance this PR in any way!

crates/burn-autodiff/src/backend.rs
```diff
@@ -86,7 +103,11 @@ impl<B: Backend, C: CheckpointStrategy> AutodiffBackend for Autodiff<B, C> {
         grads: &mut Self::Gradients,
         grad: B::FloatTensorPrimitive,
     ) {
-        grads.remove::<B>(tensor);
+        if let Err(TensorContainerError::DowncastError) = grads.remove::<B>(tensor) {
+            panic!(
+                "Downcast mismatch when retriefing tensor. If you are trying to replace the gradients for a given parameter id, make sure to use the inner backend. Gradients are not stored on the autodiff backend."
```
laggui, line 108: Typo `retriefing` -> `retrieving`.

Instead of repeating the literal string for the panic message, maybe we can keep is as a const or have a single helper function to map the message?

crates/burn-core/src/optim/grads.rs
```diff
@@ -66,15 +66,15 @@ impl GradientsParams {
     where
         B: Backend,
     {
-        self.container.get(&id).map(Tensor::from_primitive)
+        self.container.get(&id).map(Tensor::from_primitive).ok()
```
laggui, line 69: If we just use `.ok()` here it will simply ignore the downcast error 😅 

Given that the initial motivation for this improved error handling came from the `GradientsParams` usage, we should probably make the distinction here. I feel like it's probably more important at this level than at the autodiff backend level (which you added).

Then, we could add a small test for the downcast error to demonstrate incorrect usage (trying to get gradients on the autodiff wrapped backend).

crates/burn-tensor/src/tests/mod.rs
```diff
@@ -19,6 +20,7 @@ macro_rules! testgen_all {
             pub type IntType = <TestBackend as $crate::backend::Backend>::IntElem;
             pub type BoolType = <TestBackend as $crate::backend::Backend>::BoolElem;
 
+            $crate::testgen_container!();
```
laggui, line 23: Probably don't need to run this for all backends. Behavior is the same across all here, as opposed to other tests in the test macro (tensor ops implementation).

Maybe we can move this to burn-core, alongside the suggested addition for `GradientsParams` test.

laggui (changes requested):
Overall, the addition of the result makes sense.

Just have some comments on the usage. See below 🙂

VirtualNonsense:
Thank you for taking the time to reviewing and sorry for that sloppy oversight!
I'll try to account for those issues as soon as possible.

laggui:
Ha no worries! That's what reviews are for 🙂

laggui:
@VirtualNonsense also let me know when this is ready for another round of review 🙂

crates/burn-tensor/src/tensor/container.rs
```diff
@@ -9,6 +9,18 @@ use std::collections::HashMap;
...
+
+    /// Downcast mismatch when retrieving tensor.
+    /// If you are trying to retrieve the gradients for a given parameter id, make sure to use the inner backend.
+    /// Gradients are not stored on the autodiff backend.
+    DowncastError,
+}
+
 /// Contains tensor of arbitrary dimension.
 #[derive(Debug)]
 pub struct TensorContainer<ID> {
```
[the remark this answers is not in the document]
VirtualNonsense: Thank you for taking the time to review my PR!

If I unterstand you correctly you are suggest to change the `TensorContainer` to something like this:
```rust
/// Contains tensor of arbitrary dimension.
#[derive(Debug)]
pub struct TensorContainer<ID, B: Backend> {
    tensors: HashMap<ID, TensorPrimitive<B>>,
}
```
and thereby sidestepping the issue entirely?
I have to admit I'm not that good with rust yet and do not know burn well enough to properly to foresee the consequences of that change but I can try if you like :D
Should I create another PR for that version?

PR #2966, opened 2025-03-28
Fix: burn-book getting-started Use Declarations

jerryshell:
### Checklist

### Changes

`backend::backend::Wgpu` -> `backend::Wgpu`


laggui (approved):
Good catch 😅 must have missed it all this time due to the block being collapsed by default.

Thanks!

PR #2974, opened 2025-03-31
Transformers

Brooooooklyn:
### Checklist

### Related Issues/PRs

### Changes

### Testing


theoparis:
Why is this closed? I'd like to be able to use burn with huggingface llama models since candle doesn't support webgpu/AMD.

antimora:
> Why is this closed? I'd like to be able to use burn with huggingface llama models since candle doesn't support webgpu/AMD.

You should be able to. Here is how: https://github.com/tracel-ai/models/tree/main/llama-burn

This recent one uses burn-nn https://github.com/tracel-ai/models/pull/92

PR #2990, opened 2025-04-03
update cubecl

louisfd:


crates/burn-cubecl-fusion/src/matmul/optimization.rs
```diff
@@ -285,12 +285,12 @@ impl FusedMatmul {
         let rhs_strides = inputs.strides(&self.rhs);
 
         let check_layout = |strides| match matrix_layout(strides) {
```
laggui, line 287: Should `matrix_layout` also be renamed to match the distinction between `MatrixLayout` and `MatrixBatchLayout`?

laggui (approved):
LGTM! Thanks for fixing the CI for the new rust stable version 😄 

Just a minor comment for function that lives in cubecl

crates/burn-cubecl-fusion/src/matmul/optimization.rs
louisfd: You're right. However I'll do it in a subsequent PR to avoid going through another CubeCL revision just for that.

laggui: Yeah not a blocking change! We can merge this PR first

PR #2997, opened 2025-04-06
Fix gradient checkpointing

nathanielsimard:


examples/text-classification/src/model.rs
```diff
@@ -141,7 +141,9 @@ impl<B: AutodiffBackend> TrainStep<TextClassificationTrainingBatch<B>, Classific
     ) -> TrainOutput<ClassificationOutput<B>> {
         // Run forward pass, calculate gradients and return them along with the output
         let item = self.forward(item);
+        let device = item.loss.device();
```
louisfd, line 144: useless?

louisfd (changes requested):
Overall still a bit dirty but I understand the fix

PR #3019, opened 2025-04-14
Update cubecl

nathanielsimard:


syl20bnr:
Maybe you can combine this one with the compilation cache ?

PR #3024, opened 2025-04-15
Improve test tolerance assertions

laggui:
### Checklist

### Related Issues/PRs

The previous implementation had huge tolerances in f16 due to the eps checks in the comparisons (so big that `9 == 50` in f16 because the tolerance would be computed as > 2000).

### Changes

Added new `Tolerance` to compute approx equality based on absolute / relative tolerances.

Refactored all tests to use this new tolerance and adjusted the bounds based on the computation. Some tests had expected values defined without enough precision, so changed those as well. Only very few tests should be allowed to be outside of standard tolerances for f32 or f16.

Reference: https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison


crates/burn-tensor/src/tensor/bytes.rs
```diff
@@ -50,9 +50,9 @@ impl core::fmt::Debug for Bytes {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         let data = &**self;
         let fmt_data = move |f: &mut core::fmt::Formatter<'_>| {
-            if data.len() > 3 {
+            if data.len() > 20 {
                 // There is a nightly API `debug_more_non_exhaustive` which has `finish_non_exhaustive`
-                f.debug_list().entries(&data[0..3]).entry(&"...").finish()
+                f.debug_list().entries(&data[0..20]).entry(&"...").finish()
```
maxtremblay, line 55: SHould we keep to 3?

PR #3031, opened 2025-04-16
Fix fusion bug

nathanielsimard:


laggui (approved):

antimora:
A little bit of more information will be nice =D 

In all seriously, guys lets not skip steps when merging PRs. We need context here. It will during conflict resolutions/fixing other issues. The more information we will put, the less we will be pinged on discord to explain the changes.

PR #3036, opened 2025-04-16
Remove affine

maxtremblay:
Remove affine and per block quantization. This is the first toward the new quantization scheme. It is likely that per block quantization will be reintroduced later when the new scheme is implemented.


laggui (approved):
Thanks for the clean up!

We'll bring back per-block quantization in the future but not there yet with cubecl so not worth keeping the code around at this time.

For affine quantization.. turns out in practice it's not as widely used 🙂

PR #3043, opened 2025-04-17
Fix autotune measurement

nathanielsimard:


laggui (commented):
Linking to https://github.com/tracel-ai/cubecl/pull/630

Reflects new time measurement for cubecl runtimes

PR #3051, opened 2025-04-20
Restrict ONNX opset to 16 and up

### LINKED ISSUE #3046: ONNX Import Feature: Require opset_version == 16
The ONNX import feature in the Burn project should enforce `opset_version == 16` to ensure compatibility, optimize performance, and reduce development complexity. Supporting multiple opset versions increases maintenance overhead and testing requirements, so standardizing on opset_version 16 simplifies the implementation. Users with ONNX models using other opset versions should be guided to upgrade their models to opset_version 16.

### Proposed Solution
- Update the ONNX import functionality to check for `opset_version == 16` and raise an informative error if the condition is not met.
- Provide users with documentation and a script to upgrade their ONNX models, including shape inference to maintain model integrity.

#### Model Upgrade Script
Users can upgrade their ONNX models using the following Python script:

```python
import onnx
from onnx import shape_inference
from onnx import version_converter

# Load your ONNX model
model = onnx.load('onnx-tests/tests/unsqueeze/unsqueeze.onnx')

# Upgrade the model to opset 16
upgraded_model = version_converter.convert_version(model, 16)

# Apply shape inference
inferred_model = shape_inference.infer_shapes(upgraded_model)

# Save the inferred model
onnx.save(inferred_model, 'unsqueeze_inferred.onnx')
```

### Tasks
1. Add validation in the ONNX import code to enforce `opset_version == 16`.
2. Create a clear error message directing users to the model upgrade process.
3. Update the project documentation to include the opset version requirement, the upgrade script, and the reasoning for requiring opset_version 16 (reduced development complexity).
4. Add a test case to verify that models with `opset_version != 16` are rejected.

### Reasoning
Requiring `opset_version == 16` simplifies the development and maintenance of the ONNX import feature by:
- Reducing the complexity of supporting multiple opset versions, which have varying operator implementations.
- Minimizing the test matrix for validating ONNX model compatibility.
- Streamlining debugging and issue resolution by focusing on a single opset version.

### Additional Notes
- Ensure the error message includes a link to the documentation for the upgrade process.
- Consider adding a utility function in the Burn project to automate the shape

antimora:
**Related Issue:** 

Fixes #3046

**Changes:**
- Updated ONNX model import documentation for clarity and added a new "ONNX Compatibility" section.
- Required ONNX models to use opset version 16 or higher, deprecating older versions.
- Simplified documentation text and added a Python script for upgrading ONNX models to opset 16.
- Disabled outdated ONNX test cases (e.g., `clip_opset7`, `dropout_opset7`, `reduce_sum_opset11`) to align with opset 16.
- Removed unnecessary imports in test code for a cleaner codebase.

**Testing:**
- Verified exception handling for opset 7 ONNX models.
- Confirmed all existing tests pass.

**Checklist:**
- Ran `run-checks all` script.
- Updated the book with PR changes.


crates/burn-import/onnx-tests/build.rs
```diff
@@ -97,8 +97,8 @@ fn main() {
         .input("tests/reduce_mean/reduce_mean.onnx")
         .input("tests/reduce_min/reduce_min.onnx")
         .input("tests/reduce_prod/reduce_prod.onnx")
-        .input("tests/reduce_sum/reduce_sum_opset11.onnx")
-        .input("tests/reduce_sum/reduce_sum_opset13.onnx")
+        // .input("tests/reduce_sum/reduce_sum_opset11.onnx")  // Disabled for OpSet 16 below
+        // .input("tests/reduce_sum/reduce_sum_opset13.onnx") // Disabled for OpSet 16 below
```
laggui, line 101: While the opset refers to the global graph operation set, note that some operations might still have an opset version that is lower! For example, the latest `ReduceSum` op version is 13 - so it should be the same as opset 13.

It has not been updated (i.e., no breaking changes or new features) since version 13. So it should still be compatible!

laggui (commented):
It makes sense to restrict to more recent opsets. The ONNX spec is already large enough, we should try to limit the scope 😅

In this case, I would simply delete the tests and ONNX files associated with older opsets.

crates/burn-import/onnx-tests/build.rs
antimora: Fixed it. Thanks for the catch.

laggui (commented):
LGTM overall! Mostly minor comments.

Will wait for the linked PR to be merged before this review is finalized.

crates/burn-import/onnx-tests/build.rs
```diff
@@ -116,22 +113,20 @@ fn main() {
...
+        .input("tests/squeeze/squeeze.onnx")
         .input("tests/sub/sub.onnx")
         .input("tests/sub/sub_int.onnx")
         .input("tests/sum/sum.onnx")
         .input("tests/sum/sum_int.onnx")
         .input("tests/tan/tan.onnx")
         .input("tests/tanh/tanh.onnx")
         .input("tests/tile/tile.onnx")
-        .input("tests/top_k/top_k_opset_1.onnx")
+        // .input("tests/top_k/top_k_opset_1.onnx") // Disabled for OpSet 16 below
```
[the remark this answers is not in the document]
antimora: Yeah, thanks for catching. I forgot about this one.

I recreated the onnx creation script using torch with proper new tests. We have now topk working for opset16

burn-book/src/import/onnx-model.md
```diff
@@ -1,72 +1,86 @@
...
+version, you'll need to upgrade it using the ONNX version converter.
+
+### Upgrading ONNX Models
+
+There are two simple ways to upgrade your ONNX models to the required opset version:
+
+Option 1: Use the provided utility script:
+
+```
+uv run --script https://raw.githubusercontent.com/tracel-ai/burn/refs/heads/main/crates/burn-import/onnx-opset-upgrade.py
```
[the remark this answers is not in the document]
antimora: True but I am seeing more people shifting towards `uv`. We are using it consistently with Burn too.

There are a few convenient aspect of this example:

1. uv downloads the script directly and executes.
2. creates necessary virtual environment 
3. installs the dependency packages specified in the script itself.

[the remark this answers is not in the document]
laggui: Oh I am not advocating to remove the uv command entirely, just that we might want to provide the basic steps to convert the model with python 🙂

burn-book/src/import/onnx-model.md
```diff
@@ -1,72 +1,86 @@
...
+
+# Load the ONNX model
+model = onnx.load('path/to/your/model.onnx')
+
+# Upgrade to opset version 16
+converted_model = version_converter.convert_version(model, 16)
+
+# Save the upgraded model
+onnx.save(converted_model, 'upgraded_model.onnx')
+```
```
laggui, line 79: Btw, why are there so many more steps in the `onnx_opset_upgrade.py` script? Actually, most of the additional code is for debug info or user input, but what about the additional shape inference?

burn-book/src/import/onnx-model.md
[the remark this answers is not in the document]
antimora: They can download the script and run in python like `python3 onnx_opset_upgrade.py` too or they the can follow the python script steps.

Another alternative is to call `uv` as a library. We can put it behind a feature flag. This pattern can also work with dataset crate. See https://github.com/astral-sh/uv/pull/4642

burn-book/src/import/onnx-model.md
antimora: Updated the document to include that step. I added it as a safe measure in case the older opset files didn't infer shapes already. But also as a precursor to use inferred node output shapes. For now, I think we can figure out Input/Output ranks from the static shapes.

crates/burn-import/onnx_opset_upgrade.py
laggui, line 1: Btw have you successfully tested this script with some more complex ONNX models? 🙂 Would like to avoid maintaining a python script over time

laggui (approved):

crates/burn-import/onnx_opset_upgrade.py
antimora: Yes, I have tested with this 1gb [onnx](https://huggingface.co/iiiorg/piiranha-v1-detect-personal-information/resolve/refs%2Fpr%2F8/onnx/model.onnx?download=true) file reported in recent bug

PR #3056, opened 2025-04-21
Fix matmul dynamic line size support

nathanielsimard:
https://github.com/tracel-ai/cubecl/pull/634

The feature isn't available on Vulkan.


antimora (approved):
LGTM

PR #3061, opened 2025-04-22
Improve `tensor.slice(...)` to support multiple range types

### LINKED ISSUE #3048: Improve `tensor.slice(ranges)` to accept a single dim element and other range types
**Describe the bug**
I was looking at documentation for Tensor::slice and one thing feels odd to me. According to the documentation for negative values it will "Handles negative indices by wrapping around from the end of the dimension." so how to select last element in dimension without using positive index with correct position of dim? [Example](https://docs.rs/burn-tensor/0.16.1/burn_tensor/struct.Tensor.html#examples) presents that (1, -1) will return tensor from second element till second last. In Python it is addressed by omitting end part of range: `some_list[-1..]`.

Also, it is not possible to use other range operators than `core::ops::Range`. If I want to use, for example, `RangeFrom`, it will result in a compilation error.

**To Reproduce**
```
    // Using negative indices
    let tensor = Tensor::<Wgpu, 3, Int>::from_data([
        [[1, 2], [3, 4], [5, 6]],
        [[7, 8], [9, 0], [1, 2]]
    ], &device);
    println!("Tensor: {}", tensor.clone());
    let slice = tensor.slice([None, Some((-2, -1))]);
    println!("Slice: {}", slice);
```
This will produce output:
```
Slice: Tensor {
  data:
[[[3, 4]],
 [[9, 0]]],
  shape:  [2, 1, 2],
}
```
What is correct code. but there is no way to select last element
```
    let slice = tensor.slice([None, Some((-1, 0))]); // Would 0 be valid way to provide such information?
```
So I will receive:
```
Slice: Tensor {
  data:
[[[5, 6]],
 [[1, 2]]],
  shape:  [2, 1, 2],
}

```

Example of using other Range operators:
```
    let tensor = Tensor::<Wgpu, 3, Int>::from_data([
        [[1, 2], [3, 4], [5, 6]],
        [[7, 8], [9, 0], [1, 2]]
    ], &device);
    println!("Tensor: {}", tensor.clone());
    let slice = tensor.slice([..]); // Similar to 2.., ..3 and 1..=2
    println!("Slice: {}", slice);
```
Will result in compilation error:
```
error[E0277]: the trait bound `[RangeFull; 1]: RangesArg<_>` is not satisfied
   --> src/main.rs:13:30
    |
13  |     let slice = tensor.slice([..]);
    |                        ----- ^^^^ the trait `RangesArg<_>` is not implemented for `[RangeFull; 1]`
    |                        |
    |                        required by a bound introduced by this call
    |
    = help: the following other types implement trait `RangesArg<D2>`:
              [(i64, i64); D2]
              [std::ops::Range<usize>; D2]
              [std::option::Option<(i64, i64)>; D2]
note: required by a bound in `burn::tensor::Tensor::<B, D, K>::slice`
   --> /Users/adam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-tensor-0.16.1/src/tensor/api/base.rs:861:38
    |
861 |     pub fn slice<const D2: usize, R: RangesArg<D2>>(self, ranges: R) -> Self {
    |                                      ^^^^^^^^^^^^^ required by this bound in `Tensor::<B, D, K>::slice`
```

**Expected behavior**
Way to provide range that allow to select also last element using negative indexes
Accept other Range operators as input

**Additional context**
I can work on updating this code to include this cases, I found that it is caused by `RangesArg` trait.
Other Range operators looks straight forward, but I need some guidance how to tackle issue with last element using negative index. Maybe accepting `Option<i64>` next to current pair: `Option<(i64, i64)>?

laggui:
### Checklist

### Related Issues/PRs

Fixes #3048

### Changes

- Added more range types (and single index) support for slice
- Removed `(i64, i64)` and `Option<i64, i64>` tuples support (breaking change)
  - ~~I wanted to add a deprecation warning instead of directly introducing the breaking change, but there is not straightforward way to do so since the `#[deprecated]` on trait impl blocks has no effect on the user code. Either we accept the breaking change or we document it and add a runtime warning?~~ 
- New slice macro to allow heterogenous slice specification (e.g., `s![0, .., 2.., -1]`)

**Example usage**
```rust
use burn::prelude::*; // now also includes the `s!` macro

let tensor = Tensor::<B, 3>::zeros([2, 4, 2], &Default::default());
let slice = tensor.slice(s![1.., 1..=3, -1]);
assert_eq!(slice.dims(), [1, 3, 1]);

let tensor = Tensor::<B, 4>::zeros([8, 4, 2, 3], &Default::default());
let slice = tensor.slice(s![..=4, 0..=3, .., -2..]);
assert_eq!(slice.dims(), [5, 4, 2, 2]);
```

### Testing

Unit tests


crates/burn-tensor/src/tests/ops/slice.rs
```diff
@@ -180,24 +180,24 @@ mod tests {
...
         let data = TensorData::from(as_type!(FloatType: [[0.0f32, 1.0, 2.0]]));
         output.into_data().assert_eq(&data, true);
 
-        let output = tensor.clone().slice([None, Some((0, 2))]);
+        let output = tensor.clone().slice(s![.., 0..2]);
         let data = TensorData::from(as_type!(FloatType: [[0.0f32, 1.0], [3.0, 4.0]]));
         output.into_data().assert_eq(&data, true);
 
-        let output = tensor.clone().slice([None, None]);
+        let output = tensor.clone().slice(s![.., ..]);
```
antimora, line 200: Not sure if it's worth covering `let output = tensor.clone().slice([.., ..]);` homogenous case (without `s!`)

antimora:
Yeap it looks good and cleaner. It's similar to ndarray's design except that we don't support step:

https://docs.rs/ndarray/latest/ndarray/macro.s.html

(probably we should add step in the future as I have seen needed in ONNX)

laggui:
Yeah I think it was Genna that mentioned that on discord regarding ndarray. The name for the macro `s![]` is concise and always used for `tensor.slice(...)` so it made a lot of sense.

I didn't check their implementation though, but like you said they support stepped ranges too which we currently don't. I just started by extending our support to more range types, then added the macro for heterogenous slice specifiers. Was a pretty natural extension.

crates/burn-tensor/src/tests/ops/slice.rs
laggui: Hah yeah in this case the macro is not actually required. We could adjust the tests.

PR #3062, opened 2025-04-22
[hip] Move burn-hip to burn-rocm and rename backend to ROCm

syl20bnr:
### Checklist

### Testing

Test on a machine with an AMD GC with text classification example.

### DIFF
--- a/crates/burn-rocm/src/lib.rs
+++ b/crates/burn-rocm/src/lib.rs
@@ -12,11 +12,11 @@ use cubecl::hip::HipRuntime;
 
 #[cfg(target_os = "linux")]
 #[cfg(not(feature = "fusion"))]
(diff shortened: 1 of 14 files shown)


crates/burn-rocm/src/lib.rs
```diff
@@ -12,11 +12,11 @@ use cubecl::hip::HipRuntime;
 
 #[cfg(target_os = "linux")]
 #[cfg(not(feature = "fusion"))]
-pub type Hip<F = f32, I = i32, B = u8> = CubeBackend<HipRuntime, F, I, B>;
+pub type ROCm<F = f32, I = i32, B = u8> = CubeBackend<HipRuntime, F, I, B>;
```
laggui, line 15: Following the convention with other backends, should the name be `Rocm`?

nathanielsimard: Same comment

laggui (approved):

PR #3069, opened 2025-04-23
[Perf] Slice assign

wingertge:
### Checklist

### Changes

Uses `FastDivmod` and vectorization for speeding up `slice_assign`

### Testing

All tests pass


crates/burn-cubecl/src/kernel/index/slice_assign.rs
```diff
@@ -1,59 +1,112 @@
 use crate::{CubeRuntime, element::CubeElement, tensor::CubeTensor};
 use cubecl::{calculate_cube_count_elemwise, prelude::*};
+use cubecl_std::{FastDivmod, FastDivmodArgs};
 use std::ops::Range;
 
 #[cube(launch)]
```
nathanielsimard, line 6: Should we put `launch_unchecked`

crates/burn-cubecl/src/kernel/index/slice_assign.rs
```diff
@@ -1,59 +1,112 @@
...
+        let shape = last.end - last.start;
+        let offset = last.start;
+        *R::supported_line_sizes()
+            .iter()
+            .filter(|it| {
+                let it = **it as usize;
+                shape % it == 0 && strides_compatible(&tensor.strides, it) && offset % it == 0
+            })
+            .max()
+            .unwrap_or(&1)
```
nathanielsimard, line 71: Could we use a built-in function for that? like `tensor_line_size_parallel` or `tensor_line_size_perpendicular`

https://docs.rs/cubecl/latest/cubecl/fn.tensor_line_size_perpendicular.html
https://docs.rs/cubecl/latest/cubecl/fn.tensor_line_size_parallel.html

wingertge: I don't think so because it needs to be compatible with the shape like usual, but also additionally compatible with the offset and slice length.

crates/burn-cubecl/src/kernel/index/slice_assign.rs
wingertge: Should be fine yeah.

nathanielsimard (approved):

Issue #2901, opened 2025-03-12
Invalid fused shader generated


ArthurBrussee:
**Describe the bug**

When running on WASM, I see some errors when compiling wgsl shaders. This does seem like actually invalid wgsl - casting a vec4f to a f32.

```
Error while parsing WGSL: :315:23 error: no matching constructor for 'f32(vec4<f32>)'

2 candidate constructors:
 • 'f32(f32  ✗ ) -> f32'
 • 'f32() -> f32' where:
      ✗  overload expects 0 arguments, call passed 1 argument

1 candidate conversion:
 • 'f32(T  ✗ ) -> f32' where:
      ✗  'T' is 'abstract-int', 'abstract-float', 'i32', 'f16', 'u32' or 'bool'

output_2_global[id] = f32(l_175);
                      ^^^^^^^^^^


 - While validating [ShaderModuleDescriptor]
 - While calling [Device "egui+burn"].CreateShaderModule([ShaderModuleDescriptor]).
 ```

The generated shader looks like:

```
@group(0)
@binding(0)
var<storage, read> input_0_global: array<f32>;

@group(0)
@binding(1)
var<storage, read> input_1_global: array<vec4<f32>>;

@group(0)
@binding(2)
var<storage, read> input_2_global: array<vec4<f32>>;

@group(0)
@binding(3)
var<storage, read_write> output_0_global: array<vec4<f32>>;

@group(0)
@binding(4)
var<storage, read_write> output_1_global: array<vec4<f32>>;

@group(0)
@binding(5)
var<storage, read_write> output_2_global: array<f32>;

@group(0)
@binding(6)
var<storage, read> info: array<u32>;

@group(0)
@binding(7)
var<storage, read> scalars_f32: array<f32, 2>;

@group(0)
@binding(8)
var<storage, read> scalars_u32: array<u32, 4>;

const WORKGROUP_SIZE_X = 16u;
const WORKGROUP_SIZE_Y = 16u;
const WORKGROUP_SIZE_Z = 1u;

@compute
@workgroup_size(16, 16, 1)
fn elemwise_fuse(
    @builtin(global_invocation_id) global_id: vec3<u32>,
    @builtin(num_workgroups) num_workgroups: vec3<u32>,
) {
var a_0: array<u32, 4>;

var a_1: array<u32, 4>;

[...]

PR #3075, opened 2025-04-24
Fix vectorization problem with fusion on reshaped not contiguous tens…

nathanielsimard:
It should fix some compilation issue on WGPU.


laggui (approved):
Fixes #2901

Tested with the provided MWE ✅

PR #3077, opened 2025-04-24
[Perf] Interpolate optimizations

wingertge:
### Checklist

### Related Issues/PRs

Also fixes https://github.com/tracel-ai/burn/pull/3076

### Changes

Refactors interpolation to use NHWC layout to allow for potential vectorization when used with NHWC layout tensors (i.e. from convolution), as well as other perf optimizations. Also fixes a bug in `is_contiguous` and a clippy warning for slices.

### Testing

All tests pass

### DIFF
--- a/crates/burn-cubecl/src/kernel/conv/conv2d/implicit_gemm/launch.rs
+++ b/crates/burn-cubecl/src/kernel/conv/conv2d/implicit_gemm/launch.rs
@@ -18,7 +18,7 @@ use cubecl::linalg::{
 
 use crate::{
     CubeElement, CubeRuntime, FloatElement,
-    ops::{numeric::empty_device_strided, permute, permute_nchw_to_nhwc, permute_nhwc_to_nchw},
+    ops::{numeric::empty_device_strided, permute_nchw_to_nhwc, permute_nhwc_to_nchw},
     tensor::CubeTensor,
 };
 
@@ -96,7 +96,7 @@ where
     );
 
     let input = permute_nchw_to_nhwc(input);
-    let weight = permute(weight, &[0, 2, 3, 1]);
+    let weight = permute_nchw_to_nhwc(weight);
 
     let out_shape = Shape::new([batch_size, out_h, out_w, out_channels]);
     let out =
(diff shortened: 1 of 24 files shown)


crates/burn-cubecl/src/tensor/base.rs
```diff
@@ -476,6 +471,8 @@ pub(crate) fn is_contiguous(shape: &[usize], strides: &[usize]) -> bool {
             if prev_stride >= *stride {
                 return false;
             }
+        } else if *stride != 1 {
+            return false;
```
laggui, line 475: We should probably add more tests (e.g., your tensor of shape `[1, 1, 1, 9]` with strides `[72, 1, 72, 8]` which prompted this fix). Better illustrates what is expected to be contiguous or not, even in more complex cases.

Also @nathanielsimard I just realized we seem to have another `is_contiguous` fn in fusion, but the implementations are slightly diverging with the latest changes

https://github.com/tracel-ai/burn/blob/e096b0cc789520852533c053603b9804b1e6e1c3/crates/burn-cubecl-fusion/src/base.rs#L141-L169

The definition of contiguous should probably match 😅

laggui (approved):
LGTM!

Tested locally and with the super-resolution model code (ESRGAN) which prompted this PR.

Just a minor comment on the contiguous checks

crates/burn-cubecl/src/kernel/conv/conv2d/implicit_gemm/launch.rs
```diff
@@ -95,7 +95,7 @@ where
         width,
     );
 
-    let input = permute(input, &[0, 2, 3, 1]);
+    let input = permute_nchw_to_nhwc(input);
     let weight = permute(weight, &[0, 2, 3, 1]);
```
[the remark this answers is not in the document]
wingertge: I've kind of gone back and forth on this because I'm unsure about the semantics. The order of the params is the same, but the meaning is a bit different. It's not N, it's the out channels. But since the other 3 fit I think you're right, I'll change it.

crates/burn-cubecl/src/tensor/base.rs
wingertge: I added a test, but leaving this unresolved because of the fusion thing

laggui: Unrelated to this PR, so I'll merge but will open a small PR to use the same `is_contiguous` fn across the crates.

PR #3081, opened 2025-04-25
Update pytorch-model.md with a new troubleshooting help

antimora:
### Checklist

### Changes

Update pytorch import book section on troubleshooting 

### Testing

Rendered the book

### DIFF
--- a/burn-book/src/import/pytorch-model.md
+++ b/burn-book/src/import/pytorch-model.md
@@ -46,17 +46,16 @@ If you accidentally save the entire model instead of just the weights, you may e
 during import like:
 
 ```
-Failed to decode candy: DeserializeError("Serde error: other error:
+Failed to decode foobar: DeserializeError("Serde error: other error:
 Missing source values for the 'foo1' field of type 'BarRecordItem'.
 Please verify the source data and ensure the field name is correct")
 ```
 
-You can verify if your model is exported correctly by opening the `.pt` file in Netron. A properly
-exported weights file will show a flat structure of tensors, while an incorrectly exported file will
-display nested blocks representing the entire model architecture.
-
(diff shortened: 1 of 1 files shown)


laggui (changes requested):
The new section is superfluous.

The first section after the intro is already [`How to export a PyTorch model`](https://burn.dev/burn-book/import/pytorch-model.html#how-to-export-a-pytorch-model)

and the example clearly shows the `state_dict()` being saved.

```py
if __name__ == "__main__":
    torch.manual_seed(42)  # To make it reproducible
    model = Net().to(torch.device("cpu"))
    model_weights = model.state_dict()
    torch.save(model_weights, "conv2d.pt")
 ```

Instead of adding a whole new section even further down, I would simply add a sentence in the first section to highlight that only the state dict should be saved.

File: burn-book/src/import/pytorch-model.md [hunk 1]
```diff
@@ -38,10 +38,23 @@ class Net(nn.Module):
...
+Missing source values for the 'foo1' field of type 'BarRecordItem'.
+Please verify the source data and ensure the field name is correct")
+```
+
+You can verify if your model is exported correctly by opening the `.pt` file in Netron. A properly
+exported weights file will show a flat structure of tensors, while an incorrectly exported file will
+display nested blocks representing the entire model architecture.
+
 Use [Netron](https://github.com/lutzroeder/netron) to view the exported model. You should see
 something like this:
```
laggui, line 59: If you want to provide additional info, we should merge with the existing paragraph below. Some information is repeated.

laggui (changes requested):
Makes more sense now! Steps flow naturally.

Two minor comments, then will be good to go!

PR #3083, opened 2025-04-25
Make is_contiguous check common

laggui:
### Checklist

### Related Issues/PRs

Following [this comment](https://github.com/tracel-ai/burn/pull/3077#discussion_r2060270547) in #3077

### Changes

Migrated `is_contiguous` to common (including minor fix).

### Testing

Also added a unit test for shape `[3, 1]` and strides `[1, 1]` which would previously be marked as not contiguous.

### DIFF
--- a/crates/burn-common/src/lib.rs
+++ b/crates/burn-common/src/lib.rs
@@ -28,9 +28,9 @@ pub mod tensor {
     /// Check if the current tensor is contiguous.
     ///
     /// A tensor is considered contiguous if its elements are stored in memory
-    /// such that the strides are in strictly decreasing order, and the stride at
-    /// position `k` is equal to the product of the shapes of all dimensions greater
-    /// than `k`. Axes with a shape of 1 are ignored.
+    /// such that the strides are in non-increasing order, and the stride at position
+    /// `k` is equal to the product of the shapes of all dimensions greater than `k`.
+    /// Axes with a shape of 1 are ignored.
     pub fn is_contiguous(shape: &[usize], strides: &[usize]) -> bool {
         if shape.is_empty() {
             return true;
--- a/crates/burn-common/src/lib.rs
+++ b/crates/burn-common/src/lib.rs
@@ -25,6 +25,8 @@ pub mod parallel;
 
 /// Tensor utilities.
 pub mod tensor {
+    use alloc::vec::Vec;
+
     /// Check if the current tensor is contiguous.
     ///
     /// A tensor is considered contiguous if its elements are stored in memory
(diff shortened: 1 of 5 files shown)


crates/burn-common/src/lib.rs
```diff
@@ -22,3 +22,44 @@ pub mod network;
...
+            }
+
+            if i > 0 {
+                if current_num_elems_shape != stride {
+                    return false;
+                }
+
+                if prev_stride > stride {
+                    return false;
+                }
```
laggui, line 54: Previously checked for `prev_stride >= stride` but equal adjacent strides are still valid.

See added test case for tensor shape `[3, 1]` with strides `[1, 1]`, which would previously fail.

crates/burn-common/src/lib.rs
```diff
@@ -22,3 +22,44 @@ pub mod network;
...
+
+/// Tensor utilities.
+pub mod tensor {
+    /// Check if the current tensor is contiguous.
+    ///
+    /// A tensor is considered contiguous if its elements are stored in memory
+    /// such that the strides are in non-increasing order, and the stride at position
+    /// `k` is equal to the product of the shapes of all dimensions greater than `k`.
+    /// Axes with a shape of 1 are ignored.
+    pub fn is_contiguous(shape: &[usize], strides: &[usize]) -> bool {
```
laggui, line 34: I also think the check could be simplified to

```rust
pub fn is_contiguous(shape: &[usize], strides: &[usize]) -> bool {
    if shape.is_empty() {
        return true;
    }

    let mut expected_stride = 1;

    for (&shape, &stride) in shape.iter().zip(strides).rev() {
        if shape == 1 {
            continue;
        }

        if stride != expected_stride {
            return false;
        }

        expected_stride *= shape;
    }

    true
}
```

But leaving this here as a note and not a required change.

nathanielsimard: I would prefer that strategy actually, we could even create a function that could be used to create the strides AND be used for the checks:

```rust
pub fn contiguous_strides(shape: &[usize]) -> impl Iterator<Item = usize> {
    let mut current = 1;
    shape
            .iter()
            .rev()
            .map(|val| {
                current *= val;
                current
            })
            .rev()
}
```
```rust
pub fn is_contiguous(shape: &[usize], strides: &[usize]) -> bool { 
    if shape.is_empty() {
         return true;
     }
     
     for (i, expected) in contiguous_strides(shape).enumerate() {
         if expected != strides[i] {
             return false;
         }
     }
     
     true
```

nathanielsimard (changes requested):

crates/burn-common/src/lib.rs
laggui: The double `.rev()` cancel out for lazy iterator operations. Also, the first computed stride value should be 1.

I changed the suggested `contiguous_strides` implementation slightly to match.

nathanielsimard (approved):

PR #3090, opened 2025-04-27
[Perf] Convolution migration to NHWC

wingertge:
### Checklist

### Related Issues/PRs

Requires https://github.com/tracel-ai/cubecl/pull/646 to get merged first

### Changes

Reworks im2col and direct convolution to use NHWC layout, making the 2D forward pass completely NHWC. This saves unnecessary transpositions and is a prerequisite for eventually moving weights to NHWC overall. It also improves performance.

### Testing

All tests pass with both the specific algorithms and autotune enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2027,9 +2027,9 @@ dependencies = [
 
 [[package]]
 name = "deunicode"
-version = "1.6.1"
+version = "1.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc55fe0d1f6c107595572ec8b107c0999bb1a2e0b75e37429a4fb0d6474a0e7d"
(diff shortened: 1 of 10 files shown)


laggui (approved):
Ran the tests locally for cuda and wgpu/vulkan. Got a couple of failures on vulkan with f16 for convolutions, but it looks like they are also failing on main. So something broke since the 0.17 release.

With the failing tests I found a print statement for into data but that will be fixed in #3114.

Also linux-std runner keeps running out of disk space with all the updates we're doing recently. I think I'm just gonna disable the caching on this runner, more of an annoyance than anything else!

Sorry for the rant 😅 just happened to stumble upon multiple issues while testing the changes..

TL;DR: this PR looks good to me!

PR #3107, opened 2025-04-28
Port {numpy,pytorch}.meshgrid() to burn_tensor::grid

crutcher:
### Checklist

### Changes

- create `burn::grid` module.
- add `meshgrid` family of functions:
  - `meshgrid_compat`
  - `meshgrid`
  - `meshgrid_dense`
  - `meshgrid_sparse`

### Testing

Unit tests added

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -361,6 +361,12 @@ strategies.
 | `activation::softplus(tensor, beta)`             | `nn.functional.softplus(tensor, beta)`             |
 | `activation::tanh(tensor)`                       | `nn.functional.tanh(tensor)`                       |
 
+## Grid Functions
+
+| Burn API                        | PyTorch Equivalent                      |
+|---------------------------------|-----------------------------------------|
+| `grid::meshgrid_dense(tensors)` | `torch.meshgrid(tensors, indexing="ij") |
+
 ## Displaying Tensor Details
 
 Burn provides flexible options for displaying tensor information, allowing you to control the level
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -363,9 +363,10 @@ strategies.
 
 ## Grid Functions
 
-| Burn API                        | PyTorch Equivalent                      |
-|---------------------------------|-----------------------------------------|
-| `grid::meshgrid_dense(tensors)` | `torch.meshgrid(tensors, indexing="ij") |
+| Burn API                                           | PyTorch Equivalent                      |
+|----------------------------------------------------|-----------------------------------------|
+| `grid::meshgrid(tensors, GridIndexing::Matrix)`    | `torch.meshgrid(tensors, indexing="ij") |
+| `grid::meshgrid(tensors, GridIndexing::Cartesian)` | `torch.meshgrid(tensors, indexing="xy") |
 
 ## Displaying Tensor Details
 
--- a/crates/burn-tensor/src/tensor/grid/meshgrid.rs
+++ b/crates/burn-tensor/src/tensor/grid/meshgrid.rs
@@ -38,6 +38,7 @@ where
 {
     let options = options.into();
     let swap_dims = options.indexing == GridIndexing::Cartesian && N > 1;
+    let dense = options.sparsity == GridSparsity::Dense;
 
     let grid_shape: [usize; N] = tensors
         .iter()
@@ -46,7 +47,7 @@ where
         .try_into()
         .unwrap();
 
-    let result = tensors
+    tensors
         .iter()
         .enumerate()
         .map(|(i, tensor)| {
@@ -56,18 +57,16 @@ where
             // Reshape the tensor to have singleton dimensions in all but the i-th dimension
             let mut tensor = tensor.clone().reshape(coord_tensor_shape);
 
-            if options.sparsity == GridSparsity::Dense {
+            if dense {
                 tensor = tensor.expand(grid_shape);
             }
-
             if swap_dims {
-                // Swap the first two dimensions for "xy" / CartesianIndexing
                 tensor = tensor.swap_dims(0, 1);
             }
 
             tensor
         })
-        .collect::<Vec<_>>();
-
-    result.try_into().unwrap()
+        .collect::<Vec<_>>()
+        .try_into()
+        .unwrap()
 }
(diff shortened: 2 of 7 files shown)


crates/burn-tensor/src/tensor/grid/meshgrid.rs
```diff
@@ -0,0 +1,188 @@
...
+/// * In `Dense` mode, output tensors will be expanded to the full grid shape.
+///
+/// Equivalent to ``meshgrid_compat(tensors, sparsity, GridCompatIndexing::CartesianIndexing)``.
+///
+/// Users who need compatibility with the legacy `"xy"` indexing mode from
+/// NumPy and PyTorch should use `meshgrid_compat`.
+///
+/// See:
+///  - https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html
+///  - https://pytorch.org/docs/stable/generated/torch.meshgrid.html
```
laggui, line 22: If we support both "xy" and "ij" indexing, we can omit the "legacy". It also seems that this is not actually legacy behavior, but rather a pytorch API default discrepancy with the numpy implementation.

So I would remove the mention of legacy.

Also, we can simply point to the numpy reference.

crates/burn-tensor/src/tensor/grid/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/// Legacy compatibility enum for grid indexing modes.
+///
+/// NumPy (and by copying, PyTorch) used an indexing mode which swapped the first two dimensions,
+/// added to simplify some graphics and plotting tasks. As it broke the natural order of the
+/// dimensions, the behavior was flagged, and migration plans are in-flight in both libraries
+/// to make the natural ordering the default.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum GridCompatIndexing {
```
laggui, line 22: Same comment regarding "legacy" mention. The docstring can be simplified.

We can also rename the enum to `GridIndexing`.

crates/burn-tensor/src/tensor/grid/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Dimensions are in the same order as the cardinality of the inputs.
+    /// Equivalent to "ij" indexing in NumPy and PyTorch.
+    MatrixIndexing,
+
+    /// The first two dimensions are swapped.
+    /// Equivalent to "xy" indexing in NumPy and PyTorch.
+    CartesianIndexing,
```
laggui, line 29: Since the enum name is already suffixed to indicate `Indexing`, what do you think about removing the suffix in the variants?

burn-book/src/building-blocks/tensor.md
```diff
@@ -361,6 +361,12 @@ strategies.
 | `activation::softplus(tensor, beta)`             | `nn.functional.softplus(tensor, beta)`             |
 | `activation::tanh(tensor)`                       | `nn.functional.tanh(tensor)`                       |
 
+## Grid Functions
+
+| Burn API                        | PyTorch Equivalent                      |
+|---------------------------------|-----------------------------------------|
+| `grid::meshgrid_dense(tensors)` | `torch.meshgrid(tensors, indexing="ij") |
```
laggui, line 368: Will change with the suggested modifications.

e.g.,
```
grid::meshgrid(tensors, GridIndexing::Matrix)
```

laggui (changes requested):
Very clean and clearly documented!

I have a suggestion w.r.t the proposed API though. I think we can simplify this and have a single `meshgrid` function (no more dense, sparse, etc.).

While there is no overloading, we can get similar behavior via traits implementation.

I would use a `MeshgridOptions` struct and implement some defaults. Something like this:

```rust
#[derive(new, Default)]
pub struct MeshgridOptions {
    sparsity: GridSparsity,
    indexing: GridIndexing,
}

impl Default for GridSparsity {
    fn default() -> Self {
        GridSparsity::Dense
    }
}
impl Default for GridIndexing {
    fn default() -> Self {
        GridIndexing::Matrix
    }
}

pub fn meshgrid<B: Backend, const N: usize, K, O>(
    tensors: &[Tensor<B, 1, K>; N],
    options: O,
) -> [Tensor<B, N, K>; N]
where
    K: BasicOps<B>,
    O: Into<MeshgridOptions>,
{
    let options = options.into();
    // current meshgrid_compat implementation here
}

impl From<GridSparsity> for MeshgridOptions {
    fn from(value: GridSparsity) -> Self {
        MeshgridOptions {
            sparsity: value,
            ..Default::default()
        }
    }
}

impl From<GridIndexing> for MeshgridOptions {
    fn from(value: GridIndexing) -> Self {
        MeshgridOptions {
            indexing: value,
            ..Default::default()
        }
    }
}
```

Example usage:
```rust
// Default
let out = meshgrid(&[x, y, z], MeshgridOptions::default());
// Explicit indexing
let out = meshgrid(&[x, y, z], GridIndexing::Cartesian);
// Explicit sparisty
let out = meshgrid(&[x, y, z], GridSparsity::Sparse);
// Specify both
let out = meshgrid(
    &[x, y, z],
    MeshgridOptions::new(GridSparsity::Sparse, GridIndexing::Cartesian),
);
```

Let me know what you think as well!

crates/burn-tensor/src/tensor/grid/meshgrid.rs
crutcher: The docs for both numpy and pytorch mention moving away from the "xy" mode; but ok.

crutcher:
I like the options approach; that's a useful pattern. I'll update

crates/burn-tensor/src/tensor/grid/meshgrid.rs
laggui: Perhaps I missed this 🤔  I can only see the pytorch warning to follow the default numpy behavior (default cartesian vs matrix).

Can you point me to the correct section of the docs?

burn-book/src/building-blocks/tensor.md
crutcher: done.

crates/burn-tensor/src/tensor/grid/mod.rs
crutcher: done

crates/burn-tensor/src/tensor/grid/mod.rs
crutcher: done.

crates/burn-tensor/src/tensor/grid/meshgrid.rs
crutcher: just stripped it out.

crutcher:
Pretty sure this is cleaned up.

laggui (approved):
Very clean. That's much nicer.

Thank you!

PR #3108, opened 2025-04-28
Enable macro-backtrace for run-checks.sh

crutcher:
### Checklist

- [na] Made sure the book is up to date with changes in this PR.

### Changes

Enabled `macro-backtrace` via `RUSTC_BOOTSTRAP` mode.

### Testing

`./run-checks.sh all`

### DIFF
--- a/run-checks.sh
+++ b/run-checks.sh
@@ -19,4 +19,6 @@ set -e
 # If no `environment` value has been passed.
 exec_env=${1:-all}
 
-cargo xtask --execution-environment "$exec_env" validate
+RUSTC_BOOTSTRAP=1 \
+RUSTC_FLAGS="-Zmacro-backtrace" \
+  cargo xtask --execution-environment "$exec_env" validate


laggui:
Just came across your PR @crutcher 

Thanks for the suggestion! We were actually going to remove the scripts entirely (and this suggestion made me realize we still hadn't..).

The shell and powershell scripts were added before we migrated to xtask, now everything can be executed directly via a cargo command.

The suggestion is still helpful I think! So I've added a tip in both the contributor book and PR template. See #3118 for example, and let me know what you think 🙂

laggui (commented):
Just came across your PR @crutcher 

Thanks for the suggestion! We were actually going to remove the scripts entirely (and this suggestion made me realize we still hadn't..).

The shell and powershell scripts were added before we migrated to xtask, now everything can be executed directly via a cargo command.

The suggestion is still helpful I think! So I've added a tip in both the contributor book and PR template. See #3118 for example, and let me know what you think 🙂

laggui:
Just merged #3118. Will wait a bit before closing this since the scripts no longer exist, just to see if you have any further comments 🙂

PR #3109, opened 2025-04-28
Orthogonal initialization

dymat:
### Checklist

### Related Issues/PRs
No related Issues/PRs. 

### Changes
Some models need to be initialized with orthogonal weights. Pytorch provides that kind of initialization scheme ([torch.nn.init.orthogonal_](https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.orthogonal_)). This PR implements orthogonal weight initialization for burn.

### Testing
This PR comes with four tests checking the correctness of the calculations and that initialization itself is working.

### DIFF
--- a/crates/burn-core/src/nn/initializer.rs
+++ b/crates/burn-core/src/nn/initializer.rs
@@ -544,7 +544,6 @@ mod tests {
             .assert_within_range(-0.00001..0.00001);
     }
 
-
     #[test]
     fn initializer_orthogonal_correct() {
         TB::seed(0);
@@ -557,8 +556,13 @@ mod tests {
             .init([size, size], &Default::default())
             .into_value();
         let eye = Tensor::<TB, 2>::eye(size, &Default::default());
-        let diff: f32 = (eye - q.clone().transpose().matmul(q.clone()).round()).sum().into_scalar();
-        assert!(-0.00001 < diff && diff < 0.00001, "Expected Q.T @ Q to be close to identiy matrix.");
+        let diff: f32 = (eye - q.clone().transpose().matmul(q.clone()).round())
+            .sum()
+            .into_scalar();
+        assert!(
+            -0.00001 < diff && diff < 0.00001,
+            "Expected Q.T @ Q to be close to identiy matrix."
+        );
     }
 
     #[test]
--- a/crates/burn-core/src/nn/initializer.rs
+++ b/crates/burn-core/src/nn/initializer.rs
@@ -555,6 +555,7 @@ mod tests {
             .init([size, size], &Default::default())
             .into_value();
         let eye = Tensor::<TB, 2>::eye(size, &Default::default());
+
         // Q.T @ Q should be close to identity matrix
         q.clone()
             .transpose()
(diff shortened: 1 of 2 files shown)


crates/burn-core/src/nn/initializer.rs
```diff
@@ -436,4 +522,90 @@ mod tests {
...
+            .init([size, size], &Default::default())
+            .into_value();
+        let eye = Tensor::<TB, 2>::eye(size, &Default::default());
+        let diff: f32 = (eye - q.clone().transpose().matmul(q.clone()).round())
+            .sum()
+            .into_scalar();
+        assert!(
+            -0.00001 < diff && diff < 0.00001,
+            "Expected Q.T @ Q to be close to identity matrix."
+        );
```
laggui, line 565: Same here 🙂 

```suggestion
        // Q.T @ Q should be close to identity matrix
        q.clone()
            .transpose()
            .matmul(q)
            .into_data()
            .assert_approx_eq::<FT>(&eye.into_data(), Tolerance::rel_abs(1e-5, 1e-5));
```

laggui (changes requested):
Sorry for the delayed review!

Not familiar with orthogonal initialization, but the implementation LGTM.

Just a minor comment for the equality tests

/edit: also missed a small part regarding prelude usage

crates/burn-core/src/nn/initializer.rs
```diff
@@ -196,6 +237,51 @@ fn normal_draw<B: Backend, const D: usize, S: Into<Shape>>(
...
+    device: &B::Device,
+) -> (Tensor<B, 2>, Tensor<B, 2>) {
+    // Calculate the QR decomposition using Gram-Schmidt-process: https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process
+
+    let [m, n] = a.clone().dims();
+    let mut q = Tensor::<B, 2>::zeros([m, n], device);
+    let mut r = Tensor::<B, 2>::zeros([n, n], device);
+
+    for j in 0..n {
+        let mut v: Tensor<B, 1> = a.clone().slice(burn::prelude::s![.., j..=j]).squeeze(1);
```
laggui, line 251: Think we should use `burn::tensor::s!` instead of using the top-level prelude internally

crates/burn-core/src/nn/initializer.rs
```diff
@@ -196,6 +237,51 @@ fn normal_draw<B: Backend, const D: usize, S: Into<Shape>>(
...
+
+        for i in 0..j {
+            let q_i: Tensor<B, 1> = q.clone().slice(burn::prelude::s![.., i..=i]).squeeze(1);
```
laggui, line 254: Same comment for `burn::tensor::s!`

dymat: Thanks for your suggestions, @laggui 

I agree using burn::tensor::s!. I'll change that soon.

dymat:
@laggui, I guess the PR is ready now. Please excuse the wrong commit message in 7ba55f2.

laggui (approved):
Thanks for the addition 🙂 

No worries about the commit messages, they're squashed when merging anyways

PR #3123, opened 2025-04-30
Fix deadlock by really properly not capturing locks

ArthurBrussee:
Depends on https://github.com/tracel-ai/cubecl/pull/657

### DIFF
--- a/crates/burn-fusion/src/server.rs
+++ b/crates/burn-fusion/src/server.rs
@@ -3,7 +3,8 @@ use crate::{
     stream::{MultiStream, StreamId, execution::Operation},
 };
 use burn_ir::{HandleContainer, OperationIr, TensorId, TensorIr};
-use std::{future::Future, sync::Arc};
+use burn_tensor::TensorData;
+use std::sync::Arc;
 
 pub struct FusionServer<R: FusionRuntime> {
     streams: MultiStream<R>,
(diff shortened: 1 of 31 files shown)


crates/burn-fusion/src/server.rs
```diff
@@ -59,14 +59,13 @@ where
         &mut self,
         tensor: TensorIr,
         id: StreamId,
-    ) -> impl Future<Output = burn_tensor::TensorData> + 'static
+    ) -> impl Future<Output = TensorData> + Send + use<R, B>
```
nathanielsimard, line 62: I'm not familiar witht the `use<>` notation. Maybe we could create a type wrapper with a proper doc that says why we need it? Even if we're boxing the future, I don't think that's a big deal.

```rust
type CustomFuture<O, D> = Box<dyn Future<Output = O> + Send + use <D>;

fn stuff() -> CustomFuture<TensorData, (R, B)> {
    todo!();
}
```

ArthurBrussee: Yeah `use<>` is fairly new. It tells the compiler what lifetimes to include in the opaque return type. By default in the 2024 edition, all lifetimes of the input are captured. We want to say that these should NOT be captured, hence `use<>`.

Since that depends on the function inputs, you can't make that into a type alias.

If we're boxing then we don't need this. I do use for trait functions as use<> in a trait isn't stable yet (see DynFut). Could do this everywhere, but just would be a shame to need boxing for backends that don't even have any work to do for a into_data() (eg. ndarray)

nathanielsimard (approved):

PR #3127, opened 2025-04-30
Add scalar tensor operations

ArthurBrussee:
While tensor <op> scalar generally already existed, scalar <op> tensor didn't.

This is especially annoying for operations like 1.0 / tensor or a common case like (1.0 - tensor).

Due to Rust's orphan rules these sadly can't be implemented for every E: ElementConversion, but manually implementing them for primitives seems to work fine!

### DIFF
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -2,7 +2,6 @@ use alloc::vec::Vec;
 
 use crate::{alloc::borrow::ToOwned, cast::ToElement};
 
-use crate::TensorPrimitive;
 use crate::{
     BasicOps, Bool, Distribution, Element, ElementConversion, Float, Int, Shape, Tensor,
     TensorKind,
@@ -11,6 +10,7 @@ use crate::{
     check::TensorCheck,
     ops::{Device, IntTensor},
 };
+use crate::{TensorData, TensorPrimitive};
 
 /// Default RTOL value for `is_close` and `all_close`.
 pub const DEFAULT_RTOL: f64 = 1e-5;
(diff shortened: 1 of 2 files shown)


crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -4365,140 +4365,210 @@ impl<B: Backend> Numeric<B> for Float {
...
+            impl<const D: usize, B: Backend, K: Numeric<B>> core::ops::Div<Tensor<B, D, K>> for $t
+            where
+                K::Elem: Element,
+            {
+                type Output = Tensor<B, D, K>;
+
+                fn div(self, tensor: Tensor<B, D, K>) -> Self::Output {
+                    let data = TensorData::new(alloc::vec![self], [1]);
+                    let numerator = Tensor::<B, D, K>::from_data(data, &tensor.device()).unsqueeze();
+                    Tensor::div(numerator, tensor)
```
laggui, line 4494: What about using recip and scalar multiply instead?

```rust
tensor.recip().mul_scalar(self)
```

laggui (commented):
That's going to be a lot nicer to write out!

One minor comment for the scalar-tensor div, otherwise LGTM

antimora:
Probably we should mention this in https://burn.dev/burn-book/building-blocks/tensor.html

crates/burn-tensor/src/tensor/api/numeric.rs
ArthurBrussee: recip() is only for Float tensors, but then that's probably fine! I've changed it to use that now and implemented it for f32 and f64

ArthurBrussee:
> Probably we should mention this in https://burn.dev/burn-book/building-blocks/tensor.html

I've added a few lines. Didn't add notes that adding and multiplication are symmetric as it just read a bit redundant, but lmk!

crates/burn-tensor/src/tensor/api/numeric.rs
laggui: Yeah I think it makes more sense for fractions anyway. We could always add int support back if warranted but it could be ambiguous.

laggui (approved):
Thanks for filling out the docs as well!

PR #3131, opened 2025-05-01
Add burn::linalg::{vector_norm,l2_norm}

crutcher:
### Checklist

### Changes

Added `burn::linalg::{vector_norm,l2_norm}`

### Testing

Testgen tests.


crates/burn-tensor/src/tensor/linalg/vector_norm.rs
```diff
@@ -0,0 +1,147 @@
...
+{
+    match p {
+        f64::INFINITY => x.max_abs_dim(dim),
+        f64::NEG_INFINITY => x.abs().min_dim(dim),
+        0.0 => {
+            // Count non-zero elements
+            // TODO: Bool should have sum() support.
+            x.zeros_like()
+                .mask_fill(x.not_equal_elem(0.0), 1.0)
+                .sum_dim(dim)
```
laggui, line 44: For boolean sum, I'm assuming you mean for booleans treated as integers?

It should be rather straightforward in this case.
```rust
x.not_equal_elem(0.0).int().sum_dim(dim)
```

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
```diff
@@ -0,0 +1,147 @@
...
+///
+/// - `p = f64::INFINITY`: The max absolute value.
+/// - `p = f64::NEG_INFINITY`: The min absolute value.
+/// - `p = 0.0`: The count of non-zero elements.
+/// - `p = 1.0`: The sum of absolute values.
+/// - Otherwise, ``sum(abs(x)^p)^(1/p)``
+///
+/// See:
+/// - https://pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html
+/// - https://numpy.org/devdocs/reference/generated/numpy.linalg.vector_norm.html
```
laggui, line 17: Currently pointing to the devdocs instead of stable docs

crates/burn-tensor/src/tests/linalg/vector_norm.rs
```diff
@@ -0,0 +1,111 @@
...
+    #[test]
+    fn test_l2_norm() {
+        let x = TestTensor::<2>::from([[1., 2.], [3., 4.]]);
+
+        linalg::vector_norm(x.clone(), 2.0, 0)
+            .into_data()
+            .assert_approx_eq(
+                &TestTensor::<2>::from([[3.1622776601683795, 4.47213595499958]]).into_data(),
+                Tolerance::<f64>::default(),
+            );
```
laggui, line 94: Related to the failing macos test (precision / tolerance issue).

We should compare the results in the same float element type that it was computed (i.e., the backend float elem type).

Most of the backends use f32, so we can probably set a relative tolerance of 1e-5 like most other tests

```rust
type FT = FloatElem<TestBackend>;

let tolerance = Tolerance::relative(1e-5);
linalg::vector_norm(x.clone(), 2.0, 0)
            .into_data()
            .assert_approx_eq::<FT>(
                &TestTensor::<2>::from([[3.1622776601683795, 4.47213595499958]]).into_data(),
                tolerance,
            );
```

and use the same `tolerance` for the other tests below.

laggui (commented):
That's a nice addition! Good start to the linalg module 🙂 

A couple minor comments below.

Also, not sure about all the redefined functions for typical norms like `l0_norm`, `l1_norm`, etc. We could probably remove those and simply add the definitions to the docstring 🤔

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
crutcher: But then I need to convert to whatever K is

crutcher: `x.not_equal_elem(0.0).int().sum_dim(dim)` is good, thanks; but I don't know how to get from there to datatype of the input.

laggui: Oh I just read your TODO comment and the current implementation.. and the context went out the window 😅 sorry

You're right! There's not a really nice way to do this right now.. we could introduce a better way to convert between types (`.int()`, `.float()` and `.bool()` are all tied to the tensor kind).

crutcher: I couldn't work this out with the current API; let me know if you know of anything.

crutcher:
I think there's a lot of value in having common functions defined explicitly, even if they're defined in terms of other functions. I could see renaming the +inf/-inf norm functions; if you have better names.

We could also fix the lack of `.abs_min_dim()`; and drop these; idk.

But if you think about these in terms of "how readable are callers", a *lot* of literature is written wrt l1 and l2 norms; and i think at least those should have their own alias functions.

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
laggui: No that's fine, let's leave it as is. We can come back to this once we actually have a more natural conversion path between these types.

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
```diff
@@ -0,0 +1,144 @@
...
+    match p {
+        f64::INFINITY => x.max_abs_dim(dim),
+        f64::NEG_INFINITY => x.abs().min_dim(dim),
+        0.0 => x
+            .zeros_like()
+            .mask_fill(x.not_equal_elem(0.0), 1.0)
+            .sum_dim(dim),
+        1.0 => x.abs().sum_dim(dim),
+        _ => x.abs().powf_scalar(p).sum_dim(dim).powf_scalar(1.0 / p),
+    }
```
laggui, line 45: This match on an f64 is a bit awkward.. could be improved with an enum

laggui (commented):
> But if you think about these in terms of "how readable are callers", a lot of literature is written wrt l1 and l2 norms; and i think at least those should have their own alias functions.

For readability and explicitness, I agree. But the negative norm names are a bit awkward indeed, and the match on the f64 as well.

What do you think about an enum?

Could be something along the lines of..

```rust
pub enum Norm {
    L0,
    L1,
    L2,
    LInf,
    LNegInf,
    Lp(f64),
}
```

For ease of use, we could accept `p: N` where `N: Into<Norm>`.

So one could specify `Norm::L0` and other explicit norms. But also if we have `From` implementations we could have something more seamless. Just a quick example:

```rust
impl From<i32> for Norm {
    fn from(value: i32) -> Self {
        match value {
            0 => Norm::L0,
            1 => Norm::L1,
            2 => Norm::L2,
            v => Norm::Lp(v as f64)
        }
    }
}
```
could probably be added for i64, f32 and f64 for better type inference.

Lmk what you think 🙂 

Otherwise, the actual implementation LGTM. Just a matter of defining the correct interface.

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
crutcher: I'd like to point out that the cases (expect for 0.0) are ... mostly just improved numeric formulas for the base case.

laggui: Yes I know :) the comment was more about the intent and representation. Matching on f64 values for this feels somewhat brittle.

I suggested an enum because I think it more clearly express the different cases. Thought it was also aligned with your explicit function definitions for common norms used.

In addition, this pattern would allow a couple different ways to specify them (if we implement `From` for base integer and float types)
```rust
vector_norm(x, Norm::L2, dim)
vector_norm(x, 2, dim)
vector_norm(x, 2.0, dim)
```

Also, if we eventually add matrix norms it would follow a similar pattern with its own enum (example variant: frobenius).

crutcher:
Please re-review.

I *think* I've addressed the type bounds to the best of my ability.

crutcher:
Discovered, worked-around https://github.com/tracel-ai/burn/issues/3139

crutcher:
Wait on: https://github.com/tracel-ai/burn/pull/3140

crutcher:
Converted into a draft until the above bug / fix goes in.

crutcher:
Should this be moved to `burn-core`?

laggui:
> Should this be moved to `burn-core`?

I actually had a flash this weekend. Most of the "tensor ops" are just being thrown in `burn-tensor` but I think we should have a separation. Right now, `burn-core` probably makes the most sense. In the future, if we have enough ops we could have a `burn-linalg` crate (re-exported via a feature flag).

crutcher:
Done!
I've updated this now that the `.max_dim()` fix is in.

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
```diff
@@ -0,0 +1,225 @@
...
+///
+/// # Arguments
+///
+/// * `x` - The input tensor.
+/// * `dim` - The dimension to compute the norm over.
+///
+/// # Returns
+///
+/// The L:INFINITY norm of the input tensor.
+pub fn max_abs_norm<B: Backend, const D: usize, K>(
```
laggui, line 197: Much better naming than l_inf_norm 😅 

Is it called the max abs norm or just max norm typically? My linear algebra terminology is a little dusty.

crates/burn-tensor/src/tests/linalg/vector_norm.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[test]
+    fn test_max_min_abs() {
+        let x = TestTensor::<2>::from([[1., 2.], [3., 4.]]);
+
+        linalg::vector_norm(x.clone(), linalg::Norm::LInf, 0)
+            .into_data()
+            .assert_eq(&TestTensor::<2>::from([[3., 4.]]).into_data(), true);
+        linalg::max_abs_norm(x.clone(), 0)
+            .into_data()
+            .assert_eq(&TestTensor::<2>::from([[3., 4.]]).into_data(), true);
```
laggui, line 17: Technically this would test pretty much the same code path, but I have no strong opposition against the additional test here.

laggui (commented):
Just a minor comment regarding naming for the max/min norms to make sure.

Otherwise should be good to go! Thanks for going through the rounds in this PR.

crutcher:
I missed the previous discussion about the ops; I'll try and adjust

crutcher (commented):
will adjust

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
laggui: Yes of course but there are often other names used interchangeably (e.g., euclidean norm or l2 norm).

Max norm seems to [also be used](https://en.wikipedia.org/wiki/Norm_(mathematics)#Maximum_norm_(special_case_of:_infinity_norm,_uniform_norm,_or_supremum_norm)), but not sure if it is that common. Anyway, either `max_abs_norm`, `max_norm` or `infinity_norm` (without the `L` prefix) should be suitable.

laggui (approved):

PR #3042, opened 2025-04-17
Refactor quantization scheme

maxtremblay:
### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [ ] Made sure the book is up to date with changes in this PR.

### Changes

Refactor `QuantizationScheme` to a struct and add two new parameters. The quantization accumulator precision and the output mode. 

### Testing

No new tests, but the existing ones are still succeeding.

PR #3136, opened 2025-05-02
Migrate to new cubecl multi tensor handle changes

wingertge:
### Checklist

### Related Issues/PRs

Update PR for https://github.com/tracel-ai/cubecl/pull/661

### Changes

Updates to new matmul signature, fixes some bugs in quantize kernel

### Testing

Tests pass


crates/burn-cubecl/src/kernel/quantization/quantize.rs
```diff
@@ -65,64 +103,137 @@ fn quantize_per_tensor_symmetric_int8_kernel(
...
+
+    let out_layout = strided_layout(&output);
+
+    // Force vectorization to process 4 quantized values packed for 1 output value
+    let line_size = max_line_size(&tensor);
+    let cube_dim = CubeDim::default();
+    let cube_count = calculate_cube_count_elemwise(num_elems / line_size as usize, cube_dim);
+
+    match scheme {
+        QuantizationScheme::PerTensor(QuantizationMode::Symmetric, QuantizationType::QInt8) => {
```
laggui, line 207: This will change with #3042 as the quantization scheme will be more like a config struct.

But no worries, we'll make sure that this PR is merged first so you can avoid the possible conflict hell 😅

laggui (commented):
Thanks for fixing the current limitations with the quantization kernels!

Just to give you a bit more (historical) context (might help fill in the blanks lol):

The current implementation dates back to a long time ago, it hasn't changed much since the first draft that came around the initial release of CubeCL. It was also written specifically with wgpu in mind which only supported f32 and u32 (hence the packing / unpacking) because the CUDA backend was barely a thing around this time lol.

I was just discussing this recently with Max and the plan was to add int8 support. This is exactly in line with your changes, and the additions in cubecl will make this a lot easier. So really, thanks a ton 🙏 

> (from the linked cubecl PR)  
> Adds multi-tensor allocations so quantization params can be allocated along with strided tensors. The multiple tensors are allocated into the same buffer, but with separate strides.

This is awesome! It makes a lot more sense. The initial works actually dealt with multiple handles on the burn-ir side, but it was messy. And there was no easy way to do this at the time for cubecl. Tensor operations and transformations (e.g., layout changes) will be easier to handle with such a separation, especially as we add more quantization levels (e.g., per-block).

TL;DR: thanks for contribution(s) 😄 the changes LGTM

crates/burn-cubecl/src/kernel/quantization/quantize.rs
wingertge: Ah it sems like the align for copy is incorrect for offsets, we need to align to the min storage buffer offset instead. I'll have to fix that in cubecl.

nathanielsimard (approved):
Late review, but I think we can move the `QParams` struct to `burn-tensor`, since it's used in both `burn-cubecl` and `burn-cubecl-fusion`.

Issue #3139, opened 2025-05-02
.max_dim/.min_dim broken for Int tensors when gathering anything other than the last dimension.


crutcher:
**Describe the bug**

`.max_dim(dim)` is broken for *Int* tensors, but not *Float*, when dim is not the last dimension.

This also affects `.min_dim()`; probably generic issue with the common gather.

**To Reproduce**
This will backtrace when called on an Int tensor (in NdArray, unsure about others):
```
pub fn max_abs_norm<B: Backend, const D: usize, K>(
    x: Tensor<B, D, K>,
    dim: usize,
) -> Tensor<B, D, K>
where
    K: BasicOps<B> + Numeric<B>,
{
    x.abs().max_dim(dim)
}
```

This will not:
```
pub fn max_abs_norm<B: Backend, const D: usize, K>(
    x: Tensor<B, D, K>,
    dim: usize,
) -> Tensor<B, D, K>
where
    K: BasicOps<B> + Numeric<B>,
{
    let mut x = x.abs();
    let last = x.dims().len() - 1;
    if last != dim {
        x = x.swap_dims(last, dim);
    }
    x = x.max_dim(last);
    if last != dim {
        x = x.swap_dims(last, dim);
    }
    x
}
```

**Expected behavior**
dimension order agnostic behavior.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 24.04

laggui:
Closed via #3140

PR #3140, opened 2025-05-02
WIP fix to NdArray int_max_dim bug

crutcher:
See: https://github.com/tracel-ai/burn/issues/3139

This is a WIP PR; I need to add tests which wouldn't have passed before this fix.


crutcher:
See: https://github.com/tracel-ai/burn/pull/3143

laggui (approved):
Good catch 😅 Thanks for also adding a unit test for the regression.

NdArray is the only backend that doesn't overwrite the implementation, so this copy/pasta mistake never came up until now it seems!

PR #3141, opened 2025-05-02
Fix burn-cubecl with autotune disabled

wingertge:
Small fix to feature gates that were causing compile errors with autotune disabled. This was caused by the module refactor in convolution.


laggui (approved):
Thanks for the fix!

Linking to [the discord message](https://discord.com/channels/1038839012602941528/1091796857996451942/1367972028799389778) that mentions the issue

PR #3126, opened 2025-04-30
Refactor: Move op_configuration.rs from burn-import to onnx-ir

antimora:
### Checklist

- [x] Confirmed that the `run-checks all` script has been executed.
- [x] Ensured the book is updated with changes in this PR.

### Related Issues/PRs

Fixes #3091

### Changes

1. Moved `op_configuration.rs` from `burn-import` to `onnx-ir`. The `onnx-ir` module does not depend on any data types from `burn-import` or `burn`. Configurations are now organized by node type.
2. Added unit tests for the moved node configurations.
3. Relocated rank inference functions to node folders in `onnx-ir` and added corresponding unit tests.
4. Updated the developer book and README files.

### Testing

1. Existing integration tests are passing.
2. New unit tests have been added.

PR #3144, opened 2025-05-03
break down `test_onnx.rs` into test subdirectories

antimora:
### Checklist

### Related Issues/PRs

#3126

### Changes

1. Broke down test_onnx.rs into test subdirectories
2. Updated `crates/burn-import/onnx-tests/README.md` 

### Testing

All existing tests are passing


crates/burn-import/onnx-tests/README.md
```diff
@@ -1,49 +1,199 @@
...
+
+Your script should:
+
+- Import the necessary PyTorch modules
+- Define a model that uses your operator
+- Generate test inputs
+- Export the model to ONNX format
+- Run the model on test inputs and print the output
+
+Example structure:
```
laggui, line 60: If we now provide an example template for exporting onnx via pytorch, perhaps we should briefly mention that in some cases the graph can be manually defined with the node(s) to test using onnx directly. We can simply point to an existing test (e.g., `gather_1d_idx`).

laggui (approved):
Sorry for the late review!

There has been a lot of activity recently, so it's hard to keep up. Bigger PRs (even if mostly documentation or restructuring) take more time to get to.

One minor comment, otherwise LGTM. I'll approve in advance nonetheless.

I don't think the current structure was a dealbreaker for now, but splitting into separate files makes it more accessible that's for sure.

Thanks 🙂

crates/burn-import/onnx-tests/README.md
antimora: Thanks for the input. I added the example and subsection.

laggui: Oh, the embedded example was not required. Just pointing to a test with such code would have sufficed.

But I guess this is even more clear and upfront 👍

Issue #2761, opened 2025-02-01
Save a labels.txt in the artifact directory for the ImageFolderDataset (could maybe apply to other datasets)


aurelien-clu:
### Feature description

Training a model should also provide a way to know the label name and ordering without knowledge of the dataset. It would also prevent potential errors if we default to use the labels provided inside the artifact directory next to the model weights.

### Feature motivation

I am training a model that classifies kanjis with many classes (~5600) and no obvious label ordering so it could be prone to error between a prediction [0.1, ..., 0.5] and knowing that this means:

```
一: 0.1
...
龠: 0.5
```

AshAnand34:
Can you clarify on where the artifact directory would be for this scenario?

aurelien-clu:
When we train a model we can get the following directory structure:

```
├── checkpoint
├── config.json
├── experiment.log
└── train
    └── epoch-1
        ├── Accuracy.log
        ├── Loss.log
        └── Recall.log
```

Having the following to me would be nice:

```
├── checkpoint
├── config.json
├── experiment.log
├── labels.txt/json/yaml
└── train
    └── epoch-1
        ├── Accuracy.log
        ├── Loss.log
        └── Recall.log

```

As for the file format, whether it is txt (line 0 = label at index 0), YAML or JSON anything would be good I think.

PR #3145, opened 2025-05-03
Save labels.txt in artifact for imagefolderdataset

AshAnand34:
### Checklist

> [!TIP]
> Want more detailed macro error diagnostics? This is especially useful for debugging tensor-related tests:
>
> ```bash
> RUSTC_BOOTSTRAP=1 RUSTFLAGS="-Zmacro-backtrace" cargo run-checks
> ```

### Related Issues/PRs

#2761

### Changes

The issue requested adding functionality to save a labels.txt file in the artifact directory when using the ImageFolderDataset. This feature helps preserve the mapping between class indices and their corresponding label names, which is particularly valuable for:
1. Maintaining label order and names without requiring the original dataset structure
2. Reducing errors in model prediction interpretation
3. Simplifying the mapping of numerical predictions back to their corresponding labels

The solution implemented:
- Modified the with_items method in ImageFolderDataset to automatically create a labels.txt file
- Added functionality to write class names to the file in index order
- Set the default artifact directory to /tmp/burn-dataset
- Added proper error handling for file operations
- Updated the documentation to reflect this new feature

### Testing
- Existing test suite for ImageFolderDataset
- Error handling with directory creation failures and invalid paths.

### DIFF
--- a/crates/burn-dataset/Cargo.toml
+++ b/crates/burn-dataset/Cargo.toml
@@ -49,9 +49,10 @@ rand = { workspace = true, features = ["std"] }
 rmp-serde = { workspace = true }
 rusqlite = { workspace = true, optional = true }
 sanitize-filename = { workspace = true }
-serde = { workspace = true, features = ["std", "derive"] }
-serde_json = { workspace = true, features = ["std"] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
 serde_rusqlite = { workspace = true, optional = true }
+serde_yml.workspace = true
 strum = { workspace = true }
 tempfile = { workspace = true }
 thiserror = { workspace = true }
--- a/crates/burn-dataset/src/audio/speech_commands.rs
+++ b/crates/burn-dataset/src/audio/speech_commands.rs
@@ -1,5 +1,5 @@
 use crate::{
-    Dataset, HuggingfaceDatasetLoader, SqliteDataset,
+    Dataset, HuggingfaceDatasetLoader, LabeledDataset, SqliteDataset,
     transform::{Mapper, MapperDataset},
 };
 
@@ -156,6 +156,12 @@ impl Dataset<SpeechItem> for SpeechCommandsDataset {
 
     fn len(&self) -> usize {
         self.dataset.len()
+    }
+}
+
+impl LabeledDataset<SpeechItem, String> for SpeechCommandsDataset {
+    fn get_label(&self, index: usize) -> Option<String> {
+        self.dataset.get(index).map(|item| item.label.to_string())
     }
 }
 
--- a/crates/burn-dataset/src/audio/speech_commands.rs
+++ b/crates/burn-dataset/src/audio/speech_commands.rs
@@ -97,6 +97,17 @@ pub struct SpeechItem {
     pub label: SpeechCommandClass,
 }
 
+/// Format for saving labels
+#[derive(Debug, Clone, Copy)]
+pub enum LabelFormat {
+    /// Text format with one label per line
+    Txt,
+    /// JSON format with an array of labels
+    Json,
+    /// YAML format with an array of labels
+    Yaml,
+}
+
 /// Speech Commands dataset from Huggingface v0.02.
 /// See [Speech Commands dataset](https://huggingface.co/datasets/speech_commands).
 ///
--- a/crates/burn-dataset/src/dataset/base.rs
+++ b/crates/burn-dataset/src/dataset/base.rs
@@ -24,6 +24,27 @@ pub trait Dataset<I>: Send + Sync {
     }
 }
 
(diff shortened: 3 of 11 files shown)


crates/burn-dataset/Cargo.toml
```diff
@@ -49,9 +49,10 @@ rand = { workspace = true, features = ["std"] }
 rmp-serde = { workspace = true }
 rusqlite = { workspace = true, optional = true }
 sanitize-filename = { workspace = true }
-serde = { workspace = true, features = ["std", "derive"] }
-serde_json = { workspace = true, features = ["std"] }
```
laggui, line 53: We should keep pointing to workspace dependencies

crates/burn-dataset/src/audio/speech_commands.rs
```diff
@@ -97,6 +97,17 @@ pub struct SpeechItem {
...
+/// Format for saving labels
+#[derive(Debug, Clone, Copy)]
+pub enum LabelFormat {
+    /// Text format with one label per line
+    Txt,
+    /// JSON format with an array of labels
+    Json,
+    /// YAML format with an array of labels
+    Yaml,
+}
```
laggui, line 109: While the flexibility is nice, I would stick to a single format. We can remove the enum and just save a text file.

crates/burn-dataset/src/dataset/dataframe.rs
```diff
@@ -248,6 +248,101 @@ where
     extractor.fields
 }
 
+/// Dataset implementation for Polars DataFrame with labels
+///
+/// This struct provides a way to access data and labels from a Polars DataFrame
+/// as if it were a Dataset of type I with labels of type L.
+pub struct LabeledDataframeDataset<I, L> {
```
laggui, line 255: Not sure this specialization is required 🤔 a dataframe dataset can be used by the user to get the labels from the appropriate field for each item.

Maybe I am missing something. Can you provide more details if you think this should be added?

crates/burn-dataset/src/dataset/in_memory.rs
```diff
@@ -6,20 +6,38 @@ use std::{
...
-use crate::Dataset;
+use crate::{Dataset, LabeledDataset};
 
 /// Dataset where all items are stored in ram.
 pub struct InMemDataset<I> {
     items: Vec<I>,
 }
 
+/// Dataset where all items and their labels are stored in ram.
+pub struct LabeledInMemDataset<I, L> {
```
laggui, line 17: Similar argument here. The `InMemDataset` is generic over the items. So if items contain a label/class, it is entirely up to the user 🤔

crates/burn-dataset/src/dataset/base.rs
```diff
@@ -24,6 +24,66 @@ pub trait Dataset<I>: Send + Sync {
...
+    Txt,
+    /// JSON format with an array of labels
+    Json,
+    /// YAML format with an array of labels
+    Yaml,
+}
+
+/// The labeled dataset trait defines a dataset that contains labeled data.
+/// It extends the basic Dataset trait with functionality to handle labels.
+pub trait LabeledDataset<I, L>: Dataset<I>
```
laggui, line 40: Trait might be overkill

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -691,6 +692,29 @@ impl ImageFolderDataset {
...
+                .map_err(|e| ImageLoaderError::IOError(e.to_string()))?;
+        }
+
+        // Write labels to file
+        let mut labels_file = std::fs::File::create(&labels_path)
+            .map_err(|e| ImageLoaderError::IOError(e.to_string()))?;
+
+        // Sort classes by index to ensure consistent ordering
+        let mut sorted_classes: Vec<_> = classes_map.iter().collect();
+        sorted_classes.sort_by_key(|(_, idx)| *idx);
```
laggui, line 711: This would also sort classes provided by the user given that public constructors also go through this path.

When parsing folder names automatically, we already sort the names. But I don't think we should sort classes provided by the user. This could lead to unexpected behavior.

crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -691,6 +692,29 @@ impl ImageFolderDataset {
...
+
+        for (class_name, _) in sorted_classes {
+            writeln!(labels_file, "{}", class_name)
+                .map_err(|e| ImageLoaderError::IOError(e.to_string()))?;
+        }
+
```
laggui, line 717: I think I would move this to a separate method instead, e.g., `dataset.save_classes(path)` where one could pass the current artifact/experiment dir.

It is not always required, but having an opt-in method could make more sense.

laggui (changes requested):
Thanks for taking on this request!

I have a couple of comments

PR #3147, opened 2025-05-04
Extract Linear.forward to nn::functional::linear

crutcher:
### Checklist

- [n/a] Made sure the book is up to date with changes in this PR.

### Changes

This begins extracting `nn.<Module>.forward` code into common utility libs under `nn::functional`, by moving the `Linear` impl to `nn::functional::linear`.

Also provides `linear_pytorch` (name?) so there's a direct impl of the transpose semantics under pytorch.

### Testing

Tested 1d, forward (with,without) bias.


antimora:
I am trying to learn the context. 

Is there a particular reason for adding linear_pytorch? Wouldn't be it be easier to transpose the weight when importing like we do when [importing pt file here](https://github.com/tracel-ai/burn/blob/5a437b066e28e1eff9c7a3b8229776ba7c963bac/crates/burn-import/src/pytorch/adapter.rs#L24)

Also there was a recent conversation about storing different layouts possibly: https://discord.com/channels/1038839012602941528/1091796857996451942/1365333083716325437

crutcher:
@antimora I'll do my best to un-pack this without writing a novel; and I can hop on discord and talk about it this week as well.

I've been working on translating some complex vision models (such as SWIN) to Burn; and they have complex kernels that frequently rely on more primitive ideas. That's my proximate immediate drive.

In doing so, I've been bumping up against the awkwardness of not having access to the equivalent of the `torch.nn.functional` tree of non-module functions. Sure, many of these components could be forced to be sub-modules; but it gets tedious. It's also easier to test the pure functions in isolation; they don't have configs, they don't have Params, etc. If we also consider all the non-AI applications for tensors, having a robust library of common functions which aren't tied to a training/weight management scheme seems attractive.

In some places, it's clear that Burn made different choices than pytorch about how some of these functions work. Now, in this specific case, I've always hated the `x A^T + b` approach (which mainly seems to exist because someone at some point felt weird about `Ax + b`, which is the standard linear algebra form.

But!

Compat functions and option-prefill functions are essentially free.
- they are generally stripped from programs which don't use them.
- they are generally optimized out of programs which do use them.
- they memorialize a contract (and the tests for that contract) in a way which is more durable than a side-note in a document.

If you want the `linear_pytorch` removed (or renamed?); I'm ok with that; but I do generally think that Burn would benefit from more "this is probably what you want" wrapper/alias functions for implementors.

The question of if loading or processing of the Linear layer should be aligned with, or modally aligned with, the weight structure from other nn layers is more a strategy question that should depend upon how Burn does all the common layers, and I'm not trying to pick a dog in that fight at all.

But I do feel strongly about the utility of the `nn::functional` tree; and I'm willing to do a lot of work to flesh it out.

antimora:
@crutcher Thank you for explaining your motivation.

I am +1 on nn::functional module but I am not sure there is a value for `linear_pytorch` even with a different name. It's so simple in functionality and yet specific to PyTorch's way.

BTW, there is a performance hit when transposing (even changing only strides) the weight matrix. @louisfd did a benchmark test to see if there was a difference when I suggested a special flag for linear: https://github.com/tracel-ai/burn/pull/1084. This was done before CubeCL, so it could be different now.

crutcher:
I've removed the `linear_pytorch` function.

crates/burn-core/src/nn/functional/linear.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn test_linear_1d() {
+        let weight =
+            Tensor::<TestBackend, 2>::from_data([[1.0, 2.0], [3.0, 4.0]], &Default::default());
+
+        let x = Tensor::<TestBackend, 1>::from_data([1.0, 2.0], &Default::default());
+
+        linear(x.clone(), weight.clone(), None)
+            .into_data()
+            .assert_eq(
+                &Tensor::<TestBackend, 1>::from_data([7.0, 10.0], &Default::default()).into_data(),
```
antimora, line 62: You can simplify this as:

```rust
&TensorData::from([7.0, 10.0]),
```
More on this:
https://burn.dev/docs/burn/prelude/struct.TensorData.html#method.from-3

crates/burn-core/src/nn/functional/linear.rs
```diff
@@ -0,0 +1,114 @@
...
+        );
+
+        linear(x.clone(), weight.clone(), None)
+            .into_data()
+            .assert_eq(
+                &Tensor::<TestBackend, 3>::from_data(
+                    [[[7.0, 10.0], [15.0, 22.0]], [[-7.0, -10.0], [-15.0, -22.0]]],
+                    &Default::default(),
+                )
+                .into_data(),
```
antimora, line 84: Same as my previous comment using TensorData

crutcher:
@antimora I've addressed your comments.

@nathanielsimard are you sure about the move?

my thoughts on this are that fleshing out a full functional api will end up refactoring many components in burn-core; and they'll be in a split logic situation. i'm also concerned that, as we see in most std libs, many of the algorithms will be best expressed in terms of *other* algorithms; and sequencing becomes messy.

Is there any guide to how the layer cake should be stacked? Should there maybe be an intermediate layer?

Maybe:
- burn-tensor
- burn-fn-core
- burn-core (or burn-ai-core)?

I'll do what you want with this PR; but I'm concerned that we're walking into a tangle if I continue building this kind of PR.

crutcher:
@nathanielsimard I've moved the tree

crates/burn-tensor/src/tensor/nn/functional/linear.rs
laggui, line 1: I think the intention was to add the linear to the `ModuleOps` trait, similar to other "functional" modules are defined.

Then, it can be used via `module::linear` (just like conv2d can be used via `module::conv2d`).

This is defined in [`tensor/ops/modules/base.rs`](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/ops/modules/base.rs#L261).

crates/burn-tensor/src/tests/nn/functional/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod vector_norm;
```
laggui, line 1: Wrong module name from the other PR? 😅

crates/burn-tensor/src/tests/nn/functional/vector_norm.rs
laggui, line 1: With the move to the module ops, the tests would be defined in `tests/module/linear.rs`

crates/burn-tensor/src/tests/nn/functional/mod.rs
crutcher: Loly; yes o_O

crutcher:
Lifted to ModuleOps

crates/burn-tensor/src/tensor/ops/modules/base.rs
```diff
@@ -764,6 +765,39 @@ pub trait ModuleOps<B: Backend> {
...
+    /// ```math
+    /// y = x @ weight^T + [bias]
+    /// ```
+    fn linear<const D: usize>(
+        input: Tensor<B, D>,
+        weight: Tensor<B, 2>,
+        bias: Option<Tensor<B, 1>>,
+    ) -> Tensor<B, D> {
+        linear(input, weight, bias)
+    }
```
laggui, line 800: Similar to other module ops defined in the trait, this should operate on `FloatTensor<B>` (backend float tensor primitive).

So the actual default implementation should be here, and the `linear` function in the `modules/` module should simply call `B::linear` (along with dealing with the underlying primitives).

This also allows backends to implement the module op and possible optimize it (though it won't be the case now).

crates/burn-tensor/src/tensor/ops/modules/linear.rs
```diff
@@ -0,0 +1,44 @@
...
+        return linear::<B, 2>(input.unsqueeze(), weight, bias).flatten(0, 1);
+    }
+
+    let weight = weight.unsqueeze();
+    let output = input.matmul(weight);
+    match bias {
+        Some(bias) => output + bias.unsqueeze(),
+        None => output,
+    }
+}
```
laggui, line 44: This should call the backend impl.

See for example conv2d

https://github.com/tracel-ai/burn/blob/741c4b1263564b2cfeb7aa7d400d8d0e0ac8b71b/crates/burn-tensor/src/tensor/module.rs#L46-L68

Also, it can probably be added to the same file as linked above where all public module ops currently live.

laggui (changes requested):
The implementation is almost there!

See my comments below. But for this small stuff, I don't mind moving things around to complete your work myself. Especially since `unsqueeze` and `flatten` are only available via the public tensor API, not backend ops. So we just need to do `B::float_reshape` instead.

Will do this tomorrow to complete it if you'd like 🙂 otherwise feel free to complete yourself!

laggui:
Can take care of the last mile delivery tomorrow!

Had a couple things to investigate today (notably that dreaded macos CI issue).. so time ran out before I could make it here 😅

crates/burn-core/src/nn/linear.rs
```diff
@@ -65,24 +67,24 @@ impl LinearConfig {
...
-
-        match bias {
-            Some(bias) => output + bias,
-            None => output,
-        }
+        linear(
+            input,
+            self.weight.val(),
+            self.bias.as_ref().map(|b| b.val()),
+        )
```
laggui, line 87: The forward now simply calls the `module::linear` method.

crates/burn-tensor/src/tensor/module.rs
```diff
@@ -357,3 +357,40 @@ where
...
+    input: Tensor<B, D>,
+    weight: Tensor<B, 2>,
+    bias: Option<Tensor<B, 1>>,
+) -> Tensor<B, D> {
+    Tensor::new(TensorPrimitive::Float(B::linear(
+        input.primitive.tensor(),
+        weight.primitive.tensor(),
+        bias.map(|b| b.primitive.tensor()),
+    )))
+}
```
laggui, line 396: The module function which operates on `Tensor` (not backend primitives) is the one that calls the module ops trait implementation. `.tensor()` will simply return the float tensor primitive (floats can be represented via quantization too (WIP), so they're wrapped in an enum).

The result is simply wrapped in the `Tensor` struct.

laggui (commented):
@crutcher just completed the changes, see my comments below if you are curious.

Also cleaned up the vector norm tests a bit. When I ran the tests locally the lp norm test failed with f16 (tolerance difference), so I went over them after setting the half precision tolerance.

antimora (approved):
LGTM

laggui (approved):
Approving my partial contribution 😆

crutcher:
@laggui thanks for fixing this!

PR #3152, opened 2025-05-05
Replace some powf->powi

ArthurBrussee:
### Changes

Replace some powf -> powi calls. This isn't any faster currently, but could be in the future if powi is specialized.


laggui (approved):
Think the macos runner isn't picking up the job because other runners are already hanging.. looks like it might have something to do with the new cubecl config.

PR #3118, opened 2025-04-29
Fix/run checks

laggui:
### Checklist

- [ ] Confirmed that `cargo run-checks` command has been executed.
- [ ] Made sure the book is up to date with changes in this PR.

> [!TIP]
> Want more detailed macro error diagnostics? This is especially useful for debugging tensor-related tests:
>
> ```bash
> RUSTC_BOOTSTRAP=1 RUSTFLAGS="-Zmacro-backtrace" cargo run-checks
> ```

### Related Issues/PRs

_Provide links to relevant issues and dependent PRs._

### Changes

_Summarize the problem being addressed and your solution._

### Testing

_Describe how these changes have been tested._

PR #3159, opened 2025-05-06
Moved debugging tip from PR to contributor section

antimora:
The tip on the PR makes the template busy. Most submitters do not need macro debugging. It's better if we live the tip in the contributor book section that we can refer people.

### Related Issues/PRs

#3118

### Changes

1. Moved the tip from PR template to contributor section

### Testing

Rendered


antimora:
> The tip is already in the contributor book
> 
> https://github.com/tracel-ai/burn/blob/main/contributor-book/src/getting-started/setting-up-the-environment.md
> 
> As well as in the CONTRIBUTING.md (see linked PR #3118 which you reviewed).
> 
> The addition to the template was related to the suggestion in #3108 to add it by default.
> 
> I'm don't have a strong opinion on the addition or removal from the template. I think it's easy to ignore visually.

It's hard to ignore it =D Let's remove it.

Should I remove from the book, or leave it for troubleshooting so we can direct people?

laggui:
> Should I remove from the book, or leave it for troubleshooting so we can direct people?

Like I said, this is already in the book. So we can remove your addition to the book and link to the [other section](https://github.com/tracel-ai/burn/blob/main/contributor-book/src/getting-started/setting-up-the-environment.md).

We can remove it from the PR template as you suggest though and refer people to this section. Probably not useful to most people.

laggui (approved):

PR #3162, opened 2025-05-07
Update fusion.rs

ta2ozg:
refactoring fusion codebase

Instead of option <HashMap<...> was used directly checks like is_none are no longer necessary.
Removed register_inner function, integrated its content directly into client function.
Added expect() error message instead of unwrap() because Rust gives more meaningful panic message if downcast fails.
Mutex is still spin::Mutex but i guess i can switch to parking_lot::Mutex if you wish.


laggui (commented):
Thanks for taking the time to contribute!

But this is not particularly impactful.. and it doesn't compile 😅

PR #3167, opened 2025-05-07
Fix: fusion multi-block scalar index sharing

### LINKED ISSUE #3142: Loaded model.mpk gives different results across backends
**Describe the bug**
When I load a model from disk, I get very different results than at the end of my training loop. It seems that my model produces very different results across backends, including between Autodiff\<B\> backends and their respective inner backends. The differences are far greater than would be expected from floating point precision. Not all layers are causing this, but the issue seems to be with the combination of `Conv2D` followed by some `reshape` and `swap_dims` to apply `LayerNorm`.

This issue might be related to https://github.com/tracel-ai/burn/issues/2901, when I came up with a minimal example for that it was also the same part of my model, combining convolution + LayerNorm.

I've not much idea about the underlying difference here, i.e. if this is a difference between the weights being loaded, or the architecture of model.

**To Reproduce**
Minimum reproducible example @ https://paste.rs/E2ZUC.rs

1. `cargo run` - writes out a saved model and config to `/tmp/burn_backend_mre`
2. `cargo test` - some combinations pass, other fails, see screenshot

Image

The result above is on v0.17, but running on the current main branch gives the same results. These tests all pass on v0.16, excluding the `Cuda` tests which aren't supported by v0.16. These are all with `fusion`, haven't tried without.

**Expected behavior**
All tests in the above example should pass.

**Desktop**
 - OS: Ubuntu 24.04.1 LTS
 - GPU: GTX 1060 and RTX 3060

nathanielsimard:
We had a bug in the reduce fallback code when both fuse on read and fuse on write had scalars. The scalar positions need to be updated between the two kernel launch in the fallback procedure.

fix #3142


laggui (approved):
Validated against the original code in linked issue. This also fixes the `test_lp_norm` on main.

Somehow the isolated MWE from https://github.com/tracel-ai/burn/issues/3142#issuecomment-2854929622 is still incorrect with fusion

```
failures:

---- tests::test_cuda_swap_dims stdout ----

thread 'tests::test_cuda_swap_dims' panicked at src\main.rs:44:14:
Tensors are not approx eq:
  => Position 0: 0.017461976 != 0.03273773
     diff (rel = +3.04e-1, abs = +1.53e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
  => Position 1: 0.013624389 != 0.0021862322
     diff (rel = +7.23e-1, abs = +1.14e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
  => Position 2: 0.087606415 != 0.025062565
     diff (rel = +5.55e-1, abs = +6.25e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
  => Position 3: 0.091444 != 0.15015031
     diff (rel = +2.43e-1, abs = +5.87e-2), tol (rel = +1.00e-5, abs = +1.00e-5)

---- tests::test_wgpu_swap_dims stdout ----

thread 'tests::test_wgpu_swap_dims' panicked at src\main.rs:44:14:
Tensors are not approx eq:
  => Position 0: 0.01746198 != 0.03273773
     diff (rel = +3.04e-1, abs = +1.53e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
  => Position 1: 0.013624394 != 0.0021862322
     diff (rel = +7.23e-1, abs = +1.14e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
  => Position 2: 0.08760644 != 0.025062565
     diff (rel = +5.55e-1, abs = +6.25e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
  => Position 3: 0.09144402 != 0.15015031
     diff (rel = +2.43e-1, abs = +5.87e-2), tol (rel = +1.00e-5, abs = +1.00e-5)
```

Has to be a different issue, so approving this fix for the scalars.

PR #3188, opened 2025-05-12
Feat/autotune optional

nathanielsimard:
CubeCl update from https://github.com/tracel-ai/cubecl/pull/685

Also fix some fusion problem regarding broadcasting:
1. Broadcasting between multiple blocks isn't supported, and we should ensure the return value of a block match the width of the block.
2. Broadcasted output that don't match the shape of fusion shound't be taken as reference layout.


louisfd (approved):
LGTM

PR #3189, opened 2025-05-13
Burn-remote to_device function

Cielbird:


crates/burn/Cargo.toml
```diff
@@ -131,3 +132,4 @@ burn-remote = { path = "../burn-remote", version = "0.18.0", default-features =
 burn-router = { path = "../burn-router", version = "0.18.0", default-features = false, optional = true }
 burn-tch = { path = "../burn-tch", version = "0.18.0", optional = true }
 burn-wgpu = { path = "../burn-wgpu", version = "0.18.0", optional = true, default-features = false }
+burn-ir = { path = "../burn-ir", version = "0.18.0", default-features = false }
```
nathanielsimard, line 135: This should be optional, since it's not usefull all of the time.

PR #3200, opened 2025-05-15
Chain lint inheritance [was: Disable new default clippy tests]

crutcher:
### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed.
- [x] Made sure the book is up to date with changes in this PR.

### Changes

- Wired up clippy defaults to override several new default warnings that are broken in the codebase.
- Chained crates to depend upon the workspace lint configs.

PR #3206, opened 2025-05-16
[WIP] Run clippy fix on clippy::pedantic errors

crutcher:
### Checklist

### Changes

This sits on top of https://github.com/tracel-ai/burn/pull/3200

I enabled `pedantic = "deny"`; ran `cargo clippy --fix`; then disabled and cleaned up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1498,7 +1498,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.6.0"
-source = "git+https://github.com/tracel-ai/cubecl?rev=cfcbc082c8fbe4285e8d01d587132ff6720295fa#cfcbc082c8fbe4285e8d01d587132ff6720295fa"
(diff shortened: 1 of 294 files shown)


laggui (commented):
This mostly seems to fix more aggressive clippy lints along with typos / formatting in docstrings. Not entirely against, though I feel some lints are probably more opiniated than anything so we might have to be careful.

One thing I am not sure about is the usage of `#[must_use]`. I also think its usage should be informed to indicate specific types that are "important" (e.g., results that should not be ignored, or types with custom drop implementation). In other words, this is more advice to the usage than anything. I did not review the whole set of changes, but the few additions of `must_use` seemed to make sense at first glance.

@crutcher did you actually want to go through with some of these suggestions, or were you mostly fiddling with the lints?

crutcher:
I think most of these suggestions are good ideas; and I'd like to move forward by turning on *specific* extra lints, one at a time, and fixing them.

antimora:
I am closing this for now. Please free to reopen. Too many conflicts to resolve.

PR #3215, opened 2025-05-20
[CI] Disable GCP runners in test workflow

syl20bnr:
We need to use a different execution model to create and delete runners for our public repos. In the meantime we disable the GCP runners.


louisfd (approved):
LGTM

PR #3225, opened 2025-05-25
Add support onnx ceil & round

tye-singwa:
### Checklist

### Related Issues/PRs

#1714

### Changes

Adding ceil & round onnx support

### Testing

Added tests


laggui:
> @laggui I want to confirm if this autodiff fix is correct.
> 
> I have had a pending comment that I forgot to submit

Yeah the autodiff implementation previously returned the result with floor for the output (though the gradients were fine). Slipped through the cracks it seems.

PR #3231, opened 2025-05-26
Feature/inference timing utils

### LINKED ISSUE #3226: Best practices of calculating inference time in Burn
Can anyone tell me what are the best practices of calculating inference time in Burn?

I know that one has to run some dummy examples through the model to do a GPU warm-up.

There is one more thing. When I was working on PyTorch, we used some specific functions to perform synchronization between the GPU and CPU in order to calculate accurate inference time but i do not know how can we do the same thing in Burn? Can anyone guide me in this regard? I did not find yet any such thing in documentation.

chojuninengu:
# Add inference timing utilities with GPU synchronization

This PR adds utilities for measuring inference time in Burn with proper GPU synchronization, addressing issue #3226.

## Features
- `TimingConfig` struct for configuring warmup and measurement iterations
- `TimingResults` struct for comprehensive timing statistics (mean, std dev, min, max)
- `InferenceMeasurable` trait for models that can be measured
- `measure_inference_time` function that handles proper GPU synchronization
- Example showing usage with different backends (CPU, WGPU, CUDA, Metal)

## Usage Example
```rust
let device = WgpuDevice::default();
let model = YourModel::new(&device);
let input = your_input.to_device(&device);

let config = TimingConfig {
    num_warmup: 10,      // Number of warm-up iterations
    num_iterations: 100,  // Number of measured iterations
};

let results = measure_inference_time(&model, &input, &device, config);
println!("Mean inference time: {:?}", results.mean);
```

Fixes #3226

### DIFF
--- a/examples/inference_timing.rs
+++ b/examples/inference_timing.rs
@@ -1,0 +1,80 @@
+use burn::{
+    backend::{
+        libtorch::{LibTorch, LibTorchDevice},
+        wgpu::{Wgpu, WgpuDevice},
+        ndarray::{NdArray, NdArrayDevice},
+        Backend,
+    },
+    tensor::Tensor,
+};
+use burn_example::{utils::{TimingConfig, measure_inference_time, InferenceMeasurable}};
+
+// Example model that performs matrix multiplication
+struct ExampleModel<B: Backend> {
+    weights: Tensor<B, 2>,
+}
+
+impl<B: Backend> ExampleModel<B> {
+    fn new(device: &B::Device) -> Self {
+        let weights = Tensor::ones([100, 100]).to_device(device);
+        Self { weights }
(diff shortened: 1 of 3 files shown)


nathanielsimard (commented):
We already have a benchmark utility with proper synchronization that works with all backend. It's reexported from CubeCl, since that's the same logic:

https://github.com/tracel-ai/cubecl/blob/main/crates/cubecl-common/src/benchmark.rs

All benchmarks also use that trait: https://github.com/tracel-ai/burn-bench/blob/main/backend-comparison/benches/matmul.rs

I'm not against having more options, but I wouldn't introduce another way of measuring time.

Though I still think it's valuable to maybe improve it a bit, and write more docs around that.

PR #3239, opened 2025-05-28
Updating/cubecl: matmul test macros and partition

louisfd:


laggui (commented):
Cubecl update included in #3240

PR #3245, opened 2025-06-01
Add support onnx group norm

tye-singwa:
### Checklist

### Related Issues/PRs

#1714

### Changes

Adding onnx support for GroupNormalization node

### Testing

Added tests


crates/burn-import/src/burn/node/group_norm.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
+        let input = scope.tensor_use_owned(&self.input, node_position);
+        let output = &self.output.name;
+        let field = &self.field.name;
+
+        // TODO: handle self.full_precision
+        quote! {
+            let #output = self.#field.forward(#input);
+        }
```
laggui: Should be fairly straightforward to support now that we have `tensor.cast(dtype)` for float tensors.

Something like
```rust
let dtype = #input.dtype();
let #output = self.#field.forward(#input.cast(burn::tensor::DType::F32)).cast(dtype);
```

tye-singwa: Thanks for checking, added this to onnx layer norm too.

laggui: Ha, didn't realize it was also missing for layer norm. Thank you!

PR #3251, opened 2025-06-02
Adding image dimensions to ImageDatasetItem

catch-twenty-two:
### Checklist

### Related Issues/PRs

Discord discussion here:
https://discord.com/channels/1038839012602941528/1357819409267425341/1359178264530256045

### Changes

This adds original image dimensions to the ImageDataset struct items so that image tensors can be shaped correctly when necessary

### Testing

Added unit tests for testing image size

### DIFF
--- a/crates/burn-dataset/src/vision/image_folder.rs
+++ b/crates/burn-dataset/src/vision/image_folder.rs
@@ -98,6 +98,12 @@ pub struct ImageDatasetItem {
     /// Image as a vector with a valid image type.
     pub image: Vec<PixelDepth>,
 
+    /// Original source image width.
+    pub image_width: usize,
+
+    /// Original source image height.
+    pub image_height: usize,
+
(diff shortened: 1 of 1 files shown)


crates/burn-dataset/src/vision/image_folder.rs
```diff
@@ -98,6 +98,9 @@ pub struct ImageDatasetItem {
     /// Image as a vector with a valid image type.
     pub image: Vec<PixelDepth>,
 
+    /// Original source image dimensions.
+    pub image_dims: (u32, u32),
```
laggui: Might was well split into separate height and width fields to make it clear 🙂 could also be usize instead of u32

catch-twenty-two:
Happy to contribute! PR updated with requested changes

laggui (approved):

PR #3254, opened 2025-06-03
Update agumentation - WIP!

catch-twenty-two:
This is a work in progress, initially to come to an agreement on the Image augmentation interface. In that regard, it wont compile or work yet but should provide a good outline of what the final interface will look like.


crates/burn-dataset/Cargo.toml
```diff
@@ -30,7 +30,7 @@ __sqlite-shared = [
 dataframe = ["dep:polars"]
 
 [dependencies]
-burn-common = { path = "../burn-common", version = "0.18.0", optional = true, features = [
+burn-common = { path = "../burn-common", version = "0.17.0", optional = true, features = [
```
laggui, line 33: Should revert this change 😄

crates/burn-dataset/src/vision/image_ops.rs
```diff
@@ -0,0 +1,260 @@
...
+/// # Parameters
+/// - `img_data`: A tuple containing:
+///     - A 3D tensor representing the image, in shape `[3, H, W]` (channel-first format).
+///     - An `Option` containing a list of 1D tensors representing bounding boxes in `[x, y, w, h]` format,
+///       where `(x, y)` is the top-left corner of the box, and `w` and `h` are the width and height.
+///
+/// # Returns
+/// Returns a new tuple:
+/// - The vertically flipped image tensor.
+pub fn vertical_flip<B: Backend>(img_tnsr: Tensor<B, 3>) -> Tensor<B, 3> {
```
laggui, line 23: I would just name the argument `image` since it's a `Tensor`

crates/burn-dataset/src/vision/image_ops.rs
```diff
@@ -0,0 +1,260 @@
...
+/// # Parameters
+/// - `tensor_img`: A 3D tensor representing the image in `[3, H, W]` format (channel-first).
+/// - `contrast`: A `f32` value representing the percentage of contrast adjustment:
+///     - `0.0` leaves the image unchanged.
+///     - Positive values increase contrast (e.g., `20.0` increases by 20%).
+///     - Negative values decrease contrast (e.g., `-20.0` decreases by 20%).
+///
+/// # Returns
+/// A new tensor representing the image with adjusted contrast.
+pub fn contrast<B: Backend>(img_tnsr: Tensor<B, 3>, contrast: f32) -> Tensor<B, 3> {
```
laggui, line 57: Naming: `adjust_contrast`

crates/burn-dataset/src/vision/image_ops.rs
```diff
@@ -0,0 +1,260 @@
...
+///
+/// Increases the brightness of the image by adding the specified `value` to each pixel.
+///
+/// # Parameters
+/// - `tensor_img`: A 3D tensor representing the image in `[3, H, W]` format (channel-first).
+/// - `value`: The brightness adjustment value. Positive values brighten, negative values darken.
+///
+/// /// # Returns
+/// A new tensor representing the image with adjusted brigthness.
+pub fn brighten<B: Backend>(img_tnsr: Tensor<B, 3>, value: i32) -> Tensor<B, 3> {
```
laggui, line 131: Naming: `adjust_brightness`

crates/burn-dataset/src/vision/image_transforms.rs
```diff
@@ -0,0 +1,376 @@
...
+///
+/// Takes a `BoundingBox` and converts it into a 1D tensor.
+///
+/// # Parameters
+/// - `bbox`: The bounding box to be converted.
+///
+/// # Returns
+/// An `Option` containing the `[x, y, w, h]` tensor representing the bounding box,
+/// or `None` if the bounding box is invalid.
+pub fn bbox_as_tensor<B: Backend>(bbox: &BoundingBox) -> Tensor<B, 1> {
```
laggui, line 30: Duplicate?

crates/burn-dataset/src/vision/image_transforms.rs
```diff
@@ -0,0 +1,376 @@
...
+pub enum AnnotationTnsr<B: Backend> {
+    /// Image-level label.
+    Label(usize),
+    /// Multiple image-level labels.
+    MultiLabel(Vec<usize>),
+    /// Object bounding boxes.
+    BoundingBoxes(Vec<Tensor<B, 1>>),
+    /// Segmentation mask.
+    SegmentationMask,
+} // BM&JJ
```
laggui, line 18: For classic image transformations (whether geometrical or pixel-level), the image-level (single or multi) are not typically affected. So not sure the enum should have these variants 🤔

Also, I think I'd rename the enum.

crates/burn-dataset/src/vision/image_transforms.rs
```diff
@@ -0,0 +1,376 @@
...
+        let tnsr_data = TensorData::new(
+            image_as_vec_u8(self.image),
+            Shape::new([self.image_dims.0 as usize, self.image_dims.1 as usize, 3]),
+        );
+
+        let mut sample = Sample {
+            image: Tensor::<B, 3>::from_data(tnsr_data.convert::<B::FloatElem>(), &device)
+                .permute([2, 0, 1]),
+            annotations: vec![],
+        };
```
laggui, line 67: We'd like to expand the int tensor support to include unsigned ints. This will be especially useful for images, which can be loaded as u8. Most of the transforms could also be applied on int tensors in this case.

We have to keep in mind that not all images are u8 though, so while `MAX_PIXEL_VAL` works for rbg 8 images it is not always the case. Should probably have a max pixel val per dtype 🤔

crates/burn-dataset/src/vision/image_transforms.rs
```diff
@@ -0,0 +1,376 @@
...
+                AnnotationTnsr::SegmentationMask => todo!(),
+            }
+        }
+
+        Ok(sample)
+    }
+}
+
+/// doc
+pub struct Zoom {
```
laggui, line 152: Naming: if the operation is a zoom out (and the function is named accordingly `zoom_out`), then the transformation should be `ZoomOut`.

laggui (commented):
Sorry for the late review 😅 

Thanks for sharing the draft following our offline discussions! I think that's a great start 🙂 

General comments:
- I think this would belong in `burn-vision/src/transforms` rather than `burn-dataset`. The latter should remain focused on data loading, dataset abstractions, and I/O operations. The transformations are really image operations.
- I like the image_ops / image_transforms separation. But I would go a step further and have each transform (e.g., `ZoomOut`) implementation in its own module. I think it will make things cleaner.

crates/burn-dataset/src/vision/image_transforms.rs
catch-twenty-two: Yeah, I figured this is at least a place to start I'm guessing that the majority of people will use u8, so maybe that could be a later change? (baby steps)

laggui: All good! As long as we keep this in mind 🙂

catch-twenty-two:
Since this is being actively worked on in [bimm](https://docs.rs/bimm/latest/bimm/all.html), I don't know if I should continue to add to this.

laggui:
@catch-twenty-two that is entirely up to you 🙂 if the usage in bimm suits your needs, then I'll close this PR.

But if ever you come back around to work on this feel free to re-open!

PR #3263, opened 2025-06-06
Perf/fusion deadcode

nathanielsimard:
Added the Drop operation that can be fused, eliminating some writes in fused kernels, properly propagating dead code elimination. We could apply a dead code elimination algorithm inside of burn-fusion as well in the future to reduce even more computations that are not necessary, but this is a good start.

Uses CubeCL : https://github.com/tracel-ai/cubecl/pull/712

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,9 +158,9 @@ portable-atomic = { version = "1.11.0" }
 portable-atomic-util = { version = "0.2.4", features = ["alloc"] }
 
 ### For the main burn branch. ###
-cubecl = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "c8a0a681b67020cb70af89100af5433220c0246d" }
-cubecl-common = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "c8a0a681b67020cb70af89100af5433220c0246d" }
-cubecl-std = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "c8a0a681b67020cb70af89100af5433220c0246d" }
+cubecl = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "b34cffd711e25c1438967c19dc14376de8427df2" }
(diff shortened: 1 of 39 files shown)


Cargo.toml
```diff
@@ -158,9 +158,9 @@ portable-atomic = { version = "1.11.0" }
 portable-atomic-util = { version = "0.2.4", features = ["alloc"] }
 
 ### For the main burn branch. ###
-cubecl = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "cdf67a222ce973955fba439bb019139f5b33d40c" }
-cubecl-common = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "cdf67a222ce973955fba439bb019139f5b33d40c" }
-cubecl-std = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "cdf67a222ce973955fba439bb019139f5b33d40c" }
+cubecl = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "b34cffd711e25c1438967c19dc14376de8427df2" }
+cubecl-common = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "b34cffd711e25c1438967c19dc14376de8427df2" }
+cubecl-std = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "b34cffd711e25c1438967c19dc14376de8427df2" }
```
laggui: TODO: update rev once the linked cubecl PR is merged

crates/burn-cubecl-fusion/src/shared/builder.rs
```diff
@@ -36,6 +36,14 @@ impl OptimizationBuilder<FuseTrace> for FuseOptimizationBuilder {
         }
 
         match op {
+            OperationIr::Drop(tensor) => {
+                if self.num_ops == 0 {
+                    self.status = OptimizationStatus::Closed;
+                    return;
+                }
+
+                self.builder.builder.tag_dropped(tensor.id);
```
laggui: What about `register_dropped(tensor.id)`?

Just a naming preference though, so it's subjective.

Issue #2464, opened 2024-11-07
Add a PixelShuffle implementation like in pytorch


Rick-29:
In PyTorch there is a module called [PixelShuffle ](https://pytorch.org/docs/stable/generated/torch.nn.PixelShuffle.html).
I created as mall implementation of it that currently only supports 4D Tensors (I tried to follow the implementation format of the library) and wanted to share it to check if someday it could be added to the crate.
This is the full code:

```rs
use burn::{config::Config, module::Module, prelude::Backend, tensor::Tensor};

#[derive(Config, Debug)]
pub struct PixelShuffleConfig {
    #[config(default = "2")]
    upscale_factor: usize
}

#[derive(Module, Debug, Clone)]
pub struct PixelShuffle {
    upscale_factor: usize
}

impl PixelShuffleConfig {
    pub fn init(&self) -> PixelShuffle {
        PixelShuffle { upscale_factor: self.upscale_factor }
    }
}

impl PixelShuffle {
    pub fn forward<B: Backend>(&self, input: Tensor<B, 4>) -> Tensor<B, 4> {        
        let mut dims = input.dims();
        dims.reverse();
        let c = dims[2];
        let h = dims[1];
        let w = dims[0];
        if c % (self.upscale_factor * self.upscale_factor) != 0 {
            panic!("pixel_shuffle expects its input's 'channel' dimension to be divisible by the square of upscale_factor, but input.size(-3)={c} is not divisible by {}", self.upscale_factor * self.upscale_factor)
        }
        let oc = c / (self.upscale_factor * self.upscale_factor);
        let oh = h * self.upscale_factor;
        let ow = w * self.upscale_factor;
    
        let x = input.reshape([dims[3], oc, self.upscale_factor, self.upscale_factor, h, w]);  
        let x = x.permute([0, -5, -2, -4, -1, -3]);
        x.reshape([dims[3], oc, oh, ow])
    }
}
```
And here are some test that I made comparing the outputs with the ones from the pytorch implementation:
```rs
#[cfg(test)]
mod tests {
    use burn::backend::Wgpu;

    use super::*;

    #[test]
    fn test_pixel_shuffle() {
        let shuffle = PixelShuffle { upscale_factor: 3 };
        let tensor1 = Tensor::<Wgpu, 4>::random([1, 9, 4, 4], burn::tensor::Distribution::Default, &Default::default());
        let tensor1_shuffle = shuffle.forward(tensor1);
[...]

PR #3277, opened 2025-06-09
Add onnx SpaceToDepth / DepthToSpace

tye-singwa:
### Checklist

### Related Issues/PRs

#1714, #2930

### Changes

Adding support for SpaceToDepth / DepthToSpace onnx nodes

### Testing

Added a bunch of tests


tye-singwa:
Just noticed https://github.com/tracel-ai/burn/issues/2464, maybe it make sense to implement it as separate module in burn-core?

crates/burn-import/onnx-tests/tests/depth_to_space/depth_to_space.py
```diff
@@ -0,0 +1,77 @@
...
+    graph = onnx.helper.make_graph(
+        [depth_to_space],
+        f'DepthToSpace{mode}Model',
+        [input],
+        [output],
+    )
+
+    # Create the model
+    model = onnx.helper.make_model(
+        opset_imports=[onnx.helper.make_operatorsetid("", 21)],
```
laggui: Doesn't have a direct impact since the last version of this op is 13, but let's keep this to the version currently required (opset 16). Don't forget to regenerate/update the onnx files :)

crates/burn-import/onnx-tests/tests/depth_to_space/depth_to_space.py
```diff
@@ -0,0 +1,77 @@
...
+    # Ensure valid ONNX and save
+    onnx.checker.check_model(onnx_model)
+    onnx.save(onnx_model, file_name)
+    print(f"Finished exporting model to {file_name}")
+
+    # Output some test data for use in the test
+    print(f"Test input data:\n{repr(test_input)}")
+    print(f"Test input data shape: {test_input.shape}")
+    session = ReferenceEvaluator(file_name, verbose=1)
+    test_output, = session.run(None, {"input": test_input})
```
laggui, line 64: I was not aware that onnx added a reference evaluator! That's especially useful for these tests

crates/burn-import/onnx-tests/tests/depth_to_space/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+                    [[0.24, -1.91, -1.72], [-0.56, -1.01, 0.31]],
+                    [[-0.91, -1.41, 1.47], [-0.23, 0.07, -1.42]],
+                ],
+                [
+                    [[-0.54, 0.11, -1.15], [0.38, -0.6, -0.29]],
+                    [[-0.6, 1.85, -0.01], [-1.06, 0.82, -1.22]],
+                    [[0.21, -1.96, -1.33], [0.2, 0.74, 0.17]],
+                    [[-0.12, -0.3, -1.48], [-0.72, -0.46, 1.06]],
+                ],
+            ],
```
laggui, line 32: Probably could have used a simpler input for the test (like a range of ints) 😅 but this is fine

crates/burn-import/onnx-tests/tests/space_to_depth/space_to_depth.py
```diff
@@ -0,0 +1,67 @@
...
+    graph = onnx.helper.make_graph(
+        [space_to_depth],
+        'SpaceToDepthModel',
+        [input],
+        [output],
+    )
+
+    # Create the model
+    model = onnx.helper.make_model(
+        opset_imports=[onnx.helper.make_operatorsetid("", 21)],
```
laggui: Same

crates/onnx-ir/src/node/depth_to_space.rs
```diff
@@ -0,0 +1,172 @@
...
+        let [b, c, h, w] = shape
+            .try_into()
+            .expect("DepthToSpace: input tensor rank is not 4");
+        vec![
+            b,
+            c / (block_size * block_size),
+            h * block_size,
+            w * block_size,
+        ]
+    });
```
laggui, line 42: Not required, but since this will only be applied if the static shape is available we can keep it :)

laggui (commented):
Just a few minor comments, otherwise LGTM!

Thanks for the addition 🙏

crates/burn-import/onnx-tests/tests/depth_to_space/depth_to_space.py
tye-singwa: Thanks for checking, i've updated it to 16, regenerated files

crates/burn-import/onnx-tests/tests/space_to_depth/space_to_depth.py
tye-singwa: Fixed it to 16

crates/onnx-ir/src/node/depth_to_space.rs
tye-singwa: As i see across code, some operators require static shape, this is reason why i added it

laggui (approved):

PR #3278, opened 2025-06-10
KaTeX for Mathematical expressions in docstrings

BhavyeMathur:
### Checklist

### Related Issues/PRs

#2450 

### Changes

Embedding mathematical expressions in documentation is very important. This PR follows the instructions given [here](https://docs.rs/rustdoc-katex-demo/0.1.5/rustdoc_katex_demo/) to add KaTeX to this repository, allowing us to use LaTeX syntax to type out formulas that can be rendered on the docs.

To see this locally, run

```
RUSTDOCFLAGS="--html-in-header katex-header.html" cargo doc --no-deps --package burn-tensor
```

The docs should now be rendered like so:

image

I have also redone the documentation for all the activation functions and a few more.

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -40,7 +40,9 @@ pub fn leaky_relu<const D: usize, B: Backend>(
 /// [Gaussian Error Linear Units (GELUs)](https://arxiv.org/pdf/1606.08415v3.pdf).
 ///
 /// $$
-/// \text{GELU}\(x\) = x * \Phi\(x\)
+/// \text{GELU}\(x\)
+/// = x \cdot \Phi\(x\)
+/// = x \cdot \frac{1}{2}\left(1 + \text{erf}\left(\frac{x}{\sqrt{2}}\right)\right)
 /// $$
 ///
 /// where $\Phi\(x\)$ is the cumulative distribution function for the Gaussian distribution.
--- a/crates/burn-tensor/src/tensor/api/float.rs
+++ b/crates/burn-tensor/src/tensor/api/float.rs
@@ -110,7 +110,7 @@ where
     ///     let device = Default::default();
     ///
     ///     let tensor = Tensor::<B, 3>::from_data([0.0, -1.0, 2.0], &device);
-    ///     println!("{}", tensor.cosh()); // should be [1.0, 1.5430, 3.7621]
+    ///     println!("{}", tensor.cosh()); // [1.0, 1.5430, 3.7621]
     /// }
     /// ```
     pub fn cosh(self) -> Self {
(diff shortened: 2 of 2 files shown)


crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -23,15 +36,38 @@ pub fn leaky_relu<const D: usize, B: Backend>(
     )))
 }
 
-/// Applies the Gaussian Error Linear Units function as described in the paper [Gaussian Error Linear Units (GELUs)](https://arxiv.org/pdf/1606.08415v3.pdf).
+/// Applies the Gaussian Error Linear Units function as described in the paper
+/// [Gaussian Error Linear Units (GELUs)](https://arxiv.org/pdf/1606.08415v3.pdf).
+///
+/// $$
+/// \text{GELU}\(x\) = x * \Phi\(x\)
```
laggui: Could probably expand the definition for Phi(x), which would give

```
= x * 1/2 * [1 + erf(x / sqrt(2))]
```

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -49,56 +53,112 @@ where
...
+    ///
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::Tensor;
+    ///
+    /// fn example<B: Backend>() {
+    ///     let device = Default::default();
+    ///
+    ///     let tensor = Tensor::<B, 3>::from_data([0.0, -1.0, 2.0], &device);
+    ///     println!("{}", tensor.cosh()); // should be [1.0, 1.5430, 3.7621]
```
laggui: I think I would remove the "should be" for brevity, but also because we expect the results to be accurate. Otherwise, there is probably a bug anyway😅

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -49,56 +53,112 @@ where
...
+    ///
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::Tensor;
+    ///
+    /// fn example<B: Backend>() {
+    ///     let device = Default::default();
+    ///
+    ///     let tensor = Tensor::<B, 3>::from_data([0.0, -1.0, 2.0], &device);
+    ///     println!("{}", tensor.sinh()); // should be [0.0, -1.1752, 3.6269]
```
laggui: Same

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -49,56 +53,112 @@ where
...
+    ///
+    /// ```rust
+    /// use burn_tensor::backend::Backend;
+    /// use burn_tensor::Tensor;
+    ///
+    /// fn example<B: Backend>() {
+    ///     let device = Default::default();
+    ///
+    ///     let tensor = Tensor::<B, 3>::from_data([0.0, -1.0, 2.0], &device);
+    ///     println!("{}", tensor.sinh()); // should be [0.0, -0.7616, 0.9640]
```
laggui: Same

BhavyeMathur:
image
 
Expanded GELU definition ^

laggui (approved):
LGTM 🚀 

Thanks for improving the docs!

Issue #2132, opened 2024-08-08
ONNX import issue: `ceil_mode is not supported` in `avg_pool1d_config()`


SimonBrandner:
```
ERROR burn_import::logger: PANIC => panicked at /home/simon/.cargo/git/checkouts/burn-178c6829f420dae1/d770b1f/crates/burn-import/src/onnx/op_configuration.rs:374:9:
ceil_mode is not supported
```

Link to model: [drive.google.com/file/d/1E6fNKzGE3_JhEokgqqHDqJxwqw3UUku9/view?usp=drive_link](https://drive.google.com/file/d/1E6fNKzGE3_JhEokgqqHDqJxwqw3UUku9/view?usp=drive_link)

antimora:
This requires enhancement to Burn's average pool implementation.

mepatrick73:
This probably requires rewriting the pooling tensor operations in CubeCL. They are currently written in "pseudo-assembly" and I don't see a good reason to try and maintain that version.

mepatrick73:
I'm locking this behind #2140

raphaelmenges:
I get a similar error but for `avg_pool2d`:

> `thread 'main' (34885003) panicked at /Users/raphael/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/onnx-ir-0.19.0/src/node/avg_pool2d.rs:60:9: ceil_mode is not supported`

I guess it is the same "problem" 😬

laggui:
@raphaelmenges yepp, we do not support ceil mode for pooling in burn right now.. so any ONNX models that use this configuration will panic 🥲 

https://github.com/tracel-ai/burn/blob/89de0e13926291028817983d65c242906d3b649b/crates/burn-nn/src/modules/pool/avg_pool2d.rs#L33-L55

torsteingrindvik:
Same for LPIPS specifically for the squeeze variant, the others are ok:

```python
import lpips
import torch

def main(size=(1920, 1080)):
    (width, height) = size

    for net in ["alex", "vgg", "squeeze"]:
        print(f"Exporting {net}")

        loss_fn = lpips.LPIPS(net=net)
        loss_fn.eval()

        dummy_input_img0 = torch.randn(1, 3, height, width)
        dummy_input_img1 = torch.randn(1, 3, height, width)

        torch.onnx.export(
            loss_fn,
            (dummy_input_img0, dummy_input_img1),  # Model inputs as a tuple
[...]

PR #3279, opened 2025-06-10
Adding option for ceiling to compute size for pool output

catch-twenty-two:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/2132

### Changes

Adding a ceil option to max pool operations.

Example:

```
let maxpool2d3 = MaxPool2dConfig::new([2, 2])
    .with_strides([2, 2])
    .with_ceil(true)
    .init();
```
### Testing

Added Unit tests, 

Also tested on SSD: Single Shot MultiBox Detector model I am building, which is required to go from 

300x300->150x150->75x75->**38x38**

38x38 is where the maxpool size wasn't compatible in burn and it would resize to 37x37 instead. This update now allows it to resize properly to 38x38.

### DIFF
--- a/crates/burn-core/src/nn/pool/avg_pool1d.rs
+++ b/crates/burn-core/src/nn/pool/avg_pool1d.rs
@@ -28,6 +28,9 @@ pub struct AvgPool1dConfig {
     /// If the padding is counted in the denominator when computing the average.
     #[config(default = "true")]
     pub count_include_pad: bool,
+    /// Use ceil when calculating the output size.
+    #[config(default = "false")]
+    pub ceil: bool,
 }
 
 /// Applies a 1D avg pooling over input tensors.
@@ -52,6 +55,8 @@ pub struct AvgPool1d {
     pub padding: Ignored<PaddingConfig1d>,
     /// If the padding is counted in the denominator when computing the average.
     pub count_include_pad: bool,
+    /// Use ceil when calculating the pool size.
+    pub ceil: bool,
 }
 
 impl ModuleDisplay for AvgPool1d {
@@ -67,6 +72,7 @@ impl ModuleDisplay for AvgPool1d {
             .add("stride", &self.stride)
             .add("padding", &self.padding)
             .add("count_include_pad", &self.count_include_pad)
+            .add("ceil", &self.ceil)
             .optional()
     }
 }
@@ -82,6 +88,7 @@ impl AvgPool1dConfig {
             kernel_size: self.kernel_size,
             padding: Ignored(self.padding.clone()),
             count_include_pad: self.count_include_pad,
+            ceil: self.ceil,
         }
     }
 }
@@ -107,6 +114,7 @@ impl AvgPool1d {
             self.stride,
             padding,
             self.count_include_pad,
+            self.ceil,
         )
     }
 }
@@ -129,7 +137,7 @@ mod tests {
 
         assert_eq!(
             alloc::format!("{}", layer),
-            "AvgPool1d {kernel_size: 3, stride: 1, padding: Valid, count_include_pad: true}"
+            "AvgPool1d {kernel_size: 3, stride: 1, padding: Valid, count_include_pad: true, ceil: false}"
         );
     }
 }
--- a/crates/burn-core/src/nn/pool/max_pool1d.rs
+++ b/crates/burn-core/src/nn/pool/max_pool1d.rs
@@ -28,6 +28,9 @@ pub struct MaxPool1dConfig {
     /// The dilation.
     #[config(default = "1")]
     pub dilation: usize,
+    /// Use ceil when calculating the max pool size.
+    #[config(default = "false")]
+    pub ceil: bool,
 }
 
 /// Applies a 1D max pooling over input tensors.
@@ -44,6 +47,8 @@ pub struct MaxPool1d {
     pub padding: Ignored<PaddingConfig1d>,
     /// The dilation.
     pub dilation: usize,
+    /// Use ceil when calculating the max pool size.
+    pub ceil: bool,
 }
 
 impl ModuleDisplay for MaxPool1d {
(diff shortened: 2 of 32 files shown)


crates/burn-core/src/nn/pool/avg_pool1d.rs
```diff
@@ -52,6 +55,8 @@ pub struct AvgPool1d {
     pub padding: Ignored<PaddingConfig1d>,
     /// If the padding is counted in the denominator when computing the average.
     pub count_include_pad: bool,
+    /// Use ceil when calculating the pool size.
```
laggui, line 58: Not the pool size, but the output shape.

crates/burn-core/src/nn/pool/max_pool1d.rs
```diff
@@ -28,6 +28,9 @@ pub struct MaxPool1dConfig {
     /// The dilation.
     #[config(default = "1")]
     pub dilation: usize,
+    /// Use ceil when calculating the max pool size.
```
laggui, line 31: Same comment.

"Use ceil to compute the output shape"

crates/burn-core/src/nn/pool/max_pool1d.rs
```diff
@@ -44,6 +47,8 @@ pub struct MaxPool1d {
     pub padding: Ignored<PaddingConfig1d>,
     /// The dilation.
     pub dilation: usize,
+    /// Use ceil when calculating the max pool size.
```
laggui, line 50: Same

crates/burn-core/src/nn/pool/max_pool2d.rs
```diff
@@ -28,6 +28,9 @@ pub struct MaxPool2dConfig {
     /// The dilation.
     #[config(default = "[1, 1]")]
     pub dilation: [usize; 2],
+    /// Use ceil when calculating the max pool size.
```
laggui, line 31: Same 😅

crates/burn-core/src/nn/pool/max_pool2d.rs
```diff
@@ -44,6 +47,8 @@ pub struct MaxPool2d {
     pub padding: Ignored<PaddingConfig2d>,
     /// The dilation.
     pub dilation: [usize; 2],
+    /// Use ceil when calculating the max pool size.
```
laggui, line 50: +1

crates/burn-tensor/src/tests/module/avgpool1d.rs
laggui, line 1: Adding a ceil mode requires testing with ceil mode too 🙂

laggui (commented):
The config addition with the extra parameter makes sense.

But I don't think the implementations for all backends will be compatible without any modifications. Sure, for tch we can forward the ceil arg. But for cubecl the kernel probably needs to take the elements at the border into account.

Also, the `_ceil` arg is currently not forwarded for ndarray.

Some other comments:

crates/burn-core/src/nn/pool/avg_pool1d.rs
catch-twenty-two: > Not the pool size, but the output shape.

gotcha, I was following the naming convention of the original function...` pub fn calculate_pool_output_size`. 

https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/ops/modules/conv.rs#L68

Should this be refactored as well to match? I think an official naming convention wiki would be great since I have seen contradicting names in other documentation and code as well.

crates/burn-tensor/src/tests/module/avgpool1d.rs
catch-twenty-two: Agreed! Will add more tests

crates/burn-core/src/nn/pool/avg_pool1d.rs
catch-twenty-two: either way I'll make this change and leave the refactoring of the function name for another pr/issue if it's deemed necessary the whole suite of tests and a bunch of other things are named around pool size instead of shape as well.

catch-twenty-two:
> The config addition with the extra parameter makes sense.
> 
> But I don't think the implementations for all backends will be compatible without any modifications. Sure, for tch we can forward the ceil arg. But for cubecl the kernel probably needs to take the elements at the border into account.

I thought this was taken care of since all the backends seem to make a call to [the same function](https://github.com/catch-twenty-two/burn/blob/add-ceil-maxpool-ops/crates/burn-tensor/src/tensor/ops/modules/conv.rs#L68) (the function I actually updated) to get the shape, I may be mistaken though:

Examples:
https://github.com/catch-twenty-two/burn/blob/add-ceil-maxpool-ops/crates/burn-cubecl/src/kernel/pool/avg_pool2d.rs#L88

https://github.com/catch-twenty-two/burn/blob/add-ceil-maxpool-ops/crates/burn-fusion/src/ops/module.rs#L658

> 
> Also, the `_ceil` arg is currently not forwarded for ndarray.

Will fix, thanks!

> 
> Some other comments:

crates/burn-core/src/nn/pool/avg_pool1d.rs
laggui: "pool output size" or "pool output shape" are both equivalent, I have no preference. But "pool size" is not as accurate, and could be misinterpreted 🙂 (basically, just missing the "output" specifier for the size/shape)

laggui:
> I thought this was taken care of since all the backends seem to make a call to [the same function](https://github.com/catch-twenty-two/burn/blob/add-ceil-maxpool-ops/crates/burn-tensor/src/tensor/ops/modules/conv.rs#L68) (the function I actually updated) to get the shape, I may be mistaken though

Well yes, the output size is computed from the same function. But using ceil instead of floor basically ensures that every element in the input tensor is covered by the sliding window when pooling, including elements that would be partially outside the input boundaries. So there is likely some implicit padding added or logic to handle out of bounds elements in order to compute the results. For max pooling, these elements should be negative infinity so as to not affect the result. For avg pooling, probably zeros.

Although this might require asymmetric padding specification, which we do not currently support 😅. There is definitely a case to support asymmetric padding, but not sure about `ceil_mode` 🤔

PR #3281, opened 2025-06-10
tensor.roll()

crutcher:
`tensor.roll()`

### Checklist

### Related Issues/PRs

See pytorch roll:
https://docs.pytorch.org/docs/stable/generated/torch.roll.html

### Changes

- Added `burn::tensor::indexing`
- Added `ReversableIndex` (for usize / isize dims)
- Added `roll()`, `roll_dim()`

### Testing

Gen tests.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -890,15 +890,11 @@ where
             let size = self.shape().dims[dim];
             assert!(
                 0 < shift && shift < size,
-                "Expected: 0 < shift < size: found shift={}, size={}",
-                shift,
-                size,
+                "Expected: 0 < shift < size: found shift={shift}, size={size}",
             );
             assert!(
                 dim < self.shape().num_dims(),
-                "Expected: dim < num_dims: found dim={}, num_dims={}",
-                dim,
-                size,
+                "Expected: dim < num_dims: found dim={dim}, num_dims={size}",
             );
         }
 
@@ -934,9 +930,7 @@ where
         assert_eq!(
             dims.len(),
             shifts.len(),
-            "Dimensions and shifts must align; found dims={:#?}, shifts={:#?}",
-            dims,
-            shifts,
+            "Dimensions and shifts must align; found dims={dims:#?}, shifts={shifts:#?}",
         );
 
         // This is a fair amount of complexity, which could be replaced
--- a/crates/burn-tensor/src/tests/ops/roll.rs
+++ b/crates/burn-tensor/src/tests/ops/roll.rs
@@ -3,8 +3,8 @@ mod tests {
     use super::*;
     use burn_tensor::{Int, Slice, Tensor, TensorData, as_type, might_panic, s};
 
+    #[ignore = "0 size resources are not yet supported"]
     #[cfg(feature = "std")]
-    #[might_panic(reason = "0 size resources are not yet supported")]
     #[test]
     fn test_roll_empty() {
         let device = Default::default();


crutcher:
Probably needs some more tests; but I'd like a review of the `ReflectableIndex` code and if changes needed to be made there

crutcher:
# copied from Discord thread

I would like to lift the trait and utility code into burn::indexing; as we'll probably have a lot of related support machinery, including ravel/unravel support with bounds checking, as that is always part of tensor binding. Particularly for things like binding external datatypes (such as Images) into/out-of TensorData.

I do want to open a disagreement about having separate semantic types; and clarify a strategy here.

I am on board, in general, with the "new type" pattern to do param / unit separation, particularly in situations where we don't have keyword params; but I don't think that applies here. We're not discussing explicit casting into the appropriate semantic type 'dimension, index, shape' at the call location; in fact most of the work is to permit usize and isize to be used interchangeably without additional call-site syntax.

There is some protection if we make the utitily machinery require the newtypes, so you can't call something like wrap_index(size, index) and swap the params; but there's also a higher expectation of "know what I'm doing" when calling the indexing utilities explicitly; they will generally only be used by lower level implementors. And your comments aren't proposing that change anyway.

There's some additional (potential) utility in labeling the panic errors better; but those could be addressed by passing in labels to the utility mechanisms.

My preference would be to lift IndexConversion into burn::indexing, and not have seperate types for dim, index, size, and extent (what shift is here); just because, with the current auto-convert plans, the other approach puts a lot of weight on implementors without having any resulting api change for callers

crutcher:
Per discussion with @laggui : 
- Renamed IndexConversion to AsIndex
- Added more trait impls for the basic integer types
- moved to burn::indexing
- also, reordered roll fns to : (dim, shift) param order.
- added tests

crates/burn-tensor/src/lib.rs
```diff
@@ -10,6 +10,7 @@
 extern crate derive_new;
 
 extern crate alloc;
+extern crate core;
```
laggui: This should not be required..?

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -842,6 +842,208 @@ where
...
+    ///
+    /// ## Parameters
+    ///
+    /// - `dim`: The dimension to roll; supports negative indexing.
+    /// - `shift`: The number of positions to shift; supports negative values and wraps around.
+    ///
+    /// ## Returns
+    ///
+    /// A new tensor with the specified dimension rolled by the given shift amount.
+    #[must_use]
```
laggui: Since we haven't enforced this hint across tensor ops for now I think I would remove it for now (this applies for other additions in the roll changes).

See also my related comments in #3206

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -842,6 +842,208 @@ where
...
+    /// ## Panics
+    ///
+    /// Panics if the contract conditions are not met.
+    ///
+    /// ## Returns
+    ///
+    /// A new tensor with the specified dimension rolled by the given shift amount.
+    #[inline(always)]
+    #[must_use]
+    fn _unchecked_roll_dim(self, dim: usize, shift: usize) -> Self {
```
laggui: Underscore pre-fixes are not standard/common practice in Rust. Just the private visibility is enough.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -842,6 +842,208 @@ where
...
+    /// - `shift`: The number of positions to shift; supports negative values and wraps around.
+    ///
+    /// ## Returns
+    ///
+    /// A new tensor with the specified dimension rolled by the given shift amount.
+    #[must_use]
+    pub fn roll_dim<I1, I2>(self, dim: I2, shift: I1) -> Self
+    where
+        I1: AsIndex,
+        I2: AsIndex,
```
laggui: Even if both generics are bound to `AsIndex`, I would use better / more descriptive names than `I1` and `I2`. Since `D` is already used for the const rank generic, maybe `Dim` and `Shift`.

Also, I thought you wanted to change the order to `tensor.roll_dim(shift, dim)` ([discord](https://discord.com/channels/1038839012602941528/1091796857996451942/1382739189472432128))

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -842,6 +842,208 @@ where
...
+    ///   supports negative values and wraps around.
+    ///
+    /// ## Returns
+    ///
+    /// A new tensor with the specified dimensions rolled by the given shifts.
+    #[must_use]
+    pub fn roll<I1, I2>(self, dims: &[I2], shifts: &[I1]) -> Self
+    where
+        I1: AsIndex,
+        I2: AsIndex,
```
laggui: Same comment for dims/shifts order, and generics names.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -842,6 +842,208 @@ where
...
+    /// ## Panics
+    ///
+    /// Panics if the shifts and dimensions do not align, or if dimensions contain repeats.
+    ///
+    /// ## Returns
+    ///
+    /// A new tensor with the specified dimensions rolled by the given shifts.
+    #[inline(always)]
+    #[must_use]
+    fn _unchecked_roll(self, dims: &[usize], shifts: &[usize]) -> Self {
```
laggui: Same comment for dims/shifts order, and generics names.

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+        self as isize
+    }
+}
+
+/// Canonicalizes and bounds checks an index with negative indexing support.
+///
+/// ## Arguments
+///
+/// * `idx` - The dimension index to canonicalize.
+/// * `size` - The rank of the tensor.
```
laggui: Should refer to the size of the dimension, not rank of the tensor.

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// The canonicalized index.
+///
+/// ## Panics
+///
+/// * If `wrap_scalar` is false and the size is 0.
+/// * If the index is out of range for the dimension size.
+#[inline(always)]
+#[must_use]
+pub fn _canonicalize_named_index<I>(
```
laggui: Should probably be private? And without the underscore prefix.

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+/// Canonicalizes and bounds checks an index with negative indexing support.
+///
+/// ## Arguments
+///
+/// * `name` - The name of the index (for error messages).
+/// * `size_name` - The name of the size (for error messages).
+/// * `idx` - The index to canonicalize.
+/// * `size` - The size of the dimension.
```
laggui: Since this is used more generally, the doc should probably state "size of the dimension or rank of the tensor"?

crates/burn-tensor/src/tests/ops/roll.rs
```diff
@@ -0,0 +1,112 @@
...
+        let device = Default::default();
+        let input = TestTensorInt::<2>::zeros([12, 0], &device);
+
+        // Rolling an empty tensor should return the same empty tensor
+        input
+            .clone()
+            .roll(&[0, 1], &[1, 2])
+            .to_data()
+            .assert_eq(&input.to_data(), false);
+    }
```
laggui: I tested this use case locally for CUDA, which doesn't have the same restrictions on 0 size resources as wgpu. We have a separate issue when returning the data for a buffer of size 0.

Probably need to handle this special case for cubecl `into_data` method when `bytes.is_empty()`. Can take care of this before merging.

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    #[test]
+    fn test_canonicalize_dim() {
+        for idx in 0..3 {
+            let wrap_scalar = false;
+            assert_eq!(canonicalize_dim(idx, 3, wrap_scalar), idx as usize);
+            assert_eq!(
+                canonicalize_dim(-(idx + 1), 3, wrap_scalar),
+                (3 - (idx + 1)) as usize
+            );
+        }
```
laggui: I think it's preferrable to replace the loops with explicit test cases. That way, each test is immediately visible and easier to understand. More debuggable.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: ... my memory of the outcome of that discussion doesn't match the thread; I did extra work to *make* this change, because it used to be the other way ...

I'll change it back.

crates/burn-tensor/src/lib.rs
crutcher: Yeah, auto insert by refactoring tools; removed.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: I generally think extra-strict is a thing large projects like burn should adopt; but I'll rename it.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: Done.

crates/burn-tensor/src/tests/ops/roll.rs
crutcher: Done

crates/burn-tensor/src/tensor/indexing/mod.rs
crutcher: Done.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: done.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: done.

crates/burn-tensor/src/tensor/indexing/mod.rs
crutcher: done

crates/burn-tensor/src/tensor/api/base.rs
laggui: Could also be my memory 😅 but I think dim is typically after the input arguments to specify on which dim/axis the operation is applied. And that's how most of these ops are currently defined (though there might be some inconsistencies as you pointed out last time).

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -0,0 +1,293 @@
...
+{
+    canonicalize_named_index("index", "size", idx, size, wrap_scalar)
+}
+
+/// Canonicalizes and bounds checks a dimension index with negative indexing support.
+///
+/// ## Arguments
+///
+/// * `idx` - The dimension index to canonicalize.
+/// * `rank` - The size of the dimension.
```
laggui: rank != size of the dimension

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -842,6 +842,202 @@ where
...
+        {
+            let size = self.shape().dims[dim];
+            assert!(
+                0 < shift && shift < size,
+                "Expected: 0 < shift < size: found shift={}, size={}",
+                shift,
+                size,
+            );
+            assert!(
+                dim < size,
```
laggui: Didn't catch this in the previous round, is this check correct? `size` here is the size of the `dim`.. why are we checking that the dimension specified is smaller than its size?

laggui (commented):
Hmm not sure why the macos CI hangs.. might actually be related to the empty test.

Will have to investigate before we merge. Otherwise, just some minor comments.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: Hmm; yeah.

crates/burn-tensor/src/tensor/indexing/mod.rs
crutcher: fixed.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: No, you were right, I went back and looked at the thread

crates/burn-tensor/src/tests/ops/roll.rs
```diff
@@ -0,0 +1,112 @@
...
+        let device = Default::default();
+        let input = TestTensorInt::<2>::zeros([12, 0], &device);
+
+        let result = input.clone().roll(&[1, 2], &[0, 1]);
+
+        assert_eq!(result.shape().dims, &[12, 0]);
+
+        // TODO: Rolling an empty tensor should return the same empty tensor;
+        // but we have no way to compare tensor references yet.
+    }
```
laggui: After some on & off debugging via the CI in #3324 my initial hypothesis was confirmed.

The macos tests hang due to the creation of a 0-element tensor with `zeros([12, 0])`. This is not officially supported so it can cause some undefined behavior with the current implementation. Instead of marking the test as `might_panic` we should probably ignore it until this is supported e.g.

```
#[ignore = "0 size tensors are not yet supported"]
```

We could probably add a global check in [`TensorCheck::creation_ops`](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/api/check.rs#L83-L103) to restrict this for all backends 🤔

laggui (commented):
Sorry for the delay, haven't had a lot of time to look into this until yesterday 😅 

Found the culprit for hanging macos tests.

crutcher:
fixed and rebased

crates/burn-tensor/src/tests/ops/roll.rs
crutcher: done

laggui (approved):
Thanks for addressing all the comments 🙏 

Sorry again for the CI hold up

PR #3284, opened 2025-06-12
Fix onnx topological sort check

tye-singwa:
### Checklist

### Related Issues/PRs

#1714

### Changes

Checked why onnx models are loaded with error "Nodes are not topologically sorted" after using `onnx_opset_upgrade.py` script. The issue is that `position` holds position of last encounter of node by name, and there are empty nodes having `node.name == ""`. I changed it to explicit indices from iterators over nodes.

### Testing

I used scripts provided in linked issue https://github.com/tracel-ai/burn/issues/1714#issuecomment-2952373124


laggui (approved):
Thanks for fixing the issue!

Validated the code and double checked against some existing models.

PR #3297, opened 2025-06-16
Fix/fusion multiple streams

nathanielsimard:
Fixes many problems with fusion related to shared tensors across streams/threads. There were concurrency issues and memory leak issues.

### DIFF
--- a/crates/burn-fusion/src/stream/memory_checks.rs
+++ b/crates/burn-fusion/src/stream/memory_checks.rs
@@ -211,7 +211,7 @@ impl MemoryChecks {
 
                         // We assume that if nothing has been registered in the last second
                         // while being at a count of 1, it's the end.
-                        std::thread::sleep(Duration::from_secs(1));
+                        std::thread::sleep(Duration::from_secs(5));
 
                         if num_queued_moved.load(Ordering::Relaxed) <= 1 {
                             match last_analyses.take() {


crates/burn-fusion/Cargo.toml
```diff
@@ -15,14 +15,16 @@ version.workspace = true
 workspace = true
 
 [features]
-default = ["std"]
+default = ["std", "memory-checks"]
```
laggui: Probably shouldn't be enabled by default?

crates/burn-fusion/src/ops/boolean.rs
```diff
@@ -258,8 +265,10 @@ impl<B: FusionBackend> BoolTensorOps<Self> for Fusion<B> {
         }
 
         let shape: Vec<usize> = tensor.shape.clone();
-        let stream_1 = tensor.stream;
-        let stream_2 = value.stream;
+        let mut streams = OperationStreams::default();
+        streams.tensor(&tensor);
+        streams.tensor(&value);
```
laggui, line 270: Given the repeated pattern of creating default streams + registering inputs, should we add something like `OperationStreams::with_inputs(&[&tensors])` that basically does

```rust

let mut streams = Self::default();
for t in tensors {
    streams.tensor(t);
}
streams
```
?

crates/burn-fusion/src/ops/module.rs
```diff
@@ -260,7 +260,9 @@ impl<B: FusionBackend> ModuleOps<Fusion<B>> for Fusion<B> {
                     .bias
                     .as_ref()
                     .map(|bias| handles.get_float_tensor::<B>(bias));
+                println!("Get output grad ------------------ START");
                 let output_grad = handles.get_float_tensor::<B>(&args.out_grad);
+                println!("Get output grad ------------------ END");
```
laggui: Debug leftover

crates/burn-fusion/src/server.rs
```diff
@@ -96,21 +98,25 @@ where
...
     where
         B: FusionBackend<FusionRuntime = R>,
     {
-        let tensor = self.handles.get_float_tensor::<B>(tensor);
-        let tensor = B::float_to_device(tensor, device);
+        let tensor_float = self.handles.get_float_tensor::<B>(tensor);
+        self.streams
+            .mark_read(StreamId::current(), tensor, &self.handles);
+
+        let tensor = B::float_to_device(tensor_float, device);
```
laggui, line 119: Shouldn't this also apply to other types (e.g., `change_server_int`)?

crates/burn-fusion/src/stream/memory_checks.rs
```diff
@@ -0,0 +1,249 @@
...
+use std::{
+    fmt::Display,
+    sync::{
+        Arc,
+        atomic::{AtomicU64, Ordering},
+        mpsc::{Sender, SyncSender},
+    },
+    thread::JoinHandle,
+    time::Duration,
+};
```
laggui: Should use no-std compatible or make sure `memory_checks` is only available with "std" feature

crates/burn-fusion/src/stream/multi.rs
```diff
@@ -1,39 +1,142 @@
...
+    memory_checks: super::memory_checks::MemoryChecks,
+}
+
+#[derive(Debug)]
+enum DropAction {
+    SkipSharedTensor,
+    ForceSharedTensor(Vec<StreamId>, TensorId),
+    ContinueDrop,
+    NotADrop,
 }
```
laggui: `DropAction::NotADrop` is confusing.

I would remove the variant, and only call `self.handle_drop_op(id, tensor_ir)` when the operation is matched as a drop (see next comment for relevant code).

crates/burn-fusion/src/stream/multi.rs
```diff
@@ -1,39 +1,142 @@
...
+                        stream.queue.variables.remove(&tid);
+                        if stream.queue.variables.is_empty() {
+                            self.streams.remove(&stream_id);
+                        }
+                    }
+                }
+                true
+            }
+            DropAction::ContinueDrop => true,
+        };
```
laggui: Match on `OperationIr::Drop` here and remove the match in `handle_drop_op` which should just take the tensor ir for the drop op.

```rust
let sync = match repr {
    OperationIr::Drop(ref mut tensor_ir) => match self.handle_drop_op(id, tensor_ir) {
        DropAction::SkipSharedTensor => return,
        DropAction::ForceSharedTensor(stream_ids, tid) => {
            for stream_id in stream_ids {
                if let Some(stream) = self.streams.get_mut(&stream_id) {
                    stream.queue.variables.remove(&tid);
                    if stream.queue.variables.is_empty() {
                        self.streams.remove(&stream_id);
                    }
                }
            }
            true
        }
        DropAction::ContinueDrop => true,
    },
    _ => false,
};

```

crates/burn-fusion/src/stream/shared_tensors.rs
```diff
@@ -0,0 +1,310 @@
...
+/// The result from a [SharedTensor::update].
+pub enum SharedTensorUpdate {
+    /// The tensor is removed from the current stream.
+    ///
+    /// Also contains if the current stream is empty.
+    RemovedFromStream(bool),
+    /// If the tensor is shared across zero streams.
+    NoMoreStream,
+    /// If nothing has been done from the update.
+    NothingToDo,
```
laggui: Maybe rename the variants?
- `NoMoreStreams` -> `ReadyForCleanup`
- `NothingToDo` -> `NoChange`

crates/burn-fusion/src/stream/shared_tensors.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    /// Update the current shared tensor state on the given stream.
+    ///
+    /// If the shared tensor is no longer needed on the stream, we will remove it from the list of
+    /// shared streams.
+    ///
+    /// If the shared tensor is needed on no stream, we return true, indicating that the shared
+    /// tensor is safe to manually drop.
```
laggui: The `return true` comment is not accurate

crates/burn-ir/src/operation.rs
```diff
@@ -1392,6 +1392,39 @@ impl OperationIr {
             OperationIr::Drop(repr) => vec![repr],
         }
     }
+
+    /// Set the given nodes that are [read write](super::TensorStatus::ReadWrite) to
+    /// [read only](super::TensorStatus::ReadOnly) in the current operation.
+    ///
+    /// Returns the tensor that were updated with their original representation.
+    pub fn readonly(&mut self, nodes: &[TensorId]) -> Vec<TensorIr> {
```
laggui: Maybe `set_read_only` or `mark_read_only` would be more explicit?

crates/burn-fusion/src/ops/boolean.rs
nathanielsimard: Yeah we could, I'll leave the refactor for another PR.

crates/burn-fusion/src/stream/memory_checks.rs
nathanielsimard: I putted `std` in the feature flag to make it obvious

PR #3327, opened 2025-06-27
Fix tui label placement

BenFradet:
### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed. (loads of formatting issues, not sure why?)

### Changes

Fixes a minor pet peeve of mine where the plot is being updated behind the legend
Another pet peeve of mine is the validation plot being only one line going across
I'll try to fix that next unless there are objections

### Testing

before
1750948751_grim
after
1751035167_grim

### DIFF
--- a/crates/burn-train/src/renderer/tui/metric_numeric.rs
+++ b/crates/burn-train/src/renderer/tui/metric_numeric.rs
@@ -6,7 +6,7 @@ use ratatui::{
     prelude::{Alignment, Constraint, Direction, Layout, Rect},
     style::{Color, Modifier, Style, Stylize},
     text::Line,
-    widgets::{Axis, Block, Borders, Chart, Paragraph, Tabs},
+    widgets::{Axis, Block, Borders, Chart, LegendPosition, Paragraph, Tabs},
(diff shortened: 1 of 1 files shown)


laggui (approved):
Ideally we could try to keep it at the top but have some sort of margin so it never overlaps with the rendered data points. Not sure exactly how we would need to configure this with ratatui.

But for now LGTM!

> Another pet peeve of mine is the validation plot being only one line going across

This is simply due to the fact that validation only happens after each training epoch, so there is no data between these steps.

BenFradet:
> Ideally we could try to keep it at the top but have some sort of margin so it never overlaps with the rendered data points. Not sure exactly how we would need to configure this with ratatui.

I can try looking into it

> This is simply due to the fact that validation only happens after each training epoch, so there is no data between these steps.

yes, I understand that, but it still doesn't really look good imo :sweat_smile:

laggui:
> yes, I understand that, but it still doesn't really look good imo 😅

Yeah I agree that the jumps always look weird lol, we just never took the time to address it 😅

PR #3335, opened 2025-07-01
feat(examples): add Metal backend support to custom-image-dataset

TsaoLun:
## Changes

- Added Metal backend support to the custom-image-dataset example for macOS users.
- Refactored main function to reduce code duplication.

### DIFF
--- a/examples/custom-image-dataset/examples/custom-image-dataset.rs
+++ b/examples/custom-image-dataset/examples/custom-image-dataset.rs
@@ -28,13 +28,13 @@ fn main() {
         train::<Autodiff<LibTorch>>(config, device);
     }
 
-    #[cfg(all(feature = "wgpu", not(feature = "tch-gpu")))]
+    #[cfg(feature = "wgpu")]
     {
         use burn::backend::wgpu::{Wgpu, WgpuDevice};
         train::<Autodiff<Wgpu>>(config, WgpuDevice::default());
     }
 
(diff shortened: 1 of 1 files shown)


examples/custom-image-dataset/examples/custom-image-dataset.rs
```diff
@@ -1,57 +1,44 @@
...
+        train::<Autodiff<Wgpu>>(config, WgpuDevice::default());
     }
-}
 
-fn main() {
-    #[cfg(feature = "tch-gpu")]
-    tch_gpu::run();
-    #[cfg(feature = "wgpu")]
-    wgpu::run();
+    #[cfg(all(feature = "metal", not(feature = "tch-gpu"), not(feature = "wgpu")))]
```
laggui: The current pattern is incremental and inconsistent:

```rust
    #[cfg(feature = "tch-gpu")]
    // ...

    #[cfg(all(feature = "wgpu", not(feature = "tch-gpu")))]
    // ...

    #[cfg(all(feature = "metal", not(feature = "tch-gpu"), not(feature = "wgpu")))]
    // ...
```

If the goal is to select a single backend, all cfg block conditions should be mutually exclusive of the others.

But for simplicity we can keep it as `#[cfg(feature = "selected-backend-feature-flag")]` for each block.

laggui (commented):
Thanks for the addition! LGTM, just a minor comment

laggui (approved):

PR #3339, opened 2025-07-01
Update KaTeX header path in rustdoc args and move file

antimora:
### Related Issues/PRs

#3281 

### Changes

Moved 'katex-header.html' to 'assets/katex-header.html' and updated the rustdoc arguments in burn-tensor's Cargo.toml to reference the new path. This ensures documentation builds correctly with the relocated header file.

### Testing

```
RUSTDOCFLAGS="--html-in-header assets/katex-header.html" cargo doc --no-deps --package burn-tensor
```


laggui (commented):
Is there a particular issue with having it at the root? We have a bunch of other files/configs also at the root anyway. And iirc even the basic instructions for KaTeX point to the root.

If we had a `docs/` we could put it there, or next best might be `.cargo/` since this is related to cargo docs. But `assets/` holds static content assets only right now, none of which are used for the docs.

antimora:
@laggui i would like to keep the root clean and only for things that cannot be moved or for doc discoverability. Since this html is a script resource, we should move it. 

I can move the file elsewhere.

laggui:
Sure! In this case, let's start this in a new `docs/` directory.

I think we can eventually move the `burn-book` and `contributor-book` to this tree 🙂

Issue #3340, opened 2025-07-01
`is_close` incorrectly reports infinite to be close to all finite values


jonboh:
```rust
let inf = Tensor::<B, 1>::from_floats([f32::INFINITY], &device);
let zero = Tensor::<B, 1>::from_floats([0.], &device);
println!("{}", zero.is_close(inf, None, None)); //returns true
```

`is_close` should check for infinite values.

For reference, here is pytorch's behavior:
```python
inf = torch.tensor([float('inf')], dtype=torch.float32)
one = torch.tensor([1], dtype=torch.float32)
torch.isclose(inf, one) # == false
torch.isclose(inf, inf) # == true
```

I have a fix almost ready, I'll submit a PR shortly

antimora:
Thanks for filing. I worked on this so I can review your PR.

PR #3341, opened 2025-07-01
add `is_finite` tensor element wise op and fix `is_close`/`all_close` behavior for infinity values

jonboh:
This pr fixes: https://github.com/tracel-ai/burn/issues/3340

There might be a better (cleaner?) way to make these operations if `is_nan`/`is_finite` are moved to the `Numeric` trait, but as `is_nan` was done at the level of `Tensor`, I implemented `is_finite` at the same level.

I've added some extra test on `test_all_close` to check these conditions.


jonboh:
@antimora here's the PR, thanks for reviewing it :), I could not assign you as reviewer, I assume only members of the org can

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2264,6 +2276,34 @@ where
...
+    /// fn example<B: Backend>() {
+    ///    let device = B::Device::default();
+    ///    let tensor = Tensor::<B, 2>::from_data([[1.0, f64::INFINITY, 3.0], [5.0, 9.0, 6.0]], &device);
+    ///    let tensor = tensor.is_finite();
+    ///    println!("{tensor}");
+    ///    // [[true, false, true], [true, true, true]]
+    /// }
+    /// ```
+    pub fn is_finite(&self) -> Tensor<B, D, Bool> {
+        let max = Tensor::zeros_like(self) + K::Elem::MAX;
```
laggui: You can use `Tensor::full_like(value)` instead of `0 + value`.

crates/burn-tensor/src/tests/ops/close.rs
```diff
@@ -30,5 +30,14 @@ mod tests {
...
         assert!(tensor1.all_close(tensor2, None, None));
+
+        // non finite values
+        let inf_plus = TestTensor::<2>::from([[f32::INFINITY]]);
+        let one = TestTensor::<2>::from([[1.]]);
+        let inf_minus = TestTensor::<2>::from([[-f32::INFINITY]]);
+        assert!(!inf_plus.clone().all_close(inf_minus.clone(), None, None));
+        assert!(!one.clone().all_close(inf_minus.clone(), None, None));
+        assert!(!one.all_close(inf_plus.clone(), None, None));
+        assert!(inf_plus.clone().all_close(inf_plus, None, None));
```
laggui, line 41: We should also have tests for new functions (e.g., `is_finite`)

laggui (changes requested):
Thanks for contributing 🙏 

This is certainly better than what we have in place.

While `abs(x) <= Elem::MAX` improves the robustness of the current check, I think we should have an actual `is_inf` check that compares both INF and -INF.

We don't have an associated const for infinities in the element trait, but we can still use `.elem()` or `from_elem(value)` from the `ElementConversion` trait with `f64::INFINITY` and `f64::NEG_INFINITY`. Obviously this only makes sense for float tensors, so might be best to move the implementation to float tensors only. The conversion between floating point types should work correctly.

This also made me realize that `is_close` is implemented for numeric (float AND int), but it only really makes sense for float tensors.

So overall, I think the best approach would be:
- New `is_inf` float method to check against `INFINITY` and `NEG_INFINITY`
- `is_finite` should check for (not) `is_nan` and (not) is_inf` (as your current logic)

and these methods (along with `is_close`) should belong to float only imo.

Let me know what you think!

We can do this in a follow-up PR if you prefer, but I think it makes sense to improve it here.

jonboh:
I agree with the changes you propose regarding the consts for INF and -INF and having the methods defined only for float tensors. I'll push the changes in this PR once I have them ready👍

jonboh:
I've moved the `is_nan` and `contains_nan` as well, as those also only make sense on Float Tensors, and added specific tests for `is_inf` and `is_finite`

antimora:
The code changes look good to me. 

Could you also please update the book here: https://github.com/tracel-ai/burn/blob/main/burn-book/src/building-blocks/tensor.md

New entries for is_inf and is_finite. And move is_close from numeric to float?

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2204,6 +2111,129 @@ where
...
+    /// ```math
+    /// C = AB
+    /// ```
+    pub fn matmul(self, other: Tensor<B, D, K>) -> Tensor<B, D, K> {
+        check!(TensorCheck::matmul(&self, &other));
+        Tensor::new(K::matmul(self.primitive, other.primitive))
+    }
+}
+
+impl<B, const D: usize> Tensor<B, D, Float>
```
laggui: Since these operations are now strictly for float tensors, let's remove it from numeric and move it to the `Tensor<B, D>` impl in [`api/float.rs`](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/api/float.rs#L9).

With this change we can also use the high-level tensor API methods (e.g., `Float::sub(self.primitive, other.primitive)` -> `self.sub(other)`).

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2204,6 +2111,129 @@ where
...
+        let atol = atol.unwrap_or(DEFAULT_ATOL);
+
+        // check finite difference is close
+        let is_close_finite_val = Tensor::new(Float::lower_equal(
+            Float::abs(Float::sub(self.primitive.clone(), other.primitive.clone())),
+            Float::add_scalar(
+                Float::mul_scalar(Float::abs(other.primitive.clone()), rtol),
+                atol,
+            ),
+        ))
```
laggui: No need to work with the primitives here

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2227,7 +2257,10 @@ where
     pub fn is_nan(&self) -> Tensor<B, D, Bool> {
         // Check if the input tensor is NaN by comparing it to itself
         // NaN is the only value that is not equal to itself
-        Tensor::new(K::not_equal(self.primitive.clone(), self.primitive.clone()))
+        Tensor::new(Float::not_equal(
+            self.primitive.clone(),
+            self.primitive.clone(),
+        ))
```
laggui: Same here

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2258,20 +2291,61 @@ where
         // Summing the tensor will result in NaN if the tensor contains any NaN values
         // This is faster than checking each element individually
         // because it rolls up the NaN values into a single value
-        let sum = K::sum(self.primitive.clone());
+        let sum = Float::sum(self.primitive.clone());
 
         // Check if the sum is NaN by comparing it to itself
-        Tensor::new(K::not_equal(sum.clone(), sum))
+        Tensor::new(Float::not_equal(sum.clone(), sum))
```
laggui: Same 🙂

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2258,20 +2291,61 @@ where
...
+    ///    // [[false, true, false], [false, false, false]]
+    /// }
     /// ```
-    pub fn matmul(self, other: Tensor<B, D, K>) -> Tensor<B, D, K> {
-        check!(TensorCheck::matmul(&self, &other));
-        Tensor::new(K::matmul(self.primitive, other.primitive))
+    pub fn is_inf(&self) -> Tensor<B, D, Bool> {
+        self.clone()
+            .abs()
+            .equal(Tensor::full_like(self, f64::INFINITY))
```
laggui: Prefer using `equal_elem(val)` instead of creating a full tensor with the same shape.

jonboh:
Ok, I've moved the ops to api/float, I've slightly modified the functions to now consume `self`. I've modified the cases in `burn-train` where this made an additional clone necessary. I think this makes sense, as the clone would otherwise be needed inside the functions themselves and that would make, in some cases, an unneeded extra clone. 
Let me know if there's anything else to improve

laggui (approved):
LGTM! Thanks again 🙂

Issue #3260, opened 2025-06-05
Support exporting module to safetensors format


cex1gt:
### Feature Description

The `burn_import` crate currently supports reading parameters from the SafeTensors format. However, saving a Burn module's parameters to the SafeTensors format is not yet implemented, likely due to differences in how Burn tensor types are handled by the default Serde serialization.

This feature proposes moving the SafeTensors reader from `burn_import` to an official recorder implementation and adding support for SafeTensors serialization. This would enable saving Burn module parameters in the SafeTensors format.

### Feature Motivation

SafeTensors is a widely adopted weight transfer protocol, supported by frameworks such as PyTorch, FLAX, tinygrad, and others. Enabling Burn to save model weights in the SafeTensors format would facilitate exporting Burn models to these frameworks. This would provide a standardized, interoperable way to transfer Burn module parameters, simplifying the process of using Burn models in other frameworks.

### Asks

I am happy to do the work of implementing the serialization. My ask is to guide where the best place to override default serialization to support exporting SafeTensors. My initial guess would be to override how TensorData is serialized.

antimora:
yes, it's a great idea. It should be doable. Here is the general steps:

1. create a serde serializer of Burn module that flattens hierarchical modules into flatten name value pairs (levels separated by dots). This function should be general not specific to safetensors so other exporters can use it. You can see this serializer as an example this creates nested objects: https://github.com/tracel-ai/burn/blob/main/crates/burn-core/src/record/serde/ser.rs
2. implement safetensors conversion (use their API).
3. Update SafetensorsFileRecorder  save_item and save methods here: https://github.com/tracel-ai/burn/blob/1f5ccd43b3f317a5cdb8fef4fcd4ed506d4d190f/crates/burn-import/src/safetensors/recorder.rs#L28

laggui:
> Enabling Burn to save model weights in the SafeTensors format would facilitate exporting Burn models to these frameworks

While it would be easier since safetensors files can be loaded in python, you would still only get the parameters without model definition. And just like we added an adapter for the pytorch import (including safetensors), you would probably need to perform some manipulation on the other end nonetheless.

Still, I agree that it would make it a little easier just because you can load this format in python (i.e., other frameworks).

The general steps provided in the last comment should help you get started 🙂

PR #3345, opened 2025-07-03
Add safetensors export feature

jonboh:
This is still a work in progress, I started out following antimora's indications in: https://github.com/tracel-ai/burn/issues/3260#issuecomment-2950281284

I did not find a clean way to walk the Module with the Serializer, and perform the serialization only with the Recorder. The problem is that once the Serializer is finished I'd need to reconstruct each TensorData back from its `bytes`, `dtype` and `shape` fields as the serializer will walk up to the basic types. 

The alternative I found was to use the Serializer to get a mapping from ParamId to the TensorName, and use a ModuleVisitor to link to the TensorData. I think this is a bit cleaner, but the API is pretty different from the rest of the serializations, so I'm not sure, if it would be better to do it the other way, even if the logic ends up a bit more convoluted.
Any comments are appreciated :)

### DIFF
--- a/crates/burn-import/src/safetensors/serializer.rs
+++ b/crates/burn-import/src/safetensors/serializer.rs
@@ -95,7 +95,7 @@ impl std::fmt::Display for Error {
         match self {
             Error::UnsupportedType => write!(f, "Unsupported type TensorData"),
             Error::NotFromTensorData => write!(f, "The field cannot be from a TensorData"),
-            Error::Custom(msg) => write!(f, "{}", msg),
+            Error::Custom(msg) => write!(f, "{msg}"),
         }
     }
 }


crates/burn-import/Cargo.toml
```diff
@@ -49,6 +49,7 @@ thiserror = { workspace = true, optional = true }
 tracing-core = { workspace = true }
 tracing-subscriber = { workspace = true }
 zip = { workspace = true, optional = true }
+safetensors = { workspace = true }
```
antimora: Should be `optional = true` and enabled via feature flag.

antimora:
I think I know what the issue you are facing is regarding the serializer going all the way down. It's okay to stop if you come across a `TensorData` struct type and handle serialization differently. 

My recommendation: 
1. Don't use ModuleVisitor (you can get away with a serde serializer). 
2. The only intermediate data structure should be a hash type of full path name + tensor safetensor view.

crates/burn-import/Cargo.toml
```diff
@@ -26,6 +26,7 @@ safetensors = [
     "thiserror",
     "zip",
     "candle-core",
+    "dep:safetensors",
```
antimora, line 29: you don’t need to use dep: for the newer rust editions.

crates/burn-import/safetensors-tests/tests/export/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+        let serialized = to_safetensors(model.clone()).unwrap();
+        let mut file = File::create("model.safetensors").unwrap();
+        file.write_all(&serialized).unwrap();
+        let record = SafetensorsFileRecorder::<FullPrecisionSettings>::default()
+            .load(
+                LoadArgs::new("model.safetensors".into()).with_adapter_type(AdapterType::NoAdapter),
+                &device,
+            )
+            .expect("Should decode state successfully");
+        std::fs::remove_file("model.safetensors").unwrap();
```
antimora, line 41: I recommend using NamedTempFile https://github.com/tracel-ai/burn/blob/main/crates/burn-dataset/src/dataset/sqlite.rs#L713. Otherwise you'll have collisions and hanging files.

crates/burn-import/src/safetensors/export.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Generates the safetensors serialization of a [Module]
+pub fn to_safetensors<B: Backend, T: Module<B>>(module: T) -> Result<Vec<u8>, SafeTensorError> {
+    let mut ser = FlattenSerializer::new();
+    module
+        .into_record()
+        .into_item::<FullPrecisionSettings>()
+        .serialize(&mut ser)
+        .unwrap();
+    safetensors::serialize(ser.into_map(), None)
+}
```
antimora, line 20: We should also enhance SafetensorsFileRecorder's save_item method just like in 
https://github.com/tracel-ai/burn/blob/main/crates/burn-core/src/record/file.rs#L194

This way SafetensorsFileRecorder feature is symmetrical

crates/burn-import/src/safetensors/serializer.rs
```diff
@@ -0,0 +1,700 @@
...
+
+#[derive(Debug)]
+pub struct SafetensorsTensorData {
+    bytes: Vec<u8>,
+    shape: Vec<usize>,
+    dtype: DType,
+}
+
+impl safetensors::View for SafetensorsTensorData {
+    fn dtype(&self) -> safetensors::Dtype {
```
antimora, line 17: Can't create a view on TensorData directly without the need to save into `bytes`? This way there won't be an intermediate memory usage for large models. So if we have 8GB model sitting in GPU, we will not create a copy in RAM.

antimora (changes requested):
Congrats to make it work! I know dealing with Serde is not easy.

I did a quick pass and here are my quick high level comments. I will dive in more details later.

1. We should hook save_item under SafetensorsFileRecorder. 
2. We should aim not to duplicate tensor data. The tensor view should be sufficient for safetensors to pull data directly. 
3. It seems the serde serializer could be simplified and made more robust. I am still looking into the implementation, so I do not have any concrete suggestions.
4. We should verify serialization of tuple, enum and vec of modules. It might be worth to update safetensors tests with more complete test data.
5. We should add module adapter just like in SafetensorsFileRecorder because our modules are not one to one. SafetensorsFileRecorder's default adapter is pytorch.

crates/burn-import/src/safetensors/serializer.rs
jonboh: yes, you are right, I was preocuppied with making it work and I didn't try using just a reference to the original bytes. I'll try modifying it so that instead of `Vec<u8>`, we point to the original data using a `&[u8]`.

antimora:
I am working on this feature. I will make a new PR available.

PR #3362, opened 2025-07-08
Fix/ci autotune

laggui:
[WIP]

### DIFF
--- a/crates/burn-cubecl-fusion/src/matmul/optimization.rs
+++ b/crates/burn-cubecl-fusion/src/matmul/optimization.rs
@@ -137,7 +137,7 @@ impl<R: Runtime> MatmulOptimization<R> {
         fused_matmul_autotune::<R, BT>(self, context);
 
         #[cfg(not(feature = "autotune"))]
-        if self.execute_standard_fused::<BT>(context).is_err() {
+        if self.execute_fused::<BT, Simple>(context).is_err() {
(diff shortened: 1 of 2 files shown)


crates/burn-cubecl-fusion/src/shared/trace/plan.rs
```diff
@@ -150,6 +150,36 @@ pub struct HandleInput<R: Runtime> {
...
+impl<R: Runtime> HandleInput<R> {
+    pub fn new(
+        tensor_global: &TensorIr,
+        tensor_relative: &TensorIr,
+        precision: FusePrecision,
+        mut handle: CubeFusionHandle<R>,
+        mut strides: Vec<usize>,
+    ) -> Self {
+        // For rollback
+        core::mem::swap(&mut handle.strides, &mut strides);
```
nathanielsimard, line 166: We remove the strides from the handle, is it OK?

laggui: Actually we modify the strides from the handle with the new strides, and the original handle strides are preserved for the rollback.

Before this change the handle strides were directly modified in the input planner and incorrectly handled in the fusion rollback, handles would have modified strides on rollback. Caused some invalid accesses / indexing. I added a test for this: `should_support_multiple_reshapes_cloned_tensor`.

PR #3208, opened 2025-05-18
ONNX import - sanitized names 

antimora:
This pull request introduces functionality to sanitize ONNX node names, input names, and output names to ensure they are valid Rust identifiers. It also includes new tests to verify the correctness of this functionality. 

### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed.

### Related Issues/PRs

Fixes #2878

### Changes

* Added and wired sanitize function
* Added test
* Improved comments in `util.rs` to clarify the purpose of the `check_opset_version` function and adjusted its logic to handle non-standard ONNX domains gracefully.

### Testing

Added ONNX file with bad names to test the correctness and also checked user provided ONNX files in  #2878

PR #3371, opened 2025-07-11
Fix ONNX node name sanitization and add ai.onnx.ml domain support

### LINKED ISSUE #2878: bug(onnx): model created using tf2onnx panics with non-valid Ident
**Describe the bug**
I have a equinox jax model I want to import into burn. I used `jax2tf` to get a tensorflow model, then `tf2onnx` to obtain the onnx file. When I try to import this into burn it panics with the message:

```
ERROR burn_import::logger: PANIC => panicked at /home/mrobins/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-import-0.16.0/src/burn/ty.rs:123:19:
  "jax2tf_rhs_/pjit_silu_/Const_2:0" is not a valid Ident   
```

I'd imagine it is the ":" or "/" characters in the ident, which seem to be used to identify blocks and outputs in the model. Is this panic expected or a bug?

**To Reproduce**

You can get the onnx file here: https://github.com/martinjrobins/diffsol/raw/refs/heads/workspace/examples/neural-ode-weather-prediction/rhs.onnx

Then I read it in as per the onnx example: https://burn.dev/burn-book/import/onnx-model.html

**Expected behavior**

The model to import without panic

 - OS: Ubuntu
 - Browser chrome
 - Version 22.04

**Additional context**
Using burn v0.16.0

antimora:
### Checklist

### Related Issues/PRs

  - Fixes #2878 - ONNX import fails with panic on node names containing special characters

###  Changes

  Fixed ONNX import issues:

  1. Node name sanitization: Enhanced format_name function to sanitize node names with invalid identifier characters (:, /, etc.) by replacing them with underscores
  2. ONNX domain support: Added support for ai.onnx.ml domain in addition to standard ONNX domain

  This allows importing ONNX models generated by tf2onnx that contain special characters in node names.

### Testing

  - Added unit tests covering the problematic node name from the issue and various edge cases
  - Verified the original failing ONNX file now imports successfully
  - All existing tests pass

### DIFF
--- a/crates/burn-import/SUPPORTED-ONNX-OPS.md
+++ b/crates/burn-import/SUPPORTED-ONNX-OPS.md
@@ -206,32 +206,6 @@ functionality.
 | [Xor][192]                       | ✅             | ✅           |
 | [Unsqueeze][193]                 | ✅             | ✅           |
 
-## ai.onnx.ml Domain Operators
-
-The following operators from the ai.onnx.ml domain are not currently supported:
-
-| ONNX ML OP                       | Import Support | Burn Support |
-|----------------------------------|:--------------:|:------------:|
-| [ArrayFeatureExtractor][200]     | ❌             | ❌           |
-| [Binarizer][201]                 | ❌             | ❌           |
-| [CastMap][202]                   | ❌             | ❌           |
-| [CategoryMapper][203]            | ❌             | ❌           |
-| [DictVectorizer][204]            | ❌             | ❌           |
-| [FeatureVectorizer][205]         | ❌             | ❌           |
-| [Imputer][206]                   | ❌             | ❌           |
-| [LabelEncoder][207]              | ❌             | ❌           |
-| [LinearClassifier][208]          | ❌             | ❌           |
-| [LinearRegressor][209]           | ❌             | ❌           |
-| [Normalizer][210]                | ❌             | ❌           |
-| [OneHotEncoder][211]             | ❌             | ❌           |
-| [Scaler][212]                    | ❌             | ❌           |
-| [SVMClassifier][213]             | ❌             | ❌           |
-| [SVMRegressor][214]              | ❌             | ❌           |
-| [TreeEnsemble][215]              | ❌             | ❌           |
-| [TreeEnsembleClassifier][216]    | ❌             | ❌           |
-| [TreeEnsembleRegressor][217]     | ❌             | ❌           |
-| [ZipMap][218]                    | ❌             | ❌           |
-
 [1]: https://onnx.ai/onnx/operators/onnx__Abs.html "ONNX Abs"
 [2]: https://onnx.ai/onnx/operators/onnx__Acos.html "ONNX Acos"
 [3]: https://onnx.ai/onnx/operators/onnx__Acosh.html "ONNX Acosh"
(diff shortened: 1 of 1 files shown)


antimora:
The name sanitization is happening in burn-import instead of onnx-ir. This is simpler for now and updates existing sanitization in place related. Added a todo to come back. #3208 PR attempted to fix in onnx-ir but added up introducing new issues because some names relied on old unsanitized names thus breaking code.

crates/burn-import/src/burn/ty.rs
```diff
@@ -63,13 +63,31 @@ pub enum Type {
...
+                    }
+                })
+                .collect::<String>();
+
+            // Ensure the name starts with a letter or underscore
+            if sanitized.starts_with(|c: char| c.is_ascii_alphabetic() || c == '_') {
+                sanitized
+            } else {
+                format!("_{sanitized}")
+            }
```
laggui: I think we can further simplify the general check with all these conditions, e.g.

```rust
pub fn format_name(name: &str) -> String {
    let mut result = String::with_capacity(name.len());
    // Sanitize the name by replacing invalid identifier characters with underscores
    for c in name.chars() {
        if c.is_ascii_alphanumeric() || c == '_' {
            result.push(c);
        } else {
            result.push('_');
        }
    }

    // Ensure the first character is valid to start an identifier
    if !result.starts_with(|c: char| c.is_ascii_alphabetic() || c == '_') {
        result = format!("_{result}");
    }
    result
}
```

mostly removes a duplicate check for the starting character

crates/onnx-ir/src/util.rs
```diff
@@ -53,15 +53,21 @@ pub fn shape_config(curr: &Node) -> (usize, usize) {
...
+/// * If the domain is not supported
 pub fn check_opset_version(opset: &OperatorSetIdProto, min_version: i64) -> bool {
-    // For now, only empty domain (standard ONNX operators) is supported
-    if !opset.domain.is_empty() {
-        panic!("Only the standard ONNX domain is supported");
+    match opset.domain.as_str() {
+        // Standard ONNX operators
+        "" => opset.version >= min_version,
+        // ONNX ML operators - commonly used for traditional ML operators
+        "ai.onnx.ml" => opset.version >= 1, // ML operators are generally stable from version 1
```
laggui, line 62: Adding this domain means we need to extend the ONNX op support list, no? We currently only list the basic ops in `SUPPORTED-ONNX-OPS.md`, not part of `ai.onnx.ml`.

laggui (commented):
Fix looks good, just a minor comment.

What's the motivation behind extending support for `ai.onnx.ml` domain? We would need a separate list for [these ops](https://github.com/onnx/onnx/blob/main/docs/Operators-ml.md), which I'm not sure is desirable.

antimora:
> Fix looks good, just a minor comment for the implementation 🙂
> 
> What's the motivation behind extending support for `ai.onnx.ml` domain? We would need a separate list for [these ops](https://github.com/onnx/onnx/blob/main/docs/Operators-ml.md), which I'm not sure is desirable.

Sorry I should have explained. TensorFlow adds this when exporting even though no ops are used. I have discovered this when testing an onnx file in #2878

crates/onnx-ir/src/util.rs
antimora: Maybe we should mention this since it's part of the ONNX spec.

crates/burn-import/SUPPORTED-ONNX-OPS.md
```diff
@@ -206,6 +206,32 @@ functionality.
...
+| [Normalizer][210]                | ❌             | ❌           |
+| [OneHotEncoder][211]             | ❌             | ❌           |
+| [Scaler][212]                    | ❌             | ❌           |
+| [SVMClassifier][213]             | ❌             | ❌           |
+| [SVMRegressor][214]              | ❌             | ❌           |
+| [TreeEnsemble][215]              | ❌             | ❌           |
+| [TreeEnsembleClassifier][216]    | ❌             | ❌           |
+| [TreeEnsembleRegressor][217]     | ❌             | ❌           |
+| [ZipMap][218]                    | ❌             | ❌           |
+
```
laggui: Not sure I would add the table. From the looks of these operators, I don't think we'll plan to support them any time soon 😅 so might be misleading.

We can accept the domain in the check for the reason you mentioned, but I think I would keep it at that.

laggui (commented):
> Sorry I should have explained. TensorFlow adds this when exporting even though no ops are used.

Ahh ok I see, LGTM then!

But I think we should just accept the domain without listing the ops.

laggui (approved):
Thank you!

PR #3381, opened 2025-07-16
Add comprehensive Shape type support for ONNX operations

### LINKED ISSUE #3494: Error when importing the joytag image tagging ONNX model
**Describe the bug**

When running `cargo build` with this build file:

```rust
// build.rs

use burn_import::onnx::ModelGen;

fn main() {
    ModelGen::new()
        .input("model/model.onnx")
        .out_dir("model/")
        .run_from_script();
}
```

I get this error: 
```
  ...
  DEBUG onnx_ir::rank_inference: Inferring rank for node: concat2
  DEBUG onnx_ir::node::concat: Concat rank inference for node concat2
  ERROR burn_import::logger: PANIC => panicked at /home/simonp/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/onnx-ir-0.18.0/src/node/concat.rs:14:10:
  called `Option::unwrap()` on a `None` value

  --- stderr

  thread 'main' panicked at /home/simonp/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/onnx-ir-0.18.0/src/node/concat.rs:14:10:
  called `Option::unwrap()` on a `None` value
```

**To Reproduce**

1. Download the [joytag onnx model](https://huggingface.co/fancyfeast/joytag/blob/main/model.onnx) file.
2. Make a `build.rs` file with the code as described above.
3. Run `cargo build`.

**Expected behavior**

I would expect that the corresponding `model.rs` file would be generated.

 - OS: Linux; terminal is in a fedora distrobox.
 - Browser: Firefox
 - Version: burn-import 0.18

**Aditional Information**
The onnx file has opset version 17 according to the provided onnx script.
Script:
```bash
uv run --script https://raw.githubusercontent.com/tracel-ai/burn/refs/heads/main/crates/burn-import/onnx_opset_upgrade.py
```

antimora:
This PR adds comprehensive Shape type support for ONNX operations in Burn. 

Burn's ONNX implementation uses a dedicated Shape type (rather than Tensor) for the [Shape Op](https://onnx.ai/onnx/operators/onnx__Shape.html) output. This design choice enables efficient CPU-based shape manipulations since shape data is simply a small array of `usize` values. This approach avoids expensive CPU-GPU data transfers that would occur if shapes were represented as tensors.

Based on analysis of real-world ONNX models, we've identified and implemented Shape support for the following operations:

- [x] **Add** (Scalar + Shape, Shape + Shape)
- [x] **Cast** (pass through)
- [x]  **Contcat** 
- [x] **Constant** (as input for Add, Div, Mul, Sub)
- [x] **Div** (Scalar + Shape, Shape + Shape)
- [x] **Expand** (Shape type for shape input (second input)
- [x] **Gather**
- [x] **Mul** (Scalar + Shape, Shape + Shape)
- [x] **Reshape** (as an arg and input)
- [x] **Equal**
- [x] **Where**
- [x] **Slice** (as starts/ends args and input)
- [x] **Sub** (Scalar + Shape, Shape + Shape)
- [x] **Squeeze**
- [x] **Unsqueeze**
- [x] **Shape** (Shape OP on Shape input)

### Related Issues/PRs

- #2822
- #2930
- fixes #3494
- https://github.com/tracel-ai/models/issues/74
- #1882 

### Testing

WIP

### Checklist

- [ ] Confirmed that `cargo run-checks` command has been executed
- [ ] Made sure the book is up to date with changes in this PR

### Examples:

This subgraph is from Yolo11X:

ONNX graph with Shape operations

This subgraph is from Yolo4v:

image

This subgraphs is from deberta:

image

image

image

This subgraph is from joytag

image


antimora:
I am still working on this. This is a bit comprehensive ops update.

antimora:
@laggui FYI. This is my progress so far.

antimora:
Quick update I am able to convert `yolo11x_opset16.onnx`. More ops to go.

antimora:
@laggui, can you please start reviewing the PR. I still have a few ops support but current I am able to convert yolo11x_opset16.onnx without any graph simplification. 

I will be traveling but I'll have my computer with me. I might be able to do some light work. 

Feel free to fix anything yourself if easy or quick.

crates/burn-import/onnx-tests/tests/add/add_shape.py
```diff
@@ -0,0 +1,83 @@
...
+    graph_def = helper.make_graph(
+        [shape_node1, shape_node2, scalar_const, add_scalar_node, add_shapes_node],
+        'shape_add_test',
+        [input_tensor1, input_tensor2],
+        [output1, output2],
+    )
+    
+    # Create the model
+    model_def = helper.make_model(graph_def, producer_name='onnx-tests')
+    model_def.opset_import[0].version = 16
```
laggui: I think that's why they have `make_model_gen_version`

crates/burn-import/onnx-tests/tests/add/add_shape.py
```diff
@@ -0,0 +1,83 @@
...
+    # Save the model
+    onnx_name = "add_shape.onnx"
+    onnx.save(model_def, onnx_name)
+    print("Finished exporting model to {}".format(onnx_name))
+    
+    # Print test expectations
+    print("Test input1 shape: [2, 3, 4]")
+    print("Test input2 shape: [5, 6, 7]")
+    print("Expected shape_plus_scalar: [12, 13, 14] (shape1 + 10)")
+    print("Expected shape_plus_shape: [7, 9, 11] (shape1 + shape2)")
```
laggui: Just curious, what's the point of the print statements here if everything is hardcoded in the string literal? 😄

crates/burn-import/onnx-tests/tests/slice/slice_shape_runtime.py
```diff
@@ -0,0 +1,114 @@
...
+    print(f"Input shape: {input_data.shape}")
+    print(f"Shape input shape: {shape_input_data.shape}")
+    
+    # Simulate the operations
+    shape = np.array(shape_input_data.shape)  # [3, 4]
+    print(f"Extracted shape: {shape}")
+    
+    # The slice uses shape as ends, slicing axes 0 and 1
+    print(f"Slice on axes [0, 1]: starts=[0, 0], ends={shape}")
+    print(f"Expected output shape: [3, 4, 6]")
```
laggui: If you want to actually simulate the operations to get the expected results, you can use a reference evaluator with onnx.

See for example:
https://github.com/tracel-ai/burn/blob/d79e92ccee9015d28eedfe32fa4c33066316d7a1/crates/burn-import/onnx-tests/tests/space_to_depth/space_to_depth.py#L62-L67

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -1,24 +1,55 @@
...
 
     fn input_types(&self) -> Vec<crate::burn::Type> {
-        vec![self.input.clone(), self.index.clone()]
+        match &self.index {
+            GatherIndices::Runtime(index_type) => vec![self.input.clone(), index_type.clone()],
+            GatherIndices::Static(_) => vec![self.input.clone()],
+        }
     }
 
     fn forward(
```
laggui: Should probably split this into specialized functions to handle tensor input and shape input, e.g.

```rust
match &self.input {
    Type::Shape(_) => self.forward_shape_gather(scope, node_position),
    Type::Tensor(_) => self.forward_tensor_gather(scope, node_position),
    _ => panic!("Gather needs Tensor or Shape input, got {:?}!", self.input),
}
```

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -4,78 +4,273 @@ use burn::record::PrecisionSettings;
...
+                        for (i, range) in ranges.iter_mut().enumerate().take(num_dims) {
+                            let idx = proc_macro2::Literal::usize_unsuffixed(i);
+                            *range = quote! { #start_name[#idx]..#end_name[#idx] };
+                        }
+                    }
+                    _ => {
+                        // Single dimension slicing for scalars
+                        let (start_expr, end_expr) = self.get_slice_range_expressions();
+                        ranges[0] = quote! { #start_expr..#end_expr };
+                    }
```
laggui: Isn't there also the case for a 1D tensor input provided at runtime?

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -4,78 +4,273 @@ use burn::record::PrecisionSettings;
...
+                    for (i, range) in ranges.iter_mut().enumerate().take(num_dims) {
+                        let start = starts[i].to_tokens();
+                        let idx = proc_macro2::Literal::usize_unsuffixed(i);
+                        *range = quote! { #start..#end_name[#idx] };
+                    }
+                }
+                _ => {
+                    let (start_expr, end_expr) = self.get_slice_range_expressions();
+                    ranges[0] = quote! { #start_expr..#end_expr };
+                }
```
laggui: Same. Though not sure how starts would be static and ends would not

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -4,78 +4,273 @@ use burn::record::PrecisionSettings;
...
+                    for (i, range) in ranges.iter_mut().enumerate().take(num_dims) {
+                        let idx = proc_macro2::Literal::usize_unsuffixed(i);
+                        let end = ends[i].to_tokens();
+                        *range = quote! { #start_name[#idx]..#end };
+                    }
+                }
+                _ => {
+                    let (start_expr, end_expr) = self.get_slice_range_expressions();
+                    ranges[0] = quote! { #start_expr..#end_expr };
+                }
```
laggui: Same

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -4,78 +4,273 @@ use burn::record::PrecisionSettings;
...
+        };
+
+        (start_expr, end_expr)
+    }
+
+    fn generate_shape_slice(
+        &self,
+        shape: &crate::burn::ShapeType,
+        _scope: &mut Scope,
+        _node_position: usize,
```
laggui: If the args are not used, we don't need to have them 🙂

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -4,78 +4,273 @@ use burn::record::PrecisionSettings;
...
+                    let output_len = if end_val == i64::MAX {
+                        shape.rank.saturating_sub(start_val as usize)
+                    } else {
+                        (end_val as usize).saturating_sub(start_val as usize)
+                    };
+                    let output_rank = Literal::usize_unsuffixed(output_len);
+
+                    quote! {
+                        let #output: [usize; #output_rank] = #shape_name[s![#start..#end].into_ranges([#shape_len].into())[0].clone()].try_into().unwrap();
+                    }
```
laggui: Hmmm I think this could be simplified to a single block where we simply always clamp the start/end values

crates/onnx-ir/src/node/slice.rs
```diff
@@ -1,81 +1,94 @@
...
+                    return Some(SliceInput::Static(values));
+                }
+            }
+        }
+
+        // Fall back to attributes
+        for (key, value) in node.attrs.iter() {
+            if key == attr_name {
+                return Some(SliceInput::Static(value.clone().into_i64s()));
+            }
```
laggui: We can drop the slice attributes, this was only available in version 1. The last three versions of the op (10, 11 and 13) no longer take attributes. Version 1 should not be used in any model of opset 16 I believe.

crates/onnx-ir/src/node/squeeze.rs
```diff
@@ -38,24 +38,59 @@ pub fn squeeze_update_output(node: &mut Node) {
         node.attrs.get("axes").cloned().map(|v| v.into_i64s())
     };
 
-    let axes = axes.unwrap_or_else(|| panic!("Squeeze must specify an axis"));
+    // If axes is None, PyTorch squeezes all dimensions of size 1
```
laggui: You mean per ONNX spec, not pytorch?

crates/onnx-ir/src/node/squeeze.rs
```diff
@@ -38,24 +38,59 @@ pub fn squeeze_update_output(node: &mut Node) {
...
 
-    let input_rank = match &node.inputs[0].ty {
-        ArgType::Tensor(tensor) => tensor.rank,
-        ty => panic!("Squeeze: invalid input type: {ty:?}"),
-    };
+    match &node.inputs[0].ty {
+        ArgType::Tensor(tensor) => {
+            log::debug!("Squeeze input rank for {}: {}", node.name, tensor.rank);
+            let output_rank = if axes.is_empty() {
+                // When axes is empty, PyTorch squeezes all dimensions of size 1
```
laggui: Again, per ONNX spec right?

antimora:
@laggui Thanks for reviewing and catching my silly mistakes. I’ll fix them.

Regarding this comment:

> What I'm not sure about is the 1D tensor coercion to shape types when parsing the binary input types (see my comment for convert\_shape\_constants below). Maybe I missed something.

I couldn’t find the inline comment, but I think you’re referring to converting Constants's tensor to Shape. This happens when we do add/mul/div/sub on a Shape input with a constant, like in this graph:

image  

Previously, constants were stored as scalars or tensors in module attributes. Since we’re adding values to a shape, it’s more efficient to treat 1D tensors as shapes (1D arrays) rather than tensors.

Also, I’ve switched ONNX Shape storage to `[i64; N]` instead of `[usize; N]`:

* ONNX Shape outputs `i64` ([see ](https://onnx.ai/onnx/operators/onnx__Shape.html)).
* Some downstream ops (e.g., `Concat`) can create negative indices and I want the storage type remain the same through graph operations. `[i64; N]` should cover all values. For example, Mul/Add/Concat/etc that returns Shape will have `[i64; N]` storage. 

This introduces minor API changes (e.g., `expand`, `reshape`) but mostly just type adjustments.

I checked that `i64` works on all platforms, including 32-bit.

crates/onnx-ir/src/from_onnx.rs
```diff
@@ -267,6 +272,99 @@ impl OnnxGraphBuilder {
...
+            };
+
+            // Mark rank-1 tensor inputs for conversion
+            for input in &node.inputs {
+                if let ArgType::Tensor(tensor) = &input.ty {
+                    if tensor.rank == 1 {
+                        constants_to_convert.insert(input.name.clone(), shape_rank);
+                    }
+                }
+            }
```
laggui: This could coerce any other 1D tensor input for a binary op to a shape type, no? Because in the second pass you also have this match arm:

```rust
NodeType::Add | NodeType::Sub | NodeType::Mul | NodeType::Div => {
    // Update input types to match converted constants
    for input in &mut node.inputs {
        if let Some(&shape_rank) = constants_to_convert.get(&input.name) {
            if matches!(&input.ty, ArgType::Tensor(t) if t.rank == 1) {
                input.ty = ArgType::Shape(shape_rank);
                log::debug!(
                    "Updated input {} to Shape({})",
                    input.name,
                    shape_rank
                );
            }
        }
    }
}
```

so any second input to one of these binary ops will be marked/inserted as `constants_to_convert`, and then coerced to a shape type in the second pass.

Which could have unintended consequences. In your example it works because it matches the first match arm (constant node output).

Otherwise, the 1D constants treated as shapes when used with an actual shape output type makes sense as a heuristic.

laggui (commented):
> I couldn’t find the inline comment, but I think you’re referring to converting Constants's tensor to Shape. 

I also couldn't find it 😆 I guess I forgot to register the comment with the review yesterday, whoops.

Anyway, see my comment below (for real this time lol)

crates/burn-import/src/burn/node/slice.rs
antimora: There is and it explodes in combinations =P. 

Anyhow I have added it but it's messy at the moment.

antimora:
joytag_image_tagging.onnx is parsing mentioned in tracel-ai/burn#3494

antimora:
I have fixed all Shape related issues. I am able to build 
yolo11x (with current changes) and DeBERTa (with additional PR fixes that are coming in parallel).

@laggui, it should be ready to review. I realize it's a huge PR but it was helpful to fix shape related problems in one go. If you can't review all at once, you can add comments and I can fix as you're continue reviewing.

AdrianEddy:
Shouldn't this match also handle `Shape`? 
image

It panics here with the CLIP ViT-B-32 (Text) model

laggui:
@antimora will have a look at this today or tomorrow! 🙂

crates/burn-import/onnx-tests/tests/cast/cast_shape.py
```diff
@@ -0,0 +1,95 @@
...
+    # Test with onnxruntime if available
+    try:
+        import onnxruntime as ort
+        
+        # Create test data
+        test_input = np.ones((2, 3, 4), dtype=np.float32)
+        
+        # Run inference
+        sess = ort.InferenceSession(file_name)
+        results = sess.run(None, {"input": test_input})
```
laggui: We can use the `ReferenceEvaluator` from `onnx.reference` 🙂

crates/burn-import/onnx-tests/tests/cast/cast_shape_to_float.py
```diff
@@ -0,0 +1,103 @@
...
+    onnx_model = build_model()
+    file_name = "cast_shape_to_float.onnx"
+    onnx.save(onnx_model, file_name)
+    onnx.checker.check_model(file_name)
+    
+    print(f"Finished exporting model to {file_name}")
+    
+    # Test with onnxruntime if available
+    try:
+        import onnxruntime as ort
```
laggui: Same (`ReferenceEvaluator`)

crates/burn-import/onnx-tests/tests/concat/concat_shape.py
```diff
@@ -0,0 +1,122 @@
...
+    onnx_model = build_model()
+    file_name = "concat_shape.onnx"
+    onnx.save(onnx_model, file_name)
+    onnx.checker.check_model(file_name)
+    
+    print(f"Finished exporting model to {file_name}")
+    
+    # Test with onnxruntime if available
+    try:
+        import onnxruntime as ort
```
laggui: Same (`ReferenceEvaluator`)

crates/burn-import/onnx-tests/tests/concat/concat_shape_with_constant.py
```diff
@@ -0,0 +1,102 @@
...
+    onnx_model = build_model()
+    file_name = "concat_shape_with_constant.onnx"
+    onnx.save(onnx_model, file_name)
+    onnx.checker.check_model(file_name)
+    
+    print(f"Finished exporting model to {file_name}")
+    
+    # Test with onnxruntime if available
+    try:
+        import onnxruntime as ort
```
laggui: Same (`ReferenceEvaluator`)

crates/burn-import/onnx-tests/tests/reshape/reshape_3d_to_scalar.py
```diff
@@ -0,0 +1,61 @@
...
+    print("Model exported successfully to reshape_3d_to_scalar.onnx")
+    print("Model structure: Reshape([1, 1, 1] -> scalar)")
+    
+    # Verify with numpy
+    test_input = np.array([[[2.5]]], dtype=np.float32)
+    print(f"Test input shape: {test_input.shape}")
+    print(f"Test input value: {test_input}")
+    reshaped = test_input.reshape(())  # Reshape to scalar
+    print(f"Reshaped shape: {reshaped.shape}")
+    print(f"Reshaped value: {reshaped}")
```
laggui: Shouldn't the verification use the ONNX model? Otherwise the expected output could be mismatched

crates/burn-import/onnx-tests/tests/reshape/reshape_to_scalar.py
```diff
@@ -0,0 +1,60 @@
...
+    print("Model exported successfully to reshape_to_scalar.onnx")
+    print("Model structure: Reshape([1, 1] -> scalar)")
+    
+    # Verify with numpy
+    test_input = np.array([[1.5]], dtype=np.float32)
+    print(f"Test input shape: {test_input.shape}")
+    print(f"Test input value: {test_input}")
+    reshaped = test_input.reshape(())  # Reshape to scalar
+    print(f"Reshaped shape: {reshaped.shape}")
+    print(f"Reshaped value: {reshaped}")
```
laggui: Same comment regarding verification

crates/burn-import/onnx-tests/tests/reshape/reshape_with_1d_tensor.py
```diff
@@ -0,0 +1,88 @@
...
+    onnx_model = build_model()
+    file_name = "reshape_with_1d_tensor.onnx"
+    onnx.save(onnx_model, file_name)
+    onnx.checker.check_model(file_name)
+    
+    print(f"Finished exporting model to {file_name}")
+    
+    # Test with onnxruntime if available
+    try:
+        import onnxruntime as ort
```
laggui: `ReferenceEvaluator` ?

crates/burn-import/onnx-tests/tests/reshape/reshape_with_shape.py
```diff
@@ -0,0 +1,96 @@
...
+    onnx_model = build_model()
+    file_name = "reshape_with_shape.onnx"
+    onnx.save(onnx_model, file_name)
+    onnx.checker.check_model(file_name)
+    
+    print(f"Finished exporting model to {file_name}")
+    
+    # Test with onnxruntime if available
+    try:
+        import onnxruntime as ort
```
laggui: Same (`ReferenceEvaluator`)

crates/burn-import/onnx-tests/tests/slice/slice_shape_runtime.py
laggui: This wasn't changed ☝️

crates/burn-import/onnx-tests/tests/squeeze/squeeze_float.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+
+# used to generate model: squeeze_float.onnx
+# Test squeeze operation for float tensors to scalars
+
+import numpy as np
+import onnx
+from onnx import helper, TensorProto
+import onnxruntime as ort
```
laggui: Same (`ReferenceEvaluator`)

antimora:
Thanks for checking, @AdrianEddy!

I'll check it. I presume this is the onnx file: https://huggingface.co/Qdrant/clip-ViT-B-32-text/blob/main/model.onnx

crates/burn-import/onnx-tests/tests/squeeze/squeeze_shape_noop.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+    
+    # Create the model
+    model = helper.make_model(
+        graph, 
+        opset_imports=[helper.make_operatorsetid("", OPSET_VERSION)]
+    )
+    
+    # Skip shape inference to avoid validation errors
+    # onnx.checker.check_model(model, full_check=True)
```
laggui: So this is an invalid ONNX graph? Even the reference throws an error according to your script.

What's the justification behind supporting this squeeze no-op?

crates/burn-import/onnx-tests/tests/unsqueeze/unsqueeze_int_to_shape.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env python3
+
+# used to generate model: unsqueeze_int_to_shape.onnx
+
+import numpy as np
+import onnx
+from onnx import helper, TensorProto
+import onnxruntime as ort
```
laggui: Same (`ReferenceEvaluator`)

crates/burn-import/src/burn/node/cast.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// Input and output tensor types (required for codegen imports)
+    pub input_kind: Option<TensorKind>,
+    pub output_kind: Option<TensorKind>,
+}
+
+impl CastNode {
+    pub fn new(input: Type, output: Type, target_elem_type: ElementType) -> Self {
+        let (input_kind, output_kind) = match (&input, &output) {
+            (Type::Tensor(input_tensor), Type::Tensor(output_tensor)) => {
+                (Some(input_tensor.kind), Some(output_tensor.kind))
```
laggui: Why not just have this comparison directly in `register_imports`?

crates/burn-import/src/burn/node/concat.rs
```diff
@@ -1,40 +1,88 @@
...
+impl ConcatNode {
+    /// Create a new concat node
+    pub fn new(inputs: Vec<Type>, output: Type, dim: usize) -> Self {
+        Self {
+            inputs,
+            output,
+            dim,
+        }
+    }
+}
```
laggui: Couldn't we have just kept the `#[derive(new)]`?

crates/burn-import/src/burn/node/concat.rs
```diff
@@ -1,40 +1,88 @@
...
     fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
-        let dim = self.dim.to_tokens();
-        let inputs = self
-            .inputs
-            .iter()
-            .map(|t| scope.tensor_use_owned(t, node_position));
-
-        let output = &self.output.name;
+        match &self.output {
+            Type::Tensor(_) => {
```
laggui: Capture the output here

crates/burn-import/src/burn/node/concat.rs
```diff
@@ -1,40 +1,88 @@
...
+                let dim = self.dim.to_tokens();
+                let inputs = self.inputs.iter().map(|t| match t {
+                    Type::Tensor(tensor) => scope.tensor_use_owned(tensor, node_position),
+                    _ => panic!("Expected tensor input for tensor concatenation"),
+                });
+
+                let output = match &self.output {
+                    Type::Tensor(tensor) => &tensor.name,
+                    _ => panic!("Expected tensor output"),
+                };
```
laggui: Instead of having another match to capture it here.

crates/burn-import/src/burn/node/concat.rs
```diff
@@ -1,40 +1,88 @@
...
+                    }
+                }
 
-        quote! {
-            let #output = burn::tensor::Tensor::cat([#(#inputs),*].into(), #dim);
+                quote! {
+                    let #output: [i64; #output_rank] = {
+                        let mut result = alloc::vec::Vec::new();
+                        #(result.extend_from_slice(#shape_parts);)*
+                        result.try_into().unwrap()
```
laggui: Could we use `[..].concat().try_into().unwrap()` instead? (where `..` is the repeated pattern for `shape_parts`)

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -1,106 +1,199 @@
...
+                            quote! {
+                                let #output: [i64; #output_rank] = #index.to_data()
+                                    .iter::<i64>()
+                                    .map(|idx| #input_shape_name[idx as usize])
+                                    .collect::<alloc::vec::Vec<_>>()
+                                    .try_into()
+                                    .unwrap();
+                            }
+                        } else {
+                            unreachable!(
```
laggui: Maybe use panic instead? Just for semantics

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -1,106 +1,199 @@
...
         let input = match &self.input {
             Type::Tensor(in_tensor) => scope.tensor_use_owned(in_tensor, node_position),
-            Type::Shape(in_shape) => in_shape.to_tensor(),
-            _ => panic!("Gather needs Scalar or Shape input, got {:?}!", self.input),
+            _ => unreachable!(),
+        };
+        let input_rank = match &self.input {
+            Type::Tensor(in_tensor) => in_tensor.rank,
+            _ => unreachable!(),
         };
```
laggui: Use a single match instead:

```rust
let input = match &self.input {
    Type::Tensor(in_tensor) => in_tensor,
    _ => unreachable!(),
};
let input_rank = input.rank;
let input = scope.tensor_use_owned(input, node_position);
```

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -1,106 +1,199 @@
...
-                            let indices = Tensor::<B, 1, _>::from_data([#index], &*self.device);
-                            let slice = Tensor::select(#input, #dim, indices);
-                            let #output = slice.squeeze::<#output_rank>(#dim);
+
+                        if output_rank == 0 {
+                            // If output rank is 0, squeeze and keep as 1D tensor
+                            quote! {
+                                let indices = Tensor::<B, 1, _>::from_data([#index], &*self.device);
+                                let #output = Tensor::select(#input, #dim, indices);
+                            }
```
laggui: Shouldn't this be `Type::Scalar`? So not reachable technically. We could add a panic for sanity.

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -1,106 +1,199 @@
...
+                        } else {
+                            quote! {
+                                let indices = Tensor::<B, 1, _>::from_data([#index], &*self.device);
+                                let slice = Tensor::select(#input, #dim, indices);
+                                let #output = slice.squeeze::<#output_rank>(#dim);
+                            }
```
laggui: All good for now, but we could probably use `tensor.slice(...)` instead with the slice macro `s![]`, where every other dim is selected for the whole range and only the index is used for dim (e.g., `s![.., #index, .., ..]` for a 4D tensor and dim = 1).

Maybe add a note?

crates/burn-import/src/burn/node/reshape.rs
```diff
@@ -1,32 +1,235 @@
...
+                quote! {
+                    let #output_name: [i64; #output_rank] = {
+                        let reshaped = #input.reshape([#output_rank]);
+                        let data = reshaped.into_data();
+                        let values = data.as_slice::<i64>().unwrap();
+                        let mut result = [0i64; #output_rank];
+                        result.copy_from_slice(&values[..#output_rank]);
+                        result
+                    };
+                }
```
laggui: More of a question than comment, but when is this desirable? Or even possible? Not sure what it represents concretely to do tensor -> reshape -> shape output (maybe we should panic instead?). Going from tensor -> shape with `into_data()` works, but not great for sync points when model is executed on GPU.

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -1,107 +1,516 @@
...
+        output: &proc_macro2::Ident,
+    ) -> TokenStream {
+        let input = scope.tensor_use_owned(tensor, node_position);
+        let rank = tensor.rank;
+        let mut ranges = vec![quote! { .. }; rank];
+
+        // Build slice ranges based on parameter types
+        match (&self.starts, &self.ends) {
+            // Both static: simple case
+            (SliceParam::Static(starts), SliceParam::Static(ends)) => {
```
laggui, line 54: For compatibility handling all of the other input pairs is good, but they all require a sync point (`to_data()`) which is not great 😅 We need to figure out a better way to handle this.

Also, do we need to handle mismatched pairs (e.g., static + runtime combos).

crates/burn-import/src/burn/node/unsqueeze.rs
```diff
@@ -43,22 +40,93 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for UnsqueezeNode {
...
+                //    Many ONNX models use this pattern extensively for dynamic shapes, and
+                //    keeping the data as Shape type is much more efficient.
+                //
+                // 2. Zero-cost conversion: Both scalars and Shape arrays are CPU-resident,
+                //    so this is just a type system change with no runtime overhead.
+                //
+                // 3. Maintains type symmetry: This is the exact reverse of the squeeze
+                //    operation (Shape[1] -> Scalar), ensuring bidirectional compatibility.
+                //
+                // 4. Lazy GPU transfer: If a downstream operation needs this on GPU, it can
```
laggui: I wouldn't say a CPU->GPU data transfer is lazy 😅 maybe you meant "deferred GPU transfer"?

Also, not sure if this big block comment is absolutely necessary? We can probably keep it much shorter.

crates/burn-import/src/burn/node/slice.rs
```diff
@@ -1,107 +1,516 @@
...
+                let start_vec_var = quote! { start_vec };
+                let end_data_var = quote! { end_data };
+                let end_vec_var = quote! { end_vec };
+
+                // Build ranges for each dimension
+                let range_exprs: Vec<_> = (0..rank).map(|i| {
+                    let idx = proc_macro2::Literal::usize_unsuffixed(i);
+                    quote! {
+                        #start_vec_var.get(#idx).map(|&s| s as usize).unwrap_or(0)..#end_vec_var.get(#idx).map(|&e| e as usize).unwrap_or(#input_dims_var[#idx])
+                    }
```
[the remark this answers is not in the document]
antimora: it's generated code;

crates/onnx-ir/src/node/slice.rs
```diff
@@ -1,81 +1,87 @@
...
-    // Convert negative axes indices to positive (counting from the end)
-    for axis in &mut axes {
-        if *axis < 0 {
-            *axis += input_rank as i64;
-        }
+    // Validate steps if present
+    if let Some(SliceInput::Static(ref step_values)) = steps
+        && step_values.iter().any(|&x| x != 1)
+    {
+        panic!("Slice: steps other than 1 are not supported");
```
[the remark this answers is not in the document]
antimora: Want to keep separate.

crates/onnx-ir/src/node/squeeze.rs
```diff
@@ -38,24 +38,59 @@ pub fn squeeze_update_output(node: &mut Node) {
...
-    };
+    match &node.inputs[0].ty {
+        ArgType::Tensor(tensor) => {
+            log::debug!("Squeeze input rank for {}: {}", node.name, tensor.rank);
+            let output_rank = if axes.is_empty() {
+                // When axes is empty, ONNX spec squeezes all dimensions of size 1
+                // Without static shape info, we can't know which dims are size 1
+                // The output type will be corrected later if ONNX provides it
+                // TODO: Infer rank from output tensor shape based on static shape inference
+                todo!("Output tensor rank must be inferred from static shape inference");
```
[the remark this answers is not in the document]
antimora: OK. Clever idea. Not a complete solution but kinda works.

crates/burn-import/onnx-tests/tests/squeeze/squeeze_shape_noop.py
antimora: the test to still demonstrate the Shape(n) where n>1 remains unchanged behavior (the actual no-op case). I have fixed the script.

crates/burn-import/src/burn/node/gather.rs
antimora: Fixed it

crates/burn-import/src/burn/node/slice.rs
antimora: Filed async issue https://github.com/tracel-ai/burn/issues/3557

crates/burn-import/src/burn/node/reshape.rs
antimora: I removed it and it will panic.

crates/burn-import/src/burn/node/cast.rs
```diff
@@ -0,0 +1,333 @@
...
+
+impl CastNode {
+    pub fn new(input: Type, output: Type, target_elem_type: ElementType) -> Self {
+        Self {
+            input,
+            output,
+            target_elem_type,
+        }
+    }
+}
```
laggui: `#[derive(new)]` but not really important

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -387,11 +556,60 @@ mod tests {
...
+                    let tensor2 = sliced.squeeze::<1usize>(0);
+                    tensor2
+                }
+            }
+        };
+
+        assert_tokens(graph.codegen(), expected);
+    }
+
+    // Scalar output test removed - no longer supported
```
laggui: Guess it is back to being supported with your last changes? 😄

laggui (approved):
Some minor comments remaining, otherwise LGTM.

We can at least merge this big PR after and work on top of that for further improvements.

AdrianEddy:
I want to express my gratitude to @antimora (and the entire burn team too) for working on this. It's amazing to see burn getting closer and closer to supporting a wide range of ONNX models and allowing more apps to switch to burn in production. Even though this PR doesn't fix models I'm interested in yet, it's a huge step forward I'm excited to see burn getting better and better!

Thank you ❤️

antimora:
Thank you, @AdrianEddy! I’ll be working on your model next.

**burn-import** is my baby, and I want to see it grow up and become truly useful. I genuinely believe Burn’s ONNX feature will be its superpower! The community’s support has been incredibly motivating for me personally.

PR #3382, opened 2025-07-17
ci: split tests on GitHub runners and on GPU runners

syl20bnr:
### Checklist

### Changes

We execute GPU tests when the PR as label `ci:test-gpu` to avoid unnecessary GPU usage.


syl20bnr:
~~Setting this in WIP the time I add automatic labelling and add back wgpu to the GitHub runners.~~

No need for now.

PR #3383, opened 2025-07-17
fix(onnx-test): mismatch between expected and actual output types (#3370)

### LINKED ISSUE #3370: ONNX import tests fail or do not compile when switching backend from Ndarray to Autodiff<Wgpu>
**Describe the bug**
When switching the backend from burn_ndarray::NdArray<f32> to burn_autodiff::Autodiff<burn_wgpu::Wgpu>, and compiling and running the tests under crates/burn-import/onnx-tests/,  several ONNX import tests fail.

**To Reproduce**
1. in ``crates/burn-import/onnx-tests/Cargo.toml``, add:
```
[features]
backend-autodiff-wgpu = ["burn-autodiff", "burn-wgpu"]

[dependencies]
burn-autodiff = { path = "../../burn-autodiff", version = "0.18.0", default-features = false, optional = true }
burn-wgpu = { path = "../../burn-wgpu", version = "0.18.0", optional = true, default-features = false }
```

2.  in ``crates/burn-import/onnx-tests/tests/test_mod.rs``, add: 
```
#[cfg(feature = "backend-autodiff-wgpu")]
type Backend = burn_autodiff::Autodiff<burn_wgpu::Wgpu>;

#[cfg(not(feature = "backend-autodiff-wgpu"))]
type Backend = burn_ndarray::NdArray<f32>;
```

3. under ``crates/burn-import/onnx-tests/tests/``, replace:
```
type Backend = burn_ndarray::NdArray<f32>
```
with:
```
use super::super::Backend;
```

in the ``mod.rs`` files, by running script under ``crates/burn-import/onnx-tests/tests/``:
```
find ./ -name mod.rs -exec sed -i '' 's|^[[:space:]]*type Backend = burn_ndarray::NdArray<f32>;|use super::super::Backend;|' {} +
```

4. under ``crates/burn-import/onnx-tests/``, run:
```
cargo test --test test_mod --features backend-autodiff-wgpu
```

got:
```
  --> crates/burn-import/onnx-tests/tests/constant_of_shape/mod.rs:41:9
   |
41 |         assert!(f_output.equal(f_expected).all().into_scalar());
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `bool`, found `u32`

  --> crates/burn-import/onnx-tests/tests/constant_of_shape/mod.rs:42:9
   |
42 |         assert!(i_output.equal(i_expected).all().into_scalar());
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `bool`, found `u32`

  --> crates/burn-import/onnx-tests/tests/constant_of_shape/mod.rs:43:9
   |
43 |         assert!(b_output.equal(b_expected).all().into_scalar());
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `bool`, found `u32`

```

After commenting out ``pub mod constant_of_shape;``  in the  ``test_mod.rs``,  run (4) again, got:
```
running 134 tests
test argmax::tests::argmax ... FAILED
test and::tests::and ... FAILED
test ceil::tests::ceil_test ... ok
test argmin::tests::argmin ... FAILED
test add::tests::add_scalar_to_int_tensor_and_int_tensor_to_int_tensor ... FAILED
......
test result: FAILED. 50 passed; 84 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.36s
```
**Expected behavior**
All code pass compiling and testing.

 - Device: Apple M1 Pro
 - OS: mac OS 
 - Version: 14.2.1 (23C71)
 
**Additional context**
The git repo version is: 
```
commit b42a8b6556bf0f87077cdd0b2b5c0756f90ec8cb (HEAD -> main, origin/main, origin/HEAD)
Author: Jimmy Johnson <catch22@fastmail.net>
Date:   Fri Jul 11 15:13:38 2025 +0200

    Updating documentation description for nonzero and nonzero_async (#3368)
```

lucianyao:
See: https://github.com/tracel-ai/burn/issues/3370#issuecomment-3080046479

The root cause is that the WGPU backend uses `i32` and `u32` to represent integer and boolean values, respectively. However, the expected values in tests were defined using `i64` and `bool`, which align with the default types used by the NdArray backend.

To address this mismatch, new features such as `int-i32` and `bool-u32` are introduced to allow switching the expected types depending on the backend.

### Considered alternatives:
1. **Implementing `From` for `TensorData`**: This would reduce boilerplate but obscure the actual data types, which ONNX test contributors are expected to be aware of.
2. **Using helper functions to convert values**: Given that both backend type and expected values vary across tests, such helpers would have limited use and add unnecessary complexity. Furthermore, they make tests less self-contained and harder to reason about locally.

### Checklist

### Related Issues/PRs

Partially fixes #3370 

### Testing

- **Regression test**:   
  run `cargo test` in`crates/burn-import/onnx-tests`; all tests passed.
- **WGPU feature tests**: 
  run `cargo test --features backend-autodiff-wgpu-default`; all type-mismatch-related tests passed.
- **Full test suite**: 
  run `RUSTC_BOOTSTRAP=1 RUSTFLAGS="-Zmacro-backtrace" cargo run-checks` from the repository root; all test passed.

### DIFF
--- a/crates/burn-import/onnx-tests/Cargo.toml
+++ b/crates/burn-import/onnx-tests/Cargo.toml
@@ -6,6 +6,10 @@ license.workspace = true
 
 [features]
 backend-autodiff-wgpu = ["burn-autodiff", "burn-wgpu"]
+int-i32 = []
+bool-u32 = []
+
+backend-autodiff-wgpu-default = ["backend-autodiff-wgpu", "int-i32", "bool-u32"]
 
 [dev-dependencies]
 burn = { path = "../../burn" }
--- a/crates/burn-import/onnx-tests/README.md
+++ b/crates/burn-import/onnx-tests/README.md
@@ -185,6 +185,20 @@ fn test_my_new_op() {
 
 Your test will be automatically included in the main test suite through `tests/test_mod.rs`.
 
+Note:
+- Be aware that different backends use different data types. For example, the `NdArray` backend uses native `bool`, while the `Wgpu` backend uses `u32` for boolean values. Feature flags are used to select the appropriate types depending on the backend. In test code, you might write:
+
+```rust
+    #[cfg(feature = "bool-u32")]
+    let expected = TensorData::from([[1u32, 0, 1, 1]]);
+
+    #[cfg(not(feature = "bool-u32"))]
+    let expected = TensorData::from([[true, false, true, true]]);
+```
+- Refer to `Cargo.toml` and `backend.rs` for how these features are defined and used.
+
+- The `bool-u32` feature only takes effect when using the `backend-autodiff-wgpu` backend. When using `NdArray`, native `bool` is always used, regardless of feature flags.
+
 ## Best Practices for ONNX Testing
 
 ### Model Generation
@@ -234,7 +248,7 @@ This command runs all tests using the default backend: `burn_ndarray::NdArray<f3
 
 To run tests with an alternative backend (e.g. `burn_autodiff::Autodiff<burn_wgpu::Wgpu>`), use:
 ```
-cargo test --features backend-autodiff-wgpu
+cargo test --features backend-autodiff-wgpu-default
 ```
 Currently supported backends:
 *	`burn_ndarray::NdArray<f32> (default)`
@@ -248,7 +262,7 @@ cargo test --test test_mod my_new_op::test_my_new_op
 
 Run a specific test with a selected backend:
 ```
-cargo test --test test_mod my_new_op::test_my_new_op --features backend-autodiff-wgpu
+cargo test --test test_mod my_new_op::test_my_new_op --features backend-autodiff-wgpu-default
 ```
 
 ## Debugging Failed Tests
--- a/crates/burn-import/onnx-tests/tests/add/mod.rs
+++ b/crates/burn-import/onnx-tests/tests/add/mod.rs
@@ -34,6 +34,11 @@ mod tests {
         let input = Tensor::<Backend, 4, Int>::from_ints([[[[1, 2, 3, 4]]]], &device);
         let scalar = 2;
         let output = model.forward(input, scalar);
+
+        #[cfg(feature = "int-i32")]
+        let expected = TensorData::from([[[[9i32, 11, 13, 15]]]]);
+
+        #[cfg(not(feature = "int-i32"))]
         let expected = TensorData::from([[[[9i64, 11, 13, 15]]]]);
 
         output.to_data().assert_eq(&expected, true);
--- a/crates/burn-import/onnx-tests/tests/argmax/mod.rs
+++ b/crates/burn-import/onnx-tests/tests/argmax/mod.rs
@@ -18,6 +18,11 @@ mod tests {
         // Run the model
         let input = Tensor::<Backend, 2>::from_floats([[1., 2., 3.], [4., 5., 6.]], &device);
         let output = model.forward(input);
+
+        #[cfg(feature = "int-i32")]
+        let expected = TensorData::from([[2i32], [2]]);
+
+        #[cfg(not(feature = "int-i32"))]
         let expected = TensorData::from([[2i64], [2]]);
 
         output.to_data().assert_eq(&expected, true);
--- a/crates/burn-import/onnx-tests/tests/argmin/mod.rs
+++ b/crates/burn-import/onnx-tests/tests/argmin/mod.rs
@@ -21,6 +21,11 @@ mod tests {
             &device,
         );
         let output = model.forward(input);
+
+        #[cfg(feature = "int-i32")]
+        let expected = TensorData::from([[2i32], [0]]);
+
+        #[cfg(not(feature = "int-i32"))]
         let expected = TensorData::from([[2i64], [0]]);
 
         output.to_data().assert_eq(&expected, true);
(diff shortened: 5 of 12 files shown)


lucianyao:
**@antimora What’s your preferred direction? I’d love to hear your thoughts if you have an alternative in mind.**

So far, I’ve explored three possible solutions:

**1.	Using a From implementation for TensorData**

This would allow expected values to be written in one canonical form (e.g., using bool), and convert them automatically to the backend’s underlying type (e.g., u32).
It’s arguably the cleanest solution syntactically, but it obscures the concrete data types, which test authors should be aware  of, which may increase the risk of subtle or misleading test behavior.

**2.	Using helper functions to construct expected tensors**

This can reduce repetition and encapsulate some of the backend-specific logic.
However, since both the data shape and backend types vary between tests, these helpers often become fragmented or overly specialized. It also makes test expectations less self-contained, which can hurt clarity and maintainability.

**3.	Using backend-specific feature flags to define expected types**
For example:
```rust
#[cfg(feature = "backend-autodiff-wgpu")]
let expected = TensorData::from([1u32, 0u32, ...]);
```
This approach mixes data type concerns with backend features, which aren’t strictly orthogonal.
For instance, both Candle and NdArray may use i64 for integers, so using a feature like int-i64 (decoupled from backend) would enable reuse of expected values. By contrast, tying expectations directly to a backend-* feature leads to duplication and inflexibility.

Additionally, some backends support multiple valid type configurations (e.g., WGPU can use either i32 or i64 if supported). Supporting and testing these configurations cleanly requires a way to decouple backend selection from data type selection. Encoding all combinations into backend features would either be inflexible or lead to an explosion of feature flags.

Let me know if you’d prefer a different direction — happy to adjust the implementation if there’s a better balance between usability and correctness.

antimora:
What are your own thoughts? I feel I am chatting with an AI agent :-). Please dive deeper and see if you are actually improving the code and usability. You can't burden the reviewers with AI-generated solutions without thoroughly vetting them yourself first. You have to stand by them and make sure what is submitted makes sense. I am all for AI-assisted coding, but this is not the right approach.

lucianyao:
I am not a native english speaker, so I use AI to help me   rephrase it and polish it to a more clear expression and try not be rude, otherwise my English would be difficult to comprehend and sometime even offensive.

You could check this url: https://chatgpt.com/share/687a22ee-e158-800f-a3d3-5710baa078d5

Here is some pieces of the conversions:

(AI suggest using helper fn, which I already abandoned, I gave my reasons)
> the problem is that: 1. not all backends use u32 as boolean. 2. the user of the tests should be aware of what the data types are used in their backend

(I suggest using From)
> there might be some even more simple way to address it, for example, implement From for the Tensor data, so if the user use "into", I thin the cimpiler will automatically convert to the type needed, it is even more simple, however, it hides the data type which  the user should be aware of

(I did not know how to say in English these words: "稀里糊涂的" and "正交的". )
> I like to response: what is your opinion, do you prefer an alternative solution?  I have 3 solutions so far: 1) use From trait for TensorData to automatically convert the "expected value"'s type to the underline data type of the tensor. This is the most elegant solution probably if it works, however, since it convert data implicitly, the testers may do the tests 稀里糊涂的, which causes errors prone.   2) use some helper functions to do the job, however,   while new test for new backend type and new tensor structure adding in, the helper functions can be segment probably, also, it is not local addressed. 3) just use the backend feature to define the expected values, for example, #[cfg(feature="backend-autodiff-wgpu")] let expected = TensorData.from([1u32, 0u32, ...],  this problem I think is the mixing the data type with the backend, which are not 正交的, which causes difficult to extend, for example, both candle and Ndarray use i64 as integer, if using feature int-i64, we can define the expected value using i64 once, if only using the backend feature, we have to define twice.


I admit I do not very familiar with nn and even Rust ( this is the first project I am on in Rust), sometimes I ask AI if they can provide some alternative solutions, you could check all the info in the previous log to see that most of the idea come from me, the only solution provided by AI is to make some helper functions, which I think is some kind of default solution can be figure out at the first place and I abandoned it at the very beginning, however, AI gave me that solution but denied by me.

Every piece of the thought in the previous comment is from me, maybe even which is dumb, but it is from me(just check the log pls). I let AI to rephrase them, which looks too verbose maybe? but to be honest, I really sincerely wanted to discuss to figure out a better solution. I can implement the alternative solutions in the days to see the real result instead of just thinking "in the air"

btw, I do not use ai-assistant coding, I am using RustRover, which provides local simple sometimes even stupid code completion. I do ask AI if how to write some "feature"(not that "feature" in Rust), and gave me examples to learn. But I write every piece of code.

antimora:
@lucianyao, thank you for the context. We welcome contributors of all levels and backgrounds. Please continue using the tools that help you. I don't want to discourage that. My main point was about ensuring you fully understand and can explain any AI generated solutions before proposing them. These tools should augment your work not replace your understanding.

I haven't investigated this issue thoroughly yet but I believe the solution likely involves making the correct API calls to compare results. Any solution should be consistent with existing and future tests while remaining easy to use. If we're missing the necessary API functionality, we should either enhance what we have or propose new additions.

PR #3390, opened 2025-07-18
Pre-Shuffle Multithread DataLoaders on Shuffle

crutcher:
### Checklist

### Changes

Pre-shuffle MultiThreadedDataLoader before split (when shuffle is enabled) to unbias the worker datasets.

### Testing

No tests at the moment (there aren't distribution tests for DLs yet).

### DIFF
--- a/crates/burn-core/src/data/dataloader/multithread.rs
+++ b/crates/burn-core/src/data/dataloader/multithread.rs
@@ -264,12 +264,12 @@ impl<O: std::fmt::Debug> Iterator for MultiThreadsDataloaderIterator<O> {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashSet;
-
     use super::*;
     use crate::data::dataloader::FixBatchStrategy;
     use crate::data::dataloader::batcher::TestBatcher;
     use crate::data::dataset::FakeDataset;
+    use burn_dataset::InMemDataset;
+    use std::collections::HashSet;
 
     #[test]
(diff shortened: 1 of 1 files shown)


crutcher:
Added tests; with no-shuffle and shuffle behavior.

crates/burn-core/src/data/dataloader/multithread.rs
```diff
@@ -87,7 +87,17 @@ where
...
-                let datasets = PartialDataset::split(self.dataset.clone(), self.num_threads);
+                let mut dataset = self.dataset.clone();
+                if let Some(rng) = self.rng.as_ref() {
+                    // Pre-shuffle the dataset before split if shuffle is enabled.
+                    // This ensures that each thread gets a uniform random sample of the dataset.
+                    let mut rng = rng.clone();
+                    dataset = Arc::new(burn_dataset::transform::ShuffledDataset::new(
+                        dataset, &mut rng,
+                    ));
+                }
```
nathanielsimard, line 98: Did you remove the suffling after the initialization?

crutcher: @nathanielsimard that is the shuffling.

`shuffle` is enabled if there's an attached `rng`; and in that situation, we shuffle the dataset before we pass it to partition.

there's an attached test that explores both behaviors.

crutcher:
Boop?

PR #3406, opened 2025-07-21
Add SelectionDataset, refactor ShuffledDataset, and add transform tests.

crutcher:
SelectionDataset is morally a more abstract replacement for ShuffledDataset; all uses could be migrated.

### Checklist

### Changes

When looking at partitioning and over-selection; I noticed that ShuffledDataset is overly constrained, there's nothing requiring a bijection. So based upon discord thread discussions, this is SelectionDataset.

I also added tests to several transforms; and refactored ShuffledDataset for testing and generality.

### Testing

Full and additional behavior tests.

### DIFF
--- a/burn-book/src/building-blocks/dataset.md
+++ b/burn-book/src/building-blocks/dataset.md
@@ -30,14 +30,15 @@ Transformations in Burn are all lazy and modify one or multiple input datasets. 
 transformations is to provide you with the necessary tools so that you can model complex data
 distributions.
 
-| Transformation    | Description                                                                                                              |
-| ----------------- | ------------------------------------------------------------------------------------------------------------------------ |
-| `SamplerDataset`  | Samples items from a dataset. This is a convenient way to model a dataset as a probability distribution of a fixed size. |
-| `ShuffledDataset` | Maps each input index to a random index, similar to a dataset sampled without replacement.                               |
-| `PartialDataset`  | Returns a view of the input dataset with a specified range.                                                              |
-| `MapperDataset`   | Computes a transformation lazily on the input dataset.                                                                   |
-| `ComposedDataset` | Composes multiple datasets together to create a larger one without copying any data.                                     |
-| `WindowsDataset`  | Dataset designed to work with overlapping windows of data extracted from an input dataset.                               |
+| Transformation     | Description                                                                                                              |
+|--------------------|--------------------------------------------------------------------------------------------------------------------------|
+| `SamplerDataset`   | Samples items from a dataset. This is a convenient way to model a dataset as a probability distribution of a fixed size. |
+| `SelectionDataset` | Selects a subset of items by index from a dataset. Can be randomly shuffled; can be re-shuffled.                         |
+| `ShuffledDataset`  | Shuffles a wrapped dataset; This is a thin wrapper around `SelectionDataset`.                                            |
+| `PartialDataset`   | Returns a view of the input dataset with a specified range.                                                              |
+| `MapperDataset`    | Computes a transformation lazily on the input dataset.                                                                   |
+| `ComposedDataset`  | Composes multiple datasets together to create a larger one without copying any data.                                     |
+| `WindowsDataset`   | Dataset designed to work with overlapping windows of data extracted from an input dataset.                               |
 
 Let us look at the basic usages of each dataset transform and how they can be composed together.
 These transforms are lazy by default except when specified, reducing the need for unnecessary
@@ -59,9 +60,33 @@ let dataset: DbPedia = HuggingfaceDatasetLoader::new("dbpedia_14")
 let dataset = SamplerDataset<DbPedia, DbPediaItem>::new(dataset, 10000);
 ```
 
+- **SelectionDataset**: This transform can be used to select a subset of items from a dataset by
+  index. It can be initialized with a list of indices to select from the input dataset. This is
+  particularly useful when you want to create a smaller dataset from a larger one, for example, to
+  create a validation set from a training set.
+
+  The `SelectionDataset` can also be initialized with a random seed to shuffle the indices before
+  selection. This is useful when you want to randomly select a subset of items from the dataset.
+ 
+  Base dataset items may be included more than once in the selection.
+
+```rust, ignore
+let explicit = SelectionDataset::from_indicies_checked(dataset.clone(), vec![0, 1, 2, 0]);
+
+let shuffled = SelectionDataset::new_shuffled(dataset.clone(), &mut rng);
+let shuffled = SelectionDataset::new_shuffled_with_seed(dataset.clone(), 42);
+
+let mut mutable = SelectionDataset::new_select_all(dataset.clone(), vec![0, 1, 2, 0]);
+mutable.shuffle_with_seed(42);
+mutable.shuffle(&mut rng);
+```
+
+
 - **ShuffledDataset**: This transform can be used to shuffle the items of a dataset. Particularly
   useful before splitting the raw dataset into train/test splits. Can be initialized with a seed to
   ensure reproducibility.
+
+  The `ShuffledDataset` is a thin wrapper around the `SelectionDataset`.
 
 ```rust, ignore
 let dataset = ShuffledDataset<DbPedia, DbPediaItem>::with_seed(dataset, 42);


crates/burn-dataset/src/transform/random.rs
```diff
@@ -10,18 +10,29 @@ pub struct ShuffledDataset<D, I> {
     input: PhantomData<I>,
 }
 
+/// Generates a shuffled vector of indices up to a size.
```
nathanielsimard: The `ShuffleDataset` could have a `SelectionDataset` internaly to avoid implementing the same logic twice while being compatible with the older API.

crutcher: Done!

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,233 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if any index is out of bounds for the dataset.
+    pub fn new(dataset: D, indices: Vec<usize>) -> Self {
+        let size = dataset.len();
+
+        if let Some(idx) = indices.iter().find(|&i| *i >= size) {
+            panic!("Index out of bounds for wrapped dataset size: {idx} >= {size}");
+        }
```
nathanielsimard, line 85: I think we should not perform that check. Some datasets are huge, and doing a gazillion comparisons might not be OK. You could expose a function to perform the check or provide a checking strategy (out-of-bounds values are clipped, etc.).

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,233 @@
...
+    ///
+    /// * `dataset` - The original dataset to select from.
+    ///
+    /// # Returns
+    ///
+    /// A new `SelectionDataset` that selects all indices from the dataset.
+    pub fn new_select_all(dataset: D) -> Self {
+        let size = dataset.len();
+        Self::new(dataset, iota(size))
+    }
```
nathanielsimard: It's kind of bad to use, since we create an indices vector for nothing if we don't perform any transformation afterward. It should be in the doc that some transformation should be done for it to be useful.

crutcher:
Based upon comments; I made the constructor names explicit.

The full-selection/iota constructor was there for further manipulations, but we hadn't really had much more than shuffle in place. I'd always planned on implementing something like PartialDataset::Split; so I did that as well (and stole the Arc<D> trick).

crates/burn-dataset/src/transform/selection.rs
crutcher: made this much more explicit.

crates/burn-dataset/src/transform/selection.rs
crutcher: made checked and unchecked versions.

nathanielsimard (approved):
LGTM, @laggui could check before merging

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,407 @@
...
+///
+/// # Arguments
+///
+/// * `size` - The size of the dataset.
+///
+/// # Returns
+///
+/// A vector containing indices from 0 to size - 1.
+#[inline(always)]
+pub fn iota(size: usize) -> Vec<usize> {
```
laggui, line 19: Why is this public exactly?

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,407 @@
...
+///
+/// # Arguments
+///
+/// * `size` - The size of the dataset to shuffle.
+///
+/// # Returns
+///
+/// A vector of shuffled indices.
+#[inline(always)]
+pub fn shuffled_indices(size: usize, rng: &mut StdRng) -> Vec<usize> {
```
laggui, line 34: Same here, also the function is only used in tests it seems.

crates/burn-dataset/src/transform/random.rs
```diff
@@ -1,38 +1,63 @@
...
 impl<D, I> ShuffledDataset<D, I>
 where
     D: Dataset<I>,
+    I: Clone + Send + Sync,
 {
-    /// Creates a new shuffled dataset.
+    /// Creates a new selection dataset with shuffled indices.
+    ///
+    /// Selects every index of the dataset and shuffles them
+    /// using `shuffled_indices`.
```
laggui: Copy-pasta leftover? `shuffled_indices` is not used 🤔

crates/burn-dataset/src/transform/random.rs
```diff
@@ -1,38 +1,63 @@
...
-            input: PhantomData,
+            wrapped: SelectionDataset::new_shuffled(dataset, rng),
         }
     }
 
-    /// Creates a new shuffled dataset with a fixed seed.
+    /// Creates a new selection dataset with shuffled indices using a fixed seed.
+    ///
+    /// Selects every index of the dataset and shuffles them
+    /// using `shuffled_indices`; seeds the random number generator with the provided seed.
```
laggui: Same copy-pasta?

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,407 @@
...
+    {
+        let dataset = dataset.into();
+        let size = dataset.len();
+        Self::from_indices_unchecked(dataset, iota(size))
+    }
+
+    /// Creates a new selection dataset with shuffled indices.
+    ///
+    /// Selects every index of the dataset and shuffles them
+    /// using `shuffled_indices`.
```
laggui: Same 🙂

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,407 @@
...
+    {
+        let mut this = Self::new_select_all(dataset);
+        this.shuffle(rng);
+        this
+    }
+
+    /// Creates a new selection dataset with shuffled indices using a fixed seed.
+    ///
+    /// Selects every index of the dataset and shuffles them
+    /// using `shuffled_indices`; seeds the random number generator with the provided seed.
```
laggui: Same 🙂 

I guess it was re-used initially but changed through the commits!

crates/burn-dataset/src/transform/selection.rs
```diff
@@ -0,0 +1,407 @@
...
+        self.shuffle(&mut rng);
+    }
+
+    /// Creates a new dataset that is a slice of the current selection dataset.
+    ///
+    /// Slices the *selection indices* from `[start..end]`.
+    ///
+    /// Independent of future shuffles on the parent, but shares the same wrapped dataset.
+    ///
+    /// TODO: RangeArg in burn-tensor should be lifted to burn-common.
```
laggui: Good point! Would be nicer to have the syntax using more flexible ranges.

For now I would keep the TODO comment (`//`), just not part of the doc comments (`///`) 😅

crates/burn-dataset/src/transform/selection.rs
crutcher: This is public because users wishing to construct a selection are likely to want it as a utility method.

There's an argument to lift it into `burn::tensor::indexing` along with `shuffled_indices`; but I think the entire thing should be lifted above the `tensor` namespace. `burn::indexing` seems confusing; but maybe a `burn::{lib,core}::indexing` (and reorg other base support machinery there).

crates/burn-dataset/src/transform/selection.rs
crutcher: Same justification; if you, as a user, want to construct a selection before calling the new function; this is a useful utility method.

crates/burn-dataset/src/transform/random.rs
crutcher: It's not copy-pasta; but it's not as helpful as listing the proximate wrapper call. Updated.

crates/burn-dataset/src/transform/random.rs
crutcher: Same; updated.

crates/burn-dataset/src/transform/selection.rs
crutcher: Again; this is accurate. I can remove the specificity.

crates/burn-dataset/src/transform/selection.rs
crutcher: Again; it's accurate, but I'll make it less.

crates/burn-dataset/src/transform/selection.rs
crutcher: Done.

crutcher:
Replied / updated.

crutcher:
Landed changes from review; and rebased.

crates/burn-dataset/src/transform/selection.rs
laggui: Ahhh ok that makes sense, thanks for clarifying!

laggui (approved):
Thanks for clarifying! And making the changes to the book section 🙏

PR #3407, opened 2025-07-21
Add Dice-Sorenson Coefficient Metric

MathijsdeBoer:
## Vision Metrics

I saw that the current set metrics did not contain vision metrics. For this pull request, I have currently implemented the Dice-Sorenson Coefficient, and placed it in its own module to differentiate it from the other metrics. I could've added more metrics to this, but I figured I'd open a PR early, to get some feedback from the maintainers regarding organization, codestyle, etc. or if there even is an interest to include these metrics in the core Burn project. (as opposed to a spinoff)

I mainly come from the usual PyTorch scene, and am learning to use Rust in my research. I see Burn as a solid option to move some of our research into a faster language than Python for inference when we're trying to get some of it into clinical practice.

### Checklist

- [ x ] Confirmed that `cargo run-checks` command has been executed.

### Related Issues/PRs

None

### Changes

Add `DiceMetric` to `burn-train`'s `metric/vision` module.

### Testing

Unit testing code has been included to evaluate the following cases:
- No overlap
- Partial overlap
- Perfect overlap
- Empty ground truth

### LLM Notice

I used Copilot in VS Code to set up the scaffolding for this metric, based on the implementations in `acc.rs` and `recall.rs`.  All code is double checked by me and altered where I deemed it necessary.


crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+};
+use burn_core::{
+    prelude::{Backend, Tensor},
+    tensor::{ElementConversion, Float, TensorKind},
+};
+use core::marker::PhantomData;
+
+/// Input type for the [DiceMetric].
+///
+/// # Parameters
```
laggui: I think `Type Parameters` is more idiomatic when referring to generics

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+/// - `B`: Backend type.
+/// - `D`: Number of dimensions (default 2).
+/// - `K`: Tensor element type (default `Float`).
+#[derive(new)]
+pub struct DiceInput<B: Backend, const D: usize = 2, K = Float>
```
laggui: We can omit the tensor kind (`K`) for the `DiceInput`, especially since the metric is implemented for float.

but are binary masks typically provided as float? Perhaps it should be int for both outputs and targets?

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+impl Default for DiceMetricConfig {
+    fn default() -> Self {
+        Self {
+            epsilon: 1e-7,
+            include_background: false,
+        }
+    }
+}
+
+/// The Dice metric for evaluating overlap between two binary masks.
```
laggui: We can keep the `DiceMetric` name, but at least refer to `Dice coefficient` in the doc comment. Maybe link to a ref (wiki)?

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+///
+/// # Parameters
+/// - `B`: Backend type.
+/// - `D`: Number of dimensions (default 2).
```
laggui: I don't think we need to explicitly document these type parameters at this point 🙂 given that it follows from `Tensor<B, D>`

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+#[derive(Default)]
+pub struct DiceMetric<B: Backend, const D: usize = 2> {
```
laggui: I haven't used the dice coefficient as much, but is there a reason for the default `D = 2`? Wouldn't a typical usage have tensors `[B, C, H, W]` (so `D = 4`) for image segmentation?

Maybe we can omit the default rank.

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+            panic!(
+                "Outputs and targets must have the same dimensions. Got {:?} and {:?}",
+                outputs.dims(),
+                targets.dims()
+            );
+        }
+
+        let dims = outputs.dims();
+        let batch_size = dims[0];
+        let n_classes = dims[1];
```
laggui, line 130: Technically, the implementation is for any `D`, so even 1. But that will not work 😅 so if `DiceMetric` is generic over the rank, we should at least have a check for `D > 1`.

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,193 @@
...
+        if self.config.include_background && n_classes < 2 {
+            panic!("Dice metric requires at least 2 classes when including background.");
+        }
+
+        let intersection = (outputs.clone() * targets.clone()).sum();
```
laggui, line 144: This is only valid for two binary inputs, but it's only mentioned in the `DiceMetric` docs as "overlap between two binary masks". We should be more explicit, and maybe accept integer values for the multi-class targets which can be one-hot encoded?

crates/burn-train/src/metric/vision/dice.rs
MathijsdeBoer: You are completely correct. For some reason my mind misfired on defaulting on 2 _spatial_ dimensions, so `D` obviously has to be 2, right? I'll set it to default 4.

I include a dim parameter because I mainly work with 3D volumetric segmentations (medical scans), which means that ideally this metric would be functional for the much more common 2D, but also supports volumes, if need be.

crates/burn-train/src/metric/vision/dice.rs
MathijsdeBoer: Here is some Python sneaking into my code. Generally the targets would be 0.0's and 1.0's only in the classical Dice implementation. However, Soft Dice does use continuous values in the target. For example, if you have multiple labelers, who have slight disagreements with where the exact boundaries of an object are, you could encode this in your target as a probability instead. Perhaps we could split the soft dice case off into a separate metric, for user clarity?

One issue I run into (and that I've sort of left aside for this iteration of the PR) is that _generally_ your targets would be binary masks, one-hot or otherwise, but your predictions straight out of the model tend to be a floating point type, so in Rust we would have to check datatypes, and match them up.

Dice Loss implementations I see tend to use floating point data for both, to provide a smooth loss surface. So I just sort of matched on that a little. If it is preferable for the project, I can also see about enforcing int-type only values as input.

crates/burn-train/src/metric/vision/dice.rs
MathijsdeBoer: `D > 2`, at the very least. While I can 't come up with an example of this happening, perhaps it would be good to support a `[B, C, X]` tensor at the minimum.

MathijsdeBoer:
Hey there, thanks for the feedback!

I've quickly commented on some of your comments where I could form an answer, and will have a closer look at the rest when I have time.

crates/burn-train/src/metric/vision/dice.rs
laggui: > `D > 2`, at the very least

Yeah that's what I meant! In other words, maybe we can have a check in `DiceMetric::new()` to assert that `D > 2`.

crates/burn-train/src/metric/vision/dice.rs
laggui: > However, Soft Dice does use continuous values in the target. [...] Perhaps we could split the soft dice case off into a separate metric, for user clarity?

If the metric is computed in a similar fashion, it could simply be a matter of configuration. But we can leave that for the future 😄 

>  that generally your targets would be binary masks, one-hot or otherwise, but your predictions straight out of the model tend to be a floating point type, so in Rust we would have to check datatypes, and match them up.

Yeah the predictions are obviously float. But we define the `DiceInput`, so we can force the target type. Since we need to convert them to float anyway (e.g., `int_tensor.float()`), I agree that it makes sense to have the field as float. Since the fields are private, we could provide functions like `DiceInput::multiclass(float_preds, int_target)` to take care of one-hot + float conversion.

Not mandatory for now, mostly wanted to make sure of the typical usage since I'm not entirely familiar 🙂

laggui:
> Hey there, thanks for the feedback!
> 
> I've quickly commented on some of your comments where I could form an answer, and will have a closer look at the rest when I have time.

No rush 🙂 implementation looks good btw, comments were mainly directed at the API/usage and asking questions to better grasp the scope,

crates/burn-train/src/metric/vision/dice.rs
MathijsdeBoer: I've seen all sorts of conventions/assumptions across the various implementations of DSC that I've used over the years. I don't think there is a commonly accepted set of rules. As long as the convention that Burn uses is immediately obvious to the user, it should be perfectly fine.

I think I'll go ahead and make it int-only for now, pending anyone bringing this up in the future.

MathijsdeBoer:
I've tackled the points you've raised. It's pretty late here, so there is a reasonable chance that some new mistakes slipped in. All checks pass though!

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,347 @@
...
+            "Targets must have exactly {} dimensions. Got {:?}",
+            D,
+            targets.dims()
+        );
+        assert!(
+            outputs.shape() == targets.shape(),
+            "Outputs and targets must have the same shape. Got {:?} and {:?}",
+            outputs.shape(),
+            targets.shape()
+        );
```
laggui: We don't need to check `dims().len() == D`, that should always be true for a given tensor.

Also, `shape()` and `dims()` check are essentially the same thing, so we can keep only one of those.

crates/burn-train/src/metric/vision/dice.rs
```diff
@@ -0,0 +1,347 @@
...
+/// Input type for the [DiceMetric].
+///
+/// # Type Parameters
+/// - `B`: Backend type.
+/// - `D`: Number of dimensions. Should be more than, or equal to 3 (default 4).
+pub struct DiceInput<B: Backend, const D: usize = 4> {
+    /// Model outputs (predictions), as a tensor.
+    outputs: Tensor<B, D, Int>,
+    /// Ground truth targets, as a tensor.
+    targets: Tensor<B, D, Int>,
```
laggui, line 20: > I think I'll go ahead and make it int-only for now, pending anyone bringing this up in the future

Sure! We can broaden the support in the future.

laggui (commented):
Just a minor comment for the checks, otherwise looks good!

crates/burn-train/src/metric/vision/dice.rs
MathijsdeBoer: Ah, there's probably some more of my Python main background coming in. I'll simplify that section.

MathijsdeBoer: These have now been reduced to a single `assert`

PR #3415, opened 2025-07-23
Fix ndarray conv2d groups channels

laggui:
### Related Issues/PRs

From [discord](https://discord.com/channels/1038839012602941528/1045036181148139530/1397514093123866656):

```rust
use burn::module::Module;
use burn::nn::conv::{Conv2d, Conv2dConfig};
use burn::prelude::Backend;
use burn::tensor::Tensor;

#[derive(Module, Debug)]
pub struct Net<B: Backend> {
    c: Conv2d<B>,
}

impl<B: Backend> Net<B> {
    pub fn new(device: &B::Device) -> Self {
        Self {
            c: Conv2dConfig::new([2, 16], [3, 3])
                .with_stride([1, 1])
                .with_padding(burn::nn::PaddingConfig2d::Same)
                .with_groups(2)
                .init(device),
        }
    }

    pub fn forward(&self, x: Tensor<B, 4>) -> Tensor<B, 4> {
        let x = self.c.forward(x);
        x
    }
}

fn main() {
    type Backend = burn_ndarray::NdArray<f32>;

    let model = Net::<Backend>::new(&device);

    let input = Tensor::<Backend, 4>::ones([1, 2, 8, 8], &device);

    let output = model.forward(input);
}
```

```
ndarray: index [1, 1, 1, 0] is out of bounds for array of shape [1, 3, 3, 16]
ndarray: index [1, 1, 1, 8] is out of bounds for array of shape [1, 3, 3, 16]
```

### Changes

Fixed in channels in conv2d with groups

### Testing

Added unit test


wingertge:
Have you checked for performance regressions? That code is extremely sensitive to register pressure and even a single extra variable can cause spills.

If there's a meaningful regression I think it would be better to leave out groups from the SIMD implementation and dispatch the standard one instead. Groups are only used for older models and low power devices, which wouldn't benefit that much from SIMD regardless.

laggui:
Thanks for taking a look!

I actually haven't checked for performance regressions yet. But I just realized that we don't need to pass it as an argument. We can just replace the previous dim to use the correct tensor to get the input channels:

```diff
- let in_channels = x.shape()[0];
+ let in_channels = weights.shape()[0];
```

The weights are already passed as reference. No extra parameter, so shouldn't affect register allocation.

Issue #3344, opened 2025-07-03
Add functionality similiar to pytorch's torch.max/min for tensor shape [X] comparisions


catch-twenty-two:
### Feature description

Create a similar function in burn like in pytorch where if for example t1 has shape 4 and t2 has shape 4 tensor.max(t1,t2) will 'broadcast' and create a new tensor of shape 4,4 of all comparisons in t1 and t2.

e.g.:

`[3.0, 2.0, 6.0, 3.0]`
compared with:
`[[1.0], [0.5], [4.0], [7.0]]`

becomes:
```
[
    [3.0, 3.0, 4.0, 7.0],
    [2.0, 2.0, 4.0, 7.0],
    [6.0, 6.0, 6.0, 7.0],
    [3.0, 3.0, 4.0, 7.0],
]
```
### Feature motivation

Needed for an easy way to create an IOU algorithm for comparison of bounding boxes for object detection training.

### (Optional) Suggest a Solution

Example code for max_expand

[https://docs.pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics](https://docs.pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics)

laggui:
Btw, in torchland this would not be a single operation but a broadcast + max pair, e.g.

```py
x = torch.tensor([3., 2., 6., 3.])
y = torch.tensor([1., 0.5, 4., 7.])
torch.max(x.unsqueeze(1), y.unsqueeze(0))
```

but as you probably noticed we do not have such broadcasting support with `a.max_pair(b)`.

catch-twenty-two:
I actually ended up needing it for quite a few functions... So I ended up extending the functionality of Tensor and came up with this, not very flexible afa dimensions... But it's a starting point

```
use burn::{prelude::Backend, tensor::Tensor};

pub trait SSDTensorOps<B: Backend, const D: usize> {
    fn max_expand(&self, other: Tensor<B, 1>) -> Tensor<B, 2>;
    fn min_expand(&self, other: Tensor<B, 1>) -> Tensor<B, 2>;
    fn add_expand(&self, other: Tensor<B, 1>) -> Tensor<B, 2>;
    fn sub_expand(&self, other: Tensor<B, 1>) -> Tensor<B, 2>;
}

impl<B: Backend> SSDTensorOps<B, 1> for Tensor<B, 1> {
[...]

PR #3417, opened 2025-07-24
Adding option for broadcasting tensors similar to torch WIP

catch-twenty-two:
This adds functionality to broadcast two tensors with potentially different static ranks to a common rank and then preform an operation on them. Before adding more operations via macros, I wanted to make sure this would be helpful to the community since as of now I use it the following way:

```
let a = Tensor::<B, 6>::empty([7, 6, 2, 3, 1, 9], device);
let b = Tensor::<B, 4>::empty([2, 1, 7, 1], device);

let (a, b) = broadcast!(a: Tensor<B, 6>, b: Tensor<B, 4>);

let a_add_b = a.add(b);
```
More functionality can be easily added to use it as so as well:

```
let a = Tensor::<B, 6>::empty([7, 6, 2, 3, 1, 9], device);
let b = Tensor::<B, 4>::empty([2, 1, 7, 1], device);

let a = add_broadcast!(a: Tensor<B, 6>, b: Tensor<B, 4>);
```
Note: Macros were incorporated since the broadcast operation requires knowledge of what size the rank from the resulting tensors will be (which are constants) prior to compile time.

### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/3344
https://github.com/tracel-ai/burn/issues/1499

### Changes

I've added the logic to correctly shape the tensors via a broadcast function and have operations done on them.

### Testing

Multiple unit tests, and integrated into my own code base.


laggui (commented):
I think having broadcasting for tensors of different rank would address a real gap in usability 🙂

But we should be a thoughtful about the right approach (and I haven't really thought about it yet).

Some general comments on the current macro:
- it only applies left padding to both operands, but broadcasting can be a bit more complex
- the syntax is a little verbose with the type annotations vs type inference
- having a per-op macro like `add_broadcast!` is helpful to apply different broadcasting rules, but can lead to macro explosion; we can probably have an element-wise broadcast and matmul broadcast to cover almost all cases

catch-twenty-two:
Thanks for the review! 

> I think having broadcasting for tensors of different rank would address a real gap in usability 🙂
> 

So, that kinda sounds like a yay, not a nay on usefulness? 😁

> But we should be a thoughtful about the right approach (and I haven't really thought about it yet).
> 
> Some general comments on the current macro:
> 
>  -  it only applies left padding to both operands, but broadcasting can be a bit more complex
> 
Could you give me an quick example? I thought I had all the bases covered here afa the general way pytorch handles broadcasting. If there are some corner cases I'd love to account for them.

https://docs.pytorch.org/docs/stable/notes/broadcasting.html

>    - the syntax is a little verbose with the type annotations vs type inference

Agreed! I couldn't think of a better way to pass the generic constants to the actual function (since as far as I know, this is the only way to do this?) to compare which one is larger (since this has to be done at compile time) and still have it look intuitively like the rest of the burn/rust tensor type syntax 
```
let a: Tensor<B, D, K> = ect..
```

Any ideas would be greatly appreciated, as I know once this syntax is incorporated to the api, you are pretty much stuck with it.

I've since, updated it a bit, but it's still verbose:
```
broadcast!(
    a: Tensor<Backend, RANK_A>,
    b: Tensor<RANK_B>
)
```
>  - having a per-op macro like `add_broadcast!` is helpful to apply different broadcasting rules, but can lead to macro explosion; we can probably have an element-wise broadcast and matmul broadcast to cover almost all cases

Agreed, this is why I only did it for one... I just wanted to show an example of what could be done... Personally I really dislike macros, I feel that they obfuscate the code and make debugging difficult, the less the better! I'll look into incorporating matmul.

laggui:
Sorry for the late response, notification got lost in the recent flood 😅 

> So, that kinda sounds like a yay, not a nay on usefulness? 😁

Yeah I definitely see the usefulness! Just gotta nail the usage/implementation 🙂 

> Could you give me an quick example? I thought I had all the bases covered here afa the general way pytorch handles broadcasting. If there are some corner cases I'd love to account for them.

Actually this kinda ties into the third point regarding elemwise and matmul broadcast rules. For elemwise I believe it's simply left padding, it's matmul that has different rules.

Not entirely sure yet about the best approach, I'd have to think about the best way to tackle the syntax and rules 🤔

antimora:
@catch-twenty-two is it ready for a review?

catch-twenty-two:
> @catch-twenty-two is it ready for a review? 

Hey @antimora,

It is a start and I think it works well,  I have used it all through-out my projects. After talking with @laggui I tried implementing matmul but ran into issues with the requirement of knowing how many dimensions the tensors have prior to doing specific matmul operations on them. (It sounds like this is an issue that has been discussed before at length) I couldn't have done a lot of what I needed for training/image augmentation with out it so it may be better than nothing ATM, but that's up to you all.

catch-twenty-two:
@antimora If you do decide to use it, I still need to integrate the unit tests into the overall code base, so it's not ready to merge yet. But is ready for a review. Thanks!

nathanielsimard:
I think we will support the same notation with dynamic rank instead of macros

laggui (commented):
Still unsure if the macro is the right approach.

We could possibly provide something like `tensor.broadcast_with(other)`.

catch-twenty-two:
> I think we will support the same notation with dynamic rank instead of macros

@nathanielsimard This is the best case scenario! But until that happens this was the best alternative to get me unblocked.

If you have some unlreleased code you wanted to share that uses dynamic rank, I could work that into this instead of using macros...

> Still unsure if the macro is the right approach.
> We could possibly provide something like tensor.broadcast_with(other).

@laggui - I couldn't do this without macros since there isn't a concept of dynamic rank yet. Did you have a different idea that I'm not seeing?

antimora:
@catch-twenty-two 

> @laggui - I couldn't do this without macros since there isn't a concept of dynamic rank yet. Did you have a different idea that I'm not seeing?

There is a design document for dyn ranks: https://github.com/tracel-ai/burn/issues/3624 still not implemented. 

I am closing this PR since we've decided not to purse macro direction.

Issue #2649, opened 2025-01-02
CER/WER metrics


laggui:
### Discussed in https://github.com/tracel-ai/burn/discussions/2643

<div type='discussions-op-text'>

<sup>Originally posted by **pmalex** December 26, 2024</sup>
Is it possible to use these metrics in burn? I'm not found any examples so far.</div>

me-diru:
Hi @laggui ! I am new to Burn and thought working on an issue would be a nice way to learn how it works. Do you know if this issue is still relevant? thanks

laggui:
Yep, the metrics have not been implemented yet 🙂

I am not sure how widely these are used, but you can also have a look at the [current metrics in `burn-train`](https://github.com/tracel-ai/burn/tree/main/crates/burn-train/src/metric) and check if any useful/common metrics are missing.

yazanmashal03:
take

PR #3418, opened 2025-07-24
Adding CER/WER metrics

yazanmashal03:
### Checklist

### Related Issues/PRs

The PR is about Issue #2649, and implements two new metrics for sequence evaluation in NLP and related fields. These are the character error rate (CER) and word error rate(WER) metrics. These are (the same) error metrics used on the character and word levels, respectively.

### Changes

The cer.rs file provides an implementation of the Character Error Rate metric. CER measures (here we use Levenshtein using DP) the percentage of characters that are incorrect (insertions, deletions, substitutions) in the predicted sequence compared to the reference sequence.

The wer.rs file implements the Word Error Rate metric. WER is similar to CER but operates at the word level, measuring the percentage of words that are incorrect in the predicted sequence.

### Testing

Testing was done using unit tests, where each function (for example test_wer_without_padding, test_wer_with_padding) is a unit test for the WerMetric implementation. Same for CER.


crates/burn-train/src/metric/cer.rs
```diff
@@ -0,0 +1,210 @@
...
+            curr[j + 1] = if ca == cb {
+                prev[j] // match
+            } else {
+                1 + prev[j].min(prev[j + 1]).min(curr[j]) // subst/ins/del
+            };
+        }
+        core::mem::swap(&mut prev, &mut curr);
+    }
+    prev[b.len()]
+}
```
laggui: If using the same function for word error rate, we should only define it once and re-use it.

Also, I think the function would benefit from a bit more documentation and/or better variable names 😅 what argument should be the reference? and the predictions? It should be a bit more clear at first glance.

crates/burn-train/src/metric/cer.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// The character error rate metric.
+#[derive(Default)]
+pub struct CerMetric<B: Backend> {
```
laggui: Maybe `CharErrorRate` is a more descriptive name.

Also, can you add a bit more documentation to give details on the metric?

crates/burn-train/src/metric/cer.rs
```diff
@@ -0,0 +1,210 @@
...
+        let [batch_size, seq_len] = targets.dims();
+
+        let outputs_data = outputs
+            .to_data()
+            .to_vec::<i64>()
+            .expect("Failed to convert outputs to Vec");
+        let targets_data = targets
+            .to_data()
+            .to_vec::<i64>()
+            .expect("Failed to convert targets to Vec");
```
laggui: We should perform the operations on tensors as much as possible, and then update the state using the computed values.

crates/burn-train/src/metric/cer.rs
```diff
@@ -0,0 +1,210 @@
...
+            100.0 * total_edit_distance / total_target_length
+        } else {
+            0.0
+        };
+
+        self.state.update(
+            value,
+            batch_size,
+            FormatOptions::new(self.name()).unit("%").precision(2),
+        )
```
laggui, line 140: I think the state might need to keep track of the errors and total characters (or words for WER)? Otherwise aggregation might be incorrect 🤔 this would require a new state type though

crates/burn-train/src/metric/wer.rs
```diff
@@ -0,0 +1,215 @@
...
+            };
+        }
+        core::mem::swap(&mut prev, &mut curr);
+    }
+    prev[b.len()]
+}
+
+/// The word error rate metric.
+#[derive(Default)]
+pub struct WerMetric<B: Backend> {
```
laggui: Same comment regarding name (e.g. `WordErrorRate`), documentation and implementation using tensors.

laggui (changes requested):
Thanks for contributing these additional metrics!

I only have a couple of comments regarding the implementation.

crates/burn-train/src/metric/cer.rs
yazanmashal03: Yes, thank you for the input. I re-used the function, and added the respective documentation.

crates/burn-train/src/metric/cer.rs
yazanmashal03: Added relevant documentation

crates/burn-train/src/metric/cer.rs
yazanmashal03: I implemented the operations on tensors as much as possible.

crates/burn-train/src/metric/cer.rs
yazanmashal03: Hmm, that sounds correct. However, the value here includes the errors relative to the total characters, since value=total_edit_distance/total_characters * 100, so why would we need to keep the total characters?

crates/burn-train/src/metric/wer.rs
yazanmashal03: Yes sir

crates/burn-train/src/metric/cer.rs
yazanmashal03: @laggui

crates/burn-train/src/metric/cer.rs
laggui: For the current batch that's accurate, but when aggregated for an epoch it might be incorrect since this is a numeric state (not all batches have the same composition). Probably out of scope for this PR so no worries 👍

crates/burn-train/src/metric/cer.rs
```diff
@@ -0,0 +1,214 @@
...
+use burn_core::tensor::{Int, Tensor};
+use core::marker::PhantomData;
+
+/// Computes the edit distance (Levenshtein distance) between two sequences of integers.
+///
+/// The edit distance is defined as the minimum number of single-element edits (insertions,
+/// deletions, or substitutions) required to change one sequence into the other. This
+/// implementation is optimized for space, using only two rows of the dynamic programming table.
+/// 
+pub fn edit_distance(reference: &[i32], prediction: &[i32]) -> usize {
```
laggui: We can mark it as `pub(crate)` only

laggui (commented):
Sorry for the late follow-up! Didn't see this was updated. Please explicitly re-request a review so I know when changes have been applied 🙂


LGTM, just minor formatting issues to fix.

antimora:
@yazanmashal03 , if you'd like, I can take over and complete this.

yazanmashal03:
> @yazanmashal03 , if you'd like, I can take over and complete this.

Yes please, I cannot attend to this issue right now. Thanks!

laggui (approved):

Issue #1811, opened 2024-05-25
Error when importing onnx of transformers bert model 


7kanak:
I was following the onnx import example in burn  `examples/onnx-inference` . but instead of using mnist.py to generate the onnx model, i am using transformers python package to generate onnx model

```
model_id = "albert/albert-base-v2"
feature = "sequence-classification"

model = AutoModelForSequenceClassification.from_pretrained(model_id)
tokenizer = AutoTokenizer.from_pretrained(model_id)

inputs = tokenizer("onnx is wonderful", return_tensors="pt")
torch.onnx.export(
    model, 
    tuple(inputs.values()),
    f="albert.onnx",  
    input_names=['input_ids', 'attention_mask'], 
    output_names=['logits'], 
    dynamic_axes={'input_ids': {0: 'batch_size', 1: 'sequence'}, 
                  'attention_mask': {0: 'batch_size', 1: 'sequence'}, 
                  'logits': {0: 'batch_size', 1: 'sequence'}}, 
    do_constant_folding=True, 
    opset_version=16, 
    verbose=True
)
```

getting this error
```
  checking outputs    
  DEBUG burn_import::onnx::from_onnx: it's a constant    
  DEBUG burn_import::onnx::proto_conversion: Converting ONNX node with type "Unsqueeze"    
  DEBUG burn_import::onnx::from_onnx: renaming node "/albert/embeddings/Unsqueeze"    
  DEBUG burn_import::onnx::from_onnx: checking node unsqueeze3 for constants    
  DEBUG burn_import::onnx::from_onnx: checking input Argument { name: "/albert/embeddings/Constant_4_output_0", ty: Tensor(TensorType { elem_type: Int64, dim: 1, shape: Some([1]) }), value: None, passed: false } for const    
  DEBUG burn_import::onnx::from_onnx: input /albert/embeddings/Constant_4_output_0 matched constant node constant10    
  ERROR burn_import::logger: PANIC => panicked at /home/kanak/.cargo/registry/src/index.crates.io-6f17d22bba15001f/burn-import-0.13.2/src/onnx/dim_inference.rs:281:40:
  called `Option::unwrap()` on a `None` value    

  --- stderr
  thread 'main' panicked at /home/kanak/.cargo/registry/src/index.crates.io-6f17d22bba15001f/burn-import-0.13.2/src/onnx/dim_inference.rs:281:40:
  called `Option::unwrap()` on a `None` value
```

[...]

PR #3423, opened 2025-07-26
Add ONNX Attention op

Friedrich-S:
### Checklist

### Related Issues/PRs

#1714 

### Changes

Implements the ONNX `Attention` operator.

### Testing

The operator is fully tested regarding both codegen and functionality.


Friedrich-S:
Looks like I might need to replace my usage of let_chains to make the code work in Rust 1.85. Any plans for raising the MSRV to 1.88? :smile:

antimora:
> Looks like I might need to replace my usage of let_chains to make the code work in Rust 1.85. Any plans for raising the MSRV to 1.88? 😄

I think we should do it. Our goal has been to support current + previous version. Feel free to submit a PR to bump MSRV. 

CCing @laggui @nathanielsimard @syl20bnr

Friedrich-S:
I was playing around with albert from #1811. With the original opset version of 16, I was getting similar issues to those reported in the issue (plus some other strange ones that make me suspect some errors in the pytorch export; might add a comment). I then tried out opset 23 with dynamo export, which now uses the new Attention node as per https://github.com/pytorch/pytorch/issues/149662. The output from this does not contain any static shapes for the input at all. The current implementation does very strict shape checking that matches the OP spec and reference implementation, but I am wondering if I should simply remove all the static shape checking and rely purely on runtime shape checking to make it work with these dynamic scenarios.

Surprisingly, Netron does show static shapes for this case, so maybe they are getting lost somewhere in the parsing pipeline?
image

For reference, this is the data the code is receiving:
 ```
 Node {
      node_type: Attention,
      name: "attention1",
      inputs: [
          Argument {
              name: "transpose2_out1",
              ty: Tensor(
                  TensorType {
                      elem_type: Float32,
                      rank: 4,
                      static_shape: None,
                  },
              ),
              value: None,
              passed: false,
          },
          Argument {
              name: "transpose4_out1",
              ty: Tensor(
                  TensorType {
                      elem_type: Float32,
                      rank: 4,
                      static_shape: None,
                  },
              ),
              value: None,
              passed: false,
          },
          Argument {
              name: "transpose6_out1",
              ty: Tensor(
                  TensorType {
                      elem_type: Float32,
                      rank: 4,
                      static_shape: None,
                  },
              ),
              value: None,
              passed: false,
          },
          Argument {
              name: "where1_out1",
              ty: Tensor(
                  TensorType {
                      elem_type: Float32,
                      rank: 4,
                      static_shape: None,
                  },
              ),
              value: None,
              passed: false,
          },
      ],
      outputs: [
          Argument {
              name: "scaled_dot_product_attention",
              ty: Tensor(
                  TensorType {
                      elem_type: Float32,
                      rank: 0,
                      static_shape: None,
                  },
              ),
              value: None,
              passed: false,
          },
      ],
      attrs: {
          "is_causal": Int64(
              0,
          ),
          "softcap": Float32(
              0.0,
          ),
          "qk_matmul_output_mode": Int64(
              0,
          ),
      },
  }
 ```

Friedrich-S:
Cool, thanks for the detailed feedback. Looks like I will change the operator then to completely disregard the static shapes and only work with ranks (I don't think it makes sense to support both, given how complex this code is). This will thankfully also make the code quite a bit simpler.

laggui:
> (I don't think it makes sense to support both, given how complex this code is). This will thankfully also make the code quite a bit simpler.

Agreed! And I think it should work given the rank only, though I have to admit I haven't completely parsed all the different configurations.

Also, not sure if you stumbled upon it but we do have a `MultiHeadAttention` module in `burn::nn` (num heads query = num heads key value), though it's not flexible enough to cover all the configurations of the ONNX `Attention` spec.

Friedrich-S:
Yeah, I saw. Unfortunately, the ONNX Attention OP is just too powerful and supports every common variant of Attention. However, I think the implementation is relatively straightforward and the codegen is not overly complex.

Friedrich-S:
Done. Now it is completely disregarding the static shapes and does slightly more things at runtime (assembling the data for the reshape operations if necessary).
Btw, what other areas of `burn` are currently in (the greatest) need of contributions? I randomly picked the ONNX operators because I saw the parent ticket in the "Help Wanted" section in GitHub. I have some background in hardcore performance optimization, and I assume contributions in that regard would be welcome as well. Pointers to relevant tickets or discussions are welcome.

antimora:
> Done. Now it is completely disregarding the static shapes and does slightly more things at runtime (assembling the data for the reshape operations if necessary).
> 
> Btw, what other areas of `burn` are currently in (the greatest) need of contributions? I randomly picked the ONNX operators because I saw the parent ticket in the "Help Wanted" section in GitHub. I have some background in hardcore performance optimization, and I assume contributions in that regard would be welcome as well. Pointers to relevant tickets or discussions are welcome.

Thanks so much for your contribution!

I have one long wish for Burn to support 0 dim tensors (you can look up an issue, or I can later). It gives some background and why it would help with performance (hint not to copy data). Currently we use 1d with one element.

Check with the team if you decide to work on this. Not sure the current plan.

antimora:
> 0 dim tensors

https://github.com/tracel-ai/burn/issues/1689

> > Done. Now it is completely disregarding the static shapes and does slightly more things at runtime (assembling the data for the reshape operations if necessary).
> > Btw, what other areas of `burn` are currently in (the greatest) need of contributions? I randomly picked the ONNX operators because I saw the parent ticket in the "Help Wanted" section in GitHub. I have some background in hardcore performance optimization, and I assume contributions in that regard would be welcome as well. Pointers to relevant tickets or discussions are welcome.
> 
> Thanks so much for your contribution!
> 
> I have one long wish for Burn to support 0 dim tensors (you can look up an issue, or I can later). It gives some background and why it would help with performance (hint not to copy data). Currently we use 1d with one element.
> 
> Check with the team if you decide to work on this. Not sure the current plan.

This is the issue I was telling you about https://github.com/tracel-ai/burn/issues/1689. The scope excludes ONNX related changes. It can come later.

laggui:
> Btw, what other areas of `burn` are currently in (the greatest) need of contributions? I randomly picked the ONNX operators because I saw the parent ticket in the "Help Wanted" section in GitHub. I have some background in hardcore performance optimization, and I assume contributions in that regard would be welcome as well. Pointers to relevant tickets or discussions are welcome.

There are many ways and levels at which to contribute. We've been wanting to add a scan / prefix sum for a while but never had the time to add it to CubeCL (#2664). This is really more work in CubeCL than Burn, but would benefit a lot of users.

If you have an application in mind that could be a good driving factor for contributions. Lacking / unclear docs, missing feature or an actionable performance improvement.

We only added ONNX to the "Help Wanted" section (pinned) because it is easier for contributors to start out, and eventually they might find ways that benefit their own projects once they get more involved with Burn.

Friedrich-S:
Thanks for all the feedback on open projects. I will check out the 0-dim tensor stuff. I have written a GPU implementation (Vulkan) for the decoupled look-back parallel prefix sum for a different project, so I think I could potentially help here. Having such a tool available would be massively helpful, as I have found the lack of a fast prefix sum operation to be a severe bottleneck in my other projects.

I have also always wished for an indexed matmul operation, but don't have an entire picture of potential technical reasons for not having such a feature. Instead of doing a matmul on the entire tensor, index into a subset (of the batch) first and then do the matmul, but as a fused operator without any intermediate allocations. A potential use-case could be mixture-of-experts or other sparsely activated model architectures. I would be happy to create a ticket with a more detailed explanation if you don't foresee any obvious problems.

PR #3424, opened 2025-07-27
Bump MSRV from 1.85 to latest stable 1.87

Friedrich-S:
### Checklist

### Related Issues/PRs

Discussed briefly in the comments on #3423.

### Changes

Updates the MSRV from the 1.85 to the latest stable version 1.88.

### Testing

`cargo run-checks` works fine and since Rust is backwards compatible within the same edition, there should be no other major problems.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ env:
   # Note: It is not possible to define top level env vars and pass them to composite actions.
   # To work around this issue we use inputs and define all the env vars here.
 
-  RUST_PREVIOUS_VERSION: 1.88.0
+  RUST_PREVIOUS_VERSION: 1.87.0
 
   # Cargo
   CARGO_TERM_COLOR: "always"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ env:
   # Note: It is not possible to define top level env vars and pass them to composite actions.
   # To work around this issue we use inputs and define all the env vars here.
 
-  RUST_PREVIOUS_VERSION: 1.87.0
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ env:
   # Note: It is not possible to define top level env vars and pass them to composite actions.
   # To work around this issue we use inputs and define all the env vars here.
 
-  RUST_PREVIOUS_VERSION: 1.87.0
+  RUST_PREVIOUS_VERSION: 1.88.0
```
antimora: I am not sure if this should be previous version or not. But according to a related PR #2388, this value is the same as rust-version. I am guessing this should be 1.87.0

@syl20bnr @laggui can confirm this.

laggui: The `rust-version` is for the crate, and `RUST_PREVIOUS_VERSION` is used for the CI. This should stay at 1.87 until 1.89 is stable.

antimora:
> We can update the `rust-version` for the crate MSRV to 1.87. But the CI should stay on the previous stable version (1.87).

Do you mean 1.88 for rust-version?

laggui:
No 1.87 was correct 🙂

`rust-version` is the minimal supported Rust version a crate. 1.89 is not stable just yet, so if we set it to 1.88 we only support one version (no "previous" stable version).

Right now we have `rust_version: 1.85` which is a bit out of sync with our previous version CI.

Quite simply, I would keep `RUST_PREVIOUS_VERSION: 1.87.0` and bump `rust_version: 1.87` for now.

Friedrich-S:
Makes sense. I switched both to 1.87

PR #3432, opened 2025-07-29
[WIP] Proposal to establish `burn::baselib::...` collecting pre-tensor common libs; starting with indexing.

crutcher:
This is incomplete; it is a proposal to establish a `burn::baselib::...` tree; which exists pre-core and pre-tensor; but is also public for third-parties to use.

the first candidate is lifting the indexing, shape, range, and slice utils.

i've not resolved how to re-export things to maintain backwards compat; this is a rough proof of concept.


crutcher:
@laggui based upon discussions in `development` > `#general`

antimora:
We might want to review other common shape operations implemented in Candle for potential reuse or inspiration: [candle-core/src/shape.rs](https://github.com/huggingface/candle/blob/main/candle-core/src/shape.rs)

Candle’s `shape.rs` includes utilities like:

* Shape validation and broadcasting helpers.
* Indexing/slicing logic with bounds checks.
* Contiguous shape transformations (e.g., `reshape`, `permute`, `expand`).
* Stride and contiguous memory layout calculations.
* Compact methods for computing element count and dimensionality.

crutcher:
I have a bunch of indexing / traversal iterator util stuff from a java tensor impl I built; and some of it might be worth stealing:
https://github.com/crutcher/tapestry/tree/main/src/main/java/org/tensortapestry/zspace/indexing

Basically, we often need these things not just to write correct third-party code that has the same generic flexibility as the API; but to write implementations/bindings of array-like objects into the tensor implementation layers.

antimora:
@crutcher, still interested keeping it open?

antimora:
Closing this for now, since there are no activity; feel free to reopen if needed.

PR #3444, opened 2025-07-30
ADD GLU activation function

bn-c:
### Checklist

### Related Issues/PRs

Saw an activation function I want to use, but haven't been implemented yet

### Changes

Added the Gated Linear Unit activation function from [`torch.nn.GLU`](https://docs.pytorch.org/docs/stable/generated/torch.nn.GLU.html#torch.nn.GLU): 
- GLU module. 
- glu activation functon.

### Testing

~~Required feedback on how to organize and implement tests~~
Added test with random data.

### DIFF
--- a/crates/burn-core/src/nn/glu.rs
+++ b/crates/burn-core/src/nn/glu.rs
@@ -25,14 +25,13 @@ impl GLU {
     /// GLU(a,b)=a⊗σ(b) where `a` is the first half of the input matrices and `b` is the second half.
     ///
     /// **Note**:
-    /// * The input tensor along `dim` must have an even size along the specified dimension. N is divisible by 2.
-    /// * Negative indices for `dim` are not supported (unlike PyTorch's nn.GLU).
+    /// * The size of the input tensor along `dim` must be divisible by 2.
     ///
     /// ### Arguments
-    /// * `tensor` - The input tensor. With shape `[∗1,N,∗2]` where `*` means, any number of additional dimensions
+    /// * `tensor` - The input tensor.
     ///
     /// ### Returns
-    /// * Output tensor with shape `[∗1​,M,∗2​]` where M=N/2
+    /// * A tensor with the same shape as the input, except the size along `dim` is halved.
     pub fn forward<B: Backend, const D: usize>(&self, input: Tensor<B, D>) -> Tensor<B, D> {
         crate::tensor::activation::glu(input, self.dim)
     }
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -265,14 +265,13 @@ pub fn tanh<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
 /// GLU(a,b)=a⊗σ(b) where `a` is the first half of the input matrices and `b` is the second half.
 ///
 /// **Note**:
-/// * The input tensor along `dim` must have an even size along the specified dimension. N is divisible by 2.
-/// * Negative indices for `dim` are not supported (unlike PyTorch's nn.GLU).
+/// * The size of the input tensor along `dim` must be divisible by 2.
 ///
 /// ### Arguments
-/// * `tensor` - The input tensor. With shape `[∗1,N,∗2]` where `*` means, any number of additional dimensions
+/// * `tensor` - The input tensor.
 ///
 /// ### Returns
-/// * Output tensor with shape `[∗1​,M,∗2​]` where M=N/2
+/// * A tensor with the same shape as the input, except the size along `dim` is halved.
 pub fn glu<const D: usize, B: Backend>(tensor: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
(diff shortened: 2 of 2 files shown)


crates/burn-core/src/nn/glu.rs
```diff
@@ -0,0 +1,49 @@
...
+        Self { dim }
+    }
+
+    /// Applies the gated linear unit function.
+    ///
+    /// GLU(a,b)=a⊗σ(b) where `a` is the first half of the input matrices and `b` is the second half.
+    ///
+    /// # Arguments
+    ///
+    /// * `tensor` - The input tensor. With shape [∗1,N,∗2] where * means, any number of additional dimensions
```
laggui: I think we can skip the shape description, the only relevant parts are that
1) The input tensor size along `dim` must be divisible by 2.
2) The output tensor has the same shape as the input, except the size along `dim` is halved.

Feels more clear imo.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -259,3 +259,26 @@ pub fn mish<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
...
+/// $$
+/// where $a$ is the first half of the input matrices and $b$ is the second half.
+///
+/// # Arguments
+///
+/// * `tensor` - The input tensor. With shape [∗1,N,∗2] where * means, any number of additional dimensions
+/// * `dim` - The dimension on which to split the input.
+///
+/// # Returns
+/// Output tensor with shape [∗1​,M,∗2​] where M=N/2
```
laggui: Same

laggui (changes requested):
Thanks for adding the activation 🙂 

Implementation looks good! But it's missing tests.

See the tests in [`burn-tensor/src/tests/activation`](https://github.com/tracel-ai/burn/tree/main/crates/burn-tensor/src/tests/activation) for reference. Should add one for reference with `glu`. You can use pytorch as a reference to compute the expected output for a given input.

Some notes (not actionable, just for reference):
For other ops we might add autodiff tests, but again here this is a simple composition of existing ops. So not required.

Depending on the activation we might add it to the [`ActivationOps`](https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/ops/activation.rs), but in this case it's a simple activation composed of other ops. So I don't think it's necessary (just like other activations).

crates/burn-tensor/src/tensor/activation/base.rs
bn-c: fixed

crates/burn-core/src/nn/glu.rs
bn-c: fixed

crates/burn-core/src/nn/glu.rs
```diff
@@ -0,0 +1,51 @@
...
+    ///
+    /// **Note**:
+    /// * The input tensor along `dim` must have an even size along the specified dimension. N is divisible by 2.
+    /// * Negative indices for `dim` are not supported (unlike PyTorch's nn.GLU).
+    ///
+    /// ### Arguments
+    /// * `tensor` - The input tensor. With shape `[∗1,N,∗2]` where `*` means, any number of additional dimensions
+    ///
+    /// ### Returns
+    /// * Output tensor with shape `[∗1​,M,∗2​]` where M=N/2
```
laggui: This is what I meant by skipping shape description, which was a little non-standard and ambiguous.

```suggestion
    /// **Note**:
    /// * The size of the input tensor along `dim` must be divisible by 2.
    ///
    /// ### Arguments
    /// * `tensor` - The input tensor.
    ///
    /// ### Returns
    /// * A tensor with the same shape as the input, except the size along `dim` is halved.
```

I also removed the note on negative indexing as I feel this is pretty clear with the use of `dim: usize`.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -259,3 +259,32 @@ pub fn mish<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
...
+///
+/// **Note**:
+/// * The input tensor along `dim` must have an even size along the specified dimension. N is divisible by 2.
+/// * Negative indices for `dim` are not supported (unlike PyTorch's nn.GLU).
+///
+/// ### Arguments
+/// * `tensor` - The input tensor. With shape `[∗1,N,∗2]` where `*` means, any number of additional dimensions
+///
+/// ### Returns
+/// * Output tensor with shape `[∗1​,M,∗2​]` where M=N/2
```
laggui: Reflected changes from above

```suggestion
/// **Note**:
/// * The size of the input tensor along `dim` must be divisible by 2.
///
/// ### Arguments
/// * `tensor` - The input tensor.
///
/// ### Returns
/// * A tensor with the same shape as the input, except the size along `dim` is halved.
```

laggui (approved):
Thanks for adding a numerical test 🙂 

I think the shape documentation can be clearer if we simply use natural language to describe that.

Approving in advance

Issue #3316, opened 2025-06-24
Add a config entry to dropping incomplete batch for DataLoader


sammyne:
### Feature description

Just wants some config entry like `drop_last` in pytorch's `DataLoader` as
```py
DataLoader(dataset, batch_size=1, shuffle=False, sampler=None,
           batch_sampler=None, num_workers=0, collate_fn=None,
           pin_memory=False, drop_last=False, timeout=0,
           worker_init_fn=None, *, prefetch_factor=2,
           persistent_workers=False)
```

### Feature motivation

To achieve development expirence similar to pytorch.

hollermay:
Can i work on this?

laggui:
Yeah this is still open. I think it should be fairly straightforward, the current implementation always forces the last batch to be returned (even if smaller than the configured batch size).

Note that when the `num_workers > 1`, the current behavior first splits the dataset into partial dataset for each worker, which are responsible to fetch a batch. So there could be more than one batch that is not complete (see #2373).

hollermay:
On it.

hollermay:
@laggui kindly review my PR

PR #3448, opened 2025-07-30
Add drop_last option to DataLoader batching

hollermay:
Introduces a drop_last flag to DataLoaderBuilder and FixBatchStrategy, allowing incomplete batches to be optionally dropped during data loading. Updates the builder API and batch strategy logic to support this feature, improving flexibility for batch processing.

### Checklist

### Related Issues/PRs

Fixes issue: DataLoader yields as many iterations as num_workers instead of correct batch count; no drop_last support (see #3316)

### Changes

**Problem:**  
The DataLoader previously yielded one batch per worker per epoch, regardless of batch size or dataset size, leading to incorrect iteration counts. There was also no way to drop incomplete batches, unlike PyTorch’s DataLoader.

**Solution:**  
- Introduced a `drop_last` flag to `DataLoaderBuilder` and `FixBatchStrategy`.
- Updated the builder API to allow users to set `drop_last`.
- Modified the batch strategy logic so that, when `drop_last` is true, incomplete batches are dropped.
- Refactored the multithreaded DataLoader to use a shared batch queue among workers, ensuring the number of iterations per epoch is determined by dataset size and batch size, not by `num_workers`.
- This brings the DataLoader’s behavior in line with PyTorch and improves flexibility for batch processing.

### Testing

- Ran `cargo test --workspace --all-features` to ensure all tests pass.
- Manually verified that:
  - The number of iterations per epoch matches `ceil(dataset_size / batch_size)` for various `num_workers` values.
  - The `drop_last` flag correctly drops incomplete batches when enabled.
  - Changing `num_workers` only affects parallelism, not batch count.


hollermay:
@laggui kindly review

laggui:
Will take a look today!

hollermay:
Any updates?

crates/burn-core/src/data/dataloader/builder.rs
```diff
@@ -14,6 +14,7 @@ pub struct DataLoaderBuilder<B: Backend, I, O> {
     num_threads: Option<usize>,
     shuffle: Option<u64>,
     device: Option<B::Device>,
+    drop_last: bool,
```
laggui: Maybe `drop_incomplete` better describes the behavior of dropping incomplete batches (< `batch_size`) 🤔

crates/burn-core/src/data/dataloader/multithread.rs
```diff
@@ -82,47 +68,10 @@ where
...
-                if let Some(rng) = self.rng.as_ref() {
-                    // Pre-shuffle the dataset before split if shuffle is enabled.
-                    // This ensures that each thread gets a uniform random sample of the dataset.
-                    let mut rng = rng.clone();
-                    dataset = Arc::new(burn_dataset::transform::ShuffledDataset::new(
-                        dataset, &mut rng,
-                    ));
-                }
-
-                let datasets = PartialDataset::split(dataset, self.num_threads);
```
laggui: It would be much easier to simply have a new `PartialDataset::split_batch_aware(dataset, num, batch_size)` instead to distribute complete batches across multiple partial datasets. The current issue really lies in the distribution of items across the splits.

The rest of the logic/code should just work as is.

crates/burn-core/src/data/dataloader/strategy.rs
```diff
@@ -54,23 +64,22 @@ impl<I: Send + 'static> BatchStrategy<I> for FixBatchStrategy<I> {
...
-
     fn batch(&mut self, force: bool) -> Option<Vec<I>> {
         if self.items.len() < self.batch_size && !force {
             return None;
         }
-
+        if self.items.len() < self.batch_size && force && self.drop_last {
+            self.items.clear();
+            return None;
+        }
```
laggui: We don't need to add `drop_last` to the batch strategy. The `force` flag already implements the logic.

Instead, we can check if we want to force the remaining items into an incomplete batch when `drop_last` (or `drop_incomplete` if we rename it) is set for the batch dataloader. Currently, we always force:

https://github.com/tracel-ai/burn/blob/689aa77ce8aedf80f56f1bd57ec3ce15bf9fb261/crates/burn-core/src/data/dataloader/batch.rs#L167-L182

crates/burn-core/src/data/dataloader/multithread.rs
laggui: Just made this addition in #3476, which now handles even batch splits in the multi-threaded scenario. The scope was pretty clear after reviewing your PR and adding this comment.

Your PR (and other comments) still stands for handling the incomplete batches 🙂

hollermay: Thanks for #3476! I'll adjust my PR to focus on the incomplete batch handling as you suggested.

laggui: Just merged into main if you want to rebase/merge your branch btw 🙂

crates/burn-core/src/data/dataloader/strategy.rs
laggui: This comment is still relevant, I think the changes can be simplified.

hollermay:
Hey @laggui sorry for late reply! I was a little off my work, made the changes with strategy.rs. I think it must be resolved by now

laggui (commented):
> Hey @laggui sorry for late reply! I was a little off my work, made the changes with strategy.rs. I think it must be resolved by now

No problem. What about the `dataloader .iter()` changes (as per my last comments)? These are no longer required with #3476 I believe.

P.S: CI failed on the formatting

hollermay:
> > Hey @laggui sorry for late reply! I was a little off my work, made the changes with strategy.rs. I think it must be resolved by now
> 
> No problem. What about the `dataloader .iter()` changes (as per my last comments)? These are no longer required with #3476 I believe.
> 
> P.S: CI failed on the formatting

You are right it's no longer required with #3476

laggui:
Closing due to inactivity, but feel free to re-open if you ever jump back on this!

PR #3451, opened 2025-07-31
Refactor burn-train

Cielbird:
Replaced devices() with learning_strategy() in Learner

This changes the API, but clears the way for better definitions of training strategies (data parallelism...)


crates/burn-train/src/learner/strategies/base.rs
```diff
@@ -0,0 +1,124 @@
...
+        starting_epoch: usize,
+    ) -> Learner<LC>
+    where
+        TI: Send + 'static,
+        TO: Send + 'static,
+        LC: LearnerComponents<Backend = B>,
+        LC::EventProcessor: EventProcessor<ItemValid = VO, ItemTrain = TO>,
+        <LC::Model as AutodiffModule<LC::Backend>>::InnerModule: ValidStep<VI, VO>,
+        <LC as LearnerComponents>::Model: TrainStep<TI, TO>;
+}
```
nathanielsimard: Don't use the trait if it's to be implemented by the enum

crates/burn-train/src/learner/strategies/single_device.rs
```diff
@@ -0,0 +1,106 @@
...
+{
+    let LearnerDataLoaders::SingleTrainSingleValid {
+        dataloader_train,
+        dataloader_valid,
+    } = dataloaders
+    else {
+        panic!("Wrong dataloaders for strategy");
+    };
+
+    let mut epoch_train = SingleDeviceTrainEpoch::new(
```
nathanielsimard: Move the traing epoc struct in the strategies folder. Also I would probably have:

```
strategies/single/epoch.rs
strategies/single/dataloaders.rs
strategies/single/learn.rs
etc.
```

PR #3452, opened 2025-07-31
Kill `tensor^T` magic transpose marker in favor of `tensor.T()`.

crutcher:
### Checklist

### Changes

IDEs were frequently auto-importing `use burn::tensor::T` while editing other
unrelated generic rust expressions (which use `T`) in `burn` projects.

- killed `tensor^T` where `T` was a magic marker struct.
- added `tensor.T()` as an alias for `tensor.transpose()`.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -272,6 +272,7 @@ where
     }
 
     /// Alias for `transpose`.
+    #[inline(always)]
     #[allow(non_snake_case)]
     pub fn T(self) -> Tensor<B, D, K> {
         self.transpose()
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -273,7 +273,6 @@ where
 
     /// Alias for `transpose`.
     #[inline(always)]
-    #[allow(non_snake_case)]
     pub fn t(self) -> Tensor<B, D, K> {
         self.transpose()
     }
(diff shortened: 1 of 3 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -271,6 +271,13 @@ where
         Tensor::new(K::transpose(self.primitive))
     }
 
+    /// Alias for `transpose`.
+    #[inline(always)]
+    #[allow(non_snake_case)]
+    pub fn T(self) -> Tensor<B, D, K> {
```
laggui: I feel like `x.T()` reads strangely (and is not very rust-like).

Plus it is a bit inconsistent with all conventions:
- `x ^ T` is a mathematical-style notation (but not a great convention for code)
- `x.T`: numpy/jax (property, not available for rust)
- `x.t()`: pytorch land (only for 2D tensor in their case)

If we introduce a short-hand for `x.transpose()` I would go with `x.t()`, it feels more idiomatic.

crutcher: Done

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -271,6 +271,13 @@ where
         Tensor::new(K::transpose(self.primitive))
     }
 
+    /// Alias for `transpose`.
+    #[inline(always)]
+    #[allow(non_snake_case)]
```
laggui:
```suggestion
```
No longer needed

crutcher: Done

PR #3453, opened 2025-08-01
Add quantization params precision

laggui:
### Changes

- Cleaned up some duplicated code
- Made unpack flexible (was hardcoded for sign extension with U32 / QInt8)
- ~~Added QInt4 (WIP)~~ moved to a follow-up PR
- Added qparams precision

### Tests
Run `cargo test --release quant` in `burn-cuda` and `burn-wgpu`

### DIFF
--- a/crates/burn-tensor/src/tensor/quantization/scheme.rs
+++ b/crates/burn-tensor/src/tensor/quantization/scheme.rs
@@ -33,8 +33,8 @@ impl Default for QuantScheme {
             mode: QuantMode::Symmetric,
             q_type: QuantInputType::QInt8,
             q_store_type: QuantStoreType::U32,
-            q_params_precision: QuantFloatPrecision::Full,
-            acc_precision: QuantFloatPrecision::Full,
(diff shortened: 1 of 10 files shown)


crates/burn-tensor/src/tensor/quantization/scheme.rs
```diff
@@ -124,10 +136,12 @@ pub enum QuantMode {
...
 
-/// Quantization accumulator precision. This is the precision to used when accumulating values
-/// while executing algorithms such as matmul.
+/// Quantization floating-point precision.
+///
+/// This is used to represent the floating-point precision of quantization parameters like the scale(s)
+/// or the accumulation precision used during operations like matrix multiplication.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
-pub enum QuantAccPrecision {
+pub enum QuantFloatPrecision {
```
nathanielsimard, line 144: I would add F16, BF16 and F32 here. We need to be more specific.

PR #3475, opened 2025-08-04
fix: link in examples

domenicocinque:
### Changes

The link to the pytorch example was leading to a non existing page


domenicocinque:
Done! 🚀

PR #3482, opened 2025-08-05
Improve DataLoaderBuilder docs.

crutcher:
### Checklist

### Changes

Clarified the `DataLoaderBuilder::num_workers()` behavior in docs.

### DIFF
--- a/crates/burn-core/src/data/dataloader/builder.rs
+++ b/crates/burn-core/src/data/dataloader/builder.rs
@@ -131,50 +131,54 @@ where
             Some(strategy) => strategy,
             None => Box::new(FixBatchStrategy::new(1)),
         };
-        let num_threads = self.num_threads.unwrap_or(0);
-        if num_threads > 0 {
-            Arc::new(MultiThreadDataLoader::new(
+
+        if let Some(num_threads) = self.num_threads
+            && num_threads > 0
+        {
+            return Arc::new(MultiThreadDataLoader::new(
                 strategy,
                 dataset,
                 self.batcher,
                 num_threads,
                 device,
                 rng,
-            ))
-        } else {
-            Arc::new(BatchDataLoader::new(
-                strategy,
-                dataset,
-                self.batcher,
-                device,
-                rng,
-            ))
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/data/dataloader/builder.rs
```diff
@@ -151,6 +160,32 @@ mod tests {
...
+
+        #[derive(new, Clone)]
+        pub struct TestBatcher;
+
+        #[cfg(test)]
+        impl<I> Batcher<TestBackend, I, TestDevice> for TestBatcher {
+            fn batch(&self, _items: Vec<I>, device: &TestDevice) -> TestDevice {
+                *device
+            }
+        }
```
laggui: The other tests actually validate the device, hence the custom test batcher. For this test you can probably use the `TestBatcher` defined at the crate level:

https://github.com/tracel-ai/burn/blob/902578ba9385a1efd77aab60bcf20a19789def98/crates/burn-core/src/data/dataloader/batcher.rs#L26-L31

```rust
crate::data::dataloader::batcher::TestBatcher::new();
```

crates/burn-core/src/data/dataloader/builder.rs
```diff
@@ -124,24 +131,25 @@ where
             Some(strategy) => strategy,
             None => Box::new(FixBatchStrategy::new(1)),
         };
-        if let Some(num_threads) = self.num_threads {
-            return Arc::new(MultiThreadDataLoader::new(
+        let num_threads = self.num_threads.unwrap_or(0);
+        if num_threads > 0 {
```
laggui: The previous syntax (before your last commit to fix tests on prev) was nicer 😄 Once 1.89 is stable (2 days iirc) we could use the `if let` statement and set the MSRV for the CI to 1.88

crutcher: I'll just wait then; and update.

laggui: Soon #3492 🙂

PR #3485, opened 2025-08-06
Use tensor ops for clip_by_norm

laggui:
### Checklist

### Changes

Avoid possible CPU <=> GPU sync by using only tensor operations to clip

### Testing

Unit tests


crutcher:
My local tests show a train that was only using 60% GPU going to 100% GPU with only this change.

laggui:
> My local tests show a train that was only using 60% GPU going to 100% GPU with only this change.

Thanks for confirming with your tests 🙏

PR #3490, opened 2025-08-06
Shuffle indices in `SamplerDataset` for fully random order during sampling.

crutcher:
I noticed an order-biased distribution in `SamplerDataset::without_replacement()`;
and found this callout in the docs for `choose_multiple`:
// Although the elements are selected randomly, the order of elements in
// the buffer is neither stable nor fully random. If random ordering is
// desired, shuffle the result

### Checklist

### Changes

Shuffle `indices` after selection.

### DIFF
--- a/crates/burn-dataset/src/transform/sampler.rs
+++ b/crates/burn-dataset/src/transform/sampler.rs
@@ -1,4 +1,5 @@
 use crate::Dataset;
+use rand::prelude::SliceRandom;
 use rand::{Rng, SeedableRng, distr::Uniform, rngs::StdRng, seq::IteratorRandom};
 use std::{marker::PhantomData, ops::DerefMut, sync::Mutex};
 
@@ -70,6 +71,12 @@ where
                 if indices.is_empty() {
                     // Refill the state.
                     *indices = (0..self.dataset.len()).choose_multiple(rng, self.dataset.len());
+
+                    // From `choose_multiple` documentation:
+                    // Although the elements are selected randomly, the order of elements in
+                    // the buffer is neither stable nor fully random. If random ordering is
+                    // desired, shuffle the result.
+                    indices.shuffle(rng);
                 }
 
                 indices.pop().expect("Indices are refilled when empty.")


crutcher:
I went ahead and fully fixed the `without_replacement()` distribution in the over-sample case.

crutcher:
And a builder.

laggui:
Went ahead and self-requested a review for tomorrow 😅 didn't have time to get to this today

crates/burn-dataset/src/transform/sampler.rs
```diff
@@ -31,31 +33,131 @@ where
...
-            state: Mutex::new(SamplerState::WithReplacement(StdRng::from_os_rng())),
+            state: Mutex::new(state),
             input: PhantomData,
         }
     }
 
+    /// Creates a builder for the given dataset.
+    pub fn builder(dataset: D) -> SamplerDatasetBuilder<D, I> {
+        SamplerDatasetBuilder::for_dataset(dataset)
+    }
```
laggui: Following patterns across other builders, I wouldn't add a `SamplerDataset::builder`. Using `SamplerDatasetBuilder::new(dataset)` is preferable.

crates/burn-dataset/src/transform/sampler.rs
```diff
@@ -69,7 +171,19 @@ where
...
+                    }
+
+                    // From `choose_multiple` documentation:
+                    // > Although the elements are selected randomly, the order of elements in
+                    // > the buffer is neither stable nor fully random. If random ordering is
+                    // > desired, shuffle the result.
+                    //
+                    // Without this, for size~=ds.size; the indices will return
+                    // essentially in a linear order.
+                    indices.shuffle(rng);
```
laggui: Good catch on the `choose_multiple` doc. In this case, we can probably just call `choose_multiple` when selecting partial indices.

Instead of the loop, something like:

```rust
for _ in 0..self.size / self.dataset.len() {
    indices.extend(0..self.dataset.len());
}

// Add randomly sampled partial copy
let remainder = self.size % self.dataset.len();
if remainder > 0 {
    let partial = (0..self.dataset.len()).choose_multiple(rng, remainder);
    indices.extend(partial);
}

indices.shuffle(rng);
```

laggui (changes requested):
Thanks for fixing the distribution issue! And improving the docs 🙏 

Given the addition of the builder, I would keep the `SamplerDataset` public creation API the same. The more "complex" configs should be handled by the builder.

crates/burn-dataset/src/transform/sampler.rs
crutcher: Hah; I was actually doing this while you reviewed, so it's already pushed

crates/burn-dataset/src/transform/sampler.rs
crutcher: Done.

crutcher:
Based upon requests to clean up / simplify the construction API; and @laggui 's general comments about being less happy with builders; I built 3 utility option types that we can re-use; and a SamplerDatasetOptions type.

laggui (commented):
> I built 3 utility option types that we can re-use; and a SamplerDatasetOptions type

The implementation seems overly complex just to handle this. Three different enums with conversions and three trait systems for setters?

For reusability, I'd opt for a sampler config/builder instead. Inspired by part of your changes:

```rust
#[derive(Debug, Clone)]
pub struct SamplerConfig {
    pub size: SamplerSize,
    pub with_replacement: bool,
    // if we want to support passing an rng, then maybe an enum
    pub rng: Option<u64>,
}

#[derive(Debug, Clone)]
pub enum SamplerSize {
    Fixed(usize),
    Ratio(f64),
    Source,
}
```

We can then keep the current constructors for most use cases:
```rust
SamplerDataset::with_replacement(dataset, size)
SamplerDataset::without_replacement(dataset, size)
```

And for more configuration:
```rust
SamplerDataset::with_sampler(dataset, sampler) // SamplerConfig
```

crutcher:
@laggui my thoughts with the options is to then extend the whole dataset transform stack with a common set of options; because random source and size are actually needed by all of them, and having a common option setting api for specifying transforms would be helpful when integrating with config driven behavior (no branches at the call sites for different options)

this approach to an Options object is used in other places where there are multiple options in burn; like GridOptions

crutcher:
In user code, we can then say things like:
```rust
let ds = SamplerDataset::new(
  source_ds,
  SamplerDatasetOptions::default()
    .with_replacement(config.with_replacement)
    .with_rng(cfg.rng_source)
    .with_size_ratio(cfg.sample_ratio)
    );
```

The impl of SamplerDataset becomes pretty simple; but then we could also get similar tools for shuffle, etc.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+    /// Build a new rng from the system.
+    #[default]
+    System,
+
+    /// The rng is passed as an option.
+    FromRng(Box<StdRng>),
+
+    /// The rng is passed as a seed.
+    FromSeed(u64),
+}
```
laggui: Naming suggestions:
```rust
pub enum RngSource {
    Default,
    Seed(u64),
    Rng(StdRng),
}
```

The box is not necessary, and we can drop `From` prefix (should it be `SizeSource::FromRatio` otherwise?).

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+            RandomSource::FromSeed(seed) => StdRng::seed_from_u64(*seed),
+        }
+    }
+}
+
+impl From<RandomSource> for StdRng {
+    fn from(source: RandomSource) -> Self {
+        source.build()
+    }
+}
```
laggui: The `build` impl can simply go into `From`. If you want to get a `StdRng` from the enum, calling `.into()` is the way.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+where
+    T: Into<RandomSource>,
+{
+    fn from(rng: Option<T>) -> Self {
+        match rng {
+            Some(rng) => rng.into(),
+            None => Self::default(),
+        }
+    }
+}
```
laggui: Using `None` instead of specifying the default variant for the rng is confusing imo. We don't need this implementation.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+
+impl From<StdRng> for RandomSource {
+    fn from(rng: StdRng) -> Self {
+        Self::FromRng(Box::new(rng))
+    }
+}
+
+/// Helper trait to add random source setters to a type.
+pub trait WithRandomSourceSetters: Sized {
```
laggui: I don't really see the benefit of these traits. If we need the config to set the rng or size we can follow a simple convention, so I would remove them.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    /// Use an rng, built from a seed.
+    fn with_seed(self, seed: u64) -> Self {
+        self.with_rng(seed)
+    }
+}
+
+/// Helper option to describe the size of a wrapper, relative to a wrapped object.
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+pub enum WrapperSizeSource {
```
laggui: Naming: `WrapperSizeSource` -> `SizeMode` or `SizeConfig`

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+    /// Construct a source which will have a size that is ``floor(<source> * <ratio>)``.
+    pub fn ratio(ratio: f64) -> Self {
+        assert!(ratio >= 0.0, "Ratio must be positive: {ratio}");
+        Self::Ratio(ratio)
+    }
+
+    /// Construct a source which will have a fixed size.
+    pub fn fixed(size: usize) -> Self {
+        Self::Fixed(size)
+    }
```
laggui: I don't see why we need these instead of using the variants. Sure, the ratio is validated but we can do it in `evaluate_for_source`.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    /// Computes the effective size, given the source.
+    pub fn evaluate_for_source(self, source_size: usize) -> usize {
```
laggui: Suggestion: `size.resolve(source_size)`

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+impl<T> From<Option<T>> for WrapperSizeSource
+where
+    T: Into<WrapperSizeSource>,
+{
+    fn from(option: Option<T>) -> Self {
+        match option {
+            Some(option) => option.into(),
+            None => Self::default(),
+        }
+    }
```
laggui: Again, not sure why `None` should map to default. We have `default()`, or the enum variant.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -0,0 +1,481 @@
...
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum ReplacementMode {
+    /// New samples are selected with replacement.
+    #[default]
+    WithReplacement,
+
+    /// New samples are selected without replacement;
+    /// unless sampling more than the source dataset.
+    WithoutReplacement,
+}
```
laggui: This is a bool 😅 We can keep it as so in the actual sampler config, e.g.

```rust
#[derive(Debug, Clone, Default)]
pub struct SamplerConfig {
    pub size: SizeMode,
    pub with_replacement: bool,
    pub rng: RngSource,
}
```

crates/burn-dataset/src/transform/sampler.rs
```diff
@@ -1,10 +1,88 @@
...
+            None => Self::default(),
+        }
+    }
+}
+
+impl From<usize> for SamplerDatasetOptions {
+    fn from(size: usize) -> Self {
+        Self::default().with_fixed_size(size)
+    }
+}
```
laggui: Probably the only relevant `From` that we need to have for the sampler config, to maintain compatibility with the current API.

Should return a config with replacement, default rng and fixed size.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: done.

crates/burn-dataset/src/transform/options.rs
crutcher: See discussion of default below.

crates/burn-dataset/src/transform/sampler.rs
crutcher: Getting rid of ReplacementMode is a problem for defaults.

Right now, the Default::default() is WithReplacement; and that's inherrited from ReplacementMode. I could override SamplerConfig, and manually implement a Default.

Or the bool could be `without_replacement`, or some nonsesnse.

crutcher:
This isn't done, but needs review.

There's a question of how to handle defaults if we change ReplacementMode to a bool.

crates/burn-dataset/src/transform/sampler.rs
crutcher: Nm; done.

crates/burn-dataset/src/transform/options.rs
crutcher: Nm; done.

crutcher:
Nm, I think this is done.

Issue #2816, opened 2025-02-15
yolov10 ReduceMax keepdims == 0


sihaiwuhu:
Hi, there is a panic like this when building from yolov10.onnx model. How should I solve this error?

ERROR burn_import::logger: PANIC => panicked at /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-import-0.16.0/src/onnx/op_configuration.rs:1467:9:
  ReduceMax: the reduce operation must preserve the reduced dimension    

  --- stderr
  thread 'main' panicked at /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-import-0.16.0/src/onnx/op_configuration.rs:1467:9:
  ReduceMax: the reduce operation must preserve the reduced dimension
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
ubuntu@PC-PW09NVP9:/mnt/d/workspace/rust_ws/burn_demo1$

antimora:
`keepdims == 0` has been fixed by https://github.com/tracel-ai/burn/pull/3497

PR #3497, opened 2025-08-09
Extend onnx reduce op support

tye-singwa:
### Checklist

### Related Issues/PRs

#2816

### Changes

Completing support for Reduce* onnx ops:
- `keepdims` param
- multiple axes support
- `ReduceL1`, `ReduceL2`, `ReduceLogSum`, `ReduceLogSumExp`, `ReduceSumSquare`

### Testing

Added test for each op


crates/burn-import/src/burn/node/reduce.rs
```diff
@@ -0,0 +1,733 @@
...
+                );
+
+                match self.input.kind {
+                    TensorKind::Int => {
+                        quote! { #input_square_reduced.float().cast(burn::tensor::DType::F32).sqrt().int() }
+                    }
+                    TensorKind::Float => {
+                        quote! {
+                            let input_dtype = #input.dtype();
+                            #input_square_reduced.cast(burn::tensor::DType::F32).sqrt().cast(input_dtype)
```
laggui: This is to avoid overflow/underflow in lower precision right? Should probably add a small comment above the `quote! {` to give more context for the cast.

crates/burn-import/src/burn/node/reduce.rs
```diff
@@ -0,0 +1,733 @@
...
+                if keepdims {
+                    quote! { #reduced_input.expand([1; #output_rank]) }
+                } else {
+                    reduced_input
+                }
+            } else {
+                // Reducing along all specific dimensions
+                if dims.is_empty() {
+                    dims = (0..input_rank).collect();
+                }
```
laggui: `dims` is already known to be empty at this point

crates/burn-import/src/burn/node/reduce.rs
```diff
@@ -0,0 +1,733 @@
...
+        let reduced_input = dims.iter().fold(input, |tokens, dim| {
+            reduction_type.forward_reduce_by_dim(tokens, *dim)
+        });
+
+        if keepdims {
+            reduced_input
+        } else {
+            // Squeezing dimensions
+            let dims = dims.to_tokens();
+            quote! { #reduced_input.squeeze_dims(&#dims) }
```
laggui: We should be mindful about `squeeze_dims(...)` if the output is reduced to a scalar (rank 0)

laggui (commented):
Thanks for adding the other reduction types!

I think for `keepdims = 0` we need to check for the scalar output case.

crates/burn-import/src/burn/node/reduce.rs
tye-singwa: Yep, this case i missed.

crates/burn-import/src/burn/node/reduce.rs
tye-singwa: i implemented it based on official doc https://onnx.ai/onnx/operators/onnx__ReduceL2.html#function-body, not sure it is required. Will check

crates/burn-import/src/burn/node/reduce.rs
tye-singwa: Yep, it is redundant

crates/burn-import/src/burn/node/reduce.rs
laggui: I think it makes sense, but probably would be good to leave a small comment to state the requirement 🙂

antimora:
Fixed the issues myself because I needed the fixes. I'll merge after CI passes.

antimora (approved):

antimora:
@tye-singwa , thank you so much for your contribution! Your PR was very timely. 

I liked how you combined reduce ops and your tests were extensive.

PR #3498, opened 2025-08-10
docs: fix example

domenicocinque:
The example in the readme was broken

### DIFF
--- a/examples/onnx-inference/README.md
+++ b/examples/onnx-inference/README.md
@@ -76,8 +76,8 @@ https://huggingface.co/datasets/ylecun/mnist/viewer/mnist/test?row=15
        // Get a default device for the models's backend
        let device = Default::default();
 
-       // Create a new model and load the state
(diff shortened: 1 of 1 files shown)


examples/onnx-inference/README.md
```diff
@@ -76,8 +76,8 @@ https://huggingface.co/datasets/ylecun/mnist/viewer/mnist/test?row=15
        // Get a default device for the models's backend
        let device = Default::default();
 
-       // Create a new model and load the state
-       let model: Model<Backend> = Model::new(&device).load_state();
+       // Create a new model
+       let model: Model<NdArray> = Model::new(&device);
```
laggui: Actually to load the saved weights it should be `Model::default()` or `Model::from_embedded(&device)`

domenicocinque: Thanks. I updated the code and the comment (following the example in the [docs](https://burn.dev/books/burn/import/onnx-model.html))

PR #3514, opened 2025-08-12
Fix ReduceMean to support dimension reduction with keepdims=0

antimora:
The ONNX ReduceMean operator wasn't properly handling the keepdims=0 case, which should reduce the tensor rank by removing the reduced dimensions. Previously, the implementation would panic with "ReduceMean: the reduce operation must preserve the reduced dimension" when keepdims=0 was specified.

### Checklist

### Related Issues/PRs

#3494

### Changes
This PR implements full support for the ReduceMean operator with both keepdims=0 and keepdims=1, as well as multi-axis reduction:

1. Added squeeze operation: When keepdims=0, the implementation now applies squeeze_dims() after the mean reduction to remove the reduced dimensions, properly reducing the output tensor rank.
2. Fixed rank calculation: The output rank is now correctly calculated as input_rank - num_reduced_axes when keepdims=0.
3. Multi-axis support: The implementation now handles reduction over multiple axes correctly, applying mean operations sequentially in descending axis order.
4. Refactored into dedicated node: Moved ReduceMean from the generic UnaryNode to its own dedicated ReduceMeanNode for better code organization and maintainability.

### Testing

New ONNX file tests


antimora:
ahh, I should have reviewed carefully #3497 which also implements ReduceMean in one go.

Issue #3511, opened 2025-08-11
Conv2d groups backward slice out of bounds


laggui:
Initially reported [on discord](https://discord.com/channels/1038839012602941528/1091796857996451942/1404548545750499471).

```rust
use burn::backend::{Autodiff, Wgpu};
use burn::nn::conv::Conv2dConfig;
use burn::tensor::Tensor;

fn main() {
    type B = Autodiff<Wgpu>;
    let device = Default::default();

    let conv = Conv2dConfig::new([48, 12], [3, 3])
        .with_stride([2, 2])
        .with_groups(12)
        .init(&device);

    let input = Tensor::<B, 4>::ones([1, 48, 8, 8], &device);

    let output = conv.forward(input);
    output.backward();
}
```

Error:
```
thread 'main' panicked at /home/laggui/workspace/burn/crates/burn-cubecl/src/kernel/index/slice.rs:17:44:
attempt to subtract with overflow
```

(ndarray backend):
```
thread 'main' panicked at /home/laggui/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ndarray-0.16.1/src/impl_methods.rs:557:22:
Slice end 4 is past end of axis of length 1
```

Puranjay-del-Mishra:
I'm on it, fixing the range overflow in kernel/slice.rs!

PR #3515, opened 2025-08-12
cubecl/index: clamp ranges; saturating_sub fixes conv-bwd underflow

Puranjay-del-Mishra:
### Checklist

- [Yes] Confirmed that `cargo run-checks` command has been executed.
- [ Nil] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

Related issue: https://github.com/tracel-ai/burn/issues/3511

### Changes

1. The slice helpers did unchecked range math; when backprop produced a window with end < start, expressions like (dim - end) and (end - start) underflowed usize, causing the “attempt to subtract with overflow” panic.

2. Fix: clamp {start,end} to [0, dim] and use saturating_sub for (dim - end) and (end - start) (mirrored in slice_assign), so grouped conv backward with stride can’t generate invalid slices.

### Testing

Added a regression test (regression_grouped_conv_backward.rs) under the GPU backend(crates/burn-wgpu/tests:
```
use burn::backend::{Autodiff, Wgpu};
use burn::nn::conv::Conv2dConfig;
use burn::tensor::Tensor;

#[test]
fn grouped_conv_backward_wgpu_no_panic() {
    type B = Autodiff<Wgpu>;

    let conv = Conv2dConfig::new([48, 12], [3, 3])
        .with_stride([2, 2])
        .with_groups(12)
        .init(&device);

    let input = Tensor::<B, 4>::ones([1, 48, 8, 8], &device);
    let output = conv.forward(input);
    output.backward(); // repro: should not panic after fix
}
```
Ran from the repo root: 
`cargo test --test regression_grouped_conv_backward -- --nocapture; 
`
the test passes (no panic). Also ran formatting and checks: 
```
cargo fmt;
cargo fmt -- --check;
cargo run -p xtask check; 
cargo run-checks; 
```

### DIFF
--- a/crates/burn-cubecl/src/kernel/index/slice.rs
+++ b/crates/burn-cubecl/src/kernel/index/slice.rs
@@ -3,6 +3,14 @@ use burn_tensor::Shape;
 use cubecl::{calculate_cube_count_elemwise, prelude::*};
 use std::ops::Range;
 
+// clamping the start and end values to avoid overflow
+#[inline]
+fn clamp_range(dim: usize, r: &Range<usize>) -> (usize, usize) {
+    let s = r.start.min(dim);
+    let e = r.end.min(dim);
+    if e < s { (s, s) } else { (s, e) }
+}
+
 /// Slice a jit tensor with a set of ranges
 pub fn slice<R: CubeRuntime, E: CubeElement>(
     tensor: CubeTensor<R>,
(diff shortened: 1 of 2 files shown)


laggui (commented):
Thanks for trying to tackle this bug.

Your changes address the symptom, not the problem. The issue is actually with the slice provided in the `conv2d_weight_grad_groups`.

I am working on it. The slices have been fixed but there is another problem when computing the input gradients with the conv transpose due to the groups.

PR #3520, opened 2025-08-12
save_tensor_as_image utility

Cielbird:
Adds `save_tensor_as_image` to the vision crate for utilities

### DIFF
--- a/crates/burn-vision/src/show.rs
+++ b/crates/burn-vision/src/show.rs
@@ -1,0 +1,191 @@
+//! Utilities for saving tensors as images
+
+use burn_tensor::{ElementConversion, Tensor, backend::Backend};
+use image::{Rgb, RgbImage};
+use std::fs;
+use std::path::Path;
+
+/// How to save a tensor as an image
+pub struct TensorDisplayOptions {
+    /// How should the dimensions be interpreted
+    pub dim_order: ImageDimOrder,
+    /// What colors should be used
+    pub color_opts: ColorDisplayOpts,
+    /// How to handle batches
+    pub batch_opts: Option<BatchDisplayOpts>,
+    /// Output image width
+    pub width_out: usize,
+    /// Output image height
+    pub height_out: usize,
+}
+
+/// How to interpret dimensions for image tensors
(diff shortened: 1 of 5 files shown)


crates/burn-vision/src/show.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// dims: (channels, height, width)
+    CHW,
+    /// dims: (height, width, channels)
+    HWC,
+    /// dims: (batch_size, height, width)
+    NHW,
+    /// dims: (batch_size, channels, height, width)
+    NCHW,
+    /// dims: (batch_size, height, width, channels)
+    NHWC,
```
antimora: Should be camel case to be idiomatic and consistent just like with Rgb below.

crates/burn-vision/src/show.rs
```diff
@@ -0,0 +1,191 @@
...
+
+/// Save a tensor of a batch of images as an image
+///
+/// * `tensor` - Image batch with shape (N, height, width)
+/// * `opts` - Options for how to draw the tensor
+/// * `path` - The file path to use
+pub fn save_tensor_as_image<B: Backend, const D: usize>(
+    tensor: Tensor<B, D>,
+    opts: TensorDisplayOptions,
+    path: &str,
```
antimora: `path: &str` is better to have as `<P: AsRef<Path>>(path: P)`

crates/burn-vision/src/show.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    img.save(path)?;
+    Ok(())
+}
+
+/// Normalize values in 2D tensor from 0 to 1
+fn normalize<B: Backend, const D: usize>(tensor: Tensor<B, D>) -> Tensor<B, D> {
+    let min = tensor.clone().min().to_data().to_vec::<f32>().unwrap()[0];
+    let max = tensor.clone().max().to_data().to_vec::<f32>().unwrap()[0];
```
antimora: instead of to_data, etc, use to_scalar()

nathanielsimard (approved):

PR #3522, opened 2025-08-12
grid_sample ndarray implementation

Cielbird:
Added a ndarray specific implementation of the grid_sample_2d operation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1320,9 +1320,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.44"
+version = "4.5.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c1f056bae57e3e54c3375c41ff79619ddd13460a17d7438712bd0d83fda4ff8"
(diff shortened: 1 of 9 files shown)


crates/burn-ndarray/src/ops/grid_sample.rs
```diff
@@ -0,0 +1,73 @@
...
+                id % strides.0 / strides.1,
+                id % strides.1 / strides.2,
+                id % strides.2,
+            );
+
+            let sample_x = grid[(b, y, x, 0)].elem::<f64>();
+            let sample_y = grid[(b, y, x, 1)].elem::<f64>();
+
+            let sample_x = sample_x.min(1_f64);
+            let sample_y = sample_y.min(1_f64);
```
laggui: Related to #3523, is this clamped to handle -inf/+inf? If so, we probably need to handle both with something like `sample_y.min(1.0).max(-1.0)`?

Cielbird: It's clamped to keep things between [-1, 1], I missed the `.max(-1.0)`

laggui (approved):

PR #3526, opened 2025-08-12
Add affine_grid_2d

Cielbird:
Added affine_grid_2d, which generates a sampling grid given a batch of matrices.

Closely follows pytorch's [torch.nn.functional.affine_grid](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.affine_grid.html). This is only the 2D case.

This is useful in conjunction with grid_sample_2d for transforming images.

I was unsure where to put the function, lmk if there's a better spot!

### DIFF
--- a/crates/burn-tensor/src/tensor/grid/affine_grid.rs
+++ b/crates/burn-tensor/src/tensor/grid/affine_grid.rs
@@ -1,0 +1,73 @@
+use crate::ElementConversion;
+use crate::backend::Backend;
+use crate::tensor::{Int, Tensor};
+
+/// Generate a tensor with homogeonous coordinates of each element's
+/// transformed location
+///
+///
+/// See:
+///  - [torch.nn.functional.affine_grid](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.affine_grid.html)
+///
+/// * `transform` - Transformation with shape (batch_size, 2, 3)
+/// * `dims` - dimensions as (batch_size, channels, height, width)
+///
+/// # Returns
+///
+/// Tensor with shape (batch_size, height, width, 2), where dim 2 is (x, y)
+/// All coordinates are broadcast on the batch dim
+pub fn affine_grid_2d<B: Backend>(transform: Tensor<B, 3>, dims: [usize; 4]) -> Tensor<B, 4> {
+    let batch_size = dims[0];
+    let _channels = dims[1];
+    let height = dims[2];
+    let width = dims[3];
+
+    let device = &transform.device();
+
+    let x = Tensor::<B, 1, Int>::arange(0..width as i64, device)
+        .reshape([1, width, 1])
+        .expand([height, width, 1]);
+    let y = Tensor::<B, 1, Int>::arange(0..height as i64, device)
+        .reshape([height, 1, 1])
+        .expand([height, width, 1]);
+
+    // from ints (0..(width-1)) and (0..(height-1)), to (-1.0..1.0)
+    let x = x
+        .float()
+        .div_scalar(((width - 1) as f32 / 2.0).elem::<f32>())
+        .sub_scalar((1_f32).elem::<f32>());
+    let y = y
+        .float()
+        .div_scalar(((height - 1) as f32 / 2.0).elem::<f32>())
+        .sub_scalar((1_f32).elem::<f32>());
+    let w = Tensor::<B, 3>::ones([height, width, 1], device);
+
+    let grid = Tensor::cat(vec![x, y, w], 2);
+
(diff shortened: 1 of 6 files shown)


crates/burn-tensor/src/tensor/grid/affine_grid.rs
```diff
@@ -0,0 +1,73 @@
...
+///
+/// # Returns
+///
+/// Tensor with shape (batch_size, height, width, 2), where dim 2 is (x, y)
+/// All coordinates are broadcast on the batch dim
+pub fn affine_grid_2d<B: Backend>(transform: Tensor<B, 3>, dims: [usize; 4]) -> Tensor<B, 4> {
+    let batch_size = dims[0];
+    let _channels = dims[1];
+    let height = dims[2];
+    let width = dims[3];
```
laggui:
```rust
let [batch_size, _c, height, width] = dims;
```

crates/burn-tensor/src/tensor/grid/affine_grid.rs
```diff
@@ -0,0 +1,73 @@
...
+        1,
+    );
+    let transform = transform
+        .reshape([batch_size, 1, 1, 3, 3])
+        .expand([batch_size, height, width, 3, 3]);
+    let grid = grid
+        .reshape([1, height, width, 3, 1])
+        .expand([batch_size, height, width, 3, 1]);
+    println!("{}", transform.clone());
+    println!("{}", grid.clone());
```
laggui: Left over print

crates/burn-tensor/src/tensor/grid/affine_grid.rs
```diff
@@ -0,0 +1,73 @@
...
+    let result = transform.matmul(grid);
+    let result = result.reshape([batch_size, height, width, 3]);
+
+    // Homogeneous coordinates (x, y, w) to normal coordinates (x/w, y/w)
+    let mut grid = result.split_with_sizes(vec![2, 1], 3);
+    let grid_xy = grid.remove(0);
+    let grid_w = grid.remove(0);
+    let grid_w = grid_w.expand([batch_size, height, width, 2]);
+
+    grid_xy.div(grid_w)
```
laggui: Had to read up a bit on affine grid usage.

The shape manipulations for intermediate tensors to perform the matmul seem overly complex.

I think this could be simplified with element-wise operations to apply the transformation on every pixel directly 🤔 

```rust
// Broadcast to batch dimension
let x = x.unsqueeze_dim::<3>(0).expand([batch_size, height, width]); // [B, H, W]
let y = y.unsqueeze_dim::<3>(0).expand([batch_size, height, width]); // [B, H, W]

// Apply affine transform
let scale_x = transform.clone().slice(s![.., 0, 0]);
let shear_x = transform.clone().slice(s![.., 0, 1]);
let trans_x = transform.clone().slice(s![.., 0, 2]);

let shear_y = transform.clone().slice(s![.., 1, 0]);
let scale_y = transform.clone().slice(s![.., 1, 1]);
let trans_y = transform.slice(s![.., 1, 2]);

let grid_x = scale_x.mul(x.clone()).add(shear_x.mul(y.clone())).add(trans_x);
let grid_y = shear_y.mul(x).add(scale_y.mul(y)).add(trans_y);

Tensor::stack(vec![grid_x, grid_y], 3)
```

I just tested locally and the results are the same for the unit tests.

laggui (commented):
Implementation LGTM, but I think we can improve it slightly

> I was unsure where to put the function, lmk if there's a better spot!

I think in the grid module makes sense. Actually the accompanying `grid_sample` would probably have a better place there, though it seems it was added with backend implementations override in mind so we'd have to rethink that.

crates/burn-tensor/src/tensor/grid/affine_grid.rs
laggui: Nvm, looks like you already removed those since I started the review 😅

crates/burn-tensor/src/tensor/grid/affine_grid.rs
Cielbird: Looks good! I think `shear` and `scale` aren't quite accurate though because rotation affects both.

laggui: Ahhh I think you're right. We can be a bit more general about the transformation then, given an affine transform

$$
x_{tfm} = ax + by + t_{x}
$$
$$
y_{tfm} = cx + dy + t_{y}
$$

we can use `a, b, tx` and `c, d, ty` variable names (or something similar)

laggui (approved):

PR #3529, opened 2025-08-13
Add missing interpolate mode

laggui:


laggui:
Fix was included when merging #3526

PR #2995, opened 2025-04-04
Adding inital image augmentation.rs module in burn-datasets

catch-twenty-two:
Image augmentation and preprocessing utilities.

`Augmentations` provides randomized image transformations commonly used to
boost model performance through data augmentation.

# Supported Augmentations

- `random_photometric_distortion` — Adjusts brightness, contrast and hue with a set probability
- `random_zoom_out` — Pads and resizes to simulate zooming out with a set probability.
- `random_vertical_flip` — Flips images vertically with a set probability.

# Notes

- Augmentations are applied probabilistically and may vary on each call.
- Designed to be composable and easily integrated into preprocessing pipelines.
- Aims to improve model robustness to lighting, scale, and spatial orientation.

### Checklist

- [x] Confirmed that `run-checks all` script has been executed.
- [X] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/207

https://github.com/tracel-ai/burn/issues/2361

https://github.com/tracel-ai/burn/pull/2426

### Changes

This adds an initial module for image augmentation using a the image-rs library and the tensor backend where it was easily implementable. 

### Testing

Tested with actual images, also tested using unit tests.

PR #3534, opened 2025-08-13
Mnist example augmented data

Cielbird:
Transform the dataset images for the mnist dataset before training for better inference


antimora:
This PR also had some useful augmentation utils that you might want to lift as well: 

https://github.com/tracel-ai/burn/pull/3254 and this https://github.com/tracel-ai/burn/pull/2995

PR #3537, opened 2025-08-14
Dot Product Op

kikefdezl:
### Changes

Added the dot product operation to `Tensor`,

Checks that both tensors are 1D and equal in length. Then computes `mul + sum` using existing methods.

Similar to [torch.dot](https://docs.pytorch.org/docs/stable/generated/torch.dot.html).

### Testing

Added some basic unit tests for happy path and panicking.

--- 

I'm quite new to the codebase, so any support from the maintainers is greatly appreciated if needed.


crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -277,6 +277,39 @@ where
...
+    ///    let tensor = tensor1.dot(tensor2);
+    ///    println!("{tensor}");
+    ///    // [4]
+    /// }
+    /// ```
+    #[allow(clippy::should_implement_trait)]
+    pub fn dot(self, other: Self) -> Tensor<B, 1, K> {
+        check!(TensorCheck::dot(&self, &other));
+        Self::mul(self, other).sum()
+    }
```
laggui: We could simply make this available for rank 1 tensors:

```rust
impl<B, K> Tensor<B, 1, K>
where
    B: Backend,
    K: Numeric<B>,
    K::Elem: Element,
{

    pub fn dot(self, other: Self) -> Self {
        self.mul(other).sum()
    }
}
```

That way, there is also no need for a custom `TensorCheck` since 1) rank has to be 1 and 2) the shapes are checked by the mul elemwise check.

kikefdezl: You're totally right, thank you

Changes done 🙂

kikefdezl:
My bad, missed that part of the book.

Added :+1:

PR #3539, opened 2025-08-14
Debug log new name during ONNX renames

torsteingrindvik:
If some node "foo" crashes late during ONNX import it can be useful to inspect which node this relates to in e.g. netron.app.

However if the node has been renamed it's not as easy to find the original node.

Therefore also print the new name during renames such that we can find the original name easier in logs, which allows searching for the original node more easily.

The below example would have a log statement about "node_conv_9" being renamed "foo".

image

### DIFF
--- a/crates/onnx-ir/src/from_onnx.rs
+++ b/crates/onnx-ir/src/from_onnx.rs
@@ -262,7 +262,7 @@ impl OnnxGraphBuilder {
         )
         .to_lowercase();
 
-        log::debug!("renaming node {:?} to {new_name:?}", &node.name);
+        log::debug!("Renaming node {:?} to {new_name:?}", &node.name);
 
         node.name.clone_from(&new_name);
     }


laggui (approved):

antimora:
That's a good one. I also did it in this pending PR: #3381

image

We can merge it with small request: Capitalize it the message like "Renaming ..."

crates/onnx-ir/src/from_onnx.rs
```diff
@@ -262,6 +261,9 @@ impl OnnxGraphBuilder {
             node.node_type, self.node_name_counter[&node.node_type]
         )
         .to_lowercase();
+
+        log::debug!("renaming node {:?} to {new_name:?}", &node.name);
```
antimora:
```suggestion
        log::debug!("Renaming node {:?} to {new_name:?}", &node.name);
```

PR #3541, opened 2025-08-14
Proto conversion: Allow f16 tensors

torsteingrindvik:
From the proto comments: 

image

I had a model with a f16 tensor as input which failed due to the `TryFrom<TensorProto> for TensorData` missing the handling of f16.

With the knowledge that f16 can be stored in the raw data field, handle this case.

Add the bytemuck flag to the half crate such that this cast is possible.

My model no longer fails this step on these changes.

### DIFF
--- a/crates/onnx-ir/src/proto_conversion.rs
+++ b/crates/onnx-ir/src/proto_conversion.rs
@@ -34,6 +34,15 @@ impl TryFrom<TensorProto> for TensorData {
                     Data::Float32s(tensor.float_data)
                 },
             ),
+            DataType::FLOAT16 => (
+                ElementType::Float16,
+                // Convert the raw data to a vector of float16s
+                if !tensor.raw_data.is_empty() {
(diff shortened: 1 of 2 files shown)


crates/onnx-ir/src/proto_conversion.rs
```diff
@@ -34,6 +34,15 @@ impl TryFrom<TensorProto> for TensorData {
                     Data::Float32s(tensor.float_data)
                 },
             ),
+            DataType::FLOAT16 => (
+                ElementType::Float16,
+                // Convert the raw data to a vector of float16s
+                if !tensor.raw_data.is_empty() {
+                    Data::Float16s(cast_slice(&tensor.raw_data[..]).to_vec())
```
laggui, line 41: This pattern is already used in the other arms, but I just noticed we could probably try to cast the vec directly and fallback to use the slice.

```rust
try_cast_vec(tensor.raw_data)unwrap_or_else(|(_, raw_data)| cast_slice(&raw_data).to_vec());
```

Not required for this PR, but we can change that in a follow-up

PR #3544, opened 2025-08-14
Support int min/max types in clip_config

antimora:
Added handling for Int32 and Int64 types for min and max values in the clip_config function. 

Came across a case when building deberta model. This should fix it.

### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/pull/3381

### Changes

Add more datatypes. 

### Testing

Built deberta model.

### DIFF
--- a/crates/onnx-ir/src/node/clip.rs
+++ b/crates/onnx-ir/src/node/clip.rs
@@ -33,7 +33,9 @@ pub fn clip_config(node: &Node) -> (Option<f64>, Option<f64>) {
                 Data::Float16(min) => Some(f32::from(min) as f64),
                 Data::Float32(min) => Some(min as f64),
                 Data::Float64(min) => Some(min),
-                _ => panic!("Clip: only float min is supported"),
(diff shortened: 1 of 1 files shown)


crates/onnx-ir/src/node/clip.rs
```diff
@@ -33,7 +33,9 @@ pub fn clip_config(node: &Node) -> (Option<f64>, Option<f64>) {
                 Data::Float16(min) => Some(f32::from(min) as f64),
                 Data::Float32(min) => Some(min as f64),
                 Data::Float64(min) => Some(min),
-                _ => panic!("Clip: only float min is supported"),
+                Data::Int32(min) => Some(min as f64),
+                Data::Int64(min) => Some(min as f64),
+                _ => panic!("Clip: unsupported min data type {:?}", min),
```
laggui, line 38: Looks like int is also supported since version 12, so min and max should type should probably be based off the input dtype.

Totally fine for now though

PR #3549, opened 2025-08-15
Fix MNIST example model

Cielbird:


antimora:
@Cielbird 

Can we also fix the live demo link to `https://burn.dev/#demo` from `https://burn.dev/demo`? Since the main website was changed recently:

Issue #3548, opened 2025-08-15
ONNX problem: ConstantOfShape


torsteingrindvik:
I have a model which fails like so:

<img width="855" height="867" alt="Image" src="https://github.com/user-attachments/assets/f2771953-9b1f-461f-b07f-b43cea1bb1f4" />

netron view of node:

<img width="1926" height="1180" alt="Image" src="https://github.com/user-attachments/assets/b76c94ec-cbff-46c7-93f6-26d5534292e9" />

Related code:

<img width="813" height="904" alt="Image" src="https://github.com/user-attachments/assets/53c58dbf-4c95-40fe-b0ea-e6a837234255" />

Some context: https://discordapp.com/channels/1038839012602941528/1091796857996451942/1405609521568677990

I tried the Shape PR: https://github.com/tracel-ai/burn/pull/3381
It did not fix this issue.

I have a minimal test which reproduces the model node (will link in a minute when I have a PR for it)

torsteingrindvik:
Feel free to cherry-pick the test in #3550 when this is fixed.

PR #3550, opened 2025-08-15
Add minimal test showing bug with ConstantOfShape

torsteingrindvik:
Does not pass as of now, but should.

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/3548

### Changes

A minimal test of a problem found in the wild is added as a test case.

When the bug is fixed this PR should pass.


torsteingrindvik:
I tested my model on this PR with your new commits @antimora , now the model doesn't complain about `ConstantOfShape` any more. Thank you :tada:

PR #3553, opened 2025-08-15
Fixed local burn-collective ops

Cielbird:
# Fixed local burn-collective ops

Consecutive calls to all-reduce sometimes blocked, the all_reduce_ops map wasn't being correctly emptied. Same with reduce and broadcast.


nathanielsimard (changes requested):
Outdated

PR #3538, opened 2025-08-14
Infer conv spatial dims from weight rank

laggui:
### Related Issues/PRs

As reported on discord:

```
DEBUG onnx_ir::proto_conversion: Converting ONNX node with type "Conv"    
  ERROR burn_import::logger: PANIC => panicked at /home/togr/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/onnx-ir-0.18.0/src/node_remap.rs:8:74:
  called `Option::unwrap()` on a `None` value   
```

### Changes

As per ONNX spec, the kernel shape in conv nodes is optional and should be inferred from the weights input. So we must support when it is not provided.

### Testing

Added tests

PR #3554, opened 2025-08-15
onnx-ir: Infer conv2d kernel shape from weight tensor

torsteingrindvik:
If not set explicitly in attrs per spec.

### Changes

Fix a TODO in code related to inferring a shape from a Conv node.

### Testing

My model no longer crashes at this point.
Before these changes the `kernel_shape[0]` access panics due to no kernel shape being set.


torsteingrindvik:
> Part 2 of #3538 😅
> 
> I think the same logic should also be handled for conv1d/3d and conv transpose1d/2d/3d. Do you mind filling that in as well?
> 
> Otherwise I can do it in a follow-up PR.

I'll take a quick stab at filling those in!

torsteingrindvik:
Added same logic for conv1d,2d,3d and added a test for each. Seems to work out well.

Will look at the transpose ones now

torsteingrindvik:
> Part 2 of #3538 😅
> 
> I think the same logic should also be handled for conv1d/3d and conv transpose1d/2d/3d. Do you mind filling that in as well?
> 
> Otherwise I can do it in a follow-up PR.

Should be ready to go!

crates/onnx-ir/src/node/conv_transpose1d.rs
```diff
@@ -98,10 +98,25 @@ pub fn conv_transpose1d_config(curr: &Node) -> ConvTranspose1dConfig {
     let channels_in = weight_shape[1] * group;
     let channels_out = weight_shape[0];
 
+    let kernel_size = if kernel_shape.is_empty() {
+        // https://onnx.ai/onnx/operators/onnx__Conv.html#attributes
```
laggui: Copy-pasta link 😄

crates/onnx-ir/src/node/conv_transpose2d.rs
```diff
@@ -96,9 +96,24 @@ pub fn conv_transpose2d_config(curr: &Node) -> ConvTranspose2dConfig {
     // the channels are inverted in the weight tensor
     let channels: [usize; 2] = [weight_shape[1] * group, weight_shape[0]];
 
+    let kernel_size = if kernel_shape.is_empty() {
+        // https://onnx.ai/onnx/operators/onnx__Conv.html#attributes
```
laggui: Same

crates/onnx-ir/src/node/conv_transpose3d.rs
```diff
@@ -98,13 +98,28 @@ pub fn conv_transpose3d_config(curr: &Node) -> ConvTranspose3dConfig {
     // the channels are inverted in the weight tensor
     let channels: [usize; 2] = [weight_shape[1] * group, weight_shape[0]];
 
+    let kernel_size = if kernel_shape.is_empty() {
+        // https://onnx.ai/onnx/operators/onnx__Conv.html#attributes
```
laggui: Same

crates/onnx-ir/src/node/conv_transpose1d.rs
torsteingrindvik: Fixed

crates/onnx-ir/src/node/conv_transpose2d.rs
torsteingrindvik: Fixed

crates/onnx-ir/src/node/conv_transpose3d.rs
torsteingrindvik: Fixed

PR #3558, opened 2025-08-15
Add f16 and string to datatype to element type mapping in Cast node

torsteingrindvik:
Bug: Importing one of my models caused a panic due to a missing mapping in a conversion.

### Changes

Expand mapping from DataType to ElementValue in the Cast node.

### Testing

My model which uses f16 does not panic here any more.


nathanielsimard (approved):

torsteingrindvik:
> String and float16 were included #3381, which is causing merge conflicts. Feel free to close this PR or resolve the merge conflicts if you want to add the data type to the panic 🙂

No need- the conversion are in place so I'm happy. Thanks for the ping :+1:

PR #3563, opened 2025-08-17
Support dynamic shape and tensor sizes in ONNX resize

### LINKED ISSUE #1882: burn-import initializer tensor's not added to scope

antimora:
Adds support for ONNX Resize nodes where the output size is provided as a runtime shape or tensor, not just as a static constant. Updates the ONNX IR and codegen to handle dynamic 'sizes' and 'scales' inputs, adds new test models and test cases for these scenarios, and refactors the resize node handling to distinguish between static and runtime arguments.

### Checklist

### Related Issues/PRs

Fixes #1882

### Changes

- Refactors resize configuration to use structured types and support both static and runtime scales/sizes
- Updates codegen to handle runtime resize operations using tensor interpolation functions

### Testing

Adds comprehensive test coverage for resize with shape inputs and tensor inputs


crates/burn-import/src/burn/node/resize.rs
```diff
@@ -4,46 +4,145 @@ use burn::record::PrecisionSettings;
...
+        // Only create a field if we have static scales/sizes
+        // Runtime inputs mean we don't need a static interpolation field
+        let field = match (&scales, &sizes) {
+            (ResizeScales::Runtime(_), _) | (_, ResizeSizes::Runtime(_)) => {
+                None // Runtime inputs, no field needed
             }
-        } else if input.rank == 4 {
-            quote! {
-                Interpolate2d
+            (ResizeScales::Static(s), ResizeSizes::Static(z)) if !s.is_empty() || !z.is_empty() => {
```
laggui: Why not use `Option<..>` for both scales and sizes instead?

crates/burn-import/src/burn/node/resize.rs
```diff
@@ -4,46 +4,145 @@ use burn::record::PrecisionSettings;
...
+                    );
+                }
+            }
+            ResizeSizes::Runtime(Type::Tensor(tensor)) => {
+                let sizes_name = &tensor.name;
+                let mode_token = self.mode.to_tensor_interpolate_mode_token();
+                quote! {
+                    // Convert tensor to shape array
+                    let sizes_data = #sizes_name.to_data();
+                    let sizes_array = sizes_data.as_slice::<i64>().unwrap();
```
laggui: 1. Should we add warnings whenever we have a `to_data()` operation for now? This is not free if the model is loaded on GPU.
2. `as_slice::<i64>()` might fail if the backend used defaults to another dtype for int storage (e.g., `i32`). We should either use the associated backend type or explicitly convert.

antimora: Regarding "Should we add warnings whenever we have a to_data() operation for now? This is not free if the model is loaded on GPU.". Warning won't do anything because the caller can't do anything about it and we, the dev, are aware of the situation. We should fix it by introducing on device scalar and improving APIs

laggui: Yeah warnings are just to indicate a potential issue, like debug or info logs. Anyway, as you said we have an outstanding issue for that so it's ok.

laggui (approved):
Thanks for addressing my comments 🙏

PR #3572, opened 2025-08-18
Refactor burn-collective

Cielbird:
Refactoring in `burn-collective`:

* Collective operations are organized into their own seperate modules
* Better checks when calling collective operations
* Collective API now uses tensor primitives to avoid the `const D: usize`

Collective tests still pass

### DIFF
--- a/crates/burn-collective/src/local/all_reduce/tree.rs
+++ b/crates/burn-collective/src/local/all_reduce/tree.rs
@@ -2,7 +2,7 @@ use crate::PeerId;
 use burn_tensor::backend::{Backend, DeviceOps};
 use std::collections::HashMap;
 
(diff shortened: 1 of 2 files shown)


crates/burn-collective/src/local/all_reduce/tree.rs
```diff
@@ -0,0 +1,81 @@
+use crate::PeerId;
+use burn_tensor::backend::{Backend, DeviceOps};
+use std::collections::HashMap;
+
+/// Performs a all-reduce on the provided tensors in a b-tree structure with `arity`.
```
nathanielsimard: an all-reduce

nathanielsimard (commented):
LGTM

PR #3574, opened 2025-08-18
Add missing scalar conversion to to_burn

torsteingrindvik:
### Changes

`OnnxArgument` -> `TensorType` was missing some scalar match variants.

I assume these should be converted to rank 1 tensors.

### Testing

My model no longer panics regarding this match variant.

### DIFF
--- a/crates/burn-import/src/onnx/to_burn.rs
+++ b/crates/burn-import/src/onnx/to_burn.rs
@@ -1825,6 +1825,13 @@ impl From<&OnnxArgument> for TensorType {
                 rank,
                 ..
             }) => TensorType::new_bool(arg.name.clone(), *rank),
+            ArgType::Scalar(ElementType::Float16 | ElementType::Float32 | ElementType::Float64) => {
+                TensorType::new_float(arg.name.clone(), 1)
+            }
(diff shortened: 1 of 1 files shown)


crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -1825,6 +1825,13 @@ impl From<&OnnxArgument> for TensorType {
                 rank,
                 ..
             }) => TensorType::new_bool(arg.name.clone(), *rank),
+            ArgType::Scalar(ElementType::Float16 | ElementType::Float32 | ElementType::Float64) => {
+                TensorType::new_float(arg.name.clone(), 1)
+            }
+            ArgType::Scalar(ElementType::Int32 | ElementType::Int64) => {
+                TensorType::new_int(arg.name.clone(), 1)
+            }
+            ArgType::Scalar(ElementType::Bool) => TensorType::new_bool(arg.name.clone(), 1),
```
antimora, line 1834: Actually these should be converted to ScalarType

torsteingrindvik:
From discord: https://discordapp.com/channels/1038839012602941528/1407089099621273632/1407383868679655445

It seems this situation should likely not occur in the first place- it's not as straight forward as this PR.
Closing until more is known.

PR #3575, opened 2025-08-18
Implement int matmul

wingertge:
### Checklist

### Related Issues/PRs

CubeCL PR: https://github.com/tracel-ai/cubecl/pull/826

### Changes

Uses real matmul kernels for int tensors, instead of polyfilling

### Testing

Int matmul tests still pass with all numeric types


antimora:
Great. I guess we can now implement MatMulInteger ONNX operation to Burn

PR #3577, opened 2025-08-19
Unify transform usage of RngOptions.

crutcher:
### Checklist

### Related Issues/PRs

### Changes

### Testing


crutcher:
I also tried to clean up some for of `DatasetArgument<I>` which would work for:
 * `D: Dataset<I>`
 * `Arc<D> where D: Dataset<I>`
 * `Arc<dyn Dataset<I>>`

I couldn't figure it out; and decided to leave it out of this cleanup; but the wrapper types should do a better job of erasing the types of the things they are wrapping.

I'm leaning somewhat towards "wrappers take an `Arc<dyn Dataset<I>>`" and not trying to auto-build the `Arc`; it is in some ways a cleaner call-site implication.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -1,7 +1,42 @@
...
+///
+/// // Note: Stateful `From<&mut StdRng>` conversions:
+/// // via `&mut StdRng`, will also contain a snapshot of ``rng1``.
+/// let mut stateful = original_rng.clone();
+/// let with_rng3 : RngSource = (&mut stateful).into();
+/// // This advances the source rng;
+/// assert_ne!(stateful, original_rng);
+/// let created_rng: StdRng = with_rng3.into();
+/// assert_eq!(created_rng, stateful);
+/// ```
```
laggui: Although correct and thorough, this reads more like an integration test than an example for users. It tries to show every possible conversion path at the doc site.

I would keep this short to show the most idiomatic ways to use the type.

crates/burn-dataset/src/transform/options.rs
```diff
@@ -38,6 +73,21 @@ impl From<StdRng> for RngSource {
...
+}
+
+/// Users calling with a mutable rng expect state advancement,
+/// So conversion from `&mut StdRng` advances the rng before cloning.
+impl From<&mut StdRng> for RngSource {
+    fn from(rng: &mut StdRng) -> Self {
+        rng.next_u64();
+        Self::Rng(rng.clone())
+    }
+}
```
laggui, line 80: Conversion traits are useful but I think we still need to be careful about their usage. Sometimes it can implicitly pollute the code. These are typically cheap conversions without hidden behavior.

`From<&mut StdRng>` matches the intuition that `&mut` means “I’m allowing you to mutate”, but might have surprising behavior.

I would keep it explicit at the call site instead.

laggui (changes requested):

crates/burn-dataset/src/transform/options.rs
crutcher: @laggui I think this is the least-surprising thing to happen; and that the `&mut` *is* the explicit marker.

Consider situations like the a batch iterator, which re-shuffles (or re-creates) the internal dataset on every epoch restart.

Lacking state advancement, that's going to be the same shuffle every time, in a way that's going to confuse users.

Meanwhile, any call with `&mut rng` is assumed to have advanced state; by contract.

The most surprising thing to happen would be for `&mut rng` to *not* advance the state, in a way you could continue to depend on.

crates/burn-dataset/src/transform/options.rs
crutcher: I'm actually concerned that the module docs aren't long enough; this is just what I did in a hurry. This is an across the board issue; I think the rustdoc is really lacking.

I can try and flush out more discussion about when and how you'd use these.

crates/burn-dataset/src/transform/options.rs
laggui: > the `&mut` *is* the explicit marker.

Ok actually you're right here, if for some reason one wants to avoid advancing the state they can simply omit the deref mut. So it is rather explicit for all conversions. That makes sense.

crates/burn-dataset/src/transform/options.rs
laggui: Perhaps my comment was malformed. I simply meant that we should still strive for clear and concise.

What do you think of something like:
```rust
/// Defines a source for a `StdRng`.
///
/// # Examples
///
/// ```rust
/// use rand::rngs::StdRng;
/// use rand::SeedableRng;
/// use burn_dataset::transform::RngSource;
///
/// // Default via `StdRng::from_os_rng()` (`RngSource::Default`)
/// let system: RngSource = RngSource::default();
///
/// // From a fixed seed (`RngSource::Seed`)
/// let seeded: RngSource = 42.into();
///
/// // From an existing rng (`RngSource::Rng`)
/// let rng = StdRng::seed_from_u64(123);
/// let with_rng: RngSource = rng.into();
///
/// // From a snapshot of the current state (`RngSource::Rng`)
/// let rng = StdRng::seed_from_u64(123);
/// let snapshot: RngSource = (&rng).into();
///
/// // Advances the original RNG and then clones its new state
/// let mut rng = StdRng::seed_from_u64(123);
/// let stateful: RngSource = (&mut rng).into();
/// ```
```

crutcher: Done

PR #3582, opened 2025-08-19
Make Distribution::Default the Default::default().

crutcher:
### Checklist

- [na] Made sure the book is up to date with changes in this PR.

### Changes

`Distribution::Default` should be `Default`, or should be renamed.

### DIFF
--- a/crates/burn-tensor/src/tensor/distribution.rs
+++ b/crates/burn-tensor/src/tensor/distribution.rs
@@ -3,9 +3,10 @@ use rand::{Rng, RngCore, distr::StandardUniform};
 use crate::{Element, ElementConversion};
 
 /// Distribution for random value of a tensor.
-#[derive(Debug, Clone, Copy, PartialEq, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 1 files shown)


antimora (approved):
LGTM

crates/burn-tensor/src/tensor/distribution.rs
```diff
@@ -107,3 +108,16 @@ impl Distribution {
...
+    use super::*;
+
+    #[test]
+    fn test_distribution_default() {
+        let dist: Distribution = Default::default();
+
+        assert_eq!(dist, Distribution::Default);
+        assert_eq!(Distribution::default(), Distribution::Default);
+    }
+}
```
antimora, line 123: Probably this test is not needed.

crutcher: I see the point of unittests not as verifying behavior; but as hardening a codebase against automated refactoring breakage. There should be something in the tests that exercise those 3 interfaces; it's easy to add, and it potentially prevents breakages under rewrites later.

PR #3587, opened 2025-08-20
Perf/linear layout

nathanielsimard:
Allows performing linear layer in ColMajor.

Also fixes fusion and reshape optimization.

Also updated on the new main cubecl commit.


crates/burn-core/src/record/primitive.rs
```diff
@@ -155,7 +155,9 @@ where
     type Item<S: PrecisionSettings> = ParamSerde<FloatTensorSerde<S>>;
 
     fn into_item<S: PrecisionSettings>(self) -> Self::Item<S> {
-        let (id, tensor) = self.consume();
+        let (id, tensor, mapper) = self.consume();
+        println!("On Save {mapper:?}");
```
laggui: Left-over print

crates/burn-core/src/nn/rnn/gate_controller.rs
```diff
@@ -35,13 +35,15 @@ impl<B: Backend> GateController<B> {
                 d_output,
                 bias,
                 initializer: initializer.clone(),
+                layout: LinearLayout::Row,
```
laggui, line 38: Explicitly specified for performance I assume?

crates/burn-core/src/nn/swiglu.rs
```diff
@@ -64,10 +64,12 @@ impl SwiGluConfig {
         SwiGlu {
             linear_inner: LinearConfig::new(self.d_input, self.d_output)
                 .with_bias(self.bias)
+                .with_layout(LinearLayout::Col)
```
laggui: Deviates from the default row layout, was this intentional for all backends/implementations?

crates/burn-core/src/nn/rnn/gate_controller.rs
nathanielsimard: No for backward compatibility

crates/burn-core/src/nn/swiglu.rs
nathanielsimard: Yup

antimora:
is there any implication for `crates/burn-import/src/common/adapter.rs`'s `PyTorchAdapter`? Currently we transpose linear weights.

nathanielsimard:
> is there any implication for `crates/burn-import/src/common/adapter.rs`'s `PyTorchAdapter`? Currently we transpose linear weights.

Not for now since the setting is backward compatible, but we could simply remove the code that transposes the weights and generate a linear layer with the `LinearLayer::Col` config.

antimora:
> > is there any implication for `crates/burn-import/src/common/adapter.rs`'s `PyTorchAdapter`? Currently we transpose linear weights.
> 
> Not for now since the setting is backward compatible, but we could simply remove the code that transposes the weights and generate a linear layer with the `LinearLayer::Col` config.

Yeah, lets do it because it will give an extra boost for free.

PR #3593, opened 2025-08-21
fix: Use correct datatype when filling boolean tensors

wingertge:
### Checklist

`bool_zeros` and `bool_ones` incorrectly used the int element instead of the bool element.


antimora (approved):
LGTM

PR #3595, opened 2025-08-21
Add a `Tensor::<B, D, K>::unit(fill_value, device)` op for `[1; D]` tensors.

crutcher:
### Checklist

### Changes

Unit tensors come up in some algorithms; and they're trivial to construct on top of `.full()`.

Also moved the `full()/full_like()` docs.

### Testing

tests.

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -3,7 +3,7 @@
 As previously explained in the [model section](../basic-workflow/model.md), the Tensor struct has 3
 generic arguments: the backend B, the dimensionality D, and the data type.
 
-```rust , ignore
+```rust, ignore
 Tensor<B, D>           // Float tensor (default)
 Tensor<B, D, Float>    // Explicit float tensor
 Tensor<B, D, Int>      // Int tensor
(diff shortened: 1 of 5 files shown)


antimora (approved):
LGTM

crutcher:
@laggui I feel like this is a perspective difference that we've bumped into before.

As a user of these APIs, I don't see these methods as clutter. Sure, there may be a one-liner for a lot of operations, but that's assuming that the readers and writers consistently remember, apply, and interpret the one-liner.

I'd personally rather see:
```
let unit = Tensor::unit(0.5, device);
```
than:
```
let unit = Tensor::full([1; D], 0.5, device);
```

In the same way we have `eye` and related constructors.

Having the "blessed right way" simplifies the readability of callers, and opens the door to optimizations that would never be a safe one-liner. Provided that those operations are findable on the interface surfaces.

I wish it were easier to build captured namespaces in rust; `Tensor::util::unit(...)`

Is there a place where you *would* be happy hanging this sort of thing?

Additionally, in a world with code-assistants, utility unification (with examples) makes these completion tools and bots more powerful.

antimora:
> I wish it were easier to build captured namespaces in rust; `Tensor::util::unit(...)`
> 
> Is there a place where you _would_ be happy hanging this sort of thing?

We can use a trait and import it if we need , like `TensorHelper` type of trait?

Regarding the name. To me unit sounded like like it's related to a unit vector. So it might confuse some. Maybe use something like `single` or `uno`? Just some ideas.

antimora:
A couple of more shorts names in the hat:

* Tensor::mono(0.5, device)
* Tensor::solo(0.5, device)
* Tensor::point(0.5, device)
* Tensor::lonely(0.5, device) 🙂 
* Tensor::onesie(0.5, device)

Previously:
* Tensor::unit(0.5, device)
* Tensor::single(0.5, device)
* Tensor::uno(0.5, device)
* Tensor::singleton(0.5, device)

crutcher:
Maybe `scaled_unit`? as its a fast way to get `.ones([1; D]) * value`?
Maybe also have `.unit() => .ones([1;D])`?

antimora:
@laggui , @laggui are we still interested in pursuing this PR?

antimora:
Closing this PR for now; please reopen if needed.

Issue #3412, opened 2025-07-23
ONNX models with import or runtime issues in burn-import


antimora:
# Tracking: Models That Fail to Import or Run in Burn

This issue tracks ONNX models that cannot currently be built or run using `burn-import` or the Burn framework. If you encounter a model that fails to import or execute, please comment below or submit a new issue and reference this tracker.

## Checklist of Models with Known Issues

- [ ] YOLOv8n ([#2822](https://github.com/tracel-ai/burn/issues/2822))
- [ ] ModernBERT-base ([#3130](https://github.com/tracel-ai/burn/issues/3130))
- [ ] RTMW3D-x ([#2930](https://github.com/tracel-ai/burn/issues/2930), [models#74](https://github.com/tracel-ai/models/issues/74))
- [ ] Depth-Anything-v2 ([#2592](https://github.com/tracel-ai/burn/issues/2592), [#2926](https://github.com/tracel-ai/burn/issues/2926))
- [ ] XFeat ([#3110](https://github.com/tracel-ai/burn/issues/3110))
- [ ] Silero VAD ([#1941](https://github.com/tracel-ai/burn/issues/1941))
- [ ] Face Detector (MediaPipe) ([#1370](https://github.com/tracel-ai/burn/issues/1370))
- [ ] IBM Granite 4.0 Tiny Preview ([models#71](https://github.com/tracel-ai/models/issues/71))

## How to Use
- Add a comment if you find a new failing model, or if your model is fixed by a PR.
- Check off models as they become supported or fixed.

---

**For operator-level or feature gaps, please also check:**
- [SUPPORTED-ONNX-OPS.md](https://github.com/tracel-ai/burn/blob/main/crates/burn-import/SUPPORTED-ONNX-OPS.md) for coverage
- [burn-import issues](https://github.com/tracel-ai/burn/issues)

antimora:
Quick update I am able to convert `yolo11x_opset16.onnx` using this WIP PR: #3381

antimora:
RTMW3D-x is buildable with https://github.com/tracel-ai/burn/pull/3564 (still under review)

n1ght-hunter:
can we add kokoro to this. as its a very popular tts model. currently has an issue with expand1 - rank which is an already known issue.
it also uses albert internally so will probably require albert model to be working for this to work
https://github.com/thewh1teagle/kokoro-onnx/releases/download/model-files-v1.0/kokoro-v1.0.onnx
there is also the bin file for voices
[...]

PR #3599, opened 2025-08-22
ONNX model checking

antimora:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/3412

### Changes

### Testing

Torch backend passing: 

```
    Finished `release` profile [optimized] target(s) in 1m 53s
     Running `target/release/burn-import-model-checks-yolo11x`
========================================
YOLO11x Burn Model Test
========================================

Initializing YOLO11x model...
  Model initialized in 50.51ms

Loading test data from artifacts/test_data.pt...
  Data loaded in 2.49ms
  Loaded input tensor with shape: [1, 3, 640, 640]
  Loaded reference output with shape: [1, 84, 8400]

Running model inference with test input...
  Inference completed in 241.12ms

  Model output shape: [1, 84, 8400]
  ✓ Output shape matches expected: [1, 84, 8400]

Comparing model output with reference data...
  ✓ Model output matches reference data within tolerance (1e-4)!

========================================
Model test completed!
========================================
cargo run --release --features tch  2.13s user 1.86s system 3% cpu 1:53.96 total
[yolo11x]%
```


crates/burn-import/model-checks/README.md
```diff
@@ -0,0 +1,75 @@
...
+- Compile the generated code
+
+## Adding New Model Checks
+
+To add a new model for testing:
+
+1. Create a new directory under `model-checks/`
+2. Add a script to download/prepare the ONNX model
+3. Create a `build.rs` that uses `burn_import::onnx::ModelGen`
+4. Write test code in `src/main.rs`
```
laggui: This is pretty much the first `Structure` section right?

crates/burn-import/model-checks/README.md
```diff
@@ -0,0 +1,75 @@
...
+5. Ensure the model directory is excluded from the workspace in the root `Cargo.toml`
```
laggui: That already seems to be taken care of in the workspace exclude `"crates/burn-import/model-checks/*"`

crates/burn-import/model-checks/yolo11x/Cargo.lock
laggui, line 1: I don't think we should track the lock for every model, especially since these are tests.

antimora:
> Just a couple of comments.
> 
> Future note: We might eventually be able to generalize the tests to load different models, so each model just needs to provide the ONNX file and test data, although the test data shapes will differ so we would have to define those more flexibly 🤔 Just thinking ahead before we accumulate a lot of code just to test different models.

Spoke like a true engineer 😄

PR #3601, opened 2025-08-22
Derive clone for Record Items

amfaber:
# Add `#[derive(Clone)]` for Record Items when deriving Record
Hi! This small change allows all automatically derived RecordItems to implement clone

## Motivation
The clone implementation is crucial if you'd like to hold on to an in-memory, off-device representation of the model that can be put on device at any time.
The requirement of clone comes from the interface of a module being created from a Record, as well as a Record being created from a Record Item, taking their arguments by value. It is thus not possible to put the model on device and hold on to its state on CPU at the same time.
With this change, the RecordItem can be held onto and simply cloned when model instantiation is required.

Now, you might be thinking, why would anyone want this? Why hold onto the CPU-side data if the model as already been put on device? Well, we are building a GUI where users will be able to, among other things, train smaller neural networks through a UI. To this end and to conform with the rest of the GUI architecture, it would be very convenient for us to store the result of training a model in memory until the user is satisfied with its performance, at which point they can save it to disk. It should also be possible to have multiple of such models loaded at the same time, and these representations should derive Serde, such that we can save the models to disk when the user requests it.

The autogenerated RecordItem types seem perfect for this, as they just hold onto the data on CPU side at some precision, and their raison d'etre is to implement Serialize and Deserialize. However, it seems I cannot recover my module from one of these without consuming it, as described above. A clone implementation would allow me to simply clone it before putting it on device. A bit of a waste of resources, but nothing too bad and nothing in comparison to training / inference.

## Other things I considered
- I could serialize to a slice of bytes using the provided Recorder or my own serde implementation. This seems even more wasteful and less elegant than the clone impl
- I could serialize a temporary file. Upside is that it would free up ram, downside is that I need to go to disk to recover the model again (probably fine in this case), but now I also need to worry about whether these temporary files get cleaned up properly, and whether someone else cleans them up while I somehow have a reference to them. Not great.
- I considered making Clone be a required trait for the associated RecordItem type. However as its not actually necessary for normal operation, that doesn't make a whole lot of sense. This way, all the automatically derived versions are clone if you ever end up needing it, while if you implement Record manually you can do that without requiring the RecordItem to be Clone. Again, if you actually need it then you can also manually implement clone for your custom RecordItem type.

Thanks for all the great work!

### DIFF
--- a/crates/burn-core/src/record/primitive.rs
+++ b/crates/burn-core/src/record/primitive.rs
@@ -258,6 +258,7 @@ primitive!(i8);
 /// deserializing arrays of variable size,
 /// see [serde/issues/1937](https://github.com/serde-rs/serde/issues/1937)
 /// for backward compatibility reasons. Serde APIs were created before const generics.
+#[derive(Clone)]
 pub struct Array<const N: usize, T>([T; N]);
 
 impl<T: Serialize, const N: usize> Serialize for Array<N, T> {
--- a/crates/burn-core/tests/test_derive_module.rs
+++ b/crates/burn-core/tests/test_derive_module.rs
@@ -83,6 +83,37 @@ impl<B: Backend> ModuleComposed<B> {
     }
 }
 
+#[allow(dead_code)]
+mod compiletime_clone_impl_check {
+    use burn_core::{
+        module::{Module, ModuleDisplay},
+        prelude::Backend,
+        record::{PrecisionSettings, Record},
+    };
+
+    use super::*;
+    
+    type RecordItem<M, B, S> = <<M as Module<B>>::Record as Record<B>>::Item<S>;
+
+    fn implements_clone<T: Clone>() {}
+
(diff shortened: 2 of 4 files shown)


amfaber:
I hadn't realized the possibility of generic modules when using #[derive(Module)]. That complicates the implementation somewhat by requiring custom `impl Clones` with proper where clauses instead of just delegating to `#[derive(Clone)]`, but it seems like everything passes now. I added some tests to check at compiletime that the types we expect to implement Clone from the macro do so

nathanielsimard:
The reason why `Clone` wasn't implemented is because the data isn't behind a reference count, so it can be really expensive to clone. Since it's pretty much a hidden type, I don't think most people will consider cloning it, so maybe adding the implementation isn't too harmful.

nathanielsimard (approved):
LGTM

PR #3604, opened 2025-08-22
fix: Ensure output layout is the same for SIMD ops in ndarray

wingertge:
Ndarray was not taking the input strides into account in element-wise ops when creating a new output for tensors that can't be mutated. It now ensures the output is created with the same layout as the input, to ensure element positions match.

### Checklist


antimora:
@crutcher confirms the fix works. I will check as well once I am at my computer

antimora:
It fixed ndarray reported here: https://github.com/tracel-ai/burn-onnx/issues/51


image

PR #3616, opened 2025-08-26
Feature: Remove generic dependency on backends from RecordItems

amfaber:
## Remove generic dependency on backends from RecordItems
Motivated by the same usecase as I described in https://github.com/tracel-ai/burn/pull/3601, and also includes the changes from there. 

This PR is still a draft and somewhat messy - my intention is to gauge your interest in whether you would even like this change and only continue with it if you do.

As I want to store these RecordItems on my app struct, I became annoyed with their generic dependency on a <B: Backend>. This dependency seems to be entirely artificial and solely a byproduct of needing a way to name the appropriate associated types in the derive macros that produce the RecordItems.

In theory, different implementations with different backends could define different RecordItems, meaning that there would be an actual need to know what backend we're talking about when describing a RecordItem. However in practice, this would mean that the serialization / deserialization from these two different implementations would also be incompatible, which I would think would be an antipattern in burn - it would mean that a model serialized from one backend wouldn't be able to be loaded on another backend.

In practice, the RecordItems seem to always be the same regardless of backend, making models deserializable to any backend.

This means that the only motivation for the <B: Backend> generic is to name the nested RecordItems when deriving Module or Record. I propose that this can instead be done by referencing them through a single "implementor" of Backend.
DummyBackend is a typesystem trick to always provide a single backend implementor that we can always use to refer to the required associated items in the derive macro.

Let me know what you think about the core idea. I am quite happy that I can represent a cpu-side version of my model without completely type-erasing it and without infecting my structs with needless <B: Backend> generics.

I think it also enforces something that wasn't explicitly enforced before - the RecordItem /shouldn't/ need to know which backend it belongs to, as that means that it potentially can't be loaded onto other backends.

I also understand if this is too niche and you don't think it fits the project - in that case I'll probably move the DummyBackend to my "userspace" and simply transmute the RecordItem I get from burn on some backend to a RecordItem<DummyBackend> instead and just store that. I'd rather avoid that piece of unsafety if I can though

### Testing

Cargo check fails on some generated files from burn-onnx - I'll look into these and run the full test suite once I know the burn project is interested in these changes

### DIFF
--- a/crates/burn-core/src/record/primitive.rs
+++ b/crates/burn-core/src/record/primitive.rs
@@ -258,6 +258,7 @@ primitive!(i8);
 /// deserializing arrays of variable size,
 /// see [serde/issues/1937](https://github.com/serde-rs/serde/issues/1937)
 /// for backward compatibility reasons. Serde APIs were created before const generics.
+#[derive(Clone)]
 pub struct Array<const N: usize, T>([T; N]);
 
 impl<T: Serialize, const N: usize> Serialize for Array<N, T> {
--- a/crates/burn-core/tests/test_derive_module.rs
+++ b/crates/burn-core/tests/test_derive_module.rs
@@ -83,6 +83,37 @@ impl<B: Backend> ModuleComposed<B> {
     }
 }
 
+#[allow(dead_code)]
+mod compiletime_clone_impl_check {
+    use burn_core::{
+        module::{Module, ModuleDisplay},
+        prelude::Backend,
+        record::{PrecisionSettings, Record},
+    };
+
+    use super::*;
+    
+    type RecordItem<M, B, S> = <<M as Module<B>>::Record as Record<B>>::Item<S>;
+
+    fn implements_clone<T: Clone>() {}
+
+    fn basic_implements_clone<B: Backend, S: PrecisionSettings>() {
+        implements_clone::<RecordItem<ModuleBasic<B>, B, S>>();
+        implements_clone::<RecordItem<ModuleComposed<B>, B, S>>();
+    }
+
+    fn generic_implements_clone<B, S, M>()
+    where
+        B: Backend,
+        S: PrecisionSettings,
+        M: Module<B> + ModuleDisplay,
+        RecordItem<M, B, S>: Clone,
+    {
+        implements_clone::<RecordItem<ModuleWithGenericModule<B, M>, B, S>>();
+        implements_clone::<RecordItem<ModuleEnumWithGenericModule<B, M>, B, S>>();
+    }
+}
+
 mod state {
     use super::*;
 
--- a/crates/burn-core/tests/test_derive_module.rs
+++ b/crates/burn-core/tests/test_derive_module.rs
@@ -92,7 +92,7 @@ mod compiletime_clone_impl_check {
     };
 
     use super::*;
-    
+
     type RecordItem<M, B, S> = <<M as Module<B>>::Record as Record<B>>::Item<S>;
 
     fn implements_clone<T: Clone>() {}
--- a/crates/burn-derive/Cargo.toml
+++ b/crates/burn-derive/Cargo.toml
@@ -19,5 +19,5 @@ proc-macro = true
 [dependencies]
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
-syn = { workspace = true }
+syn = { workspace = true, features = ["visit-mut"] }
 derive-new = { workspace = true }
--- a/crates/burn-derive/src/record/codegen.rs
+++ b/crates/burn-derive/src/record/codegen.rs
@@ -1,6 +1,6 @@
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
-use syn::{Generics, parse_quote};
+use syn::{GenericParam, Generics, PredicateType, Type, WherePredicate, parse_quote};
 
 use crate::record::item::codegen::RecordItemCodegen;
 
@@ -86,15 +86,15 @@ impl<G: RecordItemCodegen> RecordCodegen<G> {
     /// Get the generics attached to the record item type.
     fn record_item_generics(&self) -> Generics {
         let param: syn::Generics = parse_quote! { <S: burn::record::PrecisionSettings >};
-        let mut generics = self.ty.generics.clone();
+        let mut generics = strip_backend_from_generics(&self.ty.generics);
         for param in param.params.into_iter() {
             generics.params.push(param);
         }
 
-        if !self.ty.has_backend {
-            let param: syn::TypeParam = parse_quote! { B: burn::tensor::backend::Backend };
-            generics.params.push(syn::GenericParam::Type(param));
-        }
+        // if !self.ty.has_backend {
+        //     let param: syn::TypeParam = parse_quote! { B: burn::tensor::backend::Backend };
+        //     generics.params.push(syn::GenericParam::Type(param));
+        // }
 
         generics
     }
@@ -137,4 +137,38 @@ impl RecordType {
             has_backend,
         }
     }
+}
+
+pub fn strip_backend_from_generics(orig: &Generics) -> Generics {
+    let mut g = orig.clone();
+
+    g.params = g
+        .params
+        .into_iter()
+        .filter(|p| !matches!(p, GenericParam::Type(tp) if tp.ident == "B"))
+        .collect();
+
+    if let Some(wc) = &mut g.where_clause {
+        wc.predicates = wc
+            .predicates
+            .clone()
+            .into_iter()
+            .filter(|pred| {
+                // Drop predicates where the LHS is exactly `B`
+                match pred {
+                    WherePredicate::Type(PredicateType { bounded_ty, .. }) => {
+                        if let Type::Path(tp) = bounded_ty {
+                            return !(tp.qself.is_none()
+                                && tp.path.segments.len() == 1
+                                && tp.path.segments[0].ident == "B");
+                        }
+                        true
+                    }
+                    _ => true,
+                }
+            })
+            .collect();
+    }
+
+    g
 }
--- a/crates/burn-derive/src/record/item/codegen.rs
+++ b/crates/burn-derive/src/record/item/codegen.rs
@@ -1,5 +1,8 @@
 use proc_macro2::{Ident, TokenStream};
-use syn::Generics;
+use syn::{
+    GenericArgument, Generics, TypePath,
+    visit_mut::{self, VisitMut},
+};
 
 /// Basic trait to be implemented for record generation.
 pub(crate) trait RecordItemCodegen {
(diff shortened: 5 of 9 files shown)


amfaber:
A better name for DummyBackend could be NoBackend or similar

amfaber:
I don't think moving the record type from being a generic on Record to an associated type will make any difference. We need to keep in mind what types will implement Module and Record - in both cases this includes Param and Tensor, both of which are and should be generic over B, as their data lives in the backend. It is only when we get to the Record::Item, which bottom out at types like FloatTensorSerde that we can be sure that we are completely free from any backend. At this stage the data is guaranteed to live on the CPU and implement serde.

Another way to think about this is that there should be a 1-to-1 relationship between the Record::Item's and the representation on disk, since it is the Record::Item that implements serde. Thus, if the Record::Item needs to know what backend it belongs to to function, you will not be able to serialize a model from one backend and load it with another. I think it is a goal of burn to handle that usecase though.
Just as if the serialization had been taken all the way to disk / a buffer, it should be able to encode the module into a Record::Item and decode it into a Record / Module on another backend.
This is not the case for the types that implement Record, as their data actually lives on the device.

The DummyBackend / NoBackend implementor is merely something that can stand in for any backend and be available regardless of feature set, since the serialization needs to be backend-agnostic regardless. It is solely used to be able to name the appropriate types in the derive macro - if we can come up with some other way of naming these types without the DummyBackend, that would be preferable. I don't have any ideas though

Thanks for taking a look! Let me know if I have any misunderstandings or if I am missing something

amfaber:
I think what I'm arguing for can probably be explained better using examples. Say we have a custom module type for which we
#[derive(Module)]. Just looking at the types, the proc macro will generate something like this for us on main today

```rust
#[derive(Module)]
struct MyModule<B: Backend> {
    some_field: Conv2d<B>,
}

struct MyModuleRecord<B: Backend> {
    some_field: <Conv2d<B> as Module<B>>::Record,
}

struct MyModuleRecordItem<B: Backend, S: PrecisionSettings> {
    some_field: <<Conv2d<B> as Module<B>>::Record as Record<B>>::Item<S>,
}
```

If we want to generate RecordItems that are not generic over B, we could try to refactor the Module and Record traits to use associated types instead of generics. Something like

```rust
trait ModuleV2{
    type Backend;
    type Record: RecordV2<Backend = Self::Backend>;

    // ...
}

trait RecordV2{
    type Backend;
    type Item<S: PrecisionSettings>;

    // ...
}
```

Lets imagine what our proc macro would generate if we have these trait definitions and want to generate a RecordItem that is not generic over B. We would get something like


```rust
#[derive(ModuleV2)]
struct MyModule<B: Backend>{
    some_field: Conv2d<B>,
}

struct MyModuleRecord<B: Backend>{
    some_field: <Conv2d<B> as ModuleV2>::Record,
}

struct MyModuleRecordItemV2<S: PrecisionSettings>{
    some_field: <<Conv2d<B> as ModuleV2>::Record as RecordV2>::Item<S>,
    //                   ^ Cannot name this type without using a concrete backend or
    //                     having the RecordItem be generic over B.
    //                     I argue that the RecordItem being generic over B is semantically wrong.
}
```

The issue is that, because our Module types are (and, I assume, should be) generic over B, we still cannot refer to their Records and thus RecordItems without ourselves being generic over B or using a concrete type that implements Backend as a standin for any type that implements Backend

laggui:
> Thus, if the Record::Item needs to know what backend it belongs to to function, you will not be able to serialize a model from one backend and load it with another. I think it is a goal of burn to handle that usecase though.

Ahhh thanks for the detailed response! I definitely misunderstood what you meant 😅

As you correctly pointed out, the distinction between a `Record` and a `Record::Item` is crucial. But let me clarify some details:

- `Record` is intentionally backend-specific, as the types can be intrinsically linked to a specific backend's data format. That's why we have `from_data` and `into_data` for tensors. And this is at the root of the record implementations for module parameters.
- `Record::Item` represents a backend-agnostic data structure that is serializable. This type can be moved between different backends and thus is not generic over `B`.

When record items have to be loaded or saved for a module, the `Record<B>` is essentially the bridge. The backend generic is not an artificial dependency, but required context that makes this conversion safe and possible.

So, while very clever, I think the `DummyBackend` approach essentially tries to circumvent this relationship / one-to-one mapping between a specific module and its corresponding record.

I think what you are trying to achieve can be accomplished using a `Recorder` instead 🙂 If the goal is to store the data in memory, why not use an in-memory recorder? For example, you could use `BytesRecorder` to keep the records as bytes, or implement custom in-memory recorder that stores the serialized representation in the format that you would like to keep. The recorder output can be as close to the serialized state (or even the exact same).

amfaber:
I am afraid there might still be a misunderstanding - the PR does not try to remove the dependency on <B: Backend> from Records, only from RecordItems. As you agreed, these are and should be backend-agnostic, yet the #[derive(Record)] macro today emits code that makes the RecordItems generic over <B: Backend>. Focusing on just the structs involved, the generated code today is something like

```rust
#[derive(Module)]
struct MyModule<B: Backend> {
    some_field: Conv2d<B>,
}

// Generated
struct MyModuleRecord<B: Backend> {
    some_field: <Conv2d<B> as Module<B>>::Record,
}

// Generated
struct MyModuleRecordItem<B: Backend, S: PrecisionSettings> {
    some_field: <<Conv2d<B> as Module<B>>::Record as Record<B>>::Item<S>,
} 
```

As you say, MyModuleRecord<B: Backend> is generic over the backend for good reason - the data is actually on device with this representation

MyModuleRecordItem<B: Backend, S: PrecisionSettings> on the other hand, is generic over B solely to be able to name the nested RecordItems of the fields within the macro. It is this dependency that is artificial, not the MyModuleRecord's.
However DummyBackend allows us to name the relevant inner Items without being generic over the Backend. I.e. I propose

```rust
#[derive(Module)]
struct MyModule<B: Backend> {
    some_field: Conv2d<B>,
}

// Generated
struct MyModuleRecord<B: Backend> {
    some_field: <Conv2d<B> as Module<B>>::Record,
}

// Generated
struct MyModuleRecordItem<S: PrecisionSettings> {
    some_field: <<Conv2d<DummyBackend> as Module<DummyBackend>>::Record as Record<DummyBackend>>::Item<S>,
} 
```

Crucially, MyModuleRecord<B: Backend> is still generic, and so is all of the method implementations. The /only/ place we use the DummyBackend trick is for defining the MyModuleRecordItem struct.

Now, because we only use the trick for the struct definition of MyModuleRecordItem and nowhere in the method implementations, all the parts of the system that actually interact with the Backend will still be parameterized by it - it is only the artificial dependency we've been able to manage to remove from the generated RecordItem

As to your comment about the BytesRecorder - you're completely right that I could use this to solve my problem as well - though it feels a bit icky. I could be getting out a struct that implements ser/de and simply save that in my app struct, which would make it be serialized just the same as all of my other application data (in this case using the cbor format). Using the bytes recorder, I introduce another decoding step every time I'd need to use it (minimal performance hit, but still icky), and I'd be embedding whatever format the BytesRecorder uses (msgpack as far as I recall) in my serialized app state. Its not like its human readable to begin with, so not too much of a loss either. Still, its icky and an artificial limitation.

Another solution would also be to implement the DummyBackend outside of Burn and simply std::mem::transmute'ing the generated RecordItem<B> to RecordItem<DummyBackend>, since RecordItems will always have the same representation regardless of backend, thus freeing it from the generic. This also feels icky, as I am forced to `unsafe`ly telling the compiler that these representations are identical - it'd be better if burn itself expressed this through the type system

laggui:
> I could be getting out a struct that implements ser/de and simply save that in my app struct, which would make it be serialized just the same as all of my other application data (in this case using the cbor format)

The bytes recorder was just an example implementation since we currently provide it, but even if you have a custom in-memory recorder that simply deals with the model record item, the crux of the matter is still that the item will be generic over the backend. And the generic will unnecessarily propagate to your code 🫤 so that is definitely something I overlooked earlier. Sorry for the confusion.

> I am afraid there might still be a misunderstanding - the PR does not try to remove the dependency on <B: Backend> from Records, only from RecordItems.

Yeah I understood that, no worries 😄 I also think I failed to address the explicit concern in my response: the `DummyBackend` is not a simple place-holder type. It would have to implement the `Backend` trait due to the `Module<B: Backend>` bound. This is a core trait, but it has a lot of requirements for tensor ops, primitives, device trait, etc. That's not a simple/elegant trick anymore.. it is somewhat of an implementation burden.

I also want to clarify that I am not dismissing the issue, this might be a limitation of the current module trait. We've been discussing some pain points and possible directions on discord a lot recently, especially more dynamic modules. The discussion stemmed from another origin, but the associated module record was a blocker. I think we need to have a better separation of concerns for modules and their state. So the solution might not be as simple as a `DummyBackend` (which really feels more of a circumvention for the current pattern / limitations 😅). Unless you had something else in mind?

> Using the bytes recorder, I introduce another decoding step every time I'd need to use it (minimal performance hit, but still icky), and I'd be embedding whatever format the BytesRecorder uses (msgpack as far as I recall) in my serialized app state

For serialization along with your app state, recorders can be extended to implement in the format you would like. So you could technically have a cbor recorder. Perhaps that could bridge the gap given the current limitation? I am trying to think of other solutions that could work for the current state. `Box`ing would be a pain since you will have to downcast back to a concrete type when loading the record back (e.g., to initialize a new model from the state), but if you change backends it could fail.

Btw, I do want to say thanks for engaging in these constructive discussions 🙏 the current roadmap for the next release(s) is to improve the user experience, and this is a very concrete example.

amfaber:
I also want to thank you for engaging in the discussion - the repo is very active and you guys have done incredible things on a very limited timescale. I really want to see the project succeed, so thank you :)

I'll just summarize where I see us standing now. These are statements that I would think we both agree to. Please correct me if I am wrong
- We agree that derived RecordItem items are artificially generic over backends due to how they are generated by the proc macro
- It'd be nice if that wasn't the case
- One solution to this is a placeholder struct implementing Backend, as I propose in this PR. However this trick isn't very pretty
- Another solution could be to rework the Module / Record system to allow for more dynamic records

Now, here's my opinions:
I am willing to bet it will take a fair amount of time to properly rework the Module / Record system. In the meantime, a placeholder Backend fixes the problem now right now - even though it is ugly, which I completely agree with, it _works_. For me it really comes down to a tradeoff

- Derived RecordItems being generic over Backends is a wart that is exposed to users (though realistically most won't care)
- Maintaining a placeholder struct that derives backend will be an internal wart - a slight maintenance burden, but something that is also internal to the library.

So it really comes down to whether one thinks this tradeoff is worth it, which is also why I started this PR as a draft to gauge interest. Personally, I think the maintenance cost is very minor - I think the placeholder struct should probably be called `NoBackend` and every required method for the various traits should just be `unimplemented!()`. Since burn already maintains a number of backends, this would only add a small burden of adding some boilerplate `unimplemented!()` methods when the traits change.

And who knows, other situations could arise where its useful to have a concrete Backend in other cases where the behavior from any Backend would be identical.

As for the BytesRecorder, even if I have it produce cbor into a Vec<u8> that I then put in my app state with a serde_bytes directive, I'd still effectively introduce a nested binary blob inside my serialized state. Totally doable, but it feels wrong to have to do it simply because a proc macro can't name the correct types without taking on this generic.

laggui:
> I'll just summarize where I see us standing now. These are statements that I would think we both agree to. 

Yeah I think we have the same stance on the problem. I was/am just unsure of the best resolution path (given a reasonable trade-off, which might be time sensitive in your case) 😅 so I was also trying to find possible alternatives for your use case.

> Derived RecordItems being generic over Backends is a wart that is exposed to users (though realistically most won't care)

Agreed! And as you stated, most won't care (or at least, none have cared until now 😄 ), but that doesn't mean it is correct.

>  I think the placeholder struct should probably be called `NoBackend` and every required method for the various traits should just be unimplemented!(). Since burn already maintains a number of backends, this would only add a small burden of adding some boilerplate unimplemented!() methods when the traits change.

Maintenance should be minimal as you stated, it really comes down to the amount of boilerpate required to "patch" this until it is actually addressed. All tensor/module/activation ops, a device type and tensor primitive type 🥲 (as we can see in your draft implementation)

> As for the BytesRecorder, even if I have it produce cbor into a Vec that I then put in my app state with a serde_bytes directive, I'd still effectively introduce a nested binary blob inside my serialized state. Totally doable, but it feels wrong to have to do it simply because a proc macro can't name the correct types without taking on this generic.

Hahah yeah I agree. I was trying to see what other workarounds could work for your problem.

antimora:
I will close for now since it's stale. Also I believe we will drop RecordItems in favor of something like burn-store which uses a visitor patter to serialize/deserialize tensor data.

PR #3620, opened 2025-08-26
Tensor::flatten() => AsIndex dims support.

crutcher:
### Checklist

### Changes

Tensor::flatten() => AsIndex dims support.
tests.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -546,8 +546,10 @@ where
     ///
     /// # Arguments
     ///
-    /// - `start_dim`: The starting dimension of the range to be flattened.
-    /// - `end_dim`: The ending dimension of the range to be flattened (inclusive).
+    /// - `start_dim`: The starting dimension of the range to be flattened,
+    ///   supports negative indexing.
+    /// - `end_dim`: The ending dimension of the range to be flattened (inclusive),
+    ///   supports negative indexing.
     ///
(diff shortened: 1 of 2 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -575,7 +577,13 @@ where
     ///     println!("{flattened}");
     /// }
     /// ```
-    pub fn flatten<const D2: usize>(self, start_dim: usize, end_dim: usize) -> Tensor<B, D2, K> {
+    pub fn flatten<const D2: usize>(
+        self,
+        start_dim: impl AsIndex,
+        end_dim: impl AsIndex,
```
antimora, line 583: For code consistency, I suggest using a generic parameter, as was done previously.

antimora (changes requested):

crates/burn-tensor/src/tensor/api/base.rs
crutcher: Unfortunately, that changes `.flatten()` from a 1-param to a 3-param interface; and I can't work out how to make the additional params inferable when the target rank isn't.

laggui: Both could be tied to the same generic, so 2. But still breaks the number of generics.

I think this is a valid reason to leave it as `impl AsIndex` 👍

laggui (approved):

PR #3623, opened 2025-08-27
CLIP ViT-B/32 text model ONNX verification & backend fixes

antimora:
This PR implements comprehensive fixes for CLIP ViT-B/32 text model ONNX verification and extends backend testing capability. The changes focus on resolving type inference issues with Shape types, improving binary operation broadcasting for comparison operations, and adding constant lifting optimizations to ensure proper ONNX model support across different backends.

### Checklist
- [x] Ran `cargo run-checks`
- [x] Updated the book / docs for any new public APIs or examples

### Related Issues/PRs
- Follow-up backend issues created from this work:
  - Candle issue: https://github.com/tracel-ai/burn-onnx/issues/50
  - Metal issue: https://github.com/tracel-ai/burn/issues/3634

### Changes
- Fixed Shape type propagation and binary operations to handle mixed Tensor/Shape inputs correctly
- Improved comparison operation broadcasting in the ndarray backend to support complex tensor shape combinations
- Enhanced constant lifting logic to properly track usage counts and prevent unnecessary removal of reused constants
- Other fixes

### Testing

Performed:
- Ran ONNX parity test with multiple random text batches; all matched within tolerance on primary backend(s).

The model matches the output from running ONNX runtime on tch, metal and ndarray: 

image

image

image

### DIFF
--- a/crates/burn-import/src/burn/node/range.rs
+++ b/crates/burn-import/src/burn/node/range.rs
@@ -65,20 +65,13 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for RangeNode {
                 let literal = Literal::i64_suffixed(*value);
                 quote! { #literal }
             }
-            RangeParam::Runtime(t) => {
-                match t {
-                    Type::Scalar(s) => {
-                        let name = &s.name;
-                        quote! { #name }
-                    }
-                    Type::Tensor(tensor) if tensor.rank == 0 => {
-                        // Rank 0 tensor is treated as scalar
-                        let name = &tensor.name;
-                        quote! { #name }
-                    }
-                    _ => panic!("Range parameter must be a scalar"),
+            RangeParam::Runtime(t) => match t {
+                Type::Scalar(s) => {
+                    let name = &s.name;
+                    quote! { #name }
                 }
-            }
+                _ => panic!("Range parameter must be a scalar"),
+            },
         };
 
         let limit = match &self.limit {
@@ -86,20 +79,13 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for RangeNode {
                 let literal = Literal::i64_suffixed(*value);
                 quote! { #literal }
             }
-            RangeParam::Runtime(t) => {
-                match t {
-                    Type::Scalar(s) => {
-                        let name = &s.name;
-                        quote! { #name }
-                    }
-                    Type::Tensor(tensor) if tensor.rank == 0 => {
-                        // Rank 0 tensor is treated as scalar
-                        let name = &tensor.name;
-                        quote! { #name }
-                    }
-                    _ => panic!("Range parameter must be a scalar"),
+            RangeParam::Runtime(t) => match t {
+                Type::Scalar(s) => {
+                    let name = &s.name;
+                    quote! { #name }
                 }
-            }
+                _ => panic!("Range parameter must be a scalar"),
+            },
         };
 
         let delta = match &self.delta {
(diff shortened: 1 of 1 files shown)


crates/burn-import/src/burn/node/matmul.rs
```diff
@@ -45,37 +45,59 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for MatmulNode {
...
-                        let #output = #lhs.matmul(#rhs.unsqueeze_dims(&#axes)).squeeze(#squeeze_dim);
+                        let #output = #lhs.matmul(#rhs.unsqueeze_dims(&[#(#unsqueeze_dims),*])).squeeze::<#output_rank>(#squeeze_dim);
                     }
                 } else {
+                    // General tensor broadcasting: add leading dimensions
+                    let mut rhs_expanded = rhs;
+                    for i in 0..num_unsqueezes {
+                        let current_rank = rhs_dim + i + 1;
+                        rhs_expanded = quote! { #rhs_expanded.unsqueeze::<#current_rank>() };
+                    }
```
laggui: Repeatedly calling `.unsqueeze()` on current rank + 1 is the same as calling `.unsqueeze()` once for the output rank?

crates/burn-import/src/burn/node/matmul.rs
```diff
@@ -45,37 +45,59 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for MatmulNode {
...
+                    // General tensor broadcasting: add leading dimensions
+                    let mut lhs_expanded = lhs;
+                    for i in 0..num_unsqueezes {
+                        let current_rank = lhs_dim + i + 1;
+                        lhs_expanded = quote! { #lhs_expanded.unsqueeze::<#current_rank>() };
+                    }
+
                     quote! {
-                        let #output = #lhs.unsqueeze_dims(&#axes).matmul(#rhs);
+                        let #output = #lhs_expanded.matmul(#rhs);
```
laggui: Same comment regarding unsqueeze on leading dims, which can be done with a single `.unsqueeze()` call.

crates/burn-import/src/burn/node/range.rs
```diff
@@ -1,39 +1,130 @@
...
+                let literal = Literal::i64_suffixed(*value);
+                quote! { #literal }
+            }
+            RangeParam::Runtime(t) => {
+                match t {
+                    Type::Scalar(s) => {
+                        let name = &s.name;
+                        quote! { #name }
+                    }
+                    Type::Tensor(tensor) if tensor.rank == 0 => {
```
laggui: Wouldn't that be a `Type::Scalar` already? 🤔

crates/burn-import/src/onnx/to_burn.rs
```diff
@@ -722,9 +742,27 @@ impl ParsedOnnxGraph {
...
+        let lhs = Type::from(lhs_arg);
+        let rhs = Type::from(rhs_arg);
+        let output = Type::from(output_arg);
+
+        log::debug!(
+            "mul_conversion types: lhs={:?}, rhs={:?}, output={:?}",
+            lhs,
+            rhs,
+            output
+        );
```
laggui: Are both debug statements useful?

crates/onnx-ir/src/from_onnx.rs
```diff
@@ -337,10 +343,21 @@ impl OnnxGraphBuilder {
...
                     for input in &node.inputs {
                         if matches!(&input.ty, ArgType::Tensor(t) if t.rank == 1) {
-                            shape_inputs.push((input.name.clone(), shape_rank));
+                            // Only consider constants for conversion
+                            // Check if this input is a constant (either has a value or comes from a Constant node)
+                            let is_constant = input.value.is_some()
+                                || all_nodes.iter().any(|n| {
+                                    n.node_type == NodeType::Constant
+                                        && n.outputs.iter().any(|o| o.name == input.name)
+                                });
```
laggui, line 356: This works but as I highlighted in a previous review, we're already doing multiple passes over all nodes for shape propagation and constant lifting. And now we're adding yet another pass over all the nodes.

We could improve this by tracking changed nodes or outputs so that only affected nodes are processed instead of iterating everything repeatedly.

Not required for this PR given all the other changes, but we need to keep that in mind.

antimora: Yeah, I know currently it's very messy but it works. I want to refactor the rank-inference into type-inference to have a pull mechanism on the node level instead of having a general blanket ahead-of-time transformation.

PR #3625, opened 2025-08-27
Remove D param from `BatchNorm<B, D>`.

crutcher:
Based upon discussion with @laggui

### Checklist

### Changes

`BatchNorm<B, D>` has a type/call interface distinct from the other norm layers in the tree; making norm-abstraction wrapper layers messy. Additionally, the param is used only in a run-time check in `.forward()`.

### DIFF
--- a/examples/import-model-weights/src/model.rs
+++ b/examples/import-model-weights/src/model.rs
@@ -12,10 +12,10 @@ pub struct Model<B: Backend> {
     conv1: Conv2d<B>,
     conv2: Conv2d<B>,
     conv3: Conv2d<B>,
-    norm1: BatchNorm<B, 2>,
+    norm1: BatchNorm<B>,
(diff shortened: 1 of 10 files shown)


examples/import-model-weights/src/model.rs
```diff
@@ -12,10 +12,10 @@ pub struct Model<B: Backend> {
     conv1: Conv2d<B>,
     conv2: Conv2d<B>,
     conv3: Conv2d<B>,
-    norm1: BatchNorm<B, 2>,
+    norm1: BatchNorm<B>,
     fc1: Linear<B>,
     fc2: Linear<B>,
-    norm2: BatchNorm<B, 0>,
+    norm2: BatchNorm<B>,
```
laggui, line 18: `BatchNorm<B, 0>` was always a bit confusing 😅

PR #3627, opened 2025-08-27
Reorganize activation layer sources into `nn.activation` module

crutcher:
### Checklist

### Changes

- Aligned `ActivationConfig` name spellings with matched types.
- Moved activation-related files into the `nn/activation` directory.
- Updated imports and documentation accordingly.
- Deprecated direct activation exports in `nn` module.

### DIFF
--- a/crates/burn-core/src/nn/activation/swiglu.rs
+++ b/crates/burn-core/src/nn/activation/swiglu.rs
@@ -5,8 +5,7 @@ use crate::module::{Content, DisplaySettings, Module, ModuleDisplay};
 use crate::tensor::activation::silu;
 use crate::tensor::{Tensor, backend::Backend};
 
-use crate::nn::Initializer;
-use crate::nn::{Linear, LinearConfig, LinearLayout};
+use crate::nn::{Initializer, Linear, LinearConfig, LinearLayout};
 
 /// Configuration to create a [SwiGlu](SwiGlu) activation layer using the [init function](SwiGluConfig::init).
 #[derive(Config, Debug)]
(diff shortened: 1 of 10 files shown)


crutcher:
@laggui

crutcher:
This includes the onnx code-gen, btw.

crates/burn-core/src/nn/mod.rs
```diff
@@ -22,40 +22,23 @@ pub mod interpolate;
...
-mod relu;
 mod rnn;
 mod rope_encoding;
-mod sigmoid;
-mod swiglu;
-mod tanh;
 mod unfold;
 
+#[deprecated(since = "0.19.0", note = "Use `nn.activation` module instead.")]
+pub use activation::*;
```
laggui: I think we can keep the current usage without forcing everyone to move to `nn::activation` (leaving out the deprecation note). Activations are still part of the `nn` module, and you can use one or the other path if you prefer for convenience.

crutcher:
@laggui `Linear` was just ... late-night autopilot. It's totally wrong, rolling back all of that part.

crates/burn-core/src/nn/mod.rs
crutcher: Done.

laggui (commented):
LGTM! Just need to keep this up to date with main (i.e., fix the conflicts 😄)

crutcher:
@laggui one last question; should `GLU` be in `nn.activation`; or is it ... something else?

it *isn't* part of the wrapper layer, it's call semantics are very different.

but it's ... morally? ... an activation? I can't think of a better place for it.

laggui (approved):
> should GLU be in nn.activation; or is it ... something else?

GLU is a bit atypical when compared to other activation functions.. but it is still generally considered as such.

P.S: I reverted some equivalent import changes that were introduced with the work in progress.

Issue #3639, opened 2025-08-28
Random seed doesn't probagate correctly


nathanielsimard:
Setting the seed should be on the device instead of being static on the backend trait. This will allow sending the seed to an async compute server, which requires a client to communicate with, which can be obtained for a device.

PR #3641, opened 2025-08-28
Refactor/seed

nathanielsimard:
Fix [#3639](https://github.com/tracel-ai/burn/issues/3639)


crates/burn-ndarray/src/backend.rs
```diff
@@ -65,7 +65,7 @@ impl<E: FloatNdArrayElement, I: IntNdArrayElement, Q: QuantElement> Backend for
         String::from("ndarray")
     }
 
-    fn seed(seed: u64) {
+    fn seed(_device: &Self::Device, seed: u64) {
```
laggui, line 68: There's only one device for ndarray, but should the SEED be thread-local in this case?

Issue #3598, opened 2025-08-22
SafetensorsFileRecorder: Support streaming, selection, and direct use of Safetensors APIs for memory efficiency


antimora:
### Feature description

Currently, `SafetensorsFileRecorder` eagerly loads all tensors into host memory by calling `safetensors::load`, building a full tensor map even when only a subset is needed. For large models, this is wasteful and can lead to OOMs or slow loading times.

Proposed improvements:
- Add an option to stream or selectively load tensors from safetensors files, instead of always loading everything.
- Use the low-level `safetensors` Rust APIs (`SafeTensors`, `tensor`, etc) directly to iterate over metadata and load only requested tensors.
- Enable include/exclude key filtering and immediate device upload to minimize memory use and double buffering.
- Provide a backwards-compatible API: keep eager loading as default, but allow users to opt-in to streaming and filtering.

### Feature motivation
- Memory footprint for large `.safetensors` models becomes a bottleneck (peak RAM = all tensors at once).
- Many users only need a subset of model weights (e.g., encoder, specific layers) or want to fine-tune part of a model.
- Fast startup and compatibility with constrained environments (CI, web, embedded) requires less eager allocation.
- Align Safetensors import with similar improvements for ONNX and PyTorchFileRecorder (issues #3596, #3597).

Issue #3597, opened 2025-08-22
PyTorchFileRecorder: Memory‑efficient streaming load


antimora:
### Summary

`PyTorchFileRecorder` (in `crates/burn-import/src/pytorch/recorder.rs`) currently loads *all* tensor data from a `.pt` / `.pth` archive eagerly into memory through `reader::from_file` (`crates/burn-import/src/pytorch/reader.rs`). For large PyTorch state dictionaries this results in high peak RSS and limits feasibility for large foundation models, constrained devices, CI, and WebAssembly.  

We should add a memory-efficient pathway (streaming / lazy / mmap-backed) that decodes and materializes tensor data only when required, similar to the planned ONNX lazy weight loading (see issue #3596). Additionally, part of Candle’s `.pt` reading implementation can be reused (license-compatible) to avoid re‑implementing robust streaming/pickled tensor deserialization. Attribution must be added to `NOTICES.md`.

### Current Behavior (What Happens Today)

- `PyTorchFileRecorder::<PS>::load(...)` invokes `from_file::<PS, R::Item, B>` which:
  - Opens the zip (if TorchScript archive) or raw pickle.
  - Fully deserializes all pickle objects / tensors into host memory (`candle_core::pickle` & custom adapter types in `common::candle`).
  - Applies optional key remapping & unflattening (`remap`, `unflatten`).
  - Returns an in-memory `Record` object which is then transferred to the backend device.
- No streaming, chunking, or deferred materialization is performed.
- Non-contiguous layer indices are normalized eagerly (see docs section “Automatic Handling of Non‑Contiguous Indices”).
- There is no configuration knob to choose a lazy reader vs eager reader.

### Goals

- Provide a *drop-in backward-compatible* API: existing `.load()` continues to work (perhaps still defaulting to eager for now).
- Introduce a **lazy / streaming mode**:
  - Incrementally read pickle/zip structure.
  - Defer decoding of tensor raw bytes until each parameter is requested (or optionally decode + stream directly to backend tensor buffers).
- Enable **selective layer loading** via filter predicate (e.g. regex or prefix list).
- Optionally support **direct device upload** to avoid staging full tensor on host (when backend supports).
- Provide a **progress callback / logging hooks** for large model ingestion.
- Reduce peak memory usage significantly (target: O(single largest tensor + small metadata) instead of O(total model size)).

Issue #3596, opened 2025-08-22
Implement memory-efficient ONNX weight loading (lazy/protobuf streaming)


antimora:
### Feature description

Currently, `burn-import`'s ONNX loader (`onnx_ir`) reads all ONNX weights (initializers) into memory up front. For large models, this approach is memory intensive and can cause scalability issues or OOM crashes. Instead, we should move to a strategy where weights and tensors are read from the ONNX/protobuf file only as needed during model loading or code generation, possibly using streaming/protobuf lazy parsing.

#### Current state
- `onnx_ir`/`burn-import` loads all weight tensors at once, regardless of actual need.
- This behavior is particularly problematic for large models or resource-constrained environments.
- No option exists for memory-mapped or streaming/lazy loading.

#### Proposal
- Refactor ONNX loading in `burn-import`/`onnx_ir` to support on-demand reading of weights/tensors from the ONNX file.
- Use protobuf's streaming API, or a similar mechanism, to avoid loading unnecessary data into memory.
- Consider providing both eager (current) and lazy/streaming modes for backwards compatibility.
- Update codegen and all ONNX operator implementations in `burn-import` to work with on-demand tensor access.
- Document any new APIs or usage considerations for downstream users.

### Feature motivation

- Support importing and working with extremely large ONNX models without OOM errors.
- Reduce the memory footprint for typical ONNX import scenarios.
- Enable use of `burn-import` in environments with constrained memory (e.g., embedded, wasm, CI/CD, cloud).
- Bring burn-import's ONNX handling up to par with best practices in other frameworks (cf. PyTorch, TensorFlow, ONNX Runtime).

### (Optional) Suggest a Solution

- Investigate the protobuf parsing used in `onnx_ir`, and refactor to support iterators or readers for initializers/weights.
- Use streaming reads for large tensor data blocks, and only decode weights as needed for each node/operator.
- Consider a trait or abstraction for weight access that can be implemented for both eager and lazy backends.
- Profile and benchmark memory usage before/after.
- Add regression tests for large ONNX models to ensure memory use stays low.

---

#### Context
- Related to scalability and performance limitations in current `burn-import` ONNX support.
- Not directly addressed by any existing open tickets; this is a new proposal.
- For overlapping concerns, see existing issues on ONNX import scalability, async, and backend support.

BjornTheProgrammer:
[...]

PR #3651, opened 2025-08-31
Add efficient tensor import system

antimora:
This PR is dependent on https://github.com/tracel-ai/burn/pull/3649

## Checklist

- [x] Confirmed that `cargo run-checks` command has been executed
- [x] Made sure the book is up to date with changes in this PR

## Related Issues/PRs

* https://github.com/tracel-ai/burn/pull/3649
* https://github.com/tracel-ai/burn/issues/3597
* https://github.com/tracel-ai/burn/issues/3598
* https://github.com/tracel-ai/burn/issues/3596

## Summary

Adds a new tensor import API that provides direct loading of module tensors via dot-notation paths (e.g., `encoder.layer1.weight`) from `HashMap<String, TensorView>` or streaming readers, complementing the export API.

## Motivation

- Current system requires loading entire models into memory
- Cannot selectively load specific tensors
- Need memory-efficient streaming/lazy loading for large models
- Complex Module → Record → RecordItem → Deserializer pipeline

### New API

```rust
use burn::module::ModuleImport;

// Import all tensors from exported views
let views = source_model.export_tensor_views();
target_model.import_tensor_views(views, &device);

// Import filtered tensors using regex patterns
let result = target_model.import_tensor_views_filtered(
    views,
    &device,
    &[r"^encoder\..*"]
)?;

// Import using custom predicate function
let result = target_model.import_tensor_views_with_predicate(
    views,
    &device,
    |path| path.ends_with(".weight")
);

// Check import results
assert!(result.is_success());
println!("Applied: {}, Skipped: {}", result.applied.len(), result.skipped.len());
```

### Core Components

**TensorReader trait**: Extensible interface for lazy loading from various sources
```rust
pub trait TensorReader: Send {
    fn list_tensors(&mut self) -> Result<Vec<String>, ReaderError>;
    fn tensor_metadata(&mut self, path: &str) -> Result<TensorMetadata, ReaderError>;
    fn read_tensor_view(&mut self, path: &str) -> Result<TensorView, ReaderError>;
    fn read_all_views(&mut self) -> Result<HashMap<String, TensorView>, ReaderError>;
}
```

**TensorApplier**: Maps imported tensors to module hierarchy using enhanced `ModuleMapper`

**ModuleImport trait**: High-level import methods with three approaches:
- `import_tensor_views()`: Import all tensors
- `import_tensor_views_filtered()`: Filter with regex patterns (OR union)
- `import_tensor_views_with_predicate()`: Filter with custom function

**ImportResult**: Detailed feedback on import operation
```rust
pub struct ImportResult {
    pub applied: Vec<String>,    // Successfully imported
    pub skipped: Vec<String>,    // Filtered out
    pub missing: Vec<String>,    // In module but not in source
    pub unused: Vec<String>,     // In source but not in module
    pub errors: Vec<String>,     // Failed imports
}
```

### Key Features

- **Zero-copy round-trip**: Reuses `TensorView` from export API
- **Streaming support**: `TensorReader` trait enables lazy loading from files/network
- **Regex filtering**: Multiple patterns work as OR union
- **Predicate filtering**: Full control via custom functions
- **Memory efficient**: Only loads requested tensors
- **Detailed feedback**: `ImportResult` provides comprehensive status

### Implementation Details

- **Enhanced ModuleMapper**: Added `enter_module`/`exit_module` for path tracking
- **Fixed Vec/Array mapping**: Updated `primitive.rs` to properly track indices during map operations
- **Single-pass design**: Simplified from initial two-phase approach per user feedback
- **Derive macro updates**: Modified `burn-derive` to call enter/exit methods for all module fields

## Testing

Comprehensive test coverage including:
- Vec modules with indexed paths (`layers.0.weight`)
- Array modules (`[Linear<B>; 3]`)
- Enum modules with variant names (`Small.weight`)
- Optional fields (`Option<Linear<B>>`)
- Deeply nested structures with Vecs
- Regex pattern filtering
- Custom predicate functions
- Shape mismatch detection
- Partial imports with missing tensors

## Breaking Changes

None. Fully backward compatible. The `ModuleMapper` trait changes only add default implementations.

## Notes

- Concrete reader implementations (file, network, etc.) left for future PRs as requested
- Supports all module types that export supports: Vec, Array, Enum, Option, and nested combinations
- `ImportResult` provides detailed feedback for debugging and validation
- Predicate-based filtering provides maximum flexibility without adding complex dependencies


antimora:
I have converted to draft because I want to bring changes from a PR that depends of this.

antimora:
@laggui, I think it's ready for your review. I have another follow up PR that relies on this changes for importing/exporting safetensors very efficiently.

antimora:
Thank you, @laggui, for a quick pass. 

> I think the `Persister` naming is a bit confusing and potentially misleading, but generally not having to load the whole model structure in memory will be useful 🙂

I am open to renaming or merging with ModulePersist with Record and ModulePersister with Recorder. In my proposal, I didn't want to mix both up -- that's why I had it under a different name. At high level they are the same but the details differ significantly. And spent more effort on the details for efficiency and flexibility. Also take a look at the safetensors implementation of ModulePersister: https://github.com/tracel-ai/burn/pull/3666. The implementation has influenced the design.

> 
> This is also closer to a module param container/adapter with traversal, which could be nice for an eventual `Module::iter`.
> 

I did consider using Iterator interface for fetching and applying TensorView, however, in conjunction with the visitor pattern, it's impossible with [generators](https://dev-doc.rust-lang.org/beta/unstable-book/language-features/generators.html#generators) Rust feature! Which is unfortunately only available in unstable rust. Anything else will require Macro hacks on static type. `Vec<TensorView>` is optimal. 

> I will have to give this a bit more thought to see how we can possibly unify this. Recorders are the current way to persist module parameters, so having two distinct systems creates API fragmentation.

Yes, please do spend time learning. Lets figure out how to the best way. It's a crucial part of Burn framework.

antimora:
@nathanielsimard 
> I think there are naming issues with the PR. We can't use `TensorView` since that could be confusing with a real tensor view (reshape, slice, etc.) in the framework for new users.

How about `TensorSnap`?

> 
> I would probably extract most of the code into another crate, probably `burn-import`, and only keep the newly added visitor and mapper functions. I think we should move toward a visitor-based loading and saving approach, but using the recorder and record API. This will require a lot of thought and benchmarks. Our way of loading models is really quick, and I would be surprised if a new way were faster, at least with the initial implementation. We will be able to speed it up with streaming, but right now we have optimal CPU loading, which is very easy to mess up.
> 
> Anyway, I think this PR is still valuable for loading `safetensors`, but it's probably better if it lives in another crate behind a feature flag. We could bring some of its components to the record system iteratively.

In a follow-up PR (https://github.com/tracel-ai/burn/pull/3666), I introduced `burn-persist` but since both you advocating keeping record, which I am okay with, can we perhaps introduce `burn-record` crate? We can move all implementations there. With the new `safetensors` feature, we can import AND export. So it makes more sense if it has its own crate.

Also `burn-record` can have its own benchmarks. Most improvements are memory saving, especially with large models. Serde de/serializer must load the entire model into ram before sending to GPU.

antimora:
Per our [Discord conversation](https://discord.com/channels/1038839012602941528/1091796857996451942/1414662208595427398):

1. Move ModulePersist and ModulePersister to `burn-persist` crate defined here: https://github.com/tracel-ai/burn/pull/3666
2. Rename TensorView to TensorSnapshot
3. Keep Record and Recorder independent and over time we will merge both. We do not wish to disrupt the existing users and confuse with the new ways to load/save, however, we want to improve weight loading/saving. `burn-persist` will be that experiment.
4. `burn-persist` will collect benchmarks to demonstrate the memory efficiency with the new approach.

antimora:
Closing to keep all the changes in https://github.com/tracel-ai/burn/pull/3666

PR #3666, opened 2025-09-04
Safetensor persister

antimora:
### Checklist

### Related Issues/PRs

### Changes

### Testing


antimora:
Per our [Discord conversation](https://discord.com/channels/1038839012602941528/1091796857996451942/1414662208595427398):

1. Move ModulePersist and ModulePersister to `burn-persist` crate defined here.
2. Rename TensorView to TensorSnapshot
3. Keep Record and Recorder independent and over time we will merge both. We do not wish to disrupt the existing users and confuse with the new ways to load/save, however, we want to improve weight loading/saving. `burn-persist` will be that experiment.
4. `burn-persist` will collect benchmarks to demonstrate the memory efficiency with the new approach.

crates/burn-core/src/module/base.rs
```diff
@@ -217,10 +217,49 @@ pub trait ModuleVisitor<B: Backend> {
     fn visit_int<const D: usize>(&mut self, _id: ParamId, _tensor: &Tensor<B, D, Int>) {}
     /// Visit a bool tensor in the module.
     fn visit_bool<const D: usize>(&mut self, _id: ParamId, _tensor: &Tensor<B, D, Bool>) {}
+
+    /// Called when entering a submodule.
+    fn enter_module(&mut self, _name: &str, _container_type: &str) {}
```
crutcher: This should document the form of the name; and if it's a path.

Alternatively, paths could be `&[&str]` or `&[String]`; so that the visit stack can re-use a `Vec`.

crates/burn-core/src/module/base.rs
```diff
@@ -217,10 +217,49 @@ pub trait ModuleVisitor<B: Backend> {
...
+
+    /// Called when exiting a submodule.
+    fn exit_module(&mut self, _name: &str, _container_type: &str) {}
+
+    /// Visit a float tensor with its full module path.
+    fn visit_float_with_path<const D: usize>(
+        &mut self,
+        _path: &str,
```
crutcher: Same comment about `enter_module`s name; suggestion about paths.

crates/burn-core/src/module/base.rs
crutcher: per conversation with @nathanielsimard ; having a `&dyn Any` reference here to the module in question opens a lot of things.

Also, enter/exit raise a question about pre/post order semantics; and if traversal is cancelable.

crates/burn-persist/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! Module persistence and serialization infrastructure.
+//!
+//! This module provides comprehensive functionality for persisting and loading Burn modules
+//! and their tensor data. It supports bidirectional data flow with flexible filtering,
+//! remapping, and efficient memory management through lazy materialization.
```
crutcher: Modules should, in general, provide an overview of their major components and suggested usage examples in their top-level docs.

crates/burn-persist/src/traits.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::{PathFilter, TensorSnapshot};
+use burn_core::module::Module;
+use burn_tensor::backend::Backend;
+
+/// Extension trait for modules that provides tensor persistence functionality.
+///
+/// This trait provides convenient methods to collect and apply tensor views from any Burn module.
+/// Collection operations create lightweight tensor views without immediately copying data.
+/// Apply operations apply tensor data from views to the corresponding tensors in the module.
+pub trait ModulePersist<B: Backend>: Module<B> + Clone {
```
crutcher: My biggest question is about non-tensor state. How is that handled?

crates/burn-persist/src/keyremapper.rs
```diff
@@ -0,0 +1,284 @@
...
+    ///
+    /// # Returns
+    ///
+    /// A tuple containing:
+    /// * The remapped Vec of TensorSnapshots with updated paths
+    /// * A vector of (new_path, original_path) showing the transformations
+    pub fn remap(
+        &self,
+        mut tensors: Vec<TensorSnapshot>,
+    ) -> (Vec<TensorSnapshot>, Vec<(String, String)>) {
```
crutcher: you could make this generic; and drop all references to `TensorSnapshot` from the module?
`remap(items: Vec<T>, fn(&T)=>String, fn(String, T)=>T)`?

crates/burn-persist/src/filter.rs
```diff
@@ -0,0 +1,609 @@
...
+///     .with_regex(r".*\.weight$")
+///     .with_full_path("special_tensor");
+///
+/// // Check if a path should be included
+/// if filter.matches("encoder.layer1.weight") {
+///     // This will match due to both regex patterns
+/// }
+/// ```
+#[derive(Debug, Clone, Default)]
+pub struct PathFilter {
```
crutcher: So; we generally expect these to be dotted-paths; and it seems that a filter that works natively with that approach would be preferable to one that had to explicitly model `\.`

crates/burn-persist/src/collector.rs
```diff
@@ -0,0 +1,958 @@
...
+
+/// Collects tensor views from modules without copying data.
+///
+/// This collector traverses a module hierarchy and creates lightweight views
+/// of tensors that can be materialized to `TensorData` on demand.
+///
+/// # Examples
+///
+/// ## Collect all tensors
+/// ```ignore
```
crutcher: these should be ```rust,ignore``` to get better IDE handling.

crates/burn-persist/src/adapter.rs
```diff
@@ -0,0 +1,404 @@
...
+#[cfg(feature = "std")]
+use crate::KeyRemapper;
+use crate::TensorSnapshot;
+
+/// Trait for adapting tensors between different framework conventions.
+///
+/// This trait allows conversion between different deep learning framework formats,
+/// handling both loading and saving operations. Implementations can transform
+/// tensor data, rename parameters, and modify shapes as needed.
+pub trait Adapter {
```
crutcher: "smurfing is bad" ... except when it's in a framework that's going to be used from 3rd-party libs. I think "Adapter" is too broad a name for a public trait like this.

crates/burn-persist/src/adapter.rs
```diff
@@ -0,0 +1,404 @@
...
+/// - Renames normalization parameters (PyTorch: weight/bias, Burn: gamma/beta)
+#[derive(Debug, Clone, Default)]
+pub struct PyTorchToBurnAdapter;
+
+impl Adapter for PyTorchToBurnAdapter {
+    fn adapt_tensor(&self, snapshot: &TensorSnapshot) -> Option<TensorSnapshot> {
+        let path = snapshot.full_path();
+
+        // Check if this is a linear layer weight that needs transposing
+        if is_linear_weight(&path) {
```
crutcher: Looks an awful lot like a plugin-stack model here ...

crutcher: Also; it seems that the place we have the most information is upon *applying* the weight to a Linear

crates/burn-persist/src/traits.rs
antimora: @crutcher 
> My biggest question is about non-tensor state. How is that handled?

Currently, it’s ignored by the visitor, just as it is with record serialization. This remains a big open question for discussion.

Related question. How should we go about metadata? Many exported models I’ve seen include metadata related to model configuration—examples include Safetensors, PyTorch `.pt` files (e.g., Whisper specifies model size), and formats like GGUF. Should we have a facility to import these into a special data-structure?

antimora:
The benchmarks tested 5 backends across 3 model sizes, comparing the new `SafetensorsPersister` with the old `SafetensorsFileRecorder`.


Memory Usage Comparison (New Persister vs Old Recorder)

  Peak Memory & Total Allocations

  | Backend  | Model | New Peak | Old Peak | Peak Δ | New Total | Old Total | Total Δ |
  |----------|-------|----------|----------|--------|-----------|-----------|---------|
  | NdArray  | S     | 7.4 MB   | 6.8 MB   | 8%    | 7.9 MB    | 21.0 MB   | -62%    |
  |          | M     | 122.5 MB | 112 MB   | 9%    | 134.8 MB  | 355.8 MB  | -62%    |
  |          | L     | 671.4 MB | 654.6 MB | 3%    | 983.4 MB  | 2,618 MB  | -62%    |
  | Candle   | S     | 7.4 MB   | 6.8 MB   | 8%    | 7.9 MB    | 21.0 MB   | -62%    |
  |          | M     | 122.5 MB | 112 MB   | 9%    | 134.8 MB  | 355.8 MB  | -62%    |
  |          | L     | 671.4 MB | 654.6 MB | 3%    | 983.5 MB  | 2,618 MB  | -62%    |
  | LibTorch | S     | 2.1 MB   | 5.3 MB   | -60%   | 2.7 MB    | 13.1 MB   | -80%    |
  |          | M     | 33.6 MB  | 89.8 MB  | -63%   | 45.0 MB   | 222.8 MB  | -80%    |
  |          | L     | 16.8 MB  | 654.6 MB | -97%   | 328.8 MB  | 1,636 MB  | -80%    |
  | Metal    | S     | 2.1 MB   | 5.3 MB   | -60%   | 2.7 MB    | 15.8 MB   | -83%    |
  |          | M     | 33.6 MB  | 89.8 MB  | -63%   | 45.1 MB   | 266.9 MB  | -83%    |
  |          | L     | 16.8 MB  | 654.6 MB | -97%   | 329.1 MB  | 1,964 MB  | -83%    |
  | WGPU     | S     | 2.1 MB   | 5.3 MB   | -60%   | 2.7 MB    | 15.8 MB   | -83%    |
  |          | M     | 33.6 MB  | 89.8 MB  | -63%   | 45.1 MB   | 266.9 MB  | -83%    |
  |          | L     | 16.8 MB  | 654.6 MB | -97%   | 329.1 MB  | 1,964 MB  | -83%    |

  S=Small, M=Medium, L=Large model

  Key Insights:
  - GPU backends achieve 83% memory reduction
  - LibTorch shows 80% memory reduction
  - CPU backends show 62% memory reduction
  - Large models see 97% peak memory reduction on GPU

  Speed Improvements (new/old ratio):

  | Backend           | Small Model  | Medium Model | Large Model  |
  |-------------------|--------------|--------------|--------------|
  | NdArray (CPU)     | 1.80x faster | 2.06x faster | 1.72x faster |
  | Candle            | 0.62x slower | 0.77x slower | 0.66x slower |
  | LibTorch          | 0.72x slower | 0.87x slower | 0.81x slower |
  | Metal (Apple GPU) | 3.31x faster | 3.04x faster | 3.22x faster |
  | WGPU              | 3.54x faster | 3.05x faster | 3.27x faster |


Throughput Comparison (MB/s → GB/s for better readability):

  | Backend  | Model Size | File Size | New Throughput | Old Throughput | Improvement |
  |----------|------------|-----------|----------------|----------------|-------------|
  | NdArray  | Small      | 2.63 MB   | 0.58 GB/s      | 0.32 GB/s      | 1.8x        |
  |          | Medium     | 44.9 MB   | 0.56 GB/s      | 0.27 GB/s      | 2.1x        |
  |          | Large      | 327.3 MB  | 0.56 GB/s      | 0.32 GB/s      | 1.7x        |
  | Candle   | Small      | 2.63 MB   | 0.71 GB/s      | 1.16 GB/s      | 0.6x        |
  |          | Medium     | 44.9 MB   | 0.72 GB/s      | 0.93 GB/s      | 0.8x        |
  |          | Large      | 327.3 MB  | 0.70 GB/s      | 1.05 GB/s      | 0.7x        |
  | LibTorch | Small      | 2.63 MB   | 1.32 GB/s      | 1.84 GB/s      | 0.7x        |
  |          | Medium     | 44.9 MB   | 1.30 GB/s      | 1.50 GB/s      | 0.9x        |
  |          | Large      | 327.3 MB  | 1.30 GB/s      | 1.61 GB/s      | 0.8x        |
  | Metal    | Small      | 2.63 MB   | 3.22 GB/s      | 0.97 GB/s      | 3.3x        |
  |          | Medium     | 44.9 MB   | 3.87 GB/s      | 1.27 GB/s      | 3.0x        |
  |          | Large      | 327.3 MB  | 4.24 GB/s      | 1.32 GB/s      | 3.2x        |
  | WGPU     | Small      | 2.63 MB   | 3.57 GB/s      | 1.01 GB/s      | 3.5x        |
  |          | Medium     | 44.9 MB   | 4.00 GB/s      | 1.31 GB/s      | 3.1x        |
  |          | Large      | 327.3 MB  | 4.36 GB/s      | 1.33 GB/s      | 3.3x        |


Key Findings:

  1. Memory Efficiency Champions:
    - GPU backends (Metal, WGPU): 83% reduction in total allocations, 97% reduction in peak memory for large models
    - LibTorch: ~80% reduction in total allocations across all model sizes
    - CPU backends (NdArray, Candle): Consistent 62% reduction in total allocations
  2. Throughput Performance:
    - Best: WGPU achieving 4.36 GB/s on large models (3.3x improvement)
    - Metal: Close second with 4.24 GB/s (3.2x improvement)
    - NdArray CPU: Solid 2x improvement despite being CPU-bound
  3. Outliers:
    - Candle and LibTorch show regression in speed but still benefit from memory improvements
    - Peak memory slightly higher for new persister on CPU backends, but total memory usage is dramatically lower
  4. Scalability:
    - GPU backends show better performance with larger models
    - Memory savings become more dramatic with larger models (up to 97% peak memory reduction)

The new `SafetensorsPersister` excels particularly in GPU-accelerated environments and provides universal memory efficiency improvements, making it ideal for resource-constrained deployments and large model handling.


[bench-results.txt](https://github.com/user-attachments/files/22268095/bench-results.txt)

crates/burn-persist/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+# External dependencies
+hashbrown = { workspace = true, features = ["serde"] }
+memmap2 = { workspace = true, optional = true }
+regex = { workspace = true, optional = true }
+
+# Note: safetensors requires either std or alloc feature
+# We always require std for file I/O functionality
+# change after https://github.com/huggingface/safetensors/issues/650 is released to support no_std
+# safetensors = { workspace = true }
+safetensors = { git = "https://github.com/huggingface/safetensors", rev = "0816a1a", default-features = false }
```
nathanielsimard: This can block a release, is it possible to deactivate no-std tests until that issue if merged?

crates/burn-persist/src/adapter.rs
nathanielsimard: In the scope of `burn-persist` it makes sense, do you have a better name that isn't too long?

laggui:
Great! Will review tomorrow 🙂

crates/burn-persist/Cargo.toml
antimora: yeah, there is a work-around by not passing HashMap of metadata data to safetensors (pass None), but it'll be broken no-std support.

@Narsil of safetensors has accepted my PR and merged. Hopefully, they'll release it before we do, otherwise it's impossible for us to support `no-std` using with `--no-default-features` in Burn.

crates/burn-core/src/module/base.rs
laggui: Agree with the comments above.

Eventually, instead of having separate `&str` for the name and container type, we could introduce something like:

```rust
/// A frame in the module path (name + type)
pub struct ModuleFrame {
    pub name: String,
    pub container_type: String,
}

/// The path of modules leading to a parameter
pub struct ModulePath {
    pub frames: Vec<ModuleFrame>,
}
```

And have the methods use these types instead. More docs + better types should help.

With this, we could potentially remove the enter/exit and _with_path modules, if visit/map have the module path arg (which you can choose to use or not).

Should help clarify the ambiguities raised above, and reduce the number of methods for the trait. The paths can be handled without enter/exit, and will likely simplify the `Collector` / `TensorPathCollector` into a simple module walker.

crates/burn-persist/src/adapter.rs
antimora: I can rename to ModuleAdapter, just like we had it previously. I figured in the scope adapter would be okay, but I can rename it easily.

crates/burn-persist/src/keyremapper.rs
antimora: For now lets keep it specific to the crate but if needed we can refactor.

crates/burn-core/src/module/base.rs
antimora: Refactored path handling to use `path: &[String]`

crates/burn-persist/src/filter.rs
antimora: We should probably support a subset of JsonPath expressions instead. It's hard to disambiguate regexp expressions if we escape some. Another alternative is to support glob patters like in this lib: https://docs.rs/glob/latest/glob/struct.Pattern.html

laggui (approved):

cex1gt:
Huge thank you @antimora and team!

PR #3671, opened 2025-09-05
Fix no-std Support for `burn-no-std-tests` and warning clean up

antimora:
## Summary

This PR fixes the no-std build configuration to ensure that `burn-no-std-tests` can be built without any `std` feature being enabled in its dependency tree. Previously, several issues caused `std` to be pulled in transitively, defeating the purpose of the no-std tests.

### 1. Removed Unused `Float` Trait Imports

* Eliminated conditional `Float` imports from 7 files that caused warnings.
* These imports were unnecessary when `std` was available through transitive dependencies.

### 2. Added Proper no-std Support to `burn-ndarray`

* Introduced a new `rand` module in `burn-ndarray`:

  * `NdArrayRng` type alias:

    * Uses `StdRng` for `std` builds.
    * Uses `SmallRng` for `no-std` builds.
  * `get_seeded_rng()` function returns the appropriate RNG type per configuration.
* For `no-std` builds:

  * Uses `const-random` to generate compile-time random seeds.
  * Each compilation gets a different seed, improving randomness over hardcoded values.
  * Deterministic within a single build for reproducibility.
* Updated all RNG usage to this unified approach, eliminating conditional compilation blocks.

### 3. Fixed Dependency Configurations

* Removed `std_rng` from workspace-level `rand` configuration.
* Updated `burn-core` to avoid explicitly requesting `std_rng`.
* Updated `burn-tensor` and `burn-ndarray` to use `rand` with `default-features = false`.
* Added `rand_chacha` with `default-features = false` to prevent pulling in `std`.
* Added `const-random` for compile-time random seed generation.
* `burn-ndarray` now uses `small_rng` by default, adding `std_rng` only when the `std` feature is enabled.

## Testing

* `cargo build` in `burn-no-std-tests` completes without warnings.
* Verified no `std` feature is enabled in the dependency tree via:

  ```bash
  cargo tree --no-default-features -f "{p} {f}" | grep -E " std[^_]| std$"
  ```
* Confirmed compile-time random seeds are generated for no-std builds.

## Impact

This PR ensures Burn can be used in true `no-std` environments, critical for embedded systems and other constrained targets where the standard library is unavailable.
The addition of compile-time random seed generation improves randomness in no-std builds while preserving reproducibility within a single compilation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,6 +138,7 @@ openblas-src = "0.10.12"
 rand = { version = "0.9.2", default-features = false, features = [
     "std_rng",
 ] } # std_rng is for no_std
+rand_chacha = { version = "0.9", default-features = false }
 rand_distr = { version = "0.5.1", default-features = false }
 serde = { version = "1.0.219", default-features = false, features = [
     "derive",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ bytemuck = "1.23.2"
 bytes = "1"
 candle-core = { version = "0.9.1" }
 clap = { version = "4.5.44", features = ["derive"] }
+const-random = "0.1"
 colored = "3.0.0"
 console_error_panic_hook = "0.1.7"
 csv = "1.3.1"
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -135,9 +136,8 @@ num-traits = { version = "0.2.19", default-features = false, features = [
     "libm",
 ] } # libm is for no_std
 openblas-src = "0.10.12"
-rand = { version = "0.9.2", default-features = false, features = [
-    "std_rng",
-] } # std_rng is for no_std
+rand = { version = "0.9.2", default-features = false }
+rand_chacha = { version = "0.9", default-features = false }
```
laggui: This dep is unused, shouldn't be required no?

PR #3681, opened 2025-09-08
Add NumPy-like take operation with multi-dimensional index support

antimora:
Adds a `take` operation to burn-tensor that matches NumPy's behavior, supporting multi-dimensional indices and negative dimension indexing.

Also extends select operations to support negative dimension indexing.

With this new method, we can cover ONNX Gather functionality partially (negative indices are not supported). The ONNX Gather implementation has been updated to use this new tensor operation, significantly simplifying the code.

### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/pull/3673

#### New `take` operation:
- Supports indices of any dimensionality (1D, 2D, 3D+)
- Output shape follows NumPy: input.shape[:dim] + indices.shape + input.shape[dim+1:]
- Supports negative dimension indexing (e.g., `-1` refers to last dimension)
- **Note**: Negative indices within the indices tensor are NOT supported (indices must be in range [0, dim_size))

#### Enhanced `select` and `select_assign` operations:
- Now support negative dimension indexing (e.g., `-1` refers to last dimension)
- Uses `AsIndex` trait and `canonicalize_dim` for dimension parameter

#### New `TensorCheck::take` validation:
- Validates dimension bounds
- Ensures output dimensions match expected formula: DO = D + DI - 1

#### Updated ONNX Gather implementation:
- Simplified from ~20 lines of complex logic to just 1 line using `take`
- Located in `/Users/dilshod/Projects/burn/crates/burn-import/src/burn/node/gather.rs`

### Testing

- 16 tests for take operation (removed tests for negative indices after PR feedback)
- 10 new tests for negative dimension indexing in select/select_assign operations
- All existing tests pass
- Verified on burn-ndarray backend
- All ONNX gather tests pass

### DIFF
--- a/crates/burn-tensor/src/tensor/api/check.rs
+++ b/crates/burn-tensor/src/tensor/api/check.rs
@@ -913,11 +913,10 @@ impl TensorCheck {
         if DO != expected_do {
             check = check.register(
                 "Take",
-                TensorError::new("Output dimension mismatch")
-                    .details(format!(
-                        "Expected output dimension {} (D={} + DI={} - 1) but got DO={}",
-                        expected_do, D, DI, DO
-                    )),
+                TensorError::new("Output dimension mismatch").details(format!(
+                    "Expected output dimension {} (D={} + DI={} - 1) but got DO={}",
+                    expected_do, D, DI, DO
+                )),
             );
         }
 
--- a/crates/burn-tensor/src/tests/ops/take.rs
+++ b/crates/burn-tensor/src/tests/ops/take.rs
@@ -250,13 +250,6 @@ mod tests {
         output.into_data().assert_eq(&expected, false);
     }
 
-
-
-
-
-
-
-
     #[test]
     fn should_take_with_negative_dim_2d() {
         // Test using negative dimension indexing on 2D tensor


antimora:
CC: @crutcher

crates/burn-tensor/src/tensor/api/take.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        // Calculate expected output dimensions
+        // DO = D - 1 + DI (remove 1 dim, add DI dims)
+        assert_eq!(
+            DO,
+            D + DI - 1,
+            "Output dimension mismatch: expected {} but got {}",
+            D + DI - 1,
+            DO
+        );
```
laggui: Should probably be in a new `TensorCheck::take`

crates/burn-tensor/src/tensor/api/take.rs
```diff
@@ -0,0 +1,130 @@
...
+        // Store the indices shape for reshaping later
+        let indices_shape = indices.shape();
+        let indices_dims = indices_shape.dims.clone();
+
+        // Flatten indices to 1D for processing
+        let indices_flat = if DI > 1 {
+            indices.flatten(0, DI - 1)
+        } else {
+            indices.reshape([indices_shape.num_elements()])
+        };
```
laggui: Both cases are equivalent, no? I think you can simply reshape to the number of elements, no branch.

crates/burn-tensor/src/tensor/api/take.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        // Normalize negative indices: if index < 0, add dim_size to it
+        // For example, with dim_size=3: -1 becomes 2, -2 becomes 1, -3 becomes 0
+        // This is done with: normalized = (indices % dim_size + dim_size) % dim_size
+        // Which handles both positive and negative indices correctly
+        let dim_size_scalar = dim_size as i32;
+        let normalized_indices = indices_flat
+            .clone()
```
laggui: `indices_flat` doesn't appear to be reused anywhere, so `.clone()` is not required

crates/burn-tensor/src/tensor/api/take.rs
```diff
@@ -0,0 +1,130 @@
...
+        let indices_flat = if DI > 1 {
+            indices.flatten(0, DI - 1)
+        } else {
+            indices.reshape([indices_shape.num_elements()])
+        };
+
+        // Normalize negative indices: if index < 0, add dim_size to it
+        // For example, with dim_size=3: -1 becomes 2, -2 becomes 1, -3 becomes 0
+        // This is done with: normalized = (indices % dim_size + dim_size) % dim_size
+        // Which handles both positive and negative indices correctly
```
laggui: We could possibly limit indices to positives and avoid additional normalization? Or maybe have a `mode`, similar to numpy's `take`.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1300,6 +1300,65 @@ where
...
+    /// fn example<B: Backend>() {
+    ///   let device = B::Device::default();
+    ///   let tensor = Tensor::<B, 2>::from_data([[1.0, -2.0, 3.0], [4.0, 5.0, 6.0]], &device);
+    ///   let indices = Tensor::<B, 1, Int>::from_data([0], &device);
+    ///   let tensor = tensor.select(0, indices);
+    ///   println!("{tensor}");
+    ///   //  [[1.0, -2.0, 3.0]]
+    /// }
+    /// ```
+    pub fn select(self, dim: usize, indices: Tensor<B, 1, Int>) -> Self {
```
crutcher:
```
    pub fn select<I: AsIndex>(self, dim: I, indices: Tensor<B, 1, Int>) -> Self {
      let dim = canonicalize_dim(dim, D, false);
```

crutcher: this, and other places? gives negative-indexed dims.

crates/burn-tensor/src/tensor/api/take.rs
crutcher: see `AsIndex`?

laggui: Yes but `AsIndex` doesn't apply to tensors 😅 hence why I am wondering what is the best way to handle this

antimora:
@laggui, @crutcher thank you for your reviews. I made the necessary changes per your feedback.

Notably I removed support for negative indices and used `dim: impl AsIndex` for select, select_assign, and take.

laggui (approved):
LGTM!

We should probably open an issue to replace the default bool select implementations.

antimora:
> LGTM!
> 
> We should probably open an issue to replace the default bool select implementations.

Done: https://github.com/tracel-ai/burn/issues/3697

PR #3686, opened 2025-09-08
Readme + Burn Book performance section

nathanielsimard:
Some updates to the readme + burn book

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,18 +37,18 @@ models in the cloud, then deploy on customer hardwares, which vary from user to 
 
 ### Supported Backends
 
-Most backends support all operating systems, so we don't mentionned them in the tables bellow.
+Most backends support all operating systems, so we don't mentioned them in the tables below.
 
 **GPU Backends:**
 
-|         | CUDA | ROCm | Metal | Vulkan | Wgpu | Candle | LibTorch |
-| ------- | ---- | ---- | ----- | ------ | ---- | ------ | -------- |
-| Nvidia  | ☑️   | -    | -     | ☑️     | ☑️   | ☑️     | ☑️       |
-| AMD     | -    | ☑️   | -     | ☑️     | ☑️   | -      | ☑️       |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -193,20 +35,29 @@ models in the cloud, then deploy on customer hardwares, which vary from user to
 
 </div>
 
-<br />
+### Supported Backends
+
+Most backends support all operating systems, so we don't mentionned them in the tables bellow.
```
laggui: `mentionned` -> `mentioned`

`bellow` -> `below`

README.md
```diff
@@ -193,20 +35,29 @@ models in the cloud, then deploy on customer hardwares, which vary from user to
...
+
+**GPU Backends:**
+
+|         | CUDA | ROCm | Metal | Vulkan | Wgpu | Candle | LibTorch |
```
antimora: I recommend to use WebGPU instead of Wgpu (which is implementation of WebGPU standard). People are more familiar (also search engines will up on this too). WebGPU is probably on the same level as Vulkan and CUDA

burn-book/src/performance/good-practices/asynchronous-execution.md
```diff
@@ -0,0 +1,62 @@
...
+Most Burn backends execute tensor operations in an asynchronous manner. However, the async notation
+is often not required for most tensor operations, privileging the simplicity of sync Rust.
+
+There are only a few operations that trigger synchronization of the backend, and it is very
+important to correctly handle those to optimize hardware utilization. Those operations are
+`into_data`, `into_scalar`, and `sync`. Some tensor operations might call `into_data` underneath,
+triggering a synchronization, like `to_device` for some backends.
+
+There are many different ways to optimize those syncs, and the first one is to group them into a
+single transaction. Burn provides a high-level composable API to build transactions, which will only
```
laggui: I think I would rephrase slightly to something that might be a bit more clear to others:

There are several ways to minimize synchronization overhead, one of which is to batch sync operations into a single transaction.

PR #3693, opened 2025-09-10
Remove DimSelectActivation

crutcher:
### Checklist

### Changes

This was an overly-specific idea.

If we add it back, it should be a module-generic form of, maybe, PermutationWrapped.


crutcher:
@laggui can you push the button?

PR #3695, opened 2025-09-10
Perf/data transfer

nathanielsimard:
Optimize data transfer between multiple devices.

It doesn't yet solve our problems with multi device training, but it's better!


laggui (approved):
Leaving the TODO comments aside, I think we should clean up  the `Device` / `DeviceOps` traits with the re-export.

But this is not a show stopper for this PR. Let's merge to keep it going.

Issue #1538, opened 2024-03-27
Linear algebra operations


laggui:
### Feature description

Add common linear algebra operations.

### Feature motivation

We've had a couple of users ask about support for linalg ops like SVD, QR decomposition and matrix exponential. We added matrix operations like the diagonal but don't have matrix/vector norm or determinant.

### Suggest a Solution

Maybe in a new `burn-linalg` crate, but would have to see how that would work for support on different backends.

jungerm2:
Two questions:
- Is there any roadmap for this? The lack of linalg is a major issue IMO. 
- At the moment I'm converting to ndarray and back in order to do some linalg operations. I have:
```rust
pub fn array2_to_tensor2<T, B: Backend>(arr: Array2<T>, device: &B::Device) -> Tensor<B, 2>
where
    f32: From<T>,
{
    // TODO: This is likely suboptimal
    assert!(arr.is_standard_layout());
    let shape = Shape::new(arr.dim().into());
    Tensor::<B, 1>::from_floats(
        &arr.into_raw_vec()
            .into_iter()
            .map(|i| i.into())
            .collect::<Vec<_>>()[..],
        &device,
    )
    .reshape(shape)
}

pub fn tensor2_to_array2<B: Backend>(tensor: Tensor<B, 2>) -> Array2<f32> {
    let vals: Vec<f32> = tensor.to_data().convert().value;

    Array2::from_shape_vec(tensor.dims(), vals)
        .expect("Tensor should be converted to Array of same shape.")
}
```
Which enables me to do stuff like `array2_to_tensor2(tensor2_to_array2().inv()?)`. Is this the best workaround at the moment? It's rather slow, as data is copied around (I'm using a wgpu backend), and I'm not sure how to make the above use dynamic dimensions either as tensors use `D: usize` and ndarrays use `ndarray::Dimension` such as `Ix2` and `Ix3`, etc. Any insights? Maybe there's a better bridge between backends I'm missing here...

laggui:
I'm sorry there is no roadmap for this feature at this time. We have a lot of other things we're working to improve and add like quantization.

Would love to see the support for linalg operations come to light, just can't say when it will happen as of now.

[...]

PR #3703, opened 2025-09-11
Added trace and diag with batch support for linalg crate

niklund:
### Checklist

### Related Issues/PRs

Relevant issue:
https://github.com/tracel-ai/burn/issues/1538

### Changes

The feature description of the issue is to add common linear algebra operations. There are some existing in
the linalg crate such as vector norms and cosine similarity. I  added trace and diag operation for starters. Will continue working on more operations. 
### Testing

Test cases covering rank 2, 3 and 4. Square tall and wide matrices and edge cases such as 1x1, single row/column with multiple data types. 
Tested with: cargo test -p burn-ndarray diag && cargo test -p burn-ndarray trace
cargo run-checks passes.

### DIFF
--- a/crates/burn-tensor/src/tensor/linalg/diag.rs
+++ b/crates/burn-tensor/src/tensor/linalg/diag.rs
@@ -4,7 +4,7 @@ use crate::check::TensorCheck;
 use crate::tensor::{Int, Shape, Tensor};
 use crate::{BasicOps, TensorKind};
 
-/// Returns the diag of the of a matrix.
+/// Returns the diag of a matrix.
 ///
 /// For batched inputs, returns of each matrix in the batch independently.
 ///
@@ -16,9 +16,7 @@ use crate::{BasicOps, TensorKind};
 /// * `tensor` - The input tensor with at least 2 dimensions.
 ///
 /// # Returns
-/// Tensor with rank D - 1, where the last two matrix dimensions are replaced by a single
-/// dimension containing the diagonal elements
-///
+/// A tensor of rank `D - 1`, where the last dimension contains the diagonal elements of the input.
 pub fn diag<B: Backend, const D: usize, const DO: usize, K>(
     tensor: Tensor<B, D, K>,
 ) -> Tensor<B, DO, K>
--- a/crates/burn-tensor/src/tensor/linalg/trace.rs
+++ b/crates/burn-tensor/src/tensor/linalg/trace.rs
@@ -2,7 +2,7 @@ use super::diag;
 use crate::backend::Backend;
 use crate::tensor::Tensor;
 
-/// Computes the trace of the of square matrices.
+/// Computes the trace of a matrix.
 ///
 /// For batched inputs, computes the trace of each matrix in the batch independently.
 ///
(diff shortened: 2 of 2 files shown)


crates/burn-tensor/src/tensor/linalg/trace.rs
```diff
@@ -0,0 +1,42 @@
...
+/// # Arguments
+///
+/// * `tensor` - The input tensor with at least 2 dimensions.
+///
+/// # Returns
+///
+/// Tensor with same shape as the input, where the last two dimensions are reduced to size 1,
+/// containing the trace of each matrix.
+///
+pub fn trace<B: Backend, const D: usize>(tensor: Tensor<B, D>) -> Tensor<B, D> {
```
laggui: Could this simply be implemented as a sum of tensor diag?

niklund: Yes for sure that was a big oversight not using diag for the trace. Makes total sense.

Regarding the rank preservation I implemented it like that since the vector_norms also preserved the rank so I figured that was the way to go, but I think it actually makes more sense to reduce the rank. I am completely open to any which way. Also 

> Also, we should assert that D >= 2.

Also agree with this!

Thank you for the feedback, I would appreciate input as to which way of handling the rank is preferable.

laggui: > Regarding the rank preservation I implemented it like that since the vector_norms also preserved the rank

That's a fair assumption I think. But that's only because it reduces a dim down to a scalar, and these reductions currently preserve rank.

For `diag`, the operation effectively "consumes" a dimension to return the diag vector so it makes sense to return `D - 1`. The trace is then a reduction over that, so this will match the current reduction pattern and the output will also be `D - 1` in this case.

niklund: Ah yes that makes total sense. I ended up doing it like that now. Thank you very much for your feedback!

crates/burn-tensor/src/tensor/linalg/diag.rs
```diff
@@ -0,0 +1,46 @@
...
+/// treating them as the matrix dimensions, while preserving all leading batch dimensions.
+///
+/// # Arguments
+///
+/// * `tensor` - The input tensor with at least 2 dimensions.
+///
+/// # Returns
+/// Tensor with rank D - 1, where the last two matrix dimensions are replaced by a single
+/// dimension containing the diagonal elements
+///
```
laggui:
```suggestion
/// A tensor of rank `D - 1`, where the last dimension contains the diagonal elements of the input.
```

crates/burn-tensor/src/tensor/linalg/trace.rs
```diff
@@ -0,0 +1,25 @@
...
+///
+/// # Arguments
+///
+/// * `tensor` - The input tensor with at least 2 dimensions.
+///
+/// # Returns
+///
+/// Tensor with rank D - 1, where the last two matrix dimensions are replaced by a single
+/// dimension containing the the trace for each matrix.
+///
```
laggui:
```suggestion
/// A tensor of rank `D - 1`, where the last dimension contains the sum along diagonals of the input.
```

crates/burn-tensor/src/tensor/linalg/trace.rs
```diff
@@ -0,0 +1,25 @@
+use super::diag;
+use crate::backend::Backend;
+use crate::tensor::Tensor;
+
+/// Computes the trace of the of square matrices.
```
laggui: I think the "square" here is over specified no? You also have tests that work on non square matrices.

laggui (changes requested):
Some minor comments on doc, otherwise LGTM!

laggui (approved):

Issue #3705, opened 2025-09-12
Conv Ouput Geometry issues


crutcher:
**Describe the bug**
This is on 0.18.0; haven't verified on main yet.

1. Some conv-family operations misshandle geometry for some backends.
2. NdArray is broken in the below case for `conv1d`, but `conv2d` has the same issue.
3. There is no standard geometry output contract that conv-family operations are compared to.
4. There is no *documentation* for that contract, just a vague expectation that it matches `pytorch`.

**To Reproduce**
```rust
    #[test]
    fn test_conv1d_edge_case_wgpu() {
        // This passes.
        check_conv1d_edge_case::<burn::backend::Wgpu>();
    }

    #[test]
    fn test_conv1d_edge_case_ndarray() {
        // This dies.
        check_conv1d_edge_case::<burn::backend::NdArray>();
    }

    #[allow(unused)]
    fn check_conv1d_edge_case<B: Backend>() {
        let device = Default::default();

        let input_shape = 3;
        let kernel_shape = 3;
        let stride = 1;
        let dilation = 3;
        let padding = 2;

        let batch_size = 1;
        let in_channels = 4;
        let out_channels = 8;

        let input: Tensor<B, 3> = Tensor::ones([batch_size, in_channels, input_shape], &device);
        let weight: Tensor<B, 3> = Tensor::ones([out_channels, in_channels, kernel_shape], &device);

        // See: https://docs.pytorch.org/docs/stable/generated/torch.nn.Conv1d.html
        // out_shape = floor( ((in_shape + 2*padding - dilation*(kernel_shape-1) - 1) / stride) + 1 )
        //   = floor( ((3 + 2*2 - 3*(3-1) - 1) / 1) + 1 )
        //   = floor( (3 + 4 - 6 - 1) + 1 )
        //   = floor( (7 - 7) + 1 )
        //   = floor( 1 )
        let output = conv1d(
            input,
            weight,
            None,
            ConvOptions {
                stride: [stride],
                padding: [padding],
                dilation: [dilation],
                groups: 1,
            },
        );
        assert_eq!(output.shape().dims(), [batch_size, out_channels, 1])
    }
```

crutcher:
@wingertge I think you have the most context?

crutcher:
I want something like this, for all the conv-like interfaces:
```rust
pub fn conv1d_output_shape(
    input_shape: usize,
    kernel_shape: usize,
    stride: usize,
    dilation: usize,
    padding: usize,
[...]

PR #3709, opened 2025-09-12
More aggressive conv shape functions, docs, maybe_/expect_ call patterns.

crutcher:
See https://github.com/tracel-ai/burn/issues/3705

### Changes

Flush out conv shape calculations; with `maybe_` and `expect_` variants in 1 and N-D; include both `[usize; D]` and `&[usize]` variants.

### DIFF
--- a/crates/burn-tensor/src/tensor/ops/modules/conv.rs
+++ b/crates/burn-tensor/src/tensor/ops/modules/conv.rs
@@ -73,6 +73,20 @@ pub fn maybe_conv1d_output_size(
     if x < 1 { None } else { Some(x) }
 }
 
+fn throw_conv_error(
+    msg: &str,
+    input_shape: &[usize],
+    kernel_shape: &[usize],
+    stride: &[usize],
+    padding: &[usize],
+    dilation: &[usize],
+) -> ! {
+    panic!(
+        "{}:\n input_shape:{:?}\n kernel_shape:{:?}\n stride:{:?}\n padding:{:?}\n dilation:{:?}",
+        msg, input_shape, kernel_shape, stride, padding, dilation,
+    )
+}
+
 /// Predict the output size of a 1D convolution operation.
 ///
 /// This is the ``panic``-ing variant of [`maybe_conv1d_output_size`].
@@ -101,7 +115,7 @@ pub fn maybe_conv1d_output_size(
 ///
 /// # Panics
 ///
-/// If the output size would be <= 0.
+/// On degenerate output sizes.
 pub fn expect_conv1d_output_size(
     input_size: usize,
     kernel_size: usize,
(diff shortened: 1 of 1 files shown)


crutcher:
@laggui

laggui (commented):
Checking for invalid configurations is a good idea.

Regarding the API changes, I think a better approach would be to implement `output_shape(input_shape, weight_shape, options)` to get the full shape, e.g.

```rust
/// Calculate the expected output shape `[batch_size, channels_out, spatial_dims, ..]` for a convolution.
pub fn calculate_conv_output_shape<const N: usize>(
    in_shape: &Shape,
    weight_shape: &Shape,
    options: &ConvOptions<N>,
) -> Shape {
    assert_eq!(weight_shape.num_dims(), N + 2);
    assert_eq!(in_shape.num_dims(), N + 2);

    let kernel_size = &weight_shape.dims[2..];

    let mut out_shape = in_shape.clone();
    // Spatial dims
    for (i, size_i) in out_shape.dims[2..].iter_mut().enumerate() {
        *size_i = calculate_conv_output_size(
            kernel_size[i],
            options.stride[i],
            options.padding[i],
            options.dilation[i],
            *size_i,
        );
    }
    // Output channels
    out_shape.dims[1] = weight_shape.dims[0];

    out_shape
}
```

Eventually, we could make the output shape explicit to the backend impl, e.g:
```rust
pub trait ModuleOps<B: Backend> {
  fn conv2d(
      x: FloatTensor<B>,
      weight: FloatTensor<B>,
      bias: Option<FloatTensor<B>>,
      options: ConvOptions<2>,
      output_shape: Shape,
  ) -> FloatTensor<B>;
}
```

That way, there is a single source of truth for the output shape. We can easily compute the complete output shape before calling the backend op. But that will impact some other APIs atm.

crutcher:
@laggui I'm down to provide an api for D2 + conv<D>; but I don't want to lose the raw computation.

laggui:
> @laggui I'm down to provide an api for D2 + conv; but I don't want to lose the raw computation.

Not sure I follow, what do you mean by raw computation?

antimora:
@crutcher, might have missed a notification. Still valid?

crutcher:
@laggui, @antimora I did miss the update.

`raw computation`; I want to be able to explicitly re-use the *exact same* shape calculations in library code that the `conv` family uses; so that we can tie contracts which depend upon those shapes to `conv` without those libraries being forced to try and re-implement that logic.

laggui:
> `raw computation`; I want to be able to explicitly re-use the _exact same_ shape calculations in library code that the `conv` family uses; so that we can tie contracts which depend upon those shapes to `conv` without those libraries being forced to try and re-implement that logic.

That makes sense! We'd keep the `calculate_conv_output_shape` public anyway.

@crutcher see also my PR that introduces more `Shape` manipulations / calculations #3845. I will follow this up with some changes for the conv output shape similar to my previous comment (without the explicit shape as input).

PR #3740, opened 2025-09-17
Move nn components to `burn-nn`

laggui:
### Checklist

### Changes

Moved `burn-core/src/nn` to `burn-nn` crate.

- Re-export some paths to keep backward compat (e.g., `Initializer` was moved to `burn-core::module` but is re-exported as part of `burn-nn`)
- `nn` module is always available, but we could possibly add it under a feature flag
- Added `SimpleLinear` module for testing in `burn-core` so it stays independent
- Moved module quantize tests for concrete nn modules (linear, transformer) to `burn-nn` integration tests
- Updated the publish workflows


nathanielsimard (approved):
Having the optimizers in `burn-core` feels a bit strange. Maybe we could extract them in their own crate in a following PR. What do you think?

laggui:
I agree. I was thinking of moving them to `burn-train` initially, but it should be independent of the other training utils if you don't need them.

I think it makes sense to introduce a `burn-optim` instead.

Issue #3605, opened 2025-08-23
`PyTorchFileRecorder` is unable to import snapshots from older pickle protocol versions


dcrewi:
**Relevant versions**

- burn-import version 0.18
- candle-core version 0.8.4

**What I tried**

```rust
    let path = "…";
    let load_args = LoadArgs::new(path.into());
    let record = PyTorchFileRecorder::<FullPrecisionSettings>::default()
        .load(path.into(), &device)
        .expect("Failed to decode model");
```

**Error message***

```
thread 'main' panicked at src/bin/main.rs:28:10:
Failed to decode model: DeserializeError("Candle Tensor error: invalid Zip archive: Could not find central directory end")
```

**This is not a zip file**

Looking at the hex dump of my `.pth`, the first two bytes are `80 02`. A zip file should start with `50 4B`.

**Implication**

According to [python docs](https://docs.python.org/3/library/pickle.html#data-stream-format), there are multiple pickle protocol versions and not all of them are zipfile-based. The magic bytes `80 02` indicate protocol version 2. Python2 can't load/save from/to a version higher than 2, so I'm guessing these weights were saved using python2. In any case, burn can't import them today.

**Workaround**

I loaded and re-saved the model using torch on python3. I was then able to import those weights using burn. I would argue that shouldn't be necessary and the error message is confusing.

antimora:
Currently we are relying on Candle's facility to load pt files and Candle doesn't support this: https://github.com/huggingface/candle/issues/2666

We plan to bring pt loading in-house via https://github.com/tracel-ai/burn/issues/3597. Perhaps at this we will start making fixes. Otherwise we are dependent on downstream fixes and we can't fix it now.

antimora:
https://github.com/tracel-ai/burn/pull/3741 will fix

PR #3741, opened 2025-09-17
Add PytorchStore for optimized model loading and in-house pickle reader

antimora:
This PR introduces `PytorchStore` as a modern replacement for `PytorchFileRecorder`, bringing significant memory optimizations and enhanced functionality for PyTorch model loading.

### Checklist
- [x] Confirmed that `cargo run-checks` command has been executed
- [x] Made sure the book is up to date with changes in this PR

### Related Issues/PRs
- Fixes: https://github.com/tracel-ai/burn/issues/3597
- Fixes: https://github.com/tracel-ai/burn/issues/3605
- Continuation of: [#3666](https://github.com/tracel-ai/burn/pull/3666)

#### 1. In-House Pickle Reader Implementation
- Brings pickle reader functionality in-house, removing dependency on Candle's implementation
- Bypasses Candle's tensor type constraints
- Enables direct bug fixes and feature extensions

#### 2. Enhanced Pickle Format Support
- Extended type support including boolean tensors
- Added support for legacy PyTorch file format (e.g., `ResNet18.pth`)
- Implemented lazy loading for both modern and legacy formats

#### 3. Core Infrastructure Improvements
- **Fixed Adapter transformation order**: Now properly applies adaptations after `TensorSnapshot` to correctly handle container types
- **Refactored `PytorchFileRecorder`**: Updated to use the new in-house pickle reader
- **Introduced `PytorchStore`**: New optimized loading mechanism with memory-efficient design
-  **Critical Architecture Fix**: Eliminated unnecessary module cloning that previously caused memory to triple during loading. Implemented efficient unsafe pointer operations for direct memory management

#### 4. Testing and Benchmarking
- Added comprehensive pickle format testing suite
- Updated benchmarks to compare all loading methods across multiple backends

### Testing

1. All existing pytorch-tests pass in `burn-import`
2. Extensive new test coverage for pickle functionality
3. Memory consumption benchmarks validate optimization improvements

### Benchmarks

Testing performed with a 312MB model (20 layers) across 3 backends:

Maximum Memory Usage Results

| Backend | Method | Max Memory | Ratio | Status |
|---------|--------|------------|-------|--------|
| **NdArray (CPU)** | | | | |
| | SafetensorsStore (new) | 361 MB | 1.16x | ✅ Optimized |
| | SafetensorsFileRecorder (old) | 655 MB | 2.10x | ❌ High memory |
| | PytorchStore (new) | 361 MB | 1.16x | ✅ Optimized |
| | PyTorchFileRecorder (old) | 361 MB | 1.16x | ✅ Baseline |
| **WGPU (GPU)** | | | | |
| | SafetensorsStore (new) | 34 MB | 0.11x | ✅ GPU efficient |
| | SafetensorsFileRecorder (old) | 655 MB | 2.10x | ❌ High memory |
| | PytorchStore (new) | 34 MB | 0.11x | ✅ GPU efficient |
| | PyTorchFileRecorder (old) | 344 MB | 1.10x | ⚠️ Higher than new |
| **LibTorch** | | | | |
| | SafetensorsStore (new) | 34 MB | 0.11x | ✅ Most efficient |
| | SafetensorsFileRecorder (old) | 655 MB | 2.10x | ❌ High memory |
| | PytorchStore (new) | 34 MB | 0.11x | ✅ Most efficient |
| | PyTorchFileRecorder (old) | 344 MB | 1.10x | ⚠️ Higher than new |

Key Improvements

1. Memory Efficiency Gains
- CPU backends: Reduced memory usage to 1.16x model size (down from 2.1x)
- GPU backends: Achieved remarkable 0.11x model size (only 34MB for 312MB model)
- Overall: 44% memory reduction on CPU backends

2. Cross-Format Consistency
- Both `SafetensorsStore` and `PytorchStore` show identical memory usage patterns
- Proves optimization effectiveness across different storage formats

3. GPU Memory Optimization
- New stores leverage GPU memory management effectively
- Uses only ~10% of model size in GPU memory

The new `PytorchStore` and updated `SafetensorsStore` provide substantial memory improvements while maintaining or improving load performance. These optimizations make Burn suitable for memory-constrained environments and enable loading of larger models on the same hardware.

benchmark details: [benchmark-results.txt](https://github.com/user-attachments/files/22417686/benchmark-results.txt)

### DIFF
--- a/crates/burn-store/src/adapter.rs
+++ b/crates/burn-store/src/adapter.rs
@@ -8,6 +8,7 @@ use crate::TensorSnapshot;
 
 use alloc::boxed::Box;
 use alloc::rc::Rc;
+use alloc::string::String;
 use alloc::string::ToString;
 use alloc::vec;
 


crates/burn-store/src/traits.rs
```diff
@@ -1,93 +1,104 @@
...
+            // Read the module out of self (moves it, leaving self in undefined state)
+            let module = core::ptr::read(self as *const Self);
+
+            // Map the module to create a new one with updated tensors
+            let new_module = module.map(&mut applier);
+
+            // Write the new module back to self
+            core::ptr::write(self as *mut Self, new_module);
+        }
+
```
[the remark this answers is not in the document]
antimora: Module does not have default implementation so it can't be done.

crates/burn-store/src/adapter.rs
```diff
@@ -1,201 +1,208 @@
...
-    ///
-    /// This provides a convenient way to rename multiple keys systematically.
-    #[cfg(feature = "std")]
-    fn key_remapper(&self) -> Option<KeyRemapper> {
-        None
+    // Define constants that match the stringified struct names
+    pub const LINEAR: &str = stringify!(Linear);
+    pub const BATCH_NORM: &str = stringify!(BatchNorm);
+    pub const LAYER_NORM: &str = stringify!(LayerNorm);
+    pub const GROUP_NORM: &str = stringify!(GroupNorm);
```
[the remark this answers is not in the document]
antimora: The point is that we want to keep them in sync

crates/burn-store/src/traits.rs
```diff
@@ -1,93 +1,104 @@
...
+    fn collect(
+        &self,
+        filter: Option<PathFilter>,
+        adapter: Option<Box<dyn ModuleAdapter>>,
+    ) -> Vec<TensorSnapshot> {
+        let mut collector = Collector::new(filter, adapter);
         self.visit(&mut collector);
-        collector.tensors
+        collector.into_tensors()
     }
```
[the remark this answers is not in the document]
antimora: not concerned since it's not releassed

antimora:
@crutcher, your input is appreciated. With this we can now read legacy pytorch file formats. The public API does not depart from what you have seen in SafetensorsStore, so it should be familiar.

crates/burn-store/src/adapter.rs
```diff
@@ -1,201 +1,212 @@
...
-/// Handles differences between Burn and PyTorch conventions:
-/// - Transposes linear layer weights (Burn: [in, out], PyTorch: [out, in])
-/// - Renames normalization parameters (Burn: gamma/beta, PyTorch: weight/bias)
+/// Handles:
+/// - Linear layer weight transposition (Burn: [in, out] → PyTorch: [out, in])
+/// - Normalization parameter renaming (gamma → weight, beta → bias)
 #[derive(Debug, Clone, Default)]
 pub struct BurnToPyTorchAdapter;
 
 impl ModuleAdapter for BurnToPyTorchAdapter {
```
laggui, line 88: I realize this was added as part of the previous PR, but going through the changes here I I'd recommend not including it as part of an official Burn crate as the maintenance burden would fall on us.

If you want to provide this functionality to power users, the best approach is to create a separate crate yourself that you can publish to provide these adapters.

crates/burn-store/src/pytorch/pickle_reader.rs
laggui, line 1: This obviously comes from candle's pickle loader, and I think it would be useful to point to the source (both for attribution, and for reference).

If any other code/files were added as part of the migration to maintain pickle loading on our end it should probably be clearly indicated as well.

laggui (commented):
I have a couple of high-level comments for now

crates/burn-store/src/pytorch/pickle_reader.rs
antimora: Attributions were added to Notices.md but I can add them here too. I should add that so much was changed here too. I can say originally this was copied from Candle project.

laggui: Yeah I saw the notice, but I think a local attribution can still be relevant. Esp. here the pickle reader that clearly builds upon the source (with some better separations, and extended compatibility).

crates/burn-store/src/pytorch/reader.rs
```diff
@@ -0,0 +1,963 @@
...
+    let bytes_read = file.read(&mut header)?;
+    file.seek(std::io::SeekFrom::Start(0))?;
+
+    // Only check for legacy format if we have enough bytes
+    let is_legacy_format = bytes_read >= 15
+        && header[0] == 0x80
+        && header[1] == 0x02
+        && header[2] == 0x8a
+        && header[3] == 0x0a
+        && header[14] == 0x2e;
```
laggui: For robustness, should we also check that bytes [4..14] correspond to the magic number `0x1950a86a20f9469cfc6c`?

Also, would be good to point to the source: https://github.com/pytorch/pytorch/blob/main/torch/serialization.py#L65

crates/burn-store/src/pytorch/reader.rs
```diff
@@ -0,0 +1,963 @@
...
+    tensors: &mut HashMap<String, TensorSnapshot>,
+) {
+    match obj {
+        Object::Dict(dict) => {
+            for (key, value) in dict {
+                let new_path = if path.is_empty() {
+                    key.clone()
+                } else {
+                    format!("{}.{}", path, key)
+                };
```
laggui: Could reuse a mut vec of str instead:

```rust
/// Recursively extract tensors from an object
fn extract_tensors_recursive<'a>(
    obj: &'a Object,
    path: &mut Vec<&'a str>,
    tensors: &mut HashMap<String, TensorSnapshot>,
) {
    match obj {
        Object::Dict(dict) => {
            for (key, value) in dict {
                path.push(key);
                extract_tensors_recursive(value, path, tensors);
                path.pop();
            }
        }
        Object::TorchParam(snapshot) => {
            // The TensorSnapshot already contains the data loading closure
            tensors.insert(path.join("."), snapshot.clone());
        }
        _ => {}
    }
}
```

No new string allocations at each recursion step, only at the leafs when joining.

laggui (approved):
Only have a couple minor comments, approving in advance

crates/burn-store/src/adapter.rs
antimora: Removed duplicated code and minimized the logic reducing the maintenance just to inverse operation:

image

If this becomes unmanageable, we can remove it.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -0,0 +1,1153 @@
...
+            0x86 => Ok(Self::Tuple2),
+            0x87 => Ok(Self::Tuple3),
+            0x88 => Ok(Self::NewTrue),
+            0x89 => Ok(Self::NewFalse),
+            b'h' => Ok(Self::BinGet),
+            b'j' => Ok(Self::LongBinGet),
+            b's' => Ok(Self::SetItem),
+            b'u' => Ok(Self::SetItems),
+            b'}' => Ok(Self::EmptyDict),
+            b'd' => Ok(Self::EmptyDict),
```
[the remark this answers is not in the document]
antimora: Thanks. I have this and other two missing ops.

PR #3743, opened 2025-09-17
Add torch.cross like functionality support

### LINKED ISSUE #3732: Add torch.cross
### Feature description

Add support for https://docs.pytorch.org/docs/stable/generated/torch.cross.html

### Feature motivation

I was porting some preprocessing of a pytorch model and it used cross, which Burn did not have.

SinanGncgl:
### Checklist

  - Note: Some clippy error/warnings which this PR did not touch.

### Related Issues/PRs

Fixes #3732

### Changes

Added cross product operation for tensors with:
- Support for arbitrary dimensions and broadcasting
- Implementations across all backends (candle, cubecl, ndarray)
- torch-compatible behavior

### Testing

Implementation verified with test suite covering:
- 3D vector cross products
- Cross products along arbitrary dimensions
- Broadcasting scenarios
- Higher dimensional tensors (4D)

### DIFF
--- a/crates/burn-autodiff/src/tests/cross.rs
+++ b/crates/burn-autodiff/src/tests/cross.rs
@@ -4,20 +4,6 @@ mod tests {
     use burn_tensor::might_panic;
     use burn_tensor::{TensorData, Tolerance, ops::FloatElem};
 
-    // Helper to compute expected cross product for 2-D (N × 3) tensors.
-    fn manual_cross(a: &[[f32; 3]], b: &[[f32; 3]]) -> Vec<[f32; 3]> {
-        a.iter()
-            .zip(b.iter())
-            .map(|(x, y)| {
-                [
-                    x[1] * y[2] - x[2] * y[1],
-                    x[2] * y[0] - x[0] * y[2],
-                    x[0] * y[1] - x[1] * y[0],
-                ]
-            })
-            .collect()
-    }
-
     #[test]
     fn backward_basic() {
         let device = Default::default();
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -3,12 +3,11 @@ use alloc::vec::Vec;
 use crate::alloc::borrow::ToOwned;
 
 use crate::{
-    AsIndex, BasicOps, Bool, Distribution, Element, ElementConversion, Float, Int, Shape, Tensor,
+    BasicOps, Bool, Distribution, Element, ElementConversion, Float, Int, Shape, Tensor,
     TensorKind,
     backend::Backend,
     check,
     check::TensorCheck,
-    indexing::canonicalize_dim,
     ops::{Device, IntTensor},
 };
 use crate::{DType, TensorPrimitive};
@@ -2026,25 +2025,6 @@ where
         check!(TensorCheck::matmul(&self, &other));
         Tensor::new(K::matmul(self.primitive, other.primitive))
     }
-
-    /// Computes the cross product of `self` and another tensor along a given dimension.
-    ///
-    /// Both `self` and `other` **must have size 3** along the specified `dim`,
-    /// because the cross product is only defined in three-dimensional space.
-    ///
-    /// # Arguments
-    ///
-    /// * `other` - The other tensor to take the cross product with.
-    /// * `dim`   - The dimension along which to compute the cross product.
-    ///
-    /// # Returns
-    ///
-    /// A tensor containing the cross product of `self` and `other` along `dim`.
-    pub fn cross<Dim: AsIndex>(self, other: Tensor<B, D, K>, dim: Dim) -> Tensor<B, D, K> {
-        let dim = canonicalize_dim(dim, D, false);
-        check!(TensorCheck::cross(&self, &other, dim));
-        Tensor::new(K::cross(self.primitive, other.primitive, dim))
-    }
 }
 
 impl<B, K> Tensor<B, 1, K>
@@ -3317,22 +3297,6 @@ where
     /// C = AB
     /// ```
     fn matmul(lhs: Self::Primitive, rhs: Self::Primitive) -> Self::Primitive;
-
-    /// Computes the cross product of two tensors along a given dimension.
-    ///
-    /// Both `lhs` and `rhs` **must have size 3** along the specified `dim`,
-    /// because the cross product is only defined in three-dimensional space.
-    ///
-    /// # Arguments
-    ///
-    /// * `lhs` - The left-hand-side tensor.
-    /// * `rhs` - The right-hand-side tensor.
-    /// * `dim` - The dimension along which to compute the cross product.
-    ///
-    /// # Returns
-    ///
-    /// A tensor containing the cross product of `lhs` and `rhs` along `dim`.
-    fn cross(lhs: Self::Primitive, rhs: Self::Primitive, dim: usize) -> Self::Primitive;
 }
 
 impl<B: Backend> Numeric<B> for Int {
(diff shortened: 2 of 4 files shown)


crates/burn-cubecl/src/kernel/cross.rs
```diff
@@ -0,0 +1,89 @@
...
+    let cube_count = calculate_cube_count_elemwise(num_vectors, cube_dim);
+
+    unsafe {
+        cross_kernel::launch_unchecked::<E, R>(
+            &lhs.client,
+            cube_count,
+            cube_dim,
+            linear_view(&lhs, &VECTOR_SIZE),
+            linear_view_ref(&rhs, &lhs, &VECTOR_SIZE),
+            linear_view(&output, &VECTOR_SIZE),
```
laggui: You didn't need to follow Copilot's advice here 😄 passing &3 is fine. Or you could use `let line_size = ..` and `&line_size`.

But more importantly here, I believe a vectorization of 3 is not supported right now on any cubecl runtime. 

Did you execute the tests locally? Probably doesn't give the expected results, might not even compile on CUDA due to alignment issues.

Alternative: you could use a line size of 4 and processes multiple 3-element vectors for a single unit of work.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -602,6 +602,62 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                let rhs = rhs_id.map(|id| checkpointer.retrieve_node_output(id));
+
+                binary::<B, _, _>(
+                    ops.parents,
+                    ops.node,
+                    grads,
+                    |grad| B::float_cross(grad, rhs.unwrap(), dim),
+                    |grad| {
+                        let cross_grad_lhs = B::float_cross(grad, lhs.unwrap(), dim);
+                        B::float_neg(cross_grad_lhs)
```
laggui: Wouldn't this be equivalent to:

```rust
B::float_cross(lhs.unwrap(), grad, dim)
```

We should add a test for the backward

SinanGncgl: Exactly, thanks nice one.

crates/burn-cubecl/src/kernel/cross.rs
SinanGncgl: Yes, but seems like there is an issue with fusion, thanks for the suggestion.

antimora:
Rebased

crates/burn-autodiff/src/tests/cross.rs
```diff
@@ -0,0 +1,161 @@
...
+                    x[1] * y[2] - x[2] * y[1],
+                    x[2] * y[0] - x[0] * y[2],
+                    x[0] * y[1] - x[1] * y[0],
+                ]
+            })
+            .collect()
+    }
+
+    #[test]
+    fn forward_matches_manual_cross() {
```
laggui: This is not an autodiff test so I would leave it with the other tensor ops tests.

crates/burn-autodiff/src/tests/cross.rs
```diff
@@ -0,0 +1,161 @@
...
+        let a_raw = [[1.0, 4.0, 7.0], [2.0, 5.0, 8.0], [3.0, 6.0, 9.0]];
+        let b_raw = [[9.0, 6.0, 3.0], [8.0, 5.0, 2.0], [7.0, 4.0, 1.0]];
+
+        let a = TestTensor::<2>::from_data(TensorData::from(a_raw), &device);
+        let b = TestTensor::<2>::from_data(TensorData::from(b_raw), &device);
+        // Cross along dim 0. Some backends (for example CubeCL) may not support
+        // cross on non-last dimensions and will intentionally panic with a
+        // message like "Cross product on non-last dimension not yet implemented".
+        // In that case we treat the panic as a skipped test for that backend.
+        use std::panic::{AssertUnwindSafe, catch_unwind};
```
laggui: You could use `burn_tensor::might_panic` to capture the error message just as you implemented with your catch unwind.

```rust
#[might_panic(reason = "Cross product on non-last dimension")]
#[test]
```

crates/burn-ndarray/src/lib.rs
```diff
@@ -35,6 +35,7 @@ mod tests {
     use alloc::format;
     use alloc::vec;
 
+    #[cfg(feature = "std")]
```
laggui: We should not feature gate all ndarray tensor op tests for std 😅

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -2025,6 +2026,22 @@ where
         check!(TensorCheck::matmul(&self, &other));
         Tensor::new(K::matmul(self.primitive, other.primitive))
     }
+
+    /// Computes the cross product of two tensors along a given dimension.
```
laggui: I would also explicitly mention the size 3 requirement for the cross product

crates/burn-tensor/src/tests/ops/cross.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[test]
+    fn test_cross_3d_dim0() {
+        let tensor_1 = TestTensor::<2>::from([[1.0, 0.0], [0.0, 1.0], [0.0, 0.0]]);
+        let tensor_2 = TestTensor::from([[0.0, 1.0], [0.0, 0.0], [1.0, 0.0]]);
+
+        // Some backends (for example CubeCL) may not support cross on non-last
+        // dimensions and will intentionally panic with a message like
+        // "Cross product on non-last dimension not yet implemented". In that
+        // case we treat the panic as a skipped test for that backend.
+        let res = catch_unwind(AssertUnwindSafe(|| tensor_1.cross(tensor_2, 0)));
```
laggui: Same comment for `might_panic` usage. Might need to gate the test to std since it uses `std::panic`

crates/burn-cubecl/src/kernel/cross.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[comptime] num_vectors: u32,
+) {
+    // Each thread processes one 3-element vector
+    let vector_idx = ABSOLUTE_POS;
+
+    if vector_idx >= num_vectors {
+        terminate!();
+    }
+
+    // Calculate base position - each vector occupies 3 contiguous elements
```
laggui: The inputs might not necessarily be contiguous (as per your comment), but since you used `LinearView` this should map to the correct elements for the product.

crates/burn-cubecl/src/kernel/cross.rs
```diff
@@ -0,0 +1,95 @@
...
+fn cross_kernel<E: Float>(
+    lhs: &LinearView<Line<E>>,
+    rhs: &LinearView<Line<E>>,
+    output: &mut LinearView<Line<E>, ReadWrite>,
+    #[comptime] num_vectors: u32,
+) {
+    // Each thread processes one 3-element vector
+    let vector_idx = ABSOLUTE_POS;
+
+    if vector_idx >= num_vectors {
```
laggui: Prefer using `output.is_in_bounds` with linear view so it correctly checks the mapped index.

laggui (commented):
Sorry for the late review, somehow missed that notification 😅

Just a couple of comments and then it should be good

crates/burn-ndarray/src/lib.rs
SinanGncgl: upps, you are absolutely right 🥶

SinanGncgl:
> Sorry for the late review, somehow missed that notification 😅
> 
> Just a couple of comments, and then it should be good

Thanks for all your reviews/comments :)

laggui (approved):
Fixed the last minor things myself 🙂 

Thanks for your contribution!

antimora:
One last comment: we need update burn's tensor section with cross method

Issue #3690, opened 2025-09-09
Slice op with arbitraty steps


notdanilo:
### Feature description

Support slice as described in
https://onnx.ai/onnx/operators/onnx__Slice.html

The current implementation only supports `steps=1` and `start <= end`.
We need to support arbitrary positive and negative steps, and also allow `start > end` for when we want to use `negative steps`.

### Feature motivation

Support more onnx models.

### Suggest a Solution

It was decided that the API will follow `ndarray`'s `s!` syntax.

```rust
tensor.slice(0..4); // start=0, end=4, step=1
tensor.slice(s![0..4; -1]); // start=3, end=-1, step=-1
tensor.slice(s![0..4; 2]); // start=0, end=4, step=2
tensor.slice(s![0..4; -2]); // start=3, end=-1, step=-2
tensor.slice(s![1..4; -1]); // start=2, end=-1, step=-1
```

laggui:
Just echoing from discord:

At the tensor API level it should be easy to add support by extending the slice macro to support steps, e.g.

```rust
// Slice with a step of 2
let slice = tensor.slice(s![0..; 2]); 
```

But then we need to change the backend tensor ops to have stepped ranges instead of just `Range`s, and make sure the implementations support it.

antimora:
I'll give a shot.

notdanilo:
@antimora that's great! I couldn't get any free time during the past weeks, so big thanks for the effort! :) Is the onnx import working for this?

antimora:
> [@antimora](https://github.com/antimora) that's great! I couldn't get any free time during the past weeks, so big thanks for the effort! :) Is the onnx import working for this?

Yes, the ONNX portion should also be working. I added step support for shapes as well. 

<img width="289" height="284" alt="Image" src="https://github.com/user-attachments/assets/eef88f98-f111-41c8-9da4-1d7635fafda2" />

Next I need to fix slice_assign with steps because it's required for slice with steps autodiff

PR #3748, opened 2025-09-19
Slice with steps

antimora:
## Pull Request Overview

This PR implements slice operations with step support for Burn tensors, addressing issue #3690. The feature enables strided slicing and reversing of tensor dimensions, similar to NumPy's advanced slicing capabilities.

### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/3690

### Key Changes

- Extended the `s![]` macro to support step syntax (`s![range;step]`)
- Added `SliceInfo` struct to carry range and step information through the tensor operation pipeline
- Updated all backend implementations to handle stepped slicing (with fallbacks for unsupported backends)

### Testing

New tests and existing tests pass

### DIFF
--- a/crates/burn-tensor/src/tensor/ops/modules/conv.rs
+++ b/crates/burn-tensor/src/tensor/ops/modules/conv.rs
@@ -612,9 +612,9 @@ fn conv1d_weight_grad_no_groups<B: Backend>(
 
     if weight_grad.shape() != weight_shape {
         let slices = vec![
-            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
+            Slice::from(0..weight_shape.dims[0]),
+            Slice::from(0..weight_shape.dims[1]),
+            Slice::from(0..weight_shape.dims[2]),
         ];
         weight_grad = B::float_slice(weight_grad, &slices);
     }
@@ -639,10 +639,10 @@ fn conv2d_weight_grad_no_groups<B: Backend>(
 
     if weight_grad.shape() != weight_shape {
         let slices = vec![
-            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[3] as isize), 1),
+            Slice::from(0..weight_shape.dims[0]),
+            Slice::from(0..weight_shape.dims[1]),
+            Slice::from(0..weight_shape.dims[2]),
+            Slice::from(0..weight_shape.dims[3]),
         ];
         weight_grad = B::float_slice(weight_grad, &slices);
     }
@@ -667,11 +667,11 @@ fn conv3d_weight_grad_no_groups<B: Backend>(
 
     if weight_grad.shape() != weight_shape {
         let slices = vec![
-            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[3] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[4] as isize), 1),
+            Slice::from(0..weight_shape.dims[0]),
+            Slice::from(0..weight_shape.dims[1]),
+            Slice::from(0..weight_shape.dims[2]),
+            Slice::from(0..weight_shape.dims[3]),
+            Slice::from(0..weight_shape.dims[4]),
         ];
         weight_grad = B::float_slice(weight_grad, &slices);
     }
@@ -766,10 +766,10 @@ fn conv2d_weight_grad_groups<B: Backend>(
 
         if kernel_size_1_tmp != kernel_size_1 || kernel_size_2_tmp != kernel_size_2 {
             let slices = vec![
-                Slice::new(0_isize, Some(increment_co as isize), 1),
-                Slice::new(0_isize, Some(increment_ci as isize), 1),
-                Slice::new(0_isize, Some(kernel_size_1 as isize), 1),
-                Slice::new(0_isize, Some(kernel_size_2 as isize), 1),
+                Slice::from(0..increment_co),
+                Slice::from(0..increment_ci),
+                Slice::from(0..kernel_size_1),
+                Slice::from(0..kernel_size_2),
             ];
             weight_grad_tmp = B::float_slice(weight_grad_tmp, &slices);
         }
@@ -845,11 +845,11 @@ fn conv3d_weight_grad_groups<B: Backend>(
             || kernel_size_3_tmp != kernel_size_3
         {
             let slices = vec![
-                Slice::new(0_isize, Some(increment_co as isize), 1),
-                Slice::new(0_isize, Some(increment_ci as isize), 1),
-                Slice::new(0_isize, Some(kernel_size_1 as isize), 1),
-                Slice::new(0_isize, Some(kernel_size_2 as isize), 1),
-                Slice::new(0_isize, Some(kernel_size_3 as isize), 1),
+                Slice::from(0..increment_co),
+                Slice::from(0..increment_ci),
+                Slice::from(0..kernel_size_1),
+                Slice::from(0..kernel_size_2),
+                Slice::from(0..kernel_size_3),
             ];
             weight_grad_tmp = B::float_slice(weight_grad_tmp, &slices);
         }
@@ -889,9 +889,9 @@ fn conv_transpose1d_weight_grad_no_groups<B: Backend>(
     let grad_shape = weight_grad.shape();
     if grad_shape != weight_shape {
         let slices = vec![
-            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
+            Slice::from(0..weight_shape.dims[0]),
+            Slice::from(0..weight_shape.dims[1]),
+            Slice::from(0..weight_shape.dims[2]),
         ];
         weight_grad = B::float_slice(weight_grad, &slices);
     }
@@ -917,10 +917,10 @@ fn conv_transpose2d_weight_grad_no_groups<B: Backend>(
     let grad_shape = weight_grad.shape();
     if grad_shape != weight_shape {
         let slices = vec![
-            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[3] as isize), 1),
+            Slice::from(0..weight_shape.dims[0]),
+            Slice::from(0..weight_shape.dims[1]),
+            Slice::from(0..weight_shape.dims[2]),
+            Slice::from(0..weight_shape.dims[3]),
         ];
         weight_grad = B::float_slice(weight_grad, &slices);
     }
@@ -946,11 +946,11 @@ fn conv_transpose3d_weight_grad_no_groups<B: Backend>(
     let grad_shape = weight_grad.shape();
     if grad_shape != weight_shape {
         let slices = vec![
-            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[3] as isize), 1),
-            Slice::new(0_isize, Some(weight_shape.dims[4] as isize), 1),
+            Slice::from(0..weight_shape.dims[0]),
+            Slice::from(0..weight_shape.dims[1]),
+            Slice::from(0..weight_shape.dims[2]),
+            Slice::from(0..weight_shape.dims[3]),
+            Slice::from(0..weight_shape.dims[4]),
         ];
         weight_grad = B::float_slice(weight_grad, &slices);
     }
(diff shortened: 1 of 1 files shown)


crates/burn-cubecl/src/kernel/index/slice.rs
```diff
@@ -99,3 +99,114 @@ pub(crate) fn slice_on_output<R: CubeRuntime, E: CubeElement>(
 
     output
 }
+
+/// Kernel for slicing with steps
+#[cube(launch_unchecked)]
+fn slice_with_steps_kernel<E: CubePrimitive>(
+    input: &Tensor<E>,
+    output: &mut Tensor<E>,
```
wingertge: These should be `&LinearView<E>` and `&mut LinearView<E, ReadWrite>` respectively to avoid any layout issues. you can construct the launch arg by just using `linear_view(&tensor, &line_size)`.

crates/burn-router/src/ops/op_float.rs
```diff
@@ -514,12 +516,22 @@ impl<R: RunnerChannel> FloatTensorOps<Self> for BackendRouter<R> {
...
+            .iter()
+            .map(|info| {
+                let len = info.range.end - info.range.start;
+                if info.step.unsigned_abs() == 1 {
+                    len
+                } else {
+                    len.div_ceil(info.step.unsigned_abs())
+                }
+            })
+            .collect();
```
wingertge: This shape calculation is repeated a lot, I think it should be a shared utility similar to the convolution output size

antimora: Refactored into a single central function with good amount of tests.

wingertge (commented):
I totally missed the complex indexing in the kernel, so my suggestion for linear view wasn't quite right. It's a bit complicated and uses more advanced cubecl features so I just refactored the kernel myself. I did this slice refactor before but stashed during testing and forgot to re-apply the stash.
Still need to fix the remaining lint issues but looks good otherwise.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1178,10 +1191,19 @@ where
...
+        let slice_infos = ranges.into_slice_infos(self.shape());
 
+        // Extract ranges for validation as an array
+        let ranges: [Range<usize>; D2] = slice_infos
+            .iter()
+            .map(|s| s.range.clone())
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
         check!(TensorCheck::slice::<D, D2>(&self.shape(), &ranges));
```
laggui: We don't need to collect an array of ranges just for the check. We should update the check to use `Slice`s instead, it only checks the start/end bounds.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2638,7 +2660,14 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
         ranges: &[Range<usize>],
```
laggui: We should eventually standardize all slice APIs to work with steps

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2663,10 +2692,21 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
     /// with static dispatch. It is not designed for direct usage by users, and not recommended to import
     /// or use this function directly.
     fn slice_dim(tensor: Self::Primitive, dim: usize, range: &Range<usize>) -> Self::Primitive {
```
laggui: Same (future work)

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -3558,33 +3600,132 @@ pub trait RangeArg {
     fn into_range(self, shape_dim: usize) -> Range<usize>;
 }
 
-impl<T: Into<Slice>> RangeArg for T {
+// Slice already has an into_range method, so we just delegate to it
+// But we need to use the full path to avoid ambiguity
+impl RangeArg for Slice {
```
laggui: Hmm why explicitly implement for each type instead of the blanket implementation?

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -3558,33 +3600,132 @@ pub trait RangeArg {
...
+    /// Converts into a set of SliceInfo structs for backends that support steps
+    fn into_slice_infos(self, shape: Shape) -> [SliceInfo; D2];
 }
 
-impl<const D2: usize, T: Into<Slice>> RangesArg<D2> for [T; D2] {
+impl<const D2: usize, T: Into<Slice> + Clone> RangesArg<D2> for [T; D2] {
     fn into_ranges(self, shape: Shape) -> [Range<usize>; D2] {
         // clamp the ranges to the shape dimensions
         let ranges = self
+            .clone()
```
laggui: Why `+ Clone` when `into_ranges` takes ownership and consumes itself?

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -3558,33 +3600,132 @@ pub trait RangeArg {
...
+            .into_iter()
+            .enumerate()
+            .map(|(i, range)| range.into().to_slice_info(shape.dims[i]))
+            .collect::<Vec<_>>();
+        infos.try_into().unwrap()
+    }
 }
 
-impl<T: Into<Slice>> RangesArg<1> for T {
+impl<T: Into<Slice> + Clone> RangesArg<1> for T {
```
laggui: Same

crates/burn-tensor/src/tensor/api/slice.rs
```diff
@@ -1,11 +1,92 @@
...
 use core::ops::{Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive};
 
+/// Information about a slice operation including range and step.
+#[derive(Clone, Debug, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct SliceInfo {
+    /// The range of indices to select
+    pub range: Range<usize>,
+    /// The step between selected elements (default: 1)
+    pub step: isize,
+}
```
laggui: Instead of introducing a new type, we should use the `Slice` struct which already captures this info with `start`, `end` and now `step` which you have added.

It was initially added for the `s!` slice macro support but it should be the type to represent slices all around.

crates/burn-tensor/src/tensor/api/slice.rs
```diff
@@ -16,11 +97,36 @@ use core::ops::{Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInc
...
+    [] => {
+        compile_error!("Empty slice specification")
+    };
+
+    // Single expression with step
+    [$range:expr; $step:expr] => {
+        {
+            #[allow(clippy::reversed_empty_ranges)]
+            {
+                $crate::SliceWithStep::from_range_and_step($range, $step)
```
laggui: And instead of another `SliceWithStep` struct, we can provide methods on `Slice`, e.g.:
- `Slice::new(start, end, step)`
- `Slice::from_range_stepped(range, step)`

which should cover all supported slice types creation (on top of the existing `From` implementations for all range types).

We can also get rid of the `StepValue` trait, and have the `step: S` require `num_trait::PrimInt` which is implemented for all primitive ints and provides `to_isize()` conversion.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1170,22 +1170,26 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
             }
         }
 
+        // Convert slice_infos to ranges for backward compatibility in State
+        let ranges: Vec<core::ops::Range<usize>> =
+            slice_infos.iter().map(|s| s.range.clone()).collect();
+
```
laggui: This makes the code compile since stepped slices wasn't added to `slice_assign`, but the backward pass needs to place gradients back into the original tensor at the same stepped positions.

The result will be incorrect. We should add an autodiff test specifically for that to make sure the results are correct.

laggui:
Regarding autodiff, here is a test w/ expected result that should pass:

```rust
#[test]
fn should_diff_matmul_with_slice_stepped() {
    use burn_tensor::s;
    let data_1 = TensorData::from([[1.0, 7.0], [100.0, 100.0], [2.0, 3.0], [100.0, 100.0]]);
    let data_2 = TensorData::from([[4.0, 100.0, 7.0, 100.0], [2.0, 100.0, 3.0, 15.0]]);

    let tensor_1 = TestAutodiffTensor::<2>::from_data(data_1, &device).require_grad();
    let tensor_2 = TestAutodiffTensor::from_data(data_2, &device).require_grad();

    let tensor_3 = tensor_1.clone().slice(s![0..;2, 0..2]); // [[1., 7.], [2., 3.]]
    let tensor_4 = tensor_2.clone().slice(s![0..2, 0..;2]); // [[4., 7.], [2., 3.]]
    let tensor_5 = tensor_3.clone().matmul(tensor_4);
    let grads = tensor_5.backward();

    let grad_2 = tensor_2.grad(&grads).unwrap();

    grad_1.to_data().assert_eq(
        &TensorData::from([[11., 5.], [0., 0.], [11., 5.], [0., 0.]]),
        false,
    );
    grad_2.to_data().assert_eq(
        &TensorData::from([[3., 0., 3., 0.], [10., 0., 10., 0.]]),
        false,
    );
}
```

crates/burn-tensor/src/tensor/api/slice.rs
antimora: Removed SliceInfo

crates/burn-tensor/src/tensor/api/slice.rs
antimora: Removed SliceWithStep

crates/burn-autodiff/src/ops/tensor.rs
antimora: Panicking when step != 1 for autodiff

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2585,16 +2644,21 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
...
+    fn slice(tensor: Self::Primitive, slices: &[Slice]) -> Self::Primitive;
 
-    ///  Assigns the given value to the tensor elements corresponding for the given ranges.
+    ///  Assigns the given value to the tensor elements corresponding to the given slice infos.
     ///
     /// # Arguments
     ///
     /// * `tensor` - The tensor.
-    /// * `ranges` - The ranges of the elements to select.
+    /// * `slice_infos` - The slice information for elements to assign.
```
laggui:
```suggestion
    /// * `slices` - The slice information for elements to assign.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2618,12 +2682,12 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
     /// # Arguments
     ///
     /// * `tensor` - The tensor.
-    /// * `ranges` - The ranges of the elements to fill.
+    /// * `slice_infos` - The slice information containing ranges and steps for each dimension.
```
laggui:
```suggestion
    /// * `slices` - The slice information containing ranges and steps for each dimension.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2633,25 +2697,28 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
...
 
-    /// Slices the tensor along a given dimension with the specified range.
+    /// Slices the tensor along a given dimension.
     ///
     /// # Arguments
     ///
     /// * `tensor` - The tensor to slice.
     /// * `dim` - The dimension along which to slice.
-    /// * `range` - The range of indices to slice along the specified dimension.
+    /// * `slice_info` - The slice information containing range and step.
```
laggui, line 2721:
```suggestion
    /// * `slice` - The slice information containing range and step.
```

crates/burn-tensor/src/tensor/ops/bool_tensor.rs
```diff
@@ -124,12 +124,12 @@ pub trait BoolTensorOps<B: Backend> {
     /// # Arguments
     ///
     /// * `tensor` - The tensor.
-    /// * `ranges` - The ranges to get the values from.
+    /// * `slice_infos` - The slice information including ranges and steps.
```
laggui:
```suggestion
    /// * `slices` - The slice information including ranges and steps.
```

crates/burn-tensor/src/tensor/ops/int_tensor.rs
```diff
@@ -78,12 +78,12 @@ pub trait IntTensorOps<B: Backend> {
     /// # Arguments
     ///
     /// * `tensor` - The tensor.
-    /// * `indices` - The indices.
+    /// * `slice_infos` - The slice information including ranges and steps.
```
laggui:
```suggestion
    /// * `slices` - The slice information including ranges and steps.
```

crates/burn-tensor/src/tensor/ops/modules/conv.rs
```diff
@@ -929,16 +945,14 @@ fn conv_transpose3d_weight_grad_no_groups<B: Backend>(
...
-                0..weight_shape.dims[3],
-                0..weight_shape.dims[4],
-            ],
-        );
+        let slice_infos = vec![
+            Slice::new(0_isize, Some(weight_shape.dims[0] as isize), 1),
+            Slice::new(0_isize, Some(weight_shape.dims[1] as isize), 1),
+            Slice::new(0_isize, Some(weight_shape.dims[2] as isize), 1),
+            Slice::new(0_isize, Some(weight_shape.dims[3] as isize), 1),
+            Slice::new(0_isize, Some(weight_shape.dims[4] as isize), 1),
```
laggui: Could have used `Slice::from(range)` or `range.into()` but that's also fine

crates/burn-tensor/src/tensor/ops/qtensor.rs
```diff
@@ -278,17 +277,17 @@ pub trait QTensorOps<B: Backend> {
...
     ) -> QuantizedTensor<B>;
 
-    /// Select tensor elements corresponding for the given ranges.
+    /// Select tensor elements corresponding for the given slice infos.
     ///
     /// # Arguments
     ///
     /// * `tensor` - The tensor to select from.
-    /// * `ranges` - The ranges to select.
+    /// * `slice_infos` - The slice infos specifying ranges and steps.
```
laggui:
```suggestion
    /// * `slices` - The slice infos specifying ranges and steps.
```

crates/burn-tensor/src/tensor/ops/tensor.rs
```diff
@@ -473,12 +473,12 @@ pub trait FloatTensorOps<B: Backend> {
     /// # Arguments
     ///
     /// * `tensor` - The tensor to select from.
-    /// * `ranges` - The ranges to select.
+    /// * `slice_infos` - The slice information including ranges and steps.
```
laggui:
```suggestion
    /// * `slices` - The slice information including ranges and steps.
```

PR #3767, opened 2025-09-22
`trunc`, `fmod` and `Mod` ONNX ops

antimora:
This PR implements support for the `trunc`, `fmod`, and `Mod` ONNX operations, adding floating-point modulo operations and truncation functionality to the Burn tensor library.

### Checklist

### Related Issues/PRs

- https://github.com/tracel-ai/burn/issues/3768
- https://github.com/tracel-ai/burn/issues/3769
- https://github.com/tracel-ai/burn/issues/3412 (Yolov10 model)

### Changes

- Adds `trunc()` method that truncates tensor elements toward zero (positive values like floor, negative values like ceil)
- Adds `fmod()` and `fmod_scalar()` methods for IEEE 754 compliant floating-point remainder operations where the result sign follows the dividend
- Implements ONNX `Mod` operator with support for both C-style fmod (when `fmod=1`) and Python-style remainder (when `fmod=0`) behaviors

### Testing

New unit tests


antimora:
no-std-tests are failing due to https://github.com/tracel-ai/cubecl/issues/910

crates/burn-import/onnx-tests/tests/mod/mod_broadcast.py
```diff
@@ -0,0 +1,69 @@
...
+    test_y = np.array([[[[3.0, 3.0, 3.0, 3.0],
+                         [3.0, 3.0, 3.0, 3.0],
+                         [3.0, 3.0, 3.0, 3.0]]],
+                       [[[4.0, 4.0, 4.0, 4.0],
+                         [4.0, 4.0, 4.0, 4.0],
+                         [4.0, 4.0, 4.0, 4.0]]]]).astype(np.float32)
+
+    # Broadcast x to match y's shape and apply fmod
+    test_x_broadcast = np.broadcast_to(test_x, test_y.shape)
+    result = np.fmod(test_x_broadcast, test_y)
```
laggui: When creating an ONNX graph manually, it's probably more explicit to use the session evaluator now that it's available to us. No mismatches in expectation (and contributors can see this as a reference).

crates/burn-import/onnx-tests/tests/mod/mod_broadcast_remainder.py
```diff
@@ -0,0 +1,64 @@
...
+        producer_name='onnx-tests',
+        opset_imports=[helper.make_operatorsetid("", 16)]
+    )
+
+    # Save the model
+    onnx_name = "mod_broadcast_remainder.onnx"
+    onnx.save(model_def, onnx_name)
+    print(f"Finished exporting model to {onnx_name}")
+
+    # Test with NumPy to show expected results
```
laggui: Same

crates/burn-import/onnx-tests/tests/mod/mod_fmod.py
```diff
@@ -0,0 +1,60 @@
...
+        producer_name='onnx-tests',
+        opset_imports=[helper.make_operatorsetid("", 16)]
+    )
+
+    # Save the model
+    onnx_name = "mod_fmod.onnx"
+    onnx.save(model_def, onnx_name)
+    print(f"Finished exporting model to {onnx_name}")
+
+    # Test with NumPy to show expected results
```
laggui: Same

crates/burn-import/onnx-tests/tests/mod/mod_remainder.py
```diff
@@ -0,0 +1,60 @@
...
+        producer_name='onnx-tests',
+        opset_imports=[helper.make_operatorsetid("", 16)]
+    )
+
+    # Save the model
+    onnx_name = "mod_remainder.onnx"
+    onnx.save(model_def, onnx_name)
+    print(f"Finished exporting model to {onnx_name}")
+
+    # Test with NumPy to show expected results
```
laggui: Same

crates/burn-tensor/src/tensor/api/fmod.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        // When divisor is infinity and dividend is finite:
+        // - quotient is 0, truncated is 0
+        // - but 0 * infinity = NaN, which is wrong
+        // We need to handle this case by replacing NaN with 0 when appropriate
+
+        // Check if the product is NaN due to 0 * inf
+        let is_zero_times_inf = truncated
+            .equal_elem(0.0)
+            .bool_and(other.is_finite().bool_not());
```
laggui: The check here for `is_finite().bool_not()` will match +/- inf AND NaN (which it should not according to your comments/docs. Probably meant to use `is_inf`?

crates/burn-tensor/src/tensor/api/trunc.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// ```
+    pub fn trunc(self) -> Self {
+        // Get the sign of each element: 1 for positive, -1 for negative, 0 for zero
+        let sign = self.clone().sign();
+
+        // Take absolute value and apply floor
+        let abs_floored = self.abs().floor();
+
+        // Restore the original sign
+        abs_floored * sign
```
laggui: According to the `trunc` definition the signed zero should also be preserved. Here the zero sign is lost.

Also matches the other intrinsics def like CUDA:
image


```
truncate(x) = ⌊x⌋ if x ≥ 0, and ⌈x⌉ if x < 0.
```

A more accurate impl might be:
```rust
let is_negative = self.clone().lt_elem(0.0);
let floored = self.clone().floor();
let ceiled = self.clone().ceil();
floored.mask_where(is_negative, ceiled)
```

laggui (changes requested):
Both trunc and mod are intrinsics in, we can provide a default impl like your but should probably allow backends to override the impl.

Then we can add the support for cubecl backends. Doesn't need to be implemented in this PR, but we should at least make the definition for them to be implemented at the backend level when the time comes.

antimora:
@laggui fixed per your feedback. Thank you.

Issue #3683, opened 2025-09-08
VecMat fusion autotune bug


laggui:
Running wgpu tests fails due to vecmat fusion autotune:

```sh
cargo test  --lib --bins --examples -p burn-wgpu --color=always --release
```

```
  thread 'tests::cube_fusion::autodiff::f32_ty::ad_transpose::tests::should_diff_swap_dims' panicked at /home/agent/.cargo/git/checkouts/cubecl-058c47895211d464/5a4ad7f/crates/cubecl-runtime/src/tune/local.rs:155:26:
  Should run when selected by autotune.: Unknown("RunnerError(LaunchError(Unable to launch matmul because the config is invalid: \"Lhs and Rhs must have same line size, got lhs=1 and rhs=2\"\n))")
```

When running the failing test `should_diff_swap_dims` only, from a clean cache, it passes (but if you run it using the current cache it will fail, as expected):

```sh
cargo test  --lib --bins --examples -p burn-wgpu --color=always --release should_diff_swap_dims
```

And if we remove (comment out) the `SimpleVecMat` and `DoubleVecMat` tunables when running all the tests, there are no issues.

https://github.com/tracel-ai/burn/blob/ff5f34e0f8b56b26a07e23292afde039983f4670/crates/burn-cubecl-fusion/src/matmul/tune.rs#L84-L89

It looks like the vecmat are selected by a previous test during autotune, so when running `should_diff_swap_dims` it is selected as the fastest but fails to execute at runtime due to line size mismatch.

wingertge:
It seems to be introduced by this PR, or at least the error doesn't happen before. Since the PR specifically mentions changing the swapped dimension logic it seems related.
<https://github.com/tracel-ai/burn/pull/3627>

PR #3775, opened 2025-09-23
Feat/multi streams 

nathanielsimard:
* Add a new multi-streams strategy for wgpu
* Enable multi-streams for cuda and hip


laggui:
GPU CI failure is due to the autotune issue that we discussed on discord 😅 

#3683

PR #3786, opened 2025-09-25
Add outer and outer_batch ops in linalg

huy209vn:
Pull Request
Checklist

 [v] Confirmed that cargo run-checks command has been executed.

[v] Made sure the book is up to date with changes in this PR.

Related Issues/PRs

Contributes to [#1538](https://github.com/tracel-ai/burn/issues/1538)
 — Linear algebra operations.

Changes

Added linalg::outer(x, y) to compute the outer product of two 1-D vectors.

Shapes: (m,) ⊗ (n,) → (m, n).

Equivalent to torch.outer / numpy.outer.

Added linalg::outer_batch(x, y) for batched outer products of 2-D inputs.

Shapes: (batch, m) ⊗ (batch, n) → (batch, m, n).

Each pair (x[b], y[b]) produces one (m, n) matrix.

Updated Linalg Functions table in the Book with both new ops.

Integrated new ops into burn-tensor module exports and test harness.

Testing

Added outer tests covering:

Correctness vs expected matrices

Shape asymmetry ((m,n) vs (n,m))

Zero vectors (left/right)

Negative values and integer inputs

Equivalence with reshape + matmul

Vector identity (u ⊗ v)w == u(v⋅w)

Length-one vectors

Large magnitude stability (no overflow)

NaN propagation

Added outer_batch tests covering:

Basic correctness with small batches

Shape checks (batch, m, n)

Zero batch inputs

Positive/negative value propagation

Integer batch inputs

Consistency with looping over outer

NaN propagation within batches

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -384,11 +384,13 @@ strategies.
 
 ## Linalg Functions
 
-| Burn API                               | PyTorch Equivalent                        |
-|----------------------------------------|-------------------------------------------|
-| `linalg::vector_norm(tensors, p, dim)` | `torch.linalg.vector_norm(tensor, p, dim) |
-| `linalg::diag(tensor)`                 | `torch.diag(tensor)`                      |
-| `linalg::trace(tensor)`                | `torch.trace(tensor)`                     |
+| Burn API                                   | PyTorch Equivalent                               |
+|--------------------------------------------|--------------------------------------------------|
+| `linalg::vector_norm(tensors, p, dim)`     | `torch.linalg.vector_norm(tensor, p, dim)`       |
+| `linalg::diag(tensor)`                     | `torch.diag(tensor)`                             |
+| `linalg::trace(tensor)`                    | `torch.trace(tensor)`                            |
+| `linalg::outer(x, y)`                      | `torch.outer(x, y)`                              |
+| `linalg::outer_batch(x, y)`                | `torch.einsum("bi,bj->bij", x, y)` (no direct op)|
 
 ## Displaying Tensor Details
 
--- a/crates/burn-tensor/src/tensor/linalg/outer.rs
+++ b/crates/burn-tensor/src/tensor/linalg/outer.rs
@@ -1,0 +1,43 @@
+use crate::backend::Backend;
+use crate::tensor::{BasicOps, Tensor};
+use crate::{Numeric, Shape};
+/// Computes the outer product of two 1D tensors (vectors).
+///
+/// Given `x` of shape `(m,)` and `y` of shape `(n,)`, returns a tensor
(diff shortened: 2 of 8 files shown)


laggui (changes requested):
Thanks for the addition! 

The implementation LGTM, but instead of having separate functions for vectors / batch I would only have `linalg::outer`.

We can check the input rank and adjust the shape accordingly. Panic if rank > 2, and clearly state this in the docs.

Because const generic expressions are not supported (`Tensor<B, {D+1}, K>`), that will likely require a const rank for the input and output, but it matches other APIs.

Side note: for large batched tensors, `x_col.matmul(y_row)` might be more performant but we'd have to benchmark to know. So we can keep the broadcasted elemwise atm.

huy209vn:
i merged outer with outer_batch.

burn-book/src/building-blocks/tensor.md
```diff
@@ -385,13 +385,17 @@ strategies.
...
-| `linalg::vector_norm(tensors, p, dim)` | `torch.linalg.vector_norm(tensor, p, dim) |
-| `linalg::diag(tensor)`                 | `torch.diag(tensor)`                      |
-| `linalg::trace(tensor)`                | `torch.trace(tensor)`                     |
+
+| Burn API                               | PyTorch Equivalent                                 |
+|----------------------------------------|----------------------------------------------------|
+| `linalg::vector_norm(tensors, p, dim)` | `torch.linalg.vector_norm(tensor, p, dim)`         |
+| `linalg::diag(tensor)`                 | `torch.diag(tensor)`                               |
+| `linalg::trace(tensor)`                | `torch.trace(tensor)`                              |
+| `linalg::outer(x, y)`                  | `torch.outer(x, y)` / `einsum("bi,bj->bij", …)`    |` (no direct op)|
```
laggui: Left over column? If there is no direct equivalent we usually mark it as N/A.

```suggestion
| `linalg::outer(x, y)`                  | `torch.outer(x, y)` / `einsum("bi,bj->bij", …)`    |
```

crates/burn-tensor/src/tensor/linalg/outer.rs
```diff
@@ -0,0 +1,48 @@
...
+/// - D = 2, R = 3: batched (b, m) × (b, n) → (b, m, n)
+///
+/// Panics:
+/// - if D > 2
+/// - if (D, R) is not (1,2) or (2,3)
+/// - if D = 2 and batch dimensions differ
+///
+/// Notes:
+/// - For large batched inputs, `x_col.matmul(y_row)` *might* be more performant
+///   than broadcasted elemwise multiply; benchmarking needed to confirm.
```
laggui: Tiny nitpick: I would keep this as a regular comment on the current implementation, not on the doc.

PR #3788, opened 2025-09-25
ci: add dispatch trigger publish workflow and bump xtask to 2.1.10

syl20bnr:
When this PR is merged it is possible to execute the `publish` workflow manually and set the parameter `--dry-run-only`. Handy to test that the workflow actually works after a modification.


laggui (approved):
Thanks 🙏 

Should probably update the Cargo.lock too with the new xtask version

PR #3791, opened 2025-09-25
Add uint16 to onnx-ir

TheGhostHuCodes:
### Checklist

All tests except these 4 passed. I think this might be floating-point precision issues. I'm running these tests on AppleSilicon. I don't think my changes would have affected this part of the code.

```
failures:
    tests::autodiff::ad_div::tests::test_div_complex_2
    tests::autodiff::ad_log_sigmoid::tests::should_diff_log_sigmoid
    tests::autodiff_checkpointing::ad_div::tests::test_div_complex_2
    tests::autodiff_checkpointing::ad_log_sigmoid::tests::should_diff_log_sigmoid
```

### Related Issues/PRs

None

### Changes

Added `Uint16` as a element type in `onnx-ir`. I'm using `onnx-ir` as a as an ONNX parser, and ran into an issue where a model takes a type `Uint16` input tensor. This fixes the issue on my end, and I am able to parse the ONNX model to completion.

### Testing

Manually tested on my local machine to ensure that a model using uint16 input parses.

### DIFF
--- a/crates/burn-import/src/burn/node/cast.rs
+++ b/crates/burn-import/src/burn/node/cast.rs
@@ -44,6 +44,7 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for CastNode {
                     ElementType::Int32
                     | ElementType::Int64
                     | ElementType::Int8
+                    | ElementType::Uint16
                     | ElementType::Uint8 => ScalarKind::Int64,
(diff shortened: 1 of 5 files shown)


antimora (approved):
LTGM

Thank you for you contribution.

PR #3792, opened 2025-09-26
BurnpackStore 

antimora:
### Checklist

### Related Issues/PRs

### Changes

### Testing


antimora:
Added `burn-store/examples/burnpack_inspect.rs` to generate a sample burnpack file to examine. Here is some output:

```
[burn-store]% hexdump -C /tmp/sample.burnpack | head -50
00000000  4e 52 55 42 01 00 d2 01  00 00 a2 67 74 65 6e 73  |NRUB.......gtens|
00000010  6f 72 73 a6 6c 6c 69 6e  65 61 72 31 2e 62 69 61  |ors.llinear1.bia|
00000020  73 a3 65 64 74 79 70 65  63 46 33 32 65 73 68 61  |s.edtypecF32esha|
00000030  70 65 81 18 40 6c 64 61  74 61 5f 6f 66 66 73 65  |pe..@ldata_offse|
00000040  74 73 82 19 80 00 19 81  00 6e 6c 69 6e 65 61 72  |ts.......nlinear|
00000050  31 2e 77 65 69 67 68 74  a3 65 64 74 79 70 65 63  |1.weight.edtypec|
00000060  46 33 32 65 73 68 61 70  65 82 18 80 18 40 6c 64  |F32eshape....@ld|
00000070  61 74 61 5f 6f 66 66 73  65 74 73 82 00 19 80 00  |ata_offsets.....|
00000080  6c 6c 69 6e 65 61 72 32  2e 62 69 61 73 a3 65 64  |llinear2.bias.ed|
00000090  74 79 70 65 63 46 33 32  65 73 68 61 70 65 81 18  |typecF32eshape..|
000000a0  20 6c 64 61 74 61 5f 6f  66 66 73 65 74 73 82 19  | ldata_offsets..|
000000b0  a1 00 19 a1 80 6e 6c 69  6e 65 61 72 32 2e 77 65  |.....nlinear2.we|
000000c0  69 67 68 74 a3 65 64 74  79 70 65 63 46 33 32 65  |ight.edtypecF32e|
000000d0  73 68 61 70 65 82 18 40  18 20 6c 64 61 74 61 5f  |shape..@. ldata_|
000000e0  6f 66 66 73 65 74 73 82  19 81 00 19 a1 00 6c 6c  |offsets.......ll|
000000f0  69 6e 65 61 72 33 2e 62  69 61 73 a3 65 64 74 79  |inear3.bias.edty|
00000100  70 65 63 46 33 32 65 73  68 61 70 65 81 0a 6c 64  |pecF32eshape..ld|
00000110  61 74 61 5f 6f 66 66 73  65 74 73 82 19 a6 80 19  |ata_offsets.....|
00000120  a6 a8 6e 6c 69 6e 65 61  72 33 2e 77 65 69 67 68  |..nlinear3.weigh|
00000130  74 a3 65 64 74 79 70 65  63 46 33 32 65 73 68 61  |t.edtypecF32esha|
00000140  70 65 82 18 20 0a 6c 64  61 74 61 5f 6f 66 66 73  |pe.. .ldata_offs|
00000150  65 74 73 82 19 a1 80 19  a6 80 68 6d 65 74 61 64  |ets.......hmetad|
00000160  61 74 61 a5 66 61 75 74  68 6f 72 6c 42 75 72 6e  |ata.fauthorlBurn|
00000170  20 45 78 61 6d 70 6c 65  6b 64 65 73 63 72 69 70  | Examplekdescrip|
00000180  74 69 6f 6e 78 29 53 61  6d 70 6c 65 20 66 69 6c  |tionx)Sample fil|
00000190  65 20 66 6f 72 20 65 78  61 6d 69 6e 69 6e 67 20  |e for examining |
000001a0  42 75 72 6e 70 61 63 6b  20 66 6f 72 6d 61 74 66  |Burnpack formatf|
000001b0  66 6f 72 6d 61 74 68 62  75 72 6e 70 61 63 6b 68  |formathburnpackh|
000001c0  70 72 6f 64 75 63 65 72  64 62 75 72 6e 67 76 65  |producerdburngve|
000001d0  72 73 69 6f 6e 66 30 2e  31 39 2e 30 5a 7e 3e 3d  |rsionf0.19.0Z~>=|
000001e0  e6 00 93 bd 14 7a 89 bc  d2 91 57 3d fa c4 4a 3d  |.....z....W=..J=|
000001f0  20 05 6d 3b d6 de 7e 3d  42 5c 73 3d 70 9b da bc  | .m;..~=B\s=p...|
00000200  61 e6 83 3d c0 d3 d7 3b  62 5a 44 3d 2c 44 b9 3c  |a..=...;bZD=,D.<|
00000210  a5 8c 1c bd 11 83 a7 3d  93 a1 0b bd bd 06 a2 3d  |.......=.......=|
00000220  00 30 af b9 47 7e b4 bd  64 e9 02 bd 20 eb fb bb  |.0..G~..d... ...|
00000230  48 38 b9 3c 35 0c 8d 3d  3a c1 1b 3d 30 26 ee bb  |H8.<5..=:..=0&..|
00000240  b0 be e4 3c d8 fd d4 3c  3c 48 18 bd 40 99 f4 3a  |...<...<<H..@..:|
00000250  9b b1 4c bd 5e 39 27 3d  f1 fa a3 3d e8 db c7 3c  |..L.^9'=...=...<|
00000260  88 91 ef 3c 6b f4 7c bd  a2 38 a9 bd ab cf 89 bd  |...<k.|..8......|
00000270  f5 1f 8a bd 33 3e a9 3d  30 95 29 3c 4b 99 8a 3d  |....3>.=0.)<K..=|
00000280  5e 4b a5 bd 68 d2 20 bc  d0 ed 00 3c 8a e7 28 3d  |^K..h. ....<..(=|
00000290  b0 30 0b 3c c2 b8 17 3d  6d 41 8a 3d 69 ad ae bd  |.0.<...=mA.=i...|
000002a0  e8 fe 0c 3d b1 50 b4 bd  57 2a ac 3d ce d8 1a 3d  |...=.P..W*.=...=|
000002b0  3a a3 64 3d b4 be 34 bd  80 2f a4 3b 6c 61 92 bd  |:.d=..4../.;la..|
000002c0  38 e0 e3 bc a0 f4 89 bb  73 4e 4d bd 43 8b 9f bd  |8.......sNM.C...|
000002d0  e0 7e 14 3d d1 e7 4b bd  50 9b 3e 3c c1 7b b3 bd  |.~.=..K.P.><.{..|
000002e0  48 d8 b5 bc c7 c3 a2 3d  c0 5f 53 bb eb 0d a5 bd  |H......=._S.....|
000002f0  30 a3 f8 bc f7 04 a5 3d  90 dd a9 bd fb e0 9a 3d  |0......=.......=|
00000300  a0 1d 83 3b 24 f3 aa bc  b0 09 d2 bc 54 34 aa bd  |...;$.......T4..|
00000310  4e 0e 27 bd 00 91 a8 3c  75 be b1 3d 6c 99 85 3c  |N.'....<u..=l..<|
[burn-store]%
```

antimora:
Rebased with the main; @laggui @nathanielsimard  Ready for review.

crates/burn-store/src/burnpack/base.rs
```diff
@@ -0,0 +1,197 @@
...
+use alloc::collections::BTreeMap;
+use alloc::string::String;
+use alloc::vec::Vec;
+use burn_tensor::DType;
+use byteorder::{ByteOrder, LittleEndian};
+use serde::{Deserialize, Serialize};
+
+/// Magic number identifying a Burnpack file: "BURN" in ASCII (0x4255524E)
+/// When written to file in little-endian format, appears as "NRUB" bytes
+pub const MAGIC_NUMBER: u32 = 0x4255524E;
```
laggui, line 14: Clever 😄

crates/burn-store/src/burnpack/base.rs
```diff
@@ -0,0 +1,197 @@
...
+const _: () = assert!(MAGIC_SIZE + VERSION_SIZE + METADATA_SIZE_FIELD_SIZE == HEADER_SIZE);
+
+/// Header structure for Burnpack files
+#[derive(Debug, Clone, Copy)]
+pub struct BurnpackHeader {
+    /// Magic number (4 bytes): 0x4255524E ("BURN")
+    pub magic: u32,
+    /// Format version (2 bytes)
+    pub version: u16,
+    /// Size of MessagePack metadata in bytes (4 bytes)
```
laggui: No longer messagepack

crates/burn-store/src/burnpack/base.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        Ok(Self {
+            magic,
+            version,
+            metadata_size,
+        })
+    }
+}
+
+/// Metadata structure serialized with MessagePack
```
laggui: No longer messagepack

crates/burn-store/src/burnpack/store.rs
```diff
@@ -0,0 +1,393 @@
...
+use burn_core::prelude::Backend;
+
+/// Store mode for BurnpackStore
+enum StoreMode {
+    #[cfg(feature = "std")]
+    File(PathBuf),
+    Bytes(Option<Vec<u8>>),
+}
+
+/// BurnpackStore - A Burn-specific file format store using MessagePack for metadata
```
laggui: No longer messagepack

laggui (commented):
(this is a bit more of a high level review)

There seems to be a couple of dead code paths that are just marked as `#[allow(dead_code)]`. Not sure why there is "legacy" code in new additions 😄 

Also, the memory usage of the CPU backend seems to point to some additional allocations overhead that we should try to minimze. The detailed benchmarks seem to point in that direction as well.

The current recorder approach (read all at once) is really optimized for speed, but the gap in read/write speeds should still not be as wide as we see in the benchmarks. I think we can optimize the sequential lazy read path (the common use case when loading a model) to improve this. My guess is that there are too many syscalls right now due to the seek + read for each tensor. We should probably separate the IO strategy from the tensor/snapshot object. I am not sure how it would be best yet but just giving my thoughts.

antimora:
@laggui I will investigate issue. 

Looking at the benchmark results, BurnpackStore definitely calls more alloc than NamedMpk, even though it uses less memory overall.

image

NamedMpk reads a file and loads all its content into memory. Then it transfers vec ownership to Data without copying again. 

BurnpackStore uses `TensorData::from_bytes_vec()`; I need to see if there is more efficient API.

antimora:
@laggui UPDATE:

 The 10x slowdown is caused by:

  1. LargeModel::new(&device) creates Params with random weights (but lazy via OnceCell)
  2. apply_from() → module.map() → Param::map() → .consume() → .val() forces 360MB of random weights to materialize!
  3. Then TensorSnapshots create another 327MB of loaded weights
  4. Total waste: 360MB + overhead

antimora:
@laggui 
> Also, the memory usage of the CPU backend seems to point to some additional allocations overhead that we should try to minimze. The detailed benchmarks seem to point in that direction as well.

I have figured out the root cause and additional issues:

1. ModuleMapper was initializing params instead of replacing uninitialized params. I have fixed this. Additionally partially loaded model keeps skipped params intact. 
2. Params also had on_save/on_load hooks that Nathaniel added recently for record saving which we totally forgot. Now Store's applier/collector properly execute those hooks - take the same execution path as the recorder. 

I updated the benchmark table in the PR description. BurnpackStore performs the best of all now. PytorchStore and SafetensorsStore also benefit from these fixes.

antimora:
Came across a safetensors issue (https://github.com/huggingface/safetensors/issues/649#issuecomment-3248786583) regarding whether it's possible to identify a safetensors file. Apparently, there is no magic number, which requires structure matching instead. In contrast, the Burnpack file format has a magic number now.

crates/burn-core/src/module/param/base.rs
```diff
@@ -28,61 +29,130 @@ fn new_mapper<T, F: Fn(T) -> T + Send + Sync + 'static>(func: F) -> Mapper<T> {
...
+///     let expected_require_grad = self.lazy_is_require_grad(); // NO initialization!
+///
+///     // Move to expected device, apply mapper, set require_grad
+///     new_tensor = tensor.to_device(&expected_device);
+///     new_tensor = mapper.on_load(new_tensor);
+///     new_tensor = new_tensor.set_require_grad(expected_require_grad);
+///
+///     Self::initialized(param_id, new_tensor)  // Returns initialized param
+/// }
+/// ```
```
nathanielsimard: I don't think the doc is correct, and imo not necessary, I would remove it.

crates/burn-core/src/module/param/base.rs
```diff
@@ -28,61 +29,130 @@ fn new_mapper<T, F: Fn(T) -> T + Send + Sync + 'static>(func: F) -> Mapper<T> {
...
+/// ## Performance Pattern
+///
+/// ```rust,ignore
+/// // Create uninitialized module (no tensor allocations)
 /// let module = config.init(device);
-/// // Will use the tensor allocated for the record if the same device is used.
-/// let module = module.load_record(record);
+///
+/// // Load tensors - lazy params use loaded tensors directly, bypassing init functions
+/// let module = module.load(...);
```
nathanielsimard: Same here

crates/burn-core/src/module/param/base.rs
```diff
@@ -28,61 +29,130 @@ fn new_mapper<T, F: Fn(T) -> T + Send + Sync + 'static>(func: F) -> Mapper<T> {
...
-    pub(crate) record_mapper: RecordMapper<T>,
+    /// **State Transitions:**
+    /// - Initialized params: `None`
+    /// - Uninitialized params: `Some(RwLock<Some(Uninitialized<T>)>)`
+    /// - After lazy init triggers: `Some(RwLock<None>)` (inner Option is taken)
+    ///
+    /// **Thread Safety:**
+    /// The RwLock ensures that concurrent calls to `lazy_device()` or `lazy_is_require_grad()`
+    /// during initialization will wait for the write lock to complete, preventing race conditions
+    /// between initialization and lazy queries.
```
nathanielsimard: When there is no unsafe, I wouldn't add safety comments, it's verbose and may be wrong.

crates/burn-core/src/module/param/base.rs
```diff
@@ -135,14 +205,43 @@ pub trait Parameter: Clone + core::fmt::Debug + Send {
...
+/// struct Uninitialized<P> {
+///     init: Box<dyn FnOnce(&P::Device, bool) -> P>,  // initialization function
+///     device: P::Device,                              // target device
+///     is_require_grad: bool,                          // gradient requirement
+///     shape: Shape,                                   // tensor shape
+/// }
+/// ```
+///
+/// This struct is stored inside the `initialization` field of [Param] and is consumed
+/// when the parameter is first accessed, transferring ownership to the initialization function.
```
nathanielsimard: I would remove that doc

crates/burn-core/src/module/param/base.rs
```diff
@@ -172,12 +277,23 @@ impl<T: Parameter> Param<T> {
...
+    ///
+    /// For initialized parameters, this returns a clone of the cached value.
+    /// For uninitialized parameters, this triggers initialization:
+    /// 1. Acquires write lock on the initialization state
+    /// 2. Takes the `Uninitialized` struct (replacing with `None`)
+    /// 3. Calls the initialization function
+    /// 4. Stores the result in the `OnceCell` for future access
+    ///
+    /// **Thread Safety:** OnceCell ensures initialization happens exactly once,
+    /// even with concurrent access.
```
nathanielsimard: That part is unecessary

crates/burn-core/src/module/param/base.rs
```diff
@@ -193,43 +309,64 @@ impl<T: Parameter> Param<T> {
...
+    /// This is a helper method for creating parameters while preserving the param mapper,
+    /// typically used in ModuleMapper implementations.
+    pub fn into_initialized(id: ParamId, value: T, param_mapper: ParamMapper<T>) -> Self {
+        Self {
+            id,
+            state: OnceCell::from(value),
+            initialization: None,
+            param_mapper,
+        }
+    }
```
nathanielsimard: The naming isn't correct here, into should consume a self. Maybe `from_mapped_value`

crates/burn-core/src/module/param/base.rs
```diff
@@ -267,19 +404,19 @@ impl<T: Parameter> Param<T> {
...
-    /// when the device is used for updating the tensor value, which has potentially not been
-    /// initialized yet, like loading a record.
+    /// This is critical for the load optimization: when loading tensors into an uninitialized parameter,
+    /// we need to know the target device to move the loaded tensor appropriately, but we don't want to
+    /// trigger the initialization function (which would allocate an unnecessary tensor).
     ///
-    /// # Notes
+    /// **Returns:**
+    /// - For uninitialized params: the device from the `Uninitialized` struct
+    /// - For initialized params: the actual device from the tensor
```
nathanielsimard: I would remove the Returns part

crates/burn-core/src/module/param/base.rs
```diff
@@ -293,12 +430,15 @@ impl<T: Parameter> Param<T> {
...
+    /// When loading tensors into an uninitialized parameter, we need to apply the correct gradient
+    /// setting to the loaded tensor without triggering the initialization function.
     ///
-    /// This should be used instead of [crate::tensor::Tensor::is_require_grad], since using the tensor
-    /// function requires a dereference, which triggers the initialization. This is only useful
-    /// when the boolean is used for updating the tensor value, which has potentially not been
-    /// initialized yet, like loading a record.
+    /// **Returns:**
+    /// - For uninitialized params: the gradient requirement from the `Uninitialized` struct
+    /// - For initialized params: the actual gradient requirement from the tensor
```
nathanielsimard: Same here

crates/burn-core/src/module/param/tensor.rs
```diff
@@ -72,6 +72,32 @@ impl<B: Backend, const D: usize> Param<Tensor<B, D>> {
...
 
+    /// The shape of the parameter, **without triggering initialization**.
+    ///
+    /// This is critical for shape validation during loading: when applying tensors to an
+    /// uninitialized parameter, we need to validate the shape without triggering the
+    /// initialization function (which would allocate an unnecessary tensor).
+    ///
+    /// **Returns:**
+    /// - For uninitialized params: the shape from the `Uninitialized` struct
+    /// - For initialized params: the actual shape from the tensor
```
nathanielsimard: I would remove that.

crates/burn-core/src/module/param/tensor.rs
```diff
@@ -82,51 +108,202 @@ impl<B: Backend, const D: usize> Param<Tensor<B, D>> {
...
+    ///
```
nathanielsimard: Remove

crates/burn-core/src/module/param/tensor.rs
```diff
@@ -82,51 +108,202 @@ impl<B: Backend, const D: usize> Param<Tensor<B, D>> {
...
+    /// It applies the param mapper's `on_save` transformation, which can be used
+    /// to modify the tensor before serialization (e.g., quantization, precision conversion).
+    pub fn save(&self) -> Self {
+        let mut tensor = self.val();
+        let mapper = self.param_mapper.clone();
+
+        tensor = mapper.on_save(tensor);
+
+        Self::initialized(self.id, tensor)
+    }
```
nathanielsimard: This is strange, `save` doesn't save and returns itself (the same type).

crates/burn-no-std-tests/src/burnpack.rs
```diff
@@ -0,0 +1,158 @@
...
+    // Get the serialized bytes
+    let bytes = save_store.get_bytes().expect("Failed to get bytes");
+
+    // Load from bytes
+    let mut load_store = BurnpackStore::from_bytes(Some(bytes));
+    let mut loaded_model = TestModel::<B>::new(device);
+    let result = loaded_model
+        .apply_from(&mut load_store)
+        .expect("Failed to load model");
+
```
nathanielsimard, line 58: The naming could be improved.

```rust
BurnpackStore::empty();
model.save_into(&mut store);
store.into_bytes().unwrap();
BurnpackStore::from_bytes(bytes);
loaded_model.load_from(&mut store);
```
`collect` and `apply` are not really intuitive and I woud not have discovered them using autocompletion.

crates/burn-store/README.md
```diff
@@ -282,12 +310,26 @@ if !result.errors.is_empty() {
 
 ## Benchmarks
 
+### Loading Benchmarks
+
+Compares 6 loading methods with a 312MB ResNet-18 model:
+
+**Performance Results (NdArray Backend)**:
+- **BurnpackStore**: 565ms (2x faster than SafeTensors, 361MB memory)
+- **NamedMpkFileRecorder**: 48ms (fastest, 344MB memory)
```
nathanielsimard: The old recorder with ndarray simply gives away the vector of bytes as storage for the ndarray tensor, so it's essentially a no-copy loading. We should be able to reproduce the same behavior with burnpack, maybe it already is and the problems comes from the benchmark tests

crates/burn-store/src/burnpack/base.rs
```diff
@@ -0,0 +1,197 @@
...
+// │  - Metadata size (4 bytes)       │  Size of CBOR metadata in bytes (u32)
+// ├──────────────────────────────────┤
+// │  Metadata (CBOR)                 │
+// ├──────────────────────────────────┤
+// │  - Tensor descriptors (BTreeMap) │  Sorted map of tensor metadata
+// │    Key: tensor name (string)     │  Tensor identifier (e.g., "model.layer1.weight")
+// │    Value: TensorDescriptor       │
+// │      - dtype: enum               │  Data type (F32, F64, I32, I64, U32, U64, U8, Bool)
+// │      - shape: Vec<u64>           │  Tensor dimensions
+// │      - data_offsets: (u64, u64)  │  (start, end) byte offsets relative to data section
```
nathanielsimard: Is it the correct way of using the data offsets?

```rust
let file = std::fs::File::open("file.bpk").offset(data_offsets[0]).size(data_offsets[1]);
let tensor_data = file.to_bytes();
```
Something like that?

crates/burn-store/src/burnpack/base.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub fn new(metadata_size: u32) -> Self {
+        Self {
+            magic: MAGIC_NUMBER,
+            version: FORMAT_VERSION,
+            metadata_size,
+        }
+    }
+
+    /// Serialize header to bytes
+    pub fn to_bytes(self) -> [u8; HEADER_SIZE] {
```
nathanielsimard: this would be `into_bytes` since it takes ownership of the header.

crates/burn-store/src/burnpack/base.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        Ok(Self {
+            magic,
+            version,
+            metadata_size,
+        })
+    }
+}
+
+/// Metadata structure serialized with CBOR
```
nathanielsimard, line 138: what is CBOR

crates/burn-store/src/burnpack/reader.rs
```diff
@@ -0,0 +1,337 @@
...
+use std::cell::RefCell;
+#[cfg(feature = "std")]
+use std::fs::File;
+#[cfg(feature = "std")]
+use std::io::{Read, Seek};
+#[cfg(feature = "std")]
+use std::path::Path;
+
+/// Storage backend for BurnpackReader
+pub(crate) enum StorageBackend {
```
nathanielsimard, line 26: Does the backend contain all the data of the store or simply for a tensor.

crates/burn-store/src/burnpack/reader.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// File-based storage with buffered reading
+    #[cfg(feature = "std")]
+    #[allow(dead_code)]
+    FileBuffered { file: Rc<RefCell<File>> },
+}
+
+impl StorageBackend {
+    /// Get data slice from the storage
+    /// For FileBuffered backend, start and end should be absolute file positions
+    pub(crate) fn get_slice(&self, start: usize, end: usize) -> Result<Vec<u8>, BurnpackError> {
```
nathanielsimard: I would change that function signature to the following:

```rust
fn read_into(&self, bytes: &mut [u8], offset: usize)
```

Since in the future the `Vec<u8>` could be allocated by the backend as pinned memory managed in a memory pool.

crates/burn-store/src/burnpack/reader.rs
```diff
@@ -0,0 +1,337 @@
...
+    pub(crate) metadata: BurnpackMetadata,
+    /// Storage backend
+    pub(crate) storage: StorageBackend,
+    /// Offset to the start of tensor data
+    pub(crate) data_offset: usize,
+}
+
+impl BurnpackReader {
+    /// Load from bytes
+    pub fn from_bytes(bytes: Vec<u8>) -> Result<Self, BurnpackError> {
```
nathanielsimard: We should replace all instances of `Vec<u8>` by `burn_common::Bytes`, `&[u8]` and `&mut [u8]` are OK

crates/burn-store/src/burnpack/store.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub fn default_metadata() -> BTreeMap<String, String> {
+        let mut metadata = BTreeMap::new();
+        metadata.insert("format".into(), "burnpack".into());
+        metadata.insert("producer".into(), "burn".into());
+        metadata.insert("version".into(), env!("CARGO_PKG_VERSION").into());
+        metadata
+    }
+    /// Create a new store from a file path
+    ///
+    /// By default, automatically appends `.burnpack` extension if the path doesn't have one.
```
nathanielsimard: `burnpack` extension could become `bpk` 🤩 3 letters extensions are the best and `bpk` follows the same schema as `mpk`

nathanielsimard (commented):
The file design seems good, and the overall architecture seems flexible. I have quite a few comments, mostly related to naming and documentation.

I think a lot of comments are quite bad; they describe how things work (sometimes not even accurately) instead of documenting how the code should be used.

Also, some formatting isn’t correct. We shouldn’t use `**title**` but `# title in` most places (`# Returns`, `# Safety`, etc.).

crates/burn-core/src/module/param/base.rs
antimora: Yeah, I will remove this. This is module level document and users complain about lack of documentation. This was to explain how it works. But i'll remove it.

crates/burn-store/README.md
antimora: Sorry, I forgot to update the README file. This has been fixed with the latest code. It was related to the fact that Recorder was simply updated uninitialized Params whereas BurnpackStore was initializing because ModuleMapper was initializing values right before being replaced. The latest benchmark report is in the PR description:

##### Load Time (Median) vs Memory Trade-off

| Loading Method | LibTorch | NdArray (CPU) | WGPU (GPU) |
|---|---|---|---|
| **Burnpack Store** | **36.72 ms / 16.82 MB** | **34.21 ms / 344.1 MB** | 60.31 ms / 243.4 MB |
| **NamedMpk Recorder** | 38.43 ms / 344.1 MB | 34.77 ms / 344.1 MB | 80.69 ms / 344.1 MB |
| **SafeTensors Recorder** | 125.8 ms / 654.6 MB | 932.8 ms / 654.6 MB | 166 ms / 654.6 MB |
| **SafeTensors Store** | 722.1 ms / 33.58 MB | 602.5 ms / 344.1 MB | 648.9 ms / 243.4 MB |
| **PyTorch Recorder** | 229.6 ms / 344.1 MB | 1,026 ms / 360.9 MB | 261.3 ms / 344.2 MB |
| **PyTorch Store** | 858.9 ms / 33.59 MB | 741.1 ms / 344.1 MB | 798.5 ms / 176.2 MB |

crates/burn-store/src/burnpack/base.rs
antimora: Reader (burnpack/reader.rs:260-261):
```rust
// Calculate absolute positions from relative offsets
let start = self.data_offset + descriptor.data_offsets.0 as usize;
let end = self.data_offset + descriptor.data_offsets.1 as usize;

// Then read the slice [start..end]
storage.get_slice(start, end)
```

Key points:
1. Both values are byte offsets, not offset+size
2. **Both are relative to the data section start, not the file start**
3. Length must be calculated as: `data_offsets.1 - data_offsets.0``
4. Absolute file position is: `data_section_start + data_offsets.0`

The first tensor_data starts with `(0..`. I purposefully made the offsets independent of the header size in case we add a new tensor to the existing burnpack file. Header and Metadata are independent. In theory, it'd be possible to store them separately.

antimora: Wanted to share ciborium/CBOR fileformat. It is possible to have a fixed file (lets say 10MB) filed with zeros and override a part of the file, lets say 3 MB that's only for serialized metadata, and the reader will be able to parse with no issue. This is in case if we need to preallocate some section in the beginning and override with metadata. The file spec allows for this. 

----

Based on the CBOR specification and how ciborium implements it, here are the answers to your questions:

Can you write CBOR at the beginning of a zero-filled file?

**Yes, absolutely!** If you write a CBOR-serialized object at the beginning of your 10MB zero-filled file, the deserializer will:

1. Read exactly the bytes needed for the complete CBOR data item
2. Stop after successfully deserializing that one item
3. Leave the remaining zeros (or any other trailing data) unread

In streaming applications, the decoder immediately begins decoding a new data item if data is found after the end of a previous data item, and decoders by definition only return contents from well-formed data items, not including following extraneous data.

Practical implications with ciborium:

When you use `ciborium::from_reader()`, it will:
- Read your CBOR data from the beginning
- Deserialize exactly one complete data item
- Stop reading once that item is complete
- The remaining zeros in your file won't cause any issues

This makes CBOR excellent for scenarios where you want to reserve file space upfront or append data later without rewriting the entire file.

crates/burn-store/src/burnpack/reader.rs
antimora: All data, this is so it's easy to work with bytes, mmap and file handler (abstraction).

crates/burn-store/src/burnpack/reader.rs
antimora: OK. I updated the reader APIs and also commented for the future:

```rust
  // Create lazy TensorSnapshot
  let snapshot = TensorSnapshot::from_closure(
      Rc::new(move || {
          // This closure is only called when data is actually needed
          let len = end - start;
          // TODO Should be allocated by the backend in the future
          // See https://github.com/tracel-ai/burn/pull/3792#discussion_r2416812091
          let mut data_bytes = vec![0u8; len];
          storage.read_into(&mut data_bytes, start).map_err(|e| {
              crate::TensorSnapshotError::IoError(format!(
                  "Failed to read tensor data: {}",
                  e
              ))
          })?;
          Ok(TensorData::from_bytes_vec(
              data_bytes,
              shape_for_closure.clone(),
              dtype,
          ))
      }),
```

crates/burn-store/src/burnpack/base.rs
antimora: I switched from MessagePack to CBOR for metadata serialization due to rmp-serde's limitations:

Why the change:
- rmp-serde is not no-std compatible
- The project has been inactive for over a year, so submitting a PR didn't seem worthwhile

Why CBOR:
- True no-std support: Using ciborium, which implements CBOR (IETF RFC 8949) for proper embedded system compatibility
- Recommended by Serde as a preferred serialization format
- MessagePack-inspired: CBOR is an IETF standardized format inspired by MessagePack
- Preserves enum variant information in file output

References:
- https://cbor.io/
- https://en.wikipedia.org/wiki/CBOR

antimora:
@laggui, I have addressed all @nathanielsimard feedback. It's more robust. Additionally, I have fixed potential data corruption issues (made more robust and strict). 

One of the biggest improvements is Bytes/[u8] usage. I added todos for to have a better control over byte allocation, as we transition to Backends allocator. So it should be easier now to transition to that model.

antimora:
Addressed Denial of Service (DOS) vulnerabilities related to resource exhaustion. This is for a production use with untrusted input.

PR #3793, opened 2025-09-26
Fix/autodiff/multi threads

nathanielsimard:
Not a full solution yet for mutli-device training, but now there's a possibility to reduce the amount of locking for the backward pass.

### DIFF
--- a/crates/burn-autodiff/src/runtime/graph.rs
+++ b/crates/burn-autodiff/src/runtime/graph.rs
@@ -229,10 +229,12 @@ impl GraphLocator {
     /// Registers a key for a given origin node.
     fn register_key(&mut self, origin: NodeId, key: NodeId) {
         if !self.keys.contains_key(&origin) {
+            // Ensure an entry exists for this origin
             self.keys.insert(origin, HashSet::new());
         }
 
         if origin != key {
+            // Register this node to point to the origin graph
             self.keys.get_mut(&origin).unwrap().insert(key);
         }
     }
@@ -244,8 +246,10 @@ impl GraphLocator {
         core::mem::swap(&mut state_old, &mut locked);
         main_state.server.extend(state_old.server);
 
+        // Re-map merged origin to the main graph
         self.graphs.insert(merged.origin, main.clone());
 
+        // Move all keys (node IDs) from the merged graph to the main graph
         if let Some(locator_keys) = self.keys.remove(&merged.origin) {
             for k in locator_keys.iter() {
                 self.graphs.insert(*k, main.clone());


crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,200 @@
...
+
+/// A client for managing multiple graphs using mutex-based synchronization.
+///
+/// The biggest benefit of using this client implementation is that each graph can modify its own
+/// data without blocking other graphs, which is essential for multi-device training.
+///
+/// # Notes
+///
+/// The [AutodiffServer] fully support multiple graphs with sharing nodes, however those type of
+/// graphs will be store under a single mutex-protected graph by the client, limiting
```
laggui:
```suggestion
/// The [AutodiffServer] fully supports multiple graphs with sharing nodes, however those type of
/// graphs will be stored under a single mutex-protected graph by the client, limiting
```

laggui (commented):
Just a minor typo, otherwise comments were addressed offline.

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,262 @@
...
+/// A client for managing multiple graphs using mutex-based synchronization.
+///
+/// The biggest benefit of using this client implementation is that each graph can modify its own
+/// data without blocking other graphs, which is essential for multi-device training.
+///
+/// # Notes
+///
+/// The [AutodiffServer] fully supports multiple graphs with sharing nodes, however those type of
+/// graphs will be stored under a single mutex-protected graph by the client, limiting
+/// parralelisation.
```
laggui: Typo:

```suggestion
/// parallelisation.
```

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,262 @@
...
+
+impl core::fmt::Debug for Graph {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Graph")
+            .field("origin", &self.origin)
+            .finish()
+    }
+}
+
+static STATE: spin::Mutex<Option<GraphLocator>> = spin::Mutex::new(None);
```
laggui, line 67: Mixed usage of the `stub::Mutex` (for `GraphState`) and here `spin::Mutex`

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,262 @@
...
+                self.register_key(graph.origin, node);
+            }
+
+            return graph;
+        }
+
+        self.merge(node, graphs)
+    }
+
+    fn select_many(&mut self, node: NodeId, parents: &[Parent]) -> Vec<Arc<Graph>> {
```
laggui: It would be helpful to add some comments so when we (or someone else) comes back to this code in the future the logic is explained a bit more. Suggestions below (correct if wrong):

```diff
+   /// Returns the distinct graphs (by node origin) relevant to this node and its parents.
    fn select_many(&mut self, node: NodeId, parents: &[Parent]) -> Vec<Arc<Graph>> {
        let mut graphs = HashMap::<NodeId, Arc<Graph>>::new();

        if let Some(val) = self.graphs.get(&node) {
            if parents.is_empty() {
+               // No parents introduce other graphs
                return vec![val.clone()];
            }
+           // Otherwise, collect its graph as candidate
            graphs.insert(val.origin, val.clone());
        }

        for parent in parents {
+           // Collect all graphs that the parent nodes belong to
            match self.graphs.get(&parent.id) {
                Some(graph) => graphs.insert(graph.origin, graph.clone()),
                None => continue,
            };
        }

        if graphs.is_empty() {
+           // No graphs collected from node + parents
            return match self.graphs.get(&node) {
                Some(old) => vec![old.clone()],
                None => {
                    let graph = self.new_graph(node);
                    vec![graph]
                }
            };
        }

        graphs.drain().map(|(_, v)| v).collect()
    }
```

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,262 @@
...
+                    let graph = self.new_graph(node);
+                    vec![graph]
+                }
+            };
+        }
+
+        graphs.drain().map(|(_, v)| v).collect()
+    }
+
+    fn merge(&mut self, node: NodeId, mut graphs: Vec<Arc<Graph>>) -> Arc<Graph> {
```
laggui:
```diff
    fn merge(&mut self, node: NodeId, mut graphs: Vec<Arc<Graph>>) -> Arc<Graph> {
        let main = graphs.pop().unwrap();
        self.register_key(main.origin, node);

        let mut state = main.state.lock().unwrap();

        for graph in graphs.drain(..) {
            self.merge_two(&mut state, &main, graph);
        }

+       // Update graph mapping for origin and current node
        self.graphs.insert(main.origin, main.clone());
        self.graphs.insert(node, main.clone());

        core::mem::drop(state);

        main
    }
```

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        self.graphs.insert(main.origin, main.clone());
+        self.graphs.insert(node, main.clone());
+
+        core::mem::drop(state);
+
+        main
+    }
+
+    fn register_key(&mut self, origin: NodeId, key: NodeId) {
```
laggui, line 230:
```diff
    fn register_key(&mut self, origin: NodeId, key: NodeId) {
        if !self.keys.contains_key(&origin) {
+           // Ensure an entry exists for this origin
            self.keys.insert(origin, HashSet::new());
        }

        if origin != key {
+           // Register this node to point to the origin (graph)
            self.keys.get_mut(&origin).unwrap().insert(key);
        }
    }
```

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,262 @@
...
+        if !self.keys.contains_key(&origin) {
+            self.keys.insert(origin, HashSet::new());
+        }
+
+        if origin != key {
+            self.keys.get_mut(&origin).unwrap().insert(key);
+        }
+    }
+
+    fn merge_two(&mut self, main_state: &mut GraphState, main: &Arc<Graph>, merged: Arc<Graph>) {
```
laggui, line 243:
```diff
    fn merge_two(&mut self, main_state: &mut GraphState, main: &Arc<Graph>, merged: Arc<Graph>) {
        let mut locked = merged.state.lock().unwrap();
        let mut state_old = GraphState::default();
        core::mem::swap(&mut state_old, &mut locked);
        main_state.server.extend(state_old.server);

+       // Re-map merged origin to the main graph
        self.graphs.insert(merged.origin, main.clone());

+       // Move all keys (node IDs) from the merged graph to the main graph
        if let Some(locator_keys) = self.keys.remove(&merged.origin) {
            for k in locator_keys.iter() {
                self.graphs.insert(*k, main.clone());
            }

            let locator_keys_main = self
                .keys
                .get_mut(&main.origin)
                .expect("Should be init before the merge.");
            locator_keys_main.extend(locator_keys);
        }
    }
```

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -21,20 +22,27 @@ enum NodeMemoryStatus {
 }
 
 impl GraphMemoryManagement {
+    pub fn extend(&mut self, other: Self) {
+        self.nodes.extend(other.nodes);
+        self.leaves.extend(other.leaves);
+        self.statuses.extend(other.statuses);
+    }
```
laggui, line 29: This is fine and safe for single device and multi-device data parallel.

But I think we will need better handling for other forms of parallelism where node collisions might happen (if I am reasoning correctly, since there will be multiple contributors to the same node). In this case, we will need a smarter merge I think. And if so, we should leave a comment (TODO).

laggui (commented):
Good stuff!

Should we add some tests for the graph merging logic?

Need to update the cubecl rev / merge main.

crates/burn-autodiff/src/runtime/memory_management.rs
nathanielsimard: I don't think so because when we detech a colision, we merge the graphs __before__ adding the new node.

laggui: Ahhh you're right, that should prevent any issues. Collisions at the autodiff server level should not happen in this case, so not to worry.

nathanielsimard:
@laggui 

> Should we add some tests for the graph merging logic?

I'm unsure, the behavior shouldn't change compared to before and we already have tests for multiple graphs with and without shared nodes. A real test would be to validate that we don't share the same lock for graphs that don't have shared nodes, but it's hard to test automatically (I checked manually).

laggui (approved):

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,285 @@
...
+        for parent in parents {
+            match self.graphs.get(&parent.id) {
+                Some(graph) => graphs.insert(graph.origin, graph.clone()),
+                None => continue,
+            };
+        }
+
+        if graphs.is_empty() {
+            return match self.graphs.get(&node) {
+                Some(old) => GraphAnalysis::NoCollision(old.clone()),
```
swfsql, line 196: Just a note, I think this path is not possible, cuz if the `graphs` are empty then it's already known that `self.graphs` doesn't have the current `node`

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,285 @@
...
+                None => GraphAnalysis::NoCollision(self.new_graph(node)),
+            };
+        }
+
+        if graphs.len() == 1 {
+            return GraphAnalysis::NoCollision(graphs.drain().next().unwrap().1);
```
swfsql, line 202: One note here, is that it's possible that the current `node` is not getting a graph created for it. This happens if it doesn't yet exists but if it has exactly one existing parent. Idk if that would be a problem or not, but I just thought of this when reading the code

edit: oh nvm, the caller checks if the single returned graph is the original `node` or not.

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        GraphAnalysis::Collisions(graphs)
+    }
+
+    /// Merges multiple graphs associated with a node into a single graph.
+    fn merge(&mut self, node: NodeId, mut graphs: HashMap<NodeId, Arc<Graph>>) -> Arc<Graph> {
+        let mut graphs = graphs.drain().map(|g| g.1);
+
+        let main = graphs.next().expect("At least one graph");
+        self.register_key(main.origin, node);
```
swfsql, line 213: A note for here is that the `drain` could return a different ordering I think, in case the "same" merge happens multiple times. In that case many different parents could get the `node` registered as a key. But again idk if that happens and if that would be desired or not.

crates/burn-autodiff/src/runtime/server.rs
```diff
@@ -14,12 +14,26 @@ use alloc::vec::Vec;
...
+    /// Cleans a single [node](NodeId).
+    fn clean(&mut self, node: &NodeId);
+}
+
 impl AutodiffServer {
+    pub fn extend(&mut self, other: AutodiffServer) {
+        self.steps.extend(other.steps);
+        self.actions_builder.extend(other.actions_builder);
+        self.memory_management.extend(other.memory_management);
+    }
```
swfsql, line 35: A note for the `HashMap::extend` is that if the key clashes, the previous value gets removed -- but ofc that wouldn't be relevant if the keys for different things (e.g. step) never clash, assuming such info wasn't supposed to be dropped.

Edit: in that case, a non-intersection debug assertion could be good
Edit: maybe nvm, as tensors keep having their ID increasing, then I think this won't ever happen.

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        if origin != key {
+            // Register this node to point to the origin graph
+            self.keys.get_mut(&origin).unwrap().insert(key);
+        }
+    }
+
+    /// Merges two graphs by combining their states and updating graph mappings.
+    fn merge_two(&mut self, main_state: &mut GraphState, main: &Arc<Graph>, merged: Arc<Graph>) {
```
swfsql, line 243: In this function, should the `merged`'s `NodeId` be inserted into the `main`'s key set? (assuming this would be necessary, and `merged` wasn't inserted as a key of itself)

crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -0,0 +1,285 @@
...
+        if graphs.len() == 1 {
+            return GraphAnalysis::NoCollision(graphs.drain().next().unwrap().1);
+        }
+
+        GraphAnalysis::Collisions(graphs)
+    }
+
+    /// Merges multiple graphs associated with a node into a single graph.
+    fn merge(&mut self, node: NodeId, mut graphs: HashMap<NodeId, Arc<Graph>>) -> Arc<Graph> {
+        let mut graphs = graphs.drain().map(|g| g.1);
```
swfsql, line 210: I'm thinking about this mapping, because sometimes the `g.0` can be different from `g.1` (in case the node got merged into another one), and therefore the `g.0` information is lost? I'm not sure if that would be a problem, or if the only thing that matters is the "pointed" graph origin.

crates/burn-autodiff/src/runtime/memory_management.rs
```diff
@@ -199,9 +207,9 @@ impl GraphMemoryManagement {
 
     fn identify_leaves_and_deletables(
         &self,
-        leaf_id: NodeID,
-        new_leaves: &mut HashSet<NodeID>,
-        to_delete: &mut Vec<NodeID>,
+        leaf_id: NodeId,
```
swfsql, line 210: On this method, it could be good to send the leaves as the initial `to_visit` vec. In this way, the leaves are inserted before the parents in the `to_delete` vec (maybe this could be useful).

I'm not sure if this could be important because later the `NodeCleaner` may depend on this ordering (of children getting removed before the parents).

PR #3797, opened 2025-09-27
Tensor::num_dims()

crutcher:
### Checklist

### Changes

Explicit `Tensor::num_dims()` api; which doesn't require resolved D param; to avoid `tensor.shape().num_dims()` heap copy.

### DIFF
--- a/crates/burn-autodiff/src/tensor.rs
+++ b/crates/burn-autodiff/src/tensor.rs
@@ -22,6 +22,10 @@ impl<B: Backend> TensorMetadata for AutodiffTensor<B> {
     fn shape(&self) -> burn_tensor::Shape {
         self.primitive.shape()
     }
+
+    fn num_dims(&self) -> usize {
+        self.primitive.num_dims()
+    }
 }
 
 pub type NodeRefCount = Arc<NodeID>;
--- a/crates/burn-candle/src/tensor.rs
+++ b/crates/burn-candle/src/tensor.rs
@@ -27,6 +27,10 @@ impl TensorMetadata for CandleTensor {
     fn shape(&self) -> Shape {
         Shape::from(self.tensor.dims().to_vec())
     }
+
+    fn num_dims(&self) -> usize {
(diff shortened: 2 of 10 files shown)


ArthurBrussee:
What about `rank` vs `num_dims` ?

crutcher:
@ArthurBrussee I personally prefer `Tensor::rank()`, but SOME OTHERS (ahem, pytorch, tensorflow) have hijacked `rank` to use in their crappy data parallel training implementations, to identify the master nodes.

I think it is cleaner, and is what I would prefer; but I anticipated the above push back; and going the other direction would *also* mean changing the `Shape::num_dims()` interface; which seemed a harder sell (but I'd prefer it).

laggui (commented):
I wouldn't require `num_dims` to be implemented for all backend primitives (via the `TensorMetadata` trait.

We should go through the `Shape`, but we could have an interface that returns a `&Shape` reference instead.

We could also bring the `shape(&self) -> &Shape` up to the higher level `Tensor` API, though that will likely break some user code (and this one might be annoying). So we can have a migration/deprecation path with a temporary `shape_ref(&self) -> &Shape`.

Then, `tensor.num_dims()` or `tensor.rank()` can simply be implemented on the `Tensor` struct using the shape ref.

This *will* require more changes throughout the codebase to fix what actually requires shape ownership vs just a ref right now. I can take this over if you'd like. But lmk if you have any thoughts on the above proposition (esp. for the ref/owned interfaces).

/edit: also, I think I would also prefer `tensor.rank()`. We have a const rank right now, I think Burn users are used to this nomenclature. And it would match cubecl.

crutcher:
- switched to `rank`.

WRT `Tensor::rank() := Tensor::shape_ref().rank()`; forcing the impls to uniformly have a storage location for a `shape_ref()` would require some of the impls to be wrapped with a base object; which is a larger change.

if you want to introduce that, we can; it starts to assume some impl control above the impl

laggui (approved):

PR #3805, opened 2025-09-29
fix: Fix tests with hardcoded types

wingertge:
### Checklist

### Changes

Fixes tests that were using hardcoded f32 and were panicking when running the suite on Vulkan.
Updates cubecl.

### Testing

The tests in question now pass.
There's one failure remaining but it seems to be an actual bug that need more investigating.
Remaining failure:
```
tests::cube::tensor::i8_ty::unfold::tests::test_unfold_int
```
It's trying to read out `255` as `i8`, which seems like an unsigned vs signed issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1870,7 +1870,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.7.0"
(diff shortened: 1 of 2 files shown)


laggui (approved):
Thanks for catching these 😅

PR #3806, opened 2025-09-29
Add `tensor.cumsum(dim)` operation for cumulative sum computation

antimora:
This PR implements the tensor.cumsum(dim) operation for computing cumulative sums along a specified dimension, similar to PyTorch's torch.cumsum and NumPy's np.cumsum.

### Checklist

### Related Issues/PRs

* #1722
* #2099
* #2664

### Changes
Adds cumulative sum support for all numerical tensor kinds (Float, Int) across all backends:

Core API:
- Added cumsum(dim) method to Tensor<B, D> API
- Implemented trait methods in FloatTensorOps, IntTensorOps, and QTensorOps

Backend Implementations:
- burn-ndarray: Efficient CPU implementation using Zip iteration
- burn-tch: Native PyTorch cumsum binding
- burn-candle: F32 conversion workaround for integer types + contiguity fix for flip operation
- burn-cubecl (wgpu): Custom GPU kernel with **naive** sequential scan (TODO: implement efficient parallel scan via cubecl-scan)
- burn-autodiff: Gradient computation using reversed cumsum pattern (validated against PyTorch)
- burn-fusion: Non-blocking async execution with IR operation support
- burn-router: Full IR integration with execution handlers

IR Support:
- Added Cumsum(ReduceDimOpIr) variant to BaseOperationIr enum
- Integrated cumsum into fusion and router backends via IR pipeline

### Testing
Comprehensive test coverage:

1. Basic cumsum tests
  - Forward pass validation for Float and Int tensors
  - Multiple dimensions (1D, 2D, 3D)
  - Different cumsum dimensions (dim=0, dim=1)
  - Tested on: ndarray,  tch , candle , wgpu
2. Autodiff gradient tests:
  - Gradient computation validation against PyTorch-computed expected values
  - Used PyTorch with uv python to generate reference gradients


antimora:
CC: @ArthurBrussee

ArthurBrussee:
Yay really want to see a cumsum!

Something is definitely better than nothing, but, I wonder whether this needs a simple parallel reduction kernel for cubecl first. Doing a serial reduction along the axis is quite different in performance charasterics to the point where a big array (eg. 1 million elements) probably would lock up your GPU and crash. 

The serial version could make sense for small dimensions though, so maybe good to keep it for the autotune.

crates/burn-cubecl/src/ops/numeric.rs
```diff
@@ -235,3 +235,75 @@ pub fn bitwise_xor_scalar<R: CubeRuntime, E: IntElement>(
...
 }
+
+/// Compute the cumulative sum along a dimension
+///
+/// Note: This is a naive sequential implementation along the cumsum dimension.
+/// TODO: Implement an efficient GPU-optimized parallel scan algorithm (cubecl-scan crate)
+/// References:
+/// - https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda
+/// - https://www.w3.org/TR/WGSL/#builtin-subgroupInclusiveAdd
+pub fn cumsum<R: CubeRuntime, E: CubeElement>(input: CubeTensor<R>, dim: usize) -> CubeTensor<R> {
```
laggui: Yeah this was the blocking point last time 😅 a cubecl-scan would be great.

There is actually a draft PR for cubecl https://github.com/tracel-ai/cubecl/pull/863, but it seems to be somewhat blocked.

But I guess this is a good starting point. As Arthur said it's already better than nothing! It doesn't operate on lines either. For tiny tensors this should be performant (or maybe even preferred over a more complex algo), but we should probably state the limitations.

crates/burn-cubecl/src/ops/numeric.rs
```diff
@@ -235,3 +235,75 @@ pub fn bitwise_xor_scalar<R: CubeRuntime, E: IntElement>(
...
+    #[cube(launch)]
+    fn cumsum_kernel<C: Numeric>(
+        input: &Tensor<C>,
+        output: &mut Tensor<C>,
+        dim_stride: u32,
+        #[comptime] dim_size: u32,
+    ) {
```
laggui: I'd split the kernel launch from the function

crates/burn-ir/src/operation.rs
```diff
@@ -276,6 +276,12 @@ pub enum BaseOperationIr {
     /// Cast operation, no direct operation and should be supported by fusion backend.
     Cast(UnaryOpIr),
 
+    /// Operation corresponding to:
+    ///
+    /// Float => [cumsum](burn_tensor::ops::FloatTensorOps::float_cumsum).
+    /// Int => [cumsum](burn_tensor::ops::IntTensorOps::int_cumsum).
+    Cumsum(ReduceDimOpIr),
```
laggui: Technically speaking, this is not a reduction op. It doesn't collapse any dimension.

We could probably have a more general representation instead of `ReduceDimOpIr` to represent unary ops that operate along an axis/dim e.g. `AxisOpIr ` or `DimOpIr`.

PR #3811, opened 2025-09-30
Tensor::<agg>_dims()

crutcher:

### Changes

- `Tensor::sum_dims`
- `Tensor::mean_dims`
- `Tensor::max_dims`
- `Tensor::min_dims`
- `Tensor::prod_dims`
- `Tensor::max_abs_dims`

Improved docs for `Tensor::<agg>_dim` methods.
Unified to use `Self` in numeric ops where appropriate.


laggui (approved):
LGTM!

We could follow this up by moving the default impl at the trait level to allow backends to override it.

Issue #3780, opened 2025-09-24
Unit tests are failing on Mac


antimora:
It appears ever since we disabled Mac CI, we have accumulated issues on Mac. 

See tests errors on Mac:

```
uname -a
Darwin Mac.attlocal.net 24.6.0 Darwin Kernel Version 24.6.0: 
Mon Jul 14 11:30:55 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6031 arm64
```

[tch-errors.txt](https://github.com/user-attachments/files/22519438/tch-errors.txt)
[ndarray-errors.txt](https://github.com/user-attachments/files/22519439/ndarray-errors.txt)
[wgpu-errors.txt](https://github.com/user-attachments/files/22519437/wgpu-errors.txt)

Metal errors when I run:

`[burn-wgpu]% cargo test --release --features metal`

[metal-errors.txt](https://github.com/user-attachments/files/22519777/metal-errors.txt)

antimora:
CC @laggui

TheGhostHuCodes:
I was running `cargo run-checks` on my M2 mac, and also got some FP mis-matches. This was on `main`.

```
failures:

---- tests::autodiff::ad_div::tests::test_div_complex_2 stdout ----

thread 'tests::autodiff::ad_div::tests::test_div_complex_2' panicked at crates/burn-ndarray/src/lib.rs:41:5:
Tensors are not approx eq:
  => Position 0: 0.08251953 != 0.08333334
     diff (rel = +9.77e-3, abs = +8.14e-4), tol (rel = +5.00e-3, abs = +1.00e-5)
  => Position 2: -0.056803405 != -0.05555558
     diff (rel = +2.20e-2, abs = +1.25e-3), tol (rel = +5.00e-3, abs = +1.00e-5)
  => Position 3: -0.06800783 != -0.06714284
     diff (rel = +1.27e-2, abs = +8.65e-4), tol (rel = +5.00e-3, abs = +1.00e-5)

---- tests::autodiff::ad_log_sigmoid::tests::should_diff_log_sigmoid stdout ----

thread 'tests::autodiff::ad_log_sigmoid::tests::should_diff_log_sigmoid' panicked at crates/burn-ndarray/src/lib.rs:41:5:
Tensors are not approx eq:
  => Position 3: 0.001953125 != 0
     diff (rel = +1.00e0, abs = +1.95e-3), tol (rel = +5.00e-3, abs = +1.00e-5)

[...]

PR #3816, opened 2025-10-01
Implement LU decomposition

DimitriTimoz:
### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed. (fails for the same tests as https://github.com/tracel-ai/burn/issues/3780)

### Related Issues/PRs

Another step for #1538 

### Changes

Burn increasingly needs linear algebra operations. To solve linear systems, we need to start by implementing matrix factorization algorithms. `PA = LU` is one of them.
Since `factorization` algorithms require swap operations, I exposed the `swap_slices` function in the linalg module.

### Testing

Tested with [Wikipedia examples](https://en.wikipedia.org/wiki/LU_decomposition) as well as with incorrectly structured matrices (non-square or singular matrices). I also verified that multiplying the decomposition reproduces the original matrix.

### Notes

Backend implementations should be considered, since `libTorch` provides an LU decomposition function.
More efficient LU decomposition algorithms exist, such as Block LU decomposition, and they could lead to better performance. To track future progress in performance, implementing benchmarks could be a good idea.

### DIFF
--- a/crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
+++ b/crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
@@ -1,15 +1,7 @@
-use crate::{Int, backend::Backend, cast::ToElement, linalg::swap_slices, s, tensor::Tensor};
-use alloc::{format, string::String};
-
-/// The things that can go wrong when computing the LU decomposition.
-#[derive(Debug)]
-pub enum LuError {
-    /// The input matrix is not square.
-    NotSquareMatrix(String),
-    /// The input matrix is singular.
-    SingularMatrix(String),
-}
-
+use crate::{
+    Int, backend::Backend, cast::ToElement, check, check::TensorCheck, linalg::swap_slices, s,
+    tensor::Tensor,
+};
 /// Performs PLU decomposition of a square matrix.
 ///
 /// The function decomposes a given square matrix `A` into three matrices: a permutation vector `p`,
@@ -27,16 +19,18 @@ pub enum LuError {
 /// - A 2D tensor representing the combined `L` and `U` matrices.
 /// - A 1D tensor representing the permutation vector `p`.
 ///
-pub fn lu_decomposition<B: Backend>(
-    tensor: Tensor<B, 2>,
-) -> Result<(Tensor<B, 2>, Tensor<B, 1, Int>), LuError> {
+/// # Panics and numerical issues
+/// - The function will panic if the input matrix is singular or near-singular.
+/// - The function will panic if the input matrix is not square.
+/// # Performance note (synchronization / device transfers)
+/// This function may involve multiple synchronizations and device transfers, especially
+/// when determining pivot elements and performing row swaps. This can impact performance,
+pub fn lu_decomposition<B: Backend>(tensor: Tensor<B, 2>) -> (Tensor<B, 2>, Tensor<B, 1, Int>) {
+    check!(TensorCheck::is_square::<2>(
+        "lu_decomposition",
+        &tensor.shape()
+    ));
     let dims = tensor.shape().dims::<2>();
-    if dims[0] != dims[1] {
-        return Err(LuError::NotSquareMatrix(format!(
-            "LU decomposition requires a square matrix, but got shape: {:?}",
-            tensor.shape()
-        )));
-    }
     let n = dims[0];
 
     let mut permutations = Tensor::arange(0..n as i64, &tensor.device());
@@ -55,11 +49,8 @@ pub fn lu_decomposition<B: Backend>(
         let max = tensor.clone().slice(s![p, k]).abs();
 
         // Avoid division by zero
-        if max.into_scalar().to_f32().abs() < f32::EPSILON * 10.0 {
-            return Err(LuError::SingularMatrix(format!(
-                "LU decomposition failed: matrix is singular or near-singular at column {k}",
-            )));
-        }
+        let pivot = max.into_scalar();
+        check!(TensorCheck::lu_decomposition_pivot::<B>(pivot));
 
         if p != k {
             tensor = swap_slices(tensor, s![k, ..], s![p, ..]);
@@ -84,5 +75,5 @@ pub fn lu_decomposition<B: Backend>(
         }
     }
 
-    Ok((tensor, permutations))
+    (tensor, permutations)
 }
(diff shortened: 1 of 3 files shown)


DimitriTimoz:
I consider implementing Cholesky and QR decompositions too.

crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,88 @@
...
+/// * `tensor` - A square matrix to decompose, represented as a 2D tensor.
+///
+/// # Returns
+/// A tuple containing:
+/// - A 2D tensor representing the combined `L` and `U` matrices.
+/// - A 1D tensor representing the permutation vector `p`.
+///
+pub fn lu_decomposition<B: Backend>(
+    tensor: Tensor<B, 2>,
+) -> Result<(Tensor<B, 2>, Tensor<B, 1, Int>), LuError> {
```
laggui: Right now tensor ops panic even for incompatible inputs (see `TensorCheck`), so there is no `Result` return type in the API. We should keep this uniform.

crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,88 @@
...
+    for k in 0..n {
+        // Find the pivot row
+        let p = tensor
+            .clone()
+            .slice(s![k.., k])
+            .abs()
+            .argmax(0)
+            .into_scalar()
+            .to_usize()
+            + k;
```
laggui, line 48: This host-device sync point occurs at every iteration of the loop, which will result in very suboptimal performance.

You mentioned Block LU decomposition, which appears to operate on a block of columns at a time. This should be much better in terms of "out-of-the-box" performance if it can be implemented with high-level tensor ops.

Otherwise, it will likely require specialization. But that is likely too complex for now.

crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,88 @@
...
+        let max = tensor.clone().slice(s![p, k]).abs();
+
+        // Avoid division by zero
+        if max.into_scalar().to_f32().abs() < f32::EPSILON * 10.0 {
+            return Err(LuError::SingularMatrix(format!(
+                "LU decomposition failed: matrix is singular or near-singular at column {k}",
+            )));
+        }
```
laggui: This will add a sync point even for valid matrices. Right now the above pivot sync point will make it mostly free, but still something to consider 🙂

crates/burn-tensor/src/tests/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn test_lu_2x2_decomposition() {
+        let device = Default::default();
+        let tensor = TestTensor::<2>::from_data([[4.0, 3.0], [6.0, 3.0]], &device);
+        let (result, permutations) = lu_decomposition(tensor).unwrap();
+        let expected = TestTensor::<2>::from_data([[6.0, 3.0], [2.0 / 3.0, 1.0]], &device);
+        assert_eq!(
+            result.to_data(),
+            expected.to_data(),
+            "LU decomposition result does not match expected value"
+        );
```
laggui: Prefer using the equality checks defined for `TensorData`, e.g. `result.into_data().assert_eq(&expected.into_data(), true)`.

Also, the expected result does not need to be defined as a tensor. You can simply create the `TensorData` for the result.

crates/burn-tensor/src/tests/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,138 @@
...
+        assert_eq!(
+            permutations.to_data().to_vec::<i64>().unwrap(),
+            vec![1, 2, 0],
+            "LU decomposition permutations do not match expected value"
+        );
+        assert_eq!(
+            result.to_data(),
+            expected.to_data(),
+            "LU decomposition result does not match expected value"
+        );
```
laggui: Same here. And hardcoding `i64` is not portable across backends (e.g. wgpu and cuda use i32 by default, which would break the equality check).

crates/burn-tensor/src/tests/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn test_lu_1x1_element_matrix() {
+        let device = Default::default();
+        let tensor = TestTensor::<2>::from_data([[5.0]], &device);
+        let (result, permutations) = lu_decomposition(tensor).unwrap();
+        let expected = TestTensor::<2>::from_data([[5.0]], &device);
+        assert_eq!(
+            result.to_data(),
+            expected.to_data(),
+            "LU decomposition result does not match expected value for 1x1 matrix"
+        );
```
laggui: Same

crates/burn-tensor/src/tests/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,138 @@
...
+        let device = Default::default();
+
+        let tensor = TestTensor::<2>::eye(4, &device);
+        let (result, permutations) = lu_decomposition(tensor).unwrap();
+        let expected = TestTensor::<2>::eye(4, &device);
+        assert_eq!(
+            result.to_data(),
+            expected.to_data(),
+            "LU decomposition result does not match expected value for huge matrix"
+        );
```
laggui: Same

crates/burn-tensor/src/tests/linalg/lu_decomposition.rs
```diff
@@ -0,0 +1,138 @@
...
+        let reconstructed = p.matmul(l).matmul(u);
+        let abs_difference = (reconstructed - tensor).abs();
+        let max_difference = abs_difference.max();
+        let tolerance = 1e-6;
+        assert!(
+            max_difference.clone().into_scalar().to_f32() < tolerance,
+            "Random LU decomposition reconstruction error exceeds tolerance: {} >= {}",
+            max_difference.into_scalar().to_f32(),
+            tolerance
+        );
```
laggui: Same, we have `tensor_data.assert_approx_eq(...)`.

crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
DimitriTimoz: @laggui  one solution was to not check the matrix singularity and return a decomposition with NaNs values.

laggui: @DimitriTimoz Sorry for the delay, missed that notification!

For the current implementation I think that is fine, since there is already a sync point for the pivot. But we should add a performance note to the doc comments.

PR #3817, opened 2025-10-01
Add Tensor::sum_and_squeze_dims(dims)

crutcher:
### Checklist

### Changes

added one-shot `sum_and_squeeze_dims(dims)`

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -259,6 +259,7 @@ Those operations are available for numeric tensor kinds: `Float` and `Int`.
 | `tensor.sum()`                                                  | `tensor.sum()`                                 |
 | `tensor.sum_dim(dim)`                                           | `tensor.sum(dim, keepdim=True)`                |
 | `tensor.sum_dims(dims)`                                         | `tensor.sum(dims, keepdim=True)`               |
+| `tensor.sum_and_squeeze_dims(dims)`                             | `tensor.sum(dims, keepdim=False)`              |
 | `tensor.topk(k, dim)`                                           | `tensor.topk(k, dim).values`                   |
 | `tensor.topk_with_indices(k, dim)`                              | `tensor.topk(k, dim)`                          |
 | `tensor.tril(diagonal)`                                         | `torch.tril(tensor, diagonal)`                 |
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -581,6 +581,45 @@ where
         dims.iter().fold(self, |tensor, &dim| tensor.sum_dim(dim))
     }
 
+    /// Aggregate and squeeze along the given dimensions.
+    ///
+    /// This is equivalent to ``tensor.sum_dims(dims).squeeze_dims(dims)``
+    ///
+    /// # Arguments
+    ///
+    /// * `dims` - the dimensions to aggregate; supports negative indexing.
+    ///
+    /// # Returns
+    ///
+    /// The returned tensor will have the same rank,
+    /// but the aggregated dimensions will have size 1.
+    ///
+    /// # Example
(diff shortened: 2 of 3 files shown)


ArthurBrussee:
I'd like this for sum_dim as well. Not sure about sum_and_squeeze_dim though. Maybe

```
sum_dim_squeeze
sum_dims_squeeze
```

Also so that when you type `sum_d` you'd see both in autocomplete

laggui (commented):
Instead of adding a new method per reduction, what do you think about having a new `ReductionArgs` struct instead that allows specifying `keep_dims`?

```rust
pub struct ReductionArgs<'a, I: AsIndex> {
    pub dims: &'a [I],
    pub keep_dims: bool,
}

// Preserve current default
impl<'a, I: AsIndex> From<&'a [I]> for ReductionArgs<'a, I> {
    fn from(dims: &'a [I]) -> Self {
        Self {
            dims,
            keep_dims: true,
        }
    }
}

impl<'a, I: AsIndex, const N: usize> From<&'a [I; N]> for ReductionArgs<'a, I> {
    fn from(dims: &'a [I; N]) -> Self {
        Self {
            dims: &dims[..],
            keep_dims: true,
        }
    }
}
```

and then `sum_dims` can become:

```rust
pub fn sum_dims<'a, I: AsIndex + 'a, A: Into<ReductionArgs<'a, I>>>(self, args: A) -> Self {
    let args: ReductionArgs<I> = args.into();

    let mut result = args
        .dims
        .iter()
        .fold(self, |tensor, &dim| tensor.sum_dim(dim));

    if !args.keep_dims {
        let idims = ...; // map to isize
        result = result.squeeze_dims(&idims);
    }
    result
}
```

It also follows a similar pattern to other configurable APIs.

crutcher:
@laggui I like the idea of unifying the reduction args; but I don't know how to do it. The above example isn't going to work, because the D param of the return type isn't based upon the keep_dims value.

There's probably some way to exploit traits and generics to get there without invalidating existing call sites for `sum_dims`; but it would take some work.

it could also be something like `Tensor::reduce(self, Into<ReduceArgs>)` where the reduce operation and the dims and the keep_dims result were all wrapped up, and a const D2 was exposed. And the new weirdness wouldn't break existing callers.

laggui:
> @laggui I like the idea of unifying the reduction args; but I don't know how to do it. The above example isn't going to work, because the D param of the return type isn't based upon the keep_dims value.

Whoops, that didn't make it to my code example.

We can add the `D2` const generic, and the full implementation would look like:
```rust
    pub fn sum_dims<'a, I, A, const D2: usize>(self, args: A) -> Tensor<B, D2, K>
    where
        I: AsIndex + 'a,
        A: Into<ReductionArgs<'a, I>>,
    {
        let args: ReductionArgs<I> = args.into();

        let result = args
            .dims
            .iter()
            .fold(self, |tensor, &dim| tensor.sum_dim(dim));

        if !args.keep_dims {
            // TODO: remove idims when squeeze_dims uses AsIndex.
            let idims = args
                .dims
                .iter()
                .map(|&dim| canonicalize_dim(dim, D, false) as isize)
                .collect::<Vec<_>>();
            result.squeeze_dims::<D2>(&idims)
        } else {
            assert_eq!(D, D2); // TODO: TensorCheck
            Tensor::new(result.primitive)
        }
    }
```

crutcher:
Rebased, and changed to `sum_dims_squeeze`. Looking at the other thing now.

crutcher:
I tried the `ReductionArgs` idea; and it breaks all the `sum_dims` call sites that don't explicitly declare a rank typing.

I like the idea; but if we go that way, I think it should come from inside the house, because there are going to be a lot of follow-on effects.

laggui (approved):
We can revisit the reduction args in the future, but for now this is the best approach without too much noise (even if `sum_dims` was added during this release cycle).

PR #3823, opened 2025-10-03
Fix `mask_where` broadcasted line size

### LINKED ISSUE #3807: Buffer size validation error with broadcasting when default-features = false
**Describe the bug**

WGPU backend fails with buffer size validation error when performing `mask_where` operations with implicit broadcasting between tensors of different shapes (specifically [1] and [2]). The same code works correctly with the NdArray backend.

**⚠️ CRITICAL: This bug only occurs when `default-features = false` is set in Cargo.toml. Removing this flag makes the bug disappear.**

The error occurs during compute pass validation:
```
Buffer is bound with size 4 where the shader expects 8 in group[0] compact index 2
```

**To Reproduce**

1. Create a minimal Burn project with WGPU and NdArray backends **with `default-features = false`** (see Cargo.toml configuration below)
2. Run the following test code:
```rust
use burn::backend::wgpu::{Wgpu, WgpuDevice};
use burn::tensor::{Tensor, backend::Backend};

fn test_implicit_broadcast<B: Backend>(device: &B::Device) -> bool {
    // Parameters: [3]
    let params: Tensor<B, 1> = Tensor::from_floats([1.5, 0.5, 2.0], device);
    
    // Batch: [2]
    let batch: Tensor<B, 1> = Tensor::from_floats([1.0, 2.0], device);
    let condition: Tensor<B, 1> = Tensor::from_floats([1.0, 2.0], device);
    
    // Get parameter as [1] tensor
    let w0 = params.clone().slice([0..1]);  // [1]
    let w1 = params.clone().slice([1..2]);  // [1]
    
    // The bug: mask_where with [1] + complex broadcast chain
    let mask_result = Tensor::ones_like(&batch)
        .mask_where(condition.equal_elem(2.0), w0);  // [2] mask [1]
    
    let result = batch.clone() 
        * w1.exp()                    // [2] * [1] broadcast
        * mask_result;                // [2] * [2]
    
    // Force computation
    let _ = result.to_data();
    true
}

test_implicit_broadcast::<Wgpu>(&WgpuDevice::default());
```
3. Run with `RUST_BACKTRACE=full cargo run`
4. Observe the panic with buffer size mismatch error

**Expected behavior**

The WGPU backend should handle implicit broadcasting in `mask_where` operations the same way as the NdArray backend. When masking a [2] tensor with a [1] tensor, it should automatically broadcast the [1] tensor to match the [2] shape, allocating appropriate buffer sizes for the GPU computation.

**Actual behavior**

The WGPU backend panics with:
```
wgpu error: Validation Error

Caused by:
  In ComputePass::end
    In a dispatch command, indirect:false
      Buffer is bound with size 4 where the shader expects 8 in group[0] compact index 2
```

This suggests that the WGPU backend is allocating a buffer for 1 f32 element (4 bytes) when it should allocate for 2 f32 elements (8 bytes) due to broadcasting.

**Environment:**
 - OS: macOS (Darwin 23.5.0)
 - Burn version: 0.17.1
 - WGPU version: 25.0.2 (via Burn dependency)
 - Rust toolchain: stable-aarch64-apple-darwin

**Additional context**

- The bug is specifically related to implicit broadcasting in `mask_where` operations
- The issue appears to be in the buffer size calculation for broadcasted tensors in the WGPU compute pipeline
- NdArray backend correctly handles the same broadcasting scenario
- The error occurs at the compute pass validation stage, indicating a mismatch between what the shader expects and what buffer was bound
- This is a minimal reproduction case - the original issue was discovered in a more complex model training scenario involving FSRS algorithm implementation

**Configuration Dependency:**

❌ **FAILS** with:
```toml
[dependencies.burn]
version = "0.17.1"
default-features = false
features = ["ndarray", "wgpu"]
```

✅ **WORKS** with:
```toml
[dependencies.burn]
version = "0.17.1"
features = ["ndarray", "wgpu"]
```

This suggests that one or more of Burn's default features include code paths or optimizations that correctly handle broadcasting in WGPU operations. The bug appears to be in the minimal feature set implementation.

**Workaround:**

Remove `default-features = false` from your `Cargo.toml` to enable default features, which resolves the issue.

**Minimal reproduction repository structure:**
```
Cargo.toml (with the problematic configuration):
[dependencies.burn]
version = "0.17.1"
default-features = false  # ⚠️ This triggers the bug
features = ["ndarray", "wgpu"]
```

Full reproduction code is available in the attached `main.rs` file.

---

**Summary:** This appears to be a feature-specific bug where the minimal feature set (`default-features = false`) fails to properly handle broadcasting in WGPU backend, while the default feature set handles it correctly. This suggests missing broadcast logic or buffer size calculation code when certain default features are disabled.

[burn-wgpu-broadcast-bug.zip](https://github.com/user-attachments/files/22619654/burn-wgpu-broadcast-bug.zip)

laggui:
### Checklist

### Related Issues/PRs

Fixes #3807 

### Changes

Select line size based on input, mask and value.

### Testing

Added test.

### DIFF
--- a/crates/burn-cubecl/src/kernel/mask/mask_where.rs
+++ b/crates/burn-cubecl/src/kernel/mask/mask_where.rs
@@ -4,7 +4,7 @@ use crate::{
     BoolElement, CubeRuntime,
     element::CubeElement,
     kernel::utils::{broadcast_shape, linear_view, linear_view_alias, linear_view_ref},
-    ops::{max_line_size, numeric::empty_device},
+    ops::{max_line_size, max_line_size_many, numeric::empty_device},
     tensor::CubeTensor,
 };
 
@@ -46,7 +46,7 @@ pub fn mask_where<R: CubeRuntime, E: CubeElement, BT: BoolElement>(
     strategy: MaskWhereStrategy,
 ) -> CubeTensor<R> {
     let cube_dim = CubeDim::default();
-    let line_size = max_line_size(&input);
+    let line_size = max_line_size_many(&[&input, &mask, &value], input.shape.num_dims() - 1);
     let cube_count =
         calculate_cube_count_elemwise(input.shape.num_elements() / line_size as usize, cube_dim);
 
--- a/crates/burn-tensor/src/tests/ops/mask.rs
+++ b/crates/burn-tensor/src/tests/ops/mask.rs
@@ -96,6 +96,19 @@ mod tests {
     }
 
     #[test]
+    fn should_support_mask_where_broadcast_value_small() {
+        let device = Default::default();
+        let tensor = TestTensorInt::<1>::arange(2..4, &device);
+        let mask = TestTensorBool::<1>::from_bool(TensorData::from([true, false]), &device);
+        let value = TestTensor::<1>::ones([1], &device);
+
+        let output = tensor.float().mask_where(mask, value);
+        let expected = TensorData::from([1., 3.]);
+
+        output.into_data().assert_eq(&expected, false);
+    }
+
+    #[test]
     fn should_handle_mask_where_nans() {
         let device = Default::default();
         let tensor = TestTensor::from_data(


L-M-Sherlock:
It passed my minimal reproducible example, thanks!

PR #3824, opened 2025-10-03
fix(onnx-ir): rank inference for bool ops with broadcasting

mooori:
### Checklist

### Changes

Broadcasting for ONNX binary ops `bool_and, bool_or, bool_xor` is not yet supported, see tracel-ai/burn-onnx#16. I suppose that's why currently rank inference for these ops with `same_as_input` instead of `same_as_input_broadcast` doesn't cause problems.

**When it will matter**

Currently I'm taking a stab at implementing broadcasting for ONNX `bool_and` and ran into the following issue:

- For `bool_and` of `lsh (rank 2)` with `rhs (rank 3)` the ONNX graph correctly has `output (rank 3)` as broadcasting is applied.
- Hovewer, the Rust code generated from that model wrongly infers `output (rank 2)`.

This is because currently for `And` the output rank is inferred with `same_as_input`, which equals the rank of the output to the rank of the first input. Can be fixed by instead using `same_as_input_broadcast`.


mooori:
I've sent the PR: tracel-ai/burn#3829.

Tracking down that `same_as_input` needs to be changed to `same_as_input_broadcast` was tricky, at least for me as a new contributor. Because there was no error pointing to it, just "mysteriously" the Rust code generated from the ONNX model had the wrong rank for an output tensor.

That's why I've prepared this PR here to eagerly change it to `same_as_input_broadcast`, hoping to make adding broadcast support for the remaining bool ops easier.

laggui:
I guess this can be closed now? The remaining bool support can be added in follow-ups to #3829

laggui (commented):
I guess this can be closed now? The remaining bool support can be added in follow-ups to #3829

Issue #3532, opened 2025-08-13
Expose backend supported types


laggui:
Brought up on discord.

Right now we can switch the backend types easily, which is nice, but you need prior knowledge on the supported dtypes (e.g. f16 is supported on Cuda, but not Wgpu unless you're using Vulkan).

Feature request: backend-level API to query which dtypes are supported. Similar to `CubePrimitive::is_supported(client)`.

Puranjay-del-Mishra:
@laggui 
Hey I can take this up, would appreciate if you would like to see it implemented in a certain way!

Edit:
Implementation plan for the data type support api:

1. Add DTypeSupport + HasDTypeSupport in burn and re-export from burn/src/backend.rs.
2. Implement the trait per backend type, with optional per-device caching.
3. Expose a simple call (<B as HasDTypeSupport>::dtype_support(&device)) and add a small example/test to print the supported dtypes.

laggui:
Probably just need to have a function in the backend trait, e.g.

```rust
fn supports_dtype(device: &Self::Device, dtype: DType) -> bool;
fn supports_element<E: Element>(device: &Self::Device) -> bool {
    Self::supports_dtype(device, E::dtype())
}
```

Puranjay-del-Mishra:
@laggui 
I'm working to make a PR for the dtype support. Can you help validate that the following would work for this? I'm new to Burn and would appreciate the help! :

**Current design (small & opt-in)**

```
pub trait BackendDType: Backend {
    fn supports_dtype(device: &Self::Device, dtype: DType) -> bool;
    #[inline]
    fn supports_element<E: Element>(device: &Self::Device) -> bool {
        Self::supports_dtype(device, E::dtype())
    }
}

```

- WGPU impl: F32 | I32 | Bool = true; F16 = wgpu::Features::SHADER_F16; everything else false.
- Wrapper impls: Autodiff<B> and Router<B> forward to B.

Questions:

1. Semantics: Should supports_dtype mean “math OK in kernels” (arithmetic), not just “can store & cast”? (I’m implementing it as arithmetic.)
2. Router: prefer intersection (safe everywhere) or union (pick best device per op)? I forwarded 1:1 for now.
3. Probe location: OK to keep the tiny feature_probe helper inside burn-wgpu so the root crate never touches wgpu?
4. Backends mapping: any recommended signals? :

  - CUDA/ROCm: gate F16 via device caps (SM / HIP attribute). Which minimal check do you want me to use?
[...]

PR #3826, opened 2025-10-03
slice_fill() should pick a compatible dtype

crutcher:
### Checklist

### Related Issues/PRs

Encountered `slice_fill()` failing for a F64 tensor.

### Changes

create the fill value tensor with the correct dtype.

### Testing

I need help with the tests, there's conditional availability I don't understand for backend types.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -2864,7 +2864,11 @@ pub trait BasicOps<B: Backend>: TensorKind<B> {
         let slice_shape_vec = calculate_slice_output_shape(slices, &tensor_shape.dims);
         let slice_shape = Shape::from(slice_shape_vec);
 
-        let value = Self::from_data(TensorData::from([value]), &Self::device(&tensor));
+        let value = Self::from_data_dtype(
+            TensorData::from([value]),
+            &Self::device(&tensor),
+            Self::dtype(&tensor),
+        );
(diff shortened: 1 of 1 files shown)


laggui (approved):
> I need help with the tests, there's conditional availability I don't understand for backend types.

Pending https://github.com/tracel-ai/burn/issues/3532 we can always cast the input to f32 using `.cast(DType::F32);`. This is what we currently do to test this behavior across some tests.

For the default dtype tested (f32), it will be a no-op but for f16 / bf16 tests for example it will cast the input to f32 while having a different default precision.

e.g.
https://github.com/tracel-ai/burn/blob/b198ef80a7374cb3b778b2d71b30e16a5e48afc5/crates/burn-tensor/src/tests/ops/cat.rs#L105-L119

Approving the changes nonetheless, will wait for the test.

crutcher:
I don't know how to gate this (with the current code) on the backend/supported-gtypes.

I feel like we need a `testgen_dtypes!([dtype, dtype, ...])`?

laggui (commented):
We have that but on cubecl backends, e.g. `burn_cubecl::testgen_all!([f16, f32], [i8, i16, i32, i64], [u8, u32]);`

But otherwise, we need to have a test that will run without gating for now (not supported, as per my previous comment). So we can always cast to f32, and on backends that are tested with different dtypes (like cuda/wgpu/vulkan), the cast will work.

The current tests should work, just need to run cargo fmt!

PR #3842, opened 2025-10-07
Feat/persistent memory

nathanielsimard:
This PR uses persistent memory https://github.com/tracel-ai/cubecl/pull/947 for model parameters, both for training and inference.

There is a CubeCL setting to deactivate persistent memory or to enforce it for all allocations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1875,7 +1875,6 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.7.0"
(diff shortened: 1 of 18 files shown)


nanookclaw:
Will address:
1) Will implement `LinearWarmup` and `CosineWarmup` as types that impl the existing `LrScheduler` trait rather than a new trait
2) Will add a note in the docs: warmup schedules affect learning rate only — weight decay is applied independently by the optimizer and is not modulated by warmup

Will push an update.

PR #3845, opened 2025-10-08
Refactor `Shape` manipulations

laggui:
### Checklist

### Changes

Refactored all shapes manipulations as part of `Shape` interface (deduplicate code, reduce errors)

- Interpret as slice / iter over dims
- swap, premute, repeat, cat, slice, broadcast, matmul

Also updated the custom kernel examples in the book that weren't totally up to date.

Will follow-up with burn-ir simplifications to automatically create the appropriate outputs and further reduce code duplication and possible error paths.

### Testing

Added tests


crates/burn-tensor/src/tensor/shape.rs
```diff
@@ -1,15 +1,47 @@
...
+};
+use serde::{Deserialize, Serialize};
 
 /// Shape of a tensor.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct Shape {
     /// The dimensions of the tensor.
     pub dims: Vec<usize>,
 }
```
laggui, line 14: Most of the additions are in the `Shape` impl (including tests). Lmk what you think about the interface.

PR #3847, opened 2025-10-08
Top level handling of fast-path int-power exponents.

crutcher:
### Checklist

### Changes

- introduced float_pow{i,f}_scalar{,_fallback} tower.
- special case [-2, -1, 0, 1, 2].
- introduced `tensor.square()`

Backends have a choice of which part of the tower to override; with the minimal override being `float_powf_scalar_fallback`.

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1778,7 +1778,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
         }
     }
 
-    fn float_powf_scalar_fallback(tensor: FloatTensor<Self>, value: f32) -> FloatTensor<Self> {
+    fn float_powf_scalar_impl(tensor: FloatTensor<Self>, value: f32) -> FloatTensor<Self> {
         #[derive(Debug)]
         struct PowfScalar;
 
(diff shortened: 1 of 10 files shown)


crutcher:
I commented out the `t^0` case for now; I think we want to match the IEEE 754 recommendations, but I want alignment on that.

powi(_, 0) => 0

powr(NaN, 0.0) => NaN
powr(otherwise, 0.0) => 0.0

crutcher:
there's some weirdness about conflicting requirements/checks on the import of the float traits; I think this will pass CI ...

crutcher:
IEEE 754 seems to argue that `NaN**0 == 1`, and `pytorch` also does that; so updated the 0 case.

crutcher:
@laggui , @wingertge

crutcher:
I didn't build a full-map documentation of why I built it the way that I did; but I don't think handling it in the Int specialization is sufficient.

- when a float tensor calls pow{i,f}_scalar([-2, -1, 0, 1, 2]), I want that to be accelerated by default for all backends; that's actually the case that has the most performance impact on us, not the int=>float=>int handling of an int tensor's powi_scalar,
- I want a pretty-good answer for backends that requires minimal kernel participation. As currently implemented, the tower cascades all the non-special cases back down to float float_powf_scalar_fallback, which is just the renamed float_powf_scalar.,
- I want backends that do have more specific kernels to be able to override and skip the entire tower, if they need to.

crutcher:
Updated based upon discord conv with @laggui

PR #3849, opened 2025-10-09
PyTorch reader: Add F16, BF16, and unsigned integer support

antimora:
### Checklist

### Related Issues/PRs

Originally reported: https://discord.com/channels/1038839012602941528/1166763685822333009/1425842677496283247

### Changes

**Problem:** The PyTorch pickle reader had incomplete dtype support, with missing implementations for F16, BF16, and U8 types. Additionally, element sizes were hardcoded throughout the binary parsing code, and default fallback cases silently returned incorrect f32 zeros instead of proper errors.

**Solution:**

1. Implemented Missing Data Types
  - Added F16 (half precision) reading with proper 2-byte parsing
  - Added BF16 (bfloat16) reading with proper 2-byte parsing
  - Added U8 reading for unsigned 8-bit integers
  - Added U16, U32, U64 for complete unsigned integer support
2. Replaced Hardcoded Element Sizes
  - Changed from hardcoded match statement to dtype.size() from base.rs:330-362
  - Replaced all hardcoded byte indexing (i * 8, i * 4, etc.) with element_size variable
  - Now single source of truth for element sizes
3. Fixed Storage Type Mappings
  - Separated ByteStorage (U8) from CharStorage (I8) for correct PyTorch compatibility
  - Updated both modern and legacy format parsers
4. Improved Error Handling 
  - Changed default fallback from silent f32 zeros to explicit DataError
  - Added proper zero fallback cases for all supported dtypes (F16, BF16, U8, U16, U32, U64)
5. Enhanced Test Coverage
  - Added module-level precision standards documentation
  - Enhanced mixed_types, nested_dict, checkpoint, complex_structure tests to load actual data
  - Standardized all floating-point precision comparisons (F16/BF16: 1e-2, F32: 1e-6, F64: 1e-10)
  - Added uint8 test with data verification
  - Updated test_data.py to generate uint8.pt test file

Coverage: Now supporting all DType variants:
- Floats: F32, F64, F16, BF16
- Signed Ints: I64, I32, I16, I8
- Unsigned Ints: U64, U32, U16, U8
- Bool

### Testing

- All 45 existing tests passing
- Added uint8 test with actual value verification
- Enhanced 4 existing tests (mixed_types, nested_dict, checkpoint, complex_structure) to load and verify actual tensor
data
- F16 test verifies values
- BF16 test verifies values
- Generated test data using uv run test_data.py
- Verified precision consistency across

### DIFF
--- a/crates/burn-store/src/pytorch/pickle_reader.rs
+++ b/crates/burn-store/src/pytorch/pickle_reader.rs
@@ -487,6 +487,7 @@ fn rebuild_tensor_v2(
     }
 
     // Create a TensorSnapshot with a closure that loads the actual data on-demand
+    // TODO extract this long code into stand along function (part of a bigger refactor)
     Ok(Object::TorchParam(TensorSnapshot::from_closure(
         Rc::new(move || {
             // Load data only when needed


crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -484,24 +487,16 @@ fn rebuild_tensor_v2(
     }
 
     // Create a TensorSnapshot with a closure that loads the actual data on-demand
+    // TODO extract this long code into stand along function (part of a bigger refactor)
```
laggui, line 490: Agreed 😅 especially the big match arms to read the bytes into the proper type.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -537,7 +532,9 @@ fn rebuild_tensor_v2(
                         let mut values = Vec::with_capacity(num_elements);
                         for i in 0..elements_to_read {
                             let mut bytes = [0u8; 8];
-                            bytes.copy_from_slice(&data_slice[i * 8..(i + 1) * 8]);
+                            bytes.copy_from_slice(
+                                &data_slice[i * element_size..(i + 1) * element_size],
+                            );
                             values.push(f64::from_le_bytes(bytes));
                         }
                         values.resize(num_elements, 0.0);
```
laggui, line 540: I think these match arms could be factored into a single function that is called for each dtype, given that we store the raw bytes anyway in `TensorData`.

For example:
```rust
fn tensor_from_raw(
    data_slice: &[u8],
    num_elements: usize,
    element_size: usize,
    shape: Vec<usize>,
    dtype: DType,
) -> TensorData {
    let total_bytes = num_elements * element_size;
    let mut bytes = Vec::with_capacity(total_bytes);
    bytes.extend_from_slice(data_slice);
    bytes.resize(total_bytes, 0);
    TensorData::from_bytes_vec(bytes, shape, dtype)
}
```

PR #3865, opened 2025-10-14
Fix concurrency problems

nathanielsimard:
There is currently a deadlock on the main thread involving the fusion channel mutex. To resolve this, lock the fusion channel only when the current thread has permission to execute operations on the inner backend. This prevents deadlocks, as the inner backend can register fusion tasks that might otherwise cause blocking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1852,6 +1852,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.7.0"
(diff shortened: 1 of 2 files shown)


crutcher:
I tried pulling this and running `clockmill` > `fishbowl` against it; and I'm seeing a 3/4th slow down in tics-per-second when running a pair of "sim thread / render thread".
https://github.com/crutcher/clockmill/tree/main

PR #3869, opened 2025-10-15
Add `cautious_weight_decay` to AdamW optimizer.

crutcher:
### Checklist
- [?] Made sure the book is up to date with changes in this PR.

### Changes

- Implement cautious weight decay as described in https://arxiv.org/abs/2510.12402.
- Update decay logic to support cautious mode.
- Add configuration and state handling for `cautious_weight_decay`.
- Refactor beta factors in momentum updates for readability.

### Testing

Numerical smoke test.

### DIFF
--- a/crates/burn-optim/src/optim/adamw.rs
+++ b/crates/burn-optim/src/optim/adamw.rs
@@ -12,7 +12,7 @@ use crate::{LearningRate, grad_clipping::GradientClippingConfig};
 #[allow(unused_imports)]
 use num_traits::Float as _;
 
-/// AdamW configuration.
+/// [`AdamW`] Configuration.
 #[derive(Config, Debug)]
 pub struct AdamWConfig {
     /// Parameter for AdamW.
@@ -29,6 +29,7 @@ pub struct AdamWConfig {
     weight_decay: f32,
 
     /// Cautious weight decay config.
+    ///
     /// See: https://arxiv.org/abs/2510.12402
     #[config(default = false)]
     cautious_weight_decay: bool,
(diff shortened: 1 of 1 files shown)


crutcher:
Updated to run against the MNIST example.

AdamW without CWD:
Screenshot from 2025-10-16 17-43-28

AdamW with CWD:
Screenshot from 2025-10-16 17-54-30

crates/burn-optim/src/optim/adamw.rs
```diff
@@ -60,15 +67,32 @@ impl<B: Backend> SimpleOptimizer<B> for AdamW {
...
+            tensor.clone()
+        } else if self.cautious_weight_decay {
+            // Cautious weight decay.
+            // See: https://arxiv.org/abs/2510.12402
+            let tensor_pos = tensor.clone().greater_elem(0.0);
+            let grad_pos = raw_delta.clone().greater_elem(0.0);
+            let differ = tensor_pos.not_equal(grad_pos);
+
+            // Zero out the decay where the decay is counter to the update direction.
+            tensor.clone() - tensor.mul_scalar(decay_rate).mask_fill(differ, 0.0)
```
laggui, line 91: I think this is slightly incorrect!

Looking at algorithm 2 of the paper, it appears they compare with the first moment (`m_t` in the paper, or `momentum_state.momentum_1` in the code). Not with `raw_delta`, which is the bias-corrected, scaled update.

image


Also, you could forgo the three comparisons and stick to the element-wise multiplication for same-sign check as detailed in the algo.

```rust
let same_sign = tensor.clone()
    .mul(momentum_state.moment_1.clone())
    .greater_equal_elem(0.0);
let differ = same_sign.bool_not();
```

crutcher:
Applied the changes.
image

crates/burn-optim/src/optim/adamw.rs
crutcher: Done

PR #3872, opened 2025-10-16
ONNX-IR Refactor - Op/Node-Centric Architecture

antimora:
This PR refactors the ONNX IR layer to be more op/node-centric, significantly simplifying future op additions and maintenance. The previous design made limiting assumptions that didn't scale well as more ops were added.

### Checklist

### Changes

- Simplified Op Model: Redesigned around individual operations and nodes for better scalability
- Removed TensorData Indirection: Now uses burn-tensor's native tensor data directly, enabling mmap support and better type handling
- Opset-Aware Processing: Processors now accept opset parameters, allowing flexible condition testing across different ONNX versions
- Constants-First Approach: All values treated as constants initially (with initializers and identities converted)
- Explicit Static Input Declaration: Users explicitly declare which inputs are static, rather than preemptive conversion
- Transformation Pipeline: New multi-phase transformation pipeline for better visibility into the processing flow
- Native Type Integration: Direct use of TensorData and Dtype for improved efficiency and consistency. Next we will use mmap to reduce memory overhead.
- Graph input name preservation (sanitized) for easy development and troubleshooting.

### Testing

- new onnx-ir onnx tests that cover the onnx-ir infrastructure and edge cases
- all existing tests pass
- all model-checks pass

### DIFF
--- a/crates/onnx-ir/src/node/argmax.rs
+++ b/crates/onnx-ir/src/node/argmax.rs
@@ -4,24 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__ArgMax.html>
 //!
-//! ## Attributes
-//!
-//! - `axis` (int64, default=0): The axis along which to compute the arg indices.
-//!   Accepted range is [-r, r-1] where r = rank(data).
-//! - `keepdims` (int64, default=1): Keep the reduced dimension or not. 1 means keep reduced dimension.
-//! - `select_last_index` (int64, default=0): Whether to select the last index or the first index
-//!   if the maximum value appears in multiple indices.
-//!
-//! ## Inputs
-//!
-//! - `data` (T): Input tensor to compute argmax on.
-//!
-//! ## Outputs
-//!
-//! - `reduced` (tensor(int64)): Output tensor of indices with datatype int64. Shape depends on `keepdims`:
-//!   - If keepdims=1: same rank as input, reduced dimension becomes size 1
-//!   - If keepdims=0: rank reduced by 1
-//!
 //! ## Opset Versions
 //!
 //! - **Opset 13**: Added bfloat16 to type constraints
--- a/crates/onnx-ir/src/node/argmin.rs
+++ b/crates/onnx-ir/src/node/argmin.rs
@@ -4,22 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__ArgMin.html>
 //!
-//! ## Attributes
-//!
-//! - `axis` (int, default=0): The axis along which to compute the arg indices. Accepted range is
-//!   [-r, r-1] where r = rank(data).
-//! - `keepdims` (int, default=1): Keep the reduced dimension or not. 1 means keep reduced dimension.
-//! - `select_last_index` (int, default=0): Whether to select the last index or the first index if
-//!   the minimum appears in multiple indices. 0 (default) selects the first index.
-//!
-//! ## Inputs
-//!
-//! - `data` (T): An input tensor.
-//!
-//! ## Outputs
-//!
-//! - `reduced` (tensor(int64)): Reduced output tensor with integer data type.
-//!
 //! ## Opset Versions
 //!
 //! - **Opset 11**: Initial supported version.
--- a/crates/onnx-ir/src/node/arithmetic.rs
+++ b/crates/onnx-ir/src/node/arithmetic.rs
@@ -10,19 +10,6 @@
 //! - Mul: <https://onnx.ai/onnx/operators/onnx__Mul.html>
 //! - Div: <https://onnx.ai/onnx/operators/onnx__Div.html>
 //!
-//! ## Attributes
-//!
-//! None - these operators have no attributes.
-//!
-//! ## Inputs
-//!
-//! - `A` (T): First operand, any rank
-//! - `B` (T): Second operand, any rank (must be type-compatible with A)
-//!
-//! ## Outputs
-//!
-//! - `C` (T): Result of element-wise operation (supports multidirectional broadcasting)
-//!
 //! ## Type Constraints
 //!
 //! T: Numeric tensor types (float16, float32, float64, int8, int16, int32, int64, uint8, uint16, uint32, uint64)
--- a/crates/onnx-ir/src/node/attention.rs
+++ b/crates/onnx-ir/src/node/attention.rs
@@ -4,28 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__Attention.html>
 //!
-//! ## Attributes
-//!
-//! - `is_causal` (int64, default=0): Enable causal masking (lower triangular)
-//! - `kv_num_heads` (int64, optional): Number of key/value heads (required for 3D inputs)
-//! - `q_num_heads` (int64, optional): Number of query heads (required for 3D inputs)
-//! - `qk_matmul_output_mode` (int64, default=0): Which intermediate to output (0-3)
-//! - `scale` (float, optional): Scaling factor for Q*K^T
-//! - `softcap` (float, default=0.0): Softcap value for attention weights
-//! - `softmax_precision` (int64, optional): Precision mode for softmax
-//!
-//! ## Inputs
-//!
-//! - `Q`, `K`, `V` (T): Query, key, value tensors (rank 3 or 4)
-//! - `attention_mask` (optional): Boolean or float mask
-//! - `past_key`, `past_value` (optional): KV cache inputs
-//!
-//! ## Outputs
-//!
-//! - `Y` (T): Attention output (rank 3 or 4)
-//! - `present_key`, `present_value` (optional): Updated KV cache
-//! - `qk_matmul_output` (optional): Intermediate result
-//!
 //! ## Opset Versions
 //!
 //! - **Opset 23**: Initial version with multi-head attention support (MHA, GQA, MQA variants)
--- a/crates/onnx-ir/src/node/avg_pool1d.rs
+++ b/crates/onnx-ir/src/node/avg_pool1d.rs
@@ -4,21 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__AveragePool.html>
 //!
-//! ## Attributes
-//! - `kernel_shape` (ints, required): Kernel size
-//! - `strides` (ints, default=1): Stride
-//! - `pads` (ints, default=0): Padding
-//! - `dilations` (ints, default=1): Dilation (opset 10+)
-//! - `auto_pad` (string, default="NOTSET"): Padding mode (only `NOTSET` supported)
-//! - `count_include_pad` (int, default=0): Include padding in average calculation
-//! - `ceil_mode` (int, default=0): Use ceil for output shape (not supported)
-//!
-//! ## Inputs
-//! - `X` (T): Input tensor (N x C x L)
-//!
-//! ## Outputs
-//! - `Y` (T): Output tensor
-//!
 //! ## Opset Versions
 //! - **Opset 7**: Initial AveragePool operator
 //! - **Opset 10**: Added dilations attribute support
--- a/crates/onnx-ir/src/node/avg_pool2d.rs
+++ b/crates/onnx-ir/src/node/avg_pool2d.rs
@@ -4,21 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__AveragePool.html>
 //!
-//! ## Attributes
-//! - `kernel_shape` (required, ints): Kernel size \[height, width\]
-//! - `strides` (ints, default=\[1, 1\]): Stride \[height, width\]
-//! - `pads` (ints, default=\[0, 0, 0, 0\]): Padding \[top, left, bottom, right\]
-//! - `dilations` (ints, default=\[1, 1\]): Dilation \[height, width\] (opset 10+)
-//! - `auto_pad` (string, default="NOTSET"): Padding mode (only `NOTSET` supported)
-//! - `count_include_pad` (int, default=0): Include padding in average calculation
-//! - `ceil_mode` (int, default=0): Use ceil for output shape (not supported)
-//!
-//! ## Inputs
-//! - `X` (T): Input tensor (N x C x H x W)
-//!
-//! ## Outputs
-//! - `Y` (T): Output tensor
-//!
 //! ## Opset Versions
 //! - **Opset 7**: Initial AveragePool operator
 //! - **Opset 10**: Added dilations attribute support
--- a/crates/onnx-ir/src/node/batch_norm.rs
+++ b/crates/onnx-ir/src/node/batch_norm.rs
@@ -4,25 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__BatchNormalization.html>
 //!
-//! ## Attributes
-//! - `epsilon` (float, default=1e-5): Numerical stability constant
-//! - `momentum` (float, default=0.9): Momentum for running statistics
-//! - `training_mode` (int, default=0): Training mode flag
-//!
-//! ## Inputs
-//! - `X` (T): Input tensor (N x C x ...)
-//! - `scale` (T): Scale tensor (C)
-//! - `B` (T): Bias tensor (C)
-//! - `input_mean` (T): Mean tensor (C)
-//! - `input_var` (T): Variance tensor (C)
-//!
-//! ## Outputs
-//! - `Y` (T): Normalized output tensor
-//! - `mean` (T, optional): Running mean (training mode)
-//! - `var` (T, optional): Running variance (training mode)
-//! - `saved_mean` (T, optional): Saved mean (training mode)
-//! - `saved_var` (T, optional): Saved variance (training mode)
-//!
 //! ## Opset Versions
 //! - **Opset 1-5**: Initial version with spatial attribute
 //! - **Opset 6-8**: Removed spatial attribute, added consumed_inputs
--- a/crates/onnx-ir/src/node/bernoulli.rs
+++ b/crates/onnx-ir/src/node/bernoulli.rs
@@ -4,22 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__Bernoulli.html>
 //!
-//! ## Attributes
-//!
-//! - `dtype` (int64, optional): The data type for the elements of the output tensor.
-//!   If not specified, the data type of the input tensor is used.
-//! - `seed` (float, optional): Optional seed to the random generator.
-//!   If not specified, one will be auto-generated.
-//!
-//! ## Inputs
-//!
-//! - `input` (T1): Input tensor containing probabilities in the range [0, 1].
-//!   Each value p represents the probability of drawing 1 (with probability p) or 0 (with probability 1-p).
-//!
-//! ## Outputs
-//!
-//! - `output` (T2): Output tensor with same shape as input, containing only values 0 or 1.
-//!
 //! ## Opset Versions
 //!
 //! - **Opset 15**: Initial version with dtype and seed attributes for drawing binary random numbers
--- a/crates/onnx-ir/src/node/bitshift.rs
+++ b/crates/onnx-ir/src/node/bitshift.rs
@@ -4,20 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__BitShift.html>
 //!
-//! ## Attributes
-//! - `direction` (string, required): Direction of moving bits. Can be either "RIGHT" (for right shift)
-//!   or "LEFT" (for left shift). When direction is "RIGHT", the operator moves the binary
-//!   representation toward the right side, effectively decreasing the input value. When direction
-//!   is "LEFT", bits move toward the left side, increasing the actual value.
-//!   Note: Implementation provides default "left" despite spec marking as required.
-//!
-//! ## Inputs
-//! - `X` (T, required): Tensor to be shifted
-//! - `Y` (T, required): Tensor specifying the amounts of shifting (number of bits to shift)
-//!
-//! ## Outputs
-//! - `Z` (T): Output tensor with shifted values
-//!
 //! ## Type Constraints
 //! - T in (tensor(uint8), tensor(uint16), tensor(uint32), tensor(uint64))
 //!
--- a/crates/onnx-ir/src/node/cast.rs
+++ b/crates/onnx-ir/src/node/cast.rs
@@ -4,23 +4,6 @@
 //!
 //! **ONNX Spec**: <https://onnx.ai/onnx/operators/onnx__Cast.html>
 //!
-//! ## Attributes
-//! - `to` (required): Target data type specified as an integer from the TensorProto DataType enum.
-//!   Supported types include: bool, float16, float32, float64, int8, int16, int32, int64,
-//!   uint8, uint16, uint32, uint64, string, and various float8 formats.
-//! - `saturate` (optional, since opset 19): Defines conversion behavior when input value is out of
-//!   range of destination type. Only applies for float8 conversions. Default: true.
-//! - `round_mode` (optional, since opset 21): Rounding mode for float8e8m0 conversion.
-//!   Values: "up" (default), "down", "nearest".
-//!
-//! ## Inputs
-//! - `input` (T1): Input tensor to be cast. Supports heterogeneous types including all numeric types,
-//!   bool, and string. Casting from complex types is not supported.
-//!
-//! ## Outputs
-//! - `output` (T2): Output tensor with the same shape as input but with the target data type.
-//!   Supports casting to all numeric types, bool, and string. Casting to complex types is not supported.
-//!
 //! ## Opset Versions
 //! - **Opset 1-5**: Basic cast operation with core numeric types
 //! - **Opset 6-8**: Extended type support for additional numeric types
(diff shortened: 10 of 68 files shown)


contributor-book/src/guides/onnx-to-burn-conversion-tool.md
```diff
@@ -212,85 +233,213 @@ That's it! The registry automatically generates:
...
+
+- Removes unreferenced constant nodes
+- Constructs the final `OnnxGraph` with inputs, outputs, and nodes
+
+This pipeline ensures that:
+
+- Type information flows correctly through the graph
+- Constants are accessible where needed
+- The graph is optimized (no unnecessary Identity nodes)
+- Only referenced nodes are included in the final output
```
laggui: These bullet points are essentially a repeat of the phases described just before

contributor-book/src/guides/onnx-to-burn-conversion-tool.md
```diff
@@ -212,85 +233,213 @@ That's it! The registry automatically generates:
...
+- **Separation of concerns**: Configuration extraction is separate from type inference
+- **Error handling**: All methods return `Result` for proper error propagation
+- **Opset awareness**: Methods receive opset version for version-specific behavior
+- **Type preferences**: Nodes can request specific types from their inputs (e.g., requesting Shape
+  type for dynamic operations)
 
-1.  Access the constant values during node configuration
-2.  Process operations like Conv1d that expect weights as direct inputs
-3.  Handle shape-defining inputs needed for operations like Reshape
+### ONNX-IR Modules
```
laggui: I would remove this section entirely, it doesn't really need explanations with the way it is structured IMO. It is more likely to be out of date if we forget to update it with reflecting changes.

contributor-book/src/guides/onnx-to-burn-conversion-tool.md
```diff
@@ -212,85 +233,213 @@ That's it! The registry automatically generates:
...
+
+- **`proto_conversion.rs`**: Conversion utilities from ONNX protobuf to IR structures
+
+### Burn-Import Modules
 
 The `burn-import` crate is organized into several key modules:
 
+- **`onnx/model_gen.rs`**: Public API (`ModelGen`) for converting ONNX models to Burn code
+
 ### Core Modules
```
laggui: Same here

crates/burn-import/src/burn/graph.rs
```diff
@@ -515,15 +515,22 @@ impl<PS: PrecisionSettings> BurnGraph<PS> {
...
+        input_types: &[Type],
+        output_types: &[Type],
+    ) {
+        // Handle empty graphs: use provided types directly
+        if self.nodes.is_empty() {
+            // For empty graphs, inputs pass through directly to outputs
+            self.graph_input_types.extend_from_slice(input_types);
+            self.graph_output_types.extend_from_slice(output_types);
+            return;
+        }
```
laggui, line 533: It looks like input/output types are almost always empty.

Plus, when graphs are "empty" it just means that it doesn't contain any nodes, but it still has at least an input and output, which should be represented by a name too. So similar steps should be logically applied and work by retrieving the inputs / outputs? But right now we only check those from nodes.

Simplest alternative: input should not only be the names, but also include the types or be an enum

crates/burn-import/src/burn/node/cast.rs
```diff
@@ -83,17 +81,11 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for CastNode {
...
-                    ElementType::Bool => TensorKind::Bool,
-                    ElementType::String => panic!("Cast: String type not supported"),
+                let target_kind = match self.target_dtype {
+                    DType::F32 | DType::F64 | DType::F16 => TensorKind::Float,
+                    DType::I32 | DType::I64 | DType::I8 | DType::U16 | DType::U8 => TensorKind::Int,
+                    DType::Bool => TensorKind::Bool,
+                    _ => panic!("Unsupported DType for Cast: {:?}", self.target_dtype),
                 };
 
                 if input_tensor.kind == target_kind {
```
laggui, line 91: Unrelated to the changes in this PR, but cast should support specific dtypes otherwise this might be incorrect

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -339,10 +224,8 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for GatherNode {
     }
 
     fn input_types(&self) -> Vec<crate::burn::Type> {
-        match &self.index {
-            GatherIndices::Runtime(index_type) => vec![self.input.clone(), index_type.clone()],
-            GatherIndices::Static(_) => vec![self.input.clone()],
-        }
+        // All indices are now runtime, so always return both input and index
+        vec![self.input.clone(), self.index.clone()]
```
laggui, line 228: Why exactly are they all runtime now?

crates/burn-import/src/burn/node/gemm.rs
```diff
@@ -111,8 +111,17 @@ impl OnnxIntoNode for GemmNode {
...
+        Self::new(
+            a,
+            b,
+            c,
+            output,
+            config.alpha,
+            config.beta,
+            config.trans_a,
+            config.trans_b,
+        )
```
laggui, line 124: The node should just store the config in a field instead. The same thing goes for all other nodes.

crates/burn-import/src/burn/node/linear.rs
```diff
@@ -131,33 +131,22 @@ impl OnnxIntoNode for LinearNode {
...
-                        Data::Int32s(val) => TensorData::new(val, shape).convert::<f32>(),
-                        Data::Int64s(val) => TensorData::new(val, shape).convert::<f32>(),
-                        _ => panic!("Unsupported tensor element type"),
-                    };
-
-                    Some(tensor_data)
+                ArgType::Tensor(_) | ArgType::Shape(_) | ArgType::Scalar(_) => {
+                    // For Tensor, Shape, and Scalar types, extract the underlying tensor data
+                    // onnx-ir now uses burn_tensor::TensorData directly
+                    Some(value.clone().convert::<f32>())
```
laggui: Possible improvement: we should convert to the desired dtype, not strictly f32

crates/burn-import/src/burn/node/one_hot.rs
```diff
@@ -69,8 +67,31 @@ impl OnnxIntoNode for OneHotNode {
...
+            onnx_ir::node::one_hot::OneHotValuesInput::Runtime(_) => {
+                panic!("OneHot with runtime values is not supported in burn-import")
+            }
+        };
+
+        // Derive values_type from output to preserve the correct output element type
+        // Even though onnx-ir stores values as [f32; 2], the output type is determined
+        // by the ONNX model's output element type (e.g., Int64 -> Int)
+        let mut values_type = output.clone();
+        values_type.name = format_ident!("{}_values", output.name);
```
laggui: `values_type` doesn't appear to be used anymore? So we can probably remove the field.

crates/burn-import/src/onnx/model_gen.rs
```diff
@@ -543,8 +543,12 @@ impl ParsedOnnxGraph {
             .map(|output| output.name.clone())
             .collect();
 
+        // Convert ONNX arguments to Burn types for empty graphs
+        let input_types: Vec<_> = self.0.inputs.iter().map(crate::burn::Type::from).collect();
+        let output_types: Vec<_> = self.0.outputs.iter().map(crate::burn::Type::from).collect();
+
         // Register inputs and outputs with the graph
-        graph.register_input_output(input_names, output_names);
+        graph.register_input_output(input_names, output_names, &input_types, &output_types);
```
laggui, line 552: Mirroring my comment in `register_input_output`: we should not handle this separately/differently.

crates/onnx-ir/README.md
```diff
@@ -9,15 +9,44 @@ tensor operations.
...
+  - **Graph State** (`graph_state.rs`): Manages mutable state during conversion including node
+    storage, name mappings, and tensor data
+  - **Tensor Store** (`tensor_store.rs`): Centralized storage for tensor data with ID-based access
+
+- **Processing**:
+  - **Processor** (`processor.rs`): Defines the `NodeProcessor` trait with type inference and
+    constant lifting capabilities
+  - **Registry** (`registry.rs`): Centralized registry mapping node types to their processors -
+    **add new node types here**
+  - **Node Implementations** (`node/`): Contains operation-specific processor implementations
```
laggui: I don't think the detailed code tree is valuable in the README. Users will navigate the tree with the given instructions to add new nodes, and can also refer to the contributing instructions. These types of module docs should stay local to a module IMO (and you currently already have module-level documentation, which is great!).

crates/onnx-ir/src/ir/argument.rs
```diff
@@ -0,0 +1,252 @@
...
+pub enum ValueSource {
+    /// Static constant value embedded in the argument (name="" + data_id=Some)
+    Static,
+    /// Points to a constant node output (name="constant1_out1" + data_id=None)
+    Constant,
+    /// Points to a runtime node output (name="conv1_out1" + data_id=None)
+    Dynamic,
+    /// Optional/not provided (name="" + data_id=None)
+    Optional,
+}
```
laggui, line 30: `Some`/`None` here has overlap in semantics with `data_id`. If these should be tied, then we should remove the `data_id` field and change the static variant to `Static(TensorId)`.

crates/onnx-ir/src/ir/attribute.rs
```diff
@@ -0,0 +1,181 @@
...
+        } else {
+            panic!("Expected Tensors, got {self:?}");
+        }
+    }
+}
+
+/// Convert AttributeValue to an Argument
+impl From<AttributeValue> for Argument {
+    fn from(attr: AttributeValue) -> Argument {
+        let name = "".to_string();
```
laggui: Should we generate a name instead of using an empty placeholder? iirc some constants were being mishandled with empty names before, but maybe this was fixed

crates/onnx-ir/src/ir/graph.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// The inputs of the graph.
+    pub inputs: Vec<Argument>,
+
+    /// The outputs of the graph.
+    pub outputs: Vec<Argument>,
+
+    /// Reference to GraphState to keep tensor data alive for .value() access
+    /// This ensures Arguments can access tensor data via their data_id
+    pub(crate) _graph_data: Option<std::rc::Rc<std::cell::RefCell<crate::graph_state::GraphState>>>,
```
laggui, line 23: The reliance on `_graph_data` here feels like a lifetime hack. Couldn't this have been accomplished with a builder pattern? e.g.
1. Keep `GraphState` only during graph construction in a builder.
2. Consume `GraphState` into an immutable `OnnxGraph` that owns the tensor store.
3. Update `Argument` to hold a reference directly to the tensor store instead of the full `GraphState`.

That way, lifetime is clear, and we don't need to hold onto this graph data just for `.value()` argument access.

crates/onnx-ir/src/ir/tensor_data_ext.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    /// Convert to `Vec<i64>`, handling Int32 and Int64 types
+    ///
+    /// Useful for extracting indices, shapes, or other integer arrays that need to be i64.
+    fn to_i64_vec(&self) -> Result<Vec<i64>, burn_tensor::DataError>;
+
+    /// Convert to `Vec<usize>`, handling Int32 and Int64 types
+    ///
+    /// Useful for extracting shape or dimension values.
+    fn to_usize_vec(&self) -> Result<Vec<usize>, burn_tensor::DataError>;
```
laggui: This doesn't seem to be used at all?

crates/onnx-ir/src/ir/tensor_data_ext.rs
```diff
@@ -0,0 +1,161 @@
...
+            DType::I32 => {
+                let vec_i32 = self.to_vec::<i32>()?;
+                Ok(vec_i32.into_iter().map(|v| v as i64).collect())
+            }
+            other => Err(burn_tensor::DataError::TypeMismatch(format!(
+                "Cannot convert {:?} to Vec<i64>",
+                other
+            ))),
+        }
+    }
```
laggui, line 94: This can be simplified to
```rust
match self.dtype {
    DType::I64 | DType::I32 => self.convert::<i64>().to_vec()?
    other => Err(burn_tensor::DataError::TypeMismatch(format!(
        "Cannot convert {:?} to Vec<i64>",
        other
    ))),
}
```
`convert()` is a no-op if the dtype matches the target.

Same for other vec utilities below

crates/onnx-ir/src/node/argmax.rs
```diff
@@ -1,4 +1,37 @@
...
+//!
+//! - `reduced` (tensor(int64)): Output tensor of indices with datatype int64. Shape depends on `keepdims`:
+//!   - If keepdims=1: same rank as input, reduced dimension becomes size 1
+//!   - If keepdims=0: rank reduced by 1
+//!
+//! ## Opset Versions
+//!
+//! - **Opset 13**: Added bfloat16 to type constraints
+//! - **Opset 12**: Added `select_last_index` attribute
+//! - **Opset 11**: Changed `axis` range to support negative indices [-r, r-1]
```
laggui, line 11: Hmmm I don't think we should duplicate the ONNX spec info here. This can be very error prone, and we can just reference the actual spec via the link.

Listing the supported opset versions for the implementation can be valuable though, so we should keep a short form of that.

crates/onnx-ir/src/node/argmax.rs
```diff
@@ -9,110 +42,138 @@ pub struct ArgMaxConfig {
...
+impl NodeProcessor for ArgMaxProcessor {
+    fn infer_types(
+        &self,
+        node: &mut Node,
+        opset: usize,
+        _output_preferences: &OutputPreferences,
+    ) -> Result<(), ProcessError> {
+        crate::processor::validate_opset(opset, 11)?;
+        crate::processor::validate_input_count(node, 1)?;
+        crate::processor::validate_output_count(node, 1)?;
```
laggui: We should make this more straightforward, instead of forcing users to always call functions for validation.

Suggestion: a node spec that defines the supported opset(s), input and output counts (which might actually vary based on opset version!!) and let the node processor handle the checks based on the definitions.

crates/onnx-ir/src/node/argmax.rs
```diff
@@ -9,110 +42,138 @@ pub struct ArgMaxConfig {
...
-        node.name,
-        config.keepdims,
-        node.outputs[0].ty
-    );
+        if axis < 0 {
+            axis += tensor.rank as i64;
+        }
+
+        let config = ArgMaxConfig::new(axis as usize, keepdims);
+        Ok(Some(Box::new(config)))
```
laggui, line 156: This is a bit of a weird pattern, where `extract_config` returns the config just for `convert_nodes` to assign it.

It's sort of a two-step dance: processor returns config, caller assigns it back to node. And this has to be done in order to use the config in other steps. So it's also a bit brittle.

crates/onnx-ir/src/phases/node_conversion.rs
```diff
@@ -0,0 +1,498 @@
...
+
+        let config = processor
+            .extract_config(&node, opset_version)
+            .unwrap_or_else(|e| {
+                panic!(
+                    "Failed to extract config for node {} (type: {:?}): {:?}",
+                    node.name, node.node_type, e
+                )
+            });
+        node.config = config;
```
laggui, line 90: This is brittle. And not actually required by the node conversion, feels more like an artifact of plumbing for the current pipeline.

crates/onnx-ir/src/phases/post_processing.rs
```diff
@@ -0,0 +1,341 @@
...
+            .restore_data(tensor_data_clone, next_tensor_id);
+        result
+    };
+
+    // Identity elimination
+    log::debug!("Starting Identity elimination");
+    {
+        let elimination_plan = plan_identity_elimination(&nodes);
+        apply_identity_elimination(&mut nodes, &mut outputs, elimination_plan);
+    }
```
laggui, line 175: It's not immediately clear to me why we need to remove identity nodes after the previous steps.

Are these nodes somehow added by previous steps? And if not, why can't we clean this up before?

crates/onnx-ir/src/phases/type_inference.rs
```diff
@@ -0,0 +1,200 @@
...
+/// This allows runtime preference collection (e.g., Concat requests Shape after seeing Shape inputs).
+pub(super) fn iterative_type_inference_with_preferences(nodes: &mut [Node], opset: usize) {
+    let registry = get_processor_registry();
+
+    // Track collected preferences: (producer_output_name, consumer_name, pref_type_str)
+    let mut collected_preferences: HashSet<(String, String, String)> = HashSet::new();
+
+    let max_iterations = 100; // Safety limit to prevent infinite loops
+
+    for iteration in 1..=max_iterations {
```
laggui: This seems overly complex. Is it common to have >2 iterations?

In practice, most ONNX graphs should be feed-forward with clear type propagation, no?

crates/onnx-ir/src/phases/type_inference.rs
```diff
@@ -0,0 +1,200 @@
...
+        for i in 0..nodes.len() {
+            // Get preferences for this node
+            let prefs = node_preferences
+                .get(&nodes[i].name)
+                .cloned()
+                .unwrap_or_else(crate::processor::OutputPreferences::new);
+
+            // Run type inference on this node
+            let processor = registry.get(&nodes[i].node_type);
+            let _ = processor.infer_types(&mut nodes[i], opset, &prefs);
```
laggui: Why are we ignoring the possible errors in `infer_types` result?

crates/onnx-ir/src/pipeline.rs
```diff
@@ -0,0 +1,115 @@
...
+    log::debug!(" PHASE 3: Type Inference ");
+    let opset_version = extract_opset_version(model);
+    type_inference::infer_types(&state_rc, opset_version);
+
+    log::debug!(" PHASE 4: Post-processing ");
+    let (mut nodes, inputs, mut outputs) = post_processing::post_process(&state_rc);
+
+    log::debug!(" PHASE 5: Finalization ");
+    finalization::finalize(&mut nodes, inputs, &mut outputs, state_rc)
+}
```
laggui, line 189: While the separation is great, each phase here traverses the graph nodes multiple times. The worst one probably being in type inference with an iterative process that can reach up to 100 traversals?

For small graphs, this might be negligible, but could be potentially bad for bigger / more complex models.

crates/onnx-ir/src/proto_conversion.rs
```diff
@@ -1,29 +1,261 @@
...
-use bytemuck::{cast_slice, try_cast_vec};
+use burn_tensor::DType;
+use protobuf::Enum;
 
-fn cast_vec_with_fallback<E: bytemuck::Pod>(raw_data: Vec<u8>) -> Vec<E> {
-    // Zero-copy `try_cast_vec` with fallback when alignment and size are not compatible
-    try_cast_vec(raw_data).unwrap_or_else(|(_e, raw_data)| cast_slice(&raw_data).to_vec())
-}
+/// Minimum required ONNX opset version
+pub const MIN_OPSET_VERSION: i64 = 16;
```
laggui: Why not usize? opset should be unsigned

crates/onnx-ir/src/proto_conversion.rs
```diff
@@ -33,85 +265,103 @@ impl TryFrom<TensorProto> for TensorData {
...
                 }
-                ElementType::Float64 => {
-                    if !tensor.double_data.is_empty() {
-                        Data::Float64s(tensor.double_data)
-                    } else {
-                        Data::Float64s(vec![])
-                    }
+                DType::F32 if expected_elems == 0 => {
+                    // Empty tensor with zero elements
+                    Ok(TensorData::new(Vec::<f32>::new(), shape))
```
laggui: Given that `float_data` is a `Vec<f32>`, both `DType::F32` match arms are equivalent no? Same as empty vec:
```rust
TensorData::new(tensor.float_data, shape)
```
so why do we need to check `if expected_elems == 0`?

crates/onnx-ir/src/ir/node.rs
```diff
@@ -0,0 +1,124 @@
...
+        self.get_config::<T>().unwrap_or_else(|| {
+            panic!(
+                "Node '{}' ({:?}) config is not set or has wrong type. Expected {}",
+                self.name,
+                self.node_type,
+                std::any::type_name::<T>()
+            )
+        })
+    }
+}
```
laggui, line 96: 1. This seems to be entirely an artifact of plumbing, not a genuine part of the node. The node processor returns a config via `extract_config`, caller assigns it back to node, just so it can be re-used in later steps like `infer_types` or burn-import conversion.
Ideally, config could be passed to the `infer_types`, but with your current pipeline that iterates through the nodes at each step that might not work. Instead of extracting the config during conversion, which is only useful for the subsequent type inference, type inference should extract the config and pass it along.
For `OnnxIntoNode` (the only other place where the config is retrieved again), the processor can be used to extract it again.

2. The runtime type erasure with added up/down casting requirements can be simplified. At first glance, this seems like a good place to have an associative config type on the `NodeProcessor`. This would remove the need for `NodeConfig` trait entirely.

So with 1 + 2, the node no longer stores a config (remove the field and the two methods), and the type is associated to the node processor implementation for a node.

antimora: We have discussed offline and concluded we should properly redo Graph using Node enums. This will get done in a next PR. The issue is tracked here: https://github.com/tracel-ai/burn/issues/3988

crates/onnx-ir/src/phases/node_conversion.rs
antimora: Agreed. We will redo as a graph of node enums: https://github.com/tracel-ai/burn/issues/3988

crates/onnx-ir/src/node/argmax.rs
antimora: Agreed. We will redo as a graph of node enums: https://github.com/tracel-ai/burn/issues/3988

crates/onnx-ir/src/phases/type_inference.rs
antimora: Yeah you're right. It should not take more 2-3 iterations. I can reduce from 100 to 10. We a warning to catch and re-visit this approach. 

FYI, this iterative approached was the safest and bug free compared to graph re-trigger approach, which would have been more efficient but it was complex.

crates/onnx-ir/src/phases/type_inference.rs
antimora: Good catch! I'll fix it so that the error propagates correctly. I even caught a muted error in my unit tests!

I am also refactoring so that parse_onnx returns results.

crates/onnx-ir/src/pipeline.rs
antimora: It used to be much worse and I tried to minimize traversals. There are hard cross node dependencies, such as conversation to Linear. I figured for now the code complex is more important than onnx build conversation optimization. BTW, infer_types should be at most 2-3 times. I also reduced max iteration catch to 10. 

Down the line, this could be optimized even further however the code will be more complex. This simple approach is more preferable because it is more robust against ONNX edge cases. Once we have more tests, we could undertake optmization.

crates/onnx-ir/src/phases/post_processing.rs
antimora: Rewiring is easier in the end. Also there are Identity nodes with initializers that act like Constants. BTW, there is a big of refactoring of Identify rewiring in Subgraph because it's recursive.

crates/onnx-ir/src/ir/attribute.rs
antimora: Removed the conversion entirely. It wasn't needed for attribute to be converted to argument.

crates/onnx-ir/src/ir/graph.rs
antimora: Understood. I will refactor this along with [Implement memory-efficient ONNX weight loading (lazy/protobuf streaming) ](https://github.com/tracel-ai/burn/issues/3596). This will not remain as graph data. The new new refactored will reference mmapped memory regions.

crates/burn-import/src/burn/node/gemm.rs
antimora: This was a hold out from the previous design principle: code generator uses its own abstraction layer. I recommend we loosen this restriction and code against the new clean onnx ir graph directly and don't do extra arg transformations, etc. This work should be done after: https://github.com/tracel-ai/burn/issues/3988

antimora: I filed this ticket: https://github.com/tracel-ai/burn/issues/3989

crates/burn-import/src/onnx/model_gen.rs
antimora: This will be refactored as part of https://github.com/tracel-ai/burn/issues/3989

It's a big change if we need to refactored again.

crates/burn-import/src/burn/node/gather.rs
antimora: Keeping the index as a constant arg ( as opposed to static arg) or input. Previously we were lifting them and converting to static but would create a tensor. It's better for performance since select op uses tensor.

crates/burn-import/src/burn/graph.rs
antimora: I will refactor this as part of https://github.com/tracel-ai/burn/issues/3989. This ended up bigger change for this PR to be left to be removed later.

laggui: Sure!

crates/onnx-ir/src/phases/type_inference.rs
laggui: Ok sounds like a more reasonable limit for now :+1:

PR #3874, opened 2025-10-16
feat: Matmul quant

wingertge:
### Checklist

### Related Issues/PRs

CubeCL changes: https://github.com/tracel-ai/cubecl/pull/960

### Changes

Refactors quantized matmul to use automatic dequantization on global load.

### Testing

Adds a new quantized matmul test that works with the packed value requirements (aligned to 4 on major dim).

### DIFF
--- a/crates/burn-cubecl-fusion/src/base.rs
+++ b/crates/burn-cubecl-fusion/src/base.rs
@@ -193,7 +193,7 @@ impl<R: Runtime> CubeFusionHandle<R> {
                 QuantParam::BF16 => DType::BF16,
                 QuantParam::UE8M0 | QuantParam::UE4M3 => unimplemented!("Not yet supported"),
             },
-            strides: vec![1],
+            strides: qparams.scales.strides.clone(),
             qparams: None,
         })
     }
--- a/crates/burn-cubecl/src/kernel/conv/implicit_gemm/launch.rs
+++ b/crates/burn-cubecl/src/kernel/conv/implicit_gemm/launch.rs
@@ -19,7 +19,8 @@ use cubecl::{
 };
 
 use crate::{
-    CubeElement, CubeRuntime, FloatElement, ops::numeric::empty_device_optimized, tensor::CubeTensor,
+    CubeElement, CubeRuntime, FloatElement, ops::numeric::empty_device_optimized,
+    tensor::CubeTensor,
 };
 
 /// Perform a 2D convolution using the implicit GEMM (im2col) algorithm, using cubecl tiling matmul


crates/burn-cubecl-fusion/src/matmul/builder.rs
```diff
@@ -67,8 +68,8 @@ impl<R: Runtime> OptimizationBuilder<CubeOptimization<R>> for MatmulBuilder<R> {
                 let out = self.builder.output_unhandled(&op.out);
 
                 self.matmul = Some(FusedMatmul::new(
-                    lhs,
-                    rhs,
+                    MatmulArg::Normal(lhs),
+                    MatmulArg::Normal(rhs),
```
nathanielsimard: I would be cool to support quantized

crates/burn-cubecl/src/ops/qtensor.rs
```diff
@@ -275,52 +268,18 @@ where
...
-
-        // If the above quantized matmul fail, we fallback to the dequantize-then-matmul pattern.
-        let scheme = *lhs.scheme();
-        let propagation = lhs.propagation();
-        let t1_f = <Self>::dequantize(lhs);
-        let t2_f = <Self>::dequantize(rhs);
-        let out = Self::float_matmul(t1_f, t2_f);
-
-        match propagation {
+        let out = kernel::matmul::matmul::<R, half::f16>(
```
nathanielsimard: Unsure we should hardcode f16 here

wingertge: It's what was done before so I wanted to replicate the existing behavior for now. I think the rationale was that quantized is already lower precision (and we don't have block scaled yet) so f16 is a good default. But of course it does limit the range, so not sure what the correct choice is/how we should determine this type.

nathanielsimard: Yeah I don't agree with the premise, and f16 isn't supported on the web where quantization makes a lot of sense. I would put f32 by default for now! It's the accumulation precision right? just to make sure

crates/burn-cubecl-fusion/src/matmul/builder.rs
wingertge: That would require changes to the traces right? Not sure I understand the fusion workflow enough to implement that part myself tbh 😅
But with a few pointers I could give it a try.

crates/burn-cubecl/src/ops/qtensor.rs
wingertge: Accumulation and inputs. Just set it to the float type then?

nathanielsimard: There is a big problem in the definition, we can't assume both LHS and RHS have the same precision, especially for quantization where one can be quantized and the other normal.

wingertge: burn doesn't support that right now so it's currently mostly fine. But yes, we really need separate types for lhs and rhs at some point. And support mixed inputs. The way the operation is defined right now just don't allow for that.

wingertge: I've set things up for mixed precision both in the matmul itself and the fusion implementation I'm working on rn, but the burn interface still needs to be adapted. But the groundwork is there.

crates/burn-cubecl-fusion/src/matmul/builder.rs
wingertge: I've added my best guess at an implementation, but can't test at the moment

PR #3877, opened 2025-10-17
Refactor IR to define outputs as a function of the operation

laggui:
### Checklist

### Related Issues/PRs

Follow-up to #3845

### Changes

IR changes:
- Outputs are a function of the operation; added op IR builders `create(...)` that automatically creates the output(s) with correct shape and dtype
- All ops can automatically derive `Hash` with the changes to `ScalarIr`
- Split IR `nodes()` into `inputs()` / `outputs()`
- Removed `ToDevice` op IR (unused)
- Added `CreationOpIr`, `FullOpIr`,  `CastOpIr`, `ShapeOpIr` (replaces expand, also used for reshape), `ReduceOpIr`, `MatmulOpIr` for semantics

Fusion changes:
- New `OperationStreams::with_inputs` constructor that creates a default stream and registers the inputs
- Registering an operation with now automatically handles input streams and output fusion tensor
  - No longer need `tensor_uninitialized`, just provide the interface to create new empty handle IDs

Others:
- More `Shape` methods
- Zeros/ones are now properly defined as base ops
- Cumulative ops are numeric ops (were added as base ops IR)

### Testing

Unit tests

### DIFF
--- a/crates/burn-fusion/src/client.rs
+++ b/crates/burn-fusion/src/client.rs
@@ -297,33 +297,3 @@ where
         server.resolve_server_bool::<B>(&tensor.into_ir())
     }
 }
-
-/// Extension trait to extract outputs when registering an operation.
-pub trait OperationOutput<R: FusionRuntime> {
-    /// Extract a single output tensor.
-    ///
-    /// # Panics
-    ///
-    /// Panics if the operation produced more or fewer than one output.
-    fn output(self) -> FusionTensor<R>;
-
-    /// Extract a pair of output tensors.
-    ///
-    /// # Panics
-    ///
-    /// Panics if the operation produced more or fewer than two outputs.
-    fn outputs(self) -> (FusionTensor<R>, FusionTensor<R>);
-}
-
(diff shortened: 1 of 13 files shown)


crates/burn-fusion/src/client.rs
```diff
@@ -281,3 +297,33 @@ where
...
+    /// Panics if the operation produced more or fewer than one output.
+    fn output(self) -> FusionTensor<R>;
+
+    /// Extract a pair of output tensors.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the operation produced more or fewer than two outputs.
+    fn outputs(self) -> (FusionTensor<R>, FusionTensor<R>);
+}
```
nathanielsimard: I would change that pattern to the following:

```rust
use std::path::PathBuf;

trait Output {
    fn outputs<const N: usize>(self) -> [String; N];
}

impl Output for Vec<PathBuf> {
    fn outputs<const N: usize>(self) -> [String; N] {
        let arra: [PathBuf; N] = self.try_into().unwrap();
        return arra.map(|path| format!("{:?}", path));
    }
}
```

That is a simple example, but it generalize to any length and is determined at compile time.

PR #3879, opened 2025-10-17
Add `Shape::ravel` for row-major raveling of indices.

crutcher:
- Implement `ravel` method for computing row-major index offsets.
- Add canonical index handling via `canonicalize_index`.
- Include unit tests for validation.

This should ease working with `TensorData::to_vec()`

### DIFF
--- a/crates/burn-tensor/src/tensor/indexing/mod.rs
+++ b/crates/burn-tensor/src/tensor/indexing/mod.rs
@@ -213,28 +213,32 @@ where
 
 /// Compute the ravel index for the given coordinates.
 ///
-/// This returns the row-major order raveling.
+/// This returns the row-major order raveling:
+/// * `strides[-1] = 1`
+/// * `strides[i] = strides[i+1] * dims[i+1]`
+/// * `dim_strides = coords * strides`
+/// * `ravel = sum(dim_strides)`
 ///
 /// # Arguments
-/// - `coords`: must be the same size len as `dims`.
-/// - `dims`: must be the same len as `coords`.
+/// - `indices`: the index for each dimension; must be the same length as `shape`.
+/// - `shape`: the shape of each dimension; be the same length as `indices`.
 ///
 /// # Returns
 /// - the ravel offset index.
-pub fn ravel_dims<I: AsIndex>(coords: &[I], dims: &[usize]) -> usize {
+pub fn ravel_index<I: AsIndex>(indices: &[I], shape: &[usize]) -> usize {
     assert_eq!(
-        dims.len(),
-        coords.len(),
+        shape.len(),
+        indices.len(),
         "Coordinate rank mismatch: expected {}, got {}",
-        dims.len(),
-        coords.len(),
+        shape.len(),
+        indices.len(),
     );
 
     let mut ravel_idx = 0;
     let mut stride = 1;
 
-    for (i, &dim) in dims.iter().enumerate().rev() {
-        let coord = canonicalize_index(coords[i], dim, false);
+    for (i, &dim) in shape.iter().enumerate().rev() {
+        let coord = canonicalize_index(indices[i], dim, false);
         ravel_idx += coord * stride;
         stride *= dim;
     }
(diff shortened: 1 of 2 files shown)


crutcher:
Follow on ideas, a ranked `Index` view object:
```rust
#[derive(Debug)]
pub struct TensorDataIndexView<'a, E: Element, const R: usize> {
    data: &'a TensorData,
    shape: Shape,
    _phantom: std::marker::PhantomData<&'a E>,
}

impl<'a, I: AsIndex, E: Element, const R: usize> Index<[I; R]> for TensorDataIndexView<'a, E, R> {
    type Output = E;
    fn index(&self, index: [I; R]) -> &Self::Output {
        &self.data.as_slice::<E>().unwrap()[self.shape.ravel(index)]
    }
}
```
with an op on `TensorData`:
```rust
    /// Get an indexed view of the data.
    pub fn indexed_view<'a, E: Element, const R: usize>(&'a self) -> TensorDataIndexView<'a, E, R> {
        TensorDataIndexView {
            data: self,
            shape: self.shape.clone().into(),
            _phantom: std::marker::PhantomData,
        }
    }
```

Which lets us do this:
```rust
        let state_data = state.to_data().convert::<bool>();
        let state_view = TensorDataIndexView::view(&state_data);

        self.gl.draw(args.viewport(), |c, gl| {
            for h_idx in 0..h {
                for w_idx in 0..w {
                    let is_live = state_view[[h_idx, w_idx]];
                    ...
                }
            }
        }
```

crutcher:
I extracted `ravel_dims`, so that `TensorData` views can use it directly without allocating a `Shape`.

crutcher:
with the change to expose `ravel_dims`, there's no need for the view object to store additional information; so we could skip exposing the view object and go straight to:
```rust
pub fn get<E: Element, const R: usize, I: AsIndex>(&self, coords: [I; R]) -> &E;
pub fn get_mut<E: Element, const R: usize, I: AsIndex>(&self, coords: [I; R]) -> &E;
```

It's possible that type hackery would let us do this directly with `Index` impls on `TensorData`.

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -211,9 +211,52 @@ where
     }
 }
 
+/// Compute the ravel index for the given coordinates.
+///
+/// This returns the row-major order raveling.
+///
+/// # Arguments
+/// - `coords`: must be the same size len as `dims`.
+/// - `dims`: must be the same len as `coords`.
```
laggui: These are not really descriptive w.r.t to the arguments, but rather their constraints. Coords are actually the indices for each dim.

crates/burn-tensor/src/tensor/indexing/mod.rs
```diff
@@ -211,9 +211,52 @@ where
...
+///
+/// # Returns
+/// - the ravel offset index.
+pub fn ravel_dims<I: AsIndex>(coords: &[I], dims: &[usize]) -> usize {
```
laggui: I think `ravel_dims` is a bit ambiguous. Following the above description ("compute the ravel index"), I would suggest `ravel_index` over `ravel_dims`.

crates/burn-tensor/src/tensor/shape.rs
```diff
@@ -75,6 +76,19 @@ impl Shape {
         self
     }
 
+    /// Compute the ravel index for the given coordinates.
+    ///
+    /// This returns the row-major order raveling.
+    ///
+    /// # Arguments
+    /// - `coords`: must be the same size as `self.rank()`.
```
laggui: Same comment regarding description

crates/burn-tensor/src/tensor/shape.rs
```diff
@@ -75,6 +76,19 @@ impl Shape {
...
+    ///
+    /// # Returns
+    /// - the ravel offset index.
+    pub fn ravel<I: AsIndex>(&self, coords: &[I]) -> usize {
```
laggui: `ravel_index`?

crutcher: done.

crates/burn-tensor/src/tensor/shape.rs
crutcher: done.

crates/burn-tensor/src/tensor/indexing/mod.rs
crutcher: done.

crates/burn-tensor/src/tensor/indexing/mod.rs
crutcher: done.

PR #3880, opened 2025-10-17
Extend `TensorData` with `index_view()` and `mut_index_view()`

crutcher:
This is built on https://github.com/tracel-ai/burn/pull/3879

Consider also `at()[[a, b, c]]` and `at_mut()[[a, b, c]]`

### DIFF
--- a/crates/burn-tensor/src/tensor/data.rs
+++ b/crates/burn-tensor/src/tensor/data.rs
@@ -939,9 +939,7 @@ impl<'a, E: Element, I: AsIndex> Index<&[I]> for TensorDataIndexViewMut<'a, E> {
     }
 }
 
-impl<'a, E: Element, I: AsIndex> IndexMut<&[I]>
-    for TensorDataIndexViewMut<'a, E>
-{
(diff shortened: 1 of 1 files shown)


crutcher:
Maybe this should be `&[I]` instead, and have all the rank checking dynamic. This makes it annoying to use `Vec<usize>` for indices. Unsure.

crutcher:
the view should maybe cache the slice, save some lookup time there for repeated calls?

laggui (commented):
As we discussed on discord the other week, I am not really in favor of such extensions to `TensorData`.

Adding indexing makes it try to be both a container and a computational view, which is not its responsibility.

Such manipulations should be performed on `Tensor`. I understand that currently requires moving between backends since some like CUDA don't have a CPU device, which is a bit awkward and not so ergonomic. But we are going to provide a better multi-backend support that will provide proper CPU-backed `Tensor`s that can be used for this sort of manipulation.

For now, I much prefer to add lightweight utilities to `Shape` (like the linked PR).

PR #3882, opened 2025-10-19
Feat/matvec

huy209vn:
Pull Request Template
Checklist

 Verified successful builds and tests for burn-tensor, burn-core, and burn-cubecl.

 Updated Burn book (burn-book/src/building-blocks/tensor.md) to include linalg::matvec.

 cargo run-checks partially executed — CUDA backend failed due to local env, unrelated to tensor changes.

Related Issues/PRs

N/A

Changes

Added linalg::matvec(matrix, vector) — a dedicated matrix–vector helper reusing optimized matmul kernels.
Includes strict shape validation and batch broadcasting support.

Expanded triangular masks (triu_mask, tril_mask, diag_mask) to handle batched shapes consistently.

Fixed Shape::dims padding logic to correctly pad with 1s.

Added corresponding tests and book update.

Testing

Ran:
 cargo check `
  -p burn-core `
  -p burn-tensor `
  -p burn-cubecl `
  
  ⚠️ cargo run-checks failed locally due to cudarc compile errors (CUDA 12.6+ header mismatch).
Tensor, Core, and CubeCL tests all passed. CI CUDA env should pass.


laggui:
Sorry for the delay :sweat_smile: we're in sort of a code freeze to release 0.19 (following benchmarks and tests completion for our release process)

Will review once the release is complete :pray:

crates/burn-tensor/src/tensor/api/bool.rs
```diff
@@ -144,8 +144,13 @@ where
...
 
-        // Generate and return the mask by applying the comparison to the matrix.
-        compare(matrix, 0).unsqueeze()
+        // Generate the base mask by applying the comparison to the broadcast matrix.
+        let mask = compare(matrix, 0);
+
+        // Expand the mask to match the requested shape in every dimension so that
+        // batched triangular masks report the exact same shape as provided.
+        let target_shape = shape.dims::<D>();
+        mask.expand(target_shape).unsqueeze()
```
laggui, line 153: We can skip `shape.dims()` and `unsqueeze()`, the target shape is already define.

Just doing
```rust
mask.expand(shape)
```

should be sufficient

crates/burn-tensor/src/tensor/shape.rs
```diff
@@ -63,9 +63,13 @@ impl Shape {
...
     /// Returns the dimensions of the tensor as an array.
+    ///
+    /// Missing trailing dimensions are padded with `1` to preserve the historical
+    /// behaviour of fixed-size arrays representing shapes.
     pub fn dims<const D: usize>(&self) -> [usize; D] {
         let mut dims = [1; D];
-        dims[..D].copy_from_slice(&self.dims[..D]);
+        let len = core::cmp::min(D, self.dims.len());
+        dims[..len].copy_from_slice(&self.dims[..len]);
         dims
```
laggui, line 73: Hmmm I'm not sure this is such explicit behavior.

If someone writes `shape.dims::<4>()`, they're explicitly stating "I expect this shape to have 4 dimensions". So I think it should rightfully panic if the shape does not hold 4 dimensions (as it does now, but perhaps we should actually check for it and return a more meaningful error). 

We could actually argue that truncation should behave similarly, and disallow it as well.

crates/burn-tensor/src/tests/shape.rs
```diff
@@ -0,0 +1,34 @@
...
+        assert_eq!(shape.dims::<4>(), [4, 8, 1, 1]);
+    }
+
+    #[test]
+    fn dims_handles_empty_shapes() {
+        let shape = Shape::new([]);
+
+        assert_eq!(shape.dims::<3>(), [1, 1, 1]);
+    }
+}
```
laggui, line 34: Shape tests should be defined in `burn-tensor/src/tensor/shape.rs` (with other shape struct tests).

With the argument I made above we should probably check that these tests panic instead. Lmk what you think.

huy209vn:
Hey @laggui,
I noticed this PR includes some unrelated edits that slipped in while I was iterating on matvec.
I’ve opened a new, clean version focused purely on the matvec implementation and tests. matvec #3967
I’ll close this one to keep things tidy.
Thanks for your earlier feedback!

PR #3901, opened 2025-10-22
Make ElementType Copy

TheGhostHuCodes:
### Checklist

### Related Issues/PRs

N/A
### Changes

tldr; I derived `Copy` on the `ElementType` enum.

While using the `onnx-ir` crate as a dependency I noticed that the `ElementType` enum wasn't `Copy`, causing me to clone it in various places. Not sure why it's not `Copy` the type is certainly small enough to be, so I went ahead and made it `Copy`. This unfortunately unleashed a bit of an attack of the `clones()` which I also fixed, and that's the bulk of the changes.

### Testing

Ran `cargo run-checks`, the only errors I got were FP comparison errors, same as in my last few PRs. They're not related this change.

### DIFF
--- a/crates/burn-import/src/burn/node/constant.rs
+++ b/crates/burn-import/src/burn/node/constant.rs
@@ -285,9 +285,9 @@ impl OnnxIntoNode for ConstantNode {
                         .value
                         .as_ref()
                         .expect("Scalar constant should have value");
-                    scalar_from_data(tensor.elem_type.clone(), v.data.clone())
+                    scalar_from_data(tensor.elem_type, v.data.clone())
                 } else {
-                    let kind: TensorKind = tensor.elem_type.clone().into();
+                    let kind: TensorKind = tensor.elem_type.into();
                     let rank = tensor.rank;
                     let name = node.name.clone();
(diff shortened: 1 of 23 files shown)


antimora:
Thank you for your contribution and interest. 

`onnx-ir` is undergoing a big refactor see https://github.com/tracel-ai/burn/pull/3872 (posted about it in a [message group here](https://discord.com/channels/1038839012602941528/1166763685822333009/1428195880715943997)) 

The latest change gets rid of ElementType and replaces with DType, which did not exist at the time we created Burn's ONNX. Now onnx-ir is using Data from burn-tensor and Dtype. This will eliminate unnecessary indirection and open up usage of mmap to minimize memory usage.

TheGhostHuCodes:
Oh nice! By `DType` you're talking about these? https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/element/base.rs#L271-L289

Looks like that's already `Clone`, `Copy`, and `pub` so I'm all set. I'm looking forward to this refactor!

I should probably hang out in the message group chat a bit more, I'll ask around there first next time.

PR #3672, opened 2025-09-06
Fix/implemented Matmulinteger in burn-import +Uint8 int8 elementtype

huy209vn:
## Checklist

* [x] Confirmed that `cargo run-checks` command has been executed.
* [x] Made sure the book is up to date with changes in this PR.

## Related Issues/PRs

* Implements part of *Help Wanted: Implementing ONNX Ops* #1714.

## Changes

* Implemented `MatMulInteger` in **burn-import**.
* Added `Uint8` and `Int8` element types to **onnx-ir** (ElementType, Data, conversions, debug, etc.).
* Reworked `Argument::from_initializer` to correctly handle scalar (`[]`, `[1]`) and empty (`[0]`) initializers instead of panicking.
* Centralized dtype mapping in `element_type_from_proto`, applied in `proto_conversion.rs`.
* Updated Cast logic in both **onnx-ir** and **burn-import** to support `Uint8`/`Int8`.
* Added new `MatMulIntegerNode`:

  * **onnx-ir**: output rank/type inference.
  * **burn-import**: codegen for `(A - a_zp) @ (B - b_zp)` with optional zero-points, always `int32` output.
* Registered in node enums and dispatch.
* Added ONNX models + generators for `matmulinteger` and `matmulinteger_ranks`.
* Integrated new tests into `build.rs` and `test_mod.rs`.
* Added Rust tests (compile/run + codegen golden) for `matmulinteger`.

## Testing

* Ran **onnx-tests** and **burn-import** test suites, all passing.
* New tests validate ONNX → IR → Burn record works end-to-end and codegen output matches expected.

PR #3904, opened 2025-10-22
Fix ONNX import panic on empty tensor constants

antimora:
### Context

Fixes a panic that occurs when importing ONNX models containing tensor constants with empty data fields (both `raw_data` and typed data fields like `float_data` are empty). This bug was introduced in #3672 and prevents models like `retinaface-resnet50_opset16.onnx` from being imported.

The refactoring in commit `57b53db` changed how tensor proto conversion handles empty data:

**Before:** The code would accept empty typed data fields and create tensors with the data from those fields (even if empty).

**After:** The code only handled non-empty typed data fields with `if !tensor.xxx_data.is_empty()` guards, causing empty tensors to fall through to a catch-all panic:

```
panicked at crates/onnx-ir/src/proto_conversion.rs:156:73:
called `Result::unwrap()` on an `Err` value: VariantNotFound("empty/unsupported payload for Float32")
```

### Checklist

### Related Issues/PRs

Originally reported on discord: https://discord.com/channels/1038839012602941528/1166763685822333009/1430313309416067083

### Changes

Modified the tensor proto conversion in `crates/onnx-ir/src/proto_conversion.rs` to:

1. Remove the `if !tensor.xxx_data.is_empty()` guards that prevented handling of empty data
2. Return empty vectors (`vec![]`) when data fields are empty instead of panicking
3. Add explicit handling for all `ElementType` variants (`Float16`, `Uint16`, etc.)
4. Maintain error handling for unsupported types (e.g., `String`)

### Testing

Added comprehensive test case for empty tensor constants


antimora:
CC: @AdrianEddy

laggui (approved):

PR #3906, opened 2025-10-22
Some updates to the book

louisfd:


burn-book/src/basic-workflow/conclusion.md
laggui, line 1: Might as well delete the file

burn-book/src/performance/quantization.md
```diff
@@ -80,11 +78,11 @@ values, storage format, granularity, and how the values are scaled.
 
 ```rust
-    .with_mode(QuantMode::Symmetric)    // Quantization mode
-    .with_level(QuantLevel::Block(32))  // Granularity (per-tensor or per-block)
-    .with_value(QuantValue::Q8S)        // Data type of quantized values, independent of how they're stored
-    .with_store(QuantStore::Native)     // Storage format for quantized values
-    .with_param(QuantParam::F16);       // Precision for quantization parameters
+    .with_mode(QuantMode::Symmetric)         // Quantization mode
+    .with_level(QuantLevel::Block([2, 16]))  // Granularity (per-tensor or per-block)
```
laggui: I think it should be `QuantLevel::block([2, 16])` (note the lowercase `b`, `block(...)` is a function to create `BlockSize` struct from the array).

burn-book/src/basic-workflow/conclusion.md
louisfd: I did, but must have clicked on the link in the summary which re-generated it 😅

PR #3918, opened 2025-10-24
Fix typos

laggui:


xtask/src/commands/test.rs
```diff
@@ -157,6 +159,8 @@ pub(crate) fn handle_command(
 
                     args_wgpu.features = Some(vec!["test-wgpu".into()]);
                     handle_wgpu_test("burn-core", &args_wgpu)?;
+                    handle_wgpu_test("burn-optim", &args_wgpu)?;
+                    handle_wgpu_test("burn-nn", &args_wgpu)?;
                     handle_wgpu_test("burn-vision", &args_wgpu)?;
                 }
```
antimora, line 165: Was this meant to be part of this PR?

laggui: Yeah fixing a couple of random things that we missed. But not going to merge until after the release.

PR #3923, opened 2025-10-24
Generalize linalg::outer semantics; add linalg::outer_dim

crutcher:

### Changes

Generalized dimensional semantics for `linalg::outer` through `linalg::outer_dim`.

### DIFF
--- a/crates/burn-tensor/src/tensor/linalg/outer.rs
+++ b/crates/burn-tensor/src/tensor/linalg/outer.rs
@@ -5,15 +5,20 @@ use crate::{AsIndex, Numeric};
 
 /// Computes the outer product for the last columns of 2 tensors.
 ///
-/// See also: `outer_dim`.
+/// See also: [`outer_dim`].
 ///
 /// # Arguments
-/// - `lhs`: the "row" tensor, with shape ``[..., M]``.
-/// - `rhs`: the "col" tensor, with shape ``[..., N]``.
+/// - `lhs`: the "row" tensor, with shape ``[..., i]``.
+/// - `rhs`: the "col" tensor, with shape ``[..., j]``.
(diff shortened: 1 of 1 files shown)


crates/burn-tensor/src/tensor/linalg/outer.rs
```diff
@@ -1,48 +1,64 @@
...
+/// Computes the dimensional outer product for one dimension of two tensors.
+///
+/// # Arguments
+///
+/// - `lhs`: the "row" tensor, with shape ``[..., M, ...]``.
+/// - `rhs`: the "col" tensor, with shape ``[..., N, ...]``.
+/// - `dim`: the dimension to product, supports negative indexing.
+///
+/// # Returns
+/// - the broadcast-shaped ``[..., M, N, ...]`` result.
```
laggui: I would clarify the doc just a bit to make the outer product along dim more clear.

Suggestion:

```rust
/// Computes the outer product along a specific dimension, broadcasting over others.
///
/// For the given `dim`, computes the outer product of elements along that dimension,
/// expanding it into two dimensions of size `M × N` at positions `(dim, dim + 1)`.
///
/// # Arguments
///
/// - `lhs`: left operand, with size `M` at dimension `dim`.
/// - `rhs`: right operand, with size `N` at dimension `dim`.
/// - `dim`: dimension to compute the outer product along (supports negative indexing).
///
/// # Returns
///
/// A tensor of rank `R = D + 1`, where:
///
/// ```
/// result[..., i, j, ...] = lhs[..., i, ...] * rhs[..., j, ...]
/// ```
```

crutcher: done.

Issue #3924, opened 2025-10-25
Add muon optimizer


NewBornRustacean:
What's up Burn community!
I'd like to suggest to add a new optimizer *muon*

### Feature description
Add the Muon optimizer to `burn-optim`. Muon (Momentum Orthogonalized by Newton-schulz) is a new optimizer specifically designed for training neural network hidden layers, particularly effective for large language models and transformers.

Muon combines SGD-momentum with Newton-Schulz orthogonalization, replacing each 2D parameter's update with the nearest orthogonal matrix. This approach provides:
- Faster convergence compared to Adam/AdamW
- Better stability in large-scale training
- Memory efficiency (similar to SGD-momentum)
- Automatic learning rate transfer across different model scales

**References**
- [original implementation](https://github.com/KellerJordan/Muon)
- [pytorch documentation](https://docs.pytorch.org/docs/stable/generated/torch.optim.Muon.html)

### Feature motivation
Adding Muon to Burn would benefit:
- Large language model training projects
- Users seeking faster convergence with less hyperparameter tuning
- Applications where memory efficiency is critical
- ...and more!

### (Optional) Suggest a Solution
- [draft PR for this](https://github.com/tracel-ai/burn/pull/3925)

xstraven:
very cool, just read the muon paper a couple of days ago

bn-c:
looking forward to this 🔥

PR #3925, opened 2025-10-25
Feature muon

NewBornRustacean:
#### What's new?:
- Adds a new Muon optimizer implementation (crates/burn-optim/src/optim/muon.rs).
- Implements Newton–Schulz orthogonalization, momentum integration, optional weight decay, and shape-based learning-rate adjustment (Original and MatchRmsAdamW modes).
- Includes comprehensive unit tests. 
#### Notes:
- [The original implementation](https://github.com/KellerJordan/Muon/blob/master/muon.py#L17) uses `bfloat16` and so is [pytorch implementation](https://github.com/pytorch/pytorch/blob/main/torch/optim/_muon.py#L56). 
- for now in this pr, the module tested with `f32`. `bfloat16` tests could be added soon.

#### Test summary
- `test_adjust_lr_fn_original` -  Verifies the Original learning-rate adjustment ratios for square, tall, and wide matrices.
- `test_adjust_lr_fn_match_rms_adamw` - Verifies the MatchRmsAdamW learning-rate adjustment ratios for example shapes.
- `test_1d_tensor_panics`- Ensures Newton–Schulz orthogonalization panics for 1D tensors (requires 2D).
- `test_muon_optimizer_save_load_state` - Verifies optimizer state can be saved and loaded for a Linear layer without bias.
- `test_muon_with_weight_decay` -  Ensures weight decay is applied (weights are reduced) for a Linear layer without bias.
- `test_newton_schulz_orthogonalization` - Checks Newton–Schulz produces approximately orthogonal output (A * A^T ≈ I).
- `test_tall_matrix_transpose` - Ensures tall matrices are transposed internally and shape is preserved; verifies orthogonalization changes values and wide-matrix behavior.
- `test_zero_gradient` — Confirms Muon handles zero gradients without NaNs, creates state, and weight decay still reduces values when gradients are zero.

### Related issue, readings, etc.
- #3924 
- [Keller Jordan's writeup - Muon: An optimizer for hidden layers in neural networks](https://kellerjordan.github.io/posts/muon/)
- [Keller Jordan's implementation](https://github.com/KellerJordan/Muon/tree/master?tab=readme-ov-file)
- [pytorch implementation  _muon.py](https://github.com/pytorch/pytorch/blob/main/torch/optim/_muon.py)

### DIFF
--- a/crates/burn-optim/src/optim/muon.rs
+++ b/crates/burn-optim/src/optim/muon.rs
@@ -298,9 +298,8 @@ impl<B: Backend> Muon<B> {
     /// - PyTorch: https://github.com/pytorch/pytorch/blob/main/torch/optim/muon.py
     fn zeropower_via_newtonschulz<const D: usize>(&self, g: Tensor<B, D>) -> Tensor<B, D> {
         let shape = g.shape();
-        let dims: Vec<usize> = shape.dims::<D>().into();
-        let dim_m2 = dims[D - 2];
-        let dim_m1 = dims[D - 1];
+        let dim_m2 = shape[D - 2];
+        let dim_m1 = shape[D - 1];
 
         // Step 1: Transpose if tall matrix (more rows than columns)
         let (mut x, needs_transpose) = if dim_m2 > dim_m1 {


crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,787 @@
...
+    ///    - A = X @ X^T
+    ///    - B = b*A + c*A^2
+    ///    - X = a*X + B@X
+    /// 4. Transpose back if needed
+    ///
+    /// # References
+    ///
+    /// - Original: https://github.com/KellerJordan/Muon/blob/master/muon.py
+    /// - PyTorch: https://github.com/pytorch/pytorch/blob/main/torch/optim/muon.py
+    fn zeropower_via_newtonschulz<const D: usize>(&self, g: Tensor<B, D>) -> Tensor<B, D> {
```
NewBornRustacean, line 299: this function could be the first review point 😃 

in pytorch implementation, they reject steps > 100([raise value error](https://github.com/pytorch/pytorch/blob/main/torch/optim/_muon.py#L47)) should we follow the same way?

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,787 @@
...
+            }
+        }
+    }
+}
+
+/// Muon configuration.
+///
+/// Muon is an optimizer specifically designed for 2D parameters of neural network
+/// hidden layers (weight matrices). Other parameters such as biases and embeddings
+/// should be optimized using a standard method such as AdamW.
```
nathanielsimard, line 92: Are those parameters ignored during training if you use only a single optimizer?

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,787 @@
...
+    ///    - B = b*A + c*A^2
+    ///    - X = a*X + B@X
+    /// 4. Transpose back if needed
+    ///
+    /// # References
+    ///
+    /// - Original: https://github.com/KellerJordan/Muon/blob/master/muon.py
+    /// - PyTorch: https://github.com/pytorch/pytorch/blob/main/torch/optim/muon.py
+    fn zeropower_via_newtonschulz<const D: usize>(&self, g: Tensor<B, D>) -> Tensor<B, D> {
+        assert!(
```
nathanielsimard: Unsure if this should be the default behavior. There isn't a great way yet to define multiple optimizers for a single burn module (e.g a linear layer with a bias vector). Do you have an idea: @laggui ?

crates/burn-optim/src/optim/muon.rs
laggui: Unless the model is strictly composed of 2D weights, I think Muon requires a hybrid optimization setup (with multiple optimizers).

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,787 @@
...
+        // Step 2: Normalize by Frobenius norm
+        // X = X / (||X|| + epsilon)
+        let norm = x.clone()
+            .powf_scalar(2.0)
+            .sum()
+            .sqrt()
+            .clamp_min(self.epsilon)
+            .into_scalar();
+
+        x = x.div_scalar(norm);
```
laggui: We should avoid `into_scalar` here, which brings the result to host memory and synchronizes the computation (forcing the GPU to finish all pending work).

```rust
let norm = x.clone()
    .powf_scalar(2.0)
    .sum()
    .sqrt()
    .clamp_min(self.epsilon);
x = x.div(norm);
```

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,787 @@
...
+        // Step 1: Apply momentum
+        let state_momentum = state.map(|s| s.momentum);
+        let (grad, new_momentum_state) = self.momentum.transform(grad, state_momentum);
+
+        // Step 2: Orthogonalize via Newton-Schulz
+        let update = self.zeropower_via_newtonschulz(grad);
+
+        // Step 3: Adjust learning rate based on parameter shape
+        let shape_dims: [_; D] = tensor.shape().dims();
+        let adjusted_lr = self.adjust_lr(lr, &shape_dims);
```
laggui: Don't need to get the shape array, you could simply pass the `Shape` type e.g.:

```rust
let adjusted_lr = self.adjust_lr(lr, &tensor.shape());
```

and `adjust_lr` would take a shape ref instead of a slice.

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,787 @@
...
+    /// different learning rates:
+    /// - Weight decay uses the original `lr`
+    /// - Parameter update uses the shape-adjusted `lr`
+    fn step<const D: usize>(
+        &self,
+        lr: LearningRate,
+        tensor: Tensor<B, D>,
+        grad: Tensor<B, D>,
+        state: Option<Self::State<D>>,
+    ) -> (Tensor<B, D>, Option<Self::State<D>>) {
```
laggui, line 384: Instead of panicking only via `zeropower_via_newtonschulz`, we should probably move the assertion at the beginning of `step` and add a `# Panics` note in the doc.

crates/burn-optim/src/optim/muon.rs
laggui: Hmmm yeah this does shine light on a current limitation of the high level optimizer with burn. The optimizer adaptor assumes that all parameters of a single module will follow the same optimization. But there are many such cases where hybrid optimization is desired.

We need to add a more flexible optimizer that potentially allows specifying which parameters are linked to an optimizer. Either via predicates, or parameter groups. Not sure yet in terms of API.

crates/burn-optim/src/optim/muon.rs
NewBornRustacean: @laggui @nathanielsimard  Hey! sorry to late reply 🥲  I just read your comments, thanks! 

> Are those parameters ignored during training if you use only a single optimizer?

For now, if there're other parameters muon would panic at `step()`.  

in [pytorch implementation](https://github.com/pytorch/pytorch/blob/main/torch/optim/_muon.py#L135C1-L136C1), just raise when the optimizer instantiated.


there could be 2 different ways to deal with this: 

1. handle hybrid optimization _inside_ muon   
2. keep muon with strict implementation(maybe we can validate it at `MuonConfig`), expose higher level API - facade for the multi optimizer
```rust
// maybe added to base optimizer? base.rs
hybrid_opt
  .add_optimizer("hello_world_opt", HelloworldConfg::init())
  .add_optimizer("biases_and_embeddings", AdamWConfig::new().init())
```

I'd like to suggest going for the later.

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,789 @@
...
+    /// Keller Jordan's original method: `lr * sqrt(max(1, A/B))`
+    ///
+    /// This scales the learning rate based on the aspect ratio of the weight matrix,
+    /// ensuring that tall matrices (more rows than columns) get proportionally larger
+    /// learning rates.
+    ///
+    /// # Example
+    ///
+    /// For a [1024, 512] matrix: `lr * sqrt(1024/512) = lr * 1.414`
+    Original,
```
laggui, line 43: Oh and you can use `#[derive(Default)]` and mark this variant with `#[default]` btw

NewBornRustacean: I tried adding `#[derive(Default)]` with `#[default]` but it conflicts with the `Config` derive macro which already implements `Default`. 

The `Config` macro doesn't seem to support the `#[default]` attribute on enum variants, so I've kept the explicit `#[config(default = "AdjustLrFn::Original")]` approach instead

NewBornRustacean: @laggui what's up!  I just added a commit to format the code with cargo fmt 😃 
if this PR is good enough to merge, maybe I could continue to work `hybrid optimizer` feature we talked about.
I'm planning to open an issue with a design proposal for the feature. After we agree on the approach, will begin implementing it!

[added 25. 11. 15 02:16 UTC] 
maybe this [commit](https://github.com/tracel-ai/burn/commit/0eb7eaa85eea73ac5cd1971056cf20f8cfd00d2a) deals with the hybrid part?

laggui: > I tried adding `#[derive(Default)]` with `#[default]` but it conflicts with the `Config` derive macro which already implements `Default`.
> 

Oh, I didn't realize `AdjustLrFn` derived `Config` :sweat_smile: 

It shouldn't have to actually, it's a simply enum. I suggest removing the `Config` derive, and instead derive `Default` (as well as `Clone`, and serde's `Serialize` & `Deserialize` required by the config that includes it).

crates/burn-optim/src/optim/muon.rs
laggui: Yeah I would keep Muon strict, and handle the hybrid optimization at a higher level to combine different optimizers.

crates/burn-optim/src/optim/muon.rs
laggui: > I'm planning to open an issue with a design proposal for the feature. After we agree on the approach, will begin implementing it!

Yep that's a good idea!

The commit you linked is a strategy for multi-device handling of optimizer params, so not quite :slightly_smiling_face:

crates/burn-optim/src/optim/muon.rs
```diff
@@ -0,0 +1,790 @@
...
+    use crate::{GradientsParams, Optimizer};
+    use burn::module::{Module, Param};
+    use burn::tensor::{Distribution, Shape, Tensor, TensorData};
+    use burn_nn::{Linear, LinearConfig, LinearRecord};
+
+    type TestBackend = burn_ndarray::NdArray<f32>;
+
+    const TOLERANCE: f64 = 1e-8;
+
+    fn given_linear_layer(weight: TensorData, bias: TensorData) -> Linear<TestAutodiffBackend> {
```
laggui: This appears to be unused?

NewBornRustacean: oh my bad 🥲  just removed redundant imports. 

@laggui  changed `AdjustLrFn` derive, good to go 🔢

NewBornRustacean:
@laggui hi!  just pull main and merge it(codecov says this PR is behind on main😂

laggui (approved):
Sorry for the late update

Changes LGTM!

Fixed a minor no-std issue that would have required to import `Vec` from alloc, but you can just index the shape directly to get a dimension size.

4pygmalion:
LGTM! Seom.

Issue #3927, opened 2025-10-26
(cuda) panic on memory_cleanup(): "The size should match"


swfsql:
**Describe the bug**
After a linear layer `forward` call, executing a `AutodiffBackend::memory_cleanup` on cuda panics with "The size should match".

**To Reproduce**
Please see [this gist](https://gist.github.com/swfsql/8cf371e7fdec397e6d495fe0f5f710ff).

**Expected behavior**
All pass.

**Screenshots**
(Not applicable)

**Desktop (please complete the following information):**
 - OS: Linux 6.x, cuda 12.x.

**Smartphone (please complete the following information):**
(Not applicable)

**Additional context**
- Three blocks, XYZ, are executed.
- X: A tensor with a `require_grad` test; Y: A `Param` tensor test; Z: A `Linear` model test. I didn't explore much, but I didn't find a simpler test that fails for Z.
- They all pass for ndarray, but block Z fails for cuda.
- Note: Each block probably doesn't interfere with one another. On cuda, when executing XYYZ, or XZY, it always only fails for block Z.
- The gist contains the error message.

amyssnippet:
Hello @swfsql , lemme check this issue and solve it for you asap.

nathanielsimard:
The issue will be in CubeCL, but they are already some updates recently on memory management, so I would be curious about updated bahavior.

swfsql:
@nathanielsimard yes, just tested it and it's no longer panicking. Will be closing this then, thanks!

swfsql:
Hi @nathanielsimard , just a heads-up, I re-tested the linked gist and I'm currently getting a panic on today's (2025-11-22*) main:
```
cleanup 01 X
cleanup 02 X
cleanup 03 X
cleanup 01 Y
cleanup 02 Y
cleanup 03 Y
cleanup 01 Z
cleanup 02 Z
cleanup 03 Z
cleanup 01 X

thread 'main' (13921) panicked at (...)/cubecl-matmul/src/base.rs:542:26:
called `Result::unwrap()` on an `Err` value: Unable to launch matmul because a required feature is unavailable: Types lhs=Scalar(Float(TF32)), rhs=Scalar(Float(TF32)) and/or output=Scalar(Float(F32)) not supported.


stack backtrace:
   (...)
   4: cubecl_matmul::base::launch_ref
             at (...)/cubecl-matmul/src/base.rs:534:25
   5: burn_cubecl::kernel::matmul::base::launch_matmul
[...]

PR #3932, opened 2025-10-27
fix: prevent panic in memory_cleanup on CUDA after linear forward

amyssnippet:
After a linear layer forward call in autodiff mode, executing AutodiffBackend::memory_cleanup on CUDA would panic with 'The size should match'. This occurred because memory_cleanup was called without synchronizing pending asynchronous operations, causing size mismatches in CubeCL's memory management.

The fix adds a sync() call before memory_cleanup in CubeBackend to ensure all operations are completed before cleanup. This resolves the issue for CUDA backends while having no impact on synchronous backends like ndarray.

Added a test in burn-cubecl to verify memory_cleanup works after linear operations.

Files changed:
- crates/burn-cubecl/src/backend.rs: Added sync before memory_cleanup
- crates/burn-cubecl/src/tests/memory_cleanup.rs: New test file
- crates/burn-cubecl/src/tests/mod.rs: Added test module and macro

### Checklist

### Related Issues/PRs

### Changes

### Testing

issue fixed: #3927

### DIFF
--- a/crates/burn-cubecl/src/backend.rs
+++ b/crates/burn-cubecl/src/backend.rs
@@ -66,6 +66,7 @@ where
 
     fn memory_cleanup(device: &Self::Device) {
         let client = R::client(device);
+        futures_lite::future::block_on(client.sync());
         client.memory_cleanup();
     }
 }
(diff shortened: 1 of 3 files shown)


swfsql:
Hi, I've tried running with your changes and it still panics for me. The part that panics is block Z, which has a linear layer. Your test refers to block X (although block X and Y are pretty much the same). Also, I think you accidentally removed `testgen_jit_fusion` when copying it to make `testgen_memory_cleanup`.

Issue #724, opened 2023-08-29
Support for ONNX subgraphs in Burn for conditional inference computation


antimora:
### Description

Introduce the capability to support ONNX subgraphs within the Burn framework, particularly for handling conditional inference computations. This feature aims to enable Burn to fully support modern complex ONNX models that rely on subgraphs for conditional computations.

### Motivation

1. **Complete ONNX Compatibility**: Many advanced models like Version 4 of the [Silero VAD](https://github.com/snakers4/silero-vad) make use of subgraphs for various tasks. Lack of support for ONNX subgraphs would mean incomplete compatibility between Burn and ONNX.

2. **Conditional Inference**: Subgraphs are particularly useful for models that require conditional computation during inference, allowing for more dynamic and efficient operations.

3. **Model Integrity and Flexibility**: Providing support for ONNX subgraphs ensures that Burn can accommodate more complex model architectures without requiring modifications to the original ONNX model.

4. **Elevate Burn's Capabilities**: Incorporating subgraph support would make Burn a more versatile and robust framework, increasing its appeal to a broader audience.

#### Backend Steps

1. **Subgraph Detection**: Parse the ONNX file to identify and isolate subgraphs.
  
2. **Subgraph-to-Module Conversion**: Convert these subgraphs into Burn-compatible modules, maintaining their conditional inference capabilities.

3. **Integration and Testing**: Integrate the converted modules into Burn and conduct thorough tests to ensure they work as expected.

#### Code Changes

- Some refactoring of the existing implementation will be required.
  
- However, this feature should be achievable without major architectural changes.

#### Additional Features

- **Documentation**: Offer detailed documentation and examples that show how to work with ONNX subgraphs within Burn.

- **Error Handling and Warnings**: Implement robust error handling to notify users of any potential issues during the import and conversion process.

- **Logging**: Provide logging mechanisms for debugging and monitoring the performance of subgraph handling.

hkrutzer:
Thanks for creating and open-sourcing Burn! I think there is a lot of demand for running Silero VAD inside Rust, there's several projects using Whisper that want to add a VAD. Looking forward to this feature!

willstott101:
There is a new improved version of Silero VAD (v5) https://github.com/snakers4/silero-vad/blob/master/src/silero_vad/data/silero_vad.onnx - which might adjust priorities in terms of feature support - if the model architecture has changed in any meaningful way (I am yet to test burn).

willstott101:
[...]

Issue #1941, opened 2024-06-29
Support ONNX subgraph for silero vad model


smallbraineng:
### Feature description

I'd love support for the `ConstantOfShape` (9) ONNX Operator, which isn't currently implemented according to the docs

### Feature motivation

To use [Silero VAD](https://github.com/snakers4/silero-vad) on embedded hardware / really cheaply in wasm

The operators needed for that are
```
ai.onnx;1;Not
ai.onnx;9;ConstantOfShape
ai.onnx;11;Conv
ai.onnx;13;Cast,Concat,Equal,Gather,Pad,ReduceMean,Size,Split,Sqrt,Squeeze,Unsqueeze
ai.onnx;14;Add,LSTM,Relu
ai.onnx;15;Pow,Shape
ai.onnx;16;Identity,If
com.microsoft;1;FusedConv
```

<3

antimora:
there is an outstanding pr for constant https://github.com/tracel-ai/burn/issues/1815

antimora:
The direct link file: https://github.com/snakers4/silero-vad/blob/master/files/silero_vad.onnx

Attached: 
[silero_vad.onnx.zip](https://github.com/user-attachments/files/16045486/silero_vad.onnx.zip)


In addition to missing ConstantOfShape, Burn ONNX needs to support subgraphs, see #724.

CCing @skewballfox. This is something to consider for our future design.

smallbraingames:
Ah got it, that makes sense. Looks like that will take some larger design changes?

antimora:
Yes, that's correct. But it is good that you file now we know to prioritize this feature. I have been wanting to use it as well.

smallbraingames:
Love it. Can't wait to use it and have a 10x smaller wasm

polarathene:
Status update almost a year later. The ONNX file location has [since changed](https://github.com/snakers4/silero-vad/commit/f0d880d79c50d36c9c014e282a89fe42b0683120) (July 2024), with only the latest v5 ONNX model [offered here](https://github.com/snakers4/silero-vad/tree/master/src/silero_vad/data) (_[replacing v4 at the end of June 2024](https://github.com/snakers4/silero-vad/commit/fd1f1a62b75ec5824993aa4062b55c0965b9621b)_).

So for reference here are the permalinks:

- [v4](https://github.com/snakers4/silero-vad/blob/v4.0/files/silero_vad.onnx) (Oct 2022)
[...]

PR #3936, opened 2025-10-27
Add ONNX control flow operators: `If`, `Loop`, and `Scan`

antimora:
This PR adds support for ONNX Control Flow Operators (`If`, `Loop`, and `Scan`) to the burn-import crate, enabling conditional execution, general-purpose looping, and sequential processing with state accumulation. The implementation supports deeply nested control flow structures with 100% ONNX spec compliance.

### Checklist

### Related Issues/PRs

- Fixes https://github.com/tracel-ai/burn/issues/724
- Prerequisite for https://github.com/tracel-ai/burn/issues/1941

### Changes

- `If` Operator - Conditional execution with then/else branches
- `Loop` Operator - General-purpose looping with loop-carried dependencies
- `Scan` Operator - Sequential processing with state accumulation
- Nested Subgraphs - Support for deeply nested control flow (`If`→`Loop`→`If`→`Scan`)
- 100% ONNX Spec Compliance - All operators validated against official ONNX spec
- Boolean tensor support for `ReduceMin`/`ReduceMax` operations

### Testing

New unit and onnx tests. All existing tests are passing. Few existing tests were fixed.


antimora:
Rebased with https://github.com/tracel-ai/burn/pull/3872 changes.

antimora:
@laggui, I have rebased with main. Unfortunately, I initially messed up. So I had to squash my subgraph related commits into one.

crates/onnx-ir/src/phases/post_processing.rs
```diff
@@ -126,19 +228,26 @@ fn apply_identity_elimination(
         resolved_rewire_map.insert(output.clone(), current);
     }
 
-    // Step 3: Rewire node inputs
+    // Step 3: Rewire subgraphs in control flow nodes (If/Loop/Scan)
+    // This must happen BEFORE rewiring regular node inputs to ensure subgraph
+    // inputs/outputs that reference Identity node outputs are properly updated
+    for node in nodes.iter_mut() {
+        rewire_node_subgraphs(node, &resolved_rewire_map, &output_arg_map);
+    }
```
laggui, line 236: Could likely just have a node mapping instead to point to control flow nodes that contain subgraphs. So we don't have to iterate through all nodes all the time.

For now this is fine, but just throwing it out there for the future since we keep adding iterations over all graph nodes at each step :smile:

crates/burn-import/src/burn/node/if_node.rs
```diff
@@ -0,0 +1,264 @@
...
+    fn forward(&self, scope: &mut Scope, node_position: usize) -> TokenStream {
+        let cond = match &self.condition {
+            Type::Scalar(scalar) => {
+                let name = &scalar.name;
+                quote! { #name }
+            }
+            Type::Tensor(tensor) => {
+                let cond_tensor = scope.tensor_use_owned(tensor, node_position);
+                // Convert tensor to bool - assume it's a scalar tensor
+                quote! { #cond_tensor.into_scalar().elem::<bool>() }
```
laggui, line 147: There is no other way around it if the condition is based on a runtime tensor value, but eventually I think we could document some drawbacks of ONNX import that might be optimized manually (e.g. if it is mostly data flow).

crates/burn-import/src/burn/node/if_node.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    // Convert ONNX nodes to Burn nodes
+    let burn_nodes: Vec<_> = subgraph
+        .nodes
+        .iter()
+        .map(|node| {
+            try_convert_onnx_node::<PS>(node.clone()).unwrap_or_else(|| {
+                unsupported_ops.push(node.node_type.clone());
+                panic!("Unsupported op in subgraph: {:?}", node.node_type)
```
laggui: The check for `!unsupported_ops.is_empty()` will never be reached with the panic here.. so the `unsupported_ops` vec is useless at this time.

crates/burn-import/src/burn/node/loop_node.rs
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    // Convert ONNX nodes to Burn nodes
+    let burn_nodes: Vec<_> = subgraph
+        .nodes
+        .iter()
+        .map(|node| {
+            try_convert_onnx_node::<PS>(node.clone()).unwrap_or_else(|| {
+                unsupported_ops.push(node.node_type.clone());
+                panic!("Unsupported op in loop body: {:?}", node.node_type)
```
laggui: Same comment as `If` node

crates/burn-import/src/burn/node/loop_node.rs
```diff
@@ -0,0 +1,419 @@
...
+        // Get body graph from config
+        let config = node.config::<onnx_ir::node::loop_node::LoopConfig>();
+        let body = config.body.clone();
+
+        // Loop-carried dependencies are inputs after M and cond
+        let v_initial: Vec<Type> = node.inputs.iter().skip(2).map(Type::from).collect();
+
+        // Outputs are the final values of loop-carried dependencies
+        // (and potentially scan outputs, but we'll handle those later)
+        let outputs: Vec<Type> = node.outputs.iter().map(Type::from).collect();
```
laggui: Regarding the scan outputs, I noticed that the codegen does not currently handle them?

So this doesn't cover the whole spec for the node (already better than what we have anyway :sweat_smile:), but this might fail on models that expect to use these outputs right?

crates/burn-import/src/burn/node/scan_node.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    // Convert ONNX nodes to Burn nodes
+    let burn_nodes: Vec<_> = subgraph
+        .nodes
+        .iter()
+        .map(|node| {
+            try_convert_onnx_node::<PS>(node.clone()).unwrap_or_else(|| {
+                unsupported_ops.push(node.node_type.clone());
+                panic!("Unsupported op in scan body: {:?}", node.node_type)
```
laggui: Same here :slightly_smiling_face:

crates/burn-import/src/burn/node/scan_node.rs
```diff
@@ -0,0 +1,427 @@
...
+                Type::Tensor(tensor) => {
+                    let tensor_name = &tensor.name;
+                    let iter_expr = if reverse {
+                        quote! { seq_len - i - 1 }
+                    } else {
+                        quote! { i }
+                    };
+                    slice_stmts.extend(quote! {
+                        let #var_name = #tensor_name.clone().slice([#iter_expr..#iter_expr + 1]).squeeze();
+                    });
```
laggui, line 263: Is this going to apply to the correct scan axis?

Also, shouldn't only on dimension be squeezed?

laggui (commented):
Just a couple of comments, mostly on the new control flow ops codegen

crates/burn-import/src/burn/node/if_node.rs
antimora: With more real usage, it will be interesting to see how much this is a bottleneck. Maybe there is a way to optimize hot paths for scalars, like making these available in advance. This probably especially useful for dynamic graphs Food for thought, for you and @nathanielsimard =D

crates/onnx-ir/src/phases/post_processing.rs
antimora: Once we stabilize structural refactoring (Enum node, lazy data loading), we can improve the transformation flow. I agree there is a lot of opportunities to optimize.

crates/burn-import/src/burn/node/loop_node.rs
antimora: Made the full spec now! Thanks for the catch.

crates/burn-import/src/burn/node/scan_node.rs
antimora: It looks we were missing `scan_output_axes` attribute configuration. Ooops. Fixed now. Thanks for the catch =)

antimora:
@laggui thanks for your previous detailed review. I have addressed your comments and fixed semi large issues. It should be 100% spec compliant now =D If not, lets fix them again.

PR #3938, opened 2025-10-28
ci: let CI server dispatch the test-gpu workflow

syl20bnr:


syl20bnr:
Merging because of the new CI server version in production.

PR #3941, opened 2025-10-28
Fix/qlinear

nathanielsimard:
Fix fusion configuration to fuse more.


syl20bnr:
@nathanielsimard rebase against main and the GPU tests should be triggered. Note that now the triggers are unified and all processed by our CI server which means that they won't appear in the CI checks but there will be notifications from `bot-ember`.

laggui (approved):

PR #3950, opened 2025-10-29
add fusion matmul optimization i64 support

tzemanovic:
### Checklist

### Changes

Adds i64 support to matmul fusion.

I ran into the `Unsupported precision` catch-all case with i64. It seems that simply adding a case to handle this works, at least in my external work. One thing I'm not sure about if the `row_count` for `FusedMatmulSelector::OrderedDoubleBuffering` on line 506 in this file should also be updated to handle this precision?

### Testing

Only tested externally in private code. I'd be happy to add some tests here with some guidance

### DIFF
--- a/crates/burn-cubecl-fusion/src/matmul/optimization.rs
+++ b/crates/burn-cubecl-fusion/src/matmul/optimization.rs
@@ -361,6 +361,7 @@ impl<R: Runtime> TraceRunner<R> for FusedMatmul {
         configs: &'a [FuseBlockConfig],
     ) -> Result<(), FusedMatmulError> {
         match self.out.precision() {
+            FusePrecision::F64 => self.matmul_fused::<R, f64>(client, inputs, outputs, &configs[0]),
             FusePrecision::F32 => self.matmul_fused::<R, f32>(client, inputs, outputs, &configs[0]),
             FusePrecision::Flex32 => {
                 self.matmul_fused::<R, flex32>(client, inputs, outputs, &configs[0])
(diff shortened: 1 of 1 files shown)


nathanielsimard (commented):
It's not just i64, but all int types that are not supported in fusion. This can be fixed by extending the enum match case.

tzemanovic:
> It's not just i64, but all int types that are not supported in fusion. This can be fixed by extending the enum match case.

thx, I've added all the other int types (unsigned too) and also f64, so the only one that's not supported is bool

nathanielsimard:
We fixed the problem without increasing compilation time. Thanks for pointing out the issue.

tzemanovic:
> We fixed the problem without increasing compilation time. Thanks for pointing out the issue.

hi @nathanielsimard, I appreciate your help on this! I'm trying to use the new version, but I'm still not about use matmul with i64 (using a rev after #3998 merge - 4a34da7f678341aeceb3a87d1caa11d95cfa8260), crashing on:

```
 thread 'layers::transformer::mha::test::test_mha_padding' (437328) panicked at cubecl/crates/cubecl-runtime/src/tune/tuner.rs:346:17:
    No autotune was flagged as valid for the problem.
```

The tuner loop at this location fails to find a required feature with: `"Unable to launch matmul because a required feature is unavailable: Cmma on lhs Scalar(Int(I64)) rhs Scalar(Int(I64)) and output Scalar(Int(I64)) with shape m=16, n=16, k=8 not supported."`

Should I open an issue for this?

PR #3951, opened 2025-10-29
[WIP] float_dtypes(device) -> Vec<FloatDType> sketch.

crutcher:
This is a first-pass sketch at threading introspection operations for building a B::Device => capability mapping.

This is mostly trivial, but I don't know what to do about the Router case.

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -44,6 +44,10 @@ fn unsqueeze_like<B: Backend>(
 }
 
 impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> {
+    fn float_dtypes(device: &Device<Self>) -> Vec<FloatDType> {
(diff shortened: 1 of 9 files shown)


nathanielsimard:
I think it would be better to have a `B::is_dtype_supported(device, dtype) -> bool`. It avoids an allocation and it assigns the dtype to the device. A backend can have multiple devices with different dtype support.

laggui (commented):
[cross-posting my earlier discord comment here]

Hmmm I don't think this should be defined for the `FloatTensorOps` trait. Also, in line with the [original tracking issue](<https://github.com/tracel-ai/burn/issues/3532>), I think we should have something like:

```rust
let is_supported = B::supports_dtype(&device, dtype);
```

And if we really want a matrix of supported dtypes, would need something like
```rust
pub struct BackendDTypes {
    pub float: Vec<DType>,
    pub int: Vec<DType>,
    pub bool: Vec<DType>,
}

fn supported_dtypes(&device) -> BackendDTypes;
```

but not sure if that is desirable.

PR #3957, opened 2025-10-30
Fix autodiff graph cleanup

### LINKED ISSUE #3926: Need help in avoiding memory leak (ndarray, ram)
**Describe the bug**
I think detaching or using the gradients, currently, are not clearing memory.

**To Reproduce**
Please see [this gist](https://gist.github.com/swfsql/07da79e7386b4679910291173524ea4a).

**Expected behavior**
A somewhat constant memory footprint. Currently, each train function call consumes ~1GB of ram.

**Screenshots**
(Not applicable)

(Not applicable, I think)

(Not applicable)

**Additional context**
- I've added two tests, one which detaches and drops the loss, and another that uses sgd.
- Interestingly, when using detach, it appears there is some vector reallocation since the timing has some spikes.
- I'm not sure if I'm doing something incorrectly =P

laggui:
### Checklist

### Related Issues/PRs

Fixes #3926
Closes #3947
and other reported issues related to this on discord.

### Changes

- Clean up consumed nodes in the graphs state
- Handle untracked nodes for register + cleanup

### Testing

Debugged while using the simple regression example (20 epochs only), the current (main) graph management accumulates over 343k graphs with 1-2 nodes only, for a total of over 353k nodes.

```
INFO burn_autodiff::runtime::graph: Report for 343210 graphs
INFO simple_regression::model: [Report] GraphReport { steps: 340618, nodes: 353618, leaves: 340610 }
```

Now, it only holds onto a max of 9 graphs (17 nodes) at the peak, and on the last backward:
```
INFO burn_autodiff::runtime::graph: Report for 8 graphs
INFO simple_regression::model: [Report] GraphReport { steps: 16, nodes: 16, leaves: 8 }
```

### DIFF
--- a/crates/burn-autodiff/src/runtime/graph.rs
+++ b/crates/burn-autodiff/src/runtime/graph.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use alloc::sync::Arc;
 use burn_common::stub::Mutex;
-use burn_tensor::backend::{Backend, GraphReport};
+use burn_tensor::backend::Backend;
 use hashbrown::{HashMap, HashSet};
 
 /// A client for managing multiple graphs using mutex-based synchronization.


crates/burn-autodiff/src/runtime/graph.rs
```diff
@@ -273,6 +291,35 @@ impl GraphLocator {
         self.keys.insert(origin, HashSet::new());
         graph
     }
+
+    fn mark_untracked(&mut self, node_id: NodeId) {
+        self.untracked_nodes.insert(node_id);
+    }
+
+    /// Clean up untracked nodes.
+    fn cleanup_untracked(&mut self) {
```
nathanielsimard, line 300: So untracked nodes were not cleanup from the autodiff server?

laggui: These nodes were registered just like tracked nodes

https://github.com/tracel-ai/burn/blob/139800550249623b5d8f8b5555f923b080559436/crates/burn-autodiff/src/ops/base.rs#L180-L187

With the new graph locator, many of these untracked nodes ended up in singleton graphs composed of this single node. So we collected an increasing amount of useless untracked single node graphs in the hashmap without this cleanup.

nathanielsimard (approved):

PR #3961, opened 2025-10-31
feat: Enable tuning for MMA matmul

wingertge:
### Checklist

### Related Issues/PRs

Requires https://github.com/tracel-ai/cubecl/pull/1003

### Changes

Adds lower level MMA matmul to tuning for both fused and unfused matmul.

### Testing

All tests pass and `burn-lm` works as expected (but 25-40% faster depending on the model).

### DIFF
--- a/crates/burn-cubecl-fusion/src/matmul/optimization.rs
+++ b/crates/burn-cubecl-fusion/src/matmul/optimization.rs
@@ -342,7 +342,7 @@ pub struct FusedMatmul {
 #[derive(Debug)]
 pub enum FusedMatmulError {
     LaunchError(MatmulSetupError),
-    InvalidInput,
+    InvalidInput(&'static str),
 }
 
 impl From<MatmulSetupError> for FusedMatmulError {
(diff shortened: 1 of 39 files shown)


crates/burn-cubecl-fusion/src/matmul/optimization.rs
```diff
@@ -106,11 +113,44 @@ impl MatmulVariants {
...
+            specialized_mma: selector(FusedMatmulSelector::DoubleBuffering {
+                specialized: true,
+                tile_matmul: AcceleratedTileKind::Mma,
+            }),
+            ordered: selector(FusedMatmulSelector::OrderedDoubleBuffering {
+                tile_matmul: AcceleratedTileKind::Cmma,
+            }),
+            ordered_mma: selector(FusedMatmulSelector::OrderedDoubleBuffering {
+                tile_matmul: AcceleratedTileKind::Mma,
+            }),
```
nathanielsimard, line 153: There is a lot of duplications here. I think a good refactor would be to store the `selectors` in a vector and select them based on indexing. Unsure how to actually do it, but that would be cleaner.

wingertge: Maybe just a `HashMap` or `BTreeMap`, with the selector as the key?

wingertge: I think there's something to be said for keeping it statically typed so all the variants are properly initialized. So I'd keep it for now.

nathanielsimard: I can look into, I want to refactor that project with better documentation

PR #3967, opened 2025-11-02
matvec

huy209vn:
implement matvec in burn-linalg
cleaned up from #3882


crates/burn-tensor/src/tensor/linalg/matvec.rs
```diff
@@ -0,0 +1,62 @@
+// What: Introduce a matrix-vector helper that mirrors Tensor::matmul for ergonomic access.
+// Why: Reuses the optimized matmul kernels while surfacing clear shape validation for vectors.
+// Creative: Adds batch-aware validation with minimal reshaping by unsqueezing/squeezing exactly once.
```
laggui:
```suggestion
```

PR #3974, opened 2025-11-03
feat: In-memory dataset loading of Arrow Record Batches

ap0phasi:
### Checklist

### Changes

Added `arrow` feature to In-Memory datasets that allows for loading from Arrow Record Batches, such as what is produced by `datafusion` or `duckdb-rs`. 

### Testing

Two new feature-dependent tests have been added demonstrating using `datafusion` to query the same test csv used in the existing `from_csv_rows` test. These tests can be run with `cargo test --feature arrow`, which executes an example `datafusion` query within a `tokio` runtime.

### DIFF
--- a/burn-book/src/building-blocks/dataset.md
+++ b/burn-book/src/building-blocks/dataset.md
@@ -263,6 +263,26 @@ let dataset = InMemDataset::from_csv("path/to/csv", rdr).unwrap();
 
 Note that this requires the `csv` crate.
 
+### Arrow Record Batches
+`InMemDataset` can also load data from `arrow` record batches produced by query engines such as `datafusion` and `duckdb-rs`. This requires the optional `arrow` feature which utilizes the `ar_row` and `ar_row_derive` crates.
+
+```rust, ignore
+// Build dataset from arrow recrod batches
+#[derive(Debug, Default, Clone, Serialize, ar_row_derive::ArRowDeserialize, PartialEq)]
+pub struct SampleCsvArrow {
+    column_str: String,
+    column_int: i64,
+    column_bool: bool,
+    column_float: f64,
+}
+let ctx = SessionContext::new();
+let df = ctx.read_csv("path/to/csv", CsvReadOptions::new()).await.unwrap();
+
+let record_batches: Vec<RecordBatch> = df.collect().await.unwrap()
+
+let dataset = InMemDataset::<SampleCsvArrow>::from_arrow_batches(record_batches).unwrap();
+```
+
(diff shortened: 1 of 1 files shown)


crates/burn-dataset/src/dataset/in_memory.rs
```diff
@@ -88,6 +93,26 @@ where
...
+where
+    I: ArRowDeserialize + Clone,
+{
+    /// Create from Arrow Record Batches
+    /// See:
+    /// - [Deserialize structs in a row-oriented way from Apache Arrow](https://docs.rs/ar_row_derive/latest/ar_row_derive/index.html#examples)
+    pub fn from_arrow_batches(record_batches: Vec<RecordBatch>) -> Result<Self, std::io::Error> {
+        let items: Vec<I> = record_batches
+            .into_iter()
+            .flat_map(|batch| -> Vec<I> { I::from_record_batch(batch).unwrap() })
```
laggui, line 108: We should probably propagate the result/error instead of possibly panicking w/ unwrap.

The return type `Result` should also reflect this change (not a `std::io::Error`, as opposed to the other methods).

burn-book/src/building-blocks/dataset.md
```diff
@@ -263,6 +263,26 @@ let dataset = InMemDataset::from_csv("path/to/csv", rdr).unwrap();
...
+```rust, ignore
+// Build dataset from arrow recrod batches
+#[derive(Debug, Default, Clone, Serialize, ar_row_derive::ArRowDeserialize, PartialEq)]
+pub struct SampleCsvArrow {
+    column_str: String,
+    column_int: i64,
+    column_bool: bool,
+    column_float: f64,
+}
+let ctx = SessionContext::new();
```
laggui, line 278: Should probably indicate that this is imported from `datafusion` given that it is not a burn type

laggui (changes requested):
Sorry for the delay!

Just a couple of comments, otherwise LGTM

ap0phasi:
Thanks, I agree it makes sense to have separate ArrowDataset, CsvDataset, and JsonDatasets. I'll break out a separate ArrowDataset implementation and better propagate the results/errors. Would you like me to do a separate PR for separating out CSV and JSON datasets?

laggui:
I think this PR can focus on the arrow dataset changes, and if you feel like following up for the csv and json sources another PR can be opened after :slightly_smiling_face:

laggui:
Closing due to inactivity, but feel free to re-open if you want to pick this up again! :slightly_smiling_face:

PR #3983, opened 2025-11-05
Implement `FromStr` for `Slice` with parsing and error handling

crutcher:
- Add `FromStr` implementation to parse string patterns into `Slice`.
- Introduce custom `SliceParseError` to handle parsing errors.
- Extend tests to cover various valid and invalid parsing cases.
- Include `core::str::FromStr` import to support changes.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/slice.rs
+++ b/crates/burn-tensor/src/tensor/api/slice.rs
@@ -647,6 +647,7 @@ impl SliceExpressionError {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::alloc::string::ToString;
 
     #[test]
     fn test_parse_error() {


crates/burn-tensor/src/tensor/api/slice.rs
```diff
@@ -507,10 +509,87 @@ impl From<i32> for Slice {
     }
 }
 
+/// Parse error for [`Slice`] parsing.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SliceParseError {
+    /// The value that failed to parse.
+    pub value: String,
+}
```
laggui: The error type should be an enum, and could (or should I say, should) have variants to be a bit more specific as to what failed to parse perhaps. Was the slice empty? Is the integer invalid? Are the specified steps unsupported (e.g., zero)? Was the step separator specified without any steps following? Did it not match the expected format (i.e., could not parse at all)?

crates/burn-tensor/src/tensor/api/slice.rs
```diff
@@ -507,10 +509,87 @@ impl From<i32> for Slice {
...
+            value: buf.to_string(),
+        };
+        let parse =
+            |v: &str| -> Result<isize, Self::Err> { v.parse::<isize>().map_err(|_| make_error()) };
+
+        let mut start: isize = 0;
+        let mut end: Option<isize> = None;
+        let mut step: isize = 1;
+
+        if let Some((head, tail)) = s.split_once(":") {
```
laggui: We should split on `;` to match the macro syntax.

crates/burn-tensor/src/tensor/api/slice.rs
crutcher: I'm hesitant about this; because I don't know where to stop. I can see a more informative error message, but it's hard to see how we'll have a recovery pathway (which is the usual argument for enum error cases; mechanically handling the cases).

Possible Cases?
* Invalid(source, msg)
* ParseError(source, msg)

Also, that raises the larger "should Burn have a common parse error type"?

crates/burn-tensor/src/tensor/api/slice.rs
crutcher: Oh; missed this didn't match; changed.

crutcher (commented):
Added matching `Display` impl.

crates/burn-tensor/src/tensor/api/slice.rs
laggui: Yeah even if it's only two variants for invalid vs parsing error, with this at least the reason is more clear (instead of just throwing back the value at the user). I'm good with that!

crates/burn-tensor/src/tensor/api/slice.rs
```diff
@@ -507,10 +510,122 @@ impl From<i32> for Slice {
...
+            if let Some(end) = self.end {
+                f.write_fmt(format_args!("{}", end))?;
+            }
+            if self.step != 1 {
+                f.write_fmt(format_args!(";{}", self.step))?;
+            }
+            Ok(())
+        }
+    }
+}
```
laggui, line 532: Nice!

It would be helpful to add some tests to show the expected output for some input slices.

You can use `slice.to_string()` which uses the `Display` implementation.

laggui (commented):
Sorry for the late review :sweat_smile:

crates/burn-tensor/src/tensor/api/slice.rs
crutcher: Done

crutcher:
Abstracted out a common `ExpressionError` enum.

crutcher:
I think that `1..-3` should be illegal (`sign(b - a) != sign(step)`); but there's no enforcement of that yet in `Slice`.

PR #4017, opened 2025-11-13
Refactor metric logger : epoch summary and multiple entries at once

Charles23R:
### Checklist

### Changes

We can now log multiple metric entries at a time.
Small refactor of `NumericEntry` since the naming wasn't consistent with what it was being used for.
Add `EpochSummary` to the loggers.

### Testing

Tested with the examples.


nathanielsimard (approved):
I looks good! I think we should refactor `MetricEntry` where some fields are moved into `MetricDefinition` like the name.

PR #4022, opened 2025-11-14
Add GaussianNoise layer

kul-sudo:
# Add GaussianNoise layer

### Checklist

### Motivation & What's new

I've come across an article talking about noise injection as a means of overfitting prevention and improvement of generalization. I've tested this technique several times in my project and I've been glad with the result, so I've come to the conclusion a layer like this might be helpful for others.
For more information: https://arxiv.org/abs/2307.06855

### Testing

| Test name | Goal | Comments |
| :---:   | :---: | :---: |
| `with_ad_backend_should_mark_input` | Ensure applying noise has an effect during training.  | The word "mark" isn't entirely clear to me in this context, but I've seen it used in the implementation of `Dropout`, so probably it makes sense to stick to this lexicon. | 
| `without_ad_backend_should_not_change_input` | Same as the first test, but this one makes sure the noise has no effect whatsoever during inference. | |
| `negative_std_should_panic` | According to the formula, the standard deviation can't be negative by definition. | It's important to mention that `-0.0` would panic too. However, there's a justification: if someone tries to assign a value like that instead of simply `0.0`, they're clearly unaware of what they're doing. |


kul-sudo:
Added the missing test and updated the PR description accordingly. Should be fine now, so feel free to review.

laggui (approved):

PR #4027, opened 2025-11-16
Fix incorrect example in burn-store KeyRemapper docs

SzczurekYT:
I lost over 3 hours due to this and ~~another bug that I'm going to report soon~~ 🫩
Just a docs fix so I didn't fill in the checklist.

Edit: That other bug is the fact that `PytorchReader` does not store paths into `TensorSnapshots`, but looks like it is intended/not a bug so nevermind


SzczurekYT:
Nevermind, I just realized the docs are also actually correct.
I'm dealing with some weird issues trying to port the Yolox model from burn models to burn store and thought this is the culprit.
Sorry

PR #4031, opened 2025-11-17
Refactor MetricEntry

Charles23R:
### Checklist

### Changes

Refactor `MetricEntry` : 
- Remove name and tags fields.
- Getting the Metric information should now be done via the metric definitions.
- The tag of the experiment is passed to the processor/logger.

### Testing

Unit tests and running examples.

### DIFF
--- a/crates/burn-train/src/logger/metric.rs
+++ b/crates/burn-train/src/logger/metric.rs
@@ -22,6 +22,7 @@ pub trait MetricLogger: Send {
     /// * `epoch` - Current epoch.
     /// * `split` - Current dataset split.
     /// * `iteration` - Current iteration.
+    /// * `tag` - Optional, additional tag for the split.
     fn log(
         &mut self,
         items: Vec<MetricEntry>,


crates/burn-train/src/logger/metric.rs
```diff
@@ -21,7 +22,13 @@ pub trait MetricLogger: Send {
     /// * `epoch` - Current epoch.
     /// * `split` - Current dataset split.
     /// * `iteration` - Current iteration.
```
nathanielsimard, line 24: Missing doc entry for tags

crates/burn-train/src/logger/metric.rs
```diff
@@ -21,7 +22,13 @@ pub trait MetricLogger: Send {
...
-    fn log(&mut self, items: Vec<&MetricEntry>, epoch: usize, split: Split);
+    fn log(
+        &mut self,
+        items: Vec<MetricEntry>,
+        epoch: usize,
+        split: Split,
+        tag: Option<Arc<String>>,
```
nathanielsimard, line 31: I don't recal the tags, are they applied on all the metrics or are they per-metric?

Charles23R: They are applied to all metrics for the whole duration of this epoch. In this spirit, i think it's logical to just pass it along with the epoch and split during logging. 
Also it was kind of weird because it was a list of tags, but the metric entries were always tagged only once with the evaluation label. Basically, there could only be 1 tag.

PR #4032, opened 2025-11-17
Refactor/dtype cubecl

nathanielsimard:
Reduce binary size by removing generic and using `define` attributes in cubecl kernels.

Requires: https://github.com/tracel-ai/cubecl/pull/1055

### DIFF
--- a/crates/burn-tensor/src/tensor/quantization/bytes.rs
+++ b/crates/burn-tensor/src/tensor/quantization/bytes.rs
@@ -146,7 +146,7 @@ fn read_bytes_to_i8(bytes: Bytes) -> Vec<i8> {
         // Safety,
         //
         // `Vec<u8>` can be Re-interpreted as `Vec<i8>` since they share the same alignment.
-        Err(bytes) => unsafe { core::mem::transmute(bytes.to_vec()) },
+        Err(bytes) => unsafe { core::mem::transmute::<Vec<u8>, Vec<i8>>(bytes.to_vec()) },
     }
 }
 


crates/burn-cubecl/src/kernel/conv/deform_conv_transpose2d.rs
```diff
@@ -55,9 +49,13 @@ pub(crate) fn deform_conv2d_backward<
...
-        let grad = CubeBackend::<R, E, I, BT>::float_sum_dim(out_grad.clone(), 0);
-        let grad = CubeBackend::<R, E, I, BT>::float_sum_dim(grad, 2);
-        let grad = CubeBackend::<R, E, I, BT>::float_sum_dim(grad, 3);
+        let grad =
+            reduce_dim::<R>(out_grad.clone(), 0, Default::default(), ReduceFnConfig::Sum).unwrap();
+        let grad = reduce_dim::<R>(grad, 2, Default::default(), ReduceFnConfig::Sum).unwrap();
+        let grad = reduce_dim::<R>(grad, 3, Default::default(), ReduceFnConfig::Sum).unwrap();
+        // let grad = CubeBackend::<R, E, I, BT>::float_sum_dim(out_grad.clone(), 0);
+        // let grad = CubeBackend::<R, E, I, BT>::float_sum_dim(grad, 2);
+        // let grad = CubeBackend::<R, E, I, BT>::float_sum_dim(grad, 3);
```
laggui: Dead code

crates/burn-cubecl/src/kernel/quantization/dequantize.rs
```diff
@@ -1,27 +1,31 @@
...
         DType::QFloat(scheme) => {
-            // TODO: Only works with contiguous tensors.
-            cubecl_quant::dequantize::launch_ref::<R, F>(
+            cubecl_quant::dequantize::launch_ref::<R>(
                 &values.client,
                 &values.as_handle_ref(),
                 &output.as_handle_ref(),
                 &params.as_handle_ref(),
                 &scheme,
+                tensor.dtype.into(),
```
laggui: Looks like quantization tests are failing..

~~The dequantize kernel expects a float dtype but this will yield the int storage dtype?~~

nvm the kernel signature changed in cubecl. It's actually the output dtype that is incorrect when creating `empty_device_dtype` above, it should be specified instead of using `tensor.dtype`

Issue #3989, opened 2025-11-07
Eliminate burn-import abstraction layer and use onnx-ir types directly


antimora:
Currently, burn-import duplicates onnx-ir types (Arguments, Nodes, Configs) and converts between them. This creates many lines of unnecessary conversion code.

Current: ONNX → onnx-ir → burn-import types → codegen

Proposed: ONNX → onnx-ir → codegen (directly)

Benefits:

  - Remove many LOC of conversion boilerplate
  - Single source of truth (onnx-ir)
  - Simpler maintenance (no parallel type hierarchies)
  - Clearer design: onnx-ir validates, burn-import generates code

This should be done after: https://github.com/tracel-ai/burn/issues/3988

PR #4033, opened 2025-11-18
Remove burn-import abstraction layer and use onnx-ir types directly

antimora:
This PR represents a major architectural refactoring of the ONNX import pipeline in Burn. It eliminates the abstraction layer between `burn-import` and `onnx-ir`, implementing code generation logic directly on ONNX IR types.

Fixes https://github.com/tracel-ai/burn/issues/3989

### Checklist

### Changes

**Before:** ONNX → onnx-ir → burn-import types → codegen  
**After:** ONNX → onnx-ir → codegen (directly)

##### 1. Architectural Simplification

- **Direct type usage**: Eliminates wrapper types and uses `onnx_ir::Argument` and `onnx_ir::ir::ArgType` directly
- **Reduced duplication**: Centralizes node logic in the IR library instead of maintaining parallel implementations
- **Cleaner abstractions**: Replaces trait-based indirection with direct enum dispatch via match statements

##### 2. Code Generation Improvements

- **Better type safety**: Using onnx-ir types directly ensures consistency between parsing and codegen
- **Helper utilities**: New `argument_helpers.rs` provides clean utility functions for type token generation
- **Field consolidation**: Consolidates field type/value logic for simpler maintenance

##### 3. Testing Enhancements

- **Snapshot testing**: Migration to `insta` crate for cleaner, more maintainable tests
- **Focused tests**: Testing individual node forward passes rather than full graph codegen
- **Better coverage**: Added subgraph tests for control flow operators (If, Loop, Scan)

##### 4. Developer Experience

- **Derive macros**: `NodeBuilderDerive` reduces boilerplate for test fixtures
- **Error handling**: Consistent use of let-else patterns for cleaner error paths
- **Documentation**: Good use of TODOs and FIXMEs to track technical debt

### Testing

The existing codegen tests were replaced and made more comprehensive. Also existing onnx-tests pass.

### DIFF
--- a/crates/burn-import/src/burn/node/loop_node.rs
+++ b/crates/burn-import/src/burn/node/loop_node.rs
@@ -117,8 +117,10 @@ impl<PS: PrecisionSettings + 'static> NodeCodegen<PS> for onnx_ir::node::loop_no
         let num_loop_carried_outputs = num_loop_vars;
         let num_scan_outputs = self.outputs.len() - num_loop_carried_outputs;
 
-        // Initialize loop-carried dependency variables
-        let mut init_stmts = quote! {};
+        // Get body input and output variable names
+        // Body inputs: [iter_num, cond_in, v_in...]
+        let iter_name = arg_to_ident(&self.config.body.inputs[0]);
+        let cond_in_name = arg_to_ident(&self.config.body.inputs[1]);
         let loop_var_names: Vec<_> = self
             .config
             .body
@@ -128,12 +130,30 @@ impl<PS: PrecisionSettings + 'static> NodeCodegen<PS> for onnx_ir::node::loop_no
             .map(arg_to_ident)
             .collect();
 
+        // Body outputs: [cond_out, v_out..., scan_outputs...]
+        let cond_out_name = arg_to_ident(&self.config.body.outputs[0]);
+        let loop_out_names: Vec<_> = self
+            .config
+            .body
+            .outputs
+            .iter()
+            .skip(1)
+            .take(num_loop_carried_outputs)
+            .map(arg_to_ident)
+            .collect();
+
+        // Initialize loop-carried dependency variables
+        // Only mark as mutable if the variable is actually updated (different name from output)
+        let mut init_stmts = quote! {};
         for (idx, initial_arg) in v_initial_args.iter().enumerate() {
             let var_name = &loop_var_names[idx];
             let init_value = arg_to_ident(initial_arg);
 
-            // Variables that are updated by the loop need to be mutable
-            if idx < num_loop_carried_outputs {
+            // Check if this variable will be updated (different name means it gets assigned)
+            let needs_mut = idx < num_loop_carried_outputs
+                && loop_out_names.get(idx).is_some_and(|out| var_name != out);
+
+            if needs_mut {
                 init_stmts.extend(quote! {
                     let mut #var_name = #init_value;
                 });
@@ -143,11 +163,6 @@ impl<PS: PrecisionSettings + 'static> NodeCodegen<PS> for onnx_ir::node::loop_no
                 });
             }
         }
-
-        // Initialize iteration counter and condition
-        // Body inputs: [iter_num, cond_in, v_in...]
-        let iter_name = arg_to_ident(&self.config.body.inputs[0]);
-        let cond_in_name = arg_to_ident(&self.config.body.inputs[1]);
 
         // Initialize scan output collectors if any
         let mut scan_init = quote! {};
@@ -178,26 +193,24 @@ impl<PS: PrecisionSettings + 'static> NodeCodegen<PS> for onnx_ir::node::loop_no
         let body_code =
             generate_loop_body_code::<PS>(&self.config.body, scope.scope(), node_position);
 
-        // Extract body outputs
-        let cond_out_name = arg_to_ident(&self.config.body.outputs[0]);
-        let loop_out_names: Vec<_> = self
-            .config
-            .body
-            .outputs
-            .iter()
-            .skip(1)
-            .take(num_loop_carried_outputs)
-            .map(arg_to_ident)
-            .collect();
-
         // Update loop-carried variables after iteration
+        // Skip self-assignments when body passes through a value unchanged (same name)
         let mut update_vars = quote! {};
         for (idx, out_name) in loop_out_names.iter().enumerate() {
             let var_name = &loop_var_names[idx];
-            update_vars.extend(quote! {
-                #var_name = #out_name;
-            });
+            if var_name != out_name {
+                update_vars.extend(quote! {
+                    #var_name = #out_name;
+                });
+            }
         }
+
+        // Update condition from body output (skip if same name)
+        let update_cond = if cond_in_name != cond_out_name {
+            quote! { #cond_in_name = #cond_out_name; }
+        } else {
+            quote! {}
+        };
 
         // Collect scan outputs - handle scalar vs tensor
         let mut collect_scans = quote! {};
@@ -279,6 +292,14 @@ impl<PS: PrecisionSettings + 'static> NodeCodegen<PS> for onnx_ir::node::loop_no
             quote! { (#(#output_values),*) }
         };
 
+        // cond_in only needs mut if it's actually updated (different name from output)
+        let cond_needs_mut = cond_in_name != cond_out_name;
+        let cond_init = if cond_needs_mut {
+            quote! { let mut #cond_in_name = #init_cond; }
+        } else {
+            quote! { let #cond_in_name = #init_cond; }
+        };
+
         quote! {
             #[allow(unused_variables, unused_assignments)]
             #output_decls = {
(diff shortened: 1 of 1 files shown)


crates/burn-import/src/burn/node/attention.rs
```diff
@@ -342,549 +260,499 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for AttentionNode {
...
+                let qk = q_scaled.matmul(k_transpose);
+                let capped = qk;
+                let scores = softmax(capped, 3);
+                let output = scores.matmul(v);
+                let output = output
+                    .permute([0, 2, 1, 3])
+                    .reshape([batch_size as i32, q_sequence_length as i32, -1]);
+                (output,)
+            };
+        ");
```
laggui, line 377: I see two issues with using snapshots for burn import tests:
1. For more complex nodes (such as this), writing out the expected code as a string makes it even more error prone. And even for simpler nodes, multi-line code is indentation-sensitive and gets quite noisy.
2. The tests no longer check that the correct inputs/outputs are generated (graph codegen), it only checks the forward pass code block (e.g., if the input is `Int` instead of `Float`). So we lose coverage for the scopes, though it might be partially mitigated by ONNX import integration tests for models.

crates/burn-import/src/burn/node/div.rs
```diff
@@ -72,9 +35,18 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for DivNode {
...
-            (Type::Scalar(_), Type::Scalar(_)) => quote! { #lhs / #rhs },
-            (Type::Shape(_), Type::Shape(_)) => quote! {
+            (ArgType::Tensor(_), ArgType::Scalar(_)) => quote! { #lhs.div_scalar(#rhs) },
+            (ArgType::Scalar(_), ArgType::Tensor(_)) => {
+                // Scalar / Tensor: convert scalar to tensor and divide
+                quote! {
+                    {
+                        let scalar_tensor = Tensor::<B, 1>::from_data([#lhs.elem::<B::FloatElem>()], &*self.device);
+                        scalar_tensor.div(#rhs)
+                    }
```
laggui: For floats, you can just leverage the already implemented `Div`:

https://github.com/tracel-ai/burn/blob/459cdb0dec828bbe28180123a0d8946ac4554033/crates/burn-tensor/src/tensor/api/numeric.rs#L4459-L4473 

should probably check that the scalar is a float too

crates/burn-import/src/burn/node/test_helpers.rs
```diff
@@ -0,0 +1,141 @@
...
+            scope.tensor_register_future_use(input, node_position + 1);
+        }
+    }
+
+    // Generate code using the node's forward method with ScopeAtPosition
+    let mut scope_at_pos = scope.at_position(node_position);
+    let code = node.forward(&mut scope_at_pos);
+
+    // Format the statement by wrapping in a function, formatting, then extracting
+    format_statement(code)
```
laggui: Running the generated code through a formatter means we're testing a transformed version of your codegen, no? That could hide or introduce potential issues.

crates/burn-import/src/burn/node/test_helpers.rs
```diff
@@ -0,0 +1,141 @@
...
+    //
+    // We want to extract just the "let output = ...;" part
+    extract_statement_from_function(&formatted)
+}
+
+/// Extract the statement from a formatted function body
+fn extract_statement_from_function(formatted: &str) -> String {
+    // Find the opening brace and closing brace
+    let start = formatted.find('{').map(|i| i + 1);
+    let end = formatted.rfind('}');
```
laggui: That seems a little brittle :sweat_smile:

crates/onnx-ir/onnx-ir-derive/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+/// - `input_tensor_shape(name, shape, dtype)` - Add tensor input with static shape
+/// - `input_scalar(name, dtype)` - Add scalar input
+/// - `input_shape(name)` - Add shape input
+/// - `output_tensor(name, rank, dtype)` - Add output tensor
+/// - `output_scalar(name, dtype)` - Add scalar output
+/// - `output_shape(name)` - Add shape output
+/// - `config(config)` - Set config (if node has a config field)
+/// - `build()` - Build the node
+#[proc_macro_derive(NodeBuilderDerive)]
+pub fn node_builder_derive(input: TokenStream) -> TokenStream {
```
laggui, line 31: This is strictly used in tests right? It's useful now that you only test snapshots for single node forward codegen, without going through the graph input/outputs registration.

I guess `onnx-ir` users could leverage this too, but seems like 

I would rename it simply to `NodeBuilder`, but this seems to conflict with the current `NodeBuilder` struct which is a bit of a misnomer. I suggest renaming the current struct to `RawNode`, as it more clearly communicates that the node is still being processed.

And the derive can simply be `NodeBuilder`.

crates/onnx-ir/src/lib.rs
```diff
@@ -17,3 +17,7 @@ mod util;
 pub use ir::*;
 pub use node::*;
 pub use pipeline::{OnnxIrError, parse_onnx};
+
+// Export GraphState for testing
+#[doc(hidden)]
+pub use graph_state::GraphState;
```
laggui: Hmmm don't think this should be publicly re-exported. I saw that it is required for the new constant node test, but likely just a byproduct of the design issue for `Argument` which should not hold the `GraphState`

as per my previous comment: https://github.com/tracel-ai/burn/pull/3872#discussion_r2487738073

laggui (changes requested):
The refactor to implement `NodeCodeGen` on `onnx-ir` type LGTM! That's what I was looking for in the previous PR :clap: 

```rust
impl<PS: PrecisionSettings> NodeCodegen<PS> for onnx_ir::node::abs::AbsNode {
    fn inputs(&self) -> &[Argument] {
        &self.inputs
    }

    fn outputs(&self) -> &[Argument] {
        &self.outputs
    }

    fn forward(&self, scope: &mut ScopeAtPosition<'_>) -> TokenStream {
        let input = scope.arg(self.inputs.first().unwrap());
        let output = arg_to_ident(self.outputs.first().unwrap());

        quote! {
            let #output = #input.abs();
        }
    }
}
```

very clean and simple.

I think most of the other changes should have been addressed in different PRs. That would have also made it easier to review. Please keep this in mind :pray: 

For `NodeDeriveBuilder` & `insta` codegen tests:
These changes make the tests much simpler to write for contributors, and clearly isolate only node-level codegen with less boilerplate.

But there are some drawbacks, see my comments below.

crates/burn-import/src/burn/node/attention.rs
antimora: The cool part of insta is that you don't need to write the expected code string by hand. If you run cargo insta review, you can review and confirm the expected output. I recommend trying it out yourself. It is pretty cool and fast. The failure message points how to do this.

antimora: I will fix the second part. I agree it is useless.

crates/burn-import/src/burn/node/test_helpers.rs
antimora: If not formatted, you would simply see a stream of tokens (space separated tokens in a string ).

crates/burn-import/src/burn/node/attention.rs
laggui: > The cool part of insta is that you don't need to write the expected code string by hand. If you run cargo insta review, you can review and confirm the expected output. 

Ahhh well that's good to know! Makes it less error prone then :slightly_smiling_face:

crates/onnx-ir/src/lib.rs
antimora: I can hide it again and remove the tests

crates/burn-import/src/burn/node/test_helpers.rs
laggui: Yeah the comment was linked to the statement extraction as part of the format (comment below).

The rest looks fine, it uses the same pretty please formatting as the codegen.

crates/onnx-ir/src/lib.rs
laggui: Sure, you can comment out the test until the comment is addressed.

crates/burn-import/src/burn/node/test_helpers.rs
antimora: It removes dummy function: 

image

crates/burn-import/src/burn/node/test_helpers.rs
antimora: OK. I changed to generate a proper forward function with inputs/outputs.

crates/burn-import/src/burn/node/test_helpers.rs
antimora: Changed it.

crates/burn-import/src/burn/node/test_helpers.rs
antimora: image

PR #4034, opened 2025-11-18
Fix raspberry pi pico example not compiling

BjornTheProgrammer:
### Checklist

### Related Issues/PRs

Just this PR.

### Changes

This just removes the `unsafe-assume-single-core` that ends up breaking the raspberry-pi-pico from compiling, since you can't use critical section.

### Testing

I've removed the raspberry-pi-pico exclude from the `Cargo.toml`, since it seems like it is working now.


BjornTheProgrammer:
Attempting to fix CI issue in https://github.com/embassy-rs/embassy/pull/4909, before proceeding further

BjornTheProgrammer:
Additionally xtask doc needs to probably be updated to get the check to pass properly

BjornTheProgrammer:
I've gotten https://github.com/embassy-rs/embassy/pull/4948 merged, but now it seems that xtask might have changed a little since two weeks ago. I'll look into the underlying issue and submit a PR if needed.

BjornTheProgrammer:
When running `cargo clippy` it runs for every single package in the workspace, and there doesn't seem to be a way to exclude an example or package to be excluded from compiling currently, as seen in https://github.com/rust-lang/rust-clippy/issues/9555

BjornTheProgrammer:
I'm just gonna exclude the example from the workspace like it was previously, which means that future changes still can break it. Until `cargo clippy` adds a way to exclude, I don't see an easy way to do this. At the very least, this should ensure that `0.20.0-pre.5` and `0.20.0` still works with the microcontrollers.

PR #4035, opened 2025-11-18
BUG: Fixed burn version 

Marc-AnthonyG:
### Checklist

### Changes

Pre-release right now are evaluated as patch by cargo while being breaking that cause issue for the an end user that would like to use a specific pre-release because cargo will force the user to use the latest. When trying to specify a burn version using the equal operator the main burn crate do get the right version but other sub crate use their latest pre-release as demonstrated below. Thus, at least for pre-release we need to fix version with equal operator.

**Cargo.toml**
`burn = { version = "=0.20.0-pre.1",  features = ["vision", "train"] }`

**Cargo.lock**
```

[[package]]
name = "burn"
version = "0.20.0-pre.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "3cfefc1825375ea0517bd18604ce7296061232e1155f82336cee6f8e459d0640"
dependencies = [
 "burn-autodiff",
 "burn-candle",
 "burn-collective",
 "burn-core",
 "burn-cpu",
 "burn-cuda",
 "burn-ndarray",
 "burn-nn",
 "burn-optim",
 "burn-remote",
 "burn-rocm",
 "burn-router",
 "burn-store",
 "burn-tch",
 "burn-train",
 "burn-wgpu",
]

[[package]]
name = "burn-autodiff"
version = "0.20.0-pre.2"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "2c0805025908fb5d5126d2ef99cff9debe735c1916998f84cd57e92ddaa732f1"
dependencies = [
 "burn-common",
 "burn-tensor",
 "derive-new",
 "hashbrown 0.15.5",
 "log",
 "num-traits",
 "portable-atomic",
 "spin",
]

```

### DIFF
--- a/examples/import-model-weights/Cargo.toml
+++ b/examples/import-model-weights/Cargo.toml
@@ -4,7 +4,7 @@ edition.workspace = true
 license = "MIT OR Apache-2.0"
 name = "import-model-weights"
 publish = false
-version = "=0.20.0-pre.2"
+version = "0.20.0-pre.2"
 
 [lints]
 workspace = true


laggui:
Audit failed because of the `tokenizers` crate which still uses an older version of indicatif. `number_prefix` was switched in a later release but looks like huggingface doesn't update their crates very often.

```
  error: 1 denied warning found!
  Crate:     number_prefix
  Version:   0.4.0
  Warning:   unmaintained
  Title:     number_prefix crate is unmaintained
  Date:      2025-11-17
  ID:        RUSTSEC-2025-0119
  URL:       https://rustsec.org/advisories/RUSTSEC-2025-0119
  Dependency tree:
  number_prefix 0.4.0
  └── indicatif 0.17.11
      └── hf-hub 0.4.3
          └── tokenizers 0.22.1
              ├── text-generation 0.20.0-pre.2
              └── text-classification 0.20.0-pre.2
                  └── multi-gpus 0.20.0-pre.2
```

we should ignore `RUSTSEC-2025-0119` in audit config for now, `tokenizers` is only used in the examples.

PR #4042, opened 2025-11-19
Implement Slice iterator and utility methods.

crutcher:
- Add `SliceIter` struct and implement `Iterator` for `Slice`.
- Enable `Slice` to be converted into an iterator using `IntoIterator`.
- Add utility methods: `into_vec` and `clip_size` for enhanced slice manipulation.
- Include comprehensive unit tests for new functionality.

### DIFF
--- a/crates/burn-std/src/tensor/slice.rs
+++ b/crates/burn-std/src/tensor/slice.rs
@@ -351,6 +351,7 @@ impl Iterator for SliceIter {
     }
 }
 
+/// Note: Unbounded [`Slice`]s produce infinite iterators.
 impl IntoIterator for Slice {
     type Item = isize;
     type IntoIter = SliceIter;


crates/burn-std/src/tensor/slice.rs
```diff
@@ -321,6 +321,46 @@ pub struct Slice {
...
+                    return None;
+                }
+            } else if next >= end {
+                return None;
+            }
+        }
+
+        Some(next)
+    }
+}
```
laggui, line 352: Hmmm shouldn't an iterator always produce valid indices, and thus only work with a bounded slice?

I feel like the iterator cannot be correct without normalization :thinking: 

Do you have a different interpretation of what a slice iterator should return?

crates/burn-tensor/src/tensor/api/slice.rs
```diff
@@ -348,6 +388,52 @@ impl Slice {
...
+    ///     Slice::new(0, Some(5), 1).clip_size(10),
+    ///     Slice::new(0, Some(5), 1));
+    /// assert_eq!(
+    ///     Slice::new(0, None, -1).clip_size(10),
+    ///     Slice::new(0, Some(-11), -1));
+    /// assert_eq!(
+    ///     Slice::new(0, Some(-5), -1).clip_size(10),
+    ///     Slice::new(0, Some(-5), -1));
+    /// ```
+    pub fn clip_size(self, size: usize) -> Self {
```
laggui: I don't think size is the right word here. It is setting the end bound. I would suggest something like `bound_to(end)`

laggui (changes requested):

crates/burn-std/src/tensor/slice.rs
crutcher: Iterators in rust are permitted to be infinite; and accept bounds via `.take(n)`.

crates/burn-tensor/src/tensor/api/slice.rs
crutcher: done.

crates/burn-std/src/tensor/slice.rs
laggui: Yes I know, that's not my point.

My concern is when the **slice** is not bounded. The values produced by the iterator, interpreted as indices, will be returned without normalization to the dimension size. That's what I am worried about, not the unbounded iterator.

crutcher: Yes; and that is the correct behavior in unbounded fields.

We have two choices here:
- panic
- infinite iteration

There are arguments for both. Infinite iteration has applications in some spaces, but requires special handling. Panic prevents those applications, but adds some training wheels.

Improper use of the infinite iteration will generally also produce a panic; and it seems that the question is ease of debugging that case.

If you're asking if it is a bug; I'd say it is a defensible application.

If you're telling me to make it a panic; I can.

laggui: I wasn't implying that it was a bug, or that it should panic necessarily. As you said, I think both choices are defensible.

As long as it is clearly documented :slightly_smiling_face:

crutcher: Done; added to the impl.

PR #4050, opened 2025-11-21
Move types from `burn-tensor`

laggui:
First part of a bigger refactor to split `burn-tensor` types and traits.

No runtime changes, just better crate organization by actual responsibility.

### Checklist

### Changes

New `burn-std` crate for core types and utilities used by burn crates (ours, and contributors):
- `Shape`
- `Slice`
- `AsIndex` and indexing utilities
- `Bytes` (re-exported) and `QuantizedBytes`
- `DType`
- `QuantScheme` (re-exported)

New `burn-backend` crate:
- `Element`
- `Distribution`
- and of course, `TensorData`

will eventually contain the `Backend` trait, and `burn-tensor` will only have the higher level `Tensor` struct stuff.

Crate names still opened to discussion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1965,7 +1965,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.9.0-pre.3"
-source = "git+https://github.com/tracel-ai/cubecl?rev=4f9a7d8f54271b5a144c3a827a27b2a2e19618e0#4f9a7d8f54271b5a144c3a827a27b2a2e19618e0"
+source = "git+https://github.com/tracel-ai/cubecl?rev=bf6126cb67eb802a92d1dee05bf2a1c556d68151#bf6126cb67eb802a92d1dee05bf2a1c556d68151"
 dependencies = [
  "cubecl-attention",
  "cubecl-convolution",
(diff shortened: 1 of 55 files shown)


.github/workflows/publish.yml
```diff
@@ -292,7 +292,7 @@ jobs:
   publish-burn-core:
```
nathanielsimard, line 303: Do we want to add the new 'burn-backend' crate in the publish script?

crates/burn-backend/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+workspace = true
+
+[features]
+default = ["std"]
+doc = ["default"]
+std = ["rand/std", "num-traits/std", "burn-std/std"]
+cubecl = ["burn-std/cubecl"]
+
+[dependencies]
+burn-std = { path = "../burn-std", version = "0.20.0-pre.2", default-features = false }
```
nathanielsimard: Version is wrong, should be =0.20.0-pre.3

crates/burn-core/Cargo.toml
```diff
@@ -26,18 +26,18 @@ doc = [
...
     # Doc features
-    "burn-common/doc",
+    "burn-std/doc",
     "burn-dataset/doc",
     "burn-tensor/doc",
 ]
 dataset = ["burn-dataset"]
 
-network = ["burn-common/network"]
+network = ["burn-dataset?/network"]
```
nathanielsimard, line 35: The network package is now in burn-dataset? Is there another use-case for network other than downloading datasets?

.github/workflows/publish.yml
laggui: Good catch :sweat_smile: we almost always forget to do that with new crates the first time around

crates/burn-core/Cargo.toml
laggui: Right now it is only used, even in the examples, to download datasets. I didn't feel it should be in `burn-std`, so I moved it entirely.

Lmk if you have other preferences.

PR #4051, opened 2025-11-22
burn-store enhancements for troubleshooting and new enum skip flag

### LINKED ISSUE #4036: burn-store fails to load PyTorch .pth files due to tensor name mismatch (PyTorchToBurnAdapter issue)
**Describe the bug**
When using burn-store as a burn-import replacement to load models from .pth files, an error occurs indicating many tensors are not found (especially ones ending with 'beta' or 'gamma'). Investigation suggests that the PyTorchToBurnAdapter may not be working as intended: in the `Applier` struct's `apply_tensor` method, tensor snapshots are being queried using burn-specific names, whereas the .pth file contains PyTorch names. As a result, the adapter is invoked after the failed lookup: too late to do correct name mapping.

**To Reproduce**
1. Clone this branch: https://github.com/SzczurekYT/tracel-ai-models/tree/yolox-burn-store
2. Attempt to load a model from a .pth file with burn-store as a replacement for burn-import.
3. Observe that loading fails with errors about missing tensors (notably those with 'gamma' or 'beta' in their names).

**Expected behavior**
burn-store should correctly map PyTorch tensor names to their burn equivalents with the adapter at the appropriate stage, allowing models to load from .pth files without missing tensor errors.

**Additional context**
- This problem was discussed on Discord (see [attached conversation](https://discord.com/channels/1038839012602941528/1091796857996451942/1439621330398875772))
- Failing paths often involve 'gamma' or 'beta' in their names
- Issue likely in adapter timing within `Applier::apply_tensor`
- As a side note, if 'BaseConv' appears in the path, the tensor is also not found, possibly a separate issue
- Maintainers confirmed that a fix is planned
- Easy reproduction from the shared tracel-ai-models branch

antimora:
This PR addresses issue #4036 by introducing comprehensive support for handling enum variant path mismatches when loading PyTorch models into Burn modules. The key insight is that Burn includes enum variant names in tensor paths (e.g., "feature.BaseConv.weight") while PyTorch doesn't (e.g., "feature.weight").

### Checklist

### Related Issues/PRs

Fixes #4036 

### Changes
- **Added `skip_enum_variants` parameter** to `Collector`, `Applier`, and store implementations, allowing paths to be built/matched without enum variant names for PyTorch compatibility
- **Introduced `get_alternative_param_name`** to `ModuleAdapter` trait for handling parameter naming differences (e.g., PyTorch's "weight" vs Burn's "gamma" in normalization layers)
- **Refactored error reporting** by moving `ApplyResult` and `ApplyError` to a dedicated module with enhanced diagnostics including edit-distance based suggestions and enum variant detection

### Testing

- Added comprehensive tests for enum variant path mismatch scenarios
- Included Python script to generate test data simulating PyTorch state dicts

### Screenshot

image

### DIFF
--- a/crates/burn-store/src/adapter.rs
+++ b/crates/burn-store/src/adapter.rs
@@ -312,12 +312,12 @@ mod tests {
         let adapter = PyTorchToBurnAdapter;
 
         // Linear layer weight should be transposed
-        let snapshot = create_test_snapshot("fc.weight", vec![10, 5], "Struct:Linear");
+        let snapshot = create_test_snapshot("fc.weight", vec![10, 5], module_names::LINEAR);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.shape, vec![5, 10]);
 
         // Linear layer bias should not be transposed
-        let snapshot = create_test_snapshot("fc.bias", vec![10], "Struct:Linear");
+        let snapshot = create_test_snapshot("fc.bias", vec![10], module_names::LINEAR);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.shape, vec![10]);
     }
@@ -327,12 +327,12 @@ mod tests {
         let adapter = PyTorchToBurnAdapter;
 
         // BatchNorm weight -> gamma
-        let snapshot = create_test_snapshot("norm.weight", vec![10], "Struct:BatchNorm");
+        let snapshot = create_test_snapshot("norm.weight", vec![10], module_names::BATCH_NORM);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.full_path(), "norm.gamma");
 
         // BatchNorm bias -> beta
-        let snapshot = create_test_snapshot("norm.bias", vec![10], "Struct:BatchNorm");
+        let snapshot = create_test_snapshot("norm.bias", vec![10], module_names::BATCH_NORM);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.full_path(), "norm.beta");
     }
@@ -342,7 +342,7 @@ mod tests {
         let adapter = BurnToPyTorchAdapter;
 
         // Linear layer weight should be transposed
-        let snapshot = create_test_snapshot("fc.weight", vec![5, 10], "Struct:Linear");
+        let snapshot = create_test_snapshot("fc.weight", vec![5, 10], module_names::LINEAR);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.shape, vec![10, 5]);
     }
@@ -352,12 +352,12 @@ mod tests {
         let adapter = BurnToPyTorchAdapter;
 
         // BatchNorm gamma -> weight
-        let snapshot = create_test_snapshot("norm.gamma", vec![10], "Struct:BatchNorm");
+        let snapshot = create_test_snapshot("norm.gamma", vec![10], module_names::BATCH_NORM);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.full_path(), "norm.weight");
 
         // BatchNorm beta -> bias
-        let snapshot = create_test_snapshot("norm.beta", vec![10], "Struct:BatchNorm");
+        let snapshot = create_test_snapshot("norm.beta", vec![10], module_names::BATCH_NORM);
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.full_path(), "norm.bias");
     }
@@ -393,7 +393,7 @@ mod tests {
         let adapter = PyTorchToBurnAdapter;
 
         // Without container info, adapter returns unchanged for non-norm parameters
-        let mut snapshot = create_test_snapshot("fc.weight", vec![10, 5], "Struct:Linear");
+        let mut snapshot = create_test_snapshot("fc.weight", vec![10, 5], module_names::LINEAR);
         snapshot.container_stack = None;
 
         // Without container info, no transformation occurs for linear layers


antimora:
CC @crutcher making easy to see errors. Feedback is appreciated.

crates/burn-derive/src/module/codegen_struct.rs
```diff
@@ -104,12 +105,13 @@ impl ModuleCodegen for StructModuleCodegen {
 
     fn gen_map(&self) -> TokenStream {
         let struct_name = self.name.to_string();
+        let container_type = format!("Struct:{}", struct_name);
```
laggui, line 108: What's the advantage of adding the `Struct:` prefix? Only the type name is relevant for the container no?

For enums, probably `EnumType::EnumVariant` should be adopted to know the specific variant.

crates/burn-store/README.md
```diff
@@ -103,20 +103,23 @@ use burn_store::{PyTorchToBurnAdapter, BurnToPyTorchAdapter, PytorchStore};
 // Load PyTorch .pth file directly (PyTorchToBurnAdapter is applied automatically)
 let mut store = PytorchStore::from_file("pytorch_model.pth")
     .with_top_level_key("state_dict")         // Access nested state dict
+    .skip_enum_variants(true)                 // Match PyTorch paths to Burn enum paths
```
laggui: Shouldn't this simply be the default behavior when loading pytorch models?

crates/burn-store/src/apply_result.rs
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        // Remove the first found variant and return the modified path
+        let mut result_segments = segments.clone();
+        result_segments.remove(variant_indices[0]);
+        Some(result_segments.join("."))
+    }
+
+    /// Find similar paths for a given missing path (for "Did you mean?" suggestions)
+    fn find_similar_paths(&self, missing_path: &str, max_suggestions: usize) -> Vec<String> {
```
laggui, line 130: Hey that's a pretty useful debugging tool :smile: users will love it for sure

laggui (commented):
Had a quick look before leaving, I have some more general comments

crates/burn-derive/src/module/codegen_struct.rs
antimora: We capture Enum variant name as part of name path. This is equivalent to an index of a Vec or array. Struct/Enum/Vec/Array are additional information to tell what kind of type we have.

crates/burn-store/README.md
antimora: I suppose this will be safer for burn-store since we are loading tensors are against a model instance whereas Serde deserialization creates a new instance. This means:

1. Enum variants are not picked like in serde and uses the actual instance. 
2. No new item is added or removed into/from Vec (like shape overriding).

antimora: I will make true for PytorchStore

crates/burn-store/src/apply_result.rs
antimora: Switched path suggestion to Jaro similarity from [textdistance.rs lib](https://github.com/life4/textdistance.rs) (which is no-std compatible). Jaro is better suited for our case per their suggestion:

> Jaro is included in the Elixir standard library ([String.jaro_distance](https://hexdocs.pm/elixir/1.12/String.html#jaro_distance/2)). It is used by the compiler and by mix (cargo for Elixir) to provide the "did you mean?" functionality for typos in module or command names.

This algo gives more weight to the prefix string. So there is there are two: `module1.bias` vs `boo.bias` for `module1.beta`, the alg will suggest `module1.bias`.

crates/burn-store/src/adapter.rs
```diff
@@ -252,12 +312,12 @@ mod tests {
         let adapter = PyTorchToBurnAdapter;
 
         // Linear layer weight should be transposed
-        let snapshot = create_test_snapshot("fc.weight", vec![10, 5], module_names::LINEAR);
+        let snapshot = create_test_snapshot("fc.weight", vec![10, 5], "Struct:Linear");
```
laggui: Still part of the module names though?

crates/burn-store/src/adapter.rs
```diff
@@ -252,12 +312,12 @@ mod tests {
...
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.shape, vec![5, 10]);
 
         // Linear layer bias should not be transposed
-        let snapshot = create_test_snapshot("fc.bias", vec![10], module_names::LINEAR);
+        let snapshot = create_test_snapshot("fc.bias", vec![10], "Struct:Linear");
```
laggui: Same

crates/burn-store/src/adapter.rs
```diff
@@ -267,12 +327,12 @@ mod tests {
         let adapter = PyTorchToBurnAdapter;
 
         // BatchNorm weight -> gamma
-        let snapshot = create_test_snapshot("norm.weight", vec![10], module_names::BATCH_NORM);
+        let snapshot = create_test_snapshot("norm.weight", vec![10], "Struct:BatchNorm");
```
laggui: Same

crates/burn-store/src/adapter.rs
```diff
@@ -267,12 +327,12 @@ mod tests {
...
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.full_path(), "norm.gamma");
 
         // BatchNorm bias -> beta
-        let snapshot = create_test_snapshot("norm.bias", vec![10], module_names::BATCH_NORM);
+        let snapshot = create_test_snapshot("norm.bias", vec![10], "Struct:BatchNorm");
```
laggui: Same

crates/burn-store/src/adapter.rs
```diff
@@ -282,7 +342,7 @@ mod tests {
         let adapter = BurnToPyTorchAdapter;
 
         // Linear layer weight should be transposed
-        let snapshot = create_test_snapshot("fc.weight", vec![5, 10], module_names::LINEAR);
+        let snapshot = create_test_snapshot("fc.weight", vec![5, 10], "Struct:Linear");
```
laggui: Same

crates/burn-store/src/adapter.rs
```diff
@@ -292,12 +352,12 @@ mod tests {
         let adapter = BurnToPyTorchAdapter;
 
         // BatchNorm gamma -> weight
-        let snapshot = create_test_snapshot("norm.gamma", vec![10], module_names::BATCH_NORM);
+        let snapshot = create_test_snapshot("norm.gamma", vec![10], "Struct:BatchNorm");
```
laggui: Same

crates/burn-store/src/adapter.rs
```diff
@@ -292,12 +352,12 @@ mod tests {
...
         let adapted = adapter.adapt(&snapshot);
         assert_eq!(adapted.full_path(), "norm.weight");
 
         // BatchNorm beta -> bias
-        let snapshot = create_test_snapshot("norm.beta", vec![10], module_names::BATCH_NORM);
+        let snapshot = create_test_snapshot("norm.beta", vec![10], "Struct:BatchNorm");
```
laggui: Same :sweat_smile:

crates/burn-store/src/adapter.rs
```diff
@@ -332,16 +392,16 @@ mod tests {
     fn test_no_container_info() {
         let adapter = PyTorchToBurnAdapter;
 
-        // Without container info, adapter returns unchanged
-        let mut snapshot = create_test_snapshot("fc.weight", vec![10, 5], module_names::LINEAR);
+        // Without container info, adapter returns unchanged for non-norm parameters
+        let mut snapshot = create_test_snapshot("fc.weight", vec![10, 5], "Struct:Linear");
```
laggui: Another one

PR #4064, opened 2025-11-25
Fix scatter values backward

### LINKED ISSUE #4063: Burn's scatter values backward bug
**Describe the bug**
Burn's `x.scatter(d, indices, values)` requires that `indices` and `values` have the same dimensions, but doesn't make the same requirement for `x` (and indeed there is no reason why one would want to have such a requirement, given that entries of `indices` could range over whatever bounds are appropriate for `x`).

However, while `x.scatter(1, indices, values)` works fine when `x` has dimensions different to `indices` and `values` at inference time (as expected), with e.g. Autodiff<B<f32>> for B any of NdArray, Candle, Cuda, or Cpu (but not Wgpu or Tch) the whole stack crashes after a `.backward()` due to an assertion/memory access violation depending on the backend. It happens because, internally, autodiff attempts to use the backend's ordinary `.scatter()` (to propagate gradients backwards), and since `x` and `indices` have different dimensions this immediately blows up.

I'm no expert, but I think the implementation of scatter autodiff is just broken:

https://github.com/tracel-ai/burn/blob/66b51fc5ebb514c96c874ca25ab109cc9d504749/crates/burn-autodiff/src/ops/tensor.rs#L991C1-L1014C10
```rust
            fn backward(
                self,
                ops: Ops<Self::State, 2>,
                grads: &mut Gradients,
                _checkpointer: &mut Checkpointer,
            ) {
                let (dim, indices, shape_lhs, shape_rhs, device) = ops.state;
                let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));


                binary::<B, _, _>(
                    ops.parents,
                    ops.node,
                    grads,
                    |grad| {
                        let zeros = B::float_zeros(shape_lhs, &device, grad.dtype().into());
                        B::float_scatter(dim, grad, indices_4lhs.unwrap(), zeros)
                    },
                    |grad| {
                        let zeros = B::float_zeros(shape_rhs, &device, grad.dtype().into());
                        B::float_scatter(dim, zeros, indices_4rhs.unwrap(), grad)
                    },
                );
            }
        }
```

I think a correct implementation should simply be something like (note the `rhs` in the first line):
```rust
|grad| grad
|grad| B::float_gather(dim, grad, indices_4rhs.unwrap()),
```

Given "equal size tensors" (avoiding the crash), I think the current implementation still (wrongly) zeroes entries of the tensor gradient direction which are not written, as well as permuting the value gradients according to index order which is also wrong?

Perhaps this happened because the author(s) didn't realize that Burn's scatters have "add" and not "write" semantics?

**Reproducer**
```rust
use burn::backend::{Autodiff, Candle, Cuda, NdArray, LibTorch, Wgpu};
use burn::tensor::backend::AutodiffBackend;

fn scatter_notequal_forward_to_backward<B: AutodiffBackend>() {
    let device = B::Device::default();

    let flat_len = 1000000;
    let active_len = 30000;
    let batch = 1;

    let x = Tensor::<B, 2>::zeros([batch, flat_len], &device).require_grad();

    let idx_host: Vec<i32> = (0..active_len as i32).collect();
    let indices_1d = Tensor::<B, 1, Int>::from_ints(idx_host.as_slice(), &device);\
    let indices = indices_1d.reshape([batch, active_len]);

    let values = Tensor::<B, 2>::from_data(
        TensorData::new((0..active_len).map(|i| i as f32).collect::<Vec<f32>>(), [batch, active_len]),
        &device,
    );

    // Always works
    let y = x.clone().scatter(1, indices, values);

    let loss = y.sum();

    // Crashes (depending on backend)
    let grads = loss.backward();

    // Force e.g. Cuda to synchronize and actually trigger the crash
    println!("{}", x.grad(&grads).unwrap().sum().to_data());
}

#[test]
#[should_panic]
fn try_on_ndarray() {
    // Crashed (Invalid dimension: the shape of the index tensor should be the same as the value tensor: Index [1, 3] value [1, 10])
    scatter_notequal_forward_to_backward::<Autodiff<NdArray<f32>>>();
}

#[test]
#[should_panic]
fn try_on_candle() {
    // Crashed (called `Result::unwrap()` on an `Err` value: shape mismatch in scatter (indexes, src), lhs: [1, 3], rhs: [1, 10])
    scatter_notequal_forward_to_backward::<Autodiff<Candle<f32>>>();
}

#[test]
#[should_panic]
fn try_on_cuda() {
    // Crashed (called `Result::unwrap()` on an `Err` value: DriverError(CUDA_ERROR_ILLEGAL_ADDRESS, "an illegal memory access was encountered"))
    scatter_notequal_forward_to_backward::<Autodiff<Cuda<f32>>>();
}

// #[test]
// #[should_panic]
// fn try_on_cpu() {
//     // Crashes entire process (SIGSEGV: invalid memory reference)
//     scatter_notequal_forward_to_backward::<Autodiff<Cpu<f32>>>();
// }

#[test]
fn try_on_wgpu() {
    // OK
    scatter_notequal_forward_to_backward::<Autodiff<Wgpu<f32>>>();
}

#[test]
fn try_on_tch() {
    // OK
    scatter_notequal_forward_to_backward::<Autodiff<LibTorch<f32>>>();
}
```

khoek:
Fix the bug described in https://github.com/tracel-ai/burn/issues/4063; `ops::tensor`'s `Scatter.backward()` has a broken autodiff implementation which crashes on backprop when all three arguments are not all the same size, and even if it didn't (by padding with zeros or otherwise), it's mathematically completely wrong (likely the author confused burn's "add" vs "write" `scatter()` semantics?).

Running `cargo run-checks` passes on my laptop up until building the documentation, whence it exhausts all 32GB of RAM on my laptop.

Anyone have any idea how this got through the various test suites it seems like you run (PyTorch, ONNX, etc.)?

Fixes #4063

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -986,7 +986,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
         struct Scatter;
 
         impl<B: Backend> Backward<B, 2> for Scatter {
-            type State = (usize, IntTensor<B>, Shape, Shape, B::Device);
+            type State = (usize, IntTensor<B>);
 
             fn backward(
                 self,
@@ -994,21 +994,15 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                 grads: &mut Gradients,
                 _checkpointer: &mut Checkpointer,
             ) {
-                let (dim, indices, shape_lhs, shape_rhs, device) = ops.state;
-                let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));
+                let (dim, indices) = ops.state;
+                let [_, indices_4rhs] = duplicate(&ops.parents, Some(indices));
 
                 binary::<B, _, _>(
                     ops.parents,
                     ops.node,
                     grads,
-                    |grad| {
-                        let zeros = B::float_zeros(shape_lhs, &device, grad.dtype().into());
-                        B::float_scatter(dim, grad, indices_4lhs.unwrap(), zeros)
-                    },
-                    |grad| {
-                        let zeros = B::float_zeros(shape_rhs, &device, grad.dtype().into());
-                        B::float_scatter(dim, zeros, indices_4rhs.unwrap(), grad)
-                    },
+                    |grad| grad,
+                    |grad| B::float_gather(dim, grad, indices_4rhs.unwrap()),
                 );
             }
         }
@@ -1019,13 +1013,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
             .stateful()
         {
             OpsKind::Tracked(prep) => prep.finish(
-                (
-                    dim,
-                    indices.clone(),
-                    tensor.primitive.shape(),
-                    value.primitive.shape(),
-                    B::float_device(&value.primitive),
-                ),
+                (dim, indices.clone()),
                 B::float_scatter(dim, tensor.primitive, indices, value.primitive),
             ),
             OpsKind::UnTracked(prep) => prep.finish(B::float_scatter(
--- a/crates/burn-autodiff/src/tests/gather_scatter.rs
+++ b/crates/burn-autodiff/src/tests/gather_scatter.rs
@@ -65,4 +65,40 @@ mod tests {
             .to_data()
             .assert_eq(&TensorData::from([[19., 19., 19.], [64., 64., 64.]]), false);
     }
+
+    #[test]
+    fn test_scatter_grad_partial_indices() {
+        let device = Default::default();
+        let tensor_1 = TestAutodiffTensor::from_data(
+            TensorData::from([[0.0, 1.0, 2.0, 3.0, 4.0, 5.0]]),
+            &device,
+        )
+        .require_grad();
+        let tensor_2 = TestAutodiffTensor::from_data(
+            TensorData::from([[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]]),
+            &device,
+        )
+        .require_grad();
+        let values = TestAutodiffTensor::from_data(TensorData::from([[4.0, 5.0, 6.0]]), &device)
+            .require_grad();
+        let indices = Tensor::<TestAutodiffBackend, 2, Int>::from_data(
+            TensorData::from([[2, 1, 0]]),
+            &device,
+        );
+
+        let tensor_3 = tensor_1.clone().mul(tensor_2);
+        let tensor_4 = tensor_3.clone().scatter(1, indices, values.clone());
+
+        let grads = tensor_4.backward();
+
+        let grad_1 = tensor_1.grad(&grads).unwrap();
+        let grad_2 = values.grad(&grads).unwrap();
+
+        grad_1
+            .to_data()
+            .assert_eq(&TensorData::from([[1., 2., 3., 4., 5., 6.]]), false);
+        grad_2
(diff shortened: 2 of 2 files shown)


khoek:
In addition, `float_cat`, `float_unfold`, `float_cast`, and `float_cross` actually all also look like they have mistakes (I have only checked this file). I'll send another follow up pull request later today.

Sorry to belabor the point, but surely there is some big standard test suite we wouldn't have to write which covers this sort of stuff? (And between you and me I'm now a little spooked about other autodiff-related files, so running something like that would be reassuring---does anyone happen to know of anything like this?)

laggui (changes requested):
Hey thanks for digging into the autodiff implementations!

> it's mathematically completely wrong (likely the author confused burn's "add" vs "write" scatter() semantics?).

Actually, it's correct. And you pointed to the current definition of burn's scatter:

https://github.com/tracel-ai/burn/blob/66b51fc5ebb514c96c874ca25ab109cc9d504749/crates/burn-tensor/src/tensor/api/numeric.rs#L1065-L1073

See also the numeric ops table: https://burn.dev/books/burn/building-blocks/tensor.html#numeric-operations

It's a scatter add, though I don't necessarily agree it should remain that way because it is confusing :sweat_smile: 

This has also been bugging me for a while, I think now is as good of a time as any to fix that.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -994,21 +994,15 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
-                    |grad| {
-                        let zeros = B::float_zeros(shape_lhs, &device, grad.dtype().into());
-                        B::float_scatter(dim, grad, indices_4lhs.unwrap(), zeros)
-                    },
-                    |grad| {
-                        let zeros = B::float_zeros(shape_rhs, &device, grad.dtype().into());
-                        B::float_scatter(dim, zeros, indices_4rhs.unwrap(), grad)
-                    },
+                    |grad| grad,
+                    |grad| B::float_gather(dim, grad, indices_4rhs.unwrap()),
```
laggui, line 1005: We should add a unit test for this in `burn-autodiff/src/tests/gather_scatter.rs`:

```rust
#[test]
fn test_scatter_grad_partial_indices() {
    let tensor_1 = TestAutodiffTensor::from_data(
        TensorData::from([[0.0, 1.0, 2.0, 3.0, 4.0, 5.0]]),
        &device,
    )
    .require_grad();
    let tensor_2 = TestAutodiffTensor::from_data(
        TensorData::from([[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]]),
        &device,
    )
    .require_grad();
    let values = TestAutodiffTensor::from_data(TensorData::from([[4.0, 5.0, 6.0]]), &device)
        .require_grad();
    let indices = Tensor::<TestAutodiffBackend, 2, Int>::from_data(
        TensorData::from([[2, 1, 0]]),
        &device,
    );

    let tensor_3 = tensor_1.clone().mul(tensor_2);
    let tensor_4 = tensor_3.clone().scatter(1, indices, values.clone());

    let grads = tensor_4.backward();

    let grad_2 = values.grad(&grads).unwrap();

    grad_1
        .to_data()
        .assert_eq(&TensorData::from([[1., 2., 3., 4., 5., 6.]]), false);
    grad_2
        .to_data()
        .assert_eq(&TensorData::from([[1., 1., 1.]]), false);
}
```

and we can remove the `shape_lhs, shape_rhs, device` from the state instead of ignoring them

khoek:
Great! Hopefully I've made those changes as you wanted.

khoek:
Sorry I made some noise accidentally bumping main to run CI on my own branch. Good to go now.

PR #4067, opened 2025-11-25
re-export serde macros

crutcher:
Burn clients which use `#[derive(Serialize, Deserialize)]` on structs are forced to align a serde dep with their `burn` dep, solely because the `serde` macros are not re-exported.

This permits clients to work directly with the pinned serde dep for simple derive blocks.

### DIFF
--- a/crates/burn-core/src/lib.rs
+++ b/crates/burn-core/src/lib.rs
@@ -8,9 +8,12 @@
 #[macro_use]
 extern crate derive_new;
 
-/// Re-export serde for proc macros.
+/// Re-export serde and macros.
+/// - Supports burn proc macros,
+/// - Permits clients to define serializable structs with the pinned serde.
+#[allow(unused_imports)]
 #[macro_use]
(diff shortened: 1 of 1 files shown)


laggui (commented):
Have you tested the change? I don't think there is actually a way around users having to add `serde` as an explicit dependency to their project :thinking:

because the derive macro will always expand this code with `extern crate serde as _serde`
```rust
use burn::serde::{Deserialize, Serialize};

#[derive(Serialize, Deserialize)]
struct MyStruct {
    pub field: usize,
}
```

which will break nonetheless.

So the current re-export works as long as users select a matching version, e.g.
```toml
serde = "1" # Uses any 1.x version compatible with burn
```

But maybe we could add more docs at the module re-export for `pub use serde`

crutcher:
Yeah; this doesn't work, sorry.

laggui:
No worries! I was curious if you had somehow managed a way around that :smile:

PR #4069, opened 2025-11-26
More correctness fixes in autodiff ops

khoek:
On top of #4064; fix some more autodiff correctness problems in `float_cross()`, `float_select()`, `float_select_assign()`, and `float_cat()`. Notably:

- `float_cross()`: Second gradient's sign is flipped.
- `float_select()`: The old implementation fails for repeated indices because it overwrites instead of summing; so, use Burn's `scatter` (which has add semantics) instead.
- `float_select_assign()`: The previous implementation did not mask overwritten values; do that---we treat duplicates as undefined behaviour and don't e.g. implement "last writer wins" strictly.
- `float_cat()`: CatStep zips nodes (which are `Option<NodeRef>`) with `dim_sizes`, then does a `filter_map` to drop `None`s before advancing `current_index`. That causes gradient slices to be misaligned whenever some inputs don't require grad.

It also seems to me that `float_unfold()` and `float_cast()` detach their gradients, would it be reasonable for me to "fix" this too?

### DIFF
--- a/crates/burn-autodiff/src/tests/cast.rs
+++ b/crates/burn-autodiff/src/tests/cast.rs
@@ -1,8 +1,11 @@
 #[burn_tensor_testgen::testgen(ad_cast)]
 mod tests {
     use super::*;
+    use burn_tensor::might_panic;
     use burn_tensor::{DType, Tensor, TensorData};
 
+    #[cfg(feature = "std")]
+    #[might_panic(reason = "Unsupported precision for fusion")]
     #[test]
     fn cast_keeps_gradient_flow() {
         let device = Default::default();
@@ -12,8 +15,7 @@ mod tests {
         )
         .require_grad();
 
-        let dtype = x.dtype();
-        let y = x.clone().cast(dtype);
+        let y = x.clone().cast(DType::F64);
         let z = y.sum();
 
         let grads = z.backward();


crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1073,7 +1061,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
 
                 unary::<B, _>(ops.parents, ops.node, grads, |grad| {
                     let zeros = B::float_zeros(shape, &device, grad.dtype().into());
-                    B::float_select_assign(zeros, dim, indices, grad)
```
laggui, line 1064: Another not-so-explicit op here, the select_assign actually performs a sum reduction (like scatter which performs a scatter add).. :sweat_smile: so this one was already correct

https://github.com/tracel-ai/burn/blob/080ade2b82421fd16058a6cfb6e6d54ecb7f4b41/crates/burn-tensor/src/tensor/api/base.rs#L1602-L1621

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1127,22 +1115,27 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
                     ops.node,
                     grads,
-                    |grad| grad,
-                    |grad| B::float_select(grad, dim, indices),
+                    |grad| {
+                        let zeros =
+                            B::float_zeros(value_shape.clone(), &device, grad.dtype().into());
+                        B::float_select_assign(grad, dim, indices_4lhs.unwrap(), zeros)
+                    },
+                    |grad| B::float_select(grad, dim, indices_4rhs.unwrap()),
```
laggui: This would be for select assign (pure assignment, no addition) - but again the select assign actually performs an addition

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -2419,18 +2417,22 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                let ranges_template = grad.shape().iter().map(|&v| 0..v).collect();
 
                 self.nodes
                     .into_iter()
                     .zip(self.dim_sizes)
-                    .filter_map(|(node, dim_size)| node.map(|node| (node, dim_size)))
-                    .for_each(|(node, dim_size)| {
-                        let mut ranges = ranges.clone();
-                        ranges[self.dim] = current_index..dim_size + current_index;
-                        current_index += dim_size;
```
laggui, line 2421: Good catch!

laggui (changes requested):
The select and select assign are correct based on the add semantics. I am going to make changes to the interface today to make that more clear for assign -> assign add and scatter -> scatter add.

So only the cross and cat are erroneous :slightly_smiling_face: 

Do you mind adding a unit test to capture the concatenation behavior?

> It also seems to me that float_unfold() and float_cast() detach their gradients, would it be reasonable for me to "fix" this too?

Oh, yeah! That's not expected or desired. I think cast was initially just to move the float into int, that's no longer the case and the gradients should absolutely follow.

For unfold, that would require implementing a backward step that sums overlapping windows back. Feel free to take a crack at it! I thought the current implementation just didn't allow autodiff with unfold, which would be better than an incorrect implementation. Either of the two would be an improvement :sweat_smile: 

Thanks for validating our ops!

crates/burn-autodiff/src/ops/tensor.rs
khoek: Sorry for my confusion, I agree!

crates/burn-autodiff/src/ops/tensor.rs
khoek: Sorry for my confusion, I agree!

khoek:
I believe I've added float_unfold() and float_cast() gradients, and that test you've asked for. Thanks for taking a look!

crates/burn-autodiff/src/ops/tensor.rs
laggui: No worries! I've been meaning to fix this for a while and you just reminded me of that :smile: so I opened https://github.com/tracel-ai/burn/pull/4070

crates/burn-autodiff/src/tests/cast.rs
```diff
@@ -0,0 +1,26 @@
...
+    fn cast_keeps_gradient_flow() {
+        let device = Default::default();
+        let x = Tensor::<TestAutodiffBackend, 2>::from_data(
+            TensorData::from([[1.0, 2.0], [3.0, 4.0]]),
+            &device,
+        )
+        .require_grad();
+
+        let dtype = x.dtype();
+        let y = x.clone().cast(dtype);
```
laggui: The cast doesn't apply when it's the same dtype! So you should use a different dtype.

The caveat is that we don't offer a way to check supported dtypes yet :sweat_smile: casting to `DType::F64` should work for all backends except fusion, but we could use the internal `burn_tensor::might_panic` macro to mark this.

```rust
#[might_panic(reason = "Unsupported precision for fusion")]
```

I can take care of that though no worries

laggui (approved):
Thanks :pray:

PR #4070, opened 2025-11-26
[Breaking] Add `IndexingUpdateOp` to `scatter` and `select_assign`

laggui:
### Checklist

### Related Issues/PRs

#4069 

The current nomenclature across indexing operations with assignment is confusing.
- `slice_assign`: actually performs assignment
- `select_assign`: actually performs a select add
- `scatter`: actually performs a scatter add

### Changes

- [Breaking] Added `IndexingUpdateOp` to specify what kind of assignment/update is performed
  - Only addition is available right now, but we can easily support assign and mul eventually
- Backend tensor ops renamed to `float_select_add` and `float_scatter_add` based on enum match

### DIFF
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -2,7 +2,6 @@ use alloc::vec::Vec;
 
 use crate::alloc::borrow::ToOwned;
 
-use crate::TensorPrimitive;
 use crate::canonicalize_dim;
 use crate::{
     AsIndex, BasicOps, Bool, Distribution, Element, ElementConversion, Float, Int, Shape, Tensor,
--- a/crates/burn-ir/src/builder.rs
+++ b/crates/burn-ir/src/builder.rs
@@ -2,7 +2,7 @@
 
 use alloc::vec::Vec;
 use burn_tensor::{
-    DType, Distribution, Shape, Slice, IndexingUpdateOp, calculate_matmul_output,
+    DType, Distribution, IndexingUpdateOp, Shape, Slice, calculate_matmul_output,
     ops::{
         conv::{
             calculate_conv_output_shape, calculate_conv_transpose_output_shape,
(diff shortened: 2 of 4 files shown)


crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -1081,7 +1081,34 @@ where
     /// # Warning
     /// Not all backends have runtime bound checks for the indices, so make sure the they are valid.
     /// Otherwise, out of bounds indices could lead to unexpected results instead of panicking.
+    #[deprecated(
+        since = "0.20.0",
+        note = "In a future release, `scatter` will require an update mode.\nUse `scatter_add` instead for the current additive behavior."
+    )]
     pub fn scatter(self, dim: usize, indices: Tensor<B, D, Int>, values: Self) -> Self {
+        self.scatter_add(dim, indices, values)
+    }
```
laggui: See here, with deprecation notice that will eventually take the enum in the future

crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -3776,8 +3805,11 @@ impl<B: Backend> Numeric<B> for Int {
         tensor: Self::Primitive,
         indices: B::IntTensorPrimitive,
         values: Self::Primitive,
+        update: UpdateComputation,
     ) -> Self::Primitive {
-        B::int_scatter(dim, tensor, indices, values)
+        match update {
+            UpdateComputation::Add => B::int_scatter_add(dim, tensor, indices, values),
+        }
     }
```
laggui: In the future, the tensor API method will also take the enum (like this).

And since the implementations might differ, I figured I'd change the tensor ops trait method to dispatch. But I guess this can be left as `*_scatter` and the backends have to handle all variants instead.

crates/burn-tensor/src/tensor/api/mod.rs
```diff
@@ -26,3 +26,12 @@ pub use numeric::*;
...
 pub use transaction::*;
+
+/// Computation to be used to update the existing values in indexed assignment operations (scatter/select).
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub enum UpdateComputation {
+    // Assign,
+    /// Performs an addition.
+    Add,
+    // Mul
+}
```
laggui, line 35: The enum already exists, just a different name. Don't mind changing it either.

laggui:
Renamed the enum to `IndexingUpdateOp` (closer to your suggestion) and removed the additional methods to introduce a breaking change immediately as discussed.

PR #4075, opened 2025-11-26
Manually call autodiff cleanup

swfsql:
### Checklist

### Related Issues/PRs

- Re-opens https://github.com/tracel-ai/burn/pull/3977

### Changes

- Adds the `graph_cleanup` method to the `burn::tensor::backend::AutodiffBackend` trait.
- Also changed how the ordering of node removal is made, ensuring all leaf nodes are removed first.
  - This is not necessary since a given parent can only have a single child, but I think this could be a good little change.

### Testing

Example of usage:

```rust
use burn::prelude::*;

type NdArray = burn::backend::NdArray<f32, i32>;
type NdArrayAuto = burn::backend::Autodiff<NdArray>;

fn main() {
    train::<NdArrayAuto>();
}

pub fn train<AutoB: burn::tensor::backend::AutodiffBackend>() {
    let train_device = <<AutoB as Backend>::Device>::default();
    for _ in 0..10_000_000 {
        let a: Tensor<AutoB, 2> = Tensor::zeros([2, 2], &train_device);
        let a = a.require_grad(); // an autodiff node is created
        drop(a); // the tensor is dropped but its unusable graph persists
        AutoB::graph_cleanup(); // either here, after the tensor is dropped, or after the loop
    }
    // in here, without graph_cleanup, AutoB/train_device cannot be used to cleanup the "hanging nodes"
}
```

Then by accompanying the system's ram usage, the memory usage doesn't grow (in case it's in the loop). Also, the are no vector/hashmaps/hashsets re-allocations, possibly improving runtime by a little.

---

I recently rebased the `graph-sweep` branch to `main`, as I'm using it locally.  
In the current `main`, the "hot loop stress test" (increased to 10 million iterations) eventually still has the ram going up to around ~1GB, whereas `graph-sweep` leaves it limited at ~138kB. I haven't looked much into it why some nodes are missed on `main`, but I suspect it could be related to Arc counting (but I'm not sure).

Well, the reason why I created another PR is because all updates to the `graph-sweep` branch are ignored on the previous PR, because it is closed. And If the PR was already closed and if I rebased the branch (i.e. force-pushed the branch), I can no longer re-open it (-.-)', it appears to be a Github thing. So if it is ok, and contrary to what I did, I'd like to leave this PR open, even if it's open-ended.

### DIFF
--- a/crates/burn-autodiff/src/runtime/graph.rs
+++ b/crates/burn-autodiff/src/runtime/graph.rs
@@ -127,35 +127,6 @@ pub(crate) struct GraphCleaner<'a> {
 
 impl<'a> GraphCleaner<'a> {
     pub(crate) fn cleanup_orphaned_entries() {
-        // extra cleanup procedure
-        {
-            let graphs_to_visit = {
-                let graph_locator = crate::runtime::graph::STATE.lock();
-                let graph_locator = graph_locator.as_ref().unwrap();
-                let mut graphs_to_visit = HashMap::new();
-                for (_node_id, graph) in &graph_locator.graphs {
-                    graphs_to_visit
-                        .entry(graph.origin)
-                        .or_insert_with(|| Arc::clone(graph));
-                }
-                graphs_to_visit
-            };
-
-            let mut cleaner = crate::runtime::graph::GraphCleaner::init();
-            for (_graph_origin, graph) in graphs_to_visit {
-                let mut state = graph.state.lock();
-                let server = &mut state.server;
-                server
-                    .memory_management
-                    .free_unavailable_nodes(|node_id: &NodeId| {
-                        server.steps.remove(node_id);
-                        server.actions_builder.remove(node_id);
-                        cleaner.clean(node_id);
-                    });
-            }
-            drop(cleaner);
-        }
-
         let graphs = {
             // Get the available graphs and release the lock
             match STATE.lock().as_ref() {


swfsql:
To note, this PR probably has deadlocks on it, given the github action test runs.

vadixidav:
Let me know when it is stable again and I will battle test your branch. I am running training continuously and benefit a lot for my application from graph cleanup. I will say rc4 has been better with automated graph cleanup, but it isn't perfect. It helps a whole lot to call graph cleanup after I am done running all my backwards calls.

swfsql:
@vadixidav I still haven't investigated the deadlocks, but I often rebase it to `main` (those are the force-push updates). But I'll ping you once I do -- but plz note that I'm not planning to investigate it too soon as I'm not getting the deadlocks for my own use.

And also to note, this PR as it is, you don't need to manually run it (you can if you want), but it automatically runs on every backward as well.

Edit: @laggui Also to note, if I moved the "extra cleanup procedure" from this PR to _below_  the current cleanup procedure (not above as it currently is), then it fails (deadlocks aside) in cleaning up. So this is why I believe that this is related to Arc counting on the standard `main` cleanup. But then again, I haven't checked this too closely.

swfsql:
This linked PR (https://github.com/tracel-ai/burn/pull/4342) solves the need for the extra cleanup, and in current `main` I no longer see any leaks during the loop (considering the manual cleanup call).

So now this PR is mostly kept as a drafty way to directly call to the cleanup.

laggui:
> This linked PR (#4342) solves the need for the extra cleanup, and in current `main` I no longer see any leaks during the loop (considering the manual cleanup call).
> 
> So now this PR is mostly kept as a drafty way to directly call to the cleanup.

Yeah that PR fixed a little oversight during the refactor that missed re-using the existing `remove_entry` cleanup method :grimacing: It was a good catch.

I appreciate you following up on the changes! Given this report, I think we'll eventually close this draft and you won't have to keep it up to date :smile:

laggui:
Closing before the bot marks this as stale again :smile: 

Hopefully manual cleanup will not be required in the future. But please let us know if you observe any regressions that require this workaround.

Issue #4052, opened 2025-11-22
RF-DETR onnx fails to import


torsteingrindvik:
**Describe the bug**

[RF-DETR](https://roboflow.com/model/rf-detr) fails to import.

 
**To Reproduce**

Save below as e.g. `rfdetr-export.py` and `chmod +x rfdetr-export.py` then run `./rfdetr-export.py`.
Assuming you have [uv](https://docs.astral.sh/uv/) this should just work and you will have the onnx model exported.

```python
#!/usr/bin/env -S uv run --python 3.11 --script
# /// script
# python = "3.11"
# dependencies = [
#   "torch==2.6.*",
#   "torchvision==0.21.*",
#   "onnx",
#   "rfdetr[onnxexport]",
# ]
# ///

import torch
from rfdetr import RFDETRSmall

print("Loading RF-DETR...")
model = RFDETRSmall()

print("Exporting model to ONNX...")
onnx_path = model.export()
print(f"Export complete → {onnx_path}")
```

Then try to import in `burn-import`: `cargo run --release -- rfdetr/output/inference_model.onnx output`

**Error output**

```
PANIC => panicked at crates/burn-import/src/onnx/model_gen.rs:422:33:
Failed to parse ONNX file '/home/togr/Documents/rfdetr/output/inference_model.onnx': Type inference failed: InvalidAttribute { name: "axis", reason: "axis 2 is out of bounds for rank 1" }
```

**Expected behavior**

Import succeeds.

Ran on latest Burn main: 859ca5b639761ee85c69a04721c1f607fe3068d1

antimora:
This PR will fix it: https://github.com/tracel-ai/burn/pull/4087

antimora:
Still fixing one bug for wgpu/metal: https://github.com/tracel-ai/burn/issues/4090

antimora:
Closing via https://github.com/tracel-ai/burn/pull/4087

PR #4081, opened 2025-11-28
ONNX fixes for rfdetr model

antimora:
This PR has multiple fixes and will be broken down into multiple PRs. 

This fixes https://github.com/tracel-ai/burn/issues/4052


antimora:
I broke down the change into https://github.com/tracel-ai/burn/pull/4087

PR #4082, opened 2025-11-28
Fix Expand operation to use ONNX max-semantics

antimora:
Fixes the ONNX Expand operation to properly implement max-semantics as defined in the ONNX
specification.

ONNX Expand uses max-semantics where `output_dim = max(input_dim, shape_dim)`. This means when
`shape_dim == 1` but `input_dim > 1`, the output keeps the input dimension rather than broadcasting.

Previously, Burn's Expand codegen passed the shape directly, which caused incorrect behavior when
the shape contained 1s that should preserve the input dimension.

### Checklist

### Related Issues/PRs

Partially addresses https://github.com/tracel-ai/burn/issues/4052

### Changes

- Updated `crates/burn-import/src/burn/node/expand.rs` to transform ONNX shapes to Burn's convention
  (using -1 to mean "keep input dim")
- Added `expand_max_semantics` test case that verifies the fix

### Testing

- Added `expand_max_semantics.onnx` test with input shape `[1, 1, 4, 1]` and expand shape
      `[1, 1, 1, 5]` expecting output shape `[1, 1, 4, 5]`
- Existing expand tests continue to pass

Example

```
Input shape: [1, 1, 4, 1]
ONNX shape:  [1, 1, 1, 5]
Expected:    [1, 1, 4, 5]  (max of each dimension)
```

Without this fix, the output would incorrectly be `[1, 1, 1, 5]`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -629,7 +629,7 @@ dependencies = [
  "burn-tensor",
  "burn-tensor-testgen",
  "derive-new",
- "hashbrown 0.15.5",
+ "hashbrown 0.16.1",
  "log",
  "num-traits",
  "parking_lot",
@@ -748,7 +748,7 @@ dependencies = [
  "derive-new",
  "flate2",
  "half",
- "hashbrown 0.15.5",
+ "hashbrown 0.16.1",
  "log",
  "num-traits",
  "portable-atomic",
@@ -797,7 +797,7 @@ dependencies = [
  "derive-new",
  "futures-lite",
  "half",
- "hashbrown 0.15.5",
+ "hashbrown 0.16.1",
  "log",
  "num-traits",
  "paste",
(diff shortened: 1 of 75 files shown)


crates/burn-import/src/burn/node/expand.rs
```diff
@@ -41,8 +39,24 @@ impl<PS: PrecisionSettings> NodeCodegen<PS> for onnx_ir::expand::ExpandNode {
...
+                let mut burn_shape = onnx_shape;
+                #[allow(clippy::needless_range_loop)]
+                for i in 0..#input_rank {
+                    let dim_offset = #output_rank - #input_rank + i;
+                    if burn_shape[dim_offset] == 1 && input_dims[i] > 1 {
+                        burn_shape[dim_offset] = -1;
+                    }
+                }
+                #input.expand(burn_shape)
+            };
```
laggui, line 58: If we're going to loop over the dims, why not just set the shape as the input dim when it's the case?

The shape would already fixed when passed to `input.expand(shape)`, isolating the ONNX semantics.

antimora: Agree. Cleaner and more explicit.

PR #4083, opened 2025-11-28
Fix Slice operation to handle empty ranges

antimora:
Fixes the Slice operation to correctly handle empty ranges (where `start >= end`) by returning a tensor with size 0 in that dimension instead of panicking. 

This is valid behavior in ONNX and PyTorch - slicing with an empty range should produce an empty tensor, not an error.

### Checklist

### Related Issues/PRs

Partially addresses https://github.com/tracel-ai/burn/issues/4052

### Changes

- Empty slices (where `start >= end`) are now handled at the high-level tensor API, returning empty tensors when any output dimension is zero
- Backend implementations no longer need to check for empty slices - related early returns have been removed
- Documentation updated to clarify this behavior for bool, int, and float tensor slicing traits

### Testing

- Added `slice_empty` ONNX test that slices `[1, 2, 3, 4, 5]` with range `[2:2]` expecting empty tensor `[]`
- Added unit test `should_support_empty_slice`
- Existing slice tests continue to pass

### Example

```python
import torch
x = torch.tensor([1, 2, 3, 4, 5])
result = x[2:2]  # Returns tensor([], dtype=torch.int64) with shape [0]
```


antimora:
> Instead of adapting all backend implementations I would probably change the high-level tensor API method to return `Tensor::empty` with the 0-size dim.
> 
> ```rust
> pub fn slice<const D2: usize, S>(self, slices: S) -> Self
> where
>     S: SliceArg<D2>,
> {
>     let shape = self.shape();
>     let slices = slices.into_slices(shape.clone());
> 
>     // Validate slices
>     check!(TensorCheck::slice::<D, D2>(&shape, &slices));
> 
>     // TODO: just return an empty tensor
>     if slices.iter().zip(shape.iter()).any(|(s, &d)| s.output_size(d) == 0)
>     {
>         return Self::empty(shape, &self.device());
>     }
> 
>     // Use the slice method that supports steps
>     Self::new(K::slice(self.primitive, &slices))
> }
> ```

I can do it do but I thought the underlying backend implementation will be wrong if we decide to slice primitive in other ops. 

Do you still want to do it at the higher level?

@nathanielsimard, what's your opinion on this? Probably this has some significance at the CubeCL level too.

laggui:
> I can do it do but I thought the underlying backend implementation will be wrong if we decide to slice primitive in other ops.

Empty slices are trivial, they always produce an empty tensor. I think all backends should stick to implementing the “real” slicing logic.

We could always add a note to the `{float, int, bool}_slice` doc comments to specify that backends are not required to handle empty slices themselves.

PR #4089, opened 2025-11-28
Flash Attention module

louisfd:
Expose the flash attention kernel from CubeCL into Burn. Kernel selection is still very rudimentary so don't expect peak performance yet.

### DIFF
--- a/crates/burn-tensor/src/tensor/ops/modules/attention.rs
+++ b/crates/burn-tensor/src/tensor/ops/modules/attention.rs
@@ -1,4 +1,6 @@
 use core::f32;
+#[allow(unused_imports)]
+use num_traits::Float as _;
 
 use crate::{
     ElementConversion, TensorMetadata,


crates/burn-cubecl/src/ops/module_ops.rs
```diff
@@ -195,4 +195,14 @@ where
...
         kernel::interpolate::interpolate_backward(x, grad, output_size, options)
     }
+
+    fn attention(
+        query: FloatTensor<Self>,
+        key: FloatTensor<Self>,
+        value: FloatTensor<Self>,
+        mask: Option<BoolTensor<Self>>,
+    ) -> FloatTensor<Self> {
+        kernel::attention::flash_attention(query, key, value, mask, F::dtype())
```
laggui: `out_dtype` should probably match the query / key / value dtype instead of being the default floating point dtype

crates/burn-cubecl/src/kernel/attention/base.rs
```diff
@@ -0,0 +1,53 @@
...
+            query_global: query.dtype.into(),
+            query_tile: query.dtype.into(),
+            key_global: key.dtype.into(),
+            key_stage: key.dtype.into(),
+            value_global: value.dtype.into(),
+            value_stage: value.dtype.into(),
+            key_value_tile: value.dtype.into(),
+            softmax: query.dtype.into(),
+            accumulator: out_dtype.into(),
+            mask: DType::U8.into(),
```
laggui: Should this match the `mask` dtype? e.g.

```rust
let mask_dtype = mask.as_ref().map(|m| m.dtype).unwrap_or(DType::U8);
```

laggui (commented):
Integration LGTM overall!

Just some minor comments (excluding lint fixes required by CI).

We could eventually revamp `burn-nn`'s `MultiHeadAttention` to leverage this.

antimora:
> Integration LGTM overall!
> 
> Just some minor comments (excluding lint fixes required by CI).
> 
> We could eventually revamp `burn-nn`'s `MultiHeadAttention` to leverage this.

I have filed an issue (https://github.com/tracel-ai/burn/pull/4089) to update burn-import to use Attention module so we can take advantage.

crates/burn-tensor/src/tensor/ops/modules/base.rs
```diff
@@ -848,6 +849,32 @@ pub trait ModuleOps<B: Backend> {
...
+    /// - `mask`: Optional boolean mask of shape `[batch_size, num_heads, seq_len_q, seq_len_k]`,
+    ///   here `true` indicates positions to mask (i.e. set to -∞ before softmax).
+    ///
+    /// # Returns
+    /// A tensor of shape `[batch_size, num_heads, seq_len_q, val_dim]`
+    /// representing the attended context per head.
+    ///
+    /// # Note
+    /// This implementation does not support dropout and is intended for inference or
+    /// use cases where dropout is not needed.
```
laggui: We'll have to add the autodiff implementation to expand usage to training eventually. Should help speed things up.

laggui, line 869: We should add the autodiff implementation to expand usage to training eventually. Should help speed things up over the naive attention.

And probably expand to fusion. Right now only cubecl backends without fusion will call the flash attention kernel.

PR #4094, opened 2025-11-29
Remove RefCell from onnx-ir Arguments

antimora:
This PR removes RefCell from Argument in onnx-ir, simplifying the API and improving safety. The key insight is that Arguments only read tensor data - they never mutate the underlying storage. Mutations happen at the pipeline phase level via GraphState, not through Arguments.

### Checklist

### Related Issues/PRs

- https://github.com/tracel-ai/burn/issues/3596
- https://github.com/tracel-ai/burn/pull/3872

### Changes

Architecture Changes:
- Move weight transpose to burn-import layer 
- TensorStore is now wrapped in Rc<TensorStore> (append-only during construction)
- constant_map is now Rc<HashMap<String, DataId>>
- Removed ValueStoreRef enum - Arguments now use Option<ValueStore> directly
- ValueStore is an immutable snapshot combining Rc<TensorStore> + Rc<HashMap<String, DataId>>                               |

Performance:
- Rc::make_mut only used during construction (ref count = 1)
- Post-processing preserves Rc references instead of clone-restore cycle
- No expensive tensor data copying

### Testing

All 352 onnx-tests pass, plus all unit tests in onnx-ir and burn-import.

### DIFF
--- a/crates/onnx-ir/src/ir/graph.rs
+++ b/crates/onnx-ir/src/ir/graph.rs
@@ -97,12 +97,12 @@ fn finalize_arguments_in_nodes(nodes: &mut [Node], value_store: &ValueStore) {
         }
 
         // Recursively process subgraphs
-        finalize_subgraphs_in_node(node, value_store);
+        finalize_subgraphs_in_node(node);
     }
 }
 
 /// Recursively finalize subgraphs within a node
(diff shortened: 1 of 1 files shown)


laggui:
(waiting to merge since pre-release task is on the way)

Issue #4090, opened 2025-11-28
Tensor::cat panics with "assertion failed: divisor != 0" when concatenating tensors with size 0 along concat dimension


antimora:
Location: cubecl-std/src/fast_math.rs:33

Root Cause

When Tensor::cat is called with a tensor that has size 0 along the concatenation dimension, the underlying cat_with_slice_assign implementation in
burn-tensor/src/tensor/ops/modules/cat.rs calls slice_assign with a zero-length slice.

The call chain:
1. Tensor::cat([tensor1, tensor2], dim) where tensor2.shape()[dim] == 0
2. cat_with_slice_assign computes tensor_dim_length = tensor.shape().dims[dim] → 0
3. slice_assign is called with a slice range like 5..5 (start == end)
4. In burn-cubecl/src/kernel/index/slice_assign.rs:163:
let length = (end - slice.start) as usize;  // length = 0
shape.push(FastDivmodArgs::new(&client, length as u32));  // divisor = 0
5. FastDivmodArgs::new at cubecl-std/src/fast_math.rs:33 has:
debug_assert!(divisor != 0);  // PANICS!

Minimal Reproduction

```rust
use burn::prelude::*;

type MyBackend = burn::backend::Metal;  // Also fails with Wgpu

fn main() {
    let device = Default::default();

    // Create tensors where one has size 0 on concat dimension
    let tensor1: Tensor<MyBackend, 2> = Tensor::zeros([2, 3], &device);
    let tensor2: Tensor<MyBackend, 2> = Tensor::zeros([2, 0], &device);

    // Panics with "assertion failed: divisor != 0"
    let result = Tensor::cat(vec![tensor1, tensor2], 1);
}
```

Context

This bug was discovered while running the RF-DETR model imported via burn-import. The model contains ONNX Slice operations that can produce empty tensors (e.g.,
tensor[n..MAX] where n equals the dimension size), which are then concatenated.

Suggested Fix

Two possible approaches:

1. In cat_with_slice_assign: Skip tensors with size 0 along the concat dimension (they contribute nothing to the output):
```rust
for tensor in tensors {
    let tensor_dim_length = tensor.shape().dims[dim];
    if tensor_dim_length == 0 {
        continue;  // Skip empty tensors
    }
    // ... rest of the loop
}

```

2. In slice_assign: Handle zero-length slices as a no-op (they don't modify anything):
// Early return if any dimension has length 0

```rust
if indices.iter().any(|slice| {
[...]

Issue #4093, opened 2025-11-28
CubeCL Operations Panic on Empty Tensors (Division by Zero)


antimora:
Multiple tensor operations in `burn-cubecl` panic with `assertion failed: divisor != 0` when
processing tensors with 0-sized dimensions. This occurs because CubeCL's `FastDivmod` optimization
precomputes division parameters on the host, which fails when the divisor is 0.

Empty tensors (tensors with at least one dimension of size 0) are valid mathematical objects that
arise naturally in dynamic models, but several GPU kernels don't handle them gracefully.

**Error message:**

```
thread 'main' panicked at cubecl-std/src/fast_math.rs:33:9:
assertion failed: divisor != 0
```

###  CRITICAL - Will Crash on Empty Tensors

| Operation                      | File                              | Line(s)    | Vulnerable Code                    |
| ------------------------------ | --------------------------------- | ---------- | ---------------------------------- |
| `adaptive_avg_pool2d`          | `adaptive_avg_pool2d.rs`          | 67, 73     | `(idx * input_size) / output_size` |
| `adaptive_avg_pool2d`          | `adaptive_avg_pool2d.rs`          | 62         | `sum / (num_ih * num_iw)`          |
| `adaptive_avg_pool2d_backward` | `adaptive_avg_pool2d_backward.rs` | 61, 72, 78 | Same division patterns             |
| `nearest_backward`             | `nearest_backward.rs`             | 61, 69     | `numerator / input_size`           |

###  HIGH - Division by Dimension/Config

| Operation                 | File                         | Line(s)           | Vulnerable Code                                 |
| ------------------------- | ---------------------------- | ----------------- | ----------------------------------------------- |
| `avg_pool2d`              | `avg_pool2d.rs`              | 72                | `sum / count` (count can be 0)                  |
| `avg_pool2d_backward`     | `avg_pool2d_backward.rs`     | 35, 84, 89        | `shape / line_size`, `grad / count`             |
| `pool2d` (direct)         | `pool2d.rs`                  | 63, 72-73         | `channels / line_size`, `pos / channel_lines`   |
| `adaptive_avg_pool2d`     | `adaptive_avg_pool2d.rs`     | 21-31             | `channels / line_size`, `pos % channel_lines`   |
| `conv_transpose2d`        | `transpose_direct.rs`        | 37, 47, 60        | `weight.shape / groups`, `oc / out_c_per_group` |
[...]

PR #4095, opened 2025-11-29
Handle empty tensors in cat and slice_assign ops

antimora:
Updated tensor API to treat empty tensors as no-ops in slice_assign and to filter out tensors with size 0 along the concatenation dimension in cat. Added tests to verify correct behavior when handling empty tensors and empty slice assignments. This prevents backend errors and ensures consistent results for edge cases.

### Checklist

### Related Issues/PRs

- Fixes https://github.com/tracel-ai/burn/issues/4090
- https://github.com/tracel-ai/burn/pull/4083
- https://github.com/tracel-ai/burn/issues/4052
- https://github.com/tracel-ai/burn/issues/4093

### Changes

- Empty slice assignments (where start == end or any dimension has size 0) now return the original tensor unchanged as a no-op
- The `cat` operation now filters out tensors with size 0 along the concatenation dimension before calling the backend
- Removed validation that prevented empty slice assignments, with handling moved to the API level

### Testing

Added new test cases for the change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1966,7 +1966,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.9.0-pre.3"
-source = "git+https://github.com/tracel-ai/cubecl?rev=a15c1d2735490b65d26dc3374c9810d090d9c4de#a15c1d2735490b65d26dc3374c9810d090d9c4de"
+source = "git+https://github.com/tracel-ai/cubecl?rev=86b23d0f6e73d6e1ea3f7ba97eaf876402da9b94#86b23d0f6e73d6e1ea3f7ba97eaf876402da9b94"
 dependencies = [
  "cubecl-attention",
  "cubecl-convolution",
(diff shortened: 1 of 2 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1884,8 +1895,29 @@ where
...
+        // Empty tensors don't contribute to the output and would cause issues
+        // in backend implementations (e.g., division by zero in slice_assign).
+        let first_tensor = tensors.first().expect("Tensors should not be empty");
+        let device = first_tensor.device();
+        let mut shape = first_tensor.shape();
+
+        let non_empty_tensors: Vec<_> = tensors
+            .into_iter()
+            .filter(|t| t.shape().dims[dim] > 0)
+            .collect();
```
laggui: While we're iterating over the tensors to filter them, might as well map them to the primitives.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1884,8 +1895,29 @@ where
...
+            shape.dims[dim] = 0;
+            return Self::empty(shape, &device);
+        }
+
         Self::new(K::cat(
-            tensors.into_iter().map(|vector| vector.primitive).collect(),
+            non_empty_tensors
+                .into_iter()
+                .map(|vector| vector.primitive)
+                .collect(),
```
laggui: Which will remove this entirely

PR #4097, opened 2025-11-30
Add memory-mapped ONNX loading with lazy tensor data

antimora:
Adds memory-mapped file loading for ONNX models to reduce memory usage during parsing. Raw tensor bytes are stored lazily and only copied when accessed, avoiding unnecessary allocations for unused tensors. The public API is refactored to use a builder pattern (`OnnxGraphBuilder`) for better extensibility.

NOTE: ONNX protobuf does not allow you to reference arbitrary data regions. The entire file have to be parsed. But with mmap + protobuf codegen, it's possible to optimize.

### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/3596

### Changes

**Memory-mapped loading**
- Files are memory-mapped via `memmap2` instead of read into heap memory
- `bytes::Bytes::from_owner(mmap)` keeps the mmap alive as long as tensor data references it
- Enabled by default via `mmap` feature flag

**Lazy tensor storage**
- New `LazyTensorData` type stores raw bytes + shape + dtype
- Conversion to `TensorData` deferred until `value()` is called
- Alignment-safe: copies to aligned heap memory on access

**API refactor**
- New: `OnnxGraphBuilder::new().parse_file("model.onnx")?`
- New: `OnnxGraphBuilder::new().parse_bytes(&bytes)?`
- New: `OnnxGraphBuilder::new().parse_reader(reader)?`
- Renamed: `OnnxIrError` → `Error`
- Removed: `parse_onnx()`, `parse_onnx_bytes()`, `parse_onnx_from_bytes()`

**Feature flags**
- `onnx-ir`: `mmap` (default)
- `burn-import`: `onnx-mmap` (default)

### Testing
- All existing  tests pass 
- Build works with and without mmap feature

### Benchmarks 

Key Results for CLIP Vision (336 MB)

| Metric                 | mmap ENABLED | mmap DISABLED | Improvement   |
|------------------------|--------------|---------------|---------------|
| parse_file max alloc   | 59.94 MB     | 411.7 MB      | 85% reduction |
| parse_file total alloc | 1.46 GB      | 1.812 GB      | 19% reduction |
| parse_file time        | ~76 ms       | ~98 ms        | 22% faster    |

The fix dramatically improved mmap effectiveness:
- Peak memory dropped from 411 MB to 60 MB (the 336 MB of tensor data now stays in the mmap region)
- Without mmap, parse_file uses 411.7 MB because it must copy all tensor data to heap
- With mmap, only ~60 MB of protobuf parsing overhead is allocated

### DIFF
--- a/crates/onnx-ir/src/tensor_store.rs
+++ b/crates/onnx-ir/src/tensor_store.rs
@@ -1,8 +1,8 @@
 //! Central tensor data storage with unique IDs for constants and initializers
 //!
-//! This module provides lazy tensor data storage that enables zero-copy parsing
-//! from memory-mapped ONNX files. Tensor data is stored as raw bytes and only
-//! converted to `TensorData` when accessed.
+//! This module provides tensor data storage that enables zero-copy parsing
+//! from memory-mapped ONNX files. Tensor data is stored as raw bytes references
+//! and only converted to `TensorData` when accessed.
 
 use std::collections::HashMap;
 use std::rc::Rc;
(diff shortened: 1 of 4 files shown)


crates/onnx-ir/src/tensor_store.rs
```diff
@@ -1,18 +1,87 @@
...
+/// the tensor data is actually accessed.
+#[derive(Debug, Clone)]
+pub struct LazyTensorData {
+    /// Raw bytes from protobuf (zero-copy slice of mmap'd buffer)
+    raw_bytes: bytes::Bytes,
+    /// Shape of the tensor
+    shape: Vec<usize>,
+    /// Data type of elements
+    dtype: DType,
+}
```
laggui: Right, the `raw_data` field already holds a cheaply cloneable and sliceable `bytes::Bytes`.

I think a better name would be `TensorDataRef` instead.

laggui (commented):
Nice! I think we should change the `LazyTensorData` name (see comment).

Otherwise LGTM

Issue #4096, opened 2025-11-30
[Feature] Implement Cross Attention and Windowed Attention


huy209vn:
This issue tracks support for two attention variants that are currently missing from Burn’s core modules:

Cross Attention (separate Query vs Key/Value)

Windowed/Local Attention (sliding window over the sequence)

Both are standard components in modern models, especially encoder–decoder architectures and long-sequence models.

Motivation

Cross Attention:
Required for any encoder–decoder architecture (e.g., T5, Whisper) and for conditioning mechanisms in generative models.

Windowed Attention:
Allows efficient long-sequence processing with complexity reduced from O(N²) to O(N × w).
Used in models like Mistral, long-form audio transformers, and streaming inference.

Anri-Lombard:
This is closed, right?

huy209vn:
i think windowed attention hasn't been implemented yet so no

Anri-Lombard:
Are you planning to implement it?

huy209vn:
Yeah i might..later
I have some other work right now.

antimora:
Cross attention landed in #4101, so I'm retitling this to track windowed/local attention only.

PR #4101, opened 2025-12-01
Add CrossAttention module to burn-nn

huy209vn:
part of #4096

This adds a standalone CrossAttention module supporting:

asymmetric query/context dimensions

configurable head layouts (n_heads_kv for MHA/MQA/GQA)

quantization-safe masking

optional quiet softmax

### DIFF
--- a/crates/burn-nn/src/modules/attention/cross_attention.rs
+++ b/crates/burn-nn/src/modules/attention/cross_attention.rs
@@ -22,6 +22,10 @@ use burn::{
     },
 };
 
+#[cfg(not(feature = "std"))]
+#[allow(unused_imports)]
+use num_traits::Float as _;
+
 #[derive(Config, Debug)]
 /// Configuration to create a [CrossAttention](CrossAttention) layer using the [init function](CrossAttentionConfig::init).
 pub struct CrossAttentionConfig {


antimora:
I have filed an issue (https://github.com/tracel-ai/burn/pull/4089) to update burn-import to use Attention module so we can take advantage. 

Currently burn's nn::attention lacks some of the features of ONNX

crates/burn-nn/src/modules/attention/cross_attention.rs
```diff
@@ -0,0 +1,207 @@
+//! Cross-Attention Module for Burn (Production Grade)
+//!
+//! Features:
+//! - Asymmetric Input Shapes (Query vs Context)
+//! - Grouped Query Attention (GQA) & Multi-Query Attention (MQA) support
+//! - Quantization-Safe Masking (min_float)
+//! - Sparse-Ready (quiet_softmax)
+//! - KV Caching for Streaming Inference
```
laggui, line 8: I don't see any key-value cache used?

Would be very useful though.

crates/burn-nn/src/modules/attention/cross_attention.rs
```diff
@@ -0,0 +1,207 @@
...
+
+use crate::modules::{Linear, LinearConfig};
+use crate::{Dropout, DropoutConfig};
+use burn_core as burn;
+use num_traits::Float;
```
laggui: Looks like this is unused

crates/burn-nn/src/modules/attention/cross_attention.rs
```diff
@@ -0,0 +1,207 @@
+//! Cross-Attention Module for Burn (Production Grade)
```
laggui: Minor suggestion, I think we can remove the last bit
```suggestion
//! Cross-Attention Module for Burn
```

huy209vn:
i added KV cache, tests and apply your suggestions.

huy209vn:
Should i wire flashattention into this also?

antimora:
> Should i wire flashattention into this also?

I'd suggest another PR since this is ready to be merged.

PR #4112, opened 2025-12-02
Add ceil_mode support to pooling operations (MaxPool, AvgPool)

### LINKED ISSUE #2132: ONNX import issue: `ceil_mode is not supported` in `avg_pool1d_config()`

antimora:
This PR adds comprehensive ceil_mode support to pooling operations (MaxPool1d/2d and AvgPool1d/2d) across the entire Burn framework, enabling ceiling-based output size calculation that produces larger output dimensions when input doesn't divide evenly by stride.

### Checklist

### Related Issues/PRs

Fixes #2132

### Changes

- Adds ceil_mode parameter to max_pool1d, max_pool2d, avg_pool1d, and avg_pool2d operations
- Enables ceiling-based output size calculation for pooling layers, producing larger output dimensions when input doesn't divide evenly
- Implements end-to-end support across all backends (ndarray, tch, candle, cubecl, fusion, router)

### Testing

- Unit tests for pooling modules with ceil_mode enabled
- Backward pass tests for autograd with ceil_mode
- ONNX import tests for maxpool1d_ceil_mode, maxpool2d_ceil_mode, avg_pool1d_ceil_mode, avg_pool2d_ceil_mode
- Insta snapshot tests for code generation field initialization

### DIFF
--- a/crates/burn-cubecl/src/kernel/pool/avg_pool2d.rs
+++ b/crates/burn-cubecl/src/kernel/pool/avg_pool2d.rs
@@ -21,9 +21,11 @@ impl Pool2dDirectStrategyFamily for AvgPoolStrategy {
 
 #[derive(CubeType, Debug, PartialEq, Eq, Hash, Clone, Copy)]
 pub struct AvgPoolStrategyConfig {
-    kernel_size_h: u32,
-    kernel_size_w: u32,
     count_include_pad: bool,
+    /// Total padded height (input_height + 2 * padding_0)
+    padded_h: u32,
+    /// Total padded width (input_width + 2 * padding_1)
+    padded_w: u32,
 }
 
 #[cube]
@@ -60,16 +62,17 @@ impl<N: Numeric> Pool2dDirectStrategy<N> for AvgPoolStrategy {
         *sum += result;
     }
 
-    fn set_padded_count(
+    fn count_position(
         #[comptime] config: &Self::Config,
         accumulator: &mut Self::Accumulator,
-        padded_count: u32,
+        ih: u32,
+        iw: u32,
     ) {
-        // When count_include_pad=true, use the dynamically computed padded count
-        // This correctly handles ceil_mode by only counting positions within padded bounds
-        if comptime![config.count_include_pad] {
+        // When count_include_pad=true, count positions within padded bounds
+        // (excludes ceil_mode extensions beyond the padded input)
+        if comptime![config.count_include_pad] && ih < config.padded_h && iw < config.padded_w {
             let (_sum, count) = accumulator;
-            *count = padded_count;
+            *count += 1;
         }
     }
 
@@ -141,14 +144,12 @@ pub(crate) fn avg_pool2d<R: CubeRuntime>(
             ScalarArg::new(dilation as u32),
             ScalarArg::new(padding[0] as u32),
             ScalarArg::new(padding[1] as u32),
-            ScalarArg::new(padded_0 as u32),
-            ScalarArg::new(padded_1 as u32),
         ),
         (kernel_size[0] as u32, kernel_size[1] as u32),
         AvgPoolStrategyConfig {
-            kernel_size_h: kernel_size[0] as u32,
-            kernel_size_w: kernel_size[1] as u32,
             count_include_pad,
+            padded_h: padded_0 as u32,
+            padded_w: padded_1 as u32,
         },
         output.dtype.into(),
     )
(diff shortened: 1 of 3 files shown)


crates/burn-ndarray/src/ops/maxpool.rs
```diff
@@ -113,19 +166,24 @@ pub(crate) fn max_pool2d_with_indices<E: FloatNdArrayElement, I: IntNdArrayEleme
...
+                                // Calculate index in original (unpadded) input
+                                let ih_orig = ih as i64 - (total_padding[0]) as i64;
+                                let iw_orig = iw as i64 - (total_padding[1]) as i64;
+
+                                // Clamp to valid range for index calculation
+                                let ih_clamped = ih_orig.max(0).min(x_height as i64 - 1);
+                                let iw_clamped = iw_orig.max(0).min(x_width as i64 - 1);
 
-                                index = ih * x_height as i64 + iw;
+                                index = ih_clamped * x_width as i64 + iw_clamped;
```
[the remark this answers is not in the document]
antimora: Verified with tests; the implementation is correct.

crates/burn-cubecl/src/kernel/pool/avg_pool2d.rs
```diff
@@ -79,14 +79,27 @@ pub(crate) fn avg_pool2d<R: CubeRuntime>(
...
+        ceil_mode,
+    );
+    let size_1 = calculate_pool_output_size(
+        kernel_size[1],
+        stride[1],
+        padding[1],
+        dilation,
+        x.shape[3],
+        ceil_mode,
+    );
```
laggui, line 117: For max pool just changing the output size should be enough, but I believe average pool with `ceil_mode` + `count_include_pad` might affect the results because valid pooling windows might not be the complete size.

To be confirmed.

antimora: Your concern is valid. When count_include_pad=true AND ceil_mode=true, the current implementation produces incorrect results.

For input 6x6, kernel 3×3, stride 2, padding 1, ceil_mode=true:
- Output position (3,3) window extends beyond even the user-padded region
- Valid + user-padding positions in window: 4 elements
- PyTorch: 35 / 4 = 8.75
- Current impl: 35 / 9 = 3.89

Submitting a fix.

crates/burn-cubecl/src/kernel/pool/pool2d.rs
```diff
@@ -84,14 +97,26 @@ pub fn pool2d_direct<E: Numeric, S: Pool2dDirectStrategyFamily>(
...
 
         for kw in 0..kernel_size.1 {
             let iw = ow * args.strides_1 + kw * args.dilation_1;
             let within_padding_w = iw >= args.padding_1 && iw < border_right;
+            let within_padded_w = iw < args.padded_1;
+
+            // Count positions within padded bounds
+            if within_padded_h && within_padded_w {
+                padded_count += 1;
+            }
```
laggui: I think this should be the responsibility of the pooling implementation, especially since it is only relevant for `count_include_pad` in avg pooling.

Instead of tracking an internal count in the generic pool2d implementation, and calling `set_padded_count` before `store`, I would have:

```suggestion
            // Count positions within padded bounds
            S::Pool2d::<E>::count_position(config, &mut accumulator, ih, iw);
```

So the args are no longer part of the generic pool2d kernel, but rather the avg pool2d config. And only checked for that.

```rust
#[derive(CubeType, Debug, PartialEq, Eq, Hash, Clone, Copy)]
pub struct AvgPoolStrategyConfig {
    kernel_size_h: u32,
    kernel_size_w: u32,
    count_include_pad: bool,
    /// Total padded height (input_height + 2 * padding_0)
    padded_0: u32,
    /// Total padded width (input_width + 2 * padding_1)
    padded_1: u32,
}

impl<N: Numeric> Pool2dDirectStrategy<N> for AvgPoolStrategy {
    // ...
    // feel free to use another function name that you find more relevant
    fn count_position(
        #[comptime] config: &Self::Config,
        accumulator: &mut Self::Accumulator,
        ih: u32,
        iw: u32,
    ) {
        if comptime![config.count_include_pad] {
            if ih < config.padded_0 && iw < config.padded_1 {
                let (_sum, count) = accumulator;
                *count += 1;
            }
        }
    }
}
```

laggui (approved):

PR #4119, opened 2025-12-03
Fix outer-scope variable references in ONNX subgraphs (If/Loop/Scan)

antimora:
**Problem**

ONNX control flow nodes (If, Loop, Scan) contain subgraphs that can reference values from the parent graph's scope. For example:

```
x = Conv(input, weights)      # x has shape [1, 64, 32, 32]
y = If(condition) {
    then_branch: Add(x, bias)  # References 'x' from parent!
    else_branch: Mul(x, scale)
}
```

The subgraph's Add(x, bias) needs to know the type of x for type inference, but x's type is only known after processing the parent graph's Conv
node.

Previously, when nested subgraphs were built during node conversion, the parent graph hadn't run type inference yet, causing errors like:
- Cannot squeeze 1 axes from a rank 0 tensor
- Type inference failures in nested subgraphs

**Solution**: Deferred/Lazy Building Pattern

Instead of building subgraphs immediately when parsing the ONNX proto, we now:

1. Parse phase: Store the raw GraphProto in a DeferredGraph
2. Type inference phase: When processing If/Loop/Scan nodes, all outer-scope types are now known
3. Build phase: Call deferred.build_graph_with_outer_scope(outer_scope_types) to build the subgraph with correct type information

### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/1941

### Changes
1. Parent state chain: GraphState now has an optional parent reference for recursive outer-scope lookups
2. Deferred subgraph building: Subgraphs are built lazily when outer-scope types are known
3. Filter Loop/Scan body inputs: When extracting outer-scope references, loop body's declared inputs are filtered out (they're loop-provided,
not outer-scope references)
4. Proper name matching: Filtering uses actual body input names to prevent duplicate bindings
5. Update model checks for silero-vad

### Testing

Tests added (6 new test cases):
| Model                 | Description                                                    |
|-----------------------|----------------------------------------------------------------|
| outer_scope_ref       | If with single outer-scope variable reference                  |
| outer_scope_multi_var | If with 3 outer-scope variables (y1=Relu, y2=Sigmoid, y3=Tanh) |
| outer_scope_loop      | Loop body references parent-computed variable                  |
| outer_scope_scan      | Scan body references parent-computed variable                  |

Works on ndarray, tch, and metal backends

image

Generated code: 
[silero_vad.zip](https://github.com/user-attachments/files/24040188/silero_vad.zip)

### DIFF
--- a/crates/onnx-ir/src/node/if_node.rs
+++ b/crates/onnx-ir/src/node/if_node.rs
@@ -93,29 +93,20 @@ impl NodeProcessor for IfProcessor {
             )));
         }
 
-        // Infer output types from branches
-        // When branches have different types, merge them to find the common type.
-        // For tensors with different ranks, use the lower rank (common base type).
-        // ONLY update types, preserve existing output structure (names set by add_node)
+        // Infer output types from then_branch
+        // Both branches produce the same number of outputs, so we use then_branch as the reference.
+        // At runtime, only one branch executes, so the types should be compatible.
 
         // If outputs don't exist yet, create them from branch outputs
         if node.outputs.is_empty() {
-            for (then_output, else_output) in then_outputs.iter().zip(else_outputs.iter()) {
-                let merged_ty = merge_branch_types(&then_output.ty, &else_output.ty);
-                let mut output = then_output.clone();
-                output.ty = merged_ty;
-                node.outputs.push(output);
+            for then_output in &then_outputs {
+                node.outputs.push(then_output.clone());
             }
         } else {
             // Update types for existing outputs (preserves names set by add_node)
-            for (i, (then_output, else_output)) in
-                then_outputs.iter().zip(else_outputs.iter()).enumerate()
-            {
-                let merged_ty = merge_branch_types(&then_output.ty, &else_output.ty);
-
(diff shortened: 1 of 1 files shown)


crates/onnx-ir/src/node/if_node.rs
```diff
@@ -207,6 +215,84 @@ impl NodeProcessor for IfProcessor {
...
+/// while another outputs `[N, 128]`. At runtime, only one branch executes.
+///
+/// For static type inference, we need to choose a consistent type. This function:
+/// 1. Returns the type if both branches match exactly
+/// 2. For tensors with same dtype but different ranks, uses the higher rank
+/// 3. For incompatible dtypes, logs a warning and uses then_branch
+///
+/// We prefer higher rank because lower-rank outputs are often the result of
+/// squeeze operations that can be unsqueezed to match the higher rank.
+fn merge_branch_types(then_ty: &ArgType, else_ty: &ArgType) -> ArgType {
```
laggui: I think this was added as a workaround for the previous model version that had incompatible IF branch output ranks.

As discussed offline, we should remove it.

PR #4120, opened 2025-12-03
chore: Update cubecl

wingertge:
### Checklist

### Changes

Updates to latest cubecl

### DIFF
--- a/crates/burn-nn/src/activation/softplus.rs
+++ b/crates/burn-nn/src/activation/softplus.rs
@@ -86,7 +86,7 @@ mod tests {
 
     #[test]
     fn test_softplus_with_beta() {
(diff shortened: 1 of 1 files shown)


crates/burn-nn/src/activation/softplus.rs
```diff
@@ -85,7 +86,7 @@ mod tests {
 
     #[test]
     fn test_softplus_with_beta() {
-        let device = <TestBackend as Backend>::Device::default();
+        let device: burn_ndarray::NdArrayDevice = <TestBackend as Backend>::Device::default();
```
laggui: Probably didn't mean to add this type hint (only valid for the default `TestBackend`)

wingertge: I must've accidentally double clicked the type hint, whoops 😅

PR #4126, opened 2025-12-05
Refactor `LocalCollectiveServer` for improved clarity and error handling

crutcher:
- Replace `Mutex` with `OnceLock` and enhance state management with `get_backend_client_map`.
- Introduce `LocalCollectiveClientConfig` for flexible server-client setup.
- Improve error handling in collective operations (`Register`, `AllReduce`, `Broadcast`, etc.).
- Streamline runtime initialization with `get_collective_server_runtime`.
- Update `process_*` methods to handle invalid configurations and send errors consistently.
- Rename and consolidate redundant methods for clarity (`get_server_runtime` to `get_collective_server_runtime`).

### DIFF
--- a/crates/burn-collective/src/local/server.rs
+++ b/crates/burn-collective/src/local/server.rs
@@ -130,23 +130,19 @@ pub(crate) fn get_collective_client<B: Backend>() -> LocalCollectiveClient<B> {
 }
 
 /// Global runtime.
-static SERVER_RUNTIME: OnceLock<Mutex<Arc<Runtime>>> = OnceLock::new();
-
-/// Gets a locked mutable view of the `SERVER_RUNTIME`.
-pub(crate) fn get_collective_server_runtime_mut() -> MutexGuard<'static, Arc<Runtime>> {
-    SERVER_RUNTIME
-        .get_or_init(|| {
-            Mutex::new(Arc::new(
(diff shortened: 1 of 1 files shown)


crates/burn-collective/src/local/server.rs
```diff
@@ -60,75 +63,112 @@ pub(crate) struct LocalCollectiveServer<B: Backend> {
...
+            state_map.insert(typeid, Box::new(client.clone()));
+            client
         }
-    };
+    }
+}
 
-    val.downcast_ref().cloned().unwrap()
+/// Global runtime.
+static SERVER_RUNTIME: OnceLock<Mutex<Arc<Runtime>>> = OnceLock::new();
```
nathanielsimard: Is there a reason we need the runtime inside a mutex? Since it's behind an 'Arc' we can't mutate it anyways.

crutcher: This was just a transform from `M<Option<A<T>>` to `OnceLock<M<A<T>>>`; because the init is cleaner.

The only argument for a mutex here is if we wanted to replace the runtime (which is insane to try and make work); so yeah, we should change that.

crutcher: This is done

PR #4127, opened 2025-12-05
Add slice_dyn, slice_assign_dyn, and slice_fill_dyn variants.

crutcher:
Lets us use dynamic `Vec<Slice>` for slicing operations.

### Checklist

### Testing

These are tested by the static versions; which are now simple wrappers.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1265,6 +1265,7 @@ where
     ///   - A single range for 1D slicing (e.g., `0..5`, `..`, `2..`)
     ///   - An array of ranges (e.g., `[0..2, 1..4]`)
     ///   - The [`s!`] macro output for advanced slicing with steps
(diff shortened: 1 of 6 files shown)


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1325,10 +1325,15 @@ where
         S: SliceArg<D2>,
     {
         let shape = self.shape();
-        let slices = slices.into_slices(shape.clone());
+        self.slice_dyn(&slices.into_slices(shape))
+    }
+
+    /// Dynamic version of [`Self::slice`].
+    pub fn slice_dyn(self, slices: &[Slice]) -> Self {
```
laggui: I don't think we need to provide a new entrypoint.

We should make the current implementation more flexible instead, so adjust the `SliceArg` trait accordingly.

crutcher: I can make that work if we relax the D2 param (which is checked dynamically anyway)

PR #4131, opened 2025-12-07
Add direct tensor snapshot retrieval API to ModuleStore

### LINKED ISSUE #4115: Expose get_snapshot and get_snapshots for tensor retrieval in all stores (Burnpack, Safetensors, PyTorch)
### Feature description
Expose methods to retrieve tensor snapshots from all three supported stores (Burnpack, Safetensors, and PyTorch) in burn-store. Specifically:
- `get_snapshot`: returns a single tensor snapshot
- `get_snapshots`: returns a `Vec<TensorSnapshot>`

This should enable users to programmatically access intermediate or specific tensor values (not just entire model loading), similar to the `get` method on PyTorchReader, which allows retrieving a tensor by name:
```rust
pub fn get(&self, name: &str) -> Option<&TensorSnapshot> { ... }
```

Currently, Safetensors does not expose the reader directly (safetensors_to_snapshots_lazy), but it is possible to use the safetensors library for related functionality. All stores should have a convenient API for getting one or multiple tensor snapshots.

### Feature motivation
The request arose from the need to retrieve a single tensor under a specific key without loading an entire model. Presently, workarounds like manually decoding with JSON/serde_json are used, which are not ideal. Exposing these methods would streamline workflows when debugging, extracting features, or making partial use of stored weights/tensors.

This also makes the API consistent with PyTorchFileRecorder, where users can load a record and use a wrapped tensor as needed, and improve parity across supported backends.

- Add a `get_snapshot` and `get_snapshots` method for BurnPackStore, SafetensorsStore, and PytorchStore, returning the appropriate tensor snapshot struct(s).
- Follow API conventions from PyTorchReader and Recorder for naming, error handling, and documentation.
- For Safetensors, if direct exposure is impractical, document recommended usage with the safetensors library, or expose an internal API for lazy snapshots.
- Consider trait unification or a utility trait for common snapshot retrieval/public exposure across stores (for consistency and discoverability).

antimora:
Adds three new methods to the `ModuleStore` trait for direct tensor access without requiring a module, as requested in #4115:

- `get_snapshot(name)` - Get a single tensor by name with O(log n) lookup
- `get_snapshots()` - Get all tensors as a `BTreeMap<String, TensorSnapshot>`
- `keys()` - List all tensor names without loading data

### Checklist

### Related Issues/PRs

Fixes #4115

### Changes

ModuleStore Trait (`traits.rs`)
  - Added `get_snapshot(&mut self, name: &str) -> Result<Option<&TensorSnapshot>, Self::Error>`
  - Added `get_snapshots(&mut self) -> Result<&BTreeMap<String, TensorSnapshot>, Self::Error>`
  - Added `keys(&mut self) -> Result<Vec<String>, Self::Error>`

 Implementation for All Stores
  - **BurnpackStore**: Implemented with caching via `snapshots_cache` field
  - **SafetensorsStore**: Implemented with caching for both `FileStore` and `MemoryStore` variants
  - **PytorchStore**: Implemented with caching (read-only, as saving is not supported)

 Optimizations
  - **Caching**: Parsed snapshots are cached in a `BTreeMap` after first access for O(log n) lookups
  - **Lazy loading**: Tensor data is only materialized when `to_data()` is called
  - **Fast `keys()`**: Can return keys from metadata without full snapshot parsing when cache is empty

### Testing

  - `test_store_get_snapshots` - Get all snapshots as BTreeMap
  - `test_store_get_snapshot_existing` - Get specific tensor by name
  - `test_store_get_snapshot_nested` - Get nested tensor (e.g., "nested.gamma")
  - `test_store_get_snapshot_not_found` - Handle missing tensor gracefully
  - `test_store_keys` - List all tensor names
  - `test_store_get_snapshots_from_file` - File-based access
  - `test_store_caching_behavior` - Verify caching works correctly
  - `test_store_cache_invalidation_on_save` - Verify cache cleared on save

### DIFF
--- a/crates/burn-store/src/safetensors/tests/direct_access.rs
+++ b/crates/burn-store/src/safetensors/tests/direct_access.rs
@@ -34,10 +34,6 @@ impl<B: Backend> DirectAccessTestModule<B> {
     }
 }
 
-// ============================================================================
-// Tests for MemoryStore variant
-// ============================================================================
-
 #[test]
 fn test_memory_get_all_snapshots() {
     let device = Default::default();


crates/burn-store/src/traits.rs
```diff
@@ -197,6 +199,89 @@ pub trait ModuleStore {
...
+    ///
+    /// ```rust,ignore
+    /// let mut store = BurnpackStore::from_file("model.bpk");
+    /// if let Some(snapshot) = store.get_snapshot("encoder.weight")? {
+    ///     println!("Shape: {:?}", snapshot.shape);
+    ///     println!("Dtype: {:?}", snapshot.dtype);
+    ///     let data = snapshot.to_data()?;  // Lazy load
+    /// }
+    /// ```
+    fn get_snapshot(&mut self, name: &str) -> Result<Option<&TensorSnapshot>, Self::Error>;
```
laggui, line 234: The store is a bit like a tensor container, so I think we could rename this to `.get(name)` but that's fine for now

Issue #2871, opened 2025-03-05
Memory optimization for loading weights for no_std mode


HerrMuellerluedenscheid:
Hey folks,

Thanks for building a genius framework! From a recent issue, you probably remember that I tried running the SqueezeNet example on an ESP32. I switched to an ESP32-S3 with 8 MB PSRAM. After failing to run it there due to allocation failures, I started a discussion in the #esp-rs:matrix.org chat, which turned out to be super fruitful (BIG shoutout!).

A few key findings and questions that I will try to summarize from the thread:

1. Burn starts by taking the neural network, pushing it into a Vec<u8>, and then cloning it—leading to 5 MB of RAM usage for something that’s already readable in flash. More specifically, the first part is the generated squeezenet1.rs (generated in target/<ARCH>/debug/build/squeezenet-burn-whatever/out/model):

```
impl<B: Backend> Model<B> {
    pub fn from_embedded(device: &B::Device) -> Self {
        let record = BinBytesRecorder::<HalfPrecisionSettings>::default()
            .load(EMBEDDED_STATES.to_vec(), device) // <-- here is an allocation that shouldn't be needed
            .expect("Should decode state successfully");
        Self::new(device).load_record(record)
    }
}
```

2. burn-core, Recorder::load clones the model again. Is that really necessary?
```
    fn load<R>(&self, args: Self::LoadArgs, device: &B::Device) -> Result<R, RecorderError>
    where
        R: Record<B>,
    {
        let item: BurnRecord<R::Item<Self::Settings>, B> =
            self.load_item(args.clone()).map_err(|err| { // <-- here
```

3. can burn operate directly off from the EMBEDDED_STATES, i.e. without copying the model to RAM

4. Ideally they should split their model into "readonly" stuff and "readwrite" stuff and then the "readonly" stuff is used as-is. I.e. flashed un-decoded.

Apparently there is room for some improvement to run models in `no_std` which I guess will also be beneficial when running with `std`.
Looking forward hearing your thoughts.

BjornTheProgrammer:
I have the exact same issues, and I noticed the same things with the Raspberry Pi Pico, I would be willing to tackle this issue as well with a team that I'm working with. Is there an active PR for this, or should I create one?

HerrMuellerluedenscheid:
@BjornTheProgrammer I just pushed some experiments to https://github.com/tracel-ai/burn/pull/2881. On my ESP32s3 it does not panic because of allocation failures.
[...]

PR #4132, opened 2025-12-07
Update raspberry-pi-pico example to use the Pico 2, and burnpack

BjornTheProgrammer:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/2871

### Changes

This updates the example to use the newer Raspberry Pi Pico 2. It also uses burnpack, and as such fixes some compilation issues that the example had. It also changes to default to using [elf2flash](https://github.com/BjornTheProgrammer/elf2flash) instead of [probe-rs](https://probe.rs/), since `probe-rs` requires owning additional hardware.

> [!WARNING]  
> Please note that I'm the developer behind [elf2flash](https://github.com/BjornTheProgrammer/elf2flash). I'm currently in the process of upstreaming my changes in `elf2flash` to [elf2uf2-rs](https://github.com/JoNil/elf2uf2-rs/issues/43), and until that is done, there isn't a great alternative to my project. As soon as `elf2uf2-rs` is fully updated, I would be interested in changing the docs to recommend using it.

### Testing

I've run the example on real hardware, and everything seems to be working just fine.

### DIFF
--- a/burn-book/src/advanced/no-std.md
+++ b/burn-book/src/advanced/no-std.md
@@ -1,6 +1,6 @@
 # No Standard Library
 
-In this section, you will learn how to run an onnx inference model on an embedded system, with no standard library support on a Raspberry Pi Pico. This should be universally applicable to other platforms. All the code can be found under the
+In this section, you will learn how to run an onnx inference model on an embedded system, with no standard library support on a Raspberry Pi Pico 2. This should be universally applicable to other platforms. All the code can be found under the
 [examples directory](https://github.com/tracel-ai/burn/tree/main/examples/raspberry-pi-pico).
 
 ## Step-by-Step Guide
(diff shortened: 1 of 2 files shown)


BjornTheProgrammer:
Additionally important note here is that the reason for not using [picotool](https://github.com/raspberrypi/picotool) is due to how hard it is to install it on different linux distros.

examples/raspberry-pi-pico/build.rs
```diff
@@ -33,7 +32,7 @@ fn main() {
 
     println!("cargo:rustc-link-arg-bins=--nmagic");
     println!("cargo:rustc-link-arg-bins=-Tlink.x");
-    println!("cargo:rustc-link-arg-bins=-Tlink-rp.x");
+    // println!("cargo:rustc-link-arg-bins=-Tlink-rp.x");
```
laggui: This was specific to RP2040 right?

In this case, we can either add a comment, or remove the line entirely.

BjornTheProgrammer: Yeah it was specific for the rp2040, I'll remove it entirely, since there are a lot more changes in other parts of the code.

PR #4143, opened 2025-12-09
Shape FromStr/ToString

crutcher:
Implement FromStr/ToString for Shape (like Slice).

### DIFF
--- a/crates/burn-std/src/tensor/slice.rs
+++ b/crates/burn-std/src/tensor/slice.rs
@@ -1,7 +1,5 @@
 //! Tensor slice utilities.
 
-use crate::Shape;
-use crate::indexing::AsIndex;
 #[allow(unused_imports)]
 use alloc::format;
 use alloc::string::String;


crates/burn-std/src/lib.rs
```diff
@@ -8,6 +8,7 @@
 //! and data types.
 
 extern crate alloc;
+extern crate core;
```
laggui: We don't need to declare core, it's available. I think this was required pre 2018 edition? :sweat_smile:

crates/burn-std/src/tensor/shape.rs
```diff
@@ -442,6 +464,57 @@ pub fn calculate_matmul_output(lhs: &Shape, rhs: &Shape) -> Result<Shape, ShapeE
...
+                    .trim()
+                    .parse::<usize>()
+                    .map_err(|_| ShapeExpressionError::ParseError {
+                        message: format!("Failed to parse dimension '{}'", dim_str),
+                        source: source.to_string(),
+                    })
+            })
+            .collect::<Result<Vec<usize>, ShapeExpressionError>>()?;
+
+        assert_ne!(dims.len(), 0, "Empty shape expression");
```
laggui: Should return an error instead of panicking?

crates/burn-std/src/tensor/shape.rs
```diff
@@ -556,9 +629,57 @@ impl From<Shape> for Vec<usize> {
...
+
+        assert_eq!(
+            "[a]".parse::<Shape>(),
+            Err(ShapeExpressionError::ParseError {
+                message: "Failed to parse dimension 'a'".to_string(),
+                source: "[a]".to_string()
+            })
+        );
+    }
+
```
laggui, line 716: Could probably add a test for too many braces e.g. `[[1, 2]]`. 

`strip_prefix` and `strip_suffix` will only capture/remove a single leading/trailing character, so this should panic with "Failed to parse dimension".

crates/burn-std/src/lib.rs
crutcher: I think this is an artifact of my auto-import going with `crate::{alloc, core}` in the imports, and it threading this in. Will fix.

crates/burn-std/src/tensor/shape.rs
crutcher: It's impossible; I'll add a comment.

crates/burn-std/src/tensor/shape.rs
crutcher: Added tests; I can't give effective error messages for all malformations without a grammar, which seems too weighty here.

laggui:
Whoops, looks like there might be a conflict with the merged changes from your slice PR that git cannot resolve automatically :sweat_smile: 

Should be good to go after!

PR #4150, opened 2025-12-10
Add contiguous reindexing for non-contiguous layer indices

### LINKED ISSUE #4149: Automatic Contiguous Reindexing of Non-Contiguous Layer Indices for PytorchStore and SafetensorsStore
### Feature description

Add automatic handling for non-contiguous indices in model layer names in PytorchStore and SafetensorsStore. 

Currently, PyTorchFileRecorder provides contiguous reindexing for cases where model layers have gaps in indices (see [existing discussion](https://github.com/tracel-ai/burn/issues/1420)).

**Example scenario:**
- Source keys:
  - `model.layers.0.weight`
  - `model.layers.0.bias`
  - `model.layers.2.weight`  // Gap (no index 1)
  - `model.layers.2.bias`
  - `model.layers.4.weight`
  - `model.layers.4.bias`
- Expected output after reindexing:
  - `model.layers.0.weight`
  - `model.layers.0.bias`
  - `model.layers.1.weight`  // Reindexed from 2
  - `model.layers.1.bias`
  - `model.layers.2.weight`  // Reindexed from 4
  - `model.layers.2.bias`

**Request:**
- Make contiguous reindexing of non-contiguous indices the **default behavior** for PytorchStore. 
- Provide similar support for SafetensorsStore.
- This should enhance loading and saving behavior, reducing the need for manual key remapping and avoiding issues with empty or missing layers.

### Feature motivation

- Simplifies serialization/deserialization for models where layer indices are not continuous.
- Removes the need for users to manually remap indices or tolerate empty maps. 
- Consistency across different storage backends (PyTorch, Safetensors).
- Reduces friction for integrating activations or non-parameter layers into model containers. 
- 

- Implement automatic reindexing logic for both PytorchStore and SafetensorsStore, closely matching the approach in PyTorchFileRecorder. 
- Ensure order preservation and correctness in mapping old to new indices when gaps are present.
- This should be enabled by default for PytorchStore and opt-in (or also default) for SafetensorsStore as appropriate.
- Consider providing an explicit configuration or override for users who prefer manual key mapping. 

References:
- [Issue with current behavior and user workaround](https://github.com/tracel-ai/burn/issues/1420)
- [Discussion](https://github.com/tracel-ai/burn/discussions/1413)

antimora:
Problem

When PyTorch models use `nn.Sequential` with mixed layer types (e.g., Conv2d + ReLU), the state_dict keys have non-contiguous indices because only layers with parameters are saved:

```python
self.fc = nn.Sequential(
  nn.Conv2d(...),  # fc.0.weight, fc.0.bias
  nn.ReLU(),       # No parameters (index 1 skipped)
  nn.Conv2d(...),  # fc.2.weight, fc.2.bias
  nn.ReLU(),       # No parameters (index 3 skipped)
  nn.Conv2d(...),  # fc.4.weight, fc.4.bias
)
```

Burn models expect contiguous indices (fc.0, fc.1, fc.2), causing loading failures without manual remapping.

Solution

Added a reindex_contiguous() function that automatically remaps non-contiguous indices:
```
- fc.0.weight → fc.0.weight
- fc.2.weight → fc.1.weight
- fc.4.weight → fc.2.weight
```
 
API

```rust
// PytorchStore: contiguous_reindex is ON by default
let mut store = PytorchStore::from_file("model.pth");

// Disable if needed
let mut store = PytorchStore::from_file("model.pth")
    .contiguous_reindex(false);

// SafetensorsStore: contiguous_reindex is OFF by default
let mut store = SafetensorsStore::from_file("model.safetensors")
    .contiguous_reindex(true);  // Enable for PyTorch-exported safetensors
```

### Checklist

### Related Issues/PRs

Closes #4149

### Changes
- Add automatic contiguous reindexing for `PytorchStore` and `SafetensorsStore` to handle PyTorch models with non-contiguous layer indices
- Enable reindexing by default for `PytorchStore` (common use case), disable by default for `SafetensorsStore`
- Add `contiguous_reindex(bool)` builder method to both stores

### Testing
- Unit tests for reindex_contiguous() function (6 tests)
- Integration tests for PytorchStore with non-contiguous indices
- Tests verifying behavior with reindexing enabled vs disabled
- All burn-store tests passing

### DIFF
--- a/crates/burn-store/src/keyremapper.rs
+++ b/crates/burn-store/src/keyremapper.rs
@@ -182,6 +182,8 @@ impl KeyRemapper {
 ///
 /// This function detects numeric indices in tensor paths and renumbers them
 /// to be contiguous (0, 1, 2, ...) while preserving their relative order.
+/// It handles nested sequential structures by processing ALL numeric indices
+/// in each path independently based on their position context.
 ///
 /// This is useful when loading PyTorch models that have gaps in layer numbering,
 /// such as when using `nn.Sequential` with mixed layer types (e.g., Conv2d + ReLU
@@ -189,7 +191,7 @@ impl KeyRemapper {
 ///
 /// # Example
 ///
-/// Input paths:
+/// Simple case - input paths:
 /// - `fc.0.weight`, `fc.0.bias`
 /// - `fc.2.weight`, `fc.2.bias`
 /// - `fc.4.weight`, `fc.4.bias`
@@ -198,6 +200,18 @@ impl KeyRemapper {
 /// - `fc.0.weight`, `fc.0.bias`
 /// - `fc.1.weight`, `fc.1.bias`
 /// - `fc.2.weight`, `fc.2.bias`
+///
+/// Nested case - input paths:
+/// - `feature.layers.0.conv_block.0.weight`
+/// - `feature.layers.0.conv_block.2.weight`
+/// - `feature.layers.2.conv_block.0.weight`
+/// - `feature.layers.2.conv_block.2.weight`
+///
+/// Output paths:
+/// - `feature.layers.0.conv_block.0.weight`
+/// - `feature.layers.0.conv_block.1.weight`
+/// - `feature.layers.1.conv_block.0.weight`
+/// - `feature.layers.1.conv_block.1.weight`
 ///
 /// # Arguments
 ///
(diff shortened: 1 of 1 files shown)


crates/burn-store/src/pytorch/store.rs
```diff
@@ -269,6 +274,36 @@ impl PytorchStore {
...
+    /// ```rust,no_run
+    /// # use burn_store::PytorchStore;
+    /// // Disable contiguous reindexing if your model already has contiguous indices
+    /// let store = PytorchStore::from_file("model.pth")
+    ///     .contiguous_reindex(false);
+    /// ```
+    pub fn contiguous_reindex(mut self, reindex: bool) -> Self {
+        self.contiguous_reindex = reindex;
+        self
+    }
```
laggui: I don't think `contiguous_reindex` and `reindex_contiguous` are clear :thinking: 

It's essentially mapping the layer indices, so I think it something like `map_indices_contiguous` or `map_indices_sequential` might be better?

crates/burn-store/src/keyremapper.rs
```diff
@@ -177,6 +178,140 @@ impl KeyRemapper {
...
+/// * `tensors` - Vec of TensorSnapshots to reindex
+///
+/// # Returns
+///
+/// A tuple containing:
+/// * The reindexed Vec of TensorSnapshots with updated paths
+/// * A vector of (new_path, original_path) showing the transformations
+pub fn reindex_contiguous(
+    mut tensors: Vec<TensorSnapshot>,
+) -> (Vec<TensorSnapshot>, Vec<(String, String)>) {
```
laggui: I think there might be an issue in the current logic which groups by `(prefix, suffix)` for paths with nested sequential layer groups.

I don't think it would correctly remap these for example
```
"feature.layers.0.conv_block.0.weight"
"feature.layers.0.conv_block.2.weight"
"feature.layers.2.conv_block.0.weight"
"feature.layers.2.conv_block.2.weight"
```

which could originate from a model like
```python
self.feature = nn.Sequential(
    ConvBlock(...), 
    nn.ReLU(),       # No parameters
    ConvBlock(...),,
    nn.ReLU(),       # No parameters
    ConvBlock(...),
)
```

with a `ConvBlock` defined as sequential e.g. `nn.Sequential(nn.Conv2d(1, 20, 5), nn.ReLU(), nn.Conv2d(20, 64, 5))`

antimora: You're right! 

Fixed!

PR #4152, opened 2025-12-10
Return slice for each dimension in shape

laggui:
### Checklist

### Related Issues/PRs

Reported on discord

```rust
#[test]
fn should_support_slice_assign_single_dim_slice() {
    let x = TestTensor::<3>::ones([2, 3, 1], &device);
    let values = TestTensor::<3>::zeros([1, 3, 1], &device);

    let output = x.slice_assign(s![1], values);

    output.into_data().assert_eq(
        &TensorData::from([[[1.0], [1.0], [1.0]], [[0.0], [0.0], [0.0]]]),
        false,
    );
}
```

Would panic with candle because `into_slice` only returns the dimension slice for the number of dim slices specified.

```
/home/laggui/workspace/burn/crates/burn-candle/src/ops/base.rs:215:78:
called `Result::unwrap()` on an `Err` value: slice-assign requires input with the same rank as there are ranges 3 <> 1
```

Other backends implemented their own "full slice" logic which allowed this to work.

### Changes

Changed `into_slices(shape)` to normalize the behavior. It now returns the slice for each dim in the shape.

### Testing

Unit tests

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1507,7 +1507,11 @@ where
         check!(TensorCheck::slice::<D>(&shape, &slices));
 
         let slice_shape = shape.slice(&slices).unwrap();
-        let value = Self::from_data_dtype([value.elem::<K::Elem>()], &self.device(), self.dtype());
+        let value = Tensor::<B, 1, K>::from_data_dtype(
+            [value.elem::<K::Elem>()],
+            &self.device(),
+            self.dtype(),
+        );
         let value = value.expand(slice_shape);
         self.slice_assign(&slices, value)
     }


crates/burn-std/src/tensor/slice.rs
```diff
@@ -12,22 +11,32 @@ use core::str::FromStr;
...
+    ///
+    /// Returns a [Slice] for each dimension in `shape`.
     fn into_slices(self, shape: &Shape) -> Vec<Slice>;
 }
 
-impl SliceArg for &[Slice] {
+impl<S: Into<Slice> + Clone> SliceArg for &[S] {
     fn into_slices(self, shape: &Shape) -> Vec<Slice> {
-        self.iter()
+        // TODO: should check that slice len <= shape.num_dims()
```
antimora: Todo still remains Consider:

```rust
assert!(
      self.len() <= shape.num_dims(),
      "Cannot provide {} slices for a {}-dimensional shape",
      self.len(),
      shape.num_dims()
  );
```

crates/burn-std/src/tensor/slice.rs
```diff
@@ -901,4 +903,27 @@ mod tests {
...
+
+        let slice = s![1, 0..2];
+        let slices = slice.into_slices(&shape);
+
+        assert_eq!(slices.len(), shape.len());
+
+        assert_eq!(slices[0], Slice::new(1, Some(2), 1));
+        assert_eq!(slices[1], Slice::new(0, Some(2), 1));
+        assert_eq!(slices[2], Slice::full());
+    }
```
antimora, line 977: The test is good but could be expanded to cover:
  - Empty shape dimensions (e.g., Shape::new([0, 3, 1]))
  - Full slices (s![..])
  - When slice count matches dimension count exactly

antimora (approved):
Overall look good; A couple of minor comments. 

Approving in advance.

crates/burn-std/src/tensor/slice.rs
laggui: Yeah wasn't sure if perhaps `into_slices` should return a `Result` instead but we can add this assertion for now

crates/burn-std/src/tensor/slice.rs
laggui: Done!

I also normalized the `Slice::full()` behavior in `into_slices` to return the clamped range for each dim.

laggui:
Latest changes will fix #4187

Issue #4153, opened 2025-12-10
True Zero-Copy Tensor Loading for NdArray Backend


antimora:
### Summary

Currently, even with zero-copy loading enabled at the store level, the NdArray backend still allocates and copies tensor data because ndarray::ArrayD requires owning data as Vec<T>. This proposes modifying burn-ndarray to support true zero-copy by using borrowed array views backed by external memory.

### Current Behavior (312 MB model)

| Level                   | Allocation |
|-------------------------|------------|
| Store level (zero-copy) | 53 KB ✅    |
| NdArray backend         | 327 MB ❌   |

### Expected Results

| Level                    | Allocation   |
|--------------------------|--------------|
| Store level              | 53 KB ✅      |
| NdArray (inference)      | ~0 KB ✅      |
| NdArray (first mutation) | 327 MB (COW) |

PR #4154, opened 2025-12-10
Zero-copy loading for embedded burnpack weights

### LINKED ISSUE #2871: Memory optimization for loading weights for no_std mode

### LINKED ISSUE #4123: Implement zero-copy loading for embedded burnpack weights
When using embed_states(true) with burnpack format, the generated code currently copies the entire model weights from the
binary's data section to the heap. This is wasteful since the data is already in memory.

**Current Behavior**

static EMBEDDED_STATES: &[u8] = include_bytes!("model.bpk");

// This copies all weights to the heap
let bytes = Bytes::from_bytes_vec(EMBEDDED_STATES.to_vec());
let mut store = BurnpackStore::from_bytes(Some(bytes));

For large models (e.g., 100MB+), this means:
- Unnecessary heap allocation during startup
- Peak memory usage is doubled (binary data section + heap copy)
- Slower model initialization

**Proposed Solution**

Use bytes::Bytes::from_static() with a custom AllocationController to provide zero-copy access to the embedded data.

**Implementation Plan**

1. Add ZeroCopyAllocationController in crates/burn-store/src/zero_copy.rs
  - Implements cubecl's AllocationController trait
  - Wraps bytes::Bytes for zero-copy tensor data access
  - Works with both static data (from_static) and memory-mapped files (from_owner)
2. Add BurnpackStore::from_static() method
pub fn from_static(data: &'static [u8]) -> Self
3. Update generated code to use zero-copy:
let mut store = BurnpackStore::from_static(EMBEDDED_STATES);

**Benefits**

- No heap allocation: Embedded data stays in the binary's data section
- Faster startup: No need to copy potentially large model weights
- Lower peak memory: Avoids having two copies during loading
- Extensible: Same controller can be used for memory-mapped file loading

**Implementation Reference**

Full implementation details are documented in crates/burn-store/ZERO_COPY_IMPLEMENTATION.md, including:
- Complete ZeroCopyAllocationController implementation with tests
- Required dependency additions
- BurnpackStore::from_static() method
- Updated code generation for burn-import

**Related**

- This optimization was deferred from the burnpack migration PR to keep scope manageable
- Could also be applied to BurnpackStore::from_file() using memory-mapped files in a follow-up
- https://github.com/tracel-ai/burn/pull/4122
- https://github.com/tracel-ai/burn/issues/2871#issuecomment-3607171211

antimora:
This PR implements zero-copy loading for embedded model weights in the Burnpack format, significantly improving memory efficiency and loading performance for embedded models and memory-mapped files.

### Motivation
When using embed_states(true) with burnpack format, the generated code previously copied entire model weights from the binary' data section to heap:

For large models this caused unnecessary heap allocations and slower startup.

### Solution

Tensor data is now sliced directly from static bytes without copying:

```rust
// After: zero-copy, data stays in .rodata
let mut store = BurnpackStore::from_static(EMBEDDED_STATES);
```

The implementation uses bytes::Bytes::from_static() combined with cubecl's SharedBytesAllocationController for copy-on-write  semantics.

### Checklist

### Related Issues/PRs

- Closes #4123
- Closes #2871
- Related https://github.com/tracel-ai/cubecl/pull/1102

### Changes
- Add BurnpackStore::from_static() for zero-copy loading of embedded model weights and mmap
- Add zero_copy(bool) builder method to control zero-copy behavior
- Update burn-import codegen to use from_static() instead of copying embedded data
- Add benchmarks demonstrating 600x faster loading with 6000x less memory

### Testing

Existing tests pass. Added new ones. 

### Benchmarks:

Store Only (no backend) - Pure Zero-Copy Performance

| Test             | Time     | Alloc    | Speedup                         |
|------------------|----------|----------|---------------------------------|
| file_copy        | 29.37 ms | 327.3 MB | baseline                        |
| file_zero_copy   | 48.96 µs | 53.26 KB | 600x faster, 6000x less memory  |
| static_copy      | 16.95 ms | 982 MB   | baseline                        |
| static_zero_copy | 25.81 µs | 53.12 KB | 656x faster, 18000x less memory |

NdArray Backend - Full Model Loading

| Test             | Time     | Alloc    |
|------------------|----------|----------|
| static_copy      | 21.79 ms | 1.309 GB |
| static_zero_copy | 5.545 ms | 327.3 MB |

Key takeaways:
- Store-level zero-copy is working perfectly: 53 KB vs 327-982 MB allocations
- NdArray still allocates 327 MB because it needs its own ArrayD<T> storage - this is the model's actual size
- The 327 MB for NdArray is unavoidable with the current design - eliminating it would require NdArray to use ArrayView with lifetime management (the work we started in burn-ndarray/src/storage.rs). Filed a ticket to refactor ndarray backend: https://github.com/tracel-ai/burn/issues/4153

### Notes

  - Zero-copy is opt-in via zero_copy(true) or automatic with from_static()
  - Copy remains the default for from_bytes() and from_file() for safety
  - Some backends (e.g., NdArray) still copy data internally—see #4153 for true backend zero-copy
  -   The try_detach() change I made in https://github.com/tracel-ai/cubecl/pull/1102 makes explicitly calls into_vec(), they're asking for owned memory, and a copy must happen. Copy occurs in try_detach instead of bytemuck fallback.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,7 @@ invalid_html_tags = "deny"
 atomic_float = "1"
 axum = "0.8.5"
 bytemuck = "1.24.0"
-# extra-platforms enables portable-atomic for targets without native atomics (e.g., thumbv7m-none-eabi)
-bytes = { version = "1", default-features = false, features = ["extra-platforms"] }
+bytes = { version = "1.10", default-features = false }
 candle-core = { version = "0.9.1" }
 ciborium = { version = "0.2", default-features = false }
 clap = { version = "4.5.48", features = ["derive"] }
--- a/crates/burn-std/Cargo.toml
+++ b/crates/burn-std/Cargo.toml
@@ -44,6 +44,10 @@ tokio = { workspace = true, optional = true }
 [dev-dependencies]
 dashmap = { workspace = true }
 
+# Enable extra-platforms for bytes on targets without native atomics (e.g., thumbv6m-none-eabi)
+[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
+bytes = { workspace = true, features = ["extra-platforms"] }
+
 [package.metadata.docs.rs]
 features = ["doc"]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/crates/burn-store/Cargo.toml
+++ b/crates/burn-store/Cargo.toml
@@ -81,3 +81,7 @@ harness = false
 [[bench]]
 name = "zero_copy_loading"
 harness = false
+
+# Enable extra-platforms for bytes on targets without native atomics (e.g., thumbv6m-none-eabi)
+[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
+bytes = { workspace = true, features = ["extra-platforms"] }


Cargo.toml
```diff
@@ -36,7 +36,8 @@ invalid_html_tags = "deny"
 atomic_float = "1"
 axum = "0.8.5"
 bytemuck = "1.24.0"
-bytes = "1"
+# extra-platforms enables portable-atomic for targets without native atomics (e.g., thumbv7m-none-eabi)
+bytes = { version = "1", default-features = false, features = ["extra-platforms"] }
```
nathanielsimard: Do we want extra-platforms by default?

nathanielsimard (approved):

Cargo.toml
antimora: it's not enabled by default, so should enable if atomic is missing. I'll fix it here and in CubeCL

laggui: In this case, we probably want to use something like this to enable this feature only when necessary

```
[target.'cfg(not(target_has_atomic = "ptr"))'.dependencies]
```

PR #4157, opened 2025-12-11
Reorganize and tracing::instrument collective operations.

crutcher:
### Checklist

- clean up a lot of weird in-place HashMap stuff.
- reorg some private libs
- partially unify `CollectiveTensorMap<B>` (maybe a struct later? idk)
- instrument collective ops, some of the float ops

- introduced a benchtop framework bin to run 1 all_reduce

- disabled for no-std

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -960,7 +960,6 @@ version = "0.20.0-pre.5"
 dependencies = [
  "burn-backend",
  "hashbrown 0.16.1",
- "portable-atomic-util",
  "serde",
 ]
 
@@ -984,6 +983,7 @@ dependencies = [
  "num-traits",
  "openblas-src",
  "paste",
+ "portable-atomic",
  "portable-atomic-util",
  "rand 0.9.2",
  "rayon",
--- a/crates/burn-autodiff/Cargo.toml
+++ b/crates/burn-autodiff/Cargo.toml
@@ -23,6 +23,8 @@ export_tests = [] # check checkpointer is_empty in tests
 burn-std = { path = "../burn-std", version = "=0.20.0-pre.5", default-features = false }
 burn-backend = { path = "../burn-backend", version = "=0.20.0-pre.5", default-features = false }
 
+tracing = { workspace = true, optional = true }
+
 derive-new = { workspace = true }
 spin = { workspace = true }
 parking_lot = { workspace = true, optional = true }


crutcher:
I'm struggling with the no-std deps; here's a fast failing example.
```terminaloutput
$ cargo build --color always --no-default-features --target thumbv6m-none-eabi -p burn
```

crates/burn-collective/src/local/all_reduce/base.rs
```diff
@@ -12,200 +14,90 @@ use crate::{
...
-        main_tensor = global_client
+/// Do an all-reduce in a multi-node context
+///
+/// With Tree and Centralized strategies, the all-reduce is split between a
+/// reduce (all tensors are reduced to one device), and a broadcast (the result is sent to all
+/// other devices). The all-reduce on the global level is done between both steps.
+/// Due to the nature of the Ring strategy, this separation can't be done.
+// For the Ring strategy, this isn't possible, because it is more like a
+// reduce-scatter plus an all-gather, so using a Ring strategy locally in a multi-node
+// setup may be unadvantageous.
```
nathanielsimard: Those two last lines should have 3 `/` and not 2 `/`: `//` => `///`

crates/burn-collective/src/local/all_reduce/base.rs
```diff
@@ -12,200 +14,90 @@ use crate::{
...
+    let devices = tensors
+        .iter()
+        .map(|(id, tensor)| (*id, B::float_device(tensor)))
+        .collect::<HashMap<PeerId, B::Device>>();
+
+    // For Centralized and Tree, we only need to do a reduce here, we'll do a broadcast later
+    let main_device = *tensors.keys().next().unwrap();
+
+    let mut main_tensor = match local_strategy {
+        AllReduceStrategy::Centralized => reduce_sum_centralized::<B>(tensors, &main_device),
```
nathanielsimard, line 66: It's strange that the global all reduce call `reduce_sum_centralized` and the local one call `all_reduce_sum_centralized`. When prefix by `all` it's local, and if not it's global?.

Maybe we should add proper pre/post fix?

crates/burn-cubecl/src/ops/float_ops.rs
```diff
@@ -26,6 +26,10 @@ where
     I: IntElement,
     BT: BoolElement,
 {
+    #[tracing::instrument(
```
nathanielsimard, line 29: Is it going to be duplicated by the `burn-autodiff` backend? How does that look, do we see a backtrace of the same call and the lower level function calls?

examples/dop_timer/src/main.rs
```diff
@@ -0,0 +1,395 @@
+#![allow(unused)]
```
nathanielsimard: The example can't be executed with `cargo --example reduce_timing` I think you need to create a main in `reduce_timing/examples/reduce_timing.rs` and call this file as a `lib.rs`

nathanielsimard: Also it would be great to have a README.md with the expected output of the program :)

crates/burn-collective/src/local/all_reduce/base.rs
crutcher: `reduce_sum(...) -> Tensor` and `all_reduce_sum() -> { PeerId: Tensor }` are different operations; but I do agree that there's duplication, and we probably want a local op-combinator library to avoid that duplication, even if those ops aren't shared to users.

examples/dop_timer/src/main.rs
crutcher: It is this:
```
 cargo run -p reduce_timing --release -- --tracing otel --shape "42, 3, 1024, 1024" --workers-per-device 3 --op mean
 ```
 
 I can look more at how `-p` vrs `--example` is structured.

**Comment by antimora** [comment]
Disabled for no-std and rebased with the main

**Comment by crutcher** [comment]
This is as much as I want to put in 1 PR

File: examples/dop_timer/README.md [hunk 5]
```diff
@@ -0,0 +1,13 @@
+# dop_timer
+
+This binary exists to time the behavior of distributed (local, global) collective operations.
+
+## Example
+
+1. Setup an OTEL Collector
```
nathanielsimard, line 9: Could you provide a link to setup the OTEL Collector?

crates/burn-backend/Cargo.toml
```diff
@@ -27,6 +27,7 @@ cubecl-cpu = ["cubecl", "cubecl/cpu"]
 [dependencies]
 burn-std = { path = "../burn-std", version = "0.20.0-pre.5", default-features = false }
 cubecl = { workspace = true, optional = true, default-features = false }
+tracing = { workspace = true, features = ["attributes"], optional = true }
```
laggui: Might have removed the tracing usage in burn-backend? Doesn't appear to be used in this crate

crates/burn-collective/src/local/broadcast/op.rs
```diff
@@ -1,13 +1,14 @@
...
-
+use crate::local::tensor_map::{CollectiveTensorMap, PeerDeviceMap};
 use crate::{
     BroadcastStrategy, CollectiveConfig, CollectiveError, PeerId,
     local::{broadcast_centralized, broadcast_tree},
     node::base::Node,
 };
+use burn_communication::Protocol;
+#[allow(unused_imports)]
+use burn_tensor::TensorMetadata;
```
laggui: Why allow unused?

crates/burn/Cargo.toml
```diff
@@ -69,6 +69,10 @@ std = [
     "burn-store?/std",
     "burn-nn/std",
     "burn-optim/std",
+    "dep:tracing",
+    "tracing?/default",
+    "dep:tracing-core",
+    "tracing-core?/default",
```
laggui: Shouldn't this only be enabled only in the crates that actually use it? For example, propagating `burn-autodiff/std` should be sufficient. Same for `burn-cubecl/std`.

Might be missing burn-remote and burn-collective instead?

crates/burn-collective/src/local/all_reduce/base.rs
laggui: Still the case btw.

If it's meant as a comment / note, maybe prefix with `// NOTE:` instead?

examples/dop_timer/README.md
crutcher: Done

crates/burn-collective/src/local/broadcast/op.rs
crutcher: Because the rust linter isn't smart enough to understand that `TensorMetadata` trait operations are being used by `#[tracing::instrument]`

laggui: Oh makes sense, I didn't notice it in the `tensor.as_ref().map(..)`!

crutcher: Documented.

crates/burn-backend/Cargo.toml
crutcher: Done

crates/burn/Cargo.toml
crutcher: Done, I think.
- `burn-remote` and `burn-collective` don't support `no-std`, and don't have a `std` feature.

laggui (approved):

Issue #3995, opened 2025-11-08
ONNX import fails on standard gpt2 model from Optimum with unwrap()


pavel807:
pavellabushev@PavelLabushevs-MacBook-Pro convert % cargo run       
   Compiling convert v0.1.0 (/Users/pavellabushev/coding/convert)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/convert`
 INFO burn_import::onnx::model_gen: Starting to convert ONNX to Burn
ERROR burn_import::logger: PANIC => panicked at /Users/pavellabushev/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-import-0.19.1/src/onnx/model_gen.rs:405:53:
OUT_DIR env is not set: NotPresent

thread 'main' (310821) panicked at /Users/pavellabushev/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/burn-import-0.19.1/src/onnx/model_gen.rs:405:53:
OUT_DIR env is not set: NotPresent
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

antimora:
Sure. I can assist you. Do you have a URL to the ONNX file so we can check as well? The error you pasted is about an existing file.

pavel807:
yes url 1 moment https://huggingface.co/onnx-community/gemma-3-270m-it-ONNX

antimora:
Currently ONNX import does not support external data yet, see https://github.com/tracel-ai/burn/issues/3812

Gemma ONNX contains external data: 

<img width="1335" height="504" alt="Image" src="https://github.com/user-attachments/assets/ad237867-7a0c-4eca-9385-998eb5cdc8db" />

antimora:
Submitted a PR: https://github.com/tracel-ai/burn/pull/4158

Issue #3812, opened 2025-09-30
ONNX import limited to 2GB models due to protobuf deserialization


antimora:
**Description:**
The current ONNX import implementation in burn-import cannot load models larger than 2GB due to protobuf's serialization limit. This prevents importing many modern large models like Stable Diffusion, SDXL, and other vision/language models.

**Current Implementation:**
`onnx-ir/src/from_onnx.rs:1124`:

```
let onnx_model: ModelProto =
    Message::parse_from_reader(&mut file).expect("Unable to parse ONNX file");
```

This uses protobuf deserialization which has a hard 2GB limit.

**The Issue:**
When attempting to import models like Stable Diffusion 1.5 UNet (~3.4GB), the import fails because the protobuf parser cannot deserialize files larger than 2GB into memory.

**Industry Standard Solution:**
ONNX officially supports external data format for models >2GB:
- The .onnx file contains only the graph structure (small)
- Model weights are stored in separate .onnx.data or .bin files
- The protobuf contains pointers (location, offset, length) to external weight data

**References:**
- https://onnx.ai/onnx/api/serialization.html
- https://github.com/microsoft/onnxruntime/issues/15349
- https://github.com/onnx/onnx/issues/4394

**Proposed Solution:**
Add support for ONNX external data format in onnx-ir:

1. Parse the .onnx file normally (graph structure is small)
2. For tensors with external_data field set, read from external files:
  - location: relative file path to external data
  - offset: byte offset in external file
  - length: number of bytes to read
3. Load tensor data from external files as needed

Example External Data Structure:

```
tensor {
  data_location: EXTERNAL
  external_data {
    key: "location"
    value: "model_weights.bin"
  }
  external_data {
    key: "offset"
    value: "1024"
  }
  external_data {
    key: "length"
    value: "4096"
  }

```
}

  Impact:
  Without this feature, burn-import cannot support:
  - Stable Diffusion (any version)
  - Large vision models (YOLO-X, ResNet-152, etc.)
  - Large language models
  - Any model where individual layers exceed 2GB

  Workaround:
  Currently, users would need to manually split models or use smaller model variants, but this is not always practical.

  Related Files:
  - /crates/onnx-ir/src/from_onnx.rs - Main ONNX parsing logic
  - /crates/onnx-ir/src/protos/onnx.proto - Protobuf definitions (should already have TensorProto.DataLocation and external_data fields
  from ONNX spec)

SinanGncgl:
[...]

PR #4158, opened 2025-12-11
Add ONNX external data support for models >2GB

antimora:
This PR adds support for loading ONNX models that store tensor weights in external files, which is necessary for models exceeding protobuf's 2GB size limit (also exceeding enabled by default now). The implementation includes lazy loading of external tensor data via memory mapping, security hardening against path traversal attacks, and comprehensive test coverage.

### Checklist

### Related Issues/PRs

- Closes https://github.com/tracel-ai/burn/issues/3812
- Closes https://github.com/tracel-ai/burn/issues/3995

### Changes

- **External data parsing**: New `external_data` module parses `TensorProto.external_data` fields (location, offset,
length, checksum)
- **Lazy loading**: External tensor data is loaded on-demand via memory mapping (`memmap2`) when `to_tensor_data()` is
called
- **Uniform access**: `TensorDataSource` enum abstracts over embedded vs external storage - consumers use the same
`DataId` interface regardless of where data lives
- **Base path propagation**: Pipeline functions now accept/propagate base directory for resolving relative external data
 paths
- **Security hardening**: Path traversal protection against malicious ONNX files (rejects `../`, absolute paths, null
bytes, symlink escapes)
- **Multiple external files**: Supports models with tensors split across multiple `.bin` files
- **Full dtype coverage**: F32, F64, F16, BF16, I8, I16, I32, I64, U8, U16, U32, U64, Bool

### Testing
- Unit tests for external data parsing and path resolution
- Security tests for path traversal attack vectors
- Integration tests with fixtures generated by Python script
- Tests for multiple external files, mixed embedded/external, offset handling
- All tests passing, clippy clean


mertalev:
After merging main into this branch for the gelu fix, I can confirm it compiles and runs with external data.

crates/onnx-ir/src/pipeline.rs
```diff
@@ -290,8 +325,10 @@ pub fn build_graph_from_proto_with_registry(
     graph: &crate::protos::GraphProto,
     opset_version: usize,
     name_registry: Option<crate::graph_state::NameRegistry>,
```
laggui: Looks like `build_graph_from_proto_with_registry` is always called without a provided name registry... 

I think we could clean up these functions, and only keep `build_graph_from_proto_with_base_path` since `build_graph_from_proto` is also unused.

Otherwise LGTM.

crates/onnx-ir/src/proto_conversion.rs
```diff
@@ -260,15 +267,30 @@ pub fn argument_from_initializer(initializer: &TensorProto) -> (Argument, Tensor
 /// This is the preferred path for mmap loading - it creates TensorDataRef directly
 /// from the TensorProto without going through TensorData, avoiding unnecessary copies.
 /// The tensor bytes remain as references to the mmap'd buffer until actually accessed.
+///
+/// For external data support, pass the base_path (directory containing the ONNX file).
+/// External tensor data will be loaded lazily when the TensorDataRef is accessed.
+#[allow(dead_code)]
 pub fn argument_from_initializer_lazy(
```
laggui: Why keep an unused function?

PR #4161, opened 2025-12-12
fix(onnx): Use activation function for GELU codegen instead of non-existent tensor method

### LINKED ISSUE #4160: Generated code fails to build due to missing `.gelu()` method
**Describe the bug**
When attempting to import a model using the changes from PR https://github.com/tracel-ai/burn/pull/4158, the import succeeds and generates the expected `.rs` and `.bpk` files. However, the generated code fails to build due to multiple errors indicating that the `gelu` method is not found on `burn::Tensor<B, D, K>`.

Example error message:
```
error[E0599]: no method named `gelu` found for struct `burn::Tensor<B, D, K>` in the current scope
    --> src/model/vit_so400m_14_siglip2_webli/visual.rs:1928:42
     |
1928 |         let gelu27_out1 = linear107_out1.gelu();
     |                                          ^^^^ method not found in `burn::Tensor<B, 3>`

error[E0599]: no method named `gelu` found for struct `burn::Tensor<B, D, K>` in the current scope
    --> src/model/vit_so400m_14_siglip2_webli/visual.rs:1971:42
     |
1971 |         let gelu28_out1 = linear112_out1.gelu();
     |                                          ^^^^ method not found in `burn::Tensor<B, _>`
```

**To Reproduce**
1. Import a model using the workflow enabled by PR https://github.com/tracel-ai/burn/pull/4158
2. Attempt to build the generated `.rs` code
3. Observe build failures regarding missing `gelu` methods on `Tensor`

**Expected behavior**
The generated Rust code should compile successfully. All necessary methods (such as `.gelu()`) should be available on the expected types.

**Additional context**
- Reported by a user following the merge of PR https://github.com/tracel-ai/burn/pull/3844#issuecomment-3643933050
- This problem likely affects models with GELU activations during import.
- Example error locations: `visual.rs:1928` and `visual.rs:1971`.

antimora:
### Summary

- Fix GELU code generation to use `burn::tensor::activation::gelu()` instead of `.gelu()` method
- Update ONNX test to use native GELU op (opset 20) instead of decomposed version

### Problem

The GELU node codegen was generating `input.gelu()` which doesn't exist as a tensor method. This caused compilation failures when importing ONNX models containing native GELU ops:

error[E0599]: no method named gelu found for struct burn::Tensor<B, D, K>

Fixes #4160

### Solution

Changed the codegen to use the standalone activation function, consistent with other activations like ReLU and Sigmoid:

```rust
// Before (broken)
let output = input.gelu();

// After (correct)
let output = burn::tensor::activation::gelu(input);
```

Why the existing test didn't catch this

The previous ONNX test used PyTorch export with opset 16, which decomposes GELU into primitive ops (erf, div, mul, add). This meant the GeluNode codegen path was never exercised.

Updated the test to manually construct an ONNX model with a native GELU op (opset 20) using onnx.helper, ensuring the codegen is properly tested.

### Test plan

- Unit test burn::node::gelu::tests::test_gelu_forward passes
- Integration test onnx-tests::gelu::tests::gelu passes with native GELU op


antimora:
CC @mertalev

PR #4180, opened 2025-12-13
Fix quantized tensor storage data length calculation

### LINKED ISSUE #4179: Panic when storing quantized weights with burnpack
**Describe the bug**
When saving a model with `BurnpackStore` after quantizing, the program panics:
```
thread 'main' (11961) panicked at src/bin/repro.rs:29:45:
called `Result::unwrap()` on an `Err` value: IoError("Data corruption: tensor 'weight' has inconsistent length (expected 1048576, got 262148)")
```

**To Reproduce**
- Run the following code:
  ```rs
  #![recursion_limit = "256"]
  
  use burn::{
      backend::Cpu,
      module::{Module, Quantizer},
      nn::LinearConfig,
      prelude::Backend,
      tensor::quantization::{Calibration, QuantScheme},
  };
  
  use burn_store::{BurnpackStore, ModuleSnapshot};
  
  type B = Cpu;
  
  fn main() {
      let device = <B as Backend>::Device::default();
  
      let transformer = LinearConfig::new(512, 512).init::<B>(&device);
  
      let scheme = QuantScheme::default();
      let mut quantizer = Quantizer {
          calibration: Calibration::MinMax,
          scheme,
      };
  
      let transformer = transformer.quantize_weights(&mut quantizer);
  
      let mut out_weights = BurnpackStore::from_file("test").overwrite(true);
      transformer.save_into(&mut out_weights).unwrap(); // Panic
  }
  ```

**Expected behavior**
The weights should properly saved in a quantized format.

**Screenshots**
/

 - OS: Linux (NixOS unstable)
 - Version Latest commit as of today

**Additional context**
The backtrace is unfortunately not very useful as it points towards the unwrap in the main.rs file.

antimora:
### Checklist

### Related Issues/PRs

Closes #4179

### Changes

- Fix TensorSnapshot::data_len() to correctly calculate byte size for quantized tensors
- Add DType::data_bytes(shape) method that accounts for quantization parameters
- Add QPARAM_ALIGN constant for quantization parameter alignment

### Testing
  - Added test_store_quantized_module_round_trip - tensor-level Q8S quantization
  - Added test_store_quantized_module_block_level - block-level quantization
  - Added test_data_bytes_* unit tests for all quantization configurations
  - All existing tests pass

### DIFF
--- a/crates/burn-std/src/tensor/quantization.rs
+++ b/crates/burn-std/src/tensor/quantization.rs
@@ -6,6 +6,10 @@ pub use cubecl_common::quant::scheme::{
 };
 
 /// Alignment (in bytes) for quantization parameters in serialized tensor data.
+///
+/// NOTE: This is currently f32-based since scales were originally always f32.
+/// With `QuantParam` now supporting different precisions (F16, BF16, etc.),
+/// this alignment may need to be revisited in the future.
 pub const QPARAM_ALIGN: usize = core::mem::align_of::<f32>();
 
 use alloc::vec::Vec;


crates/burn-std/src/tensor/dtype.rs
```diff
@@ -138,6 +151,39 @@ impl DType {
...
+                    params_shape(&Shape::from(shape.to_vec()), scheme.level).num_elements();
+
+                let aligned_value_bytes = value_bytes.div_ceil(QPARAM_ALIGN) * QPARAM_ALIGN;
+                let scale_bytes = num_params * quant_param_size(scheme.param);
+
+                aligned_value_bytes + scale_bytes
+            }
+            _ => num_elements * self.size(),
+        }
+    }
```
laggui: I think I would keep this in burn-store for now, though we should eventually consolidate this with the burn-cubecl code that already performs similar checks.

crates/burn-std/src/tensor/quantization.rs
```diff
@@ -116,14 +119,14 @@ impl QuantizedBytes {
         match scheme.level {
             QuantLevel::Tensor => {
                 let scale_bytes = bytemuck::bytes_of(&scales[0]);
-                bytes.extend_from_byte_slice_aligned(scale_bytes, align_of::<f32>());
+                bytes.extend_from_byte_slice_aligned(scale_bytes, QPARAM_ALIGN);
```
laggui, line 126: iirc the scales were initially always f32, but now we have `QuantParam` precision that can be changed. But you can leave this as is for this PR.

Only affects ndarray implementation for now, which is only tested in f32 for reference. But burn-cubecl `into_data` doesn't go through `QuantizedBytes` :sweat_smile: will have to clean this up.

antimora: Added a comment to capture for the future

PR #4189, opened 2025-12-15
Add `flatten_dims` method to `Shape` and refactor tensor flattening API

crutcher:
- Introduce `Shape::flatten_dims` for collapsing specified dimensions into one.
- Update tensor flattening logic in base API to use the new `flatten_dims` method.
- Add unit test for `Shape::flatten_dims` functionality.

### Checklist

### DIFF
--- a/crates/burn-std/src/tensor/shape.rs
+++ b/crates/burn-std/src/tensor/shape.rs
@@ -125,7 +125,6 @@ impl Shape {
     /// # Example
     ///
     /// ```rust
-    ///
     /// use burn_std::Shape;
     ///
     /// fn example() {


crates/burn-std/src/tensor/shape.rs
```diff
@@ -105,6 +107,59 @@ impl Shape {
...
+    ///
+    /// # Returns
+    ///
+    /// A new `Shape` instance with the specified range of dimensions flattened.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    ///
+    /// use burn_tensor::Shape;
```
laggui: `Shape` is not in burn_tensor anymore :smile:

crates/burn-std/src/tensor/shape.rs
```diff
@@ -1,9 +1,14 @@
...
 
+use super::indexing::ravel_index;
+use super::{AsIndex, Slice, SliceArg};
+use crate::canonicalize_dim;
 #[allow(unused_imports)]
 use alloc::format;
 use alloc::string::String;
 use alloc::string::ToString;
+#[allow(unused_imports)]
+use alloc::vec;
```
laggui: It is used, so why `allow(unused_imports)`?

crutcher: done.

crates/burn-std/src/tensor/shape.rs
crutcher: done.

PR #4196, opened 2025-12-15
Refactor: Replace `canonicalize_dim` with `expect_dim`

crutcher:
- Update occurrences of `canonicalize_dim` across tensor API methods and `indexing.rs` with `expect_dim`.
- Simplify dimensional validation by using `NegativeWrap::expect_dim`.
- Adjust imports accordingly, removing `canonicalize_dim` references.
- Introduce `NegativeWrap` struct for streamlined dimension and index wrapping with improved error handling.

### Checklist

### DIFF
--- a/crates/burn-std/src/tensor/indexing.rs
+++ b/crates/burn-std/src/tensor/indexing.rs
@@ -23,12 +23,12 @@ pub trait AsIndex: Debug + Copy + Sized {
     /// Converts into a slice index.
     fn index(self) -> isize;
 
-    /// Short-form `NegativeWrap::expect_index(idx, size)`.
+    /// Short-form [`IndexWrap::expect_index(idx, size)`].
     fn expect_elem_index(self, size: usize) -> usize {
         IndexWrap::expect_elem(self, size)
     }
 
-    /// Short-form `NegativeWrap::expect_dim(idx, size)`.
+    /// Short-form [`IndexWrap::expect_dim(idx, size)`].
     fn expect_dim_index(self, size: usize) -> usize {
         IndexWrap::expect_dim(self, size)
     }
@@ -121,8 +121,8 @@ impl IndexWrap {
 
     /// Set the policy for wrapping 0-size ranges.
     ///
-    /// - when size == 0:
-    ///   - if wrap_scalar; then size == 1
+    /// When ``size`` == 0:
+    ///   - if `wrap_scalar`; then ``size == 1``
     ///   - otherwise; an error.
     pub fn with_wrap_scalar(self, wrap_scalar: bool) -> Self {
         Self {
@@ -141,12 +141,12 @@ impl IndexWrap {
         expect_wrap(idx, size, self.kind, self.wrap_scalar)
     }
 
-    /// Short-form `NegativeWrap::index().expect_wrap(idx, size)`.
+    /// Short-form [`NegativeWrap::index().expect_wrap(idx, size)`].
     pub fn expect_elem<I: AsIndex>(idx: I, size: usize) -> usize {
         Self::index().expect_wrap(idx, size)
     }
 
-    /// Short-form `NegativeWrap::dim().expect_wrap(idx, size)`.
+    /// Short-form [`NegativeWrap::dim().expect_wrap(idx, size)`].
     pub fn expect_dim<I: AsIndex>(idx: I, size: usize) -> usize {
         Self::dim().expect_wrap(idx, size)
     }


crates/burn-std/src/errors.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct BoundsError {
+    /// The name/type of the index.
+    pub index_name: String,
+    /// The index value.
+    pub index: String,
+    /// The name/type of the bounds.
+    pub bounds_name: String,
+    /// The bounds value.
+    pub bounds: String,
+}
```
laggui, line 54: Instead of stringy types, I would make this more restricted e.g.

```rust
pub struct BoundsError {
    pub kind: IndexKind,
    pub index: isize,
    pub size: usize,
}
```

With `IndexKind` enum
```rust
pub enum IndexKind {
    Elem,
    Dim,
}
```

given `index_name` is either "index" or "dimension index" (i.e., what are we indexing).

crates/burn-std/src/tensor/indexing.rs
```diff
@@ -79,109 +94,148 @@ impl AsIndex for u8 {
...
-    Index: AsIndex,
-{
-    canonicalize_named_index("index", "size", idx, size, wrap_scalar)
+/// Wraps an index with negative indexing support.
+#[derive(Debug)]
+pub struct NegativeWrap {
+    index_name: &'static str,
+    size_name: &'static str,
+    wrap_scalar: bool,
+}
```
laggui, line 103: The name `NegativeWrap` might be a little underspecified. We can re-use the suggested `IndexKind`, and have

```rust
pub struct CanonicalIndex {
    kind: IndexKind,
    wrap_scalar: bool,
}
```

and then `CanonicalIndex::elem()` or `CanonicalIndex::dim()` constructors.

With other functions renamed accordingly, since `try_wrap` / `expect_wrap` remain a bit ambiguous.

crates/burn-std/src/errors.rs
crutcher: BoundsError; i specifically wanted to avoid valued bounds for this object. I want to re-use it as a reporting mechanism in other places; and modeling every possible bounds error (when we currently foresee no programatic consumers of these errors) seemed overkill.

In general, structured errors should only exist when there is a valid reason to consume that structure.

When we throw errors as a result of programming mistakes (as opposed to runtime variances); the correct structural mechanism was to have generated a correct request in the first place; and it seems unlikely that callers who failed to do that, will somehow meaningfully unroll from the structured error.

The above suggestion immediately has issues when we want multi-dim bounds; or the issue is range containment, or slice membership.

crates/burn-std/src/tensor/indexing.rs
crutcher: None of the current users call this constructor, or even import it.

As the short-cut path is added to `AsIndex`; almost all callers are just:
```rust
let dim = dim.expect_dim(D);
```

crates/burn-std/src/errors.rs
laggui: > BoundsError; i specifically wanted to avoid valued bounds for this object. I want to re-use it as a reporting mechanism in other places; and modeling every possible bounds error (when we currently foresee no programatic consumers of these errors) seemed overkill.

I don’t see how the string-based `BoundsError` helps with multi-dim, slice, or range errors. It doesn’t model those cases, it merely allows printing arbitrary text.

If these really are just programming errors with no programmatic consumers, then why not just use a formatted string (panic, or error) and skip the `BoundsError` struct entirely?

If you want a reusable error type that handles multiple scenarios, then you'd need an explicit error type that can actually model those cases e.g.

```rust
pub enum BoundsError {
    Index { .. },
    Range { .. },
    MultiDim { .. },
    // etc.
}
```

crates/burn-std/src/tensor/indexing.rs
laggui: > As the short-cut path is added to `AsIndex`; almost all callers are just:
> 
> ```rust
> let dim = dim.expect_dim(D);
> ```

Yeah I like this change, I was only commenting w.r.t the name which is not as clear as the previous "canonicalize index" you had. 

The changes to the field of the struct were merely to reflect the changes in my other comment.

crates/burn-std/src/errors.rs
crutcher: @laggui I *don't* want to handle multiple error types, programatically on the consumer side. I don't think we have a use case for them yet, and I don't think we have a design; so the "You get an Error impl, and some text" is the plan I'm arguing for.

crates/burn-std/src/tensor/indexing.rs
crutcher: `IndexWrap`?

crutcher: done.

crates/burn-std/src/errors.rs
crutcher: done.

laggui (approved):

PR #4201, opened 2025-12-16
Fix checkpoint and summary log level

J-F-Liu:
### Checklist

### Related Issues/PRs

None

### Changes

Change `Saving checkpoint` and `Removing checkpoint` log level to TRACE, since they are very noisy in log file.
Log summary output instead of `println`, since there is no easy way to extract the result summary.

### Testing

Does not affect testing.

### DIFF
--- a/crates/burn-train/src/renderer/tui/renderer.rs
+++ b/crates/burn-train/src/renderer/tui/renderer.rs
@@ -367,6 +367,7 @@ impl Drop for TuiMetricsRenderer {
             self.reset().unwrap();
 
             if let Some(summary) = &self.summary {
+                println!("{summary}");
                 log::info!("{summary}");
             }
         }


crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -367,7 +367,7 @@ impl Drop for TuiMetricsRenderer {
             self.reset().unwrap();
 
             if let Some(summary) = &self.summary {
-                println!("{summary}");
+                log::info!("{summary}");
```
laggui, line 371: I think displaying the summary on stdout is still a useful signal post-training, but agree that it could be logged.

So why not both? :slightly_smiling_face:

laggui (commented):
LGTM, just one comment for the summary:

crates/burn-train/src/renderer/tui/renderer.rs
J-F-Liu: Yes, both is better, code updated.

PR #4204, opened 2025-12-16
Perf/reduce cpu + Fix OOB

nathanielsimard:
Requires CubeK: https://github.com/tracel-ai/cubek/pull/31

### DIFF
--- a/crates/burn-cubecl/src/kernel/reduce/tune.rs
+++ b/crates/burn-cubecl/src/kernel/reduce/tune.rs
@@ -72,7 +72,7 @@ pub fn autotune_reduce<R: CubeRuntime>(
                     )| {
                         let strategy = ReduceStrategy {
                             routine: routine.clone(),
-                            line_size: line_size.clone(),
+                            line_size,
                         };
                         cubek::reduce::reduce::<R>(
                             &input.client,


crates/burn-cubecl-fusion/src/optim/reduce/optimization.rs
laggui, line 1: Missing fusion without autotune changes:

```rust
        if arg
            .execute_fused::<BT>(
                context,
                ReduceStrategy::FullUnit(RoutineStrategy::Strategy(UnitStrategy)),
            )
            .is_err()
        {
            arg.execute_fallback::<BT>(context);
        }
```

lines 214-223

crates/burn-cubecl-fusion/Cargo.toml
```diff
@@ -28,6 +28,7 @@ burn-ir = { path = "../burn-ir", version = "=0.20.0-pre.5", default-features = f
 burn-std = { path = "../burn-std", version = "=0.20.0-pre.5" }
 cubecl = { workspace = true }
 cubek = { workspace = true, features = ["matmul", "reduce", "quantization"] }
+half = { workspace = true, optional = true }
```
laggui, line 33: f16 / bf16 are re-exported in burn-std and burn-backend, we could re-use those instead.

PR #4205, opened 2025-12-16
fix: handle 1D slope when importing prelu from onnx

mertalev:
## Pull Request Template

The ONNX importer's generated code assumes PRelu nodes always have a static slope. This causes a shape mismatch error when attempting to load the Burn model as it expects [1] rather than the true parameter count. Additionally, the shape of the slope can be something like [64, 1, 1] rather than [64].

### Checklist

### Changes

* Added `with_num_parameters` to the code-gen for parameters > 1
* Flattened slope shape to 1D during import
* Added `PReluConfig` for consistency with how other ops are tested

### Testing

The ONNX model I encountered this issue with loads successfully after making this change.


crates/onnx-ir/src/node/prelu.rs
```diff
@@ -25,18 +25,25 @@
...
+    pub num_parameters: usize,
+}
+
 /// Node representation for PRelu operation
 #[derive(Debug, Clone, NodeBuilder)]
 pub struct PReluNode {
     pub name: String,
     pub inputs: Vec<Argument>,
     pub outputs: Vec<Argument>,
+    pub config: PReluConfig,
```
antimora: `PReluConfig` should be removed.

Node configuration is meant to more ONNX configuration rather Burn's module configurations. 

ONNX-IR is a higher abstraction for ONNX graph. Meaning anyone (outside Burn) can also consume.

crates/onnx-ir/src/node/prelu.rs
```diff
@@ -100,10 +107,24 @@ impl NodeProcessor for PReluProcessor {
...
+            .get(1)
+            .and_then(|slope| {
+                if let ArgType::Tensor(tensor) = &slope.ty {
+                    tensor.static_shape.as_ref()
+                } else {
+                    None
+                }
+            })
+            .map(|shape| shape.iter().product())
+            .unwrap_or(1);
```
antimora: This should be in burn-import layer.

crates/burn-import/src/burn/node/prelu.rs
```diff
@@ -12,17 +12,16 @@ impl NodeCodegen for onnx_ir::prelu::PReluNode {
...
-                let #name = PReluConfig::new()
-                    .init(device);
-            },
-        ))
+        let init = if num_parameters > 1 {
+            let n = num_parameters.to_tokens();
+            quote! { let #name = PReluConfig::new().with_num_parameters(#n).init(device); }
+        } else {
+            quote! { let #name = PReluConfig::new().init(device); }
+        };
```
antimora: Always init with `with_num_parameters(#n)` and don't rely on `PReluConfig`'s defaults so we have a single source.

mertalev:
> Thanks for fixing the problem!
> 
> This is an interesting case where we have Burn configuration but not ONNX. You'll need to arrange your code a bit to keep onnx-ir more Burn independent.
> 
> Would be a great if you could also add an onnx test to avoid a regression in the future.

No problem! I made the changes and added an ONNX test as well.

crates/burn-import/src/burn/node/prelu.rs
```diff
@@ -33,7 +44,9 @@ impl NodeCodegen for onnx_ir::prelu::PReluNode {
         // Alpha (slope) tensor at input index 1
         if let Some(alpha_input) = self.inputs.get(1) {
             let alpha_path = format!("{}.alpha", field_name);
-            if let Some(snapshot) = create_lazy_snapshot(alpha_input, &alpha_path, "PRelu") {
+            if let Some(mut snapshot) = create_lazy_snapshot(alpha_input, &alpha_path, "PRelu") {
+                // Squeeze dimensions to 1D
+                snapshot.shape = vec![num_parameters(self)];
```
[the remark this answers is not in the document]
mertalev: The serializer only uses the raw bytes of the data, not `data.shape`. It uses the snapshot for the shape information.

antimora (approved):
LGTM Thank you for fixing the issues. 

This is a well-implemented bug fix with proper test coverage.

Ready to merge but I'll give a chance for @laggui to review as well. But it looks good from my end.

PR #4207, opened 2025-12-16
fix: handle negative indices in onnx gather op

mertalev:
## Pull Request Template

The ONNX imported code doesn't handle negative indices correctly. It casts -1 to usize, which then overflows when adding + 1.

### Checklist

### Changes

I made it resolve the actual index based on shape inference to avoid adding any runtime overhead. Where the index can't be statically resolved, it falls back to a runtime check.

### Testing

The change works for the the ONNX model where I encountered this issue.

### Related

https://github.com/tracel-ai/burn/pull/3749

### DIFF
--- a/crates/burn-std/src/tensor/slice.rs
+++ b/crates/burn-std/src/tensor/slice.rs
@@ -622,6 +622,12 @@ impl From<i32> for Slice {
     }
 }
 
+impl From<i64> for Slice {
+    fn from(i: i64) -> Self {
+        Slice::index(i as isize)
+    }
+}
+
 impl Display for Slice {
     fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
         if self.step == 1


crates/burn-import/src/burn/node/gather.rs
```diff
@@ -134,6 +135,22 @@ fn forward_scalar_gather(node: &onnx_ir::gather::GatherNode) -> proc_macro2::Tok
...
+/// - The index is a constant scalar value, AND
+/// - If the index is negative, the input has a known static shape at the gather axis
+fn resolve_scalar_index(index_arg: &Argument, input_arg: &Argument, axis: usize) -> Option<usize> {
+    let idx = index_arg.value()?.scalar_i64().ok()?;
+
+    if idx >= 0 {
+        Some(idx as usize)
+    } else {
+        let dim_size = input_arg.ty.static_shape()?.get(axis)?;
+        Some((*dim_size as i64 + idx) as usize)
```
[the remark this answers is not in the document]
mertalev: A negative index like that would be invalid, right? There isn't much to be done about that besides maybe throwing an error during the import.

crates/onnx-ir/src/ir/argument.rs
```diff
@@ -292,4 +292,27 @@ impl Argument {
...
+        let mut tensor_store = TensorStore::new();
+        let data_id = tensor_store.store(data_ref);
+
+        let value_store = ValueStore::new(Rc::new(tensor_store), Rc::new(HashMap::new()));
+
+        Self {
+            name: name.into(),
+            ty: ArgType::Scalar(DType::I64),
+            value_source: ValueSource::Static(data_id),
+            value_store: Some(value_store),
```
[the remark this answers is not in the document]
mertalev: It doesn't seem to really matter whether it uses a shared or isolated store for unit testing purposes.

crates/burn-import/src/burn/node/gather.rs
```diff
@@ -175,22 +192,39 @@ fn forward_tensor_gather(
...
+                        } else {
+                            let index = arg_to_ident(index_arg);
+                            quote! {
+                                let idx = if #index < 0 {
+                                    (#input.dims()[#dim] as i64 + #index) as usize
+                                } else {
+                                    #index as usize
+                                };
+                            }
+                        };
```
laggui: Using `tensor.slice(...)` with the slice macro `s!` should handle negative indexing. So there is no need to handle this explicitly in the codegen. We can replace the old `(#index as usize)..((#index as usize) + 1)` simply with `#index`. It takes care of heterogeneous slice args e.g. `s![.., -1, ..]`

That should simplify the code :slightly_smiling_face:

mertalev: Oh, that's good to know! So do you want me to replace the runtime check path with this, or should I replace the whole thing? I assumed the resolved index would give more optimization opportunities.

laggui: You can leave the resolved index for const scalars, but I don't think you need to handle negatives at that level. `resolve_scalar_index` could always return a signed int, which is valid when used with the `s!` macro.

mertalev: That makes sense, the code is much simpler now :)

laggui:
Looks like we missed i64 for Slice conversion :sweat_smile: 

https://github.com/tracel-ai/burn/blob/8886037d9dd5d7eed23652e747613aba4a1839f0/crates/burn-std/src/tensor/slice.rs#L607-L623

From usize, isize, i32 but not i64 right now... we can add the implementation or use i32

PR #4208, opened 2025-12-16
Perf/improve reduce autotuning + plane non uniform control flow check

nathanielsimard:
Requires: https://github.com/tracel-ai/cubek/pull/33 https://github.com/tracel-ai/cubek/pull/35 https://github.com/tracel-ai/cubecl/pull/1108

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1991,7 +1991,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.9.0-pre.5"
-source = "git+https://github.com/tracel-ai/cubecl?rev=ce6959019fe0cd07df6031335180f0a0d73eee2d#ce6959019fe0cd07df6031335180f0a0d73eee2d"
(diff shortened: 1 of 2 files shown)


crates/burn-cubecl-fusion/src/optim/reduce/tune.rs
```diff
@@ -35,27 +35,69 @@ pub fn fused_reduce_autotune<R: Runtime, BT: CubeElement>(
...
-            })),
+        enum ReduceProps {
+            GreatWithLowReduceCount,
+            GreatWithHighReduceCount,
+            Balanced,
+        }
+
+        for (name, strategy, props) in [
+            (
+                "fused_unit",
```
laggui, line 59: Should the names at least refer to "reduce" (e.g. "fused_reduce_unit")?

nathanielsimard: I don't think so because it is written in the file reduce.json.log

laggui: Fair enough!

PR #4215, opened 2025-12-17
Nest generated code to permit `#[allow(...)]` pragma.

crutcher:
Some existing codegen generated unused variables / statements. Those could be stripped by the generators; but will be stripped by the compiler anyway.

### Checklist


crutcher:
@antimora , @laggui are your feelings about this *specific* issue stronger or weaker than the objection you'd raise if I turned on "all warnings are errors" at compile time?

Because I'm trying to squelch enough warnings that I can sell that.

antimora:
> @antimora , @laggui are your feelings about this _specific_ issue stronger or weaker than the objection you'd raise if I turned on "all warnings are errors" at compile time?
> 
> Because I'm trying to squelch enough warnings that I can sell that.

Yeah I am okay treating warnings as errors but I think it should be enforced at xtask level. This will allow for development proceed till all warnings are removed.

laggui:
I think rejecting all warnings as errors should be handled in the CI, similar to code quality which runs `cargo clippy --no-deps --color=always -- --deny warnings`.

In a "perfect" world, that would be the approach. It enforces consistency without disrupting local workflow, where warnings are useful signals but shouldn't block development (which is iterative).

Preventing merges for such warnings, sure. But preventing iterative development, not great.

laggui:
Will close this PR specifically given the above comments, but feel free to keep the discussion going.

PR #4219, opened 2025-12-17
Unify AsIndex (+ AsShape) handling.

crutcher:
- Simplify / Unify / Broaden AsIndex; introduce AsShape.
- Expand `From<T> for Shape`

### Checklist

### DIFF
--- a/crates/burn-std/src/tensor/shape.rs
+++ b/crates/burn-std/src/tensor/shape.rs
@@ -14,6 +14,8 @@ use core::{
 };
 use serde::{Deserialize, Serialize};
 
+pub use crate::tensor::index_conversion::AsSize;
+
 /// Shape of a tensor.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct Shape {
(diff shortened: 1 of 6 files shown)


crates/burn-std/src/tensor/shape.rs
```diff
@@ -616,70 +618,32 @@ impl DerefMut for Shape {
...
-impl From<Shape> for Vec<usize> {
-    fn from(shape: Shape) -> Self {
-        shape.dims
-    }
-}
+impl<I> ShapeSource for &[I] {}
+impl<I> ShapeSource for &Vec<I> {}
+impl<I> ShapeSource for Vec<I> {}
+impl<I, const D: usize> ShapeSource for [I; D] {}
+impl<I, const D: usize> ShapeSource for &[I; D] {}
```
laggui, line 646: I would remove the marker trait.

I realize this is to restrict the conflicting `From<Shape>` implementations given that `Shape` implements  `IntoIterator`, but I think the correct approach would be to remove the `IntoIterator` implementation for shape instead.

`shape.into_iter()` is already available via the slice (deref impl), so the only thing we lose here is by-value iteration
```rust
for d in shape {
    // ...
}
```

but consuming a shape just to iterate dims is rarely intentional I think.

crutcher: I've rebased/repushed the follow-on PR here:
https://github.com/tracel-ai/burn/pull/4221/changes

See the interaction with the `ReshapeArgs` code; but there are a number of places where it is already useful to treat `Shape` as a stream of usize.

I'd argue that `Shape.into_iter()` should be a preserved operation (and it is used in the reshape code); whereas the `From<X> into Shape` is something we have less-broad expectations for.

If this is a convincing argument; can you just approve that PR (which includes this one)?

crutcher: I see this as a conflict between:
- how widely usable should Shape be, vrs
- how restrictive should `args similar to Shape` be

I'd rather have `Shape` be a good rust citizen, and have the shape-like-args space be a little more restrictive, if that is the tradeoff.

laggui:
Given the cross-reference in discussion, and since changes are also included in the other PR I'll close this one.

PR #4221, opened 2025-12-18
Unify ReshapeArgs / Shape.reshape(args)

crutcher:
Extension to: https://github.com/tracel-ai/burn/pull/4219

### Checklist

- Leverage the AsIndex/AsShape unification to unify `ReshapeArgs`
- Move most of the logic into `Shape::reshape(args)`.

TODO:
- move/duplicate the reshape tests to `burn-std`.

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -106,7 +106,7 @@ pub fn prelu<const D: usize, B: Backend>(
         &alpha.shape()
     ));
 
-    let weight: Tensor<B, D> = if alpha.dims()[0] == 1 {
+    let weight = if alpha.dims()[0] == 1 {
         // if there is only 1 weight, then reshape it to (1,1,1... D times) so that the rank is D
         alpha.reshape([1; D])
     } else {
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -370,7 +370,7 @@ where
     /// ```
     pub fn reshape<const D2: usize, S: ReshapeArgs<D2>>(self, shape: S) -> Tensor<B, D2, K> {
         // Convert reshape args to shape
-        let shape = shape.reshape_check::<D2>(self.shape());
+        let shape = shape.into_shape::<D2>(self.shape());
         Tensor::new(K::reshape(self.primitive, shape))
     }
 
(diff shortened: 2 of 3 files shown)


crutcher:
Discussion about conflict here:
https://github.com/tracel-ai/burn/pull/4219#discussion_r2639886585

crates/burn-std/src/tensor/shape.rs
```diff
@@ -594,70 +657,32 @@ impl DerefMut for Shape {
...
-impl From<Shape> for Vec<usize> {
-    fn from(shape: Shape) -> Self {
-        shape.dims
-    }
-}
+impl<I> ShapeSource for &[I] {}
+impl<I> ShapeSource for &Vec<I> {}
+impl<I> ShapeSource for Vec<I> {}
+impl<I, const D: usize> ShapeSource for [I; D] {}
+impl<I, const D: usize> ShapeSource for &[I; D] {}
```
laggui: Marker traits are for semantics, not for plumbing. `ShapeSource` doesn't encode meaningful properties, it exists solely to restrict the blanket `From<T> for Shape` impl to make the type system happy.

But it is not required, as I mentioned in https://github.com/tracel-ai/burn/pull/4219#discussion_r2639886585

And to answer this specifically:
Also,

> it is already useful to treat `Shape` as a stream of usize.
> 
> I'd argue that `Shape.into_iter()` should be a preserved operation (and it is used in the reshape code); whereas the `From<X> into Shape` is something we have less-broad expectations for.

I'm not arguing against that, and `Shape.into_iter()` is still available via the slice implementation so we do not lose this like I said. We only lose by-value iteration where the container is consumed, but for a shape this wouldn't make sense anyway. Treating it as a stream of usize is still fully possible.

crates/burn-std/src/tensor/shape.rs
```diff
@@ -455,14 +459,73 @@ impl Shape {
     }
 
     /// Reshape this shape to the target shape.
-    pub fn reshape(&self, target: Shape) -> Result<Shape, ShapeError> {
-        if self.num_elements() != target.num_elements() {
-            return Err(ShapeError::IncompatibleShapes {
-                left: self.clone(),
-                right: target,
-            });
+    pub fn reshape<T>(&self, args: T) -> Result<Shape, ExpressionError>
```
laggui: Agree with the move to reduce and reuse the reshape ergonomics, but I think we should stick to the same error type in `Shape` operations.

We should probably  have an `InvalidShape` variant with a reason string to replace `Empty` and generalize to other invalid shapes.

laggui (changes requested):
Continuing the discussion from the other PR on marker trait.

`AsIndex` and `AsSize` changes still good to go otherwise.

crates/burn-std/src/tensor/shape.rs
crutcher: I've not been able to break the cycle between ShapeSource and RankedReshapeArgsSource and Shape::reshape.

If you know how to do this, can you update the PR, or offer a branch PR?

crates/burn-std/src/tensor/shape.rs
crutcher: Ok, but there are cases where it isn't an invalid shape.

The shape expression supports `-1` and `0`; and what is invalid is it can't be resolved effectively. If you want a structured error in that situation, it can't hold a Shape, or even a Vec<usize>. It needs either a Vec<isize>, or a message String.

crates/burn-std/src/tensor/shape.rs
laggui: Commit [56b5821](https://github.com/tracel-ai/burn/pull/4221/commits/56b58210d8fb7b62f59b7b2873c7969eea73bfff) should have fixed the marker traits. I just need to update the `tensor.reshape` API again to keep it compatible. Will do that a bit later.

PR #4223, opened 2025-12-18
Burn train api refactor

Charles23R:
### Checklist

### Related Issues/PRs

### Changes

### Testing

### DIFF
--- a/crates/burn-train/src/learner/supervised/paradigm.rs
+++ b/crates/burn-train/src/learner/supervised/paradigm.rs
@@ -178,7 +178,6 @@ impl<SC: SupervisedLearningComponentsTypes> SupervisedTraining<SC> {
         strategy: <SC::PC as ParadigmComponentsTypes>::CheckpointerStrategy,
     ) -> Self
     where
-        // CS: CheckpointingStrategy + 'static,
         <SC::PC as ParadigmComponentsTypes>::CheckpointerStrategy: 'static,
     {
         self.checkpointer_strategy = Box::new(strategy);
--- a/crates/burn-train/src/metric/processor/minimal.rs
+++ b/crates/burn-train/src/metric/processor/minimal.rs
@@ -9,7 +9,6 @@ use std::sync::Arc;
 ///   - Computing and storing metrics in an [event store](crate::metric::store::EventStore).
 #[allow(dead_code)]
 #[derive(new)]
-// TODO : DELETE
 pub(crate) struct MinimalEventProcessor<T: ItemLazy, V: ItemLazy> {
     metrics: MetricsTraining<T, V>,
     store: Arc<EventStoreClient>,


crates/burn-train/src/learner/supervised/paradigm.rs
```diff
@@ -149,9 +173,13 @@ where
...
 
     /// Update the checkpointing_strategy.
-    pub fn with_checkpointing_strategy<CS>(mut self, strategy: CS) -> Self
+    pub fn with_checkpointing_strategy(
+        mut self,
+        strategy: <SC::PC as ParadigmComponentsTypes>::CheckpointerStrategy,
+    ) -> Self
     where
-        CS: CheckpointingStrategy + 'static,
+        // CS: CheckpointingStrategy + 'static,
```
nathanielsimard: deadcode

crates/burn-train/src/learner/supervised/paradigm.rs
```diff
@@ -366,133 +401,104 @@ where
...
-        if let LearningStrategy::MultiDevice(devices, _) = &learning_strategy
-            && devices.len() == 1
-        {
-            return LearningStrategy::SingleDevice(devices[0].clone());
+        match self.training_strategy {
+            TrainingStrategy::SingleDevice(device) => {
+                let single_device: SingleDevicetrainingStrategy<SC> =
+                    SingleDevicetrainingStrategy::new(device);
+                single_device.train(
+                    // self.learner,
```
nathanielsimard: deadcode

crates/burn-train/src/metric/processor/minimal.rs
```diff
@@ -9,6 +9,7 @@ use std::sync::Arc;
 ///   - Computing and storing metrics in an [event store](crate::metric::store::EventStore).
 #[allow(dead_code)]
 #[derive(new)]
+// TODO : DELETE
```
nathanielsimard: Still valid?

crates/burn-train/src/learner/supervised/paradigm.rs
```diff
@@ -17,90 +11,107 @@ use crate::metric::processor::{
...
-    M::InnerModule: ValidStep<VI, VO>,
-    O: Optimizer<M, B>,
-    S: LrScheduler,
-    TI: Send + 'static,
-    VI: Send + 'static,
-    TO: ItemLazy + 'static,
-    VO: ItemLazy + 'static,
-{
+/// Structure to configure and launch supervised learning trainings.
+pub struct SupervisedTraining<SC: SupervisedLearningComponentsTypes> {
```
laggui, line 36: I'm not sure if `SupervisedTraining` might be too restrictive, given that it might be possible to leverage the same training loop even in other contexts (e.g., self-supervised or unsupervised which still have similar training loop, just the signal is different). I know they all mention "supervised" but people generally make the distinction :smile: 

But we have time to think about this before introducing this breaking change in a major release.

Otherwise, LGTM!!

PR #4225, opened 2025-12-18
Add tracing feature flag

laggui:
### Checklist

### Changes

Added `tracing` feature flag so it is not directly tied to std usage

### DIFF
--- a/crates/burn-autodiff/Cargo.toml
+++ b/crates/burn-autodiff/Cargo.toml
@@ -15,15 +15,15 @@ version.workspace = true
 workspace = true
 
 [features]
-default = ["std"]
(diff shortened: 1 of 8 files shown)


crutcher:
Can I argue against doing this? I've been working on exploiting the `tracing/max_level_???` machinery that is already built into `tracing`.

Please take a look at the example/readme/Cargo.toml in:
https://github.com/tracel-ai/cubecl/pull/1112

laggui:
Copying my response from discord since you cross-posted..

> I still think having tracing as an optional dependency is desirable. `tracing/max_level_*` still requires tracing to exist in the dep graph and the macro will still expand (but just compile to no-ops) if I understand correctly.
>
> So they are orthogonal in mechanism, but complementary when used together.

PR #4230, opened 2025-12-19
Fix panic for Expand onnx operation with scalar input

GregorySech:
This pull request fixes a panic that occurred when the ONNX Expand operation receives a scalar input instead of a tensor. The fix properly handles scalar inputs by converting them to rank-1 tensors during code generation.

### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn-onnx/issues/7

### Changes

- Modified type inference in `ExpandProcessor::infer_types` to accept scalar inputs and propagate their data types
- Updated code generation in `ExpandNode::forward` to convert scalar inputs to rank-1 tensors with appropriate tensor kinds (Float, Int, Bool)
- Added comprehensive test coverage including unit tests and an end-to-end ONNX model test

### Testing

1. burn-import/onnx-tests/tests/expand/expand_scalar.py generates expand_scalar.onnx, an ONNX graph in the fashion of expand_tensor but with a scalar as input. 
2. The expand_scalar test checks that the output of the model generated from expand_scalar.onnx has the correct shape.
3. Code gen tests


GregorySech:
I created this pull request to get a bit of feedback. I think I should extract the logic to create the temporary Tensor for the scalar.
I'm also unsure if it's okay to just use' burn::tensor::Int' for all those different specialisations of' int'. Same for floating point types.

laggui:
CC'ing @antimora since you self-assigned the linked issue.

antimora:
> I created this pull request to get a bit of feedback. I think I should extract the logic to create the temporary Tensor for the scalar. I'm also unsure if it's okay to just use' burn::tensor::Int' for all those different specialisations of' int'. Same for floating point types.

I think for this one it's better if make changes on your PR directly because there are a few things to consider. This is related to tensor creation with dtype

GregorySech:
> > I created this pull request to get a bit of feedback. I think I should extract the logic to create the temporary Tensor for the scalar. I'm also unsure if it's okay to just use' burn::tensor::Int' for all those different specialisations of' int'. Same for floating point types.
> 
> I think for this one it's better if make changes on your PR directly because there are a few things to consider. This is related to tensor creation with dtype

@antimora let me know if you have any directions that I should implement. 
I guess that tensor creation from a dtype based on a scalar might be somewhat useful for other ops.

antimora:
@GregorySech sorry I thought I could handle it before my travels (be back on 20th). 

Recently I have discovered that tensor creation from native value was defaulting to backends default dtype. You need to make sure you pass scalars dtype when creating a tensor from native scalar. You can find some existing examples. @laggui is also aware of this issue. 

Please proceed with this PR and assume I am not working on it. I will have access to my computer after Jan 20th. 

Thank you for your contribution.

GregorySech:
Hi, wait for the merge, i need to push a fix to handle the corner case where onnx cannot infer the rank of the output but the input is scalar. Currently it’s still panicing but it can be handled

crates/burn-onnx/src/burn/node/expand.rs
```diff
@@ -15,37 +15,57 @@ impl NodeCodegen for onnx_ir::expand::ExpandNode {
...
+                _ if dtype.is_float() => quote! {},
+                _ => quote! { , Int },
+            };
+            let init = quote! {
+                let input = Tensor::<B, 1 #kind>::from_data_dtype(
+                    burn::tensor::TensorData::from([#input]),
+                    &*self.device,
+                    #dtype_tokens
+                );
+            };
```
laggui, line 35: Why not use `Tensor::<B, 1, #kind>::full` instead? Just a small suggestion.

laggui (changes requested):
> Hi, wait for the merge, i need to push a fix to handle the corner case where onnx cannot infer the rank of the output but the input is scalar. Currently it’s still panicing but it can be handled

Sounds good! Will request changes just to block any accidental merge :smile: 

Otherwise the current changes LGTM

antimora:
closing this in favor of https://github.com/tracel-ai/burn-onnx/pull/41. Now all onnx related stuff: https://github.com/tracel-ai/burn-onnx

PR #4233, opened 2025-12-19
Remove linux-only constraint for cpu

louisfd:
Tested on macos

### DIFF
--- a/crates/burn-backend/src/lib.rs
+++ b/crates/burn-backend/src/lib.rs
@@ -73,7 +73,7 @@ mod cube_cuda {
     impl DeviceOps for CudaDevice {}
 }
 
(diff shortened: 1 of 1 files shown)


crates/burn-backend/src/lib.rs
```diff
@@ -73,7 +73,7 @@ mod cube_cuda {
     impl DeviceOps for CudaDevice {}
 }
 
-#[cfg(all(feature = "cubecl-cpu", target_os = "linux"))]
+#[cfg(all(feature = "cubecl-cpu"))]
```
laggui:
```suggestion
#[cfg(feature = "cubecl-cpu")]
```

laggui (approved):
Let's gooooo

PR #4236, opened 2025-12-21
fix: grid sample using excessive memory

mertalev:
## Pull Request Template

The int_expands and float_expands caused memory usage to explode: it failed trying to allocate 10.5GB (1 * 3 * 640 * 480 * 2859 elements).

### Checklist

### Changes

Changes the logic to use flattened indexing instead of a huge expand.

### Testing

I no longer get an OOM crash with this change, and the results are equivalent given the tests pass.

### DIFF
--- a/crates/burn-backend/src/backend/ops/modules/grid_sample.rs
+++ b/crates/burn-backend/src/backend/ops/modules/grid_sample.rs
@@ -11,7 +11,7 @@ use burn_std::{Shape, Slice};
 ///
 /// # Arguments
 ///
-/// * `tensor` - The tensor being sampled from, shape (N, C, H_in, W_in)
+/// * `tensor` - The tensor being sampled from, must be contiguous with shape (N, C, H_in, W_in)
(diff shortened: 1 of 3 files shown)


mertalev:
This is not very relevant to the PR, but out of curiosity, I had Claude write a dedicated CubeCL wgpu kernel for grid sampling (not included in this PR) to see what kind of impact this would have over the generic tensor implementation. The interaction with the `fusion` feature is very interesting (and somewhat surprising):

  | Config                | Execution time (sum of 30 async iterations) |
  |----------------------|--------------------|
  | Reference, no fusion| 28.738ms              |
  | Reference, fusion    | 17.341ms              |
  | Kernel, no fusion    | 6.415ms              |
  | Kernel, fusion       | 17.657ms              |

I had to triple check the features were set correctly. The performance benefit of fusion is real, but it seems to significantly hurt the specialized kernel. Is this a known characteristic of `fusion`? This is on an M4 Max, for reference.

Edit: Oh, the cause is that the fusion backend takes a different path where the default implementation calls the reference one. The IR needed to be updated to use the specialized version.

laggui (approved):

PR #4242, opened 2025-12-22
feat: cubecl grid sample kernel

mertalev:
### Checklist

### Changes

This PR adds the CubeCL kernel to improve performance over the reference implementation. It adds `GridSample2d` to the IR as well.

As part of this, I added a reflection test to confirm parity with PyTorch and learned that the existing reflection handling was incorrect (especially for `align_corners=false`). I updated all implementations to use a triangle wave formula to get matching results with PyTorch now.

### Testing

A project that uses Burn as a dependency builds and outputs correctly with this PR. I also ran a microbenchmark of 100 calls with one sync on an M4 Max with the following results:

|                    | Min (ms) | Max (ms) | Median (ms) | Avg (ms) |
  |--------------------|----------|----------|-------------|----------|
  | Reference (fusion) | 1.661    | 3.731    | 1.751       | 1.776    |
  | Kernel             | 0.344    | 1.116    | 0.376       | 0.414    |

I tested a few different implementation strategies and this performed the best on my M4 Max. However, I have not tested performance on any other hardware.

I used Claude Opus 4.5 to navigate the codebase, implement the kernel with different strategies and write tests, while I benchmarked, optimized and refactored the code.

### DIFF
--- a/crates/burn-cubecl/src/kernel/grid_sample/base.rs
+++ b/crates/burn-cubecl/src/kernel/grid_sample/base.rs
@@ -76,7 +76,9 @@ pub(crate) fn fetch_with_zeros<F: Float>(
     w: i32,
 ) -> F {
     let in_bounds = x >= 0 && x < w && y >= 0 && y < h;
-    let idx = base + (y as u32) * stride_h + (x as u32) * stride_w;
+    let x_clamped = Min::min(Max::max(x, 0), w - 1) as u32;
+    let y_clamped = Min::min(Max::max(y, 0), h - 1) as u32;
+    let idx = base + y_clamped * stride_h + x_clamped * stride_w;
     select(in_bounds, input[idx], F::new(0.0))
 }
 


crates/burn-cubecl/src/kernel/grid_sample/base.rs
```diff
@@ -0,0 +1,161 @@
...
+    stride_h: u32,
+    stride_w: u32,
+    y: i32,
+    x: i32,
+    h: i32,
+    w: i32,
+) -> F {
+    let in_bounds = x >= 0 && x < w && y >= 0 && y < h;
+    let idx = base + (y as u32) * stride_h + (x as u32) * stride_w;
+    select(in_bounds, input[idx], F::new(0.0))
```
laggui, line 82: `select` executes both branches, and then selects a value based on the condition. So both branches must be valid. Out-of-bounds accesses should not be done here.

Docs:
```
Executes both branches, then selects a value based on the condition. This should be branchless, but might depend on the compiler.

# Safety
Since both branches are evaluated regardless of the condition, both branches must be valid regardless of the condition. Illegal memory accesses should not be done in either branch.
```

You could clamp the x / y values like you did in `fetch_with_border`.

laggui (changes requested):
Just a small fix needed, otherwise LGTM

Thanks for the contribution!

laggui (approved):

crates/burn-cubecl/src/kernel/grid_sample/base.rs
mertalev: Good to know that's UB! I clamped it as suggested

PR #4248, opened 2025-12-24
docs: add examples for bool tensor operations

qburke:
## Summary

Added documentation examples for four boolean tensor operations in `burn-tensor`:
- `bool_not` - Boolean inversion
- `bool_and` - Logical AND operation  
- `bool_or` - Logical OR operation
- `bool_xor` - Logical XOR operation

## Changes

Added doc examples with:
- Clear descriptions of what each operation does
- Arguments and Returns sections
- Code examples
- Expected output in comments

Contributes to #2450

### DIFF
--- a/crates/burn-tensor/src/tensor/api/bool.rs
+++ b/crates/burn-tensor/src/tensor/api/bool.rs
@@ -44,15 +44,9 @@ where
     ///
     /// fn example<B: Backend>() {
     ///     let device = Default::default();
-    ///     let tensor = Tensor::<B, 2, Bool>::from_bool(
-    ///         [[true, false, true],
-    ///          [false, true, false]].into(),
-    ///         &device,
-    ///     );
+    ///     let tensor = Tensor::<B, 2, Bool>::from_bool([[true, false], [false, true]].into(), &device);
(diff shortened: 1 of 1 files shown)


crates/burn-tensor/src/tensor/api/bool.rs
```diff
@@ -35,21 +35,141 @@ where
...
+    ///     let b = Tensor::<B, 2, Bool>::from_bool(
+    ///         [[true, false],
+    ///          [true, false]].into(),
+    ///         &device,
+    ///     );
+    ///     let result = a.bool_and(b);
+    ///     println!("{result}");
+    ///     // [[true, false],
+    ///     //  [false, false]]
+    /// }
```
laggui, line 78: Just to make the examples a bit more compact, we can fix the formatting following cargo fmt

```rust
fn example<B: Backend>() {
    let a = Tensor::<B, 2, Bool>::from_bool([[true, true], [false, false]].into(), &device);
    let b = Tensor::<B, 2, Bool>::from_bool([[true, false], [true, false]].into(), &device);
    let result = a.bool_and(b);
    println!("{result}"); // [[true, false], [false, false]]
}
```

for all examples

PR #4250, opened 2025-12-26
Implement QR decomposition

RustyBamboo:
### Checklist

### Related Issues/PRs

#1538

### Changes

Would be nice to have more linalg operations out-of-the-box in burn. This adds a householder QR decomposition.

Also replaced the Gram-Shmidt method with a call to this one in `initializer` module.

### Testing

Generated examples in python and decomposed them via `numpy.linalg.qr`. Then compared burn implementation against the expected output from numpy (up to a degree-of-freedom in sign). Also check that our `Q` is indeed orthogonal.

### Notes

- There is a lot of syncing here with host and eventually will benefit from a block-based QR decomposition or something lower-level
- Can't be autodiffed
- Wasn't sure what the best way to check against numerical precession... mimicked safemin from LAPACK

### DIFF
--- a/crates/burn-tensor/src/tensor/linalg/qr_decomposition.rs
+++ b/crates/burn-tensor/src/tensor/linalg/qr_decomposition.rs
@@ -1,0 +1,129 @@
+use alloc::vec::Vec;
+
+use crate::{
+    DType, Element, ElementConversion, backend::Backend, cast::ToElement, linalg::outer, s,
+    tensor::Tensor,
+};
+
+struct Householder<B: Backend> {
+    v: Tensor<B, 1>,
+    tau: B::FloatElem,
+}
+
+fn eps_for_dtype(dtype: DType) -> f64 {
+    match dtype {
+        DType::F16 => 1e-3,
+        DType::BF16 => 1e-2,
+        DType::F32 => 1e-7,
+        DType::F64 => 1e-15,
+        _ => 1e-7,
+    }
+}
+
+/// Performs QR decomposition of a matrix using Householder reflections.
+///
+/// The input matrix `A` is factored into `Q` and `R` such that `A = Q * R`,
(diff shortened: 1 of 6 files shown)


crates/burn-tensor/src/tensor/linalg/qr_decomposition.rs
```diff
@@ -0,0 +1,129 @@
...
+/// # Returns
+///
+/// A tuple containing:
+/// - `Q`: a matrix of shape `[m, k]`
+/// - `R`: a matrix of shape `[k, n]`
+///
+/// where `m` and `n` are the input dimensions and `k = min(m, n)`.
+pub fn qr_decomposition<B: Backend>(tensor: Tensor<B, 2>) -> (Tensor<B, 2>, Tensor<B, 2>) {
+    let device = tensor.device();
+    let [m, n] = tensor.shape().dims::<2>();
```
laggui, line 37: Could use the short-hand `tensor.dims::<2>()` instead

laggui (commented):
> - There is a lot of syncing here with host and eventually will benefit from a block-based QR decomposition or something lower-level

In the meantime, we should add a performance note similar to the one in LU decomposition:
https://github.com/tracel-ai/burn/blob/8bd616f107cb49a45195ee83b07d03b3d12735fa/crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs#L25-L27

- Wasn't sure what the best way to check against numerical precession... mimicked safemin from LAPACK 

For tests, you correctly used the `Tolerance` in checks. But unsure what is the typical error tolerance for such algos.

Also, please fix the formatting issue as pointed out by the CI. You can just use cargo fmt.

laggui:
Closing due to inactivity, but feel free to re-open if you want to continue this at some point.

Issue #3712, opened 2025-09-14
Relax Element Trait constraints


skewballfox:
### Feature description

Somehow Relax the constraints for the Element trait to make it possible to define elements outside of burn-tensor

### Feature motivation

This came up primarily in #3608 in the process of trying to move complex element definitions to the crate `burn-complex`. The current major blocker is `ElementConversion`, ~~which has explicit function calls per primitive type~~ which relys on [`ToElement`](https://github.com/tracel-ai/burn/blob/dd2a02905ea4a29ce2100e432e49d63b365546ab/crates/burn-tensor/src/tensor/element/cast.rs#L18) which has explict calls per type. This made it impossible to define an Element outside of `burn-tensor`. `ElementComparison` I believe depends on `ElementConversion`.

This currently blocks implementing `basicOps` (and thus quite a few other traits) on anything outside of burn tensor if the underlying dtype is not one of the already defined and supported primitives.

### (Optional) Suggest a Solution

completely spitballing:
- remove the ElementConversion, ElementComparision constraint from element. Make the basic ops that require members of a new trait (maybe `InteropOps`? lol). Really just shuffles the problem around though.
- find a way to rework the functions in element conversion to more like the signature of mem::transmute (i.e. `convert<E1,E2>(E1)-> E2` where both are Elements, but elements are not required to cover every implementor. The core primitives will cover all the other primitives in burn-tensor. You can have 2 variants, convert which is fast but unsafe in that it will panic if the required conversion method doesn't exists, and then `try_convert<E1,E2>(E1)-> Result<E2>` that will throw an error. This might be tricky converting external elements from burn-tensor elements since that would probably violate the orphan rule
- get rid of ElementConvert, and make the ops that currently require it require a From<E>, Into<E> specific to the 2 elements in question (the current element and  product element).

I'm open to other possibilities, and if there are downsides or tradeoffs to the above approaches I'm not seeing, please let me know.

skewballfox:
from the discord, crutcher mentioned a few situations where relaxing the element constraints could allow us to represent useful primitives: unsigned variants, log-space, value plus complex confidence. frequency is apparently useful for spiking nets. He also mentioned "complex but values stored as dims", but I'm not how that might be representable as an individual element

laggui:
Cross-posting my comment here as well:

[...]

Issue #3265, opened 2025-06-07
Complex numbers


ifsheldon:
### Feature description

Support for complex numbers in tensors.

### Feature motivation

I saw your presentation on Scientific Computing in Rust and come to know burn. As I skim through the doc and issue list, it seems there's no complex number support yet. I'm writing some code for quantum computing simulation, and it'd be super cool to writing it in Rust, but I need complex numbers.

### (Optional) Suggest a Solution

Perhaps we can add the most simplistic support for it by treating a complex number as a 2D vector with specific ops.

prakash-shekhar:
Trying to do something similar. Anybody mind if I work on this?

prakash-shekhar:
Seeing two options on how to do this:

1. Treat complex as a special case of float tensors and basically reuse the existing FloatKind and FloatTensorPrimitive, and just handle complex-specific behavior in the ops.
2. Make complex a first-class kind and introduce a ComplexKind, add ComplexTensorPrimitive and ComplexTensorOps. It’s more upfront work for backends, but keeps type system clean.

Any thoughts on what would be better @laggui ?

laggui:
Complex numbers have different semantics than floats. I would opt for a different kind (option 2).

skewballfox:
Since there are multiple possible ways of implementing complex tensors (mentioned [here](https://github.com/tracel-ai/burn/issues/788#issuecomment-1999171154)), and we want to support at least 2 of them (complex as a real and imaginary tensor and a tensor of complex primitives), what naming scheme for the types would make it clear which is which? 

I'm guessing we don't want to wrap it in an enum but rather treat them as distinct kinds that support the same ops, partially because I suspect they won't be used interchangeably.

skewballfox:
once #3608 is merged, I think we should close this and then track the remaining progress via #5366

antimora:
Marked blocked on #3608. Once it merges, this can close in favor of the tracking issue #5366.

PR #4255, opened 2025-12-27
Make `ElementComparison` optional for dtypes

skewballfox:
So for making custom element types, not every type will have all the current properties required of elements. An example of this is the fact that complex numbers don't have a linear ordering. Outside of ordering alone, making it to where groups of ops are optional also lowers the amount of initial work necessary to add a new data type, or to make a new variant of an existing data type. 

### Checklist

### Related Issues/PRs

- #3712 
- reduces work for #3265 

### Changes

- I separated out elementComparison from element (may rename to `ElementOrdering`) from Element
- I added a new trait to Element for equality checks
- I split out all methods from numeric that require ordering into an Orderable trait
    - right now, mainly due to one-hot functions, I have made it a supertrait of numeric, i.e. in order for a element tensor to be orderable, it has to implement numeric. I think though I could either rework onehot to not require E to be orderable, or to somehow make onehot it's own thing that's only an option if both traits are implemented. 
 - I think I need to make some changes to how libtorch is structured to be closer to the API of the other backends. I was able to get this to build by disabling torch altogether, though I still need to check whether test pass.

### Testing

confirmed the project builds, still need to run the actual checks

### DIFF
--- a/crates/burn-backend/src/element/base.rs
+++ b/crates/burn-backend/src/element/base.rs
@@ -9,7 +9,10 @@ use burn_std::flex32;
 
 use super::cast::ToElement;
 
-/// Element trait for tensor.
+/// Core element trait for tensor values.
+///
+/// This trait defines the minimal set of capabilities required for a type to be
+/// stored and manipulated as a tensor element across all backends.
 pub trait Element:
     ToElement
     + ElementRandom
@@ -31,7 +34,13 @@ pub trait Element:
     fn dtype() -> DType;
 }
 
-/// Ordered element trait for tensor.
+/// Ordered element trait for tensor values.
+///
+/// This trait extends [`Element`] with ordering semantics, enabling comparison
+/// and order-dependent operations in generic Rust implementations.
+///
+/// Backends that implement these operations entirely at the device level do
+/// not rely on this trait. It only constrains the scalar type for generic Rust code.
 pub trait ElementOrdered: Element + ElementComparison {}
 
 /// Element conversion trait for tensor.
@@ -78,7 +87,7 @@ pub trait ElementComparison {
     fn cmp(&self, other: &Self) -> Ordering;
 }
 
-/// Element ordering trait.
+/// Element limits trait.
 pub trait ElementLimits {
     /// The minimum representable value
     const MIN: Self;


crates/burn-backend/src/backend/base.rs
```diff
@@ -83,12 +83,12 @@ pub trait Backend:
     /// Tensor primitive to be used for all float operations.
     type FloatTensorPrimitive: TensorMetadata + 'static;
     /// Default float element type.
-    type FloatElem: Element;
+    type FloatElem: Element + ElementComparison;
```
laggui: iirc `ElementComparison` was added specially for sorting ops, which have a default cpu implementation that operates on `TensorData` elements. We could likely remove the stricter `ElementComparison` bound, and only require `ElementEquality`, except for the data sorting implementation

https://github.com/tracel-ai/burn/blob/a969aad43512d1e8228c0a2b500795441b9a1073/crates/burn-backend/src/backend/ops/sort.rs#L370-L373

Would have to check, but I believe everything else should still compile and work.

laggui (commented):
Splitting `Orderable` from `Numeric` makes sense!

`Backend` implementations will still require float/int ops with orderable methods (e.g., `float_greater`) but can be left out as unimplemented, and for custom tensor kinds the API can be restricted so as to not expose the high-lever tensor methods that way. 

Not sure why the changes to `LibTorch` elements were necessary for these changes though :thinking: we could accept different int elem types if we want to extend support, but unclear why it's needed for this PR specifically

And maybe `Ordered` could be a better name instead of `Orderable`?

crates/burn-backend/src/backend/base.rs
skewballfox: > iirc ElementComparison was added specially for sorting ops, which have a default cpu implementation that operates on TensorData elements. We could likely remove the stricter ElementComparison bound, and only require ElementEquality, except for the data sorting implementation

right now, `{Float,Int}TensorOps` bundles methods that rely on both `Orderable` and `Numeric`, hence the added constraint. How would removing the bound work? Would we just remove the ordering ops from `{Float,Int}TensorOps` traits and then leave it to the stuff defined under orderable, tensorData sort impl, or would we take a similar approach to what was done to numeric and define a float,int ordering trait that would be implemented directly on the end element types (just moving the existing methods to  a new home)? or were you thinking of another approach entirely?

skewballfox:
> Splitting `Orderable` from `Numeric` makes sense!

What should be done about the onehot methods that need both?

> `Backend` implementations will still require float/int ops with orderable methods (e.g., `float_greater`) but can be left out as unimplemented, and for custom tensor kinds the API can be restricted so as to not expose the high-lever tensor methods that way.

sort of. The first approach I took you would have issues converting from element types without ordering to element types with ordering. I left ordering off of bool tensors mainly to test that conversion still worked.

> Not sure why the changes to `LibTorch` elements were necessary for these changes though 🤔 we could accept different int elem types if we want to extend support, but unclear why it's needed for this PR specifically

I can remove the constraint for int (since ElementComparison is already implemented for i64), but I had to add a type for to make it so that other element types didn't have the additional constraint applied to floats

> And maybe `Ordered` could be a better name instead of `Orderable`?

sure

skewballfox:
@laggui I think this is ready for review.

I can try to remove the ordering contstraint entirely from the backend float trait, but I could use pointers as to what would be required. Also, while I get sorting being cpu only what about clamping (limiting values to ranges)?

laggui:
> @laggui I think this is ready for review.

Awesome! Will self-assign to make sure I review the PR this week.

> I can try to remove the ordering contstraint entirely from the backend float trait, but I could use pointers as to what would be required. Also, while I get sorting being cpu only what about clamping (limiting values to ranges)?

Sorting is CPU-only solely for backends that don't override the default implementation, which operates directly on the `TensorData`. So that's where the constraint came from originally: the default implementation on elements in `TensorData`. That's why I said it's not required for the `Backend` element specifically, it was introduced globally but did not need to. All ordered requirements should be described/required by the trait that you just added imo, elements don't need to propagate the ordering requirement. In short: `FloatElem` should only require `Element`, no need to have `+ <ordered element trait>`. Hopefully that clears things up a bit :sweat_smile:

skewballfox:
@laggui  Sounds good to me

Issue #1370, opened 2024-02-26
Asymmetric padding is not supported


AdrianEddy:
**Describe the bug**
I'm trying to run the face detector and landmarker model from [MediaPipe](https://developers.google.com/mediapipe/solutions/vision/face_landmarker).
I first converted the `face_detector.tflite` to ONNX by running `python -m tf2onnx.convert --tflite face_detector.tflite --output face_detector.onnx --opset 15` and I'm trying to run the resulting ONNX models in Burn

When trying to run the ONNX to Burn conversion, I'm getting this log and a panic:
```
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   WARN burn_import::onnx::dim_inference: Must implement dimension inference for Pad
   INFO burn_import::onnx::from_onnx: Finished parsing ONNX file: src/model/face_detector.onnx
  DEBUG burn_import::burn::graph: Registering node => 'transpose'
  ERROR burn_import::logger: PANIC => panicked at ~\.cargo\registry\src\index.crates.io-6f17d22bba15001f\burn-import-0.12.1\src\onnx\op_configuration.rs:493:9:
  Asymmetric padding is not supported
```

build.rs contents:
```rust
use burn_import::onnx::ModelGen;
fn main() {
    ModelGen::new()
        .input("src/model/face_detector.onnx")
        .out_dir("model/")
        .run_from_script();
}
```

**To Reproduce**
Models I'm trying to run: 
[face_landmarker.zip](https://github.com/tracel-ai/burn/files/14410513/face_landmarker.zip)

**Expected behavior**
The conversion succeeds and I can run the model

**Desktop (please complete the following information):**
 - OS: Windows 11, Rust 1.76

[...]

Issue #2676, opened 2025-01-09
Support asymmetric padding for "same" padding with even kernel size


dfsfdfse:
**Describe the bug**
PaddingConfig::Same doesn't return the same input shape when kernel_size = [2, 2]

**Additional context**
```rust
use burn::backend::Wgpu;
use burn::backend::wgpu::WgpuDevice;
use burn::tensor::Distribution;

let device = WgpuDevice::BestAvailable;
let data: Tensor<Wgpu, 4> = Tensor::random([1, 3, 16, 16], Distribution::Default, &device);
let pool = MaxPool2dConfig::new([2, 2]).with_padding(PaddingConfig2d::Same).init();
let v = pool.forward(data);
println!("{:?}", v.shape());
```

laggui:
Thanks for flagging!

For even kernel sizes the "same" padding scheme actually requires asymmetric padding in order to produce the same output shape.

We only support symmetric padding right now, might be a good idea to add a check in the meantime 😅 

If you try a kernel size of 2, 4, 6 or any other even number with same padding I believe you will observe the same behavior.

antimora:
Fix submitted: https://github.com/tracel-ai/burn/pull/4263

PR #4263, opened 2025-12-29
[BREAKING] Add asymmetric padding support for conv and pool operations

antimora:
- Adds support for asymmetric (explicit) padding in Conv1d, Conv2d, Conv3d, MaxPool1d, MaxPool2d, AvgPool1d, and AvgPool2d operations
- Extends PaddingConfig1d and PaddingConfig2d with Explicit variants that specify per-side padding values

### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn-onnx/issues/5
Fixes https://github.com/tracel-ai/burn/issues/1370
Related https://github.com/tracel-ai/burn/issues/2676
ONNX-side changes: https://github.com/tracel-ai/burn-onnx/pull/45

### Changes

burn-nn

- Extended PaddingConfig1d with Explicit(left, right) variant
- Extended PaddingConfig2d with Explicit(top, left, bottom, right) variant
- Added forward pass implementations for asymmetric padding in all conv and pool modules
- Added comprehensive unit tests for asymmetric padding

### Testing

- Unit tests for asymmetric padding in burn-nn (Conv1d, Conv2d, MaxPool1d, MaxPool2d, AvgPool1d, AvgPool2d)

### Breaking / Migration

PaddingConfig's Explicit variant expanded. For a quick migration of existing code, duplicate each padding value:
  - Explicit(n) -> Explicit(n, n)
  - Explicit(h, w) -> Explicit(h, w, h, w)
  - Explicit(d, h, w) -> Explicit(d, h, w, d, h, w)

### Follow-up / left-overs
* https://github.com/tracel-ai/burn/issues/4362

### DIFF
--- a/crates/burn-cubecl/src/kernel/conv/backward_data/tune.rs
+++ b/crates/burn-cubecl/src/kernel/conv/backward_data/tune.rs
@@ -115,7 +115,6 @@ fn create_key<R: CubeRuntime, const N: usize>(
         padding,
         dilation,
         groups,
-        ..
     } = options.clone();
 
     let lhs_stride_align = if out_grad.strides[dim_c] == 1 {
--- a/crates/burn-cubecl/src/kernel/conv/backward_weight/tune.rs
+++ b/crates/burn-cubecl/src/kernel/conv/backward_weight/tune.rs
@@ -118,7 +118,6 @@ fn create_key<R: CubeRuntime, const N: usize>(
         padding,
         dilation,
         groups,
-        ..
     } = options.clone();
 
     let lhs_stride_align = if out_grad.strides[dim_c] == 1 {
--- a/crates/burn-cubecl/src/kernel/conv/forward/tune.rs
+++ b/crates/burn-cubecl/src/kernel/conv/forward/tune.rs
@@ -117,7 +117,6 @@ fn create_key<R: CubeRuntime, const N: usize>(
         padding,
         dilation,
         groups,
-        ..
     } = options.clone();
 
     let lhs_stride_align = if input.strides[dim_c] == 1 {


crates/burn-nn/src/modules/conv/conv2d.rs
```diff
@@ -168,16 +168,33 @@ impl<B: Backend> Conv2d<B> {
...
+                &self.kernel_size,
+                &self.stride,
+            );
+            conv2d(
+                input,
+                self.weight.val(),
+                self.bias.as_ref().map(|bias| bias.val()),
+                ConvOptions::new(self.stride, padding, self.dilation, self.groups),
+            )
+        }
```
laggui: I think this should be applied at the functional `conv1d` level, otherwise there is a discrepancy when calling the module function directly. Same for the other nn module changes in the PR.

And perhaps we should add a note that asymmetric padding involves an explicit pad operation.

crates/burn-nn/src/padding.rs
```diff
@@ -25,7 +30,29 @@ impl PaddingConfig1d {
         match self {
             Self::Valid => 0,
             Self::Same => same_padding(),
-            Self::Explicit(value) => *value,
+            Self::Explicit(left, right) => {
+                if left != right {
+                    panic!("Asymmetric padding should be handled separately via is_asymmetric()")
+                }
+                *left
+            }
```
laggui: This only handles *explicit* asymmetric padding, but when `Same` padding is used with even kernel sizes the resulting padding could be asymmetric.

So this further reinforces my previous comment, explicit padding check should be applied at the module function level instead (after actual padding sizes have been computed).

laggui (changes requested):
Just a couple of changes needed, on top of the compilation error due to breaking `Explicit(...)` enum variant change.

khengari77:
Hi @antimora @laggui,

I've also been looking into this to support models like `Facial-Landmark-Detection` which use asymmetric padding and I was working on.

Since native support in `burn-nn` involves breaking changes to the `Explicit` enum variants, I experimented with an alternative **decomposition strategy strictly within `burn-import`**.

Instead of changing the runtime `PaddingConfig`, my approach handles it during ONNX import code generation:
1.  `onnx-ir`: Parses asymmetric padding into a new internal `Asymmetric` variant (non-breaking for burn-core).
2.  `burn-import`: When generating code for `Conv2d`/`Pool2d`:
    *   Configures the layer with `PaddingConfig2d::Valid`.
    *   Injects a `tensor.pad(...)` call immediately before the layer in the generated `forward` method.

This allows existing Burn versions to run these models without waiting for a breaking release or modifying the core tensor API. I have a branch with the initial code as a PoC. I might need to add the tests but it's pretty ready already

khengari77:
> Hi @antimora @laggui,
> 
> I've also been looking into this to support models like `Facial-Landmark-Detection` which use asymmetric padding and I was working on.
> 
> Since native support in `burn-nn` involves breaking changes to the `Explicit` enum variants, I experimented with an alternative **decomposition strategy strictly within `burn-import`**.
> 
> Instead of changing the runtime `PaddingConfig`, my approach handles it during ONNX import code generation:
> 
> 1. `onnx-ir`: Parses asymmetric padding into a new internal `Asymmetric` variant (non-breaking for burn-core).
> 2. `burn-import`: When generating code for `Conv2d`/`Pool2d`:
>    
>    * Configures the layer with `PaddingConfig2d::Valid`.
>    * Injects a `tensor.pad(...)` call immediately before the layer in the generated `forward` method.
> 
> This allows existing Burn versions to run these models without waiting for a breaking release or modifying the core tensor API. I have a branch with the initial code as a PoC. I might need to add the tests but it's pretty ready already

Here's my commit on my fork. I think this is the path is the one of least resistance and less breaking changes: 258d47b2f972ff0e3a6e5c5bc5745f1db3f00049

antimora:
@khengari77 thanks for looking into this as well. I thought going this route but it only benefits ONNX users and if users want to shift to Burn API then they need to do padding manually. 

I believe for this release we are okay to introduce a breaking change.

khengari77:
> @khengari77 thanks for looking into this as well. I thought going this route but it only benefits ONNX users and if users want to shift to Burn API then they need to do padding manually. 
> 
> I believe for this release we are okay to introduce a breaking change.

Actually I think you can create an abstraction over padding that internally decomposes in the same way. Again this will not introduce any breaking changes. Anyways if you see that the other choice is better in the long term then that's something else. Thanks for your response.

laggui:
> Actually I think you can create an abstraction over padding that internally decomposes in the same way. Again this will not introduce any breaking changes. Anyways if you see that the other choice is better in the long term then that's something else. Thanks for your response.

You mean having an `Asymmetric` variant? e.g.

```rust
#[derive(Config, Debug, PartialEq)]
pub enum PaddingConfig1d {
    Same,
    Valid,
    Explicit(usize), // i.e., symmetric
    Asymmetric(usize, usize),
}

#[derive(Config, Debug, PartialEq)]
pub enum PaddingConfig2d {
    Same,
    Valid,
    Explicit(usize, usize), // i.e., symmetric
    Asymmetric(usize, usize, usize, usize),
}
```

that's not a bad alternative, but if we go with a separation between explicit symmetric and asymmetric padding I would still rename `Explicit` -> `Symmetric` which is also a breaking change.

khengari77:
> but if we go with a separation between explicit symmetric and asymmetric padding I would still rename Explicit -> Symmetric which is also a breaking change

@laggui Quite what I was trying to say. I think you can add `Symmetric` side by side with `Explicit` and just mark it as deprecated and then remove it a few versions later. Just an opinion.

laggui:
One problem though with `Symmetric` and `Asymmetric` is that symmetry is a property that emerges from the values provided, not a fundamentally different kind of padding operation. Both represent explicit padding values. If you have `PaddingConfig1d::Asymmetric(2, 2)`, that's just symmetric anyway, which is a bit redundant.

So I think having a single `Explicit` variant is still preferable, and we can provide constructor methods e.g. `PaddingConfig1d::symmetric(pad)` as a shorthand.

antimora:
@laggui, ready for your review; I have updated/merged with burn-onnx rename.

crates/burn-backend/src/backend/ops/modules/base.rs
```diff
@@ -98,9 +98,14 @@ pub struct ConvOptions<const N: usize> {
...
     pub stride: [usize; N],
 
-    /// Padding.
+    /// Padding at the start of each dimension (top/left for 2D).
     pub padding: [usize; N],
 
+    /// Padding at the end of each dimension (bottom/right for 2D).
+    /// If `None`, uses the same values as `padding` (symmetric).
+    /// If `Some`, enables asymmetric padding where start != end.
+    pub padding_out: Option<[usize; N]>,
```
laggui: Hmmm not sure this is required to allow asymmetric padding, but I'll have to review this more carefully on Monday because too many changes have been made since my last review comment

antimora: @laggui ready for your review. I have separated ONNX related code to burn-onnx.

laggui: Yeah I still think this should not be added.

`padding_out` doesn't relate to convolutions (as opposed to conv transpose), even as a workaround to handle asymmetric padding. `ConvOptions` is provided to the backend convolution operation, which doesn't handle logical padding (asymmetric) but only the symmetric padding (supported by kernel implementations).

I'd keep `ConvOptions` as is, but the functional API can apply explicit asymmetric padding before dispatching to `B::conv1d(...)` like you currently implemented. We just need to handle the options differently, e.g.

```rust
pub fn conv1d<B>(
    x: Tensor<B, 3>,
    weight: Tensor<B, 3>,
    bias: Option<Tensor<B, 1>>,
    options: impl Into<LogicalConvOptionsType>,
) -> Tensor<B, 3>
```

(dummy name for the new options, but it would support `From<ConvOptions<1>>`  which is the current symmetric form, and also asymmetric padding specifier of some kind)

antimora: @laggui , okay, it should be addressed now.

crates/burn-cubecl/src/kernel/conv/backward_data/tune.rs
```diff
@@ -115,6 +115,7 @@ fn create_key<R: CubeRuntime, const N: usize>(
         padding,
         dilation,
         groups,
+        ..
```
laggui: Left overs?

crates/burn-cubecl/src/kernel/conv/backward_weight/tune.rs
```diff
@@ -118,6 +118,7 @@ fn create_key<R: CubeRuntime, const N: usize>(
         padding,
         dilation,
         groups,
+        ..
```
laggui: Same

crates/burn-cubecl/src/kernel/conv/forward/tune.rs
```diff
@@ -117,6 +117,7 @@ fn create_key<R: CubeRuntime, const N: usize>(
         padding,
         dilation,
         groups,
+        ..
```
laggui: Same

crates/burn-ir/src/operation.rs
```diff
@@ -1316,6 +1316,7 @@ impl From<Conv1dOptionsIr> for ConvOptions<1> {
         ConvOptions {
             stride: val.stride,
             padding: val.padding,
+
```
laggui: Same

crates/burn-ir/src/operation.rs
```diff
@@ -1327,6 +1328,7 @@ impl From<Conv2dOptionsIr> for ConvOptions<2> {
...
```
laggui: Same

crates/burn-ir/src/operation.rs
```diff
@@ -1338,6 +1340,7 @@ impl From<Conv3dOptionsIr> for ConvOptions<3> {
...
```
laggui: Same

crates/burn-nn/src/modules/conv/conv3d.rs
```diff
@@ -154,6 +154,15 @@ impl<B: Backend> Conv3d<B> {
...
     /// - output: `[batch_size, channels_out, depth_out, height_out, width_out]`
     pub fn forward(&self, input: Tensor<B, 5>) -> Tensor<B, 5> {
+        // Asymmetric 3D padding is not currently supported because burn's pad API
+        // only supports padding the last 2 dimensions
+        if self.padding.is_asymmetric() {
+            panic!(
+                "Asymmetric 3D padding is not currently supported. \
+                burn's pad API only supports 2D padding (last two dimensions)."
+            );
+        }
```
laggui: In this case I would remove the provisions from `PaddingConfig3d` entirely. More explicit, and users won't mistakenly set asymmetric padding.

crates/burn-nn/src/padding.rs
```diff
@@ -10,22 +25,62 @@ pub enum PaddingConfig1d {
...
+            Self::Same => calculate_same_padding(kernel_size, stride, length),
+            Self::Explicit(left, right) => (*left, *right),
+        }
+    }
+
+    /// Calculate symmetric padding for 1D operations.
+    /// Returns a single padding value (same for both sides).
+    /// Panics if asymmetric padding is detected.
+    #[allow(dead_code)] // Kept for API consistency with 2D/3D; may be used by future modules
     pub(crate) fn calculate_padding_1d(
```
laggui: Why keep dead code? It was restricted to the crate anyway, `calculate_padding_1d_pair` can be renamed to remove this.

crates/burn-nn/src/padding.rs
```diff
@@ -10,22 +25,62 @@ pub enum PaddingConfig1d {
...
+        }
+    }
+
+    /// Returns the explicit padding values (left, right).
+    /// Panics if not Explicit padding. For computed padding, use `calculate_padding_1d_pair()`.
+    pub fn as_tuple(&self) -> (usize, usize) {
+        match self {
+            Self::Explicit(left, right) => (*left, *right),
+            _ => panic!("as_tuple() only works with Explicit padding"),
         }
```
laggui: Where is this used? Outside of tests I couldn't find, and not sure where it would be useful. So I'd remove.

crates/burn-nn/src/padding.rs
```diff
@@ -10,22 +25,62 @@ pub enum PaddingConfig1d {
...
+    /// Applies explicit padding values.
+    /// Format: (left, right)
+    /// For symmetric padding, use the same value for both (e.g., `Explicit(1, 1)`).
+    Explicit(usize, usize),
 }
 
 impl PaddingConfig1d {
+    /// Calculate padding as (left, right) pair for 1D operations.
+    /// For `Same` padding, this computes the actual asymmetric padding if needed.
+    pub fn calculate_padding_1d_pair(
```
laggui: Also why is this public and not restricted to the crate?

crates/burn-nn/src/padding.rs
```diff
@@ -37,29 +92,68 @@ pub enum PaddingConfig2d {
...
+    /// Applies explicit padding values.
+    /// Format: (top, left, bottom, right)
+    /// For symmetric padding, use matching values (e.g., `Explicit(1, 1, 1, 1)`).
+    Explicit(usize, usize, usize, usize),
 }
 
 impl PaddingConfig2d {
+    /// Calculate padding as ((top, bottom), (left, right)) pairs for 2D operations.
+    /// For `Same` padding, this computes the actual asymmetric padding if needed.
+    pub fn calculate_padding_2d_pairs(
```
laggui: Same (make it `pub(crate)` only)

crates/burn-nn/src/padding.rs
```diff
@@ -37,29 +92,68 @@ pub enum PaddingConfig2d {
...
 
+    /// Returns the explicit padding values (top, left, bottom, right).
+    /// Panics if not Explicit padding. For computed padding, use `calculate_padding_2d_pairs()`.
+    pub fn as_tuple(&self) -> (usize, usize, usize, usize) {
         match self {
-            Self::Same => same_padding(),
-            Self::Valid => [0, 0],
-            Self::Explicit(v1, v2) => [*v1, *v2],
+            Self::Explicit(top, left, bottom, right) => (*top, *left, *bottom, *right),
+            _ => panic!("as_tuple() only works with Explicit padding"),
```
laggui: Also unused?

crates/burn-nn/src/padding.rs
```diff
@@ -71,11 +165,40 @@ pub enum PaddingConfig3d {
...
+    /// Applies explicit padding values.
+    /// Format: (front, top, left, back, bottom, right)
+    /// For symmetric padding, use matching values (e.g., `Explicit(1, 1, 1, 1, 1, 1)`).
+    Explicit(usize, usize, usize, usize, usize, usize),
 }
 
 impl PaddingConfig3d {
+    /// Calculate padding as ((front, back), (top, bottom), (left, right)) pairs for 3D operations.
+    /// For `Same` padding, this computes the actual asymmetric padding if needed.
+    pub fn calculate_padding_3d_pairs(
```
laggui: Same (`pub(crate)`)

crates/burn-nn/src/padding.rs
```diff
@@ -84,18 +207,294 @@ impl PaddingConfig3d {
...
+    /// Panics if not Explicit padding. For computed padding, use `calculate_padding_3d_pairs()`.
+    pub fn as_tuple(&self) -> (usize, usize, usize, usize, usize, usize) {
         match self {
-            Self::Same => same_padding(),
-            Self::Valid => [0, 0, 0],
-            Self::Explicit(v1, v2, v3) => [*v1, *v2, *v3],
+            Self::Explicit(front, top, left, back, bottom, right) => {
+                (*front, *top, *left, *back, *bottom, *right)
+            }
+            _ => panic!("as_tuple() only works with Explicit padding"),
```
laggui: Same (unused)

laggui (changes requested):
Much better with the `PaddedConvOptions` and `impl Into<PaddedConvOptions<N>>` 👍 

Still some leftover stuff that should probably removed, but once cleaned up should be good!

crates/burn-cubecl/src/kernel/conv/backward_data/tune.rs
antimora: Fixed. Removed the leftover `...` in the `ConvOptions` destructuring. All 4 fields are now explicitly listed so it's no longer needed.

crates/burn-cubecl/src/kernel/conv/backward_weight/tune.rs
antimora: Fixed.

crates/burn-cubecl/src/kernel/conv/forward/tune.rs
antimora: Fixed.

crates/burn-ir/src/operation.rs
antimora: Fixed. Removed the leftover blank lines.

crates/burn-ir/src/operation.rs
antimora: Fixed.

crates/burn-ir/src/operation.rs
antimora: Fixed.

crates/burn-nn/src/modules/conv/conv3d.rs
antimora: Done. Simplified `PaddingConfig3d::Explicit` to 3 symmetric fields `(depth, height, width)` instead of 6. Removed the `is_asymmetric()` check from conv3d forward — it was redundant since `calculate_padding_3d` already enforces symmetry.

crates/burn-nn/src/padding.rs
antimora: Removed. `calculate_padding_1d` is gone entirely. `calculate_padding_1d_pair` is the only 1D calculate function now.

crates/burn-nn/src/padding.rs
antimora: Removed `as_tuple()` from all three PaddingConfig types.

crates/burn-nn/src/padding.rs
antimora: Changed to `pub(crate)`.

crates/burn-nn/src/padding.rs
antimora: Done.

crates/burn-nn/src/padding.rs
antimora: Removed `is_asymmetric()` from all three PaddingConfig types — none were used outside of tests.

crates/burn-nn/src/padding.rs
antimora: Done.

crates/burn-nn/src/padding.rs
antimora: Removed.

crates/burn-backend/src/backend/ops/modules/base.rs
antimora: Agreed and done. `ConvOptions` is kept as-is (symmetric only). The functional API now uses `impl Into<PaddedConvOptions<N>>` which wraps `ConvOptions` with optional asymmetric padding. `From<ConvOptions<N>>` is implemented for backward compatibility. The functional layer handles explicit pad before dispatching to `B::conv1d/conv2d`.

Also addressed all the cleanup comments — removed leftovers in burn-cubecl tune files, burn-ir, simplified PaddingConfig3d to symmetric-only, removed dead code/unused methods from padding.rs, made calculate functions `pub(crate)`.

antimora: Resolved — `ConvOptions` no longer has `padding_out`. Asymmetric padding is handled via `PaddedConvOptions` at the functional layer only.

PR #4268, opened 2025-12-31
Checkpointer interrupt

Charles23R:
### Problem

An error when checkpointing didn't stop the training, so sometimes a whole training iteration is executed for nothing even **after** the checkpointer has already failed.

### Checklist

### Changes

- `AsyncCheckpointer` can be assigned an interrupter to stop the training.
- Interrupter can be given a "reason" string when stopping, improving logging.
- Updated burn-train's training loops.
- Fix a warning for projects that do not use ddp.

### Testing

Tested with examples.

### DIFF
--- a/crates/burn-train/src/learner/supervised/strategies/ddp/strategy.rs
+++ b/crates/burn-train/src/learner/supervised/strategies/ddp/strategy.rs
@@ -125,10 +125,10 @@ impl<SC: SupervisedLearningComponentsTypes + Send + 'static> SupervisedLearningS
             .expect("Distributed data parallel main worker failed");
 
         if interrupter.should_stop() {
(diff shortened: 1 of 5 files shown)


crates/burn-train/src/learner/base.rs
```diff
@@ -193,4 +200,10 @@ impl Interrupter {
     pub fn should_stop(&self) -> bool {
         self.state.load(Ordering::Relaxed)
     }
+
+    /// Get the message associated with the interrupt.
+    pub fn get_message(&self) -> Option<String> {
```
laggui, line 205: Could also have been `interrupter.reason()` instead to get the stop reason, but LGTM nonetheless

Issue #2876, opened 2025-03-07
`bincode` v2.0.0 error: cannot find decode_borrowed_from_slice


BeadiestStar64:
**Describe the bug**
When I do a `cargo build` with Burn as a dependency, cargo reports that Burn Core cannot find `bincode::serde`.

**To Reproduce**
1. Execute `cargo add burn --features wgpu`.
2. Execute `cargo test`.

**Expected behavior**
Pass the test without any problems.

**Screenshots**
![Image](https://github.com/user-attachments/assets/4322d33a-d862-4991-a3da-466192233d70)

**Desktop (please complete the following information):**
 - OS: Windows 11 Home 24H2
 - Rust Edition: 2021

laggui:
`bincode` version `2.0.0` was officially released yesterday, which introduces some breaking changes.

Burn uses `2.0.0-rc.3`, but cargo dependency resolver treats pre-release versions (-rc.3) as lower priority than stable releases... so it fetches the `2.0.0` release (which is not compatible).

For the released Burn versions, please add an explicit dependency to the pre-release version:

```toml
bincode = "=2.0.0-rc.3"
bincode_derive = "=2.0.0-rc.3"
```

The new `bincode` release also only supports 1.85 MSRV, so for now we'll have to explicitly pin the release candidate on main and eventually upgrade to the last release.

BeadiestStar64:
Thank you so much for your detailed explanation and solution.
As you suggested, adding an explicit dependency on the pre-release version of `bincode` quickly resolved the issue👍

laggui:
Re-opening in favor of #2880 since an explanation and solution has been given here.

alessandromazza98:
is there an eta for the fix on main branch?

laggui:
For some reason cargo was not resolving to bincode v2.0 locally or on the CI so I thought the pinned rc3 had been merged, but it's still contained in another PR.

Just opened a PR for this fix specifically. Will update to 2.0 once rust 1.86 comes out because we always support current (1.85) and prev (1.84), and bincode 2.0 has MSRV 1.85.

No ETA for a release at this time, we have a couple of things we still want to do to wrap up the next release 🙂

laggui:
This should no longer be an issue with the 0.16.1 patch we just released!

Would have hoped to have 0.17 by now, but we were not satisfied enough to release (software + performance is hard apparently 😅).

So we decided to finally bring a patch update mostly for the breaking `bincode` issue.

[...]

PR #4272, opened 2026-01-04
Pinned bincode workspace dependency to the exact version =2.0.0-rc.3 …

bigmark222:
…in Cargo.toml

### Checklist

### Related Issues/PRs
- Repro: `git checkout v0.14.0 && rm Cargo.lock && cargo publish --dry-run -p burn-core` fails with E0425 (bincode API change).

### Changes
- Pin `bincode` in burn-core to `=2.0.0-rc.3` to prevent Cargo from resolving 2.0.1+, which removes `decode_borrowed_from_slice` and breaks publish/build when no lockfile is present.
- Scope: 0.14.x line only; minimal change to restore publishability.

### Testing
- `cargo run-checks`
- `cargo publish --dry-run -p burn-core` (succeeds with the exact pin)

Additional info just for clarity:
- Cargo.lock isn’t used by library consumers; when downstream builds burn-core, Cargo re-resolves deps. Any lock in the repo is ignored.
- cargo publish also re-resolves unless you force --locked. With the lock present it holds bincode at 2.0.0-rc.3, so the break is masked; without it, Cargo picks bincode 2.0.1 and build/publish fails.
- The repro removes Cargo.lock to mirror the published experience: fresh resolution pulls bincode 2.0.1, which lacks decode_borrowed_from_slice, causing E0425.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -109,7 +109,7 @@ github-device-flow = "0.2.0"
 os_info = "3.8.2"
 wsl = "0.1.0"
 
-bincode = { version = "2.0.0-rc.3", features = [
+bincode = { version = "=2.0.0-rc.3", features = [
     "alloc",
     "serde",
 ], default-features = false }


bigmark222:
Thanks for the context and the link to #2876.

We’re still on 0.14.0 because our downstream crates are pinned to that release, and moving to 0.16+ is a larger migration we haven’t scheduled yet.

Pinning bincode in my own project doesn’t fix the publish issue in burn-core itself. When running cargo publish --dry-run -p burn-core (and for downstream crates), Cargo re-resolves dependencies without a lockfile and pulls in bincode 2.0.1, which then fails due to decode_borrowed_from_slice. That’s what motivated this PR for the 0.14.x line - to keep that release publishable until we can move to a newer Burn version.

If you’d prefer, I’m happy to switch this over to the API update (decode_from_slice) instead of the pin. Either approach on the 0.14 branch would unblock publishing.

laggui:
Pinning is fine!

> Pinning bincode in my own project doesn’t fix the publish issue in burn-core itself.

Ahhh yeah this is a fix in user-space (binary project), won't work for libs. Your users would have to apply this in their project.

Hopefully you can take advantage of the improvements in the latest releases soon! The upgrade is worth it. In the meantime, I'll try to trigger a patch publish for 0.14.1 later with the pinned rev.

bigmark222:
That makes sense. Thanks for the clarification, and for taking a look.

Agreed on the distinction; a user-space pin helps binaries, but doesn’t address library publishability, which is what prompted this PR on the 0.14.x line. A patch publish with the pinned rev would fully unblock that.

Appreciate you triggering a 0.14.1 patch release. We’ll plan the upgrade to newer Burn releases once our downstream crates are ready. Looking forward to taking advantage of the improvements there.

laggui:
@bigmark222 had to update some things in our tooling but just published a patch 0.14.1 for `burn-core`. All other crates remain at version 0.14.0. Just tested locally and burn correctly pulls the updated patch, there is no bincode build error.

PR #4273, opened 2026-01-04
refactor: Migrate to `usize` indexing

wingertge:
### Checklist

### Related Issues/PRs

Migrates to changes in https://github.com/tracel-ai/cubecl/pull/1127 and https://github.com/tracel-ai/cubek/pull/51

### Changes

Migrates cubecl kernels and fusion to `usize` indexing

### Testing

The test suite runs successfully, though 64-bit indexing is not yet enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,11 +182,11 @@ portable-atomic-util = { version = "0.2.4", features = ["alloc"] }
 ### For the main burn branch. ###
 cubecl = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "7cd8e9c869d363edcbf989fd905eb4b28df938ab" }
 cubecl-common = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "7cd8e9c869d363edcbf989fd905eb4b28df938ab" }
-# cubek = { git = "https://github.com/tracel-ai/cubek", default-features = false, rev = "0d9a635229d3cabfa8297ddc967ff4e783be348c" }
+cubek = { git = "https://github.com/tracel-ai/cubek", default-features = false, rev = "8097a621dfd3a6a89f8d0433994e8c1adba377c2" }
 ### For local development. ###
 # cubecl = { path = "../cubecl/crates/cubecl", default-features = false }
 # cubecl-common = { path = "../cubecl/crates/cubecl-common", default-features = false }
-cubek = { path = "../cubek/crates/cubek", default-features = false }
+# cubek = { path = "../cubek/crates/cubek", default-features = false }
 ### For the release. ###
 # cubecl = { version = "=0.9.0-pre.6", default-features = false }
 # cubecl-common = { version = "=0.9.0-pre.6", default-features = false }


crates/burn-backend-tests/tests/common/autodiff.rs
```diff
@@ -30,5 +30,5 @@ test_float_elem_variant!(
     bf16,
     burn_tensor::bf16,
     "../autodiff/mod.rs",
-    ["vulkan", "metal"] // ["cuda", "rocm"] TODO
+    ["metal"] // ["cuda", "rocm"] TODO, ["vulkan"] only supports bf16 for matmul
```
laggui: Whoops, I added bf16 for vulkan but clearly the tests don't pass :sweat_smile: thanks for fixing.

[Unrelated to this PR]

I vaguely remember adding it when refactoring the tests since it was a supported global type for vulkan in cubecl. Also reflected by `B::supports_dtype`:

https://github.com/tracel-ai/burn/blob/868270962cf4ee3ef0338b23a2fff264725522cc/crates/burn-wgpu/src/lib.rs#L117-L145

maybe we should have a better way to represent the actual supported types?

https://github.com/tracel-ai/burn/blob/868270962cf4ee3ef0338b23a2fff264725522cc/crates/burn-cubecl/src/backend.rs#L83-L88

That way, tested dtypes can actually reflected supported dtypes.

wingertge: The `supports_type` method on the CubeCL side only checks if a type is supported in any way (in this case, it's supported for conversion, as a type for buffers, and for dot product on Intel, along with tensor core instructions). It's kinda tough though because there's no good way to express that in just a single boolean (hence why the `TypeUsage` enum set exists in CubeCL).

This is how it's registered for Vulkan
```rust
if let Some(bfloat16) = ext_feat.bfloat16 {
    if bfloat16.shader_b_float16_type == TRUE {
        register(
            ElemType::Float(FloatKind::BF16).into(),
            TypeUsage::Conversion | TypeUsage::Buffer,
        );
    }
    if bfloat16.shader_b_float16_dot_product == TRUE {
        register(
            ElemType::Float(FloatKind::BF16).into(),
            TypeUsage::DotProduct.into(),
        );
    }
}
```

wingertge: So it's supported for matmul and casting, but none of the other ops.

laggui: > The `supports_type` method on the CubeCL side only checks if a type is supported in any way

Yeah and for the first draft I simply mirrored that for the backends, but I think it should be refined.

> It's kinda tough though because there's no good way to express that in just a single boolean (hence why the TypeUsage enum set exists in CubeCL).

That's a good point. It's still useful to query backend supported types for burn, so maybe we should also define an enum similar to `TypeUsage`? (without atomics, and perhaps consolidate conversion / buffer into "storage" variant or similar).

wingertge: Regardless, that's something for a separate issue/PR I think.

laggui: Oh yes entirely, that's why I prefaced my initial comment with `[Unrelated to this PR]` :smile: 

Just wanted to get your thoughts since this was related to the bf16 change

PR #4283, opened 2026-01-06
Clean learning api

Charles23R:
### Checklist

### Changes

- Simplified traits for the ComponentsTypes.
- Decoupled LearningData and ParadigmData.
- Renamed some stuff related to learning.
- Deleted the LearningParadigm trait.
- Renamed SupervisedTraining's `run` function to `launch`.
- The Learner now does the actual learning.

### Testing

Compiles and runs.

### DIFF
--- a/crates/burn-train/src/components.rs
+++ b/crates/burn-train/src/components.rs
@@ -1,4 +1,4 @@
-use crate::{TrainStep, ValidStep, metric::ItemLazy};
+use crate::{InferenceStep, TrainStep};
 use burn_core::{module::AutodiffModule, tensor::backend::AutodiffBackend};
 use burn_optim::{Optimizer, lr_scheduler::LrScheduler};
 use std::marker::PhantomData;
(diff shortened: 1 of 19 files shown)


burn-book/src/basic-workflow/training.md
```diff
@@ -94,7 +94,7 @@ for our model.
 #     }
 # }
 # 
-impl<B: AutodiffBackend> TrainStep<MnistBatch<B>, ClassificationOutput<B>> for Model<B> {
+impl<B: AutodiffBackend> LearningStep<MnistBatch<B>, ClassificationOutput<B>> for Model<B> {
```
laggui: Hmmm unsure if it's worth renaming `TrainStep` to `LearningStep`.

It's a breaking change that doesn't really improve much, and we still refer to "training" in other types and docs (e.g., `TrainLoader` and `dataloader_train`). I think train / valid are still preferred and widely understood counterparts.

crates/burn-train/src/components.rs
```diff
@@ -1,76 +1,87 @@
...
-/// A reference to the training split [DataLoader](DataLoader).
-pub type TrainLoader<LC, LD> = Arc<dyn DataLoader<TrainBackend<LC>, InputTrain<LD>>>;
-/// A reference to the validation split [DataLoader](DataLoader).
-pub type ValidLoader<LC, LD> = Arc<dyn DataLoader<ValidBackend<LC>, InputValid<LD>>>;
+/// The model of the learner.
+pub type LearnerModel<LC> = <LC as LearningComponentsTypes>::Model;
+/// The non-autodiff model of the learner.
+pub type ValidModel<LC> = <LC as LearningComponentsTypes>::InnerModel;
+/// Type for training input.
+pub(crate) type LearnerInput<LC> =
```
laggui: Concrete example here, where `LeanerInput` is documented as "training input". So why not just `TrainInput` if we decide to keep `TrainStep` too.

crates/burn-train/src/learner/base.rs
```diff
@@ -1,108 +1,162 @@
...
+    /// # Arguments
+    ///
+    /// * `optim`: Optimizer used for learning.
+    /// * `lr`: The learning rate used for this step.
+    /// * `grads`: Multiple gradients associated to each parameter in the current model.
+    ///
+    /// # Returns
+    ///
+    /// The updated model.
+    pub fn optimize_multi(&mut self, grads: MultiGradientsParams) {
```
laggui: Doesn't return anything

burn-book/src/basic-workflow/training.md
Charles23R: I spoke about this with Nath, I'm gonna rework this PR

laggui: Sounds good!

Issue #3632, opened 2025-08-28
Tensors: Allow choosing dtype on creation


torsteingrindvik:
I have a backend which defaults to f32.

I have a model which wants f16.

I should be able to `Tensor::empty(...)` and get a f16 directly, doing `Tensor::empty(...).cast(...)` seems wasteful.

So in general I'd like to be able to create any supported dtype.

Some context: https://discordapp.com/channels/1038839012602941528/1059209073784008835/1410599054873985024

nathanielsimard:
Should be fixed after #3642

PR #4285, opened 2026-01-07
Add empty/zeros/ones/full TensorOptions

laggui:
### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/3632

### Changes

Added `TensorOptions` to allow specifying device and dtype for creation ops.

Implemented `From<&Device<B>> for TensorOptions<B>` which preserves current usage too with a single `&device` arg in creation ops which resolves dtype to `opt.dtype_or(K::Elem::dtype())`.

### Testing

Added tests

### DIFF
--- a/crates/burn-tensor/src/tensor/api/options.rs
+++ b/crates/burn-tensor/src/tensor/api/options.rs
@@ -84,3 +84,10 @@ impl<B: Backend> From<&Device<B>> for TensorCreationOptions<B> {
         TensorCreationOptions::new(device.clone())
     }
 }
+
+impl<B: Backend> From<(&Device<B>, DType)> for TensorCreationOptions<B> {
+    /// Convenience conversion for a specified `(&device, dtype)` tuple.
+    fn from(args: (&Device<B>, DType)) -> Self {
+        TensorCreationOptions::new(args.0.clone()).with_dtype(args.1)
+    }
+}
--- a/crates/burn-backend-tests/tests/tensor/float/ops/full.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/full.rs
@@ -1,5 +1,5 @@
 use super::*;
-use burn_tensor::{DType, TensorData, TensorOptions};
+use burn_tensor::{DType, TensorData};
 
 #[test]
 fn test_data_full() {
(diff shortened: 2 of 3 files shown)


crates/burn-tensor/src/tensor/api/options.rs
```diff
@@ -0,0 +1,90 @@
+use burn_backend::{Backend, Element, tensor::Device};
+use burn_std::DType;
+
+/// Options for tensor creation.
+///
+/// This struct allows specifying the `device` and/or data type (`dtype`) when creating a tensor.
+#[derive(Debug, Clone)]
+pub struct TensorOptions<B: Backend> {
```
nathanielsimard: Unsure about the name and the API. If we were willing to have a long name, that could be called `TensorCreationOptions`, and we could have the following:

```impl TensorOptions {
      pub fn new(device);
      pub fn device(&mut self, device);
      pub fn dtype(&mut self, dtype);
      pub fn default() // Defautl device.
}
```
Comments?

laggui: Yeah we can have `TensorCreationOptions` instead, it's more explicit.

Already have the default implementation which currently returns the default device + backend float elem dtype. But I agree with your suggested API, it's nicer to only set the device and allow specifying the dtype via e.g. `TensorCreationOptions::new(device).with_dtype(dtype)` when required. Though I think I'll follow the other APIs we have for config / options that take ownership of self and return the updated struct.

laggui:
@nathanielsimard I renamed to `TensorCreationOptions`, the default only sets the default device and we have `.with_dtype(dtype)` and `.with_device(device)` now.

I think the current usage is clean, but for the dtype materialization I see 2 options:
- Keep `Option<DType>` for the struct, along with `.dtype_or(dtype)` (similar to `some_opt.unwrap_or(default)`) to resolve on creation
- Force the `dtype` field on `TensorCreationOptions` to be set / materialized, in which case we need a builder pattern that resolves the full type

Lmk if you have a preference

crates/burn-tensor/src/tensor/api/options.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// use burn_tensor::TensorCreationOptions;
+    ///
+    /// fn example<B: Backend>(device: B::Device) {
+    ///     let options: TensorCreationOptions<B> = (&device).into();
+    /// }
+    /// ```
+    fn from(device: &Device<B>) -> Self {
+        TensorCreationOptions::new(device.clone())
+    }
+}
```
nathanielsimard, line 86: I would add an implementation for `From<(&Device<B>, DTpe)>`. I think that would be clean.

Issue #2406, opened 2024-10-23
where is dataframe dataset


wangjiawen2013:
Hi,
The dataframe dataset has already been merged into burn and is demonstrated in burns book:
![image](https://github.com/user-attachments/assets/99e4c664-cbf1-49c5-88cc-c08f1f09b67b)

But I cannot find it in burn's docs and examples, where is it ?

HessTaha:
Hello, 

The code is separated in differents crates.
You Can find the dataset crate under the crates folder:  https://github.com/tracel-ai/burn/tree/main/crates%2Fburn-dataset

laggui:
Right now there are no examples which make use of the `DataframeDataset` (which is under the `dataframe` feature flag for the dataset module).

I agree that it would be nice to have at least a snippet for its usage. I'll convert this issue to track this addition.

antimora:
Yeah it'd be great if we have example. Meanwhile you can see some basic tests:

https://github.com/tracel-ai/burn/blob/main/crates/burn-dataset/src/dataset/dataframe.rs

josemanuel657:
I assume this issue should be closed?

laggui:
Thanks for bringing this up. It has been addressed.

PR #4298, opened 2026-01-09
docs: add DataframeDataset example using Polars

SameerVers3:
### Checklist

### Related Issues/PRs

Contributes to #2406

### Changes

The `DataframeDataset` API was missing documentation and examples. Added a new example to the `custom-csv-dataset` demonstrating how to use `DataframeDataset` with Polars as the backend, using the same diabetes dataset for consistency.

### Testing

- Example runs successfully with `cargo run --example dataframe-dataset --features dataframe`
- No breaking changes introduced in other crates.


examples/custom-csv-dataset/src/dataframe_dataset.rs
```diff
@@ -0,0 +1,134 @@
+use burn_dataset::{DataframeDataset, Dataset};
+use polars::prelude::*;
+use serde::Deserialize;
+use std::{fs::File, io::copy, path::Path, sync::Arc};
+
+/// Diabetes patient record for DataframeDataset.
+/// Note: Field types must match Polars column types exactly.
+/// Polars reads integers as i64 and floats as f64 by default.
+#[derive(Deserialize, Debug, Clone)]
+pub struct DiabetesPatient {
```
laggui: Couldn't we set the scheme types instead of duplicating the struct to change the default types?

examples/custom-csv-dataset/src/dataframe_dataset.rs
```diff
@@ -0,0 +1,134 @@
...
+            .finish()?
+            .collect()?;
+
+        let dataset = DataframeDataset::new(df)?;
+
+        Ok(Self { dataset })
+    }
+    /// Download the CSV file from its original source on the web.
+    /// Panics if the download cannot be completed or the content of the file cannot be written to disk.
+    fn download() -> String {
```
laggui: We should extract the method from the other dataset struct as a function and re-use it

SameerVers3: Yes you're right, in the following commit I extracted the method into utils and reused it in both dataset implementations.

examples/custom-csv-dataset/src/dataframe_dataset.rs
SameerVers3: Done in the following commit. Now using Polars schema with type casting to match the shared struct instead of duplicating.

PR #4309, opened 2026-01-12
ensure that tensor is owned on iter_dim call

tzemanovic:
### Checklist

### Changes

Changes introduced in https://github.com/tracel-ai/burn/pull/4178 affected `iter_dim` in a way that may blow up memory usage on ndarray, e.g.:

- `iter_dim` get called with a borrowed tensor
- each iteration calls `tensor.clone().slice(...)` (via `impl Iterator for DimIter`)
- call to slice invokes `NdArrayStorage::into_shared`/`into_owned` (via macro in burn-ndarray/src/tensor.rs)
- for a borrowed tensor this ends up calling `view.to_owned()` which clones the whole tensor

I've added a call to `DimIter` constructor to `ensure_owned` on the tensor, so that `NdArrayStorage::into_owned` on each iter is a no-op.

The call goes via a few layers on indirection due to tensor primitive typing and there may be a better way to achieve this that I missed.

### Testing

Tested externally by running a test with `iter_dim` before and after and using heaptrack to profile memory usage.

### DIFF
--- a/crates/burn-ndarray/src/ops/base.rs
+++ b/crates/burn-ndarray/src/ops/base.rs
@@ -58,9 +58,9 @@ impl<E> NdArrayOps<E>
 where
     E: Copy + Debug + Element + crate::AddAssignElement,
 {
-    pub fn slice(tensor: SharedArray<E>, slices: &[Slice]) -> SharedArray<E> {
+    pub fn slice(tensor: ArrayView<E, IxDyn>, slices: &[Slice]) -> SharedArray<E> {
         let slices = Self::to_slice_args_with_steps(slices, tensor.shape().num_dims());
-        tensor.slice_move(slices.as_slice()).into_shared()
+        tensor.slice_move(slices.as_slice()).to_shared()
     }
 
     pub fn slice_assign(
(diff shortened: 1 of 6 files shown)


tzemanovic:
> That's a good catch!
> 
> However, adding `ensure_owned` to the `BasicOps` trait leaks a backend-specific optimization detail into the common API.
> 
> I think we should fix the ndarray slice implementation directly.
> 
> Instead of forcing the whole input `to_owned()` (as per this PR, via `ensure_owned`), we can slice the view and then call `to_owned()` on the result.
> 
> This way, each iteration in `DimIter` only copies the slice, not the entire tensor.
> 
> What do you think?

Thanks! Yeah, I have same feeling about leaking the backend-specifics to the high-level api, so I would prefer to fix this directly in ndarray too. Will try impl your suggestion

tzemanovic:
> > That's a good catch!
> > However, adding `ensure_owned` to the `BasicOps` trait leaks a backend-specific optimization detail into the common API.
> > I think we should fix the ndarray slice implementation directly.
> > Instead of forcing the whole input `to_owned()` (as per this PR, via `ensure_owned`), we can slice the view and then call `to_owned()` on the result.
> > This way, each iteration in `DimIter` only copies the slice, not the entire tensor.
> > What do you think?
> 
> Thanks! Yeah, I have same feeling about leaking the backend-specifics to the high-level api, so I would prefer to fix this directly in ndarray too. Will try impl your suggestion

I've done the suggested change. It is much cleaner, thanks again!

PR #4310, opened 2026-01-12
fix: Packed quantized matmul

wingertge:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/cubecl/pull/1134
https://github.com/tracel-ai/cubek/pull/58

### Changes

Updates to packed quantized tensor fixes in `cubecl` and `cubek`

### Testing

All failing quantized matmul tests now pass

### DIFF
--- a/crates/burn-std/src/tensor/dtype.rs
+++ b/crates/burn-std/src/tensor/dtype.rs
@@ -1,6 +1,5 @@
 //! Tensor data type.
 
-use cubecl::e2m1x2;
 use serde::{Deserialize, Serialize};
 
 use crate::tensor::quantization::{QuantScheme, QuantStore, QuantValue};
@@ -96,7 +95,7 @@ impl DType {
                 },
                 QuantStore::PackedU32(_) => core::mem::size_of::<u32>(),
                 QuantStore::PackedNative(_) => match scheme.value {
-                    QuantValue::E2M1 => core::mem::size_of::<e2m1x2>(),
+                    QuantValue::E2M1 => core::mem::size_of::<u8>(),
                     _ => 0,
                 },
             },


crates/burn-cubecl/src/kernel/matmul/base.rs
```diff
@@ -84,17 +85,30 @@ pub(crate) fn launch_matmul<R: CubeRuntime>(
...
+        Some((data, scale)) => {
+            // Extremely hacky fix to ensure naive can run in every case
+            if matches!(strategy, Strategy::Naive)
+                && matches!(rhs.scheme().level, QuantLevel::Block(_))
+            {
+                rhs = dequantize(rhs.clone(), lhs.dtype);
+                (
+                    lhs.dtype,
+                    MatmulInputHandleRef::new(rhs.as_handle_ref(), rhs.dtype.into()),
+                )
```
nathanielsimard, line 123: We don't support quantized rhs for naive? I would add that info here just to make sure it's clear.

crates/burn-std/src/tensor/dtype.rs
```diff
@@ -93,7 +94,11 @@ impl DType {
                         0
                     }
                 },
-                QuantStore::U32 => core::mem::size_of::<u32>(),
+                QuantStore::PackedU32(_) => core::mem::size_of::<u32>(),
+                QuantStore::PackedNative(_) => match scheme.value {
+                    QuantValue::E2M1 => core::mem::size_of::<e2m1x2>(),
+                    _ => 0,
```
nathanielsimard, line 99: Is zero the right default here?

crates/burn-cubecl/src/kernel/matmul/base.rs
wingertge: We do for some cases. Maybe naive should actually use a separate launch and dequantize only if the launch fails, let me try to implement that version.

wingertge: Basically because naive enforces row-major lhs and col-major rhs it can't support row-major block scaled rhs because then the scales end up perpendicular to the contiguous dim.

crates/burn-std/src/tensor/dtype.rs
wingertge: No but we can't panic in this code and `0` is what the existing failure case uses. It should be `unreachable!()`

laggui:
@wingertge my bad, thought the Burn PR was only missing the rev updates. Thanks for updating :pray: 

last PR to merge before we start the release process

wingertge:
> @wingertge my bad, thought the Burn PR was only missing the rev updates. Thanks for updating 🙏
> 
> last PR to merge before we start the release process

There's another fairly critical fix here https://github.com/tracel-ai/cubecl/pull/1140 that should be merged and the revision bumped before a release. It currently causes all 2D broadcast tensors to return wildly incorrect results when read.

PR #4318, opened 2026-01-13
Performance tweaks to the lp_norm code.

crutcher:
Refactor vector norm implementations to unify logic, improve accuracy, and add specialized handling for all norm types.

### DIFF
--- a/crates/burn-tensor/src/tensor/linalg/vector_norm.rs
+++ b/crates/burn-tensor/src/tensor/linalg/vector_norm.rs
@@ -126,6 +126,7 @@ pub fn vector_norm<B: Backend, const D: usize>(
 /// * 0.0
 /// * 1.0
 /// * 2.0
+/// * 2 * N for integral N,
 /// * f64::INFINITY,
 /// * f64::NEG_INFINITY,
 ///


crates/burn-tensor/src/tensor/linalg/vector_norm.rs
```diff
@@ -182,6 +207,33 @@ pub fn l2_norm<B: Backend, const D: usize>(x: Tensor<B, D>, dim: usize) -> Tenso
...
+/// # Arguments
+///
+/// * `x` - The input tensor.
+/// * `p` - The exponent of the Lp norm.
+/// * `dim` - The dimension to compute the norm over.
+///
+/// # Returns
+///
+/// The ``L(p)`` norm of the input tensor.
+pub fn _lp_norm<B: Backend, const D: usize>(x: Tensor<B, D>, p: f64, dim: usize) -> Tensor<B, D> {
```
laggui: I wouldn't make this internal function public. And for naming an underscore prefix is unconventional, I'd use something more descriptive like `lp_norm_general` or `lp_norm_impl`

crutcher: @laggui the tests aren't in the crate tho.

I changed it to `lp_norm_impl`; though I disagree about the `_<name>` pattern. It is a big help in tab-completion in IDEs. I hid it in a public `implementation` mod.

laggui: Oh I didn't realize you added tests that compare expected results.. so I now see the motivation.

But I don't agree that it should be public just for that reason. We can just check the expected values match (which they do) and leave it at that.

The review UI on github doesn't show the full context, but I actually now see that your changes are collapsing to  `vector_norm` (which I forgot about truthfully)

https://github.com/tracel-ai/burn/blob/df2cd5d2bd8743c1d53d96c41a94878bf6d9089a/crates/burn-tensor/src/tensor/linalg/vector_norm.rs#L81-L95

which is supposed to be the high-level dispatch. Why are both APIs collapsing to the same thing?

Re `_<name>` pattern, I don't really see your point. In rust that pattern used here communicates the wrong semantic intent imo, as it usually signals unused code that you do not want the compiler to warn about.

crutcher: ### `_{name}`
we don't need to fight through it, but it's about what happens in an IDE when you are typing `{prefix}<tab>`.

### `vector_norm` collapsing back to dispatch in `lp_norm`.
This is a question of api mechanics.

`vector_norm` provides a formalized selection API, whereas `lp_norm` is free-range exponent based. But in both cases, user libs which parameterize over exponentiation (rather than just calling `l1/l2_norm` directly) will be penalized when that parameterization selects the common exponent cases.

The expected callers of `vector_norm` and `lp_norm` *don't* know in advance what the exponent will be, generally, they're getting it from some other module or config, and using the interface to provide generality (and avoid exactly this dispatch).

We don't want them to pay a performance price for the choice of generality.

laggui: Yeah I understand the tradeoffs, and it is worth having these, hence the `vector_norm` selection. My question was moreso "why have two functions that do the same thing"?

Currently, `vector_norm` is the general API that dispatches to the respective norm functions and `lp_norm` is the general Lp norm implementation for `p not in [0, 1, 2, inf, -inf]`. But your change essentially makes `lp_norm` equivalent to `vector_norm`, so both entrypoints are the same. It's essentially an alias.

> `vector_norm` provides a formalized selection API, whereas `lp_norm` is free-range exponent based. 

If by that you mean `vector_norm` requires an enum `vector_norm(x, Norm::L1, dim)` that's not true, the `From<..>` implementations for `Norm` already bridges this conversion.

I'm just trying to understand the motivation :sweat_smile: and if in the end we do go with this alias, the duplicate match in `lp_norm` can be simplified to:

```rust
pub fn lp_norm<B: Backend, const D: usize>(x: Tensor<B, D>, p: f64, dim: usize) -> Tensor<B, D> {
    vector_norm(x, p, dim)
}
```

crutcher: I can see the argument for one being a simple alias of the other; but the type conversion and dual naming is still useful.

There are arguments for both being visible:
* `vector_norm` - this is an interface in both `numpy` and `pytorch`. It is expected that we'll have this interface.
* `lp_norm` - the other libs call this `norm`, but that form also has some case-degeneracy that `lp_norm` doesn't have; so the the distinction is useful.

Basically, the public API is a guide to getting users to select the right code. Aliases, and small call-type transformations to allow us to have a broader audience have almost no cost to us (and don't even appear in most link files), while broadening the scope of our users significantly.

laggui: > Basically, the public API is a guide to getting users to select the right code.

Agreed. So we're just keeping `vector_norm` around to for with the numpy and pytorch interfaces. Not necessarily a fan, but you're right that users should be able to find this even if it's pretty close to `lp_norm` in name anyway.

In any case, if they're the same then they should call the same code, and not duplicate the logic. The match arm really is equivalent to this (as per my previous comment):

```rust
pub fn lp_norm<B: Backend, const D: usize>(x: Tensor<B, D>, p: f64, dim: usize) -> Tensor<B, D> {
    vector_norm(x, p, dim)
}
```

so I would at least change that like I said.


And I wouldn't have a public internal module nonetheless

crutcher: @laggui updated.

Issue #2943, opened 2025-03-22
LoRA Wrapper for Existing Models


gilaroni:
Does a LoRA (Low-Rank Adaptation) wrapper for existing models already exist? If not, is it technically possible to construct one?

snowmead:
Hello, I would like to tackle this issue if available. If so, any pointers would be much appreciated!

nathanielsimard:
@snowmead: Unsure what would be the best strategy, but maybe keep the LoRA vectors in `Param<Tensor>`. Then we could use a module mapper to initialize the vectors for a specific model path. Finally, in the linear layer, we could check if there are LoRA vectors present and adapt the forward function accordingly.

snowmead:
@nathanielsimard 

Opened a PR with a couple of [questions](https://github.com/tracel-ai/burn/pull/4321#issuecomment-3747737772) which I believe you could answer as I was unsure what exactly was the best approach.

laggui:
Completed in #5139

PR #4321, opened 2026-01-14
feat: Add LoRA (Low-Rank Adaptation) for parameter-efficient fine-tuning

snowmead:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/2943

### Changes

Added a new `lora` module in the `burn-nn` crate which implements a considerable set of features inspired by the PEFT library.

#### Features

- Training LoRA adapters on linear layers with frozen base weights
- Configurable rank, alpha scaling, and dropout
- RSLoRA scaling option (`alpha / sqrt(rank)` instead of `alpha / rank`)
- Multiple initialization methods: Kaiming, Gaussian, Zeros
- Bias training modes: `None` (fully frozen) or `LoraOnly` (train bias only)
- Merging LoRA weights into base layer for zero-overhead inference
- Saving adapters (LoRA matrices + config) independently of base model
- Loading adapters onto fresh base models
- Runtime adapter swapping without reloading base weights

### Not implemented for the sake of concision

I figure these can be implemented in separate PRs and this particular change is to layout the foundations.

- **Layers**: Only `Linear` supported (Conv, Embedding)
- **Variants**: No QLoRA, DoRA, LoRA+, or AdaLoRA
- **Multi-adapter**: No fusion or simultaneous adapters
- **Convenience**: No auto-wrap by pattern or architecture presets

### Testing

An example crate `examples/lora-finetuning/` is implemented to test the features and are visualized using the existing `SupervisedTraining` implementation.

Non-exhaustive list of features tested:

i. Saving adapters which saves lora matrices and the configuration
ii. Loading adapters which load the lora matrices and the configurations and apply them to the base model to train 
iii. Roundtrip verification ensuring that a model loaded from disk (base model + adapters) produces identical inference results to the original trained mode

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4870,6 +4870,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "lora-finetuning"
+version = "0.20.0-pre.6"
+dependencies = [
+ "burn",
+ "log",
+ "rand 0.9.2",
+ "serde",
+]
+
+[[package]]
 name = "lru"
 version = "0.12.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/burn-book/src/SUMMARY.md
+++ b/burn-book/src/SUMMARY.md
@@ -39,3 +39,4 @@
   - [Custom Optimizer]()
   - [WebAssembly](./advanced/web-assembly.md)
   - [No-Std](./advanced/no-std.md)
+  - [LoRA Fine-tuning](./advanced/lora.md)
--- a/burn-book/src/advanced/lora.md
+++ b/burn-book/src/advanced/lora.md
@@ -1,0 +1,238 @@
+# LoRA (Low-Rank Adaptation)
+
+LoRA enables parameter-efficient fine-tuning by freezing pre-trained weights and adding small
+trainable low-rank matrices. This technique is particularly useful when:
+
+- You have a large pre-trained model and limited GPU memory
+- You want to create multiple specialized adapters from one base model
+- You need fast training with reduced parameter count
+
+All the code can be found under the
+[examples directory](https://github.com/tracel-ai/burn/tree/main/examples/lora-finetuning).
+
+## Key Concepts
+
+| Concept | Description |
+|---------|-------------|
+| **Rank** | Size of low-rank matrices (smaller = fewer parameters, typical: 4-64) |
+| **Alpha** | Scaling factor for LoRA contribution (typical: 2x rank) |
+| **Scaling** | Computed as `alpha / rank`, controls how much LoRA affects output |
+| **Adapter** | The trainable LoRA weights that can be saved/loaded independently |
+
+## Basic Usage
+
+### Configuration
+
+Configure LoRA with rank, alpha, and optional dropout:
+
+```rust,ignore
+use burn::nn::lora::{LoraConfig, LoraBias};
+
+// Configure LoRA
+let lora_config = LoraConfig::new(4)   // rank = 4
+    .with_alpha(8.0)                    // scaling = alpha/rank = 2.0
+    .with_dropout(0.0)
+    .with_bias(LoraBias::None);
+```
+
+### Applying LoRA to Layers
+
+Apply LoRA to any `Linear` layer using the `LoraAdaptable` trait:
+
+```rust,ignore
+use burn::nn::lora::LoraAdaptable;
+
+// Apply LoRA to a linear layer
+let lora_linear = linear.with_lora(&lora_config, &device);
+
+// The base weights are frozen, only LoRA matrices are trainable
+let output = lora_linear.forward(input);
+```
+
+### Merging for Inference
+
+After training, merge LoRA weights back into the base layer for zero-overhead inference:
+
+```rust,ignore
+// Merge LoRA into base weights (zero inference overhead)
+let merged_linear = lora_linear.merge();
+
+// Now it's a regular Linear layer with updated weights
+let output = merged_linear.forward(input);
+```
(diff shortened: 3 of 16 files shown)


snowmead:
Open questions:

i. Unsure if the lora section in the burn-book should be under `advanced` - "burn-book/src/advanced/lora.md"

ii. Should `LoraConfig` instead accept an `Initializer` as the [`init`](https://github.com/snowmead/burn/blob/b05cbb4baf784bc8fe5a2be89415912fc897d40b/crates/burn-nn/src/lora/config.rs#L91-L93) field type for complete flexibility and control or should I keep the [`LoraInit`](https://github.com/snowmead/burn/blob/b05cbb4baf784bc8fe5a2be89415912fc897d40b/crates/burn-nn/src/lora/config.rs#L21-L33) enum?
```rust
pub struct LoraConfig {
      pub init_a: Initializer,
      pub init_b: Initializer,
      // ...
  }
```

snowmead:
@nathanielsimard 
> I believe we should implement a more "generalized" approach to LoRA. Specifically, the logic for the update should reside within the `Param<Tensor<B, 2>>` type itself. In the current Linear layer implementation, parameters are retrieved using the `self.weights.val()` function. If this function were responsible for performing the weight update, calculating W = X + A B , then any module using 2D tensors as parameters could support LoRA adaptations natively, without requiring modifications to the existing module logic.

Yes I assumed this would be another approach. So do you mean something like this? 
```rust
let output = self.weight.val();  // Returns W + A @ B * scaling (if LoRA attached/enabled)
```

Would lora matrices/`Param`s and configurations be nested inside the base `Param`, since having this one layer up would not be generalized as you describe.

Something like this?
```rust
 pub struct Param<T: Parameter> {
      pub id: ParamId,
      // Base weight W (frozen)
      state: OnceCell<T>,                    

      // LoRA state
      lora_a: Option<Param<T>>,
      lora_b: Option<Param<T>>,
      lora_scaling: f64,

      // ... other fields
  }
```

But this would drastically change the visiting logic traversal, and would turn this into a recursive type which is probably a terrible idea.

Did you have something clever/cleaner in mind?

nathanielsimard:
> 
> Something like this?
> 
> ```rust
>  pub struct Param<T: Parameter> {
>       pub id: ParamId,
>       // Base weight W (frozen)
>       state: OnceCell<T>,                    
> 
>       // LoRA state
>       lora_a: Option<Param<T>>,
>       lora_b: Option<Param<T>>,
>       lora_scaling: f64,
> 
>       // ... other fields
>   }
> ```
> 
> But this would drastically change the visiting logic traversal, and would turn this into a recursive type which is probably a terrible idea.
> 
> Did you have something clever/cleaner in mind?

I would not make it recursive, but yes I would put extra tensors in the paramter, but hardcoded for simplicity:


```rust
  trait Parameter {
     type LoraParams
  }
  pub struct Param<T: Parameter> {
       pub id: ParamId,
       // Base weight W (frozen)
       state: OnceCell<T>,                    
 
       // LoRA state
       loras: Option<P::LoraParams>,
       // ... other fields
   }
```

Then when we implement parameter for `Tensor` we can select the right lora parameters.
I'm not saying this is an easy feature to implement, but if we can make it work, the generalization is great, users won't have to modify their modules to support Lora training.

snowmead:
@nathanielsimard 

> > Something like this?
> > ```rust
> >  pub struct Param<T: Parameter> {
> >       pub id: ParamId,
> >       // Base weight W (frozen)
> >       state: OnceCell<T>,                    
> > 
> >       // LoRA state
> >       lora_a: Option<Param<T>>,
> >       lora_b: Option<Param<T>>,
> >       lora_scaling: f64,
> > 
> >       // ... other fields
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But this would drastically change the visiting logic traversal, and would turn this into a recursive type which is probably a terrible idea.
> > Did you have something clever/cleaner in mind?
> 
> I would not make it recursive, but yes I would put extra tensors in the paramter, but hardcoded for simplicity:
> 
> ```rust
>   trait Parameter {
>      type LoraParams
>   }
>   pub struct Param<T: Parameter> {
>        pub id: ParamId,
>        // Base weight W (frozen)
>        state: OnceCell<T>,                    
>  
>        // LoRA state
>        loras: Option<P::LoraParams>,
>        // ... other fields
>    }
> ```
> 
> Then when we implement parameter for `Tensor` we can select the right lora parameters. I'm not saying this is an easy feature to implement, but if we can make it work, the generalization is great, users won't have to modify their modules to support Lora training.

Completely agree with you, I attempted this originally because I thought maybe it wouldn't be such a burden for users, but also because I didn't think we would have wanted LoRA embedded in the Parameters directly since I believed the Parameter shouldn't know about any particular training methods.

I will look into the architecture a little more in the direction you were thinking, devise something feasible with the same level of control and features so we can discuss about it further before implementation.

Thanks for your input.

Redhawk18:
Are there any updates? I liked to help out if needed.

antimora:
Closing it as stale. Please open a new PR with the fresh rebase. If anyone else wants to pick, you may also submit a new PR. 

CC @Redhawk18, you can open a new one

Redhawk18:
@snowmead Are you still working on this?

snowmead:
Hey @Redhawk18 not actively, a lot has come up lately and couldn't find time to dedicate to it.

Redhawk18:
Thanks, don't delete your branch maybe I'll have the time to work on this as well.

Issue #4326, opened 2026-01-14
Scalar APIs incorrectly downcast to low-precision type even when dtype is higher precision


wingertge:
The `fusion::tensor::f16::ty::ops::full::test_tensor_full_options` fails because the value is cast to the default dtype (`f16` here) even for `f32` output. This also applies to all other scalar operations and fixing it will require changes to the `Backend` trait API.

PR #4337, opened 2026-01-16
Add `Scalar` runtime literal

laggui:
### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/4326

The scalar operations currently used the backend's default element type, which could potentially convert the literal to the wrong type.

### Changes

Added `Scalar` runtime literal to represent the value based on the provided `DType`.
```rust
/// A scalar element.
pub enum Scalar {
    Float(f64),
    Int(i64),
    UInt(u64),
    Bool(bool),
}
```

The representation is created at the high-level tensor API, so all backend ops traits now require `Scalar` instead of an elem associated type or `E: ElementConversion`, e.g.:

```rust
impl<B, const D: usize, K> Tensor<B, D, K> {
    pub fn add_scalar<E: ElementConversion>(self, other: E) -> Self {
        let other = Scalar::new(other, &self.dtype());
        Self::new(K::add_scalar(self.primitive, other))
    }
}
```

```diff
pub trait Numeric<B: Backend>: BasicOps<B> {
-    fn add_scalar<E: ElementConversion>(lhs: Self::Primitive, rhs: E) -> Self::Primitive;
+    fn add_scalar(lhs: Self::Primitive, rhs: Scalar) -> Self::Primitive;
```

```diff
pub trait FloatTensorOps<B: Backend> {
-    fn float_add_scalar(lhs: FloatTensor<B>, rhs: FloatElem<B>) -> FloatTensor<B>;
+    fn float_add_scalar(lhs: FloatTensor<B>, rhs: Scalar) -> FloatTensor<B>;
```

I also simplified the `ScalarIr` to have a similar representation for `Float`/`Int`/`UInt`/`Bool` only. Looks similar to `Scalar`, but kept the types disjoint for now since they serve different roles.

### Testing

Unit tests

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/full.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/full.rs
@@ -22,7 +22,6 @@ fn test_tensor_full_options() {
     let tensor = TestTensor::<2>::full([2, 3], 2.1, (&Default::default(), DType::F32));
     assert_eq!(tensor.dtype(), DType::F32);
 
-    // TODO: `Tensor::full` should not use `FloatElem<B>` for `fill_value`.
     tensor
         .into_data()
         .assert_eq(&TensorData::from([[2.1, 2.1, 2.1], [2.1, 2.1, 2.1]]), false);


wingertge:
I don't know if this is the right approach, since scalars are on CPU I think the easiest solution is to use the appropriate type that can contain every other type (i.e. `f64`) for the intermediate storage and then convert down in the backend. CubeCL does this automatically with `InputScalar`.

laggui:
> I don't know if this is the right approach, since scalars are on CPU I think the easiest solution is to use the appropriate type that can contain every other type (i.e. `f64`) for the intermediate storage and then convert down in the backend. CubeCL does this automatically with `InputScalar`.

That was my first thought too, but it might be lossy depending on the tensor dtype and provided scalar. For example, neither `i64` or `u64` can represent the full range of values from one another. That's why I went with a `Scalar` enum instead, which can then be lowered in the backends.

Pretty similar to `InputScalar` actually, though in CubeCL we use a tagged byte buffer to preserve the original representation until it's used.

But still debating, so that's exactly why I wanted to get some feedback before making the change across all scalar methods :sweat_smile:

wingertge:
> That was my first thought too, but it might be lossy depending on the tensor dtype and provided scalar. For example, neither `i64` or `u64` can represent the full range of values from one another. That's why I went with a `Scalar` enum instead, which can then be lowered in the backends.

Yeah I realized ints can be either signed or unsigned but use the same API. Maybe a reduced hybrid of the approaches then, similar to `ConstantValue` in `cubecl-ir`?
```rust
pub enum ConstantValue {
    Int(i64),
    Float(f64),
    UInt(u64),
    Bool(bool),
}
```
Tagged byte buffers don't really make sense here, they only make sense for `InputScalar` because those need to be transferred into a byte array anyways so it just moves the conversion up a bit. That's not necessarily the case for these scalars.

laggui:
> Tagged byte buffers don't really make sense here, they only make sense for `InputScalar` because those need to be transferred into a byte array anyways so it just moves the conversion up a bit. That's not necessarily the case for these scalars.

Yeah, agreed.

> Yeah I realized ints can be either signed or unsigned but use the same API. Maybe a reduced hybrid of the approaches then, similar to `ConstantValue` in `cubecl-ir`?
> 
> ```rust
> pub enum ConstantValue {
>     Int(i64),
>     Float(f64),
>     UInt(u64),
>     Bool(bool),
> }
> ```

Ahhh yep that's a good balance!

antimora:
With these changes @laggui, @nathanielsimard, are we closer to https://github.com/tracel-ai/burn/issues/1689#issuecomment-2078212769? We talked about having native scalars and tensors with zero ranks. 

Also, does this PR impact ONNX Scalar types?

laggui:
These changes directly address incorrect element type handling for tensor scalar operations. Previously we always represented scalars with the default backend element type, but tensors can be created with a different dtype or cast. So it would possible lose precision (see linked issue, where `2.1` f32 value was being converted to f16 which cannot represent the value exactly).

The public facing API still accepts a generic `E: Element` for scalars, which are converted to this representation (see OP summary). So there is no impact for ONNX.

It also doesn't directly make any progress towards 0 size tensors (scalar), though the addition of a `Scalar` representation does make things more flexible across the board. So I guess we made a tiny step forward lol

Issue #4312, opened 2026-01-13
Image quality metrics


torsteingrindvik:
# Feature description

Add metrics for assessing image quality, including deep learning metrics.

- [x] L1/MAE loss, [torch](https://docs.pytorch.org/docs/stable/generated/torch.nn.L1Loss.html), [torch-image-metrics](https://github.com/mdipcit/torch-image-metrics?tab=readme-ov-file#-supported-metrics)
  - Closed via  https://github.com/tracel-ai/burn/pull/4341
- [x] L2/MSE loss, [torch](https://docs.pytorch.org/docs/stable/generated/torch.nn.MSELoss.html), [torch-image-metrics](https://github.com/mdipcit/torch-image-metrics?tab=readme-ov-file#-supported-metrics)
  - Closed via https://github.com/tracel-ai/burn/pull/4341 
- [x] PSNR, [Wikipedia](https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio), [scikit-image](https://scikit-image.org/docs/0.25.x/api/skimage.metrics.html#skimage.metrics.peak_signal_noise_ratio), [PyTorch-ignite](https://docs.pytorch.org/ignite/generated/ignite.metrics.PSNR.html#ignite.metrics.PSNR), [torch-image-metrics](https://github.com/mdipcit/torch-image-metrics?tab=readme-ov-file#-supported-metrics)
  - Closed via https://github.com/tracel-ai/burn/pull/4379
- [x] SSIM, [Wikipedia](https://en.wikipedia.org/wiki/Structural_similarity_index_measure), [scikit-image](https://scikit-image.org/docs/0.25.x/api/skimage.metrics.html#skimage.metrics.structural_similarity), [PyTorch-ignite](https://docs.pytorch.org/ignite/generated/ignite.metrics.SSIM.html#ignite.metrics.SSIM), [torch-image-metrics](https://github.com/mdipcit/torch-image-metrics?tab=readme-ov-file#-supported-metrics)
  - Closed via https://github.com/tracel-ai/burn/pull/4396
- [x] MS-SSIM,  [lightning.ai](https://lightning.ai/docs/torchmetrics/stable/image/multi_scale_structural_similarity.html), [pytorch-msssim](https://github.com/VainF/pytorch-msssim)
  - Closed via https://github.com/tracel-ai/burn/pull/4555
[...]

PR #4341, opened 2026-01-16
Implement Mean(L(P) Norm Error)Loss

softmaximalist:
### Checklist

### Related Issues/PRs

### Changes

Implemented a generalized Mean(L(P) Norm Error)Loss in the new file `crates/burn-nn/src/loss/mean_lp_error.rs`. It accepts positive p values where p = 1 results in MAE and p = 2 results in MSE. Additionally implemented 3 forward methods: one for no reduction, one where all elements are aggregated, and one where users can pass a slice of dimensions to reduce over.

### Testing

Added 18 test cases to `crates/burn-nn/src/loss/mean_lp_error.rs` covering different values for p, different tensor shapes, different forwards methods, and different reduction methods. 

### Note

The existing `crates/burn-nn/src/loss/mse.rs` file is not deleted as it can affect many users and existing code.


crates/burn-nn/src/loss/mean_lp_error.rs
```diff
@@ -0,0 +1,622 @@
...
+/// let targets: Tensor<Backend, 2> = /* ground truth */;
+///
+/// // Compute loss with mean reduction
+/// let reduced_loss = mse.forward(predictions, targets, Reduction::Auto);
+///
+/// // Compute loss with no reduction
+/// let unreduced_loss = mse.forward_no_reduction(predictions, targets);
+/// ```
+#[derive(Module, Clone, Debug)]
+pub struct MeanLpErrorLoss {
```
laggui: I think it would be more clear to go with a name like `LpLoss` instead.

1. The `Mean` prefix is not accurate, given that the loss can be parameterized by a `Reduction` arg
2. `Error` is a bit redundant with `Loss` when spelled out entirely

And thus adjust the doc accordingly.

laggui (changes requested):
Thanks for the addition!

I suggest clearing up the name, and perhaps we could add some friendly constructors for standard losses e.g. `LpLossconfig::mae() / l1()` and `LpLossConfig::mse() / l2()` (not all four, but either mae/mse or l1/l2 combos).

Lmk what you think!

crates/burn-nn/src/loss/mean_lp_error.rs
softmaximalist: Great suggestions! I created two constructors l1() and l2() and renamed the structs to LpLoss and LpLossConfig. Also added new tests to make sure the new constructors work the same as setting p = 1.0 or p = 2.0.

PR #4357, opened 2026-01-21
docs: Clarify design goals and performance trade-offs of LU decomposi…

red1239109-cmd:
…tion

## What does this PR do?
This PR adds a design note to the `lu_decomposition` documentation to explicitly state that this implementation prioritizes **numerical correctness** and **backend-agnostic behavior** over raw performance.

## Why is it important?
The current implementation uses an unblocked algorithm, which is excellent for stability and broad backend support but may not match the performance of BLAS-backed or blocked implementations on specific hardware.

By clarifying this design choice in the documentation, we can:
1. **Manage user expectations** regarding performance characteristics.
2. **Guide future contributors** to focus on stability and correctness rather than attempting aggressive optimizations (e.g., forcing blocked LU) that might compromise backend compatibility.

### Checklist

### Related Issues/PRs

### Changes

### Testing

### DIFF
--- a/crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
+++ b/crates/burn-tensor/src/tensor/linalg/lu_decomposition.rs
@@ -2,6 +2,7 @@ use crate::{
     Int, backend::Backend, cast::ToElement, check, check::TensorCheck, linalg::swap_slices, s,
     tensor::Tensor,
 };
+
 /// Performs PLU decomposition of a square matrix.
 ///
 /// The function decomposes a given square matrix `A` into three matrices: a permutation vector `p`,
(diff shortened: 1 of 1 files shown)


laggui (changes requested):
Hey, I'm not sure this change clarifies much actually :sweat_smile: 

> 1. **Manage user expectations** regarding performance characteristics.

Doesn't the performance note already address that?

> 2. **Guide future contributors** to focus on stability and correctness rather than attempting aggressive optimizations (e.g., forcing blocked LU) that might compromise backend compatibility.

Actually, contributions are more than welcomed for backend-optimized implementations!

The current implementation is just the default for now as it has not been specialized.

PR #4373, opened 2026-01-23
Feat/device policy

### LINKED ISSUE #3642: Default tensor dtype can be configured on the device instead of by a type.
Instead of defining the default dtype on the backend implementation type, we could set it as a context.

```rust
type Backend = Cuda<bf16>;
```

to

```rust
let device: Device<Cuda> = Default::default();
device.dtype_global(DType::BF16);
let tensor = Tensor::random(...);
assert_eq!(tensor.dtype(), DType::BF16);
```

or

```rust
let device: Device<Cuda> = Default::default();
device.with_dtype(DType::BF16, || {
    let tensor = Tensor::random(...);
    assert_eq!(tensor.dtype(), DType::BF16);
});
```

The challenge would be to define the default behavior: does the default dtype impact the current device across all threads, which could have unintended side effects? It could also be based on the `StreamId` (so the thread ID on native). The scoped approach also fixes the problem of duplicating all tensor initialization APIs to allow for a custom default dtype.

laggui:
### Checklist

### Related Issues/PRs

Fixes #3642

### Changes

Step towards moving away from backend default element type generics (e.g., `Cuda<f32, i32>`).

This PR introduces a new API to set per-device default element type(s), e.g.
```rust
// Update the device policy
set_default_dtypes::<B>(&device, DType::F16, DType::I32).expect("Should be supported");

// All float tensors created after this will use F16 by default
let tensor = Tensor::<B, 2>::zeros([2, 3], &device);
```

This is currently set as a public function, but once the multi backend lands it could be a method on a new general device type.

### Testing

Unit tests

### DIFF
--- a/crates/burn-tensor/src/device.rs
+++ b/crates/burn-tensor/src/device.rs
@@ -196,7 +196,8 @@ pub fn set_default_dtypes<B: Backend>(
     check_dtype_support::<B>(device, float_dtype)?;
     check_dtype_support::<B>(device, int_dtype)?;
 
-    Ok(set_default_dtypes_unchecked(device, float_dtype, int_dtype))
+    set_default_dtypes_unchecked(device, float_dtype, int_dtype);
+    Ok(())
 }
 
 /// Sets the default floating-point data type for the device.
@@ -228,7 +229,8 @@ pub fn set_default_float_dtype<B: Backend>(
     let dtype = dtype.into();
     check_dtype_support::<B>(device, dtype)?;
 
-    Ok(set_default_float_dtype_unchecked(device, dtype))
+    set_default_float_dtype_unchecked(device, dtype);
+    Ok(())
 }
 
 /// Sets the default integer data type for the device.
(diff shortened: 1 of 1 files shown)


crates/burn-tensor/src/device.rs
```diff
@@ -0,0 +1,470 @@
...
+///     let tensor = Tensor::<B, 2>::zeros([2, 3], &device);
+///     // All int tensors created after this will use I32 default
+///     let tensor = Tensor::<B, 2, Int>::zeros([2, 3], &device);
+/// }
+/// ```
+pub fn set_default_dtypes<B: Backend>(
+    device: &B::Device,
+    float_dtype: impl Into<FloatDType>,
+    int_dtype: impl Into<IntDType>,
+) {
```
laggui: I was thinking these methods could potentially return a result, but right now the `B::supports_dtype` isn't granular enough. Think it would make sense to have something a bit more granular like cubecl's `TypeUsage` (but higher-level, so no atomics etc.).

And then define how we should handle that for global default dtype setting, based on type usage.

But for now, it can just be set and the backend would certainly panic on an unsupported dtype.

Lmk if you have any thoughts!

nathanielsimard: Yeah a result would make sense

PR #4379, opened 2026-01-25
Implement the PSNR vision metric

softmaximalist:
### Checklist

### Related Issues/PRs

### Changes

Implemented the PSNR vision metric in the `crates/burn-train/src/metric/vision/psnr.rs` file. It computes the per-image PSNR values and then average them across all the images in the batch. Since the PSNR (in dB) is defined as following:

$$
PSNR = 10 \cdot \log_{10}\left(\frac{MAX_{I}^2}{MSE}\right)
$$

the MSE value is clamped a minimum value of `epsilon` to avoid division by 0. The `epsilon` field of the `PsnrMetricConfig` struct is set to `1e-10` by default. Users can set a custom/different epsilon value as well. 

### Testing

Added 18 test cases to the `psnr.rs` file. Tests cover different input shapes, different PSNR values, different image representation (where maximum possible pixel values are different), multichannel tensors, running average, setting custom name and/or epsilon, panics, etc. The test cases verify actual expected values rather than just ranges (which is approach taken by the other open PR).

### Note

- Right after creating this PR, I noticed that there is another open PR also implementing PSNR (https://github.com/tracel-ai/burn/pull/4377). 
- Below are the main differences (that I noticed) between my PR and the other PR:
    - My test cases verify very specific expected values while the other PR verifies if the psnr is within some range. 
    - The other PR implements a default method for the metric which I don't. This is because the PSNR metric always needs have the `max_pixel_val` set but this value depends on the specific image format used. Depending on whether the images are normalized to [0, 1] range, or they are 8, 16, 32, etc bits, this value changes. Hence, in my code, the user who is aware of their specific image formats must always set this value. 
    - My code only accepts 4d tensors with the shape [N, C, H, W] while the other PR accepts tensors with more than 1 dimension (dim >= 2)
        - I designed the code this way since I did not want the users to misuse the metric due to ambiguity. 
        - I might be wrong but I believe the implementation in the other PR is flawed. While the code in the other PR accepts tensors with more than 1 dimension, their `update()` method always treats the first dimension (dim 0) as the batches. This does not work in many different cases such as when the input tensors have shapes such as [H, W], [C, H, W], etc which are still valid shapes in their code. 
    - While the approach taken for computing MSE and also dealing with division by zero are different between the two PRs, they should not affect the actual functionality of the code.

### DIFF
--- a/crates/burn-train/src/metric/vision/psnr.rs
+++ b/crates/burn-train/src/metric/vision/psnr.rs
@@ -7,7 +7,6 @@ use burn_core::{
     prelude::{Backend, Tensor},
     tensor::ElementConversion,
 };
-use burn_nn::loss::{LpLoss, LpLossConfig};
 use core::marker::PhantomData;
 use std::f64::consts::LN_10;
 
@@ -19,9 +18,9 @@ use std::f64::consts::LN_10;
 /// - `H`: Height
 /// - `W`: Width
 pub struct PsnrInput<B: Backend> {
-    /// Model output (predictions/reconstructions) images with shape [N, C, H, W].
+    /// Model output (predictions/reconstructions) images with shape `[N, C, H, W]`.
     outputs: Tensor<B, 4>,
-    /// Ground truth images with shape [N, C, H, W].
+    /// Ground truth images with shape `[N, C, H, W]`.
     targets: Tensor<B, 4>,
 }
 
@@ -33,8 +32,8 @@ impl<B: Backend> PsnrInput<B> {
     /// `H` is the height of the image, and `W` is the width of the image.
     ///
     /// # Arguments
-    /// - `outputs`: The model output images with shape [N, C, H, W].
-    /// - `targets`: The ground truth images with shape [N, C, H, W].
+    /// - `outputs`: The model output images with shape `[N, C, H, W]`.
+    /// - `targets`: The ground truth images with shape `[N, C, H, W]`.
     ///
     /// # Returns
     /// A new instance of `PsnrInput`.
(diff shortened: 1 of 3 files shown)


crates/burn-train/src/metric/vision/psnr.rs
```diff
@@ -0,0 +1,615 @@
...
+        let dims = item.outputs.dims();
+        let batch_size = dims[0];
+        let outputs = item.outputs.clone();
+        let targets = item.targets.clone();
+
+        // Compute per-image MSE by reducing over all dimensions except batch (dims 1, 2, 3)
+        // Resulting shape: [N, 1, 1, 1]
+        let mse_per_image = self
+            .l2_loss
+            .forward_reduce_dims(outputs, targets, &[1, 2, 3]);
```
laggui: While re-use is generally encouraged, I think in this case a more functional approach is more idiomatic. Avoids pulling a `burn-nn` dependency in `burn-train` just to compute the MSE. It can be implemented in two lines:

```rust
let diff = item.outputs.clone().sub(item.targets.clone());
let mse_per_image = diff.powi_scalar(2).mean_dims(&[1, 2, 3]);
```

crates/burn-train/src/metric/vision/psnr.rs
```diff
@@ -0,0 +1,615 @@
...
+    /// Creates a new PSNR metric with the given configuration.
+    ///
+    /// # Example
+    /// ```ignore
+    /// let config = PsnrMetricConfig::new(1.0);
+    /// let metric = PsnrMetric::<B>::new(config);
+    /// ```
+    pub fn new(config: PsnrMetricConfig) -> Self {
+        Self {
+            name: MetricName::new("PSNR".to_string()),
```
laggui: Can different PSNR variants be reported for a single run? If so, I would make the name parameterized by the config, e.g.

```rust
format!("PSNR@{}", config.max_pixel_val)
```

to avoid collisions if multiple PSNR variants are logged in the same run (e.g. `PSNR@1.0` vs `PSNR@255`)

softmaximalist:
Nice suggestions! I updated the code to not rely on burn-nn for MSE computation and also updated the metric name.

PR #4389, opened 2026-01-26
chore: Enable macos CI

dcvz:
Enables macOS CI (tested on my fork) -- for some reason the workflow only had it enabled on the test-gpu job which is manually triggered.. didn't move it.


dcvz:
@laggui yes, I can do that in a follow-up because that may require fixing this at least gets macos running again.
also thanks, fixed the link!

PR #4391, opened 2026-01-27
fix(burn-cubecl-fusion): use output as reference layout fallback when no input available

### LINKED ISSUE #4347: Tensor Handle Panic with Fusion Feature
**Describe the bug**

  During training with Burn 0.20, a panic occurs related to tensor handle management when using the `fusion` feature. This affects both CUDA and W

  Error message:

  thread '<unnamed>' (143357) panicked at /home/user/.cargo/registry/src/index.crates.io-xxx/burn-ir-0.20.0/src/handle.rs:88:32:
  Should have handle for tensor TensorId { value: 497010 }

  thread 'main' (141484) panicked at /home/user/.cargo/registry/src/index.crates.io-xxx/burn-train-0.20.0/src/checkpoint/async_checkpoint.rs:123:14:
  Can send message to checkpointer thread.: SendError { .. }

  thread 'main' (141484) panicked at /home/user/.cargo/registry/src/index.crates.io-xxx/burn-train-0.20.0/src/checkpoint/async_checkpoint.rs:162:14:
  Can send the end message to the checkpointer thread.: SendError { .. }

  thread 'main' panicked at library/core/src/panicking.rs:238:5:
  panic in a destructor during cleanup
  thread caused non-unwinding panic. aborting.

  **To Reproduce**

  1. Create a model using `Autodiff<Fusion<CubeBackend<CudaRuntime>>>` or `Autodiff<Fusion<Wgpu>>` backend
  2. Configure features in Cargo.toml:
     ```toml
     [features]
     cuda = ["burn/cuda", "burn/fusion"]
     wgpu = ["burn/wgpu", "burn/fusion"]
  3. Train with SupervisedTraining API:
    - batch_size: 4-10
    - Input tensor shape: [batch, 1, 32768] (audio samples)
    - Multiple epochs with checkpointing enabled
  4. Panic occurs randomly during training, typically between epoch 30-60

  Expected behavior

  Training should complete without panic when using the fusion feature.

  Workaround

  Disabling the fusion feature resolves the issue:
  cuda-no-fusion = ["burn/cuda"]  # Works
  wgpu-no-fusion = ["burn/wgpu"]  # Works

  Desktop (please complete the following information):
  - OS: Ubuntu Linux 24.04 (kernel 6.14.0-37-generic)
  - GPU: NVIDIA GeForce RTX 4070 (12GB)
  - CUDA Version: 13.0
  - Driver Version: 580.95.05
  - Rust Edition: 2024
  - Burn Version: 0.20.0

  **Additional context**

  - Error occurs with both CUDA and WGPU backends when fusion is enabled
  - Panic often triggers during checkpoint saving, causing secondary errors:
  thread 'main' panicked at burn-train-0.20.0/src/checkpoint/async_checkpoint.rs:123:14:
  Can send message to checkpointer thread.: SendError { .. }
  - Model architecture: TCN (Temporal Convolutional Network) with ~100K parameters
  - Training uses ESR (Error-to-Signal Ratio) loss function
  - The issue appears to be in burn-ir handle management, which is shared between fusion backends

Dreaming-Codes:
## Summary

Fixes #4347

When loading saved models with the fused Wgpu backend, the fusion planner couldn't find a valid reference layout because there were no input tensors in the data flow. This caused a `ReferenceNotFound` panic.

## Changes

Added a fallback in `select_reference_from_inputs` that uses an output tensor from the current block as the reference layout when no input reference is available.

- Only considers outputs written by the current block
- Respects `ref_layout_setting` (Concrete vs VirtualShape based on contiguity)
- Follows the same pattern as input reference selection

### DIFF
--- a/crates/burn-cubecl-fusion/src/engine/launch/output.rs
+++ b/crates/burn-cubecl-fusion/src/engine/launch/output.rs
@@ -164,6 +164,7 @@ impl<'a, R: Runtime> OutputPlanner<'a, R> {
                     self.blocks[i].settings.ref_layout,
                     block,
                     &plan.handle_inputs,
+                    &plan.handle_outputs,
                 );
             } else {
                 Self::add_layout_info_inputs(block, &plan.handle_inputs);
@@ -185,6 +186,7 @@ impl<'a, R: Runtime> OutputPlanner<'a, R> {
         ref_layout_setting: RefLayoutSetting,
         block: &mut BlockPlan<'_>,
         handle_inputs: &[HandleInput<R>],
+        handle_outputs: &[HandleOutput<R>],
     ) {
         if let Some(input_ref) = block.potential_reference_input.take() {
             match input_ref {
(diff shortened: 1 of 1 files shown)


nathanielsimard:
The outputs are actually checked first when looking for a reference layout, so I don't think this will fix the problem.

Dreaming-Codes:
@nathanielsimard 
> The outputs are actually checked first when looking for a reference layout, so I don't think this will fix the problem.

Not sure if it's the same issue of #4347  then cause it fixed a similar issue for me in https://github.com/Dreaming-Codes/joy-o-meter . If you try running that without my fork it crashes

nathanielsimard:
I have a proper fix in an upcoming PR that passes the shape and strides from the host when we can't use tensor metadata for them.

Dreaming-Codes:
> I have a proper fix in an upcoming PR that passes the shape and strides from the host when we can't use tensor metadata for them.

Perfect, I'll close this then

PR #4396, opened 2026-01-28
Implement SSIM vision metric

softmaximalist:
### Checklist

### Related Issues/PRs

### Changes

Implemented the SSIM vision metric with the following parameters
- data_range
- k1
- k2
- window_size
- sigma

The metric implements the following simplified version of the SSIM formula.

$$
\text{SSIM}(x, y) = \frac{(2\mu_x\mu_y + c_1)(2\sigma_{xy} + c_2)}{(\mu_x^2 + \mu_y^2 + c_1)(\sigma_x^2 + \sigma_y^2 + c_2)}
$$

The mean, variance, and covariance values in the formula above are computed by performing convolution with a Gaussian kernel. The implementation specifically uses separable Gaussian convolution for efficiency. Instead of applying a single 2D convlution with a $K$ by $K$ kernel, it applies two 1D convolutions. This reduces the computational complexity from $\mathcal{O}(K^2)$ to $\mathcal{O}(2K)$ for each pixel. Note that the SSIM is computed for each image first. Then, these SSIM values are averaged across all the images in the batch. 

### Testing

Implemented 24 test cases covering different values for the 5 paremeters mentioned above. It also covers some mathematical properties of the SSIM metric such as symmetry. Metric methods such as `clear` and `attributes` are also tested in addition to panics.

### DIFF
--- a/crates/burn-train/src/metric/vision/ssim.rs
+++ b/crates/burn-train/src/metric/vision/ssim.rs
@@ -192,10 +192,10 @@ impl<B: Backend> SsimMetric<B> {
     /// Creates a new SSIM metric with the given configuration.
     ///
     /// # Note
-    /// The metric name format is "SSIM\_dr{}\_w{}\_σ{:.1}\_k{:.3}\_{:.3}"
+    /// The metric name format is "SSIM (dr={}, w={}, σ={})"
     /// where dr is the data range, w is the window size, sigma is the
-    /// standard deviation, the two values following k are k1 and k2 respectively.
-    /// For example, the metric name might be "SSIM\_dr1.0\_w11\_σ1.5\_k0.01\_0.03".
+    /// standard deviation. For example, the metric name might be
+    /// "SSIM (dr=1.0, w=11, σ=1.5)".
     ///
     /// # Example
     /// ```ignore


crates/burn-train/src/metric/vision/ssim.rs
```diff
@@ -0,0 +1,857 @@
...
+    /// ```ignore
+    /// let ssim_config = SsimMetricConfig::new(1.0);
+    /// let ssim_metric = SsimMetric::<B>::new(ssim_config);
+    /// ```
+    pub fn new(config: SsimMetricConfig) -> Self {
+        Self {
+            name: MetricName::new(format!(
+                "SSIM_dr{}_w{}_σ{:.1}_k{:.3}_{:.3}",
+                config.data_range, config.window_size, config.sigma, config.k1, config.k2
+            )),
```
laggui, line 210: That might be a little verbose 😅 

I think it should remain readable, functional and representative of what users typically change. k1 and k2 are algorithm parameters, rarely changed from what it seems.

Suggestion:
```rust
format!("SSIM (dr={}, w={}, σ={})", config.data_range, config_window_size, config.sigma)
```

e.g. `SSIM (dr=1.0, w=11, σ=1.5)` is already better.

softmaximalist: Good suggestion. I was not sure how detailed it should be so I just tried to include all the parameters 😂

PR #4399, opened 2026-01-28
add KLDivLoss and batch_mean in reduction

donjuanplatinum:
### Checklist

### Related Issues/PRs

### Changes

1. KLDivLoss and their tests
2. Reduction::BatchMean
3. changing cosine_embedding,huber,lp_loss,mse,poisson match reduction
### Testing

This is the pytorch test file:
```python
import torch
import torch.nn.functional as F

torch.set_printoptions(precision=10)

def get_high_precision_results():
    predict = torch.tensor([[-1.0, -0.5], [-2.0, -0.2]], dtype=torch.float64)
    targets = torch.tensor([[0.4, 0.6], [0.1, 0.9]], dtype=torch.float64)

    print("=== Scenario: LogTarget = False (Double Precision) ===")
    print(f"No Reduction:\n{F.kl_div(predict, targets, reduction='none', log_target=False)}")
    print(f"Sum: {F.kl_div(predict, targets, reduction='sum', log_target=False).item():.12f}")
    print(f"BatchMean: {F.kl_div(predict, targets, reduction='batchmean', log_target=False).item():.12f}")
    print(f"Mean (PyTorch standard): {F.kl_div(predict, targets, reduction='mean', log_target=False).item():.12f}")

    predict_log = torch.tensor([-1.0, -2.0], dtype=torch.float64)
    targets_log = torch.tensor([-0.5, -1.5], dtype=torch.float64)

    print("\n=== Scenario: LogTarget = True (Double Precision) ===")
    print(f"No Reduction: {F.kl_div(predict_log, targets_log, reduction='none', log_target=True)}")
    print(f"Sum: {F.kl_div(predict_log, targets_log, reduction='sum', log_target=True).item():.12f}")
    print(f"BatchMean: {F.kl_div(predict_log, targets_log, reduction='batchmean', log_target=True).item():.12f}")

    predict_ad = torch.tensor([[-1.0, -0.5]], dtype=torch.float64, requires_grad=True)
    targets_ad = torch.tensor([[0.4, 0.6]], dtype=torch.float64)
    loss = F.kl_div(predict_ad, targets_ad, reduction='sum', log_target=False)
    loss.backward()
    print("\n=== Gradients (log_target=False) ===")
    print(f"Predict Grad: {predict_ad.grad}")

if __name__ == "__main__":
    get_high_precision_results()

```

and this is the test results
image

image

### DIFF
--- a/crates/burn-nn/src/loss/kldiv.rs
+++ b/crates/burn-nn/src/loss/kldiv.rs
@@ -57,6 +57,9 @@ impl ModuleDisplay for KLDivLoss {
 
 impl KLDivLoss {
     /// Compute the criterion on the input tensor.
+    ///
+    /// `Reduction::Auto` behaves as `Reduction::BatchMean`,`Reduction::Mean` dose not align with the math definition.
+    ///
     /// # Shapes
     ///
     /// - predictions: \[batch_size,num_targets\]
@@ -74,10 +77,7 @@ impl KLDivLoss {
                 let batch_size = loss.dims()[0] as f32;
                 loss.sum().div_scalar(batch_size)
             }
-            Reduction::Mean => {
-                // log::warn!("reduction: 'Reduction::Mean' divides the total loss by both the batch size and the support size.'Reduction::BatchMean' divides only by the batch size, and aligns with the KL div math definition.");
-                loss.mean()
-            }
+            Reduction::Mean => loss.mean(),
             Reduction::Sum => loss.sum(),
         }
     }


crates/burn-nn/src/loss/kldiv.rs
```diff
@@ -0,0 +1,200 @@
...
+        reduction: Reduction,
+    ) -> Tensor<B, 1> {
+        let loss = self.forward_no_reduction(predictions, targets);
+        match reduction {
+            Reduction::BatchMean | Reduction::Auto => {
+                let batch_size = loss.dims()[0] as f32;
+                loss.sum().div_scalar(batch_size)
+            }
+            Reduction::Mean => {
+                // log::warn!("reduction: 'Reduction::Mean' divides the total loss by both the batch size and the support size.'Reduction::BatchMean' divides only by the batch size, and aligns with the KL div math definition.");
```
laggui: Dead code

crates/burn-nn/src/loss/kldiv.rs
```diff
@@ -0,0 +1,200 @@
...
+/// Kullback-Leibler Divergence Loss
+///
+/// KL Divergence shows the difference between two probability distributions by measuring information loss
+///
+/// KLDivLoss =
+/// ```tex
+/// y_{true} \cdot (\log{y_{true}} - \log{y_{pred}})
+///     ```
+/// By Default,this loss expects the argument `inp` in the log-space, and the output will be applied `batchmean`.
+/// The argument `target` may also be provided in the log-space if `log_target` = true
```
laggui: You mention a default for batch mean, but the `KLDivLoss` doesn't have a default 😅 the forward method takes a reduction arg instead. There is also no `inp` argument.

So I would remove the mention, and clean up the doc:


```suggestion
/// By default, the loss expects the input in the log-space.
/// The targets may also be provided in the log-space if `log_target` is true.
```

crates/burn-nn/src/loss/kldiv.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl KLDivLoss {
+    /// Compute the criterion on the input tensor.
+    /// # Shapes
+    ///
+    /// - predictions: \[batch_size,num_targets\]
+    /// - targets: \[batch_size,num_targets\]
+    /// - output: \[1\]
```
laggui, line 67: I would add a note regarding the `reduction` arg to warn that using `Reduction::Mean` does not align with the math definition.

crates/burn-nn/src/loss/kldiv.rs
donjuanplatinum: > You mention a default for batch mean, but the `KLDivLoss` doesn't have a default 😅 the forward method takes a reduction arg instead. There is also no `inp` argument.
> 
> So I would remove the mention, and clean up the doc:

I’m sorry — I copied the documentation from Candle’s implementation and didn’t notice the issue.

PR #4402, opened 2026-01-28
opt(burn-cubecl): Optimized tensors by default

wingertge:
### Checklist

### Changes

Ensure kernels work for non-contiguous tensors, and makes tensors optimized by default. This improves performance by aligning rows to cache lines.

### Testing

All tests pass and `burn-lm` works perfectly.

### DIFF
--- a/crates/burn-candle/src/backend.rs
+++ b/crates/burn-candle/src/backend.rs
@@ -284,11 +284,11 @@ mod tests {
         assert!(B::supports_dtype(&device, DType::I64));
         assert!(B::supports_dtype(&device, DType::U32));
         assert!(B::supports_dtype(&device, DType::U8));
+        assert!(B::supports_dtype(&device, DType::I32));
+        assert!(B::supports_dtype(&device, DType::I16));
 
         assert!(!B::supports_dtype(&device, DType::U64));
         assert!(!B::supports_dtype(&device, DType::U16));
-        assert!(!B::supports_dtype(&device, DType::I32));
-        assert!(!B::supports_dtype(&device, DType::I16));
         assert!(!B::supports_dtype(&device, DType::I8));
         assert!(!B::supports_dtype(&device, DType::Bool));
         assert!(!B::supports_dtype(


crates/burn-candle/src/ops/base.rs
```diff
@@ -49,12 +49,19 @@ pub fn into_data(tensor: CandleTensor) -> Result<TensorData, ExecutionError> {
     }
 
     match tensor.tensor.dtype() {
+        candle_core::DType::F8E4M3
+        | candle_core::DType::F6E2M3
+        | candle_core::DType::F6E3M2
+        | candle_core::DType::F4
+        | candle_core::DType::F8E8M0 => todo!("Not yet supported"),
```
laggui: Someone pointed this out this morning in https://github.com/tracel-ai/burn/issues/4398

Looks like a patch release introduced new variants, which can potentially break user code..

I think we should have a catch all instead for the unsupported types (in case it breaks again.. :sweat_smile: )

crates/burn-candle/src/tensor.rs
```diff
@@ -19,6 +19,13 @@ impl TensorMetadata for CandleTensor {
             candle_core::DType::F16 => DType::F16,
             candle_core::DType::F32 => DType::F32,
             candle_core::DType::F64 => DType::F64,
+            candle_core::DType::I16 => DType::I16,
+            candle_core::DType::I32 => DType::I32,
+            candle_core::DType::F8E4M3
+            | candle_core::DType::F6E2M3
+            | candle_core::DType::F6E3M2
+            | candle_core::DType::F8E8M0
+            | candle_core::DType::F4 => todo!("Not yet supported"),
```
laggui: Same here.

And for I32 / I16 support we should extend this:

https://github.com/tracel-ai/burn/blob/8d381158e6d662f5acba9df35c3911c7e9db767e/crates/burn-candle/src/tensor.rs#L93-L107

which is used to check for dtype support in some ops like cast

crates/burn-collective/src/local/server.rs
```diff
@@ -271,6 +271,7 @@ impl<B: Backend> LocalCollectiveServer<B> {
                 .unwrap();
             return;
         }
+        #[allow(clippy::unnecessary_unwrap)]
```
laggui: Was fixed on main! The clippy attribute should no longer be necessary 🙂

crates/burn-cubecl/src/kernel/pool/adaptive_avg_pool2d.rs
```diff
@@ -42,24 +43,23 @@ fn adaptive_avg_pool2d_direct<E: Numeric>(
...
+            let index_input = index_input_base + index_input_2 + index_input_3;
             sum += input[index_input / input.line_size()];
         }
     }
 
     let num_ih = ih_end - ih_start;
     let num_iw = iw_end - iw_start;
 
-    output[ABSOLUTE_POS] = sum / Line::cast_from(num_ih * num_iw);
+    output[(b, oh, ow, c)] = sum / Line::cast_from(num_ih * num_iw);
```
laggui, line 61: Nice! `decompose_linear` and indexing via `[b, oh, ow, c]` makes it easier to follow at glance

crates/burn-cubecl/src/kernel/conv/direct.rs
```diff
@@ -317,7 +317,10 @@ pub fn conv_direct<R: CubeRuntime, const N: usize>(
 }
 
 #[cube]
-pub(crate) fn div_mod_seq(pos: u32, shape: &Sequence<FastDivmod<u32>>) -> (u32, Sequence<u32>) {
+pub(crate) fn decompose_linear<I: FastDivmodInt>(
+    pos: I,
+    shape: &Sequence<FastDivmod<I>>,
+) -> (I, Sequence<I>) {
```
laggui: Might be worth moving to another module now that it is used more generally? e.g. `kernel/utils.rs` which already has other views and layout utils

crates/burn-candle/src/tensor.rs
wingertge: Done 👍

laggui:
> ❌ Workflow **test-gpu.yml** Execution Error!

I always forget that the bot doesn't work when trying to execute from forks :smiling_face_with_tear: 

/edit: oh I forgot with the added candle support we need to flip the conditions for the new supported types:

https://github.com/tracel-ai/burn/blob/933fdf4f69ebef1aeea06cf78f17a1d447904bbb/crates/burn-candle/src/backend.rs#L290-L291

PR #4403, opened 2026-01-29
Implements: LPIPS matrics for Image quality

koreaygj:
### Checklist

### Related Issues/PRs

[Image quality metrics #4312 ](https://github.com/tracel-ai/burn/issues/4312)

### Changes

Follow this references 
[lightning.ai](https://lightning.ai/docs/torchmetrics/stable/image/learned_perceptual_image_patch_similarity.html#), [torch-image-metrics](https://github.com/mdipcit/torch-image-metrics?tab=readme-ov-file#-supported-metrics)

Supported Networks: VGG16, AlexNet, SqueezeNet
  
  Configuration Parameters

  - net - Backbone network type (VGG, Alex, Squeeze)
  - normalize - Whether to normalize input from [0,1] to [-1,1]

  Implementation Details

  The LPIPS metric computes perceptual distance using deep features:

  $$\text{LPIPS}(x, y) = \sum_{l} \frac{1}{H_l W_l} \sum_{h,w}  w_l \odot
  (\hat{f}^l_{x,hw} - \hat{f}^l_{y,hw}) ^2_2$$

  Where:
  - $\hat{f}^l$ = unit-normalized features at layer $l$
  - $w_l$ = learned linear weights (1x1 conv)
  - $H_l, W_l$ = spatial dimensions at layer $l$

Implementation

  1. Extracting multi-scale deep features from backbone network
  2. Computing normalized L2 distance between features
  3. Applying learned linear weights (1x1 conv) and spatial averaging

  Supports loading pretrained weights from PyTorch lpips library via `lpips_key_remaps()`

### Testing

Test Name | Description | Backbone
-- | -- | --
test_lpips_identical_images_zero_loss | Identical images → Loss is exactly 0 | VGG
test_lpips_different_images_positive_loss | Different images → Loss is greater than 0 | VGG
test_lpips_symmetry | Verification that $LPIPS(a, b) = LPIPS(b, a)$ | VGG
test_lpips_forward_mean_reduction | Validates output shape when using mean reduction | VGG
test_lpips_forward_no_reduction | Validates output shape when no reduction is applied | VGG
test_lpips_output_non_negative | Ensures the loss value is always $\ge 0$ | VGG
test_lpips_alex_identical_images_zero_loss | Identical images → Loss is exactly 0 | AlexNet
test_lpips_alex_different_images_positive_loss | Different images → Loss is greater than 0 | AlexNet
test_lpips_squeeze_identical_images_zero_loss | Identical images → Loss is exactly 0 | SqueezeNet
test_lpips_squeeze_different_images_positive_loss | Different images → Loss is greater than 0 | SqueezeNet
display_vgg | Validates the display string format | VGG
display_alex | Validates the display string format | AlexNet
display_squeeze | Validates the display string format | SqueezeNet

### Example

**I make the examples for compareing pytorch. If Don't need it I remove this examples** 

File / Directory | Description
-- | --
src/main.rs | Loads PyTorch weights and compares Burn vs. PyTorch outputs.
compare.py | Generates weights and runs the PyTorch side of the comparison.
img/*.raw | Test images stored in 64x64 f32 raw format.
weights/*.pt | Pretrained PyTorch weights (VGG, AlexNet, SqueezeNet).

Network | Test Case | Burn Value | PyTorch Value | Match
-- | -- | -- | -- | --
VGG | zeros vs ones | 0.453837 | 0.453837 | ✓
VGG | horizontal vs vertical | 0.667304 | 0.667304 | ✓
VGG | horizontal vs diagonal | 0.350121 | 0.350121 | ✓
AlexNet | zeros vs ones | 0.813955 | 0.813955 | ✓
AlexNet | horizontal vs vertical | 0.674660 | 0.674660 | ✓
AlexNet | horizontal vs diagonal | 0.269207 | 0.269207 | ✓
SqueezeNet | zeros vs ones | 0.642093 | 0.642093 | ✓
SqueezeNet | horizontal vs vertical | 0.617866 | 0.617866 | ✓
SqueezeNet | horizontal vs diagonal | 0.250041 | 0.250041 | ✓

### DIFF
--- a/crates/burn-train/src/metric/vision/lpips/metric.rs
+++ b/crates/burn-train/src/metric/vision/lpips/metric.rs
@@ -258,8 +258,6 @@ pub struct LpipsSqueeze<B: Backend> {
 impl<B: Backend> LpipsVgg<B> {
     /// Compute LPIPS distance without reduction using VGG backbone.
     pub fn forward_no_reduction(&self, input: Tensor<B, 4>, target: Tensor<B, 4>) -> Tensor<B, 1> {
-        let [_batch, _, _, _] = input.dims();
-
         // Preprocess inputs
         let (input, target) = preprocess_inputs(input, target, self.normalize);
 
@@ -285,8 +283,6 @@ impl<B: Backend> LpipsVgg<B> {
 impl<B: Backend> LpipsAlex<B> {
     /// Compute LPIPS distance without reduction using AlexNet backbone.
     pub fn forward_no_reduction(&self, input: Tensor<B, 4>, target: Tensor<B, 4>) -> Tensor<B, 1> {
-        let [_batch, _, _, _] = input.dims();
-
         // Preprocess inputs
         let (input, target) = preprocess_inputs(input, target, self.normalize);
 
@@ -312,8 +308,6 @@ impl<B: Backend> LpipsAlex<B> {
 impl<B: Backend> LpipsSqueeze<B> {
     /// Compute LPIPS distance without reduction using SqueezeNet backbone.
     pub fn forward_no_reduction(&self, input: Tensor<B, 4>, target: Tensor<B, 4>) -> Tensor<B, 1> {
-        let [_batch, _, _, _] = input.dims();
-
         // Preprocess inputs
         let (input, target) = preprocess_inputs(input, target, self.normalize);
 


crates/burn-nn/src/loss/lpips.rs
```diff
@@ -0,0 +1,1167 @@
...
+/// ```
+#[derive(Module, Debug)]
+#[module(custom_display)]
+pub struct Lpips<B: Backend> {
+    /// VGG feature extractor (used when net=Vgg)
+    vgg: Option<VggFeatureExtractor<B>>,
+    /// AlexNet feature extractor (used when net=Alex)
+    alex: Option<AlexFeatureExtractor<B>>,
+    /// SqueezeNet feature extractor (used when net=Squeeze)
+    squeeze: Option<SqueezeFeatureExtractor<B>>,
```
laggui: Instead of all these optional fields which are tied to the `LpipsNet` field, I would make `Lpips` and enum with variants for vgg / alexnet / squeezenet.

koreaygj:
Thanks for your feedback! 

LPIPS is used both as a **loss function during training** and as an **evaluation metric**. I'll integrate it into `burn-train` so it can be properly utilized in both training and validation loops.

@laggui As you suggested, 
I'm planning to : 
1.  Feature gate the module
2.  Auto-download weights from the offical LPIPS source on initialization


Does `burn-train/metric/vision `sound like the right place for this? Since it needs pretrained weights + network download, I think it fits better there than `burn-nn/loss`. 

Let me know what you think!

laggui:
Yep that sounds like a good plan! We could probably put all vision metrics in the `vision` module behind the feature flag (`vision`).

I would also split the implementation for the metric and backbones, e.g.
```
lpips/
├── alexnet.rs
├── metric.rs
├── mod.rs
├── squeezenet.rs
└── vgg.rs
```

Oh and just a side note, we moved the `burn-import` code to the [burn-onnx](https://github.com/tracel-ai/burn-onnx/tree/main/crates) repo. We are phasing it away to use the `burn-store` adapters instead which are more flexible and provide better debugging utilities for mapping between pytorch <> burn structs.

koreaygj:
Thanks for your suggestion!

Here is my plan..
- [x] Move LPIPS to `burn-train/metric/vision`
- [x] Split into separeate backbone files (vgg.rs, alexnet.rs, squeezenet.rs
- [x] Refactor to use `burn-store` instead of burn-import
- [x] Add vision feature flag
- [x] Auto-download pretrained weights on initialization

koreaygj:
Please review this implements. @laggui

koreaygj:
Thanks for the reply! I'm sorry if I was pushing you while you've been so busy. 😂
I've just rebased my branch to include the CI fixes.

koreaygj:
Rebase onto latest main to fix rl module compilation errors
@laggui

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,834 @@
...
+
+        // Compute distance for each layer
+        let device = feats0[0].device();
+        let mut total_loss = Tensor::zeros([batch], &device);
+
+        total_loss = total_loss.add(compute_layer_distance(&feats0[0], &feats1[0], &self.lin0));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[1], &feats1[1], &self.lin1));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[2], &feats1[2], &self.lin2));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[3], &feats1[3], &self.lin3));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[4], &feats1[4], &self.lin4));
```
laggui: I would replace this pattern by stack + sum, e.g.

```rust
let layer_distances = vec![compute_layer_distances(...), ...];

Tensor::stack(layer_distances, 1).sum_dim(1).squeeze_dim(1)
```

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,834 @@
...
+        total_loss = total_loss.add(compute_layer_distance(&feats0[5], &feats1[5], &self.lin5));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[6], &feats1[6], &self.lin6));
```
laggui: Same

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,834 @@
...
+
+        // Compute distance for each layer
+        let device = feats0[0].device();
+        let mut total_loss = Tensor::zeros([batch], &device);
+
+        total_loss = total_loss.add(compute_layer_distance(&feats0[0], &feats1[0], &self.lin0));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[1], &feats1[1], &self.lin1));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[2], &feats1[2], &self.lin2));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[3], &feats1[3], &self.lin3));
+        total_loss = total_loss.add(compute_layer_distance(&feats0[4], &feats1[4], &self.lin4));
```
laggui: Same

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,834 @@
...
+
+    // Compute squared difference
+    let diff = feat0_norm.sub(feat1_norm);
+    let diff_sq = diff.clone().mul(diff);
+
+    // Apply linear layer (learned weights)
+    // Shape: [batch, C, H, W] -> [batch, 1, H, W]
+    // Note: With pretrained weights, lin weights are positive.
+    // We use abs() to ensure non-negative output with random weights.
+    let weighted = lin.forward(diff_sq).abs();
```
laggui: If I understand correctly, LPIPS is not yet a valid metric for random weights. So perhaps we should not force `abs()` as per the standard implementation?

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,834 @@
...
+        let device = Default::default();
+
+        let image1 = TestTensor::<4>::zeros([1, 3, 32, 32], &device);
+        let image2 = TestTensor::<4>::ones([1, 3, 32, 32], &device);
+
+        let lpips: Lpips<TestBackend> = LpipsConfig::new().init(&device);
+        let distance = lpips.forward(image1, image2, Reduction::Mean);
+
+        let distance_value = distance.into_data().to_vec::<f32>().unwrap()[0];
+        assert!(distance_value >= 0.0, "LPIPS should be >= 0");
```
laggui: Related to previous comment:

For random init, non-negativity is not guaranteed right? If so, we should test non-negativity only with pretrained models.

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,834 @@
...
+    }
+
+    // =========================================================================
+    // Pretrained Weights Tests (requires network)
+    // =========================================================================
+
+    /// Test VGG pretrained weights download and loading.
+    /// Run with: cargo test -p burn-train --features vision test_lpips_pretrained_vgg -- --ignored
+    #[test]
+    #[ignore = "requires network access to download pretrained weights"]
```
laggui: The issue with  adding ignored tests is that they will almost never be executed 😅 so not great for an automated workflow that should flag regressions. I think we should remove the ignore, CI has network access.

crates/burn-train/src/metric/vision/lpips/weights.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    // Download weights (synchronous, tokio runtime handled internally)
+    let bytes = download_file_as_bytes(url, message);
+
+    // Write to temporary file (PytorchStore requires file path)
+    let temp_dir = std::env::temp_dir();
+    let temp_path = temp_dir.join(format!("lpips_{:?}.pth", net));
+    let mut file = std::fs::File::create(&temp_path).expect("Failed to create temp file");
+    file.write_all(&bytes).expect("Failed to write weights");
+    drop(file);
```
laggui: I think we should download the pre-trained weights to a cache dir, similar to MNIST so it doesn't have to be re-downloaded if the temp dir is wiped.

https://github.com/tracel-ai/burn/blob/b7b388bac18ca1ccb57be362fe95a774e7a0acea/crates/burn-dataset/src/vision/mnist.rs#L120-L129

And the download fn should check if it already exists to avoid downloading every time.

crates/burn-train/src/metric/vision/lpips/metric.rs
koreaygj: You're right. I added abs() to pass the non-negativity test with random weights. Will remove it.
thanks for your feedback!

koreaygj:
Applied all feedback. Thanks! 👍

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,810 @@
...
+    /// Whether to normalize input
+    pub(crate) normalize: bool,
+}
+
+impl<B: Backend> LpipsVgg<B> {
+    /// Compute LPIPS distance without reduction using VGG backbone.
+    pub fn forward_no_reduction(&self, input: Tensor<B, 4>, target: Tensor<B, 4>) -> Tensor<B, 1> {
+        let [_batch, _, _, _] = input.dims();
+
+        // Preprocess inputs
```
laggui:
```suggestion
        // Preprocess inputs
```

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,810 @@
...
+            .squeeze_dim::<1>(1)
+    }
+}
+
+impl<B: Backend> LpipsAlex<B> {
+    /// Compute LPIPS distance without reduction using AlexNet backbone.
+    pub fn forward_no_reduction(&self, input: Tensor<B, 4>, target: Tensor<B, 4>) -> Tensor<B, 1> {
+        let [_batch, _, _, _] = input.dims();
+
+        // Preprocess inputs
```
laggui:
```suggestion
        // Preprocess inputs
```

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -0,0 +1,810 @@
...
+            .squeeze_dim::<1>(1)
+    }
+}
+
+impl<B: Backend> LpipsSqueeze<B> {
+    /// Compute LPIPS distance without reduction using SqueezeNet backbone.
+    pub fn forward_no_reduction(&self, input: Tensor<B, 4>, target: Tensor<B, 4>) -> Tensor<B, 1> {
+        let [_batch, _, _, _] = input.dims();
+
+        // Preprocess inputs
```
laggui:
```suggestion
        // Preprocess inputs
```

PR #4404, opened 2026-01-29
Add device dtype usage

laggui:
### Changes

Expanded `B::supports_dtype(&device, dtype)` to have better granularity for dtype usage. A dtype is defined as generally supported when storage and arithmetic are supported.

An accelerated-only dtype should not be used to set the default float/int data types for a device (#4373), but can be used in specific operations.

https://github.com/tracel-ai/burn/blob/3bdee8a84c360f534ffeb85a5bebb66245e424a1/crates/burn-backend/src/backend/base.rs#L351-L382

Opening this for discussion, wanna make sure we get the semantics / definition right.

### Testing

Tested for my device (RTX 4050 laptop):

**Cuda**
```
F32: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=true, mma=true, scaled_mma=false, dot_only=false

F16: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=true, stmatrix=false, cmma=true, mma=true

BF16: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=true, stmatrix=false, cmma=true, mma=true

I32: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=true, mma=true
```

**Vulkan**
```
F32: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=true, mma=false

F16: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=true, mma=false

BF16: EnumSet(Storage | Accelerated)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=true, mma=false

I32: EnumSet(Storage | Arithmetic | Accelerated)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=true, mma=false
```

**WebGpu**
```
F32: EnumSet(Storage | Arithmetic)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=false, mma=false

F16: EnumSet(Storage | Arithmetic)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=false, mma=false

BF16: EnumSet()
CubeCL features: ldmatrix=false, stmatrix=false, cmma=false, mma=false

I32: EnumSet(Storage | Arithmetic)
CubeCL features: ldmatrix=false, stmatrix=false, cmma=false, mma=false
```

I listed the CubeCL features related to `Accelerated` for explicitness.

### DIFF
--- a/crates/burn-backend/src/backend/base.rs
+++ b/crates/burn-backend/src/backend/base.rs
@@ -361,24 +361,23 @@ pub enum DTypeUsage {
     /// The type supports general-purpose arithmetic and common tensor
     /// operations (e.g. elementwise ops, reductions, etc.).
     Arithmetic,
-    /// The type is supported by specialized hardware instructions.
+    /// The type is supported by hardware-accelerated execution paths.
     ///
-    /// This primarily indicates support for high-performance matrix multiplication via
-    /// specialized instructions (e.g., MMA/CMMA). It may also represent restricted
-    /// specialized operations (such as dot products) on types that otherwise lack
-    /// general arithmetic support.
+    /// This typically indicates support for accelerator-backed compute units (e.g., tensor
+    /// cores executing MMA instructions) for high-performance operations such as matrix
+    /// multiplication and operations that lower to it.
     ///
     /// # Notes
     /// - A type can be both [`Arithmetic`](DTypeUsage::Arithmetic) and
-    ///   [`Specialized`](DTypeUsage::Specialized) if it supports general-purpose operations
+    ///   [`Accelerated`](DTypeUsage::Accelerated) if it supports general-purpose operations
     ///   *and* accelerated paths.
-    /// - If a type is only marked as `Specialized` (without `Arithmetic`), it means it is
-    ///   not generally usable for standard tensor operations and can only participate in the
-    ///   specialized or restricted instructions.
+    /// - If a type is marked as `Accelerated` but not `Arithmetic`, it is not
+    ///   suitable for general-purpose tensor operations and may only be used
+    ///   in specific accelerated operations.
     ///
-    /// `Specialized` is a **flag**, not a detailed descriptor. It does not enumerate which
-    /// exact instructions are available.
-    Specialized,
+    /// `Accelerated` is a **flag**, not a detailed descriptor. It does not enumerate which
+    /// operations are accelerated or which accelerator features are available.
+    Accelerated,
 }
 
 /// A set of [DTypeUsage] representing the total capabilities of a data type on a device.
--- a/crates/burn-cubecl/src/backend.rs
+++ b/crates/burn-cubecl/src/backend.rs
@@ -102,31 +102,22 @@ where
         let props = client.properties();
         let storage = dtype.into();
         let usage = props.type_usage(storage);
-        let is_arithmetic = usage.contains(TypeUsage::Arithmetic);
 
         let mut out = DTypeUsageSet::new();
 
-        // General purpose
         if usage.is_superset(TypeUsage::Buffer | TypeUsage::Conversion) {
             out |= DTypeUsage::Storage;
         }
 
-        if is_arithmetic {
+        if usage.contains(TypeUsage::Arithmetic) {
             out |= DTypeUsage::Arithmetic;
         }
 
-        // A type is specialized if:
-        // - It has specialized MMA hardware support
-        // - OR it only supports dot products but lacks general arithmetic (restricted specialization)
         let has_mma = |cfg: &MmaConfig| {
             cfg.a_type == storage || cfg.b_type == storage || cfg.cd_type == storage
         };
-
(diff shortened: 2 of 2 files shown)


laggui:
@wingertge you're probably the best person to ask for feedback here since you've introduced these features at the CubeCL level :slightly_smiling_face: 

I think `Storage` and `Arithmetic` are quite straightforward, but want to make sure `Specialized` is well specified. It should be a flag, not an enumeration, otherwise the signal will be too noisy for users I believe.

wingertge:
`ldmatrix`/`stmatrix` isn't really relevant, that's just an alternate loading strategy for MMA instructions and is selected or not selected based on support automatically. It's faster, but not so much faster that it impacts type selection.

I would break types into just storage/conversion, arithmetic, and matmul only. Dot product isn't really relevant outside of graphics and is only supported on a couple of Intel cards I believe (the extension is tagged as Intel).

Scaled matmul is a big performance boost but also not necessarily required since there's other ways to handle quantized tensors (and it's not yet implemented in the kernels anyways). Regardless, I think that would be a separate query on the quantization scheme, since it also requires a specific type for scales, and a specific block scale scheme (32-wide row major for lhs and 32-wide col major for rhs).

laggui:
Good points, that makes sense!

> `ldmatrix`/`stmatrix` isn't really relevant, that's just an alternate loading strategy for MMA instructions and is selected or not selected based on support automatically. It's faster, but not so much faster that it impacts type selection.

Right, I wrote the condition as a logical OR but if MMA is available, it will use the best loading strategy anyway. So MMA support is the important part, will remove the `ldmatrix`/`stmatrix`.

> Scaled matmul is a big performance boost but also not necessarily required since there's other ways to handle quantized tensors (and it's not yet implemented in the kernels anyways). Regardless, I think that would be a separate query on the quantization scheme, since it also requires a specific type for scales, and a specific block scale scheme (32-wide row major for lhs and 32-wide col major for rhs).

Yeah it is not required for `Specialized` to be true, but I agree it would make more sense to query on a quantization type (e.g. scheme) instead. QFloat is not a general dtype anyway, quantization is a different form of specialization. Will remove :+1: 

> I would break types into just storage/conversion, arithmetic, and matmul only. Dot product isn't really relevant outside of graphics and is only supported on a couple of Intel cards I believe (the extension is tagged as Intel).

That's the only part I'm unsure about. Storage/conversion and arithmetic are already represented, but I was hesitant to name the flag after a specific operation like matmul. Right now, it is definitely the biggest driving factor. But I didn't want to bake the specialization policy into the enum.

I see `Specialized` as a capability flag, not necessarily a detailed descriptor. It keeps `DTypeUsage` focused on high-level usability, while CubeCL feature queries handle more granular, operation-specific kernel decisions (lower level). Essentially striking a balance for Burn users.

So even if the dot product is niche, it would currently fall under the restricted specialized types that can perform certain math primitives but cannot be used for general purpose tensor operations.

We could potentially expose more granular queries in Burn in the future (e.g. checking for matmul acceleration), but I'd see those as device properties parameterized by dtype, rather than something that defines dtype usability itself.

That was a bit longer to explain than I thought :sweat_smile: anyway, lmk if that makes sense or if you see it differently!

wingertge:
> That's the only part I'm unsure about. Storage/conversion and arithmetic are already represented, but I was hesitant to name the flag after a specific operation like matmul. Right now, it is definitely the biggest driving factor. But I didn't want to bake the specialization policy into the enum.

I think that boat sailed the moment dedicated accelerators were implemented for matrix multiplication. Maybe matmul isn't the correct name though because it does also apply to convolution and attention. But we do have to deal with the fact that matrix multiplication (and anything that maps to it) is executed on completely separate accelerators, with entirely separate and significantly broader type support. Most of the "matmul only" types were created specifically for tensor cores, including things like fp8 and fp6 that aren't yet implemented in burn. So I don't think dot product should be grouped with those types here.

Not sure what the option should be named, there needs to be some kind of documentation on the operations that they apply to. It really depends on whether the inputs are fed into tensor cores or not.

laggui:
> I think that boat sailed the moment dedicated accelerators were implemented for matrix multiplication. Maybe matmul isn't the correct name though because it does also apply to convolution and attention.

Hahah yeah I actually had those (conv and attention) in mind too, even if they can map to a matmul.

Maybe `Accelerated` is a better name? Essentially specifies "this dtype can participate in accelerator-backed compute paths that are distinct from the general-purpose arithmetic pipeline".

> So I don't think dot product should be grouped with those types here.

Ok I agree with the distinction for accelerators. Dot product would be a niche specialized instruction (limited set), while matmul really is an accelerated path (hardware backed). So I'll remove that.

PR #4410, opened 2026-01-30
Add configurable activation and layer_norm_eps to transformer layers

antimora:
## Summary

- Add `gelu_approximate` free function to `burn-tensor::activation` (tanh-based GELU matching PyTorch `gelu(approximate="tanh")`) with precomputed `sqrt(2/pi)` constant
- Add `approximate` field to `Gelu` module (`Gelu::new()` for exact, `Gelu::new_approximate()` for tanh approximation)
- Replace hardcoded `Gelu` in `PositionWiseFeedForward` with configurable `activation: ActivationConfig` (default: `Gelu`)
- Add `activation` and `layer_norm_eps` config fields to both `TransformerEncoderConfig` and `TransformerDecoderConfig`
- Fix decoder not passing `initializer` to `PositionWiseFeedForwardConfig`

## Motivation

Models like ALBERT, GPT-2, and others use different FFN activations (approximate GELU, ReLU, SwiGLU). Previously `PositionWiseFeedForward` hardcoded `Gelu`, forcing downstream models to reimplement the entire feed-forward layer. Now any `ActivationConfig` variant can be used.

Similarly, `layer_norm_eps` was hardcoded to the `LayerNormConfig` default (1e-5), but models like ALBERT specify different values in their configs.

## Test plan

- [x] `Gelu::new_approximate()` forward test with values verified against PyTorch `torch.nn.functional.gelu(x, approximate="tanh")`
- [x] `ActivationConfig::GeluApproximate` wrapper test
- [x] All existing burn-nn tests pass (252 tests)
- [x] Display test updated for new Gelu struct format
- [x] albert-burn pretrained integration tests pass with correct logits

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -72,8 +72,7 @@ pub fn gelu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
     Tensor::from_primitive(TensorPrimitive::Float(B::gelu(tensor.primitive.tensor())))
 }
 
-/// Applies the tanh-based approximate GELU function element-wise, matching
-/// PyTorch's `gelu(approximate="tanh")` and HuggingFace's `gelu_new`.
+/// Applies the tanh-based approximate GELU function element-wise.
 ///
 #[cfg_attr(
     doc,


antimora:
CCing @crutcher

crates/burn-nn/src/modules/transformer/encoder.rs
```diff
@@ -225,19 +232,24 @@ impl<B: Backend> TransformerEncoderLayer<B> {
...
             .with_initializer(config.initializer.clone())
             .with_dropout(config.dropout)
+            .with_activation(config.activation.clone())
             .init(device);
 
         Self {
             mha,
+            pwff,
             norm_1,
             norm_2,
```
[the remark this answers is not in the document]
antimora: The feedback is not valid. The current order already matches the struct definition. Copilot misread it.

antimora:
The ALBERT model implementation in [tracel-ai/models#92](https://github.com/tracel-ai/models/pull/92) depends on this PR (configurable activation, layer_norm_eps, and the norm ordering fix). It currently points to a feature branch with these changes. Once this is merged, we can switch albert-burn to use the upstream burn crate.

crates/burn-nn/src/activation/gelu_approximate.rs
```diff
@@ -0,0 +1,69 @@
...
+use burn::module::Module;
+use burn::tensor::Tensor;
+use burn::tensor::backend::Backend;
+
+/// Applies the approximate Gaussian Error Linear Units function element-wise
+/// using the tanh approximation (matches HuggingFace `gelu_new`).
+///
+/// `GeluApproximate(x) = 0.5 * x * (1 + tanh(sqrt(2/π) * (x + 0.044715 * x³)))`
+#[derive(Module, Clone, Debug, Default)]
+pub struct GeluApproximate;
```
laggui: Don't think this should be a separate `nn` module. It is still `Gelu`, just parameterized differently.

I would add an optional approximation field, which can hold the tanh variant. We keep `Gelu::new()` without approximation by default, and can add another constructor for the tanh approx.

`burn::tensor::activation::gelu_approximate` can remain as is.

laggui (changes requested):
> * **Fix norm layer ordering bug**: encoder and decoder applied norm layers in wrong order relative to their field names (`norm_2` was used for attention, `norm_1` for FFN), causing silent weight mismatches when loading pre-trained models. Now `norm_1` → first sub-layer, `norm_2` → second, etc.

Actually, it's not a bug. The ordering might look weird, but it is applied as `norm_1` -> `norm_2` -> `norm_3` for post layernorm (`norm_first = false`). Only when `norm_first = true` does that become `norm_3` -> `norm_1` -> `norm_2`.

The numbering looks odd in that case, but functionally it’s consistent. I would not introduce a breaking change for that.

crates/burn-nn/src/activation/gelu_approximate.rs
antimora: Thanks for the feedback! That makes sense. I'll fold it into the existing Gelu module with an optional approximation field.

Regarding the norm ordering: you're right, I've reverted those changes. The numbering follows the post-norm path order and is consistent.

laggui: If the field names are an issue to load a model, you can remap the keys :wink:

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -72,6 +72,33 @@ pub fn gelu<const D: usize, B: Backend>(tensor: Tensor<B, D>) -> Tensor<B, D> {
     Tensor::from_primitive(TensorPrimitive::Float(B::gelu(tensor.primitive.tensor())))
 }
 
+/// Applies the tanh-based approximate GELU function element-wise, matching
+/// PyTorch's `gelu(approximate="tanh")` and HuggingFace's `gelu_new`.
```
laggui:
```suggestion
/// Applies the tanh-based approximate GELU function element-wise.
```

PR #4416, opened 2026-01-31
chore: deprecate burn-candle backend

antimora:
## Summary

- Add deprecation notices to `burn-candle` crate (`#![deprecated]`, updated Cargo.toml description, README)
- Remove all internal usages of `burn-candle` from `burn`, `burn-backend-tests`, `burn-vision`, `burn-store`, `xtask`, and CI
- Remove Candle from README backend tables and docs
- Keep `burn-candle` in the publish workflow so existing users see deprecation warnings on upgrade; full removal planned for a future release

Users should migrate to `burn-cubecl` (CUDA/ROCm/Vulkan/Metal/WebGPU), `burn-ndarray`, or `burn-tch`.

## Test plan

- [ ] `cargo check` passes on workspace
- [ ] `cargo check -p burn-candle` shows deprecation warning
- [ ] Verify no remaining internal feature flag references to candle (outside `burn-candle` crate itself)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -708,7 +708,6 @@ name = "burn-backend-tests"
 version = "0.21.0"
 dependencies = [
  "burn-autodiff",
- "burn-candle",
  "burn-cpu",
  "burn-cubecl",
  "burn-cuda",
@@ -1112,7 +1111,6 @@ dependencies = [
 name = "burn-store"
 version = "0.21.0"
 dependencies = [
- "burn-candle",
  "burn-core",
  "burn-cuda",
  "burn-ndarray",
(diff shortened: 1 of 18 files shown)


theduke:
Out of curiosity, why is the candle backend being removed?

laggui:
It's being deprecated because it lags behind other backends in operator support and performance.

It was a great backend early on, but as Burn evolved it became increasingly redundant with other backends, and its limitations now outweigh the benefits of continued maintenance.

Focusing on fewer, more capable backends lets us move faster and provide a better experience overall.

Users can still use candle as a backend (provided the implementation is forked & kept up to date even with known limitations), but it will be removed from burn-maintained backends.

PR #4436, opened 2026-02-02
Use CubeK radix sort kernels

ArthurBrussee:
Use the kernels from https://github.com/tracel-ai/cubek/pull/77 to actually sort

This only works for 1D sorting. Multi-dimensional sorting will still fallback to CPU sorting for now. This is even the case in PyTorch I think, but, in the future we could probably add a GPU version here as well, at least for 2D cases.

I've kept the API the same for now. In the future it'd be great to have the option to sort arbitrary (K, V) pairs as well besides just indices.

Should also look at autotuning sometime, varying the number of items per thread seems


antimora:
No activity; closing; ask to reopen if needed.

PR #4446, opened 2026-02-03
chore: update workflows

syl20bnr:
This PR update the workflows to the latest Tracel GH actions v8.

It also disabled auto-update for Tracel GH actions by dependabot because update should be automatic already in actions when the major version does no change, i.e. `@v8` means latest version of `v8` already. If the major version number changes then it means it is likely to break and dependabot has not real way to verify this (for instance when it upgrades actions in `publish.yml` it cannot run it to verify... setting us up for possible failure later when we publish a new version).

It also deletes the workflows that are not useful.

### DIFF
--- a/.github/workflows/vulnerabilities.yml
+++ b/.github/workflows/vulnerabilities.yml
@@ -8,7 +8,7 @@ on:
       - 'v*.*.*'
 
 env:
-  CAREFUL_VERSION: "0.4.8"
+  CAREFUL_VERSION: "0.4.9"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -34,7 +34,7 @@ invalid_html_tags = "deny"
 atomic_float = "1"
 axum = "0.8.8"
 bytemuck = "1.25.0"
-bytes = { version = "1.10", default-features = false }
+bytes = { version = "1.11.1", default-features = false }
```
laggui, line 37: Should probably commit the updated `Cargo.lock` with this change?

.github/workflows/vulnerabilities.yml
```diff
@@ -5,14 +5,9 @@ on:
...
-      - 'v*.*.*' # Run when a new version is being published
+      - 'v*.*.*'
 
 env:
-  #
-  # Dependency versioning
-  #
-
-  # careful version
   CAREFUL_VERSION: "0.4.8"
```
Luni-4: The latest version is `0.4.9`

PR #4447, opened 2026-02-03
Burn rl

Charles23R:
### Checklist

### Related Issues/PRs

-

### Changes

- Add `burn-rl` crate with basic RL utils and traits.
- Add the `RLTraining` to burn-train to launch a RL training. Implemented a basic off policy strategy.
- Create a DQN agent example.

### Testing

Tested mostly with DQN example. Need to write more tests.

### DIFF
--- a/xtask/src/commands/test.rs
+++ b/xtask/src/commands/test.rs
@@ -118,6 +118,9 @@ pub(crate) fn handle_command(
                         "burn-router".to_string(),
                         "burn-tch".to_string(),
                         "burn-wgpu".to_string(),
+                        // dqn-agent example relies on gym-rs dependency which requires SDL2.
+                        // It would be good not to remove the gym-rs dependency in the future.
+                        "dqn-agent".to_string(),
                     ]);
 
                     // Burn remote tests don't work on windows for now


crates/burn-rl/src/agent/base.rs
```diff
@@ -0,0 +1,419 @@
...
+        receiver
+            .recv()
+            .expect("AsyncPolicy should be able to receive policy state.")
+    }
+
+    fn load_record(self, _record: <Self::PolicyState as PolicyState<B>>::Record) -> Self {
+        // Not needed for now
+        todo!()
+    }
+}
```
Charles23R: Create tests for Autobatching logic

crates/burn-core/src/record/base.rs
```diff
@@ -7,7 +7,7 @@ use serde::{Serialize, de::DeserializeOwned};
 /// Trait to define a family of types which can be recorded using any [settings](PrecisionSettings).
 pub trait Record<B: Backend>: Send {
     /// Type of the item that can be serialized and deserialized.
-    type Item<S: PrecisionSettings>: Serialize + DeserializeOwned;
+    type Item<S: PrecisionSettings>: Serialize + DeserializeOwned + Clone;
```
laggui, line 10: I don't think the `Clone` is required, no?

crates/burn-core/src/record/recorder.rs
```diff
@@ -285,7 +285,7 @@ mod tests {
     #[test]
     #[should_panic]
     fn err_when_invalid_item() {
-        #[derive(new, Serialize, Deserialize)]
+        #[derive(new, Serialize, Deserialize, Clone)]
```
laggui, line 288: And subsequently, this.

crates/burn-optim/src/optim/simple/record/base.rs
```diff
@@ -19,7 +19,7 @@ where
 }
 
 /// [Optimizer adaptor](crate::optim::simple::adaptor::OptimizerAdaptor) record item.
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
```
laggui, line 22: Same

crates/burn-optim/src/optim/simple/record/v1.rs
```diff
@@ -56,7 +56,7 @@ impl<O: SimpleOptimizer<B>, B: Backend> Clone for AdaptorRecordV1<O, B> {
 }
 
 /// [Optimizer adaptor](crate::optim::simple::adaptor::OptimizerAdaptor) record item.
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
```
laggui, line 59: Same

examples/dqn-agent/src/training.rs
```diff
@@ -0,0 +1,120 @@
...
+    //     num_envs: 1,
+    //     autobatch_size: 1,
+    //     replay_buffer_size: 10_000,
+    //     train_interval: 1,
+    //     eval_interval: 3_000,
+    //     eval_episodes: 5,
+    //     train_batch_size: 128,
+    //     train_steps: 1,
+    //     warmup_steps: 0,
+    // };
```
laggui: Dead code

examples/dqn-agent/src/training.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    let policy_model = MlpNet::<B>::new(&model_config, &device);
+    let optimizer = AdamWConfig::new()
+        .with_grad_clipping(Some(GradientClippingConfig::Value(100.0)))
+        .init();
+    let agent = DqnLearningAgent::new(policy_model, optimizer, dqn_config);
+    let learner = RLTraining::new(ARTIFACT_DIR, CartPoleWrapper::new)
+        // .metrics((Loss, ...), RlLogFrequency::{TrainStep, EnvStep, EpisodeEn})
```
laggui: Dead code

examples/dqn-agent/src/training.rs
```diff
@@ -0,0 +1,120 @@
...
+        .metrics_train_step((LossMetric::new(),))
+        .metrics_env_step((ExplorationRateMetric::new(),))
+        .metrics_episode((EpisodeLengthMetric::new(), CumulativeRewardMetric::new()))
+        .with_file_checkpointer(CompactRecorder::new())
+        .num_steps(40_000)
+        // .checkpoint(10_000)
```
laggui: Same

crates/burn-core/src/record/base.rs
laggui: Nevermind, this is required given the current `Module` derive limitations, for the `DqnLearningRecord`.

We need to improve the module derive instead, but that should not block the current state.

crates/burn-train/src/renderer/base.rs
```diff
@@ -22,19 +22,27 @@ pub trait MetricsRendererTraining: Send + Sync {
...
     fn update_valid(&mut self, state: MetricState);
 
-    /// Renders the training progress.
+    /// Updates the training status.
     ///
     /// # Arguments
     ///
     /// * `item` - The training progress.
-    fn render_train(&mut self, item: TrainingProgress);
+    fn update_status_train(
```
laggui: I'd keep the `render_train` / `render_valid` naming given the trait is `MetricsRendererTraining`.

Only the new TUI wrapper actually sends a status message for the renderer.

crates/burn-train/src/renderer/base.rs
```diff
@@ -128,16 +140,13 @@ pub enum MetricState {
...
-    pub progress: Progress,
-
-    /// The epoch.
-    pub epoch: usize,
+    pub progress: Option<Progress>,
 
-    /// The total number of epochs.
-    pub epoch_total: usize,
+    /// The progress of the whole training.
+    pub global_progress: Progress,
```
laggui, line 134: Makes sense to be more flexible! So even supervised training could opt out of epochs over a dataset, and instead work directly with iterations.

But we should probably revisit the `progress` / `global_progress` before the release, not immediately clear what they should represent. Not required for this PR.

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -38,80 +39,173 @@ type PanicHook = Box<dyn Fn(&std::panic::PanicHookInfo<'_>) + 'static + Sync + S
...
+                    Err(mpsc::TryRecvError::Empty) => (),
+                    Err(mpsc::TryRecvError::Disconnected) => {
+                        log::error!("Renderer thread disconnected.");
+                        break;
+                    }
+                }
+
+                // Render
+                if renderer.last_update.elapsed() >= tick_rate {
+                    renderer.render().unwrap()
```
laggui: Now that the renderer is in another thread, the main thread might not cleanly handle the case when `render` fails.

We could probably `log::error` on `Err`, and break from the loop e.g.

```rust
if let Err(e) = renderer.render() {
    log::error!("Render error: {e}");
    break;  // exit
}
```

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -38,80 +39,173 @@ type PanicHook = Box<dyn Fn(&std::panic::PanicHookInfo<'_>) + 'static + Sync + S
...
+impl MetricsRendererEvaluation for TuiMetricsRendererWrapper {
     fn update_test(&mut self, name: EvaluationName, state: MetricState) {
-        self.update_metric(TuiSplit::Test, TuiGroup::Named(name.name), state);
+        self.sender
+            .send(TuiRendererEvent::MetricsUpdate((
+                TuiSplit::Test,
+                TuiGroup::Named(name.name),
+                state,
+            )))
+            .expect("Can send tui renderer event.");
```
laggui: Instead of panicking on send failures for render events, we could  warn and continue training
```rust
if let Err(e) = self.sender.send(event) {
    log::warn!("Failed to send TUI event: {e}");
}
```

So it's a bit more robust to TUI failures, and a training won't just stop. Users can still interrupt the training if they want to stop after seeing TUI failures.

/edit: whoops this is the evaluation, but my comment remains 😅

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -38,80 +39,173 @@ type PanicHook = Box<dyn Fn(&std::panic::PanicHookInfo<'_>) + 'static + Sync + S
...
+        &mut self,
+        item: EvaluationProgress,
+        progress_indicators: Vec<ProgressType>,
+    ) {
+        self.sender
+            .send(TuiRendererEvent::StatusUpdateTest((
+                item,
+                progress_indicators,
+            )))
+            .expect("Can send tui renderer event.");
```
laggui: Same

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -38,80 +39,173 @@ type PanicHook = Box<dyn Fn(&std::panic::PanicHookInfo<'_>) + 'static + Sync + S
...
+impl MetricsRendererTraining for TuiMetricsRendererWrapper {
     fn update_train(&mut self, state: MetricState) {
-        self.update_metric(TuiSplit::Train, TuiGroup::Default, state);
+        self.sender
+            .send(TuiRendererEvent::MetricsUpdate((
+                TuiSplit::Train,
+                TuiGroup::Default,
+                state,
+            )))
+            .expect("Can send tui renderer event.");
```
laggui: Same

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -38,80 +39,173 @@ type PanicHook = Box<dyn Fn(&std::panic::PanicHookInfo<'_>) + 'static + Sync + S
...
 
     fn update_valid(&mut self, state: MetricState) {
-        self.update_metric(TuiSplit::Valid, TuiGroup::Default, state);
+        self.sender
+            .send(TuiRendererEvent::MetricsUpdate((
+                TuiSplit::Valid,
+                TuiGroup::Default,
+                state,
+            )))
+            .expect("Can send tui renderer event.");
```
laggui: Same.

And now I realized the pattern can probably be encapsulated in a function that can be reused

```rust
fn send_event(&self, event: TuiRendererEvent) {
    if let Err(e) = self.sender.send(event) {
        log::warn!("Failed to send TUI event: {e}");
    }
}
```

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -38,80 +39,173 @@ type PanicHook = Box<dyn Fn(&std::panic::PanicHookInfo<'_>) + 'static + Sync + S
...
+impl MetricsRenderer for TuiMetricsRendererWrapper {
     fn manual_close(&mut self) {
-        loop {
-            self.render().unwrap();
-            if self.interuptor.should_stop() {
-                return;
-            }
-            std::thread::sleep(Duration::from_millis(100));
-        }
+        let _ = self.handle_join.take().unwrap().join();
```
laggui, line 160: I think we should also `.join()` on `Drop`. Otherwise the thread might become detached

PR #4460, opened 2026-02-05
Implement basic RNN module

aditya0by0:
## Summary

Adds a vanilla RNN module to burn-nn, which is needed to implement RNN ONNX operator(https://github.com/tracel-ai/burn-onnx/issues/11).

## Changes

- Add `Rnn` and `BiRnn` module.

## Testing

All 7 tests pass:
- Forward pass tests (single input, batched)
- Backward pass test
- Bidirectional test
- Initializer test
- Display tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,7 +935,7 @@ dependencies = [
  "tar",
  "tempfile",
  "thiserror 2.0.18",
- "zip 7.2.0",
+ "zip 7.4.0",
 ]
 
 [[package]]
@@ -1155,7 +1155,7 @@ dependencies = [
  "serde",
  "tempfile",
  "textdistance",
- "zip 7.2.0",
+ "zip 7.4.0",
 ]
 
 [[package]]
@@ -1373,7 +1373,7 @@ dependencies = [
  "candle-metal-kernels",
  "candle-ug",
  "cudarc 0.19.0",
- "float8 0.6.0",
+ "float8 0.6.1",
  "gemm 0.19.0",
  "half",
  "libc",
@@ -1389,7 +1389,7 @@ dependencies = [
  "safetensors 0.7.0",
  "thiserror 2.0.18",
  "yoke 0.8.1",
- "zip 7.2.0",
+ "zip 7.4.0",
 ]
 
 [[package]]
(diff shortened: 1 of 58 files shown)


antimora:
is `Vanilla` technical term? Why not re-use the existing RNN module? we need to make sure we don't have duplicate codes.

aditya0by0:
Hi @antimora, I'm relatively new to Rust and the Burn library.

Regarding your comment: 
> Why not re-use the existing RNN module? we need to make sure we don't have duplicate codes.

I looked through crates/burn-nn/src/modules/rnn, but I only see LSTM, GRU, and GateController. Could you clarify which existing module or struct you are referring to? I want to make sure I don't miss an existing abstraction.

antimora:
> Hi @antimora, I'm relatively new to Rust and the Burn library.

Good! Welcome! 

> 
> Regarding your comment:
> 
> > Why not re-use the existing RNN module? we need to make sure we don't have duplicate codes.
> 
> I looked through crates/burn-nn/src/modules/rnn, but I only see LSTM, GRU, and GateController. Could you clarify which existing module or struct you are referring to? I want to make sure I don't miss an existing abstraction.

Yeah it appears I was mistaken. Sorry. I thought we had an RNN module but we have a rust rrn module. Prior research was outdated.

but "vanilla" is still problematic. I think we should rename to something like:

1. `rnn/simple.rs` 
2. `rnn/basic.rs` (I prefer this)

@laggui, we need your vote here too.

aditya0by0:
> Must fix before merge:
> 
> 1. Rename vanilla.rs -> basic.rs (consistency) or something else
> 2. Update mod.rs comment from "Vanilla Recurrent Neural Network" -> "Recurrent Neural Network" or "Basic RNN"
> 
> Should fix:  4. Add test for reverse=true configuration 5. Add test for clip functionality

I have implemented the above changes. 

For the below change, I understand that `forward_iter` is private method for RNN and hence it can't be called from BiRNN. But still my code works and all the test cases pass. I wonder why?
> . Add pub(crate) to forward_iter 

In Lstm too, I can see the same thing: 
https://github.com/tracel-ai/burn/blob/34e22739947afb989d5eb47f649181879c1c95bc/crates/burn-nn/src/modules/rnn/lstm.rs#L216

PR #4466, opened 2026-02-05
fix: use all dilation entries in `max_pool2d_with_indices_backward`

fcasal:
### Changes

This PR fixes a likely typo in the crates/burn-cubecl/src/kernel/pool/max_pool2d_backward.rs file to use all dilation entries to build `PoolBackwardArgsLaunch` instead of using `dilation[0]` twice. I don't know anything about this code but the current version seems wrong. I noticed all tests use the same value for both dilation[0] and dilation[1], which is why this was probably never found


laggui (approved):
Good catch 😅 

That must have slipped through since it's almost always `[1, 1]` (default) or symmetric..

PR #4470, opened 2026-02-06
Feature flag + Tests for RL in burn-rl and burn-train

Charles23R:
### Checklist

- [] Made sure the book is up to date with changes in this PR.

### Changes

Added a `rl` feature flag to burn/burn-train for RL (default flag).
Added some unit tests missing in last PR.
Naming and API changes for transition buffer and agent/environment loops.

### Testing

Tested feature flag on mnist and dqn-agent examples.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,7 +21,7 @@ jobs:
       - publish-burn-ndarray
     uses: tracel-ai/github-actions/.github/workflows/publish-crate.yml@v9
     with:
-      crate: burn-vision
+      crate: burn-rl
       dry-run-only: ${{ github.event_name == 'workflow_dispatch' && inputs.dry-run-only || false }}
     secrets:
       CRATES_IO_API_TOKEN: ${{ secrets.CRATES_IO_API_TOKEN }}


crates/burn-rl/src/lib.rs
```diff
@@ -18,4 +18,142 @@ pub use transition_buffer::*;
...
 
 #[cfg(test)]
-pub(crate) mod tests {}
+pub(crate) mod tests {
+    use crate::{Batchable, Policy, PolicyState, TestBackend};
+
+    use burn_core::record::Record;
+    use burn_core::{self as burn};
+
+    /// Mock policy for testing
```
nathanielsimard, line 27: Maybe add a bit of docs on how to use that policy in tests

Charles23R:
> Not directly related to the changes in this PR, but we need to add the new `burn-rl` crate to the publish workflow before any pre-release.
> 
> https://github.com/tracel-ai/burn/blob/main/.github/workflows/publish.yml
> 
> Otherwise, changes LGTM

Could you check if I did it correctly? :)

.github/workflows/publish.yml
```diff
@@ -362,6 +375,7 @@ jobs:
       - publish-burn-core
       - publish-burn-optim
       - publish-burn-collective
+      - publish-burn-rl
```
laggui, line 378: Technically we should also add `publish-burn-rl` to the main burn crate publish job as it is a direct dependency with the new `rl` feature flag, but it already requires `publish-burn-train` which itself requires `publish-burn-rl`. So the dependency graph is fine for publish.

PR #4471, opened 2026-02-07
add LBFGS optimizer

donjuanplatinum:
### Checklist

### Related Issues/PRs

### Changes
1. add LBFGS optimizer
2. change  a comment error : returnerd to returned

### Testing
image

this is the test torch code:

#### test_cubic_interpolate
```python
import numpy as np

import torch
import torch.nn as nn
torch.set_printoptions(precision=10)
np.set_printoptions(precision=10, suppress=True)
def _cubic_interpolate_f64(x1, f1, g1, x2, f2, g2, bounds=None):
    x1, f1, g1, x2, f2, g2 = map(np.float64, [x1, f1, g1, x2, f2, g2])
    
    if bounds is not None:
        xmin_bound, xmax_bound = map(np.float64, bounds)
    else:
        xmin_bound, xmax_bound = (x1, x2) if x1 <= x2 else (x2, x1)

    # d1 = g1 + g2 - 3*(f1-f2)/(x1-x2);
    d1 = g1 + g2 - 3.0 * (f1 - f2) / (x1 - x2)
    d2_square = d1**2 - g1 * g2
    
    if d2_square >= 0:
        d2 = np.sqrt(d2_square)
        if x1 <= x2:
            min_pos = x2 - (x2 - x1) * ((g2 + d2 - d1) / (g2 - g1 + 2.0 * d2))
        else:
            min_pos = x1 - (x1 - x2) * ((g1 + d2 - d1) / (g1 - g2 + 2.0 * d2))
        return np.clip(min_pos, xmin_bound, xmax_bound)
    else:
        return (xmin_bound + xmax_bound) / 2.0

# 执行测试用例
print(f"{'Case':<12} | {'Result (f64)':<20}")
print("-" * 40)

# 1. Basic
res1 = _cubic_interpolate_f64(-1.0, 1.0, -2.0, 1.0, 1.0, 2.0)
print(f"{'basic':<12} | {res1:.16f}")

# 2. Bound
res2 = _cubic_interpolate_f64(0.0, 0.25, -1.0, 1.0, 0.25, 1.0, bounds=(0.6, 1.0))
print(f"{'bound':<12} | {res2:.16f}")

# 3. Fallback
res3 = _cubic_interpolate_f64(0.0, 0.0, 10.0, 1.0, 5.0, 10.0, bounds=(0.0, 1.0))
print(f"{'fallback':<12} | {res3:.16f}")

# 4. Asymmetric
res4 = _cubic_interpolate_f64(0.0, 1.0, -5.0, 1.0, 0.5, 1.0)
print(f"{'asymmetric':<12} | {res4:.16f}")

res5_1 = _cubic_interpolate_f64(1.231232145, -0.12567458754, 9.1231243007, 8.239105015, -100.9012398021, 123201321.0293982)
res5_2 = _cubic_interpolate_f64(1.231232145, -0.12567458754, 9.1231243007, 8.239105015, -100.9012398021, 123201321.0293982,bounds = (-4.4,4.4))
print(f"{'random_1 ':<12} | {res5_1:.16f}")
print(f"{'random_1_2 ':<12} | {res5_2:.16f}")
```
#### test_strong_wolfe_direct_comparison
```python
from torch.optim.lbfgs import _strong_wolfe
def complex_objective(x, t, d):
    curr_x = (x + t * d)
    
    x2 = curr_x * curr_x
    x4 = x2 * x2
    f = (x4 - 2.0 * x2 + curr_x).sum()
    
    g = 4.0 * (x2 * curr_x) - 4.0 * curr_x + 1.0
    
    return f, g
device = 'cpu'
x = torch.tensor([2.1321912957], device=device)
d = torch.tensor([0.91312321],  device=device)
t_initial = torch.tensor(1.213132, device=device)
f_init, g_init = complex_objective(x, 0, d)
gtd_init = g_init.dot(d)
print(f"Initial State: f={f_init.item():.10f}, gtd={gtd_init.item():.10f}")
f_final, g_final, t_final, evals = _strong_wolfe(
    complex_objective, x, t_initial, d, f_init, g_init, gtd_init,
    c1=1e-4,  
    c2=0.9,  
    max_ls=10
)
print(f"f_final {f_final},g_final {g_final},t_final {t_final},evals {evals}")
print(f"Final Step Size (t): {t_final:}")
print(f"Final Function Value: {f_final.item():.10f}")
print(f"Total Evaluations: {evals}")
print(f"New Position x: {(x + t_final * d).item():.10f}")
```
#### test_lbfgs_strong_wolfe_comparison
```python
def test_lbfgs_pytorch_comparison():
    device = torch.device("cpu")
    x_data = torch.tensor([[1.0], [2.0], [3.0]], device=device)
    y_true = torch.tensor([[3.0], [5.0], [7.0]], device=device)

    model = nn.Linear(1, 1)
    with torch.no_grad():
        model.weight.fill_(0.5)
        model.bias.fill_(0.1)

    optimizer = torch.optim.LBFGS(
        model.parameters(),
        lr=0.001,
        max_iter=20,      
        history_size=100,
        line_search_fn='strong_wolfe'
    )
    criterion = nn.MSELoss(reduction='sum')
    def closure():
        optimizer.zero_grad()
        output = model(x_data)
        loss = criterion(output, y_true)
        loss.backward()
        return loss

    initial_loss = closure().item()
    print(f"Initial Loss: {initial_loss:.10f}")

    optimizer.step(closure)
    with torch.no_grad():
        final_output = model(x_data)
        final_loss = criterion(final_output, y_true).item()  
    print(f"Final Loss after 1 step: {final_loss:.10f}")
    print(f"Final Weight: {model.weight.item():.10f}")
    print(f"Final Bias: {model.bias.item():.10f}")
```

### DIFF
--- a/crates/burn-optim/src/optim/lbfgs.rs
+++ b/crates/burn-optim/src/optim/lbfgs.rs
@@ -53,7 +53,7 @@ fn cubic_interpolate(
         (min_bound + max_bound) / 2.0
     }
 }
-/// Auxiliray Struct For Strong_Wolfe
+/// Auxiliary Struct For Strong_Wolfe
 struct LineSearchSample<B: Backend> {
     // step size
     t: f64,
@@ -464,15 +464,15 @@ pub struct LBFGSState<B: Backend> {
     pub history_s: Vec<Tensor<B, 1>>,
     /// Historical gradient difference vectors
     pub history_y: Vec<Tensor<B, 1>>,
-    /// 搜索方向
+    /// Search direction
     pub d: Option<Tensor<B, 1>>,
-    /// 上一次的步长 t
+    /// Step size from the previous iteration
     pub t: Option<f64>,
-    /// 上一次的扁平梯度
+    /// Flattened gradient from the previous iteration
     pub prev_flat_grad: Option<Tensor<B, 1>>,
-    /// 上一次的 Loss (用于断点检查)
+    /// Loss value from the previous iteration
     pub prev_loss: Option<f64>,
-    /// 全局迭代次数 (用于判断是否需要执行初次初始化)
+    /// Global iteration count
     pub g_iter: usize,
 }
 
@@ -594,7 +594,8 @@ impl<B: Backend + AutodiffBackend> LBFGS<B> {
                 // multiplied by the gradient
                 let num_old = self.state.history_s.len();
                 let mut q = flat_grad.clone().neg();
-                let mut alphas = vec![0.0; num_old];
+                let mut alphas: Vec<Tensor<B::InnerBackend, 1>> =
+                    vec![Tensor::zeros([1], &flat_grad.device()); num_old];
 
                 if num_old > 0 {
                     // multiply by initial Hessian
@@ -602,30 +603,33 @@ impl<B: Backend + AutodiffBackend> LBFGS<B> {
                     for i in (0..num_old).rev() {
                         let s = &self.state.history_s[i];
                         let y = &self.state.history_y[i];
-                        let rho = 1.0 / y.clone().dot(s.clone()).into_scalar().to_f64();
-                        alphas[i] = rho * s.clone().dot(q.clone()).into_scalar().to_f64();
-                        q = q.sub(y.clone().mul_scalar(alphas[i]));
+                        let rho = y.clone().dot(s.clone()).powf_scalar(-1.0);
+                        let alpha = rho.clone().mul(s.clone().dot(q.clone()));
+                        alphas[i] = alpha.clone();
+                        q = q.sub(y.clone().mul(alpha));
                     }
 
                     let last_s = &self.state.history_s[num_old - 1];
                     let last_y = &self.state.history_y[num_old - 1];
-                    let ys = last_y.clone().dot(last_s.clone()).into_scalar().to_f64();
-                    let yy = last_y.clone().dot(last_y.clone()).into_scalar().to_f64();
-                    let h_diag = ys / yy;
+                    let ys = last_y.clone().dot(last_s.clone());
+                    let yy = last_y.clone().dot(last_y.clone());
+                    let h_diag = ys.div(yy);
 
-                    let mut r = q.mul_scalar(h_diag);
+                    let mut r = q.mul(h_diag);
 
-                    for ((s, y), &alpha) in self
+                    for ((s, y), alpha) in self
                         .state
                         .history_s
                         .iter()
                         .zip(self.state.history_y.iter())
-                        .zip(alphas.iter())
+                        .zip(alphas.into_iter())
                         .take(num_old)
                     {
-                        let rho = 1.0 / y.clone().dot(s.clone()).into_scalar().to_f64();
-                        let beta = rho * y.clone().dot(r.clone()).into_scalar().to_f64();
-                        r = r.add(s.clone().mul_scalar(alpha - beta));
+                        let rho = y.clone().dot(s.clone()).powf_scalar(-1.0);
+
+                        let beta = rho.mul(y.clone().dot(r.clone()));
+
+                        r = r.add(s.clone().mul(alpha.sub(beta)));
                     }
                     d = r;
                 } else {


crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,935 @@
...
+        let mut x_flat = flatten_params_inner::<B, M>(&module);
+
+        let opt_cond =
+            flat_grad.clone().abs().max().into_scalar().to_f64() <= self.config.tolerance_grad;
+        // optimal condition
+        if opt_cond {
+            return (module, loss);
+        }
+
+        // tensors cached in state (for tracing)
```
laggui: I think the pytorch comment "for tracing" doesn't apply here 😄 it likely refers to their specific `torch.jit.trace`

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,935 @@
...
+        let mut d = flat_grad.clone().neg();
+        let mut t = lr;
+        let mut prev_flat_grad: Option<Tensor<B::InnerBackend, 1>> = None;
+        let mut n_iter = 0;
```
laggui: These variables should be retrieved from persistent state, not reinitialized on each `step()` call.

In the [reference implementation](https://github.com/pytorch/pytorch/blob/v2.10.0/torch/optim/lbfgs.py#L212), the global state is updated after each step:

```py
state["d"] = d
state["t"] = t
state["old_dirs"] = old_dirs
state["old_stps"] = old_stps
state["ro"] = ro
state["H_diag"] = H_diag
state["prev_flat_grad"] = prev_flat_grad
state["prev_loss"] = prev_loss
```

and this state is used to initialize the following step state.

Furthermore, the history is explicitly cleared if `n_iter == 1` (lines 535-538), which should be a global condition.

This makes the implementation batch-local, which I believe is incorrect.

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,935 @@
...
+    x2: f64,
+    f2: f64,
+    g2: f64,
+    bounds: Option<(f64, f64)>,
+) -> f64 {
+    // Compute bounds of interpolation area
+    let (min_bound, max_bound) = bounds.unwrap_or(if x1 <= x2 { (x1, x2) } else { (x2, x1) });
+    // Code for most common case: cubic interpolation of 2 points
+    // with function and derivative values for both
+    // Soulution in this case (where x2 is the farthest point)
```
laggui: Typo: Solution

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,935 @@
...
+/// LBFGS Configuration.
+#[derive(Config, Debug)]
+pub struct LBFGSConfig {
+    /// maximal number of iterations per optimization step (default: 20)
+    #[config(default = 20)]
+    pub max_iter: usize,
+    /// update history size (default: 100).
+    #[config(default = 100)]
+    pub history_size: usize,
+    /// ermination tolerance on first order optimality (default: 1e-7).
```
laggui: Typo: Termination

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,935 @@
...
+    pub tolerance_grad: f64,
+    /// termination tolerance on function value/parameter changes (default: 1e-9).
+    #[config(default = 1e-9)]
+    pub tolerance_change: f64,
+    /// maximal number of function evaluations per optimization step (default: max_iter * 1.25).
+    #[config(default = "None")]
+    pub max_eval: Option<usize>,
+    ///  either ‘strong_wolfe’ or None (default: None).
+    #[config(default = "LineSearchFn::None")]
+    pub line_search_fn: LineSearchFn,
```
laggui: Doc comments are typically capitalized (first letter)

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,935 @@
...
+        Self {
+            history_s: Vec::new(),
+            history_y: Vec::new(),
+        }
+    }
+}
+
+/// L-BFGS optimizer.
+#[derive(Clone)]
+pub struct LBFGS<B: Backend + AutodiffBackend> {
```
laggui, line 525: I think we should add more documentation.

1. We should point to the clear reference for the implementation (pytorch -> minFunc)
2. We should add a performance `# Note` since this optimizer is memory intensive

donjuanplatinum:
> Thanks for contributing!
> 
> I think the implementation has some mistakes for the global optimizer state.
> 
> This also cannot be used with `Optimizer` and `SimpleOptimizer` traits which are designed around parameter-local (typically first-order optimizers like SGD/Adam).
> 
> LBFGS requires a closure-based interface (to evaluate loss multiple times for line search) and maintains global state across all parameters. We might need a new trait for such methods, but this is a special case. So not sure how that might fit. With the current implementation, users can only leverage LBFGS in their own training loop.


The current trait interface isn't a great fit for the L-BFGS implementation, so I’ve implemented it as a struct method for now. However, if the trait becomes more generalized in future updates, wrapping this interface should be straightforward. Also, I have refactored the code according to your suggestions.
:)

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,972 @@
...
+            x2 - (x2 - x1) * ((g2 + d2 - d1) / (g2 - g1 + 2.0 * d2))
+        } else {
+            x1 - (x1 - x2) * ((g1 + d2 - d1) / (g1 - g2 + 2.0 * d2))
+        };
+        min_pos.max(min_bound).min(max_bound)
+    } else {
+        (min_bound + max_bound) / 2.0
+    }
+}
+/// Auxiliray Struct For Strong_Wolfe
```
laggui: Typo:
```suggestion
/// Auxiliary Struct For Strong_Wolfe
```

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,972 @@
...
+    /// 搜索方向
+    pub d: Option<Tensor<B, 1>>,
+    /// 上一次的步长 t
+    pub t: Option<f64>,
+    /// 上一次的扁平梯度
+    pub prev_flat_grad: Option<Tensor<B, 1>>,
+    /// 上一次的 Loss (用于断点检查)
+    pub prev_loss: Option<f64>,
+    /// 全局迭代次数 (用于判断是否需要执行初次初始化)
+    pub g_iter: usize,
```
laggui: Could you translate the comments please? 🙏

crates/burn-optim/src/optim/lbfgs.rs
```diff
@@ -0,0 +1,972 @@
...
+                        .state
+                        .history_s
+                        .iter()
+                        .zip(self.state.history_y.iter())
+                        .zip(alphas.iter())
+                        .take(num_old)
+                    {
+                        let rho = 1.0 / y.clone().dot(s.clone()).into_scalar().to_f64();
+                        let beta = rho * y.clone().dot(r.clone()).into_scalar().to_f64();
+                        r = r.add(s.clone().mul_scalar(alpha - beta));
```
laggui: Didn't leave this comment last time around, but we generally try to avoid `into_scalar` which introduces synchronization points.

For the conditions (e.g. `ys > 1e-10`) there is no other choice anyway, but to compute values we should keep those as tensor-tensor operations.

I think we could preserve the scalar as a 1D tensor with a single element, and use `mul(val.unsqueeze())` instead. Lmk what you think!

laggui (changes requested):
Thanks for addressing my comments! I think the implementation is much closer 👍 

Just have some minor comments on doc/typos, and a potential performance improvement.

donjuanplatinum:
> Thanks for addressing my comments! I think the implementation is much closer 👍
> 
> Just have some minor comments on doc/typos, and a potential performance improvement.

thank for your comments laggui! :)

Issue #4473, opened 2026-02-07
ClassificationOutput does not implement Adaptor for CerInput or WerInput for Character and Word error rates


nitrogar:
**Describe the bug**
ClassificationOutput does not implement Adapter for CerInput or WerInput for Character and Word error rates. therefore it seems I cant use WordErrorRate or CharErrorRate in Learner

**To Reproduce**

**Expected behavior**
WER and CER should be as easy as other metric to use, just add the metrics

nitrogar:
or maybe it should be implemented in MultiLabelClassificationOutput ?

softmaximalist:
Will work on this. The existing outputs do not work with CER and WER.

PR #4474, opened 2026-02-08
Add SequenceOutput struct for sequence prediction outputs

softmaximalist:
### Checklist

### Related Issues/PRs

Closes https://github.com/tracel-ai/burn/issues/4473

### Changes

Added a new file crates/burn-train/src/learner/sequence.rs which implements `SequenceOutput` struct. This will allow users to use  sequence-based metrics (e.g., CER, WER). Also adapted it for CER, WER, and Loss metrics. Additionally, adapted ClassificationOutput for AUROC. 

### Testing

None

### DIFF
--- a/burn-book/src/building-blocks/metric.md
+++ b/burn-book/src/building-blocks/metric.md
@@ -94,6 +94,10 @@ impl<B: Backend> Adaptor<AccuracyInput<B>> for ClassificationOutput<B> {
 }
 ```
 
+You can also open an issue on the [GitHub repository](https://github.com/tracel-ai/burn) when your task type is 
+not covered by the built-in output structs. However, since creating an output struct for your data is simple, 
+it is recommended to try creating your own output struct first. 
+
 # Custom Metric
 
 Generating your own custom metrics is done by implementing the `Metric` trait.


crates/burn-train/src/learner/sequence.rs
```diff
@@ -0,0 +1,57 @@
...
+use burn_ndarray::NdArray;
+
+/// Sequence prediction output adapted for CER, WER, and Loss metrics.
+#[derive(new)]
+pub struct SequenceOutput<B: Backend> {
+    /// The loss.
+    pub loss: Tensor<B, 1>,
+
+    /// The predicted token indices. Shape: \[batch_size, seq_length\].
+    pub output: Tensor<B, 2, Int>,
```
laggui: To make it compatible with other token prediction metrics (e.g., accuracy and perplexity), it should probably hold the raw logits too. Otherwise, the already sampled predictions cannot be used. And you can only compute CER or WER for the provided model output.

```rust
pub struct SequenceOutput<B: Backend> {
    pub loss: Tensor<B, 1>,
    /// Raw logits `[batch, seq_len, vocab_size]`
    pub logits: Tensor<B, 3>,
    /// Optional predicted tokens `[batch, seq_len]`
    pub predictions: Option<Tensor<B, 2, Int>>,
    /// Ground-truth targets `[batch, seq_len]`
    pub targets: Tensor<B, 2, Int>,
}
```

That way, logits can be used for other metrics. If the sampled `predictions` are not provided, we can default to argmax of logits on last dim.

And targets can be flattened based on expected metric input.

Lmk what you think.

burn-book/src/building-blocks/metric.md
```diff
@@ -30,38 +30,49 @@ throughout the training process. We currently offer a restricted range of metric
...
 }
-
-impl<B: Backend> Adaptor<LossInput<B>> for ClassificationOutput<B> {
-    fn adapt(&self) -> LossInput<B> {
-        LossInput::new(self.loss.clone())
-    }
-}
 ```
 
+If your task type is not covered by the built-in output structs, consider opening an issue on the [GitHub repository](https://github.com/tracel-ai/burn).
```
laggui: I'm not sure this should be the default action. Users can still use their own output type, and implement the adaptor in their project. Without any requirements from the upstream project (us).

crates/burn-train/src/learner/sequence.rs
softmaximalist: Good point. I will update the code.

burn-book/src/building-blocks/metric.md
softmaximalist: Would it be better to provide instructions on how users can use their own output type/struct while also mentioning that they can also have the option of creating an issue? 

I thought it would be better if contributors are aware of the output struct/types users usually need so we can update existing output types (to make them more flexible) or add new output types to make it easier for future users. Let me know what you think.

laggui: Yeah I think instructions for custom types would be a good start.

> I thought it would be better if contributors are aware of the output struct/types users usually need so we can update existing output types (to make them more flexible) 

Yeah and I agree! I don't think it should be removed. But it should come after the example to adapt for your own output type. So users don't automatically resolve to feature requests on github :slightly_smiling_face: 

/edit: just saw you removed this note entirely, feel free to add it at the end of the example! Code LGTM now.

softmaximalist:
I updated the burn book to re-add the note about opening github issues. The PR should be ready now.

PR #4486, opened 2026-02-09
fix overflow in int_abs_elem for i64 min value

Olexandr88:
Avoid potential overflow when calling abs on i64::MIN in int_abs_elem. Using saturating_abs ensures safe behavior for edge values without changing behavior for normal inputs.


Olexandr88:
Added regression test as requested. PTAL

laggui:

The test should be for all backends (under `burn-backend-tests`), but that's on me it wasn't very explicit in my comment :sweat_smile: I can take this PR over the finish line if needed

CI is struggling due to GH issues https://www.githubstatus.com/

Olexandr88:
Thanks! Happy to move the test to burn-backend-tests if needed  otherwise feel free to take it over.
CI failures seem GitHub-related

PR #4490, opened 2026-02-09
Update revs

nathanielsimard:
Update revs to latest cubecl and cubek


laggui:
Whoops just saw this PR. Revs had to be updated for #4467 so main is up to date now

PR #4497, opened 2026-02-10
Fix: create multiple elemwise fused block

nathanielsimard:
It was creating invalid kernels.


nathanielsimard:
> LGTM
> 
> A small regression test would be great though

Very hard to so somehow, I wasn't able to come up with a trace that triggers a problem, but the fusion code isn't designed to support what the condition avoid.

PR #4502, opened 2026-02-10
feat: Enable 64-bit indexing for kernels

wingertge:
### Checklist

### Changes

Implements 64-bit indexing for all `cubecl` kernels, when the handle is large enough to require it.

### Testing

Testing isn't currently implemented, since it's difficult. Will think about ways to add it in the future, but it's hard to run since any tests that hit the 64-bit path would necessarily use 4-16GB of VRAM, depending on the type.

### DIFF
--- a/crates/burn-rl/src/lib.rs
+++ b/crates/burn-rl/src/lib.rs
@@ -76,9 +76,7 @@ pub(crate) mod tests {
             (MockAction::batch(actions), contexts)
         }
 
-        fn update(&mut self, _update: Self::PolicyState) {
-            ()
-        }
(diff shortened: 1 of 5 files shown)


laggui:
> but it's hard to run since any tests that hit the 64-bit path would necessarily use 4-16GB of VRAM, depending on the type.

Yeah our current tests are usually minimal given the CI resources :sweat_smile: even locally my laptop only has 6GB VRAM lol

Redhawk18:
I can setup a github runner on my school computer with a 7700x and 3090ti if that helps. I might have access to a computer with a 7700x and 5080 soon aswell.

Issue #4488, opened 2026-02-09
Improve performance on replay buffer sampling.


Charles23R:
### Feature description

Instead of storing transitions in the replay buffer, store a tensor for states, actions, etc. and enable fast sampling with tensor ops.

### Feature motivation

Faster.

### (Optional) Suggest a Solution

Instead of one buffer (Vec) of transitions, the replay buffer has a field for every element to store (state, action, reward, etc.) each field is a tensor of size (batch_size, dim). When sampling, use the `gather` op. These tensors can still be on a cpu backend.

arferreira:
I'd like to take this on. I'll look into the current replay buffer implementation and put up a PR. Any preferences on whether this should be behind a feature flag or a full replacement?

nathanielsimard:
It can be a full replacement

arferreira:
Thanks! I've looked into the tensor API and plan to use select for batch sampling form pre-allocated tensors. Will put up a PR soon.

softmaximalist:
@Charles23R I think this issue can be closed.

PR #4504, opened 2026-02-11
Replace Vec-based TransitionBuffer with tensor-backed storage

arferreira:
### Checklist

### Related Issues/PRs

#4488

### Changes

Replaced Vec-based TransitionBuffer with tensor-backed storage.
Each transition field (states, actions, rewards, etc.) is now a pre-allocated 2D tensor.
Sampling uses `Tensor::select` with random indices instead of CPU-side iteration.
Removed `rand` dependency, index generation goes through the tensor API.
Breaking change to Transition, TransitionBatch, and TransitionBuffer signatures.

### Testing

Existing async_policy tests pass. Added tests for push, circular overwrite, sample shapes, and oversized batch panic.

### DIFF
--- a/crates/burn-rl/src/transition_buffer/base.rs
+++ b/crates/burn-rl/src/transition_buffer/base.rs
@@ -1,28 +1,31 @@
 use burn_core::{Tensor, prelude::Backend, tensor::Distribution};
+use derive_new::new;
+
+use super::SliceAccess;
 
 /// A state transition in an environment.
-#[derive(Clone)]
-pub struct Transition<B: Backend> {
+#[derive(Clone, new)]
+pub struct Transition<B: Backend, S, A> {
     /// The initial state.
-    pub state: Tensor<B, 1>,
+    pub state: S,
     /// The state after the step was taken.
-    pub next_state: Tensor<B, 1>,
+    pub next_state: S,
     /// The action taken in the step.
-    pub action: Tensor<B, 1>,
+    pub action: A,
     /// The reward.
-    pub reward: f32,
+    pub reward: Tensor<B, 1>,
     /// If the environment has reached a terminal state.
-    pub done: bool,
+    pub done: Tensor<B, 1>,
 }
 
 /// A batch of transitions.
-pub struct TransitionBatch<B: Backend> {
+pub struct TransitionBatch<B: Backend, SB, AB> {
     /// Batched initial states.
-    pub states: Tensor<B, 2>,
+    pub states: SB,
     /// Batched resulting states.
-    pub next_states: Tensor<B, 2>,
+    pub next_states: SB,
     /// Batched actions.
-    pub actions: Tensor<B, 2>,
+    pub actions: AB,
     /// Batched rewards.
     pub rewards: Tensor<B, 2>,
     /// Batched flags for terminal states.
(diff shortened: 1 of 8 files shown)


crates/burn-rl/src/transition_buffer/base.rs
```diff
@@ -1,319 +1,201 @@
...
-        } else {
-            self.buffer[self.cursor] = item;
-            self.cursor = (self.cursor + 1) % self.capacity;
+    /// Add a transition, overwriting the oldest if full.
+    pub fn push(&mut self, transition: Transition<B>) {
+        let idx = self.write_head % self.capacity;
+
+        self.states = self
+            .states
+            .clone()
```
nathanielsimard: Instead of cloning I think you can use inplace:

```rust
self.states.inplace(|states| states.slice_assign(idx..idx+1, transition.state.unsqueeze_dim(0));
```

Same for others.

arferreira: Done, switched to inplace for all the slice_assign calls in push

laggui (dismissed):
Going through PRs, this looks good to merge!

Just need to update the merge conflicts with the main branch (Cargo.lock)

arferreira:
Should I update burn-train to work with the new API in this PR, or would you prefer to handle that separately?

laggui:
Oh yeah I thought the changes were reflected at the call sites, must have misread that :sweat_smile: 

The PR should be self-contained, so the changes must be applied in burn-train.

arferreira:
Okay, to update burn-train I need State/Action types to be convertible to `Tensor<B, 1>`, and I need to know their dimensions at buffer creation time. I'm thinking of adding something like `state_dim()/action_dim()` to the Environment trait and a conversion bound on State/Action. Does that align with the design you have in mind, or do you want a different approach?

laggui:
Might make sense to have `State` and `Action` traits for the associated types.

CC @Charles23R to get his thoughts as he has done the work for burn-rl

Charles23R:
@arferreira For `burn-rl`, our vision is to have an API that is generic over pretty much any RL scenario. With that in mind, I don't think converting any state/action to `Tensor<B, 1>` will be feasible/desired unfortunately. Think of an environment where the state is 2-dimensionnal like an Atari screen representation, or even more complex like an audio-visual state.

In that sense, I think a good solution would be to keep the `TransitionBatch` and `TransitionBuffer` generic over `SB` and `SA`, i.e. a "batch" of states and a "batch" of actions. Those generics could implement a trait (let's call it `SliceAccess` for now) defining `select()` and `slice_assign()` for those types. You can imagine how this trait would be easy to implement for "tensorized" states/actions (see `examples/dqn-agent` for example). You can also use this example to test out how these abstractions can be implemented in practice.

Anyways, don't hesitate to reach out if this isn't clear, and thanks for the good work so far!

arferreira:
Implemented the SliceAccess trait as suggested by @Charles23R. Here's what changed:

- Added `SliceAccess<B>` trait in `burn-rl` with a default `Tensor<B, 2>` impl
- Restored generic `TransitionBatch<B, SB, AB>` and `TransitionBuffer<B, SB, AB>` with lazy-init tensor storage
- Restored `LearnerTransitionBatch<B, P>` type alias
- Updated `off_policy.rs` to use the new buffer API
- Added `SliceAccess` impls for the DQN example types (`ObservationTensor`, `DiscreteActionTensor`)

Issue #4506, opened 2026-02-12
Burn API limitations affecting ONNX operator coverage


antimora:
## Context

While expanding integration test coverage for [burn-onnx](https://github.com/tracel-ai/burn-onnx), we identified several places where Burn's API limits what ONNX operators can support. These are not bugs in burn-onnx; they are upstream API gaps that prevent full ONNX spec compliance.

## 1. Pad: only supports last 2 spatial dimensions

ONNX `Pad` allows padding any dimension, but burn-onnx rejects models that pad batch or channel dimensions because Burn's pad API is designed for spatial (H, W) dimensions only.

**ONNX spec**: Pad can apply to any dimension of any rank tensor.
**Current behavior**: Error: `"Pad: padding will only be applied to the last two dimensions"`

## 2. Pad constant_value: float-only

ONNX `Pad` with `mode=constant` accepts int and bool constant values, but Burn only supports f32.

**ONNX spec**: `constant_value` can be any numeric or bool type matching the input tensor.
**Current behavior**: Error: `"only float values are currently supported for constant value"`

## 3. Resize: only Interpolate1d and Interpolate2d exist

ONNX `Resize` works on any spatial rank. Burn provides `Interpolate1d` (rank 3) and `Interpolate2d` (rank 4) modules, but nothing for 3D spatial data (rank 5).

**ONNX spec**: Resize supports 1D, 2D, and 3D spatial data.
**Current behavior**: Rank 5 (3D spatial) falls back to a slower runtime path rather than using a dedicated module.

## 4. LayerNorm: only normalizes the last dimension

ONNX `LayerNormalization` has an `axis` attribute that specifies the first normalization dimension; normalization spans `[axis, ..., rank-1]`. Burn's `LayerNorm` always normalizes the last dimension.

**ONNX spec**: `axis` can be any value in `[-rank, rank)`.
**Current behavior**: Error: `"LayerNorm: normalization is only supported on the last axis right now"`

## Impact

These limitations cause burn-onnx to reject valid ONNX models that use these features. They are documented in burn-onnx with TODO/FIXME comments referencing the Burn API as the blocker.

None of these are urgent, but addressing any of them would directly expand ONNX model compatibility.

antimora:
Linking https://github.com/tracel-ai/burn/issues/4269

antimora:
[...]

PR #4507, opened 2026-02-12
feat: support padding on arbitrary dimensions

### LINKED ISSUE #4269: Support padding on more dimensions
`tensor.pad(...)` is currently limited to the last two dimensions (typically for images, requires `(left, right, top, bottom)`).

Linked to this ONNX import failure: https://github.com/tracel-ai/burn-onnx/issues/18#issuecomment-3701698923

antimora:
## Summary

- Generalize `tensor.pad()` to accept `(before, after)` pairs for any dimension, not just the last two
- Add `IntoPadding` trait for backward compatibility: the old `(left, right, top, bottom)` tuple still works alongside the new `&[(before, after)]` slice/array/Vec formats
- All existing callers compile without changes

## Motivation

1st item is done in #4506 (item 1: Pad only supports last 2 spatial dimensions)
Closes #4269
Unblocks https://github.com/tracel-ai/burn-onnx/pull/153

ONNX `Pad` allows padding any dimension, but Burn's pad API was limited to the last two (height/width). This prevented `burn-onnx` from supporting models that pad batch or channel dimensions.

## Changes

| File | Change |
|------|--------|
| `burn-tensor/.../pad.rs` | `IntoPadding` trait, `expand_padding` helper, generalized `pad()` signature and internal functions |
| `burn-tensor/.../mod.rs` | Re-export `IntoPadding` |
| `burn-backend/.../base.rs` | Updated `PadMode` docs |
| `burn-backend-tests/.../padding.rs` | 10 new tests for N-dim padding |

## API

```rust
// Old API still works (backward compatible)
tensor.pad((1, 1, 1, 1), PadMode::Constant(0.0));

// New: pad arbitrary dimensions with (before, after) pairs
tensor.pad([(1, 1), (2, 2)], PadMode::Constant(0.0));

// Fewer pairs than dims pads only the last N dimensions
tensor.pad([(1, 1)], PadMode::Reflect);

// Pad batch dimension of a 4D tensor
tensor.pad([(1, 0), (0, 0), (0, 0), (0, 0)], 0.0);
```

## Test plan

- [x] All 16 existing padding tests pass unchanged (backward compat verified)
- [x] 10 new tests covering: pairs API, single-dim pad, all-dims 4D pad, batch-dim padding for all 3 modes, too-many-pairs panic
- [x] Tested with `cargo test-ndarray padding` (33/33 passed)

### DIFF
--- a/crates/burn-tensor/src/tensor/api/pad.rs
+++ b/crates/burn-tensor/src/tensor/api/pad.rs
@@ -7,39 +7,75 @@ use super::Numeric;
 
 /// Trait for types that can be used as padding specifications.
 ///
-/// Padding is specified as `(before, after)` pairs per dimension.
-/// If fewer pairs than dimensions are provided, they apply to the **last** N dimensions
-/// (earlier dimensions are left unpadded).
-pub trait IntoPadding {
-    /// Converts into a list of `(before, after)` padding pairs.
-    fn into_padding(self) -> Vec<(usize, usize)>;
+/// Padding is specified as `(before, after)` pairs per dimension, returned as a
+/// fixed-size array `[(usize, usize); D]`. If fewer pairs than dimensions are provided,
+/// they apply to the **last** N dimensions (earlier dimensions are left unpadded).
+pub trait IntoPadding<const D: usize> {
+    /// Converts into a fixed-size array of `(before, after)` padding pairs.
+    fn into_padding(self) -> [(usize, usize); D];
 }
 
-impl IntoPadding for Vec<(usize, usize)> {
-    fn into_padding(self) -> Vec<(usize, usize)> {
-        self
-    }
-}
-
-impl IntoPadding for &[(usize, usize)] {
-    fn into_padding(self) -> Vec<(usize, usize)> {
-        self.to_vec()
-    }
-}
-
-impl<const N: usize> IntoPadding for [(usize, usize); N] {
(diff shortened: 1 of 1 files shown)


crates/burn-tensor/src/tensor/api/pad.rs
```diff
@@ -5,6 +5,44 @@ use crate::{Element, ElementConversion, Tensor, backend::Backend, ops::PadMode};
...
 
+/// Trait for types that can be used as padding specifications.
+///
+/// Padding is specified as `(before, after)` pairs per dimension.
+/// If fewer pairs than dimensions are provided, they apply to the **last** N dimensions
+/// (earlier dimensions are left unpadded).
+pub trait IntoPadding {
+    /// Converts into a list of `(before, after)` padding pairs.
+    fn into_padding(self) -> Vec<(usize, usize)>;
+}
```
laggui: Instead of returning a `Vec` and combining that with `expand_padding`, why not return an array for `into_padding`?

Would avoid vec allocations, and always return a fixed size array as a contract.

antimora: Good call! Updated `IntoPadding` to be generic over `D` and return `[(usize, usize); D]` directly. Each impl now constructs the fixed-size array in place, no Vec allocations. The separate `expand_padding` helper is gone too.

Issue #4415, opened 2026-01-31
Add support for runtime backend selection in Burn


HernandoR:
### Feature description

Add support for runtime backend selection in Burn, enabling applications to dynamically detect available hardware (e.g., with/without of a GPU) at execution time and dispatch computations to the most appropriate backend (e.g., WGPU for GPU acceleration, NdArray/Tch CPU for GPU-less platforms). This feature should retain Burn’s existing compile-time backend guarantees for performance where needed, while adding a flexible runtime dispatch layer that works with the full ecosystem of Burn backends (NdArray, WGPU, Tch, etc.) and core abstractions (Tensor, Module, Optimizer).

### Feature motivation

Burn currently selects backends exclusively at compile time via Rust generics and feature flags, which creates friction for cross-platform and hybrid Python/Rust applications that target heterogeneous environments:
1. Hardware agnostic deployment: Many production/edge platforms have mixed GPU/GPU-less hardware (e.g., cloud instances with optional GPUs, embedded devices, user-facing Python apps). Compiling separate binaries for each backend (GPU/CPU) is cumbersome and scales poorly with additional backends.

2. Python/Rust hybrid workloads: When accelerating Python applications with Burn (e.g., converting large numpy arrays to Burn tensors for computation), developers cannot know the end user’s hardware at compile time. A runtime backend switch allows a single distribution (e.g., a Python wheel) to support both GPU and CPU execution. i.e. #4411 

3. Graceful degradation: Applications can fall back to a CPU backend (e.g., NdArray) if a GPU backend (e.g., WGPU) fails to initialize (e.g., no compatible GPU, missing drivers) instead of crashing entirely.

#### DynamicBackend
Implement a runtime backend dispatch layer built on Burn’s existing generic backend traits (Backend, TensorBackend, etc.) with two key components:
1. A unified DynamicBackend enum/struct: Wraps all supported compile-time backends (NdArray, WGPU, Tch) and implements Burn’s core backend traits by dispatching method calls to the underlying concrete backend instance. This acts as a single "dynamic" backend that the rest of the application can use generically.

2. Hardware detection utilities: Add a set of top-level helper functions (e.g., burn::backend::detect_available(), burn::backend::is_gpu_available()) that check for hardware/driver support (e.g., WGPU adapter existence, Tch CUDA support) and return a list of usable backends.

#### Workflow for End Users

1. Compile the application with all desired backends enabled via feature flags (e.g., --features ndarray, wgpu, tch).

[...]

PR #4508, opened 2026-02-12
[Feat] Global backend `Dispatch`

### LINKED ISSUE #3628: Default backend implementation
With the goal of removing some boilerplate, I think it would be good to start considering a default backend. Right now, it would greatly improve backend selection for most applications, and potentially, we could make it the default everywhere. I think it’s important to consider these things before Burn 1.0. The reason the `Backend` trait was included in the user interface was to allow custom operations to be added to Burn via backend extensions and to reduce the compilation time/binary size of deployed models. Also, I thought that maybe some third-party closed-source backends would be implemented, but I don’t think it’s really important anymore since we have many in-house backends. Here’s what I have in mind:

```rust
use burn_core::{
    prelude::Backend,
    tensor::ops::{Device, FloatTensor, FloatTensorOps, IntTensorOps},
};

pub struct MultiBackend;

impl Backend for MultiBackend {
    type Device = MultiDevice;
    type FloatTensorPrimitive = MultiFloatTensor;
    type IntTensorPrimitive = MultiIntTensor;
}

pub enum MultiFloatTensor {
    #[cfg(feature = "ndarray")]
    NdArray(FloatTensor<crate::backend::NdArray>),
    #[cfg(feature = "wgpu")]
    Wgpu(FloatTensor<crate::backend::Wgpu>),
    #[cfg(feature = "autodiff")]
    Autodiff(FloatTensor<burn_autodiff::Autodiff<MultiBackend>>),
}

pub enum MultiIntTensor {
    #[cfg(feature = "ndarray")]
    NdArray(IntTensor<crate::backend::NdArray>),
    #[cfg(feature = "wgpu")]
    Wgpu(IntTensor<crate::backend::Wgpu>),
    #[cfg(feature = "autodiff")]
    Autodiff(IntTensor<burn_autodiff::Autodiff<MultiBackend>>),
}

pub enum MultiDevice {
    #[cfg(feature = "ndarray")]
    NdArray(Device<crate::backend::NdArray>),
    #[cfg(feature = "wgpu")]
    Wgpu(Device<crate::backend::Wgpu>),
    #[cfg(feature = "autodiff")]
    Autodiff(Box<Device<MultiBackend>>),
}

fn float_mul(lhs: MultiFloatTensor, rhs: MultiFloatTensor) -> MultiFloatTensor {
    ops!(float(lhs, rhs) => float_mul)
}

fn int_mul(lhs: MultiIntTensor, rhs: MultiIntTensor) -> MultiIntTensor {
    ops!(int(lhs, rhs) => int_add)
}

macro_rules! ops {
    (float($($t:ident),*) => $op:ident) => {
        ops!($($t),* | FloatTensorOps | $op | MultiFloatTensor)
    };

    (int($($t:ident),*) => $op:ident) => {
        ops!($($t),* | IntTensorOps | $op | MultiIntTensor)
    };

    ($($t:ident),* | $trait:ident | $op:ident | $tensor:ident) => {
        match ($($t,)*) {
            #[cfg(feature = "ndarray")]
            ($($tensor::NdArray($t),)*) => {
                use burn_ndarray::NdArray;

                $tensor::NdArray(<NdArray as $trait<NdArray>>::$op($($t,)*))
            }
            #[cfg(feature = "wgpu")]
            ($($tensor::Wgpu($t),)*) => {
                use burn_wgpu::Wgpu;

                $tensor::Wgpu(<Wgpu as $trait<Wgpu>>::$op($($t,)*))
            }
            #[cfg(feature = "autodiff")]
            ($($tensor::Autodiff($t),)*) => {
                use burn_autodiff::Autodiff;

                $tensor::Autodiff(<Autodiff<MultiBackend> as $trait<Autodiff<MultiBackend>>>::$op($($t,)*))
            }
            _ => panic!("Invalid device."),
        }

    };
}
#[cfg(feature = "wgpu")]
use burn_core::tensor::ops::IntTensor;
pub(crate) use ops;

```

With this implementation, the primitive tensor becomes an enum, and we match on its backend type for each operation. This is extremely lightweight and a no-op when only a single backend is activated. The backend could also support autodiff, but it would require a recursive type definition. I'm unsure if that part would work out, but I think it will. If it doesn't, the backend is still valuable, but we can't remove the generic backend from the code since it won't work for training anymore.

For backend extensions, it only means that the new ops trait would need to be implemented for `MultiBackend`, which isn’t a big deal, since the `ops` macro could almost dispatch it correctly.

laggui:
### Checklist

### Related Issues/PRs

Closes #3628 

Also captures #4415 but doesn't close it entirely

### Changes

- Moved `BoolTensorOps` element ops to `Scalar` (still used `BoolElem`)
- Added `Dispatch` multi backend implementation with a set of flexible macros to dispatch based on a backend list
- `Backend::ad_enabled` now takes a `device: &Self::Device` arg
- Added `inner()` method to `DeviceOps` trait (defaults to a no-op)

Example usage:
```rust
let device = DispatchDevice::Cuda(Default::default());
let device_cpu = DispatchDevice::Cpu(Default::default());

let zeros_cuda = Tensor::<Dispatch, 2>::zeros([128, 128], &device);
let zeros = zeros_cuda.clone().to_device(&device_cpu);
```

Automatic device selection + autodiff:
```rust
fn select_device() -> DispatchDevice {
    // ...
}

fn main() {
    let device = select_device();
    training::run::<Dispatch>(DispatchDevice::autodiff(device));
}
```

I initially had a compilation error when multiple wgpu backends were enabled (mutually exclusive atm), but changed it to a warning and doesn't enable any wgpu backends.

e.g. on CI (due to the workspace):
```
warning: burn-dispatch@0.21.0-pre.1: Only one WGPU backend can be enabled at once. Detected: [vulkan, webgpu]. No WGPU backend will be available in this build. This is expected in workspace builds. For production, enable only one of: metal, vulkan, or webgpu.
```

TODO
- [ ] `DispatchDevice` method to set default dtype
- [ ] Better cross-backend `to_device` implementations (currently, all go through `into_data` + `from_data`)
- [ ] Remove `Backend` default elem type generics (following #4404 and #4373)
- [ ] Consolidate `Backend` trait tensor primitives into a single primitive

### Testing

mnist example w/ `Dispatch` backend

### DIFF
--- a/crates/burn-dispatch/src/backend.rs
+++ b/crates/burn-dispatch/src/backend.rs
@@ -386,7 +386,7 @@ impl DispatchTensor {
             #[cfg(feature = "tch")]
             DispatchTensor::LibTorch(tensor) => DispatchDevice::LibTorch(tensor.device()),
             #[cfg(feature = "autodiff")]
-            DispatchTensor::Autodiff(tensor) => DispatchDevice::new_autodiff(tensor.device()),
+            DispatchTensor::Autodiff(tensor) => DispatchDevice::autodiff(tensor.device()),
         }
     }
 }
--- a/crates/burn-dispatch/src/device.rs
+++ b/crates/burn-dispatch/src/device.rs
@@ -59,7 +59,7 @@ pub enum DispatchDevice {
 // This tuple struct mainly restricts users from creating Autodiff(Autodiff) devices.
 /// A wrapper that enables automatic differentiation for a [`DispatchDevice`].
 ///
-/// Use [`DispatchDevice::new_autodiff`] to construct this type.
+/// Use [`DispatchDevice::autodiff`] to construct this type.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct AutodiffDevice(pub(crate) Box<DispatchDevice>);
 
@@ -168,7 +168,7 @@ const TYPE_ID_BASE: u16 = 10;
 
 impl DispatchDevice {
     /// Creates a new [`DispatchDevice`] with [automatic differentiation](Autodiff) enabled.
-    pub fn new_autodiff(device: impl Into<DispatchDevice>) -> DispatchDevice {
+    pub fn autodiff(device: impl Into<DispatchDevice>) -> DispatchDevice {
         let device = device.into();
         DispatchDevice::Autodiff(AutodiffDevice(Box::new(device)))
     }
--- a/examples/mnist/examples/mnist.rs
+++ b/examples/mnist/examples/mnist.rs
@@ -42,5 +42,5 @@ fn select_device() -> DispatchDevice {
 
 fn main() {
     let device = select_device();
-    training::run::<Dispatch>(DispatchDevice::new_autodiff(device));
+    training::run::<Dispatch>(DispatchDevice::autodiff(device));
 }


crates/burn-dispatch/src/ops/transaction.rs
```diff
@@ -0,0 +1,6 @@
+use burn_backend::ops::TransactionOps;
+
+use crate::Dispatch;
+
+// TODO?
+impl TransactionOps<Self> for Dispatch {}
```
nathanielsimard: We should dispatch to the underlying implementation.

PR #4509, opened 2026-02-12
allow flash attention with causal

louisfd:
Causal attention (and out of bounds which was the real problem) is fixed in dd100f1ec956f97c1b1fb0c8170244783a17eb33

### DIFF
--- a/crates/burn-cubecl-fusion/src/optim/reduce_broadcasted/launch.rs
+++ b/crates/burn-cubecl-fusion/src/optim/reduce_broadcasted/launch.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use cubecl::{
     Runtime,
-    ir::{AddressType, ElemType, FloatKind, StorageType},
+    ir::{ElemType, FloatKind, StorageType},
     prelude::{ComputeClient, ScalarArg, SequenceArg},
     server::LaunchError,
     std::CubeOptionArgs,


antimora:
@louisfd  you can also uncomment this test for metal in crates/burn-backend-tests/tests/tensor/float/module/attention.rs:

image

crates/burn-cubecl-fusion/src/optim/matmul/optimization.rs
```diff
@@ -455,6 +455,7 @@ impl FusedMatmulLaunch<'_> {
             rhs_strides,
             out_strides,
             dtypes.as_global_elems(),
+            AddressType::default(),
```
laggui: Cubecl changes hadn't made their way into main yet, but they have now that https://github.com/tracel-ai/burn/pull/4502 was merged.

So if you update with main this should be resolved.

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
     "crates/burn-store/pytorch-tests",
     "crates/burn-store/safetensors-tests",
     "crates/burn-collective/multinode-tests",
-    "examples/*",
+    # "examples/*",
```
laggui: Probably committed that by accident?

Cargo.toml
```diff
@@ -176,7 +176,7 @@ portable-atomic-util = { version = "0.2.5", features = ["alloc"] }
 ### For the main burn branch. ###
 cubecl = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "11c32460621652077898d3309430711ab3d7944d" }
 cubecl-common = { git = "https://github.com/tracel-ai/cubecl", default-features = false, rev = "11c32460621652077898d3309430711ab3d7944d" }
-cubek = { git = "https://github.com/tracel-ai/cubek", default-features = false, rev = "e42d33198091d78e145268849f6269ebac57362f" }
+cubek = { git = "https://github.com/tracel-ai/cubek", default-features = false, rev = "b121ed35e19bf9e0935dd563e97cab6e7a76005e" }
```
laggui, line 179: Merged the updates myself, and bumped to the latest rev at the same time

PR #4511, opened 2026-02-12
Perf: Improve fusion score

nathanielsimard:
Instead of using the number of operations fused as the score, use the number of IO saved.

### DIFF
--- a/crates/burn-fusion/src/stream/multi.rs
+++ b/crates/burn-fusion/src/stream/multi.rs
@@ -267,7 +267,7 @@ impl<R: FusionRuntime> MultiStream<R> {
                 .analyse(current, node, streams, &self.streams);
             match analysis {
                 SharedTensorAnalysis::SharedFromCurrentStream => {
-                    shared_analysis.current.push(node.id);
+                    shared_analysis.current.push((node.id, node.status));
                 }
                 SharedTensorAnalysis::NotShared => {}
                 SharedTensorAnalysis::SharedFromExistingStream {
@@ -295,7 +295,7 @@ impl<R: FusionRuntime> MultiStream<R> {
         nodes: &[&TensorIr],
     ) {
         // If we only have current tensors that are shared, we're safe to not sync the timelines.
-        if analysis.new.is_empty() && analysis.existing.is_empty() {
+        if analysis.new.is_empty() && analysis.existing.is_empty() && analysis.current.is_empty() {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


crates/burn-fusion/src/search/block.rs
```diff
@@ -224,7 +224,8 @@ fn find_best_optimization_index<O>(
     for (i, optimization) in optimizations.iter().enumerate() {
         let properties = optimization.properties();
 
-        if properties.ready && properties.score >= best_score {
+        // A score of zero is worse than fusing.
+        if properties.ready && properties.score > best_score {
```
laggui, line 228: The change from `>=` to `>` for a score of zero seems to be the issue for failing tests

```
test fusion::tensor::multi_threads::should_handle_multi_threads_consumed ... FAILED
test fusion::tensor::multi_threads::should_handle_multi_threads_consumed_no_wait ... FAILED
```

Previously it would have returned an index here, now it goes straight to `ExecutionStrategy::Operations`.

I checked locally real quick, when calling `into_data()` looks like it's still trying to execute optimizations when draining the stream

```
thread '<unnamed>' (622904) panicked at /home/laggui/workspace/burn/crates/burn-ir/src/handle.rs:88:32:
Should have handle for tensor TensorId { value: 0 }
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: burn_ir::handle::HandleContainer<H>::get_handle
   3: burn_cubecl_fusion::engine::launch::input::InputPlanner<R>::run
   4: burn_cubecl_fusion::engine::launch::base::FuseTraceLauncher<R,Runner>::launch
   5: burn_cubecl::fusion::<impl burn_fusion::backend::Optimization<burn_cubecl::fusion::FusionCubeRuntime<R,BT>> for burn_cubecl_fusion::optim::base::CubeOptimization<R>>::execute
   6: burn_fusion::stream::execution::ordering::OrderedExecution<R>::execute_optimization
   7: burn_fusion::stream::queue::execution::QueueExecution<R>::execute_strategy
   8: burn_fusion::stream::queue::execution::<impl burn_fusion::stream::queue::base::OperationQueue<R>>::execute
   9: burn_fusion::stream::execution::processor::Processor<O>::process
  10: burn_fusion::stream::multi::MultiStream<R>::drain
  11: burn_fusion::client::GlobalFusionClient<R>::read_tensor_float
  12: burn_fusion::ops::tensor::<impl burn_backend::backend::ops::tensor::FloatTensorOps<burn_fusion::backend::Fusion<B>> for burn_fusion::backend::Fusion<B>>::float_into_data::{{closure}}::{{closure}}
```

So perhaps the fuser being more strict (>) creates a synchronization gap mid-stream? The single op creates a handle, but it's no longer available to future optimization attempts (out of the optimization path)?

PR #4518, opened 2026-02-13
feat: add align_corners support to InterpolateOptions

### LINKED ISSUE #4510: Add align_corners support to InterpolateOptions
This blocks https://github.com/tracel-ai/burn-onnx/pull/168

## Summary

`InterpolateOptions` currently only has a `mode` field. It lacks support for `align_corners` (and more broadly, coordinate transformation modes), which causes significant numerical errors when importing ONNX models that use `coordinate_transformation_mode: "align_corners"` in Resize nodes.

## Problem

ONNX's Resize operator supports a `coordinate_transformation_mode` attribute with values like `half_pixel` (default), `align_corners`, `asymmetric`, `pytorch_half_pixel`, etc. The most commonly used are `half_pixel` and `align_corners`.

Burn's interpolation always uses `half_pixel` semantics with no way to specify alternatives. The two modes use fundamentally different coordinate mappings:

| Mode | Coordinate mapping |
|---|---|
| `half_pixel` | `(out_coord + 0.5) / out_size * in_size - 0.5` |
| `align_corners` | `out_coord / (out_size - 1) * (in_size - 1)` |

## Real-world impact

**Depth-Anything-v2** (a widely-used monocular depth estimation model) uses `align_corners` bilinear interpolation in its DPT head for all 5 upsampling stages. When imported via burn-onnx, the output has:

- **Mean absolute error: 0.59** (on values ranging 1.5-4.2)
- **Max absolute error: 1.08**

The error at each individual Resize step is ~0.05 mean / ~0.35 max for a 2x upsample, but it compounds through cascaded Resize+Conv stages.

This isn't a niche issue. `align_corners=True` is the default in PyTorch's `F.interpolate` for many vision models, and transformer-based architectures (DPT, SegFormer, etc.) commonly use it in their upsampling heads.

### 1. `InterpolateOptions` (burn-backend)

`crates/burn-backend/src/backend/ops/modules/base.rs`:

```rust
pub struct InterpolateOptions {
    pub mode: InterpolateMode,
    pub align_corners: bool,  // new field, default false
}
```

Note: `GridSampleOptions` already has `align_corners: bool` with the correct semantics, so there's precedent in the codebase.

### 2. ndarray backend implementation

`crates/burn-ndarray/src/ops/interpolate.rs`:

The bilinear/bicubic interpolation kernels need to use different source coordinate computation based on the flag:

```rust
// Current (half_pixel):
let src = (dst as f32 + 0.5) * (in_size as f32 / out_size as f32) - 0.5;

// align_corners:
let src = dst as f32 * ((in_size - 1) as f32 / (out_size - 1) as f32);
```

### 3. `Interpolate2dConfig` / `Interpolate1dConfig` (burn-nn)

Add the config option so the module API also supports it:

```rust
pub struct Interpolate2dConfig {
    pub output_size: Option<[usize; 2]>,
    pub scale_factor: Option<[f32; 2]>,
    pub mode: InterpolateMode,
    pub align_corners: bool,  // new field, default false
}
```

### 4. Other backends

CubeCL/wgpu/tch backends would also need the coordinate mapping update. The `tch` backend can pass it directly to libtorch which already supports it.

## Context

This was discovered while adding a Depth-Anything-v2 model check to [burn-onnx](https://github.com/tracel-ai/burn-onnx). The burn-onnx codegen already extracts the `coordinate_transformation_mode` from ONNX models but has no way to emit it in the generated Burn code.

antimora:
## Summary

Closes #4510

- Adds `align_corners: bool` field to `InterpolateOptions` (default `true` for backward compatibility)
- When `true`: corner pixels are aligned between input/output (`src = dst * (in_size - 1) / (out_size - 1)`)
- When `false`: half-pixel coordinate mapping (`src = (dst + 0.5) * in_size / out_size - 0.5`)
- Updated all backends: ndarray, CubeCL, tch, and IR layer
- Added `align_corners` to `Interpolate1dConfig` / `Interpolate2dConfig` module configs

## Test plan

- [x] All existing interpolation tests pass (backward compatible, default = `true`)
- [x] New `test_upsample_half_pixel` tests for bilinear and bicubic with `align_corners=false`
- [x] `cargo check` passes for burn-backend, burn-ir, burn-ndarray, burn-cubecl, burn-fusion, burn-nn, burn-autodiff

### DIFF
--- a/crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
+++ b/crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
@@ -46,10 +46,10 @@ fn interpolate_bicubic_kernel<F: Float>(
     let yw = Line::empty(line_size).fill(F::cast_from(frac - y_in_f));
 
     // Clamp indices in float space to handle negative coordinates from half_pixel
-    let y0 = f32::clamp(y_in_f - 1.0, 0.0, input_height_f) as usize;
-    let y1 = f32::clamp(y_in_f, 0.0, input_height_f) as usize;
-    let y2 = f32::clamp(y_in_f + 1.0, 0.0, input_height_f) as usize;
-    let y3 = f32::clamp(y_in_f + 2.0, 0.0, input_height_f) as usize;
+    let y0 = clamp(y_in_f - 1.0, 0.0, input_height_f) as usize;
+    let y1 = clamp(y_in_f, 0.0, input_height_f) as usize;
+    let y2 = clamp(y_in_f + 1.0, 0.0, input_height_f) as usize;
+    let y3 = clamp(y_in_f + 2.0, 0.0, input_height_f) as usize;
 
     let input_width = input.shape(2) - 1;
     let input_width_f = input_width as f32;
@@ -66,10 +66,10 @@ fn interpolate_bicubic_kernel<F: Float>(
     let xw = Line::empty(line_size).fill(F::cast_from(frac - x_in_f));
 
     // Clamp indices in float space to handle negative coordinates from half_pixel
-    let x0 = f32::clamp(x_in_f - 1.0, 0.0, input_width_f) as usize;
-    let x1 = f32::clamp(x_in_f, 0.0, input_width_f) as usize;
-    let x2 = f32::clamp(x_in_f + 1.0, 0.0, input_width_f) as usize;
-    let x3 = f32::clamp(x_in_f + 2.0, 0.0, input_width_f) as usize;
+    let x0 = clamp(x_in_f - 1.0, 0.0, input_width_f) as usize;
+    let x1 = clamp(x_in_f, 0.0, input_width_f) as usize;
+    let x2 = clamp(x_in_f + 1.0, 0.0, input_width_f) as usize;
+    let x3 = clamp(x_in_f + 2.0, 0.0, input_width_f) as usize;
 
     let index_base = b * input.stride(0) + c * input.stride(3);
     let in_stride_y = input.stride(1);
--- a/crates/burn-cubecl/src/kernel/interpolate/bilinear.rs
+++ b/crates/burn-cubecl/src/kernel/interpolate/bilinear.rs
@@ -38,7 +38,7 @@ fn interpolate_bilinear_kernel<F: Float>(
     } else {
         let in_size = input.shape(1) as f32;
         let out_size = output.shape(1) as f32;
-        f32::clamp(
+        clamp(
             (y as f32 + 0.5) * (in_size / out_size) - 0.5,
             0.0,
             in_size - 1.0,
@@ -61,7 +61,7 @@ fn interpolate_bilinear_kernel<F: Float>(
     } else {
         let in_size = input.shape(2) as f32;
         let out_size = output.shape(2) as f32;
-        f32::clamp(
+        clamp(
             (x as f32 + 0.5) * (in_size / out_size) - 0.5,
             0.0,
             in_size - 1.0,


crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
```diff
@@ -31,31 +32,44 @@ fn interpolate_bicubic_kernel<F: Float>(
...
 
-    let y0 = select(y_in != 0, y_in - 1, 0);
-    let y1 = y_in;
-    let y2 = clamp_max(y_in + 1, input_height);
-    let y3 = clamp_max(y_in + 2, input_height);
+    // Clamp indices in float space to handle negative coordinates from half_pixel
+    let y0 = f32::clamp(y_in_f - 1.0, 0.0, input_height_f) as usize;
+    let y1 = f32::clamp(y_in_f, 0.0, input_height_f) as usize;
+    let y2 = f32::clamp(y_in_f + 1.0, 0.0, input_height_f) as usize;
+    let y3 = f32::clamp(y_in_f + 2.0, 0.0, input_height_f) as usize;
```
wingertge: If you're going to use function syntax anyways, the freestanding `clamp` is cleaner than the associated function

crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
```diff
@@ -31,31 +32,44 @@ fn interpolate_bicubic_kernel<F: Float>(
...
 
-    let x0 = select(x_in != 0, x_in - 1, 0);
-    let x1 = x_in;
-    let x2 = clamp_max(x_in + 1, input_width);
-    let x3 = clamp_max(x_in + 2, input_width);
+    // Clamp indices in float space to handle negative coordinates from half_pixel
+    let x0 = f32::clamp(x_in_f - 1.0, 0.0, input_width_f) as usize;
+    let x1 = f32::clamp(x_in_f, 0.0, input_width_f) as usize;
+    let x2 = f32::clamp(x_in_f + 1.0, 0.0, input_width_f) as usize;
+    let x3 = f32::clamp(x_in_f + 2.0, 0.0, input_width_f) as usize;
```
wingertge: Same here
```suggestion
    // Clamp indices in float space to handle negative coordinates from half_pixel
    let x0 = clamp(x_in_f - 1.0, 0.0, input_width_f) as usize;
    let x1 = clamp(x_in_f, 0.0, input_width_f) as usize;
    let x2 = clamp(x_in_f + 1.0, 0.0, input_width_f) as usize;
    let x3 = clamp(x_in_f + 2.0, 0.0, input_width_f) as usize;
```

crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
wingertge:
```suggestion
    let y0 = clamp(y_in_f - 1.0, 0.0, input_height_f) as usize;
    let y1 = clamp(y_in_f, 0.0, input_height_f) as usize;
    let y2 = clamp(y_in_f + 1.0, 0.0, input_height_f) as usize;
    let y3 = clamp(y_in_f + 2.0, 0.0, input_height_f) as usize;
```

crates/burn-cubecl/src/kernel/interpolate/bilinear.rs
```diff
@@ -45,10 +54,19 @@ fn interpolate_bilinear_kernel<F: Float>(
...
+        let denominator = clamp_min(output.shape(2) - 1, 1) as f32;
+        x as f32 * (numerator / denominator)
+    } else {
+        let in_size = input.shape(2) as f32;
+        let out_size = output.shape(2) as f32;
+        f32::clamp(
+            (x as f32 + 0.5) * (in_size / out_size) - 0.5,
+            0.0,
+            in_size - 1.0,
+        )
```
wingertge: And here
```suggestion
        clamp(
            (x as f32 + 0.5) * (in_size / out_size) - 0.5,
            0.0,
            in_size - 1.0,
        )
```

wingertge (changes requested):

crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
antimora: Fixed in becf35b. Switched to freestanding `clamp()` for all instances in both bicubic and bilinear kernels.

crates/burn-cubecl/src/kernel/interpolate/bicubic.rs
antimora: Fixed in same commit.

crates/burn-cubecl/src/kernel/interpolate/bilinear.rs
antimora: Fixed in same commit.

wingertge (approved):

Issue #4459, opened 2026-02-05
Adding support for CTC and RNNT loss functions


nitrogar:
### Feature description
I am doing speech to text application, and I noticed the missing CTC and RNNT loss functions which are mandatory for most stt model training.

an implementation of CTC and RNNT compatible with torch implementation to support model inference if the model was trained on pytorch
### Feature motivation
training and inferencing Speech to Text models

for Speech to Text applications

softmaximalist:
I will work on CTC loss.

muzammil35:
I see that the PR from @softmaximalist is merged, so in that case is this issue closed?

laggui:
It's only partially complete. The PR implemented CTC loss but did not extend it for RNNT.

softmaximalist:
I believe this issue should be closed now that both losses have been implemented.

dnvt:
For anyone looking for a CTC training example — I published a speech training pipeline that uses burn-nn’s CTCLoss in a real training loop with LibriSpeech data. Shows dynamic batching, checkpoint management, and evaluation: https://github.com/dnvt/burn-speech-training

PR #4529, opened 2026-02-16
Implement CTC loss

softmaximalist:
### Checklist

### Related Issues/PRs

[Adding support for CTC and RNNT loss functions #4459](https://github.com/tracel-ai/burn/issues/4459)

### Changes

Added a vectorized implementation of the CTC loss using the log-sum-exp trick for numerical stability. While the `crates/burn-nn/src/loss/ctc.rs` file contains only the forward pass where the code iterates over time steps t, the computation graph used by the autodiff is not broken so autodiff and gradient computation works automatically. Thus, there is no need to implement the backward pass from the original CTC paper. 

### Tests

Tests were added for configurations, panics, etc. Additionally, PyTorch comparison tests were also implemented where the expected losses and expected gradient of the test cases are the values produced by PyTorch. 

#### PyTorch Script

<details>
<summary>Click here to view the code snippet</summary>

```Python
import torch
import torch.nn.functional as F
import math


def generate_logits(T, N, C):
    """Deterministic logits: sin((t*7 + n*13 + c*3) * 0.1)."""
    data = []
    for t in range(T):
        for n in range(N):
            for c in range(C):
                data.append(math.sin((t * 7 + n * 13 + c * 3) * 0.1))
    return torch.tensor(data, dtype=torch.float32).reshape(T, N, C)


def run_test(name, T, N, C, targets_flat, target_shape, input_lengths,
             target_lengths, blank=0, reduction='none'):
    print(f"\n{'=' * 70}")
    print(f"  {name}")
    print(f"  T={T}, N={N}, C={C}, blank={blank}, reduction={reduction}")
    print(f"{'=' * 70}")

    logits = generate_logits(T, N, C).requires_grad_(True)
    log_probs = F.log_softmax(logits, dim=2)

    targets = torch.tensor(targets_flat, dtype=torch.long).reshape(target_shape)
    il = torch.tensor(input_lengths, dtype=torch.long)
    tl = torch.tensor(target_lengths, dtype=torch.long)

    loss = F.ctc_loss(log_probs, targets, il, tl, blank=blank,
                      reduction=reduction, zero_infinity=False)

    if reduction == 'none':
        print(f"\nLoss (per sample): {loss.tolist()}")
        loss_for_backward = loss.sum()
    else:
        print(f"\nLoss ({reduction}): {loss.item():.10f}")
        loss_for_backward = loss

    loss_for_backward.backward()
    grad = logits.grad

    # Print gradient as a flat Rust-pasteable array
    flat = grad.flatten().tolist()
    print(f"\nGradient (flat, {len(flat)} values):")
    # Print in rows of 10 for readability
    for i in range(0, len(flat), 10):
        chunk = flat[i:i+10]
        print("  " + ", ".join(f"{v:.10f}" for v in chunk) + ",")


run_test(
    "comparison_uniform_input_lengths",
    T=5, N=3, C=4,
    targets_flat=[1, 2, 0,  1, 0, 0,  3, 2, 1],
    target_shape=[3, 3],
    input_lengths=[5, 5, 5],
    target_lengths=[2, 1, 3],
)

run_test(
    "comparison_repeated_labels",
    T=8, N=4, C=6,
    targets_flat=[1, 1, 2, 0,  2, 3, 2, 1,  5, 0, 0, 0,  1, 2, 3, 4],
    target_shape=[4, 4],
    input_lengths=[8, 8, 8, 8],
    target_lengths=[3, 4, 1, 4],
)

run_test(
    "comparison_long_sequence",
    T=10, N=2, C=8,
    targets_flat=[1, 3, 5, 7, 2,  2, 4, 6, 1, 3],
    target_shape=[2, 5],
    input_lengths=[10, 10],
    target_lengths=[5, 5],
)

run_test(
    "comparison_mixed_input_lengths",
    T=12, N=3, C=5,
    targets_flat=[1, 4, 2, 0,  3, 1, 0, 0,  2, 4, 1, 3],
    target_shape=[3, 4],
    input_lengths=[12, 7, 10],
    target_lengths=[3, 2, 4],
)

run_test(
    "comparison_sum_reduction",
    T=5, N=3, C=4,
    targets_flat=[1, 2, 0,  1, 0, 0,  3, 2, 1],
    target_shape=[3, 3],
    input_lengths=[5, 5, 5],
    target_lengths=[2, 1, 3],
    reduction='sum',
)

run_test(
    "comparison_mean_reduction",
    T=5, N=3, C=4,
    targets_flat=[1, 2, 0,  1, 0, 0,  3, 2, 1],
    target_shape=[3, 3],
    input_lengths=[5, 5, 5],
    target_lengths=[2, 1, 3],
    reduction='mean',
)
```
</details>

### Benchmark Results
Given the complexity of the CTC loss, a nested loop version of it was implemented first as follows (the code is from the forward method):
```Rust
let mut loss_tensor = Tensor::<B, 1>::zeros([batch_size], &device);
for n in 0..batch_size {
    let input_len = input_lengths.clone().slice(s![n]).into_scalar().to_usize();
    let target_len = target_lengths.clone().slice(s![n]).into_scalar().to_usize();
    let blank_inserted_target = blank_inserted_targets
        .clone()
        .slice(s![n..n + 1, ..])
        .squeeze_dim::<1>(0);

    for t in 1..input_len {
        for s in 0..(2 * target_len + 1) {
            let log_alpha_t_s = self.compute_log_alpha_t_s::<B>(...)
```

This would require a separate implementation for the backward pass so the submitted vectorized version was implemented to simplify the code and improve performance. Since the PyTorch CPU implementation uses nested loop where the loop over the samples (batch dimension) is parallelized, the nested loop version was also modified to implement a parallelized loop version. Note that CTC 0 is the submitted vectorized version, CTC 1 is the nested loop version, and CTC 2 is the nested loop version with parallelization (using rayon). 

#### CTC Loss 0 Performance Benchmark (NdArray CPU)

| Config | Avg Time | Throughput |
| :--- | :--- | :--- |
| Small  B=2  T=10  C=5   L=3 | 0.22 ms | 8953 samples/s |
| Medium B=4  T=50  C=20  L=10 | 1.66 ms | 2405 samples/s |
| Large  B=8  T=150 C=50  L=30 | 7.18 ms | 1113 samples/s |
| Long T B=2  T=500 C=30  L=20 | 12.40 ms | 161 samples/s |
| Wide C B=4  T=50  C=100 L=10 | 1.35 ms | 2961 samples/s |
| Big B  B=32 T=50  C=20  L=10 | 3.52 ms | 9094 samples/s |

#### CTC Loss 1 Performance Benchmark (NdArray CPU)

| Config | Avg Time | Throughput |
| :--- | :--- | :--- |
| Small  B=2  T=10  C=5   L=3 | 0.49 ms | 4082 samples/s |
| Medium B=4  T=50  C=20  L=10 | 15.75 ms | 254 samples/s |
| Large  B=8  T=150 C=50  L=30 | 287.58 ms | 28 samples/s |
| Long T B=2  T=500 C=30  L=20 | 173.84 ms | 12 samples/s |
| Wide C B=4  T=50  C=100 L=10 | 16.14 ms | 248 samples/s |
| Big B  B=32 T=50  C=20  L=10 | 130.21 ms | 246 samples/s |

#### CTC Loss 2 Performance Benchmark (NdArray CPU)

| Config | Avg Time | Throughput |
| :--- | :--- | :--- |
| Small  B=2  T=10  C=5   L=3 | 0.64 ms | 3131 samples/s |
| Medium B=4  T=50  C=20  L=10 | 8.06 ms | 496 samples/s |
| Large  B=8  T=150 C=50  L=30 | 99.24 ms | 81 samples/s |
| Long T B=2  T=500 C=30  L=20 | 96.33 ms | 21 samples/s |
| Wide C B=4  T=50  C=100 L=10 | 9.30 ms | 430 samples/s |
| Big B  B=32 T=50  C=20  L=10 | 44.38 ms | 721 samples/s |

#### Benchmark Script

The command `cargo test benchmark_ctc[x]_performance --release -- --nocapture` was used to benchmark the CTC [X] implementation by running the following test case.

<details>
<summary>Click here to view the code snippet</summary>

```Rust
#[test]
fn benchmark_ctc0_performance() {
    use burn::tensor::activation::log_softmax;
    use std::time::Instant;

    fn run_config(
        label: &str,
        batch_size: usize,
        time_steps: usize,
        num_classes: usize,
        target_len: usize,
        iterations: u32,
        warmup: u32,
    ) {
        let device = NdArrayDevice::Cpu;
        let ctc = CTCLossConfig::new().init();

        let logits = Tensor::<TestBackend, 3>::random(
            [time_steps, batch_size, num_classes],
            Distribution::Uniform(0.0, 1.0),
            &device,
        );
        let log_probs = log_softmax(logits, 2);

        let targets = Tensor::<TestBackend, 2>::random(
            [batch_size, target_len],
            Distribution::Uniform(1.0, (num_classes - 1) as f64),
            &device,
        )
        .int();

        let input_lengths =
            Tensor::<TestBackend, 1, Int>::full([batch_size], time_steps as i64, &device);
        let target_lengths =
            Tensor::<TestBackend, 1, Int>::full([batch_size], target_len as i64, &device);

        for _ in 0..warmup {
            let _ = ctc.forward(
                log_probs.clone(),
                targets.clone(),
                input_lengths.clone(),
                target_lengths.clone(),
            );
        }

        let start = Instant::now();
        for _ in 0..iterations {
            let loss = ctc.forward(
                log_probs.clone(),
                targets.clone(),
                input_lengths.clone(),
                target_lengths.clone(),
            );
            let _ = loss.into_data();
        }
        let duration = start.elapsed();
        let avg_ms = (duration.as_secs_f64() * 1000.0) / iterations as f64;
        let throughput = (batch_size as f64 * iterations as f64) / duration.as_secs_f64();

        println!(
            "  [{:<40}]  avg: {:>8.2} ms   throughput: {:>8.0} samples/s",
            label, avg_ms, throughput
        );
    }

    println!("\n=== CTC Loss Performance Benchmark (NdArray CPU) ===");
    println!(
        "  {:<42}  {:>12}   {:>20}",
        "Config", "Avg Time", "Throughput"
    );
    println!("  {}", "-".repeat(80));

    run_config("Small  B=2  T=10  C=5   L=3", 2, 10, 5, 3, 50, 5);
    run_config("Medium B=4  T=50  C=20  L=10", 4, 50, 20, 10, 20, 5);
    run_config("Large  B=8  T=150 C=50  L=30", 8, 150, 50, 30, 5, 2);
    run_config("Long T B=2  T=500 C=30  L=20", 2, 500, 30, 20, 5, 2);
    run_config("Wide C B=4  T=50  C=100 L=10", 4, 50, 100, 10, 10, 3);
    run_config("Big B  B=32 T=50  C=20  L=10", 32, 50, 20, 10, 10, 3);

    println!();
}
```
</details>

### Note

The benchmark is very simple and implemented to make sure that vectorized implementation is faster than loop versions. If there are any issues with any part of this PR, please let me know.

### DIFF
--- a/crates/burn-nn/src/loss/ctc.rs
+++ b/crates/burn-nn/src/loss/ctc.rs
@@ -365,7 +365,7 @@ impl CTCLoss {
         let s_is_not_blank_mask = l_prime_s.clone().not_equal_elem(self.blank as i64);
         let s_not_equal_s_minus_2_mask = l_prime_s.not_equal(l_prime_s_minus_2);
 
-        // The 2 leftmost columns of temp_l_prime_mask should only contain false.
+        // The 2 leftmost columns of the returned mask should only contain false.
         // These are invalid positions since s - 2 is a valid index only when s >= 2.
         let col_indices = Tensor::<B, 1, Int>::arange(0..(max_l_prime_len as i64), device)
             .reshape([1, max_l_prime_len])


crates/burn-nn/src/loss/ctc.rs
```diff
@@ -0,0 +1,1712 @@
...
+    ) -> Tensor<B, 1> {
+        let ctc_loss_tensor =
+            self.forward(log_probs, targets, input_lengths, target_lengths.clone());
+
+        match reduction {
+            Reduction::Auto | Reduction::Mean => {
+                // Following PyTorch's behavior where the output losses are divided
+                // by the target lengths and then the mean over the batch is taken
+                let target_lengths_float = target_lengths.float();
+                ctc_loss_tensor.div(target_lengths_float).mean()
```
laggui, line 240: This is a helpful comment, but I think we should also move it up to the documentation as a note.

crates/burn-nn/src/loss/ctc.rs
```diff
@@ -0,0 +1,1712 @@
...
+            shift_by == 1 || shift_by == 2,
+            "The parameter shift_by must 1 or 2"
+        );
+
+        let [rows, cols] = org_2d_tensor.dims();
+        let padding_shape = [rows, shift_by];
+        let padding_tensor = if org_2d_tensor.dtype().is_float() {
+            Tensor::<B, 2, K>::full(padding_shape, f32::NEG_INFINITY, device)
+        } else {
+            Tensor::<B, 2, K>::full(padding_shape, 0, device)
```
laggui, line 346: Could using 0 potentially conflict with a valid class index and `blank` ? Might have to use a different value here?

crates/burn-nn/src/loss/ctc.rs
```diff
@@ -0,0 +1,1712 @@
...
+            .squeeze_dim::<1>(1);
+        let log_likelihood = self.log_sum_exp(log_alpha_last_blank, log_alpha_last_label, &device);
+        let mut ctc_loss_tensor = log_likelihood.neg();
+
+        if self.zero_infinity {
+            let inf_mask = ctc_loss_tensor.clone().is_inf();
+            ctc_loss_tensor = ctc_loss_tensor
+                .clone()
+                .mask_where(inf_mask, ctc_loss_tensor.clone().zeros_like());
+        }
```
laggui, line 188: Should we also check for NaNs? Or is still specifically an Inf thing?

crates/burn-nn/src/loss/ctc.rs
softmaximalist: I added this to the docs of the forward_with_reduction method (to the description of the `reduction` parameter).

crates/burn-nn/src/loss/ctc.rs
softmaximalist: Nice catch! The only place where `right_shift_2d_tensor` is used on an int tensor is in the `create_l_prime_mask` method. 

Note that the mask returned by `create_l_prime_mask` should contain `false` for the 2 leftmost columns. This is because these are invalid positions (`s-2` is a valid index only when $s \geq 2$). Hence, I created another mask to apply when creating the returned `l_prime_mask`:

```Rust
let col_indices = Tensor::<B, 1, Int>::arange(0..(max_l_prime_len as i64), device)
            .reshape([1, max_l_prime_len])
            .expand([batch_size, max_l_prime_len]);
let s_greater_than_1_mask = col_indices.greater_equal_elem(2);

s_is_not_blank_mask
    .bool_and(s_not_equal_s_minus_2_mask)
    .bool_and(s_greater_than_1_mask)
```

Since the 2 leftmost columns of the returned are always `false`, using 0 in `right_shift_2d_tensor` should not cause any issues (we can use number instead and the code should work the same).

crates/burn-nn/src/loss/ctc.rs
softmaximalist: This part only checks for infinities and I believe this is how CTC loss works in PyTorch. Regarding NaN values, I am not sure what is the right approach to deal with them. In the case of CTC loss, since we use the log-sum-exp trick, we deal with negative infinities and the tensor returned by `compute_log_alpha_t_s` can contain neg-infs. In other words, the tensors containing some inf values are expected. On the other hand, I believe NaN values would be much rare and they might indicate a more serious issue with the data/tensors passed in and/or the actual implementation of the CTC loss. 

Let me know what you think. I can modify the code if you think it is better to explicitly handle NaNs.

laggui: Ok sounds fair, we can keep it as is!

crates/burn-nn/src/loss/ctc.rs
laggui: Ahhh ok thanks for the clarification.

PR #4545, opened 2026-02-19
fix: resolve macOS build and test failures

antimora:
## Summary

`cargo build` and `cargo test` fail on macOS out of the box. This PR fixes three issues:

- **Vulkan feature breaks workspace build**: `examples/guide` had `burn/vulkan` in its default features. Due to workspace feature unification, this forced `cubecl-wgpu` to build with `spirv`, which panics without the Vulkan SDK. macOS uses Metal natively, so vulkan should not be required by default. Moved to an opt-in feature, matching the pattern used by other examples.

- **CUDA tests fail on macOS**: `burn-cuda` has a test that tries to load the CUDA shared library, which doesn't exist on macOS (NVIDIA dropped macOS support). Gated the test module with `#[cfg(not(target_os = "macos"))]`.

- **WGPU dtype test fails on macOS**: Without the `metal` feature flag, wgpu still uses Metal at runtime on macOS. The fallback test branch assumed a non-Metal GPU and asserted F64 support, which Metal doesn't provide. Added a macOS-specific branch matching actual runtime capabilities.

## Test plan

- [x] `cargo check --workspace` succeeds on macOS without Vulkan SDK
- [x] `cargo test -p burn-cuda --lib` passes on macOS (test is skipped)
- [x] `cargo test -p burn-wgpu --lib` passes on macOS
- [ ] CI passes on all platforms

### DIFF
--- a/examples/guide/Cargo.toml
+++ b/examples/guide/Cargo.toml
@@ -11,6 +11,7 @@ workspace = true
 
 [features]
 default = ["burn/default", "burn/tui"]
+# Opt-in for macOS users with the Vulkan SDK installed (enabled by default on other platforms)
 vulkan = ["burn/vulkan"]
 
 [dependencies]


examples/guide/Cargo.toml
laggui, line 22: It would be nice to keep vulkan by default for the example, but there is no way to have those platform-dependent.

Maybe we could enable on `burn` instead?
```
[target.'cfg(not(target_os = "macos"))'.dependencies]
burn = { version = "...", default-features = false, features = [..., "vulkan"] }

[target.'cfg(target_os = "macos")'.dependencies]
burn = { version = "...", default-features = false, features = [...] }
```

What do you think?

laggui (commented):
LGTM, but maybe we can keep vulkan on the example?

antimora:
@laggui Good idea! Added a platform-conditional dependency:

```toml
[target.'cfg(not(target_os = "macos"))'.dependencies]
burn = { path = "../../crates/burn", features = ["vulkan"] }
```

This keeps vulkan by default on Linux/Windows while excluding it on macOS. The `vulkan` feature is still available for macOS users who have the Vulkan SDK installed via `--features vulkan`.

examples/guide/Cargo.toml
laggui, line 15: Maybe just add a comment that it is specifically for macos users that want to enable it?

antimora:
Added the comment, thanks!

PR #4547, opened 2026-02-19
Add Smooth L1 loss

softmaximalist:
### Checklist

### Related Issues/PRs

### Changes

Implemented the Smooth L1 loss with a single struct field `beta`. Since Smooth L1 loss becomes L1 loss when beta = 0, beta must be set to a positive value. For beta = 0, users can use LpLoss (which has a constructor for L1 loss). I believe this is different from PyTorch which states in its documentation that beta must be non-negative. 

### Testing

Added 17 test cases covering different configurations and different forward methods.

### DIFF
--- a/crates/burn-nn/src/loss/smooth_l1.rs
+++ b/crates/burn-nn/src/loss/smooth_l1.rs
@@ -210,11 +210,7 @@ impl SmoothL1Loss {
         sorted_dims.sort();
 
         // Reduce over specified dimensions
-        let mut result = error;
-        for &dim in sorted_dims.iter().rev() {
-            result = result.mean_dim(dim);
-        }
-        result
+        error.mean_dims(sorted_dims.as_slice())
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/burn-nn/src/loss/smooth_l1.rs
```diff
@@ -0,0 +1,524 @@
...
+
+        // Sort the dimensions to ascending order
+        let mut sorted_dims = dims.to_vec();
+        sorted_dims.sort();
+
+        // Reduce over specified dimensions
+        let mut result = error;
+        for &dim in sorted_dims.iter().rev() {
+            result = result.mean_dim(dim);
+        }
```
laggui: We actually have `tensor.mean_dims(..)` which could be used with the `sorted_dims`.

softmaximalist: I used this `mean_dim` approach in the implementation of `LpLoss` so I didn't notice it. I will update both smooth_l1.rs and lp_loss.rs to use `mean_dims` instead.

laggui: Didn't even realize in `LpLoss` :smile: thanks!

PR #4554, opened 2026-02-21
Attention: add autotune gate

### LINKED ISSUE #4553: burn-cubecl: missing #[cfg(feature = "autotune")] on AttentionStrategy::Autotune match arm
## Description

In `crates/burn-cubecl/src/kernel/attention/base.rs` at rev `26dbd53b`, the `AttentionStrategy::Autotune` enum variant is correctly gated behind `#[cfg(feature = "autotune")]` (line 31), but the corresponding match arm at line 85 is not:

```rust
// Line 31 - correctly gated
#[cfg(feature = "autotune")]
Autotune,

// Line 85 - missing gate
AttentionStrategy::Autotune => {
    attention_autotune(query, key, value, mask, attn_bias, options, out)
}
```

## How to reproduce

Any downstream crate that depends on `burn` with `default-features = false` and enables a GPU backend (`wgpu`, `metal`) without explicitly enabling `autotune` will fail to compile:

```
error[E0599]: no variant or associated item named `Autotune` found for enum `AttentionStrategy`
```

Burn's own CI passes because test configurations enable `autotune` transitively through default features.

## Fix

Add `#[cfg(feature = "autotune")]` to the match arm at line 85.

louisfd:
Fix #4553

### DIFF
--- a/crates/burn-cubecl/src/kernel/attention/base.rs
+++ b/crates/burn-cubecl/src/kernel/attention/base.rs
@@ -82,6 +82,7 @@ pub fn attention<R: CubeRuntime>(
             );
             Ok(out)
         }
+        #[cfg(feature = "autotune")]
         AttentionStrategy::Autotune => {
             attention_autotune(query, key, value, mask, attn_bias, options, out)
         }


antimora (approved):
The fix works with burn-onnx. 

Thank you!

PR #4555, opened 2026-02-21
Add multi-scale SSIM for image quality assessment

softmaximalist:
### Checklist

### Related Issues/PRs

### Changes

Implemented multi-scale SSIM vision metric for image quality assessment. 

It uses Gaussian kernel to perform separable convolution. Before performing convolution, the image tensors are reflect padded (same as torchmetrics). While the original paper [Multi-scale Structural Similarity for Image Quality Assessment](https://www.cns.nyu.edu/pub/eero/wang03b.pdf) mentions applying low-pass filter and then downsampling by 2, the implementation simply performs 2d average pooling (same as torchmetrics). 

The default beta values are [0.0448, 0.2856, 0.3001, 0.2363, 0.1333] (5 scales) which are from the original paper. Users can provide custom beta values which must sum up to 1. The alpha, beta, and gamma values for each iteration are set to be equal to each other and the length of the betas vector is the number of scales. 

The original paper defines:

$$
\begin{aligned}
l(x, y) &= \frac{2\mu_x \mu_y + C_1}{\mu_x^2 + \mu_y^2 + C_1} \\
c(x, y) &= \frac{2\sigma_x \sigma_y + C_2}{\sigma_x^2 + \sigma_y^2 + C_2} \\
s(x, y) &= \frac{\sigma_{xy} + C_3}{\sigma_x \sigma_y + C_3}
\end{aligned}
$$
 
and MS-SSIM is defined as:

$$
\text{MS-SSIM}(x, y) = [l_M(x, y)]^{\alpha_M} \cdot \prod_{j=1}^{M} [c_j(x, y)]^{\beta_j} \cdot [s_j(x, y)]^{\gamma_j}
$$

Note that $[c_j(x, y)]^{\beta_j} \cdot [s_j(x, y)]^{\gamma_j} = [c_j(x, y)s_j(x, y)]^{\beta_j}$ since $\beta_j = \gamma_j$. Then, we can simplify $c_j(x, y)s_j(x, y)$ as follows:

$$
\begin{aligned}
s(x, y) &= \frac{\sigma_{xy} + C_3}{\sigma_x\sigma_y + C_3} \\
&= \frac{\sigma_{xy} + \frac{C_2}{2}}{\sigma_x\sigma_y + \frac{C_2}{2}} \\
&= \frac{2\left(\sigma_{xy} + \frac{C_2}{2}\right)}{2\left(\sigma_x\sigma_y + \frac{C_2}{2}\right)} \\
&= \frac{2\sigma_{xy} + C_2}{2\sigma_x\sigma_y + C_2}
\end{aligned}
$$

Multiply $c(x, y)$ and the simplified $s(x, y)$, canceling the common $(2\sigma_x\sigma_y + C_2)$ term:

$$
\begin{aligned}
c(x, y) \cdot s(x, y) &= \left( \frac{2\sigma_x\sigma_y + C_2}{\sigma_x^2 + \sigma_y^2 + C_2} \right) \cdot \left( \frac{2\sigma_{xy} + C_2}{2\sigma_x\sigma_y + C_2} \right) \\
&= \frac{2\sigma_{xy} + C_2}{\sigma_x^2 + \sigma_y^2 + C_2}
\end{aligned}
$$

Thus, instead of computing the construct and structure terms separately, the code computes this simplified product of construct and structure terms. 

### Testing

Added 18 test cases covering different configurations, panics, ms-ssim scores. Additionally, the tests cover batch averaging, running average, ms-ssim scores with custom betas vector, etc.

### DIFF
--- a/crates/burn-train/src/metric/vision/ms_ssim.rs
+++ b/crates/burn-train/src/metric/vision/ms_ssim.rs
@@ -201,8 +201,8 @@ impl MsSsimMetricConfig {
 
     /// Sets custom betas for the scales. The length of the betas vector
     /// determines the number of scales used in the MS-SSIM computation.
-    /// If you to make different parameter settings comparable, the betas
-    /// vector should sum to 1 as per the original paper. However, note 
+    /// If you want to make different parameter settings comparable, the betas
+    /// vector should sum to 1 as per the original paper. However, note
     /// that this is not a strict requirement.
     ///
     /// # Default value
@@ -426,9 +426,10 @@ impl<B: Backend> Metric for MsSsimMetric<B> {
         // Shape: [N, C]
         let batch_size = dims[0];
         let channels = dims[1];
-        let mut ms_ssim_tensor = Tensor::<B, 2>::ones([batch_size, channels], &item.outputs.device());
+        let mut ms_ssim_tensor =
+            Tensor::<B, 2>::ones([batch_size, channels], &item.outputs.device());
 
-        for i in 0..betas.len() {
+        for (j, beta_j) in betas.iter().enumerate() {
             // Compute mu_x and mu_y
             let mu_x = self.gaussian_separable_conv(x.clone());
             let mu_y = self.gaussian_separable_conv(y.clone());
@@ -448,10 +449,9 @@ impl<B: Backend> Metric for MsSsimMetric<B> {
             // Compute cs_map = (2σxy + C2) / (σx² + σy² + C2)
             // This is mathematically equivalent to c(x,y) * s(x,y) when C3 = C2 / 2
             let contrast_structure = (cov_xy * 2.0 + c2) / (var_x + var_y + c2);
-            let beta_j = betas[i];
 
             // Include luminance at the last scale
-            if i == betas.len() - 1 {
+            if j == betas.len() - 1 {
                 // Compute l(x, y) = (2μxμy + C1) / (μx² + μy² + C1)
                 let luminance: Tensor<B, 4> =
                     (2 * mu_x * mu_y + c1) / (square_of_mu_x + square_of_mu_y + c1);
@@ -459,14 +459,14 @@ impl<B: Backend> Metric for MsSsimMetric<B> {
                 let ssim_spatial_mean = ssim.mean_dims(&[2, 3]).reshape([batch_size, channels]);
                 // Clamp to avoid negative values before raising to power (prevents NaNs)
                 let ssim_mean_clamped = ssim_spatial_mean.clamp_min(0.0);
-                ms_ssim_tensor = ms_ssim_tensor * ssim_mean_clamped.powf_scalar(beta_j);
+                ms_ssim_tensor = ms_ssim_tensor * ssim_mean_clamped.powf_scalar(*beta_j);
             } else {
                 let contrast_structure_spatial_mean = contrast_structure
                     .mean_dims(&[2, 3])
                     .reshape([batch_size, channels]);
                 // Clamp to avoid negative values before raising to power (prevents NaNs)
                 let c_s_mean_clamped = contrast_structure_spatial_mean.clamp_min(0.0);
-                ms_ssim_tensor = ms_ssim_tensor * c_s_mean_clamped.powf_scalar(beta_j);
+                ms_ssim_tensor = ms_ssim_tensor * c_s_mean_clamped.powf_scalar(*beta_j);
 
                 x = avg_pool2d(x, [2, 2], [2, 2], [0, 0], false, false);
                 y = avg_pool2d(y, [2, 2], [2, 2], [0, 0], false, false);
@@ -861,12 +861,6 @@ mod tests {
     #[should_panic(expected = "All beta values must be non-negative")]
     fn test_ms_ssim_negative_betas() {
         let _ = MsSsimMetricConfig::new(1.0).with_betas(vec![0.3, 0.3, -0.1, 0.5]);
-    }
-
-    #[test]
-    #[should_panic(expected = "The sum of the betas must be 1.0")]
-    fn test_ms_ssim_invalid_betas_sum() {
-        let _ = MsSsimMetricConfig::new(1.0).with_betas(vec![0.5, 0.3]);
     }
 
     #[test]


crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+/// should be at least `kernel_size * 2^(scales-1)` (e.g., 11 × 2^4 = 11 * 16 = 176 for default kernel size).
+/// If your images are smaller, reduce the kernel size or number of scales.
+///
+/// # Example
+/// ```rust,ignore
+/// // Create input for RGB images
+/// let preds: Tensor<B, 4> = /* tensor */;
+/// let targets: Tensor<B, 4> = /* tensor */;
+/// let input = MsSsimInput::new(outputs, targets);
+/// ```
```
antimora, line 36:
```suggestion
+/// let input = MsSsimInput::new(preds, targets);
```
The example declares `preds` on line 33 but passes `outputs` here.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+            "Shape mismatch: preds {:?} targets {:?}",
+            preds.dims(),
+            targets.dims()
+        );
+        Self { preds, targets }
+    }
+}
+
+/// Configuration for the [MsMsSsimMetric].
+#[derive(Debug, Clone)]
```
antimora, line 68:
```suggestion
+/// Configuration for the [MsSsimMetric].
```
Typo: `MsMsSsimMetric` should be `MsSsimMetric`.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+    pub kernel_size: usize,
+    /// The MS-SSIM metric involves applying convolution to the input tensors using a Gaussian kernel.
+    /// This is the standard deviation of the Gaussian kernel. Default is 1.5.
+    pub sigma: f32,
+    /// The number of channels in the input images (e.g., 1 for grayscale, 3 for RGB).
+    /// This is used to create the appropriate convolution kernels. Default is 3.
+    pub channels: usize,
+    /// The weights/betas for each scale in the MS-SSIM computation. We
+    /// The length of this vector determines the number of scales.
+    /// Default is \[0.0448, 0.2856, 0.3001, 0.2363, 0.1333\] (5 scales).
```
antimora, line 96:
```suggestion
+    /// The weights/betas for each scale in the MS-SSIM computation.
```
Dangling "We" - looks like the sentence was left incomplete during an edit.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+        assert!(pixel_range > 0.0, "pixel_range must be positive");
+        Self {
+            k1: 0.01,
+            k2: 0.03,
+            pixel_range,
+            kernel_size: 11,
+            sigma: 1.5,
+            channels: 3,
+            betas: vec![0.0448, 0.2856, 0.3001, 0.2363, 0.1333],
+        }
```
antimora, line 136: The default betas sum to `1.0001` in exact arithmetic (`0.0448 + 0.2856 + 0.3001 + 0.2363 + 0.1333`). The `with_betas()` assertion uses `(sum - 1.0).abs() < 1e-4`, so passing these exact same values through `with_betas()` could panic (0.0001 is not < 0.0001).

Either loosen the tolerance (e.g. `1e-3`), or nudge the last beta to `0.1332` so the sum is exactly `1.0000`.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+/// ```rust,ignore
+/// // Create input for RGB images
+/// let preds: Tensor<B, 4> = /* tensor */;
+/// let targets: Tensor<B, 4> = /* tensor */;
+/// let input = MsSsimInput::new(outputs, targets);
+/// ```
+pub struct MsSsimInput<B: Backend> {
+    /// Model predictions with shape [N, C, H, W].
+    preds: Tensor<B, 4>,
+    /// Ground truth targets with shape [N, C, H, W].
```
antimora, line 40: Naming inconsistency: the existing `SsimInput` in this same module uses `outputs` / `targets`, but this uses `preds` / `targets`. For users who work with both metrics, having to remember different names for the same concept is a friction point. Consider aligning with the existing convention (`outputs`).

Similarly, `SsimMetricConfig` uses `data_range` and `window_size`, while this uses `pixel_range` and `kernel_size`. Consistent naming across the module would make the API more ergonomic.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+    }
+}
+
+/// Configuration for the [MsMsSsimMetric].
+#[derive(Debug, Clone)]
+pub struct MsSsimMetricConfig {
+    /// A parameter of SSIM used to stabilize the luminance comparison.
+    /// Default is 0.01.
+    pub k1: f32,
+    /// A parameter of SSIM used to stabilize the contrast comparison.
```
antimora, line 73: `SsimMetricConfig` uses `f64` for `k1`, `k2`, `data_range`, and `sigma`. This config uses `f32` for the same parameters. This creates a type mismatch when users configure both metrics side by side. Consider using `f64` to stay consistent.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+impl<B: Backend> Metric for MsSsimMetric<B> {
+    type Input = MsSsimInput<B>;
+
+    fn name(&self) -> MetricName {
+        self.name.clone()
+    }
+
+    fn update(&mut self, item: &Self::Input, _metadata: &MetricMetadata) -> SerializedEntry {
+        let dims = item.preds.dims();
+        let scales = self.config.betas.len();
```
antimora, line 391: Consider adding a runtime check that the input channels match the configured `channels`:
```rust
assert_eq!(
    dims[1], self.config.channels,
    "Input has {} channels but metric was configured for {}",
    dims[1], self.config.channels
);
```
Unlike `SsimMetric` which infers channels from the input, this metric pre-computes kernels for a fixed channel count. A mismatch would produce a confusing conv error deep in the backend.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+        self.state.current_value()
+    }
+
+    fn running_value(&self) -> NumericEntry {
+        self.state.running_value()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
antimora, line 511: Good test coverage overall. A couple of gaps compared to the existing SSIM test suite:
- No `#[should_panic]` tests for negative/zero `k1`, `k2`, or `sigma` (SSIM has `test_ssim_negative_k1`, `test_ssim_negative_k2`, `test_ssim_negative_sigma`, `test_ssim_zero_sigma`)
- No symmetry test (`MS-SSIM(x, y) == MS-SSIM(y, x)`) or value-range test

Not blockers, but worth adding for parity.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+impl<B: Backend> MsSsimMetric<B> {
+    /// Creates a new MS-SSIM metric with the given configuration.
+    ///
+    /// # Arguments
+    /// - `config`: Configuration for the metric (data range, kernel size, etc.)
+    /// - `device`: Device to place the Gaussian kernels on
+    ///
+    /// # Note
+    /// The default metric name format is "MS-SSIM (pr={}, k={}, σ={})"
+    /// where dr is the pixel range, w is the kernel size, and σ is the
```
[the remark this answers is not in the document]
antimora: Agree. Good catch - I missed this one.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+    /// - If `betas` is empty.
+    pub fn with_betas(mut self, betas: Vec<f32>) -> Self {
+        assert!(!betas.is_empty(), "betas vector cannot be empty");
+
+        let sum: f32 = betas.iter().sum();
+        assert!(
+            (sum - 1.0).abs() < 1e-4,
+            "The sum of the betas must be 1.0, but got {}",
+            sum
+        );
```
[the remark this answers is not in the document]
antimora: Agree. Negative betas passed to `powf_scalar` could produce NaN or unexpected values. Worth guarding against.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+/// # Example
+/// ```rust,ignore
+/// // Create input for RGB images
+/// let preds: Tensor<B, 4> = /* tensor */;
+/// let targets: Tensor<B, 4> = /* tensor */;
+/// let input = MsSsimInput::new(outputs, targets);
+/// ```
+pub struct MsSsimInput<B: Backend> {
+    /// Model predictions with shape [N, C, H, W].
+    preds: Tensor<B, 4>,
```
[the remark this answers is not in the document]
antimora: Agree, same point as my comment above.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+        Self { preds, targets }
+    }
+}
+
+/// Configuration for the [MsMsSsimMetric].
+#[derive(Debug, Clone)]
+pub struct MsSsimMetricConfig {
+    /// A parameter of SSIM used to stabilize the luminance comparison.
+    /// Default is 0.01.
+    pub k1: f32,
```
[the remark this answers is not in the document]
antimora: Agree. This applies to all the config fields (k1, k2, pixel_range, sigma, betas) - same point as my comment above.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+    /// where `max_pixel_val` is the maximum possible pixel value and `min_pixel_val`
+    /// is the minimum possible pixel value.
+    ///
+    /// - For normalized images in range [0, 1], it should be set to `1.0 - 0.0 = 1.0`
+    /// - For normalized images in range [-1, 1], it should be set to `1.0 - (-1.0) = 2.0`
+    /// - For 8-bit images in range [0, 255], it should be set to `255.0 - 0.0 = 255.0`
+    pub pixel_range: f32,
+    /// The MS-SSIM metric involves applying convolution to the input tensors using a Gaussian kernel.
+    /// This is the kernel/kernel size of the Gaussian kernel. Default is 11.
+    pub kernel_size: usize,
```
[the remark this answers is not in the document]
antimora: Agree.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+    pub k2: f32,
+    /// The range of the pixel values in images which can be computed as following:
+    /// `let pixel_range = max_pixel_val - min_pixel_val;`
+    /// where `max_pixel_val` is the maximum possible pixel value and `min_pixel_val`
+    /// is the minimum possible pixel value.
+    ///
+    /// - For normalized images in range [0, 1], it should be set to `1.0 - 0.0 = 1.0`
+    /// - For normalized images in range [-1, 1], it should be set to `1.0 - (-1.0) = 2.0`
+    /// - For 8-bit images in range [0, 255], it should be set to `255.0 - 0.0 = 255.0`
+    pub pixel_range: f32,
```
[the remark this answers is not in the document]
antimora: Agree.

crates/burn-train/src/metric/vision/ms_ssim.rs
softmaximalist: The reason I changed the field names here is because I personally think the names `kernel_size` and `pixel_range` are more straightforward and descriptive. Same reasoning for the `preds` field. Since I did not want to modify the field names in other metrics in this PR, I left them other metrics unchanged. 

Can I keep using `kernel_size` and `pixel_range` and modify the corresponding fields in SsimInput as well? 

Regarding the `preds`, I personally find `preds` to be more descriptive. Should I modify the all the other metrics to use the name `preds` or should I update ms-ssim to use `outputs` just like the other metrics?

crates/burn-train/src/metric/vision/ms_ssim.rs
softmaximalist: Nice catch! These values are from the original paper and torchmetrics also use these exact values. Should I update the last beta to 0.1332 or should I loosen the tolerance?

crates/burn-train/src/metric/vision/ms_ssim.rs
softmaximalist: The SSIM metric PR was one of my earliest/oldest one so I just used `f64`. However, after working on the repo and other PRs, I noticed that float tensors generally use the `f32` type by default. Hence, I thought it would be better to use `f32` for these fields since they are actually used to perform tensor operations on the input tensors. 

Would it be better to keep using `f32` instead of `f64`? If so, can/should I modify SSIM to use `f32` as well?

crates/burn-train/src/metric/vision/ms_ssim.rs
laggui, line 220: Is this an actual constraint? Looking at reference implementations I haven't seen this anywhere else. I think this only holds for the defaults, right?

If so, the check could likely be removed.

crates/burn-train/src/metric/vision/ms_ssim.rs
```diff
@@ -0,0 +1,796 @@
...
+            let var_y = (mu_of_y_squared - square_of_mu_y.clone()).clamp_min(0.0);
+
+            // Cov(X, Y) = E(XY) - E(X)E(Y)
+            let mu_of_xy = self.gaussian_separable_conv(x.clone() * y.clone());
+            let cov_xy = mu_of_xy - (mu_x.clone() * mu_y.clone());
+
+            // Compute cs_map = (2σxy + C2) / (σx² + σy² + C2)
+            // This is mathematically equivalent to c(x,y) * s(x,y) when C3 = C2 / 2
+            let contrast_structure = (cov_xy * 2.0 + c2) / (var_x + var_y + c2);
+            let beta_j = betas[i];
```
laggui: Could iterate over `betas` instead and use `enumerate()` if you need an index too.

crates/burn-train/src/metric/vision/ms_ssim.rs
softmaximalist: I got it from section 3.1 of the [original paper](https://www.cns.nyu.edu/pub/eero/wang03b.pdf):

ms_ssim

softmaximalist:
@antimora I have replied to some of your comments to explain my reasoning regarding the choice of the field names and float type (f32 vs f64). Let me know what you think.

crates/burn-train/src/metric/vision/ms_ssim.rs
laggui: Ahhh thanks for the pointers!

This reads more as a convention for comparability, not a mathematical requirement for the metric to be valid though.

Based on the formula I think it might still be valid, and someone might have legitimate reasons to use unnormalized betas. But I am not familiar enough with the metric and its application to take a hard stance on that :smile:

It would be good to document that `with_betas` expects *normalized* values though.

softmaximalist: Good point. For some reason, my mind was treating the weights here like weights in a weighted average. I will clearly document that the betas should be normalized for comparability but it can sum up to any positive number otherwise.

crates/burn-train/src/metric/vision/ms_ssim.rs
antimora: I agree that `kernel_size` and `pixel_range` are more descriptive. Updating the existing SSIM metric to match sounds good - better to improve naming across the board than force worse names on new code.

For `preds` vs `outputs`, I would lean toward keeping `outputs`. It is used across multiple metrics (`SsimInput`, `PsnrInput`) and is more general - covers cases where the tensor is not strictly a "prediction" (e.g., image reconstruction, super-resolution). Changing all existing metrics for this is higher churn for marginal benefit. But either way is fine as long as it is consistent across the module.

crates/burn-train/src/metric/vision/ms_ssim.rs
antimora: I would loosen the tolerance to `1e-3`. Keeping the original paper values is better for reproducibility - people reading the code can match them to the Wang et al. paper directly. The tolerance is just a sanity check for user-provided values, not a precision requirement, and `1e-3` still catches clearly wrong inputs (e.g., betas summing to 0.5).

crates/burn-train/src/metric/vision/ms_ssim.rs
antimora: Fair point. These config values end up in tensor operations (`c1`, `c2` computations, `powf_scalar`, etc.) which operate in f32. Using f64 in config then truncating to f32 at computation time is pointless precision. For values like `k1=0.01` and `k2=0.03`, f32 is more than enough.

I would say keep f32 here and update SSIM to f32 in a separate PR.

softmaximalist:
All the comments have been addressed.

PR #4556, opened 2026-02-21
Implement HardShrink, SoftShrink and Shrink Activations

aditya0by0:
### Checklist

### Related Issues/PRs

These activations are needed to implement Shrink ONNX operator https://github.com/tracel-ai/burn-onnx/issues/158

### Changes

Soft and Hard Shrink Activations

### Testing

- test_soft_shrink_forward
- test_soft_shrink_with_lambd_and_bias
- display
- test_hard_shrink_forward
- test_hard_shrink_forward_with_lambd
- display


aditya0by0:
@antimora, Please review.

crates/burn-nn/src/activation/hardshrink.rs
```diff
@@ -0,0 +1,100 @@
...
+    use crate::TestBackend;
+    use burn::tensor::TensorData;
+
+    #[test]
+    fn test_hard_shrink_forward() {
+        let device = <TestBackend as Backend>::Device::default();
+        let model: HardShrink = HardShrinkConfig::new().init();
+        let input = Tensor::<TestBackend, 2>::from_floats(
+            TensorData::from([[0.5, -0.5, -1.0], [8.0, 0.3, 0.0]]),
+            &device,
```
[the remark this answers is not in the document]
laggui: Also, when using `from_data` you can pass the array without creating a `TensorData::from(...)`

crates/burn-nn/src/activation/hardshrink.rs
```diff
@@ -0,0 +1,100 @@
...
+        let expected = TensorData::from([[0.0_f32, 0.0, -1.0], [8.0, 0.0, 0.0]]);
+        assert_eq!(out.into_data(), expected);
+    }
+
+    #[test]
+    fn test_hard_shrink_with_lambd() {
+        let device = <TestBackend as Backend>::Device::default();
+        let model: HardShrink = HardShrinkConfig::new().with_lambd(0.2).init();
+        let input = Tensor::<TestBackend, 2>::from_floats(
+            TensorData::from([[0.1, -0.1, -0.3], [0.5, 0.1, 0.0]]),
```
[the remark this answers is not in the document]
laggui: Same

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -560,3 +560,71 @@ $$
...
+/// Applies the HardShrink function element-wise.
+///
+#[cfg_attr(
+    doc,
+    doc = r#"
+$$
+\text{hard\_shrink}(x) =
+ \begin{cases}
+     x & \text{if } x > \lambd \newline
+     x & \text{if } x < -\lambd \newline
```
antimora: `\lambd` isn't a valid LaTeX command, this'll render as garbage in the docs. Should be `\lambda`. Same issue in the `soft_shrink` docs below where `\bias` should be `\text{bias}` (or just `b`).

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -560,3 +560,71 @@ $$
...
+    doc = "`hard_shrink(x) = x if x > lambd, x if x < -lambd, 0 otherwise`"
+)]
+/// # Arguments
+/// - `lambd`: the lambd value for the Hard Shrink formulation. Default is 0.5.
+pub fn hard_shrink<const D: usize, B: Backend>(tensor: Tensor<B, D>, lambd: f64) -> Tensor<B, D> {
+    let mask = tensor.clone().abs().lower_equal_elem(lambd);
+    tensor.mask_fill(mask, 0)
+}
+
+// Applies the SoftShrink function element-wise.
```
antimora: This needs to be `///` (doc comment), not `//`. As-is the first line won't show up in the rendered documentation.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -560,3 +560,71 @@ $$
...
+    // [0.7, -0.7, 0.3, -0.3, 0.5, -0.5], lambda=0.5, bias=0.3
+    // abs_tensor = [0.7, 0.7, 0.3, 0.3, 0.5, 0.5]
+    let abs_tensor = tensor.clone().abs();
+    // sign = [1, -1, 1, -1, 1, -1]
+    let sign = tensor.clone().sign();
+    // first element: 0.7 - 1*0.3 = 0.4
+    // second element: -0.7 - (-1)*0.3 = -0.4
+    // shrunk = [0.4, -0.4, 0, 0, 0.2, -0.2]
+    let shrunk = tensor.clone().sub(sign.mul_scalar(bias));
+    // mask = [false, false, true, true, true, true]
```
antimora: Two things here:

1. The `tensor.clone()` on line 626 is unnecessary since `tensor` isn't used after this point - you can just consume it directly with `tensor.sub(sign.mul_scalar(bias))`.

2. These step-by-step worked-example comments feel more like debug/development notes. None of the other activation functions in this file have this style of commentary, and the doc comment above already describes the behavior. I'd remove them.

crates/burn-nn/src/activation/softshrink.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// Configuration to create a [SoftShrink](SoftShrink) layer using the [init function](SoftShrinkConfig::init).
+#[derive(Config, Debug)]
+pub struct SoftShrinkConfig {
+    /// The lambd value for the Soft Shrink formulation. Default is 0.5
+    #[config(default = "0.5")]
+    pub lambd: f64,
+    /// The bias value for the Soft Shrink formulation. Default is equal to lambd.
+    #[config(default = "0.5")]
```
antimora: The doc on line 32 says "Default is equal to lambd" but it's actually hardcoded to `0.5` - if someone sets `lambd = 0.3`, bias will still be `0.5` unless they explicitly change it. Should just say "Default is 0.5" to avoid the confusion.

crates/burn-nn/src/activation/mod.rs
```diff
@@ -32,12 +32,14 @@ pub(crate) mod gelu;
 pub(crate) mod glu;
 pub(crate) mod hard_sigmoid;
 pub(crate) mod hard_swish;
+pub(crate) mod hardshrink;
```
antimora: Nit: the existing two-word activation files in this directory use underscores (`hard_sigmoid.rs`, `hard_swish.rs`, `leaky_relu.rs`). For consistency these should probably be `hard_shrink.rs` and `soft_shrink.rs`.

crates/burn-nn/src/activation/hardshrink.rs
```diff
@@ -0,0 +1,100 @@
...
+///
+/// Applies the Hard Shrink function element-wise:
+/// `hard_shrink(x) = x if |x| > lambd else 0`
+///
+/// Should be created with [HardShrinkConfig](HardShrinkConfig).
+#[derive(Module, Clone, Debug)]
+#[module(custom_display)]
+pub struct HardShrink {
+    /// The lambd value for the Hard Shrink formulation.
+    pub lambd: f64,
```
laggui: Pretty sure this is a language historical artifact from pytorch, where `lambda` is a reserved python keyword 😄 

We can use `lambda` as the field name, as well as in the docs.

crates/burn-nn/src/activation/softshrink.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// Applies the Soft Shrink function element-wise:
+/// `soft_shrink(x) = x - bias if x > lambd, x + bias if x < -lambd, 0 otherwise`
+///
+/// Should be created with [SoftShrinkConfig](SoftShrinkConfig).
+#[derive(Module, Clone, Debug)]
+#[module(custom_display)]
+pub struct SoftShrink {
+    /// The lambd value for the Soft Shrink formulation.
+    pub lambd: f64,
```
laggui: Same comment for `lambda`

crates/burn-tensor/src/tensor/activation/base.rs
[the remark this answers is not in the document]
laggui: The bot is right 👍

crates/burn-nn/src/activation/softshrink.rs
```diff
@@ -0,0 +1,118 @@
...
+    use crate::TestBackend;
+    use burn::tensor::TensorData;
+
+    #[test]
+    fn test_soft_shrink_forward() {
+        let device = <TestBackend as Backend>::Device::default();
+        let model: SoftShrink = SoftShrinkConfig::new().init();
+        let input = Tensor::<TestBackend, 2>::from_floats(
+            TensorData::from([[0.5, -0.5, -1.0], [8.0, 0.3, 0.0]]),
+            &device,
```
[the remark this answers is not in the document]
antimora: not required.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -560,3 +560,71 @@ $$
...
+    not(doc),
+    doc = "`soft_shrink(x) = x - bias if x > lambd, x + bias if x < -lambd, 0 otherwise`"
+)]
+/// # Arguments
+/// - `lambd`: the lambd value for the Soft Shrink formulation. Default is 0.5.
+/// - `bias`: the bias value for the Soft Shrink formulation. Usually bias is equal to lambd.
+pub fn soft_shrink<const D: usize, B: Backend>(
+    tensor: Tensor<B, D>,
+    lambd: f64,
+    bias: f64,
```
laggui, line 640: Don't think a separate bias is required for soft shrink (see main review comment)?

laggui:
@antimora looks like I started the review at the same time :smile: comments are pretty similar, except for the `soft_shrink` vs general `shrink` behavior. Lmk if my assessment is correct here.

crates/burn-nn/src/activation/hardshrink.rs
aditya0by0: Sure will switch to `lambda`, had decided to keep it that way because even [onnx ](https://onnx.ai/onnx/operators/onnx__Shrink.html) used the same convention.

laggui: Yeah totally understandable! Seems they carried that over lol

aditya0by0:
> Thanks for contributing!
> 
> If I understand correctly, I think ONNX `Shrink` is just a generalized implementation that handles both soft and hard shrink.
> 
> For `bias = 0` it acts like HardShrink, and `bias = lambda`, it is like SoftShrink.
> 
> So if we add both `hard_shrink` and `soft_shrink`, we don't need soft shrink to have a separate bias no?
> 
> And for any other bias values, we could have a general shrink (reused in `soft_shrink` possibly).

If I understand it correctly, we agree to add **another** ActivationConfig/Activation for general `Shrink` operation which internally calls `burn::tensor::activation::soft_shrink` with given bias value. 

For Soft shink, we remove the `bias` parameter and the  `burn::tensor::activation::soft_shrink` will be called with `bias=lambda`

antimora:
@aditya0by0 Yeah that's exactly right. To spell it out:

- **`SoftShrink`** (nn module): just `lambda`, no separate `bias`. Calls `soft_shrink(input, lambda, lambda)` under the hood.
- **`HardShrink`** (nn module): stays as-is with just `lambda`.
- **`Shrink`** (nn module, new): has both `lambda` and `bias` params for the general ONNX case. Calls `soft_shrink(input, lambda, bias)` internally. This covers the arbitrary bias values that don't map to either hard or soft shrink.

At the tensor activation level, `soft_shrink(tensor, lambda, bias)` stays as the general implementation (maybe rename to just `shrink`?). `hard_shrink` can stay separate since its implementation is simpler (no sign/subtraction needed).

laggui:
Yep, agreed with the above.

> At the tensor activation level, `soft_shrink(tensor, lambda, bias)` stays as the general implementation (maybe rename to just `shrink`?).

Yeah we should probably rename to just `shrink(tensor, lambda, bias)`, and to match the module struct we could have a shorthand `soft_shrink(tensor, lambda)` that just calls `shrink(tensor, lambda, lambda)` essentially?

burn-book/src/building-blocks/module.md
```diff
@@ -227,29 +227,32 @@ Burn comes with built-in modules that you can use to build your own modules.
...
+| `InstanceNorm`    | `nn.InstanceNorm1d`, `nn.InstanceNorm2d` etc. |
+| `LayerNorm`       | `nn.LayerNorm`                                |
+| `LeakyRelu`       | `nn.LeakyReLU`                                |
+| `Linear`          | `nn.Linear`                                   |
+| `Prelu`           | `nn.PReLu`                                    |
+| `Relu`            | `nn.ReLU`                                     |
+| `Selu`            | `nn.SELU`                                     |
+| `SoftShrink`      | `nn.Softshrink`                               |
+| `Softsign`        | `nn.Softsign`                                 |
+| `Shrink`          | _No direct equivalent_                        |
```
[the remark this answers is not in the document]
aditya0by0: I think the activations which don't have a pytorch equivalent goes in the bottom irrespective of alphabetical order.

PR #4557, opened 2026-02-22
Fix(lpips): load ImageNet backbone weights for pretrained models

koreaygj:
### Summary

- Add `TAR` format support to `burn-store` for legacy `PyTorch` models (`AlexNet`, `SqueezeNet`)
- Fix `LPIPS` to load `ImageNet` backbone weights separately from `LPIPS` linear weights
- Fix key remapping to match `PyTorch`'s original model structure

---

### Checklist

---

### Related Issues/PRs

- Fixes weight loading issues introduced in #4403
- Related to #4312 (Image quality metrics)

---

### Changes

`LPIPS` pretrained weights were not loading correctly. The original implementation only loaded `LPIPS` linear layer weights but not the `ImageNet` backbone weights (`VGG16`/`AlexNet`/`SqueezeNet`). Additionally, `AlexNet` and `SqueezeNet` backbone weights are stored in `TAR` format which was not supported.

---

### Solution

1. **`burn-store`**: Added `TAR` format support for legacy `PyTorch` models (pre-1.6)
   - Implemented `TarSource` in `lazy_data.rs`
   - Added `TAR` file detection and loading in `reader.rs`
2. **`burn-train/lpips`**: Fixed pretrained weights loading
   - Now loads both `ImageNet` backbone weights and `LPIPS` linear layer weights separately
   - Fixed key remapping to match `PyTorch`'s original model structure (`features.X` instead of `net.sliceX`)
   - Added backbone weight `URLs` from `PyTorch` official repository

---

### Testing

```bash
# LPIPS pretrained tests (all pass)
cargo test -p burn-train --lib --features vision test_lpips_pretrained

# Results:
# - VGG: 0.4112574 ✅
# - AlexNet: 0.37245688 ✅
# - SqueezeNet: 0.1753163 ✅

# burn-store tests
cargo test -p burn-store --features pytorch

### DIFF
--- a/crates/burn-store/src/pytorch/lazy_data.rs
+++ b/crates/burn-store/src/pytorch/lazy_data.rs
@@ -322,10 +322,7 @@ impl TarSource {
                             other => {
                                 return Err(std::io::Error::new(
                                     std::io::ErrorKind::InvalidData,
-                                    format!(
-                                        "Expected Class for storage type, got {:?}",
-                                        other
-                                    ),
+                                    format!("Expected Class for storage type, got {:?}", other),
                                 ));
                             }
                         };
@@ -351,9 +348,8 @@ impl TarSource {
                 pos += 8;
 
                 // Determine element size from storage type
-                let element_size = storage_type_to_element_size(&storage_type).map_err(|e| {
-                    std::io::Error::new(std::io::ErrorKind::InvalidData, e)
-                })?;
+                let element_size = storage_type_to_element_size(&storage_type)
+                    .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
 
                 let data_size = num_elements * element_size;
 
@@ -391,7 +387,12 @@ impl TarSource {
     }
 
     /// Read a range of data for a specific storage key (avoids double allocation)
-    pub fn read_file_range(&self, key: &str, offset: usize, length: usize) -> std::io::Result<Vec<u8>> {
+    pub fn read_file_range(
+        &self,
+        key: &str,
+        offset: usize,
+        length: usize,
+    ) -> std::io::Result<Vec<u8>> {
         let storage_key = key.split('/').next_back().unwrap_or(key);
 
         if let Some(&(storage_offset, storage_size)) = self.storage_map.get(storage_key)
--- a/crates/burn-store/src/pytorch/pickle_reader.rs
+++ b/crates/burn-store/src/pytorch/pickle_reader.rs
@@ -384,7 +384,13 @@ fn rebuild_tensor(
     };
     let shape = parse_shape_arg(&args[2], "rebuild_tensor")?;
 
-    rebuild_tensor_impl(storage_info, storage_tuple, storage_offset, shape, data_source)
+    rebuild_tensor_impl(
+        storage_info,
+        storage_tuple,
+        storage_offset,
+        shape,
+        data_source,
+    )
 }
 
 /// Modern _rebuild_tensor_v2 function for PyTorch >= 1.6.
@@ -420,7 +426,13 @@ fn rebuild_tensor_v2(
     // args[4] is requires_grad (unused)
     // args[5] is backward_hooks (unused)
 
-    rebuild_tensor_impl(storage_info, storage_tuple, storage_offset, shape, data_source)
+    rebuild_tensor_impl(
+        storage_info,
+        storage_tuple,
+        storage_offset,
+        shape,
+        data_source,
+    )
 }
 
 /// Helper to convert storage type name to DType.

### CODE REVIEW

**Comment by koreaygj** [comment]
Currently, `burn-train --features vision` tests are not included in CI, so LPIPS pretrained test failures were not caught. Adding `vision feature tests` to CI should be considered in a follow-up PR.

File: crates/burn-store/src/pytorch/lazy_data.rs [hunk 1]
```diff
@@ -28,35 +30,31 @@ pub struct ZipSource {
...
-///      c. Infer boundaries from the gaps between storages
-///    - However, the TensorSnapshot abstraction hides storage keys in closures
-///    - This would require deep modifications to the pickle parsing logic
+/// Older PyTorch/torchvision models (pre-1.6) use TAR format instead of ZIP.
+/// The TAR archive contains:
+/// - `sys_info`: System info pickle (endianness, type sizes)
+/// - `pickle`: OrderedDict mapping tensor names to storage keys
+/// - `tensors`: Tensor metadata pickles (unused, metadata is embedded in pickle)
+/// - `storages`: Storage count + sequential (metadata pickle, element count, raw data)
+pub struct TarSource {
```
laggui, line 41: @antimora you specifically introduced support for legacy pytorch format.. do you know about another legacy format? 😅

koreaygj: Well known formats(`zip` `tar` `legacy pickle` `simple pickle`) are already implemented.
Other legacy format like `torchscripts`, `big-endian` and`full model save` exist but are rarely used in practice. 😄

koreaygj:
> > Currently, `burn-train --features vision` tests are not included in CI, so LPIPS pretrained test failures were not caught. Adding `vision feature tests` to CI should be considered in a follow-up PR.
> 
> Well that explains why it wasn't caught 😅 thanks

Current CI `xtask/src/commands/test.rs` includes spesial feature tests for `burn-dataset`, `burn-core` and `burn-vision`, but `burn-train` is not included.

If you accept to change it, I will open PR.
I can add`burn-train` vision tests in `xtask/src/commands/test.rs`
like below!

```rust
  // burn-train vision
  helpers::custom_crates_tests(
      vec!["burn-train"],
      handle_test_args(&["--features", "vision"],
  args.release),
      None,
      None,
      "std vision",
  )?;
```

crates/burn-store/src/pytorch/lazy_data.rs
laggui: Sorry I think my message might have been unclear. I was referring to this "new" legacy format (`TarSource`) that you just added for support.

laggui:
> I can add`burn-train` vision tests in `xtask/src/commands/test.rs`

Yeah we should do that! Can even be part of this PR if you want since it's related.

crates/burn-store/src/pytorch/lazy_data.rs
koreaygj: My mistake.! I thought the question was for me😅

laggui: No worries :+1:

crates/burn-store/src/pytorch/lazy_data.rs
```diff
@@ -271,6 +271,140 @@ impl LegacyMultiStorageSource {
...
+                        8
+                    } else if storage_type.contains("Half") || storage_type.contains("Short") {
+                        2
+                    } else if storage_type.contains("Byte")
+                        || storage_type.contains("Char")
+                        || storage_type.contains("Bool")
+                    {
+                        1
+                    } else {
+                        4 // Default to float (4 bytes)
```
antimora: Bug: `"BFloat16Storage".contains("Half")` returns `false`, so BFloat16 falls through to `4`. BFloat16 is 2 bytes. This will silently compute wrong offsets and corrupt tensor data.

The `rebuild_tensor` / `rebuild_tensor_v2` code has the correct explicit mapping (`"BFloat16Storage" => DType::BF16`). Consider extracting a shared `storage_type_to_dtype(name) -> DType` helper and using `dtype.size()` here instead of a separate if/else chain with `contains` checks.

crates/burn-store/src/pytorch/lazy_data.rs
```diff
@@ -271,6 +271,140 @@ impl LegacyMultiStorageSource {
...
 
+impl TarSource {
+    /// Create a new TAR source by parsing storages data.
+    ///
+    /// # Arguments
+    /// * `_tensors_data` - Unused; tensor metadata is embedded in the pickle entry
+    /// * `storages_data` - Raw storages blob with structure:
+    ///   - Count pickle (number of storages)
+    ///   - For each storage: metadata pickle + u64 num_elements + raw binary data
+    pub fn new(_tensors_data: &[u8], storages_data: Vec<u8>) -> std::io::Result<Self> {
```
antimora: `_tensors_data` is never used. It propagates up to `LazyDataSource::from_tar` and the callsite in `reader.rs`, making the API misleading. Remove it.

crates/burn-store/src/pytorch/lazy_data.rs
```diff
@@ -319,6 +467,15 @@ impl LazyDataSource {
...
                 source.read_file_range(key, offset, length)
             }
+            Self::Tar(source) => {
+                // For TAR format, read the full data and slice it
+                let source = source
+                    .lock()
+                    .unwrap_or_else(|poisoned| poisoned.into_inner());
+                let data = source.read(key)?;
+                let end = (offset + length).min(data.len());
+                Ok(data[offset..end].to_vec())
```
antimora: Double allocation: `source.read(key)` allocates a `Vec<u8>` for the full storage, then this line allocates again for the slice. Since `TarSource` already holds the full blob in memory (`storages_data`), add a `read_range` method on `TarSource` that slices directly from `self.storages_data[storage_offset + offset..]`.

crates/burn-store/src/pytorch/lazy_data.rs
```diff
@@ -271,6 +271,140 @@ impl LegacyMultiStorageSource {
...
+                        let key = match &tuple[0] {
+                            super::pickle_reader::Object::Int(i) => i.to_string(),
+                            super::pickle_reader::Object::String(s) => s.clone(),
+                            _ => continue,
+                        };
+                        // tuple[1] is device (e.g., "cpu")
+                        // tuple[2] is storage type class
+                        let stype = match &tuple[2] {
+                            super::pickle_reader::Object::Class { name, .. } => name.clone(),
+                            _ => "FloatStorage".to_string(),
```
antimora: Silently defaulting to `"FloatStorage"` when the storage type is not a `Class` will mask real parsing errors. Return an error instead.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -301,6 +305,448 @@ fn rebuild_parameter(
     Ok(tensor)
 }
 
+/// Legacy _rebuild_tensor function for PyTorch < 1.6.
+/// Same as rebuild_tensor_v2 but with fewer arguments: (storage, storage_offset, size, stride)
+fn rebuild_tensor(
```
antimora, line 359: This entire function (~440 lines, through line 748) is a near-exact copy of `rebuild_tensor_v2` below. The only difference is 4 args vs 5 args (v2 adds `requires_grad` and `backward_hooks`).

The TODO at line 936 acknowledges this. Please extract the shared logic: parse the storage args into a struct, build the closure once. Both functions should be thin wrappers. ~400 lines of duplication makes this harder to maintain and will cause bugs when one copy gets updated without the other.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -301,6 +305,448 @@ fn rebuild_parameter(
...
+                "DoubleStorage" => DType::F64,
+                "HalfStorage" => DType::F16,
+                "BFloat16Storage" => DType::BF16,
+                "LongStorage" => DType::I64,
+                "IntStorage" => DType::I32,
+                "ShortStorage" => DType::I16,
+                "CharStorage" => DType::I8,
+                "ByteStorage" => DType::U8,
+                "BoolStorage" => DType::Bool,
+                _ => DType::F32, // Default to F32
```
antimora: Defaulting to `DType::F32` for an unknown storage type will silently load garbage data. This should return an error. Same issue at lines 439, 449, and the duplicated copies at 836, 879, 889.

(Once you deduplicate `rebuild_tensor` / `rebuild_tensor_v2`, there will only be one place to fix.)

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -301,6 +305,448 @@ fn rebuild_parameter(
...
+    let (dtype, storage_key) = if let Some(tuple) = storage_tuple {
+        // Direct tuple access
+        if tuple.len() >= 3 {
+            let storage_type = match &tuple[1] {
+                Object::String(s) => s.as_str(),
+                Object::Class {
+                    module_name: _,
+                    name,
+                } => name.as_str(),
+                _ => "FloatStorage",
```
antimora: Same concern: defaulting to `"FloatStorage"` when the object type is unexpected will silently produce wrong results. Return an error for unexpected types.

crates/burn-store/src/pytorch/reader.rs
```diff
@@ -792,6 +818,137 @@ fn load_legacy_pytorch_file_with_metadata(
     Ok((tensors, metadata))
```
antimora, line 818: This is the biggest gap: there are **no integration tests** for the TAR format. The existing test suite in `tests/reader/mod.rs` has 47 tests covering ZIP and legacy formats with Python-generated fixtures. TAR needs the same treatment.

Since modern `torch.save()` cannot produce TAR files (this format predates PyTorch 0.1.10), you will need a Python script that manually constructs the TAR archive structure, similar to how `create_legacy_with_offsets.py` works.

At minimum, please add tests for:
- TAR format detection (`is_tar_file()`)
- Loading a float32 tensor from TAR and verifying values
- Loading multiple tensors (weight + bias) with correct shapes
- Loading different dtypes (float32, float64, int64)

crates/burn-train/src/metric/vision/lpips/metric.rs
```diff
@@ -730,11 +732,12 @@ mod tests {
         let image2 = TestTensor::<4>::ones([1, 3, 64, 64], &device);
         let distance = lpips.forward(image1, image2, Reduction::Mean);
         let distance_value = distance.into_data().to_vec::<f32>().unwrap()[0];
+        println!("LPIPS VGG distance (black vs white): {}", distance_value);
```
antimora: Debug `println!` left in test code. Same at lines 787-789, 795-797, 815-817. Also the weight-printing block at lines 783-798 should be removed before merging.

crates/burn-train/src/metric/vision/lpips/squeezenet.rs
```diff
@@ -70,7 +70,7 @@ impl<B: Backend> FireModule<B> {
 #[derive(Module, Debug)]
 pub struct SqueezeFeatureExtractor<B: Backend> {
     /// Conv1: 3 -> 64, kernel 3x3, stride 2
-    conv1: Conv2d<B>,
+    pub conv1: Conv2d<B>,
```
antimora: `conv1` was changed from private to `pub` only to support the debug weight printing in the test. Once that debug block is removed, revert this back to private.

crates/burn-train/src/metric/vision/lpips/weights.rs
```diff
@@ -56,71 +84,153 @@ fn get_cache_dir() -> PathBuf {
...
+    // Then load LPIPS linear layer weights
+    lpips = load_lpips_weights(lpips, net, &lpips_cache_path);
 
     lpips
 }
+
+/// Load ImageNet pretrained backbone weights.
+fn load_backbone_weights<B: Backend>(
+    lpips: Lpips<B>,
+    _net: LpipsNet,
```
antimora: `_net` is unused since you match on the `Lpips` enum variant. Remove it. Same for `load_lpips_weights` at line 189.

crates/burn-store/src/pytorch/lazy_data.rs
koreaygj: Resolved!

crates/burn-store/src/pytorch/lazy_data.rs
koreaygj: Resolved.

crates/burn-store/src/pytorch/lazy_data.rs
koreaygj: Resolved.

crates/burn-store/src/pytorch/lazy_data.rs
koreaygj: Resolved.

crates/burn-store/src/pytorch/pickle_reader.rs
koreaygj: Resolved.

crates/burn-store/src/pytorch/pickle_reader.rs
koreaygj: Resolved.

crates/burn-store/src/pytorch/pickle_reader.rs
koreaygj: Resolved.

crates/burn-train/src/metric/vision/lpips/metric.rs
koreaygj: Removed.

crates/burn-train/src/metric/vision/lpips/squeezenet.rs
koreaygj: Removed.

crates/burn-train/src/metric/vision/lpips/weights.rs
koreaygj: Removed.

crates/burn-store/src/pytorch/reader.rs
koreaygj: Thanks for the suggestion! Added 8 TAR format tests:

  1. `test_tar_format_detection` - TAR file detection
  2. `test_tar_float32_tensor` - float32 tensor loading
  3. `test_tar_float64_tensor` - float64 tensor loading
  4. `test_tar_int64_tensor` - int64 tensor loading
  5. `test_tar_multiple_tensors` - multiple tensors (weight + bias)
  6. `test_tar_multi_dtype` - mixed dtypes
  7. `test_tar_2d_tensor_shape` - 2D shape verification
  8. `test_tar_metadata` - metadata verification

koreaygj:
> Thanks for working on this! TAR format support is a useful addition since a number of older torchvision models (AlexNet, SqueezeNet) on PyTorch Hub are in this format, and the two-step weight loading approach for LPIPS is the right design (separate ImageNet backbone weights + LPIPS linear layer weights).
> 
> That said, there are several issues that need to be addressed before this can be merged. The biggest one: **there are no integration tests for the TAR format itself**. `burn-store` has ~47 test functions covering ZIP and legacy formats with Python-generated fixtures, but zero tests for TAR loading. We need the same treatment here.
> 
> Summary of issues (details in inline comments):
> 
> 1. **Missing integration tests** for TAR format in `burn-store` (critical)
> 2. **~440 lines of copy-pasted code** between `rebuild_tensor` and `rebuild_tensor_v2`
> 3. **BFloat16 bug** in `TarSource` element size mapping (will silently corrupt data)
> 4. **Silent default to F32** in multiple places instead of returning errors for unknown storage types
> 5. **Unused parameter**, wasteful allocation in `read_range`, and debug `println!` left in tests

Thanks for your review!
I fix all your feedback 😊

- [x] Missing integration tests for TAR format in burn-store (critical)
- [x] ~440 lines of copy-pasted code between rebuild_tensor and rebuild_tensor_v2
- [x] BFloat16 bug in TarSource element size mapping (will silently corrupt data)
- [x] Silent default to F32 in multiple places instead of returning errors for unknown storage types
- [x] Unused parameter, wasteful allocation in read_range, and debug println! left in tests

antimora (approved):
Looks good, all the feedback has been addressed. Thanks for the thorough rework!

PR #4572, opened 2026-02-25
Add `#[module(constant)]` and `#[module(skip)]` attributes

laggui:
### Checklist

### Changes

Improved the module derive to handle non-module fields and generics

- Add module attributes to handle non-module / parameter types
  -  `#[module(constant)]`: treated as persistent values (useful for configs, persistent tensor values, etc.)
  - `#[module(skip)]`: treated as runtime state (non-persistent)
- Deprecate `Ignored<T>` fields (in favor of `#[module(skip)]`)
- Deprecated `ConstantRecord`, a misnomer that did not persist any value (empty record), in favor of `EmptyRecord`
- Added `ValueRecord` for constant values

Example:
```rust
#[derive(Module, Debug)]
pub struct ModuleWithAttributes<B: Backend, M: Module<B>> {
    /// A normal parameter.
    weight: Param<Tensor<B, 2>>,
    /// A nested enum module with generics.
    nested: ModuleEnumWithGenericModule<B, M>,
    /// By default, primitives were not persistent (same as `#[module(skip)]`).
    other_prob: f64,
    /// By default, tensors were not persistent and not visited/mapped (same as `#[module(skip)]`).
    tensor: Tensor<B, 1>,
    /// A persistent config value.
    #[module(constant)]
    dropout_prob: f64,
    /// A field that is recomputed at runtime.
    #[module(skip)]
    cached_mask: Option<Tensor<B, 2>>,
    /// A field that contains some debug state.
    #[module(skip)]
    debug_state: String,
}
```

Modules now use `#[module(skip)]` instead of `Ignored<T>`:
```rust
#[derive(Module, Debug)]
#[module(custom_display)]
pub struct Conv1d<B: Backend> {
    // ...
    /// Padding configuration.
    #[module(skip)]
    pub padding: PaddingConfig1d,
}
```

This preserves the previous behavior, but users can now mark configs as `#[module(constant)]` to persist the values.

TODO:
- [Breaking] Tensors and primitives should not be empty records by default, but constants (allow de/serialization)

### Testing

Module derive tests

### DIFF
--- a/crates/burn-candle/src/lib.rs
+++ b/crates/burn-candle/src/lib.rs
@@ -2,7 +2,7 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![allow(unused)] // TODO remove when backend filled
 #![deprecated(
-    since = "0.21.0-pre.1",
+    since = "0.21.0",
     note = "burn-candle is deprecated and will be removed in a future release. Use burn-cubecl (CUDA/ROCm/Vulkan/Metal/WebGPU), burn-ndarray, or burn-tch instead."
 )]
 
--- a/crates/burn-core/src/module/display.rs
+++ b/crates/burn-core/src/module/display.rs
@@ -5,7 +5,7 @@ use alloc::{
     vec::Vec,
 };
 use core::any;
-use core::fmt::{Display, Write};
+use core::fmt::{Debug, Display, Write};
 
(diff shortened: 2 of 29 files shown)


laggui:
Closing in favor of #4618.

We've decided against adding the support for `#[module(constant)]` for now. All other improvements were preserved.

PR #4578, opened 2026-02-27
Add evaluator summary

### LINKED ISSUE #4577: Evaluator does not print out summary
**Describe the bug**
When trying to evaluate my model, I noticed that whatever I do, there is just no summary at the end that gets printed out.

**To Reproduce**
This is my code:
```rust
impl<B: AutodiffBackend> Model<B> {
    pub fn eval(
        &self,
        config: TrainingConfig,
        database: Database,
        dataset: StockDataset,
        artifacts: impl AsRef<Path>,
        device: B::Device,
    ) {
        B::seed(&device, config.seed);

        let batcher = StockBatcher::new(database.clone());

        assert!(!dataset.is_empty(), "Datasets must not be empty");

        let dataloader = DataLoaderBuilder::new(batcher.clone())
            .batch_size(config.batch_size)
            .num_workers(config.num_workers)
            .set_device(device)
            .build(dataset);

        let eval = EvaluatorBuilder::new(artifacts)
            .metric_numeric(SharpeRatioMetrics::default())
            .metric_numeric(PrecisionMetric::multiclass(1, ClassReduction::Macro))
            .metric_numeric(AccuracyMetric::new())
            .metric_numeric(LossMetric::new())
            .summary()
            .build(self.clone());

        eval.eval("Evaluation", dataloader);

        // TODO: this doesn't give a summary?
    }
}
```

**Expected behavior**
A summary is printed out, like training does.

laggui:
### Checklist

### Related Issues/PRs

Fixes #4577

### Changes

The `EvaluatorBuilder` exposed a `.summary()` setter, but the setting was not propagated to the evaluator and, more importantly, there were no provisions to correctly capture the test split metrics (with tags).

This PR implements the (tagged) test metrics collection for the summary. Also changed:
- Test split now captures an optional tag: `Split::Test(Option<Arc<String>>)`
  - Updated all logger methods to take only a `&split` reference instead of `tag: Option<&String>, split: Split`
- Generalized `TuiRendererEvent` for process end since the summary can be passed for both train and test end

### Testing

Enabled `.summary()` for mnist example

### DIFF
--- a/crates/burn-train/src/logger/metric.rs
+++ b/crates/burn-train/src/logger/metric.rs
@@ -244,7 +244,6 @@ impl MetricLogger for FileMetricLogger {
             value.sync()
         }
 
-        // TODO: should also handle tagged metrics
         let file_path = self.file_path(name, Some(epoch), split);
 
         let mut errors = false;


crates/burn-train/src/logger/metric.rs
```diff
@@ -225,27 +230,22 @@ impl MetricLogger for FileMetricLogger {
...
         epoch: usize,
-        split: Split,
+        split: &Split,
     ) -> Result<Vec<NumericEntry>, String> {
         if let Some(value) = self.loggers.get(name) {
             value.sync()
         }
 
-        let file_path = self.file_path(None, name, Some(epoch), split);
+        // TODO: should also handle tagged metrics
```
Charles23R: Doesn't it already handle it? Since the tag is included in the split.

laggui: Yup! I forgot to remove the comment :sweat_smile: thanks

PR #4593, opened 2026-03-02
Refactor/device handle

nathanielsimard:


crates/burn-backend/src/data/tensor.rs
```diff
@@ -21,15 +23,15 @@ pub struct TensorData {
     pub bytes: Bytes,
 
     /// The shape of the tensor.
-    pub shape: Vec<usize>,
+    pub shape: Shape,
```
laggui, line 29: This will break records (de/serialize). Will have to implement an adapted de/serialization (note to self).

crates/burn-cubecl/src/kernel/quantization/quantize.rs
```diff
@@ -13,15 +13,16 @@ where
 {
     let output = empty_qtensor_optimized(tensor.shape(), *scheme, &tensor.device);
     let (out_values, out_params) = output.clone().quantized_handles().unwrap();
```
laggui, line 15: Just fixed the handles in https://github.com/tracel-ai/cubecl/pull/1209/changes/711e23bc74453abdc597c0cd5f7954dfde423410

Quantization tests now passing

crates/burn-backend/src/data/tensor.rs
laggui: Fixed in https://github.com/tracel-ai/burn/pull/4603

PR #4595, opened 2026-03-03
Add Gram Matrix Loss for vision tasks

softmaximalist:
### Checklist

### Related Issues/PRs

Related to [Image quality metrics #4312](https://github.com/tracel-ai/burn/issues/4312)

### Changes

Implemented Gram Matrix Loss. 
- Created a feature flagged `crates/burn-nn/src/loss/pretrained` directory to hold all loss functions that requires pretrained models
- Uses VGG19: downloads pretrained pytorch weights and caches them
- VGG19 model can be configured to use max pooling or average pooling
- Can be configured to use any vector of layer weights (as long as it contains 5 non-negative weights)
- Offers two forward methods (with reduction and without reduction)
- Modified xtask behavior to run all tests locally but to exclude tests that downloads pretrained pytorch weights in GitHub CI

### Testing

| Test Name | Description | Local-Only |
| :--- | :--- | :---: |
| `test_gram_matrix_loss_config_invalid_length` | Verifies that initializing the config with anything other than exactly 5 layer weights triggers a panic. | No |
| `test_gram_matrix_loss_config_negative_weights` | Verifies that initializing the config with any negative layer weights triggers a panic. | No |
| `test_gram_matrix_loss_config_valid_weights` | Checks that valid weights (exactly 5 non-negative floats) are correctly stored in the config. | No |
| `test_gram_matrix_loss_1_channel_panic` | Ensures the loss function panics if the input tensors have 1 channel (e.g., Grayscale) instead of the expected 3. | No |
| `test_gram_matrix_loss_4_channel_panic` | Ensures the loss function panics if the input tensors have 4 channels (e.g., RGBA) instead of the expected 3. | No |
| `test_gram_matrix_loss_zero_for_identical_inputs` | Confirms that the computed loss is essentially `0.0` when the prediction and target images are exactly the same (using randomly initialized VGG19 weights). | No |
| `test_gram_matrix_loss_greater_than_zero_for_different_inputs` | Confirms that the computed loss is strictly positive (`> 0.0`) when the prediction and target images are different (using randomly initialized VGG19 weights). | No |
| `test_gram_matrix_loss_forward_no_reduction_shape` | Validates that `forward_no_reduction` returns a 1D tensor with a length exactly equal to the `batch_size`. | No |
| `test_gram_matrix_loss_reduction_sum_vs_mean` | Checks the math of the reduction types by ensuring that `Reduction::Sum` equals `Reduction::Mean` multiplied by the batch size. | No |
| `test_gram_matrix_loss_with_avg_pool` | Verifies that the loss function executes successfully and returns a positive loss when configured to use Average Pooling instead of Max Pooling. | No |
| `test_gram_matrix_loss_autodiff` | Ensures that gradients are properly calculated for the input `predictions` tensor, and verifies that the VGG19 model parameters are correctly frozen (no gradients computed for them). | No |
| `test_gram_matrix_loss_pretrained_weights_identical_inputs` | Tests that the loss is `0.0` for identical inputs, specifically using the **actual downloaded pretrained VGG19 weights** rather than random initialization. | **Yes** |
| `test_gram_matrix_loss_pretrained_weights_different_inputs` | Tests that the loss is positive for different inputs, specifically using the **actual downloaded pretrained VGG19 weights** rather than random initialization. | **Yes** |

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,143 +1,97 @@
-# Contributing to the Burn Project
+# Contributing to Burn
 
-Welcome to the Burn project! We're thrilled that you're considering contributing to our project.
-Every contribution helps, and we're delighted to have you on board.
+Welcome to the Burn community! We're glad you're interested in contributing.
 
-## How to Contribute?
+## How to Contribute
 
-Here are some steps to guide you through the process of contributing to the Burn project:
+The best way to get started is to look at [open issues](https://github.com/tracel-ai/burn/issues)
+and find one that interests you. Issues labeled `good first issue` are a great starting point for
+new contributors.
 
-### Step 1: Review the Issue Tickets
+If you have an idea that isn't covered by an existing issue, open one first to discuss the approach.
+This helps align expectations and avoids wasted effort on both sides.
 
-Before you start working on a contribution, please take a moment to look through the open issues in
-the [issue tracker](https://github.com/tracel-ai/burn/issues) for this project. This will give you an
-idea of what kind of work is currently being planned or is in progress.
+For questions, discussions, or just to say hello, join us on
+[Discord](https://discord.gg/uPEBbYYDB6). The [Contributor Book](https://burn.dev/contributor-book/)
+covers architecture, environment setup, and guides for common tasks.
 
-### Step 2: Get Familiar with the Project Architecture
+## Pull Requests
 
-It's crucial to have an understanding of the [project's architecture](https://github.com/tracel-ai/burn/tree/main/contributor-book/src/project-architecture). Familiarize
-yourself with the structure of the project, the purpose of different components, and how they
-interact with each other. This will give you the context needed to make meaningful contributions.
+Every pull request should have a descriptive title, a description covering what you changed, why,
+how you tested it, and a link to the relevant issue (if applicable). Prefer small, focused PRs over
+large ones that bundle unrelated changes.
 
-### Step 3: Fork and Clone the Repository
+Draft pull requests are considered not yet ready for review.
 
-Before you can start making changes, you'll need to fork the Burn repository and clone it to your
-local machine. This can be done via the GitHub website or the GitHub Desktop application. Here are
-the steps:
+CI checks should pass before requesting review, though the signal isn't always accurate. If you have
+questions or need early feedback, let us know on the PR or on
+[Discord](https://discord.gg/uPEBbYYDB6).
 
-1. Click the "Fork" button at the top-right of this page to create a copy of this project in your
-   GitHub account.
-2. Clone the repository to your local machine. You can do this by clicking the "Code" button on the
-   GitHub website and copying the URL. Then open a terminal on your local machine and type
-   `git clone [the URL you copied]`.
+### Change Ownership
 
-### Step 4: Create a New Branch
+The core principle behind all contributions: **PR authors must understand, justify, and explain
+every change they propose.** After a PR is accepted, both the reviewer and the author should be
+confident it improves the codebase.
 
-It's a good practice to create a new branch for each contribution you make. This keeps your changes
-organized and separated from the main project, which can make the process of reviewing and merging
-your changes easier. You can create a new branch by using the command
-`git checkout -b [branch-name]`.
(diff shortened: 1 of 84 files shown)


crates/burn-nn/src/loss/pretrained/gram_matrix/gram_matrix_loss.rs
```diff
@@ -0,0 +1,544 @@
...
+        let device = &predictions.device();
+
+        let pred_processed = self.preprocess_input(predictions);
+        let target_processed = self.preprocess_input(targets);
+
+        // Both vectors contain 5 entries since there are 5 layers
+        // Both feature map tensors already have the shape [N, C, H * W]
+        let (pred_features, pred_normalization_factors) =
+            self.feat_extractor.forward(pred_processed);
+        let (target_features, _) = self.feat_extractor.forward(target_processed);
```
laggui: The feature extractor should only extract features. It doesn't need to compute the scale factors.

This should be computed with the loss. Should be easy enough to check the feature map shape there.

crates/burn-nn/src/loss/pretrained/gram_matrix/gram_matrix_loss.rs
```diff
@@ -0,0 +1,544 @@
...
+
+        // Create final loss accumulator tensor
+        let mut final_loss = Tensor::<B, 1>::zeros([batch_size], device);
+
+        // Compute and add the weighted loss for each layer to the final loss tensor.
+        // Note that the loss tensor for each layer and the final loss tensors
+        // contains a loss value for each sample in the batch.
+        for (i, pred_f) in pred_features.iter().enumerate() {
+            let target_f = &target_features[i];
```
laggui: We can zip predicted and target features instead, e.g. 

```rust
for (pred_f, target_f) in features.into_iter().zip(targets) {
    // ...
}
```

And potentially even better: collect each each layer's loss into a vec of tensors, then sum them all at once. So zip + map + collect -> stack + sum_dim

crates/burn-nn/src/loss/pretrained/gram_matrix/gram_matrix_loss.rs
```diff
@@ -0,0 +1,544 @@
...
+/// let device = Default::default();
+/// // Uses max pool by default
+/// let loss_fn = GramMatrixLossConfig::new(vec![1.0, 1.0, 1.0, 1.0, 1.0]).init::<B>(&device);
+/// ```
+///
+/// # Reference
+/// [Image Style Transfer Using Convolutional Neural Networks](https://www.cv-foundation.org/openaccess/content_cvpr_2016/papers/Gatys_Image_Style_Transfer_CVPR_2016_paper.pdf)
+#[cfg_attr(docsrs, doc(cfg(feature = "pretrained")))]
+#[derive(Module, Debug)]
+pub struct GramMatrixLoss<B: Backend> {
```
laggui, line 114: Hmmm I'm a bit unsure if this should be the name.

The style transfer loss is computed via the gram matrix, but it is not *the* loss per se.

Looks like the linked paper refers to this loss as the "style loss" (as opposed to "content loss"). So `StyleLoss` might be a more appropriate name?

softmaximalist: While the loss was originally referred to as the Style Loss (given their context of neural style transfer) in papers published around 2015~2016, recent papers such as [FILM: Frame Interpolation for Large Motion (2022)](https://www.ecva.net/papers/eccv_2022/papers_ECCV/papers/136670244.pdf) and [Sharp Monocular View Synthesis in Less Than a Second (2026)](https://arxiv.org/pdf/2512.10685) started using this loss for tasks other than neural style transfer and are now referring to it as Gram Matrix Loss.

Hence, I chose the name Gram Martrix Loss. Let me know if I should keep this name or change it to Style Loss.

laggui: Ahhh ok, thanks for the references!

In this case, we can keep the `GramMatrixLoss` name.

softmaximalist:
> Implementation LGTM overall.
> 
> Just a couple of comments

All the comments have been addressed.

PR #4601, opened 2026-03-06
feat: Introduce Lanczos3 interpolation method

ovr:
### Checklist

### Related Issues/PRs

There are not issues/PRs for that right now.

### Changes

I am working on research releated to temporal upscalers, and there is no way right now to do lancoz interpolation. So, let's make it possible! I tried to align it as much as possible close to TF/JAX/PIL

### Testing

Tested with python library, just to compare results. https://gist.github.com/ovr/3006835f9a8529ba2af2875b649a5a14


burn-book/src/building-blocks/module.md
```diff
@@ -251,8 +251,8 @@ Burn comes with built-in modules that you can use to build your own modules.
 | `RmsNorm`         | _No direct equivalent_                        |
 | `SwiGlu`          | _No direct equivalent_                        |
 | `ThresholdedRelu` | _No direct equivalent_                        |
-| `Interpolate1d`   | _No direct equivalent_                        |
-| `Interpolate2d`   | _No direct equivalent_                        |
+| `Interpolate1d`   | `nn.Upsample`                                 |
```
ovr: It looks the same https://docs.pytorch.org/docs/stable/generated/torch.nn.Upsample.html

crates/burn-cubecl/src/kernel/interpolate/lanczos3.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    output[out_idx] = result;
+}
+
+#[cube]
+fn lanczos3_weight(x: f32) -> f32 {
+    let abs_x = f32::abs(x);
+    let mut result = 0.0f32;
```
ovr, line 106: result,  because it doesn't support return

ovr:
Burn now matches TF/JAX/PIL with max diff 4.9e-5! after https://github.com/tracel-ai/burn/pull/4601/commits/f7b1aeb4b0b72088f3e202e415ecebc0d91af5ec 👯

burn-book/src/building-blocks/module.md
laggui: I think it was marked as N/A since we have separate interfaces, while upsample handles 1D/2D/3D. But that's fair!

PR #4623, opened 2026-03-11
Implement RNNT loss

cong-or:
Adds RNN Transducer (RNNT) loss to complete the second half of #4459 (CTC was merged in #4529).                                                                                             
                                                                
The implementation computes the forward variable over the (T, U) alignment lattice in log-space, following the algorithm from [Graves 2012](https://arxiv.org/abs/1211.3711). Backward gradients come from Burn's autodiff — no manual gradient code needed.            
                                                                                                                                                                                              
  ### Changes                                                                      

  - `crates/burn-nn/src/loss/rnnt.rs` — `RNNTLossConfig` / `RNNTLoss` with `forward` and `forward_with_reduction`
  - `crates/burn-nn/src/loss/mod.rs` — module registration
  - `burn-book/src/building-blocks/module.md` — added to the Loss table

  ### API

  ```rust
  let rnnt = RNNTLossConfig::new().init(); // blank=0, fused_log_softmax=true

  // logits: [B, T, U+1, V] from joiner network
  let loss = rnnt.forward(logits, targets, logit_lengths, target_lengths);
  ```
  fused_log_softmax controls whether log_softmax is applied internally (like CTC's behavior) or expects pre-computed log-probs.

  Tests

  - Unit tests with hand-derived expected values (uniform probs, empty target)
  - Comparison tests against torchaudio.functional.rnnt_loss checking both forward loss and backward gradients
  - Variable-length masking verified by asserting zero gradients in padded regions

  Closes the RNNT portion of #4459.

### DIFF
--- a/crates/burn-nn/src/loss/rnnt.rs
+++ b/crates/burn-nn/src/loss/rnnt.rs
@@ -82,10 +82,7 @@ impl RNNTLoss {
         let mut alpha = self.init_alpha(&lpl, b, max_up1, &device);
         alpha = neg_inf.clone().mask_where(u_mask.clone(), alpha);
 
-        let logit_lengths_exp = logit_lengths
-            .clone()
-            .reshape([b, 1])
-            .expand([b, max_up1]);
+        let logit_lengths_exp = logit_lengths.clone().reshape([b, 1]).expand([b, max_up1]);
 
         for t in 1..max_t {
             let new = self.step_alpha(&alpha, &lpb, &lpl, t);
@@ -323,7 +320,7 @@ impl RNNTLoss {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use burn::tensor::{Tolerance, TensorData};
+    use burn::tensor::{TensorData, Tolerance};
     use burn_ndarray::{NdArray, NdArrayDevice};
 
     type B = NdArray<f32>;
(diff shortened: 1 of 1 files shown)


crates/burn-nn/src/loss/rnnt.rs
```diff
@@ -0,0 +1,836 @@
...
+/// Configuration for [RNNTLoss](RNNTLoss).
+#[derive(Config, Debug)]
+pub struct RNNTLossConfig {
+    /// Index of the blank label in the vocabulary. Default: `0`.
+    #[config(default = 0)]
+    pub blank: usize,
+    /// If `true`, applies `log_softmax` to the last dimension internally.
+    /// If `false`, the input must already be log-probabilities. Default: `true`.
+    #[config(default = true)]
+    pub fused_log_softmax: bool,
```
laggui: I see this matches the args to torchaudio's RNNT loss, but to keep it consistent with other losses / metrics I would rename it to `logits`.

Essentially, treat the inputs as logits, applying log-softmax when computing the loss (equivalent to `fused_log_softmax = true`. If `false`, inputs are assumed to already be log-probabilities.

crates/burn-nn/src/loss/rnnt.rs
```diff
@@ -0,0 +1,836 @@
...
+        for t in 1..max_t {
+            let new = self.step_alpha(&alpha, &lpb, &lpl, t);
+            let new = neg_inf.clone().mask_where(u_mask.clone(), new);
+
+            // Keep old alpha for samples where t >= logit_lengths[b]
+            let valid = logit_lengths
+                .clone()
+                .greater_elem(t as i64)
+                .reshape([b, 1])
+                .expand([b, max_up1]);
```
laggui: 1. If it's supposed to be for `>= logit_lengths`, you should use `greater_equal_elem`

2. The only thing that changes is `t` for the comparison, so you could have reshape + expand outside the loop and just apply `greater_equal_elem` for the mask in the loop

crates/burn-nn/src/loss/rnnt.rs
```diff
@@ -0,0 +1,836 @@
...
+        assert!(
+            (a - e).abs() < tol,
+            "[{}] expected {:.6}, got {:.6} (diff {:.6})",
+            i,
+            e,
+            a,
+            (a - e).abs()
+        );
+    }
+}
```
laggui: Should use `tensor.into_data().assert_approx_eq(...)` instead

crates/burn-nn/src/loss/rnnt.rs
cong-or: Done, renamed to logits with updated docs to match the pattern in BinaryCrossEntropyLoss and CrossEntropyLoss.

crates/burn-nn/src/loss/rnnt.rs
cong-or: On `greater_equal_elem` — the code was correct, the comment was just misleading. valid is where we update alpha, so we need `logit_lengths > t`. With >=, shorter sequences process one extra invalid step — `variable_lengths_b3` fails with loss values off by ~1.5. 

I Rewrote the comment to be clearer.

cong-or:
> Thanks for contributing this implementation!
> 
> I only have a couple of minor comments

@laggui Thanks for the review. I have addressed the issues raised.

crates/burn-nn/src/loss/rnnt.rs
cong-or: Done, switched to assert_approx_eq and dropped the custom assert_close helper.

laggui:
> @laggui Thanks for the review. I have addressed the issues raised.

Looks like the code needs to run through `cargo fmt` first. Should be good afterwards.

cong-or:
> > @laggui Thanks for the review. I have addressed the issues raised.
> 
> Thanks for addressing my comments!
> 
> Looks like the code needs to run through `cargo fmt` first. Should be good afterwards.

thanks - just ran cargo fmt, hopefully ci will be green!

cong-or:
> LGTM!
> 
> Failing macos test is not due to your changes, should be related to the latest cubecl changes.

Thanks for the review! Let me know if there's anything else needed on my end.

PR #4644, opened 2026-03-18
feat: add FID vision metric           

cong-or:
Implements FID (Frechet Inception Distance) for #4312.

Ports the pytorch-fid InceptionV3 feature extractor and computes the Frechet distance between real/generated image distributions. Pretrained weight loading follows the same pattern as LPIPS and DISTS — downloads the pytorch-fid weights and maps them via PytorchStore with key remapping.

Uses Newton-Schulz iteration for the matrix square root since burn doesn't have an eigendecomposition op. Went with the symmetric formulation (sqrtm(S @ sigma2 @ S) instead of sqrtm(sigma1@ sigma2)) so that Newton-Schulz actually converges.

  Basic usage:

  ```rust
  let fid = FidConfig::new().init_pretrained(&device);
  let score = fid.forward(real_images, generated_images);
```
  Also added a docs entry in the burn-book.

### DIFF
--- a/crates/burn-train/src/metric/vision/fid/metric.rs
+++ b/crates/burn-train/src/metric/vision/fid/metric.rs
@@ -4,11 +4,13 @@ use burn::config::Config;
 use burn::module::Module;
 use burn::tensor::Tensor;
 use burn::tensor::backend::Backend;
+use burn::tensor::linalg;
 
 use super::inception::InceptionV3FeatureExtractor;
 
 const IMAGENET_MEAN: [f32; 3] = [0.485, 0.456, 0.406];
 const IMAGENET_STD: [f32; 3] = [0.229, 0.224, 0.225];
+const EPS: f64 = 1e-6;
 
 /// Configuration for [Fid].
 ///
@@ -94,15 +96,9 @@ impl<B: Backend> Fid<B> {
 
 fn imagenet_normalize<B: Backend>(x: Tensor<B, 4>) -> Tensor<B, 4> {
     let device = x.device();
-    let [batch, _, h, w] = x.dims();
 
-    // Build [1, 3, 1, 1] broadcastable tensors
-    let mean = Tensor::<B, 1>::from_floats(IMAGENET_MEAN, &device)
-        .reshape([1, 3, 1, 1])
-        .expand([batch, 3, h, w]);
-    let std = Tensor::<B, 1>::from_floats(IMAGENET_STD, &device)
-        .reshape([1, 3, 1, 1])
-        .expand([batch, 3, h, w]);
+    let mean = Tensor::<B, 1>::from_floats(IMAGENET_MEAN, &device).reshape([1, 3, 1, 1]);
+    let std = Tensor::<B, 1>::from_floats(IMAGENET_STD, &device).reshape([1, 3, 1, 1]);
 
     x.sub(mean).div(std)
 }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: crates/burn-cubecl/src/kernel/conv/conv_transpose2d/transpose_direct.rs
Comment (laggui) [inline crates/burn-cubecl/src/kernel/conv/conv_transpose2d/transpose_direct.rs:1]: The formatting fixes have been applied to main. No changes should be necessary in burn-cubecl for this PR.

File: crates/burn-train/src/metric/vision/fid/metric.rs [hunk 1]
```diff
@@ -0,0 +1,376 @@
...
+    let device = x.device();
+    let [batch, _, h, w] = x.dims();
+
+    // Build [1, 3, 1, 1] broadcastable tensors
+    let mean = Tensor::<B, 1>::from_floats(IMAGENET_MEAN, &device)
+        .reshape([1, 3, 1, 1])
+        .expand([batch, 3, h, w]);
+    let std = Tensor::<B, 1>::from_floats(IMAGENET_STD, &device)
+        .reshape([1, 3, 1, 1])
+        .expand([batch, 3, h, w]);
```
laggui: The `.expand()` is not necessary. Broadcasting on a tensor of shape `[1, 3, 1, 1]` should work.

With all the recent vision metrics added that leverage an imagenet-pretrained feature extractor, we should really add a public function or module for normalization. Should simplify some of the duplicated code.

crates/burn-train/src/metric/vision/fid/metric.rs
```diff
@@ -0,0 +1,376 @@
...
+fn matrix_sqrt_newton_schulz<B: Backend>(a: Tensor<B, 2>, num_iterations: usize) -> Tensor<B, 2> {
+    let [d, _] = a.dims();
+    let device = a.device();
+
+    let norm_a = a.clone().mul(a.clone()).sum().sqrt();
+    let norm_val = norm_a.clone().into_data().to_vec::<f32>().unwrap()[0] as f64;
+
+    if norm_val < 1e-10 {
+        return Tensor::zeros([d, d], &device);
+    }
```
laggui: Is this check required? Or could we simply clamp the `norm_a` value?

Also, why `1e-10`? Is it a standard eps value used for FID? Looks like you also used `1e-6` later. Should these be configurable?

crates/burn-train/src/metric/vision/fid/metric.rs
```diff
@@ -0,0 +1,376 @@
...
+            .into_data()
+            .to_vec::<f32>()
+            .unwrap()[0] as f64;
+        let tr2 = trace_2d(sigma2.clone())
+            .into_data()
+            .to_vec::<f32>()
+            .unwrap()[0] as f64;
+        ((tr1 + tr2) / (2.0 * d as f64)).max(1e-10)
+    };
+    let reg = Tensor::<B, 2>::eye(d, &device).mul_scalar(1e-6 * avg_variance);
```
laggui: Instead of calling `into_data()` (which forces a sync on GPU), this could be done entirely with tensor ops.

Something like
```rust
let avg_variance = trace_2d(sigma1.clone())
    .add(trace_2d(sigma2.clone()))
    .div_scalar(2.0 * d as f64)
    .clamp_min(1e-10);

let reg = Tensor::<B, 2>::eye(d, &device)
    .mul_scalar(1e-6)
    .mul(avg_variance.unsqueeze());
```

crates/burn-train/src/metric/vision/fid/metric.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    mean_term.add(trace).reshape([1])
+}
+
+/// Sum of diagonal elements of a 2D tensor.
+fn trace_2d<B: Backend>(m: Tensor<B, 2>) -> Tensor<B, 1> {
+    let [d, _] = m.dims();
+    let eye = Tensor::<B, 2>::eye(d, &m.device());
+    m.mul(eye).sum()
+}
```
laggui: We have a `linalg::trace` function that should be used instead.

cong-or: @laggui  Rebased, cubecl stuff is gone now. Also fixed the other things — no more .expand() in normalize, got rid of the into_data() syncs, using linalg::trace now, and cleaned up the eps values into one constant.

cong-or:
> Thanks for addressing my comments!

Thanks for the review @laggui

PR #4645, opened 2026-03-18
Perf/burn fusion overhead

nathanielsimard:
This PR introduces UnfusedOp, an arena-backed wrapper for fusion operations that reduces allocation overhead by storing small operations in a thread-local arena instead of heap-allocating via Arc. Operations larger than the arena slot size fall back to Arc allocation. The client now selects the optimal transfer strategy based on operation size.

Other changes include replacing HashMap with SmallVec in OperationStreams for better cache locality on the typically small number of tensor-stream mappings, refactoring stream draining to use StreamId::executes, and re-enabling the fusion feature in the text-classification example.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,27 +98,12 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
-dependencies = [
- "anstyle",
- "anstyle-parse 0.2.7",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is_terminal_polyfill",
- "utf8parse",
-]
-
-[[package]]
-name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
- "anstyle-parse 1.0.0",
+ "anstyle-parse",
  "anstyle-query",
  "anstyle-wincon",
  "colorchoice",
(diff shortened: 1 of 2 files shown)


crates/burn-fusion/src/arena.rs
```diff
@@ -0,0 +1,479 @@
...
+    alive: Option<Arc<RwLock<bool>>>,
+    cursor: usize,
+}
+
+/// The initialized reserved memory.
+pub struct ReservedMemory<const MAX_ITEM_SIZE: usize> {
+    data: Arc<UnsafeCell<Bytes<MAX_ITEM_SIZE>>>,
+    alive: Arc<RwLock<bool>>,
+    drop_fn: fn(&mut Bytes<MAX_ITEM_SIZE>),
+}
```
laggui: I think the drop responsibility should not be "shared". The drop logic of the arena and reserved memory are intertwined. It forces coordination via a flag (`alive`) and ref counts, which seems a bit fragile.

After reviewing the logic a bit more, I think we could simplify this to a single owner:

```rust
pub struct Arena<const MAX_ITEM_COUNT: usize, const MAX_ITEM_SIZE: usize> {
    /// The arc here is only to have a stable pointer, since the buffer can be drop when the thread
    /// is done, but some bytes might still live, so the Arc handles that gracefully.
    buffer: Vec<Arc<UnsafeCell<Bytes<MAX_ITEM_SIZE>>>>,
    cursor: usize,
}

pub struct ReservedMemory<const MAX_ITEM_SIZE: usize> {
    /// Sole instance: drop fires exactly once when the last ReservedMemory clone is dropped.
    _drop_guard: Arc<DropGuard<MAX_ITEM_SIZE>>,
}

struct DropGuard<const MAX_ITEM_SIZE: usize> {
    data: Arc<UnsafeCell<Bytes<MAX_ITEM_SIZE>>>,
    drop_fn: fn(&mut Bytes<MAX_ITEM_SIZE>),
}

impl<const MAX_ITEM_SIZE: usize> Drop for DropGuard<MAX_ITEM_SIZE> {
    fn drop(&mut self) {
        let bytes_mut = unsafe { self.data.get().as_mut().unwrap() };
        (self.drop_fn)(bytes_mut);
    }
}
```

When the last `ReservedMemory` clone drops, `Arc<DropGuard>` hits zero, and the drop implementation  calls drop_fn. It should only happen once.

Seems to work locally, and the logic is simplified that way. But lmk if I missed something. Otherwise, looks good.

PR #4650, opened 2026-03-18
All reduce in backward

Charles23R:
## Pull Request Template

NOTE : 
- All-reduce for fusion is not implemented. Will do in a followup PR.
- Need to rework the way batch norm works/interacts with DDP cause it doesn't work right now.
- We would need some benchmarks for DDP. Further improvements can also be made (e.g. buckets for all-reduce).

### Checklist

### Changes

Option to dynamically call all_reduce on the gradients during the backward pass as soon as the gradient is ready. To do so, use the `grad_distributed` method on a module. The communication server handles the all_reduce calls for all the devices and also syncs the compute streams with the collective calls.

### Testing

New tests in `test_module_derive` and tested with `text-classification` example.

### DIFF
--- a/crates/burn-backend/src/backend/base.rs
+++ b/crates/burn-backend/src/backend/base.rs
@@ -12,7 +12,7 @@ use crate::tensor::{BoolTensor, FloatTensor, IntTensor, QuantizedTensor};
 use crate::{QTensorPrimitive, TensorData, TensorMetadata};
 
 #[cfg(feature = "distributed")]
-use crate::distributed::{DistributedParamId, DistributedParams};
+use crate::distributed::{DistributedBackend, DistributedParamId, DistributedParams};
 
 use super::DeviceOps;
 
@@ -205,6 +205,7 @@ impl core::fmt::Debug for ExecutionError {
     }
 }
 
+#[cfg(not(feature = "distributed"))]
 /// Trait that allows a backend to support autodiff.
 pub trait AutodiffBackend: Backend {
     /// The inner backend type.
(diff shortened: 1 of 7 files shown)


crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -661,6 +663,30 @@ $$\text{erf}\(x\) = \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2} dt$$
...
+        Self::new(primitive)
+    }
+
+    /// Returns the distributed parameters if the tensor was marked as distributed.
+    pub fn distributed_params(&self) -> Option<DistributedParams> {
+        match &self.primitive {
+            TensorPrimitive::Float(tensor) => B::float_distributed_params(tensor),
+            TensorPrimitive::QFloat(_tensor) => todo!(),
+        }
+    }
```
laggui: I don't think these should be exposed from the high-level tensor API

crates/burn-core/src/module/sharder.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Describes how the module is distributed across multiple devices.
+pub struct ModuleSharder;
+
+impl<B: Backend> ModuleMapper<B> for ModuleSharder {
+    fn map_float<const D: usize>(&mut self, param: Param<Tensor<B, D>>) -> Param<Tensor<B, D>> {
+        let (id, tensor, mapper) = param.consume();
+        let tensor = tensor.set_distributed_params(DistributedParamId::from(id.val()));
+        Param::from_mapped_value(id, tensor, mapper)
+    }
+}
```
laggui: I think we could potentially move this to burn-train only for now

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1531,6 +1536,17 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+    fn float_set_distributed_params(
+        tensor: FloatTensor<Self>,
+        param_id: DistributedParamId,
+    ) -> FloatTensor<Self> {
+        tensor.grad_distributed(param_id)
+    }
+
+    fn float_distributed_params(tensor: &FloatTensor<Self>) -> Option<DistributedParams> {
+        tensor.node.distributed_params.clone()
+    }
```
laggui: Would probably be a better fit under the autodiff backend trait instead, or as another trait that requires the autodiff backend trait.

crates/burn-backend/src/backend/base.rs
```diff
@@ -66,6 +66,7 @@ pub trait Backend:
     FloatTensorOps<Self>
     + BoolTensorOps<Self>
     + IntTensorOps<Self>
+    + CommunicationTensorOps<Self>
```
laggui: I don't think it should be required by the `Backend` trait. It probably makes more sense to have this requirement for burn-collective instead, and we could have this conditionally for autodiff backends that support this.

Right now burn-collective uses `burn-tensor` but it really doesn't really on the high-level `Tensor` API (outside of tests), it really should just use `burn-backend`.

crates/burn-core/src/module/base.rs
```diff
@@ -14,7 +15,25 @@ pub type Devices<B> = Vec<Device<B>>;
...
+            ) -> Param<Tensor<B, D>> {
+                let (id, tensor, mapper) = param.consume();
+                let func = $item;
+                let tensor = func(tensor, self.map_args.clone());
+                Param::from_mapped_value(id, tensor, mapper)
+            }
+        }
+        let mut mapper = Mapper { map_args: $args };
+        $module.map(&mut mapper)
+    }};
```
laggui: Some leftover that is not needed?

PR #4651, opened 2026-03-19
Add Adan optimizer implementation with tests

sepcnt:
### Checklist

### Changes

Add Adan optimizer implementation: https://arxiv.org/abs/2208.06677

### Testing

Aligned with Python implementation of Adan.


crates/burn-optim/src/optim/adan.rs
```diff
@@ -0,0 +1,461 @@
...
+            },
+            weight_decay: self.weight_decay,
+            no_prox: self.no_prox,
+        };
+
+        let mut optim = OptimizerAdaptor::from(optim);
+        if let Some(config) = &self.grad_clipping {
+            optim = optim.with_grad_clipping(config.init());
+        }
+        optim
```
laggui, line 124: I think this differs slightly from the reference implementation, which clips by global norm across all parameters.

But in the paper it seems to be an implementation detail, which might vary in practice:

> We clip the global gradient norm to 5 for ResNet and do not clip the gradient for ViT, Swin

So it's probably fine.

PR #4653, opened 2026-03-19
[Breaking] Use device settings to provide output dtype

laggui:
### Checklist

### Related Issues/PRs

Follow-up to #4649

### Changes

All backend ops that create a new tensor (either creation ops, or ops that return a different tensor kind e.g. `float_argmin`) receive the expected output dtype, so backends no longer rely on their own logic. Instead, he higher level methods check the device settings and provide the dtype accordingly.

Renamed `DevicePolicy` to `DeviceSettings` + moved it into `burn-backend` + added bool dtype.

### Testing

Unit tests pass

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -1874,7 +1874,6 @@ where
         // Use the given dtype when provided, otherwise default device dtype
         let opt = options.into();
         let dtype = opt.resolve_dtype::<K>();
-        println!("from_data {dtype:?} {}", K::name());
         Self::new(K::from_data(data, &opt.device, dtype))
     }
 


crates/burn-backend/src/backend/device.rs
```diff
@@ -15,3 +41,528 @@ pub trait DeviceOps: Clone + Default + PartialEq + Send + Sync + core::fmt::Debu
...
+}
+
+/// Key for the registry: physical device type + device id
+type RegistryKey = (DeviceId, TypeId);
+
+/// Global registry mapping devices to their settings.
+///
+/// Each value is wrapped in a `OnceLock` to enforce that settings are initialized only once
+/// per device.
+static REGISTRY: LazyLock<RwLock<HashMap<RegistryKey, Arc<OnceLock<DeviceSettings>>>>> =
```
nathanielsimard, line 85: I don't understand why we have a global registry AND thread local storage for a cache.

laggui: The global registry is for the shared state. This is required to access any device settings in a program.

The thread-local storage is an optimization to avoid the `RwLock` on the hashmap once the settings have been initialized.

Every time we call an operation that creates a new tensor or returns a tensor of a different kind (e.g. float argmax/argmin, or float comparison ops that return a bool), we need to get the device settings. The thread local cache is just a fast path that avoids the global read lock.

wingertge:
This breaks tests because threads are persistent and f16 tests reuse the f32 device settings stored in the `THREAD_LOCAL` cache

laggui:
> This breaks tests because threads are persistent and f16 tests reuse the f32 device settings stored in the `THREAD_LOCAL` cache

Yeah I'm working on the tests right now :sweat_smile: sorry

The initial goal was to refactor them to use the dispatch backend, but there were a bunch of inconsistencies across backends that implicitly used the associated elem types.

Had to split the work into different PRs otherwise the work area was way too big to review.

The device settings should only be initialized once anyway, which means each suite should be isolated by dtype.

Lmk if you have any comments on the device settings changes in this PR though! Or DM on discord.

PR #4655, opened 2026-03-21
feat: add multinomial (categorical distribution) sampling for 2D tensors

### LINKED ISSUE #1121: [Feature Request] Categorical Distribution Sampling
### Feature description

Categorical distribution sampling across multiple dimensions like the Pytorch multinomial function.

### Feature motivation

I am trying to write a RL (PPO) algorithm in an environment where the agent must choose an action. For this I need to sample from the probabilities of each action generated by the network. The sampling should be done on both single dimension arrays and batched arrays. I have searched for this, but I don't think this is implemented at the moment.

There are 4 other distributions implemented, but I don't see the Categorical. I would really appreciate if this was implemented.

Thank you!

majiayu000:
### Checklist

### Related Issues/PRs

Fixes #1121

### Changes

Add `Tensor<B, 2>::multinomial(num_samples, replacement)` method that samples from categorical distributions defined by rows of a 2D float tensor, returning integer category indices.

Algorithm uses inverse CDF via cumsum:
1. Normalize weights and compute cumulative sum along categories
2. Draw uniform random samples and broadcast-compare against cumsum
3. Count categories below each random threshold to get sampled indices

Key properties:
- Works across all backends automatically — no backend trait changes needed
- Matches PyTorch `torch.multinomial` semantics
- Supports unnormalized weights (auto-normalizes)
- `replacement=false` with `num_samples > 1` panics for now (future work)

### Testing

Added 9 tests in `crates/burn-backend-tests/tests/tensor/float/ops/multinomial.rs`:
- Output shape verification
- Deterministic sampling (single nonzero weight)
- Index range validation
- Batch correctness (per-row distributions)
- Statistical distribution check (equal weights)
- Single category edge case
- Unnormalized weights
- Panic on unsupported without-replacement mode

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/categorical.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/categorical.rs
@@ -63,11 +63,8 @@ fn categorical_indices_in_range() {
     let probs = TestTensor::<2>::from([[0.2, 0.2, 0.2, 0.2, 0.2]]);
     let samples = probs.categorical(100);
 
-    let data: TensorData = samples.into_data();
-    let values = data.to_vec::<i64>().unwrap();
-    for &v in &values {
-        assert!(v >= 0 && v < num_categories, "index {v} out of range");
-    }
+    let data = samples.into_data();
+    data.assert_within_range::<IntElem>(0..num_categories);
 }
 
 #[test]
@@ -110,7 +107,7 @@ fn categorical_statistical_distribution() {
     let samples = probs.categorical(num_samples);
 
     let data: TensorData = samples.into_data();
-    let values = data.to_vec::<i64>().unwrap();
+    let values = data.to_vec::<IntElem>().unwrap();
 
     let count_zero = values.iter().filter(|&&v| v == 0).count();
     let ratio = count_zero as f64 / num_samples as f64;


crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -1080,3 +1080,84 @@ $$\text{erf}\(x\) = \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2} dt$$
         Tensor::new(primitive)
     }
 }
+
+impl<B: Backend> Tensor<B, 2> {
```
laggui: The current implementation is restricted to `Tensor<B, 2>`, I would make it more flexible over any `const D` instead.

We can treat the last dimension as the number of categories and all leading dims as a batch. That way, we can support 1D, 2D and 3D which are the typical use cases.

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -1080,3 +1080,84 @@ $$\text{erf}\(x\) = \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2} dt$$
...
+    ///         [[0.0, 1.0, 0.0], [0.0, 0.0, 1.0]],
+    ///         &device,
+    ///     );
+    ///     let samples = probs.multinomial(4, true);
+    ///     // First row always samples index 1, second row always samples index 2
+    ///     println!("{samples}");
+    /// }
+    /// ```
+    pub fn multinomial(self, num_samples: usize, replacement: bool) -> Tensor<B, 2, Int> {
+        if !replacement && num_samples > 1 {
```
laggui: Should probably check for `num_samples == 0` which should be invalid.

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -1080,3 +1080,84 @@ $$\text{erf}\(x\) = \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2} dt$$
...
+            panic!(
+                "multinomial: sampling without replacement with num_samples > 1 \
+                 is not yet supported"
+            );
+        }
```
laggui: Technically, sampling with replacement for `num_samples == 1` is just like sampling without replacement 😅 

I think we could simply restrict the current implementation to no replacement.

crates/burn-tensor/src/tensor/api/float.rs
```diff
@@ -1080,3 +1080,84 @@ $$\text{erf}\(x\) = \frac{2}{\sqrt{\pi}} \int_0^x e^{-t^2} dt$$
...
+
+        // Expand dimensions for broadcasting:
+        //   cumsum: [batch, categories, 1]
+        //   uniform: [batch, 1, num_samples]
+        let cumsum_3d: Tensor<B, 3> = cumsum.unsqueeze_dim(2);
+        let uniform_3d: Tensor<B, 3> = uniform.unsqueeze_dim(1);
+
+        // Count categories where cumsum < uniform (inverse CDF)
+        let mask: Tensor<B, 3, Bool> = cumsum_3d.lower(uniform_3d);
+        let indices = mask.int().sum_dim(1).squeeze_dim::<2>(1);
```
laggui: This uses the inverse CDF, but I think it would be good to have fast-paths for the gumbel max / top-k trick (which requires unnormalized log probabilities).

This would have to be a follow-up PR of course, but what do you think the best API for that would be?

Perhaps a builder-style distribution sampler, e.g.

```rust
// Defaults to num_samples = 1, replacement = false, and from_logits uses the fast-path for log probs 
Categorical::from_logits(logits).sample()

// Defaults to num_samples = 1, replacement = false, and general inverse CDF
Categorical::new(weights).sample()
```

crates/burn-tensor/src/tensor/api/float.rs
majiayu000: Done. Generalized to `Tensor<B, D>` — the last dimension is treated as categories and all leading dimensions as batch. Flattens to 2D internally for the algorithm, then reshapes back. Added tests for 1D and 3D inputs.

crates/burn-tensor/src/tensor/api/float.rs
majiayu000: Done. Added `assert!(num_samples > 0, ...)` check and a corresponding `#[should_panic]` test.

crates/burn-tensor/src/tensor/api/float.rs
majiayu000: Done. Removed the `replacement` parameter entirely. The current implementation just draws independent samples (no replacement concept exposed).

crates/burn-tensor/src/tensor/api/float.rs
majiayu000: Great idea! I think the builder-style API you suggest is clean and extensible:

```rust
Categorical::from_logits(logits).sample()
Categorical::new(weights).num_samples(k).sample()
```

`from_logits` could use the Gumbel-max trick (add Gumbel noise to logits, then top-k) which avoids the cumsum and is numerically more stable for log-probs. `new(weights)` would keep the current inverse CDF path.

Happy to work on this as a follow-up PR once this one lands.

laggui: Yeah let's do this in another PR :slightly_smiling_face:

P.S. I think you missed the cargo fmt with the changes

PR #4675, opened 2026-03-25
Fix fusion kernel vector_size mismatch on f16 output writes

AdrianEddy:
### Checklist

### Changes
Fix a bug where fused kernels corrupted memory when writing to output tensors whose `vector_size` differs from the kernel's computation width (`config.width`)

# Bug details

When a fusion kernel runs at a certain vectorization width (e.g. `config.width = 4`, meaning it processes 4 elements at a time as a `vec4<f16>`), it assumed all output tensors shared that same vector size. However, some output tensors — notably BatchNorm's inplace running-mean/running-var statistics — are registered as alias tensors with `vector_size = 1` (scalar).

The old `write` function would `Vector::cast_from(value)` and store it directly into `tensor.tensor[offset]`, which emitted a SPIR-V store of e.g. `vec4<f16>` into a slot typed as scalar `f16`. This wrote 4 half-floats (8 bytes) into a 1-element (2-byte) slot, overwriting 6 bytes of adjacent memory. The corruption would appear as Inf/NaN values in downstream tensors.

This only manifested with `f16` because `f32` tensors on WGPU tend to all receive the same vectorization, so the mismatch never occurred in practice.

**Trigger pattern:** Two parallel conv + BatchNorm branches whose results are added. The BatchNorm running-stats outputs create the vector_size=1 aliases inside a fused block running at width=4.

# Fix
- In `write()`, check at comptime whether the output tensor's `vector_size` matches `config.width`
- When they match: use the existing direct-store path (with an added `set_polyfill` for correct typing)
- When they differ: call new `write_output_aligned()` which decomposes the vector and writes element-by-element, mirroring the existing `read_input_aligned` approach on the read path
- The `write_output_aligned` function handles both same-as-ref layouts (simple offset arithmetic) and unknown layouts (per-element offset computation respecting broadcast dimensions)

### Testing

Two new tests were added. These tests run the same operation on fused and unfused backend and compare the results. The results should match. Without this fix, they don't match.

- [x] `fusion_f16_two_branch_conv_bn_add_matches_reference` — verifies fused output matches unfused reference for conv+BN two-branch add (the minimal repro)
- [x] `fusion_f16_two_branch_conv_manual_bn` — verifies fused output matches unfused reference for conv + manual batch-norm with broadcast parameters
- [x] Existing fusion tests still pass

The tests and the fix were written by Claude Code. While the tests make sense (obviously fused/non-fused backends should produce the same results), the fix is more complicated, so please review all claims here carefully

PS. For some reason the tests without fix are causing stack overflow. I had to set env variable `RUST_MIN_STACK=167772160` before `cargo test`

### DIFF
--- a/crates/burn-cubecl-fusion/src/engine/codegen/io.rs
+++ b/crates/burn-cubecl-fusion/src/engine/codegen/io.rs
@@ -447,7 +447,7 @@ pub fn write<C: Scalar, N: Size>(
             if comptime![output_vs != config.width] {
                 // Output tensor has a different vector_size than the computation width.
                 // Write element-by-element to avoid SPIR-V type mismatch.
-                write_output_aligned(outputs, locals, pos, ref_pos, value, layout, config);
+                write_output_aligned(inputs, outputs, locals, pos, ref_pos, value, layout, config);
             } else {
                 // Vector sizes match - set polyfill to output type and write directly.
                 set_polyfill::<DynElem, DynSize>(comptime![tensor.ty]);
@@ -477,6 +477,7 @@ pub fn write<C: Scalar, N: Size>(
 /// differs from the computation width. Mirrors [read_input_aligned] for the write path.
 #[cube]
 fn write_output_aligned<C: Scalar, N: Size>(
+    inputs: &GlobalArgs,
     outputs: &mut GlobalArgs,
     locals: &LocalArgs,
     #[comptime] pos: usize,
@@ -509,18 +510,16 @@ fn write_output_aligned<C: Scalar, N: Size>(
 
             #[unroll]
             for i in 0..config.width {
-                let flat_ref = base_flat + i;
-                let mut offset = 0;
-
-                #[unroll]
-                for d in 0..config.rank {
-                    let ogwl = flat_ref / locals.ref_strides[d];
-                    let tensor_ro = outputs.tensors.index(pos);
-                    offset += ogwl % tensor_ro.tensor.shape(d) * tensor_ro.tensor.stride(d);
-                }
-
                 let tensor_ro = outputs.tensors.index(pos);
-                offset /= tensor_ro.tensor.vector_size();
+                let offset = index_offset_with_layout(
+                    inputs,
+                    tensor_ro,
+                    locals,
+                    base_flat + i,
+                    None,
+                    config.rank,
+                    None,
+                );
                 let output = outputs.tensors.index_mut(pos);
                 output.tensor[offset] = Vector::cast_from(value[i]);
             }
@@ -587,11 +586,12 @@ fn get_offset(
     #[comptime] config: &FuseBlockConfig,
     #[comptime] transform: Option<Transform>,
 ) -> usize {
+    let offset_ref = ref_pos * locals.ref_vector_size;
     index_offset_with_layout(
         inputs,
         tensor,
         locals,
-        ref_pos,
+        offset_ref,
         range,
         config.rank,
         transform,
@@ -721,7 +721,7 @@ fn index_offset_with_layout(
     inputs: &GlobalArgs,
     tensor: &GlobalTensor,
     locals: &LocalArgs,
-    index: usize,
+    offset_ref: usize,
     #[comptime] range: Option<(usize, usize)>,
     #[comptime] rank: usize,
     #[comptime] transform: Option<Transform>,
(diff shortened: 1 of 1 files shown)


crates/burn-cubecl-fusion/src/engine/codegen/io.rs
```diff
@@ -464,6 +473,61 @@ pub fn write<C: Scalar, N: Size>(
...
+    #[comptime] layout: LayoutInfo,
+    #[comptime] config: &FuseBlockConfig,
+) {
+    let tensor = outputs.tensors.index(pos);
+    set_polyfill::<DynElem, DynSize>(comptime![tensor.ty]);
+
+    match layout {
+        LayoutInfo::SameAsRef | LayoutInfo::IsRef => {
+            let offset = (ref_pos * config.width) / tensor.tensor.vector_size();
+            let tensor = outputs.tensors.index_mut(pos);
```
nathanielsimard: Instead of naming the variable tensor, output would be better I thiknk `output.tensor[i] = ...` we can avoid `tensor.tensor` I think

crates/burn-cubecl-fusion/src/engine/codegen/io.rs
```diff
@@ -464,6 +473,61 @@ pub fn write<C: Scalar, N: Size>(
...
+                for d in 0..config.rank {
+                    let ogwl = flat_ref / locals.ref_strides[d];
+                    let tensor_ro = outputs.tensors.index(pos);
+                    offset += ogwl % tensor_ro.tensor.shape(d) * tensor_ro.tensor.stride(d);
+                }
+
+                let tensor_ro = outputs.tensors.index(pos);
+                offset /= tensor_ro.tensor.vector_size();
+                let tensor = outputs.tensors.index_mut(pos);
+                tensor.tensor[offset] = Vector::cast_from(value[i]);
```
nathanielsimard: Is it possible to use `index_offset_with_layout` instead of doing the index math here?

crates/burn-cubecl-fusion/Cargo.toml
```diff
@@ -51,6 +51,8 @@ serde = { workspace = true }
 
 [dev-dependencies]
 cubecl = { workspace = true, features = ["test-runtime"] }
+burn = { path = "../burn", version = "=0.21.0-pre.2", features = ["vulkan", "fusion"] }
+half = "2.7"
```
nathanielsimard: I don't think we need to commit this.

AdrianEddy: this is needed for the tests, no? Is there any other way?

crates/burn-cubecl-fusion/src/engine/codegen/io.rs
AdrianEddy: Not directly. There are two key incompatibilities:

- Scalar vs. vector indexing: `index_offset_with_layout` takes a vectorized index and internally computes `index * locals.ref_vector_size` to get the flat scalar position. The manual code needs per-scalar-element offsets (`ref_pos * config.width + i` for each `i`). Since `ref_vector_size` may differ from `config.width` (that's the whole reason `write_output_aligned` exists), you can't cleanly express each scalar flat index as `some_integer * ref_vector_size`.

- Single offset vs. per-element offsets: `index_offset_with_layout` returns one offset for a vector position. The manual loop needs `config.width` independent offsets because broadcast dimensions can cause multiple reference elements to alias to the same output position — you can't derive them from a single base offset + stride walk (unlike the read path in `read_input_aligned`, which assumes no aliasing).

Compared to `read_input_aligned`, the critical difference is:

- Inputs can have stride=0 in broadcast dimensions, so the stride walk offset + i * 0 correctly reads the same element for all positions.
- Outputs are real allocated tensors with non-zero strides, so the stride walk would write to out-of-bounds positions when the output's last dim is smaller than the reference's.

This also means `read_input_aligned` might itself have an issue in the Unknown case if the input layout differs from the reference in ways beyond simple broadcasting (e.g., transposed strides causing row-crossing).

<--- This is from Claude, I have not verified these claims, because I don't have enough knowledge of the internals here

AdrianEddy: I slightly changed `index_offset_with_layout` so it can be reused here

crates/burn-cubecl-fusion/Cargo.toml
nathanielsimard: Ha nvm

Issue #4686, opened 2026-03-29
WGPU nearest-neighbor interpolation produces wrong output for certain spatial dimensions


wkrettek:
**Describe the bug**
burn::tensor::module::interpolate with InterpolateMode::Nearest produces incorrect output on the WGPU backend for certain input
  spatial dimensions. The NdArray backend produces correct results for the same inputs.

**To Reproduce**
Compare WGPU vs NdArray for 2x nearest-neighbor upsampling on [1, 64, H, H] tensors.

  Broken sizes (max_diff ~1.0): 214, 218, 220, 227, 230, 235, 242, 245, 255
  Working sizes (max_diff = 0.0): 200, 210, 212, 216, 224, 240, 256

Minimal reproduction: https://github.com/wkrettek/burn-interpolate-repro

**Expected behavior**
Ndarray and Wgpu backends should produce the same output.

**Screenshots**
<img width="1014" height="724" alt="Image" src="https://github.com/user-attachments/assets/b704b039-399e-4ace-9746-ca7ada0fbb0a" />

**Desktop (please complete the following information):**
 - OS: CachyOS
- RTX 3080 Ti
- Version Burn 0.20.1, cubecl 0.9.0
- NVIDIA GeForce RTX 3080 Ti, driver 595.58.03, Vulkan 1.4.341 
- Linux 6.19.10

**Additional context**

**Root cause**

  The nearest-neighbor kernel in crates/burn-cubecl/src/kernel/interpolate/nearest.rs uses f32 arithmetic to map output coordinates to
   input coordinates:

  let y = y as f32 * (h_in / h_out);
  let x = x as f32 * (w_in / w_out);

  For certain dimension pairs, the GPU's f32 division produces a result that lands just below an integer boundary. The subsequent `as
  usize` truncation then selects the wrong input pixel. For example, a value that should be exactly 3.0 computes as 2.9999998 and
  truncates to 2.

antimora:
Fixed by #4687. Closing.

PR #4687, opened 2026-03-29
fix: use integer arithmetic for nearest-neighbor coordinate scaling

wkrettek:
The WGPU nearest-neighbor interpolation kernel used f32 division to map output coordinates to input coordinates. For certain spatial dimensions, GPU f32 division produces values that land just below an integer boundary, causing the truncation to select the wrong input pixel.

Replace with integer arithmetic (y * in_h / out_h) which is exact for all valid nearest-neighbor coordinate mappings.

### Checklist

### Changes

Addresses #4686. Use integer math instead of floating point. 

### Testing

Created a minimal [reproduction](https://github.com/wkrettek/burn-interpolate-repro) and tested the change fixes the broken behavior from before.


wkrettek:
Okay I added an appropriate test. It's a little awkward because failure varied with values of `h` so it's a little bit tuned to this issue specifically. Let me know if the comments are annoying I can remove them.

Thanks!

PR #4696, opened 2026-03-30
Fix: defer CPU staging buffer drops with PendingDropQueue

nathanielsimard:
Fix from CubeCL: https://github.com/tracel-ai/cubecl/pull/1255


louisfd:
More recent update done afterwards

PR #4697, opened 2026-03-31
Fix attention_fallback NaN for fully-masked rows

### LINKED ISSUE #4694: attention_fallback produces NaN for fully-masked rows
## Summary

`attention_fallback` produces NaN output for query rows where all key positions are masked (by bool mask + causal mask combined). The naive softmax computes `-inf - (-inf) = NaN`, which propagates through the entire row.

## Root cause

In `crates/burn-backend/src/backend/ops/modules/attention.rs`, the softmax is:

```rust
let max_per_dim = B::float_max_dim(attention_scores.clone(), 3);
let minus_max = B::float_sub(attention_scores, max_per_dim);
let numerator = B::float_exp(minus_max);
let sum_exp = B::float_sum_dim(numerator.clone(), 3);
let softmax = B::float_div(numerator, sum_exp);
```

When all positions in a row are `-inf` (fully masked):
1. `max = -inf`
2. `minus_max = -inf - (-inf) = NaN`
3. `numerator = exp(NaN) = NaN`
4. `sum = NaN`
5. `softmax = NaN / NaN = NaN`

The correct output for a fully-masked row is 0 (no position contributes to the output).

## Reproduction

`test_attention_all_options` in `burn-backend-tests` triggers this when `is_causal=true` combined with a random bool mask. For early query rows (small `qi`), the causal mask allows very few key positions. If the random bool mask also masks those positions, the entire row is `-inf`, producing NaN in the fallback but 0 from backends with NaN-safe flash attention.

The test compares `attention()` against `attention_fallback()`, so backends that correctly return 0 for fully-masked rows will fail the comparison against the NaN fallback output.

## Suggested fix

Two clamps make the softmax NaN-safe without affecting normal rows:

```rust
let max_per_dim = B::float_max_dim(attention_scores.clone(), 3);
let max_per_dim = B::float_clamp_min(max_per_dim, (-1e4).into());
let minus_max = B::float_sub(attention_scores, max_per_dim);
let numerator = B::float_exp(minus_max);
let sum_exp = B::float_sum_dim(numerator.clone(), 3);
let sum_exp = B::float_clamp_min(sum_exp, (1e-6).into());
let softmax = B::float_div(numerator, sum_exp);
```

- **Clamp max to `-1e4`**: `-inf - (-1e4) = -inf` (not NaN), then `exp(-inf) = 0`
- **Clamp sum to `1e-6`**: `0 / 1e-6 = 0` (not `0 / 0 = NaN`)

Normal rows are unaffected: legitimate max values are much larger than `-1e4`, and the sum is always `>= 1.0` (since the max element contributes `exp(0) = 1`).

antimora:
## Summary

- Fix NaN output in `attention_fallback` when all positions in a softmax row are masked (`-inf`)
- Add two clamps to the softmax computation: floor the max (prevents `-inf - (-inf) = NaN`) and floor the sum (prevents `0 / 0 = NaN`)
- Add regression tests for fully-masked rows with bool-mask-only and causal+bool-mask paths

## Root cause

When both bool mask and causal mask combine to mask every key position for a query row, all attention scores become `-inf`. The naive softmax then computes `max = -inf`, `shift = -inf - (-inf) = NaN`, which propagates through the entire row. Backends with NaN-safe flash attention return 0 for these rows, causing a mismatch with the fallback.

## Fix

Two clamps make the softmax NaN-safe without affecting normal rows:
- `SOFTMAX_MAX_FLOOR (-1e4)`: clamps max so `-inf - (-1e4) = -inf` (not NaN), then `exp(-inf) = 0`
- `SOFTMAX_SUM_EPS (1e-6)`: clamps sum so `0 / 1e-6 = 0` (not `0 / 0 = NaN`)

Normal rows are unaffected: legitimate max values are much larger than `-1e4`, and sum is always `>= 1.0`.

## Test plan

- [x] `test_attention_fully_masked_rows_no_nan` - bool mask only, asserts no NaN and output near 0
- [x] `test_attention_fully_masked_rows_causal_no_nan` - causal + bool mask, same assertions
- [x] All 10 attention tests pass with `cargo test -p burn-backend-tests -- attention`

Fixes #4694

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/module/attention.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/module/attention.rs
@@ -5,6 +5,7 @@ use burn_tensor::Tolerance;
 use burn_tensor::module::attention;
 use burn_tensor::module::attention_fallback;
 use burn_tensor::ops::AttentionModuleOptions;
+use num_traits::{Float, cast::cast};
 
 #[test]
 fn test_attention_no_mask() {
@@ -349,12 +350,13 @@ fn test_attention_fully_masked_rows_no_nan() {
 
     let output_data = output.into_data();
     let values = output_data.as_slice::<FloatElem>().unwrap();
+    let tol: FloatElem = cast(1e-4f64).unwrap();
     assert!(
         !values.iter().any(|v| v.is_nan()),
         "Fully-masked rows should produce 0, not NaN"
     );
     assert!(
-        values.iter().all(|v| v.abs() < 1e-4),
+        values.iter().all(|v| v.abs() < tol),
         "Fully-masked rows should produce values near 0"
     );
 }
@@ -408,6 +410,7 @@ fn test_attention_fully_masked_rows_causal_no_nan() {
 
     let output_data = output.into_data();
     let values = output_data.as_slice::<FloatElem>().unwrap();
+    let tol: FloatElem = cast(1e-4f64).unwrap();
     assert!(
         !values.iter().any(|v| v.is_nan()),
         "Fully-masked rows should produce 0, not NaN"
@@ -415,13 +418,13 @@ fn test_attention_fully_masked_rows_causal_no_nan() {
     // Row 0 (indices 0..head_dim) should be ~0 since it's fully masked
     let row0 = &values[..head_dim];
     assert!(
-        row0.iter().all(|v| v.abs() < 1e-4),
+        row0.iter().all(|v| v.abs() < tol),
         "Fully-masked row 0 should produce values near 0"
     );
     // Rows 1-3 should have non-zero output (they have valid positions)
     let rest = &values[head_dim..];
     assert!(
-        rest.iter().any(|v| v.abs() > 1e-4),
+        rest.iter().any(|v| v.abs() > tol),
         "Non-masked rows should produce non-zero output"
     );
 }


antimora:
Addressing the review comments:

**Re: clamp floor could change results for legitimate finite rows**

The clamp only activates when `max_per_dim` is `-inf` (every position in the row is masked). For any row with at least one unmasked position, the max is a finite attention score from `QK^T * scale`, which is always >> `-1e4` in practice (bounded by `head_dim * scale * max_input_magnitude`).

A conditional approach using `float_is_inf` would be more theoretically precise, but it would add complexity (extra tensor ops, branching) for a case that cannot occur with real inputs. The unconditional clamp is the standard approach used in production softmax implementations for this reason.

The constants are now named (`SOFTMAX_MAX_FLOOR` / `SOFTMAX_SUM_EPS`) with doc comments explaining when they activate, making it easy to adjust if needed.

**Re: causal test should exercise the actual combination scenario**

Good point. Updated in a276660 - the test now masks only key 0 via bool mask. Combined with causal masking, row 0 is fully masked (can only attend to key 0, which is bool-masked) while rows 1-3 still have valid positions. The test verifies row 0 is ~0 and the remaining rows have non-zero output.

laggui:
Just need to fix cargo fmt, but LGTM otherwise!

antimora:
> Good one, I also had to do a clamp_min for that in flash attention:
> 
> ```rust
>     /// Replaces each value `v` (v >= 0) in a row with `1/v`.
>     ///
>     /// If `v = 0`, the result is set to `0` instead of `1/0`.
>     /// This occurs when the entire row is masked, meaning it should
>     /// contribute no information, and ensures numerical stability.
>     pub fn recip_inplace(&mut self) {
>         for i in 0..self.num_rows {
>             let row_val = self.vals[i];
> 
>             let epsilon = E::new(FULLY_MASKED_ROW_THRESHOLD);
>             let not_masked = E::cast_from(row_val >= epsilon);
>             let safe_val = clamp_min(row_val, epsilon);
>             let recip = safe_val.recip();
>             self.vals[i] = not_masked * recip;
>         }
>     }
> ```

It's very common op and I think we should expose some static values at dtype level. I filed an issue to improve: https://github.com/tracel-ai/burn/issues/4700

antimora:
@laggui unrelated mac tests are failing

laggui:
Yeah we are working on fixing those :+1:

Issue #3754, opened 2025-09-21
Module cloning is inefficient for CPU devices


antimora:
**Describe the bug**
Module cloning is not cheap for CPU-based devices. While working on PytorchStore benchmarks, it was observed that cloning modules incurs a noticeable overhead, especially when overriding modules. The current workaround relies on unsafe blocks in the applier to avoid cloning, which is not ideal.

There may be an underlying issue in the ndarray backend, or it could be related to module initialization during cloning (i.e., cloning a module that isn't initialized causes the init).

**To Reproduce**
1. Attempt to clone a module on a CPU device within the current backend implementation.
2. Memory doubles for CPU but not GPU based device

**Expected behavior**
Module cloning should be efficient and not incur significant overhead, especially for CPU devices.

**Additional context**
- Discussion from PR [#3741](https://github.com/tracel-ai/burn/pull/3741) suggests this could be an ndarray bug in the backend.
- Nathaniel noted that cloning a module that isn't initialized will cause the init.
- Unsafe workaround in applier is currently relied upon to avoid cloning.

PR #4701, opened 2026-03-31
Make Param<T> Sync for parallel model inference

antimora:
## Summary

- Replace `core::cell::OnceCell` (!Sync) in `Param<T>` with a thread-safe `SyncOnceCell` wrapper over `std::sync::OnceLock` / `spin::Once`
- Add `Sync` bound to initialization closures in `Uninitialized`, `Param::uninitialized`, and `init_mapper`
- Models can now be shared across threads via `&model` for parallel inference (e.g. with rayon)

### Problem

`Param<T>` used `core::cell::OnceCell<T>` for lazy parameter initialization. `OnceCell` is `!Sync`, which made all modules `!Sync`, preventing parallel inference patterns like:

```rust
let results: Vec<_> = inputs.into_par_iter().map(|input| {
    model.forward(input)  // ERROR: model is !Sync
}).collect();
```

### Fix

`SyncOnceCell` wraps `OnceLock` (std) or `spin::Once` (no_std) with a unified API. The overhead on the already-initialized fast path is a single atomic load, negligible compared to tensor operations.

A local wrapper is used instead of `burn_std::stub::SyncOnceCell` because that version requires `T: Debug` on all methods and lacks `get()`.

### Breaking change

`Param::uninitialized` and `Param::init_mapper` now require `F: ... + Sync` on top of the existing `Send` bound. In practice all closures passed to these methods capture only `Sync` types (configs, shapes, devices), so no real-world breakage is expected.

## Test plan

- [x] New compile-time test `param_is_sync` verifies `Param<Tensor<B, 2>>: Sync`
- [x] New concurrent test `param_concurrent_lazy_init` verifies 4 threads calling `val()` simultaneously on an uninitialized `Param` with no panics
- [x] All 12 param tests pass
- [x] `burn-nn` and `burn-store` compile (downstream callers of `Param::uninitialized` and `init_mapper`)
- [x] Rayon parallel inference example compiles and runs successfully

### DIFF
--- a/crates/burn-core/src/module/param/base.rs
+++ b/crates/burn-core/src/module/param/base.rs
@@ -463,14 +463,9 @@ mod tests {
 
         // Share across threads via &param (requires Sync).
         std::thread::scope(|s| {
-            let handles: Vec<_> = (0..4)
-                .map(|_| s.spawn(|| param.val()))
-                .collect();
+            let handles: Vec<_> = (0..4).map(|_| s.spawn(|| param.val())).collect();
 
-            let results: Vec<_> = handles
-                .into_iter()
-                .map(|h| h.join().unwrap())
-                .collect();
+            let results: Vec<_> = handles.into_iter().map(|h| h.join().unwrap()).collect();
 
             // All threads must get the same value.
             let expected = results[0].to_data();


nathanielsimard:
I'm unsure if it's a good idea actually. The `OnceCell` is not the only place which limits `Sync`. I'm unsure if tensor handle is `Sync` regardless of how initialization is done.

antimora:
@nathanielsimard 
> I'm unsure if it's a good idea actually. The `OnceCell` is not the only place which limits `Sync`. I'm unsure if tensor handle is `Sync` regardless of how initialization is done.

I have a real use case, I need to process data lots of data in parallel using CPU based backend. I think once params initialized, I should be able to run models in parallel, right?

nathanielsimard:
> @nathanielsimard
> 
> > I'm unsure if it's a good idea actually. The `OnceCell` is not the only place which limits `Sync`. I'm unsure if tensor handle is `Sync` regardless of how initialization is done.
> 
> I have a real use case, I need to process data lots of data in parallel using CPU based backend. I think once params initialized, I should be able to run models in parallel, right?

Let me check, the new memory management in CubeCL is Send/Sync so that should be fine.

nathanielsimard:
> > @nathanielsimard
> > > I'm unsure if it's a good idea actually. The `OnceCell` is not the only place which limits `Sync`. I'm unsure if tensor handle is `Sync` regardless of how initialization is done.
> > 
> > 
> > I have a real use case, I need to process data lots of data in parallel using CPU based backend. I think once params initialized, I should be able to run models in parallel, right?
> 
> Let me check, the new memory management in CubeCL is Send/Sync so that should be fine.

It's fine too in `burn-fusion`, we can make Param Sync!

ArthurBrussee:
I'm using a model in parallel - I thought the preferred option was just to clone the `Module` / `Param` / `Tensor` and send it to a thread?

antimora:
Cloning doesn't work with rayon either. The closure in `par_iter().map(|input| { model.clone(); ... })` captures `&model`, and rayon requires closures to be `Send + Sync`. Since `Param<T>` was `!Sync`, even the clone-per-thread pattern fails to compile.

The only workaround before this PR was manual `std::thread::spawn` with pre-cloned models - no rayon, no `thread::scope`, no `Arc`.

Additionally, cloning has its own issues - `Param::clone()` forces lazy initialization and doubles memory on CPU backends (see #3754). Sharing via `&model` avoids both problems.

This PR fixes the root cause: `Param<T>` now implements `Sync` (when `T: Sync`, which all backends satisfy). Both patterns work after this change:
- `&model` shared reference (zero-copy, natural rayon pattern)
- `model.clone()` per thread (also compiles now since the closure can capture `&model`)

ArthurBrussee:
"even the clone-per-thread pattern fails to compile" - why? Rayon has a method https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.map_with that only requires Send for the init?

And sounds like https://github.com/tracel-ai/burn/issues/3754 needs fixing, other backends don't have this issue and share memory.

It just seems odd to make Param Sync while Tensor isn't by design

antimora:
Good point about `map_with`, I wasn't aware of that API. You're right that it works today since it only requires `Send + Clone`.

That said, `map_with` is rayon-specific. Without `Sync` on `Param`, things like `std::thread::scope`, `Arc<Model>`, or simply passing `&model` across a thread boundary don't work either. This PR fixes the general case.

On the Tensor point though, `TensorMetadata` already requires `Clone + Send + Sync + Debug`, so tensors are `Sync` for all backends. It was only `Param<T>` breaking the chain because of `core::cell::OnceCell`. Aligning `Param` with the rest of the type system seems like the right thing to do.

Agreed that #3754 should be fixed separately. This PR just removes the need to clone in the first place.

nathanielsimard:
> I'm using a model in parallel - I thought the preferred option was just to clone the `Module` / `Param` / `Tensor` and send it to a thread?

It is, but since every tensor is still going to be cloned when calling `param.val()` before making any tensor operation, it is still safe to `Sync` if we use `OnceLock` for tensor init.

antimora:
@nathanielsimard does it mean we are good to go and merge it?

nathanielsimard (approved):

Issue #4659, opened 2026-03-22
quantization: reshaping block-quantized tensors silently zeros data on cube backends


sdroege:
**Describe the bug**

Taking a quantized tensor and reshaping it loses data. For example, reshaping a `[32]` tensor into `[2, 32]` sets the second half of the elements to zero.

This is probably related to using `QuantLevel::block([32])` as using `16` (or lower numbers) or `QuantLevel::Tensor` solves it.
See also further comments in the code example next to the quant scheme.

Transposing the quantized tensor in addition sets all elements to zero, which might make all this related to https://github.com/tracel-ai/burn/issues/4491

Note that this is with latest git main as of now (7c3be3be027645026e41c29233d8a089d5a28ef8) but also happens with the latest stable release.

**To Reproduce**

```rust
fn main() {
    let device = Default::default();

    let rows = 2;
    let cols = 16;
    let n_elements = rows * cols;
    let original = (0..n_elements)
        .map(|i| (i as f32) / n_elements as f32)
        .collect::<Vec<_>>();

    let scheme = QuantScheme::default()
        .with_level(QuantLevel::block([32])) // Using 16 makes it work so it's probably related to this
        .with_mode(QuantMode::Symmetric)
        .with_value(QuantValue::Q8S) // Also affects Q4S, Q2S sets all elements to 0
        .with_store(QuantStore::PackedU32(0))
        .with_param(QuantParam::F32); // Using F16 sets the second half to ridiculously large values

    let t1 = Tensor::<Wgpu, 1>::from_data(TensorData::new(original, [rows * cols]), &device);
    let t2 = t1.clone().quantize_dynamic(&scheme);
    let t1 = t1.reshape::<2, _>([rows, cols]);
    let t2 = t2.reshape::<2, _>([rows, cols]);

    // let t2 = t2.transpose(); // sets all elements to zero

    let s1 = t1.to_data();
    let s2 = t2.dequantize().to_data();

    for (i, (a, b)) in Iterator::zip(
        s1.as_slice::<f32>().unwrap().iter(),
        s2.as_slice::<f32>().unwrap().iter(),
    )
    .enumerate()
    {
        println!("{i:02} {a:1.4} {b:1.4}");
    }
}
```

This outputs

```
00 0.0000 0.0000
01 0.0312 0.0305
02 0.0625 0.0610
03 0.0938 0.0915
04 0.1250 0.1220
05 0.1562 0.1526
06 0.1875 0.1907
[...]

PR #4704, opened 2026-03-31
Add more checks for quantized tensor reshape

laggui:
### Checklist

### Related Issues/PRs

#4659

### Changes

- Refactor reshape analysis to use `Shape` and `Strides` + added tests
- Added more validation checks to `q_reshape` operation based on the current support
- Fixed block reshape when the block size essentially covers all tensor elements (acting as per-tensor quantization, like the MWE in the linked issue)

There are still some use cases to cover and fix, for example the `should_quantize_dequantize_per_tensor_reshaped_q4s_packed` which is currently marked as `#[ignore]` (incorrect output values). Likely an issue in the cubek implementation.

### Testing

Unit tests


crates/burn-cubecl/src/ops/base.rs
```diff
@@ -309,20 +309,93 @@ pub fn reshape<R: CubeRuntime>(mut tensor: CubeTensor<R>, shape: Shape) -> CubeT
...
+        QuantLevel::Block(block_size)
+            if block_size.len() == 1 && shape_last < (block_size[0] as usize) =>
+        {
+            // If the new last dimension is smaller than the block size,
+            // it means a single block now spans across multiple rows.
+            if scales.meta.shape().num_elements() > 1 {
+                unimplemented!("Reshape would split a block across multiple rows.");
+            }
+            // Exception: allow if there is exactly 1 block total (essentially per-tensor quantization)
+            scales.meta.shape().clone()
```
laggui, line 396: Maybe we should detect that in `quantize` / `quantize_dynamic` to force users to explicitly use `QuantLevel::Tensor` instead?

crates/burn-cubecl/src/ops/base.rs
```diff
@@ -309,20 +309,93 @@ pub fn reshape<R: CubeRuntime>(mut tensor: CubeTensor<R>, shape: Shape) -> CubeT
...
+                    && block_size.len() > 1
+                {
+                    // Split reshape (e.g. [32, 4] -> [32, 2, 2]): only valid if
+                    // reshaped dimension is aligned with the block boundaries.
+                    unimplemented!(
+                        "Split reshape of ND block-quantized tensor is not yet supported."
+                    );
+                }
+            }
+            _ => unreachable!(),
```
antimora: This `unreachable!()` is correct today since `HighlyPermuted`/`SmallerRank` always return `Recompute` and `NoChange` returns `NoChange`, so they can't appear inside the `UpdateStrides` guard. But it's a bit fragile since it relies on an invariant in a different module, and if it ever fires the default message gives no context.

Maybe something like:
```rust
other => unreachable!(
    "ReshapeAnalysis::{other:?} should never produce UpdateStrides"
),
```
At least then if the invariant breaks you immediately know which variant did it.

crates/burn-cubecl/src/ops/base.rs
```diff
@@ -309,20 +309,93 @@ pub fn reshape<R: CubeRuntime>(mut tensor: CubeTensor<R>, shape: Shape) -> CubeT
...
+                if let QuantLevel::Block(block_size) = scheme.level
+                    && block_size.len() > 1
+                    && !is_unsqueeze
+                {
+                    // General reshape (e.g. [32, 4] -> [16, 8]): only valid if
+                    // reshaped dimension is aligned with the block boundaries.
+                    unimplemented!("Reshape of ND block-quantized tensor is not yet supported.");
+                }
+            }
+            ReshapeAnalysis::Broadcasted => {}
```
antimora: The `IsContiguous` and `Split` arms both have ND-block guards, but `Broadcasted` silently passes everything through. I think this is actually fine since broadcasting only prepends unit dims which can't affect block boundaries, but a short comment here would save the next reader from going through the same reasoning.

crates/burn-cubecl/src/ops/base.rs
```diff
@@ -347,8 +420,17 @@ pub fn q_reshape<R: CubeRuntime>(mut tensor: CubeTensor<R>, shape: Shape) -> Cub
...
         }
-        (ReshapeAction::NoChange, ReshapeAction::NoChange) => {}
-        _ => {
+        // Any action to recompute
+        (ReshapeAction::Recompute, _) | (_, ReshapeAction::Recompute) => {
+            if let QuantLevel::Block(block_size) = scheme.level
+                && block_size.len() > 1
+            {
+                // Original block boundaries no longer align with the layout, would have to be recomputed
+                panic!(
```
antimora: This is the only place that uses `panic!()` - the other four checks all use `unimplemented!()`. Since the comment says "would have to be recomputed" it sounds like this is also a "not yet supported" case rather than a fundamentally invalid one. Might be worth making them consistent (either all `unimplemented!` or all `panic!`). The test's `#[should_panic]` message would need updating if you change it.

crates/burn-cubecl/src/ops/base.rs
```diff
@@ -347,8 +420,17 @@ pub fn q_reshape<R: CubeRuntime>(mut tensor: CubeTensor<R>, shape: Shape) -> Cub
...
+                    "Cannot reshape a block-quantized tensor when the reshape requires recomputing the buffer."
+                );
+            }
+
             tensor = kernel::into_contiguous(tensor);
             *tensor.meta = Metadata::new(shape, contiguous_strides(&shape_values));
```
antimora, line 444: For 1D blocks this falls through to `into_contiguous` without any block-boundary check. Is `into_contiguous` aware of quantization block structure? If it just reorders raw elements, the block-to-scale mapping could get corrupted silently.

If this path is safe for 1D blocks, a comment explaining why would be helpful. Also worth adding a test that actually exercises this (a highly permuted 1D-block-quantized tensor going through Recompute), since right now all the Recompute tests hit the ND-block panic instead.

crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
```diff
@@ -0,0 +1,139 @@
...
+#[should_panic = "Cannot reshape a block-quantized tensor when the reshape requires recomputing the buffer"]
+fn quantize_2d_block_reshape_should_be_valid() {
+    should_quantize_dequantize_per_block_arange_reshaped(
+        QuantLevel::Block(BlockSize::new([2, 4])),
+        QuantValue::Q8S,
+        QuantStore::PackedU32(0),
+        [4, 8],
+        [32], // invalid shape for 2D block boundaries
+    )
+}
```
antimora, line 115: Nit: the name `quantize_2d_block_reshape_should_be_valid` reads like this reshape should succeed, but the test expects a panic. Maybe something like `quantize_2d_block_reshape_recompute_should_panic`?

crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
```diff
@@ -0,0 +1,139 @@
...
+
+#[test]
+#[ignore] // FIXME: should work like tensor-level
+fn should_quantize_dequantize_per_block_reshaped_global_block_q4s_packed() {
+    should_quantize_dequantize_per_block_arange_reshaped(
+        QuantLevel::Block(BlockSize::new([32])),
+        QuantValue::Q4S,
+        QuantStore::PackedU32(0),
+        [32],
+        [2, 16],
```
antimora, line 64: These Q4S paths produce incorrect results at runtime, but there's no runtime guard preventing users from hitting them. Worth considering adding `unimplemented!` guards for Q4S packed reshape in the implementation until the cubecl fix lands, so users get a clear error instead of silently wrong data.

crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
```diff
@@ -0,0 +1,139 @@
...
+
+#[test]
+fn should_quantize_dequantize_per_block_reshaped_1d_q8s_native() {
+    if supports_native() {
+        should_quantize_dequantize_per_block_arange_reshaped(
+            QuantLevel::Block(BlockSize::new([32])),
+            QuantValue::Q8S,
+            QuantStore::Native,
+            [32],
+            [2, 16],
```
antimora, line 89: Minor: the `if supports_native()` guard means this test silently passes (green checkmark, zero assertions) on backends that don't support native quantization. Could be surprising when looking at CI results. An `#[ignore]` with a condition or at least a `println!` when skipping would make the skip visible.

crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
```diff
@@ -0,0 +1,139 @@
...
+    should_quantize_dequantize_per_block_arange_reshaped(
+        QuantLevel::Block(BlockSize::new([32])),
+        QuantValue::Q8S,
+        QuantStore::PackedU32(0),
+        [2, 32],
+        [4, 16],
+    )
+}
+
+// TODO: add tests for ND reshape split (validation should panic) and broadcasted
```
antimora: A few more test cases that would help:
- Multi-block successful reshape (all current success tests use exactly 1 block, e.g. `[4, 32] -> [1, 4, 32]` with block_size 32)
- Packed dimension alignment failure (`#[should_panic]` for the check at base.rs:323)
- ND-block unsqueeze (should succeed per the `is_unsqueeze` exemption, but nothing tests it)

laggui:
@antimora thanks for the comments, will address those soon!

@wingertge lmk if I missed anything, given that you've reworked this recently for b better shape/strides handling

wingertge:
@laggui This stuff is hard to reason about in the abstract, so as long as the tests pass and `burn-lm` doesn't panic I'd say it looks good. More checks isn't a bad idea, I didn't add them originally because it was kinda of a Temporary Fix™ and was supposed to be properly redone at some point 😅

laggui:
Yeah I agree, spent a bit of time myself going through the existing logic to validate what was going on :sweat_smile: just wanted to make sure I didn't forget anything obvious, thanks!

> Temporary Fix™

Love that hahaha

Temporary fixes have a habit of sticking around

laggui:
Oh and I validated with `burn-lm` as well! Though on main last I checked the output was incorrect, there must have been a regression somewhere else. I tested with these changes on top of the previous pre-release to make sure and all was good.

crates/burn-cubecl/src/ops/base.rs
laggui: You're right, missed that for 1D blocks. It should only be valid if the block covers the whole tensor (essentially like tensor-level).

Will adjust the check to be more defensive

crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
laggui: The issue with `#[ignore]` is that it's not dynamic, so for backends like cuda or vulkan that support native quantization we should not ignore. I could switch to  `#[might_panic]` but the output gets suppressed by passing test so we can't do much without `-- --nocapture`.

PR #4707, opened 2026-03-31
Feat/add rfft

Sublime12:
### Checklist

### Changes

Add the rfft implementation 
### Testing
cargo test-wgpu-nofuse fft
cargo test-cpu-nofuse fft

### DIFF
--- a/crates/burn-tensor/src/tensor/signal/fft.rs
+++ b/crates/burn-tensor/src/tensor/signal/fft.rs
@@ -52,11 +52,10 @@ pub fn rfft<B: Backend, const D: usize>(
     dim: usize,
 ) -> (Tensor<B, D>, Tensor<B, D>) {
     check!(TensorCheck::check_dim::<D>(dim));
-    // I did not see a tensor check for this assert
-    assert!(
-        signal.shape()[dim].is_power_of_two(),
-        "The dimension in which the fft is done must be a power of two"
-    );
+    check!(TensorCheck::check_is_power_of_two::<D>(
+        &signal.shape(),
+        dim
+    ));
     let (spectrum_re, spectrum_im) = B::rfft(signal.primitive.tensor(), dim);
     (
         Tensor::new(TensorPrimitive::Float(spectrum_re)),
--- a/crates/burn-tensor/src/tensor/api/check.rs
+++ b/crates/burn-tensor/src/tensor/api/check.rs
@@ -864,6 +864,21 @@ impl TensorCheck {
         check
     }
 
+    pub(crate) fn check_is_power_of_two<const D: usize>(shape: &Shape, dim: usize) -> Self {
+        let mut check = Self::Ok;
+        let dim_size = shape[dim];
+
+        if !dim_size.is_power_of_two() {
+            check = check.register(
+                "Check Is Power of two",
+                TensorError::new("The provided dimension size must be a power of two.")
+                    .details(format!("The length of dimension {dim} is {dim_size}.")),
+            );
+        }
+
+        check
+    }
+
     pub(crate) fn check_dim<const D: usize>(dim: usize) -> Self {
         let mut check = Self::Ok;
 


crates/burn-tensor/src/tensor/signal/fft.rs
```diff
@@ -0,0 +1,56 @@
...
+/// fn example<B: Backend>() {
+///     let device = B::Device::default();
+///     let signal = Tensor::<B, 1>::from_floats([1.0, 2.0, 3.0, 4.0], &device);
+///     let (real, imag) = burn_tensor::signal::rfft(signal, 0);
+/// }
+/// ```
+pub fn rfft<B: Backend, const D: usize>(
+    signal: Tensor<B, D>,
+    dim: usize,
+    // options: impl Into<TensorCreationOptions<B>>,
```
louisfd: dead code

crates/burn-cubecl/src/kernel/fft/base.rs
```diff
@@ -0,0 +1,39 @@
+use crate::ops::numeric::empty_device_dtype;
+use crate::{CubeRuntime, tensor::CubeTensor};
+use burn_backend::TensorMetadata;
+use cubecl::prelude::*;
+use cubek::fft::rfft_launch;
+
+/// launch the fft kernel
+pub fn rfft<R: CubeRuntime>(signal: CubeTensor<R>, dim: usize) -> (CubeTensor<R>, CubeTensor<R>) {
+    let dtype = f32::as_type_native_unchecked().storage_type();
+
```
antimora, line 14: This hardcodes the kernel dtype to f32 regardless of what the signal tensor actually is. If someone passes an f16 or f64 tensor, the kernel will interpret the memory with the wrong byte layout while the output tensors (allocated with `signal.dtype` below) expect the correct type. This will silently produce garbage.

Should derive from the signal's actual dtype instead.

crates/burn-cubecl/src/kernel/fft/base.rs
```diff
@@ -0,0 +1,39 @@
+use crate::ops::numeric::empty_device_dtype;
+use crate::{CubeRuntime, tensor::CubeTensor};
+use burn_backend::TensorMetadata;
```
antimora: Leftover commented-out import. Looks like the recent `remove dead comments` commit (40f25af) missed this one.

crates/burn-cubecl/src/kernel/fft/base.rs
```diff
@@ -0,0 +1,39 @@
...
+use cubecl::prelude::*;
+use cubek::fft::rfft_launch;
```
antimora, line 5: `TensorMetadata` is imported but never used. This will cause a compiler warning and likely fail `cargo run-checks`.

crates/burn-cubecl/src/kernel/fft/base.rs
```diff
@@ -0,0 +1,39 @@
...
+    rfft_launch(
+        &signal.client.clone(),
+        signal.binding(),
+        output_re.clone().binding(),
+        output_im.clone().binding(),
+        dim,
+        dtype,
+    )
+    .unwrap();
+
```
antimora, line 41: Bare `.unwrap()` here hides all error context if the kernel launch fails. Other kernel launches in this crate use `.expect("Kernel to never fail")` or propagate with `?`. At minimum swap this for an `.expect("rfft kernel launch failed")` so the panic message is actually useful.

crates/burn-cubecl/src/kernel/fft/base.rs
```diff
@@ -0,0 +1,39 @@
...
+    let output_im = empty_device_dtype(
+        signal.client.clone(),
+        signal.device.clone(),
+        output_shape.clone(),
+        signal.dtype,
+    );
+
+    rfft_launch(
+        &signal.client.clone(),
+        signal.binding(),
```
antimora, line 34: The clone here is unnecessary since you immediately borrow the result. `&signal.client` works fine.

crates/burn-tensor/src/tensor/signal/fft.rs
```diff
@@ -0,0 +1,55 @@
...
+///     let signal = Tensor::<B, 1>::from_floats([1.0, 2.0, 3.0, 4.0], &device);
+///     let (real, imag) = burn_tensor::signal::rfft(signal, 0);
+/// }
+/// ```
+pub fn rfft<B: Backend, const D: usize>(
+    signal: Tensor<B, D>,
+    dim: usize,
+) -> (Tensor<B, D>, Tensor<B, D>) {
+    let (spectrum_re, spectrum_im) = B::rfft(signal.primitive.tensor(), dim);
+    (
```
antimora, line 60: No bounds check on `dim`. If someone passes `dim=5` on a 2D tensor, this will panic somewhere deep in the backend with a confusing index-out-of-bounds error.

Other tensor ops in the codebase use `check!(TensorCheck::check_dim::<D>(dim))` for this. Would be good to add the same validation here so users get a clear error message at the API boundary.

crates/burn-tensor/src/tensor/signal/fft.rs
```diff
@@ -0,0 +1,55 @@
+use burn_backend::Backend;
+
+use crate::Tensor;
+use crate::TensorPrimitive;
+
+/// Computes the 1-dimensional discrete Fourier Transform of real-valued input.
+///
+/// Since the input is real, the Hermitian symmetry is exploited, and only the
```
antimora, line 10: Worth adding a note somewhere in this doc that autodiff is not yet supported. Right now if someone uses `rfft` in a training loop, they'll hit a `todo!()` panic during the backward pass with just "Unsupported operation" and no hint about what went wrong.

crates/burn-backend/src/backend/ops/modules/base.rs
```diff
@@ -1056,6 +1056,13 @@ pub trait ModuleOps<B: Backend> {
         attn_bias: Option<FloatTensor<B>>,
         options: AttentionModuleOptions,
     ) -> FloatTensor<B>;
+
+    ///Real-valued fast Fourier transform.
```
antimora: Nit: missing space after `///`. Should be `/// Real-valued fast Fourier transform.`

crates/burn-backend/src/backend/ops/modules/base.rs
```diff
@@ -1056,6 +1056,13 @@ pub trait ModuleOps<B: Backend> {
...
+    ///
+    /// Computes the discrete Fourier transform of a real-valued input along the given dimension.
+    /// The transform is applied independently for each slice along `dim`, returning the non-redundant
+    /// frequency components as separate real and imaginary tensors.
+    fn rfft(signal: FloatTensor<B>, dim: usize) -> (FloatTensor<B>, FloatTensor<B>);
```
antimora, line 1067: The return type doc doesn't specify which element is real and which is imaginary. Since both are `FloatTensor<B>`, a caller reading just the trait has to guess. A brief `# Returns` note ("first element is real, second is imaginary") would help.

crates/burn-cubecl/src/kernel/mod.rs
```diff
@@ -29,6 +29,8 @@ pub use crate::cubecl::prelude::KernelMetadata;
 pub mod attention;
 /// Convolution kernels
 pub mod conv;
+/// fft algorithms
+pub mod fft;
```
antimora, line 33: Nit: the other module comments here use title case ("Convolution kernels", "Grid sampling kernels"). This should be `/// FFT algorithms` to match.

crates/burn-backend-tests/tests/cubecl/fft.rs
```diff
@@ -0,0 +1,58 @@
+use super::*;
```
antimora, line 1: The tests here are well-chosen (sine/cosine to isolate real vs imaginary channels, multi-row to verify independent processing), but there are some gaps worth filling:

- All tests use `dim=1`. A dim=0 test would catch stride/indexing bugs on the other axis.
- All signals are length 8 (power of 2). If the kernel requires power-of-2, that constraint should be validated and documented. If it doesn't, a non-power-of-2 test would prove it.
- No explicit shape assertion on the output. The value checks implicitly verify shape, but an explicit `assert_eq!(re.shape(), ...)` makes failures easier to diagnose.
- Edge cases like all-zeros, constant signal, or length-1/length-2 signals would be quick to add and would cover boundary conditions.

crates/burn-cubecl/src/kernel/fft/base.rs
```diff
@@ -0,0 +1,39 @@
...
+use cubek::fft::rfft_launch;
+
+/// launch the fft kernel
+pub fn rfft<R: CubeRuntime>(signal: CubeTensor<R>, dim: usize) -> (CubeTensor<R>, CubeTensor<R>) {
+    let dtype = f32::as_type_native_unchecked().storage_type();
+
+    let signal_shape = signal.shape();
+    let mut output_shape = signal_shape.clone();
+    output_shape[dim] = output_shape[dim] / 2 + 1;
+
```
antimora, line 18: Related to the missing dim validation above: if `dim` is out of range, `output_shape[dim]` will panic here with a generic index error. The validation should happen before we get to this point, but if it doesn't, this is where it blows up.

crates/burn-tch/src/ops/module.rs
```diff
@@ -476,4 +477,8 @@ impl<E: TchElement> ModuleOps<Self> for LibTorch<E> {
             false,
         ))
     }
+
+    fn rfft(_signal: FloatTensor<Self>, _dim: usize) -> (FloatTensor<Self>, FloatTensor<Self>) {
+        todo!("Not supported for now in LibTorch")
```
antimora, line 482: PyTorch/LibTorch actually has `torch.fft.rfft`, so this could potentially be implemented rather than stubbed.

crates/burn-cubecl/src/kernel/fft/base.rs
Sublime12: This one is mandatory because we do signal.shape() which is implemented for CubeTensor in TensorMetadata

crates/burn-tch/src/ops/module.rs
Sublime12: Thank you for your review
I updated my mr

crates/burn-cubecl/src/kernel/fft/base.rs
Sublime12: signal.binding() on the next line also owns the signal
I get a compile error if I remove the cloning

laggui (approved):
LGTM! I think all comments have been addressed.

Should be easy to add the IR for fusion, even if it will not be fused at least the backends will support it. We can do that in a follow up PR 👍

PR #4709, opened 2026-03-31
Add native impementations for scatter_nd / gather_nd; provide autodiff for assign & add

### LINKED ISSUE #4433: Add native scatter_nd tensor operation
## Summary

Burn currently lacks a native `scatter_nd` operation. The ONNX `ScatterND` operator (opset 11+) updates a copy of a data tensor at positions specified by multi-dimensional index tuples.

## Motivation

The `burn-onnx` crate now supports importing ONNX models that use `ScatterND`, but the generated code has to fall back to converting tensors to raw data (`to_data()`), performing the scatter in a CPU loop, and converting back (`from_data()`). This works correctly but forces a CPU roundtrip that defeats GPU acceleration.

A native Burn operation would allow the scatter to execute on the backend device (GPU, etc.) without data transfer overhead.

## Proposed API

A possible API (naming open for discussion):

```rust
// Updates output[indices[idx]] = updates[idx] for each idx in the index space
// indices is a tensor of k-tuples specifying positions in data
let output = data.scatter_nd(indices, updates);

// With reduction (for handling duplicate indices)
let output = data.scatter_nd_add(indices, updates);
let output = data.scatter_nd_mul(indices, updates);
```

## ONNX Spec Reference

- [ScatterND](https://onnx.ai/onnx/operators/onnx__ScatterND.html)
- 3 inputs: `data` (rank r), `indices` (rank q, last dim = k), `updates` (rank q + r - k - 1)
- Output: same shape as data
- Reduction modes: none, add, mul, max, min
- This is the inverse of `GatherND`

## Current Workaround

The `burn-onnx` crate generates imperative Rust code that operates on raw tensor data via `to_data()`/`from_data()`, iterating over index tuples and performing element-wise updates. This is correct but not GPU-friendly.

### LINKED ISSUE #4469: Add gather_nd operation to Tensor API
## Summary

Add a `gather_nd` method to the Tensor API that gathers slices from a tensor using multi-dimensional index tuples. This is the N-dimensional generalization of `gather` and the inverse of `scatter_nd` (if one existed).

## Motivation

The ONNX `GatherND` operator is common in detection and NLP models. Without a native burn operation, the burn-onnx importer currently has to:
1. Flatten the data tensor to 1D
2. Compute flat indices on CPU
3. Upload the flat indices as a tensor
4. Call `select(0, flat_indices)` 
5. Reshape back

This works but is suboptimal. A native `gather_nd` could be implemented more efficiently at the backend level (e.g., a single GPU kernel that directly indexes into the N-dimensional tensor without flattening).

## API Sketch

```rust
/// Gather slices from `self` using multi-dimensional index tuples.
///
/// `indices` is an integer tensor where the last dimension contains
/// index tuples of length `k`. Each tuple indexes into the first `k`
/// dimensions of `self` (after `batch_dims` leading batch dimensions).
///
/// Output shape: indices.shape[..last] + self.shape[batch_dims + k..]
pub fn gather_nd(self, indices: Tensor<B, DI, Int>, batch_dims: usize) -> Tensor<B, DO, K>
```

## Prior Art

- **TensorFlow**: [`tf.gather_nd`](https://www.tensorflow.org/api_docs/python/tf/gather_nd) - first-class operation
- **PyTorch**: No dedicated function; achieved via advanced indexing (`tensor[idx[:, 0], idx[:, 1]]`)
- **ONNX**: [`GatherND`](https://onnx.ai/onnx/operators/onnx__GatherND.html) operator (opset 11+)

## Context

burn-onnx added GatherND support in https://github.com/tracel-ai/burn-onnx/pull/99 using the flatten+select workaround described above.

cu9hue:
### Checklist

### Related Issues/PRs

Resolves #4469  and partially resolves #4433 

Combined for convenience since one is needed for the gradient of the other, and vice versa

### Changes

- Add implementations for `scatter_nd` (Assign, Add, Mul, Min, Max) and `gather_nd` forward pass
  - Note that candle backend doesn't have native implementation for `scatter_nd` Mul, Min, Max. This implementation will raise an error if any of these are used
- Autodiff for `gather_nd` and `scatter_nd` (Assign, Add). Deferring Mul, Min, Max autodiff to another PR as that would be very meaty, and this PR is big as is.

### Testing

Unit tests for core functionality

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1125,8 +1125,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         _checkpointer: &mut Checkpointer,
                     ) {
                         let (indices, values_shape, device) = ops.state;
-                        let [indices_4lhs, indices_4rhs] =
-                            duplicate(&ops.parents, Some(indices));
+                        let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));
 
                         binary::<B, _, _>(
                             ops.parents,
@@ -1134,11 +1133,8 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                             grads,
                             |grad| {
                                 // Zero out the scattered positions in grad
-                                let zeros = B::float_zeros(
-                                    values_shape,
-                                    &device,
-                                    grad.dtype().into(),
-                                );
+                                let zeros =
+                                    B::float_zeros(values_shape, &device, grad.dtype().into());
                                 B::float_scatter_nd(
                                     grad,
                                     indices_4lhs.unwrap(),
@@ -1186,10 +1182,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
         }
     }
 
-    fn float_gather_nd(
-        data: FloatTensor<Self>,
-        indices: IntTensor<B>,
-    ) -> FloatTensor<Self> {
+    fn float_gather_nd(data: FloatTensor<Self>, indices: IntTensor<B>) -> FloatTensor<Self> {
         #[derive(Debug)]
         struct GatherNd;
 
@@ -1229,9 +1222,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                 ),
                 B::float_gather_nd(data.primitive, indices),
             ),
-            OpsKind::UnTracked(prep) => {
-                prep.finish(B::float_gather_nd(data.primitive, indices))
-            }
+            OpsKind::UnTracked(prep) => prep.finish(B::float_gather_nd(data.primitive, indices)),
         }
     }
 
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1054,6 +1054,187 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
         }
     }
 
+    fn float_scatter_nd(
+        data: FloatTensor<Self>,
+        indices: IntTensor<B>,
+        values: FloatTensor<Self>,
+        reduction: burn_backend::tensor::ScatterNdReduction,
+    ) -> FloatTensor<Self> {
+        use burn_backend::tensor::ScatterNdReduction;
+
+        match reduction {
+            ScatterNdReduction::Add => {
+                #[derive(Debug)]
+                struct ScatterNdAdd;
+
+                impl<B: Backend> Backward<B, 2> for ScatterNdAdd {
+                    type State = IntTensor<B>;
+
+                    fn backward(
+                        self,
+                        ops: Ops<Self::State, 2>,
+                        grads: &mut Gradients,
+                        _checkpointer: &mut Checkpointer,
+                    ) {
+                        let indices = ops.state;
+                        let [_, indices_4rhs] = duplicate(&ops.parents, Some(indices));
+
+                        binary::<B, _, _>(
+                            ops.parents,
+                            ops.node,
+                            grads,
+                            |grad| grad,
+                            |grad| B::float_gather_nd(grad, indices_4rhs.unwrap()),
+                        );
+                    }
+                }
+
+                match ScatterNdAdd
+                    .prepare::<C>([data.node, values.node])
+                    .compute_bound()
+                    .stateful()
+                {
+                    OpsKind::Tracked(prep) => prep.finish(
+                        indices.clone(),
+                        B::float_scatter_nd(
+                            data.primitive,
+                            indices,
+                            values.primitive,
+                            ScatterNdReduction::Add,
+                        ),
+                    ),
+                    OpsKind::UnTracked(prep) => prep.finish(B::float_scatter_nd(
+                        data.primitive,
+                        indices,
+                        values.primitive,
+                        ScatterNdReduction::Add,
+                    )),
+                }
+            }
+            ScatterNdReduction::Assign => {
+                #[derive(Debug)]
+                struct ScatterNdAssign;
+
+                impl<B: Backend> Backward<B, 2> for ScatterNdAssign {
+                    type State = (IntTensor<B>, Shape, B::Device);
+
+                    fn backward(
+                        self,
+                        ops: Ops<Self::State, 2>,
+                        grads: &mut Gradients,
+                        _checkpointer: &mut Checkpointer,
+                    ) {
+                        let (indices, values_shape, device) = ops.state;
+                        let [indices_4lhs, indices_4rhs] =
+                            duplicate(&ops.parents, Some(indices));
+
+                        binary::<B, _, _>(
+                            ops.parents,
+                            ops.node,
+                            grads,
+                            |grad| {
+                                // Zero out the scattered positions in grad
+                                let zeros = B::float_zeros(
+                                    values_shape,
+                                    &device,
+                                    grad.dtype().into(),
+                                );
+                                B::float_scatter_nd(
+                                    grad,
+                                    indices_4lhs.unwrap(),
+                                    zeros,
+                                    ScatterNdReduction::Assign,
+                                )
+                            },
+                            |grad| B::float_gather_nd(grad, indices_4rhs.unwrap()),
+                        );
+                    }
+                }
+
+                match ScatterNdAssign
+                    .prepare::<C>([data.node, values.node])
+                    .compute_bound()
+                    .stateful()
+                {
+                    OpsKind::Tracked(prep) => prep.finish(
+                        (
+                            indices.clone(),
+                            values.primitive.shape(),
+                            B::float_device(&data.primitive),
+                        ),
+                        B::float_scatter_nd(
+                            data.primitive,
+                            indices,
+                            values.primitive,
+                            ScatterNdReduction::Assign,
+                        ),
+                    ),
+                    OpsKind::UnTracked(prep) => prep.finish(B::float_scatter_nd(
+                        data.primitive,
+                        indices,
+                        values.primitive,
+                        ScatterNdReduction::Assign,
+                    )),
+                }
+            }
+            _ => panic!(
+                "scatter_nd backward is not supported for {:?} reduction. \
+                 Only Assign and Add support autograd. Use these reductions in \
+                 inference-only contexts or detach the tensor first.",
+                reduction,
+            ),
+        }
+    }
+
+    fn float_gather_nd(
+        data: FloatTensor<Self>,
+        indices: IntTensor<B>,
+    ) -> FloatTensor<Self> {
+        #[derive(Debug)]
+        struct GatherNd;
+
+        impl<B: Backend> Backward<B, 1> for GatherNd {
+            type State = (IntTensor<B>, Shape, B::Device);
+
+            fn backward(
+                self,
+                ops: Ops<Self::State, 1>,
+                grads: &mut Gradients,
+                _checkpointer: &mut Checkpointer,
+            ) {
+                let (indices, data_shape, device) = ops.state;
+
+                unary::<B, _>(ops.parents, ops.node, grads, |grad| {
+                    let zeros = B::float_zeros(data_shape, &device, grad.dtype().into());
+                    B::float_scatter_nd(
+                        zeros,
+                        indices,
+                        grad,
+                        burn_backend::tensor::ScatterNdReduction::Add,
+                    )
+                });
+            }
+        }
+
+        match GatherNd
+            .prepare::<C>([data.node])
+            .compute_bound()
+            .stateful()
+        {
+            OpsKind::Tracked(prep) => prep.finish(
+                (
+                    indices.clone(),
+                    data.primitive.shape(),
+                    B::float_device(&data.primitive),
+                ),
+                B::float_gather_nd(data.primitive, indices),
+            ),
+            OpsKind::UnTracked(prep) => {
+                prep.finish(B::float_gather_nd(data.primitive, indices))
+            }
+        }
+    }
+
     fn float_select(
         tensor: FloatTensor<Self>,
         dim: usize,
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1058,13 +1058,13 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
         data: FloatTensor<Self>,
         indices: IntTensor<B>,
         values: FloatTensor<Self>,
-        reduction: burn_backend::tensor::ScatterNdReduction,
+        reduction: burn_backend::tensor::IndexingUpdateOp,
     ) -> FloatTensor<Self> {
-        use burn_backend::tensor::ScatterNdReduction;
+        use burn_backend::tensor::IndexingUpdateOp;
 
         if matches!(
             reduction,
-            ScatterNdReduction::Mul | ScatterNdReduction::Min | ScatterNdReduction::Max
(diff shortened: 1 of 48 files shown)


crates/burn-cubecl/src/kernel/index/scatter_nd.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    let data_idx = base_offset + slice_offset;
+    let val_idx = ABSOLUTE_POS;
+
+    let result = Op::BinaryOp::<T, Const<1>>::execute(
+        Vector::cast_from(data[data_idx]),
+        Vector::cast_from(values[val_idx]),
+    );
+    data[data_idx] = result[0];
```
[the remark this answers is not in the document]
cu9hue: intentional. matching ONNX (and PyTorch, and others) semantics. this means that only unique indices guarantee determinism. duplicate indices are non-deterministic, but do not return an error

will update documentation to make this behavior more explicit though

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1810,6 +1810,130 @@ where
...
+    ///
+    /// `indices` is an M-dimensional integer tensor whose last dimension (K) indexes into the
+    /// first K dimensions of `self`. The `values` tensor has shape
+    /// `[indices.shape[0..M-1]..., self.shape[K..D]...]`.
+    ///
+    /// # Note
+    ///
+    /// When `indices` contains duplicate entries, the result is non-deterministic on GPU
+    /// backends (matching ONNX ScatterND semantics). For deterministic accumulation with
+    /// duplicates, use [`scatter_nd_add`](Self::scatter_nd_add) on CPU backends.
```
antimora: The existing `scatter` method above has a `# Warning` section about out-of-bounds indices possibly leading to unexpected results (since not all backends do runtime bounds checks). That same warning applies here, maybe even more so since the GPU kernel uses `launch_unchecked`. Could you add a similar warning to `scatter_nd` and `gather_nd`?

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1810,6 +1810,130 @@ where
...
+        Self::new(K::scatter_nd(
+            self.primitive,
+            indices.primitive,
+            values.primitive,
+            ScatterNdReduction::Add,
+        ))
+    }
+
+    /// Multi-dimensional scatter with multiplicative reduction.
+    pub fn scatter_nd_mul<const M: usize, const DV: usize>(
```
antimora: The `scatter_nd` and `scatter_nd_add` docs mention non-determinism with duplicate indices on GPU, but `scatter_nd_mul`, `scatter_nd_min`, and `scatter_nd_max` are missing this. They use the same non-atomic GPU kernel, so the same caveat applies.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1051,6 +1051,178 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                    ),
+                    OpsKind::UnTracked(prep) => prep.finish(B::float_scatter_nd(
+                        data.primitive,
+                        indices,
+                        values.primitive,
+                        ScatterNdReduction::Assign,
+                    )),
+                }
+            }
+            _ => panic!(
```
antimora: This panic happens during `backward()`, which can be a long time after the user called `scatter_nd_mul`. The forward pass succeeds, then the model crashes hours later during the first backward pass.

Worth considering: if the tensor requires grad and the reduction is Mul/Min/Max, could we panic here in the forward path instead? Failing fast and close to the call site would save users a lot of debugging time.

crates/burn-tensor/src/tensor/api/check.rs
```diff
@@ -910,6 +910,114 @@ impl TensorCheck {
...
+            check = check.register(
+                ops,
+                TensorError::new(format!(
+                    "Output rank DV={DV} does not match expected M-1+D-K = {expected_dv}"
+                )),
+            );
+        }
+
+        // Suppress unused warning
+        let _ = data_shape;
```
antimora: Minor: `data_shape` is passed in and then suppressed. `scatter_nd` uses it to validate slice dimension sizes (the `data_shape[k..]` check). Could `gather_nd` do the same for consistency? Not a blocker, just noticed the asymmetry.

crates/burn-backend-tests/tests/tensor/float/ops/gather_scatter_nd.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    let gathered: TestTensor<2> = data.clone().gather_nd(indices.clone());
+    let zeros = TestTensor::<2>::zeros([3, 3], &device);
+    let reconstructed: TestTensor<2> = zeros.scatter_nd_add(indices, gathered);
+
+    reconstructed.into_data().assert_eq(
+        &TensorData::from([[1.0, 2.0, 3.0], [0.0, 0.0, 0.0], [7.0, 8.0, 9.0]]),
+        false,
+    );
+}
```
antimora, line 297: Test coverage looks good for the happy paths. A few things that would make this stronger:

- A test with duplicate indices for `scatter_nd_add` (verifying the values accumulate correctly on CPU)
- A test for K=2 on 3D data (partial slice gather/scatter, the intermediate case between K=1 and K=D)
- An autodiff test with K=D (scalar-level indexing) since the current autodiff tests all use K=1

These are the untested code paths I can see.

antimora (commented):
Nice work on this! The algorithms are correct, the autodiff backward passes check out mathematically, and the implementation is consistent across all backends. A few things I noticed below.

antimora (changes requested):

crates/burn-autodiff/src/ops/tensor.rs
cu9hue: great idea, will do

crates/burn-tensor/src/tensor/api/check.rs
cu9hue: I think this one is inherently assymetric. `gather_nd`'s output shape is computed from data_shape internally, unlike `scatter_nd`'s user-supplied output shape. There is nothing to validate against it. The data_shape parameter is for the signature consistency

antimora (approved):
Thanks for the thorough updates in the latest commit - docs, warnings, and test coverage all look good now.

One item still open: the fail-fast for unsupported autodiff reductions (Mul/Min/Max). Right now the panic fires during `backward()`, which can be a long time after the user called `scatter_nd_mul`. Moving the check to the forward path so it fails close to the call site would save users debugging time. You mentioned you'd do this - just a reminder.

Approving from my side. @laggui should still review the CubeCL kernel implementation.

laggui:
Will try to take a look this afternoon! At the latest Tuesday as I won't be available on Easter Monday :slightly_smiling_face: 

Thank you both!

cu9hue:
I've added the fail-fast check, it's `crates/burn-autodiff/src/ops/tensor.rs`, L1062-1072

I believe it's the earliest place where we know both the reduction type and whether gradient is tracked :)

crates/burn-backend/src/backend/ops/int_tensor.rs
```diff
@@ -187,6 +187,21 @@ pub trait IntTensorOps<B: Backend> {
...
     ) -> IntTensor<B>;
 
+    /// Multi-dimensional scatter for int tensors.
+    fn int_scatter_nd(
+        _data: IntTensor<B>,
+        _indices: IntTensor<B>,
+        _values: IntTensor<B>,
+        _reduction: crate::tensor::ScatterNdReduction,
+    ) -> IntTensor<B> {
+        panic!("int_scatter_nd is not implemented for this backend")
```
laggui: Small nit: I would use `unimplemeted!` instead if we add these default implementations

crates/burn-backend/src/tensor/ops/mod.rs
```diff
@@ -19,3 +19,18 @@ pub enum IndexingUpdateOp {
...
+    Assign,
+    /// Add values to existing.
+    Add,
+    /// Multiply existing values.
+    Mul,
+    /// Take element-wise minimum.
+    Min,
+    /// Take element-wise maximum.
+    Max,
+}
```
laggui: Why not use the same enum as other indexing ops (`IndexingUpdateOp`)?

I don't think we should have separate enums just because the current support is not the same.

crates/burn-cubecl/src/kernel/index/gather_nd.rs
```diff
@@ -0,0 +1,98 @@
...
+#[cube(launch_unchecked, address_type = "dynamic")]
+fn gather_nd_kernel<T: Numeric, I: Int>(
+    data: &Tensor<T>,
+    indices: &Tensor<I>,
+    output: &mut Tensor<T>,
+    data_strides: &Tensor<u32>,
+    slice_size: usize,
+    k: usize,
+    #[define(T, I)] _dtypes: [StorageType; 2],
+) {
```
laggui, line 25: We shouldn't have a dedicated tensor to handle the strides manually. You can use `data.stride(dim)` instead, and for the indices access pattern we could use `LinearView` (for `indices` and `output`).

You can take a look at the other cubecl kernels is this module for examples.

crates/burn-cubecl/src/kernel/index/scatter_nd.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// scatter_nd GPU kernel.
+///
+/// Each thread handles one element across all update slices.
+/// Work items = num_updates * slice_size.
+///
+/// `data_strides` is a 1D tensor of length K holding the strides for converting
+/// K-dimensional index tuples into flat offsets.
+#[cube(launch_unchecked, address_type = "dynamic")]
+fn scatter_nd_kernel<T: Numeric, I: Int, Op: BinaryOpFamily>(
```
laggui, line 19: Same comment as the gather_nd kernel

crates/burn-tensor/src/tensor/api/check.rs
laggui: If it's not actually used for the tensor check, we can simply remove it from `TensorCheck::gather_nd`

laggui (changes requested):
Thanks for contributing the implementations!

Overall great job, I only have a couple of comments

crates/burn-backend/src/tensor/ops/mod.rs
cu9hue: will do. my main concern was the blast radius of merging these together

laggui:

CI is failing now that you've updated to the main branch, as it now runs burn-flex tests (which doesn't implement the gather nd op).

The cubecl implementation already looks much better. I think we could change `k` to a `#[comptime]` and unroll the loop. And might have to use launch checked in case of OOB indices (that might also have to be changed for our existing kernels anyway, so unrelated).

/edit: whoops, accidentally clicked close instead of comment :sweat_smile: 

I checked locally and there is only one failing test (cuda only):
```
failures:

---- tensor::float::ops::gather_scatter_nd::test_scatter_nd_k1_high_rank stdout ----

thread 'tensor::float::ops::gather_scatter_nd::test_scatter_nd_k1_high_rank' (56541) panicked at crates/burn-backend-tests/tests/common/../tensor/float/ops/gather_scatter_nd.rs:261:24:
Tensors are not eq:
  => Position 6: 0 != 30
  => Position 7: 0 != 40
```

Will come back around to this a bit later.

Great work though :pray:

cu9hue:
no worries, I noticed that I missed burn-flex this morning. should be good now in CI/CD :)

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1821,6 +1821,163 @@ where
...
+    ///
+    /// `indices` is an M-dimensional integer tensor whose last dimension (K) indexes into the
+    /// first K dimensions of `self`. The `values` tensor has shape
+    /// `[indices.shape[0..M-1]..., self.shape[K..D]...]`.
+    ///
+    /// # Note
+    ///
+    /// When `indices` contains duplicate entries, the result is non-deterministic on GPU
+    /// backends (matching ONNX ScatterND semantics). For deterministic accumulation with
+    /// duplicates, use [`scatter_nd_add`](Self::scatter_nd_add) on CPU backends.
```
antimora: The wording "use `scatter_nd_add` on CPU backends" implies the *choice of reduction* is what provides determinism. On GPU, `scatter_nd_add` uses the same non-atomic RMW as the other reductions (see `burn-cubecl/src/kernel/index/scatter_nd.rs`), so it's equally non-deterministic with duplicates there. The actual guarantee is "CPU backends are deterministic regardless of reduction." Suggest rewording to something like: `For deterministic accumulation with duplicates, run on a CPU backend.`

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1821,6 +1821,163 @@ where
         ))
     }
 
+    /// Multi-dimensional scatter: update `self` at locations given by `indices` using assignment.
+    ///
+    /// `indices` is an M-dimensional integer tensor whose last dimension (K) indexes into the
```
antimora: Minor doc nit: "last dimension (K)" conflates the *size* of the last axis with a *count of leading data dims*. K is the value of `indices.shape[M-1]`, not an axis index, so bracket notation like `self.shape[K..D]` reads ambiguously on first pass. Something like `The size of indices's last axis (call it K) indexes the leading K dims of self; the batch shape indices.shape[0..M-1] is preserved. values has shape indices.shape[0..M-1] ++ self.shape[K..D].` avoids the ambiguity.

While here, worth explicitly stating the constraints `K <= D` and `M >= 1` (currently only enforced inside `TensorCheck`, users reading rustdoc have to infer them from the example shapes).

crates/burn-candle/src/ops/base.rs
```diff
@@ -82,6 +82,122 @@ pub fn empty(shape: Shape, device: &CandleDevice, dtype: candle_core::DType) ->
...
+    let slice_offsets = candle_core::Tensor::arange(0i64, slice_size as i64, device).unwrap();
+    base_offsets
+        .unsqueeze(1)
+        .unwrap()
+        .broadcast_add(&slice_offsets.unsqueeze(0).unwrap())
+        .unwrap()
+        .reshape((num_updates * slice_size,))
+        .unwrap()
+        .to_dtype(candle_core::DType::U32)
+        .unwrap()
```
antimora: Not a correctness issue, but `flatten_nd_indices` has ~15 `.unwrap()`s in a row. Any candle error (dtype/device mismatch, OOM, shape issue) panics with whatever internal message candle produces, no `scatter_nd` / `gather_nd` context. A user seeing the panic won't know it came from this op. Wrapping the chain with `.map_err(|e| ...)` and bubbling up a `"candle scatter_nd: {e}"`-style message would make failures much easier to diagnose. Same applies to the `.unwrap()`s in `scatter_nd`/`gather_nd` below.

antimora (commented):
Ran a pass over the latest PR head. Most of what I found is already covered by laggui, and Copilot - agreeing with those rather than restating:

- Agree with Copilot's underflow concern in `check.rs` when `k > D`. One additional edge case in the same functions: `indices_shape[M - 1]` at lines 928 / 994 will also underflow if `M == 0` (a scalar indices tensor). Same fix shape: early-return after registering a rank-validation error.
- Adding the OOB-indices warning to the public docs and on moving the autograd Mul/Min/Max check to the forward path.
- Agree with laggui on using `IndexingUpdateOp` rather than a parallel enum, and on `data.stride(j)` + `LinearView` for the cubecl kernels (already applied in the latest commits - nice cleanup).

A few items I didn't see raised elsewhere, left as inline comments.

antimora (approved):
It seems you have addressed all feedback.

I am approving your PR.

Please rebase to the latest and resolve any conflicts.

laggui (approved):
Sorry for the delay, I just had a bit of time to investigate the scatter nd issue that was failing.

I've updated the implementations for both gather/scatter nd to be stride-aware, all tests are passing locally on vulkan, wgpu and cuda (f32 and f16) now.

To align with the `scatter` API, I also changed the tensor method to `tensor.scatter_nd(indices, values, update)` and removed the other duplicates.

Thanks again for the great work 🙏

PR #4717, opened 2026-04-02
[Breaking] Remove `Tensor` backend generic and add high-level `Device` struct

### LINKED ISSUE #3628: Default backend implementation

### LINKED ISSUE #4750: [Breaking] Remove backend generics from high-level API
Now that we have a `Dispatch` backend (enum static dispatch) we can remove the backend generics from the high-level `Tensor` API to use this as the default backend.

Example module:
```rust
#[derive(Module, Debug)]
pub struct BatchNorm {
    /// The learnable weight gamma.
    pub gamma: Param<Tensor<1>>,
    /// The learnable weight beta.
    pub beta: Param<Tensor<1>>,
    /// The running mean.
    pub running_mean: RunningState<Tensor<1>>,
    /// The running variance.
    pub running_var: RunningState<Tensor<1>>,
    /// Momentum used to update the metrics.
    pub momentum: f64,
    /// A value required for numerical stability.
    pub epsilon: f64,
}
```

Although a big breaking change, it should remove a lot of friction with the current API, and allow for easy runtime backend selection.

This also means a CPU device can always be available (unless disabled), making host-side tensor manipulations easily accessible. Beforehand, one would have to explicitly enable another backend (e.g. ndarray) and move the data between the "primary" backend -> tensor data -> ndarray which was annoying.

This change should also bring a new high-level device struct to handle the default precision settings and device operations (e.g. `device.sync()`).

laggui:
### Checklist

### Changes

- **Remove `B` from `Tensor`:** since `Dispatch` is the default backend for user-facing code, `Tensor<B, D, K>` becomes `Tensor<D, K>`.

```rust
let device: Device = CudaDevice::default().into(); // cuda device
let device = Device::default(); // default device

let x = Tensor::<2>::zeros([3, 4], &device);
```

Example module:
```rust
#[derive(Module, Debug)]
pub struct BatchNorm {
    /// The learnable weight gamma.
    pub gamma: Param<Tensor<1>>,
    /// The learnable weight beta.
    pub beta: Param<Tensor<1>>,
    /// The running mean.
    pub running_mean: RunningState<Tensor<1>>,
    /// The running variance.
    pub running_var: RunningState<Tensor<1>>,
    /// Momentum used to update the metrics.
    pub momentum: f64,
    /// A value required for numerical stability.
    pub epsilon: f64,
}
```

The `DispatchDevice` enum dispatches based on which Cargo feature flags are enabled (`cuda`, `ndarray`, `vulkan`, etc.). When only one backend feature is enabled the compiler optimizes the `match` away entirely; with multiple backends enabled the overhead is minimal enum dispatch.

- **Add high-level `Device`:** the compute backend is now selected via the device, which can be `sync`d and `seed`ed, among other operations. Autodiff is now a property of the `Device`. Users can call `.autodiff()` on any device to opt into gradient tracking.

```rust
let device = Device::default().autodiff(); // enables gradient tracking
device.seed(seed);
device.sync();
```

The default data types for tensors created on a device can be configured and checked:

```rust
device.set_default_dtypes(DType::F16, DType::I32).expect("Should support data type");

let settings = device.settings();
settings.float_dtype
```

- **Add `#[backend_extension]` proc macro:** users can still extend supported backends as before, but should mark their backend extension trait so dispatch can correctly forward the calls.

```rust
/// We create our own Backend trait that extends the Burn backend trait.
#[backend_extension(Autodiff, Wgpu)]
pub trait Backend: burn::tensor::backend::Backend {
    fn fused_matmul_add_relu(
        lhs: FloatTensor<Self>,
        rhs: burn::tensor::ops::FloatTensor<Self>,
        bias: <Self as burn::tensor::backend::Backend>::FloatTensorPrimitive,
    ) -> FloatTensor<Self>;
}
```

The backends can also be feature gated if user-code has feature flags, e.g.:
```rust
#[backend_extension(
    Flex: cfg(feature = "flex"),
    Wgpu: cfg(feature = "wgpu"),
    Cuda: cfg(feature = "cuda"),
)]
/// Vision ops on float tensors
pub trait FloatVisionOps: Backend {
    /// Perform Non-Maximum Suppression on bounding boxes.
    fn nms(
        boxes: FloatTensor<Self>,
        scores: FloatTensor<Self>,
        options: NmsOptions,
        out_dtype: IntDType,
    ) -> IntTensor<Self> {
        // ...
    }
}
```

- **Remove `burn-collective`:** collectives have been refactored for the distributed backend, this crate was no longer up-to-date and its interface was between the backend API and high-level API.

- **Feature gate `burn-optim`:** optim requires autodiff, which is not supported on all targets. So I moved it behind a new `optim` feature flag (enabled by default). This fixes no-std builds.

### Related Issues/PRs

- Closes #4750
- Related:
  - #3628
  - #4415
- Preparatory:
  - #4508
  - #4629
  - #4653
  - #4666
- Follow-up: #4752 #4934

TODO:
- move `TensorKind` back to the high-level API
- validate burn-dispatch distributed 
- burn-remote
- configurable default device priority (currently hardcoded, feature based)
- third-party backend integrations

### Testing

Unit tests and examples


crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -35,14 +34,13 @@ use serde::{Serialize, Serializer};
...
 /// use burn_tensor::Tensor;
 /// use burn_tensor::Int;
 ///
-/// fn example<B: Backend>() {
+/// fn example() {
 ///     let device = Default::default();
 ///
-///     let tensor = Tensor::<B, 2>::from_data(
+///     let tensor = Tensor::< 2>::from_data(
 ///         [
```
antimora, line 47: Stray space inside the angle brackets: `Tensor::< 2>` is invalid Rust. The `B,`-strip regex left a leading space behind. This pattern shows up 222 times under `crates/burn-tensor/src/` (verified via `grep -c 'Tensor::< '`). Most are in compiled `rust` blocks, so `cargo test --doc` will fail.

```sh
sed -i '' 's/Tensor::< /Tensor::</g' $(grep -rl 'Tensor::< ' crates/burn-tensor/src/)
```

crates/burn-nn/src/modules/conv/conv2d.rs
```diff
@@ -153,14 +154,14 @@ impl<B: Backend> Conv2d<B> {
     ///
     /// // Assuming backend type alias `B`
     /// let device = Default::default();
-    /// let conv = Conv2dConfig::new([3, 8], [3, 3]).init::<B>(&device);
+    /// let conv = Conv2dConfig::new([3, 8], [3, 3]).init::(&device);
     ///
```
antimora, line 158: Empty turbofish from the same `B,`-strip rewrite as the `Tensor::< N>` cluster. Should be `init(&device)` (or `init::<Backend>(&device)` if the example still needs a generic). The doctest is `rust,ignore` so CI won't catch it, but this is the public example for `Conv2d::init`.

crates/burn-tensor/src/device.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// // x.backward() is now available
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics if autodiff is already enabled on this device.
+    #[cfg(feature = "autodiff")]
+    pub fn autodiff(mut self) -> Self {
+        match self.dispatch {
+            DispatchDevice::Autodiff(_) => unimplemented!("Only first-order autodiff is supported"),
```
antimora, line 127: `unimplemented!` reads as `feature not yet implemented`. The intent here is `this composition is invalid; autodiff is already enabled` (the doc above says exactly that). `panic!("autodiff is already enabled on this device")` matches the documented contract and won't lead users to wait for a future release that lets them stack autodiff wrappers.

antimora (commented):
Few things worth pulling out:

- 222 doctests in `burn-tensor` have `Tensor::< N>` with a stray space (regex scar from stripping `B,`). Most are in compiled `rust` blocks, so `cargo test --doc` will fail. `grep -rl 'Tensor::< ' crates/burn-tensor/src/` to locate them.
- `conv2d.rs:158` has the matching `init::(&device)` empty turbofish from the same rewrite. Grepping `::(` across the changed files turns up no other survivors.
- The `Tensor` struct doc at `crates/burn-tensor/src/tensor/api/base.rs:28` still reads `A tensor with a given backend, shape and data type.` That's the old type model and the first line of the most user-facing rustdoc.
- Removing `B` from `Tensor` turns three compile-time invariants into runtime panics: backend identity, autodiff status, dtype. Probably the right ergonomic call, but `Tensor<D, K>` rustdoc doesn't acknowledge it. A short `Constraints checked at runtime` paragraph on the struct would save users a round of confusion.
- `Device::set_default_dtypes(&self, ...)` mutates the process-wide registry through `&self`. The self-TODO at `device.rs:290` flags it. Either resolve to `&mut self` or document that the lock is global and one-shot, since the PR description's own example `.expect("Should support data type")` mislabels the `AlreadyInitialized` error.
- `burn-backend-extension` has no module-level `//!` doc and the README is one paragraph; the macro example doesn't show the required `use` block, so users can't copy-paste it. `burn-cpu/src/lib.rs` also has no `//!`.

Pre-existing in burn-dispatch (not introduced here, but came up while tracing the new code paths and worth a follow-up):
- `binary_float_arms!(Float, Float)` at `crates/burn-dispatch/src/macros.rs:786` has no same-backend mixed-kind arm. `let g = ad_t.inner(); ad_t + g` falls through to the `not on the same backend` panic even though both tensors are on the same physical backend.
- `expect("Should read data")` in `crates/burn-dispatch/src/ops/{tensor,int_tensor,bool_tensor,qtensor}.rs` swallows real backend errors on cross-device transfers. CUDA OOM and wgpu adapter loss surface as the same opaque message.
- Three bare `unimplemented!()` in `crates/burn-dispatch/src/ops/distributed.rs:134,155,159`. Anyone going through `Device` to DDP will hit a no-message panic.

crates/burn-tensor/src/device.rs
laggui: Actually what you're suggesting applies to `todo!()`, not `unimplemented!()`:

> The difference between `unimplemented!` and `todo!` is that while `todo!` conveys an intent of implementing the functionality later and the message is "not yet implemented", `unimplemented!` makes no such claims. Its message is "not implemented".

laggui:
Thanks for catching these! I fixed the doc issues and added a small note on the `Tensor` struct for runtime-selected device and backend.

> * `Device::set_default_dtypes(&self, ...)` mutates the process-wide registry through `&self`. The self-TODO at `device.rs:290` flags it. Either resolve to `&mut self` or document that the lock is global and one-shot, since the PR description's own example `.expect("Should support data type")` mislabels the `AlreadyInitialized` error.

I wasn't sure yet about this method, but yeah I think `&mut self` makes more sense. Also, the short example with `.expect("Should support data type")` was for `UnsupportedDType` error.

For other follow-up notes it could be worth mentioning in #4934 which is the meta-issue for follow-up work related to this PR.

laggui:
Macos CI failure is unrelated, seems to happen intermittently in the past few days where a dependency fails to compile with their custom build script.

```
  error: failed to run custom build command for `libsqlite3-sys v0.35.0`
  
  Caused by:
    process didn't exit successfully: `/Users/admin/actions-runner/_work/burn/burn/target/release/build/libsqlite3-sys-d3e09acbf05b5f80/build-script-build` (exit status: 126)
    --- stderr
    /Users/admin/actions-runner/_work/burn/burn/target/release/build/libsqlite3-sys-d3e09acbf05b5f80/build-script-build: /Users/admin/actions-runner/_work/burn/burn/target/release/build/libsqlite3-sys-d3e09acbf05b5f80/build-script-build: cannot execute binary file
```

It also happened on this [CI job from main](https://github.com/tracel-ai/burn/actions/runs/25748152418/job/75618207204).

The linux-std runner successfully executes all tests but it looks like it might be running out of resources for code coverage? From raw logs:

```
2026-05-13T15:30:49.6689356Z [1m[92m     Running[0m `target/xtask/debug/xtask coverage generate --ignore '/*,xtask/*,examples/*' --profile release`
2026-05-13T15:30:49.6816034Z Environment: development
2026-05-13T15:30:49.6816736Z Context: std
2026-05-13T15:30:49.6817470Z ##[group]Grcov
2026-05-13T15:30:49.6819661Z Command line: grcov . --binary-path ./target/release/ -s . -t lcov -o lcov.info --branch --ignore-not-existing --ignore /* --ignore xtask/* --ignore examples/*
2026-05-13T15:34:29.7877841Z ##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
2026-05-13T15:34:30.8085145Z ##[error]The operation was canceled.
2026-05-13T15:34:30.8838836Z Cleaning up orphan processes
```

Looks like the process is being killed.

PR #4721, opened 2026-04-02
Add dtype-aware numerical precision info (finfo equivalent)

### LINKED ISSUE #4367: DType-aware epsilon handling
Some operations, modules or optimizers use constants for numerical stability.

But epsilon is not scale-free, and a single literal like `1e-5` silently becomes wrong once you change dtype.

Related: https://github.com/tracel-ai/burn/issues/4235

We could introduce a `DType` stability eps, e.g. `dtype.default_eps()` which users could still explicitly override in configs (e.g. `AdamConfig`).

Originated from [this discord convo](https://discord.com/channels/1038839012602941528/1059209073784008835/1463616390505304125).

### LINKED ISSUE #4700: Add dtype-aware numerical precision info (finfo equivalent)
## Summary

Burn lacks a way to query numerical precision constants (epsilon, min, max, smallest normal) for a given float dtype. This forces code to hardcode magic constants for numerical stability, which may not be appropriate across all precisions (f16, bf16, f32, f64).

This is needed at two levels:
1. **Backend/tensor level** - for ops like `attention_fallback` that work with `FloatTensor<B>`
2. **Element/kernel level** - for CubeCL kernels and similar code that operates on individual elements of type `E` and needs `E::epsilon()` or `E::tiny()` directly

## Motivation

In #4694 (PR #4697), fixing NaN in `attention_fallback` required hardcoded clamp constants (`-1e4`, `1e-6`). These work for f32 but are not derived from dtype properties. Other places in the codebase also hardcode precision-dependent values:

- All norm modules use `epsilon = 1e-5`
- Cosine similarity uses `DEFAULT_EPSILON = 1e-8`
- Softmax activation uses max-normalization without a dtype-aware floor

CubeCL kernels face the same problem at the element level. For example, a `recip_inplace` that needs to safely handle zero-valued rows (from fully-masked attention) must hardcode a threshold constant:

```rust
pub fn recip_inplace(&mut self) {
    for i in 0..self.num_rows {
        let row_val = self.vals[i];
        // Hardcoded threshold - should be E::tiny() or E::epsilon()
        let epsilon = E::new(FULLY_MASKED_ROW_THRESHOLD);
        let not_masked = E::cast_from(row_val >= epsilon);
        let safe_val = clamp_min(row_val, epsilon);
        let recip = safe_val.recip();
        self.vals[i] = not_masked * recip;
    }
}
```

With dtype-aware constants, this would use `E::tiny()` or `E::epsilon()` instead of a hardcoded value, automatically adapting to f16/bf16/f32/f64.

## Prior art

All major ML frameworks expose this via a `finfo(dtype)` API modeled after NumPy:

- **PyTorch**: `torch.finfo(dtype).eps`, `.min`, `.max`, `.tiny` (smallest normal)
- **JAX**: `jnp.finfo(dtype)` (same fields)
- **TensorFlow**: `tf.experimental.numpy.finfo(dtype)`

Common usage patterns:
```python
# Masked fill with dtype-aware minimum
attn = attn.masked_fill(mask, torch.finfo(attn.dtype).min)

# Dtype-aware epsilon for numerical stability
norm = x / (x.std(dim=-1, keepdim=True) + torch.finfo(x.dtype).eps)
```

### Element level (for kernels)

Add associated constants or methods to `FloatElement` / CubeCL's numeric element trait:

```rust
trait FloatElement: Element {
    /// Machine epsilon: smallest value where 1.0 + eps != 1.0
    const EPSILON: Self;
    /// Largest representable finite value
    const MAX: Self;
    /// Most negative representable finite value
    const MIN: Self;
    /// Smallest positive normal value
    const TINY: Self;
}
```

This lets kernel code write `E::EPSILON` or `E::TINY` and get the right value for whatever precision the kernel is instantiated with.

### Backend/tensor level

Add a `FloatDTypeInfo` struct accessible from `FloatDType` or `Backend`:

```rust
pub struct FloatDTypeInfo {
    pub eps: f64,
    pub max: f64,
    pub min: f64,
    pub tiny: f64,
}
```

This could be queryable via `DeviceSettings` or a method on `FloatDType`. The exact API design is open for discussion.

## Impact

Would allow numerical stability code across the codebase (attention, normalization, activation functions, CubeCL kernels) to use dtype-appropriate constants instead of hardcoded values, automatically adapting to different precisions.

antimora:
## Summary

Adds dtype-aware numerical precision info, equivalent to NumPy's `finfo` / PyTorch's `torch.finfo`, and replaces hardcoded constants that silently underflow to zero on f16/bf16.

Closes #4700
Closes #4367

### New API

**Runtime query** (`burn-std`): `FloatDTypeInfo` struct and `finfo()` on both `FloatDType` and `DType`:

```rust
let info = DType::F16.finfo().unwrap();
// info.eps   - machine epsilon (smallest value where 1.0 + eps != 1.0)
// info.max   - largest representable finite value
// info.min   - most negative representable finite value
// info.tiny  - smallest positive normal value
```

**Compile-time generic** (`burn-backend`): `FloatElementPrecision` trait for generic Rust code and CubeCL kernels:

```rust
fn stabilize<E: FloatElementPrecision>(x: E) {
    // E::EPSILON, E::TINY available as associated constants
}
```

Implemented for: `f32`, `f64`, `f16`, `bf16`, `flex32`.

### Replaced hardcoded constants

These constants underflow to zero on f16 (`MIN_POSITIVE` ~6.1e-5), silently defeating the numerical stability they provide:

| File | Before | After | Purpose |
|------|--------|-------|---------|
| `attention.rs` | `SOFTMAX_MAX_FLOOR = -1e4` | `finfo.min` | Prevent `-inf - (-inf) = NaN` |
| `attention.rs` | `SOFTMAX_SUM_EPS = 1e-6` | `finfo.tiny` | Prevent `0 / 0 = NaN` |
| `kldiv.rs` | `epsilon = 1e-8` | `finfo.tiny` | Prevent `log(0)` |
| `cosine_similarity.rs` | `DEFAULT_EPSILON = 1e-8` | `finfo.tiny` | Prevent division by zero |
| `grad_clipping.rs` | `1e-6` | `finfo.tiny` | Prevent division by zero |

### Not changed (intentionally)

- **Config defaults** (norm modules `1e-5`, optimizer epsilon `1e-5`/`1e-8`): user-overridable, match PyTorch defaults
- **`DEFAULT_RTOL`/`DEFAULT_ATOL`**: match NumPy's `is_close` spec
- **Metric constants** (FID, PSNR, DISTS, Dice): run on CPU f64, not dtype-sensitive

## Test plan

- [x] 8 unit tests covering all float dtypes, `DType` delegation, non-float returns `None`, and cross-dtype invariants
- [x] `cargo check` passes for burn-std, burn-backend, burn-tensor, burn-nn, burn-optim
- [x] `cargo check --features cubecl` passes (flex32 impl)
- [x] All existing tests pass (grad_clipping, cosine_similarity, kldiv, burn-backend-tests)

### DIFF
--- a/crates/burn-backend/src/element/base.rs
+++ b/crates/burn-backend/src/element/base.rs
@@ -95,44 +95,6 @@ pub trait ElementLimits {
     const MAX: Self;
 }
 
-/// Precision properties for floating-point element types.
-///
-/// Provides associated constants analogous to NumPy's `finfo` fields,
-/// usable in generic Rust code and CubeCL kernels.
-pub trait FloatElementPrecision: Element {
-    /// Machine epsilon: smallest value such that `1.0 + EPSILON != 1.0`.
-    const EPSILON: Self;
-    /// Smallest positive normal value (equivalent to `tiny` in NumPy/PyTorch).
-    const TINY: Self;
-}
-
-impl FloatElementPrecision for f64 {
-    const EPSILON: Self = f64::EPSILON;
-    const TINY: Self = f64::MIN_POSITIVE;
-}
-
-impl FloatElementPrecision for f32 {
-    const EPSILON: Self = f32::EPSILON;
-    const TINY: Self = f32::MIN_POSITIVE;
-}
-
-impl FloatElementPrecision for f16 {
-    const EPSILON: Self = f16::EPSILON;
-    const TINY: Self = f16::MIN_POSITIVE;
-}
-
-impl FloatElementPrecision for bf16 {
-    const EPSILON: Self = bf16::EPSILON;
-    const TINY: Self = bf16::MIN_POSITIVE;
-}
-
-#[cfg(feature = "cubecl")]
-impl FloatElementPrecision for flex32 {
-    // Flex32 computes at reduced precision; use f16 limits for safety.
-    const EPSILON: Self = flex32::from_f32(f16::EPSILON.to_f32_const());
-    const TINY: Self = flex32::from_f32(f16::MIN_POSITIVE.to_f32_const());
-}
-
 /// Macro to implement the element trait for a type.
 #[macro_export]
 macro_rules! make_element {
--- a/crates/burn-nn/src/loss/kldiv.rs
+++ b/crates/burn-nn/src/loss/kldiv.rs
@@ -94,7 +94,7 @@ impl KLDivLoss {
                     .dtype()
                     .finfo()
                     .unwrap_or(burn::tensor::FloatDType::F32.finfo())
-                    .tiny;
+                    .min_positive;
                 let log_target = targets.clone().clamp(epsilon, 1.0).log();
                 targets.mul(log_target.sub(predictions))
             }
--- a/crates/burn-optim/src/grad_clipping/base.rs
+++ b/crates/burn-optim/src/grad_clipping/base.rs
@@ -75,12 +75,12 @@ impl GradientClipping {
         threshold: f32,
     ) -> Tensor<B, D> {
         let norm = Self::l2_norm(grad.clone());
-        let tiny = grad
+        let min_positive = grad
             .dtype()
             .finfo()
             .unwrap_or(burn::tensor::FloatDType::F32.finfo())
-            .tiny;
-        let clip_coef = threshold / norm.add_scalar(tiny);
+            .min_positive;
+        let clip_coef = threshold / norm.add_scalar(min_positive);
         let clip_coef_clamped = clip_coef.clamp_max(1.0);
         grad.mul(clip_coef_clamped.unsqueeze())
     }
--- a/crates/burn-std/src/tensor/dtype.rs
+++ b/crates/burn-std/src/tensor/dtype.rs
@@ -129,7 +129,7 @@ impl DType {
     /// Returns float precision info if this is a float dtype, `None` otherwise.
     ///
     /// Analogous to `torch.finfo(dtype)` or `numpy.finfo(dtype)`.
-    pub const fn finfo(&self) -> Option<FloatDTypeInfo> {
+    pub const fn finfo(&self) -> Option<FloatInfo> {
         match self {
             DType::F64 => Some(FloatDType::F64.finfo()),
             DType::F32 => Some(FloatDType::F32.finfo()),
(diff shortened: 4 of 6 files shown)


crates/burn-tensor/src/tensor/linalg/cosine_similarity.rs
```diff
@@ -28,7 +26,10 @@ pub fn cosine_similarity<B: Backend, const D: usize>(
     dim: i32,
     eps: Option<B::FloatElem>,
 ) -> Tensor<B, D> {
-    let eps = eps.unwrap_or_else(|| B::FloatElem::from_elem(DEFAULT_EPSILON));
+    let eps = eps.unwrap_or_else(|| {
+        let tiny = x1.dtype().finfo().expect("float tensor").tiny;
```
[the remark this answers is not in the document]
antimora: Good catch. Fixed in 27ea755 - now falls back to f32 finfo when dtype is QFloat, since quantized tensors dequantize to the device's default float type (typically f32) before any math operation. This avoids the panic without needing to dequantize just to read the dtype.

crates/burn-optim/src/grad_clipping/base.rs
```diff
@@ -75,7 +75,8 @@ impl GradientClipping {
         threshold: f32,
     ) -> Tensor<B, D> {
```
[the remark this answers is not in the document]
antimora: Fixed in 27ea755. Gradients are never quantized in practice, but added a fallback to f32 finfo for safety. This is simpler than dequantizing just to query the dtype.

crates/burn-nn/src/loss/kldiv.rs
```diff
@@ -90,7 +90,7 @@ impl KLDivLoss {
         match self.log_target {
             true => targets.clone().exp().mul(targets.sub(predictions)),
             false => {
-                let epsilon = 1e-8;
+                let epsilon = targets.dtype().finfo().expect("float tensor").tiny;
```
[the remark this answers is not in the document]
antimora: Fixed in 27ea755 with the same pattern - falls back to f32 finfo for QFloat dtypes. This is lighter than dequantizing the tensor just to read its dtype.

swfsql:
@antimora just wanted to link to some tests I was doing with epsilon and other small values over discord, made on f16 which is pretty harsh with the range. I'm not certain of this, but it may be preferable to not fallback to very small numbers such as tiny and epsilon directly for the actual operations.  
In that message I try a rule-of-thumb formula, but still different places, in practice, would prefer to skew the eps value -- so I don't think there is a catch-all value/formula.

---

Inline [message](https://discord.com/channels/1038839012602941528/1059209073784008835/1444701505461555301):

> I think that a good practice is to at least let the number be a positive non-subnormal one (grok said some archs flush f16 subnormals, so it could flush back to zero). Is there a way to reach those values from the Backend trait? (ideally from `type FloatElement: Element`)
> 
> Actually, dividing by the smallest positive normal value may explode the result to inf, which can be bad also. Maybe a good rule of thumb for eps on division is doing some operation on the minimum exponents (for f16 it is -14, for f32 it is -126, for f64 it is -1022).
> Maybe something like `stable_eps = 10^(-(-exp_min * 2.3)^0.35)` could work. And making some log-space average of that to epsilon (which takes the mantissa into account) might work too.
> 
> Something like this (pseudo):
> ```rust
> fn stable_epsilon<T: Float>() -> T {
>     let exp = T::MIN_EXP as f64;
>     let raw = -(-exp * 2.3).powf(0.35);
>     let mach = (T::EPSILON as f64).log10();
>     let exponent = (raw + mach) / 2.0;
>     T::from_f64(10f64.powf(exponent))
> }
> ```
> 
> This results in `stable_eps` :
> - `f16`: `7.12e-4`
> - `bf16`: `2.09e-5`
> - `f32`: `8.16e-8`
> - `f64`: `4.07e-16`
> - `f128`: `1.45e-37`

---

In hindsight, I think having a formula for those numbers is not really necessary, I was just testing if it was somewhat possible. And ofc, those exact resulting values don't really matter.

antimora:
Thanks for sharing the research, @swfsql. You raise a fair point about `tiny` being aggressive in division contexts (e.g., `1/f16::tiny` ≈ 16384, close to f16 max).

For this PR, the main goal was eliminating hardcoded constants like `1e-8` that silently underflow to zero on f16/bf16. Using `tiny` from `finfo` is a step up from that. The division-guarding sites (cosine_similarity, grad_clipping) use `clamp_min(tiny)` on norms that are sums of squares, so they'd only hit `tiny` with near-zero inputs, where a large-but-finite result beats NaN.

If we see real overflow issues in practice, we can revisit with something like `eps` or a derived value for those specific sites. For now, keeping it simple with well-known IEEE constants seems like the right tradeoff.

crates/burn-backend/src/element/base.rs
```diff
@@ -95,6 +95,44 @@ pub trait ElementLimits {
...
+/// Precision properties for floating-point element types.
+///
+/// Provides associated constants analogous to NumPy's `finfo` fields,
+/// usable in generic Rust code and CubeCL kernels.
+pub trait FloatElementPrecision: Element {
+    /// Machine epsilon: smallest value such that `1.0 + EPSILON != 1.0`.
+    const EPSILON: Self;
+    /// Smallest positive normal value (equivalent to `tiny` in NumPy/PyTorch).
+    const TINY: Self;
+}
```
laggui: I don't think we need a trait for this info. The `num_traits::Float` already exposes `epsilon()` and `min_positive_value` for all floating point types that can be used.

crates/burn-std/src/tensor/dtype.rs
```diff
@@ -162,6 +176,65 @@ pub enum FloatDType {
...
+pub struct FloatDTypeInfo {
+    /// Machine epsilon: smallest value such that `1.0 + eps != 1.0`.
+    pub eps: f64,
+    /// Largest representable finite value.
+    pub max: f64,
+    /// Most negative representable finite value.
+    pub min: f64,
+    /// Smallest positive normal value (called `tiny` in NumPy/PyTorch).
+    pub tiny: f64,
+}
```
laggui, line 194: I think we can drop the `DType` here and simply use `FloatInfo` as the name.

And we should prefer something like `min_positive` over `tiny` (even numpy only keeps `tiny` as backward compat for `smallest_normal`). And small nit: maybe `epsilon` instead to keep it aligned with the naming used in the float consts.

crates/burn-backend/src/element/base.rs
antimora: Fixed

crates/burn-std/src/tensor/dtype.rs
antimora: Fixed

Issue #4609, opened 2026-03-08
Add native det() (determinant) tensor operation


antimora:
## Summary

Add a native `det()` function to compute the determinant of a square matrix (or batch of square matrices). This would be the tensor-level equivalent of `numpy.linalg.det` / `torch.linalg.det`.

## Motivation

The ONNX importer (`burn-onnx`) now supports the `Det` operator (tracel-ai/burn-onnx#240), but the implementation has to manually compose `lu_decomposition` + `diag` + `prod` + permutation sign counting. This has two problems:

1. **Singular matrices panic.** `lu_decomposition` panics when partial pivoting encounters a pivot with magnitude <= 1e-6. A proper `det()` should return 0 (or a near-zero value) for singular matrices, not abort.

2. **No batched support.** The ONNX importer has to loop over batch dimensions one matrix at a time, which is inefficient. A native `det()` could handle batched inputs natively via the backend.

## Proposed API

```rust
// In burn::tensor::linalg, alongside lu_decomposition
pub fn det<B: Backend, const D: usize>(tensor: Tensor<B, D>) -> Tensor<B, { D - 2 }> {
    // Input shape: [*, M, M], output shape: [*]
    ...
}
```

## Notes

- PyTorch provides `torch.linalg.det` with batched support
- NumPy provides `numpy.linalg.det`
- Once available, the ONNX `Det` operator implementation can be simplified to a single `linalg::det()` call

softmaximalist:
I can work on this.

softmaximalist:
I am about to open a PR that implements det(). There is an issue with testing. I believe the tests for linalg functions implemented in the `crates/burn-backend-tests/tests/tensor/float/linalg` directory are also run in f16 mode in the CI. Hence, I am locally running my tests in normal mode (just using `cargo test`) as well in f16 mode using vulkan. While most of my tests passes in normal mode, some of them (4 test cases) fails sometimes (not always). In one of the test cases, the code behaves weirdly in f16 where prod_dim() on tensor values 1 to 8 sometimes produces 1.0 instead of the correct value which is 8! = 40320.0. I initially thought that this particular test case was failing due to a bug in prod_dim() or some related code, I am no longer sure if that is the case because when in normal (not f16), the test case always passes and never behaves like this. 

[...]

PR #4738, opened 2026-04-07
Optimize and update LU decomposition function

softmaximalist:
### Checklist

### Related Issues/PRs

Related to [Add native det() (determinant) tensor operation #4609](https://github.com/tracel-ai/burn/issues/4609) and [Linear algebra operations #1538](https://github.com/tracel-ai/burn/issues/1538)

### Changes

- Optimized LU decomposition by using the block LU algorithm (with partial pivoting) for medium and large matrices
- Tensor row swapping via the `scatter` and `gather` methods is slow so used `select` method to apply all the swaps at once
- Added batch support (most parts of the implementation strictly uses tensor operations so batch-level parallelism should be achieved for some/most parts except when applying permutations to a tensor where the code iterates over batches)
- Added support for general rectangular matrices (square, tall, and wide matrices should work)
- Updated handling of zero (or near-zero) pivots to not panic
- Added two public functions `linalg::lu` (returns full tensors) and ~~`linalg::lu_factor` (returns a single tensor for LU and permutations in compact form)~~ for API parity with PyTorch's `torch.linalg.lu` and `torch.linalg.lu_factor`

### Testing

Added 26 tests cases covering different matrix sizes, different tensor ranks, etc.

---

#### 2D Input Tensors - NdArray
| Size | Version | Fastest | Slowest | Median | Mean |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **10** | Old | **202.7 µs** | **420.8 µs** | **213.5 µs** | **216.7 µs** |
| | New | 474.4 µs | 2.298 ms | 823.5 µs | 776.4 µs |
| **100** | Old | 17.66 ms | **22.78 ms** | 18.32 ms | 18.49 ms |
| | New | **11.58 ms** | 23.67 ms | **17.9 ms** | **17.13 ms** |
| **500** | Old | 511.2 ms | 653.9 ms | 533.5 ms | 537.6 ms |
| | New | **210.9 ms** | **264.8 ms** | **236.8 ms** | **237.4 ms** |
| **1000** | Old | 2.316 s | 2.612 s | 2.475 s | 2.462 s |
| | New | **751.2 ms** | **2.135 s** | **810.4 ms** | **834.4 ms** |
| **3000** | Old | 37.68 s | 42.49 s | 39.16 s | 39.22 s |
| | New | **6.641 s** | **9.873 s** | **7.008 s** | **7.204 s** |

#### 2D Input Tensors - Wgpu Fusion*
| Size | Version | Fastest | Slowest | Median | Mean |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **10** | New | 8.246 ms | 90.36 ms | 9.272 ms | 23.87 ms |
| **100** | New | 78.39 ms | 170.5 ms | 85.59 ms | 88.06 ms |
| **500** | New | 453.6 ms | 765 ms | 482.4 ms | 487.5 ms |
| **1000** | New | 945.1 ms | 2.317 s | 991.2 ms | 1.018 s |
| **3000** | New | 3.685 s | 4.498 s | 3.949 s | 3.946 s |

#### 3D Input Tensors - NdArray
| Size | Version | Fastest | Slowest | Median | Mean |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **10** | Old | 1.475 ms | 3.918 ms | **1.706 ms** | **1.784 ms** |
| | New | **1.323 ms** | **3.747 ms** | 1.945 ms | 2.022 ms |
| **100** | Old | 123.1 ms | 137.6 ms | 128.6 ms | 129 ms |
| | New | **78.12 ms** | **104.6 ms** | **86.39 ms** | **87.38 ms** |
| **500** | Old | 3.684 s | 4.17 s | 3.962 s | 3.94 s |
| | New | **1.452 s** | **2.796 s** | **1.526 s** | **1.568 s** |
| **1000** | Old | 18.11 s | 19.84 s | 19.22 s | 19.12 s |
| | New | **6.364 s** | **9.097 s** | **6.68 s** | **6.791 s** |

#### 3D Input Tensors - Wgpu Fusion*
| Size | Version | Fastest | Slowest | Median | Mean |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **10** | New | 8.951 ms | 135.9 ms | 9.577 ms | 11.54 ms |
| **100** | New | 90.26 ms | 152.1 ms | 95.88 ms | 98.86 ms |
| **500** | New | 615.5 ms | 799.7 ms | 640.5 ms | 643.4 ms |
| **1000** | New | 1.675 s | 2.42 s | 1.722 s | 1.779 s |

**Notes:**
- Wgpu Fusion stands for: `Fusion<burn_cubecl::backend::CubeBackend<cubecl_wgpu::runtime::WgpuRuntime, f32, i32, u32>>`
- All benchmarks were run with exactly 100 samples and 100 iterations.
- A batch size of 8 was used for the 3D Input Tensors. 
- The old function was benchmarked only on NdArray backend since it is very slow on the Wgpu backend and it panics sometimes due to the scalar handling.
- The benchmarking code is available [here](https://github.com/softmaximalist/burn-benchmarks/blob/main/linalg/lu/benches/lu_bench.rs).
- While the above benchmarking was done with the `lu_factor` (since the both the first version/implementation of lu decomposition and `lu_factor` returned tensors in compact format), the `lu_factor` function have been removed from this PR following the discussion below.


softmaximalist:
It seems like the following test failed on macos:
---- autodiff::base::multithread::should_behave_the_same_with_multithread stdout ----

This is not caused by my PR right?

antimora:
> It seems like the following test failed on macos: ---- autodiff::base::multithread::should_behave_the_same_with_multithread stdout ----
> 
> This is not caused by my PR right?

I re-run the CI job; lets see if it was flaky.

crates/burn-tensor/src/tensor/linalg/lu.rs
```diff
@@ -0,0 +1,563 @@
...
+/// be as fast as highly tuned specialized libraries, especially for very large
+/// matrices or large batch sizes.
+///
+/// # Example
+/// ```rust,ignore
+/// use burn::tensor::Tensor;
+/// use burn::backend::NdArray;
+/// use burn::tensor::linalg;
+///
+/// type B = NdArray<f32>;
```
laggui: Prefer examples that are generic over the backend, e.g.

```rust
fn example<B: Backend>() {
    // ...
}
```

crates/burn-tensor/src/tensor/linalg/lu.rs
```diff
@@ -0,0 +1,563 @@
...
+    device: &B::Device,
+) -> (Tensor<B, D>, Tensor<B, D>) {
+    let dims = tensor.dims();
+    let n_rows = dims[D - 2];
+    let n_cols = dims[D - 1];
+    let piv_nums = if n_rows == n_cols {
+        n_rows
+    } else {
+        n_rows.min(n_cols)
+    };
```
laggui: The condition is redundant, you can just always use `n_rows.min(n_cols)`

crates/burn-tensor/src/tensor/linalg/lu.rs
```diff
@@ -0,0 +1,563 @@
...
+/// Scales the `k`-th column below the diagonal by the pivot element A_{kk}.
+fn update_kth_column<B: Backend, const D: usize>(tensor: Tensor<B, D>, k: usize) -> Tensor<B, D> {
+    let a_kk = tensor.clone().slice_dim(D - 2, k).slice_dim(D - 1, k);
+    let a_rho_k = tensor.clone().slice_dim(D - 2, k + 1..).slice_dim(D - 1, k);
+
+    // Replace near-zero pivots with epsilon to prevent division by zero.
+    let epsilon = 1e-4;
+    let is_zero_mask = a_kk.clone().abs().lower_elem(epsilon);
+    let safe_a_kk = a_kk.mask_fill(is_zero_mask, epsilon);
+    let updated_column = a_rho_k / safe_a_kk;
```
laggui: Is this the correct approach? I feel like this hides the singularity, and it's ok here if the operation returns NaN or Inf 🤔

softmaximalist: Good point. PyTorch provides two main function `torch.linalg.lu` and `torch.linalg.lu_factor` where `lu` completes the decomposition and returns valid P, L, U such that A = PLU even for singular matrices. On the other hand, `lu_factor` panics for singular matrices. 

My initial code (both functions) mimicked the behavior of `lu`. Since `torch.linalg.lu_factor` seems to be designed to feed its output directly into `torch.linalg.lu_solve`, an error is raised eagerly to prevent a silent bad result downstream. Thus, I have updated my code to so that `linalg::lu` does not panic and `linalg::lu_factor` panics for singular input tensors. I have added a new test case (`test_lu_factor_singular_zero_pivot`) which checks if `linalg::lu_factor` panics with the expected message.

laggui: > Since `torch.linalg.lu_factor` seems to be designed to feed its output directly into `torch.linalg.lu_solve`

I was thinking about that earlier.. we don't actually have a solver. So what's the purpose in having `lu_factor` right now?

I think we should drop it and keep `linalg::lu` only. Thoughts?

softmaximalist: I added `lu_factor` because PyTorch provides it and some users may like that it returns lu and pivot tensors in a compact format (it might be come in handy in some cases). Also, I am thinking of adding more linear algebra operations including lu_solve (at least supporting the most common operations that are also supported by PyTorch). 

If you think we don't need `lu_factor` right now, I can remove it and add it back whenever I implement `lu_solve`. Let me know what you think.

laggui: I think I'd only keep `lu` until we have the solver too.

Thanks for understanding :pray:

softmaximalist: I removed `lu_factor`.

PR #4747, opened 2026-04-07
Add linear op to ModuleOps for fused matmul+bias

### LINKED ISSUE #4737: feat: add linear to ModuleOps for fused matmul+bias
## Summary

`nn::Linear::forward` currently calls `burn_tensor::tensor::module::linear`, which decomposes into `matmul` + `add`:

```rust
let output = input.matmul(weight);
match bias {
    Some(bias) => output.add(bias),
    None => output,
}
```

This means backends cannot fuse the multiply and bias addition. Adding a `linear` method to `ModuleOps` (similar to how `conv2d` already fuses weight application + bias) would let backends optimize this common pattern.

## Motivation

- `linear` is the most common layer in transformer models. Even small gains compound across dozens of layers.
- A fused `linear` op avoids allocating the intermediate matmul result before adding bias.
- Backends can implement gemm with beta=1 to accumulate bias in-place, or use hardware-specific fused instructions.
- This follows the same pattern already established by `conv1d`/`conv2d`/`conv3d`, which all accept an optional bias and fuse it at the backend level.

## Proposed API

```rust
// In ModuleOps trait
fn linear(
    input: FloatTensor<B>,
    weight: FloatTensor<B>,
    bias: Option<FloatTensor<B>>,
) -> FloatTensor<B> {
    // Default: matmul + add
}

fn linear_x_backward(
    weight: FloatTensor<B>,
    output_grad: FloatTensor<B>,
) -> FloatTensor<B> {
    // Default: output_grad.matmul(weight)
}

fn linear_weight_backward(
    input: FloatTensor<B>,
    output_grad: FloatTensor<B>,
) -> FloatTensor<B> {
    // Default: input^T.matmul(output_grad)
}

fn linear_bias_backward(
    output_grad: FloatTensor<B>,
) -> FloatTensor<B> {
    // Default: sum over batch dims
}
```

The backward methods are needed because a fused `linear` op is opaque to autodiff (same reason conv ops need explicit backward methods). All methods would have default implementations using matmul/sum, so backends only need to override `linear` for the forward fusion benefit. Backends that can also fuse backward (e.g., cuBLAS-style fused bias gradient) can optionally override those too.

antimora:
## Summary

- Add `linear`, `linear_x_backward`, `linear_weight_backward`, and `linear_bias_backward` methods to the `ModuleOps` trait with default implementations using matmul/sum
- Update `burn_tensor::module::linear` to delegate to `B::linear` instead of inline matmul+add
- Wire through all backend layers: autodiff, fusion, router, and IR

This lets backends fuse the matmul and bias addition into a single operation (e.g., GEMM with beta=1). All methods have default implementations, so backends only need to override what they can optimize. Filed tracel-ai/cubecl#1280 for adding GEMM support to cubecl's matmul kernels.

Follows the same pattern established by `conv1d`/`conv2d`/`conv3d`.

Closes #4737

## Test plan

- [x] `cargo check --workspace` passes
- [x] `cargo test -p burn-backend --lib` passes
- [ ] Run existing linear layer tests to verify backward pass correctness
- [ ] Verify numerical equivalence between old (matmul+add) and new (B::linear) paths


crates/burn-backend/src/backend/ops/modules/linear.rs
```diff
@@ -0,0 +1,76 @@
...
+        for _ in 0..ndims - 2 {
+            grad = B::float_sum_dim(grad, 0);
+        }
+        let shape = grad.shape();
+        let d0 = shape[shape.num_dims() - 2];
+        let d1 = shape[shape.num_dims() - 1];
+        B::float_reshape(grad, Shape::new([d0, d1]))
+    } else {
+        grad
+    }
```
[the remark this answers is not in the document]
antimora: Fixed. Moved the unsqueeze from tensor/module.rs into the default linear::linear() impl. The ModuleOps::linear contract now receives weight as [d_input, d_output] and bias as [d_output] (matching how conv ops handle their weights). The default impl handles broadcasting internally via reshape. Backward ops naturally return shapes matching the original 2D/1D inputs.

crates/burn-ir/src/builder.rs
```diff
@@ -617,6 +617,69 @@ impl_ir_create!(
...
+    LinearBiasBackwardOpIr {
+        output_grad: TensorIr,
+    },
+    shape = {
+        // db: [d_output]
+        let d_output = output_grad.shape[output_grad.shape.num_dims() - 1];
+        Shape::from(alloc::vec![d_output])
+    },
+    dtype = output_grad.dtype
+);
```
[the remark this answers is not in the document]
antimora: Fixed in the same change. Weight is now [d_input, d_output] and bias is [d_output] at the ModuleOps level, so the backward IR shapes ([d_input, d_output] and [d_output]) are correct. The default forward impl handles the unsqueeze internally.

crates/burn-tensor/src/tensor/module.rs
```diff
@@ -481,17 +481,15 @@ pub fn linear<B: Backend, const D: usize>(
...
-    let output = input.matmul(weight);
-    match bias {
-        Some(bias) => output.add(bias),
-        None => output,
-    }
+    Tensor::new(TensorPrimitive::Float(B::linear(
+        input.primitive.tensor(),
+        weight.primitive.tensor(),
+        bias.map(|b| b.primitive.tensor()),
+    )))
```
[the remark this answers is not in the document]
antimora: Fixed. Removed the unsqueeze from this function. B::linear now receives the original 2D weight and 1D bias, and the default impl handles broadcasting internally. This makes the contract consistent end-to-end: backward ops return gradients with the same shapes as their forward inputs.

crates/burn-autodiff/src/ops/module.rs
```diff
@@ -56,6 +56,116 @@ impl<B: Backend, C: CheckpointStrategy> ModuleOps<Autodiff<B, C>> for Autodiff<B
         panic!("Can't differentiate embedding backward.");
     }
```
laggui, line 57: Might want to mark the linear backward ops similarly

crates/burn-autodiff/src/ops/module.rs
```diff
@@ -56,6 +56,116 @@ impl<B: Backend, C: CheckpointStrategy> ModuleOps<Autodiff<B, C>> for Autodiff<B
...
+
+        match bias {
+            Some(bias) => match LinearWithBias
+                .prepare::<C>([x.node.clone(), weight.node.clone(), bias.node.clone()])
+                .compute_bound()
+                .stateful()
+            {
+                OpsKind::Tracked(mut prep) => {
+                    let x_state = prep.checkpoint(&x);
+                    let weight_state = prep.checkpoint(&weight);
```
laggui: We might have to conditionally checkpoint the tensors, similar to what matmul is doing. There could be cases where gradients are not tracked for one or the other (e.g. frozen layers).

crates/burn-backend/src/backend/ops/modules/linear.rs
```diff
@@ -0,0 +1,106 @@
...
+    // Sum over all dims except the last (the output feature dim).
+    // float_sum_dim preserves rank (keepdim), so sum each dim at its index.
+    for dim in 0..ndims - 1 {
+        grad = B::float_sum_dim(grad, dim);
+    }
+
+    let shape = grad.shape();
+    let d_output = shape[ndims - 1];
+    B::float_reshape(grad, Shape::new([d_output]))
+}
```
laggui, line 106: It is a bit annoying that this change forces a fixed backward decomposition that is now manually implemented instead of reusing the autodiff system 😅

crates/burn-fusion/src/ops/module.rs
```diff
@@ -32,6 +32,137 @@ macro_rules! make_ops {
 }
 
 impl<B: FusionBackend> ModuleOps<Fusion<B>> for Fusion<B> {
+    fn linear(
+        x: FloatTensor<Self>,
+        weight: FloatTensor<Self>,
+        bias: Option<FloatTensor<Self>>,
+    ) -> FloatTensor<Self> {
```
laggui: This will break fusion in its current state, so it's a regression.

The current implementation captures a matmul and (optionally) elemwise add, which are handled in burn-cubecl-fusion.

The linear op IR is not.

laggui (changes requested):

crates/burn-autodiff/src/ops/module.rs
antimora: Good catch, done. Added panic impls for linear_x_backward, linear_weight_backward, and linear_bias_backward in the Autodiff<B, C> impl, matching the embedding_backward pattern. If someone calls a backward op on an autodiff tensor directly (outside the normal linear backward path), it now panics loudly instead of silently computing wrong gradients.

crates/burn-autodiff/src/ops/module.rs
antimora: Done. Switched State to `(Option<NodeId>, Option<NodeId>)` and gate the checkpoints on the other input's tracked flag, mirroring float_matmul:

```rust
let x_tracked = x.is_tracked();
let weight_tracked = weight.is_tracked();
// x is only needed to compute the weight gradient, and vice versa.
let x_state = weight_tracked.then(|| prep.checkpoint(&x));
let weight_state = x_tracked.then(|| prep.checkpoint(&weight));
```

So a frozen weight no longer forces x to be checkpointed, and a frozen x no longer forces weight to be. Bias doesn't need any state since its gradient is just a reduction of output_grad.

crates/burn-backend/src/backend/ops/modules/linear.rs
antimora: Agreed, it's not elegant. The manual decomposition is the same trade-off the conv ops already make: a fused forward op is opaque to autodiff, so we have to spell out the backward by hand. The alternative would be something like a retro_forward that re-runs the decomposed matmul+add and lets autodiff trace through it, but that gives up the memory/fusion benefits of having a single op, which is the whole motivation for the issue.

Happy to explore a cleaner pattern if you have one in mind, otherwise I'd suggest landing this shape (matching conv) and revisiting the abstraction if/when enough of these fused ops accumulate to justify it.

crates/burn-fusion/src/ops/module.rs
antimora: Good point, dropped the override. Removed the `linear`, `linear_x_backward`, `linear_weight_backward`, and `linear_bias_backward` impls from `burn-fusion/src/ops/module.rs` so they fall back to the default ModuleOps decomposition (matmul + add for forward, matmul / swap_dims / sum_dim / reshape for the backwards). That keeps the primitive IR nodes visible to burn-cubecl-fusion, so fusion still works.

Left the `LinearOpIr` / `LinearXBackwardOpIr` / etc. definitions alone since burn-router still uses them (router forwards IR opaquely to a remote backend and doesn't decompose). If/when burn-cubecl-fusion gains a fused linear pattern matcher, we can add the fusion override back at that point.

laggui (approved):
LGTM! Thanks for addressing my comments.

Sorry for the delayed follow-up 😅 it's been a busy week.

Issue #4683, opened 2026-03-27
Burn-RL examples not compiled because of SDL dependency


johnny-smitherson:
### Feature description

Bring back the RL examples from the burn-rl crate, disabled by default from builds here https://github.com/tracel-ai/burn/blob/ed72d2b125a364aff18aed2a53396c128e01cb42/Cargo.toml#L19

This is disabled because of the outdated dependency https://github.com/MathisWellmann/gym-rs/pull/14

This does not build on modern linux e.g. arch linux, where CMAKE and SDL included by sdl-sys are very out of date and do not compile.

### Feature motivation

Build on RL example with more implementations e.g. https://github.com/yunjhongwu/burn-rl-examples

### (Optional) Suggest a Solution

Remove rendering by removing sdl feature, e.g. https://github.com/yunjhongwu/burn-rl-examples/issues/54

or change dependency to modern fork e.g. https://github.com/MathisWellmann/gym-rs/pull/14

I have done this in a fork https://github.com/tracel-ai/burn/compare/main...johnny-smitherson:burn:rl-examples-fix

NathanTheProgrammer:
I'll work on getting a fix on this one :)

PR #4756, opened 2026-04-08
Fix/dqn agent no sdl

NathanTheProgrammer:
### Related Issues/PRs

[Burn-RL examples not compiled because of SDL dependency #4683](https://github.com/tracel-ai/burn/issues/4683)

### Changes

Removed the use of gym-rs, which fixes the build issues with the RL example.

### Testing

Tested that the example complies (AKA using "cargo run -p dqn-agent")


examples/dqn-agent/Cargo.toml
```diff
@@ -31,8 +31,8 @@ burn = { path = "../../crates/burn", features = [
     "ndarray",
     # "autotune",
 ], default-features = false }
-# Just for this example.
-gym-rs = { version = "0.3.1", branch = "main", git = "https://github.com/MathisWellmann/gym-rs" }
+# Removing gym-rs because it pulls in SDL which causes big no nos to happen in linux
+# gym-rs = { version = "0.3.1", branch = "main", git = "https://github.com/MathisWellmann/gym-rs" }
```
BjornTheProgrammer: I would just remove it without leaving a comment as to why. The PR and issue already explain why it is needed.

Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
     "crates/burn-store/pytorch-tests",
     "crates/burn-store/safetensors-tests",
     "crates/burn-collective/multinode-tests",
+    "examples/dqn-agent",
```
BjornTheProgrammer: This should already be fine with `examples/*` catching the `example/dqn-agent`. Instead remove `examples/dqn-agent` from the `exclude = [...]`

BjornTheProgrammer (changes requested):
I don't have review authority, but these are just somethings that I notice right away. Great work!

Cargo.toml
NathanTheProgrammer: lol didn't see that

examples/dqn-agent/Cargo.toml
NathanTheProgrammer: done

examples/dqn-agent/src/env.rs
```diff
@@ -80,22 +74,27 @@ impl Environment for CartPoleWrapper {
...
+        self.state[3] += 0.01;
+
         StepResult {
-            next_state: CartPoleState::from(action_reward.observation),
-            reward: action_reward.reward.into_inner(),
-            done: action_reward.done,
+            next_state: CartPoleState::new(self.state),
+            reward: 1.0,
+            done: false,
             truncated: self.step_index >= Self::MAX_STEPS,
```
nathanielsimard: I don't think that's right, the action has no effect on the cart pole.

Charles23R: I would add that `done` is always false

urmzd:
Hey folks - there's also the option to use this fork: https://github.com/urmzd/gymnasia which leverages macroquad instead - which cross compiles to almost all architectures when used; and by default is feature gated.  Additionally, I've added a headless option - which might fit the use case here a bit better

examples/dqn-agent/src/env.rs
NathanTheProgrammer: Small update to this, my initial thought process was to just remove gym rs, so I updated it so now it actually does a simple direction update, and it terminates based on the step count

NathanTheProgrammer: I also noticed the comment about the fork I'll check it out lol

However I personally prefer removing deps since in the case gym-rs is only used in this one example

laggui: @NathanTheProgrammer removing the dep is a good approach, but we should preserve the functionality of the example. In this case, the task is to balance a pole on top of a moving cart.

That means the physics of the cartpole should remain :sweat_smile:

We can either port the environment implementation for the cartpole, or use another dependency instead

examples/dqn-agent/src/main.rs
laggui, line 1: The entrypoint for the example is already present in `examples/dqn-agent.rs` btw!

laggui:
Closed due to inactivity; but feel free to re-open if you want to keep the work going!

PR #4766, opened 2026-04-10
Fix flaky initializer_normal_init test

### LINKED ISSUE #4705: Flaky test: initializer_normal_init fails intermittently due to tight variance tolerance
## Description

The test `module::initializer::tests::initializer_normal_init` in `burn-core` fails intermittently due to a tight statistical tolerance on the variance of randomly-initialized tensors.

## Error

```
Expected variance to be between 1.0 += 0.1, but got 0.89897823
```

## Reproduction

Run the test repeatedly:
```bash
for i in $(seq 1 50); do cargo test -p burn-core --lib -- module::initializer::tests::initializer_normal_init 2>&1 | grep -E "ok|FAILED"; done
```

It fails roughly 1 in 10-20 runs.

## Root Cause

The test initializes a tensor with `Initializer::Normal { mean: 0.0, std: 1.0 }` and checks that the sample variance falls within `1.0 +/- 0.1`. With a finite sample size, the sample variance naturally fluctuates and occasionally lands outside this window.

## Suggested Fix

Widen the tolerance (e.g., `+/- 0.2`) or increase the tensor size to reduce variance of the sample variance.

## Context

Observed while working on #3754. This test is unrelated to that PR.

leohenon:
### Checklist

   ### Related Issues/PRs

   Closes #4705

   ### Changes

   Bump the sample size in `initializer_normal_init` from 1,000 to 10,000.

   With only 1,000 samples the ±0.1 tolerance is about 2.2σ, so it flakes roughly 1 in 37 runs. Bumping to 10,000 pushes it to ~7σ, which kills the flakiness without weakening the test.

   ### Testing

   Ran the test 30 times.

   ```bash
   for i in $(seq 1 30); do cargo test -p burn-core --lib -- module::initializer::tests::initializer_normal_init; done

### DIFF
--- a/crates/burn-core/src/module/initializer.rs
+++ b/crates/burn-core/src/module/initializer.rs
@@ -342,7 +342,7 @@ mod tests {
 
         let (mean, std) = (0.0, 1.0);
         let normal: Tensor<TB, 1> = Initializer::Normal { mean, std }
-            .init([1000], &Default::default())
+            .init([10000], &Default::default())
             .into_value();
         let (var_act, mean_act) = normal.var_mean(0);
 


laggui (approved):
Thanks 😄

PR #4781, opened 2026-04-12
Fix autotune context, remove unsafe code

ArthurBrussee:
sum_dim (and other autotuned reductions) crashed on wasm because forked benchmark outputs weren't being persisted to the real context. It paniced with something about a missing TensorId. The root cause is that create_key called input.context() which set executed = true on the original and suppressed the drop-time handle persistence path. On native this was fine because of the blocking autotune flow, but on wasm which uses async benchmarks it crashed with these missing handles.

- Forked contexts now don't try to hold a raw ptr anymore the original context. Instead there is one shared handle collection to push to, and the final context drop checks which handles are fresh and registers them. `SharedNewHandles` used unsafe to remove a lock, I've just used a spin::Mutex, on a single thread a _single_ atomic operation doesn't seem worth optimizing. 
- The original context now just holds an &mut so that rust can reason about mutability & aliasing etc. The only `unsafe` remaining is that we erase the lifetime of the reference.
- Merge Context and ContextOwned, as `Context` is essentially just a &ContextOwned. Functions passed around &mut Context which really was a reference to a reference. This means the bit of unsafe `transmute` casts away one less lifetime, easier to reason about.
- Fix some unsoundess: .context() could be called twice, producing two aliased &mut Contexts


ArthurBrussee:
I think the macOS test is flaky, not entirely sure, but the same one fails on main with enough runs.

crates/burn-cubecl-fusion/src/tune.rs
```diff
@@ -1,290 +1,245 @@
...
+                context,
+                new_handles,
+                optimization,
+                ..
+            } => (context.fork(), new_handles.clone(), optimization.clone()),
+        };
+        let original_ids = forked.handles().handle_ids().copied().collect();
+        // Each new fork resets the handles saved by the previous execution,
+        // so no memory leak is created by keeping discarded handles.
+        // SAFETY: sequential execution, no concurrent access.
```
nathanielsimard: I don't think we need to clone all the `handle ids`. During the drop, we can use the handle container of the original context to do the checks.

ArthurBrussee: Few refactors later....

we're now collecting everything to the original context instead and then on the final drop figuring out what we need to register. The last unsafe transmute is gone after https://github.com/tracel-ai/cubecl/pull/1293.

Like you mentioned we can maybe get rid of this registration dance entirely in the future.

crates/burn-fusion/src/stream/context.rs
```diff
@@ -54,49 +67,59 @@ impl Default for OperationConverter {
...
+    /// will swap them back on drop.
+    pub(crate) fn new(
+        converter: &'a mut OperationConverter,
+        handles: &'a mut HandleContainer<H>,
+    ) -> Self {
+        let context = Context {
+            tensors: core::mem::take(&mut converter.tensors_relative2global),
+            scalars: core::mem::take(&mut converter.scalars),
+            shapes_relative2global: core::mem::take(&mut converter.shapes_relative2global),
+            handles: core::mem::take(handles),
```
nathanielsimard, line 91: So there is no allocation, you simply move the data from the converter into the context, and when the context is dropped you move back the data to the converter?

That's actually a good idea, great way to remove the lifetime and a bunch of unsafe code.

crates/burn-cubecl/src/kernel/matmul/tune/base.rs
```diff
@@ -37,29 +38,37 @@ pub fn matmul_autotune<R: CubeRuntime>(
...
+                PRIORITY_MEDIUM
+            }
+        });
 
-            // In some case when a relayout can be fused (no call to into_contiguous) it's better
-            // to use accelerated matmul.
-            //
-            // TODO: Actually implement good gemv with fused relayout.
-            } else if matches!(key.analysis.kind, MatmulKind::MatVec | MatmulKind::VecMat) {
+        let mma = TuneGroup::<MatmulAutotuneKey>::new("mma", |key| {
```
nathanielsimard: There are changes from main that were not merged correctly here. The is no `mma` and `cmma` now, they all fall within the same `accelerated` group.

crates/burn-cubecl/src/kernel/matmul/tune/base.rs
```diff
@@ -105,46 +114,10 @@ pub fn matmul_autotune<R: CubeRuntime>(
...
-                            PRIORITY_HIGH
-                        }
-                    }
-                    // TODO: Actually do the correct relayout here.
-                    //
-                    // Every algo will need to relayout, in this case, we should take the optimal
-                    // kernel with a gemv.
-                    MatrixBatchLayout::HighlyPermuted => PRIORITY_HIGH,
-                }
+        let gemv = TuneGroup::<MatmulAutotuneKey>::new("gemv", |key| {
```
nathanielsimard: Same thing here the `gemv` is quite different now.

crates/burn-cubecl-fusion/src/tune.rs
```diff
@@ -1,292 +1,188 @@
...
+        context: &'a mut Context<CubeFusionHandle<R>>,
+        /// Set by [`TuneInput::execute`] when the closure has run on the real context. If
+        /// true at drop time, the staging drain is skipped — the real context has already
+        /// been written to directly, so there's nothing to merge.
+        executed: bool,
     },
+    /// Owned fork used by benchmark runs and by the wasm try-all fallback. Mutations stay
+    /// local; [`Drop`] captures the fork's full handle set into the shared staging area so
+    /// the original can filter-and-register at drain time.
+    Fork(Box<Context<CubeFusionHandle<R>>>),
```
nathanielsimard: I would not use the clone input generator for autotuning. I would create a new generator for `TuneInput` that creates a context that doesn't track new handles. I would keep the current `Fork` which only runs when the real function is called on real cloned inputs.

ArthurBrussee: Handle collection is now an Option<> and only used when it's actually needed, lmk!

nathanielsimard (approved):

PR #4791, opened 2026-04-13
Make cubecl normal distribution test reproducible

leohenon:
Seed `normal_respects_68_95_99_rule` in `crates/burn-backend-tests/tests/cubecl/normal.rs` to make it reproducible.

Tested with:
```bash
   cargo test -p burn-backend-tests --test cubecl --features cube cube::kernel::normal -- --nocapture
 ```


laggui (approved):
Didn't notice it wasn't seeded 😅 

Not strictly required but it does help reproducibility. So failures are not necessarily tied to a statistical anomaly but a regression.

Issue #4774, opened 2026-04-10
Move generic tests from burn-flex into burn-backend-tests


antimora:
## Proposal

Move generic (not flex-specific) tests out of `crates/burn-flex/src/ops/*.rs` and into `crates/burn-backend-tests/tests/tensor/...`, so the shared suite runs them against every backend instead of only burn-flex.

## Motivation

`crates/burn-flex/src/ops/` currently has ~500 `#[test]` functions spread across the ops modules. A sizable chunk are plain "does `relu` work on a 1d f32 tensor" style smoke tests that test backend-agnostic behavior. They live next to the implementation for historical reasons (flex is new and it was convenient during development), but they do not exercise anything specific to FlexTensor's layout system, and the same tests in `burn-backend-tests` would catch the same regressions while also covering ndarray, wgpu, cuda, etc.

Three costs of the current state:

1. Coverage gap. A bug in the shared dispatch logic (e.g. tracel-ai/burn#4771, tracel-ai/burn#4772) can sail past backend-tests because it happens to only land in burn-flex unit tests, and backend-tests has no equivalent coverage.
2. Duplication risk. When a new bug is filed, it's natural to add the regression test at the fix site (in burn-flex). That leaves the other backends uncovered. I hit exactly this pattern when fixing #4771 and ended up writing the tests twice before dedup.
3. Test-run velocity. `cargo test -p burn-flex` is bloated with tests that don't need a flex-specific build to run.

## What counts as "flex-specific"

Should stay in `burn-flex`:

- Tests that exercise FlexTensor's stride system directly: `test_binary_mul_non_contiguous_transposed`, `test_binary_add_non_contiguous_narrowed`, `test_bool_and_flipped`, `test_bool_and_both_flipped`, `test_bool_into_int_flipped_2d`, etc. These probe negative/permuted/narrowed strides and how the op handles them - something the backend-tests `TestTensor` API cannot express directly.
- Tests that pin SIMD/Rayon chunk boundaries: `test_softmax_simd_body_row`, `test_softmax_multi_chunk_rayon`. These guard against off-by-one bugs at the boundaries of SIMD lanes or worker chunks, which only exist in flex.
- Tests that pin flex-specific dtype storage: `test_bool_into_int_u8` (checks `BoolStore::U8` backing).
- Tests for flex-internal helpers (layout, strided iterator, SIMD intrinsics).

Should move to `burn-backend-tests`:

[...]

PR #4805, opened 2026-04-15
burn-flex: wire fused softmax and layer_norm into backend trait hooks

antimora:
## Summary
- Override `ActivationOps::softmax` and `ModuleOps::layer_norm` on Flex so `burn_tensor::activation::softmax` and `burn::nn::LayerNorm::forward` (wired in #4797) hit the fused kernels instead of the primitive-op decomposition.
- Extend the fused `softmax` to handle non-last-axis via `swap_dims` + fuse + swap back.
- Extend the fused `layer_norm` to cover f64 (new monomorphized `with_beta`/`no_beta` row kernels, `ROWS_PER_TASK` rayon chunking, shared `welford_f64`) and f16/bf16 (cast-fuse-cast through the f32 kernel).

## Benchmark results (Flex, median)

Softmax via `B::softmax` vs decomposition:
| Case | Decomposed | Fused | Speedup |
|---|---|---|---|
| last-axis 12x512x512 | 5.14 ms | 752 us | 6.8x |
| last-axis 2x256x2048 | 1.86 ms | 410 us | 4.5x |
| non-last-axis 4x1024x512 | 23.8 ms | 6.94 ms | 3.4x |

LayerNorm via `B::layer_norm` vs decomposition (ConvNeXt + transformer, f32):
| Case | Decomposed | Fused | Speedup |
|---|---|---|---|
| c48 HW=244x224 | 1.60 ms | 243 us | 6.6x |
| c96 HW=122x112 | 698 us | 169 us | 4.1x |
| c192 HW=61x56 | 344 us | 79.7 us | 4.3x |
| c384 HW=30x28 | 169 us | 61.2 us | 2.8x |
| d768 seq=512 b=2 | 418 us | 96.2 us | 4.3x |

NdArray numbers are effectively unchanged between the decomposed and trait-routed bench groups (it doesn't override the hook), confirming the wins come from the Flex fused kernels plugging into the new hooks.

## Safety
- Added dtype assertions on gamma/beta vs input to prevent the f16/bf16 cast path from silently reinterpreting mismatched bytes.

## Test plan
- [x] `cargo test -p burn-flex --features simd,rayon` - 605 passing (7 new)
- [x] New tests cover: f64 `with_beta` multi-chunk, f64 `no_beta`, f16/bf16 via f32 cast, rank-4 non-last-axis softmax (dim=0), softmax dim-out-of-range panic, gamma dtype mismatch panic
- [x] `cargo bench -p burn-flex --bench softmax_ops --features simd,rayon`
- [x] `cargo bench -p burn-flex --bench norm_ops --features simd,rayon`

### DIFF
--- a/crates/burn-flex/benches/softmax_ops.rs
+++ b/crates/burn-flex/benches/softmax_ops.rs
@@ -1,0 +1,102 @@
+//! Benchmarks comparing `B::softmax` against a manual decomposition baseline.
+//! Covers last-axis and non-last-axis cases.
+//!
+//! Run with:
+//! ```bash
+//! cargo bench --bench softmax_ops --features simd,rayon
+//! ```
+
+use burn_flex::Flex;
+use burn_ndarray::NdArray;
+use burn_tensor::{Element, Tensor, TensorData, backend::Backend};
+use divan::{AllocProfiler, Bencher};
+
+#[global_allocator]
+static ALLOC: AllocProfiler = AllocProfiler::system();
+
+fn main() {
+    println!("Softmax Benchmarks: fused (via B::softmax) vs decomposed");
+    println!();
+    divan::main();
+}
+
+fn make_tensor_3d<B: Backend, E: Element + From<f32>>(
+    d0: usize,
+    d1: usize,
+    d2: usize,
+) -> Tensor<B, 3> {
+    let data: Vec<E> = (0..d0 * d1 * d2)
+        .map(|i| E::from((((i % 997) as f32) / 997.0) - 0.5))
+        .collect();
+    Tensor::from_data(TensorData::new(data, [d0, d1, d2]), &Default::default())
+}
+
+/// `softmax(x, dim) = exp(x - max) / sum(exp(x - max))`, matching the
+/// `ActivationOps::softmax` default in burn-backend.
+fn decomposed_softmax<B: Backend, const D: usize>(x: Tensor<B, D>, dim: usize) -> Tensor<B, D> {
+    let max = x.clone().detach().max_dim(dim);
+    let shifted = x - max;
+    let exp = shifted.exp();
+    let sum = exp.clone().sum_dim(dim);
+    exp / sum
+}
+
+macro_rules! bench_backend {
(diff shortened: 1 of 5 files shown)


crates/burn-flex/src/ops/activation.rs
```diff
@@ -530,6 +539,26 @@ pub fn layer_norm(
 ) -> FloatTensor<Flex> {
     let rank = input.shape().num_dims();
     assert!(rank >= 1, "layer_norm: input must have at least one dim");
+    // gamma/beta storage is reinterpreted as the input's element type in the
+    // half-precision cast path (see `layer_norm_via_f32`). A dtype mismatch
+    // there would silently reinterpret bytes rather than panic, so reject
+    // it up front.
```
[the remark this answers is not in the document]
antimora: Good catch, thanks. Verified `FlexTensor::storage::<E>()` at crates/burn-flex/src/tensor.rs:172-182 does assert `E::dtype() == self.dtype` and would panic on mismatch, so "silently reinterpret bytes" was wrong. The early `assert_eq!` still earns its place (clearer `layer_norm`-specific message, fails before `to_contiguous()`), but the justification comment needed fixing. Applied your suggestion verbatim in ad461eb98 and also corrected the parallel comment on `test_layer_norm_gamma_dtype_mismatch_panics` to reflect the real behavior.

crates/burn-flex/benches/softmax_ops.rs
laggui, line 1: Just a heads up: with #4717 we'll need to remove the burn-tensor dependency from backends.

These benchmarks will have to be moved in time. Along with tests as you captured in #4774.

laggui (approved):
The failing CI test on macos is unrelated to the changes.

I've noticed it has been failing intermittently since #4760. Will open an issue to track it.

crates/burn-flex/benches/softmax_ops.rs
antimora: Filed https://github.com/tracel-ai/burn/issues/4810

PR #4813, opened 2026-04-15
Add native det (determinant) tensor operation

### LINKED ISSUE #4609: Add native det() (determinant) tensor operation

softmaximalist:
### Checklist

### Related Issues/PRs

Closes [Add native det() (determinant) tensor operation #4609](https://github.com/tracel-ai/burn/issues/4609)

### Changes

- Implemented `linalg::det` tensor operation
- Includes batch support and correct handling of singular matrices (returns det of 0)

### Testing

Added 31 tests cases that covers the base cases (1x1, 2x2, 3x3) as well larger tensors. The test cases also covers determinant properties, singular matrices, tensors with 1 or more leading batch dimensions, panic tests, etc.


softmaximalist:
The following test case (which is not related to my PR) seems to have failed:
`cube::kernel::select_assign::select_add_should_work_with_multiple_workgroups_3d_dim2`

Leaving a comment so a maintainer can rerun the CI or take appropriate action.

crates/burn-tensor/src/tensor/linalg/det.rs
```diff
@@ -0,0 +1,157 @@
...
+    // Update f16 to f32
+    let original_dtype = tensor.dtype();
+    let needs_upcast = original_dtype == DType::F16 || original_dtype == DType::BF16;
+    let working_float_dtype: FloatDType;
+    if needs_upcast {
+        working_float_dtype = FloatDType::F32;
+        tensor = tensor.cast(working_float_dtype);
+    } else {
+        working_float_dtype = original_dtype.into()
+    };
```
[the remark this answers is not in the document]
softmaximalist: Good point. While we don't have to worry about non-float dtypes since the input tensor has the type `Tensor<B, D>`, a quantized float tensor is still valid. Hence, I have added a tensor check to reject a quantized input tensor.

crates/burn-tensor/src/tensor/linalg/lu.rs
```diff
@@ -119,7 +120,8 @@ fn block_lu_with_partial_piv<B: Backend, const D: usize, const D1: usize>(
     let n_rows = dims[D - 2];
     let n_cols = dims[D - 1];
     let piv_nums = n_rows.min(n_cols);
-    let mut global_piv = create_permutation_tensor::<B, D>(piv_nums, dims, &device);
+    let dtype = tensor.dtype().into();
+    let mut global_piv = create_permutation_tensor::<B, D>(piv_nums, dims, dtype, &device);
```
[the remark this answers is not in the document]
softmaximalist: The input tensor passed to `compute_lu_decomposition` will never be f16 or bf16 (due to the upcasting) or a quantized tensor (due to TensorCheck rejecting it). Hence, the det function should work fine. 

The issues mentioned in the comment are relevant when the `linalg::lu` function is called directly on a quantized float tensor or on a tensor with f16 or bf16 dtype. This should be updated in a separate PR as it is not directly relevant to the current PR.

crates/burn-tensor/src/tensor/linalg/lu.rs
```diff
@@ -228,7 +230,8 @@ fn standard_lu_with_partial_piv<B: Backend, const D: usize, const D1: usize>(
     let n_rows = dims[D - 2];
     let n_cols = dims[D - 1];
     let piv_nums = n_rows.min(n_cols);
-    let mut permutations = create_permutation_tensor::<B, D>(piv_nums, dims, device);
+    let dtype = tensor.dtype().into();
```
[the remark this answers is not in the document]
softmaximalist: The input tensor passed to `compute_lu_decomposition` will never be f16 or bf16 (due to the upcasting) or a quantized tensor (due to TensorCheck rejecting it). Hence, the det function should work fine. 

The issues mentioned in the comment are relevant when the `linalg::lu` function is called directly on a quantized float tensor or on a tensor with f16 or bf16 dtype. This should be updated in a separate PR as it is not directly relevant to the current PR.

crates/burn-tensor/src/tensor/linalg/det.rs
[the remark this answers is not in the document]
softmaximalist: Note that I added the following test case to test the new check in the det TensorCheck:
```rust
#[test]
#[should_panic(expected = "The input tensor must have a real float dtype.")]
fn test_det_quantized_tensor() {
        .with_value(QuantValue::Q8S)
        .with_store(QuantStore::Native);

    let tensor = TestTensor::<3>::from_data([[[5.0]]], &device);
    let qtensor = tensor.quantize_dynamic(&scheme);
    let det_tensor = det::<TestBackend, 3, 2, 1>(qtensor);
    let expected = TestTensor::<1>::from_data([5.0], &device);
    let tolerance = Tolerance::default().set_half_precision_absolute(5e-3);
    det_tensor
        .into_data()
        .assert_approx_eq::<FloatElem>(&expected.into_data(), tolerance);
}
```

While this works fine when run by itself, it causes errors in many (previously passing) test cases when all the tests cases (including the new one above) are run in f16 using vulkan. Enabling backtrace gives the following output:

```
thread 'Device-4-0' (84876) panicked at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:1952:22:
wgpu error: Validation Error

Caused by:
  In Queue::write_buffer
    Buffer with '' label is invalid


stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/e408947bfd200af42db322daf0fadfe7e26d3bd1/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/e408947bfd200af42db322daf0fadfe7e26d3bd1/library/core/src/panicking.rs:80:14
   2: wgpu::backend::wgpu_core::default_error_handler
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:687:5
   3: wgpu::backend::wgpu_core::ErrorSinkRaw::handle_error_or_return_handler
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:671:21
   4: wgpu::backend::wgpu_core::ContextWgpuCore::handle_error_inner
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:299:18
   5: wgpu::backend::wgpu_core::ContextWgpuCore::handle_error_nolabel
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:332:14
   6: <wgpu::backend::wgpu_core::CoreQueue as wgpu::dispatch::QueueInterface>::write_buffer
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/backend/wgpu_core.rs:1952:22
   7: wgpu::api::queue::Queue::write_buffer
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wgpu-29.0.1/src/api/queue.rs:158:20
   8: cubecl_wgpu::compute::stream::WgpuStream::write_to_buffer
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-wgpu-0.10.0-pre.3/src/compute/stream.rs:403:18
   9: cubecl_wgpu::compute::stream::WgpuStream::create_uniform
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-wgpu-0.10.0-pre.3/src/compute/stream.rs:378:14
  10: cubecl_wgpu::compute::schedule::BindingsResource::into_resources
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-wgpu-0.10.0-pre.3/src/compute/schedule.rs:126:31
  11: cubecl_wgpu::compute::stream::WgpuStream::enqueue_task
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-wgpu-0.10.0-pre.3/src/compute/stream.rs:121:43
  12: <cubecl_wgpu::compute::schedule::ScheduledWgpuBackend as cubecl_runtime::stream::scheduler::SchedulerStreamBackend>::enqueue
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-wgpu-0.10.0-pre.3/src/compute/schedule.rs:141:16
  13: cubecl_runtime::stream::scheduler::SchedulerMultiStream<B>::execute_schedules_interleave
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-runtime-0.10.0-pre.3/src/stream/scheduler.rs:248:21
  14: cubecl_runtime::stream::scheduler::SchedulerMultiStream<B>::execute_streams
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-runtime-0.10.0-pre.3/src/stream/scheduler.rs:201:51
  15: <cubecl_wgpu::compute::server::WgpuServer as cubecl_runtime::server::base::ComputeServer>::read
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-wgpu-0.10.0-pre.3/src/compute/server.rs:301:24
  16: cubecl_runtime::client::ComputeClient<R>::do_read::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-runtime-0.10.0-pre.3/src/client.rs:103:51
  17: <cubecl_common::device::handle::channel::ChannelDeviceHandle<S> as cubecl_common::device::handle::base::DeviceHandleSpec<S>>::submit_blocking::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:78:28
  18: cubecl_common::device::handle::channel::ChannelDeviceHandle<S>::submit_inner::{{closure}}::{{closure}}::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:190:37
  19: cubecl_common::stream_id::StreamId::executes
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/stream_id.rs:43:24
  20: cubecl_common::device::handle::channel::ChannelDeviceHandle<S>::submit_inner::{{closure}}::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:190:25
  21: cubecl_common::device::handle::channel::ChannelService::act_on::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:366:13
  22: std::thread::local::LocalKey<core::cell::RefCell<T>>::with_borrow::{{closure}}
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:724:26
  23: std::thread::local::LocalKey<T>::try_with
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:513:12
  24: std::thread::local::LocalKey<T>::with
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:477:20
  25: std::thread::local::LocalKey<core::cell::RefCell<T>>::with_borrow
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:724:14
  26: cubecl_common::device::handle::channel::ChannelService::act_on
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:361:16
  27: cubecl_common::device::handle::channel::ChannelDeviceHandle<S>::submit_inner::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:185:19
  28: <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panic/unwind_safe.rs:274:9
  29: std::panicking::catch_unwind::do_call
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:581:40
  30: __rust_try
  31: std::panicking::catch_unwind
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:544:19
  32: std::panic::catch_unwind
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panic.rs:359:14
  33: cubecl_common::device::handle::channel::task::Task::init::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:463:39
  34: core::ops::function::FnOnce::call_once
             at /home/myusername/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
  35: cubecl_common::device::handle::channel::task::Task::run
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:481:13
  36: cubecl_common::device::handle::channel::custom_channel::Server::execute_tasks
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:763:22
  37: cubecl_common::device::handle::channel::custom_channel::Server::start
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:747:26
  38: cubecl_common::device::handle::channel::custom_channel::DeviceClient::new::{{closure}}
             at /home/myusername/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cubecl-common-0.10.0-pre.3/src/device/handle/channel.rs:611:28
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace
```

Hence, I have removed this test case for now.

antimora (commented):
Nice addition. Core `det` implementation is correct: batched singular-mask logic, permutation-sign parity, and the closed-form paths for N=1/2/3 all check out. Main concerns below cluster around scope creep into `lu.rs` and documenting the numerical behavior the function silently applies.

On prior Copilot feedback: agreed that the `tensor.dtype().into()` issue at lu.rs:124 / lu.rs:233 can stay deferred, since `det` upcasts f16/bf16 and now rejects QFloat before calling `compute_lu_decomposition`. Your reasoning holds for this PR.

Two tests worth adding if easy (happy to defer either):

1. Mixed singular/non-singular batch through the LU branch. Shape `[B>=2, N>=4, N]` with one singular and one well-conditioned matrix, asserting only the singular entry returns 0. The current batched singular tests all have batch size 1 or hit the 3x3 fast path, so the `any_dim` + `mask_fill` reduction at det.rs:156-158 is never exercised with partial singularity.

2. Explicit dtype round-trip for f16/bf16, asserting `det_tensor.dtype() == original_dtype`. The cast-back logic is duplicated across four return paths (lines 89, 99, 117, 164) and `into_data()` comparisons drop dtype, so CI would not catch one of those branches being removed.

crates/burn-tensor/src/tensor/linalg/lu.rs
```diff
@@ -360,7 +365,8 @@ fn update_kth_column<B: Backend, const D: usize>(tensor: Tensor<B, D>, k: usize)
     // A singular matrix will have a pivot of exactly 0.
     // Due to partial pivoting, if the pivot is 0, all elements below it are also 0.
     // We replace 0 with 1.0 to avoid NaN when dividing 0.0 / 0.0.
-    let is_zero_mask = a_kk.clone().equal_elem(0.0);
+    let epsilon = 1e-7;
```
antimora: This epsilon widening is the biggest concern in the PR, and it's a change to the public `linalg::lu` API semantics rather than to `det`.

Before: `equal_elem(0.0)` replaced only exactly-zero pivots with 1.0. That substitution was safe because partial pivoting guaranteed the column below a zero pivot was also zero (which is what the comment on lines 365-367 still describes: `0.0 / 0.0 -> 1.0 / 1.0 = 1.0`, factorization stays well-defined).

After: any `|pivot| <= 1e-7` is rewritten to 1.0. The invariant no longer holds: a well-conditioned matrix with a small-scale pivot (say 5e-8) still has nonzero entries below, and `update_kth_column` now divides them by 1.0 instead of 5e-8. `P * L * U` no longer reconstructs `A`, silently, for every caller of `linalg::lu`.

The threshold is also wrong per-dtype: 1e-7 is coarse for F64 (which has ~1e-16 precision) and too tight for BF16.

Since `det` upcasts f16/bf16 to f32 and already handles singularity via its own mask at det.rs:151-158, this widening isn't needed for this PR. Suggest reverting to `equal_elem(0.0)` here and keeping singularity handling inside `det`. If a wider guard is genuinely needed for `lu`, it deserves its own PR with dtype-aware epsilon, docstring coverage, and a regression test that asserts `P*L*U == A` for a nonsingular matrix with a small pivot.

Also: the comment on lines 365-367 (`"pivot of exactly 0"`, `"all elements below it are also 0"`) is now stale and actively misleading with respect to what the code does.

crates/burn-tensor/src/tensor/linalg/det.rs
```diff
@@ -0,0 +1,168 @@
...
+/// - `D1`: Must be set to `D - 1`.
+/// - `D2`: Must be set to `D - 2`.
+///
+/// # Panics
+/// This function will panic if:
+/// - The generic parameters do not satisfy `D - 1 == D1`.
+/// - The generic parameters do not satisfy `D - 2 == D2`.
+/// - The input tensor rank `D` is less than 3.
+/// - The last two dimensions of the input tensor are not equal.
+///
```
antimora, line 29: A few additions worth making to the docstring, since the function silently applies numerical behavior users can't see from the signature:

1. The `# Panics` list should include the QFloat case (added in the new `TensorCheck::det` at check.rs:1403-1409). Right now a user hitting it gets a panic with no heads-up from the docs.

2. The `# Performance Note` below says the function "relies on the LU decomposition function under the hood," but this is only true for N >= 4. The 1x1 / 2x2 / 3x3 paths use closed-form expressions. A reader optimizing tiny batched determinants may think they're paying for LU setup.

3. No mention that F16/BF16 inputs are internally upcast to F32 and cast back (lines 75-82). The cast-back can underflow to zero for nonsingular matrices whose determinant falls outside F16's range (a 10x10 matrix of entries ~0.1 has det ~1e-10 -> F16 zero). Worth a short `# Numerical behavior` note so callers who care can stay in F32.

4. No mention of the singular-matrix epsilon clamp at lines 151-158 (|diagonal| <= 1e-5 for F32, 1e-10 for F64 -> det forced to 0). These are absolute thresholds, so a well-scaled near-singular matrix with a legitimate small-but-nonzero determinant will return exactly 0 with no diagnostic. Worth documenting.

crates/burn-tensor/src/tensor/linalg/det.rs
```diff
@@ -0,0 +1,168 @@
...
+        .mask_fill(odd_mask, -1.0)
+        .squeeze_dim(D1 - 1);
+
+    // Compute the determinant of U
+    let u_diag = linalg::diag::<B, D, D1, _>(lu);
+    let mut u_det = u_diag.clone().prod_dim(D1 - 1).squeeze_dim(D1 - 1);
+    let epsilon = match tensor.dtype() {
+        DType::F32 => 1e-5,
+        DType::F64 => 1e-10,
+        _ => 1e-5,
```
antimora: After the upcast on lines 75-82, `tensor.dtype()` here can still be F32, F64, or Flex32. Flex32 silently falls into `_ => 1e-5`, which happens to be correct-ish since Flex32 is F32-precision, but it's easy to read this as unreachable-default rather than intentional coverage.

Two options: (a) make it explicit with `DType::F32 | DType::Flex32 => 1e-5`, or (b) derive the epsilon from the `FloatDType`'s precision metadata so all float dtypes are handled uniformly. Either one keeps future readers from second-guessing whether the fallthrough is an oversight.

crates/burn-backend-tests/tests/tensor/float/linalg/det.rs
```diff
@@ -0,0 +1,486 @@
...
+    det_tensor
+        .into_data()
+        .assert_approx_eq::<FloatElem>(&expected.into_data(), tolerance);
+}
+
+#[test]
+fn test_det_20x20_singular_zero_row() {
+    let device = Default::default();
+    let mut tensor = TestTensor::random([1, 20, 20], Distribution::Default, &device);
+    tensor = tensor.slice_fill(s![.., .., 16], 0.0);
```
antimora, line 97: Name says `_zero_row` but `s![.., .., 16]` with shape `[1, 20, 20]` indexes the last dim (columns), so this zeros column 16 of every matrix. Still yields a singular matrix so the assertion holds, but the name is misleading when grepping or debugging. Either rename to `_zero_column` or change the slice to `s![.., 16, ..]` to actually zero a row.

crates/burn-tensor/src/tensor/linalg/det.rs
antimora: Follow-up on this: `burn_std::FloatDType::finfo()` would be a strict improvement over the hardcoded constants here. It returns `f32::EPSILON` (~1.19e-7), `f64::EPSILON` (~2.22e-16), and importantly `f16::EPSILON` (~9.77e-4) for Flex32, so the fall-through case resolves itself naturally.

```rust
let eps = tensor.dtype().finfo().expect("float dtype").epsilon;
let threshold = (n as f64).sqrt() * eps;
```

One caveat: raw machine epsilon isn't quite the right semantic for a singularity threshold, it's the intrinsic precision of the type. A threshold usually wants scaling: LAPACK-style is `N * max(|A|) * eps`. The fully scale-aware version would be `u_diag.abs().max() * (n as f64) * eps`, which also fixes the "matrix scaled by 1000x suddenly looks singular" failure mode the current absolute `1e-5` has.

Even the plain `finfo().epsilon` (without scaling) is strictly better than the current constants though: self-documenting, dtype-correct, and Flex32 falls out for free.

crates/burn-tensor/src/tensor/linalg/lu.rs
antimora: On whether to use `finfo` here instead: I'd say no. `finfo().epsilon` is still much larger than the "pivot with legitimate small scale" values that would get silently corrupted, so it has the same problem as `1e-7`, just per-dtype.

`finfo().min_positive` (smallest normal, e.g. `f32::MIN_POSITIVE ≈ 1.18e-38`) would narrowly catch only denormal underflow without touching real pivots, but at that point `equal_elem(0.0)` is simpler and equally correct for the original NaN-avoidance goal. `finfo` fits well in `det`'s singularity mask (see the reply on det.rs:154), but the pivot-replacement here is better off staying exact-zero.

crates/burn-backend-tests/tests/tensor/float/linalg/det.rs
softmaximalist: Good catch. The `test_det_3x3_singular_zero_row` test case has same issue. I will correct both to use `col`.

softmaximalist: Fixed in c91297e

crates/burn-tensor/src/tensor/linalg/det.rs
softmaximalist: Addressed in 6f1fedb

crates/burn-tensor/src/tensor/linalg/det.rs
softmaximalist: Fixed in 4e561a6

crates/burn-tensor/src/tensor/linalg/lu.rs
softmaximalist: Fixed in c5cf318

softmaximalist:
> Two tests worth adding if easy (happy to defer either):
> 
> 1. Mixed singular/non-singular batch through the LU branch. Shape `[B>=2, N>=4, N]` with one singular and one well-conditioned matrix, asserting only the singular entry returns 0. The current batched singular tests all have batch size 1 or hit the 3x3 fast path, so the `any_dim` + `mask_fill` reduction at det.rs:156-158 is never exercised with partial singularity.
> 2. Explicit dtype round-trip for f16/bf16, asserting `det_tensor.dtype() == original_dtype`. The cast-back logic is duplicated across four return paths (lines 89, 99, 117, 164) and `into_data()` comparisons drop dtype, so CI would not catch one of those branches being removed.

Added in 7546258

crates/burn-tensor/src/tensor/linalg/det.rs
```diff
@@ -0,0 +1,182 @@
...
+/// - `D1`: Must be set to `D - 1`.
+/// - `D2`: Must be set to `D - 2`.
+///
+/// # Panics
+/// This function will panic if:
+/// - The generic parameters do not satisfy `D - 1 == D1`.
+/// - The generic parameters do not satisfy `D - 2 == D2`.
+/// - The input tensor rank `D` is less than 3.
+/// - The last two dimensions of the input tensor are not equal.
+/// - The input is a quantized tensor with with dtype `DType::QFloat`.
```
antimora: "with with" -> "with"

antimora (approved):
Small typo issue but looks good overall; 

Approving in advance.

crates/burn-tensor/src/tensor/linalg/det.rs
laggui:
```suggestion
/// - The input is a quantized tensor with dtype `DType::QFloat`.
```

PR #4818, opened 2026-04-16
Prep for Group Multi Optimizers

crutcher:
I am working on composite optimizers; and realized that there was no safe way to get access to the existing SimpleOptimizers.

- expose OptimizerAdaptor::optim() (to Clone configs built the old way).
- expose {Optim}Config::init_simple() (to build SimpleOptimizer).


crutcher:
This is to prevent the need to do this:
```rust
/// A single optimizer instance and its assigned parameter set.
#[derive(Clone)]
pub struct GroupOptimizer<B, O>
where
    B: AutodiffBackend,
    O: SimpleOptimizer<B::InnerBackend>,
{
    pub params: HashSet<ParamId>,
    pub optim: O,
    phantom: PhantomData<B>,
}

#[allow(unused)]
struct XOptimizerAdaptor<O, M, B>
where
    O: SimpleOptimizer<B::InnerBackend>,
    M: AutodiffModule<B>,
    B: AutodiffBackend,
{
    optim: O,
    records: HashMap<ParamId, AdaptorRecord<O, B>>,
    module: PhantomData<M>,
    grad_clipping: Option<GradientClipping>,
}

impl<B, O> GroupOptimizer<B, O>
where
    B: AutodiffBackend,
    O: SimpleOptimizer<B::InnerBackend>,
{
    pub fn new(
        params: HashSet<ParamId>,
        optim: O,
    ) -> Self {
        Self {
            params,
            optim,
            phantom: PhantomData,
        }
    }

    /// Build a [`GroupOptimizer`] from a [`OptimizerAdaptor`].
    pub fn from_adaptor<M>(
        params: HashSet<ParamId>,
        adaptor: &OptimizerAdaptor<O, M, B>
    ) -> Self
    where M: AutodiffModule<B>
    {
        let optim: O = unsafe {
            let adaptor: &XOptimizerAdaptor<O, M, B> = std::mem::transmute(adaptor);
            adaptor.optim.clone()
        };

        Self::new(
            params,
            optim,
        )
    }
}
```

crates/burn-optim/src/optim/adagrad.rs
```diff
@@ -73,23 +73,30 @@ impl<B: Backend> SimpleOptimizer<B> for AdaGrad {
...
     ///
     /// # Returns
     ///
-    /// Returns an optimizer that can be used to optimize a module.
-    pub fn init<B: AutodiffBackend, M: AutodiffModule<B>>(
-        &self,
-    ) -> OptimizerAdaptor<AdaGrad, M, B> {
-        let optim = AdaGrad {
+    /// The base [`SimpleOptimizer`] utility type.
+    pub fn init_simple(&self) -> AdaGrad {
```
laggui: `SimpleOptimizer` is the trait. The constructor probably shouldn't mention that it returns a simple optimizer, rather it returns the "raw" optimizer type (not wrapped by `OptimizerAdaptor`).

So in this case, I'd simply make the public method `pub fn build(&self) -> AdaGrad`.

The same goes for the other optimizers.

laggui (commented):
LGTM! Just a minor comment regarding the "raw" optimizer constructors.

crates/burn-optim/src/optim/adagrad.rs
crutcher: Done!

laggui (approved):

laggui:
Looks like clippy lints got an update with the latest rust version, will fix this in a separate PR.

PR #4819, opened 2026-04-16
Add ctc_loss backend trait hook + tch and cubecl impls

### LINKED ISSUE #4817: CTCLoss: fuse the forward-backward loop to reduce per-timestep kernel dispatch overhead
## Summary

The current `CTCLoss` implementation in `burn-nn/src/loss/ctc.rs` loops over timesteps sequentially (line 154: `for t in 1..max_input_length`), issuing ~40 small tensor operations per iteration (two `right_shift_2d_tensor` calls, two `log_sum_exp` calls with ~12 ops each, a `gather`, `add`, and `mask_where`). For typical ASR sequence lengths (T = 1000-2000 frames), this results in ~40,000-80,000 individual kernel dispatches per loss call.

On GPU backends (wgpu/cubecl, LibTorch-MPS), each dispatch has fixed overhead from CPU-side submission and GPU-side scheduling. The sequential dependency (timestep `t` depends on `t-1`) prevents the backend from batching or overlapping these kernels. In profiling, this manifests as oscillating CPU/GPU utilization during the loss computation — the GPU is underutilized because it spends time waiting between tiny kernels rather than doing sustained compute.

## Current behavior

Each CTC forward pass on a batch with `max_input_length = T` executes roughly:
- `T` iterations of the alpha recursion
- Per iteration: 2x `log_sum_exp` (each ~12 tensor ops for the NaN-safe path), 2x `right_shift_2d_tensor` (slice + cat), 1x `gather`, 1x `add`, 1x `mask_where`
- Total: ~40T kernel dispatches

The `log_sum_exp` helper (line 402) is particularly heavy — it handles `-inf` entries safely via masking, which requires ~12 ops per call. For the common case where most entries are finite, a simpler `max + log(1 + exp(diff))` with a single `-inf` guard would halve the op count.

### 1. Simplify `log_sum_exp` (low effort, ~2x fewer ops per timestep)

The current implementation creates multiple intermediate mask tensors to handle `-inf` entries. A more compact formulation:

```rust
fn log_sum_exp(a, b) -> Tensor {
    let max_val = a.clone().max_pair(b.clone());
    let sum = (a - max_val.clone()).exp() + (b - max_val.clone()).exp();
    max_val + sum.log()
}
```

This produces `-inf + log(0+0) = -inf` naturally when both inputs are `-inf`, without explicit masking. For the mixed case (`-inf` + finite), `exp(-inf - finite) = 0`, so the result is `finite + log(0 + 1) = finite`. Correct in all cases, ~4 ops instead of ~12.

### 2. Fused cubecl kernel for the alpha recursion (high effort, eliminates the loop)

The forward-backward algorithm has a well-known parallel structure along the label axis (each `s` position can be computed independently given `t-1`). A single cubecl kernel that loops over `t` internally — reading `alpha[t-1]` from shared memory and writing `alpha[t]` — would replace ~40T host-side dispatches with a single kernel launch. This is what PyTorch's native `CTCLoss` CUDA kernel does.

This would be a larger contribution but would make Burn's CTC loss competitive with PyTorch's for real ASR workloads (T > 500).

## Environment

- burn 0.21.0-pre.3 (git rev b62be4dd)
- Observed on both wgpu (Metal/cubecl) and LibTorch-MPS backends

antimora:
Closes #4817.

`ModuleOps::ctc_loss` is added with a default implementation (the existing
algorithm, lifted from `burn-nn` to `burn-backend`). burn-nn now delegates.

Backend overrides:
- `burn-tch`: dispatches to `tch::Tensor::ctc_loss_tensor` (PyTorch's native kernel).
- `burn-cubecl`: fused single-launch alpha recursion - one cube per batch
  element, alpha state in shared memory, time loop sequential inside the
  kernel with `sync_cube()` between steps. Falls back to the default impl
  when `2*S+1 > 1024`.

Drive-by: `burn-tch::int_from_data` now accepts I32/I16/I8/U8 (was I64-only),
unblocking int-using cross-backend tests on tch.

## Test plan
- [x] burn-nn ctc tests pass (15 total, includes PyTorch grad comparisons)
- [x] cross-backend tests pass on ndarray, tch, wgpu (with and without fusion)
- [x] burn-tch unit test for the LibTorch dispatch
- [x] `cargo check --workspace` clean


crates/burn-cubecl/src/kernel/ctc.rs
```diff
@@ -0,0 +1,189 @@
...
+        cube_count,
+        cube_dim,
+        log_probs.into_tensor_arg(),
+        targets.into_tensor_arg(),
+        input_lengths.into_tensor_arg(),
+        target_lengths.into_tensor_arg(),
+        output.clone().into_tensor_arg(),
+        blank as u32,
+        max_l_prime,
+    );
```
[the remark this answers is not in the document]
antimora: Fixed in 411e51e99 (the `dtype generics, autodiff, preconditions` commit). The launch now uses `ctc_loss_kernel::launch::<R>(...)` with `#[define(F, I)] _dtypes: [StorageType; 2]` in the kernel signature, so F and I are resolved at launch time from `log_probs.dtype` / `targets.dtype`. f16/bf16 and i64/i32 targets all dispatch through the correctly compiled types.

crates/burn-backend-tests/tests/tensor/float/module/ctc.rs
```diff
@@ -0,0 +1,56 @@
...
+    let target_lengths = TestTensorInt::<1>::from([2]);
+
+    let loss = ctc_loss(log_probs, targets, input_lengths, target_lengths, 0);
+
+    let expected = burn_tensor::TensorData::from([3.0f32 * 2.0f32.ln()]);
+    loss.into_data()
+        .assert_approx_eq::<FloatElem>(&expected, Tolerance::rel_abs(1e-3, 1e-3));
+}
+
+#[test]
```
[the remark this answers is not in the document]
antimora: Added `test_ctc_loss_empty_target` in b4446d24e. Used a `[1, 1]` targets shape with `target_lengths=[0]` rather than `[1, 0]` since not every backend supports zero-sized dimensions; the target slot is never read because `target_lengths[0] == 0`. Also added `test_ctc_loss_grad_empty_target` in the autodiff suite and `test_ctc_loss_unreachable_target_is_inf` to pin down the `+inf` contract for the `target_length > input_length` case.

crates/burn-backend/src/backend/ops/modules/ctc.rs
```diff
@@ -0,0 +1,362 @@
...
+    // Extract final log-likelihood: log_sum_exp(alpha[2*U], alpha[2*U-1])
+    let last_blank_idx = B::int_mul_scalar(target_lengths.clone(), 2.into());
+    let last_blank_idx = B::int_reshape(last_blank_idx, Shape::new([batch_size, 1]));
+    let last_label_idx = B::int_sub_scalar(last_blank_idx.clone(), 1.into());
+
+    let log_alpha_last_blank = B::float_gather(1, log_alpha.clone(), last_blank_idx);
+    let log_alpha_last_blank = B::float_reshape(log_alpha_last_blank, Shape::new([batch_size]));
+
+    let log_alpha_last_label = B::float_gather(1, log_alpha, last_label_idx);
+    let log_alpha_last_label = B::float_reshape(log_alpha_last_label, Shape::new([batch_size]));
```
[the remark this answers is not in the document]
antimora: Fixed in 411e51e99. `extract_loss` now clamps the label index with `B::int_clamp_min(last_blank_idx - 1, 0)` to prevent the underflow, then masks batches where `target_length == 0` to -inf so the label-position contribution degenerates to `log_sum_exp(alpha[0], -inf) = alpha[0]`. Covered by the new `test_ctc_loss_empty_target` in b4446d24e.

crates/burn-backend/src/backend/ops/modules/ctc.rs
```diff
@@ -0,0 +1,362 @@
...
+        let dest_col = 2 * s + 1;
+        result = B::int_slice_assign(
+            result,
+            &[
+                Slice::full(),
+                Slice::new(dest_col as isize, Some(dest_col as isize + 1), 1),
+            ],
+            label,
+        );
+    }
```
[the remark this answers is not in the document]
antimora: Good catch, fixed in b4446d24e. Replaced the O(S) loop with a single strided assignment equivalent to `result[:, 1::2] = targets`:

```rust
B::int_slice_assign(
    result,
    &[Slice::full(), Slice::new(1, None, 2)],
    targets.clone(),
)
```

That collapses `S` backend calls into one. All CTC tests still pass across ndarray, wgpu, and wgpu+fusion.

crates/burn-cubecl/src/kernel/ctc.rs
```diff
@@ -0,0 +1,189 @@
...
+    // Compute l'[s-2] class (only meaningful for s >= 2)
+    let mut l_class_m2 = blank_u;
+    if active && s >= 2 && (s - 2) % 2 == 1 {
+        l_class_m2 = u32::cast_from(targets[n * tgt_n + ((s - 2 - 1) / 2) * tgt_s]) as usize;
+    }
+    let skip_allowed = active && s >= 2 && l_class != blank_u && l_class != l_class_m2;
+
+    let mut alpha = SharedMemory::<F>::new(max_lp);
+    let neg_inf = F::min_value();
+    let one = F::new(1.0);
```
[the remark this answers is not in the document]
antimora: Valid concern, fixed in b4446d24e. The fused kernel now uses `F::new(f32::NEG_INFINITY)` as the impossible-alignment sentinel (expressed via `F::new` rather than the `F::NEG_INFINITY` const because the trait const runs into an `IntoMut` bound inside `#[cube]` expansion). Every `log_sum_exp` guards the `-inf - -inf = NaN` case:

```rust
let lse = if mx == neg_inf {
    mx
} else {
    mx + (one + (mn - mx).exp()).ln()
};
```

Since `mx` is the max, `mx == -inf` implies `mn == -inf`, so returning `mx` is correct. Added `test_ctc_loss_unreachable_target_is_inf` (cross-backend) that passes `target_length > input_length` and asserts the loss is `+inf`. That test exercises the fused cubecl path on wgpu / wgpu+fusion and confirms `is_inf` now fires, which means burn-nn's `zero_infinity` masking works against the fused backend.

antimora:
Root cause of the broken `Autodiff<Wgpu>` training path (loss ~0.18 instead of ~1250) is tracel-ai/cubecl#68: `F::new(f32::NEG_INFINITY)` inside a `#[cube]` kernel body emits `f32(-inf)` in the generated WGSL, which isn't a valid WGSL identifier. Shader compilation fails silently, the kernel never runs, and the output buffer reads back as its uninitialized contents.

This wasn't caught by CI because `burn-backend-tests` routes through `Dispatch`, which had no `ctc_loss` override - so every test fell through to the tensor-ops default path instead of exercising the fused kernel.

Commit `c784d9be9` adds `ctc_loss` / `ctc_loss_backward` overrides to `Dispatch`, which makes CI actually exercise the kernel on whichever backend the harness is configured for.

Commit `1c2c7f2ea` works around the cubecl issue: internal `-inf` sentinel becomes `F::new(-1.0e30_f32)`, and unreachable alignments emit true `+inf` at the final output via runtime `exp(1000 * input_len)` so downstream `is_inf` checks still fire for `zero_infinity` masking.

antimora:
@louisfd @nathanielsimard @wingertge

crates/burn-nn/src/loss/ctc.rs
```diff
@@ -246,6 +198,41 @@ impl CTCLoss {
...
+            assert!(tl >= 0, "target_lengths[{i}] = {tl} must be non-negative");
+            assert!(
+                tl as usize <= max_target_len,
+                "target_lengths[{i}] = {tl} exceeds the targets tensor width {max_target_len}"
+            );
+            assert!(
+                il >= tl,
+                "input_lengths[{i}] = {il} must be >= target_lengths[{i}] = {tl} (no valid CTC alignment otherwise)"
+            );
+        }
```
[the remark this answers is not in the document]
antimora: Gated `length_assertions` behind `cfg(debug_assertions)` to avoid the device-to-host sync in release builds. Also added the `input_lengths[i] <= max_input_length` check (see comment 3102404633 below).

crates/burn-cubecl/src/kernel/ctc.rs
```diff
@@ -0,0 +1,645 @@
...
+    let cube_dim = CUBE_DIM_X as usize;
+    let alpha_cap = alpha_capacity as usize;
+    let blank_u = blank as usize;
+
+    let target_len = u32::cast_from(target_lengths[n]) as usize;
+    let input_len = u32::cast_from(input_lengths[n]) as usize;
+    let l_prime_len = 2 * target_len + 1;
+
+    let lp_t = log_probs.stride(0);
+    let lp_n = log_probs.stride(1);
```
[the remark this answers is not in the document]
antimora: The module layer (`CTCLoss::forward`) now validates `input_lengths[i] <= max_input_length` in `length_assertions` (debug builds). This catches invalid lengths before they reach any backend kernel. Adding a second validation layer at the cubecl launch site would duplicate the check; the module layer is the correct boundary for user-facing preconditions.

crates/burn-cubecl/src/kernel/ctc.rs
```diff
@@ -0,0 +1,645 @@
...
+/// `alpha[t-1]` before any thread writes `alpha[t]`, one to publish the new row
+/// before the next iteration. This collapses what would otherwise be roughly
+/// `40 * T` host-side dispatches into a single kernel launch.
+///
+/// Impossible alignments are represented as `F::NEG_INFINITY` so that when an
+/// entire sequence has no valid alignment (e.g. `target_length > input_length`)
+/// the forward loss comes out as `+inf`, which is what `zero_infinity` masking
+/// in `burn-nn` detects via `is_inf`. Each `log_sum_exp` guards `mx == -inf`
+/// explicitly to avoid the `-inf - -inf = NaN` case that would otherwise
+/// propagate through the recursion.
```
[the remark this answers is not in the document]
antimora: Updated the doc comment to accurately describe the sentinel/threshold approach instead of claiming `F::NEG_INFINITY`.

crates/burn-nn/src/loss/ctc.rs
```diff
@@ -246,6 +198,41 @@ impl CTCLoss {
...
+    /// - `target_lengths[i] >= 0`
+    /// - `input_lengths[i] >= target_lengths[i]` (otherwise no valid CTC
+    ///   alignment exists; without this check the loss silently degenerates
+    ///   to +inf, which `zero_infinity` would then mask)
+    fn length_assertions<B: Backend>(
+        &self,
+        input_lengths: Tensor<B, 1, Int>,
+        target_lengths: Tensor<B, 1, Int>,
+        max_target_len: usize,
+    ) {
```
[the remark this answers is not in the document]
antimora: Added `input_lengths[i] <= max_input_length` assertion in `length_assertions`, gated behind `cfg(debug_assertions)` along with the rest of the per-element checks.

antimora:
Rebased

crates/burn-autodiff/src/ops/module.rs
```diff
@@ -1884,6 +1884,87 @@ impl<B: Backend, C: CheckpointStrategy> ModuleOps<Autodiff<B, C>> for Autodiff<B
...
+    ) -> FloatTensor<Autodiff<B, C>> {
+        #[derive(Debug)]
+        struct CtcLoss;
+
+        impl<B: Backend> Backward<B, 1> for CtcLoss {
+            type State = (
+                NodeId,
+                burn_backend::tensor::IntTensor<B>,
+                burn_backend::tensor::IntTensor<B>,
+                burn_backend::tensor::IntTensor<B>,
```
nathanielsimard: Can we import IntTensor to reduce the size of the type?

crates/burn-backend/src/backend/ops/modules/base.rs
```diff
@@ -1132,6 +1132,64 @@ pub trait ModuleOps<B: Backend> {
...
+    /// * `targets` - Target label indices of shape `[N, S]`
+    /// * `input_lengths` - Actual input sequence lengths per batch element `[N]`
+    /// * `target_lengths` - Actual target lengths per batch element `[N]`
+    /// * `grad_loss` - Upstream gradient w.r.t. the per-sample loss `[N]`
+    /// * `blank` - Index of the blank label
+    ///
+    /// # Returns
+    ///
+    /// Gradient w.r.t. `log_probs` of shape `[T, N, C]`
+    fn ctc_loss_backward(
```
nathanielsimard, line 1195: Ok this is very big. I don't think we should provide a default implementation for the backward pass. Maybe we should provide a way to flag if the backward pass is implemented, so the autodiff can use this flag to use it, and if not use the normal `ctc_loss_default` implementation for the forward, which will automatically calculate the backward pass. I think that would help maintenance and doesn't reduce performance much, since most backends should implement a custom kernel anyways.

crates/burn-fusion/src/ops/module.rs
```diff
@@ -1631,4 +1631,130 @@ impl<B: FusionBackend> ModuleOps<Fusion<B>> for Fusion<B> {
...
+    fn ctc_loss(
+        log_probs: FloatTensor<Fusion<B>>,
+        targets: IntTensor<Fusion<B>>,
+        input_lengths: IntTensor<Fusion<B>>,
+        target_lengths: IntTensor<Fusion<B>>,
+        blank: usize,
+    ) -> FloatTensor<Fusion<B>> {
+        // CTC is non-fuseable (it's one dense kernel call on backends that
+        // implement it, or a 40*T op chain on the default path). Going through
+        // the fusion stream's default path produces wrong results on lazy
```
nathanielsimard: If there is a bug with fusion, we should flag it and fix it. I don't think having a workaround is good here.

crates/burn-fusion/src/ops/module.rs
```diff
@@ -1631,4 +1631,130 @@ impl<B: FusionBackend> ModuleOps<Fusion<B>> for Fusion<B> {
...
+        let client = log_probs.client.clone();
+        let batch_size = log_probs.shape[1];
+        let out = TensorIr::uninit(
+            client.create_empty_handle(),
+            Shape::new([batch_size]),
+            log_probs.dtype,
+        );
+
+        let desc = CustomOpIr::new(
+            "ctc_loss",
```
nathanielsimard: I think we should create a CtcLoss in the IR instead of a custom op.

crates/burn-autodiff/src/ops/module.rs
antimora: Fixed in 69c0fe6. The `IntTensor` alias is already imported at the top of the file, so the function now uses it directly in the parameters, the `Backward::State` tuple, and the inner `CtcLoss` impl block.

crates/burn-backend/src/backend/ops/modules/base.rs
antimora: Done in 69c0fe6. Added `has_ctc_loss_backward() -> bool { false }` to `ModuleOps`. Autodiff now branches on it: when `false`, it calls `ctc_loss_default::<Self>(...)`, which is built entirely from differentiable tensor ops, so autodiff derives the gradient automatically. Backends with native kernels (`burn-tch`, `burn-cubecl`) override both methods and the flag.

The default `ctc_loss_backward` body is now `unreachable!`. Removed the now-dead `ctc_loss_backward_default`, `compute_log_beta_full`, `left_shift`, and `create_l_prime_skip_forward_mask` helpers. The shared composition helper `ctc_grad_from_alpha_beta_default` stays since the cubecl backward kernel still uses it.

`Fusion<B>` and `Dispatch` got matching forwarders. Dispatch returns `false` unconditionally because per-tensor backend choice is a runtime decision but autodiff queries the flag statically; differentiating through the decomposed forward is correct for every inner backend.

All CTC tests still pass on libtorch and metal+fusion (the same fusion-over-cubecl topology that exposed the original wrong-output bug).

crates/burn-fusion/src/ops/module.rs
antimora: Reframed in 69c0fe6. The previous comment misnamed the situation: there was no fusion bug, just an architectural mismatch from decomposing a 40T-op chain through fusion's stream model. Routing CTC as its own non-fuseable IR node is the correct pattern, the same one attention, rfft, and the conv ops already use. With the new typed IR variants, the comment now describes the standard pattern instead of a workaround.

crates/burn-fusion/src/ops/module.rs
antimora: Done in 69c0fe6. Added `CtcLoss(CtcLossOpIr)` and `CtcLossBackward(CtcLossBackwardOpIr)` variants to `ModuleOperationIr`, with `create()` builders via `impl_ir_create!` and matching arms in `inputs()`, `outputs()`, `mark_read_only()`, the `to_relative` converter in `burn-fusion::stream::context`, and the dispatch in `burn-router`'s `runner.rs`. Fusion now uses `make_ops!(CtcLossOps, CtcLossOpIr, ...)` exactly like `attention` and `rfft`. `CustomOpIr` is no longer used for CTC.

crates/burn-flex/src/ops/module.rs
```diff
@@ -748,6 +748,26 @@ impl ModuleOps<Flex> for Flex {
...
+        // GPU backends), so the decomposed alpha algorithm runs at native CPU
+        // speed without needing a monolithic flex-specific implementation.
+        burn_backend::ops::ctc::ctc_loss_default::<Flex>(
+            log_probs,
+            targets,
+            input_lengths,
+            target_lengths,
+            blank,
+        )
+    }
```
laggui: This is already the default implementation, so there is no need to override it?

laggui (commented):
Removing the default backward implementation was the good way to go!

I'll assume the cubecl kernels are correct given the tests, just a minor comment otherwise.

crates/burn-flex/src/ops/module.rs
antimora: Good catch, removed in 04677d5. The body was identical to the trait default; letting the default handle it drops the duplicate surface.

laggui (approved):
Just fixed some minor conflicts with main

PR #4839, opened 2026-04-20
Device service usage

nathanielsimard:
Update from cubecl and cubek main


laggui (approved):

laggui:
`should_diff_swap_dims` is failing locally (both with and without fusion), might be related to recent matmul changes.

PR #4842, opened 2026-04-20
Add Blackman window function to signal module

### LINKED ISSUE #4783: Add Blackman window to signal module
## Summary

Add `blackman_window` to `burn::tensor::signal`, alongside the existing `hamming_window` and `hann_window`.

## Motivation

The `burn::tensor::signal` module already provides `hamming_window` and `hann_window`. Blackman is the third standard window function and is needed to implement the ONNX [`BlackmanWindow`](https://onnx.ai/onnx/operators/onnx__BlackmanWindow.html) operator in burn-onnx (see tracel-ai/burn-onnx#161). ONNX ships all three (Hamming, Hann, Blackman) as sibling ops, and burn-onnx already has the first two — Blackman is the missing piece.

## Proposed API

Mirror the existing window functions exactly:

```rust
pub fn blackman_window<B: Backend>(
    size: usize,
    periodic: bool,
    options: impl Into<TensorCreationOptions<B>>,
) -> Tensor<B, 1>;
```

## Reference behavior

- torch.blackman_window: <https://pytorch.org/docs/stable/generated/torch.blackman_window.html>
- ONNX BlackmanWindow spec: <https://onnx.ai/onnx/operators/onnx__BlackmanWindow.html>

## Mathematical definition

`w[n] = 0.42 - 0.5 * cos(2πn/N) + 0.08 * cos(4πn/N)`

where `N = size` when `periodic` is true, or `N = size - 1` when `periodic` is false.

Same `size == 0` / `size == 1` edge-case handling as the existing window functions.

## Implementation sketch

Should be a near-copy of `hann_window.rs`, just with two cosine terms instead of one:

```rust
Tensor::<B, 1, Int>::arange(0..size_i64, &opt.device)
    .float()
    .mul_scalar(angular_increment)
    .cos()
    .mul_scalar(-0.5)
    .add_scalar(0.42)
    + arange * (2 * angular_increment) cos * 0.08
```

(actual code would fuse the two cos terms more cleanly).

## Related

- tracel-ai/burn-onnx#161 — tracks the ONNX-side signal ops; BlackmanWindow is blocked on this.
- tracel-ai/burn#4782 — companion request for STFT.

softmaximalist:
### Checklist

### Related Issues/PRs

Related to [Implement signal processing operators (DFT, STFT, MelWeightMatrix, window functions) #161](https://github.com/tracel-ai/burn-onnx/issues/161) and closes [Add Blackman window to signal module #4783](https://github.com/tracel-ai/burn/issues/4783)

### Changes

Added Blackman window function using the following formula instead to only compute cosine once instead of twice from the original formula.

$$
\begin{aligned}
w[n] &= 0.42 - 0.5\cos\left(\frac{2\pi n}{N - 1}\right) + 0.08\cos\left(\frac{4\pi n}{N - 1}\right) \\
w[n] &= 0.42 - 0.5\cos\left(\frac{2\pi n}{N - 1}\right) + 0.08\cos\left(2 \left(\frac{2\pi n}{N - 1}\right)\right) \\
w[n] &= 0.42 - 0.5\cos\left(\frac{2\pi n}{N - 1}\right) + 0.08\left(2\cos^2\left(\frac{2\pi n}{N - 1}\right) - 1\right) \\
w[n] &= 0.34 - 0.5\cos\left(\frac{2\pi n}{N - 1}\right) + 0.16\cos^2\left(\frac{2\pi n}{N - 1}\right)
\end{aligned}
$$

### Testing

Added 8 test cases covering sizes of 0, 1, and 8 with `periodic` set to `true` and `false`, explicitly setting dtype, and panics.

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -440,13 +440,13 @@ strategies.
 
 ## Signal Processing Functions
 
-| Burn API                                           | PyTorch Equivalent                     |  
-| -------------------------------------------------- | -------------------------------------- |  
-| `signal::blackman_window(size, periodic, options)` | `torch.blackman_window(size, periodic)`|
-| `signal::hamming_window(size, periodic, options)`  | `torch.hamming_window(size, periodic)` |
-| `signal::hann_window(size, periodic, options)`     | `torch.hann_window(size, periodic)`    |
-| `signal::irfft(spectrum_re, spectrum_im, dim)`     | `torch.fft.irfft(input, dim=dim)`      |
-| `signal::rfft(signal, dim)`                        | `torch.fft.rfft(signal, dim=dim)`      |  
+| Burn API                                           | PyTorch Equivalent                      |
+| -------------------------------------------------- | --------------------------------------- |
+| `signal::blackman_window(size, periodic, options)` | `torch.blackman_window(size, periodic)` |
+| `signal::hamming_window(size, periodic, options)`  | `torch.hamming_window(size, periodic)`  |
+| `signal::hann_window(size, periodic, options)`     | `torch.hann_window(size, periodic)`     |
+| `signal::irfft(spectrum_re, spectrum_im, dim)`     | `torch.fft.irfft(input, dim=dim)`       |
+| `signal::rfft(signal, dim)`                        | `torch.fft.rfft(signal, dim=dim)`       |
 
 ## Displaying Tensor Details
 
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -440,13 +440,13 @@ strategies.
 
 ## Signal Processing Functions
 
-| Burn API                                           | PyTorch Equivalent                    |  
-| -------------------------------------------------- | ------------------------------------- |  
-| `signal::blackman_window(size, periodic, options)` | `torch.hamming_window(size, periodic)`|
-| `signal::hamming_window(size, periodic, options)`  | `torch.hamming_window(size, periodic)`|
-| `signal::hann_window(size, periodic, options)`     | `torch.hann_window(size, periodic)`   |
-| `signal::irfft(spectrum_re, spectrum_im, dim)`     | `torch.fft.irfft(input, dim=dim)`     |
-| `signal::rfft(signal, dim)`                        | `torch.fft.rfft(signal, dim=dim)`     |  
+| Burn API                                           | PyTorch Equivalent                     |  
+| -------------------------------------------------- | -------------------------------------- |  
+| `signal::blackman_window(size, periodic, options)` | `torch.blackman_window(size, periodic)`|
+| `signal::hamming_window(size, periodic, options)`  | `torch.hamming_window(size, periodic)` |
+| `signal::hann_window(size, periodic, options)`     | `torch.hann_window(size, periodic)`    |
+| `signal::irfft(spectrum_re, spectrum_im, dim)`     | `torch.fft.irfft(input, dim=dim)`      |
+| `signal::rfft(signal, dim)`                        | `torch.fft.rfft(signal, dim=dim)`      |  
 
 ## Displaying Tensor Details
 
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -438,6 +438,16 @@ strategies.
 | `linalg::vector_norm(tensor, p, dim)`              | `torch.linalg.vector_norm(tensor, p, dim)`          |
 | `linalg::vector_normalize(tensor, norm, dim, eps)` | `nn.functional.normalize(tensor, p, dim, eps)`      |
 
+## Signal Processing Functions
+
+| Burn API                                           | PyTorch Equivalent                    |  
+| -------------------------------------------------- | ------------------------------------- |  
+| `signal::blackman_window(size, periodic, options)` | `torch.hamming_window(size, periodic)`|
+| `signal::hamming_window(size, periodic, options)`  | `torch.hamming_window(size, periodic)`|
+| `signal::hann_window(size, periodic, options)`     | `torch.hann_window(size, periodic)`   |
+| `signal::irfft(spectrum_re, spectrum_im, dim)`     | `torch.fft.irfft(input, dim=dim)`     |
+| `signal::rfft(signal, dim)`                        | `torch.fft.rfft(signal, dim=dim)`     |  
+
 ## Displaying Tensor Details
 
 Burn provides flexible options for displaying tensor information, allowing you to control the level
--- a/crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
@@ -75,10 +75,3 @@ fn test_blackman_window_size_8_symmetric() {
         .into_data()
         .assert_approx_eq::<FloatElem>(&expected, tolerance);
 }
-
-#[test]
-#[should_panic(expected = "The argument `size` should be less than or equal to `i64::MAX`.")]
-fn test_blackman_window_invalid_size_panics() {
-    let window_len = (i64::MAX as usize) + 1;
-    let _: TestTensor<1> = blackman_window(window_len, false, &Default::default());
-}
--- a/crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
@@ -40,14 +40,7 @@ fn should_handle_blackman_window_size_1_periodic() {
 fn should_support_blackman_window_size_8_periodic() {
     let tensor: TestTensor<1> = blackman_window(8, true, &Default::default());
     let expected = TensorData::from([
-        -2.9802e-08,
-        6.6447e-02,
-        3.4000e-01,
-        7.7355e-01,
-        1.0000e+00,
-        7.7355e-01,
-        3.4000e-01,
-        6.6447e-02,
+        0.0, 6.6447e-02, 3.4000e-01, 7.7355e-01, 1.0000e+00, 7.7355e-01, 3.4000e-01, 6.6447e-02,
     ]);
 
     // Positions 0 and 7 have values close to 0, hence setting absolute tolerance
@@ -61,14 +54,7 @@ fn should_support_blackman_window_size_8_periodic() {
 fn should_support_blackman_window_size_8_symmetric() {
     let tensor: TestTensor<1> = blackman_window(8, false, &Default::default());
     let expected = TensorData::from([
-        -2.9802e-08,
-        9.0453e-02,
-        4.5918e-01,
-        9.2036e-01,
-        9.2036e-01,
-        4.5918e-01,
-        9.0453e-02,
-        -2.9802e-08,
+        0.0, 9.0453e-02, 4.5918e-01, 9.2036e-01, 9.2036e-01, 4.5918e-01, 9.0453e-02, 0.0,
     ]);
 
     // Positions 0 and 7 have values close to 0, hence setting absolute tolerance
--- a/crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
@@ -1,0 +1,84 @@
+use super::*;
+use burn_tensor::signal::blackman_window;
+use burn_tensor::{DType, TensorData, Tolerance};
+
+#[test]
+fn test_blackman_window_options_dtype() {
+    let tensor: TestTensor<1> = blackman_window(4, true, (&Default::default(), DType::F32));
+    assert_eq!(tensor.dtype(), DType::F32);
+}
+
+#[test]
+fn test_blackman_window_size_0_symmetric() {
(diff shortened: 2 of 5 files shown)


crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
```diff
@@ -0,0 +1,84 @@
...
+        .into_data()
+        .assert_approx_eq::<FloatElem>(&expected, tolerance);
+}
+
+#[test]
+#[should_panic(expected = "The argument `size` should be less than or equal to `i64::MAX`.")]
+fn test_blackman_window_invalid_size_panics() {
+    let window_len = (i64::MAX as usize) + 1;
+    let _: TestTensor<1> = blackman_window(window_len, false, &Default::default());
+}
```
antimora: `(i64::MAX as usize) + 1` overflows `usize` on 32-bit targets (e.g. wasm32, which Burn supports), so this test will not compile portably. Either gate with `#[cfg(target_pointer_width = "64")]`, or drop it - on 64-bit `i64::MAX + 1` as a `usize` is still representable so the panic path is only reachable on 64-bit anyway, and neither `hamming_window` nor `hann_window` bothers to cover it. If you keep the test and later change the `expect` message (see my comment on the impl), remember to update the `should_panic(expected = ...)` in lockstep.

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
```diff
@@ -0,0 +1,84 @@
...
+        1.0000e+00,
+        7.7355e-01,
+        3.4000e-01,
+        6.6447e-02,
+    ]);
+
+    let tolerance = Tolerance::default().set_half_precision_absolute(1e-3);
+    tensor
+        .into_data()
+        .assert_approx_eq::<FloatElem>(&expected, tolerance);
```
antimora, line 50: Both sibling tests use `set_half_precision_relative(1e-2)` with a `// Metal has less precise trigonometric functions.` comment (`hamming_window.rs:13`, `hann_window.rs:11`). Switching to `_absolute(1e-3)` changes semantics near zero: the first expected element `-2.9802e-08` passes absolute trivially but would likely fail relative. Was that an intentional workaround for the float artifact at n=0 (from the `cos^2` rewrite), or incidental? If intentional, a one-line comment would help; if incidental, align with siblings. Also applies to line 73.

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
```diff
@@ -0,0 +1,84 @@
+use super::*;
+use burn_tensor::signal::blackman_window;
+use burn_tensor::{DType, TensorData, Tolerance};
+
+#[test]
+fn test_blackman_window_options_dtype() {
```
antimora: Minor: sibling window tests use `should_support_*` / `should_handle_*` naming (see `hamming_window.rs`, `hann_window.rs`). Consider renaming all eight tests here to match - easier to grep across the window family and keeps the test listing sorted consistently.

crates/burn-tensor/src/tensor/signal/blackman_window.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+
+use crate::{Tensor, TensorCreationOptions, check, check::TensorCheck};
+
+/// Creates a 1D Blackman window tensor.
+///
+/// # Equation
+/// The standard Blackman window is defined as:
+/// `w[n] = 0.42 - 0.5 * cos(2πn / N) + 0.08 * cos(4πn / N)`
+/// where `N` is `size` if periodic, or `size - 1` if symmetric.
```
antimora: `hann_window.rs:10-21` and `hamming_window.rs:10-21` use `#[cfg_attr(doc, doc = r#"$$...$$"#)]` so rustdoc renders the formula with KaTeX (with a plain-text fallback for non-doc builds). Consider switching to that pattern for consistency with the sibling window docs.

```suggestion
#[cfg_attr(
    doc,
    doc = r#"
$$w_n = 0.42 - 0.5 \cos\left(\frac{2\pi n}{N}\right) + 0.08 \cos\left(\frac{4\pi n}{N}\right)$$

where $N$ = `size` when `periodic` is `true`, or $N$ = `size - 1` when `periodic` is `false`.
"#
)]
#[cfg_attr(
    not(doc),
    doc = "`w_n = 0.42 - 0.5 * cos(2πn/N) + 0.08 * cos(4πn/N)` where N = size (periodic) or N = size-1 (symmetric)"
)]
```

crates/burn-tensor/src/tensor/signal/blackman_window.rs
```diff
@@ -0,0 +1,88 @@
...
+///     // Creating a window with default dtype
+///     let device = B::Device::default();
+///     let window_tensor = blackman_window::<B>(5, true, &device);
+///     // Output: [0.0, 0.20077015, 0.84922993, 0.8492298, 0.2007701]
+///
+///     // Creating a window with explicit dtype
+///     let device = B::Device::default();
+///     let window_tensor_f64 = blackman_window::<B>(5, true, (&device, DType::F64));
+///     // Output: [0.0, 0.20077015, 0.84922993, 0.8492298, 0.2007701]
+/// }
```
antimora, line 58: Both the F32 and F64 calls show identical `// Output: [...]` literals. That's accidentally correct: `arange(...).float()` computes in the backend's default `FloatElem` (usually F32), and `cast(dtype)` at the end of the function only widens the storage to F64 - it does not recompute at higher precision. So `DType::F64` buys wider storage, not more significant digits. I'd either drop the F64 example to avoid implying otherwise, or add a one-line note to that effect. Separately, the first element is actually about `-2.98e-8` (see the tests), not `0.0`, so the comment is misleading either way.

crates/burn-tensor/src/tensor/signal/blackman_window.rs
```diff
@@ -0,0 +1,88 @@
...
+/// The standard Blackman window is defined as:
+/// `w[n] = 0.42 - 0.5 * cos(2πn / N) + 0.08 * cos(4πn / N)`
+/// where `N` is `size` if periodic, or `size - 1` if symmetric.
+///
+/// # Arguments
+/// - `size`: Size of the returned 1D window tensor.
+/// - `periodic`: If `true`, the window is treated as periodic (i.e., `N = size`).
+///   If `false`, the window is symmetric (i.e., `N = size - 1`).
+/// - `options`: Controls the output device and optional dtype. Accepts:
+///     - `&device` - uses the devices' default float dtype
```
antimora: Nit: "devices' default" -> "device's default" (singular possessive).

crates/burn-tensor/src/tensor/signal/blackman_window.rs
```diff
@@ -0,0 +1,88 @@
...
+    if size == 0 {
+        return Tensor::<B, 1>::empty(shape, opt).cast(dtype);
+    }
+
+    if size == 1 {
+        return Tensor::<B, 1>::ones(shape, opt).cast(dtype);
+    }
+
+    let size_i64 = i64::try_from(size)
+        .expect("The argument `size` should be less than or equal to `i64::MAX`.");
```
antimora: Siblings use `"HannWindow size doesn't fit in i64 range."` / `"HammingWindow size doesn't fit in i64 range."` (`hann_window.rs:58`, `hamming_window.rs:58`). For family-level consistency, consider `"BlackmanWindow size doesn't fit in i64 range."`. If you change this, the `#[should_panic(expected = ...)]` in the test file needs to match.

burn-book/src/building-blocks/tensor.md
```diff
@@ -438,6 +438,16 @@ strategies.
...
 
+## Signal Processing Functions
+
+| Burn API                                           | PyTorch Equivalent                     |  
+| -------------------------------------------------- | -------------------------------------- |  
+| `signal::blackman_window(size, periodic, options)` | `torch.blackman_window(size, periodic)`|
+| `signal::hamming_window(size, periodic, options)`  | `torch.hamming_window(size, periodic)` |
+| `signal::hann_window(size, periodic, options)`     | `torch.hann_window(size, periodic)`    |
+| `signal::irfft(spectrum_re, spectrum_im, dim)`     | `torch.fft.irfft(input, dim=dim)`      |
+| `signal::rfft(signal, dim)`                        | `torch.fft.rfft(signal, dim=dim)`      |
```
antimora: Two small formatting nits vs the adjacent Linalg table:
- Lines 443, 444, and 449 have trailing hard-break spaces (two spaces at EOL); none of the other rows in this table or the neighboring ones do.
- Line 445 is missing a space before the trailing `|`, so the right column edge doesn't line up with 446-449.

antimora (commented):
Nice addition - math checks out and the tests cover the code paths. The Copilot review's five inline comments all look addressed in `bbd3437`, so I'm not restating those.

A few things worth flagging before merge, most inline:

1. **Portability**: the `#[should_panic]` test for `size > i64::MAX` does not compile on 32-bit targets (wasm32 is supported), so it needs a `target_pointer_width = "64"` gate or removal.
2. **Consistency with `hamming_window.rs` / `hann_window.rs`** as the template: test naming (`test_*` vs `should_support_*` / `should_handle_*`), tolerance API (`_absolute` vs `_relative`), rustdoc style (plain `# Equation` vs `cfg_attr(doc, ...)` LaTeX), and `expect` wording.
3. **Rustdoc `DType::F64` example**: the compute happens in the backend's default `FloatElem` and `cast(dtype)` only widens storage at the end, so requesting `DType::F64` does not buy extra precision here. The identical output shown for F32 and F64 is accidentally correct. Either drop the F64 case from the example or note that dtype controls storage, not compute.

Nothing blocking, but items 1 and 2 are worth picking up.

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
softmaximalist: The `.set_half_precision_relative(1e-2)` used for sibling tests will still result in both size 8 test cases failing. Here is the printed message when the test cases fail with `.set_half_precision_relative(1e-2)`:

```
=> Position 0: 0.00024414063 != -0
   diff (rel = +1.00e0, abs = +2.44e-4), tol (rel = +5.00e-3, abs = +1.00e-5)
=> Position 7: 0.06689453 != 0.066467285
   diff (rel = +6.39e-3, abs = +4.27e-4), tol (rel = +5.00e-3, abs = +1.00e-5)
     
=> Position 0: 0.00024414063 != -0
   diff (rel = +1.00e0, abs = +2.44e-4), tol (rel = +1.00e-2, abs = +1.00e-5)
=> Position 7: 0.00024414063 != -0
   diff (rel = +1.00e0, abs = +2.44e-4), tol (rel = +1.00e-2, abs = +1.00e-5)
```

Since the failing entries/positions have values very close to 0, I loosened up the absolute tolerance.

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
softmaximalist: I followed the naming convention used throughout the repo but I can update it to match other window functions.

crates/burn-tensor/src/tensor/signal/blackman_window.rs
softmaximalist: Good point. Regarding the first element, the expected values are from pytorch. I have updated the near-zero values to 0 to avoid confusion. The returned tensor's first element is 0.0 in Burn so I have kept the values in the example.

burn-book/src/building-blocks/tensor.md
softmaximalist: Addressed in 7d4f2e8

crates/burn-tensor/src/tensor/signal/blackman_window.rs
softmaximalist: Fixed in d1c2514

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
softmaximalist: Addressed in 4a0507d

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
softmaximalist: Addressed in 4c71a7c

crates/burn-backend-tests/tests/tensor/float/ops/blackman_window.rs
softmaximalist: Addressed in 602c274

crates/burn-tensor/src/tensor/signal/blackman_window.rs
softmaximalist: Addressed in 8fd6bab

crates/burn-tensor/src/tensor/signal/blackman_window.rs
softmaximalist: Addressed in 7124f8f and 3b7431b

crates/burn-tensor/src/tensor/signal/blackman_window.rs
softmaximalist: Addressed in f577981

antimora (approved):
LGTM

Thanks for addressing my feedback

Issue #4852, opened 2026-04-20
burn-fusion: index-out-of-bounds cascade after upstream device panic


antimora:
## Summary

Once a cubecl worker thread panics (e.g. from an unimplemented op), all subsequent ops on the same runtime panic with:

```
thread 'Device-4-0' panicked at crates/burn-fusion/src/stream/execution/ordering.rs:65:38:
index out of bounds: the len is 0 but the index is 0
```

https://github.com/tracel-ai/burn/blob/main/crates/burn-fusion/src/stream/execution/ordering.rs#L65

The root panic leaves the stream state empty, then the next op's scheduling hits the unchecked index access on a zero-length vec.

## Repro

Trigger any cubecl-backed op that panics (e.g. #4850 cross on non-last dim, #4851 FFT shared memory overflow) and observe the cascade in subsequent benches:

```sh
cd crates/burn-backend-tests && cargo bench-metal --bench cross_unfold_ops -- --test
```

First bench panics at `burn-cubecl/src/kernel/cross.rs:63` (legitimately). Every bench after that in the same binary panics at `burn-fusion/src/stream/execution/ordering.rs:65` with the index-out-of-bounds even though those benches target unrelated ops (unfold, etc.).

## Impact

The cascade makes it hard to distinguish the real backend gap from downstream noise. In bench sweeps you see N failures where only the first is meaningful; the rest are collateral from a poisoned stream.

## Proposed fix

Either:
1. Guard the index access with a bounds check and return a clearer error (`StreamExhausted` or similar) instead of panicking.
2. Propagate the upstream panic as a stream-level error so subsequent ops short-circuit with the original cause instead of a confusing bounds panic.
3. Let `Backend::sync` surface the prior panic so the caller observes the real root cause.

## Context

Surfaced during the flex-bench migration to `burn-backend-tests` (#4810), while classifying backend gaps in #4846-#4851. Not a blocker for the migration itself - the bench summary shows the first panic clearly - but the duplicated noise makes triaging other failures harder.

laggui:
Duplicate. Closing in favor of https://github.com/tracel-ai/burn/issues/4827

Issue #4851, opened 2026-04-20
metal / wgpu: FFT kernel exceeds shared memory limit at n >= 4096


antimora:
## Summary

The cubecl-based FFT kernel requests more shared memory than Metal allows for FFT sizes >= 4096:

```
Launch(Too many resources were requested during launch
Too much shared memory requested.
Requested 131072 bytes, maximum 32768 bytes available.
```

Metal's shared memory per threadgroup is 32 KB; the kernel requests 128 KB at n=4096, 512 KB at n=16384. Pattern repeats at n=65536 (same 512 KB ceiling from tiling).

## Repro

```sh
cd crates/burn-backend-tests && cargo bench-metal --bench fft_ops -- --test
```

Affected benches: `rfft_1d::n_4096`, `n_16384`, `n_65536` and the corresponding `irfft_1d` sizes. Smaller sizes (n=256, 1024) work.

## Expected

Either:
1. Tile the FFT so shared memory stays within the platform limit at every size.
2. Fall back to a global-memory kernel when the shared-memory budget is exceeded.
3. Error cleanly at dispatch time instead of letting the launch fail inside the runtime.

## Context

Surfaced during the flex-bench migration to `burn-backend-tests` (#4810). May also affect wgpu (WebGPU's per-workgroup shared-memory limit is 16 KB by default), though I've only confirmed it on Metal so far.

antimora:
Retested on main with Metal: `rfft` works for n = 4096, 16384 and 65536. Closing.

Issue #4850, opened 2026-04-20
burn-cubecl: Cross product on non-last dimension not implemented


antimora:
## Summary

`burn-cubecl` panics with `not implemented: Cross product on non-last dimension not yet implemented for CubeCL backend`:

https://github.com/tracel-ai/burn/blob/main/crates/burn-cubecl/src/kernel/cross.rs#L63

Affects all cubecl-family backends: cuda, wgpu, metal, vulkan, rocm.

## Repro

```sh
cd crates/burn-backend-tests && cargo bench-metal --bench cross_unfold_ops -- --test
```

(Also reproducible on cuda/wgpu/rocm.)

The bench at `cross_unfold_ops.rs` uses `Tensor::cross(t, dim)` with various `dim` values, including non-last dims.

## Side effect

Once the cross kernel panics on the cubecl worker thread, subsequent ops against the same runtime hit a cascading panic in `burn-fusion/src/stream/execution/ordering.rs:65` (`index out of bounds: the len is 0 but the index is 0`) because the stream state is poisoned. Probably worth making fusion resilient to upstream panics regardless of whether cross itself is fixed, but it may be a separate issue.

## Context

Surfaced during the flex-bench migration to `burn-backend-tests` (#4810).

Issue #4849, opened 2026-04-20
tch: deform_conv2d not supported by Torch bindings


antimora:
## Summary

`burn-tch` panics with `not implemented: Torch bindings don't support deform_conv2d` when `module::deform_conv2d` is called against LibTorch:

https://github.com/tracel-ai/burn/blob/main/crates/burn-tch/src/ops/module.rs#L427

## Repro

```sh
cd crates/burn-backend-tests && cargo bench-tch --bench deform_conv_ops -- --test
```

All 8 deformable conv benches skipped.

## Context

`tch`/LibTorch exposes `torchvision::ops::deform_conv2d` via the `tch::vision` crate, but it's not wired up in burn-tch. Surfaced during the flex-bench migration to `burn-backend-tests` (#4810).

Options:
1. Wire the torchvision binding through.
2. Implement via the decomposed path (im2col + deformable offsets + gemm).
3. Document as an explicit gap and return a clearer error than `unimplemented!`.

PR #4853, opened 2026-04-21
Migrate benchmarks from burn-flex to burn-backend-tests

### LINKED ISSUE #4810: Migrate benchmarks from burn-flex to burn-backend-tests
As discussed, we need to migrate the benchmarks currently in burn-flex to a more appropriate location. Along with tests referenced in #4774, these benchmarks should be part of burn-backend-tests. 

**Decisions/Context:**
- "These benchmarks will have to be moved in time. Along with tests as you captured in #4774."
- Do not remove the benchmarks if they are still useful, but they must be moved out of burn-flex.

**Action Items:**
- Identify all relevant benchmarks in burn-flex.
- Move benchmarks to the new home (suggested: burn-backend-tests).
- Ensure all references and dependencies are updated.
- Communicate any deprecated or obsolete benchmarks that should not be migrated.

Linked reference: #4774

/cc @laggui

antimora:
Closes #4810. Sibling to #4812 (test migration).

## What

- Moves 22 divan benches from `crates/burn-flex/benches/` to `crates/burn-backend-tests/benches/`.
- Each bench now runs against the feature-selected `TestBackend` (`burn_dispatch::Dispatch`) - same pattern as tests. To compare backends, run twice with different feature sets and diff.
- Drops the `Flex (naive)` attention variant (flex-internal API) and the `realfft` comparison in fft_ops.
- Adds `cargo bench-<backend>` aliases mirroring the `test-<backend>` ones.

## Sync and panic handling

- `BencherExt::bench_synced` wraps divan's `bench()` with a `Backend::sync` call before returning, so async GPU backends time actual execution rather than dispatch.
- `common::try_setup` catches panics in op-heavy bench setup (e.g. `make_qtensor` calls `quantize_dynamic`, `fft_ops` irfft input uses `rfft`). Failed benches are recorded and replaced with a no-op; a summary prints at the end of each bench binary instead of aborting mid-run.

## Backend gaps surfaced

Filed #4846 (tch int_from_data), #4847 (tch quantization), #4848 (tch rfft), #4849 (tch deform_conv), #4850 (cubecl cross non-last-dim), #4851 (metal FFT shared memory), #4852 (burn-fusion cascade).

## Test plan

- [x] `cargo bench-flex --no-run` and `cargo bench-ndarray --no-run` compile
- [x] `cargo bench-flex --bench softmax_ops -- --test` runs the full tree with sync active
- [x] `cargo bench-metal` and `cargo bench-tch` surface failures via the skipped-bench summary without aborting
- [x] `cargo test -p burn-flex` still passes (realfft kept as dev-dep for in-tree correctness tests)


crates/burn-flex/Cargo.toml
```diff
@@ -57,111 +57,21 @@ num-traits = { workspace = true }
...
 
 # no_std targets without atomic CAS (enabled via `critical-section` feature)
-once_cell = { version = "1", default-features = false, optional = true }
+once_cell = { workspace = true, optional = true }
 portable-atomic = { workspace = true, optional = true }
 portable-atomic-util = { workspace = true, optional = true }
 
 [dev-dependencies]
 burn-ndarray = { path = "../burn-ndarray", version = "=0.21.0-pre.3" }
 burn-tensor = { path = "../burn-tensor", version = "=0.21.0-pre.3" }
```
laggui, line 73: Looks like these are still required for permuted/strided ops tests, the only ones remaining in burn-flex. They could stay in burn-flex, but we need to remove the `burn-tensor` usage and use the backend methods instead of the high-level API.

Though this will only be a blocker once we merge #4717.

laggui (approved):

PR #4866, opened 2026-04-22
Update/cubecl to client

Charles23R:
### Checklist

### Related Issues/PRs

Depends on https://github.com/tracel-ai/cubecl/pull/1304

### Changes

- Updates to cubecl's `to_client` api
- Updates to cubecl collective ops initialization
- Updates to fusion server/client

### Testing

Unit tests + text-cla + benchmarks
image


crates/burn-fusion/src/client.rs
```diff
@@ -367,12 +374,32 @@ where
...
+            .server
+            .utilities()
+            .downcast::<FusionUtilities>()
+            .expect("Can downcast to `FusionUtilities`");
+        let id = CommunicationId::from(device_ids);
+        if utilities.initialized_comms.read().unwrap().contains(&id) {
+            self.flush_queue();
+            let mut initialized_comms = utilities.initialized_comms.write().unwrap();
+            initialized_comms.insert(id);
+        }
```
nathanielsimard, line 407: Would it be possible to call ensure_collective_init using the inner backend?

Charles23R: We need to initialize the communication on the first call to a collective operation. Initializing is blocking for the server, so we need to make sure to flush right away so that other devices don't end up stuck on an initialization call from another device. 

This is already handled by cubecl, but since fusion adds another layer of streams and asynchronous submits, we also needed to add some logic here to flush the fusion server. 

Maybe there is a way to avoid this by design when handling collective calls? We can also chat about this offline as it can be quite complex/confusing.

PR #3608, opened 2025-08-24
Complex Backend

skewballfox:
It's a bit early but could definitely use feedback on what works and doesn't in terms of the design

The goals are:
1. keep the implementation for complex numbers out of the core backend trait, and make it a decorator similar to autodiff.
2. don't constrain backends to a specific layout, but make it possible to define layout dependent shared behavior (like butterfly operations)
3. Don't make the ComplexBackend trait be a supertrait of Backend, while allowing ComplexBackend to be usable with backend (done)
4. implement split tensors which can work with any backend, and use flex-tensor for the first backend to support interleaved complex numbers

### Checklist

- [ ] Confirmed that `cargo run-checks` command has been executed.
- [ ] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

- continuation of #3330 
- would resolve #3265 
- required for #788 

### Changes

main changes in relation to the goals are:
1. made a new `burn-complex` crate that will house a lot of the shared definitions. I'm guessing most of the stuff other than the `ComplexTensorBackend` trait and dtype for complex numbers will be moved here
2. Added a dummy trait `ComplexLayout` that will be implemented on unit structs that indicate what type of complex layout is in use for an implementation, which allows implementors to define functions and traits only meant to be used for a specific data layout.

### Testing

TODO

### Notes

My current plan is to get all ops implemented for split and burn-flex, write test for the ops, and then once it's almost ready to merge, stash the ndarray implementation and make that a separate PR. It's just way easier to implement for flex. Ndarray macros do not spark joy.

PR #4868, opened 2026-04-23
Split Associated Types from Backend into Backend Core

skewballfox:
So if you are trying to make a custom backend that can exist semi-independently from Backend, The reliance on backend for type mapping for traits like Numeric and BasicOps makes implementing backend required in order to unambiguously reference the tensor primitives for an existing backend. This separates out the types for Backend into it's own trait `BackendCore` which is part of the Backend Supertrait. 

### Checklist

### Related Issues/PRs

related to #3608 

### Changes

- split types from backend into BackendCore
- make ElementLimit not part of Element, but part can be referenced in ElementOrdered (change included by accident)
- split out `register_transaction` from basicOps and move it into `TransactionOp`
    - I am totally open to alternatives for this, I couldn't work out how to make TransactionOp not depend on backend, sort of reoganizing the to_data methods into DataOps. I'm kind of wondering if maybe TransactionOps can move it's implementation to the kind, the way BasicOps is implemented on Int, Float, Bool (where it's passed backend)


crates/burn-backend/src/backend/base.rs
```diff
@@ -16,6 +16,30 @@ use crate::distributed::{DistributedParamId, DistributedParams};
 
 use super::DeviceOps;
 
+/// The mapping of types used by Backend and traits like Numeric, BasicOps
+pub trait BackendCore {
```
laggui: Hmmm I would change the name at least. Perhaps `BackendTypes`?

crates/burn-backend/src/element/base.rs
```diff
@@ -18,7 +18,6 @@ pub trait Element:
     + ElementRandom
     + ElementConversion
     + ElementEq
-    + ElementLimits
```
laggui, line 21: Why remove it from the core element trait? The limits are still useful outside of ordered (e.g. to check for valid values to be cast).

crates/burn-backend/src/tensor/ops/base.rs
```diff
@@ -2,12 +2,27 @@ use alloc::vec::Vec;
...
+pub trait TransactionOp<B: Backend>: TensorKind<B> {
+    /// Read the data from the tensor using a transaction.
+    ///
+    /// # Remarks
+    ///
+    /// This is a low-level function used internally by the library to call different backend functions
+    /// with static dispatch. It is not designed for direct usage by users, and not recommended to import
+    /// or use this function directly.
+    fn register_transaction(tr: &mut TransactionPrimitive<B>, tensor: Self::Primitive);
+}
```
laggui, line 27: Splitting into a transaction ops trait makes sense. The primitive could be defined for `B: BackendTypes` instead of backend, as it's essentially a container for the tensor primitives part of a transaction.

crates/burn-backend/src/backend/ops/tensor.rs
```diff
@@ -4,13 +4,13 @@ use super::repeat_dim::repeat_with_slice_assign;
...
 use crate::tensor::{BoolTensor, Device, Float, FloatTensor, IntTensor};
-use crate::{Backend, Distribution, TensorData, get_device_settings};
+use crate::{Backend, BackendCore, Distribution, TensorData, get_device_settings};
 use crate::{ExecutionError, Scalar, TensorMetadata, TensorPrimitive};
 use alloc::vec::Vec;
 use burn_std::{BoolDType, FloatDType, IntDType, Shape, Slice};
 
 /// Operations on float tensors.
-pub trait FloatTensorOps<B: Backend> {
+pub trait FloatTensorOps<B: Backend + BackendCore> {
```
laggui: The `Backend` supertrait already includes the backend types trait, we probably don't need `Backend + BackendCore`.

crates/burn-autodiff/src/backend.rs
```diff
@@ -37,7 +37,9 @@ impl<B: Backend, C: CheckpointStrategy> Backend for Autodiff<B, C> {
     type BoolElem = B::BoolElem;
 
     type QuantizedTensorPrimitive = B::QuantizedTensorPrimitive;
+}
 
+impl<B: Backend, C: CheckpointStrategy> Backend for Autodiff<B, C> {
```
laggui, line 42: This is the separation I was expecting. For most backends it will just split the concerns, but for decorators that only need to use the associated types it should help. That's totally fine.

crates/burn-backend/src/element/base.rs
skewballfox: That accidentally made it in. complex elements aren't ordered, but it's also related to the code review from antimora

> MIN = (E::MIN, E::MIN) actually produces the value with the largest magnitude, not the smallest, and MAX is similar. Since complex comparisons (in burn-tensor/src/tensor/element/complex.rs) sort by magnitude, these constants are backwards as neutral elements for max/min reductions. Since reduction ops (complex_sum, cumsum, etc.) are wired up, a max/argmax path would silently return wrong results. Either drop ElementLimits for Complex<E> entirely (since there's no natural order on C) or pick conventions that match the sort order used elsewhere.

skewballfox: You can add it to the generic constraints directly on the functions that need them

laggui (approved):
LGTM!

I just cleaned up some of the usage to prefer the type aliases instead.

PR #4873, opened 2026-04-24
All reduce backward

Charles23R:
### Checklist

### Changes

Added the backward function to `Autodiff`'s `all_reduce` + some unit tests.

### Testing

Unit tests


crates/burn-autodiff/src/ops/distributed.rs
```diff
@@ -37,10 +41,42 @@ impl<B: DistributedBackend, C: CheckpointStrategy> DistributedBackend for Autodi
...
+            fn backward(
+                self,
+                ops: Ops<Self::State, 1>,
+                grads: &mut crate::grads::Gradients,
+                _checkpointer: &mut crate::checkpoint::base::Checkpointer,
+            ) {
+                unary::<B, _>(ops.parents, ops.node, grads, |grad| {
+                    B::all_reduce(grad, ops.state.0, ops.state.1).resolve()
+                });
+            }
```
laggui, line 62: Might want to add a comment just for clarity, in case others stumble upon this and assume that the backward scaling is wrong for `ReduceOperation::Mean`.

Something like this maybe?
```rust
// Backward uses the same reduce op: local gradients are synchronized via the backend, which handles scaling (e.g., ncclAvg for mean).
```

laggui (approved):

crates/burn-autodiff/src/ops/distributed.rs
Charles23R: Thanks! Added your comment as is + a warning if we ever support other reduce ops

PR #4878, opened 2026-04-25
Fix more cases for 0 sized tensors.

ArthurBrussee:
shape == 0, the subtraction underflows and panics under overflow checks (e.g. debug builds, wasm tests).

On native this is hidden because ChannelDeviceHandle swallows worker-thread panics with catch_unwind. on wasm ReentrantMutexDeviceHandle propagates the panic and additionally poisons the device handle (the service cell stays None after the panic skips its replace), turning every later test into "State ... is already borrowed by the current thread". Fix: an empty tensor has no elements that can overlap, so return true early when any dim is zero.

For slice assign, the kernel does no useful work when `value` has 0 elements anyway, so just early-return. After https://github.com/tracel-ai/cubecl/pull/1297 this wouldn't panic anymore anyway, but, might as well skip the work here entirely.

### DIFF
--- a/crates/burn-cubecl/src/kernel/index/slice_assign.rs
+++ b/crates/burn-cubecl/src/kernel/index/slice_assign.rs
@@ -100,9 +100,11 @@ pub(crate) fn slice_assign<R: CubeRuntime>(
     indices: &[burn_backend::Slice],
     value: CubeTensor<R>,
 ) -> CubeTensor<R> {
+    let num_elems = value.meta.num_elements();
+
     // Empty assignments have no work — launching a kernel here divides by
     // zero in `FastDivmod` registration when a shape dim is 0.
-    if value.meta.num_elements() == 0 {
+    if num_elems == 0 {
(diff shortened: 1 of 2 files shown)


laggui (commented):
Instead of being backend-specific logic in cubecl, we could add the early return check in the high-level API? The contract should be the same for all.

We should probably add more tests to cover 0 sized tensor operations.

/edit: going over pending PRs, I just realized this shouldn't trigger for slice_assign in any case as it currently stands. If the value tensor is empty, the slice range should be empty, and it's a no-op:

https://github.com/tracel-ai/burn/blob/bf2e8b7e41fdbe8002e406e4bf7defbe68589d97/crates/burn-tensor/src/tensor/api/base.rs#L1447-L1456

If the slice range is not empty but the value tensor is, it would panic with a tensor operation error due to the mismatch.

But other kernels like select and select_assign don't have this check for 0 sized tensors which should be handled.

Issue #4875, opened 2026-04-24
CNN training is much slower than PyTorch on the same MPS backend


JacobCote:
## Summary

I recently had performance issues with a vision model so I created a somewhat small minimal repro setup. I benchmarked a small CNN training loop on Apple MPS using equivalent Burn and PyTorch repros. Even when both use the same backend family (burn-tch / libtorch vs PyTorch on MPS) and similar model, loss, optimizer, and tensor shapes, the Burn version is significantly slower.

## Results

Average results after a 20 step warmup: 

- Burn: `514.61 ms/step`
- PyTorch: `165.40 ms/step`

## Environment

- Apple M2
- 16 GB RAM
- macOS 15.3.1
- PyTorch `2.9.0`
- Burn `0.21.0-pre.2`

## Burn repro

Create a new Cargo project with this `Cargo.toml`:

```toml
[package]
name = "cnn_min_repro"
version = "0.1.0"
edition = "2024"

[dependencies]
burn = { version = "0.21.0-pre.2" }

[features]
default = []
tch-gpu = ["burn/tch", "burn/train"]
```

Put this code in `src/main.rs` and run with:

```bash
cargo run --release --no-default-features --features tch-gpu
```

```rust
#![recursion_limit = "256"]

use std::time::Instant;

use burn::backend::Autodiff;
use burn::backend::libtorch::{LibTorch, LibTorchDevice};
use burn::module::Module;
use burn::nn::conv::{Conv2d, Conv2dConfig};
use burn::nn::loss::{MseLoss, Reduction};
use burn::nn::pool::{MaxPool2d as Pool2d, MaxPool2dConfig as Pool2dConfig};
use burn::nn::PaddingConfig2d;
use burn::optim::{AdamConfig, GradientsParams, Optimizer};
use burn::tensor::activation::{relu, sigmoid};
use burn::tensor::backend::Backend;
use burn::tensor::Tensor;

type MyBackend = Autodiff<LibTorch<f32>>;

const BATCH_SIZE: usize = 32;
const INPUT_SIZE: usize = 256;
const GRID_SIZE: usize = 16;
const LEARNING_RATE: f64 = 3e-4;
const WARMUP_STEPS: usize = 20;
const MEASURE_STEPS: usize = 40;

#[derive(Debug, Module)]
struct ConvNet<B: Backend> {
    conv1: Conv2d<B>,
    conv2: Conv2d<B>,
    conv3: Conv2d<B>,
[...]

PR #4880, opened 2026-04-26
Override conv2d backward in burn-tch via at::convolution_backward FFI…

koreaygj:
## Summary

overrides the three conv2d backward methods in `burn-tch` to call `at::convolution_backward` directly via a small C++ FFI shim — the same path PyTorch's autograd uses internally on MPS.

### Checklist

### Related Issues/PRs

#4875 

### Changes

`burn-tch` did not implement `conv2d_x_backward` / `conv2d_weight_backward` / `conv2d_bias_backward`, so it inherited the default in `burn-backend`, which expresses conv backward via forward primitives:
- `conv_transpose2d` for input gradient
- `swap_dims + conv2d + swap_dims` for weight gradient
- `swap_dims + reshape + sum_dim + reshape` for bias gradient

This bypasses libtorch's specialized MPS kernels (`mps_convolution_backward_input` / `_weight`) — exactly the kernels PyTorch's autograd dispatches to for `loss.backward()`.

### Testing

Reproduced and measured #4875 

| Run                          | ms/step (40 steps × 5 runs avg) | vs PyTorch |
|------------------------------|--------------------------------:|-----------:|
| Burn (default)               | 169.3                           | 2.23× slower |
| **Burn (this PR)**           | **81.1**                        | **1.07×** |
| PyTorch (async, reference)   | 75.9                            | 1.0× |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1656,6 +1656,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cnn_min_repro"
+version = "0.21.0-pre.3"
+dependencies = [
+ "burn",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "codespan-reporting"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


antimora:
Honestly I don't think we should add C code into burn repo; This should live upstream in tch crate

koreaygj:
Would a Rust-only override be acceptable? It improves the example to ~105 ms @antimora

antimora:
> Would a Rust-only override be acceptable? It improves the example to ~105 ms @antimora

I think this belongs in the upstream project https://github.com/LaurentMazare/tch-rs 

The project owner (@LaurentMazare) is responsive to PRs

koreaygj:
@antimora Okay then, I closed this PR

laggui:
@koreaygj thanks for investigating the linked issue :pray: 

But I agree with the other comments, we should not add more c++ code here. We could probably implement autodiff backend for libtorch instead to leverage autograd, but that's a bigger change.

PR #4881, opened 2026-04-26
Add ParamId::try_deserialize()

crutcher:
Deserialization should not panic on failure, or should have non-panic pathways.

### DIFF
--- a/crates/burn-core/src/module/param/id.rs
+++ b/crates/burn-core/src/module/param/id.rs
@@ -52,7 +52,7 @@ impl ParamId {
     /// # Panics
     /// On invalid id format
     pub fn deserialize(encoded: &str) -> ParamId {
-        Self::try_deserialize(&encoded).expect("Invalid id.")
+        Self::try_deserialize(encoded).expect("Invalid id.")
     }
 
     /// Deserialize a param id.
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/module/param/id.rs
```diff
@@ -48,28 +48,31 @@ impl ParamId {
...
-                    // Hash the 128-bit uuid to 64-bit
-                    // Though not *theoretically* unique, the probability of a collision should be extremely low
-                    let mut hasher = DefaultHashBuilder::default().build_hasher();
-                    // let mut hasher = DefaultHasher::new();
-                    hasher.write(id.as_bytes());
-                    hasher.finish()
-                }
-                Err(_) => panic!("Invalid id. {err}"),
+            Err(_) => match uuid::Uuid::try_parse(encoded) {
+                Ok(id) => Some(ParamId::deserialize(&id.to_string())),
```
laggui: This will indefinitely recurse 😬 curious if you ran the tests because `param_serde_deserialize_legacy_uuid` should hit this

We can have a `try_deserialize` function but we should keep the code you deleted to handle uuids.

laggui (changes requested):

laggui (approved):

PR #4889, opened 2026-04-27
Added explicit libtorch download switch and removed default download-libtorch behavior

LuoZheng2002:
### Checklist

### Changes

Previously burn-tch unconditionally enables "download-libtorch" feature on tch. Although this is only a fallback and only takes effect when no libtorch is available or specified using environment variables, it will cause inconvenience when debugging why cuda is not available when we try to manually specify a libtorch distribution.

I added a "download-libtorch" feature to burn, and it propagates to burn-tch and then to tch, so we can precisely control tch's fallback download behavior. In this way if we fail to manually specify a libtorch distribution but we think we did, the compilation will crash complaining neither "download-libtorch" is enabled nor we have properly specified a libtorch distribution, making it clear the problem is at compile time instead of dynamic linking time.

There is a behavior change: if the user does not specify LIBTORCH or LIBTORCH_USE_PYTORCH=1, previously it would use tch's fallback libtorch but now the compiling will crash because the user did not add "download-libtorch" feature to burn. I think this is acceptable because the user can quickly fix the problem by reading tch crate's error message.

I added a dedicated section to the Burn book describing how to properly set up the libtorch backend, and included a common gotcha regarding how to properly pass environment variables to cargo commands (I got stuck setting up libtorch with cuda for a long time due to the combo of mandatory download-libtorch fallback and my silly misunderstanding regarding linux environment variable passing).

### Testing
I ran these validations:
- cargo metadata --no-deps
  - Verified Cargo manifests and feature wiring are valid after edits.
- cargo check -p burn --features "tch"
  - Failed in my environment because LibTorch is not preinstalled (which is intended behavior for failing fast).
- cargo check -p burn --features "tch,download-libtorch"
  - Passed, confirming the new opt-in download path works.


LuoZheng2002:
The modification in auto test scripts is beyond my capability (make burn-tch to always be tested with "build-libtorch" feature on). Apart from the compile errors from torch-sys, there are also some inherent errors like dead code that is irrelevant with this pull request. Therefore, the pull request may need core contributors' assistance to pull off.

For addressing the test problems, an alternative way may be to do the opposite: use a "no-libtorch-build" feature to disable tch's "build-libtorch", so that burn-tch can compile with empty feature, but I think it's logically less clean. Ideally it's better for tch crate to pass the compile and report informative errors at runtime if anything related to libtorch fails, so that it does not interfere with downstream crates' syntax and format checking, and do not introduce blanket errors until the user actually uses relevant features.

laggui (commented):
Sorry for the late response, I really thought I had left a comment when I first saw this PR 😅 

While I agree that the automatic fallback might create a confusing failure mode in this case, I don't think removing it from the out-of-the-box setup (even when specifying a cuda version to download) is a good trade-off.

The fallback only happens if `LIBTORCH` env variable is not set and there is no system-wide installation (based on `tch-rs`'s build script). So I assume `LIBTORCH` wasn't properly set so it could be visible to the build script somehow.

I think a better approach would be to add some diagnostic message about which libtorch source was selected. This would be in the build script.

PR #4891, opened 2026-04-27
extract common float ops in a new trait

skewballfox:
### Checklist

### Related Issues/PRs
sort of related to #3608 

### Changes

Extracted all the trig related functions that were previously implemented directly on tensor, and only for Float. Mainly trying to reduce or preferably eliminate the number of ops I'd have to expose via some trait [here](https://github.com/tracel-ai/burn/pull/3608/changes#diff-2e294d684008edf99606b2a8fcb9e3751731d5fa736bf53e169638b5108daf07R1512), I could probably add sqrt ln, and a few others to the trait, but might want to rename it something like `FloatLike`?

### Testing


laggui (changes requested):
>  I could probably add sqrt ln, and a few others to the trait, but might want to rename it something like FloatLike?

Yeah I don't think we should group them by domain necessarily, having `Trigonometric` means we would also need to have `Log` and `Exp` kinds.

Perhaps we could have a `FloatMathOps`?

skewballfox:
> Perhaps we could have a FloatMathOps?

Done

laggui (approved):

PR #4894, opened 2026-04-28
Add A-FINE image quality metric

Capataina:
### Checklist

### Related Issues/PRs

Ticks one box on the meta-issue #4312. Original scoping comment + your answers are in that thread.

### Changes

A-FINE is a full-reference image quality metric that runs both the distorted and reference images through CLIP ViT-B/32, then through a naturalness head, a fidelity head, two logistic calibrators, and an adapter that fuses them into one (0, 100) score. Paper: https://arxiv.org/abs/2503.11221. PyIQA reference: https://github.com/chaofengc/IQA-PyTorch/blob/main/pyiqa/archs/afine_arch.py.

Lives at `crates/burn-train/src/metric/vision/afine/`. CLIP ViT is inlined under the metric, matching what LPIPS, DISTS, and FID did. ~1830 LOC across 8 Rust files plus the burn-book row.

A few things worth flagging:

- The CLIP backbone uses QuickGELU (`x * sigmoid(1.702 * x)`), not erf-based GELU. Substituting silently shifts every MLP activation, so QuickGELU lives as its own private module here, with the coefficient pinned at 1.702 as a labeled constant.
- The attention block is a custom fused-QKV one (`qkv_proj: Linear(d_model, 3 * d_model)` + `chunk(3, -1)` at forward) so the CLIP checkpoint maps to it one-to-one. burn-nn's MultiHeadAttention uses separate Q/K/V Linears, which would force a pre-split at load.
- The naturalness head's MLP uses the erf-based `burn_nn::Gelu`, NOT QuickGELU. PyIQA's reference is explicit on this distinction.
- `c1 = c2 = 1e-10` in the fidelity head's SSIM-like ratios. DISTS uses `1e-6` but A-FINE's CLIP feature magnitudes are unit-scale-ish, so the smaller eps is what the trained checkpoint expects.

Two non-obvious things in the pretrained loader the parity test caught:

The five 0-D scalar params (NR/FR `yita3`, `yita4`, adapter `k`) are stored with `shape=()` in the checkpoint but burn's `Param<Tensor<B, 1>>` expects `(1,)`. PytorchStore silently drops these, leaving them at random init. Loaded manually via `PytorchReader::with_top_level_key(...).get(name)`.

The CLIP checkpoint has both `positional_embedding` (text encoder, shape `[77, 512]`) and `visual.positional_embedding` (shape `[50, 768]`). After my remap renames the visual one to `positional_embedding`, both keys collide and HashMap iteration order picks which wins. Pre-rename of the text encoder's key first dodges the collision.

### Question on weight hosting

The URL points at the PyIQA author's personal HF mirror (`chaofengc/IQA-PyTorch-Weights/resolve/main/afine.pth`). I tried pretty hard to find an org-hosted alternative and couldn't:

- Not on any HF org I checked (PyIQA-related, q-future, the authors' institutions)
- Not in IQA-PyTorch's GitHub releases (50+ other metric weights are there, just not `afine.pth`)
- Not on Zenodo, figshare, ModelScope, OSF, or Kaggle
- Authors' canonical release is Google Drive only, which reqwest can't follow cleanly

I noticed LPIPS, DISTS, and FID currently pull weights from personal-account GitHub raw URLs, so personal hosting isn't unprecedented in burn. But I remember you mentioned org hosting was preferred when we scoped this on the issue, so figured I'd flag rather than assume. Happy to go with whatever you prefer: keep this URL, have tracel-ai mirror `afine.pth`, or another option.

### Testing

Always-on property tests: forward shape, batch processing, finite output on constant inputs, asymmetry from the adapter's exponent term, `image_size` validation, display format.

Two `#[ignore = "downloads pre-trained weights"]` tests, matching the LPIPS/DISTS/FID convention:

- `test_afine_pretrained` loads real weights and checks the score is finite and meaningfully differs from a random-init metric on the same input. Catches silent partial loads where a regex remap is wrong.
- `test_afine_pretrained_parity` matches a captured PyIQA reference scalar on a deterministic `arange/(N-1)` vs `arange/N` input pair to within ~5e-5 absolute. Catches QuickGELU-vs-GELU swaps, fused-QKV transpose-direction bugs, channel-order mistakes, and `c1=c2=1e-10` epsilon drift that the property tests miss.

`cargo run-checks` green: fmt, clippy, audit, typos, doc, full burn-train with `--features vision` (123 passing, 7 ignored).

### DIFF
--- a/crates/burn-train/src/metric/vision/afine/clip_vit.rs
+++ b/crates/burn-train/src/metric/vision/afine/clip_vit.rs
@@ -104,6 +104,22 @@ impl ClipVisualEncoderConfig {
     }
 }
 
+/// Output of [`ClipVisualEncoder::forward_with_features`].
+///
+/// `features` are the per-block patch tokens A-FINE consumes. `cls` is
+/// the post-`ln_post` class-token embedding, present only when the
+/// caller requests it; the metric path skips the slice + LayerNorm to
+/// avoid wasted work.
+#[derive(Debug)]
+pub struct ClipOutput<B: Backend> {
+    /// Per-block patch tokens, length `num_layers`, each
+    /// `[batch, num_patches, embed_dim]`.
+    pub features: Vec<Tensor<B, 3>>,
+    /// Post-`ln_post` class-token embedding, `[batch, embed_dim]`. `Some`
+    /// when `forward_with_features` is called with `return_cls = true`.
+    pub cls: Option<Tensor<B, 2>>,
+}
+
 /// CLIP ViT-B/32 image encoder.
 ///
 /// Consumes a normalized RGB image and returns either the class-token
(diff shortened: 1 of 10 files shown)


laggui:
Just letting you know I am aware of this PR :pray: but haven't gotten around to it yet; bigger PRs require a bit more context switching :sweat_smile: 

It's still on my stack, will review sometime this week!

Thanks for the contribution

crates/burn-train/src/metric/vision/afine/clip_vit.rs
```diff
@@ -0,0 +1,344 @@
...
+            let patch_features = x.clone().slice([0..batch, 1..seq_len, 0..embed]);
+            features.push(patch_features);
+        }
+
+        // ln_post is applied only to the class-token output.
+        let cls = x.slice([0..batch, 0..1, 0..embed]).reshape([batch, embed]);
+        let cls = self.ln_post.forward(cls);
+
+        (cls, features)
+    }
```
laggui: For the metric implementation the cls output is discarded, so we could skip this computation.

We can preserve it for completeness (and you already have a test that covers this), so `forward_with_features` could return an output struct with optional cls output e.g.

```rust
pub struct ClipOutput<B: Backend> {
    pub features: Vec<Tensor<B, 3>>,
    pub cls: Option<Tensor<B, 2>>,
}
```

laggui (changes requested):
Pretty clean implementation! Just one minor comment but otherwise LGTM

Capataina:
hey, I made the changes you requested, took some time so sorry for that; does this look right now?

laggui (approved):
LGTM!

Your response time was totally fine btw 😄

Capataina:
Exciting! Looking forward for further contributions :)

PR #4899, opened 2026-04-29
Fix cubecl cross product on non-last dimension

### LINKED ISSUE #4850: burn-cubecl: Cross product on non-last dimension not implemented

dschulmeist:
### Related Issues/PRs

Closes #4850.

### Changes

The cubecl cross kernel reads each 3-vector from contiguous memory along the innermost axis, so it could only run when the cross dim was the last one. For other dims it bailed out with `unimplemented!()`, panicking on every cubecl-family backend (cuda, wgpu, metal, vulkan, rocm) and poisoning the fusion stream.

Fix in `crates/burn-cubecl/src/kernel/cross.rs`: permute the cross dim to the last position, contiguize the operands (`swap_dims` is metadata-only), run the existing kernel, and permute the result back. Recursion terminates after one swap because the inner call always hits `dim == ndims - 1`.

```rust
if dim != ndims - 1 {
    let last = ndims - 1;
    let lhs = into_contiguous(swap_dims(lhs, dim, last));
    let rhs = into_contiguous(swap_dims(rhs, dim, last));
    let result = cross(lhs, rhs, last);
    return swap_dims(result, dim, last);
}
```

The two `into_contiguous` copies are the price of reusing the existing kernel for non-last-dim cases. A bespoke strided kernel would avoid them but is a much bigger change; happy to follow up if the perf cost matters.

The cascading `burn-fusion` panic mentioned in the issue goes away as a side effect once `cross` stops panicking. Making fusion resilient to upstream panics in general is left as the follow-up the issue itself flags.

### Testing

`cargo test-metal --test tensor cross` — 10/10 pass, including:
- Existing `test_cross_3d_dim0` (drops the `might_panic` guard).
- New `test_cross_4d_middle_dim` — cross on a middle dim of a rank-3 tensor.
- New `test_cross_non_last_dim_broadcast` — `[3,1]` × `[3,4]` cross dim=0, validated against the permuted last-dim equivalent.
- New `test_cross_non_last_dim_matches_permuted_last_dim`.

`cargo test-metal --test autodiff cross` — 18/18 pass:
- Existing `different_dim` (drops the `might_panic` guard).
- New `backward_non_last_dim` — gradients flow through cross on dim 0 (the autodiff rule recurses into `float_cross` with the same dim, so this also exercises the non-last-dim forward path on the gradient pass).

`cargo test-metal-nofuse --test cubecl cross` — 7/7 pass.

Also re-ran the same tests on `wgpu` (10/10) and `cpu` (10/10) to make sure nothing else regressed. The kernel fix lives in shared cubecl code, so cuda/vulkan/rocm get the same fix without separate changes.

### Checklist

- [x] Made sure the book is up to date with changes in this PR. (no book changes needed.)

### DIFF
--- a/crates/burn-backend-tests/tests/autodiff/cross.rs
+++ b/crates/burn-backend-tests/tests/autodiff/cross.rs
@@ -1,9 +1,6 @@
 use super::*;
 use burn_tensor::{TensorData, Tolerance};
 
-#[cfg(feature = "std")]
-use burn_backend_tests::might_panic;
-
 #[test]
 fn backward_basic() {
     let device = AutodiffDevice::new();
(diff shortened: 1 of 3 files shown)


laggui (approved):
LGTM!

Thanks for adding extra tests on non last dims as well

Issue #4784, opened 2026-04-12
Add complex-to-complex FFT (fft / ifft) to signal module


antimora:
## Summary

Add `fft` and `ifft` (complex-to-complex) to `burn::tensor::signal`, alongside the existing `rfft` / `irfft` (real-to-complex).

## Motivation

`burn::tensor::signal` currently provides only real FFT variants:
- `rfft`: real input → complex output (onesided)
- `irfft`: complex input (onesided) → real output

This is enough for most audio use cases, but the ONNX [`DFT`](https://onnx.ai/onnx/operators/onnx__DFT.html) operator requires two more behaviors that the real variants can't express:

1. **Complex-to-complex forward DFT** — when the ONNX input has trailing dim = 2 (interleaved real/imag), the operator is defined as a full complex FFT. Not a superset of `rfft`; the input is already complex.
2. **Inverse DFT** — ONNX's `inverse=1` is the complex-to-complex inverse FFT. It is not the same as `irfft`: `irfft` assumes onesided complex input with conjugate symmetry and returns a real signal, whereas ONNX IDFT takes full complex input and returns full complex output.

As a result, burn-onnx's DFT implementation explicitly rejects both cases (see the error messages in [`dft.rs`](https://github.com/tracel-ai/burn-onnx/blob/main/crates/onnx-ir/src/node/dft.rs) — "Burn's irfft is the inverse of rfft (onesided real FFT), which differs from ONNX's complex-to-complex inverse DFT" and "complex-to-complex DFT is not supported by Burn's current signal API"). Any ONNX model that round-trips through the frequency domain or that passes complex tensors through DFT is currently unconvertible.

This is blocking full DFT support in burn-onnx (see tracel-ai/burn-onnx#161).

## Proposed API

Mirror the existing `rfft` signature but accept and return complex-encoded tensors (trailing dim = 2 for real/imag):

```rust
/// Forward complex FFT along `axis`.
///
/// Input:  [..., N, ..., 2] — real/imag pair in the trailing dim.
/// Output: [..., N, ..., 2] — same shape; frequency bins in `axis`.
pub fn fft<B: Backend, const D: usize>(
    input: Tensor<B, D>,
    axis: usize,
) -> Tensor<B, D>;

/// Inverse complex FFT along `axis`.
///
/// Input:  [..., N, ..., 2]
/// Output: [..., N, ..., 2]
pub fn ifft<B: Backend, const D: usize>(
    input: Tensor<B, D>,
[...]

PR #4631, opened 2026-03-15
Add HannWindow operator to burn-tensor

walkinggo:
### Checklist

- [ ] Confirmed that `cargo run-checks` command has been executed. *(Note: Ran targeted checks instead due to local CUDA SDK limitations, see Testing section below for details).*
- [x] Made sure the book is up to date with changes in this PR. *(Note: N/A, this is an API addition and the Rustdoc is fully documented).*

### Related Issues/PRs

Addresses part of #4531 (Specifically the `HannWindow` signal processing operator).

### Changes

Added the `HannWindow` operator to the Tensor API to support signal processing models and align with ONNX specifications. 

**Detailed implementation:**
1. **`crates/burn-tensor/src/tensor/api/float.rs`**:
   - Added `Tensor<B, 1>::hann_window(size, periodic, options)` as an associated function under a new `impl<B> Tensor<B, 1>` block.
   - Implemented the Hann window formula: $w_n = 0.5 - 0.5 \times \cos(2\pi n / N)$ using existing tensor operations (`arange`, `float`, `cos`, `mul_scalar`, `add_scalar`, `cast`).
   - Supported both `periodic` ($N = size$) and `symmetric` ($N = size - 1$) modes.
   - Handled edge cases perfectly: `size = 0` returns an empty tensor, and `size = 1` returns `[1.0]` (preventing division-by-zero panics and matching PyTorch/NumPy behavior).
   - Used `f64` precision for the angular increment calculation to ensure high accuracy before casting to the target dtype.
   - Added dual-state documentation (`cfg_attr`) for clean KaTeX math rendering in Rustdoc and readable plain-text fallbacks in IDE hover hints.

2. **`crates/burn-backend-tests/tests/tensor/float/ops/hann_window.rs` & `mod.rs`**:
   - Registered the new test module.
   - Added 6 comprehensive test cases covering `periodic` mode, `symmetric` mode, dtype options, empty tensors, and the `size=1` boundary for both modes.
   - Used `assert_approx_eq` with `Tolerance` for cross-backend floating-point stability.

### Testing

Due to my local CUDA SDK being an older version, the `cudarc` binding in `cubecl-cuda` failed to find `CUtensorMapL2promotion` and `CUtensorMapFloatOOBfill`. Therefore, I had to skip the global `cargo run-checks` command. 

Instead, I manually and successfully verified the code using targeted commands for the active backends:

| Check | Command | Result |
| :--- | :--- | :--- |
[...]

Issue #4477, opened 2026-02-09
Expand IndexingUpdateOp to support Assign, Mul, Max, Min for scatter/select_assign


antimora:
## Feature description

`IndexingUpdateOp` currently only supports `Add` (`Assign` and `Mul` are commented out in the source):

```rust
// burn-backend/src/tensor/ops/mod.rs
pub enum IndexingUpdateOp {
    // Assign,
    Add,
    // Mul
}
```

Adding `Assign`, `Mul`, `Max`, and `Min` variants would enable `scatter()` and `select_assign()` to cover all standard reduction modes.

## Feature motivation

The ONNX `ScatterElements` operator supports five reduction modes: `none` (assign), `add`, `mul`, `max`, `min`. In `burn-onnx`, we can only use `Tensor::scatter()` for the `add` case. For `none`, we work around it with a gather-diff-scatter trick (`data.scatter(dim, indices, updates - data.gather(dim, indices), Add)`). For `mul`/`max`/`min`, we fall back to an element-by-element loop that flattens tensors, decomposes indices, and does per-element `slice_assign` -- ~30 lines of generated code per model instead of a single `scatter()` call.

With all five variants available:

| Reduction | Current codegen | With full IndexingUpdateOp |
|-----------|----------------|---------------------------|
| none | gather-diff-scatter workaround | `data.scatter(dim, idx, updates, Assign)` |
| add | `data.scatter(dim, idx, updates, Add)` | same |
| mul | 30-line element loop | `data.scatter(dim, idx, updates, Mul)` |
| max | 30-line element loop | `data.scatter(dim, idx, updates, Max)` |
| min | 30-line element loop | `data.scatter(dim, idx, updates, Min)` |

This would also benefit `ScatterND` which has the same reduction modes and the same element-by-element fallback.

## Suggest a Solution

Uncomment and extend `IndexingUpdateOp`:

```rust
pub enum IndexingUpdateOp {
    Assign,
    Add,
    Mul,
    Max,
    Min,
}
```

Then implement the corresponding logic in each backend's `scatter` and `select_assign` operations.

## Context

- `burn-onnx` PR implementing ScatterElements: https://github.com/tracel-ai/burn-onnx/issues/125
- ONNX ScatterElements spec: https://onnx.ai/onnx/operators/onnx__ScatterElements.html
- ONNX ScatterND spec: https://onnx.ai/onnx/operators/onnx__ScatterND.html

laggui:
[...]

PR #4903, opened 2026-04-30
Adding complex to complex FFT implementation

RunjiaChen:
# Add complex-to-complex FFT (`cfft`) to the signal module

## Checklist

- [x] Confirmed that `cargo run-checks` command has been executed. *(Note: Ran targeted checks instead due to local CUDA SDK limitations, see Testing section below for details).*
- [ ] Made sure the book is up to date with changes in this PR. *(Note: N/A — this is an API addition and the Rustdoc is fully documented).*

## Related Issues/PRs

Extends the FFT signal processing API added alongside `rfft` / `irfft`. This PR adds the missing complex-to-complex direction (Related to #4784).

## Changes

Added a **complex-to-complex 1D FFT** (`cfft`) to the Tensor signal API, built entirely from existing frontend operations with zero backend modifications.

#### `crates/burn-tensor/src/tensor/signal/fft.rs`

**`cfft(signal_re, signal_im, dim, n)` — public API**

- Accepts two tensors (real and imaginary parts of a complex signal) and returns two tensors (real and imaginary parts of the full N-bin complex spectrum).
- Validates that `signal_re` and `signal_im` have the same shape.
- Delegates all power-of-two and dimension checks to the existing `rfft()` function.
- Computes the full complex FFT by leveraging the linearity of the DFT:
  ```
  FFT(x + iy) = FFT(x) + i · FFT(y)
              = (Xr - Yi) + i(Xi + Yr)
  ```
- Calls `rfft()` twice (once for the real part, once for the imaginary part), extends each half-spectrum to full length via `hermitian_extend`, then combines with simple arithmetic.
- Includes full Rustdoc with a code example.

**`hermitian_extend(half_re, half_im, dim, full_len)` — private helper**

- Extends the `N/2 + 1` bin half-spectrum returned by `rfft` to the full `N`-bin spectrum using Hermitian symmetry: `X[k] = conj(X[N-k])` for `k > N/2`.
- Implemented purely with existing tensor operations: `narrow`, `flip`, `neg`, `Tensor::cat`.
- Handles edge cases: for `N <= 2`, the half-spectrum already covers all bins and no extension is needed.

We need the extension function because hermitian symmetry no longer applies to complex valued signals. Thus, we return all N bins. 
---

#### `crates/burn-backend-tests/tests/cubecl/fft.rs`

Added 10 comprehensive test cases:

| Test | What it verifies |
|------|-----------------|
| `cfft_output_has_n_bins` | Output shape is N, not N/2+1 |
| `cfft_pure_real_input` | Matches known DFT of [1,2,3,4] → [10, -2+2i, -2, -2-2i] |
| `cfft_pure_imaginary_input` | Verifies FFT(i·x) = i·FFT(x) identity |
| `cfft_complex_exponential` | Single-frequency exp(i2πn/4) produces a delta at bin 1 |
| `cfft_zeros` | Zero input → zero output |
| `cfft_dim1_2d_tensor` | Works along dim=1 on a batched 2D tensor |
| `cfft_with_n_padding` | Zero-padding via `n: Some(4)` on a length-2 signal |
| `cfft_length_1` | Edge case: N=1 (single sample = identity) |
| `cfft_length_2` | Edge case: N=2 (X[0]=a+b, X[1]=a-b) |
| `cfft_rejects_mismatched_shapes` | Panics when signal_re / signal_im shapes differ |

All expected values are hand-computed from the DFT definition and cross-verified.

## Testing

I do not have CUDA on my device (Mac), so I followed the same methods of testing as #4631.

| Check | Command | Result |
|-------|---------|--------|
| Formatting | `cargo fmt -p burn-tensor -p burn-backend-tests -- --check` | Passed ✅ |
| Linting | `cargo clippy -p burn-tensor -p burn-backend-tests --features cube -- -D warnings` | Passed ✅ |
| Testing | `cargo test -p burn-backend-tests --features flex,cube --test cubecl -- cfft` | 10/10 Passed ✅ |

### DIFF
--- a/crates/burn-backend-tests/tests/cubecl/fft.rs
+++ b/crates/burn-backend-tests/tests/cubecl/fft.rs
@@ -661,4 +661,3 @@ fn cfft_rejects_mismatched_shapes() {
     let im = TestTensor::<1>::from([1.0, 2.0]);
     let _ = cfft(re, im, 0, None);
 }
-
--- a/crates/burn-backend-tests/tests/cubecl/fft.rs
+++ b/crates/burn-backend-tests/tests/cubecl/fft.rs
@@ -667,36 +667,16 @@ fn cfft_dim0_2d_tensor() {
     // Apply cfft along dim=0 on a 2D tensor (4 rows, 2 columns)
     // Column 0: complex exponential exp(i·2π·n/4) → DFT = [0, 4, 0, 0]
     // Column 1: pure real [1, 2, 3, 4] → DFT = [10, -2+2i, -2, -2-2i]
-    let re = TestTensor::<2>::from([
-        [1.0, 1.0],
-        [0.0, 2.0],
-        [-1.0, 3.0],
-        [0.0, 4.0],
-    ]);
-    let im = TestTensor::<2>::from([
-        [0.0, 0.0],
-        [1.0, 0.0],
-        [0.0, 0.0],
-        [-1.0, 0.0],
-    ]);
+    let re = TestTensor::<2>::from([[1.0, 1.0], [0.0, 2.0], [-1.0, 3.0], [0.0, 4.0]]);
+    let im = TestTensor::<2>::from([[0.0, 0.0], [1.0, 0.0], [0.0, 0.0], [-1.0, 0.0]]);
 
     let (cfft_re, cfft_im) = cfft(re, im, 0, None);
 
     assert_eq!(cfft_re.dims(), [4, 2]);
     assert_eq!(cfft_im.dims(), [4, 2]);
 
-    let expected_re = TensorData::from([
-        [0.0, 10.0],
-        [4.0, -2.0],
-        [0.0, -2.0],
-        [0.0, -2.0],
-    ]);
-    let expected_im = TensorData::from([
-        [0.0, 0.0],
-        [0.0, 2.0],
-        [0.0, 0.0],
-        [0.0, -2.0],
-    ]);
+    let expected_re = TensorData::from([[0.0, 10.0], [4.0, -2.0], [0.0, -2.0], [0.0, -2.0]]);
+    let expected_im = TensorData::from([[0.0, 0.0], [0.0, 2.0], [0.0, 0.0], [0.0, -2.0]]);
 
     cfft_re
         .into_data()
@@ -707,7 +687,7 @@ fn cfft_dim0_2d_tensor() {
 }
 
 #[test]
-fn cfft_with_n_smaller_than_signal() {
+fn cfft_with_n_truncation() {
     // Signal length 8, truncated to n=4 → DFT of [1+0i, 2+0i, 3+0i, 4+0i]
     // Trailing values are discarded, not included in the transform.
     let re = TestTensor::<1>::from([1.0, 2.0, 3.0, 4.0, 99.0, 99.0, 99.0, 99.0]);
--- a/crates/burn-backend-tests/tests/cubecl/fft.rs
+++ b/crates/burn-backend-tests/tests/cubecl/fft.rs
@@ -570,8 +570,8 @@ fn cfft_zeros() {
 #[test]
 fn cfft_dim1_2d_tensor() {
     // Apply cfft along dim=1 on a 2D tensor
-    // Row 0: pure real [1, 0, -1, 0] → DFT = [0, 2, 0+0i, -2+0i] extended
-    // Row 1: complex exponential at bin 1
+    // Row 0: pure real [1, 2, 3, 4] → DFT = [10, -2+2i, -2, -2-2i]
+    // Row 1: complex exponential exp(i·2π·n/4) → DFT = [0, 4, 0, 0]
     let re = TestTensor::<2>::from([[1.0, 2.0, 3.0, 4.0], [1.0, 0.0, -1.0, 0.0]]);
     let im = TestTensor::<2>::from([[0.0, 0.0, 0.0, 0.0], [0.0, 1.0, 0.0, -1.0]]);
 
--- a/crates/burn-backend-tests/tests/cubecl/fft.rs
+++ b/crates/burn-backend-tests/tests/cubecl/fft.rs
@@ -661,3 +661,70 @@ fn cfft_rejects_mismatched_shapes() {
     let im = TestTensor::<1>::from([1.0, 2.0]);
     let _ = cfft(re, im, 0, None);
 }
+
+#[test]
+fn cfft_dim0_2d_tensor() {
+    // Apply cfft along dim=0 on a 2D tensor (4 rows, 2 columns)
+    // Column 0: complex exponential exp(i·2π·n/4) → DFT = [0, 4, 0, 0]
+    // Column 1: pure real [1, 2, 3, 4] → DFT = [10, -2+2i, -2, -2-2i]
+    let re = TestTensor::<2>::from([
+        [1.0, 1.0],
+        [0.0, 2.0],
+        [-1.0, 3.0],
+        [0.0, 4.0],
+    ]);
+    let im = TestTensor::<2>::from([
+        [0.0, 0.0],
+        [1.0, 0.0],
+        [0.0, 0.0],
+        [-1.0, 0.0],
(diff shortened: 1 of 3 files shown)


RunjiaChen (commented):
I have read through the changes, and approved them

Sublime12 (approved):
lgtm
Just need to fix typos errors

crates/burn-tensor/src/tensor/signal/fft.rs
```diff
@@ -149,3 +149,105 @@ pub fn irfft<B: Backend, const D: usize>(
...
+
+    // FFT(z) = FFT(x) + i·FFT(y)
+    //        = (Xr + i·Xi) + i·(Yr + i·Yi)
+    //        = (Xr - Yi) + i·(Xi + Yr)
+    (xr - yi, xi + yr)
+}
+
+/// Extend a half-spectrum from [`rfft`] (`N/2 + 1` bins) to the full `N`-bin
+/// spectrum using Hermitian symmetry: `X[k] = conj(X[N-k])` for `k > N/2`.
+fn hermitian_extend<B: Backend, const D: usize>(
```
antimora: This is the same algorithm as `reconstruct_full_spectrum` in `stft.rs` (~line 170), just generalized over `D` and `dim`:

```rust
// stft.rs (existing, 2D-only, dim=1 hardcoded)
let re_neg = re.clone().narrow(1, 1, n_neg).flip([1]);
let im_neg = im.clone().narrow(1, 1, n_neg).flip([1]).neg();
```

Since they're in the same module and the new version is strictly more useful, can we have `stft` delegate to this one (make `hermitian_extend` `pub(super)`)? Otherwise we'll have two copies of the same logic that drift over time.

antimora (commented):
Math and algorithm look right, tests cover the meaningful cases. Two things:

- See the inline comment - there's already a Hermitian-extend helper in `stft.rs` that this duplicates.
- Test gaps: no `cfft` along `dim=0` of a 2D tensor (rfft has one), and no truncation case (`n` smaller than the signal). Padding is covered, truncation isn't.

Two-rfft is fine performance-wise. If this ever gets hot we can switch to the single-rfft trick (pack `x + iy` into one complex signal of length N and unpack via even/odd symmetry), but no need to block on that.

antimora (changes requested):

crates/burn-tensor/src/tensor/signal/fft.rs
RunjiaChen: Oops, I didnt actually read through the stft.rs file. Do I apply the changes, or wait for the feedback from the other reviewer, should he have anything additional to add?

crates/burn-tensor/src/tensor/signal/fft.rs
```diff
@@ -149,3 +149,105 @@ pub fn irfft<B: Backend, const D: usize>(
     );
     Tensor::new(TensorPrimitive::Float(signal))
 }
+
+/// Computes the 1-dimensional discrete Fourier Transform of complex-valued input.
+///
+/// Internally calls [`rfft`] on the real and imaginary parts separately,
+/// extends each half-spectrum to the full `N`-bin spectrum via Hermitian
+/// symmetry, then combines: `FFT(x + iy) = FFT(x) + i·FFT(y)`.
```
laggui: To fix typo flagging `iy` -> `it` (incorrect) you should rewrite it as `i·y`, it also follows your `i·FFT(y)`.

Also: we have KaTeX available in burn-tensor, you could write the equation with that. We use `cfg_attr` so the doc is also available locally, e.g.

```rust
#[cfg_attr(doc, doc = "$$\\text{ReLU}\\(x\\) = \\(x\\)^+ = \\max\\(0, x\\)$$")]
#[cfg_attr(not(doc), doc = "`ReLU(x) = max(0, x)`")]
```

crates/burn-tensor/src/tensor/signal/fft.rs
RunjiaChen: Following you advice, I deleted the function `reconstruct_full_spectrum` and replaced it with `hermitian_extend` instead.

RunjiaChen:
Hello @antimora , I have made the changes according to your requests. May I know if there is anything else I need to do? Thank you!

laggui (approved):
Looks like all comments have been addressed.

Will wait before merging in case there is pending feedback from antimora, but should be good otherwise.

crates/burn-backend-tests/tests/cubecl/fft.rs
```diff
@@ -468,3 +468,243 @@ fn rfft_2d_with_n_padded() {
...
+    cfft_im
+        .into_data()
+        .assert_approx_eq::<FloatElem>(&expected, Tolerance::absolute(1e-4));
+}
+
+#[test]
+fn cfft_dim1_2d_tensor() {
+    // Apply cfft along dim=1 on a 2D tensor
+    // Row 0: pure real [1, 0, -1, 0] → DFT = [0, 2, 0+0i, -2+0i] extended
+    // Row 1: complex exponential at bin 1
```
antimora: This one got missed from Copilot's earlier flag - the lead-in comment still has Row 0 and Row 1 swapped relative to the actual data (Row 0 of `re` is `[1, 2, 3, 4]`, Row 1 is `[1, 0, -1, 0]`). The expected outputs and the DFT comments below are correct, just this lead-in is stale.

```suggestion
    // Row 0: pure real [1, 2, 3, 4] → DFT = [10, -2+2i, -2, -2-2i]
    // Row 1: complex exponential exp(i·2π·n/4) → DFT = [0, 4, 0, 0]
```

crates/burn-tensor/src/tensor/signal/fft.rs
```diff
@@ -149,3 +150,118 @@ pub fn irfft<B: Backend, const D: usize>(
...
+) -> (Tensor<B, D>, Tensor<B, D>) {
+    let half_len = half_re.dims()[dim]; // N/2 + 1
+
+    // For N <= 2, the half-spectrum already covers all bins
+    if full_len <= half_len {
+        return (half_re, half_im);
+    }
+
+    // Mirror bins: reverse of bins 1..N/2, with conjugated imaginary part
+    // This produces X[N/2+1], X[N/2+2], ..., X[N-1]
```
antimora: Picking up Copilot's earlier nit - `1..N/2` is ambiguous (Rust exclusive range vs. math closed interval). Since `mirror_len = N/2 - 1`, being explicit is clearer:

```suggestion
    // Mirror bins: reverse of bins 1..N/2-1 (skipping the Nyquist bin),
    // with conjugated imaginary part. This produces X[N/2+1], X[N/2+2], ..., X[N-1]
```

antimora (approved):
Looks good - the dedup with `stft.rs` is exactly what I had in mind, and the new `cfft_dim0_2d_tensor` / `cfft_with_n_smaller_than_signal` tests cover the gaps. Approving.

Two non-blocking comment nits inline - feel free to fix in a follow-up.

RunjiaChen:
Thank you! I have applied the changes. I think my contributions has mostly existed in the frontend space. I haven't had the opportunity to do the backend (which seems to be where most of the heavy lifting takes place). I don't really have any background parallel computing (such as CUDA or Metal), but I want to pick up some basic principles through contributing to this library. May I ask where is a good place (or issue) to start? Thank you!

laggui:
In this case, our backends rely on [CubeCL](https://github.com/tracel-ai/cubecl). The kernels either live in [`burn-cubecl](https://github.com/tracel-ai/burn/tree/main/crates/burn-cubecl/src)` or [CubeK](https://github.com/tracel-ai/cubek).

There is some documentation, but it is a bit limited right now :sweat_smile: feel free to hop on the discord server and ask questions in the cubecl channels if you go down that route.

Perhaps a good first issue could be https://github.com/tracel-ai/burn/issues/4477. The kernels already exist, but we only support scatter add. You could check the recently added scatter_nd implementation for reference (https://github.com/tracel-ai/burn/pull/4709). Might be a good starting point to extend an existing kernel than to try to implement something from scratch.

PR #4905, opened 2026-04-30
Add Clone + 'static bounds to LrScheduler::Record and derive Clone for scheduler records

crutcher:
### Checklist

- [NA] Made sure the book is up to date with changes in this PR.

### Changes

Made LrScheduler:Record match SimpleOptimizer:State Record requirements


laggui (approved):
Makes sense!

PR #4906, opened 2026-05-01
add cubecl integration to topk

Sublime12:
### Checklist

### Changes
add topk to tensor

### Testing
```sh
# activate cubecl tests in cubecl.rs then
cargo test-wgpu topk
cargo test-wgpu-nofuse topk
```


crates/burn-tensor/src/tensor/api/orderable.rs
```diff
@@ -231,9 +231,11 @@ where
...
     /// }
     /// ```
-    pub fn topk(self, k: usize, dim: usize) -> Self {
-        let k_indices = Tensor::arange(0..k as i64, &self.device());
-        self.sort_descending(dim).select(dim, k_indices)
+    pub fn topk(self, dim: usize, k: usize) -> Self {
+        assert!(self.shape()[dim] > k);
+        Tensor::new(K::topk(self.primitive, dim, k))
+        //let k_indices = Tensor::arange(0..k as i64, &self.device());
+        //self.sort_descending(dim).select(dim, k_indices)
```
louisfd: dead code

louisfd (approved):
LGTM

crates/burn-tensor/src/tensor/api/orderable.rs
```diff
@@ -231,9 +231,11 @@ where
     ///   // [[12.0], [6.0]]
     /// }
     /// ```
-    pub fn topk(self, k: usize, dim: usize) -> Self {
-        let k_indices = Tensor::arange(0..k as i64, &self.device());
-        self.sort_descending(dim).select(dim, k_indices)
+    pub fn topk(self, dim: usize, k: usize) -> Self {
```
laggui: That's a breaking change to the public facing API. I would preserve the order of arguments to `topk(k, dim)`.

laggui (changes requested):
I wouldn't break the `tensor.topk` API just for the order of k and dim. Instead we should have the new `tensor.argtopk` match this order.

For other backends we can keep the previous implementation as the default.

laggui (approved):

PR #4909, opened 2026-05-02
add autodiff for scatter_nd min/max/mul

cu9hue:
### Checklist

### Related Issues/PRs

Builds on top of #4709 

Closes issue #4433 

### Changes

At the moment backward pass throws for `scatter_nd` min, max, and mul. this provides an implementation for the backward pass

### Testing

Unit tests

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1062,18 +1062,6 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
     ) -> FloatTensor<Self> {
         use burn_backend::tensor::IndexingUpdateOp;
 
-        if matches!(
-            reduction,
-            IndexingUpdateOp::Mul | IndexingUpdateOp::Min | IndexingUpdateOp::Max
-        ) && (!data.node.requirement.is_none() || !values.node.requirement.is_none())
-        {
-            panic!(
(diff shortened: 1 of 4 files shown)


crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1185,36 +1173,177 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                        data.primitive,
+                        indices,
+                        values.primitive,
+                        IndexingUpdateOp::Mul,
+                    )),
                 }
+            }
+            IndexingUpdateOp::Min | IndexingUpdateOp::Max => {
+                // Unique indices are required for this backward formula; duplicate indices have
+                // undefined behavior for scatter_nd Min/Max gradients.
```
laggui, line 1254: I think the `Mul` implementation might return incorrect gradients for duplicate indices too. The current implementation for `data[idx]` would return `grad * val_2` (because assign requires unique indices) instead of `grad * val_1 * val_2`.

IMO we could just add a note to the `scatter_nd`  doc that explicitly states duplicate indices are not supported in general.

laggui (changes requested):
Thanks for extending the support!

laggui (approved):
Going through the PRs before we make the official 0.21, wanted to make sure this landed so I updated the doc myself!

cu9hue:
aww thank you! was about to get to it tonight

PR #4937, opened 2026-05-09
feat(metric): add BLEU score training metric

kimjune01:
Part of #544.

Adds sentence-level BLEU score metric for training, following
Papineni et al. (2002). Configurable max n-gram order (default BLEU-4)
and optional pad token stripping.

Same implementation pattern as existing CER/WER metrics.
9 tests covering edge cases (empty, single token, perfect match,
brevity penalty, custom n-gram order).


crates/burn-train/src/metric/bleu.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// Sets the maximum n-gram order.
+    ///
+    /// Common values: 1 (BLEU-1), 2 (BLEU-2), 4 (BLEU-4, default).
+    pub fn with_max_n(mut self, max_n: usize) -> Self {
+        assert!(max_n >= 1, "max_n must be at least 1");
+        self.max_n = max_n;
+        self.name = Arc::new(format!("BLEU-{max_n}"));
+        self
+    }
```
laggui, line 115: The formatted name should match in the default case (`BLEU-4`) too.

I don't think `with_max_n` should mutate an already constructed `BleuScore` metric. It should be a constructor instead. And if want to keep `BleuScore::new()` w/ default 4 it can simply call that constructor with `max_n` of 4.

crates/burn-train/src/metric/bleu.rs
```diff
@@ -0,0 +1,375 @@
...
+            let ref_count = ref_ngrams.get(ngram).copied().unwrap_or(0);
+            clipped += count.min(ref_count);
+            total += count;
+        }
+
+        if total == 0 {
+            // No n-grams of this order in candidate (sequence too short).
+            // Skip this order entirely instead of returning 0.
+            continue;
+        }
```
laggui: Hmm wouldn't this inflate the scores? Why skip instead of returning 0?

crates/burn-train/src/metric/bleu.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        if clipped == 0 {
+            // Zero precision at this order means the whole score is 0.
+            return 0.0;
+        }
```
laggui: I think it's become standard to add an optional smoothing. We should probably have that.

crates/burn-train/src/metric/bleu.rs
```diff
@@ -0,0 +1,375 @@
...
+                        .unwrap_or(target_seq.len());
+                    &target_seq[..len]
+                }
+                None => target_seq,
+            };
+
+            total_bleu += sentence_bleu(output_seq, target_seq, self.max_n);
+        }
+
+        let value = total_bleu / batch_size as f64;
```
laggui: The metric should accumulate the values for the batch, not compute the average of averages. The reported score will not be correct.

For corpus-level score, it should also track these to compute the metric for a complete epoch / corpus. To get this right though it would require a different state than `NumericMetricState` and an extension to `NumericEntry`. This can be left out of the scope for this PR, as long as we document that the epoch score aggregation will be a bit inaccurate.

The metric should still be fixed to compute the correct score at the batch level at least. And eventually with the correct state we'll be able to extend it for correct epoch aggregation.

laggui (changes requested):
Thanks for contributing this metric!

crates/burn-train/src/metric/bleu.rs
```diff
@@ -0,0 +1,529 @@
...
+
+        let precision = if clipped == 0 {
+            // Apply smoothing to zero-match precisions.
+            match smoothing {
+                BleuSmoothing::None => return 0.0,
+                BleuSmoothing::AddEpsilon(eps) => *eps / total as f64,
+                BleuSmoothing::Exponential => {
+                    let k = n + 1;
+                    1.0 / (1 << k) as f64 / total as f64
+                }
```
laggui, line 197: This doesn't match the method 3 (also used by SacreBLEU, per your comment).

I believe `k` should be a stateful multiplier that doubles for every n-gram order with zero matches.

References:
- https://aclanthology.org/W14-3346.pdf
- https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/metrics/bleu.py#L282

laggui (changes requested):

One minor discrepancy, otherwise should be good to go.

Oh, don't forget to cargo fmt.

kimjune01:
Thanks @laggui — fixed the smoothing in 805f6cc. `Exponential` now tracks a stateful multiplier that doubles on every n-gram order with zero matches, matching method 3 in Chen & Cherry (2014) and SacreBLEU's implementation. Also ran `cargo fmt`.

laggui (approved):

laggui:
I can take this over the line due to the breaking changes we merge in main

laggui:
Thanks for updating :pray:  I just had to fix the cargo fmt. CI failure is unrealted.

PR #4938, opened 2026-05-10
Fix TensorContainer downcast panic on backend mismatch (#2924, #3969)

### LINKED ISSUE #2924: Grad.get leads to panic due to unwrap() instead of returning the None option

### LINKED ISSUE #3969: Panic when getting module gradients with autodiff backend
**Describe the bug**
When trying to get gradients for a specific module with `GradientsParams::get` and an `AutodiffBackend`, the code panics because it tries to unwrap a `None` `Option`. After digging a bit into the code, I've seen this is because `GradientsParams::from_grads` uses `B::InnerBackend` to register. Of course, I can use the inner backend directly when calling `GradiensParams::get`, but this isn't intuitive (at least to me) and there's no indication that it should be called like that unless you dig into the source code.

**To Reproduce**
```
use burn::{Tensor, backend::{Autodiff, NdArray, ndarray::NdArrayDevice}, nn::{Linear, LinearConfig}, optim::GradientsParams};

fn main() {
    type MyBackend = NdArray;
    type MyAutodiffBackend = Autodiff<MyBackend>;
    let device = NdArrayDevice::default();
    let linear: Linear<MyAutodiffBackend> = LinearConfig::new(3, 3).init(&device);
    let input: Tensor<MyAutodiffBackend, 1> = Tensor::from_data([1.0, 0.0, -1.0], &device);
    let output = linear.forward(input);
    let loss = output.powi_scalar(2).sum();
    let grads = loss.backward();
    let grads = GradientsParams::from_grads(grads, &linear);
    let weight_grads = grads.get::<MyAutodiffBackend, 2>(linear.weight.id);
    println!("{:?}", weight_grads);
}
```

**Expected behavior**
It should not panic, or at least indicate that while the id was found, it does not correspond to the requested backend.

Capataina:
this fixes the .unwrap() panic in TensorContainer's get, get_mut_ref, and remove when an entry exists for an id but is stored for a different backend. you hit it when you pass B: AutodiffBackend to GradientsParams::get instead of B::InnerBackend. same root cause behind both #2924 and #3969, hit 8 months apart by two different reporters, so it's not a one-off, it's a real ergonomics gap.

the three accessors now return a Result with two variants, NotFound and TypeMismatch, both carrying the debug-formatted id. the public GradientsParams::get and remove (and Gradients::get and remove on the autodiff side) keep their existing Option return type though. internally they just match on the Result, mapping NotFound to None and panicking on TypeMismatch with a descriptive message that points at the inner-backend convention. this is the second of the two options laggui suggested on the original PR #2965 back in march, which avoids cascading Result through every public caller.

happy to switch to nathanielsimard's idea of making TensorContainer generic over the backend if you'd rather, that's the path #2965 was reviewed for but never got an answer on, which is what eventually killed it. marked this one as draft until you tell me which way to go.

one bonus thing I noticed while I was in there: the old remove did HashMap::remove first and then downcast on the boxed entry, which means on a type mismatch the entry was already gone from the HashMap by the time the unwrap panicked. so the old code was actually leaking the tensor on the failure path on top of panicking. the new version peeks the type via Any::is first and only commits to removing once the downcast is guaranteed to succeed. small but worth fixing, there's a test for it.

on the tests:

- the lucasmdjl reproducer from #3969 ported verbatim into burn-optim's test module (Linear plus Autodiff plus GradientsParams::get with the wrong backend, asserts the panic message contains "type mismatch")
- the equivalent on the remove path
- the peek-before-remove correctness test
- a repeated-remove test (second remove returns NotFound, not a panic)
- an idempotence test for wrong-backend reads (multiple bad gets don't corrupt anything)
- the existing happy-path test still passes

ran the workspace tests, clippy, fmt, rustdoc, all clean.

credit to @VirtualNonsense for #2965, the error enum shape is basically the same idea, just narrower in scope.

Closes #2924, closes #3969

edit: broke the close sentences in two so github autoparses both

### DIFF
--- a/crates/burn-autodiff/src/distributed.rs
+++ b/crates/burn-autodiff/src/distributed.rs
@@ -29,7 +29,9 @@ impl<B: DistributedBackend> DistributedRegistration for DistributedGradientRegis
             *n_required -= 1;
 
             if *n_required == 0 {
-                let tensor_ref = container.get_mut_ref::<B>(&id.value).unwrap();
+                let tensor_ref = container.get_mut_ref::<B>(&id.value).expect(
+                    "distributed: tensor freshly registered for this node id should be retrievable",
+                );
                 let tensor_ref = TensorRef(tensor_ref.get_mut_ref());
                 B::submit_gradient_sync(tensor_ref, sharded_params.clone());
             }
--- a/crates/burn-autodiff/src/grads.rs
+++ b/crates/burn-autodiff/src/grads.rs
@@ -1,6 +1,6 @@
 use burn_backend::{
     Backend, TensorMetadata, TensorPrimitive,
-    tensor::{FloatTensor, TensorContainer},
+    tensor::{FloatTensor, TensorContainer, TensorContainerError},
 };
 
 #[cfg(feature = "distributed")]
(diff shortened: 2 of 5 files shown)


Capataina:
fair enough, taking this one down. #4717 removes the backend generic from the user API so the unwrap never gets reached anyway. thanks for the pointer.

PR #4955, opened 2026-05-14
Bool tensor API improvements

Cielbird:
## Bool tensor API improvements

Minor improvements to the boolean tensor API

- Uses `Into<TensorData>` in `from_bool`. Same as in `from_ints` and `from_floats`. 
- Implemented bitwise operators (& | ^) for boolean tensors. 

When working with masks, this helps make the code less verbose.


laggui (approved):
LGTM!

Updated with main & fixed the merge conflicts

Issue #4945, opened 2026-05-11
Implement FFT ops backward


laggui:
FFT operations are not supported with autodiff.

Related: #4707 #4719

https://github.com/tracel-ai/burn/blob/ee2b832191fd182f5a56ecbcca8a5ff86fc93e94/crates/burn-autodiff/src/ops/module.rs#L1975-L1990

Issue #3591, opened 2025-08-21
Linking errors on Win 10 & Nvidia RTX 5090


AndrejMitrovic:
Hi,

I'm getting linking errors while building any of the examples in `main`, for example:

```
$ cargo run --example guide

          libcubecl_hip-7e3feabb3f09ba63.rlib(cubecl_hip-7e3feabb3f09ba63.cubecl_hip.2d088e512feaf181-cgu.14.rcgu.o) : error LNK2019: unresolved external symbol hipEventCreateWithFlags referenced in function _ZN10cubecl_hip7compute5fence5Fence3new17had250f213efb2314E␍
          libcubecl_hip-7e3feabb3f09ba63.rlib(cubecl_hip-7e3feabb3f09ba63.cubecl_hip.2d088e512feaf181-cgu.14.rcgu.o) : error LNK2019: unresolved external symbol hipEventRecord referenced in function _ZN10cubecl_hip7compute5fence5Fence3new17had250f213efb2314E␍
          libcubecl_hip-7e3feabb3f09ba63.rlib(cubecl_hip-7e3feabb3f09ba63.cubecl_hip.2d088e512feaf181-cgu.14.rcgu.o) : error LNK2019: unresolved external symbol hipStreamWaitEvent referenced in function _ZN10cubecl_hip7compute5fence5Fence4wait17h72b8c11ed962012fE␍
          libcubecl_hip-7e3feabb3f09ba63.rlib(cubecl_hip-7e3feabb3f09ba63.cubecl_hip.2d088e512feaf181-cgu.02.rcgu.o) : error LNK2001: unresolved external symbol hipStreamWaitEvent␍
```

This appears to be an [AMD library](https://rocm.docs.amd.com/projects/hipfort/en/develop/how-to/using-hipfort.html) it's linking to, but I'm not sure why it would try to link with this library when I'm running Nvidia hardware.

Let me know if I should re-file this under https://github.com/tracel-ai/cubecl/issues, or if you need more info or further steps I could take to debug this.

Thanks!

nathanielsimard:
It's strange, but it's normal to link those dependencies when running an example from the workspace, that's how Rust workspaces work. However if you import burn as a dependency it shound't cause any problem.

antimora:
As explained above, this comes from workspace feature unification when running examples inside the repo, not from using burn as a dependency. Closing.

PR #4956, opened 2026-05-14
autodiff for rfft/irfft

cofinite:
## Autodiff for `rfft` and `irfft`

Apologies in advance if something is off about the PR, I'm new to this 🥹 

### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed.*
- [x] Made sure the book is up to date with changes in this PR.**

*~~Unable to finish due to eventual linker errors (https://github.com/tracel-ai/burn/issues/3591 related?).~~ Has now been successfully run, but seemed to skip the autodiff tests?
**I'm not aware of any changes to be made.

### Related Issues/PRs

Solves https://github.com/tracel-ai/burn/issues/4945

### Changes

Implemented the missing module ops `rfft` and `irfft` in the autodiff backend.

### Testing

`cargo run-checks` was unable to run the tests because the `ndarray` backend has not implemented `rfft` yet (these tests are marked as `ignore` for now). The tests were run manually on `Autodiff<Wgpu>` instead.

### DIFF
--- a/crates/burn-autodiff/src/ops/module.rs
+++ b/crates/burn-autodiff/src/ops/module.rs
@@ -1,3 +1,5 @@
+use alloc::{vec, vec::Vec};
+
 use crate::Autodiff;
 use crate::checkpoint::base::Checkpointer;
 use crate::checkpoint::strategy::CheckpointStrategy;
@@ -7,9 +9,11 @@ use crate::ops::{Backward, Ops, unary};
 use crate::tensor::AutodiffTensor;
 
 use burn_backend::Backend;
+use burn_backend::TensorMetadata;
 use burn_backend::ops::attention::attention_fallback;
 use burn_backend::ops::*;
 use burn_backend::tensor::{FloatTensor, IntTensor};
+use burn_std::Slice;
 
 use super::OpsKind;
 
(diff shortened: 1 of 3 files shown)


crates/burn-backend-tests/tests/autodiff/rfft.rs
```diff
@@ -0,0 +1,52 @@
+use super::*;
+use burn_tensor::TensorData;
+use burn_tensor::Tolerance;
+use burn_tensor::signal;
+
+#[test]
+#[ignore = "rfft is not supported for ndarray"]
```
Sublime12: to run the tests, you need to remove the ignore

crates/burn-backend-tests/tests/autodiff/rfft.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    TensorData::assert_approx_eq::<FloatElem>(
+        &prod.to_data(),
+        &loss.to_data(),
+        Tolerance::default(),
+    );
+}
+
+#[test]
+#[ignore = "rfft is not supported for ndarray"]
```
Sublime12: same for here

cofinite:
I've removed the `ignore` on the tests as per your request, though as stated in the OP, they are expected to fail due to the `ndarray` backend not having `rfft` support:
```
---- autodiff::base::rfft::round_trip stdout ----

thread 'autodiff::base::rfft::round_trip' (11032) panicked at crates\burn-ndarray\src\ops\module.rs:387:9:
not yet implemented: rfft is not supported for ndarray

---- autodiff::base::rfft::should_diff_rfft stdout ----

thread 'autodiff::base::rfft::should_diff_rfft' (25908) panicked at crates\burn-ndarray\src\ops\module.rs:387:9:
not yet implemented: rfft is not supported for ndarray
```
It did not seem prudent to attempt using a different backend within the test suite itself, but I manually checked that the tests run successfully on `Autodiff<Wgpu>`.

crates/burn-backend-tests/tests/autodiff/rfft.rs
```diff
@@ -0,0 +1,50 @@
+use super::*;
+use burn_tensor::TensorData;
+use burn_tensor::Tolerance;
+use burn_tensor::signal;
+
+#[test]
+fn should_diff_rfft() {
```
laggui, line 11: The tests can be feature-gated instead 🙂 iirc the implementation is missing for ndarray and candle, so you can add:

```rust
#[cfg(not(any(feature = "ndarray", feature = "candle")))]
```

crates/burn-autodiff/src/ops/module.rs
```diff
@@ -1973,23 +1977,155 @@ impl<B: Backend, C: CheckpointStrategy> ModuleOps<Autodiff<B, C>> for Autodiff<B
...
+            .prepare::<C>([signal.node.clone()])
+            .compute_bound()
+            .stateful()
+        {
+            OpsKind::Tracked(prep) => prep.finish(state, spectrum),
+            OpsKind::UnTracked(prep) => prep.finish(spectrum),
+        };
+
+        let re = Self::float_slice(spectrum.clone(), &slices_re);
+        let im = Self::float_slice(spectrum.clone(), &slices_im);
```
laggui, line 2052: I think slice / cat should match the FFT `dim` no? Your tests only used dim = 0 so it might only pass in this particular case.

Might be a good idea to add tests for other dimensions too.

crates/burn-autodiff/src/ops/module.rs
```diff
@@ -1973,23 +1977,155 @@ impl<B: Backend, C: CheckpointStrategy> ModuleOps<Autodiff<B, C>> for Autodiff<B
...
+                    grads.register::<B>(node.id, grad_re);
+                }
+                if let Some(node) = node_im {
+                    grads.register::<B>(node.id, grad_im);
+                }
+            }
+        }
+
+        let signal = B::irfft(spectrum_re.primitive, spectrum_im.primitive, dim, n);
+        let n_fft = n.unwrap_or(signal.shape()[dim]);
```
laggui, line 2100: I *think* this is incorrect, and should be inferred from input spectrum size `2 * (spectrum_size - 1)` (when `n` is `None` for `irfft`).

laggui (changes requested):
Thanks for tackling the missing implementation!

crates/burn-backend-tests/tests/autodiff/rfft.rs
cofinite: Done!

crates/burn-autodiff/src/ops/module.rs
cofinite: It ends up not making a difference which dimension we choose here, because the tensor is immediately after sliced back in half along the same dimension. I've added a test to confirm this.

crates/burn-autodiff/src/ops/module.rs
cofinite: As I was investigating this, I ran into some issues regarding cases where `n` is `Some` more generally. Will look further into that later.

crates/burn-autodiff/src/ops/module.rs
laggui: Ahhh sorry you're right, somehow missed this was out of the actual backward implementation, but a workaround for single-output autodiff node (current design limitation) to accept the a split-output operation.

Thanks for adding tests nonetheless!

crates/burn-autodiff/src/ops/module.rs
cofinite: There's currently a separate issue where if `n` differs from the input signal length, the gradient from `rfft` has the wrong length (`n` instead of the original signal length). We would need to perform the same kind of zero-padding/truncation that `rfft` itself performs, but on the gradient. I'll likely have time to implement this over the weekend.

cofinite: The issue is also present in `irfft`. Both ops will have to store the length of their input in the state and then resize the gradient to match.

cofinite: > I _think_ this is incorrect, and should be inferred from input spectrum size `2 * (spectrum_size - 1)` (when `n` is `None` for `irfft`).

If `n` is `None`, the output signal length will be the FFT size that was used in the op. The expression you suggested will also result in the same value, so we could use either. I've kept this unchanged for now.

cofinite:
I've added more tests for edge cases, which all run successfully under `Autodiff<Wgpu>`:
```
running 7 tests
test tests::round_trip_with_some_n_greater ... ok
test tests::round_trip_with_dim_nonzero ... ok
test tests::round_trip_inverse_with_some_n_greater ... ok
test tests::round_trip ... ok
test tests::should_diff_rfft ... ok
test tests::round_trip_inverse_with_some_n_less ... ok
test tests::round_trip_with_some_n_less ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.30s
```
This implementation is, _I think_, future-proof for if/when `rfft`/`irfft` get support for non-power-of-two FFT sizes.

laggui (approved):
LGTM!

Just had to add a condition to ignore lower float precision in the tests.

Issue #4934, opened 2026-05-08
[Refactor] Backend generic removal


laggui:
Meta-issue following #4750

- [ ] Update documentation and book
- [x] Move `TensorKind` and others back to `burn-tensor` (this is a user level API type)
  - [x] Remove primitive from tensor API (private), moved traits should no longer be generic over backend
- [x] #4752
- [x] #4972 
- [x] Integrate distributed backend w/ tests
- [x] Integrate remote backend
- [x] #5035
- [x] Switch dispatch `default_backend` to flex (unsure; currently some limitations for no-std targets w/o atomic CAS)
- [ ] Configurable `Device` priority (currently hardcoded, feature based)
  - Default should be "best available"; ~~right now vulkan / metal / webgpu are mutually exclusive features~~ (#5001)
  - `DeviceType` in prelude?
- [ ] Third-party backend integrations with `Dispatch`

PR #4958, opened 2026-05-14
Replace tensor primitive type

laggui:
### Checklist

### Related Issues/PRs

Follow-up to #4957 
Part of #4934
Required for #4752

### Changes

Removed tensor kind associated type for the tensor primitive, replaced by a `PrimitiveKind` enum.


crates/burn-tensor/src/bridge/ops/bool.rs
```diff
@@ -9,209 +9,233 @@ use burn_std::{DType, ExecutionError, IndexingUpdateOp, Shape, Slice};
 use crate::{
     Bool, Device,
     bridge::{BasicAutodiffOps, BasicOps, TransactionOp},
-    ops::{BoolTensor, IntTensor},
+    ops::PrimitiveKind,
```
nathanielsimard: Instead of `PrimitiveKind` we could use `BridgeTensor`

PR #4960, opened 2026-05-15
[Breaking] Add multiclass and multi-label support to AUROC metric

manucouto1:
### Checklist

### Related Issues/PRs

Part of #544 (training metrics). Addresses the multiclass ROC AUC
extension raised in the issue thread.

### Changes

AUROC was restricted to binary classification and panicked on more
than two classes. This PR generalizes it:

- One-vs-Rest AUROC for binary, multiclass and multi-label, via the
  Mann–Whitney U statistic, aggregated with a `ClassReduction`
  (Macro = mean of per-class AUROC, Micro = pooled). `AurocMetric`
  holds a `ClassReduction` field directly (no extra config wrapper).
- `AurocMetric` now uses `ConfusionStatsInput`, so it is implicitly
  adapted for `ClassificationOutput` and `MultiLabelClassificationOutput`
  (same mechanism as Precision/Recall/FBetaScore).
- Public API aligned with the other classification metrics:
  `binary()`, `multiclass(class_reduction)`, `multilabel(class_reduction)`.
- Per-class degenerate columns (only positives or only negatives in the
  batch) are excluded from the Macro mean; if every class is degenerate
  the batch AUROC is undefined and reported as `0.5` (chance level) with
  a warning — rather than `NaN` (which would poison the running mean,
  since `NumericMetricState` aggregates without NaN filtering) or `0.0`
  (the worst score, misleading for an undefined batch).
- Removed the now-dead `AurocInput` type and its `Adaptor` impl.
- Book updated: AUROC added to `MultiLabelClassificationOutput`'s
  adapted metrics and to the `ConfusionStatsInput` footnote.

Review feedback addressed: dropped the `RankingMetricConfig` wrapper in
favour of a plain `ClassReduction` field; renamed the internal method to
`compute_auc`.

**Breaking:** `AurocInput` removed (use the built-in adaptors or
`ConfusionStatsInput`); the old no-arg `AurocMetric::new()` removed
(use `binary()`); metric name is now `AUROC [Macro|Micro]`.

Follow-up (out of scope): the pairwise implementation is O(n²·c) memory;
a sort-based O(n log n) version could replace it.

### Testing

- Parametrized `#[rstest]` suite (binary / multiclass / multi-label ×
  Micro/Macro) with inputs and expected values produced by an
  independent reference equivalent to scikit-learn's `roc_auc_score`
  (distinct scores, so the statistic is unambiguous); plus edge cases:
  perfect separation, chance level, Macro dropping a degenerate class,
  all-degenerate → 0.5, and name-by-reduction.
- `cargo test -p burn-train --lib metric::` passes; `cargo fmt --check`
  and `cargo clippy -p burn-train` are clean.
- The macOS CI failure is a pre-existing, flaky `burn-vision` / cubecl
  autotune panic unrelated to this change (the same `main` commit passes
  on re-run; this PR only touches `burn-train` metrics). Full
  `cargo run-checks` can't complete locally because the
  `wasm32-unknown-unknown` target isn't installed; CI covers the full
  matrix.

### DIFF
--- a/crates/burn-train/src/metric/auroc.rs
+++ b/crates/burn-train/src/metric/auroc.rs
@@ -2,100 +2,136 @@ use core::f64;
 
 use super::MetricMetadata;
 use super::state::{FormatOptions, NumericMetricState};
-use crate::metric::{Metric, MetricName, Numeric, SerializedEntry};
-use burn_core::tensor::{Int, Tensor};
+use crate::metric::{
+    ClassReduction, ConfusionStatsInput, Metric, MetricName, Numeric, SerializedEntry,
+};
+use burn_core::tensor::{Bool, Tensor};
+use std::sync::Arc;
 
-/// The Area Under the Receiver Operating Characteristic Curve (AUROC, also referred to as [ROC AUC](https://en.wikipedia.org/wiki/Receiver_operating_characteristic)) for binary classification.
+/// The Area Under the Receiver Operating Characteristic Curve (AUROC, also
+/// referred to as [ROC AUC](https://en.wikipedia.org/wiki/Receiver_operating_characteristic)).
+///
+/// Supports binary, multiclass and multi-label classification through a
+/// One-vs-Rest decomposition, aggregated with the configured
+/// [class reduction](ClassReduction).
 #[derive(Clone)]
 pub struct AurocMetric {
     name: MetricName,
     state: NumericMetricState,
-}
-
-/// The [AUROC metric](AurocMetric) input type.
-#[derive(new)]
-pub struct AurocInput {
-    outputs: Tensor<2>,
-    targets: Tensor<1, Int>,
+    class_reduction: ClassReduction,
 }
 
 impl Default for AurocMetric {
     fn default() -> Self {
-        Self::new()
+        Self::new(Default::default())
     }
 }
 
 impl AurocMetric {
-    /// Creates the metric.
-    pub fn new() -> Self {
+    fn new(class_reduction: ClassReduction) -> Self {
+        let state = Default::default();
+        let name = Arc::new(format!("AUROC [{:?}]", class_reduction));
+
         Self {
-            name: MetricName::new("AUROC".to_string()),
-            state: Default::default(),
+            state,
+            class_reduction,
+            name,
         }
     }
 
-    fn binary_auroc(&self, probabilities: &Tensor<1>, targets: &Tensor<1, Int>) -> f64 {
-        let n = targets.dims()[0];
+    /// AUROC metric for binary classification.
+    #[allow(dead_code)]
+    pub fn binary() -> Self {
+        Self::new(ClassReduction::default())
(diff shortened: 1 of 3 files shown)


crates/burn-train/src/metric/auroc.rs
```diff
@@ -2,100 +2,136 @@ use core::f64;
...
+            .bool_not()
+            .argwhere()
+            .squeeze_dim::<1>(1);
 
-        let area_under_curve = self.binary_auroc(&probabilities, &input.targets);
+        if keep.dims()[0] == 0 {
+            log::warn!(
+                "AUROC is undefined: no class has both positive and negative samples in the batch."
+            );
+            return 0.0;
```
nathanielsimard: We could return NaN or 0.5 instead of zero here.

manucouto1: Good catch. Went with 0.5 instead of NaN so one degenerate batch doesn't turn the epoch's running metric into NaN (NumericMetricState aggregates without NaN filtering).

crates/burn-train/src/metric/classification.rs
```diff
@@ -7,6 +7,21 @@ pub struct ClassificationMetricConfig {
...
+}
+
+impl From<ClassificationMetricConfig> for RankingMetricConfig {
+    fn from(config: ClassificationMetricConfig) -> Self {
+        Self {
+            class_reduction: config.class_reduction,
+        }
+    }
+}
+
```
laggui: I don't think this config is necessary. The `Auroc` can simply have a `ClassReduction` field instead, no?

crates/burn-train/src/metric/auroc.rs
```diff
@@ -2,100 +2,137 @@ use core::f64;
...
+        let correct_pairs = reduce(si.clone().greater(sj.clone()).float() * valid.clone());
+        let tied_pairs = reduce(si.equal(sj).float() * valid);
 
         (correct_pairs + 0.5 * tied_pairs) / num_pairs
     }
-}
 
-impl Metric for AurocMetric {
-    type Input = AurocInput;
+    fn aucroc(&self, predictions: &Tensor<2>, targets: &Tensor<2, Bool>) -> f64 {
```
laggui: Small nit: would rename to something more meaningful e.g. `compute_auc` as it is an `AurocMetric` method anyway.

laggui (commented):
Thanks for contributing!

Just a few minor comments

crates/burn-train/src/metric/classification.rs
manucouto1: Yeah, I agree — that’s simpler that way.

crates/burn-train/src/metric/auroc.rs
manucouto1: Done — renamed to compute_auc.

laggui (approved):

PR #4962, opened 2026-05-15
Fix/shared tensor

### LINKED ISSUE #4947: burn-fusion: cross-thread HandleContainer race with shared Tensor + Wgpu backend
## Summary

`burn-fusion` panics with `Should have handle for tensor TensorId { value: N }` when a `Tensor` is shared between OS threads via an `Arc<RwLock<_>>` and one thread keeps swapping the value while others keep reading it.

## Failure signature

```
thread 'DSU-4-0' panicked at burn-ir/src/handle.rs:97:
  Should have handle for tensor TensorId { value: N }

thread 'DSU-4-0' panicked at burn-fusion/src/stream/execution/ordering.rs:65:
  index out of bounds: the len is 0 but the index is 0

thread '<reader>' panicked at burn-fusion/src/client.rs:175:
  called `Result::unwrap()` on an `Err` value: CallError
```

All three panics fire on burn-fusion's \`DSU-*\` device service thread (the runner where \`MultiStream::register\` runs); the `CallError` then propagates back to whichever reader was waiting on the device queue.

## Repro

`Cargo.toml\:
```toml
[package]
name = "burn-fusion-race"
version = "0.1.0"
edition = "2024"

[dependencies]
burn = { git = "https://github.com/tracel-ai/burn", features = ["wgpu", "fusion"] }
```

`tests/race.rs`:
```rust
use burn::backend::wgpu::{Wgpu, WgpuDevice};
use burn::prelude::*;
use std::sync::{Arc, RwLock};

type B = Wgpu;

const ITERS: usize = 10_000;
const READERS: usize = 8;

#[test]
fn cross_thread_handle_race() {
    let device = WgpuDevice::DefaultDevice;
    let shared = Arc::new(RwLock::new(Tensor::<B, 2>::ones([512, 512], &device)));

    let writer = {
        let shared = shared.clone();
        let device = device.clone();
        std::thread::spawn(move || {
            for i in 0..ITERS {
                let cur = shared.read().unwrap().clone();
                let next = cur * 1.0001_f32 + Tensor::full([512, 512], i as f32 * 0.001, &device);
                *shared.write().unwrap() = next;
            }
        })
    };

    let readers: Vec<_> = (0..READERS)
        .map(|_| {
            let shared = shared.clone();
            std::thread::spawn(move || {
                for _ in 0..ITERS {
                    let snap = shared.read().unwrap().clone();
                    let _ = (snap.clone() * snap).sum().into_data();
                }
            })
        })
        .collect();

    writer.join().expect("writer thread panicked");
    for (i, r) in readers.into_iter().enumerate() {
        r.join().unwrap_or_else(|e| panic!("reader {i} panicked: {e:?}"));
    }
}
```

Run with `cargo test`. Reducing `ITERS` or `READERS` lowers the failure rate; at the defaults it fails consistently on my macbook.

nathanielsimard:
Fix https://github.com/tracel-ai/burn/issues/4947

Most of the new code is more tests and documentation. The real solution removes a lot of code, since we leverage properties of the fusion IR mechanism to "remove" shared tensors from stream processing.


laggui (approved):

laggui:
CI is being weird on macos (has been failing to build some deps intermittently for a week).
Linux CI gets killed on codecov (OOM?), has been happening since we merged the big breaking change. Likely related to the amount of code / symbols with `burn-tensor` -> `burn-dispatch` -> backend crates. Need to fix.

But not related to this PR.

PR #4963, opened 2026-05-16
feat(metric): add AUC-PR (Average Precision) training metric

manucouto1:
### Checklist

### Related Issues/PRs

Part of #544 (training metrics).

Follow-up to #4960 (multiclass/multi-label AUROC), now **merged**. This PR
has been rebased onto `main` and is standalone — its diff contains only the
AUC-PR metric.

### Changes

Adds `AucPrMetric`, computing **Average Precision** — the standard
non-interpolated estimator of the area under the precision-recall curve,
equivalent to scikit-learn's `average_precision_score` (not the biased
trapezoidal AUC).

- One-vs-Rest for binary, multiclass and multi-label, aggregated with a
  Micro/Macro `ClassReduction`.
- Holds a plain `ClassReduction` field and consumes `ConfusionStatsInput`,
  so it is implicitly adapted for `ClassificationOutput` and
  `MultiLabelClassificationOutput` (same mechanism as
  Precision/Recall/FBetaScore/AUROC).
- Degenerate columns (no positive sample) yield `NaN` and are dropped
  from the Macro mean; if every column is degenerate the metric is 0.0
  with a warning.
- Book updated (metric table + `ConfusionStatsInput` footnote).

Additive, non-breaking (new metric).

### Testing

- Parametrized `#[rstest]` suite: binary/multiclass/multi-label ×
  Micro/Macro, with inputs and expected values produced by an
  independent reference equivalent to `sklearn.metrics.average_precision_score`
  (distinct scores, so the step-wise estimator matches exactly).
- `cargo test -p burn-train --lib metric::` passes.
- `cargo fmt`/`cargo clippy` on `burn-train` pass.
- Full `cargo run-checks` can't complete locally (`wasm32-unknown-unknown`
  target not installed in my environment); CI covers the full matrix.

### DIFF
--- a/burn-book/src/building-blocks/metric.md
+++ b/burn-book/src/building-blocks/metric.md
@@ -11,6 +11,7 @@ throughout the training process. We currently offer a restricted range of metric
 | Recall              | Calculate recall in percentage                                                              |
 | FBetaScore          | Calculate F<sub>β </sub>score in percentage                                                 |
 | AUROC               | Calculate the area under curve of ROC in percentage                                         |
+| AUC-PR              | Calculate the area under the precision-recall curve (average precision) in percentage       |
 | Loss                | Output the loss used for the backward pass                                                  |
 | CharErrorRate (CER) | Calculate Character Error Rate in percentage                                                |
 | WordErrorRate (WER) | Calculate Word Error Rate in percentage                                                     |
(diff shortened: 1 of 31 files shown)


laggui (commented):
The implementation LGTM at the batch level but the epoch aggregation will be incorrect when using a `NumericMetricState`. This is actually an existing issue for auroc, precision and recall metrics.

This would require a new metric state type. For this metric and auroc we probably need to accumulate the running predictions and targets. For others we would need TPs, FPs and FNs.

Do you want to fix the `AucPrMetric` in this PR? The others are existing issues so we should tackle them separately.

manucouto1:
Sounds good, I'll fix it in this PR. Btw the confusion matrix would be easy to derive from these accumulated values too, so I'm happy to take on the other fix as well in a separate PR if you'd like.

manucouto1:
Fixed the epoch aggregation for `AucPrMetric`. Summary of the approach:

- Added a `PredictionAccumulatorState` that keeps the per-batch predictions/targets and lets the metric recompute AP over the whole epoch on each update (rank-based metrics can't be derived from a per-batch mean).
- The remaining issue was that the store (`NumericMetricsAggregate`) re-aggregates the logged per-batch points as a sample-weighted mean, which is still wrong here. I first prototyped a `NumericEntry::Snapshot` variant, but that encodes an aggregation *policy* in the *value* type, which felt off. Instead I made it a metric attribute: `NumericAggregation { Mean, Last }` on `NumericAttributes`, and the store takes the last value for `Last` metrics. `NumericEntry` stays a plain value type.
- Scope: this PR only touches AUC-PR. The shared infra (`PredictionAccumulatorState`, `NumericAggregation`) is ready for AUROC/precision/recall to adopt in their own follow-up PRs, as you suggested.

Two things worth flagging:
- **Memory/cost**: the accumulator grows O(N·classes) per epoch and AP is recomputed each update. Fine for typical sizes; just noting it.
- **DDP**: "last value" is correct for single-device; multi-worker would need cross-worker accumulation, out of scope here.

laggui (approved):
Thanks for addressing my comments! That's more accurate.

The only caveat now is that the epoch-level metric is recomputed at every batch (as you mention in the memory/cost point). I think we'll have to rethink the current design, as some metrics only make sense over the whole distribution (epoch) and/or are more expensive to compute. In such cases, the update should only accumulate the required statistics, and the metric should only be computed at the end. That's definitely out of scope though, we need to address this separately.

PR #4967, opened 2026-05-17
feat(metric): add ROUGE-L score metric

### LINKED ISSUE #544: Training metrics

jaweed3:
### Checklist:

- Confirmed that cargo run-checks command has been executed.
- Made sure the book is up to date with changes in this PR.

### Related Issues/PRs:

Closes #544 (ROUGE score item)

### Changes:

- Adds RougeLScore, a training metric that computes the F1 score based on the longest common subsequence (LCS) between predicted and reference token sequences.
- New `crates/burn-train/src/metric/rouge.rs` implementing Metric + Numeric traits
- Registered in `crates/burn-train/src/metric/mod.rs`
- Supports configurable pad token for padded sequences
- Follows same pattern as BleuScore, CharErrorRate, WordErrorRate

### Testing:

8 unit tests covering:

- Perfect match (F1 = 100%)
- No match (F1 = 0%)
- Partial subsequence overlap
- Shorter candidate with padding
- Padding stripping
- Batch averaging
- Both sequences empty
- Clear reset


crates/burn-train/src/metric/rouge.rs
```diff
@@ -0,0 +1,282 @@
...
+
+            if ref_len == 0.0 && cand_len == 0.0 {
+                total_f1 += 100.0;
+                continue;
+            }
+
+            if ref_len == 0.0 || cand_len == 0.0 {
+                total_f1 += 0.0;
+                continue;
+            }
```
laggui, line 122: += 0 is a no-op here, we can just continue

laggui (commented):
Just a minor comment, otherwise I think the implementation is a good starting point for rouge (without any other additional configuration).

Thanks for contributing!

crates/burn-train/src/metric/rouge.rs
jaweed3: fixed

laggui (approved):

PR #4974, opened 2026-05-19
chore: Update to reference refactor

wingertge:
### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/cubek/pull/279

### Changes

Updates to breaking changes upstream

### Testing

Tests all pass


crates/burn-backend-tests/tests/tensor/float/ops/arg.rs
```diff
@@ -44,167 +46,3 @@ fn test_argmin_2d_dim1() {
...
 }
-
-#[test]
-fn test_argmax_flipped() {
-    // Flip [1, 5, 3, 2, 4] -> [4, 2, 3, 5, 1]; max is at index 3.
-    let tensor = TestTensor::<1>::from([1.0, 5.0, 3.0, 2.0, 4.0]);
-    let output = tensor.flip([0]).argmax(0);
-
-    output.into_data().assert_eq(&TensorData::from([3]), false);
-}
```
laggui, line 55: I think git automatically deleted those from a merge/rebase, we can keep the additional arg tests.

laggui (commented):
Looks like there was a bug introduced with fusion, I get a bunch of `Type to be registered` related to `DynElem` type.

```
thread 'DSD-0-0' (216234) panicked at /home/laggui/.cargo/git/checkouts/cubecl-058c47895211d464/a8ddb49/crates/cubecl-core/src/frontend/element/typemap.rs:251:48:
Type to be registered
```

It's not entirely clear at first glance what is the root cause though

crates/burn-backend-tests/tests/tensor/float/ops/arg.rs
wingertge: yeah I noticed when looking over the PR again, it's fixed now

wingertge:
> Looks like there was a bug introduced with fusion, I get a bunch of `Type to be registered` related to `DynElem` type.
> 
> ```
> thread 'DSD-0-0' (216234) panicked at /home/laggui/.cargo/git/checkouts/cubecl-058c47895211d464/a8ddb49/crates/cubecl-core/src/frontend/element/typemap.rs:251:48:
> Type to be registered
> ```
> 
> It's not entirely clear at first glance what is the root cause though

`burn-backend-tests` run fusion by default nowadays right? Cause I can't seem to replicate those on either CUDA or Vulkan

laggui:
Yeah running `cargo test-vulkan` or `cargo test-cuda` under `burn-backend-tests` will run with fusion by default. You have to specify `-nofuse` (e.g. `test-vulkan-nofuse`).

That's weird, I get the error almost immediately.

```sh
~/workspace/burn/crates/burn-backend-tests$ cargo test-vulkan
     Running unittests src/lib.rs (/home/laggui/workspace/burn/target/release/deps/burn_backend_tests-213f75f9872ec59e)

running 0 tests

     Running tests/autodiff.rs (/home/laggui/workspace/burn/target/release/deps/autodiff-471f19d95afae98a)

running 488 tests

thread 'DSD-4-0' (24253) panicked at /home/laggui/.cargo/git/checkouts/cubecl-058c47895211d464/a8ddb49/crates/cubecl-core/src/frontend/element/typemap.rs:251:48:
Type to be registered

thread 'DSD-4-0' (24253) panicked at /home/laggui/.cargo/git/checkouts/cubecl-058c47895211d464/a8ddb49/crates/cubecl-core/src/frontend/element/typemap.rs:251:48:
Type to be registered
test autodiff::base::add::should_diff_add_scalar ... FAILED

thread 'DSD-4-0' (24253) panicked at /home/laggui/.cargo/git/checkouts/cubecl-058c47895211d464/a8ddb49/crates/cubecl-core/src/frontend/element/typemap.rs:251:48:
Type to be registered

thread 'DSD-4-0' (24253) panicked at /home/laggui/.cargo/git/checkouts/cubecl-058c47895211d464/a8ddb49/crates/cubecl-core/src/frontend/element/typemap.rs:251:48:
Type to be registered
test autodiff::base::add::test_add_complex_1 ... FAILED
test autodiff::base::add::should_diff_add ... FAILED
[...]
test result: FAILED. 259 passed; 219 failed; 10 ignored; 0 measured; 0 filtered out; finished in 90.97s
```

with vulkan or cuda.

And ofc macos runners have been no help recently, failing to just setup rust :sweat_smile:

wingertge:
> Yeah running `cargo test-vulkan` or `cargo test-cuda` under `burn-backend-tests` will run with fusion by default. You have to specify `-nofuse` (e.g. `test-vulkan-nofuse`).

Turns out I'm just an idiot 😅 I was using a custom test command so I can toggle `--release` on or off and it was based on the aliases when I wrote it. Obviously won't update when fusion is added 🤦 I can definitely reproduce now, looking into it.

laggui:
Ahhh ok good to hear it's reproducible. I have a multiple clones of the repo so I thought maybe I ran the tests on the wrong one first :smile:

crates/burn-cubecl-fusion/src/optim/reduce/args.rs
```diff
@@ -209,8 +211,34 @@ impl CubeType for FusedReduceState {
     type ExpandType = FusedReduceStateExpand;
 }
 
+impl IntoExpand for FusedReduceStateExpand {
```
nathanielsimard, line 214: I would add a comment here on why we do a manual implementation

nathanielsimard (approved):

crates/burn-cubecl-fusion/src/optim/reduce/args.rs
wingertge: I feel like it's kind of obvious, it's because the expand type isn't generated so it can't be automatically implemented 😅 Same as the `CubeType` impl above that.

wingertge:
@laggui should be fixed now, tests pass on Vulkan and CUDA at least

laggui:
> @laggui should be fixed now, tests pass on Vulkan and CUDA at least

Confirmed :pray: thanks!

> I was using a custom test command so I can toggle `--release` on or off and it was based on the aliases when I wrote it.

If it's a test command you find useful we can update/add the aliases btw!

On a related note, what config do you use for clippy? It always seems that some lints eventually creep in with changes that are not flagged by our clippy command, but your PRs always fix them. Might be a good idea to update the config for the repo as well.

wingertge:
> > I was using a custom test command so I can toggle `--release` on or off and it was based on the aliases when I wrote it.
> 
> If it's a test command you find useful we can update/add the aliases btw!

It's a custom nushell script so I can get good completions and stuff. But an alias for non-release tests would help, then I could use that and don't need to rely on copy pasting the args from the aliases.

> On a related note, what config do you use for clippy? It always seems that some lints eventually creep in with changes that are not flagged by our clippy command, but your PRs always fix them. Might be a good idea to update the config for the repo as well.

Just `clippy` with `--all-targets`. The lints are always in the tests so I assume it's just missing `--all-targets`, I think it only checks the regular code by default and not tests.

PR #4977, opened 2026-05-19
perf(core)!: improve compile times via opaque inner types to break dependency chain

nathanielsimard:
Drastically improves compilation time. For a tensor-only application with only Burn as a dependency, incremental compilation with the `--release` flag drops from 15s to <1s, and around 1s with all backends activated.

I've checked Miri manually, but plan on extracting the unsafe part of the code inside cubecl-common that is already well equip to automatically test unsafe code on its CI.


crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -636,3 +611,47 @@ pub fn shrink<const D: usize>(tensor: Tensor<D>, lambda: f64, bias: f64) -> Tens
     let mask = abs_tensor.lower_equal_elem(lambda);
     shrunk.mask_fill(mask, 0)
 }
+
+// =========================================================================
+// Non-generic implementation helpers (outlined from the generic API).
+// =========================================================================
```
laggui, line 618: For contributors that stumble upon this code, we might want to add a note that explains why these implementations were factored out into a separate function so the pattern is duplicated when a method requires a dispatch method to be called.

/edit: the pattern is repeated and explained only in some places, maybe we could just have a burn-tensor module-wide doc for that instead?

crates/burn-tensor/src/tensor/api/transaction.rs
```diff
@@ -20,16 +20,77 @@ use burn_dispatch::Dispatch;
...
+type TransactionInner = MaybeUninit<TransactionPrimitive<Dispatch>>;
+
+/// Storage for [`Transaction`]. Holds the raw bytes of a
+/// [`TransactionPrimitive<Dispatch>`].
+///
+/// Intentionally has no type-level alignment marker (e.g.
+/// `[TransactionInner; 0]`), since that would re-introduce a `burn_dispatch`
+/// dependency in the type itself and undermine the compile-time goal of this
+/// obfuscation. Alignment must therefore be handled at access sites
+/// (TODO: bring back proper alignment without leaking the type).
```
laggui, line 35: I think this was left as a comment before the `obfuscate_type` macro was added?

crates/burn-tensor/src/device.rs
```diff
@@ -90,24 +117,236 @@ impl PartialEq for Device {
...
+
+    /// CUDA device.
     ///
-    /// # Example
+    /// Pass [`DeviceIndex::Default`] (or `DeviceIndex::Default` via
+    /// `Default::default()`) for the default CUDA device, or
+    /// [`DeviceIndex::Specified(i)`](DeviceIndex::Specified) / [`DeviceIndex::new(i)`](DeviceIndex::new)
+    /// to target hardware index `i`.
+    #[cfg(feature = "cuda")]
+    pub fn cuda(index: DeviceIndex) -> Self {
```
laggui: For constructors that take a device index we could have `index: impl Into<DeviceIndex>` with some conversions implemented for i32/i64/usize? So we'd have this instead:

```diff
- Device::cuda(DeviceIndex::new(0))
+ Device::cuda(0)
```

PR #4980, opened 2026-05-20
Feat: Add training and evaluation progress logger and add them to the event processors

Zatiji:
### Changes
Problem: There was no way to observe training and evaluation progress from outside the learner — users had no hook to persist progress events (e.g., to a file, database, or external service) beyond what the TUI/CLI renderer displays.

Solution: Two new observer traits, `TrainingProgressLogger` and `EvaluationProgressLogger`, define a structured call sequence that fires at well-defined lifecycle points:
- **TrainingProgressLogger**: called on training start (with total_epochs), at the start/end of each split (train/valid), on every processed batch (with items_processed), at the end of each full epoch, and at training end.
- **EvaluationProgressLogger**: called on evaluation start, at the start/end of each test split (with name and total_items), on every processed batch, and at evaluation end.

To support this, the` LearnerEvent` and `EvaluatorEvent` enums were extended:

- `LearnerEvent::Star`t now carries total_epochs.
- New variants `StartSplit(usize)` and `EndSplit(usize)` replace the previous overloaded use of EndEpoch for per-split bookkeeping.
- New `EvaluatorEvent::StartTest(EvaluationName, usize)` carries the split name and item count.
- Both `FullEventProcessorTraining` / `FullEventProcessorEvaluation` and `MinimalEventProcessor` were updated to carry an `optional Box<dyn TrainingProgressLogger>` / `Box<dyn EvaluationProgressLogger>`, wired in via a builder method `with_progress_logger(). EvaluatorBuilder` exposes the same method for the evaluation path.

### Testing
The MNIST example was extended with two concrete implementations (`FileTrainingProgressLogger` and `FileEvaluationProgressLogger`) that append structured log lines to files in append mode. Both are wired into the training and evaluation pipelines in `examples/mnist/src/training.rs`, exercising the full call sequence end-to-end during a real training run.


crates/burn-train/src/evaluator/builder.rs
```diff
@@ -120,6 +122,24 @@ impl<EC: EvaluatorComponentTypes> EvaluatorBuilder<EC> {
...
 
+    /// Register a progress logger to track and store evaluation progress.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use burn_train::logger::FileProgressLogger;
+    ///
+    /// let evaluator = EvaluatorBuilder::new(...)
+    ///     .with_progress_logger(FileProgressLogger::new("eval_progress.log"));
```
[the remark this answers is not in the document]
Charles23R: This is misleading as the file progress logger doesn't exist in burn. I don't think we need to provide an example, you can get inspiration from the other functions docs in the builder

examples/mnist/src/training.rs
```diff
@@ -104,8 +105,12 @@ pub fn run(device: Device) {
             Aggregate::Mean,
             Direction::Lowest,
             Split::Valid,
-            StoppingCondition::NoImprovementSince { n_epochs: 5 },
+            StoppingCondition::NoImprovementSince { n_epochs: 2 },
```
Charles23R: oversight

examples/mnist/src/training.rs
```diff
@@ -104,8 +105,12 @@ pub fn run(device: Device) {
...
         ))
+        .with_progress_logger(
+            FileTrainingProgressLogger::new("./training_progress.log")
```
Charles23R: I would put the log file in the artifact directory. Same on line 174

examples/custom-learning-strategy/src/training.rs
```diff
@@ -206,7 +206,7 @@ impl<LC: LearningComponentsTypes> SupervisedLearningStrategy<LC> for MyCustomLea
 
                 event_processor.process_valid(LearnerEvent::ProcessedItem(item));
             }
-            event_processor.process_valid(LearnerEvent::EndEpoch(epoch));
+            event_processor.process_valid(LearnerEvent::EndSplit(epoch));
```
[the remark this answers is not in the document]
Charles23R: Valid. There is the same problems at some places in burn also I think. We need to make sure that all the strategies call the new events correctly.

crates/burn-train/src/metric/processor/full.rs
```diff
@@ -107,10 +128,18 @@ impl<T: ItemLazy> EventProcessorEvaluation for FullEventProcessorEvaluation<T> {
                 definitions
                     .iter()
                     .for_each(|definition| self.renderer.register_metric(definition.clone()));
+                if let Some(logger) = &mut self.progress_logger {
+                    logger.start(EVALUATOR_TEST_SPLITS);
```
Charles23R: On veut probablement passer le nombre de test splits dans `EvaluatorEvent::Start`

crates/burn-train/src/metric/processor/full.rs
```diff
@@ -139,9 +168,16 @@ impl<T: ItemLazy> EventProcessorEvaluation for FullEventProcessorEvaluation<T> {
...
 
                 let indicators = self.progress_indicators(&progress);
+                if let Some(logger) = &mut self.progress_logger {
+                    logger.update_test(progress.progress.items_processed);
+                }
                 self.renderer.render_test(progress, indicators);
             }
             EvaluatorEvent::End(summary) => {
+                if let Some(logger) = &mut self.progress_logger {
+                    logger.end_test();
```
Charles23R, line 175: We probably want to add a `EndTest` event. Right now it's harcdcoded to a single test split, which won't always be the case.

crates/burn-train/src/metric/processor/base.rs
```diff
@@ -8,11 +8,18 @@ use crate::{
...
+        total_epochs: usize,
+    },
     /// Signal that an item have been processed.
     ProcessedItem(TrainingItem<T>),
-    /// Signal the end of an epoch.
+    /// Signal the start of a split (train or valid), carrying the total number of items in that split.
+    StartSplit(usize),
+    /// Signal the end of a split (train or valid), carrying the current epoch number.
+    EndSplit(usize),
+    /// Signal the end of a full epoch (after both train and valid splits have completed).
```
Charles23R: Nitpicking, but if we want to be generic over the possible splits, the doc should reflect that.

crates/burn-train/src/learner/supervised/strategies/multi/strategy.rs
```diff
@@ -77,6 +78,7 @@ impl<LC: LearningComponentsTypes> SupervisedLearningStrategy<LC> for MultiDevice
                 &mut event_processor,
                 &training_components.interrupter,
             );
+            event_processor.process_train(LearnerEvent::EndEpoch(epoch));
```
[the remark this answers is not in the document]
Charles23R: We need to send the new events accurately in all training strategies implemented in burn-train. I see that you did it correctly for `SingleDeviceTrainingStrategy`, but there is also `MultiDevice` and `DDP`.

Charles23R (changes requested):

Charles23R (approved):
LGTM!

PR #4982, opened 2026-05-20
Update cubek + Interpolate nearest exact mode

SamuelBelanger:
### Checklist

### Related Issues/PRs

### Changes

### Testing

### DIFF
--- a/crates/burn-std/src/ops.rs
+++ b/crates/burn-std/src/ops.rs
@@ -203,12 +203,17 @@ impl UnfoldOptions {
     }
 }
 
-/// Algorithm used for upsampling.
+/// Algorithm used.
 #[derive(new, Debug, Clone, serde::Deserialize, serde::Serialize)]
 pub enum InterpolateMode {
-    /// Nearest-neighbor interpolation.
+    /// Nearest-neighbor floor interpolation.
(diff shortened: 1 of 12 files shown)


louisfd (approved):

crates/burn-std/src/ops.rs
```diff
@@ -203,13 +203,18 @@ impl UnfoldOptions {
...
 pub enum InterpolateMode {
-    /// Nearest-neighbor interpolation.
+    /// Nearest-neighbor floor interpolation.
+    /// Matches the legacy behavior of OpenCV’s INTER_NEAREST. It results in a bottom-right shift when resizing.
     /// <https://en.wikipedia.org/wiki/Nearest-neighbor_interpolation>
     Nearest,
 
+    /// Nearest-neighbor exact interpolation.
+    /// <https://en.wikipedia.org/wiki/Nearest-neighbor_interpolation>
+    NearestExact,
```
laggui, line 216: Ahhh I forgot about this different algorithm implementation.

This can yield subtle differences in the FP input, and I'm sure this was the result of many train <> inference discrepancies for vision models being exported from pytorch etc. It's nice that we actually have both 😄

PR #4983, opened 2026-05-20
nn: add LinearTernary — ternary-quantized linear layer {-1, 0, +1}

simeon-kepp:
## What

Adds `LinearTernary` and `LinearTernaryConfig` to `crates/burn-nn/src/modules/linear.rs` — a drop-in replacement for `Linear` that stores weights quantized to `{-1, 0, +1}`.

## Why

Ternary-weight models (BitNet b1.58 and successors) train directly in `{-γ, 0, +γ}` rather than quantizing post-training. At ≥50% sparsity, zero weights are dead MACs — no multiply needed. Providing the primitive in `burn-nn` lets any model author write:

```rust
let layer = LinearTernaryConfig::new(d_in, d_out).init(&device);
```

and immediately get a ternary-weight layer that works on every existing backend today. Backends can later specialize `linear()` on ternary-valued weight tensors to realize the compute savings at the hardware level.

## Design

```rust
let layer = LinearTernaryConfig::new(in_features, out_features)
    .with_bias(true)           // default true
    .with_threshold(None)      // default: mean(|w|) — BitNet b1.58 §3.1
    .init(&device);
```

Quantization at init time (weights are frozen after this):

```
w_ternary = sign(w)   if |w| > threshold
            0         otherwise
```

- `threshold=None` (default): uses mean absolute weight value as threshold (BitNet b1.58 arxiv.org/abs/2402.17764 §3.1)
- `threshold=Some(0.0)`: all weights become ±1, no zeros
- `threshold=Some(1e9)`: all weights become 0 (degenerate case, useful for tests)

Stored as `float32` — forward pass is identical to `Linear`, works on every backend today.

## Tests

Seven unit tests in the existing `linear.rs` test module:

| Test | Checks |
|------|--------|
| `ternary_weights_are_ternary` | all values in `{-1, 0, 1}` after init |
| `ternary_output_shape` | `[batch, d_output]` correct |
| `ternary_threshold_zero_no_zeros` | `threshold=0.0` → no zero weights |
| `ternary_threshold_large_all_zeros` | `threshold=1e9` → all zeros |
| `ternary_forward_matches_manual` | hand-computed 2×2 result |
| `ternary_no_bias` | `with_bias(false)` → `None` |
| `ternary_display` | display string format |

All 7 pass (`cargo test -p burn-nn ternary`).

## Reference

- BitNet b1.58: https://arxiv.org/abs/2402.17764
- Related prior art: tinygrad `TernaryLinear` (tinygrad/tinygrad#16281), candle `LinearTernary` (huggingface/candle#3554)

### DIFF
--- a/crates/burn-nn/src/modules/linear.rs
+++ b/crates/burn-nn/src/modules/linear.rs
@@ -2,7 +2,7 @@ use burn_core as burn;
 
 use burn::config::Config;
 use burn::module::Param;
-use burn::module::{Content, DisplaySettings, Initializer, Module, ModuleDisplay};
+use burn::module::{Content, DisplaySettings, Initializer, Module, ModuleDisplay, ParamId};
 use burn::tensor::module::linear;
 use burn::tensor::{Device, Tensor};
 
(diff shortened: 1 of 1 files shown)


laggui (commented):
Quantization should not require a new layer type. We can already quantize the weights with a specified `QuantScheme`, we should extend this parameterization instead.

Additionally, this implementation still keeps the weights as float, it only changes the values to be ternary. There are no memory benefits.

simeon-kepp:
@laggui Fair critique on both points — let me address them properly.

**On the layer type:** You're right. Adding a new layer type is the wrong abstraction. I missed that `QuantScheme` + `QuantValue` is exactly the right place for this.

**On memory:** The float32 storage was a mistake in this PR specifically, not in the concept. Here's what I actually wanted to demonstrate:

```rust
// QuantValue::Q2S + PackedU32 IS ternary: {-1, 0, +1} fits in 2 bits, 16 values per u32
    .with_value(QuantValue::Q2S)
    .with_store(QuantStore::PackedU32(0))
    .with_level(QuantLevel::Block(BlockSize::B16));
```

That's already in your type system. 2-bit packed: 16× smaller than float32, 4× smaller than int8. The scale factor per block becomes the BitNet γ. Real memory reduction, within your existing abstractions.

**What I should have submitted instead:**

A zero-skip optimization in the Q2S GEMV/matmul path. When a packed u32 block is all zeros (scale = 0 or all packed bits = 0), the entire block contributes nothing to the dot product — skip the multiply-accumulate entirely. At 50–75% block sparsity (common in BitNet-trained models), this gives 1.5–2× throughput on top of the memory savings.

Would a PR targeting:
1. Docs/example showing `QuantValue::Q2S` as the right way to represent ternary weights via `Linear::quantize()`
2. A follow-up zero-skip specialization in the Q2S matmul path

...be the right shape for Burn? Happy to revise.

PR #4986, opened 2026-05-21
fix(sgd): correct SgdConfig::init doc and use impl Optimizer return type

### LINKED ISSUE #868: `SgdConfig` has bad documentation and a bad signature for `init`
Currently the documentation for the `init` method for `SgdConfig` states:

> Creates a new [SgdConfig](https://docs.rs/burn/latest/burn/optim/struct.SgdConfig.html) with default values.

This is wrong, as `init` methods of configs should create the thing they are configurating. Indeed this is what `init` for `SgdConfig` does.

Additionally the return type for the method is `OptimizerAdaptor<Sgd<...>, M, B>`, even though this may change in the future if `Sgd` were to not need an adaptor, or if optimizer adaptors were ever changed or scrapped. shouldn't ideally the return type be an `impl Optimizer`?

jaweed3:
Fixes SgdConfig::init:
- Correct doc: init creates optimizer, not config
- Use impl Optimizer<M> return type instead of concrete OptimizerAdaptor

Closes #868.

### DIFF
--- a/crates/burn-optim/src/optim/sgd.rs
+++ b/crates/burn-optim/src/optim/sgd.rs
@@ -48,7 +48,7 @@ impl SgdConfig {
         }
     }
 
-    /// Creates a new [SgdConfig](SgdConfig) with default values.
+    /// Initializes the SGD optimizer from the configuration.
     pub fn init<M: AutodiffModule>(&self) -> OptimizerAdaptor<Sgd, M> {
         let mut optim = OptimizerAdaptor::from(self.build());
         if let Some(config) = &self.gradient_clipping {


laggui (changes requested):
I'd keep the `OptimizerAdaptor` return type for now, it breaks the uniformity across the other optimizers. We probably need to revisit this as a whole instead of just changing the return type for one optimizer init.

The doc changes LGTM though 👍

jaweed3:
reverted the return type change, kept only the doc fix for now. thanks for the review!

laggui (approved):
Thanks!

PR #4987, opened 2026-05-21
fix(train): guard TUI metric navigation against empty state

### LINKED ISSUE #4964: TUI panics on arrow-key navigation before first metric is recorded
**Describe the bug**

The `burn-train` TUI metrics dashboard panics with `attempt to calculate the
remainder with a divisor of zero` if the **Left or Right arrow key** is pressed
*before any numeric metric has been recorded* — i.e. during the startup /
backend-autotune window, before the first training iteration produces metrics.

The panic kills the TUI rendering thread. Training keeps running on its own
thread, but the live dashboard is permanently dead and every subsequent metric
event logs `WARN burn_train::renderer::tui::renderer: Failed to send TUI event:
sending on a closed channel` (≈3 lines per iteration) for the rest of the run,
flooding the log.

Root cause is in `crates/burn-train/src/renderer/tui/
```rust
fn next_metric(&mut self) {
    self.selected = (self.selected + 1) % self.data.len();   // % 0  -> panic
}

fn previous_metric(&mut self) {
    if self.selected > 0 {
        self.selected -= 1;
    } else {
        self.selected = self.data.len() - 1;                 // 0usize - 1 -> panic
    }
}
```

`on_event` dispatches `KeyCode::Right -> next_metric()` and
`KeyCode::Left -> previous_metric()` with no empty-state guard, even though
`view()` already special-cases the empty state by returning
`NumericMetricView::None` when `self.names.is_empty()`. Before the first
`push()`, `self.data` (and `self.names`) is empty, so:

- `next_metric` does `% 0` → panic (`remainder with a divisor of zero`).
- `previous_metric` does `0usize - 1` → panic (subtract overflow / underflow).

**To Reproduce**

1. Run any training that uses the default renderer with the `tui` feature on a
   backend that autotunes at startup (e.g. `wgpu`), so there is a multi-second
   gap before the first metrics render — for example the
   [basic-workflow guide](https://burn.dev/books/burn/basic-workflow/) / MNIST
   example.
2. While the initial autotuning is still running (no metric/plot has appeared on
   the dashboard yet), press the **Right arrow** (or **Left arrow**) key.
3. The process panics:
   `panicked at .../burn-train-0.21.0/src/renderer/tui/metric_numeric.rs:163:25:
   attempt to calculate the remainder with a divisor of zero`.
4. The TUI thread dies; the remainder of the run logs
   `Failed to send TUI event: sending on a closed channel`.

(On Windows the navigation handler fires on key *release* rather than press;
same outcome.)

**Expected behavior**

Arrow-key navigation before any metric has been recorded should be a harmless
no-op, consistent with `view()` returning `NumericMetricView::None` when there
are no metrics. The TUI must never panic as a result of navigation input,
regardless of timing.

**Screenshots**

Not applicable (CLI). Log excerpt:

```
2026-05-16T16:41:47.382739Z ERROR burn_train::learner::application_logger: PANIC => panicked at C:\Users\…\burn-train-0.21.0\src\renderer\tui\metric_numeric.rs:163:25:
attempt to calculate the remainder with a divisor of zero
2026-05-16T16:41:48.876327Z  WARN burn_train::renderer::tui::renderer: Failed to send TUI event: sending on a closed channel
… (repeats ~3×/iteration for the rest of training)
```

- OS: Windows 11 Home 26200
- Browser: N/A (terminal CLI application)
- Version: `burn` / `burn-train` `0.21.0` (latest published; bug also present on
  `main`). Backend: `wgpu` (Vulkan, NVIDIA GeForce RTX 4090 Laptop GPU). Rust
  stable, edition 2024.

SAY-5:
### Checklist

### Related Issues/PRs

Fixes #4964.

### Changes

The TUI dashboard's Left/Right arrow handlers call `next_metric` and `previous_metric` in `renderer/tui/metric_numeric.rs`, which do `(selected + 1) % data.len()` and `data.len() - 1` unconditionally. Before the first metric is recorded (e.g. during backend autotune at startup), `data` is empty, so the first arrow press panics:

- Right arrow: `attempt to calculate the remainder with a divisor of zero` at `metric_numeric.rs:163`.
- Left arrow: `attempt to subtract with overflow` (`0usize - 1`).

The panic kills the TUI render thread; the remainder of the run floods the log with `Failed to send TUI event: sending on a closed channel`.

Return early when `data` is empty in both methods. This matches the empty state that `view()` already handles by returning `NumericMetricView::None`, and is a no-op for any non-empty state.

### Testing

- New unit test `metric_navigation_on_empty_state_is_a_no_op` in `renderer/tui/metric_numeric.rs` exercises both methods on the default (empty) state and asserts they do not panic. The test fails on `main` with the exact panic from the issue at `metric_numeric.rs:163` and passes with the fix.
- `cargo test -p burn-train --lib` — all 7 TUI-renderer tests pass (the new one plus the 6 existing).
- `cargo clippy -p burn-train --lib --tests --no-deps -- -D warnings` — clean.
- `cargo fmt --check -p burn-train` — clean.

### DIFF
--- a/crates/burn-train/src/renderer/tui/metric_numeric.rs
+++ b/crates/burn-train/src/renderer/tui/metric_numeric.rs
@@ -160,20 +160,22 @@ impl NumericMetricsState {
     }
 
     fn next_metric(&mut self) {
-        self.selected = (self.selected + 1) % {
-            let this = &self;
-            this.data.len()
-        };
+        let len = self.data.len();
+        if len == 0 {
+            return;
+        }
+        self.selected = (self.selected + 1) % len;
     }
 
     fn previous_metric(&mut self) {
+        let len = self.data.len();
+        if len == 0 {
+            return;
+        }
         if self.selected > 0 {
             self.selected -= 1;
         } else {
-            self.selected = ({
-                let this = &self;
-                this.data.len()
-            }) - 1;
+            self.selected = len - 1;
         }
     }
 
@@ -322,5 +324,24 @@ impl NumericMetricView<'_> {
             }
             Self::None => {}
         };
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn metric_navigation_on_empty_state_is_a_no_op() {
+        // Pressing Right or Left before any metric has been recorded must not
+        // panic. Previously this triggered `(0 + 1) % 0` and `0usize - 1` on
+        // the empty `data` map.
+        let mut state = NumericMetricsState::default();
+
+        state.next_metric();
+        state.previous_metric();
+
+        assert_eq!(state.selected, 0);
+        assert!(state.data.is_empty());
     }
 }


laggui (approved):
Thanks for fixing this

PR #4994, opened 2026-05-21
feat(dispatch): add remote backend

nathanielsimard:
Bring `burn-remote` to the new `Dispatch` backend and opaque `Device`:

  - Add `DispatchDevice::Remote(RemoteDevice)` and `DispatchTensorKind::Remote(BackendTensor<Remote>)`,
    gated by `feature = "remote"` on `burn-dispatch` and `burn-tensor`. `Remote` slots into the dispatch
    macros like any other backend; ops route through `BackendRouter` → `RouterTensor` →
    `OperationIr` over the existing WebSocket protocol.
  - Introduce an `inst!` helper macro so `Remote` (non-generic) coexists with the still-`<f32>`-generic
    backends in the dispatch macros, and align `NdArray`'s `cfg` in `backend_matrix!` with its variant
    `cfg` (a pre-existing asymmetry that broke `--features remote` builds).
  - Wire `Autodiff<Remote>` through all autodiff arms in `backend.rs`. Autodiff stays client-side: the
    server sees a flat op stream and doesn't know whether it's a forward or backward pass.
  - Expose `Device::remote(&str)` on the high-level opaque `Device`. Server side intentionally remains
    generic over `B: BackendIr` so user-defined backends can act as remote targets.
  - Cross-device transfers fall out of `backend_matrix!`: Local↔Remote uses the existing `TensorData`
    round-trip body; Remote↔Remote uses the optimized direct server→server transfer via `data_service`.

  ### Error handling

  `ProtocolClient::connect` now returns `Result<Channel, Error>` instead of `Option<Channel>` so the
  underlying cause flows through to callers. Connection failures in `burn-remote` worker and
  `data_service` panic with `Failed to open remote '<role>' channel to <address>: <error>. Is a
  burn-remote server running at that address?` rather than the previous `"Server to be accessible"`.
  The `parse_ws_address` helper is also lenient now — `Device::remote("localhost:3000")` auto-prepends
  `ws://`.

### Linked Issues

#4934


laggui (approved):
LGTM! Just need to fix cargo fmt

PR #4995, opened 2026-05-22
Scroll TUI metric tabs to keep selected visible

LucaCappelletti94:
This is a practical and aesthetic fix that makes the tabs in the TUI scrollable, which is important when you have many metrics and you do not understand anymore which is the metric currently being displayed.

Before:

Screenshot 2026-05-22 alle 10 07 30

After:

Note the chevrons on the sides suggesting that there is more stuff hidden in there.

Screenshot 2026-05-22 alle 10 08 05
Screenshot 2026-05-22 alle 10 08 14

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1223,6 +1223,7 @@ dependencies = [
  "tracing-appender",
  "tracing-core",
  "tracing-subscriber",
+ "unicode-width",
 ]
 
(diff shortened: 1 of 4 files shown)


laggui (approved):
Nice! Thanks for giving a bit of attention to the training TUI 😄

PR #4998, opened 2026-05-22
Add mouse support to TUI metric navigation

LucaCappelletti94:
The TUI was keyboard-only with no on-screen hint otherwise. Hovering a metric title (top tab strip or left-pane "Metrics" list) now underlines it, and clicking switches the chart, mirroring Left/Right. Reusing the existing selected-tab `UNDERLINED` modifier keeps the meaning consistent: underlined = selectable.

The `‹` `›` chevrons added by #4995 were visual-only. Clicking them now cycles to the previous/next metric (same as the arrow keys), and hovering lightens the foreground from `DarkGray` to `Gray`.

Mouse capture is torn down on every exit path including the panic hook, otherwise the shell keeps eating mouse-protocol bytes after a crash.

### DIFF
--- a/crates/burn-train/src/renderer/tui/metric_text.rs
+++ b/crates/burn-train/src/renderer/tui/metric_text.rs
@@ -4,17 +4,21 @@ use crate::{
     renderer::tui::{TuiGroup, TuiSplit},
 };
 use ratatui::{
-    prelude::{Alignment, Rect},
+    crossterm::event::{Event, MouseButton, MouseEventKind},
+    prelude::{Alignment, Position, Rect},
     style::{Color, Style, Stylize},
     text::{Line, Span},
     widgets::{Block, Borders, Paragraph, Wrap},
 };
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, ops::Range, sync::Arc};
 
 #[derive(Default)]
 pub(crate) struct TextMetricsState {
     data: BTreeMap<String, MetricGroup>,
     names: Vec<MetricName>,
+    hovered: Option<MetricName>,
+    last_pane_rect: Option<Rect>,
+    last_header_rows: Vec<(MetricName, Range<u16>)>,
 }
 
 struct MetricGroup {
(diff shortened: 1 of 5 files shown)


crates/burn-train/src/renderer/tui/metric_text.rs
```diff
@@ -4,17 +4,21 @@ use crate::{
...
-use std::{collections::BTreeMap, sync::Arc};
+use std::{collections::BTreeMap, ops::Range, sync::Arc};
 
 #[derive(Default)]
 pub(crate) struct TextMetricsState {
     data: BTreeMap<String, MetricGroup>,
     names: Vec<MetricName>,
+    hovered: Option<MetricName>,
+    last_pane_rect: Option<Rect>,
+    last_header_rows: Vec<(MetricName, Range<u16>)>,
```
laggui: We could probably use a pattern similar to `NumericMetricsState` with a `TextHitState` here? e.g.

```rust
#[derive(Default)]
struct TextHitState {
    hovered: Option<MetricName>,
    rect: Option<Rect>,
    header_rows: Vec<(MetricName, Range<u16>)>,
}

#[derive(Default)]
pub(crate) struct TextMetricsState {
    data: BTreeMap<String, MetricGroup>,
    names: Vec<MetricName>,
    pane: TextHitState,
}
```

crates/burn-train/src/renderer/tui/renderer.rs
```diff
@@ -460,6 +463,16 @@ impl TuiMetricsRenderer {
...
                     }
 
+                    Ok(event @ Event::Mouse(_)) => {
+                        if self.popup.is_empty() {
+                            self.metrics_numeric.on_event(&event);
+                            if let Some(clicked) = self.metrics_text.on_event(&event) {
+                                self.metrics_numeric.select_by_name(&clicked);
+                            }
+                        }
+                        self.draw().ok();
```
laggui: We should probably check if the mouse event captured requires a redraw or not. Otherwise any mouse event (esp. move) will fire continuously and redraw unnecessarily.

laggui (commented):
LGTM! Just some minor comments

crates/burn-train/src/renderer/tui/renderer.rs
LucaCappelletti94: Ok, I will double check, altough this is how I do it in other TUIs I made.

crates/burn-train/src/renderer/tui/metric_text.rs
LucaCappelletti94: Done!

laggui (approved):

PR #5001, opened 2026-05-22
Refactor/wgpu backends

nathanielsimard:
Requires: https://github.com/tracel-ai/cubecl/pull/1346

Fixes the conflicts between backends, keeps the same behavior with the wgpu backend. Essentially, you can now actually benchmark the difference between webgpu and metal with the same binary, which wasn't possible before. The wgpu backend will still leverage the "best" compiler for the current device based on the activated feature flag, so that workflow still works.


laggui (approved):
Nice! So now users can choose between `Wgpu` (auto) or any explicit specializations (`Vulkan`, `Metal`, `WebGpu`).

There will be a conflict with the changes in #5000 which was merged. Need to re-introduce the explicit variants in dispatch since the backend type aliases are now distinct on the compiler (but they no longer require the default element types!).

PR #5002, opened 2026-05-22
Fix ndarray remainder broadcasting

### LINKED ISSUE #4712: ndarray: remainder does not support broadcasting
## Description

`NdArrayMathOps::remainder` uses `Zip::from(&mut out).and(&rhs)` without broadcasting, causing a panic when operands have the same rank but different dimension sizes. CubeCL backends are not affected since `launch_binop` computes `broadcast_shape` for all binary ops including remainder.

```rust
// crates/burn-ndarray/src/ops/base.rs
pub fn remainder(lhs: SharedArray<E>, rhs: SharedArray<E>) -> SharedArray<E> {
    let mut out = lhs.into_owned();
    Zip::from(&mut out).and(&rhs).for_each(|out_elem, &b| {
        // ...
    });
    out.into_shared()
}
```

### Reproduction

```rust
use burn::prelude::*;
use burn_ndarray::NdArray;

type B = NdArray;

fn main() {
    let a = Tensor::<B, 2, Int>::from_ints([[10, 20, 30]], &device); // [1, 3]
    let b = Tensor::<B, 2, Int>::from_ints([[7]], &device);          // [1, 1]
    let result = a.remainder(b); // panics: Zip dimension mismatch
}
```

```
Zip: Producer dimension mismatch, expected: IxDynImpl(Inline(2, [1, 3, 0, 0])), got: IxDynImpl(Inline(2, [1, 1, 0, 0]))
```

### Context

This was hit in practice with YOLO v10/v26x models via burn-onnx (tracel-ai/burn-onnx#281). The ONNX `Mod` operator supports numpy-style broadcasting, so operands like `[1, 300]` and `[1, 1]` are valid.

Other ndarray binary ops (`add`, `mul`, `sub`, `div`) work because they use ndarray's native operators (`+`, `*`, `-`, `/`) which handle broadcasting. `remainder` uses `Zip` directly instead.

### Suggested fix

Broadcast both operands before zipping, similar to how `elementwise_op` or `broadcast_for_comparison` handle it:

```rust
pub fn remainder(lhs: SharedArray<E>, rhs: SharedArray<E>) -> SharedArray<E> {
    let broadcast_dim = /* compute max shape */;
    let lhs = lhs.broadcast(broadcast_dim.clone()).unwrap();
    let rhs = rhs.broadcast(broadcast_dim).unwrap();
    Zip::from(&lhs).and(&rhs).map_collect(|&a, &b| {
        let a_f = a.to_f64();
        let b_f = b.to_f64();
        let r = a_f - b_f * (a_f / b_f).floor();
        r.elem()
    }).into_shared()
}
```

puneetdixit200:
## Summary

- broadcast ndarray remainder operands to their common shape before zipping
- reuse the shared binary broadcast helper for remainder and comparison-style operations
- add an ndarray backend regression test for `[1, 3]` remainder `[1, 1]`

Fixes #4712

## Tests

- `cargo test -p burn-backend-tests --features ndarray --test tensor remainder_broadcast -- --nocapture`
- `cargo test -p burn-backend-tests --features ndarray --test tensor remainder -- --nocapture`
- `cargo fmt --all --check`
- `cargo clippy -p burn-ndarray -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/int/ops/remainder.rs
+++ b/crates/burn-backend-tests/tests/tensor/int/ops/remainder.rs
@@ -15,6 +15,18 @@ fn should_support_int_remainder_basic() {
 }
 
 #[test]
+fn should_support_int_remainder_broadcast() {
+    let device = Default::default();
+    let lhs = TestTensorInt::<2>::from_data(TensorData::from([[10, 20, 30]]), &device);
+    let rhs = TestTensorInt::<2>::from_data(TensorData::from([[7]]), &device);
+
+    let output = lhs.remainder(rhs);
+    let expected = TensorData::from([[3, 6, 2]]);
+
+    output.into_data().assert_eq(&expected, false);
+}
+
+#[test]
 fn should_support_int_remainder_basic_scalar() {
     let data = TensorData::from([-3, -2, -1, 1, 2, 3]);
     let device = Default::default();
--- a/crates/burn-ndarray/src/ops/base.rs
+++ b/crates/burn-ndarray/src/ops/base.rs
@@ -673,9 +673,9 @@ macro_rules! dispatch_unary_simd {
     ($elem: ty, $op: ty, $lhs: expr, $($ty: ty),*) => {{ $lhs }};
 }
 
-// Helper function to broadcast two tensors to a common shape for comparison operations
+// Helper function to broadcast two tensors to a common shape for binary operations
 // Returns broadcasted views that can be safely zipped
-fn broadcast_for_comparison<'a, E: Copy, S1, S2>(
+fn broadcast_for_binary_ops<'a, E: Copy, S1, S2>(
     lhs: &'a ndarray::ArrayBase<S1, ndarray::IxDyn>,
     rhs: &'a ndarray::ArrayBase<S2, ndarray::IxDyn>,
 ) -> (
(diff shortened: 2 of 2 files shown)


laggui (approved):
Thanks for addressing this issue!

puneetdixit200:
The current red jobs look unrelated to the remainder change:

- `linux-std-tests (stable)` passed the test suites and was canceled during `cargo xtask coverage generate` with `The runner has received a shutdown signal` / `The operation was canceled`.
- `macos-std-tests (stable)` starts failing in `burn-backend-tests --test autodiff` after a WGPU/DSD shader module validation error (`Device::create_shader_module_passthrough`, labels like `unary_float_f_f32_n_2` / `matmul_entry...`). The remaining autodiff failures look like a cascade after that backend/device error.

I tried to rerun the failed jobs, but GitHub rejected it with `Must have admin rights to Repository`. Could a maintainer rerun the failed jobs when convenient?

laggui:
Totally unrelated, that's why I merged it. The macos CI runners have been unable to run for at least a week :sweat_smile: looks like they're back now but we might have to fix some issues. The linux-std-tests fail on code coverage (the runner gets killed, probably using too much memory or disk space).

PR #5006, opened 2026-05-25
burn-book: add confusion matrix custom metric example

MihirMohapatra:
## Summary
Adds a concrete custom metric example to the burn-book metric page showing how to accumulate a confusion matrix alongside standard classification metrics.

## Changes
- Added a new section to urn-book/src/building-blocks/metric.md:
  - Adapting ClassificationOutput into a custom ConfusionMatrixInput
  - Implementing ConfusionMatrixMetric with Metric
  - Registering the custom metric with LearnerBuilder alongside accuracy

## Why
Accuracy alone can hide class-specific failure modes. A confusion matrix provides immediate diagnostics for classification errors.

## Validation
- cargo run-checks was started locally but timed out in this environment before completion.
- Change is documentation-only and scoped to a single file.

## Notes
No framework/runtime behavior was modified; this PR only extends docs with a practical example.

### DIFF
--- a/burn-book/src/building-blocks/metric.md
+++ b/burn-book/src/building-blocks/metric.md
@@ -223,3 +223,84 @@ impl<B: Backend> Numeric for LossMetric<B> {
     }
 }
 ```
+
+## Example: Confusion Matrix Metric for Classification
+
+For classification, you can keep `AccuracyMetric` and add a custom confusion matrix metric to inspect
+which classes are commonly misclassified.
+
+The metric input can be adapted from `ClassificationOutput<B>`:
+
+```rust,ignore
+#[derive(new)]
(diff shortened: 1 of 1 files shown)


laggui (commented):
Thanks for the contribution! I think the motivation here is valid, but I'm not sure the right solution is to add this example to the docs.

The metrics page already demonstrates the important customization points:
- defining a custom metric input
- adapting outputs with Adaptor
- implementing Metric
- maintaining metric state

Burn already has `ConfusionStats` that powers metrics like precision and recall. I think a better direction would be to add a built-in confusion matrix metric with the proper state.

laggui (commented):
Closing due to inactivity. Feel free to re-open if you want to pick this up again.

PR #5009, opened 2026-05-25
Update dqn example + FIX Dispatch's from_inner

Charles23R:
### Checklist

### Changes

- Fixed a bug in `Dispatch`'s `from_inner` where the checkpointing strategy was set to `None`, which is not possible for autodiff tensors.
- Updated the `dqn-agent` example to reflect recent changes in burn.
- Utilize the dispatch device changes to enable training/inference on two different devices for RL. For example, you can now train on a GPU while doing rollout collection and validation on the CPU. The `inference_device` can be set using the `RLTraining` paradigm.

### Testing

Test the `dqn-agent` example with tch, wgpu, flex, cuda.


crates/burn-dispatch/src/backend.rs
```diff
@@ -549,6 +463,12 @@ impl AutodiffBackend for Dispatch {
                 panic!("Autodiff should not wrap an autodiff tensor.")
             }
         };
+
+        let checkpointing = if let Some(strategy) = checkpointing {
+            Some(strategy)
+        } else {
+            Some(crate::CheckpointingStrategy::None)
+        };
```
laggui, line 557: Ahhhh I see the issue, the checkpointing was likely `None` because the tensor originated entirely from the inner backend.

Technically we'd have to make sure the checkpointing matches the autodiff device on downstream ops but I already see an issue in the existing code. No need to address in this PR, this patch is sufficient.

examples/dqn-agent/cubecl.toml
laggui, line 1: Now that we have introduced the `burn.toml` config we should prefer using this in the examples. You can check the `text-classification` example for reference, but essentially it just requires the sections to be prefixed with `[cubecl.*]`.

laggui (approved):

PR #5012, opened 2026-05-27
Refactor: Transform all Progress items into a global progress struct

Zatiji:
### Checklist

### Related Issues/PRs

No related issues

### Changes

**Refactors progress tracking across the training pipeline to use a unified OverallProgress type.**

Previously, the renderer interface exposed TrainingProgress, EvaluationProgress, and ProgressType as separate ad-hoc structs, each carrying slightly different representations of the same underlying data. This made it hard to pass consistent progress information across supervised and RL training, and forced renderers to implement render_train/render_valid/render_test methods on top of the existing TrainingProgressLogger/EvaluationProgressLogger traits.

**Changes made:**

- Added unit: String to `Progress` (burn-core) to describe what is being measured (e.g. "items", "epochs").
- Introduced `OverallProgress` in `burn-train/logger/progress.rs`, bundling a `global_progress` (full training run) and a `split_progress` (current phase within a run).
- Replaced `update_split(items_processed: usize)` with `update_split(progress: &OverallProgress)` in TrainingProgressLogger, giving loggers access to both progress levels.
- Added `log_event_training(event: String)` to TrainingProgressLogger for custom counter events (e.g. RL envStep count).
- Removed `TrainingProgress`, `EvaluationProgress`, `ProgressType` and the` render_train`/`render_valid`/`render_test` renderer methods. MetricsRenderer now directly requires `TrainingProgressLogger` + `EvaluationProgressLogger`, eliminating the redundancy.
- Updated the TUI renderer, CLI renderer, and RL processor to use the new types.

### Testing

Tested by running the MNIST example with a custom `FileProgressLogger` that writes progress events to disk and confirmed that both training and validation splits report correct `split_progress` and `global_progress` values across epochs.


crates/burn-core/src/data/dataloader/base.rs
```diff
@@ -12,6 +12,9 @@ pub struct Progress {
 
     /// The total number of items that need to be processed.
     pub items_total: usize,
+
+    /// The unit of measurement for the progress (e.g., "items", "batches").
+    pub unit: String,
```
Charles23R: As it was not required before, this could be an `Option`

crates/burn-train/src/logger/progress.rs
```diff
@@ -33,13 +35,16 @@ pub trait TrainingProgressLogger: Send {
...
-    fn update_split(&mut self, items_processed: usize);
+    fn update_split(&mut self, progress: &OverallProgress);
 
     /// Called at the end of a training split.
     fn end_split(&mut self);
 
     /// Called at the end of training, whether it completed successfully or was interrupted.
     fn end(&mut self);
+
+    /// Log a custom counter event, such as the number of iterations accomplished (or more).
```
Charles23R: I don't think the doc is up to date with what the function does

crates/burn-train/src/logger/progress.rs
```diff
@@ -49,29 +54,54 @@ pub trait TrainingProgressLogger: Send {
...
+    fn update_test_progress(&mut self, progress: &OverallProgress);
 
     /// Called at the end of a test split.
     fn end_test(&mut self);
 
     /// Called at the end of evaluation.
-    fn end(&mut self);
+    fn end_global_progress(&mut self);
+
+    /// Log a custom counter event, such as the number of iterations accomplished (or more).
```
Charles23R: See other comment

crates/burn-train/src/logger/progress.rs
```diff
@@ -33,13 +35,16 @@ pub trait TrainingProgressLogger: Send {
     fn start_split(&mut self, split: &str, total_items: usize);
 
     /// Log the progress of the current training step.
-    fn update_split(&mut self, items_processed: usize);
+    fn update_split(&mut self, progress: &OverallProgress);
```
Charles23R: We talked about this offline, but we should go back to items_processed

crates/burn-train/src/metric/processor/metrics.rs
```diff
@@ -225,40 +224,11 @@ impl<T: ItemLazy, V: ItemLazy> MetricsTraining<T, V> {
...
-        }
-    }
-}
-
 impl<T> From<&TrainingItem<T>> for MetricMetadata {
     fn from(item: &TrainingItem<T>) -> Self {
         Self {
             progress: item.progress.clone(),
-            global_progress: item.global_progress.clone(),
+            global_progress: item.progress.clone(),
```
Charles23R: We can remove `global_progress` from `MetricMetadata` altogether

crates/burn-train/src/renderer/tui/status.rs
```diff
@@ -23,78 +26,126 @@ enum Mode {
...
     }
+    /// Update counters from a progress event.
+    pub(crate) fn update_counter(&mut self, event: String) {
+        *self.event_counters.entry(event).or_insert(0) += 1;
+    }
+
+    /// Reset per-split counters at the end of a split.
+    pub(crate) fn reset_counters(&mut self) {
+        for (key, val) in self.event_counters.iter_mut() {
+            if key != "TrainStep" && key != "EnvStep" {
```
Charles23R: Better not have domain rules in the tui, let's just remove that if condition and just reset to 0.

crates/burn-train/src/renderer/tui/status.rs
```diff
@@ -23,78 +26,126 @@ enum Mode {
...
+}
+
 impl StatusView {
-    fn new(progress_indicators: &[ProgressType], mode: &Mode) -> Self {
+    fn new(
+        progress: Option<&OverallProgress>,
+        mode: &Mode,
+        event_counters: &BTreeMap<String, usize>,
+    ) -> Self {
+        let is_rl =
```
Charles23R: Once again, no domain rules in the renderer. Let's just remove `is_rl` and let it have the default behaviour for now. Same with the "Episode/EpisodeEnd" stuff.

crates/burn/Cargo.toml
```diff
@@ -23,6 +23,7 @@ workspace = true
 
 [features]
 default = [
+    # "rl",
```
Charles23R: testing artifact

Cargo.toml
```diff
@@ -15,7 +15,7 @@ members = [
 exclude = [
     "examples/notebook",
     "examples/raspberry-pi-pico",
-    "examples/dqn-agent",         # gym-rs
+    # "examples/dqn-agent",         # gym-rs
```
Charles23R: testing artifact

crates/burn-train/src/metric/processor/full.rs
```diff
@@ -15,6 +15,8 @@ pub struct FullEventProcessorTraining<T: ItemLazy, V: ItemLazy> {
     renderer: Box<dyn MetricsRenderer>,
     store: Arc<EventStoreClient>,
     progress_logger: Option<Box<dyn TrainingProgressLogger>>,
+    current_epoch: usize,
+    total_epochs: usize,
```
Charles23R, line 18: I think these aren't used anymore

Charles23R: ah, nevermind i didn't see

Charles23R (approved):
LGTM! Just want to add, this lays the ground for even more simplifications that will offload a lot of processing from the training loop. It will be even simpler for users to create their own custom training loops.

crates/burn-core/src/data/dataloader/base.rs
Zatiji: Should I do a constructor that only takes two parameters (to not break the contract for components that use the Progress object outside of Burn)? Or is only putting the unit attribute optional enough?

crates/burn-train/src/logger/progress.rs
```diff
@@ -49,29 +57,57 @@ pub trait TrainingProgressLogger: Send {
...
+///
+/// `global_progress` spans the full training run (e.g., epochs completed out of total),
+/// while `split_progress` tracks the current phase (e.g., batches within the current epoch).
+#[derive(Debug, Clone)]
+pub struct OverallProgress {
+    /// Progress across the entire training run.
+    pub global_progress: Progress,
+    /// Progress within the current phase (epoch or evaluation split).
+    pub split_progress: Progress,
+}
```
laggui, line 103: Just a small suggestion. I think your doc points to a better name for this new struct: `ProgressSnapshot`.

We could probably drop the `_progress` suffix from both fields

```rust
#[derive(Debug, Clone)]
pub struct ProgressSnapshot {
    /// Progress across the entire training run.
    pub global: Progress,
    /// Progress within the current phase (epoch or evaluation split).
    pub split: Progress, 
}
```

crates/burn-core/src/data/dataloader/base.rs
Charles23R: Let's just put it optional. I'm thinking most use cases will still have a unit anyway.

PR #5019, opened 2026-05-28
perf(flex): bulk-copy inner-contiguous run in to_contiguous

jkaczman:
### Checklist

### Related Issues/PRs

Not applicable.

### Changes

I use a ShuffleNetV2 model with burn-flex, and noticed that the reshape operation was taking an unusually long time compared to ONNX Runtime. After inspecting and adding performance tests, I noticed that it fell into the fallback un-optimized path for contiguous copy. The reason it needs a contiguous copy is because shuffle channels do reshape -> permute -> reshape, and as such, requires a contiguous copy on that second reshape. 

When inspecting the strides that it checks against, I noticed that the last (innermost) stride was 1. That means the innermost dim is laid out contiguously in the source memory. When an innermost dim is contiguous like this, that means we can bulk-copy / vectorize that copy instead of walking the entire tensor element-by-element (like the fallback does).

Thus, I added a branch in the if-else that checks this in contiguous_copy, that if the last stride is 1 and all strides are positive (and if dimensions are greater than 2 due to the earlier branches), that it will use a bulk-copy instead of the slow fallback path. 

On the fallback path, it was running before at around **16 GB/s**. The new one hits **150 GB/s** in this case against a **170 GB/s** memcpy ceiling on my machine. On a Raspberry Pi Zero 2W running with my model, the speed increases for inference on a 416x416 image from ~2000ms to ~1500ms per run (25% faster). It is now within 12% of optimized ORT (and I will follow up with another PR soon that should fully close that gap after I'm finished investigating).

I did not touch any of the existing 1D/2D paths. This only affects 3+ dimensions in scope that would've normally fallen into that fallback case.

### Testing

Added a test "test_to_contiguous_3d_inner_stride1_matches_naive", which checks to see that the sped-up version that now runs within this inner-contiguous case is semantically correct and matches the naive solution. This was derived based on my original case of noticing burn-flex not being performant based on falling into the fallback path (the ShuffleNet channel shuffle). All tests in burn-flex (including the new one) pass.

### DIFF
--- a/crates/burn-flex/src/tensor.rs
+++ b/crates/burn-flex/src/tensor.rs
@@ -377,6 +377,11 @@ impl FlexTensor {
             copy_2d_tiled(
                 &mut dst, src, offset, shape[0], shape[1], strides[0], strides[1],
             );
+        } else if all_positive && strides.last().copied() == Some(1) {
+            // Since the inner stride is 1, we can bulk-copy
+            // the contiguous inner run for each outer index.
+            unsafe { dst.set_len(n) };
+            copy_inner_contiguous_run(&mut dst, src, offset, shape, strides);
         } else {
             // General fallback: covers negative strides (flipped
             // tensors) and ND layouts that can't collapse to ≤2D.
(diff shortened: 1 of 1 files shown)


laggui (approved):
LGTM! Thanks for adding an isolated test case as well

PR #5020, opened 2026-05-28
Fallback implementation for quantized ops (no native kernel)

### LINKED ISSUE #5005: burn-cubecl: q_slice, q_gather, q_select, q_expand are unimplemented!() on CUDA — supports_dtype(QFloat) advertises false support
**Describe the bug**

`burn-cubecl 0.21` panics at `crates/burn-cubecl/src/ops/qtensor.rs:274` (`q_slice`) on the CUDA backend, with `not implemented`. The same file leaves `q_gather`, `q_select`, and `q_expand` as `unimplemented!()` too. `q_matmul` *is* wired through to native PTX kernels via cubecl, so isolated quantized matmul works — but any model that does anything else with QFloat tensors (slicing for KV cache, gather/select for tokenization paths, expand for broadcasts) crashes mid-forward.

**To Reproduce**
Hit while running a Whisper-tiny INT8 model on CUDA via WhisperForge:

```
wforge transcribe -a audio.wav -m tiny_en_int8 --device cuda
```

Stack:
```
thread 'DSU-0-0' panicked at .../burn-cubecl-0.21.0/src/ops/qtensor.rs:274:9:
not implemented
```

Full context and trace: https://github.com/bevsxyz/WhisperForge/issues/1

## Impact

`burn_cuda::Cuda::supports_dtype(device, DType::QFloat(<default_scheme>))` returns `true`, so downstream crates reasonably assume the backend can run quantized inference end-to-end. With the surrounding tensor-manipulation ops as stubs, any real model (anything past pure-matmul microbenchmarks) crashes. This makes `supports_dtype(QFloat)` an unreliable signal for "can run quantized inference on this backend" — it answers tensor placement, not op coverage.

In WhisperForge we fell into this trap: gated the FP32 dequant fallback on `supports_dtype(QFloat)`, tested on real CUDA hardware, panic on first slice. Reverted; INT8 stays a 4× on-disk saving with FP32 inference everywhere.

## Suggested resolution

Mirror #4847's framing: either implement the missing q-ops, or have `supports_dtype` return `false` for `QFloat` on backends where the op set is incomplete so downstream crates can make the right routing decision. A reasonable default impl (`dequantize → run the float op → requantize`) would unblock `q_slice`/`q_gather`/`q_select`/`q_expand` without native kernels.

## Related

- #4847 — `tch: quantization ops not implemented` (same shape, different backend)
- #464 — 4/8-bit inference umbrella enhancement
- #4659 — `q_reshape` data loss on WGPU (adjacent: incomplete QFloat op coverage)

Andy2887:
This is a draft PR and the changes are still in progress.

### Checklist

### Related Issues/PRs
Fix #5005

### Changes
Implements a fallback mechanism for `q_slice`, `q_gather`, `q_select`, and `q_expand` to prevent runtime panics when executing quantized inference. 

Since a native quantized slice kernel is not yet available, this PR introduces a robust fallback pattern:
1. **Dequantize**
2. Execute the native **`float_slice`** operation.
3. **Requantize**

### Testing
`crates/burn-backend-tests/tests/cubecl/quantization/slice.rs`: validates the correctness of `q_slice`.

### DIFF
--- a/crates/burn-autodiff/src/ops/qtensor.rs
+++ b/crates/burn-autodiff/src/ops/qtensor.rs
@@ -67,38 +67,11 @@ impl<B: Backend, C: CheckpointStrategy> QTensorOps<Self> for Autodiff<B, C> {
         unimplemented!()
     }
 
-    fn q_gather(
-        _dim: usize,
-        _tensor: QuantizedTensor<Self>,
-        _indices: IntTensor<Self>,
-    ) -> QuantizedTensor<Self> {
-        unimplemented!()
-    }
-
-    fn q_select(
-        _tensor: QuantizedTensor<Self>,
-        _dim: usize,
-        _indices: IntTensor<Self>,
-    ) -> QuantizedTensor<Self> {
-        unimplemented!()
-    }
-
-    fn q_slice(
-        _tensor: QuantizedTensor<Self>,
-        _slices: &[burn_std::Slice],
-    ) -> QuantizedTensor<Self> {
-        unimplemented!()
-    }
-
     fn q_argmax(tensor: QuantizedTensor<Self>, dim: usize, out_dtype: IntDType) -> IntTensor<Self> {
         B::q_argmax(tensor, dim, out_dtype)
     }
 
     fn q_argmin(tensor: QuantizedTensor<Self>, dim: usize, out_dtype: IntDType) -> IntTensor<Self> {
         B::q_argmin(tensor, dim, out_dtype)
-    }
-
-    fn q_expand(_tensor: QuantizedTensor<Self>, _shape: Shape) -> QuantizedTensor<Self> {
-        unimplemented!()
     }
 }
--- a/crates/burn-backend-tests/tests/cubecl/quantization/mod.rs
+++ b/crates/burn-backend-tests/tests/cubecl/quantization/mod.rs
@@ -2,6 +2,7 @@ pub use super::*;
 
 mod quantize_dequantize;
 mod reshape;
+mod slice;
 
 fn supports_native() -> bool {
     let name = format!("{:?}", burn_tensor::Device::default());
--- a/crates/burn-backend-tests/tests/cubecl/quantization/slice.rs
+++ b/crates/burn-backend-tests/tests/cubecl/quantization/slice.rs
@@ -1,0 +1,124 @@
+use super::*;
+use burn_tensor::Tolerance;
+use burn_tensor::{Shape, SliceArg, quantization::QuantScheme, s};
+
+/// Slice a quantized tensor (exercising `q_slice`) and compare against slicing the same
+/// quantized data in full precision (dequantized) - within a permissive tolerance.
+fn should_quantize_slice_dequantize_arange<const D: usize, S>(shape: [usize; D], slices: S)
+where
+    S: SliceArg + Clone,
+{
+    let numel = Shape::from(shape).num_elements() as i64;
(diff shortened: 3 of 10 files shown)


laggui (commented):
You could adopt a pattern similar to what we already have in the default implementations for some `QTensorOps`, so the macro can be reused to implement the different ops:

https://github.com/tracel-ai/burn/blob/1b6dd60361115a0efb5b67c43cb75cd856b30498/crates/burn-backend/src/backend/ops/qtensor.rs#L311-L321

And for the tests, we already have some defined here that need to be enabled:
https://github.com/tracel-ai/burn/blob/1b6dd60361115a0efb5b67c43cb75cd856b30498/crates/burn-backend-tests/tests/tensor/float/quantization/ops/mod.rs#L6-L8

but we might have to change some of the inputs. So it's fine if you define new tests for now.

Andy2887:
> You could adopt a pattern similar to what we already have in the default implementations for some `QTensorOps`, so the macro can be reused to implement the different ops:
> 
> https://github.com/tracel-ai/burn/blob/1b6dd60361115a0efb5b67c43cb75cd856b30498/crates/burn-backend/src/backend/ops/qtensor.rs#L311-L321
> 

Thanks for the pointer — I hadn't noticed dequant_op_flow! / dequant_op_quant! macros and QTensorOps trait already uses them for the default impls of q_add, q_mul, etc. That definitely cleans things up.
                                                                                                                                           
  Before I rewrite this, I wanted to check on where the fallback should live. I see two options:                                           
   
  1. Add the default to QTensorOps in crates/burn-backend/src/backend/ops/qtensor.rs, following the same pattern as q_add / q_gather /  q_repeat_dim, and drop the unimplemented!() stubs in crates/burn-cubecl/src/ops/qtensor.rs. **In this case, we implement a fallback implementation for the quantized tensor ops across all backends.**                                                     
  2. Add the fallback only in crates/burn-cubecl/src/ops/qtensor.rs (like what I'm doing now for q_slice). **In this case, we're only implementing fallback for CubeCL quantized tensor ops.**
                                                                                                                                           
  Do you have a preference? Happy to go either way — just want to make sure I match the project's direction.

@laggui

laggui:
Sorry for the delay!

Might as well implement it in `burn-backend`, so you can reuse the existing macro when applicable.

Andy2887:
So here's a summary of what I did:

### Implemented fallback implementations for q_slice, q_gather, q_select, and q_expand. 

Also deleted the backend specific implementations of these operations if they have "unimplemented!()" so the fallback defined in the trait is used.

Screenshot 2026-06-01 at 8 18 08 PM

### For the tests, I only added "flex". Candle, router, tch, cubecl and autodiff are not included. 

Screenshot 2026-06-01 at 8 16 08 PM

For candle, tch, router and autodiff, this is because other quantized methods are largely unimplemented for those backends. Running the tests, which include tests for every quantized operators, against any of these backends would fail. 

For CubeCL, PackedU32 requires the last dim to be a multiple of the pack factor (4 int8s per u32), which most of these test shapes violate.

### I modified `maxmin` tests to use a slightly looser tolerance.

Screenshot 2026-06-01 at 8 17 01 PM

This is because flex backend targets f16, which has lower precision than f32. When we enable "flex" in step 2, we also bring in f16 to our tests. Previously passing max/min tests failed not because of logic error, but because f16 lost more precision when quantized.

laggui (approved):

PR #5028, opened 2026-06-01
fix(derive): clearer error when generic used in both module and skip fields

### LINKED ISSUE #5007: Module derive: Confusing error message for a generic type used on both a module field and a skipped field
When a type deriving `Module` uses a generic type on a module field and a skipped field at the same time, the code doesn't compile. This makes sense and is consistent with the existing logic that disallows skipping a field that uses a generic type with the `Module` trait bound ([parse_module_field_type](https://github.com/tracel-ai/burn/blob/0034f64bf0827b1156057c7b09ba4f093a77bfa4/crates/burn-derive/src/module/codegen_struct.rs#L430-L432) in `crates/burn-derive/src/module/codegen_struct.rs`).

However, the error message is kind of confusing. For example, for the following type:
```rs
#[derive(Module, Debug)]
struct MyMod<T> {
    a: T,
    #[module(skip)]
    b: T,
}
```
the first error is one that complains about the missing `Module` trait bound:
```
error[E0277]: the trait bound `T: burn::prelude::Module` is not satisfied in `MyModRecord<T>`
  --> examples/my-demo/examples/my-demo.rs:6:8
   |
 6 | struct MyMod<T> {
   |        ^^^^^^^^ within `MyModRecord<T>`, the trait `burn::prelude::Module` is not implemented for `T`
   |
note: required because it appears within the type `MyModRecord<T>`
  --> examples/my-demo/examples/my-demo.rs:6:8
   |
 6 | struct MyMod<T> {
   |        ^^^^^
   = note: required for `<MyMod<T> as burn::prelude::Module>::Record` to implement `Record`
note: required by a bound in `burn::prelude::Module::Record`
  --> crates/burn-core/src/module/base.rs:71:18
   |
71 |     type Record: Record;
   |                  ^^^^^^ required by this bound in `Module::Record`
```

The cause is that skipping a field rewrites the kinds of all its generic types to `GenericKind::Skip`, which prevents the `Module` trait bound from being added to those generics ([GenericsParser::from_ast](https://github.com/tracel-ai/burn/blob/0034f64bf0827b1156057c7b09ba4f093a77bfa4/crates/burn-derive/src/module/codegen.rs#L188-L214) in `crates/burn-derive/src/module/codegen.rs`).

But if I add the `Module` trait bound to `T`, the existing check mentioned above will fail instead:
```
error: Fields with module generics should not be marked as 'skip'.
 --> examples/my-demo/examples/my-demo.rs:8:14
  |
8 |     #[module(skip)]
  |              ^^^^
```

This is definitely not a big issue, but probably a clearer error message would be more helpful. I've tried a fix in my local environment, which turns the message to:
```
error: Generic type should not be used on a module field and a skipped field at the same time
 --> examples/my-demo/examples/my-demo.rs:6:14
  |
6 | struct MyMod<T> {
  |              ^
```

If it's worth a fix, I can make a PR for it.

Versions:\
&nbsp;&nbsp;&nbsp;&nbsp;Burn: HEAD of main branch (0034f64bf)\
&nbsp;&nbsp;&nbsp;&nbsp;Rust: Stable 1.95.0

jaweed3:
When a generic type parameter is used in both a module field and a `#[module(skip)]` field, the derive macro now emits a clear error pointing to the generic declaration, instead of a confusing compiler error about a missing `Module` trait bound.

**Before:** Two confusing errors depending on whether `T: Module` is added:
- Without bound: cryptic `the trait bound `T: Module` is not satisfied`
- With bound: `Fields with module generics should not be marked as 'skip'`

**After:** Single clear error at the generic declaration:
```
error: Generic type `T` is used in both a module field and a skipped field. Consider removing `#[module(skip)]` or using a different type for one of the fields.
```

Fixes #5007

### DIFF
--- a/crates/burn-derive/src/module/codegen_struct.rs
+++ b/crates/burn-derive/src/module/codegen_struct.rs
@@ -397,13 +397,9 @@ pub(crate) fn parse_module_fields(
     };
 
     for ident in module_generics.intersection(&skip_generics) {
-        if let Some(param) = ast.generics.params.iter().find_map(|p| {
-            if let syn::GenericParam::Type(tp) = p {
-                if tp.ident == *ident {
-                    return Some(tp);
-                }
-            }
-            None
+        if let Some(param) = ast.generics.params.iter().find_map(|p| match p {
+            syn::GenericParam::Type(tp) if tp.ident == *ident => Some(tp),
+            _ => None,
         }) {
             return Err(syn::Error::new(
                 param.ident.span(),
--- a/crates/burn-derive/src/module/codegen_struct.rs
+++ b/crates/burn-derive/src/module/codegen_struct.rs
@@ -377,17 +377,44 @@ pub(crate) fn parse_module_fields(
     generics: &mut ModuleGenerics,
 ) -> syn::Result<Vec<ModuleField>> {
     let mut fields = Vec::new();
+    let mut module_generics = HashSet::new();
+    let mut skip_generics = HashSet::new();
 
     match &ast.data {
         syn::Data::Struct(struct_data) => {
             for field in struct_data.fields.iter() {
                 let field_type = parse_module_field_type(field, generics)?;
+                if field_type.is_module {
+                    module_generics.extend(field_type.generic_idents.iter().cloned());
+                } else if matches!(field_type.attr, Some(ModuleFieldAttribute::Skip)) {
+                    skip_generics.extend(field_type.generic_idents.iter().cloned());
+                }
                 fields.push(ModuleField::new(field.clone(), field_type));
             }
         }
         syn::Data::Enum(_) => panic!("Only struct can be derived"),
         syn::Data::Union(_) => panic!("Only struct can be derived"),
     };
+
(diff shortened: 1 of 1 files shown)


crates/burn-derive/src/module/codegen_struct.rs
```diff
@@ -377,17 +377,40 @@ pub(crate) fn parse_module_fields(
...
+
+    for ident in module_generics.intersection(&skip_generics) {
+        if let Some(param) = ast.generics.params.iter().find_map(|p| match p {
+            syn::GenericParam::Type(tp) if tp.ident == *ident => Some(tp),
+            _ => None,
+        }) {
+            return Err(syn::Error::new(
+                param.ident.span(),
+                format!(
+                    "Generic type `{ident}` is used in both a module field and a skipped field. \
```
laggui: I'd make a minor adjustment on the wording:

```suggestion
                    "Generic type `{ident}` should not be used on both a module field and a skipped field. \
```

laggui (changes requested):

laggui (approved):

PR #5029, opened 2026-06-01
fix(remote): correct multi-streams and drastically improve performance

nathanielsimard:
Fixed the correctness of the remote backend while drastically improving its performance. Now on local compute, I can get near-native performance while streaming the data onto the websocket and using it for every tensor operation. That really confirms the current architecture doesn't have sync points that break the lazy execution model.

There are many potential improvements that I want to add to that backend later on to improve performance even further and actually make it useful, namely burn-fusion support and an easy-to-use backend extension for distributed dataset loading.


crates/burn-ir/src/builder.rs
```diff
@@ -443,6 +443,235 @@ impl_ir_create!(
...
+    ) -> Self {
+        let out = TensorIr::uninit(new_id(), input.shape.clone(), input.dtype);
+        SortOpIr {
+            input,
+            dim,
+            descending,
+            out,
+        }
+    }
+}
```
laggui: We should reuse the `impl_ir_create!` macro:

```suggestion
impl_ir_create!(
    SortOpIr {
        input: TensorIr,
        dim: usize,
        descending: bool
    },
    shape = input.shape.clone(),
    dtype = input.dtype
);
```

crates/burn-ir/src/builder.rs
```diff
@@ -443,6 +443,235 @@ impl_ir_create!(
...
+    ) -> Self {
+        let out = TensorIr::uninit(new_id(), input.shape.clone(), out_dtype);
+        ArgSortOpIr {
+            input,
+            dim,
+            descending,
+            out,
+        }
+    }
+}
```
laggui: Right now arg reductions use the same operation ir, just different output dtype. We should follow the same convention, whether we introduce a `ReduceDimArgOpIr` or remove `ArgSortOpIr`.

If we remove `ArgSortOpIr` we can have the same create pattern:
```rust
impl_ir_create!(
    SortOpIr {
        input: TensorIr,
        dim: usize,
        descending: bool
    },
    shape = input.shape.clone(),
    dtype = input.dtype
    // Additional constructor for argument reduction
    create_arg(ind_dtype: DType)
);
```

crates/burn-ir/src/builder.rs
```diff
@@ -443,6 +443,235 @@ impl_ir_create!(
...
+
+impl LayerNormOpIr {
+    /// Create a layer-norm IR. Epsilon is stored as `to_bits()` so the struct stays `Hash`.
```
laggui: `ScalarIr::Float` already handles this
https://github.com/tracel-ai/burn/blob/73e4c4604ff211090033fbf943d3b87d3d22e204/crates/burn-ir/src/scalar.rs#L6-L25

The IR can simply represent eps as a scalar:

```rust
pub struct LayerNormOpIr {
    /// Input tensor.
    pub input: TensorIr,
    /// Scale (gamma) parameter.
    pub gamma: TensorIr,
    /// Optional shift (beta) parameter.
    pub beta: Option<TensorIr>,
    /// Numerical-stability epsilon.
    pub epsilon: ScalarIr,
    /// Output tensor.
    pub out: TensorIr,
}
```

crates/burn-ir/src/operation.rs
```diff
@@ -3379,3 +3614,294 @@ impl<O: core::fmt::Debug> OperationOutput<O> for Vec<O> {
...
+/// Operation IR for reduce-all/reduce-any along a specified dim.
+#[derive(Clone, Debug, Hash, PartialEq, Serialize, Deserialize)]
+pub struct ReduceBoolDimOpIr {
+    /// Input tensor.
+    pub input: TensorIr,
+    /// Dim along which to reduce.
+    pub axis: usize,
+    /// Output tensor (bool dtype).
+    pub out: TensorIr,
+}
```
laggui: Why not reuse `ReduceOpIr` and `ReduceDimOpIr`? At the IR level these should be the same.

laggui (commented):
I think we should rename the `Runner` types as well which are a bit confusing

- `RunnerClient` -> `RouterClient`
- `RunnerChannel` -> `RouterChannel`
- `Runner` -> `Executor` (or any other suggestion?)

PR #5036, opened 2026-06-02
Feat/multi device remote

nathanielsimard:
Lets a single burn-remote server host multiple devices, with clients selecting one via Device::remote(addr, index) and enumerating them via Device::enumerate. Includes same-host (no round-trip)
  tensor transfers, cross-device all-reduce, and one worker thread per device session.

  Robustness fixes in this branch:
  - Client no longer hangs when the server goes down — pending requests fail fast on disconnect.
  - Sessions release backend memory on close (memory_cleanup + purge of stranded transfers).
  - Out-of-range device index is now a hard error instead of silently falling back to device 0.

  Cleanup:
  - Server start path deduplicated (sync/async share one macro) with a backend-default fallback for backends that can't enumerate hardware.
  - Removed leftover debug logging; added server "started" log.


crates/burn-backend-tests/tests/common/tensor.rs
```diff
@@ -9,7 +9,7 @@ mod clone_invariance;
 #[path = "../tensor/multi_threads.rs"]
 mod multi_threads;
 
-#[cfg(feature = "distributed")]
+#[cfg(feature = "cuda")]
```
laggui: We should probably keep a `distributed` feature for tests, otherwise testing on cuda without NCCL will break, failing to run any tests at all.

crates/burn-dispatch/src/backend.rs
```diff
@@ -651,19 +647,22 @@ impl AutodiffBackend for Dispatch {
                 DispatchTensorKind::Cuda(tensor) => {
                     tensor.as_autodiff().node.distributed_params.clone()
                 }
+                #[cfg(feature = "remote")]
+                DispatchTensorKind::Remote(tensor) => {
+                    tensor.as_autodiff().node.distributed_params.clone()
+                }
```
laggui, line 653: Now that's it's not strictly behind a `distributed` feature, we could technically reuse one of the dispatch macros. But I think there will still be some refactors / cleanups in the future so we can keep it explicit for now.

crates/burn-dispatch/src/backend.rs
```diff
@@ -839,9 +841,25 @@ impl Dispatch {
                 .map(|i| LibTorchDevice::Cuda(i).into())
                 .collect(),
             #[cfg(feature = "remote")]
-            // Remote devices are user-supplied addresses, not enumerable hardware.
+            // Remote devices are keyed by a network address, which the type-id-only
+            // `enumerate` can't carry. Use [`Dispatch::enumerate_remote`] to list the devices
+            // behind a given address.
             DispatchDeviceId::Remote => Vec::new(),
```
laggui, line 847: Even other devices currently assume a single type-id 0 which is insufficient (e.g. there is more than one Wgpu device type), so I think we could refactor the `Dispatch::enumerate` to have another specifier / filter. For remote that would capture the address.

Not a blocker, this can be done in a follow-up.

crates/burn-router/src/client/base.rs
```diff
@@ -42,6 +42,8 @@ pub trait RouterClient: Clone + Send + Sync + Sized {
     fn read_tensor_async(&self, tensor: TensorIr) -> DynFut<Result<TensorData, ExecutionError>>;
     /// Sync the interpreter, ensure that all computations are finished.
     fn sync(&self) -> Result<(), ExecutionError>;
+    /// TODO: Docs
```
laggui: TODO 😄

crates/burn-tensor/src/device.rs
laggui: I think we can remove this comment, the new `DeviceFilter` makes it pretty clear that you can query/filter multiple `DeviceType`s.

examples/text-classification/examples/ag-news-train.rs
```diff
@@ -106,10 +108,44 @@ mod wgpu {
...
+    /// Address of the `burn-remote` server to train against.
+    const ADDRESS: &str = "ws://localhost:3000";
+
+    /// List every device the remote server hosts and train across all of them.
+    #[cfg(not(feature = "ddp"))]
+    pub fn run() {
+        let mut devices = Device::enumerate(DeviceType::remote(ADDRESS));
+        // devices
+        //     .configure(DeviceConfig::default().float_dtype(ElemType::dtype()))
+        //     .unwrap();
```
laggui: Commented out but likely required to configure the selected `ElemType` as the default

examples/text-classification/examples/ag-news-train.rs
```diff
@@ -106,10 +108,44 @@ mod wgpu {
...
+
+        crate::launch_single(devices.into_vec().pop().unwrap());
+        // crate::launch(ExecutionStrategy::MultiDevice(
+        //     devices.into_vec(),
+        //     burn::train::MultiDeviceOptim::OptimSharded,
+        // ));
```
laggui: Dead code

laggui (commented):
LGTM on the remote side, just some minor comments

PR #5039, opened 2026-06-02
fix(tui): replace panic with clean shutdown on user exit

### LINKED ISSUE #4657: In Tui mode, a panic occurred when exiting.
=== PANIC ===
A fatal error happened, you can check the experiment logs here => '/tmp/guide/experiment.log'
 =============
 
 thread '<unnamed>' (3761708) panicked at /Users/caojinyu/.cargo/registry/src/mirrors.ustc.edu.cn-5857e57f01837ef8/burn-train-0.20.1/src/renderer/tui/renderer.rs:345:9:
Killing training from user input.
 === PANIC ===
A fatal error happened, you can check the experiment logs here => '/tmp/guide/experiment.log'
=============

On MacBook Pro m1 8G RAM, macOS 15.7.4

jaweed3:
When the user kills training in TUI mode (via kill popup or Ctrl+C), the renderer was calling `panic!()` instead of shutting down cleanly, producing a scary error message:

```PANIC: thread panicked at ... renderer.rs:345:9:\nKilling training from user input.```

**Changes:**- `send_event`: return early instead of panicking when kill signal is received- 
`KillPopupAccept::call`: return `true` instead of panicking 

(consistent with `QuitPopupAccept`)

Fixes #4657

### DIFF
--- a/crates/burn-train/src/renderer/tui/renderer.rs
+++ b/crates/burn-train/src/renderer/tui/renderer.rs
@@ -114,6 +114,8 @@ impl TuiMetricsRendererWrapper {
 
     fn send_event(&self, event: TuiRendererEvent) {
         if self.kill_signal.lock().unwrap().try_recv().is_ok() {
+            self.interrupter
+                .stop(Some("Killing training from user input."));
             return;
         }
         if let Err(e) = self.sender.send(event) {
@@ -175,7 +177,9 @@ impl MetricsRendererEvaluation for TuiMetricsRendererWrapper {
 impl MetricsRenderer for TuiMetricsRendererWrapper {
     fn manual_close(&mut self) {
         self.send_event(TuiRendererEvent::ManualClose);
-        let _ = self.handle_join.take().unwrap().join();
+        if let Some(handle) = self.handle_join.take() {
+            let _ = handle.join();
+        }
     }
 
     fn register_metric(&mut self, definition: MetricDefinition) {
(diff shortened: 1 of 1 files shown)


laggui (commented):
Removing the panics and returning instead would simply bypass any shutdown, and the TUI and training would keep going. Have you tried it?

The difference with `QuitPopupAccept` is that it explicitly signals the training interrupter to stop, and the training components actually check for `interrupter.should_stop()` to break the training loop.

When we forcefully terminate, we still have to stop immediately. Right now the panic handles that, but it's a bit noisy given the background and main threads both panic.

jaweed3:
pushed a new approach based on your review. instead of panicking:

- KillPopupAccept now signals the interrupter (same as quit) + sends Close directly to the renderer event loop, so the renderer shuts down cleanly
- send_event returns early on kill signal without panicking, training stops when the interrupter check fires between batches
- wrapper Drop always sends Close in drop, if it was already sent by KillPopupAccept, thats harmless

main tradeoff vs the old panic approach: kill now stops at the next batch boundary (like quit), instead of mid-batch. and code after `launch()` still runs since theres no panic / `process::exit`. i think this is fine, the real win is no more noisy double-panic ,but lmk if you need the truly-immediate-stop behavior preserved.

laggui (commented):
Sorry for the delay; I was on leave for the past few weeks 😅 

Replacing the `panic!`s with an `interrupter.stop` signal doesn't terminate the program immediately, it just stops the training loop. So any following code (e.g. evaluation) can still execute.

laggui:
Closing in favor of #5223 which includes a very simple fix to silently unwind and display a single message to signal the training was killed.

I usually send the suggestions as a comment but this one was so minimal :smile: it was quicker to implement and push.

Issue #5046, opened 2026-06-04
[feat] logical Any/All reduction (enables efficient bool_any / contains_nan)


zhan-wei-919:
### Feature description

Add an `Any` (logical-OR) reduction instruction to `cubek-reduce`, i.e. a reduction that
returns `1` if **any** element along the reduced axis is non-zero, and `0` otherwise.
Optionally add the mirror `All` (logical-AND) reduction in the same change.

Concretely:

- A new `Any` instruction implementing `ReduceInstruction` (and `ReduceFamily`).
- New `ReduceOperationConfig::Any` / `ReduceOperation::Any` variants so it is reachable
  through the dynamic dispatch path that Burn uses.
- (Optional) the same for `All`.

### Feature motivation

`cubek-reduce` currently offers `Sum`, `Prod`, `Mean`, `Max`, `Min`, `Arg*`, `TopK`, but has
no boolean/logical reduction. As a result, boolean reductions in Burn (`bool_any`, `bool_all`)
are emulated as **cast bool → int → `Sum` → compare** (`int_greater_elem(sum, 0)` for `any`,
`int_equal_elem(sum, num_elems)` for `all`). This has two downsides:

1. **Extra work.** It launches an additional comparison kernel and uses a wide
   integer accumulator just to recover a 0/1 answer that an OR/AND reduction produces directly.

2. **It motivated a real bug.** `Tensor::contains_nan()` used to be implemented as
   `sum().is_nan()`, relying on NaN propagating through `+`. But for narrow-exponent floats
   (e.g. f16) a *finite* tensor's partial sums can overflow to `±Inf`, and merging blocks then
   yields `Inf + (-Inf) = NaN` — a **false positive** even though no element is NaN:

   ```rust
   // CUDA f16, shape [12, 13, 4, 64, 64], values in roughly [-59, 0]
   // contains_nan(sum) = true
   // is_nan().any()    = false   <-- correct
   // cpu_first_nan     = None
   // cpu_max_abs       = 59.03125
   ```

   `contains_nan()` was fixed to `is_nan().any()`, which is exact per element and reduces a
   clean 0/1 mask. That fix already routes through `bool_any`, so giving `bool_any` a proper
   OR reduction makes the whole `contains_nan` path both correct *and* efficient — a single
   reduction kernel with a narrow accumulator and no overflow concern.

A first-class `Any` reduction is also generally useful on its own (boolean masks, presence
checks, early-out conditions) and rounds out the reduction op set.

[...]

Issue #304, opened 2023-04-22
Add TransformerDecoder


antimora:
### Feature description

Add TransformerDecoder.

### Feature motivation

Currently Burn has TransformerEncoder but missing TransformerDecoder needed for [OpenAI's Whisper model](https://github.com/openai/whisper). 

### (Optional) Suggest a Solution

(TBD)

antimora:
@nathanielsimard I suppose this task is done, right?

antimora:
Completed by https://github.com/burn-rs/burn/pull/316

PR #5051, opened 2026-06-06
feat(cubecl): wire logical any/all reductions

zhan-wei-919:
### Checklist

### Related Issues/PRs

Depends on tracel-ai/cubek#302
Related with tracel-ai/burn#5046

### Changes

This PR wires Burn's CubeCL backend to cubek's logical `any` / `all` reductions.

Changes include:

  - update `cubek` to the cubek PR commit from tracel-ai/cubek#302
  - update `cubecl`, `cubecl-common`, and `cubecl-zspace` to the matching cubek dependency revision
  - implement CubeCL backend overrides for:
    - `bool_any` / `bool_all`
    - `float_any` / `float_all`
    - `int_any` / `int_all`
  - route logical reductions through cubek `ReduceOperationConfig::Any` / `All`
  - add large-dimension float tests for `any_dim` / `all_dim`
  - adapt `GlobalMemoryStrategy` initialization to the newer cubek interpolate API

  This makes Burn's `contains_nan()` path use a real logical reduction through `is_nan().any()` on CubeCL instead of relying on sum based behavior.

  ### Testing

  ```bash
  cargo check -p burn-cubecl
  cargo test -p burn-backend-tests --features flex test_any_dim_large -- --nocapture
  cargo test -p burn-backend-tests --features flex test_all_dim_large -- --nocapture
  cargo test -p burn-backend-tests --features flex test_any -- --nocapture
  cargo test -p burn-backend-tests --features flex test_all -- --nocapture
  cargo test -p burn-backend-tests --features flex contains_nan -- --nocapture
  cargo fmt --all -- --check
  git diff --check

### DIFF
--- a/crates/burn-cubecl/src/kernel/reduce/base.rs
+++ b/crates/burn-cubecl/src/kernel/reduce/base.rs
@@ -2,7 +2,6 @@
 use super::{autotune_reduce, autotune_sum};
 use crate::{
     CubeRuntime,
-    kernel::cast,
     ops::numeric::{empty_device_contiguous_dtype, zeros_client},
     tensor::CubeTensor,
 };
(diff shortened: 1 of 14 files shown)


crates/burn-cubecl/src/kernel/reduce/base.rs
```diff
@@ -131,6 +132,45 @@ pub fn reduce<Run: CubeRuntime>(
...
+    let out_bool = DType::Bool(out_dtype);
+
+    let reduced = match dim {
+        Some(d) => reduce_dim::<Run>(tensor, None, d, Default::default(), config),
+        None => reduce::<Run>(tensor, None, Default::default(), config),
+    }
+    .expect("Any/All reduce on a valid axis cannot fail");
+
+    let backing = elem_type_to_dtype(dtype_to_elem_type(out_bool));
+    let mut out = cast(reduced, backing);
```
laggui: We can avoid the cast and instead require the output precision for `ReduceOperationConfig::{Any, All}`. Similar to argmax/argmin. It makes more sense anyway than a "logical" encoded float / int.

The underlying u8/u32 types both implement `Numeric` to it should be compatible with the cubek reduce traits, and always only have the `DType::Bool(out_dtype)` relabeled.

laggui (changes requested):

crates/burn-cubecl/src/kernel/reduce/base.rs
zhan-wei-919: That makes sense! but I'm a bit busy recently, will get back this weekend.

laggui: No worries! You can request another review when you're done :slightly_smiling_face:

zhan-wei-919:
Finished. Thanks for your advices

Any/All now require the output precision like ArgMax/ArgMin — the kernel writes the 0/1 flags directly into the bool backing storage This went in on the cubek side as #304

I also rebased on main and bumped cubek to 9deb872, which crosses the matmul reorganization from #305/#306 — the import path updates in burn-cubecl/burn-cubecl-fusion are included here

SamuelBelanger (approved):
LGTM

laggui (approved):

PR #5059, opened 2026-06-08
Add enable flags for MultiHeadAttention configuration Linear::bias.

crutcher:
There exist users in the wild with no bias on some of these components (such as Whisper)

### Checklist

### DIFF
--- a/crates/burn-nn/src/modules/attention/mha.rs
+++ b/crates/burn-nn/src/modules/attention/mha.rs
@@ -43,19 +43,19 @@ pub struct MultiHeadAttentionConfig {
 
     /// Enable bias for the key [Linear] layer.
     #[config(default = true)]
-    pub enable_key_bias: bool,
+    pub key_bias: bool,
 
     /// Enable bias for the value [Linear] layer.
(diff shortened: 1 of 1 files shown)


crates/burn-nn/src/modules/attention/mha.rs
```diff
@@ -40,6 +40,22 @@ pub struct MultiHeadAttentionConfig {
...
+    #[config(default = true)]
+    pub enable_value_bias: bool,
+
+    /// Enable bias for the query [Linear] layer.
+    #[config(default = true)]
+    pub enable_query_bias: bool,
+
+    /// Enable bias for the output [Linear] layer.
+    #[config(default = true)]
+    pub enable_output_bias: bool,
```
laggui: Small nit: I'd remove the `enable_` prefix. Something like `key_bias` is clear enough for intent, and follows the current linear `bias: bool` flag.

Otherwise LGTM

crutcher: Done!

laggui (approved):

PR #5064, opened 2026-06-10
feat(store): extract burnpack format to burn-pack + add minimal record

nathanielsimard:
* Extract burnpack format into a minimal crate.
* Add a new RecordNext type for serialization of Burn parameters.


crates/burn-core/src/store/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+    fn into_record_next(self) -> RecordNext {
+        let mut collector = Collector::default();
+        self.visit(&mut collector);
+        RecordNext::from_tensors(collector.tensors)
+    }
+
+    /// Apply a [`RecordNext`] to this module, returning the loaded module.
+    ///
+    /// Honors the record's [`DTypePolicy`], `validate`, and `allow_partial` settings.
+    fn apply_record(self, record: RecordNext) -> Result<Self, RecordError>
```
laggui: Given that `load_record_next` is the panicking wrapper to this method, I think it would make more sense to rename this fallible method to `try_load_record_next` instead.

The `Applier` is really just a `RecordMapper` anyway (we could also rename it accordingly). I think it "loading" a record makes more sense to users as well.

PR #5065, opened 2026-06-10
hypot function

skewballfox:
## Pull Request Template

This will be useful for any complex backends that use split representation. has a default implementation for backends that don't have an intrinsic that tries to do something close to what is happening in libc.

There is a [simd implementation](https://github.com/ChillFish8/cfavml/pull/15/changes#diff-b59d9b3bf35c7796ff8dd9e664fee8a0a365110699777378bcb39f7694354f44) I wrote before that I could try to adapt to burn to provide a better default. If I remember correctly it almost all inputs it matched the standard libraries output and specific input combinations would cause a 1 ulp deviation. It would involve an unsafe cast from float to int (to do a bitwise xor), and would be off by 1 bit from std on certain inputs, but it is branchless so it should work on a gpu. without the bitwise ops, I could probably implement it with a log_with_base function (need base 2)

NOTE: the broadcast was passing for flex but failing for ndarray, and it looks as though the issue was how ndarray was attempting to broadcast the arrays to compatible shapes. I added a function to create the broadcast shape for both from the current shapes (rather than just trying to broadcast both sides to the other's shape)

### Checklist

### Related Issues/PRs

### Changes

### Testing

added a few test mirroring the structure found for other ops

### DIFF
--- a/crates/burn-backend/src/backend/ops/tensor.rs
+++ b/crates/burn-backend/src/backend/ops/tensor.rs
@@ -1128,7 +1128,7 @@ pub trait FloatTensorOps<B: Backend> {
     /// A tensor with the same shape as `tensor` with square root values.
     fn float_sqrt(tensor: FloatTensor<B>) -> FloatTensor<B>;
 
-    /// Returns a new tensor with square root values.
+    /// Returns a new tensor with the Euclidean distance values.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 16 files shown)


crates/burn-backend/src/backend/ops/tensor.rs
```diff
@@ -1128,6 +1128,54 @@ pub trait FloatTensorOps<B: Backend> {
...
+        let dtype = lhs.dtype();
+        let abs_lhs = B::float_abs(lhs);
+        let abs_rhs = B::float_abs(rhs);
+
+        let diff = B::float_sub(abs_rhs.clone(), abs_lhs.clone());
+        let max = B::float_add(
+            abs_lhs.clone(),
+            B::float_clamp_min(diff.clone(), 0.0.into()),
+        );
+        let min = B::float_sub(abs_lhs, B::float_clamp_min(diff, 0.0.into()));
```
laggui: Shouldn't this be `abs_rhs`?


```suggestion
        let min = B::float_sub(abs_rhs, B::float_clamp_min(diff, 0.0.into()));
```

crates/burn-backend/src/backend/ops/tensor.rs
```diff
@@ -1128,6 +1128,54 @@ pub trait FloatTensorOps<B: Backend> {
...
+
+        // Clamp max to at least epsilon to avoid 0/0; result will be 0 anyway
+        // since min <= max, so (min/clamped_max)^2 won't blow up meaningfully.
+        let max_safe = B::float_clamp_min(max.clone(), 1e-38.into());
```
laggui: This seems awfully close to `f32::MIN_POSITIVE`. We should rely on `FloatDType::finfo` instead to get `min_positive` based on the float dtype.

crates/burn-backend/src/backend/ops/tensor.rs
```diff
@@ -1128,6 +1128,54 @@ pub trait FloatTensorOps<B: Backend> {
...
+
+        let ratio = B::float_div(min, max_safe);
+        let ratio_sq = B::float_mul(ratio.clone(), ratio);
+
+        let inner = B::float_add(B::float_ones(shape, &device, dtype.into()), ratio_sq);
```
laggui: Instead of creating a tensor full of ones prefer using `float_add_scalar`.

crates/burn-backend/src/backend/ops/tensor.rs
```diff
@@ -1128,6 +1128,54 @@ pub trait FloatTensorOps<B: Backend> {
     /// A tensor with the same shape as `tensor` with square root values.
     fn float_sqrt(tensor: FloatTensor<B>) -> FloatTensor<B>;
 
+    /// Returns a new tensor with square root values.
```
laggui: Copy/pasta mistake?

laggui (changes requested):
Looks like the default implementation might be inaccurate (failing on macos CI).

skewballfox:
the issue should be fixed, I tested with wgpu locally

laggui (approved):

PR #5071, opened 2026-06-12
Fix split_strides anchoring on unit dims

ArthurBrussee:
`split_strides` (used by reshape's `ReshapeAnalysis::Split` path) anchors its stride walk on the old shape's trailing dim. A size-1 trailing dim can never anchor a group — `dim_prod == shape[old_idx]` compares a real dim against 1 and never matches — so the walk multiplies the unit dim's stride instead of advancing, and that stride leaks onto the next real dim.

Unit dims also carry arbitrary strides: `repeat_dim` on a size-1 dim legitimately returns a broadcast view with stride 0. Combining the two:

```
[26, 1] strides [1, 0]   --unsqueeze-->   [26, 1, 1] strides [0, 0, 0]
```

Every index along dim 0 now aliases element 0. Any consumer that trusts strides misbehaves; the way this surfaced in practice was `scatter(Add)` with index tensors built via `unsqueeze_dim` + `repeat_dim` (a `[k, 1, 1]` index view): all `k` updates landed on `indices[0]` and none on the other rows.

The fix skips unit dims when anchoring (at the start and after each group advance), so a real new dim can never inherit a unit dim's stride. Stride values assigned to size-1 *new* dims don't matter for addressing.

### Testing

On the repro commit (fix reverted):

```
test tensor::tests::test_split_strides_trailing_unit_dim_broadcast_view ... FAILED
  left: [0, 0, 0]   right: [1, 1, 1]
test tensor::tests::test_split_strides_trailing_unit_dims_arbitrary_strides ... FAILED
  left: [32, 32, 32, 32, 32]   right: [1, 1, 1, 1, 1]
```

and both backend tests fail on wgpu.

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/gather_scatter.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/gather_scatter.rs
@@ -171,3 +171,25 @@ fn scatter_should_panic_on_mismatch_of_shapes() {
 
     tensor.scatter(0, indices, values, IndexingUpdateOp::Add);
 }
+
+#[test]
+fn should_scatter_add_3d_dim0_with_index_view() {
+    // Index tensor expanded from 1D via unsqueeze + repeat_dim — a common
+    // pattern to scatter whole rows along dim 0. The repeat of a unit dim
+    // is a broadcast view; reshaping it must keep stride 1 on dim 0,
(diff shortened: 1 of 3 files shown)


laggui (approved):
Good catch 😅

Issue #4991, opened 2026-05-21
Explicit Stream Creation


nathanielsimard:
## Current behavior

Threads are automatically assigned to a stream on creation. When a user spawns a thread and sends their Burn module to it for processing, that thread gets its own stream. Because each stream owns a separate memory pool, memory-heavy workloads that move across streams cause every pool to grow to accommodate the peak — which manifests as a memory leak. I believe this is what's happening in Burn today.

## Proposed change

Decouple stream creation from threads and make it explicit. The stream API can mirror the thread API closely, and in the default case a stream can still be backed by a thread — so migration stays cheap. But making it explicit unlocks two things:

* Alternative backends. The same API could be implemented on top of Tokio instead of OS threads, which would be a better fit in some scenarios (async-heavy workloads, runtimes already built on Tokio, etc.).
* Wasm support. It becomes possible to support multiple streams on Wasm targets, where threads aren't available.

## Why this helps with the memory issue

Because streams are no longer created implicitly per thread, users (and Burn internally) can reuse a single stream across many threads when appropriate, keeping memory-heavy work pinned to one pool instead of fanning it out across many.

Andy2887:
Hi! Is anyone working on this issue? If not, I would like to work on it. I'll open a draft PR this weekend.

nathanielsimard:
@Andy2887 go ahead I don't think there's someone working on it!

Andy2887:
The PR for CubeCL is created. Once approved, my next step is to make the following changes to Burn:

1. Update Burn to use the latest version of CubeCL so that the stream API is accessible to users.
2. Add integration tests to the new stream API
3. Update documentation to include information about how to set stream for a specific thread.
4. (Optional) create a wrapper method (`execute_on_stream(stream_id, || ...)`) that wraps `stream_id.executes(|| ...)`.- This is because `stream_id` functions more as an identifier. Having a wrapper function that manages execution provides better readability.

---

Edit: this is no longer the most updated plan, since the previous PR is rejected.

Andy2887:
Hi @nathanielsimard . Sorry for misunderstanding your intent! I should have communicated more.

I rethink my solution. Here's what I have in mind right now.

### Implement a stream handler

The stream handler contains a StreamId object (and optionally a worker thread will be created if `Thread` is set, see below). Users could explicitly create streams and perform operations on streams through a stream handler. 

1. The stream handler could be cloned and shared with multiple threads. 

[...]

PR #5073, opened 2026-06-13
fix(data): reuse persistent dataloader workers across epochs (#4792)

### LINKED ISSUE #4792: GPU memory leak during training
**Describe the bug**
There is a GPU memory leak affecting model training, causing multiple backends to crash by running out of memory. This occurs in `0.21.0-pre.3` (and the current `main` branch), but does not occur in `0.21.0-pre.2`. 

**To Reproduce**
1. `git checkout v0.21.0-pre.3`
2. Run the CIFAR-10 example, e.g. `cargo run --example custom-image-dataset --release --features wgpu`
3. Observe the memory leak using e.g. `nvtop`, screenshot below (yellow line is GPU memory usage). The memory jumps up at the start of the epoch for both train and validation passes.
4. This usually causes an out of memory panic in the backend, but if you have lots of VRAM you might not run out before training finishes, depending on model size, number of epochs etc.

This issue occurs for me on `0.21.0-pre.3` but not on `0.21.0-pre.2`. I have found the issue happening with backends `wgpu`, `cuda`, and `vulkan` both with this example, and another unrelated image model, on both GTX 1060 3GB and RTX 3060 12 GB GPUs. CPU backends are unaffected.

**Expected behavior**
Expect memory to stay near constant after the dataset loads, and training run not to panic.

**Screenshots**
Image

 - OS: Ubuntu 24.04
 - Burn Version `0.21.0-pre.3`
 - CUDA version: observed with 13.0 and 13.2

**Additional context**
I have tried bisecting the issue on the burn main branch and resolved that at `3cd6e90e4` there was no issue, but by `6cb763107` there was an issue. I couldn't resolve inbetween because there was another error which I'm assuming was due to mismatch between burn/cube branches. It's possible that this is an issue with memory pooling in CubeCL but I'm not sure. I don't know if this issue is architecture specific, but I have only tested on 2x image related networks.

I have found this issue happening on `dev` and `release` profiles, as well as with/without features `fusion` / `autotune` enabled.

manucouto1:
### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed (see note below).

### Related Issues/PRs

Fixes #4792 (GPU memory leak during training). Related: #4991 (Explicit Stream Creation).

### Changes

`MultiThreadDataLoader::iter()` spawned a fresh set of worker threads on every
call — i.e. once per epoch. Each new OS thread is implicitly assigned its own
CubeCL stream, and every stream owns a separate GPU memory pool. CubeCL does
not reclaim the pool of a stream whose thread has terminated, so each epoch
accumulated another set of pools. This is the GPU memory leak in #4792, which
grows until the device OOMs (the `max_streams = 1` workaround masks it by
collapsing every thread onto a single pool).

This PR spawns the workers **once** into a persistent pool owned by the
dataloader and reuses them for every `iter()` call:

- Workers live for the lifetime of the dataloader, so each keeps a stable
  stream (and memory pool) across all epochs — memory stays constant.
- Each `iter()` hands every worker a fresh per-epoch channel; the worker runs
  one pass into it, sends `Done`, then waits for the next pass.
- Dropping the pool closes the command channels, which terminates and joins the
  workers.
- Dropping an iterator early no longer kills the workers; they simply resume on
  the next pass.

No public API change.

Note on a separate issue: while validating this, a smaller, *independent*
host-memory (RSS) growth was observed on both `main` and this branch (≈ one
staged batch per batch), so it is unrelated to the per-stream pool fix here and
likely a staging-buffer concern worth its own investigation.

### Testing

- New unit tests in `multithread.rs`: `..._multiple_epochs` (re-iterating the
  same loader keeps yielding the full dataset) and `..._resumes_after_early_drop`
  (an early-dropped iterator does not kill the workers).
- `cargo test -p burn-core --lib --features dataset` passes; `cargo fmt` and
  `cargo clippy -p burn-core` are clean.
- Validated end-to-end on an RTX 3090 (CUDA backend) with a synthetic
  multi-epoch training loop: on `main`, VRAM grows ~390 MiB/epoch and the run
  OOMs; with this change VRAM stays flat across 40 epochs.
- Note: full `cargo run-checks` is not run locally (wasm target unavailable in
  my environment); CI covers the full matrix.

### DIFF
--- a/crates/burn-core/src/data/dataloader/multithread.rs
+++ b/crates/burn-core/src/data/dataloader/multithread.rs
@@ -7,7 +7,7 @@ use rand::{Rng, SeedableRng};
 
 use super::batcher::Batcher;
 use super::{BatchDataLoader, BatchStrategy, DataLoader, DataLoaderIterator, Progress};
-use std::sync::{Arc, Mutex, OnceLock, mpsc, mpsc::SyncSender};
+use std::sync::{Arc, OnceLock, mpsc, mpsc::SyncSender};
 use std::thread;
 
 const MAX_QUEUED_ITEMS: usize = 100;
@@ -29,7 +29,7 @@ pub struct MultiThreadDataLoader<I, O> {
 
     // Spawned once and reused across every `iter()` call so each worker keeps a
     // stable CubeCL stream (and its memory pool) instead of leaking one per epoch (#4792).
-    workers: OnceLock<Mutex<WorkerPool<O>>>,
+    workers: OnceLock<WorkerPool<O>>,
 }
 
 /// A message that can be sent between threads.
@@ -177,7 +177,7 @@ where
     }
 
     /// Lazily spawns the persistent worker pool (once) and returns it.
-    fn workers(&self) -> &Mutex<WorkerPool<O>> {
+    fn workers(&self) -> &WorkerPool<O> {
         self.workers.get_or_init(|| {
             let dataloaders = self.initialize();
             let item_counts: Vec<usize> = dataloaders.iter().map(|d| d.num_items()).collect();
@@ -212,11 +212,11 @@ where
                 handles.push(handle);
             }
 
-            Mutex::new(WorkerPool {
+            WorkerPool {
                 senders,
                 handles,
                 item_counts,
-            })
+            }
         })
     }
 }
@@ -227,7 +227,7 @@ where
     O: Send + 'static + std::fmt::Debug,
 {
     fn iter<'a>(&'a self) -> Box<dyn DataLoaderIterator<O> + 'a> {
-        let workers = self.workers().lock().unwrap();
+        let workers = self.workers();
 
         let (sender, receiver) = mpsc::sync_channel::<Message<O>>(MAX_QUEUED_ITEMS);
         let unit: Option<String> = Some("items".to_string());
@@ -243,7 +243,6 @@ where
 
         // Drop our sender so the channel disconnects once every worker is done.
         drop(sender);
-        drop(workers);
 
         Box::new(MultiThreadsDataloaderIterator::new(
             receiver,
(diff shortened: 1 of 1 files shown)


crates/burn-core/src/data/dataloader/multithread.rs
```diff
@@ -26,6 +26,10 @@ pub struct MultiThreadDataLoader<I, O> {
 
     // The lazily initialized data loaders
     dataloaders: OnceLock<Vec<BatchDataLoader<I, O>>>,
+
+    // Spawned once and reused across every `iter()` call so each worker keeps a
+    // stable CubeCL stream (and its memory pool) instead of leaking one per epoch (#4792).
+    workers: OnceLock<Mutex<WorkerPool<O>>>,
```
laggui: Is the `Mutex` really necessary here? I think this could just be `OnceLock<WorkerPool<O>>`, and `self.workers()` could simply return a reference to the pool.

laggui (commented):
While this doesn't directly address the streams/threads decoupling, this makes sense as an immediate fix for the described issue.

The persistent pools stops the per-epoch thread/stream spawning, allowing memory pools to be correctly reused.

Just a minor comment below.

crates/burn-core/src/data/dataloader/multithread.rs
manucouto1: Good catch — you're right, the `Mutex` isn't needed. I've dropped it: `workers` is now `OnceLock<WorkerPool<O>>` and `workers()` returns `&WorkerPool<O>` directly.

Since `MultiThreadDataLoader` is `Send + Sync`, sharing `&self` (and therefore `&self.workers`) across threads is already safe — the `OnceLock` handles the one-time init, and after that `iter()` only reads the pool and sends on the channels, so there's nothing to lock. The only case the `Mutex` was implicitly serializing was iterating the *same* loader concurrently from multiple threads, which isn't a real usage pattern (train/valid use separate loaders).

All multithread tests still pass.

laggui (approved):
Thanks for addressing this!

PR #4989, opened 2026-05-21
quantization: add Calibration::Ternary for BitNet b1.58 ternary weight quantization

simeon-kepp:
## What

Adds `Calibration::Ternary` to the existing `Calibration` enum in `burn-std`. No new layer types.

This directly addresses the feedback from #4983: instead of a standalone `LinearTernary` layer, the right place for ternary calibration is the existing `QuantScheme` parameterization.

## Why

BitNet b1.58 (arxiv:2402.17764 §3.1) trains weights in `{-γ, 0, +γ}` where γ = mean(|W|). The existing `Calibration::MinMax` maps `[min, max]` to the quantized range. Ternary calibration needs `[-γ, +γ]` instead — symmetric, and derived from the mean absolute value rather than the extremes.

The natural storage for ternary weights is `QuantValue::Q2S` + `QuantStore::PackedU32`: 3 states fit in 2 bits, so 16 weights pack into one u32 — **16× smaller than f32, 4× smaller than int8**.

## Usage

```rust
use burn::tensor::quantization::{Calibration, QuantValue, QuantStore, QuantScheme};

// Quantize a Linear layer to 2-bit ternary using BitNet b1.58 threshold
let layer = Linear::new(cfg, &device)
    .quantize(
        &QuantScheme::default()
            .with_value(QuantValue::Q2S)
            .with_store(QuantStore::PackedU32(0)),
        &Calibration::Ternary { threshold: None },  // None => mean(|W|)
    );

// Or supply a threshold explicitly
let layer = layer.quantize(
    &scheme,
    &Calibration::Ternary { threshold: Some(0.5) },
);
```

## Changes

- `crates/burn-std/src/tensor/quantization.rs`: add `Calibration::Ternary { threshold: Option<f32> }` variant with doc comment
- `crates/burn-backend/src/quantization/scheme.rs`: implement `compute_range` for `Ternary` (per-tensor and per-block)
- `crates/burn-backend-tests/tests/tensor/float/quantization/calibration.rs`: 3 new tests (per-tensor auto threshold, per-tensor explicit threshold, per-block auto threshold)

## Tests

| Test | What it checks |
|------|---------------|
| `ternary_calibration_range_per_tensor_auto_threshold` | `mean(\|W\|)` = 0.45 for `[-0.9, -0.3, 0.0, 0.6]` → range `[-0.45, 0.45]` |
[...]

PR #5075, opened 2026-06-14
quantization: native multiply-free ternary matmul for Q2S (BitNet b1.58)

simeon-kepp:
## What

Adds a native `q_matmul` fast path on the **ndarray** backend for BitNet b1.58 ternary weights. A `Q2S` symmetric per-tensor weight (values in `{-1, 0, +1}`) times an `f32` activation is computed by **add / subtract / skip** instead of dequantizing the weight to floats and multiplying — the single per-tensor scale γ is applied once per output element at the end.

Follows up #4989 (which added `Calibration::AbsMean` to *compute* the ternary range): that made it possible to **store** ternary weights; this makes the **compute** exploit them.

## Why

Every quantized matmul path today dequantizes back to float and runs a normal float matmul, so `Q2S` weights buy a memory/bandwidth saving but **no compute saving**. BitNet b1.58's whole premise is that ternary weights replace multiply-accumulate with add/subtract/skip. This is the first implementation of that compute path in burn — a reference in the ndarray backend that the accelerated (cubecl) backends can mirror.

For weight `[K, N]` × activation `[.., K]` (i.e. `Linear`), the inner loop does:

```
+1 -> add the activation,  -1 -> subtract,  0 -> skip
```

i.e. `0` multiplies in the inner loop (vs `M·N·K`), `M·N` scale multiplies at the end, and the zeros are never touched.

## Scope / safety

- Fires only for `Q2S` / `Symmetric` / per-tensor weights, an `f32` activation, and a 2D weight. **Every other case falls through to the existing `dequantize → float_matmul` path, byte-for-byte unchanged** (Q8, per-block, f16, batched weights, quantization propagation, ...).
- Matches the dequantize-then-matmul reference to within f32 rounding (the only difference is summation reassociation, `γ·Σ(±a)` vs `Σ(±γ·a)`).

## Tests

`crates/burn-backend-tests/tests/tensor/float/quantization/ops/extended/ternary_matmul.rs` (runs on ndarray + flex):

- `q2s_ternary_matmul_matches_dequantized` — native path `==` dequantize-then-float-matmul, `Tolerance::relative(2e-2)`.
- `q2s_ternary_matmul_all_zero_weight_is_zero` — an all-zero weight yields exactly `0`.

## Follow-ups (not in this PR)

- `nn::Linear` currently dequantizes a quantized weight to a full float tensor before matmul (via `into_float`), so it does not yet route through `q_matmul` — wiring that up would let a ternary `Linear` use this path. Happy to follow up.
- The natural next step is the same skip/add/sub kernel on the cubecl backends, where it turns the memory saving into a real speed/energy win.


laggui (commented):
Thanks for the contribution.

As a reference implementation that might make sense, but, I don't think we should merge new feature work into the ndarray backend at this point since it's planned for deprecation.

I would rather focus future optimization efforts on the cubecl backends and possibly the flex backend instead.

For those reasons, I'm leaning towards closing this PR.

simeon-kepp:
quick context that might change the picture a bit.

we're a 7-person ternary AI research team (rfi-irfos, vienna) actively building a full ternary inference stack from scratch, specifically around weight sparsity and multiply-free compute. this PR is not a one-off contribution, it's one piece of a larger body of work we intend to maintain ourselves.

on the maintenance concern: we'd be happy to own the ndarray ternary path going forward and keep it in sync. if the worry is that it becomes dead weight, that's on us, and we'll make sure it doesn't.

one thing worth mentioning: are you familiar with @sparseskip? it's a culling technique we developed that skips zero-weight positions at the standard deviation level before the kernel is even called. ternary matmul is the natural home for it since ~50% of weights are zero in a trained BitNet b1.58 model. the ndarray path is the reference we need to validate the skip logic before porting it to cubecl.

so the ask would be: keep this open while we wire up @sparseskip into the path and show the actual zero-skip savings, not just the add/sub substitution. if the numbers look good, the case for a cubecl port becomes obvious.

happy to share the whitepaper if useful.

laggui:
Ahh in this case I don't have any issues adding this as a reference implementation. Just have to update the code to make the CI happy :slightly_smiling_face:

simeon-kepp:
clippy::collapsible_if fixed — the nested if-let is now collapsed with the `&&` syntax. CI needs a maintainer approval to run on fork PRs. @laggui could you trigger it? happy to address any further feedback.

laggui (approved):
Sorry for the delay; I was on leave for the past few weeks 😅 

I resolved the conflicts so we can merge

simeon-kepp:
Thanks for resolving the conflicts @laggui! The CI failures were the two new `ternary_matmul` tests panicking in `quantize()`: since the `burn-backend-tests`/`burn-dispatch` restructure, the ndarray backend is built **without** `export_tests` in the backend test target, so the `#[cfg(not(feature = "export_tests"))]` arm only accepted `Q8F | Q8S` and `Q2S` fell into the `unimplemented!()` arm before the ternary fast path could run.

Pushed `0ee1620c` which adds `Q2S` to the non-test cfg arm — legitimate since `Q2S` is now a first-class native scheme in this backend (i8 storage, multiply-free `q_matmul` path), not just a test reference.

Verified locally on the PR head: the 2 ternary tests pass, full quantization suite 217 passed / 0 failed, fmt + clippy clean.

laggui:
Thanks for fixing the issue, didn't notice after I updated. Will merge when the CI completes!

simeon-kepp:
Quick cross-link for anyone following the ternary/BitNet b1.58 work here: the training-side counterpart just landed in Keras — keras-team/keras#22960 (TernaryDense + Dense.quantize("ternary")). Storage/training compression there, this kernel here is the actual compute-side multiply-free path. Figured it's worth linking the two together for anyone bouncing between ecosystems.

PR #5078, opened 2026-06-15
fix: forward features flag in burn Cargo.toml for burn-vision

Marc-AnthonyG:
### Checklist

### Changes

Feature flag for backend was not properly forwarded toward burn-vision making only flex working with vision without having to override the burn-vision dependency.

### Testing

Tested against local project using metal.


laggui (commented):
LGTM.

Looks like this prompted another feature flag issue in the examples check. The default features are not propagated to `burn-vision`, therefore `fusion` (enabled by default) only propagates to the backends (e.g. wgpu).

If we update `burn/Cargo.toml` to carry the default features to `burn-vision?/default` it should work.

laggui (approved):

PR #5080, opened 2026-06-16
Update readme

nathanielsimard:


README.md
```diff
@@ -13,17 +13,101 @@
...
 </div>
 
 <div align="left">
 
-Burn is both a tensor library and a deep learning framework optimized for numerical computing, model
-inference and model training. Burn leverages Rust to perform optimizations normally only available
-in static-graph frameworks, offering optimal speed without impacting flexibility.
+Training and inference usually live in separate worlds. You train a model in Python, then export it
+to an inference engine like vLLM, ONNX, or TensorRT to run it. That export step is brittle and lossy,
+and it rules out some architectures and use cases.
```
laggui: I'd make two changes: change the second-person address (you) for an authoritative voice, and make sure we don't conflate ONNX (the open format) with ONNX Runtime (the engine).

```suggestion
Training and inference usually live in separate worlds. Models are typically trained in Python then
exported to an open format like ONNX or optimized for production engines like vLLM, ONNX Runtime, or
TensorRT. This export step is often brittle and lossy, ruling out complex architectures and advanced
deployment use cases.
```

README.md
```diff
@@ -13,17 +13,101 @@
...
+
+Burn doesn't separate the two. Everything is lowered to the same multi-platform tensor operations, so
+the code you train is the code you run. That makes things like on-device personalization and
+federated learning straightforward, and you can go from prototype to deployment in a single codebase.
```
laggui: I think *unifying* is punchier, and it clearly states what Burn is solving (instead of not doing).

Suggestion with a minor reformulation:

```suggestion
Burn unifies the two. By executing multi-platform tensor operations via a single, unified API, the
exact code used for training is the exact code that runs in production. This makes workloads like
on-device personalization and federated learning straightforward, while enabling teams to go from
prototype to deployment in a single codebase.
```

README.md
```diff
@@ -13,17 +13,101 @@
...
+
+Burn keeps the feel of PyTorch, with dynamic shapes and graphs, but JIT-compiles streams of tensor
```
laggui: Small nit: should we say "keeps the feel" or "preserves the intuitive ergonomics" (a bit more precise about the feel)?

README.md
```diff
@@ -13,17 +13,101 @@
...
+operations, performing automatic kernel fusion. You get the flexibility of dynamic graphs
+without the performance drop.
+
+## Rust for Research?
+
+Rust used to be a poor fit for research: long builds broke the fast edit-compile-run loop that draws
+people to Python in the first place. Not anymore. Burn is built around incremental compilation, so
+changing model code recompiles in under 5 seconds, even in release mode. You get a Python-like
+feedback loop with the speed and safety of Rust.
```
laggui: Poor fit or tough sell?

Small reformulation to replace "you" with an authoritative tone, like the rest:

```suggestion
Rust used to be a tough sell for research: long compilation times disrupted the fast
edit-compile-run loop that draws researchers to Python. Burn changes this paradigm. Designed around
incremental compilation, modifying model code recompiles in under 5 seconds, even in release mode.
This delivers a Python-like feedback loop with the speed and safety of Rust.
```

README.md
```diff
@@ -13,17 +13,101 @@
...
+|               | `burn-rl`                                             | Reinforcement learning building blocks                                                                                                                            |
+|               | `burn-dataset`                                        | Dataset loading, transforms and ready-made sources                                                                                                                |
+| Models        | [models](https://github.com/tracel-ai/models)         | Curated pre-trained models and examples built with Burn                                                                                                           |
+| Tooling       | [burn-bench](https://github.com/tracel-ai/burn-bench) | Benchmark and compare backends, tracking performance over time                                                                                                    |
+
+Burn's own backends (CUDA, ROCm, Metal, Vulkan, WebGPU, LibTorch and pure-Rust CPU/`no_std`) compose
+with autodiff, fusion and remote-execution decorators. See [Supported Backends](#supported-backends)
+below for the full matrix.
```
laggui: Having LibTorch and Flex (pure-Rust CPU) in the same list as our cubecl backends might read like they also support fusion when they don't.

Maybe we could draw the line a bit more cleanly around CubeCL / accelerated backends for that?
```
Burn's backends (CUDA, ROCm, Metal, Vulkan, WebGPU, LibTorch, and pure-Rust CPU/`no_std`) 
compose with autodiff and remote-execution decorators, with automatic kernel 
fusion enabled by default for all accelerated backends.
```

Feel free to tweak otherwise.

README.md
nathanielsimard: I like "preserves the intuitive ergonomics"

laggui (approved):

PR #5083, opened 2026-06-16
Refactor Record - Serialization & Deserialization of Module, Optimizer & LrScheduler

nathanielsimard:
Revamps how Burn saves and loads state. The legacy record system is replaced by a new binary serialization format, **burnpack**, and a `RecordState` representation that decomposes module and optimizer state into named tensors and scalars. `burn-core`'s old `record` module gives way to a new `store` module, and the optimizer subsystem is moved into its own `burn-optim` crate that builds on the new state system.


crates/burn-core/src/module/base.rs
```diff
@@ -143,70 +139,47 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
-    /// The file extension is automatically added depending on the file recorder provided, you
-    /// don't have to specify it.
-    fn save_file<FR, PB>(
-        self,
-        file_path: PB,
-        recorder: &FR,
-    ) -> Result<(), crate::record::RecorderError>
+    /// The record can be saved to a burnpack file or byte buffer and applied back with
+    /// [`load_record`](Module::load_record).
+    fn into_record(self) -> crate::store::ModuleRecord
```
laggui, line 151: I think we should keep `module.save_file(path)` and `module.load_file(path)` for convenience. Modules always load or convert to a `ModuleRecord`, so there is no ambiguity. Since `ModuleRecord` actually has some configurations (dtype policy, allow partial and validation), the load argument could be wrapped by some load config type, and passing a path would keep all other configs as default.

We could also have a fallible `try_load_file` which goes through `try_load_record`.

crates/burn-core/src/module/base.rs
```diff
@@ -143,70 +139,47 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
-    /// The recorder should be the same as the one used to save the module, see
-    /// [save_file](Self::save_file).
-    ///
-    /// ## Notes
-    ///
-    /// The file extension is automatically added depending on the file recorder provided, you
-    /// don't have to specify it.
-    fn load_file<FR, PB>(
+    /// Honors the record's [`DTypePolicy`](crate::store::DTypePolicy), `validate`, and
+    /// `allow_partial` settings.
```
laggui, line 162: I think I only saw one test for allow partial, and none for the dtype policy and validation. We should add some.

crates/burn-core/Cargo.toml
```diff
@@ -116,7 +113,6 @@ num-traits = { workspace = true }
 rmp-serde = { workspace = true, optional = true }
```
laggui: With the previous file recorders gone, we can remove the dependencies they required like bincode, flate2, rmp-serde, etc.

crates/burn-train/src/checkpoint/base.rs
```diff
@@ -11,19 +11,72 @@ pub enum CheckpointerError {
...
 
+/// A record that can be saved to and loaded from a burnpack file.
+///
+/// Implemented for the burnpack record types used during training: the module
+/// ([`ModuleRecord`]), the optimizer ([`OptimizerRecord`]) and the learning rate scheduler
+/// ([`LrSchedulerRecord`]).
+///
+/// Records are device-free: a checkpoint is just file-backed bytes. Device placement is decided
+/// when a record is applied (the module keeps its existing parameter device; optimizer state
+/// migrates to each parameter's device on the next step), not when the checkpoint is loaded.
```
laggui, line 31: Ha, I just fixed an issue related to the device placement in #5084. This will conflict with the fix, but it should no longer be an issue with loading checkpoints now! Which makes a lot more sense anyway.

crates/burn-core/tests/test_derive_module.rs
```diff
@@ -123,234 +123,10 @@ impl ModuleWithAttributes<ModuleBasic, PaddingConfig> {
...
-        implements_clone::<RecordItem<ModuleWithGenericModule<M>, S>>();
-        implements_clone::<RecordItem<ModuleEnumWithGenericModule<M>, S>>();
-    }
-}
-
 pub fn test_device() -> Device {
     burn_tensor::Device::flex()
 }
 
-mod state {
```
laggui, line 159: We should preserve such tests with the new module records, not simply delete them. They validate the module records for different module types (nested, enums, etc.), as well as state and initialization behavior.

crates/burn-core/src/store/mod.rs
```diff
@@ -76,24 +73,24 @@ struct RecordTensor {
...
+/// ([`save`](ModuleRecord::save) / [`into_bytes`](ModuleRecord::into_bytes)) or apply it back with
+/// [`Module::load_record`]. Load-time behavior is
+/// configured with the builder methods; they are ignored when saving.
 ///
 /// The save-side dtype is intentionally not configurable: use `module.cast(dtype)` before
 /// taking the record. The record stores whatever dtype the module currently holds.
 #[derive(Clone)]
-pub struct RecordNext {
+pub struct ModuleRecord {
     tensors: Vec<RecordTensor>,
```
laggui, line 85: Open question:

Right now plain tensors (not params) are still not captured by module records. This preserves the old behavior, but should they be allowed as buffers (not trainable)? I know it can be useful in practice.

nathanielsimard: Eventually

PR #5087, opened 2026-06-17
Fix fusion vectorization for reduce broadcasted

### LINKED ISSUE #5060: [fusion] Fused broadcast generates invalid vectorizations
Fused broadcast reduce can generate invalid assignment from a scalar to a vector. `tensor::float::module::attention::test_attention_all_options` for sure reproduces it, but there are others.
```
//Fuse on write begin;
%485 = load(%185) : (ptr<f32, local>) -> (f32); @ uniform
%486 = %485 < %17 : (f32, f32) -> (bool); @ uniform
%487 = cast<u8>(%486) : (bool) -> (u8); @ uniform
%488 = cast<bool>(%487) : (u8) -> (bool); @ uniform
%490 = load(%185) : (ptr<f32, local>) -> (f32); @ uniform
%489 = %488 ? %18 : %490 : (bool, f32, f32) -> (f32); @ uniform
store(%64, %489) : (ptr<vector<f32, 16>>, f32) -> ();
//Fuse on write end;
```

nathanielsimard:
fix #5060

### DIFF
--- a/crates/burn-backend-tests/tests/fusion/mod.rs
+++ b/crates/burn-backend-tests/tests/fusion/mod.rs
@@ -4,6 +4,7 @@ mod fusion_f16_broadcast;
 mod fusion_f16_write_vectorization;
 mod fusion_shape;
 mod int_bitwise;
+mod reduce_broadcast_vectorization;
 mod reduce_broadcasted;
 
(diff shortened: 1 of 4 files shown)


Charles23R:
Closed because of staleness.

PR #5088, opened 2026-06-17
Remote backend fusion: client-side op-graph caching 

nathanielsimard:
Adds a fusion path for the remote backend that caches recurring op-graphs on the client and replays them on the server by id, instead of re-streaming the full op sequence every invocation. For workloads with a stable structure (e.g. a model block per training step), this turns a recurring computation into one registration plus cheap per-invocation replays, cutting network traffic substantially.


crates/burn-cubecl/src/ops/bool_tensor.rs
```diff
@@ -49,9 +49,10 @@ impl<R: CubeRuntime> BoolTensorOps<Self> for CubeBackend<R> {
             data.dtype,
             DType::Bool(BoolStore::U8) | DType::Bool(BoolStore::U32)
         ) {
-            unimplemented!("Unsupported dtype for `bool_from_data` {:?}", data.dtype);
+            super::from_data(data.convert_dtype(DType::U32), device)
```
laggui: I think this is no longer required since #5094 (just merged earlier), as bool storage type is resolved automatically in `Tensor::from_data` now. Unless this was specifically for some internal backend usage issue?

crates/burn-ir/src/operation.rs
```diff
@@ -2140,6 +2151,72 @@ impl OperationIr {
...
+            OperationIr::Int(repr) => repr.for_each_tensor_mut(f),
+            OperationIr::Float(_dtype, repr) => repr.for_each_tensor_mut(f),
+            OperationIr::Module(repr) => repr.for_each_tensor_mut(f),
+            OperationIr::Init(repr) => repr.for_each_tensor_mut(f),
+            OperationIr::Custom(repr) => repr.for_each_tensor_mut(f),
+            OperationIr::Drop(repr) => f(repr),
+            OperationIr::Distributed(repr) => repr.for_each_tensor_mut(f),
+            OperationIr::Activation(repr) => repr.for_each_tensor_mut(f),
+        }
+    }
```
laggui: Potential improvement: the three `for_each_*` methods could be consolidated into a single `visit_mut` if we express this a visitor pattern over the IR, e.g. 

```rust
pub trait IrVisitorMut {
    fn visit_tensor_mut(&mut self, _tensor: &mut TensorIr) {}
    fn visit_scalar_mut(&mut self, _scalar: &mut ScalarIr) {}
    fn visit_range_mut(&mut self, _range: &mut Slice) {}
}
```

Which could collapse the three massive match arms into a single clear `visit_mut` mapping

```rust
impl OperationIr {
    pub fn visit_mut(&mut self, v: &mut impl IrVisitorMut) {
        match self {
            OperationIr::BaseFloat(repr)
            | OperationIr::BaseInt(repr)
            | OperationIr::BaseBool(repr) => repr.visit_mut(v),
            // Other variants implemented similarly...
            OperationIr::Drop(tensor) => v.visit_tensor_mut(tensor),
        }
    }
}

impl BaseOperationIr {
    pub fn visit_mut(&mut self, v: &mut impl IrVisitorMut) {
        match self {
            BaseOperationIr::Slice(repr) => {
                v.visit_tensor_mut(&mut repr.tensor);
                v.visit_tensor_mut(&mut repr.out);
                repr.ranges.iter_mut().for_each(|r| v.visit_range_mut(r));
            }
            BaseOperationIr::SliceAssign(repr) => {
                v.visit_tensor_mut(&mut repr.tensor);
                v.visit_tensor_mut(&mut repr.value);
                v.visit_tensor_mut(&mut repr.out);
                repr.ranges.iter_mut().for_each(|r| v.visit_range_mut(r));
            }
            // Other variants traverse their specific fields
        }
    }
}
```

And then have a `ReplayVisitor` for the graph.

PR #5092, opened 2026-06-18
fix(train): avoid overflow in TUI progress when extending training

### LINKED ISSUE #5091: Training Extension Causes Panic in UI code as a result of invalid progress rehydration
ref issue [5082](https://github.com/tracel-ai/burn/issues/5082), which fixed the device issue but there appears to be another problem for extending training, this time in the UI code.

If I train a/my model to, say, 3 epochs, everything is fine (reproducible for both backends I'm testing with, namely CUDA and torch-cuda).

However, if I train my model to 2 epochs and then extend the training to 3 epochs, I get the following error:
```
thread 'train-renderer' (475626) panicked at /home/some-folder/Projects/burn/crates/burn-train/src/renderer/tui/progress.rs:199:21:
attempt to subtract with overflow
```

The error consistently appears part way through the extension epoch.  I'm wondering if this is an issue with rehydration of progress state for continuation of training.

The panic is here:
```
        let epoch = progress.global.items_processed - starting_epoch;
```
where items_processed is 1, but the starting_epoch is 2, underflowing epoch (usize).

SAY-5:
### Checklist

### Related Issues/PRs

Fixes #5091

### Changes

When training is extended from a checkpoint, the global progress can briefly report fewer completed epochs than `starting_epoch`. The TUI progress estimator subtracted `starting_epoch` from `items_processed` (and from `items_total`/`ignore_num_items`) with plain unsigned subtraction, so this underflowed and panicked with "attempt to subtract with overflow" in `init` and `calculate_progress`.

Switched those subtractions to `saturating_sub` and clamped the current epoch to at least 1, so the estimator reports a sensible progress value instead of crashing the `train-renderer` thread.

### Testing

Added a regression test in `progress.rs` that drives `calculate_progress` with `items_processed < starting_epoch`; it panicked before the change and now returns the expected ratio. Also ran `cargo test -p burn-train`, `cargo fmt --check`, and `cargo clippy -p burn-train`.

### DIFF
--- a/crates/burn-train/src/renderer/tui/progress.rs
+++ b/crates/burn-train/src/renderer/tui/progress.rs
@@ -196,7 +196,11 @@ impl ProgressEstimate {
     }
 
     fn init(&mut self, progress: &ProgressSnapshot, starting_epoch: usize) {
-        let epoch = progress.global.items_processed - starting_epoch;
+        let epoch = progress
+            .global
+            .items_processed
+            .saturating_sub(starting_epoch)
+            .max(1);
         let local = &progress.split;
 
         let epoch_items = (epoch - 1) * local.items_total;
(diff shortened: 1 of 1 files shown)


laggui (changes requested):
Thanks for trying to tackle the issue!

As it has been pointed out, I think this only fixes the symptom but not the actual problem when resuming training from a specific epoch.

PR #5094, opened 2026-06-19
fix(cubecl): accept native bool store in bool_from_data

yasumorishima:
## Problem

Model loaders such as `burn-store` deserialize boolean tensors as `DType::Bool(BoolStore::Native)` (the safetensors reader maps `BOOL` to `BoolStore::Native`, and the applier preserves the stored dtype via `Tensor::from_data(data, (device, snapshot.dtype))`). However, the cubecl backend's `bool_from_data` only accepted `BoolStore::U8` / `BoolStore::U32` and panicked on `Native`:

```
not implemented: Unsupported dtype for `bool_from_data` Bool(Native)
  crates/burn-cubecl/src/ops/bool_tensor.rs
```

As a result, **loading any module that contains a bool tensor onto a cubecl backend (CUDA, wgpu, CPU) panics**, even though the `tch` and `candle` backends accept native bool. This is reproducible without a GPU on the `burn-cpu` backend (same `CubeBackend`).

## Fix

Convert `BoolStore::Native` to `BoolStore::U8` on ingestion in `bool_from_data` (a bool is a single byte, so `U8` is the natural storage), then proceed as before. Other stores are unchanged; genuinely unsupported dtypes still panic.

## Test

Added a regression test on the CPU cubecl backend (`burn-cpu`) that builds `BoolStore::Native` `TensorData` and asserts `bool_from_data` accepts it. Before this change the test panicked.

- `cargo test -p burn-cpu`: 2 passed
- `cargo clippy -p burn-cpu --all-targets -- -D warnings`: clean
- `cargo fmt`: applied

## Notes

`supports_dtype(Bool(Native))` intentionally stays `false`: this change only makes *ingestion* robust by converting to a supported store; it does not claim native-bool *operations* on cubecl runtimes.

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/bool/ops/init.rs
+++ b/crates/burn-backend-tests/tests/tensor/bool/ops/init.rs
@@ -29,3 +29,23 @@ fn should_support_bool_ones() {
         .into_data()
         .assert_eq(&TensorData::from([[true, true], [true, true]]), false);
 }
+
+#[test]
+fn should_load_bool_from_data_forcing_native_store() {
+    use burn_tensor::{BoolStore, DType};
+
+    // Model loaders (e.g. `burn-store`) reconstruct tensors with `Tensor::from_data`,
+    // forcing the serialized snapshot dtype. Bool tensors are serialized as
+    // `BoolStore::Native`, a storage layout that cubecl runtimes don't support. Forcing
+    // it used to panic in `bool_from_data`; `resolve_dtype` must now fall back to the
+    // device's default bool storage. Regression test for tracel-ai/burn#5094.
+    let device = Default::default();
+    let data = TensorData::new(vec![true, false, true], [3]);
+    assert_eq!(data.dtype, DType::Bool(BoolStore::Native));
+
+    let tensor = TestTensorBool::<1>::from_data(data, (&device, DType::Bool(BoolStore::Native)));
+
+    tensor
+        .into_data()
+        .assert_eq(&TensorData::from([true, false, true]), false);
+}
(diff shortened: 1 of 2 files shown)


laggui (changes requested):
I believe this fixes the symptom, not the issue. U8 is not supported for all cubecl backends (e.g. wgpu), so even converting the native -> u8 won't completely resolve this.

`Tensor::<D, Bool>::from_data(bool_data, &device)` should correctly resolve the bool dtype to use for the device provided based on its settings.

https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/api/base.rs#L1894-L1909

The issue originates from the provided args in `burn-store`:

https://github.com/tracel-ai/burn/blob/7349bc8e3cc8f66bdd8b50d25ebac26be080c7b1/crates/burn-store/src/applier.rs#L236

which forces the tensor snapshot dtype, even if it might not be supported. `BoolStore` variants are backend-specific storage layouts rather than semantic types, so I think the correct fix would be to check for bool variants in `resolve_dtype`:

https://github.com/tracel-ai/burn/blob/7349bc8e3cc8f66bdd8b50d25ebac26be080c7b1/crates/burn-tensor/src/tensor/api/options.rs#L57-L67

So either check if it's supported, or always use the device default.

yasumorishima:
Thanks for the review — you're right, the original change only masked the symptom in one backend. I've reworked it to fix the root cause:

- Reverted the `bool_from_data` change in `burn-cubecl` (back to the original `U8`/`U32` guard).
- `resolve_dtype` now always resolves `Kind::Bool` to the device's default bool storage (`settings.bool_dtype`), ignoring any forced dtype. Since `BoolStore` variants are backend-specific storage layouts rather than semantic types, a variant forced from a serialized snapshot (as `burn-store`'s applier does) shouldn't override what the target device actually supports. `Float`/`Int` still honor an explicitly provided dtype.
- Moved the regression test to the cross-backend `burn-backend-tests` suite (`tensor/bool/ops/init.rs`): it loads a bool tensor with a forced `BoolStore::Native` dtype, which panicked on the cubecl CPU backend before this change.

Verified on the cubecl CPU backend (`--no-default-features --features cpu`): the test panics without the fix and passes with it, and the `ops::init` tests (incl. the float/int explicit-dtype cases) stay green.

laggui (approved):
Thanks for addressing my comments. This is a good alternative for bool store.

PR #5096, opened 2026-06-20
[wip] fp8 support

skewballfox:
### Checklist

### Related Issues/PRs

### Changes

### Testing


laggui (commented):
I don't think E4M3/E5M2 should be introduced as ordinary `FloatDType`s. Although they are floating-point encodings, their practical use is primarily as scaled low-precision representations consumed by a limited set of accelerated operations, not as general-purpose, unscaled compute types.

Burn now represents this distinction through `DType::QFloat`, `QuantScheme`, and `QuantValue::E4M3/E5M2`. `TensorData` can preserve the encoded FP8 values together with their quantization parameters, while `QTensorOps` provides the right boundary for fallback behavior and specialized backend implementations.

Making FP8 a regular float dtype would imply support across the normal float tensor operation surface and create a second representation overlapping with `QFloat`. Formats such as MXFP8 and NVFP4 reinforce that scaling and storage layout are integral parts of these low-precision representations.

Given this, I don't think further changes to this PR would lead to something we can merge. I'm leaning toward closing it and handling FP8 through focused additions to the existing `QFloat` path.

PR #5098, opened 2026-06-22
update cube

louisfd:


laggui (commented):
If you update with main the audit failure should be fixed via #5099.

PR #5100, opened 2026-06-22
Use CBOR as a serialization format across the crates

nathanielsimard:
WIP: The goal is to remove some dependencies


laggui:
Related: #5544 

/edit: the linked PR only changes `rmp-serde` -> `ciborium` for fusion. This changes dataset/communication/remote crates.

Issue #3864, opened 2025-10-14
`cumprod` gradient produces NaN when input contains zeros


antimora:
## Problem Description

The current `cumprod` autodiff implementation (see https://github.com/tracel-ai/burn/pull/3819) produces NaN gradients when the input tensor contains zeros. This is because the gradient computation uses division by the input values:

```rust
// Current implementation in burn-autodiff/src/ops/tensor.rs
let grad_times_output = B::float_mul(grad, output.clone());
let grad_result = /* reverse cumsum of grad_times_output */;
B::float_div(grad_result, input)  //  Division by zero causes NaN
```

### Example

```rust
use burn::tensor::Tensor;
use burn_autodiff::Autodiff;

let tensor = Tensor::<Autodiff<Backend>, 1>::from_data([2.0, 0.0, 3.0, 4.0], &device)
    .require_grad();
let output = tensor.clone().cumprod(0);
let grads = output.sum().backward();
let grad = tensor.grad(&grads).unwrap();

// Expected: [1.0, 32.0, 0.0, 0.0]
// Actual:   [1.0, NaN, NaN, NaN]
```

## Expected Behavior

The gradient should be computed correctly even when the input contains zeros, matching PyTorch's behavior:

```python
import torch

x = torch.tensor([2.0, 0.0, 3.0, 4.0], requires_grad=True)
y = torch.cumprod(x, dim=0)  # [2, 0, 0, 0]
y.sum().backward()
print(x.grad)  # tensor([1., 32., 0., 0.])
```

## Root Cause

The gradient formula `grad_input[i] = sum_{j>=i}(grad_output[j] * output[j] / input[i])` involves division by `input[i]`, which fails when `input[i] = 0`.

## Mathematical Solution

The correct zero-safe gradient formula is:

```
grad_input[i] = left[i] * tail[i]

where:
  left[i] = prod(input[0:i])  // exclusive left cumulative product
  tail[i] = sum_{j>=i}(grad[j] * prod(input[i+1:j+1]))
```

The key insight is that this formula **doesn't involve `input[i]` in any division**, making it well-defined even when `input[i] = 0`.

The `tail[i]` term can be computed efficiently via **reverse accumulation**:

```
tail[n-1] = grad[n-1]
tail[i] = grad[i] + input[i+1] * tail[i+1]  // for i < n-1
```

### Verification

This formula has been verified to match PyTorch's output for all test cases including:
- Zeros at the beginning, middle, and end
- Multiple zeros
[...]

PR #5106, opened 2026-06-24
feat(autodiff): implement prod and prod_dim

### LINKED ISSUE #1458: Implement prod and prod_dim for Autodiff backend
Currently prod and prod_dim are using unoptimized default implementation (log, exp, sum). We should implement more optimized version for Autodiff backend.

leonardwetzel:
### Checklist

### Related Issues/PRs

Closes #1458. Shares the zero-gradient limitation already tracked for `cumprod` in #3864.

### Changes

The Autodiff backend never overrode `float_prod`/`float_prod_dim`, so it fell through to
the default implementation in `burn-backend` (`crates/burn-backend/src/backend/ops/tensor.rs`),
which computes the product as `exp(sum(log(x)))`. On the Autodiff path that has two problems:

- `log` is only defined for positive inputs, so `prod` returns `NaN` for any tensor that
  contains a negative value, and the gradient blows up around zeros. The base backends
  compute the real signed product (e.g. `[[1, -2, 3], [5, 9, 6]].prod() == -1620`), so
  `Autodiff<NdArray>` and `NdArray` disagreed on the exact same input.
- It builds three graph nodes (`log -> sum -> exp`), each with its own saved state and
  backward pass, for what is a single reduction.

This PR adds dedicated `float_prod` and `float_prod_dim` ops. The forward calls the backend's
real product, and the backward uses the analytic gradient `grad * prod(x) / x`
(the leave-one-out product). The forward now agrees with the other backends on negatives,
and it is one graph node instead of three. The structure follows the existing `Sum` and
`CumProd` ops in the same file.

Important limitation: the gradient divides by the input, so it returns `NaN` when an element is
exactly zero. This is the same issue `cumprod` already has (#3864). I kept that out as it takes more care and to keep this change focused. Note it is still strictly better than the old default, 
which also produced `NaN` on zeros and additionally on every negative input. The zero cases are covered
by `#[ignore]`d tests carrying the expected reference values, so they are ready to enable
once the zero-safe path lands.

### Testing

Added gradient tests under `crates/burn-backend-tests/tests/autodiff/aggregation.rs`:
`prod`, `prod_dim` over both a trailing and a leading dim, the negative-input cases (which
also assert the forward value), and the `#[ignore]`d single-zero / multiple-zeros cases with
PyTorch reference gradients. Expected values were checked by hand against the leave-one-out
product.

```
cargo test -p burn-backend-tests --test autodiff --features ndarray
cargo fmt --check
cargo clippy -p burn-autodiff --features export_tests
```

All pass; the two zero-input tests are `#[ignore]`d as described above.

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -2020,6 +2020,95 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
         }
     }
 
+    fn float_prod(tensor: FloatTensor<Self>) -> FloatTensor<Self> {
+        #[derive(Debug)]
+        struct Prod;
+
+        impl<B: Backend> Backward<B, 1> for Prod {
+            // Saves the input and the output product so backward can compute
+            // `grad * prod(x) / x` without recomputing the reduction.
+            type State = (B::FloatTensorPrimitive, B::FloatTensorPrimitive);
+
+            fn backward(
+                self,
+                ops: Ops<Self::State, 1>,
+                grads: &mut Gradients,
+                _checkpointer: &mut Checkpointer,
+            ) {
+                let (input, output) = ops.state;
+
+                unary::<B, _>(ops.parents, ops.node, grads, |grad| {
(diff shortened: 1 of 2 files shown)


laggui (approved):
LGTM! Perhaps we should have a tracking issue for the zeros case, separate from the cumprod op.

PR #5109, opened 2026-06-25
Fix checkpointing + minor stuff

Charles23R:
### Checklist

### Changes

- Fix bug where starting_epoch wasn't used by progress loggers.
- Fix test split metric logging in the `FileMetricLogger`.
- Update db-pedia-train in text-classification example.
- Update all examples to NOT "remove existing artifacts before [training] to get an accurate learner summary". Now, only the logs are removed. This prevented checkpointing by default since it deleted the artifacts. Also, This could delete a whole directory if the path already contained other stuff. We could add a `reset` feature flag or something if we really wanted to keep the behaviour but I think we can keep it like this.

### Testing

Tested with examples.

### DIFF
--- a/burn-book/src/basic-workflow/training.md
+++ b/burn-book/src/basic-workflow/training.md
@@ -148,6 +148,8 @@ Book.
 Let us move on to establishing the practical training configuration.
 
 ```rust , ignore
+# use std::path::PathBuf;
+#
 # use crate::{
 #     data::{MnistBatch, MnistBatcher},
 #     model::{Model, ModelConfig},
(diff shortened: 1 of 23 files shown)


Zatiji (approved):
looks good to me

fuzing:
@Charles23R - thanks for your PR, it's much appreciated.

It doesn't quite fix the issue enumerated (and discussed) here:
[PR-5092](https://github.com/tracel-ai/burn/pull/5092)
[Issue-5091](https://github.com/tracel-ai/burn/issues/5091)

It fixes the underflow issue in the UI, but doesn't rectify the storage of follow-on validation and training logs.

After applying the PR, when running training continuation, all follow-on checkpointed artifacts appear to be correctly numbered and stored in the checkpoint folder, however training and validation logs are first stored in epoch-1 and then the correct follow-on folder.

So, for example, if I train a model for 10 epochs then everything is fine. Then, when continuing the training with follow-on epochs:

- The first follow-on epoch (i.e. 11 in this case) validation and training logs are stored in train/epoch-1 and valid/epoch-1, clobbering the original logs
- The second and subsequent follow-on epochs (i.e. 12, 13, etc) are stored in the correct folders (i.e. train/epoch-12, valid/epoch-12, train/epoch-13, valid/epoch-13 and so on)

Charles23R:
@fuzing thanks for pointing it out, I'll look into it today

Issue #5108, opened 2026-06-25
`any`/`all` reductions bypass the dedicated cubek reduce on fusion backends


zhan-wei-919:
## Summary

cubek has a fused single-pass Any/All instruction (max/min-of-flags), wired into the bare `CubeBackend`. But Wgpu/Cuda default to `Fusion<CubeBackend>`, and `any`/`all` aren't wired into the fusion layer → they fall back to burn's built-in sum-emulation (`into_int` / `sum` / `greater`). The dedicated instruction is dead code on the default backends.

## Root cause

`AnyDim`/`AllDim` are `BaseOperationIr` (bool output), but the reduce fuser only recognizes `NumericOperationIr`'s `*Dim`; and Fusion doesn't even record `any` as `AnyDim` at all (it uses the default decomposition).

In detail:

- `Fusion<B>` doesn't override `*_any`/`*_all`, so it uses the default trait impls in `burn-backend` (`backend/ops/tensor.rs::float_any`, `backend/ops/bool_tensor.rs::bool_any`) which decompose into elementwise + a `Sum`.
- The bool→int/float cast in that decomposition is **not fused into the reduce**, so it materializes a full intermediate tensor to global memory (~3× DRAM traffic for a memory-bound reduce).
- The reduce fuser (`burn-cubecl-fusion/src/optim/reduce/fuser.rs`) only matches `NumericOperationIr::*Dim`. `any`/`all` are `BaseOperationIr::AnyDim`/`AllDim` (bool output, `burn-ir/src/operation.rs`), so they're never picked up as a fusible reduction.

## Benchmark

`RAW` = `CubeBackend` (dedicated reduce). `FUSED` = `Fusion<CubeBackend>` (current default; sum-emulation). Measured in median ms, reduce over dim 1, RTX 5060 Ti.

| shape | wgpu FUSED `any` | wgpu RAW `any` | cuda FUSED `any` | cuda RAW `any` |
|--------------|------------------|----------------|------------------|----------------|
| 4096×4096 | 0.90 | **0.36** | 0.58 | **0.17** |
| 1048576×64 | 3.01 | **1.02** | 2.80 | **0.94** |
| 256×262144 | 2.81 | **0.82** | 2.19 | **0.78** |
> This was tested by Claude
---

I'm happy to take this on and open a PR to fix it.

laggui:
[...]

PR #5113, opened 2026-06-27
perf(fusion): use cubek's dedicated any/all reduce instruction

zhan-wei-919:
`Fusion<CubeBackend>` (the default Wgpu/Cuda backend) didn't override `*_any`/`*_all`, so they fell back to the default `into_int -> sum ->compare` emulation. The reduce fuser also only matched `NumericOperationIr::*Dim`, never `BaseOperationIr::AnyDim/AllDim`, so cubek's dedicated single-pass Any (max-of-flags) / All (min-of-flags) instruction was dead code on the default backends — leaving `any`/`all` ~2-7x slower than necessary for a memory-bound reduce.

- burn-cubecl-fusion: add `ReduceInstruction::Any/All`, map them to `ReduceOperationConfig::Any/All`, and recognize `BaseOperationIr::AnyDim/AllDim` in the reduce fuser (and the broadcasted fuser).
- burn-fusion: override `{bool,float,int}_{any,all}{,_dim}` for `Fusion<B>`. The `*_dim` variants record `AnyDim/AllDim` so they fuse through cubek's instruction; the whole-tensor variants record `BaseOperationIr::Any/All` and run via the bare-backend reduce, the same way whole-tensor `sum`/`prod` are handled.

No IR or cubek changes were needed: the IR variants and their relativization already existed (used by the bare backend and burn-router).
Tested on wgpu and cuda (RTX 5060 Ti): new fusion correctness tests in `tests/fusion/reduce_logical.rs` plus the existing any/all suite pass on both backends; measured 2.2-7.6x faster than the sum-emulation it replaces.

### Checklist

- [x ] Confirmed that `cargo run-checks` command has been executed.
- [ x] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

#5108 

### Testing

| shape         | wgpu before | wgpu after | cuda before | cuda after |
|---------------|-------------|------------|-------------|------------|
| 4096×4096     | 0.90        | 0.36   | 0.58        | 0.17  |
| 1048576×64    | 3.01        | 1.02   | 2.80        | 0.94   |
| 256×262144    | 2.81        | 0.82   | 2.19        | 0.78   |

 Added `tests/fusion/reduce_logical.rs` covering the fused `any`/`all` path for bool / float / int inputs, both `*_dim` and whole-tensor reductions, against a CPU reference. The reduced dimension is deliberately wide (257) so the reduction spans several plane/cube blocks that must merge, and the float/int inputs mix magnitudes and signs (incl. negatives) to exercise the bool→flag normalization inside the instruction.

Verified on:
- **CUDA** (RTX 5060 Ti, CUDA 13.0) — new `reduce_logical` suite (10/10) and the existing `any`/`all` tensor tests pass.
- **wgpu** — same suites pass.

### DIFF
--- a/crates/burn-backend-tests/tests/fusion/reduce_logical.rs
+++ b/crates/burn-backend-tests/tests/fusion/reduce_logical.rs
@@ -1,0 +1,196 @@
+//! Fusion-path correctness for `any`/`all` reductions.
+//!
+//! Under `Fusion<CubeBackend>` these route through the recorded `AnyDim`/`AllDim`
+//! IR into cubek's dedicated Any (max-of-flags) / All (min-of-flags) instruction,
+//! instead of the default `bool->int->sum->compare` emulation. The reduced
+//! dimension is kept wide enough to exercise the multi-block (plane/cube) merge.
+
+use super::*;
+use burn_tensor::TensorData;
+
+const ROWS: usize = 4;
+// Wide reduced dim so the reduction spans several plane/cube blocks that must merge.
+const COLS: usize = 257;
+
+/// Deterministic mask: `(r + c) % stride == 0`. With `stride == 3` every row has a
+/// mix of set/unset flags; `stride == ROWS * COLS + 1` makes the whole tensor all-false.
+fn mask(rows: usize, cols: usize, stride: usize) -> Vec<bool> {
+    (0..rows * cols)
+        .map(|i| (i / cols + i % cols) % stride == 0)
+        .collect()
+}
+
+/// Non-zero <-> flag set; mixes magnitudes/signs so the flag normalization is exercised.
+fn floats(m: &[bool]) -> Vec<f32> {
+    m.iter()
+        .enumerate()
+        .map(|(i, &b)| {
+            if b {
+                if i % 2 == 0 { 2.0 } else { -3.0 }
+            } else {
+                0.0
+            }
+        })
+        .collect()
+}
+
+fn ints(m: &[bool]) -> Vec<i32> {
+    m.iter()
+        .enumerate()
+        .map(|(i, &b)| {
+            if b {
+                if i % 2 == 0 { 2 } else { -3 }
+            } else {
+                0
+            }
+        })
+        .collect()
+}
+
+fn expect_any_dim(m: &[bool], rows: usize, cols: usize) -> TensorData {
+    let v: Vec<bool> = (0..rows)
+        .map(|r| (0..cols).any(|c| m[r * cols + c]))
+        .collect();
+    TensorData::new(v, [rows, 1])
+}
+
+fn expect_all_dim(m: &[bool], rows: usize, cols: usize) -> TensorData {
+    let v: Vec<bool> = (0..rows)
+        .map(|r| (0..cols).all(|c| m[r * cols + c]))
+        .collect();
+    TensorData::new(v, [rows, 1])
+}
+
+// ---- dim reductions ----------------------------------------------------------
+
+#[test]
+fn bool_any_dim_fused() {
+    let device = Default::default();
+    let m = mask(ROWS, COLS, 3);
+    let tensor = TestTensorBool::<2>::from_data(TensorData::new(m.clone(), [ROWS, COLS]), &device);
+    device.sync().unwrap();
+
+    let actual = tensor.any_dim(1).into_data();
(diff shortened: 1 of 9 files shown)


zhan-wei-919:
Hi, just a friendly ping — this PR has been open for a bit over a week. Would appreciate a review when someone has time. Happy to make any changes needed!

crates/burn-backend-tests/tests/fusion/reduce_logical.rs
```diff
@@ -0,0 +1,196 @@
...
+    let v: Vec<bool> = (0..rows)
+        .map(|r| (0..cols).all(|c| m[r * cols + c]))
+        .collect();
+    TensorData::new(v, [rows, 1])
+}
+
+// ---- dim reductions ----------------------------------------------------------
+
+#[test]
+fn bool_any_dim_fused() {
```
nathanielsimard, line 73: Every *_dim test uses mask(ROWS, COLS, 3), i.e. (r + c) % 3 == 0 with COLS = 257. Because every row contains at least one set flag and at least one unset flag, the reduced result is constant across rows:

any_dim expected: [True, True, True, True]
all_dim expected: [False, False, False, False]
bool_reshape_then_any_dim: all True

So bool_any_dim_fused, float_any_dim_fused, int_any_dim_fused, bool_reshape_then_any_dim, and float_elemwise_then_any_dim all pass against a kernel that unconditionally returns true, and the three *_all_dim_fused tests pass against one that unconditionally returns false. These are exactly the tests covering the code the PR adds — the only test with real true/false discrimination is bool_any_all_whole, which runs on the non-fused path.

The fix is small: make the pattern row-dependent so the expected vector varies. For example, gate on the row so some rows are all-false and some are all-true, and keep one mixed row:

crates/burn-cubecl-fusion/src/optim/reduce/fuser.rs
```diff
@@ -257,6 +257,20 @@ impl<R: Runtime> OperationFuser<CubeOptimization<R>> for ReduceFuser<R> {
                         self.on_elemwise_read(operation);
                     }
                 };
+            } else if let OperationIr::BaseBool(BaseOperationIr::AnyDim(op))
```
nathanielsimard: The new else if let chain with or-patterns reads differently from the two if let ... { match op { ... } } blocks above it.

crates/burn-fusion/src/ops/unary.rs
```diff
@@ -119,6 +119,150 @@ macro_rules! reduce_int_ops {
     };
 }
 
+#[allow(missing_docs)]
+#[macro_export(local_inner_macros)]
+macro_rules! reduce_bool_ops {
```
nathanielsimard: We could probably reduce the number of macros and consolidate them together. There is a lot of duplication

nathanielsimard (commented):
1. We should add test for the `reduce_broadcasted` case.

crates/burn-backend-tests/tests/fusion/reduce_logical.rs
zhan-wei-919: I updated the mask to include all-false, all-true, and mixed rows, so the expected results are no longer constant. I also added a test covering the reduce-broadcasted path for both any and all

crates/burn-cubecl-fusion/src/optim/reduce/fuser.rs
zhan-wei-919: Refactored this to match the base operation first, then handle AnyDim and AllDim in an inner match, consistent with the numeric branches above

crates/burn-fusion/src/ops/unary.rs
zhan-wei-919: Consolidated the duplicated reduce macros into a single reduce_ops! macro and updated the existing call site

zhan-wei-919:
Thanks for the review! I’ve addressed all the comments and added the requested test coverage

laggui (approved):
Thanks for addressing the comments, LGTM!

Oh and sorry for delayed review; I took a couple of weeks of paternity leave so my colleagues had to deal with the increased bandwidth 😅

zhan-wei-919:
Thanks for the review! And congratulations on the new baby — hope you
enjoyed the time off! 🎉

Guillaume Lagrange ***@***.***> 于2026年7月20日周一 22:25写道：

Issue #4832, opened 2026-04-17
Add AdaptiveAvgPool3d module (adaptive 3D average pooling)


antimora:
Burn supports adaptive 1D and 2D average pooling (`AdaptiveAvgPool1d`, `AdaptiveAvgPool2d`) but is missing `AdaptiveAvgPool3d`, which is needed for global or output-size-targeted pooling of volumetric data. ONNX and PyTorch provide this, and it is widely required in medical and video workflows.

**Requested feature:**
- Implement `AdaptiveAvgPool3d` for the nn library, mirroring the configuration and API style of the 1D/2D adaptive avg pool modules.
- Add documentation and comprehensive unit tests.

**Motivation:**
- Many 3D architectures rely on adaptive pooling (e.g. output size matches input constraints or is reduced to [1,1,1] as "global" pooling).
- ONNX backend test coverage and model imports need this operator.

Reference: [burn-onnx#343](https://github.com/tracel-ai/burn-onnx/issues/343)

jaweed3:
Implemented in #5115 , AdaptiveAvgPool3d module + all backends (ndarray, cubecl, tch, candle, flex, autodiff, fusion, router, dispatch) with forward and backward tests.

laggui:
Completed in #5115, just missing #5288.

PR #5115, opened 2026-06-28
feat: add AdaptiveAvgPool3d module (#4832)

jaweed3:
Implements the 3D adaptive average pooling operation, closing #4832.

### Added
- `AdaptiveAvgPool3d` module in `burn-nn`
- `tensor::module::adaptive_avg_pool3d` public API
- `ModuleOps::adaptive_avg_pool3d` + `adaptive_avg_pool3d_backward` trait methods
- IR structs, enum variants, builder macros, and visitor impls
- Backend implementations: ndarray (native Array5), cubecl (CPU fallback), tch (native), candle (panic), flex (all 4 dtypes), autodiff (backward with checkpointing), fusion, router, dispatch
- Forward tests (3) and autodiff backward tests (4)

### DIFF
--- a/crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
+++ b/crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
@@ -39,6 +39,44 @@ fn should_quantize_dequantize_per_block_arange_reshaped<const D1: usize, const D
     output.assert_approx_eq::<FloatElem>(&output_ref, Tolerance::permissive());
 }
 
+fn should_quantize_dequantize_reshaped_roundtrip<const D1: usize, const D2: usize>(
+    level: QuantLevel,
+    value: QuantValue,
+    store: QuantStore,
+    shape: [usize; D1],
+    new_shape: [usize; D2],
+) {
+    let numel = Shape::from(shape).num_elements() as i64;
+    let device = Default::default();
+
+    let scheme = QuantScheme::default()
+        .with_level(level)
+        .with_value(value)
+        .with_store(store);
+
+    let input = TestTensorInt::arange(0..numel, &device)
+        .float()
+        .div_scalar(numel)
+        .reshape::<D1, _>(shape);
+
+    // Reference: quantize → dequantize → reshape (quantized roundtrip, then reshape)
+    let reference = input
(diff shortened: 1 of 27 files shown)


crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
laggui, line 1: The quantization changes are unrelated. I'd leave them out of this PR.

crates/burn-cubecl/src/kernel/pool/base.rs
```diff
@@ -328,6 +332,148 @@ pub(crate) fn adaptive_avg_pool2d_backward<R: CubeRuntime>(
     permute_nhwc_to_nchw(output)
 }
 
+pub(crate) fn adaptive_avg_pool3d<R: CubeRuntime>(
+    input: CubeTensor<R>,
+    output_size: [usize; 3],
+) -> CubeTensor<R> {
+    // ponytail: CPU fallback until cubek adds pool3d support
```
laggui: I think I'd rather have no support at all than this CPU fallback in this case. Keep the issue opened, and perhaps link it to a new issue in cubek so it can be tracked over there too.

laggui (changes requested):
Sorry for the delay; I was on leave for the past few weeks.

Small note: this doesn't completely close the linked issue given the CubeCL fallback implementation.

crates/burn-cubecl/src/kernel/pool/base.rs
jaweed3: ok. will remove support for CPU fallback. also keeping the issue opened after this PR.

crates/burn-backend-tests/tests/cubecl/quantization/reshape.rs
jaweed3: ok, i'll leave them out of this PR :)

jaweed3:
Updated per maintainer feedback: removed CPU fallback implementation for CubeCL backend (now uses todo stubs) and reverted unrelated quantization changes. CI failure due to unused variable warnings has been fixed.

laggui (approved):

Just had to fix the cargo fmt

PR #5116, opened 2026-06-29
fix metric logging when checkpoint + burn-train integration tests

Charles23R:
### Checklist

### Related Issues/PRs

#5109 
#5091  

### Changes

- Added a store event for the start of a split, fixing the default behaviour of always logging metrics to first epoch by default.
- Added integration tests to burn-train (only testing checkpointing behaviour for now, but can be extended in the future).

### Testing

New integration tests
Tested with MNIST example.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1219,6 +1219,7 @@ dependencies = [
  "serde",
  "sysinfo 0.38.4",
  "systemstat",
+ "tempfile",
  "thiserror 2.0.18",
  "tracing-appender",
  "tracing-core",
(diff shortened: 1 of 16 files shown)


Charles23R:
@fuzing If I could also get your review on that so we can close #5091 it would be appreciated :smile:

fuzing:
@Charles23R - I just merged locally and re-tested, and everything seems to work like a charm.  Thanks so much for your work on this, it's much appreciated!

Zatiji (approved):
looks good to me

PR #5118, opened 2026-07-01
Add R2 Score metric for regression tasks

Infinty-ux:
This PR adds the R2 Score (Coefficient of Determination) metric to \urn-train\. This helps users evaluate regression models during training, addressing the lack of built-in regression metrics (e.g. #544).


Charles23R (changes requested):
Looks good! The only things missing are some tests for the metric.

Charles23R:
Closed in favor of #5119

PR #5119, opened 2026-07-01
Add MSE and MAE metrics for regression tasks

Infinty-ux:
This PR adds the Mean Squared Error (MSE) and Mean Absolute Error (MAE) metrics to \urn-train\. This extends the regression metric suite (following #5118) to allow users to directly evaluate regression and image quality models during training.


Charles23R (changes requested):
We only need some tests for the metrics. Great work!

Charles23R:
Since R2 score is also included in this PR, i will close #5118 in favor of this.

laggui:
Closing due to inactivity, but let me know if you want to re-open when ready

PR #5139, opened 2026-07-08
Feat/lora qlora

Charles23R:
### Checklist

### Changes

- Lora + qlora
- New example in text-classification

### Testing

Unit tests + new lora finetune example


examples/text-classification/src/finetune.rs
```diff
@@ -0,0 +1,95 @@
...
+        config.seq_length,
+    )
+    .init(&strategy.main_device().clone().autodiff());
+
+    // Load pre-trained weights.
+    let model = model.load_file("model.bpk");
+
+    // Apply LoRA to the attention module's query, value, output and feed-forward weights.
+    let r = 8.0;
+    let mut model = model.apply_lora(&mut LoraMapper::new(LoraConfig::new(r as usize, 2.0 * r)));
```
nathanielsimard, line 50: Should we use a param matcher to apply the lora adaptor?

Charles23R: I tried a few combinations but the best performance/fastest training by far is when i applied LoRA on all params. Probably due to the fact this model has so few layers.

nathanielsimard (approved):

PR #5141, opened 2026-07-08
fix(autodiff): zero-safe cumprod gradient (no division by input)

### LINKED ISSUE #3864: `cumprod` gradient produces NaN when input contains zeros

AnayGarodia:
Fixes #3864.

## Problem

The `cumprod` backward pass computed `reverse_cumsum(grad * output) / input`, dividing by the input and producing **NaN** gradients whenever the input contains zeros. Four autodiff tests for this case were `#[ignore]`d with a TODO pointing at this issue.

## Fix

Replace the division with the standard zero-safe formula `grad_input[i] = left[i] * tail[i]`, where:

- `left[i] = prod(input[0..i])` — exclusive prefix product (built by prepending `1` along `dim`, dropping the last element, then an inclusive cumprod)
- `tail[i] = grad[i] + input[i+1] * tail[i+1]` — reverse accumulation along `dim`

This never divides by `input[i]`, so it stays finite at zeros and matches `torch.cumprod`'s backward.

## Testing

Un-ignored the four zero-input tests (`zero_in_middle`, `zero_at_start`, `zero_at_end`, `multiple_zeros`), whose expected gradients are taken from PyTorch — they now pass.

```
cargo test -p burn-backend-tests --features ndarray,std --test autodiff cumprod   # 12 passed, 0 ignored
cargo clippy -p burn-autodiff --all-targets -- -D warnings                        # clean
cargo fmt --check                                                                 # clean
```

Verified numerically against PyTorch references, e.g. input `[2,0,3,4]` → grad `[1,32,0,0]` (was NaN). The fix uses only backend-agnostic tensor ops; I ran it against the ndarray backend.

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -2162,58 +2162,76 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                 let (input, dim) = ops.state;
 
                 unary::<B, _>(ops.parents, ops.node, grads, |grad| {
-                    // zero-safe cumprod gradient (#3864).
+                    // grad_input[i] = sum_{j>=i}(grad[j] * prod_{k<=j, k!=i} input[k])
                     //
-                    // grad_input[i] = left[i] * tail[i] where
-                    //   left[i] = prod(input[0..i])                        (exclusive prefix product)
-                    //   tail[i] = sum_{j>=i}(grad[j] * prod(input[i+1..=j]))
+                    // Where input[i] != 0 this reduces to the classic
+                    //   reverse_cumsum(grad * output) / input
+                    // but the division makes it NaN at zeros (#3864). Split each lane
+                    // along `dim` at its FIRST zero (the approach PyTorch's
+                    // cumprod_backward takes):
                     //
-                    // no division by input[i], so it stays finite when the input has zeros -
-                    // the old `reverse_cumsum(grad * output) / input` gave NaN there.
+                    // * before the first zero: every input in the formula is nonzero,
+                    //   so the classic division form is exact;
+                    // * at the first zero: the omitted products keep every other
+                    //   factor, i.e. reverse_cumsum(grad * cumprod(input with that
+                    //   zero replaced by one)) evaluated at that position;
+                    // * after the first zero: every omitted product still contains
+                    //   that zero, so the gradient is exactly zero.
                     //
-                    // tail via the reverse recurrence
-                    //   tail[n-1] = grad[n-1]
-                    //   tail[i]   = grad[i] + input[i+1] * tail[i+1]   (i < n-1)
-                    // evaluated sequentially along `dim`.
+                    // All three regions are computed with cumsum/cumprod and
+                    // elementwise masks, so the whole gradient stays a fixed number
+                    // of parallel kernels, with no data-dependent branching.
 
-                    let shape = input.shape();
-                    let ndims = shape.num_dims();
-                    let n = shape[dim];
-                    let device = grad.device();
+                    let ndims = input.shape().num_dims();
                     let dtype = grad.dtype();
+                    let bool_dtype = get_device_settings::<B>(&grad.device()).bool_dtype;
 
-                    // Slice a single index `i` along `dim`, keeping the dimension.
-                    let slice_at = |tensor: FloatTensor<B>, i: usize| {
+                    let reverse = |tensor: FloatTensor<B>| {
                         let mut slices = vec![Slice::full(); ndims];
-                        slices[dim] = Slice::new(i as isize, Some(i as isize + 1), 1);
+                        slices[dim] = Slice::with_step(0, None, -1);
                         B::float_slice(tensor, &slices)
                     };
+                    let reverse_cumsum =
+                        |tensor: FloatTensor<B>| reverse(B::float_cumsum(reverse(tensor), dim));
 
-                    // left[i] = prod(input[0..i]); left[0] = 1.
-                    // prepend a 1 along `dim`, drop the last element, then inclusive cumprod.
-                    let mut ones_dims: Vec<usize> = (0..ndims).map(|d| shape[d]).collect();
-                    ones_dims[dim] = 1;
-                    let ones = B::float_ones(Shape::from(ones_dims), &device, dtype.into());
-                    let mut head_slices = vec![Slice::full(); ndims];
-                    head_slices[dim] = Slice::new(0, Some(n as isize - 1), 1);
-                    let input_head = B::float_slice(input.clone(), &head_slices);
-                    let shifted = B::float_cat(vec![ones, input_head], dim);
-                    let left = B::float_cumprod(shifted, dim);
+                    // zeros_seen[i] counts zeros in input[0..=i]; comparing against
+                    // 0.5/1.5 sidesteps exact float equality on the running count.
+                    let zero_mask = B::bool_into_float(
+                        B::float_equal_elem(input.clone(), 0.into(), bool_dtype),
+                        dtype.into(),
+                    );
+                    let zeros_seen = B::float_cumsum(zero_mask.clone(), dim);
+                    let before_first = B::bool_into_float(
+                        B::float_lower_elem(zeros_seen.clone(), 0.5.into(), bool_dtype),
+                        dtype.into(),
+                    );
+                    let at_first = B::float_mul(
+                        zero_mask.clone(),
+                        B::bool_into_float(
+                            B::float_lower_elem(zeros_seen, 1.5.into(), bool_dtype),
+                            dtype.into(),
+                        ),
+                    );
 
-                    // tail[i] via the reverse recurrence, collected front-to-back.
-                    let mut tail_parts: Vec<FloatTensor<B>> = Vec::with_capacity(n);
-                    let mut acc = slice_at(grad.clone(), n - 1);
-                    tail_parts.push(acc.clone());
-                    for i in (0..n - 1).rev() {
-                        let grad_i = slice_at(grad.clone(), i);
-                        let input_next = slice_at(input.clone(), i + 1);
-                        acc = B::float_add(grad_i, B::float_mul(input_next, acc));
-                        tail_parts.push(acc.clone());
-                    }
-                    tail_parts.reverse();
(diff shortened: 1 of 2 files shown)


Charles23R:
How does the performance compare with the previous implementation on GPU? This is probably way slower. We should at the very least keep the original, faster implementation if there are no 0s in the input, and keep a note to reimplement with a scan/fold operation eventually like mentioned in #5141

AnayGarodia:
Good call on the perf question — the sequential recurrence was indeed the weak point, and benchmarking it on wgpu/Metal surfaced a second problem: a few hundred chained slice/mul/add ops made the fusion pass fall over (repeated `elemwise_fuse` shader-generation failures, then a stack overflow at n=1024).

So I dropped the recurrence entirely and reworked the gradient into the masked parallel form PyTorch's `cumprod_backward` uses. Each lane is split at its first zero along `dim`:

- **before** the first zero: the classic `reverse_cumsum(grad * output) / input` is exact (every divisor is nonzero),
- **at** the first zero: `reverse_cumsum(grad * cumprod(input with that zero set to 1))` evaluated at that position,
- **after** it: exactly zero.

The regions are computed with `cumsum`/`cumprod` plus elementwise masks, so it's a constant number of kernel launches for every input — no data-dependent branching, no readback, no O(n) loop, and rows with no zeros reduce to the classic formula. Multiple zeros per lane are handled (everything past the first zero is zero).

Perf on an Apple M-series GPU (Metal via wgpu, fwd+bwd+sync, 50 iters — noisy laptop numbers, treat as ballpark):

| shape | old (main) | this branch |
|---|---|---|
| 32x128, no zeros | 5.5 ms | 2.9 ms |
| 32x1024, no zeros | 3.5 ms | 9.7 ms |
| 8x4096, no zeros | 6.1 ms | 9.2 ms |
| 32x1024, with zeros | 6.6 ms (NaN grads) | 3.4 ms (correct) |

Within noise on this machine, both are flat in n; the new form adds ~8 small elementwise kernels of constant overhead. A scan/fold primitive would still be nice for other ops, but this no longer depends on it. All 12 cumprod autodiff tests pass (zero-at-start/middle/end, multiple zeros, checkpointing).

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -2160,33 +2160,78 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                    // replaces exactly that zero with one before the cumprod.
+                    let input_one = B::float_add(input, at_first.clone());
+                    let omitted =
+                        reverse_cumsum(B::float_mul(grad, B::float_cumprod(input_one, dim)));
 
-                    B::float_div(grad_result, input)
+                    B::float_add(
+                        B::float_mul(classic, before_first),
+                        B::float_mul(omitted, at_first),
+                    )
```
laggui: I think we can get away with just one cumprod call by deriving output from output_one instead of computing it separately:

```rust
// Replace ONLY the first zero with 1
let input_one = B::float_add(input.clone(), at_first.clone());
// Compute cumprod only once
let output_one = B::float_cumprod(input_one, dim);
// Derive standard output: output = output_one * before_first
let output = B::float_mul(output_one.clone(), before_first.clone());

// Then, just like your approach:
let input_safe = B::float_add(input, zero_mask);
let classic = B::float_div(
    reverse_cumsum(B::float_mul(grad.clone(), output)),
    input_safe,
);
let omitted = reverse_cumsum(B::float_mul(grad, output_one));

B::float_add(
    B::float_mul(classic, before_first),
    B::float_mul(omitted, at_first),
)
```

This should be faster in your benchmarks than the current approach.

laggui (commented):
Thanks for addressing the initial comments. I think we can still improve the current approach a bit.

crates/burn-autodiff/src/ops/tensor.rs
AnayGarodia: Applied in 97eb3002. `output_one` now feeds both the masked classic output and the omitted-product path, so the backward pass uses one cumprod instead of two. `cargo test -p burn-backend-tests --test autodiff cumprod` passes all 12 base/checkpointing cases, including the zero-position and multiple-zero regressions.

laggui (approved):

PR #5149, opened 2026-07-10
feat: add soft margin loss

Bellman281:
### Checklist

### Related Issues/PRs

No existing issue — adds `SoftMarginLoss` for parity with `torch.nn.SoftMarginLoss`.

### Changes

Adds `SoftMarginLoss`, following the same `Config`-free `forward` /
`forward_no_reduction` pattern as `MseLoss`. Per-element loss is
`log(1 + exp(-target * logit))` = `softplus(-target * logit)`, with the target
tensor expected to hold values in `{-1, 1}`. No new dependencies; backward is
handled by autodiff.

### Testing

`cargo test -p burn-nn soft_margin` — known-answer test covering the un-reduced
output plus `Mean` and `Sum` reductions.

### DIFF
--- a/crates/burn-nn/src/loss/soft_margin.rs
+++ b/crates/burn-nn/src/loss/soft_margin.rs
@@ -8,7 +8,8 @@ use burn::tensor::activation::softplus;
 
 /// Calculate the soft margin loss between the input logits and the target tensor.
 ///
-/// The target values are expected to be `-1` or `1`. The loss for each element is
+/// This criterion is for a two-class (binary) classification task, which is why the
+/// target values are expected to be `-1` or `1`. The loss for each element is
 /// `log(1 + exp(-target * logit))`, matching `torch.nn.SoftMarginLoss`.
 #[derive(Module, Debug)]
 pub struct SoftMarginLoss;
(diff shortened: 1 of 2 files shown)


crates/burn-nn/src/loss/soft_margin.rs
```diff
@@ -0,0 +1,89 @@
+use burn_core as burn;
+
+use crate::loss::reduction::Reduction;
+
+use burn::module::Module;
+use burn::tensor::Tensor;
+use burn::tensor::activation::softplus;
+
+/// Calculate the soft margin loss between the input logits and the target tensor.
```
Charles23R, line 9: I would specify that it is specifically for a two-class classification task (hence the +1 / -1)

Charles23R (dismissed):

Charles23R (approved):

crates/burn-nn/src/loss/soft_margin.rs
Bellman281: @Charles23R 
"Good point — updated the doc to specify this is a two-class (binary) classification criterion, which is why targets are ±1. Thanks!",

Bellman281 (commented):
"Good point — updated the doc to specify this is a two-class (binary) classification criterion, which is why targets are ±1. Thanks!",

PR #5150, opened 2026-07-10
feat: add RReLU activation module

Bellman281:
### Checklist

### Related Issues/PRs

No existing issue — adds RReLU for parity with `torch.nn.RReLU`.

### Changes

Adds the **Randomized Leaky ReLU** (RReLU) activation as an `nn` module
(`RRelu` / `RReluConfig`), following the same train/eval convention as `Dropout`:

- **Training** (autodiff backend): the negative slope is sampled element-wise
  from `Uniform(lower, upper)`.
- **Evaluation**: a fixed midpoint slope `(lower + upper) / 2` is used, identical
  to `LeakyReLU`.
- Defaults: `lower = 1/8`, `upper = 1/3` (matching PyTorch).

Implemented as a module (not a `burn-tensor` functional) because the behaviour is
training-dependent — consistent with how `Dropout` is structured. No new
dependencies; the backward pass is handled by autodiff.

(Note: `RRelu` is an `nn` module like `Dropout`, so there is no functional
activation-table entry in the book to update.)

### Testing

`cargo test -p burn-nn rrelu` — four tests: eval golden values (`= leaky_relu`
midpoint), a training-mode randomness check on an autodiff device, module
display, and bounds validation.


Charles23R (approved):

Bellman281 (commented):
Thanks for the review, @Charles23R I pushed a commit applying cargo fmt, which fixes the failing Format check. Since this is a fork PR the workflows need a maintainer to kick off — could you approve/re-run when you get a chance? 🙏

PR #5153, opened 2026-07-10
8 bit AdamW optimizer

Redhawk18:
### Checklist

### Related Issues/PRs

Decreases memory usage, a major problem in Burn.

### Changes

Adds two impls of 8 bit optimizers based on Tim Dettmers, the tensor operation one is mostly just an artifact of in progress work. The 8 bit kernel is the one that improves results overall. 

### Testing

Unit testing everywhere, then the optimizers were compared directly with [this integration benchmark that also produces figures.](https://github.com/Terascale-All-sensing-Research-Studio/optimizer-benchmark)

Loss for the synthetic benchmark
Ones Loss, Learning Rate: 1e-4
Ones Loss, Learning Rate: 1e-5
Ones Loss, Learning Rate: 1e-6

Wikitext 2 benchmark
WikiText Training Loss
WikiText Training Peak Memory
WikiText Validation Loss

**Table: Comparison of optimizer implementations at learning rate 1e-4: average step time (with standard deviation), optimizer state size, and average process memory.**

| Optimizer | Step time (ms ± µs) | State size (GB) | Process video memory (GB) |
|---|:---:|:---:|:---:|
| AdamW (fp32) | 141 ± 675 | 11.988 | 38.80 |
| 8-bit AdamW (tensor-op) | 1992 ± 8000 | 6.041 | 35.31 |
| 8-bit AdamW (kernel) | 66 ± 650 | 3.044 | 20.88 |


laggui (commented):
Sorry for the delayed response time; this landed while I was on leave :sweat_smile:

I'm glad you managed to reach these gains for your use case!

We should instead pursue a generic approach to quantized optimizers that reuses the existing quantization mechanisms, operates through the `Tensor` API, and avoids requiring custom backend-specific code in `burn-optim` for each optimizer.

Redhawk18:
Based on your commets I'm not even sure you read the code then. There was no code for quantization of optim state, thats why I wrote all of it myself. As seen in my results burn-optim and the tensor api has some design flaws to where this optimization is impossible without the backend traits. I know its unideal, but my results with the Tensor API were terrible.

laggui:
I did read the code. I’m not claiming that the generic infrastructure already exists or that this optimization is currently achievable efficiently through the `Tensor` API. That's why I said we should pursue this generic approach instead, I never claimed it was already available.

The absence of quantized optimizer-state support doesn't make an optimizer-specific CubeCL kernel the right abstraction to upstream into `burn-optim`. We should solve this generically so other optimizers can benefit, rather than maintaining a separate backend-specific implementation for each optimizer.

This is still a useful implementation, and Burn should remain extensible enough to support optimizations like this in user code. That doesn't mean every specialized implementation belongs upstream, is all :slightly_smiling_face: 

The generic approach will require more thought. We need to keep `burn-optim` entirely on the backend-erased `Tensor` API and make optimizer state precision composable at the reusable state-component level. The scope is bigger than the proposed changes here.

PR #5164, opened 2026-07-15
feat(burn-store): add FloatCastAdapter for target-driven float dtype casting

### LINKED ISSUE #5163: burn-store: no adapter can cast a BF16 checkpoint (or any dtype) to the loading module's float dtype
**Feature request / gap**

`HalfPrecisionAdapter` is the only dtype-conversion adapter in burn-store, and it has two properties that make it unable to cover the common "load a stock checkpoint into a reduced-precision model" case:

1. The direction is inferred from the **source** dtype and only handles F32 ↔ F16. A **BF16** checkpoint — the storage dtype of most current model releases — passes through unconverted, and F64/Flex32 likewise.
2. It only converts tensors belonging to a fixed allowlist of `burn-nn` module types. Params of **custom modules** keep the source dtype.

Consequence: loading a BF16 safetensors checkpoint into an f16 model (e.g. `Wgpu<f16>` to halve weight memory) silently leaves the params carrying the source dtype instead of the module's — there's no error at load time, things go wrong downstream. We hit this in practice loading a BF16 diffusion-transformer checkpoint into an f16 module tree and had to write a custom adapter.

**Proposal**

A small *target-driven* adapter alongside `HalfPrecisionAdapter`:

```rust
let adapter = FloatCastAdapter::to(<B::FloatElem as Element>::dtype());
store.with_from_adapter(PyTorchToBurnAdapter.chain(adapter));
```

Every float snapshot (F64/F32/Flex32/F16/BF16) is converted to the given dtype regardless of module type; non-float tensors and already-matching snapshots pass through. It composes with the existing adapters via `chain`.

We've been running exactly this adapter in production for real-model loading (BF16 → F16, ~600 tensors) and it's ~60 lines following `HalfPrecisionAdapter`'s closure idiom. PR incoming.

laurigates:
Closes #5163.

### What

Adds `FloatCastAdapter`, a `ModuleAdapter` that casts **every float snapshot** (F64, F32, Flex32, F16, BF16) to one caller-specified target dtype. Non-float tensors (ints, bools, quantized) and snapshots already in the target dtype pass through unchanged. The constructor validates the target is a float dtype. Re-exported from `burn_store` alongside the existing adapters.

```rust
// load a stock BF16/F32 checkpoint into a model running at the backend's precision
let adapter = PyTorchToBurnAdapter.chain(FloatCastAdapter::to(<B::FloatElem as Element>::dtype()));
store.with_from_adapter(adapter);
```

### Why

`HalfPrecisionAdapter` infers direction from the **source** dtype and only handles F32 ↔ F16, and only for a fixed allowlist of `burn-nn` module types. So the common pairing — a **BF16** checkpoint (the storage dtype of most current model releases) loaded into an **f16** model (e.g. `Wgpu<f16>` to halve weight memory) — silently keeps the source dtype in the loaded params, as do all params of custom modules. Nothing errors at load time; things go wrong downstream. Details in #5163.

`FloatCastAdapter` is *target-driven*, which is the semantics that case needs: "whatever the checkpoint holds, my module runs at dtype X." We have been running exactly this adapter in production for real-model loading (BF16 → F16 across ~600 tensors of a 12B diffusion transformer, plus tiny-fixture parity tests) in [loractl](https://github.com/laurigates/loractl) (`CastFloatsAdapter` there); this PR upstreams it in burn-store style.

### How it was tested

Six new unit tests in `adapter.rs`: BF16→F16 with exact value round-trip, conversion of arbitrary/absent module types (the `HalfPrecisionAdapter` gap), pass-through on already-target dtype, non-float skip (I64), `chain` after `PyTorchToBurnAdapter` (transpose + cast), and `#[should_panic]` on a non-float target.

Ran locally on this branch: `cargo test -p burn-store` (229 lib + 35 doc tests pass), `cargo clippy -p burn-store --all-targets` clean, `cargo fmt` clean. (Scoped to the touched crate rather than the full `cargo run-checks` workspace sweep — happy to run more if CI wants it.)

### Checklist

- [x] Confirmed that `cargo fmt` / clippy / tests pass for the touched crate
- [x] Linked issue: #5163

### DIFF
--- a/crates/burn-store/src/adapter.rs
+++ b/crates/burn-store/src/adapter.rs
@@ -309,6 +309,85 @@ impl ModuleAdapter for HalfPrecisionAdapter {
     }
 }
 
+/// Adapter that casts every float tensor to one target dtype.
+///
+/// Unlike [`HalfPrecisionAdapter`], which infers the direction from the
+/// source dtype (F32 ↔ F16 only) and converts a fixed list of module types,
+/// this adapter is *target-driven*: every float snapshot (F64, F32, Flex32,
+/// F16, BF16) is converted to the given dtype regardless of which module it
+/// belongs to. Non-float tensors (ints, bools, quantized) and tensors already
+/// in the target dtype pass through unchanged.
+///
+/// This is the right tool for loading a stock checkpoint into a module whose
+/// backend runs at a different float precision — e.g. BF16 safetensors into
+/// an F16 model. `HalfPrecisionAdapter` passes BF16 tensors through
+/// unconverted and skips custom module types entirely, so the loaded params
+/// silently keep the source dtype instead of the module's.
+///
+/// # Examples
+///
+/// Cast everything to the backend's float element on load:
+/// ```rust,ignore
+/// let adapter = FloatCastAdapter::to(<B::FloatElem as Element>::dtype());
+/// // store.with_from_adapter(adapter);
+/// ```
+///
+/// Chain after a framework adapter:
+/// ```rust
+/// # use burn_store::{FloatCastAdapter, ModuleAdapter, PyTorchToBurnAdapter};
+/// # use burn_core::tensor::DType;
+/// let adapter = PyTorchToBurnAdapter.chain(FloatCastAdapter::to(DType::F16));
+/// ```
+#[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


Charles23R (approved):
LGTM!

Issue #786, opened 2023-09-08
Fold tensor operation


nathanielsimard:
### Feature description

Support the fold operation. See the PyTorch documentation for full description: https://pytorch.org/docs/stable/generated/torch.nn.Fold.html#torch.nn.Fold

charitarthchugh:
Hi,
I would like to work on this issue, is it still good to work on?

I plan to mirror the semantics of how unfold was implemented in https://github.com/tracel-ai/burn/pull/819

antimora:
Implemented as `fold4d` / `Fold4d` in #5173. Closing.

PR #5173, opened 2026-07-17
feat: add fold4d / Fold4d module (inverse of unfold4d)

Bellman281:
### Checklist

### Related Issues/PRs

### Changes

### Testing


Bellman281:
### Checklist

### Related Issues/PRs

Closes #786.

### Changes

Adds a fold operation (`col2im`), the inverse of the existing `Unfold4d` / `unfold4d`.

- `burn_tensor::module::fold4d(x, output_size, kernel_size, options)` — combines an
  array of sliding local blocks `[N, C * kH * kW, L]` back into a tensor
  `[N, C, H, W]`, summing the values of blocks that overlap. Matches
  `torch.nn.functional.fold`.
- `Fold4d` / `Fold4dConfig` in `burn-nn`, mirroring `Unfold4d`, exposing
  `output_size`, `kernel_size`, `stride`, `padding`, and `dilation`.

**Implementation.** `fold4d` is the adjoint of `unfold4d`. Since `unfold4d` is a
`conv2d` with a one-hot "im2col" kernel, `fold4d` reuses that exact kernel through
`conv_transpose2d`, with `padding_out = (size + 2·pad − dil·(k−1) − 1) mod stride` to
recover the requested `output_size`. It is composed entirely from existing ops, so the
backward pass comes for free through autodiff and no backend / IR / router changes are
needed.

No book change: `Unfold4d` has no entry in the module reference, so there is nothing to
mirror for `Fold4d`.

### Testing

- **Known-answer test:** folds a `[1, 4, 4]` column tensor into a `3×3` output with a
  `2×2` kernel; expected values were computed independently with a numpy `col2im`
  reference.
- **Round-trip test:** verifies `fold4d(unfold4d(x))` equals a numpy-computed reference
  (overlaps summed), confirming `fold` is the true adjoint of `unfold`.
- The `conv_transpose2d` + `padding_out` approach was cross-checked against a direct
  `col2im` implementation across 7 configurations (varying stride, padding, dilation,
  and channels), all matching exactly.
- `cargo run-checks` passes (fmt, clippy, typos, tests).

Bellman281:
Thanks! All CI checks are green (macOS just finishing up), and codecov/patch is 100%. 
@Charles23R @nathanielsimard  whenever you have a moment, this is ready for review — it adds fold4d / Fold4d as the inverse of Unfold4d (closes #786), built as its adjoint via conv_transpose2d so autodiff is free and no backend changes were needed. 🙏

crates/burn-tensor/src/tensor/module.rs
```diff
@@ -266,6 +266,125 @@ pub fn unfold4d(x: Tensor<4>, kernel_size: [usize; 2], options: UnfoldOptions) -
...
+/// * `x` - Input columns of shape
+///   `[batch_size, channels * kernel_size_0 * kernel_size_1, number_of_blocks]`.
+/// * `output_size` - The spatial size `[height, width]` of the folded output tensor.
+/// * `kernel_size` - The size of the sliding blocks.
+/// * `options` - The stride, padding and dilation of the matching unfold.
+///
+/// # Returns
+///
+/// A tensor of shape `[batch_size, channels, output_size_0, output_size_1]`.
+pub fn fold4d(
```
Charles23R, line 287: Following the same pattern as `unfold4d` defined above, this method should just dispatch to the right backend's implementation. 
This code should be in the *backend*'s default implementation, in `crates/burn-backend/src/backend/ops/modules/base.rs`. This allows for different backends to use custom implementations. 
Moreover, this will finally allow us to define the autodiff backend's `fold` and `unfold` operations as complementary of each other.

Charles23R (changes requested):

crates/burn-tensor/src/tensor/module.rs
Bellman281: @Charles23R 
Thanks, that makes sense. I'll move the composed implementation into ModuleOps::fold4d as the default method in burn-backend/.../modules/base.rs and make the public fold4d a thin dispatch, mirroring unfold4d. Since the default composes conv_transpose2d (already first-class), it flows through the router/dispatch/autodiff backends without new IR entries.

One scope question: would you like the explicit autodiff fold/unfold adjoint (fold's backward = unfold and vice-versa) in this same PR, or as a follow-up? The default composed impl is already differentiable through conv_transpose2d, so I'm happy to keep this PR focused on the trait-method refactor and do the adjoint separately — whichever you prefer.

Charles23R: if you want to plug the explicit autodiff adjoint in this PR, go ahead :) Although as you say, the default is already differentiable so it would juste be a bonus.

Bellman281: @Charles23R 

Done — moved the composed impl into ModuleOps::fold4d as a default method and made the public fold4d a thin dispatch, mirroring unfold4d. It reuses the same one-hot weight builder as unfold4d, and since it composes conv_transpose2d it stays differentiable and flows through the router/dispatch/autodiff backends with no new IR entries.

For the explicit autodiff adjoint: I'd like to do it as a focused follow-up rather than in this PR — I noticed burn-autodiff currently has no custom unfold4d op either (there's a // TODO: Support a custom unfold4d operation…), so the clean move is a separate PR that adds custom autodiff ops for both fold and unfold as each other's gradients, with its own tests. Keeps this one to the refactor you asked for. Sound good?

crates/burn-nn/src/modules/fold.rs
```diff
@@ -0,0 +1,168 @@
...
+            input,
+            self.output_size,
+            self.kernel_size,
+            UnfoldOptions::new(self.stride, self.padding, self.dilation),
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
laggui, line 99: I'd also add tests to the `burn-backend-tests/tests/tensor/float/module/` for the functional form. No need to repeat the same valued tests, I'd actually recommend adding non-unit stride and/or non-zero padding.

laggui (commented):
LGTM! Thanks for addressing the requested changes.

Just a minor request before we merge

Bellman281:
Added functional fold4d tests in burn-backend-tests/tests/tensor/float/module/fold4d.rs — a shape check plus two value cases with non-unit stride (and one with non-zero padding), validated against torch.nn.functional.fold. Thanks!

laggui (approved):

PR #5176, opened 2026-07-17
feat: add MultiMarginLoss

Bellman281:
### Checklist

### Related Issues/PRs

Part of the PyTorch loss-parity effort, following `SoftMarginLoss` (#5149),
`MarginRankingLoss` (#5151), `HingeEmbeddingLoss`, and `TripletMarginLoss`.

### Changes

Adds `MultiMarginLoss` / `MultiMarginLossConfig`, matching `torch.nn.MultiMarginLoss`: for
each sample, `(1/C) · Σ_{i≠y} max(0, margin − x[y] + x[i])^p` over classes (defaults
`margin = 1.0`, `p = 1`; `p ∈ {1, 2}`). Uses `gather` for the correct-class score and composes
existing ops, so autodiff is free. Follows the `MarginRankingLoss` config/module pattern; added
to the book's loss table.

### Testing

Known-answer tests for `p = 1` (`forward_no_reduction` + `Mean` + `Sum`) and `p = 2`, verified
against PyTorch reference values, plus a display test. `cargo run-checks` passes.


crates/burn-nn/src/loss/multi_margin.rs
```diff
@@ -0,0 +1,202 @@
...
+        let summed = input
+            .sub(correct)
+            .add_scalar(self.margin)
+            .clamp_min(0.0)
+            .powi_scalar(self.p)
+            .sum_dim(1);
+
+        // The correct class (i == y) always contributes clamp(margin)^p = margin^p (margin >= 0),
+        // so subtract it once to exclude it, then average over the classes.
+        // p is restricted to {1, 2}, so compute margin^p without `f64::powi` (std-only).
```
Charles23R, line 128: I assume this PR was AI-assisted. This is minor, but AIs tend to add a lot of comments that are not really necessary like this last one about std-only. Just something to look out for in the future!

Charles23R (approved):

PR #5177, opened 2026-07-17
fix(store): restore persisted ParamId on load_record and Applier apply (#5130)

### LINKED ISSUE #5130: [🤖] `Module::load_record` discards persisted `ParamId`s, orphaning optimizer state across save/load cycles
**Describe the bug**
The burnpack format persists each tensor's `param_id` — `burn-pack/src/lib.rs` even labels it *"training-state identity"* — but nothing restores it on load: `ModuleRecordMapper` (`burn-core/src/store/mod.rs`, `map_kind!`) keeps the target module's fresh `init()`-minted id and drops the record's id on the floor (its `new()` only keeps `(path, data)`).

Meanwhile `ModuleOptimizer` keeps its per-param states in a `HashMap<ParamId, DynState>`, and neither `load_record` nor `step` ever prunes entries whose id no longer exists in the module.

So for any resume-from-checkpoint across a process boundary (fresh `init()` ⇒ fresh random ids):

1. **Adam moments silently reset** — the loaded states never match any gradient's id, so the first `step` re-creates every state from scratch.
2. **The optimizer record grows without bound** — the orphaned entries are retained and re-serialized alongside the new cohort: **+1 full-model state cohort (2× model size for AdamW) per relaunch**, all of it also materialized on `Device::default()` at every load.

**To Reproduce**
<details> 
  <summary>Reproduction</summary>

- `load_record_discards_param_ids_and_orphans_optimizer_state` — save model+optim after one step, reload into a fresh `init()`, step once: the loaded id differs from the saved one, Adam `time` restarts at 1, and the optimizer record now holds 4 tensors (dead cohort + new) instead of 2. The assertions document the buggy behavior and should fail once fixed.
- `workaround_restoring_param_ids_keeps_optimizer_state_alive` — the user-level fix: read the `param_id`s back out of the record with `burn_pack::Reader` and stamp them onto the loaded module by path with a `ModuleMapper` (`Param::from_mapped_value` keeps value/mapper/`require_grad`). Moments then carry over (`time` 1 → 2) and the record stays at 2 tensors.

```toml
[dependencies.burn]
git = "https://github.com/tracel-ai/burn.git"
rev = "d028234e2ccd75c5ca57d1c71ae74456e0e15c4f" # 2026-06-25 main
default-features = false
features = ["flex", "autodiff", "optim", "std"]

[dependencies.burn-pack]
git = "https://github.com/tracel-ai/burn.git"
rev = "d028234e2ccd75c5ca57d1c71ae74456e0e15c4f"
default-features = false
features = ["std"]
```

```rs
#[cfg(test)]
mod param_id_discarded_on_load_record {
    use burn::module::{Module, ModuleMapper, Param, ParamId};
    use burn::optim::{AdamWConfig, GradientsParams, ModuleOptimizer};
    use burn::prelude::*;
    use burn::store::ModuleRecord;
    use burn::tensor::Bytes;
    use std::collections::HashMap;

    #[derive(Module, Debug)]
    struct Tiny {
        w: Param<Tensor<1>>,
    }

    fn init(device: &Device) -> Tiny {
        Tiny {
            w: Param::from_data([1.0, 2.0], device),
        }
    }

    /// One AdamW step on `loss = sum(w)`, creating/advancing the state for `w`.
    fn step_once(model: Tiny, optim: &mut ModuleOptimizer) -> Tiny {
        let grads = model.w.val().sum().backward();
        let grads = GradientsParams::from_grads(grads, &model);
        optim.step(1e-2, model, grads)
    }

    /// The Adam `time` counter of every per-param state in the optimizer, sorted.
    /// A freshly created state starts at 1; a carried-over state keeps counting.
    fn adamw_times(optim: &ModuleOptimizer) -> Vec<u64> {
        let reader = burn_pack::Reader::from_bytes(optim.into_bytes().unwrap()).unwrap();
        let mut times: Vec<u64> = reader
            .scalars()
            .iter()
            .filter(|(key, _)| key.ends_with(".momentum.time"))
            .map(|(_, scalar)| u64::try_from(*scalar).unwrap())
            .collect();
        times.sort();
        times
    }

    /// BUG. Burnpack records persist each tensor's `param_id` ("training-state
    /// identity" per the format docs), but `load_record` discards them: the loaded
    /// module keeps the ids freshly minted by `init()` (`burn-core store/mod.rs`,
    /// `ModuleRecordMapper` uses the target's `param.id`). `ModuleOptimizer` state is
    /// keyed by `ParamId` and entries are never pruned, so across a checkpoint
    /// save→load ("process relaunch") the entire loaded optimizer state is orphaned:
    /// Adam moments silently reset, and every relaunch appends one dead full-model
    /// state cohort (2× model size for AdamW) to the re-saved optimizer record.
    ///
    /// The assertions document the CURRENT (buggy) behavior — this test should fail
    /// once the persisted ids are restored on load.
    #[test]
    fn load_record_discards_param_ids_and_orphans_optimizer_state() {
        let device = Device::default().autodiff();

        // "Process 1": one training step, then checkpoint model + optimizer.
        let model = init(&device);
        let saved_id = model.w.id;
        let mut optim: ModuleOptimizer = AdamWConfig::new().init();
        let model = step_once(model, &mut optim);
        let model_bytes = model.into_record().into_bytes().unwrap();
        let optim_bytes = optim.into_bytes().unwrap();

        // "Process 2": resume from the records.
        let record = ModuleRecord::from_bytes(model_bytes).unwrap();
        let model = init(&device).load_record(record);
        // The persisted id is discarded — the fresh `init()` id survives instead:
        assert_ne!(saved_id, model.w.id);

        let mut optim = AdamWConfig::new().init().from_bytes(optim_bytes).unwrap();
        let _model = step_once(model, &mut optim);

        // Consequence 1: the loaded state was orphaned — the step created a FRESH
        // state (`time` restarts at 1 next to the dead cohort's 1) instead of
        // continuing the loaded moments at time = 2.
        assert_eq!(vec![1, 1], adamw_times(&optim));
        // Consequence 2: the re-saved record holds the dead cohort + the new one
        // (2 tensors per AdamW state: moment_1 + moment_2).
        assert_eq!(4, optim.to_record().len());
    }

    // ---- user-level fix ------------------------------------------------------------

    /// Stamp the record's persisted ids back onto the loaded module, matched by module
    /// path (the same dotted path the record collector writes). Values, param mappers,
    /// and `require_grad` are untouched.
    fn restore_param_ids<M: Module>(model: M, record_bytes: Bytes) -> M {
        let ids: HashMap<String, ParamId> = burn_pack::Reader::from_bytes(record_bytes)
            .unwrap()
            .into_tensors()
            .unwrap()
            .into_iter()
            .filter_map(|t| t.param_id.map(|id| (t.name, ParamId::from(id))))
            .collect();
        struct Stamper {
            path: Vec<String>,
            ids: HashMap<String, ParamId>,
        }
        impl ModuleMapper for Stamper {
            fn enter_module(&mut self, name: &str, _container_type: &str) {
                self.path.push(name.to_string());
            }
            fn exit_module(&mut self, _name: &str, _container_type: &str) {
                self.path.pop();
            }
            fn map_float<const D: usize>(&mut self, param: Param<Tensor<D>>) -> Param<Tensor<D>> {
                match self.ids.get(&self.path.join(".")) {
                    Some(&id) => {
                        let (_fresh_id, value, mapper) = param.consume();
                        Param::from_mapped_value(id, value, mapper)
                    }
                    None => param,
                }
            }
        }
        model.map(&mut Stamper {
            path: Vec::new(),
            ids,
        })
    }

    /// USER-LEVEL FIX: same relaunch cycle as above, but the persisted ids are read
    /// back out of the record (burn-pack keeps them) and stamped onto the loaded
    /// module. The optimizer state stays associated: moments carry over (time 1 → 2)
    /// and the record does not grow.
    #[test]
    fn workaround_restoring_param_ids_keeps_optimizer_state_alive() {
        let device = Device::default().autodiff();

        // "Process 1": identical checkpoint.
        let model = init(&device);
        let saved_id = model.w.id;
        let mut optim: ModuleOptimizer = AdamWConfig::new().init();
        let model = step_once(model, &mut optim);
        let model_bytes = model.into_record().into_bytes().unwrap();
        let optim_bytes = optim.into_bytes().unwrap();

        // "Process 2": resume, then restore the persisted ids.
        let record = ModuleRecord::from_bytes(model_bytes.clone()).unwrap();
        let model = init(&device).load_record(record);
        let model = restore_param_ids(model, model_bytes);
        assert_eq!(saved_id, model.w.id);

        let mut optim = AdamWConfig::new().init().from_bytes(optim_bytes).unwrap();
        let _model = step_once(model, &mut optim);

        // The loaded moments carried over, and nothing accreted.
        assert_eq!(vec![2], adamw_times(&optim));
        assert_eq!(2, optim.to_record().len());
    }
}
```

</details>

**Expected behavior**
Loading a module record restores each parameter's persisted `ParamId`, so that resuming from a checkpoint in a new process leaves the ParamId-keyed optimizer state associated with its parameters — Adam moments continue instead of resetting, and the re-saved optimizer record keeps its size instead of accreting a dead state cohort per relaunch.

**Additional context**
- *(Analysis and reproduction prepared by Claude).*
- The first test passing asserts the incorrect behavior.
- Observed in the wild (training with a periodic-restart loop): an optimizer record at **1.13 GB = 8 dead AdamW cohorts for a 70 MB model**, growing by exactly `2 × model_size` (+ ~13 KB metadata) at each restart boundary.
- Suggested fix (Claude): Have `ModuleRecordMapper` (and the `burn-store` `Applier`) pass the record's persisted id to `Param::transform_for_load` instead of the target's fresh id. Additionally/alternatively, `ModuleOptimizer::load_record` could drop (or at least warn about) entries whose `ParamId` matches no parameter once the module is known.

jaweed3:
## Checklist

- [x] Confirmed that `cargo test -p burn-core -p burn-store` passes.
- [ ] The book does not need changes for this fix.

## Related Issues

Closes #5130.

## Changes

**`burn-core/src/store/mod.rs`**: `ModuleRecordMapper` now stores `HashMap<String, (ParamId, TensorData)>` instead of discarding the id. `map_kind!` passes the record's `ParamId` to `transform_for_load` instead of the target's fresh `param.id`.

**`burn-store/src/applier.rs`**: `apply_tensor` returns the persisted `ParamId` from `snapshot.tensor_id`. `map_float`/`map_int`/`map_bool` use it instead of `param.id`, so optimizer state keyed by `ParamId` survives save/load cycles.

## Testing

- `load_record_preserves_param_id` — round-trip save/load: loaded `ParamId`s equal originals.
- `applier_preserves_param_id` — snapshot with known id → `Applier::map_float` → loaded param carries it.
- All 234 tests pass across both crates.

### DIFF
--- a/crates/burn-core/src/store/mod.rs
+++ b/crates/burn-core/src/store/mod.rs
@@ -275,10 +275,13 @@ impl ModuleVisitor for Collector {
     }
 }
 
-/// Mapper that loads recorded tensors back onto matching parameters by module path.
+/// Mapper that loads recorded tensors back onto matching parameters by module path,
+/// restoring the persisted [`ParamId`] so optimizer state (keyed by id) survives
+/// save/load cycles.
 struct ModuleRecordMapper {
     path: Vec<String>,
-    tensors: HashMap<String, TensorData>,
+    /// Map from module path to (persisted ParamId, tensor data).
+    tensors: HashMap<String, (ParamId, TensorData)>,
     dtype_policy: DTypePolicy,
     missing: Vec<String>,
     errors: Vec<String>,
@@ -289,7 +292,7 @@ impl ModuleRecordMapper {
         let tensors = record
             .tensors
             .into_iter()
-            .map(|t| (t.path, t.data))
+            .map(|t| (t.path, (t.id, t.data)))
             .collect();
         Self {
             path: Vec::new(),
@@ -303,6 +306,8 @@ impl ModuleRecordMapper {
     /// Look up the recorded tensor for the current path and build the tensor to load,
     /// or `None` (recording it as missing / errored) to leave the parameter unchanged.
     ///
+    /// Returns the tensor and the persisted [`ParamId`] on a hit.
+    ///
     /// `module_dtype` is only evaluated on a hit under [`DTypePolicy::CastToModule`], so a
     /// missing parameter never materializes its current value just to read a dtype.
     fn take<const D: usize, K: Basic>(
@@ -310,10 +315,10 @@ impl ModuleRecordMapper {
         device: &Device,
         target_shape: Shape,
         module_dtype: impl FnOnce() -> DType,
-    ) -> Option<Tensor<D, K>> {
+    ) -> Option<(Tensor<D, K>, ParamId)> {
         let path = self.path.join(".");
-        let data = match self.tensors.remove_entry(&path) {
-            Some(data) => data.1,
+        let (id, data) = match self.tensors.remove_entry(&path) {
+            Some(entry) => entry.1,
             None => {
                 self.missing.push(path);
                 return None;
@@ -334,7 +339,7 @@ impl ModuleRecordMapper {
             return None;
         }
 
-        Some(Tensor::from_data(data, (device, dtype)))
+        Some((Tensor::from_data(data, (device, dtype)), id))
     }
 }
 
@@ -347,11 +352,10 @@ macro_rules! map_kind {
             &mut self,
             param: Param<Tensor<D, $kind>>,
         ) -> Param<Tensor<D, $kind>> {
-            let id = param.id;
             let device = param.lazy_device();
             let shape = param.lazy_shape();
             match self.take(&device, shape, || param.val().dtype()) {
-                Some(tensor) => param.transform_for_load(tensor, id),
+                Some((tensor, record_id)) => param.transform_for_load(tensor, record_id),
                 None => param,
             }
         }
@@ -530,6 +534,33 @@ mod tests {
         // `bias` shape mismatch is reported as a validation error by default.
         let result = tiny_wrong_bias_shape(&device).try_load_record(record);
         assert!(matches!(result, Err(RecordError::Validation(_))));
+    }
+
+    #[test]
+    fn load_record_preserves_param_id() {
+        let device = Default::default();
+        let model = Tiny::new([[1.0, 2.0], [3.0, 4.0]], [5.0, 6.0], &device);
+
+        // Capture original ParamIds before saving.
+        let weight_id = model.weight.id;
+        let bias_id = model.bias.id;
+
+        let bytes = model.into_record().into_bytes().unwrap();
+        let record = ModuleRecord::from_bytes(bytes).unwrap();
+
+        // Load into a fresh model (new init() = different ParamIds).
+        let loaded = Tiny::new([[0.0; 2]; 2], [0.0; 2], &device).load_record(record);
+
+        // The loaded model should have the ORIGINAL ParamIds from the record,
+        // not the fresh ones from init().
+        assert_eq!(
+            loaded.weight.id, weight_id,
+            "weight ParamId should be restored from record"
+        );
+        assert_eq!(
+            loaded.bias.id, bias_id,
+            "bias ParamId should be restored from record"
+        );
     }
 
     #[test]
--- a/crates/burn-store/src/applier.rs
+++ b/crates/burn-store/src/applier.rs
@@ -286,9 +286,7 @@ impl ModuleMapper for Applier {
 
         // Try to apply snapshot with shape validation
         match self.apply_tensor(&target_device, target_shape) {
-            Some((tensor, snapshot_id)) => {
-                param.transform_for_load(tensor, snapshot_id)
-            }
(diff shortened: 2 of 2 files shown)


Charles23R (approved):
Nice catch, thanks!

PR #5182, opened 2026-07-17
feat: add GaussianNLLLoss

Bellman281:
### Checklist

### Related Issues/PRs

Part of the PyTorch loss-parity effort, following `SoftMarginLoss` (#5149),
`MarginRankingLoss` (#5151), `HingeEmbeddingLoss`, `TripletMarginLoss`, and `MultiMarginLoss`.

### Changes

Adds `GaussianNLLLoss` / `GaussianNLLLossConfig`, matching `torch.nn.GaussianNLLLoss`:
`0.5 · (log(var) + (input − target)² / var)` element-wise, with `var` clamped to `eps`
(default `1e-6`) and an optional `full` term `0.5·log(2π)` (default off). Composed from existing
ops so autodiff is free. Follows the `MarginRankingLoss` config/module pattern; added to the
book's loss table.

### Testing

Known-answer tests for the base and `full = true` cases (`forward_no_reduction` + `Mean` + `Sum`)
verified against PyTorch reference values, plus a display test. `cargo run-checks` passes.


Charles23R (approved):

crates/burn-nn/src/loss/gaussian_nll.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// - var:    `[...dims]` (predicted variance, positive)
+    /// - output: `[...dims]`
+    pub fn forward_no_reduction<const D: usize>(
+        &self,
+        input: Tensor<D>,
+        target: Tensor<D>,
+        var: Tensor<D>,
+    ) -> Tensor<D> {
+        // Clamp the variance for numerical stability.
+        let var = var.clamp_min(self.eps);
```
laggui, line 110: Since values should always be clamped to `eps`, we could add a small test to validate that. For example compute the loss for a var smaller than eps, and check that the result is the same as with var = eps.

Otherwise LGTM

Bellman281: @laggui 
Added test_gaussian_nll_loss_clamps_var_to_eps — passes a variance below eps and checks the loss matches var = eps, validating the clamp. Thanks!

laggui (approved):

PR #5191, opened 2026-07-18
fix(cosine): remove erroneous restart, match PyTorch CosineAnnealingLR behavior

### LINKED ISSUE #5128: CosineAnnealingLrScheduler displays non-standard behavior
I'm using the following setup:

```
    let cosine_config = CosineAnnealingLrSchedulerConfig::new(max_lr, 250)
        .with_min_lr(min_lr);
```
where ``max_lr = 1e-3`` and ``min_lr = 1e-6``

And get the following recurring half-cosine LR waveform:

Image

For pytorch, documentation suggests that once the cosine wave reaches its minima, the LR remains there for the duration (i.e. without restarting the cycle - although I'm unsure what happens in the case where the minima approaches 0.0 which would be an unfortunate place to continue training, as it would be for Burn without configuring a .with_min_lr(), as the default minima would be 0.0 also).

Burn's current implementation which cycles a half cosine wave between max and min does not seem like an acceptable or expected outcome, and clearly could contribute to training instability.

DeathSurfing:
## Summary

Fixes #5128

The `CosineAnnealingLrScheduler` was wrapping its iteration counter with `% (self.num_iters + 1)`, causing the learning rate to abruptly jump back to `max_lr` after every `num_iters + 1` steps. This created an erroneous half-cosine restarting waveform instead of the standard cosine annealing schedule.

## Changes

- **Line 92:** Removed the `% (self.num_iters + 1)` wrapping from `step()`. The `current_iter` now increases monotonically (via `wrapping_add(1)`), matching PyTorch's `CosineAnnealingLR` which explicitly states *"performs cosine annealing without restarts, so `T_cur = t` and increases monotonically with each call to `step()`"*
- **Doc updates:** Config and struct docs now describe the scheduler as performing cosine annealing **without** warm restarts, with the iteration counter increasing monotonically
- **Test update:** `test_lr_change` now verifies a full 5-step smooth cosine curve rather than the previous 4-step sequence that expected an artificial restart

## Verification

- All 7 cosine tests pass (`cargo test -p burn-optim -- lr_scheduler::cosine`)
- All 95 burn-optim tests pass (`cargo test -p burn-optim`)
- The fix is a one-line code change (removing `% (self.num_iters + 1)`); the rest is documentation and test updates

## Reference

[PyTorch CosineAnnealingLR docs](https://pytorch.org/docs/2.12/generated/torch.optim.lr_scheduler.CosineAnnealingLR.html)

### DIFF
--- a/crates/burn-optim/src/lr_scheduler/cosine.rs
+++ b/crates/burn-optim/src/lr_scheduler/cosine.rs
@@ -6,12 +6,17 @@ use crate::RecordState;
 use crate::lr_scheduler::module_lr_scheduler::ModuleLrScheduler;
 use burn::config::Config;
 
-/// The configuration for creating a [Cosine Annealing learning rate scheduler with warm
-/// restarts](CosineAnnealingLrScheduler).
+/// The configuration for creating a [Cosine Annealing learning rate
+/// scheduler](CosineAnnealingLrScheduler).
 ///
-/// This scheduler returns the learning rate `initial_lr` at the first step, then changes it by
-/// following a cosine function. After `num_iters` iterations, the learning rate is reset to
-/// `initial_lr`.
+/// This scheduler uses a cosine annealing schedule without warm restarts,
+/// where the learning rate follows a cosine curve from `initial_lr` down to
+/// `min_lr` over `num_iters` steps. After `num_iters` steps, the learning rate
+/// continues along the cosine curve without restarting.
+///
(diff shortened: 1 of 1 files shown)


laggui (approved):
Thanks for the fix, this appears to be the correct SGDR implementation without restarts.

We could re-introduce a *correct* SGDR with restarts via `CosineAnnealingWarmRestartsLrScheduler` (bit of a mouthful 😅) but we would have t o correctly handle `T0` and `Tmult` (which the previous version didn't).

Issue #5195, opened 2026-07-19
scatter_nd fails on non-trivial indices


artyompal:
**Describe the bug**

The CubeCL `scatter_nd` kernel computes the `values` offset incorrectly when updates are spread across multiple batch dimensions. It uses:

```rust
val_offset = update_idx * values.stride(0)
// then trailing dims with values.stride(1 + dim)
```

That only works if all updates vary along dim 0. When `values.shape[0] == 1` (common in ONNX ScatterND layouts), `stride(0)` equals the full tensor volume, so `update_idx >= 1` reads out of bounds. On CUDA this can surface as `CUDA_ERROR_ILLEGAL_ADDRESS`.

**To Reproduce**

1. Build Burn with a CubeCL backend (CUDA or ROCm).
2. Run a `scatter_nd` where values have a leading batch dim of size 1, e.g.:
   - `data`: `[1, 2, 4, 2]`
   - `indices`: `[1, 2, 4, 1, 4]` (K=4)
   - `values`: `[1, 2, 4, 1]`
3. Observe OOB validation / illegal address, or wrong output vs NdArray.

**Expected behavior**

`scatter_nd` should match the CPU backends: map `update_idx` across all batch dims of `values` (not only dim 0), then apply `slice_offset` over trailing slice dims using `values.stride(dim)`. No out-of-bounds reads when `shape[0] == 1`.

**Screenshots**

N/A

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: N/A
 - Version: 0.21, 0.22-pre.

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

**Additional context**

- Hit while running a transformer ONNX model on CUDA (`ScatterND` with values shaped like `[1, 4, 2050, 258, 1]`).
- Flex/NdArray use `val_offset = n * slice_size` (contiguous layout); the GPU kernel must be stride-aware and decompose the full batch index.

nathanielsimard:
I think this is fixed right? @artyompal @laggui

laggui:
Yup, just merged

PR #5196, opened 2026-07-19
Fix scatter_nd values offset for multi-dim batches (#5195)

artyompal:
This PR fixes scatter_nd with non-trivial and non-contiguous indices.

### Checklist

### Related Issues/PRs

https://github.com/tracel-ai/burn/issues/5195

### Changes

The kernel indexed values as `update_idx * values.stride(0)` and then walked trailing dims with `stride(1 + dim)`. That only works when all updates vary along dim 0. When `values.shape[0] == 1` (common in ONNX ScatterND layouts), stride(0) is the full tensor volume, so `update_idx >= 1` reads out of bounds.

Instead I decomposed `update_idx` right-to-left over all batch dims and `slice_offset` over the trailing slice dims, using `values.stride(dim)` for each.

### Testing

I've added a test `crates/burn-backend-tests/tests/cubecl/scatter_nd.rs`. It tests both contiguous and non-contiguous indices.

### DIFF
--- a/crates/burn-cubecl/src/kernel/index/scatter_nd.rs
+++ b/crates/burn-cubecl/src/kernel/index/scatter_nd.rs
@@ -54,16 +54,38 @@ fn scatter_nd_kernel<T: Numeric, I: Int, Op: BinaryOpFamily>(
         data_slice_offset += coord * data.stride(k + dim);
     }
 
-    // Decompose slice_offset over values' dims (1..n_v), with update_idx for dim 0
+    // values_shape covers values dims 1..rank.
+    // Decompose update_idx over batch dims (right-to-left), then slice_offset over
+    // trailing slice dims. Never use `update_idx * values.stride(0)` alone — when
+    // shape[0] == 1 that stride is the full volume and OOB-reads values.
     let val_rank = values_shape.len().comptime();
-    let mut val_offset = update_idx * values.stride(0);
+    let values_rank = val_rank + 1;
+    let batch_rank = values_rank - slice_rank;
+
+    let mut val_offset = 0usize;
+    let mut remainder_u = update_idx;
+    #[unroll]
+    for i in 0..batch_rank {
+        let dim = batch_rank - i - 1;
+        if dim == 0 {
+            let shape0 = values.shape(0);
+            let coord = remainder_u % shape0;
+            remainder_u = remainder_u / shape0;
+            val_offset += coord * values.stride(0);
+        } else {
+            let (rem, coord) = values_shape[dim - 1].div_mod(remainder_u);
+            remainder_u = rem;
+            val_offset += coord * values.stride(dim);
+        }
+    }
+
     let mut remainder_v = slice_offset;
     #[unroll]
-    for i in 0..val_rank {
-        let dim = val_rank - i - 1;
-        let (rem, coord) = values_shape[dim].div_mod(remainder_v);
+    for i in 0..slice_rank {
+        let dim = values_rank - i - 1;
+        let (rem, coord) = values_shape[dim - 1].div_mod(remainder_v);
         remainder_v = rem;
-        val_offset += coord * values.stride(1 + dim);
+        val_offset += coord * values.stride(dim);
     }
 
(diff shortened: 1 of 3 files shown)


crates/burn-cubecl/src/kernel/index/scatter_nd.rs
```diff
@@ -111,7 +133,7 @@ pub(crate) fn scatter_nd<R: CubeRuntime>(
     };
 
     let data_slice_shape = shape_divmod_range(&tensor, k..data_shape.num_dims());
-    // values dims 1.. (skip the num_updates leading dim)
+    // values dims 1.. (dim 0 is read from values.shape(0) in the kernel)
     let values_slice_shape = shape_divmod_range(&values, 1..values.meta.shape.num_dims());
```
laggui: First, we should pass all dimensions of values so `FastDivmod` is available for dim 0 too

```rust
let values_shape = shape_divmod_range(&values, 0..values.meta.shape.num_dims());
```

crates/burn-cubecl/src/kernel/index/scatter_nd.rs
```diff
@@ -54,16 +54,38 @@ fn scatter_nd_kernel<T: Numeric, I: Int, Op: BinaryOpFamily>(
...
-    for i in 0..val_rank {
-        let dim = val_rank - i - 1;
-        let (rem, coord) = values_shape[dim].div_mod(remainder_v);
+    for i in 0..slice_rank {
+        let dim = values_rank - i - 1;
+        let (rem, coord) = values_shape[dim - 1].div_mod(remainder_v);
         remainder_v = rem;
-        val_offset += coord * values.stride(1 + dim);
+        val_offset += coord * values.stride(dim);
     }
```
laggui, line 79: Then, in the kernel we can decompose `update_idx` over batch dimensions and `slice_offset` over trailing slice dimensions:

```rust

    // Decompose slice_offset over data's trailing dims (k..n) - unchanged
    let slice_rank = data_slice_shape.len().comptime();
    let mut data_slice_offset = 0usize;
    let mut remainder = slice_offset;
    #[unroll]
    for i in 0..slice_rank {
        let dim = slice_rank - i - 1;
        let (rem, coord) = data_slice_shape[dim].div_mod(remainder);
        remainder = rem;
        data_slice_offset += coord * data.stride(k + dim);
    }

    // values_shape covers all values dims (0..values_rank).
    let values_rank = values_shape.len().comptime();
    let batch_rank = values_rank - slice_rank;

    let mut val_offset = 0usize;
    let mut remainder_u = update_idx;
    #[unroll]
    for i in 0..batch_rank {
        let dim = batch_rank - i - 1;
        let (rem, coord) = values_shape[dim].div_mod(remainder_u);
        remainder_u = rem;
        val_offset += coord * values.stride(dim);
    }

    let mut remainder_v = slice_offset;
    #[unroll]
    for i in 0..slice_rank {
        let dim = slice_rank - i - 1;
        let val_dim = batch_rank + dim;
        let (rem, coord) = values_shape[val_dim].div_mod(remainder_v);
        remainder_v = rem;
        val_offset += coord * values.stride(val_dim);
    }
```

laggui (commented):
I think we can avoid this branching patch if we actually generalize the original implementation instead.

crates/burn-cubecl/src/kernel/index/scatter_nd.rs
artyompal: Done

crates/burn-cubecl/src/kernel/index/scatter_nd.rs
artyompal: Done

artyompal:
What guarantees are there for input arguments? I saw that indices can be non-contiguous, but can values be also non-contiguous?

laggui (approved):

laggui:
> What guarantees are there for input arguments? I saw that indices can be non-contiguous, but can values be also non-contiguous?

Yeah values can also be non-contiguous. The kernel should handle that as it explicitly uses the strides for every dimension to compute offsets.

PR #5198, opened 2026-07-20
chore: change loop logic for data iterators and batchers

Zatiji:
### Checklist

### Related Issues/PRs

_No issues_

### Changes

1. **`Dataset::get` now returns `Result<I, E>` instead of silently swallowing errors.** : The `Dataset<I, E = DatasetError>` trait's `get` method now has the signature `get() -> Result<I, E>`, replacing the previous pattern where implementations`.unwrap()`'d internally and could panic with no context, or in some cases silently discarded real errors. `Err` is reserved for genuine retrieval failures on an in-bounds index (I/O errors, deserialization failures, corrupt rows, etc). Out-of-bounds indices still **panic**, matching `Vec`/slice indexing conventions, since a dataset is expected to know its own length.

2. **Simplified the batching loop logic in `BatchDataloaderIterator::next`** (`crates/burn-core/src/data/dataloader/batch.rs`):  now that `Dataset::get` has a real `Result`-based contract instead of the previous `Option`-based one.

3. **Fixed a class of silent, permanent hangs in the multi-threaded/multi-device orchestration code.**: Several places in the training/data-loading pipeline wait for a *fixed count* of completion signals from worker threads before continuing. Previously, if a worker thread panicked (e.g. on a real error now surfaced by the `Dataset::get` change above) *before* sending its completion signal, the orchestrator would block on `recv()`/`join()` forever with no error ever surfaced to the user. Fixed in three places by catching the panic on the worker side and explicitly propagating it as an error message/result instead of letting the worker die silently:

   - `MultiThreadDataLoader`'s worker pool: workers now catch panics from `Dataset::get` and send an explicit `Message::Error`, which the orchestrator turns into an immediate, clear panic instead of hanging on `Message::Done` counting. 
   
   - `MultiDevicesTrainStep`: per-device worker threads now catch panics from `TrainStep::step` and report them via a `WorkerMessage::Error`, instead of leaving the orchestrator's `recv()` loop blocked on an output that will never arrive.
   
    - `DdpTrainingStrategy::fit`'s worker join logic: replaced the sequential pattern with a concurrent join. One small "reaper" thread per worker forwards its `.join()` result over a channel, and the orchestrator reacts to the first result to arrive, regardless of order. 

**Other information**:

_**Known follow-up, not addressed in this PR**: the DDP strategy fix above only stops the *orchestrator* thread from hanging silently. It does not stop a live-but-innocent peer device thread from getting permanently stuck inside `DistributedSyncClient::submit_sync_collective`'s `rx.recv()` waiting for a collective-sync contribution from a device that already died. That stuck thread is currently only reaped because the orchestrator's panic tears down the whole process; in a context where this code isn't running on the process's main thread, it would leak forever even after the error is reported._

_The root cause is that all devices share clones of a single `mpsc::Sender` funneling into one server thread (`DistributedSyncClient::new` / `DistributedSyncServer`), so the server has no way to distinguish "a device is just slow" from "a device is dead"_

### Testing

- ran `cargo test`


crates/burn-core/src/data/dataloader/multithread.rs
```diff
@@ -194,13 +208,29 @@ where
...
-                            while let Some(item) = iterator.next() {
-                                let progress = iterator.progress();
-
-                                // Consumer dropped the receiver: abandon this pass,
-                                // don't terminate the worker.
-                                if sender.send(Message::Batch(index, item, progress)).is_err() {
-                                    break;
+                            loop {
+                                let next =
+                                    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
Charles23R: We should just return a result in the iterator instead of unwrap()-ing directly.

crates/burn-core/src/data/dataloader/batch.rs
```diff
@@ -161,8 +165,11 @@ impl<I, O> Iterator for BatchDataloaderIterator<I, O> {
     type Item = O;
 
     fn next(&mut self) -> Option<O> {
-        while let Some(item) = self.dataset.get(self.current_index) {
+        while self.current_index < self.len {
+            let index = self.current_index;
```
Charles23R: Nitpick, but this intermediate variable isn't really useful

crates/burn-dataset/src/dataset/dataframe.rs
```diff
@@ -82,14 +82,24 @@ where
...
     ///
-    /// # Returns
+    /// # Panics
     ///
-    /// An Option containing the item if it exists, or None if it doesn't
-    fn get(&self, index: usize) -> Option<I> {
-        let row = self.df.get_row(index).ok()?;
+    /// Panics if `index >= len()`.
+    fn get(&self, index: usize) -> Result<I, DataframeDatasetError> {
+        if index >= self.len {
```
Charles23R: The assert! macro does exactly this

crates/burn-dataset/src/dataset/sqlite.rs
```diff
@@ -192,40 +197,44 @@ impl<I> SqliteDataset<I> {
...
     I: Clone + Send + Sync + DeserializeOwned,
 {
     /// Get an item from the dataset.
-    fn get(&self, index: usize) -> Option<I> {
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index >= len()`.
+    fn get(&self, index: usize) -> Result<I> {
+        if index >= self.len {
```
Charles23R: assert! here too

crates/burn-dataset/src/transform/partial.rs
```diff
@@ -82,15 +83,19 @@ where
...
 where
-    D: Dataset<I>,
+    D: Dataset<I, E>,
     I: Clone + Send + Sync,
+    E: Error + Send + Sync + 'static,
 {
-    fn get(&self, index: usize) -> Option<I> {
+    fn get(&self, index: usize) -> Result<I, E> {
         let index = index + self.start_index;
         if index < self.start_index || index >= self.end_index {
```
Charles23R: assert!

crates/burn-dataset/src/transform/sampler.rs
```diff
@@ -291,9 +292,12 @@ where
     D: Dataset<I>,
     I: Send + Sync,
 {
-    fn get(&self, index: usize) -> Option<I> {
+    fn get(&self, index: usize) -> Result<I, DatasetError> {
         if index >= self.size {
```
Charles23R: assert!

crates/burn-dataset/src/transform/window.rs
```diff
@@ -1,22 +1,25 @@
...
     /// # Returns
     ///
-    /// A `Vec<I>` representing the window.
+    /// A `Vec<I>` representing the window, or `None` if the window doesn't fit within the
+    /// collection.
     fn window(&self, current: usize, size: NonZeroUsize) -> Option<Vec<I>>;
 }
 
 impl<I, T: Dataset<I> + ?Sized> Window<I> for T {
     fn window(&self, current: usize, size: NonZeroUsize) -> Option<Vec<I>> {
```
Charles23R: We could propagate the error here

crates/burn-dataset/src/transform/window.rs
```diff
@@ -144,8 +160,14 @@ where
     /// # Returns
     ///
     /// A vector representing the window.
```
Charles23R, line 165: We should document the panic! especially as it is a breaking change

Charles23R (changes requested):
Minor stuff. Also, We should test a whole training loop with the examples, and DDP.

crates/burn-core/src/data/dataloader/base.rs
```diff
@@ -21,6 +22,14 @@ pub struct Progress {
 pub trait DataLoaderIterator<O>: Iterator<Item = O> {
     /// Returns the progress of the data loader.
     fn progress(&self) -> Progress;
+
+    /// Pulls the next item, surfacing a dataset retrieval failure as an `Err` instead of
+    /// panicking. Implementations that can't fail on their own (e.g. iterators that only
+    /// relay already-built batches) can rely on the default, which just delegates to
+    /// [`Iterator::next`].
+    fn try_next(&mut self) -> Result<Option<O>, DatasetError> {
```
Charles23R: What i meant was that the `Iterator`'s `Item` could be `Result<O, DatasetError>` (So next() returns an `Option<Result<O, DatasetError>>`). This would eliminate the need for try_next, the dataset errors would surface at the iterator level and could be handled by the user.

crates/burn-dataset/src/transform/window.rs
Charles23R: I think we can switch the order here too : `Result<Option<Vec<I>>, DatasetError>` => `Option<Result<Vec<I>, DatasetError>>`

crates/burn-dataset/src/transform/window.rs
```diff
@@ -71,20 +78,34 @@ impl<'a, I> WindowsIterator<'a, I> {
...
+            if let Some(window) = self.dataset.window(self.current, self.size)? {
+                return Ok(Some(window));
+            }
+            self.current += 1;
+        }
+        Ok(None)
+    }
 }
 
 impl<I> Iterator for WindowsIterator<'_, I> {
```
Charles23R, line 92: And here, wrap I in a result instead if the `try_next`

Charles23R (changes requested):

Charles23R (approved):

PR #5201, opened 2026-07-20
feat: add functional scaled_dot_product_attention

Bellman281:
### Checklist

### Related Issues/PRs

None — standalone addition. It is the first of the upstream primitives needed for a native Burn SAM2/SAM3 inference port (functional attention over caller-supplied q/k/v).

### Changes

Adds a functional `scaled_dot_product_attention(query, key, value, attn_mask, is_causal, scale)` to `burn-nn`, mirroring [`torch.nn.functional.scaled_dot_product_attention`](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html).

Burn already ships `MultiHeadAttention` and `CrossAttention`, but there is no standalone functional SDPA over caller-supplied q/k/v. Models that build their own q/k/v — e.g. RoPE-rotated or bias-added attention — call attention functionally and currently have to re-derive `softmax(QKᵀ·scale + mask)·V` by hand.

The implementation follows Burn's existing attention idiom (`matmul` · scale · optional mask · `softmax(dim=-1)` · `matmul`), is generic over tensor rank `D`, and supports:

- an optional **additive** `attn_mask` (added to the scores before softmax),
- `is_causal` (applies a lower-triangular causal mask via `triu_mask`),
- an optional `scale` (defaults to `1 / sqrt(d_k)`).

Design note: `attn_mask` is additive (matching PyTorch's float-mask path) rather than boolean, to avoid clashing with the "`true` = masked" boolean convention used in `MultiHeadAttention`. Happy to switch to a boolean mask if preferred for consistency.

### Testing

Added three unit tests in `sdpa.rs`, each validated against `torch.nn.functional.scaled_dot_product_attention` on identical inputs:

- `sdpa_no_mask` — plain attention (reference `[[2, 3], [2, 3]]`),
- `sdpa_is_causal` — causal masking (reference `[[1, 2], [2, 3]]`),
- `sdpa_additive_mask` — additive `-inf` mask (reference `[[1, 2], [2, 3]]`).

`cargo test -p burn-nn scaled_dot_product` passes; `cargo run-checks` (fmt / clippy / tests) is clean.

### DIFF
--- a/crates/burn-nn/src/modules/attention/sdpa.rs
+++ b/crates/burn-nn/src/modules/attention/sdpa.rs
@@ -2,6 +2,7 @@ use burn_core as burn;
 
 use burn::tensor::activation::softmax;
 use burn::tensor::{Bool, Tensor};
+use num_traits::Float as _;
 
 /// Functional scaled dot-product attention, mirroring
 /// [`torch.nn.functional.scaled_dot_product_attention`](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html).
@@ -36,7 +37,7 @@ pub fn scaled_dot_product_attention<const D: usize>(
     scale: Option<f64>,
 ) -> Tensor<D> {
     let d_k = query.dims()[D - 1];
-    let scale = scale.unwrap_or(1.0 / (d_k as f64).sqrt());
+    let scale = scale.unwrap_or_else(|| 1.0 / (d_k as f64).sqrt());
 
     // [..., seq_len_q, seq_len_kv]
     let scores = query.matmul(key.transpose()).mul_scalar(scale);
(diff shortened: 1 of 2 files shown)


crates/burn-nn/src/modules/attention/sdpa.rs
```diff
@@ -0,0 +1,121 @@
...
+        // i.e. the future positions a causal mask must block.
+        let mask = Tensor::<2, Bool>::tril_mask([seq_q, seq_kv], 0, &scores.device());
+        scores.mask_fill(mask.unsqueeze::<D>(), f32::NEG_INFINITY)
+    } else if let Some(mask) = attn_mask {
+        scores.add(mask)
+    } else {
+        scores
+    };
+
+    softmax(scores, D - 1).matmul(value)
```
nathanielsimard, line 62: This should use the backend's attention operation rather than reimplementing it with basic tensor ops. I also don't think it belongs in this crate, the functional operations are re-exported from the tensor crate under `burn::tensor::module::attention`, and they don't need the neural network utilities of this crate.

PR #5204, opened 2026-07-21
feat: add sequential learning rate scheduler

### LINKED ISSUE #5127: Ability to chain LR schedulers to complement LR composition
Burn currently provides ```ComposedLRScheduler``` which provides a mechanism for composing different learning rate schedulers that emits a single LR.  They can be combined/composed with 3 different composition reductions, namely Avg, Sum and Prod.

This said, Burn currently seems to lack a mechanism for swapping out new LR Schedulers at different milestones of the training journey.  In my experience, this is a far more widely used use-case.

Pytorch provides a pair of LR combinators, namely:

1. Chaining Schedulers:
How it works: Each scheduler adjusts the learning rate obtained by the preceding one. At every step, the new learning rate is the product of all active scaling factors.
Implementation: You pass a list of schedulers to ``torch.optim.lr_scheduler.ChainedScheduler``.
Use case: Applying continuous exponential decay alongside a warm-up phase or a constant learning rate period.

2. Sequential Schedulers
How it works: You define a list of schedulers, and each one takes over the learning rate schedule only when a specific milestone (epoch number) is reached.
Implementation: You pass your list of schedulers and a list of milestone epochs to ``torch.optim.lr_scheduler.SequentialLR``.
Use case: Switching abruptly from a linear warm-up phase to a cosine annealing schedule.

Clearly, burn's analog for ``1`` above is its ``ComposedLrSchedulerConfig``, but there appears to be no analog for ``2``, despite this being a more common use-case.

Note in pytorch the ``milestones`` initializer for transitioning between schedulers (which in Burn's case could be epochs, or some number of iterations which would allow more granular control, but either way would be fine):

```
# The LinearLR will run for the first 5 epochs, then CosineAnnealingLR takes over
combined_scheduler = lr_scheduler.SequentialLR(
    optimizer, 
    schedulers=[warmup_scheduler, cosine_scheduler], 
    milestones=[5]
)
```

Burn should provide an analog for pytorch's ``SequentialLR`` to handle non-overlapping LR scheduler types/configurations.

A very typical use-case requires a warmup phase (linear?) for some number of epochs/iterations, followed by a transition to an alternate (e.g. cosine) for the remainder of training.

AnayGarodia:
### Checklist

### Related Issues/PRs

Closes #5127.
Depends on the scheduler config and record infrastructure merged in #5121.

### Changes

Adds `SequentialLrSchedulerConfig` and `SequentialLrScheduler` for running non-overlapping learning rate schedules at step milestones. Only the active child scheduler advances; each later child begins from its own initial learning rate when its milestone is reached.

The scheduler participates in `LrSchedulerConfig`, can be nested inside composed or module schedulers, validates scheduler/milestone cardinality and ordering, and checkpoints the global step plus every child state. The learner book now includes a linear-warmup-to-cosine example.

The shared `LrSchedulerConfig::build` path removes duplicated exhaustive matches from composed and module schedulers, which also keeps nested scheduler support consistent.

### Testing

- `cargo test -p burn-optim` (101 passed)
- `cargo test -p burn-optim --doc` (1 passed, 3 ignored)
- `cargo clippy -p burn-optim --lib --all-features -- -D warnings`
- `git diff --check`
- `cargo run-checks` progressed successfully through dependency audit, formatting, workspace Clippy with warnings denied, typo checks, and the full optimized workspace build. It was stopped during subsequent custom all-feature backend rebuilds after the relevant workspace gates passed.

### DIFF
--- a/burn-book/src/building-blocks/learner.md
+++ b/burn-book/src/building-blocks/learner.md
@@ -43,6 +43,23 @@ The `launch` method will start the training and return the trained model once fi
 Again, please refer to the [training section](../basic-workflow/training.md) for a relevant code
 snippet.
 
+## Sequential learning rate schedules
+
+Use `SequentialLrSchedulerConfig` when different schedulers should run during non-overlapping parts
+of training. Milestones count scheduler steps: in this warmup example, the linear scheduler produces
+the first 1,000 learning rates and the cosine scheduler takes over on step 1,000.
+
+```rust,ignore
+let lr_scheduler = SequentialLrSchedulerConfig::new(
+    vec![
+        LinearLrSchedulerConfig::new(1e-6, 1e-3, 1_000).into(),
+        CosineAnnealingLrSchedulerConfig::new(1e-3, 9_000).into(),
+    ],
+    vec![1_000],
+)
+.init()?;
+```
+
 ## Multiple optimizers
 
 It's common practice to set different learning rates, optimizer parameters, or use different optimizers entirely, for different parts
--- a/burn-book/src/building-blocks/learner.md
+++ b/burn-book/src/building-blocks/learner.md
@@ -36,29 +36,13 @@ The `SupervisedLearning` struct must be created with the training and validation
 
 When the training is configured to your liking, you can then move forward to running the training. The
 `launch` method requires a learner object providing: the model, the optimizer and the learning rate scheduler. Note
-that the latter can be a simple float if you want it to be constant during training.
+that the latter can be a simple float if you want it to be constant during training. See the
+[learning rate scheduler section](./lr-scheduler.md) for the available schedulers.
 
 The `launch` method will start the training and return the trained model once finished.
 
 Again, please refer to the [training section](../basic-workflow/training.md) for a relevant code
 snippet.
-
-## Sequential learning rate schedules
-
-Use `SequentialLrSchedulerConfig` when different schedulers should run during non-overlapping parts
-of training. Milestones count scheduler steps: in this warmup example, the linear scheduler produces
-the first 1,000 learning rates and the cosine scheduler takes over on step 1,000.
-
-```rust,ignore
-let lr_scheduler = SequentialLrSchedulerConfig::new(
-    vec![
-        LinearLrSchedulerConfig::new(1e-6, 1e-3, 1_000).into(),
-        CosineAnnealingLrSchedulerConfig::new(1e-3, 9_000).into(),
-    ],
-    vec![1_000],
-)
-.init()?;
-```
 
 ## Multiple optimizers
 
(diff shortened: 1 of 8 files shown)


burn-book/src/building-blocks/learner.md
```diff
@@ -43,6 +43,23 @@ The `launch` method will start the training and return the trained model once fi
...
+```rust,ignore
+let lr_scheduler = SequentialLrSchedulerConfig::new(
+    vec![
+        LinearLrSchedulerConfig::new(1e-6, 1e-3, 1_000).into(),
+        CosineAnnealingLrSchedulerConfig::new(1e-3, 9_000).into(),
+    ],
+    vec![1_000],
+)
+.init()?;
+```
```
laggui: This is essentially the only subsection describing a particular scheduler in more details. I think it would make more sense to have a new page entirely for schedulers, similar to how metrics have their own.

crates/burn-optim/src/lr_scheduler/sequential.rs
```diff
@@ -0,0 +1,216 @@
...
+    milestones: Vec<usize>,
+    step: usize,
+}
+
+impl SequentialLrScheduler {
+    fn active_scheduler(&self) -> usize {
+        self.milestones
+            .iter()
+            .take_while(|milestone| self.step >= **milestone)
+            .count()
```
laggui: Since `milestones` is guaranteed to be sorted in strictly increasing order, we can use `partition_point`:

```suggestion
        self.milestones.partition_point(|&m| self.step >= m)
```

crates/burn-optim/src/lr_scheduler/sequential.rs
```diff
@@ -0,0 +1,216 @@
...
+            self.step = state.step;
+        }
+
+        self.schedulers = self
+            .schedulers
+            .clone()
+            .into_iter()
+            .enumerate()
+            .map(|(index, scheduler)| scheduler.load_record(record.record(&index.to_string())))
+            .collect();
```
laggui: Small nit: we can avoid cloning all underlying schedulers
```suggestion
        let schedulers = core::mem::take(&mut self.schedulers);
        self.schedulers = schedulers
            .into_iter()
            .enumerate()
            .map(|(index, scheduler)| scheduler.load_record(record.record(&index.to_string())))
            .collect();
```

laggui (changes requested):
Just some minor comments, otherwise implementation LGTM!

Thanks for the addition.

AnayGarodia:
thanks for the review! moved the sequential scheduler docs to their own book page, switched to partition_point and mem::take as suggested. ci should rerun on the new push

laggui (approved):
Thanks for addressing my comments

PR #5207, opened 2026-07-21
feat(tensor): add `extract` method; simplify `inplace` implementation.

crutcher:
### Checklist

- [n/a] Made sure the book is up to date with changes in this PR.

### Changes

Formalized `Tensor::swap()` and `Tensor::release()`; rebuilt `Tensor::inplace()`

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -95,7 +95,7 @@ where
     K: Basic,
 {
     /// Swap the contents of two tensors.
-    pub fn swap(&mut self, other: &mut self) {
+    pub fn swap(&mut self, other: &mut Self) {
         core::mem::swap(self, other);
     }
 
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -94,6 +94,21 @@ impl<const D: usize, K> Tensor<D, K>
 where
     K: Basic,
 {
+    /// Swap the contents of two tensors.
+    pub fn swap(&mut self, other: &mut self) {
+        core::mem::swap(self, other);
+    }
+
+    /// Drop the current value, and replace it with `Tensor::empty([0; D])`
+    ///
+    /// Returns the old value.
+    #[allow(clippy::unused_must_use)]
+    pub fn release(&mut self) -> Self {
+        let mut z = Tensor::empty([0; D], &self.device());
+        self.swap(&mut z);
+        z
+    }
+
     /// Executes an operation on the tensor and modifies its value.
     ///
     /// # Notes
@@ -105,11 +120,8 @@ where
     /// want to mutate a tensor by using owned operations. A plausible usage would be to
     /// update the weights of a mutable model reference.
     pub fn inplace<F: FnOnce(Self) -> Self>(&mut self, func: F) {
-        let mut tensor_owned = Tensor::empty([0; D], &self.device());
-        core::mem::swap(&mut tensor_owned, self);
-
-        let mut tensor_new = func(tensor_owned);
-        core::mem::swap(&mut tensor_new, self);
+        let mut z = func(self.release());
+        self.swap(&mut z);
     }
 
     /// Returns the number of dimensions of the tensor.
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -94,22 +94,16 @@ impl<const D: usize, K> Tensor<D, K>
 where
     K: Basic,
 {
-    /// Swaps the tensors referenced by `self` and `other`.
-    ///
-    /// No data is copied or computed, only which underlying tensor
-    /// each reference points to is swapped.
+    /// Swap the contents of two tensors.
     pub fn swap(&mut self, other: &mut Self) {
         core::mem::swap(self, other);
     }
 
-    /// Takes ownership of the tensor out of `self`, leaving an empty
-    /// zero-shape placeholder tensor in its place.
+    /// Drop the current value, and replace it with `Tensor::empty([0; D])`
     ///
-    /// This is analogous to [`Option::take`] / [`core::mem::take`]: it lets you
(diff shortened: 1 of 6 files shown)


crutcher:
This lets us write mutable attribute mutators which don't fit into `.inplace()`, but where we want to be sure we don't hold the reference (and break optim). Here's a real example from my codebase:

```rust
        let queue = self.queue.release();
        let ext = Tensor::cat(vec![queue, stream], 1);
        self.queue = ext.clone().slice_dim(1, -(win_len as isize)..);

        // `analyze` yields `steps + 1` frames at `hop_size` offsets; frame
        // `s + 1` is the queue state after hop `s`, and frame 0 (the
        // pre-push queue) is dropped.
        //
        // [batch, steps + 1, n_bins, 2]
        let x = self.coef.analyze(ext);
        // [batch, steps, n_bins, 2]
        let x = x.slice_dim(1, 1..);
        // [steps, batch, n_bins, 2]
        let x = x.swap_dims(0, 1);
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -94,6 +94,21 @@ impl<const D: usize, K> Tensor<D, K>
 where
     K: Basic,
 {
+    /// Swap the contents of two tensors.
```
laggui: I'd improve the doc slightly, e.g.

```rust
/// Swaps the tensors referenced by `self` and `other`.
///
/// No data is copied or computed, only which underlying tensor each reference points to changes.
```

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -94,6 +94,21 @@ impl<const D: usize, K> Tensor<D, K>
...
+    pub fn swap(&mut self, other: &mut Self) {
+        core::mem::swap(self, other);
+    }
+
+    /// Drop the current value, and replace it with `Tensor::empty([0; D])`
+    ///
+    /// Returns the old value.
+    #[allow(unused_must_use)]
+    pub fn release(&mut self) -> Self {
```
laggui: I think `release` could be a bit misleading to some users/readers. This is analogous to `core::mem::take` (or `Option::take`), so I think it would make sense to rename it is such `tensor.take()` and slightly update the doc, e.g.

```rust
/// Takes ownership of the tensor out of `self`, leaving an empty
/// zero-shape placeholder tensor in its place.
///
/// This is analogous to [`Option::take`] / [`core::mem::take`]: it lets you
/// obtain an owned `Tensor` from behind a `&mut Tensor` so you can call
/// owned operations on it.

**Review by laggui — CHANGES_REQUESTED** [review]
That makes sense! Just have some minor comments on docs + naming to make things a bit clearer.

File: crates/burn-tensor/src/tensor/api/base.rs (hunk 1 above)
Comment (crutcher) [reply crates/burn-tensor/src/tensor/api/base.rs]: done

File: crates/burn-tensor/src/tensor/api/base.rs (hunk 2 above)
Comment (crutcher) [reply crates/burn-tensor/src/tensor/api/base.rs]: done.

Comment (crutcher) [reply crates/burn-tensor/src/tensor/api/base.rs]: No, `take` conflicts with an existing selection operation. rolled it back.

- `release`?
- `drop`?

File: crates/burn-tensor/src/tensor/api/base.rs (hunk 1 above)
Comment (crutcher) [reply crates/burn-tensor/src/tensor/api/base.rs]: Rolled back with the other change; needs to be re-applied once we pick a name.

**Review by laggui — COMMENTED** [review]
That's my bad 😅 forgot about the existing `tensor.take` API.

Continuing the points discussed on discord:
- I don't think we should have `tensor.swap(&mut other)`. There is no functional gain to using `core::mem::swap` here, it's universally understood
- Since `take` isn't available, we could have it as `extract()` or `take_value()`, e.g.

```rust
let t = tensor.extract();
// or
let t = tensor.take_value();
```

Lmk

crutcher:
@laggui half the time, it *is* the intent to just release; sometimes we want to keep it, sometimes we don't. When the tensor is attached to something (say a mut struct), where we can't get rid of the reference tensor, and do want to update it.

`core::mem::swap` is going to stop being well-defined if/when we move tensors to true async handles; *because* we'll need references to them to sit in async notification pools. W/e. I don't see the objection from having a standardized / blessed answer; you're saying "there's a contract that this is legal, and will continue to work".

`take` and `release` are the common terms in smart pointers in other libs. I honestly don't understand the objection to `release`. I'll go with 'extract' so we avoid confusion between `take` and `take_value`.

crutcher:
@laggui I don't agree about dropping `Tensor::swap`, but I did it.
I picked `Tensor::extract`, and pushed.

laggui:
Thanks for adjusting after all. Appreciate it :pray: 

I'll try to elaborate a bit just to try to make my stance a bit more clear

> `take` and `release` are the common terms in smart pointers in other libs. I honestly don't understand the objection to `release`. I'll go with 'extract' so we avoid confusion between `take` and `take_value`.

`release()` doesn't exist in Rust, that's for C++ smart pointers, where you relinquish ownership and return a raw pointer while leaving the original smart pointer empty/null. The equivalent would be `into_raw` for pointer extraction, or `take()` / `replace()` when swapping contents out of a mutable reference. I just think that using `release()` to return a managed, owned `Tensor` confuses raw-handle semantics with value extraction. 

> `core::mem::swap` is going to stop being well-defined if/when we move tensors to true async handles; _because_ we'll need references to them to sit in async notification pools. W/e. I don't see the objection from having a standardized / blessed answer; you're saying "there's a contract that this is legal, and will continue to work".

`Tensor` handles are already cheap value wrappers around shared handles (the underlying backend primitive / resource, like `CubeTensor`) with non-blocking operations. It works and will continue to work with `core::mem::swap`. But maybe I'm misunderstanding the point you're trying to make. FWIW, I'm not entirely against `tensor.swap` (the contract is much more clear than `tensor.release` :smile:), I just don't think it's needed.

laggui (approved):
I just rebased onto upstream/main to clean up the history of duplicated cherry-picked commits

Issue #5192, opened 2026-07-18
NdArray<f64> silently produces F32 tensors after any NdArray<f32> use on the same device (device dtype locked on first use)


laurigates:
## Minimal reproduction (burn 0.21.0)

```rust
use burn::backend::NdArray;
use burn::tensor::Tensor;

type F32B = NdArray<f32>;
type F64B = NdArray<f64>;

let device = Default::default();
let _lock = Tensor::<F32B, 1>::zeros([4], &device); // first tensor use: locks the device float dtype to F32
let t = Tensor::<F64B, 1>::zeros([4], &device);     // f64 backend, same device key

assert_eq!(format!("{:?}", t.to_data().dtype), "F64"); // FAILS: dtype is F32
t.into_data().to_vec::<f64>().unwrap();                 // TypeMismatch("Invalid target element type (expected F32, got F64)")
```

Observed output:

```
F64B tensor dtype after F32B ran first: F32
to_vec::<f64> result: Err(TypeMismatch("Invalid target element type (expected F32, got F64)"))
```

Run the two backends in the opposite order and the failure inverts (`NdArray<f32>` tensors come back F64). In a test binary that uses both widths, which side fails depends on test scheduling — with the default parallel test runner the failures are nondeterministic run-to-run.

## What I expected

Either `Tensor::<NdArray<f64>, _>::zeros(...)` produces an F64 tensor (the backend's `FloatElem` type parameter is honored), or tensor creation fails loudly when the type parameter conflicts with the device's locked dtype.

## What happens instead

`burn-backend`'s per-device settings registry (`crates/burn-backend/src/backend/device.rs`) locks the default float dtype on the first tensor operation — per the `DeviceSettings` docs: *"If an operation (like creating a tensor) occurs before manual initialization, the settings are permanently locked to their default values."* Both `NdArray<f32>` and `NdArray<f64>` resolve to the same `NdArrayDevice::Cpu` device key, so whichever backend runs first locks the width for the whole process, and the other backend's tensors are silently created at the wrong width. The failure then surfaces far from the cause — e.g. as a `TypeMismatch` on `into_data().to_vec::<f64>()`, or as wrong-precision arithmetic that never errors at all.

Explicit re-initialization returns `DeviceError::AlreadyInitialized`, but the *implicit* lock-then-mismatch path has no error at the point where the conflicting tensor is created.

## Context

[...]

PR #5209, opened 2026-07-22
docs: document per-device default dtypes; make AlreadyInitialized error actionable

laurigates:
### Checklist

> Note: `cargo run-checks` was not run here — the workspace requires rustc 1.95 and the two changed
> crates were checked individually (`cargo check -p burn-std` passes). Docs-only book change plus a
> single error-message string; happy to run full checks on request.

### Related Issues/PRs

- Addresses tracel-ai/burn#5192 — `NdArray<f64>` silently produced F32 tensors after any
  `NdArray<f32>` use on the same device (device dtype locked on first use).

### Changes

Follow-up to the maintainer note on #5192: the mixed-width footgun came from backends' *default
associated element types*, which were removed in #5000. On `main` the float width is now a runtime,
**per-device** setting held in a global registry that locks on first use — so the original silent
mismatch is structurally gone, and configuring a device after it is locked already returns
`DeviceError::AlreadyInitialized` (the issue's "preferred" fix). What was missing was documentation
of the new semantics and a helpful error message.

- **Book (`building-blocks/tensor.md`)**: new **Data types** section explaining that the element type
  is not part of the `Tensor` type but a per-device default. Covers inspecting defaults
  (`Device::settings`), changing them (`Device::configure` / `set_default_dtypes`) *before* the first
  tensor op, the "lock on first use" semantics, and how to work with more than one precision via
  explicit per-tensor `cast` (rather than expecting two default float widths on one device).
- **`DeviceError::AlreadyInitialized`**: the message now explains that defaults lock on the first
  tensor operation and how to avoid it, instead of only stating that settings were already
  initialized.

This implements the issue's "minimum" ask (document that the width becomes non-authoritative /
mixed-width backends cannot coexist) on the current design, and sharpens the error users hit when
they configure a device too late.

### Testing

- `cargo check -p burn-std` passes (with rustc 1.95, required by the workspace).
- Book change is documentation only; all code snippets are `rust, ignore` and mirror real API usage
  in the codebase (e.g. `tensor.cast(FloatDType::F64)` per `burn-tensor` doctests).
- Existing `device` registry tests already assert the `AlreadyInitialized` variant is returned; the
  change only edits the human-readable message string.

https://claude.ai/code/session_015oxLsZydYJrw3AWBePecc2

---
_Generated by [Claude Code](https://claude.ai/code/session_015oxLsZydYJrw3AWBePecc2)_

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -15,10 +15,11 @@ backend implementations.
 
 ### Data types
 
-The element type used for a tensor is not encoded in the `Tensor` type. Instead, each device keeps a
-set of **default data types** that determine the width used when a tensor is created without an
-explicit dtype (for example with `Tensor::zeros`, `Tensor::ones` or `from_floats`). You can inspect
-the current defaults with [`Device::settings`](https://docs.rs/burn/latest/burn/tensor/struct.Device.html):
+A tensor has an element type, but unlike its `Float`/`Int`/`Bool` kind that type is a runtime
+property rather than a generic parameter on `Tensor`. When a tensor is created without an explicit
+dtype (for example with `Tensor::zeros`, `Tensor::ones` or `from_floats`), the element type comes
+from a set of **default data types** that each device keeps. You can inspect the current defaults
+with [`Device::settings`](https://docs.rs/burn/latest/burn/tensor/struct.Device.html):
 
 ```rust, ignore
 use burn::tensor::Device;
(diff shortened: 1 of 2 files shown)


burn-book/src/building-blocks/tensor.md
```diff
@@ -13,6 +13,52 @@ Tensor<B, D, Bool>     // Bool tensor
 Note that the specific element types used for `Float`, `Int`, and `Bool` tensors are defined by
 backend implementations.
 
+### Data types
+
+The element type used for a tensor is not encoded in the `Tensor` type. Instead, each device keeps a
```
laggui: This framing is a bit confusing. A tensor has an element type; it's just a runtime property rather than a generic type parameter.

burn-book/src/building-blocks/tensor.md
```diff
@@ -13,6 +13,52 @@ Tensor<B, D, Bool>     // Bool tensor
...
+Because the default width is a per-device property, you cannot have two different default float widths
+active on the same device within a single process. If you need values at more than one precision, set
+the default once up front and create the other tensors with an explicit dtype instead, using
+[`cast`](https://docs.rs/burn/latest/burn/tensor/struct.Tensor.html#method.cast):
+
+```rust, ignore
+// device defaults to f32
+let x = Tensor::<Backend, 2>::zeros([2, 3], &device); // f32
+let x_f64 = x.cast(FloatDType::F64);                  // explicit f64
+```
```
laggui: This is incorrect. `cast` is used to convert an existing tensor. If your goal is to create a tensor with a different dtype, you should create it with the desired dtype directly:

```rust
Tensor::<2>::zeros([2, 3], (&device, DType::F64))
```
the `(&device, dtype)` tuple is just a conversion for `TensorCreationOptions`.

On main the `Tensor` no longer takes a backend generic but the whole book needs to be updated for that anyway, so it's out of scope for this PR.

laggui (changes requested):
Just some comments on the added doc.

On a side note, we need to update the book with the recent breaking changes (no more backend generics) and improvements.

laurigates:
Thanks for the review! Both points addressed in 1712b0acd:

**1. Element-type framing (line 18).** You're right — reworded so it no longer claims the type "isn't encoded" in `Tensor`. It now reads: a tensor *has* an element type, but that type is a runtime property rather than a generic parameter on `Tensor`, and when a tensor is created without an explicit dtype the element type comes from the device's default data types.

**2. `cast` misused as a creation path (line 60).** Corrected. The example now creates a tensor at a specific dtype directly via the `(&device, dtype)` `TensorCreationOptions` tuple:

```rust
let x_f64 = Tensor::<Backend, 2>::zeros([2, 3], (&device, DType::F64));
```

and a separate, smaller snippet keeps `cast` for its actual purpose — converting an *existing* tensor:

```rust
let x_f64 = x.cast(FloatDType::F64);
```

Kept the `Tensor::<Backend, D>` form for now since the broader backend-generic book update is out of scope here, as you noted — happy to fold this section into that pass whenever it happens.

laggui (approved):
Thanks! LGTM

PR #5210, opened 2026-07-22
fix(backend): error on conflicting default element widths per device (0.21 patch)

laurigates:
### Checklist

> Note: this is a targeted fix cut from the `v0.21.0` tag (the base branch `release/0.21` points at
> `v0.21.0`), intended for a possible `0.21.x` patch release per the maintainer's note on #5192. Full
> `cargo run-checks` was not run; the affected crates were checked and tested directly with rustc 1.95
> (see Testing). No book change is needed on this line — the API is unchanged.

### Related Issues/PRs

- Fixes tracel-ai/burn#5192 — `NdArray<f64>` silently produced F32 tensors after any `NdArray<f32>`
  use on the same device.
- Follows the maintainer's suggestion on #5192 to "introduce a patch for 0.21 with one of the
  suggested fixes" (the preferred one: throw a clear error on a dtype conflict). On `main` the
  associated element types were removed in #5000, so this footgun no longer exists there; the
  companion PR #5209 documents the current per-device dtype semantics for `main`.

### Changes

The device settings registry (`crates/burn-backend/src/backend/device.rs`) is keyed by physical
device (`DeviceId` + device `TypeId`), **not** by backend type, so backends that differ only in their
associated element types — e.g. `NdArray<f32>` and `NdArray<f64>` — share the same entry. Settings
lock on first use via `OnceLock`, so whichever backend touches a device first imposes its element
width on the others: a later `NdArray<f64>` tensor was silently created at F32, surfacing as a
`TypeMismatch` far downstream (or as silent precision loss).

This detects the conflict and fails loudly at the point of use:

- Each registry entry now records, for settings locked *implicitly* by the first tensor operation,
  the backend-derived default `(float, int)` dtypes (`StoredSettings::implicit_default`).
- A later access by a backend whose own defaults differ panics with an actionable message naming both
  widths and how to avoid the conflict.
- Settings configured *explicitly* via `set_default_dtypes` carry no implicit default, so intentional
  overrides (e.g. mixed precision) are **never** flagged as conflicts — no false positives.
- The check runs on every access, including the lock-free thread-local cache path (so a cache hit
  can't bypass it).

The decision is isolated in a small pure helper (`conflicting_lock`) so it is unit-testable without a
full backend.

### Testing

Run with rustc 1.95 (required by the workspace):

- `cargo test -p burn-backend --lib backend::device` — 10 passed, including 3 new unit tests for
  `conflicting_lock` (matching width is fine; a differing implicit width conflicts; an explicit
  override never conflicts) and all pre-existing registry tests.
- `cargo test -p burn-ndarray --test dtype_conflict` — new regression test reproducing the exact
  `NdArray<f32>` then `NdArray<f64>` scenario from the issue: the first backend locks the device to
  F32, and the second now panics with `Conflicting default data types` instead of silently returning
  F32.

https://claude.ai/code/session_015oxLsZydYJrw3AWBePecc2

---
_Generated by [Claude Code](https://claude.ai/code/session_015oxLsZydYJrw3AWBePecc2)_

### DIFF
--- a/crates/burn-backend/src/backend/device.rs
+++ b/crates/burn-backend/src/backend/device.rs
@@ -78,11 +78,30 @@ impl DeviceSettings {
 /// Key for the registry: physical device type + device id
 type RegistryKey = (DeviceId, TypeId);
 
+/// Settings stored in the registry, along with the provenance needed to detect
+/// conflicting element widths.
+///
+/// The registry is keyed by *physical device* ([`DeviceId`] + device [`TypeId`]), not by backend
+/// type. Backends that only differ in their associated element types (e.g. `NdArray<f32>` and
+/// `NdArray<f64>`) therefore share the same entry. Because the settings are locked on first use,
+/// whichever backend runs first would otherwise silently impose its element width on the others.
+///
+/// To catch that mistake we remember, for settings that were locked *implicitly* (by the first
+/// tensor operation rather than an explicit [`set_default_dtypes`] call), the backend-derived
+/// default `(float, int)` dtypes used to initialize them. A later access by a backend whose own
+/// defaults differ is a conflict. Explicit configuration stores [`None`], so intentional overrides
+/// (like enabling mixed precision) never trigger a false positive.
+#[derive(Debug, Clone, Copy)]
+struct StoredSettings {
+    settings: DeviceSettings,
+    implicit_default: Option<(FloatDType, IntDType)>,
+}
+
 /// Global registry mapping devices to their settings.
 ///
 /// Each value is wrapped in a `OnceLock` to enforce that settings are initialized only once
 /// per device.
-static REGISTRY: LazyLock<RwLock<HashMap<RegistryKey, Arc<OnceLock<DeviceSettings>>>>> =
+static REGISTRY: LazyLock<RwLock<HashMap<RegistryKey, Arc<OnceLock<StoredSettings>>>>> =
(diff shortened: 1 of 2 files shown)


laurigates:
@laggui this is the 0.21 patch you raised in #5192 — "We could introduce a patch for 0.21 with one of the suggested fixes." It targets `release/0.21`.

Approach: the registry entry now records whether the settings were locked *implicitly* (by the first tensor operation) and which backend-derived `(float, int)` defaults did the locking. A later access from a backend whose own defaults differ panics at the point of use, naming both widths. An explicit `set_default_dtypes` stores `None` as the provenance, so deliberate overrides such as mixed precision don't trip it. Regression test in `crates/burn-ndarray/tests/dtype_conflict.rs`.

The one red check, `code-quality (stable)`, is `cargo audit` failing on advisories in transitive dependencies — RUSTSEC-2026-0195 (`quick-xml`), RUSTSEC-2026-0185 (`quinn-proto`), and yanked `aes` 0.9.0 / `enumset` 1.1.12 / `spin` 0.10.0. This branch changes two Rust source files and no manifest.

laggui:
Sorry for the delayed reply.

While I could merge this on the `release/0.21` branch, there is currently no plan to publish another 0.21 release. I'll revisit this patch if our release plans change, but right now our efforts are on the 0.22 release.

Thanks again for the docs improvements in #5209!

PR #5215, opened 2026-07-22
feat: support negative sort and top-k dimensions

mattisonchao:
## Motivation

Burn tensor APIs increasingly support negative dimension indexing, but sort and top-k operations still required usize dimensions. This made common last-axis calls such as dim = -1 inconsistent with the rest of the public tensor API.

## Modifications

- Changed sort, descending sort, indexed sort, argsort, topk, and topk_with_indices public APIs to accept AsIndex dimensions.
- Resolve each dimension once before existing validation and usize backend calls.
- Added float and integer regression coverage across all eight APIs.


laggui (approved):
Thanks for the PRs on this!

I think all the smaller PRs related to negative indexing support across tensor methods could have been merged into one given the context.

PR #5218, opened 2026-07-22
fix(burn-train): correctly aggregate per-epoch metrics that require global statistics

### LINKED ISSUE #2758: Incorrect aggregation for classification metrics
Hi everyone, I've noticed that currently (all?) the metrics in the library are updated using the `update` method of `NumericMetricState` [link](https://github.com/tracel-ai/burn/blob/6d8dd6920ebd66bf56d072d35fa6ca74c54e54cd/crates/burn-train/src/metric/state.rs#L63-L90).
However, when using metrics such as `PrecisionMetric` and `RecallMetric`, this aggregation approach can lead to incorrect results.

Consider the Recall metric with two batches of data:
| Batch | True Positives (TP)  | True Negatives (TN)  | False Positives (FP)  | False Negatives (FN) | Recall       |
|-------|----------------------|----------------------|-----------------------|----------------------|--------------|
| **1** | 40                   | 50                   | 10                    | 50                   | 0.44 (40/90) |
| **2** | 20                   | 90                   | 30                    | 10                   | 0.67 (20/30) |

With the current implementation, the final recall at the end of an epoch is computed as `(150*0.44+150*0.67)/(150+150)~0.55`.
However, if we computed recall using the full dataset, the correct value would be: `(40+20)/(40+20+50+10)=0.5`.

This discrepancy arises because the current approach averages batch-wise recall scores instead of aggregating the confusion matrix values (TP, FP, TN, FN).

To ensure accurate precision and recall calculations, we should store the confusion matrix as part of the metric state. This approach is used in `torchmetrics` [link](https://github.com/Lightning-AI/torchmetrics/blob/5fc2e0bb2ea19817fc5d8645bdb4a975acd9c613/src/torchmetrics/classification/stat_scores.py#L184-L190)

This however is not possible for more complex metric such as `AurocMetric` where the input and target tensors are required. I think this is the reason why `torchmetrics` [doesn't provide](https://github.com/Lightning-AI/torchmetrics/blob/5fc2e0bb2ea19817fc5d8645bdb4a975acd9c613/src/torchmetrics/classification/auroc.py#L537-L541) an `update` method for this metric.

To summarize, metrics such as precision and recall should be computed by aggregating confusion matrix values instead of averaging per-batch scores.

For metrics such as the AUROC instead, If we maintain an approximation, it should be well documented to avoid misunderstanding with the users; otherwise, we could remove the update method (following `torchmetrics` practice).

Thanks!

laggui:
### Checklist

### Related Issues/PRs

Fixes #2758

Previously, metrics like Precision, Recall, and AUROC incorrectly assumed that global epoch-level scores could be computed by averaging individual batch-level metrics (weighted by sample count in `NumericMetricState`). This approach is biased for metrics requiring global statistics and sometimes entirely incorrect for metrics like AUROC and AUC-PR which cannot be meaningfully evaluated per-batch.

### Changes

- **Updated `Numeric` Trait:** Changed `value()` and `running_value()` to return `Option<NumericEntry>` (returning `None` for epoch-only metrics mid-epoch) and added `final_value()` to retrieve the computed metric at epoch completion.
- **Split `Metric` Execution:** Updated the `Metric` trait to decouple batch steps (`update`) from epoch-level finalization (`compute`). Updated `NumericMetricState` with `compute_update` and `compute_final` methods so that global aggregation is explicitly triggered at the end of each epoch (`end_epoch_train` in the event processor).
- **Added `ConfusionStatsState`:** Introduced a state structure that aggregates raw confusion matrix statistics (TP, FP, FN) across batches, enabling accurate global metric evaluation.
- **Updated `AurocMetric` & `AucPrMetric` via `PredictionAccumulatorState`:** Refactored `AurocMetric` to use `PredictionAccumulatorState` alongside `AucPrMetric`, accumulating raw prediction/target tensors during `update` and evaluating AUROC once over the full dataset during `compute`.

### Testing

Unit tests and mnist example

### DIFF
--- a/crates/burn-train/src/metric/perplexity.rs
+++ b/crates/burn-train/src/metric/perplexity.rs
@@ -232,7 +232,6 @@ impl Metric for PerplexityMetric {
         NumericAttributes {
             unit: None,
             higher_is_better: false,
-            ..Default::default()
         }
         .into()
     }
--- a/crates/burn-train/src/metric/perplexity.rs
+++ b/crates/burn-train/src/metric/perplexity.rs
@@ -90,20 +90,20 @@ impl PerplexityState {
         SerializedEntry::new(formatted, serialized)
     }
 
-    fn value(&self) -> NumericEntry {
+    fn value(&self) -> Option<NumericEntry> {
         let perplexity = if self.total_tokens > 0 {
             (self.sum_nll / self.total_tokens as f64).exp()
         } else {
             f64::INFINITY
         };
 
-        NumericEntry::Aggregated {
+        Some(NumericEntry::Aggregated {
             aggregated_value: perplexity,
             count: self.total_tokens,
-        }
+        })
     }
 
-    fn running_value(&self) -> NumericEntry {
+    fn running_value(&self) -> Option<NumericEntry> {
         self.value()
     }
 }
@@ -216,6 +216,10 @@ impl Metric for PerplexityMetric {
         )
     }
 
+    fn compute(&mut self) -> SerializedEntry {
+        todo!()
+    }
+
     fn clear(&mut self) {
         self.state.reset()
     }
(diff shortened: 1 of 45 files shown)


crates/burn-train/src/metric/perplexity.rs
```diff
@@ -216,6 +216,10 @@ impl Metric for PerplexityMetric {
         )
     }
 
+    fn compute(&mut self) -> SerializedEntry {
+        todo!()
```
Charles23R: Was this intentionally left out of the current PR?

Charles23R (approved):

PR #5226, opened 2026-07-23
feat(tensor): support negative dimensions

mattisonchao:
## Motivation

Recent changes added negative-dimension support to activations, cumulative operations, FFTs, sorting, statistics, and norms. Several remaining high-level tensor APIs still use a mix of `usize`, signed integer types, and custom argument handling. Standardizing these public boundaries makes negative indexing available consistently while keeping backend and IR boundaries canonicalized as `usize`.

## Modifications

- Migrate the remaining Tensor base and orderable dimension parameters to `AsIndex`.
- Support `AsIndex` element types in `flip`, `squeeze_dims`, and `one_hot_fill`.
- Broaden `MovedimArgs` integer support while preserving its public trait shape.
- Normalize insertion dimensions against `D + 1` and fix middle insertion when `D2 > D + 1`.
- Add negative-dimension tests and update API documentation.

## Compatibility

This intentionally changes 21 public method signatures. Most existing typed values and integer literals remain source-compatible. Untyped empty inputs now require annotations, for example `flip([] as [isize; 0])` and `squeeze_dims(&[] as &[isize])`.

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/slice.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/slice.rs
@@ -21,7 +21,7 @@ fn should_support_slice_dim_1d() {
 }
 
 #[test]
-#[should_panic(expected = "The provided dimension exceeds the tensor dimensions")]
+#[should_panic(expected = "Index { kind: Dimension, index: 1, bounds: 0..1 }")]
 fn should_panic_when_slice_dim_1d_bad_dim() {
     let data = TensorData::from([0.0, 1.0, 2.0]);
     let tensor = TestTensor::<1>::from_data(data.clone(), &Default::default());
--- a/crates/burn-backend-tests/tests/cubecl/reduce.rs
+++ b/crates/burn-backend-tests/tests/cubecl/reduce.rs
@@ -27,7 +27,7 @@ fn reduction_argtopk_simple() {
     let device = Default::default();
 
     let tensor = TestTensor::<2>::from_data([[1, 7, 3], [8, 2, 8]], &device);
-    let actual = tensor.argtopk(2, 1);
+    let actual = tensor.argtopk(2, -1_i64);
     let expected = TestTensor::<2>::from_data([[1, 2], [0, 2]], &device);
 
     let output_shape = Shape::new([2, 2]);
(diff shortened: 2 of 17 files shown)


mattisonchao:
### Compatibility question

This PR intentionally standardizes the remaining public Tensor dimension APIs on `AsIndex`, including APIs such as `flip`, `squeeze_dims`, and `one_hot_fill` that already supported negative dimensions through concrete signed types. That creates a small but real source-compatibility cost:

- Untyped empty inputs such as `flip([])` and `squeeze_dims(&[])` need an explicit element type.
- Rare inference-heavy or exact function-item usages may need type annotations.

Normal integer literals and values using the previous concrete types remain accepted. The required fixes are local and mechanical.

Would these breaking changes be acceptable for the consistency gained here? The project currently states:

> Burn is currently in active development, and there will be breaking changes. While any resulting issues are likely to be easy to fix, there are no guarantees at this stage.

My preference is to keep the complete Tensor `AsIndex` migration together in this PR. If maintainers prefer a narrower compatibility tradeoff, I can retain the existing concrete signatures for APIs that already support negative dimensions.

crates/burn-backend-tests/tests/tensor/float/ops/slice.rs
```diff
@@ -21,7 +21,7 @@ fn should_support_slice_dim_1d() {
 }
 
 #[test]
-#[should_panic(expected = "The provided dimension exceeds the tensor dimensions")]
+#[should_panic(expected = "Index { kind: Dimension, index: 1, bounds: 0..1 }")]
```
laggui, line 24: It would be nice to preserve the `TensorCheck` error message formatting.

This can be done in a follow-up PR, as it would likely require updating `AsIndex` in cubecl to have the `try_*` equivalents of the dim wrapping methods (essentially calling `IndexWrap::try_wrap` instead) to tie into `TensorCheck` when it fails.

Are you up for it? Otherwise we can simply open an issue to track it.

laggui (approved):
I'm ok with the require type hints for empty slices when using `flip` or `squeeze_dims`. For `flip` it's essentially a no-op anyway, and `squeeze_dims` with an empty slice is just like calling `squeeze` (applies to all singleton dims).

crates/burn-backend-tests/tests/tensor/float/ops/slice.rs
mattisonchao: No worries, let me work on it. :)

mattisonchao: https://github.com/tracel-ai/cubecl/pull/1445 

Please help take a look.

PR #5227, opened 2026-07-23
feat(burn-nn): support negative dimensions

mattisonchao:
## Motivation

Several `burn-nn` abstractions still require positive dimensions even though their underlying tensor operations support negative indexing. Allowing dimensions such as `-1` keeps loss reduction, cosine similarity, and LSTM state manipulation consistent with the Tensor API and makes dimension choices less dependent on rank-specific positive indices.

## Modifications

- Accept `AsIndex` in `LstmState::squeeze_dim`, `unsqueeze_dim`, `stack`, and `chunk`.
- Accept `AsIndex` element types in `LpLoss::forward_reduce_dims` and `SmoothL1Loss::forward_reduce_dims`.
- Normalize and sort loss reduction dimensions before applying the reduction.
- Change `CosineSimilarityConfig::dim` and `CosineSimilarity::dim` from `usize` to `isize`, then normalize during `forward`.
- Add negative-dimension tests and update documentation.

## Compatibility

This intentionally changes six public method signatures and two public cosine dimension fields. Existing typed dimension values and integer literals remain compatible for the generic methods. Untyped empty loss dimension slices now need an annotation such as `&[] as &[usize]`. Typed `usize` values passed to the cosine configuration need conversion to `isize`, and the serialized configuration schema changes from unsigned to signed.

### DIFF
--- a/crates/burn-nn/src/loss/lp_loss.rs
+++ b/crates/burn-nn/src/loss/lp_loss.rs
@@ -1,7 +1,7 @@
 use super::Reduction;
 use burn::config::Config;
 use burn::module::Module;
-use burn::tensor::Tensor;
+use burn::tensor::{AsIndex, Tensor};
 use burn_core as burn;
 
 /// Configuration for the [Lp Loss](LpLoss) module.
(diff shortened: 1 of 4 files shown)


mattisonchao:
### Compatibility question

The source-compatibility impact in this PR comes from two Rust API details:

- `forward_reduce_dims`: changing `&[usize]` to `&[impl AsIndex]` removes the fixed element type that previously let Rust infer an untyped `&[]`. Non-empty literals and typed slices remain accepted.
- Cosine similarity: supporting a persistent negative dimension requires changing the stored public `dim` fields from `usize` to a signed type. This also changes the generated `with_dim` setter and the signedness of the serialized configuration schema.

The four `LstmState` methods also change from concrete `usize` parameters to `impl AsIndex`; normal calls remain source-compatible, while rare unconstrained inference or exact function-item usages may need annotations.

Would these breaking changes be acceptable for this `burn-nn` API consistency? The project currently states:

> Burn is currently in active development, and there will be breaking changes. While any resulting issues are likely to be easy to fix, there are no guarantees at this stage.

The common migration is mechanical: annotate empty slices and convert typed cosine `usize` dimensions to `isize`.

PR #5228, opened 2026-07-24
feat(burn-optim): add projective flag to WeightDecayConfig

Vrinda12-tech:
## Summary

Adds a `projective: bool` flag to `WeightDecayConfig`. When enabled, weight 
decay is skipped for tensors with rank &lt; 2 (biases, LayerNorm parameters, 
scalars). This is standard practice in modern transformer training.

## Motivation

Applying weight decay to 1D parameters destabilizes training in architectures 
like transformers. PyTorch users typically solve this via parameter groups. 
Burn does not yet support parameter groups, so this flag provides the standard 
heuristic with minimal API surface.

## Changes

- `WeightDecayConfig`: added `projective: bool` with `#[config(default = false)]`
- `WeightDecay`: stores the flag; `transform()` applies a compile-time rank 
  check (`D &lt; 2`) to zero out decay when projective mode is active
- `rmsprop.rs`, `sgd.rs`: updated struct literals to include the new field

## Tests

- `projective_decay_skips_1d_tensors` — 1D tensors bypass decay
- `projective_decay_applies_to_2d_tensors` — 2D tensors receive decay
- `projective_off_does_not_skip_1d_tensors` — backward compatibility verified

## Breaking Changes

None. Default is `false`; existing behavior is unchanged.

## Verification

- `cargo test -p burn-optim`: 98 passed, 0 failed
- `cargo clippy -p burn-optim`: clean
- `cargo fmt -p burn-optim`: clean

### DIFF
--- a/crates/burn-optim/src/optim/decay.rs
+++ b/crates/burn-optim/src/optim/decay.rs
@@ -8,8 +8,14 @@ use burn::tensor::Tensor;
 /// Configuration to create [weight decay](WeightDecay).
 #[derive(Config, Debug)]
 pub struct WeightDecayConfig {
-    /// L2 penalty.
+    /// The weight decay penalty
     pub penalty: f32,
+
+    /// When true, skip weight decay on tensors with rank < 2.
+    /// This applies to biases (rank 1), LayerNorm parameters (rank 1),
+    /// and scalars (rank 0). Matches PyTorch's common heuristic.
+    #[config(default = false)]
+    pub projective: bool,
 }
 
 /// State of [weight decay](WeightDecay).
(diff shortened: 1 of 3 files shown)


laggui (commented):
This is a useful training convention, but it is a parameter grouping policy rather than a property of weight decay itself. The proposed solution bakes a parameter classification heuristic into the optimizer implementation.

Instead, this should be accomplished via the recently added `ParamGroup` (#5121), which allows users to assign different optimizer configurations (e.g., weight_decay = 0.0 vs weight_decay = 0.01) based on parameter names or module paths (predicates).

If you want to spin this PR into an example that achieves this using parameter groups, let me know and it can be re-opened :slightly_smiling_face: or feel free to open a new PR if you prefer

PR #5234, opened 2026-07-25
fix(examples/modern-lstm) swapped lstm args

ra1u:
Fix argument order in examples/modern-lstm/model
https://github.com/ra1u/burn/blob/5dce4daac5b2443392ac7fbcb4cd6a0288a7747e/crates/burn-nn/src/modules/rnn/lstm/lstm_state.rs#L28


laggui (approved):
Good catch 😬

PR #4900, opened 2026-04-29
Add argtopk (Cubecl backend)

Sublime12:
Add argtopk

### Checklist

- [ ] Confirmed that `cargo run-checks` command has been executed.
- [ ] Made sure the book is up to date with changes in this PR.

### Changes
add argtopk to float, int and q tensors

### Testing
```sh
cargo test-wgpu-nofuse argtopk
```

PR #5236, opened 2026-07-26
feat(backend): add default int_argtopk impl and native tch support

### LINKED ISSUE #5233: int_argtopk has no default impl, unlike float_argtopk
`int_argtopk` is the only op in the ordering family without a default implementation, so it panics on ndarray, candle, flex and tch.

## Current behaviour

```rust
let tensor = Tensor::<1, Int>::from_data([1, 2, 3, 4, 5], &device);
tensor.argtopk(3, 0); // panics on ndarray / candle / flex / tch
```

Stubs on `main` (5dce4daa):

- `crates/burn-ndarray/src/ops/int_tensor.rs:328` - `unimplemented!("argtopk not implemented for ndarray")`
- `crates/burn-candle/src/ops/int_tensor.rs:399` - `panic!("argtopk not implemented for candle backend")`
- `crates/burn-flex/src/ops/int.rs:610` - `panic!("argtopk not implemented for flex")`
- `crates/burn-tch/src/ops/int_tensor.rs:343` - `panic!("argtopk not implemented for torch")`

Only cubecl implements it. The float equivalent works everywhere.

## Why this looks unintended

The float and int ordering traits mirror each other exactly. `argmax`/`argmin` are required on both sides, everything sort-derived is defaulted on both sides:

| `FloatTensorOps` | | `IntTensorOps` | |
| --- | --- | --- | --- |
| `float_argmax` | required | `int_argmax` | required |
| `float_argmin` | required | `int_argmin` | required |
| `float_argsort` | default | `int_argsort` | default |
| `float_sort` | default | `int_sort` | default |
| `float_topk` | default | `int_topk` | default |
| `float_max_dim` | default | `int_max_dim` | default |
| `float_argtopk` | **default** | `int_argtopk` | **required** (here) |

`int_argtopk` is the single exception.

History:

- #4900 added both `float_argtopk` and `int_argtopk`, both declared without a default, both stubbed with panics across the CPU backends.
- #5160 gave `float_argtopk` a default (`crates/burn-backend/src/backend/ops/tensor.rs:1433`) and removed the ndarray/candle/flex stubs. The int side wasn't updated. The change was bundled into a commit titled "Fix matmul bounds check when launch", so the int half is easy to have missed.

Nothing about the int path makes this harder: `int_argsort` already has a working default at `crates/burn-backend/src/backend/ops/int_tensor.rs:1432`.

## Proposed fix

Add the matching default to `IntTensorOps::int_argtopk` (`crates/burn-backend/src/backend/ops/int_tensor.rs:939`), composed the same way as `float_argtopk` (`int_argsort` + `int_select`), and remove the three stubs shadowing it. Give tch a native override calling the existing `TchOps::argtopk` (`crates/burn-tch/src/ops/base.rs:549`), which `float_argtopk` already uses in the same crate.

Two behaviours would deliberately match the current float implementation rather than being changed here:

- **Index dtype** follows `int_argsort` (the input's int dtype), which is what cubecl's native `int_argtopk` already returns.
- **Tie-breaking** goes through `argsort`, which is unstable, so ties may not resolve to the lowest index as the doc comment states. This is already the case for `float_argtopk` on ndarray/candle/flex today. Keeping int consistent with float seems better than having the two diverge; happy to file that separately.

Implementation attached since it was already working locally, happy to adjust the approach if you'd prefer something different.

IssaAlBawwab:
### Checklist

### Related Issues/PRs

Closes #5233.

Follows #5160, which made the same change on the float side: added a default for `float_argtopk` and removed the ndarray, candle and flex stubs. Both ops were originally introduced together in #4900, each declared without a default.

### Changes

`int_argtopk` was declared without a default, so it panicked on ndarray, candle, flex and tch. Only cubecl implemented it, while `float_argtopk` works everywhere.

- Added a default for `IntTensorOps::int_argtopk`, composed as `int_argsort` followed by `int_select`, mirroring `float_argtopk`.
- Removed the ndarray, candle and flex stubs so they inherit it.
- Gave tch a native override calling the existing `TchOps::argtopk`.
- Added `tensor.argtopk(k, dim)` to the book's tensor op table.

Index dtype and tie-breaking deliberately match current `float_argtopk` behaviour rather than changing here. Rationale in #5233.

### Testing

Three tests added to `crates/burn-backend-tests/tests/tensor/int/ops/topk.rs`: `test_argtopk_1d`, `test_argtopk` and `test_argtopk_supports_negative_dim_int`. Expected indices match what the adjacent `topk_with_indices` tests already assert for the same inputs, so the two paths cross-check each other: `int_topk_with_indices` sorts once and slices both halves, while `int_argtopk` argsorts and slices.

- `cargo test-ndarray topk`: 11 passed, 0 failed
- `cargo test-flex topk`: 11 passed, 0 failed
- `cargo run-checks`: passed

candle inherits the default and compiles, but `burn-backend-tests` has no candle feature so it is not exercised by the suite.

tch is verified indirectly. `cargo test-tch` aborts at startup on unmodified `main` with `UnsupportedDType { device: "LibTorch(Cpu)", dtype: I32 }`: the harness hardcodes `IntElem = i32` in `crates/burn-backend-tests/tests/tensor.rs` while burn-tch maps only I64. Setting `IntElem = i64` locally lets the suite run, and all three argtopk tests pass on LibTorch alongside the existing topk tests. CI does not run `burn-backend-tests` against tch. Happy to file that separately if it is not already known.

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -268,6 +268,7 @@ Those operations are available for numeric tensor kinds: `Float` and `Int`.
 | `tensor.argmin(dim)`                                            | `tensor.argmin(dim)`                          |
 | `tensor.argsort(dim)`                                           | `tensor.argsort(dim)`                         |
 | `tensor.argsort_descending(dim)`                                | `tensor.argsort(dim, descending=True)`        |
+| `tensor.argtopk(k, dim)`                                        | `tensor.topk(k, dim).indices`                 |
 | `tensor.bool()`                                                 | `tensor.bool()`                               |
 | `tensor.clamp(min, max)`                                        | `torch.clamp(tensor, min=min, max=max)`       |
 | `tensor.clamp_max(max)`                                         | `torch.clamp(tensor, max=max)`                |
--- a/crates/burn-backend-tests/tests/tensor/int/ops/topk.rs
+++ b/crates/burn-backend-tests/tests/tensor/int/ops/topk.rs
@@ -58,6 +58,26 @@ fn test_topk() {
 }
 
 #[test]
+fn test_argtopk_1d() {
+    let tensor = TestTensorInt::<1>::from([1, 2, 3, 4, 5]);
+
+    let indices = tensor.argtopk(3, /*dim*/ 0);
+
+    let expected = TensorData::from([4, 3, 2]);
+    indices.into_data().assert_eq(&expected, false);
+}
+
+#[test]
+fn test_argtopk() {
+    let tensor = TestTensorInt::<3>::from([[[1, 4, 7], [2, 5, 6]], [[3, 0, 9], [8, 2, 7]]]);
+
+    let indices = tensor.argtopk(2, /*dim*/ 2);
+
+    let expected = TensorData::from([[[2, 1], [2, 1]], [[2, 0], [0, 2]]]);
+    indices.into_data().assert_eq(&expected, false);
+}
+
+#[test]
 fn test_topk_with_indices_supports_negative_dim_int() {
     let tensor = TestTensorInt::<3>::from([[[1, 4, 7], [2, 5, 6]], [[3, 0, 9], [8, 2, 7]]]);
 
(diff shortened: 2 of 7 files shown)


laggui (approved):
Thanks!

Good catch on the book, looks like we missed the addition the first time around.

PR #5238, opened 2026-07-27
feat(tensor): add mask_select (boolean mask indexing)

### LINKED ISSUE #3669: Boolean mask indexing
### Feature description

Boolean mask indexing, similar to:

```py
inds = scores > self.conf_threshold
boxes, landmarks, scores = boxes[inds], landmarks[inds], scores[inds]
```

From [discord](https://discord.com/channels/1038839012602941528/1038839013735399547/1413497627772719317)

Yashiru:
# PR : masked_select (boolean mask indexing)

**Title:** `feat(tensor): add mask_select (boolean mask indexing)`

**Base:** `tracel-ai:main` ← **Compare:** `Yashiru:feat/masked-select`

---

### Checklist

### Related Issues/PRs

Closes #3669.

### Changes

This implements the approach I proposed in #3669: mirror the existing `nonzero`/`argwhere` design
and keep the PR small. Digging into the codebase refined it in two ways: no new shared helper is
needed at all (the op composes entirely from existing operations), and I ended up diverging from
my own naming proposal to follow a repo convention I discovered along the way (see the naming note
below).

Adds `mask_select`, the equivalent of numpy `tensor[mask]` / `torch.masked_select(tensor, mask)`:
it returns a 1D tensor containing the elements of `self` where the same-shaped `Bool` mask is
`true`, in the order of the flattened input tensor.

Because the number of selected elements is data-dependent, the output shape can't be known without
reading the mask, so this makes the same synchronous-read tradeoff already accepted by `argwhere`
and `nonzero`. A `mask_select_async` variant is provided for non-blocking contexts (e.g. wasm),
mirroring the `nonzero`/`nonzero_async` pattern.

The implementation is a small composition of existing ops, with **no backend trait changes**:

```rust
pub async fn mask_select_async(self, mask: Tensor<D, Bool>) -> Tensor<1, K> {
    check!(TensorCheck::mask_select(&self.shape(), &mask.shape()));
    let indices = mask.flatten::<1>(0, D - 1).argwhere_async().await.squeeze_dim::<1>(1);
    self.flatten::<1>(0, D - 1).select(0, indices)
}
```

Notes:
- Purely additive: no changes to existing APIs, no breaking change.
- Lives in the `K: Basic` block, so it works for `Float`, `Int`, and `Bool` tensors.
- The gradient composes from the existing `reshape`/`select` backwards; no autodiff code is needed
  (covered by autodiff tests).
- `mask` must have the same shape as `self`, enforced by a `TensorCheck` (a mismatched mask would
  otherwise select the wrong elements, or read out of bounds on backends without bounds checks).
  Unlike `torch.masked_select`, the mask is not broadcast; that could be added later if wanted.
- An all-`false` mask yields an empty `[0]` tensor with no special-casing (via `argwhere`'s
  `[0, 1]` shape).
- The docs explicitly flag each call as a host-device synchronization point, with guidance to keep
  it on final results rather than inside performance-critical loops.

**Naming note:** in the issue I proposed `masked_select` for PyTorch parity, but while studying
the codebase I noticed burn deliberately renamed `masked_fill` to `mask_fill` and `torch.where` to
`mask_where`, so I followed the local `mask_*` convention with `mask_select`. Discoverability is
preserved by the book table, which maps it to `tensor.masked_select(mask)` exactly like the
`mask_fill` row. Happy to switch back if you prefer PyTorch-parity naming, it's a quick change.

**Possible follow-ups**, out of scope here: the main cost on GPU backends is inherited from the default
`bool_argwhere` (full mask readback + host loop). A device-side `bool_argwhere` for cubecl (stream
compaction, count-only readback) would speed up `argwhere`, `nonzero`, and `mask_select` at once;
a dedicated fused kernel could come on top of it. I'd be happy to tackle that next if there's
interest.

### Testing

- `tensor/float/ops/mask_select.rs`: 1D, 2D, 3D (flattened order across dims), single-element
  (exercises the `squeeze_dim` path), empty mask (forces execution via `into_data`), transposed /
  flipped / narrowed inputs (non-contiguous views), and a `#[should_panic]` shape-mismatch case.
- `tensor/int/ops/mask_select.rs` and `tensor/bool/ops/mask_select.rs`: `Int` and `Bool` kinds
  (the `Bool` source goes through the dedicated bool select dispatch path).
- `autodiff/mask_select.rs`: gradients flow to the selected positions (zeros elsewhere), plus the
  all-`false` mask backward (zero-size scatter-add).
- The doc example runs as a doctest.
- `cargo run-checks` passes locally.

### DIFF
--- a/burn-book/src/building-blocks/tensor.md
+++ b/burn-book/src/building-blocks/tensor.md
@@ -215,6 +215,7 @@ Those operations are available for all tensor kinds: `Int`, `Float`, and `Bool`.
 | `tensor.into_primitive()`                            | N/A                                                                       |
 | `tensor.into_scalar()`                               | `tensor.item()`                                                           |
 | `tensor.mask_fill(mask, value)`                      | `tensor.masked_fill(mask, value)`                                         |
+| `tensor.mask_select(mask)`                           | `tensor.masked_select(mask)`                                              |
 | `tensor.mask_where(mask, value_tensor)`              | `torch.where(mask, value_tensor, tensor)`                                 |
 | `tensor.movedim(src, dst)`                           | `tensor.movedim(src, dst)`                                                |
 | `tensor.narrow(dim, start, length)`                  | `tensor.narrow(dim, start, length)`                                       |
--- a/crates/burn-backend-tests/tests/autodiff/mask_select.rs
+++ b/crates/burn-backend-tests/tests/autodiff/mask_select.rs
@@ -1,0 +1,50 @@
+use super::*;
+use burn_tensor::TensorData;
+
+#[test]
+fn test_mask_select_grad() {
+    // Gradients must flow back to the selected positions and be zero elsewhere.
+    let device = AutodiffDevice::new();
+    let tensor_1 = TestTensor::<2>::from_data(
+        TensorData::from([[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]),
+        &device,
+    )
+    .require_grad();
+    let mask = TestTensorBool::<2>::from_data(
+        TensorData::from([[true, false, true], [false, true, false]]),
+        &device,
+    );
+
+    // Selected values x = [0, 2, 4]; y = x * x, so dy/dx = 2x = [0, 4, 8].
+    let tensor_2 = tensor_1.clone().mask_select(mask);
+    let tensor_3 = tensor_2.clone() * tensor_2;
+
+    let grads = tensor_3.backward();
+
+    let grad_1 = tensor_1.grad(&grads).unwrap();
+
+    grad_1
+        .into_data()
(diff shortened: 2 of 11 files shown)


laggui (approved):
LGTM as a first implementation.

We should at least add the tensor ops trait with dispatch even if most backends will currently rely on the argwhere + select default. `burn-tch` can override the default, and other backends can eventually do the same.

Probably worth a separate follow-up issue.

Yashiru:
Thanks for the review and the merge!

Opened #5248 as the follow-up for promoting `mask_select` to a backend tensor op with dispatch,
with the default staying on the `argwhere` + `select` composition and a native `burn-tch` override
to start. I'll follow up with a PR there.

PR #5239, opened 2026-07-27
Skip unsupported accelerated kernels during autotune

SamuelBelanger:
## Changes

- **Matmul** (`burn-cubecl`, `burn-cubecl-fusion`): each accelerated strategy now carries its `TileMatmulKind` (`Cmma`/`Mma`). Its group priority returns `PRIORITY_NEVER` when `TileMatmulKind::supported_sizes` is empty for the key's element types (`elem_lhs`/`elem_rhs`/`elem_out` via `MatmulElems::from_globals`).
- **Attention** (`burn-cubecl`): the accelerated flash-attention tunables return `PRIORITY_NEVER` when the client exposes no `matmul.cmma`, `matmul.mma`, or `tma` features.
- The compute client is cloned into the `TUNER.init` closure so priority functions can query device properties.

Negative intra-group priority is dropped by `TunePlan` before the tunable is ever handed to the benchmark loop, so unsupported kernels are never compiled. The win is autotune time on devices without CMMA/MMA (currently they compile every accelerated variant only to fail or run slowly).

## Notes 

- Matmul TMA strategies are pruned by their tile kind only; TMA support itself is not checked, so a CMMA-capable device without TMA still compiles them.

### DIFF
--- a/crates/burn-cubecl-fusion/src/optim/matmul/tune.rs
+++ b/crates/burn-cubecl-fusion/src/optim/matmul/tune.rs
@@ -13,7 +13,8 @@ use cubecl::{
     tune::{LocalTuner, Tunable, TunableSet, TuneGroup, local_tuner},
 };
 use cubek::matmul::{
-    definition::MatmulKind,
+    components::tile::TileMatmulKind,
+    definition::{MatmulElems, MatmulGlobalElems, MatmulKind},
     strategy::{MatmulAutotuneKey, MatmulGlobalScale, should_tune_double_buffering},
 };
 use serde::{Deserialize, Serialize};
@@ -32,9 +33,10 @@ pub fn fused_matmul_autotune<R: Runtime>(
     optimization: MatmulOptimizationTuneArg<R>,
     context: &mut Context<CubeFusionHandle<R>>,
 ) {
+    let tune_client = optimization.info.client.clone();
     static TUNER: LocalTuner<FusedMatmulAutotuneKey, CubeTuneId> = local_tuner!();
 
-    let tunables = TUNER.init(|| {
+    let tunables = TUNER.init(move || {
         const PRIORITY_MAX: i8 = 3;
         const PRIORITY_HIGH: i8 = 2;
         const PRIORITY_MEDIUM: i8 = 1;
@@ -172,6 +174,10 @@ pub fn fused_matmul_autotune<R: Runtime>(
 
         // Accelerated matmuls
         for tile_matmul in [AcceleratedTileKind::Cmma, AcceleratedTileKind::Mma] {
+            let tm = match tile_matmul {
+                AcceleratedTileKind::Cmma => TileMatmulKind::Cmma,
+                AcceleratedTileKind::Mma => TileMatmulKind::Mma,
+            };
             for (selector, double_buf, extra_group) in [
                 (
                     FusedMatmulSelector::Simple {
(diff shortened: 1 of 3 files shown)


ThierryCantin-Demers:
`tile_matmul_supported` checks the global dtypes, but the kernel runs on the register dtypes `adjust_dtypes` derives. f32 gets promoted to tf32, so as written this reports every f32 problem as unsupported and `PRIORITY_NEVER`s the accelerated candidates, dropping the tf32 cmma path we use today. Fix is to run `adjust_dtypes(client, &mut elems, tile_matmul.requires_accelerator())` before the check, in both copies.

Careful though: that fix alone takes `burn-backend-tests --features cuda matmul` from 41 pass / 0 fail to 40 / 1 (`test_float_matmul_vecmat_transposed_fused`). The `accelerated` group gives `PRIORITY_MAX` to `MatVec | VecMat`, so the newly-eligible f32 candidates get tried first on a vecmat, all fail, and the plan empties. The broken check was accidentally keeping vecmat off tensor cores, so this needs pairing with a MatVec/VecMat gate.

Will have to do a bit more investigating to find the proper solution.

SamuelBelanger:
> `tile_matmul_supported` checks the global dtypes, but the kernel runs on the register dtypes `adjust_dtypes` derives. f32 gets promoted to tf32, so as written this reports every f32 problem as unsupported and `PRIORITY_NEVER`s the accelerated candidates, dropping the tf32 cmma path we use today. Fix is to run `adjust_dtypes(client, &mut elems, tile_matmul.requires_accelerator())` before the check, in both copies.
> 
> Careful though: that fix alone takes `burn-backend-tests --features cuda matmul` from 41 pass / 0 fail to 40 / 1 (`test_float_matmul_vecmat_transposed_fused`). The `accelerated` group gives `PRIORITY_MAX` to `MatVec | VecMat`, so the newly-eligible f32 candidates get tried first on a vecmat, all fail, and the plan empties. The broken check was accidentally keeping vecmat off tensor cores, so this needs pairing with a MatVec/VecMat gate.
> 
> Will have to do a bit more investigating to find the proper solution.

Good, I fixed all three places (matmul, fused matmul and attention).
Two changes:

1. `tile_matmul_supported` runs `adjust_dtypes` before supported_sizes, so it queries the register types the kernel actually runs on.
2. Unsupported → `PRIORITY_MIN` instead of `PRIORITY_NEVER`. Negative within-group priorities get discarded by TunePlan, 0 keeps them as the last bucket, so they're effectively never picked but the plan can't end up empty.

I don't have a CUDA device. could you run `burn-backend-tests --features cuda matmul` (plus attention) on this?

ThierryCantin-Demers:
Ran it on a 4070 Ti SUPER at `3b8889134` with all autotune caches cleared (`crates/*/target`, `~/.cache/cubecl`, `target/autotune`). `--features cuda matmul` is 41/0 on `tests/tensor.rs` and 18/0 on both autodiff targets, 0 failures anywhere. `test_float_matmul_vecmat_transposed_fused` passes, so disregard my regression warning above — that 40/1 was a stale autotune cache on my end.

I was also wrong about `PRIORITY_MIN` being weaker than `PRIORITY_NEVER`. The tuner only descends to the next batch when a whole batch fails to queue anything, so a priority-0 tunable sits in the last bucket and never gets compiled while anything above it works. Your version keeps the autotune-time win and avoids the empty-plan panic.

`--features cuda attention` is 13/0 on f32 and 12/1 on f16 (`test_attention_all_options`), but that failure is pre-existing noise, not yours: it passes when run alone, and running the same filter three times on main's attention gate with caches cleared gave three different failure sets (`{all_options}`, `{bias_broadcast_heads_only, bool_mask_broadcast, all_options}`, `{bool_mask_broadcast}`). Attention accuracy is nondeterministic on CUDA depending on which routine autotune lands on.

Also worth noting the matmul fix matches what cubek already does one level down: f7f703c5 changed `CmmaRoutine::select` to key its `MmaConfig` lookup on the register accumulator instead of the global out type, with `mandated_acc()` deriving it from `MatmulElems::from_globals`. So the gate is now asking the same question the launch path asks.

LGTM.

laggui (approved):

PR #5240, opened 2026-07-27
Feat/custom fusion optimization

nathanielsimard:
Fusion optimizations for cubecl backends are now open for extension: a single FusedOperation trait (typed name/state, run with fallback access) that built-in and user-defined optimizations implement alike, wrapped in a CubeOptimization type that hides the erasure, with serialized execution-plan state restored by name through the same mechanism. A registry in burn-cubecl — seeded with the built-ins (ElementWise, Matmul, Reduce, ReduceBroadcasted, NHWCRelayout) as ordinary providers — lets programs register/remove optimizations per runtime at startup, sealing once the fusion service starts, with burn-fusion left untouched.


laggui (approved):
Very nice!

Would be great to revamp the backend extension section in the book and include this.

PR #5242, opened 2026-07-28
Tensor::empty_like(self) -> Tensor

crutcher:
### Checklist

Symmetry.


laggui (approved):
Towards a more unified API!

PR #5243, opened 2026-07-28
refactor(burn-tensor): remove duplicate dimension checks

mattisonchao:
## Motivation
The Tensor API now normalizes dimension arguments through `AsIndex` before reaching backend calls. Several `TensorCheck` helpers repeated the same bounds validation immediately after `try_dim_index`, which makes dimension validation ownership less clear and leaves dead validators after `burn-nn` delegates into Tensor APIs.

This cleanup keeps the remaining shape and rank validation in `TensorCheck`, while relying on `AsIndex` and `unwrap_dim_index` for dimension bounds.

## Modifications
- Removed redundant dimension bounds checks after `try_dim_index` in aggregation, sorting, selection, stacking, concatenation, gather/scatter, cross, and take paths.
- Deleted no-longer-used `TensorCheck` helpers for duplicate dimension validation.
- Kept validators that still check output rank, empty inputs, and shape compatibility.

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -186,7 +186,7 @@ $$
 /// # Panics
 /// - If `dim` is outside [-D, D)
 pub fn softmax<const D: usize>(tensor: Tensor<D>, dim: impl AsIndex) -> Tensor<D> {
-    let dim = unwrap_dim_index(dim.try_dim_index(D));
+    let dim = unwrap_dim_index(dim.try_dim_index(D), "Softmax");
     Tensor::new(softmax_impl(tensor.primitive, dim))
 }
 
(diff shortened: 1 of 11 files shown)


laggui (commented):
Cleanup looks good to me!

I think it would be helpful to preserve the operation origin, maybe we could have `unwrap_dim_index` take an origin str to register with the `TensorCheck`.
```rust
pub(crate) fn unwrap_dim_index<E>(result: Result<usize, E>, op: &str) -> usize
where
    E: core::fmt::Display,
{
    match result {
        Ok(dim) => dim,
        Err(error) => {
            macros::check!({
                TensorCheck::Ok.register(op, TensorError::new(error.to_string()))
            });
            unreachable!()
        }
    }
}
```

This way, the error message would be:
```diff
=== Tensor Operation Error ===
- Operation: 'Dimension Indexing'
+ Operation: 'Select'
  Reason:
    1. BoundsError: dimension -3 out of bounds: 0..2
```

(if we take `select` as an example)

Can you make this change across what you've unified so far?
Thanks

mattisonchao:
HI, @laggui  

Sorry for the unexpected force-push. 
I applied the comment. Could you please help review it?

laggui (approved):
LGTM. Thanks for addressing my comment 🙂

PR #5245, opened 2026-07-28
feat(tensor): support assign indexing updates for floats and ints

mattisonchao:
## Motivation
`scatter` and `select_assign` accept `IndexingUpdateOp`, but the bridge/router/fusion path only allowed `Add`, even though overwrite-style indexed updates are already supported by `scatter_nd`. This follows the `ScatterElements` and native `scatter_nd` / `gather_nd` work from tracel-ai/burn-onnx#131 and tracel-ai/burn#4709 by adding `Assign` support to the plain indexed update APIs for float and int tensors.

Bool tensors keep the existing `Add` / logical-or behavior, and `Mul`, `Min`, and `Max` remain explicitly unimplemented for these plain APIs so they can be handled in separate operation-focused PRs.

## Modifications
- Add update-aware float/int backend trait methods that delegate `Add` to the existing `*_add` implementations by default.
- Forward `IndexingUpdateOp` through bridge, dispatch, router, and fusion for `scatter` and `select_assign`.
- Implement `Assign` for ndarray float/int `scatter` and `select_assign`.
- Support float autodiff gradients for `scatter(..., Assign)` and `select_assign(..., Assign)`, and delegate int autodiff indexed updates to the inner backend.
- Document `Add` vs `Assign` behavior and duplicate-index semantics for the public tensor APIs.
- Add ndarray-scoped backend tests and autodiff tests for `Assign` on `scatter` and `select_assign`.

### DIFF
--- a/crates/burn-autodiff/src/ops/int_tensor.rs
+++ b/crates/burn-autodiff/src/ops/int_tensor.rs
@@ -223,6 +223,16 @@ impl<B: Backend, C: CheckpointStrategy> IntTensorOps<Self> for Autodiff<B, C> {
         B::int_scatter_add(dim, tensor, indices, value)
     }
 
+    fn int_scatter(
+        dim: usize,
+        tensor: IntTensor<B>,
+        indices: IntTensor<B>,
+        value: IntTensor<B>,
+        update: IndexingUpdateOp,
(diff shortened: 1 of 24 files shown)


laggui (approved):
LGTM.

Related: #4477

PR #5259, opened 2026-07-30
feat(tensor): add batched SVD decomposition to linalg

sehaxe:
## What this PR does

Adds `linalg::svd` to burn as a full Golub-Kahan pipeline with LAPACK-grade accuracy, no native kernels:

1. **Golub-Kahan bidiagonalization** with Householder reflections (`A = U1 B V1^T`, `B` upper bidiagonal).
2. **Bidiagonal QR with shifts** (LAPACK `dbdsqr`, Wilkinson shift via `dlas2`, Givens via `dlartg`) with deflation and sweep limits. Since `B` is only `2n` scalars, this stage runs on the host.
3. The Givens rotations are applied to the accumulated U1/V1 factors directly (no extra matmuls), factors are sorted descending, numerical zeros masked relative to `sigma_max`, wide inputs go through the transpose, f16/bf16 upcast like `det`.

Earlier iterations of this PR used power iteration + deflation (then one-sided Jacobi); both are unreliable with clustered/repeated singular values and were inaccurate on fused CUDA because of the autotune issue below.

**Design choice**: the whole pipeline runs on the host over tensor data (`into_data` / `from_data`). On fused CUDA, tensor-op implementations pay ~0.3-0.7 ms of dispatch overhead per operation (measured), so a ~2000-op bidiagonalization costs seconds; the host path is deterministic, backend-independent and ~200x faster at 128x128. Sorting, zero-masking, permutation and the wide-input factor swap also run on the host: GPU gather/argsort on views turned out to be nondeterministic for equal keys and can fail with "strides are not supported" on fused CUDA.

## Accuracy vs PyTorch (RTX 5060 Ti, f32, release build)

Reconstruction error `max|A - U S Vt|`:

| size   | torch CPU | torch GPU | burn (ndarray) | burn (CUDA) |
|--------|-----------|-----------|----------------|-------------|
| 4x4    | 5.4e-7    | 1.2e-6    | 4.8e-7         | 3.6e-7      |
| 32x32  | 2.9e-6    | 2.7e-5    | 3.2e-6         | 5.7e-6      |
| 128x128| 4.5e-6    | 1.1e-4    | 8.3e-6         | 1.1e-5      |

burn is consistently more accurate than torch GPU (which uses cuSOLVER f32) and on par with torch CPU. For reference, the one-sided Jacobi variant from an earlier iteration measured 1.1e-3 reconstruction error at 8x8 on CUDA. Singular values and factors are compared against torch.linalg.svd reference values in tests (tolerance 5e-4, signs aligned since they are arbitrary in SVD).

## Performance vs PyTorch (same machine, release build)

Per-call time, single matrix, 15 sweeps:

| size   | torch CPU | torch GPU | burn (ndarray) | burn (CUDA) |
|--------|-----------|-----------|----------------|-------------|
| 4x4    | 12 us     | 231 us    | 3.4 us         | 40 us       |
| 8x8    | 17 us     | 218 us    | 6.5 us         | 60 us       |
| 16x16  | 59 us     | 359 us    | 31 us          | 72 us       |
| 32x32  | 223 us    | 331 us    | 158 us         | 210 us      |
| 64x64  | 1.26 ms   | 1.53 ms   | 1.04 ms        | 1.22 ms     |
| 128x128| 2.49 ms   | 3.10 ms   | 10.0 ms        | 9.6 ms      |

burn is faster than torch GPU on every size up to 64x64 (torch pays fixed kernel-launch overhead; a host copy + CPU compute has none), then falls behind: within ~2.7-3x of cuSOLVER at 128x128 and ~15x on tall 512x128, with no native kernel (torch calls cuSOLVER, tracked in #1538). The tensor-op bidiagonalization that preceded this design took 525 ms (ndarray) / 1.9 s (CUDA) at 128x128 in the same bench.

**Why the CUDA backend is host-computed.** `svd` runs on the CPU on every backend, including CUDA: the input is pulled with `into_data`, the pipeline runs, the result is written back. The GPU sits idle for this call. This is a deliberate trade-off, not an omission:

- A tensor-op bidiagonalization pays 0.3-0.7 ms of fusion dispatch overhead per operation, so ~2000 ops cost ~1.9 s at 128x128 (measured, reverted).
- `matmul` cannot be used on this GPU at all: cubecl autotune picks BF16-accumulating kernels (see below), which silently corrupts results.
- So the host path is the only route with acceptable accuracy today; the data transfer is negligible (64 KB each way at 128x128) and small sizes still beat torch GPU because there is no launch overhead.

Supporting CUDA is about the burn contract: `linalg::svd` must work on any backend, and it does - transparently, on GPU-resident tensors. When a native kernel or accurate matmul becomes available (#1538), the computation moves to the GPU with no API change.

## The cubecl autotune accuracy issue (root cause found)

The intermittent test failures in earlier iterations of this PR were real inaccuracies, not nondeterminism: on this GPU (RTX 5060 Ti), cubecl autotune picks matmul kernels that accumulate in BF16 for many shapes. Measured in isolation:

- `[k,k] @ [k,1]`: error <= 1e-4 for k <= 16, then jumps to 0.03-0.75 for k >= 24 (BF16 mantissa).
- First reported as an 8x8-only flake; in fact 8x8 Jacobi converged to 1.1e-3 with the autotune-picked kernel and to ~3e-7 with `CUBECL_AUTOTUNE_CACHE=off`.

`svd` sidesteps this entirely: it never calls `matmul` (not even for the U/V assembly) and runs on the host, so accuracy is backend-independent.

## Notes

- Two `Device::sync()` calls (before and after the host stage) on fused CUDA: cubecl host reads fail with "strides are not supported" while other kernels are still queued (this also covers tests that never read their outputs). No-op on eager backends.
- Input views (clone/transpose) are materialized with `reshape` before the host read, for the same reason.
- Numerical zeros are detected relative to `sigma_max` (sigma <= 10 * eps * sigma_max, with `eps` the dtype machine epsilon), because absolute epsilons behave differently across backends.
- Host pipeline: bidiagonalization is O(m n^2) scalar math; dbdsqr converges in ~2.5 QR steps per singular value on average. All norms/rotations/shifts are scale-invariant (LAPACK-style): inputs with entries up to f32::MAX and down to subnormals (e.g. diag(1e38, 1, 1) or diag(1e-40, 1, 1)) stay finite and accurate.
- The dbdsqr Wilkinson shift is computed in f64 for f32 inputs: the f32 closed form carries ~1e-5 error, which exceeds the deflation threshold and stalls 2x2 blocks with nearly-equal singular values (about 2.5% of random tall 512x128 matrices lost ~6% of the Frobenius norm before this fix; 100-matrix sweeps now worst-case at 1.6e-5).
- Empty inputs (zero leading dimension) return empty reduced SVD tensors instead of panicking.
- Known cubecl limitation (not specific to this PR): parallel test threads doing host reads on the same CUDA device can race; the CI runs GPU tests with `--test-threads=1` (xtask remote), which is stable.

## Testing

27 tests in `burn-backend-tests/tests/tensor/float/linalg/svd.rs`:

- Reconstruction: square, rectangular, wide, 1x1, single row/column, batched (including mixed and wide batches)
- Factor orthonormality (U^T U = I, Vt Vt^T = I)
- Known values: diagonal, identity, rank-1, zero, clustered, ill-conditioned (Hilbert 5x5), random
- Ordering (descending, non-negative), scaling and transpose invariance
- Convergence: more sweeps improve accuracy (fixed input, deterministic across backends)
- f16/bf16 dtype roundtrip (upcast like `det`)
- Empty matrices (zero leading dimension)
- Panics on invalid input (bad generic rank, 1-D)
- torch.linalg.svd reference values (LAPACK gesdd): singular values and U/Vt factors, signs aligned

Plus 21 unit tests in `crates/burn-tensor/src/tensor/linalg/svd.rs` covering the host pipeline directly: f64 accuracy vs numpy/LAPACK (1e-12), rank-1, 1x1, batched, f32/f64 extreme scales (1e38/1e-40, overflow/underflow), a seeded 60-matrix f32 sweep over tall 512x128 inputs, a fixed tall matrix, and a 2x2 closed-form check. An ignored stress test exercises 4x128..256x256 random, rank-deficient, clustered, zero and extreme-scale matrices on ndarray and CUDA with host-computed reconstruction/orthogonality/Frobenius checks (matmul in the harness itself is avoided for the same autotune reason).

Green on ndarray and CUDA (incl. fusion config); 5+ consecutive serial CUDA runs and 8+ serial ndarray runs. Ignored benchmark `bench_svd_vs_torch` times all sizes on the active backend.

## Things we tried and reverted during development

- **Tensor-op bidiagonalization + host dbdsqr**: ~1.9 s at 128x128 on CUDA: each tensor op costs 0.3-0.7 ms of fusion dispatch overhead, ~2000 ops dominate. Reverted in favor of the all-host pipeline.
- **GPU argsort/gather for the sort and permutation**: nondeterministic for equal sigma (parallel sort) and view-based reads that can fail with "strides are not supported" on fused CUDA. Reverted in favor of a stable host sort.
- **Global zero-perturbation in dbdsqr** (diagonal AND superdiagonal set to eps * sigma_max): on inputs like diag(1e38, 1, 1) in f32 the superdiagonal floor becomes 1e31, the sweeps diverge to NaN. Now only diagonal exact zeros are perturbed; superdiagonal zeros deflate naturally.
- **One-sided Jacobi**: slow (n^2 rotations, full-tensor gather/scatter) and its small matmuls hit the BF16 autotune kernels (8x8 -> 1.1e-3 on CUDA).
- **Matmul-based Gram reductions** (`ones @ x` instead of powf+sum_dim): ~8x faster in isolation, but degraded to ~9e-4 in the test binary (BF16 accumulation).
- **Delta write-back** (`a += new - old`, 2 scatter-adds instead of 4): faster, but loses precision for tiny rotation angles.

**Question for maintainers**: can autotune exclude inaccurate kernel candidates (BF16 accumulation) per shape, or is this a cubecl issue to fix upstream? `svd` currently sidesteps it by never calling `matmul`.

## Benchmark environment

- CPU: AMD Ryzen 5 5500 (6 cores / 12 threads)
- GPU: NVIDIA GeForce RTX 5060 Ti 16 GB (driver 610.43.03)
- torch 2.11.0+cu128 (CPU: LAPACK gesdd, GPU: cuSOLVER)
- burn: release build, burn-ndarray and burn-cuda (CubeCL), 15 sweeps, f32


crates/burn-tensor/src/tensor/linalg/svd.rs
```diff
@@ -0,0 +1,829 @@
...
+        for (t, &src) in order.iter().enumerate() {
+            for i in 0..m {
+                pu[i * n + t] = u[b * m * n + i * n + src];
+            }
+            for i in 0..n {
+                pvt[t * n + i] = vt[b * n * n + src * n + i];
+            }
+            // read from the untouched slot: sigma is rewritten in place below
+            let v = sigma[b * n + src];
+            sorted[t] = if v.abs() <= smax * F::from(1e-6).unwrap() {
```
laggui: I think the eps value should be derived from `F` instead. This could be throwing away a lot of precision in e.g. `f64`.

crates/burn-tensor/src/tensor/linalg/svd.rs
```diff
@@ -0,0 +1,829 @@
...
+///    shrinking submatrices.
+/// 2. **Bidiagonal QR with shifts** (the LAPACK `dbdsqr` algorithm) to
+///    diagonalize `B`. Only `2n` scalars are involved, so this stage runs
+///    on the host over the tensor data (deterministic, no kernels) and
+///    converges in ~2.5 QR steps per singular value. The accumulated Givens
+///    rotations rebuild `U` and `V` from the bidiagonalization factors.
+///
+/// Both stages are exact to machine precision; the number of QR iterations
+/// is governed by a convergence criterion (the `sweeps` argument is kept
+/// for API compatibility and bounds the iteration count).
```
laggui: I think "exact to machine precision" might be somewhat overstated. And "converges in ~2.5 QR steps" is more of an empirical average, not a guarantee. We could slightly reformulate as:




```suggestion
/// # Algorithm
///
/// Two stages, mirroring the LAPACK `gesvd` / `dbdsqr` structure:
///
/// 1. **Golub-Kahan bidiagonalization** using Householder reflections (`A = U1 B V1^T`).
/// 2. **Implicitly shifted bidiagonal QR iteration** (LAPACK `dbdsqr`) to diagonalize `B`.
///
/// The algorithm is backward stable to within a small multiple of machine precision.
/// Convergence typically requires ~2.5 QR sweeps per singular value on average, bounded
/// by the `sweeps` argument.
```

crates/burn-tensor/src/tensor/linalg/svd.rs
```diff
@@ -0,0 +1,829 @@
...
+                assert!(x.is_finite(), "sigma not finite: {:?}", s);
+            }
+            let err = recon_err::<f32>(&u, &s, &vt, &a, 3, 3);
+            assert!(err.is_finite(), "recon not finite {err}");
+            assert!(err <= a[0].abs().max(1.0) * 1e-4, "recon err {err}");
+        }
+    }
+}
+
+#[test]
```
laggui: The remaining tests should probably still be encapsulated in the `mod tests` above

crates/burn-tensor/src/tensor/linalg/svd.rs
```diff
@@ -0,0 +1,829 @@
...
+    #[test]
+    fn test_svd_host_f64() {
+        // 4x3 full-rank matrix with known torch f64 values
+        let a = [
+            1.0f64, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0,
+        ];
+        let (u, s, vt) = svd_host::<f64>(&a, 4, 3, 1, 30, false);
+        // svd_host returns sigma in diagonalization order; sort before comparing.
+        let mut ss = s.clone();
+        ss.sort_by(|x, y| y.partial_cmp(x).unwrap());
```
laggui: Doesn't the algo already sort the values in descending order? Maybe a stale comment?

laggui (changes requested):
I think this is a good default implementation to start, even if it is entirely on CPU.

sehaxe:
thanks a lot for the review! for now i will make it a draft and when i resolve those inconsistencies will open it again

sehaxe:
i resolved the issues you mentioned, thanks. maybe there are other suggestions? i personally dont like that it is just cpu, but we could figure out how to make gpu make sense

crates/burn-tensor/src/tensor/linalg/svd.rs
```diff
@@ -0,0 +1,820 @@
...
+
+    // Empty matrix (a zero leading dimension): the reduced SVD is empty too.
+    // Skip the pipeline (bidiagonalization would index out of bounds).
+    if m == 0 || n == 0 {
+        let mut du = [1; D];
+        du[..(D - 2)].copy_from_slice(&dims[..(D - 2)]);
+        du[D - 2] = n_rows;
+        du[D - 1] = 0;
+        let mut ds = [1; D1];
+        ds[D1 - 1] = 0;
```
laggui: For an empty batched input it seems the returned `ds` would be incorrectly constructed with a leading dim of 1, while the other tensors correctly retain the batch dimension. We should probably add similar logic:

```rust
let mut ds = [1; D1];
if D1 >= 2 {
    ds[..(D - 2)].copy_from_slice(&dims[..(D - 2)]);
}
ds[D1 - 1] = 0;
```

laggui (changes requested):
I think there's one remaining issue to fix, otherwise should be good to go.

>  i personally dont like that it is just cpu, but we could figure out how to make gpu make sense

That'd be great, you'd have to write a cubecl kernel for that. And wire it up through some tensor op trait for backends to override with a custom implementation.

sehaxe:
Rebased on main (0.22.0-pre.2) and optimized the host pipeline. Same algorithm (Golub-Kahan + dbdsqr, LAPACK structure), faster implementation:

## What changed

- **Closed forms for 1x1 / 2x2**: exact Blinn rotation formulas, skipping the full pipeline for the smallest sizes.
- **Row-major Givens application**: the rotations act on column pairs, which are stride-`n` in row-major storage but contiguous rows after transposing U1/V1 once up front. The column permutation of U and row permutation of Vt are folded into the final copy, dropping the intermediate permute buffers.
- **Allocation-free bidiagonalization**: reflection scratch (w, wta, aw) allocated once instead of per Householder step; sort scratch reused across the batch.
- **Fix**: empty batched input now keeps the batch dims on `S` (was `[1, ..]`), matching U/Vt.
- **Fix**: `bench_svd_realistic` passed the 3D batch cases (8x64x64, 16x128x128) through `svd::<2, 1>` and panicked on reshape; they now go through `svd::<3, 2>`.

## Speed (ndarray, release, RTX 3090 box)

| size | before | after | speedup |
|---|---|---|---|
| 1x4x4 | 0.004 ms | 0.004 ms | 1.0x |
| 8x4x4 | 0.019 | 0.019 | 1.0x |
| 1x32x32 | 0.174 | 0.153 | 1.1x |
| 8x32x32 | 1.426 | 1.155 | 1.2x |
| 1x128x128 | 9.55 | 5.30 | 1.8x |
| 8x128x128 | 76.4 | 42.0 | 1.8x |
| 1x512x512 | 1457 | 274 | 5.3x |
| 8x512x512 | 11819 | 2195 | 5.4x |

Accuracy unchanged: max reconstruction error ~1e-5 across all sizes (stress suite: rank-deficient, clustered, scaled 1e20/1e-30, zeros, f16/bf16, 256x256 vs torch f64).

## Testing

- `burn-tensor` unit tests: 196 passed.
- `burn-backend-tests` svd: 27/27 on ndarray and CUDA (`--test-threads=1`, same as the GPU CI runner uses via xtask).
- Stress + torch-reference tests (ignored by default): all green.

The empty-batch `S` dims and the 3D bench case were pre-existing issues; the parallel-CUDA flakiness in unrelated tests (lu/qr/maxpool/quantization) reproduces on main without this PR.

crates/burn-tensor/src/tensor/linalg/svd.rs
```diff
@@ -0,0 +1,957 @@
...
+        dv[D - 1] = n_cols;
+        let u_t = Tensor::<D>::from_data(TensorData::new(Vec::<f32>::new(), du), &device);
+        let s_t = Tensor::<D1>::from_data(TensorData::new(Vec::<f32>::new(), ds), &device);
+        let vt_t = Tensor::<D>::from_data(TensorData::new(Vec::<f32>::new(), dv), &device);
+        let (u_t, s_t, vt_t) = if needs_upcast || original_dtype == DType::F64 {
+            (
+                u_t.cast(original_dtype),
+                s_t.cast(original_dtype),
+                vt_t.cast(original_dtype),
+            )
```
laggui: Prefer using `Tensor::empty(shape, (&device, dtype))` instead

laggui (commented):
That's a bigger update than expected 😄 

Please apply the suggested clippy fixes highlighted by the failing CI

sehaxe:
All review threads are resolved and the CI clippy failures are fixed. Summary of the latest round:

## What changed since the last review

- **`FloatTensorOps::float_svd` trait hook**: the host pipeline (Golub-Kahan bidiagonalization + LAPACK-style `dbdsqr` + Givens assembly) moved from `burn-tensor` into `burn-backend` as the default implementation of a new `float_svd` method on the tensor-op trait, and `linalg::svd` now dispatches through the bridge. All backends inherit the host default and can override it with a native or fused implementation without changing the public API. `burn-dispatch` delegates to the active backend and the fusion backend resolves to its inner backend, so the hook is actually reachable.
- **Clippy (CI `code-quality`, `--deny warnings`)**: fixed everything the CI flagged — `needless_range_loop` in the norm accumulation and the wta zero-fill, `type_complexity` (factored the rotation list into `GivensRotation<F>`), `too_many_arguments` on the public `svd_postprocess`, and an accidental pub re-export of `argwhere`/`sort` helpers that tripped `missing_docs`. `cargo clippy --no-deps -- --deny warnings` is clean.
- **`Tensor::empty` for empty batched outputs** instead of `from_data` with an empty vec (review suggestion).
- The empty-batch `S` dims fix from the previous round is in (review thread resolved).
- **Small-input regression fixed**: the batch-threaded path queried `available_parallelism` (a ~40us syscall) even for single matrices; it now only runs when `batch > 1`. 4x4 went from ~90us back to ~3us.

## Speed

All measurements on this machine: **RTX 5060 Ti 16GB**, release build, best-of-N timings. Three versions of the PR for reference: the first submission (`b672fafb`, power-iteration SVD), the original Golub-Kahan implementation before the optimizations (`e19d33ad`), and the current head.

| size | first version | before optimizations | current (ndarray) | current (CUDA) |
|---|---|---|---|---|
| 1x4x4 | 5.12 ms | 0.003 ms | 0.003 ms | 0.068 ms |
| 1x32x32 | 38.3 | 0.168 | 0.110 | 0.155 |
| 1x128x128 | 370 | 9.49 | 2.81 | 2.58 |
| 1x512x512 | 12601 | 1448 | 119.5 | 109.4 |
| 8x512x512 | — (no batch support) | 11652 | 277.7 | 214.4 |

That is **105x faster than the first submission** and **12.1x faster than the pre-optimization version** on 1x512x512 (42x on the 8x512 batch). The CUDA column uses the host pipeline with device round-trips.

The latest round added two optimizations inside the bidiagonalization:

- **8-accumulator dot products** (`blk_sum`): plain `s = s + ...` reductions are forced to run scalar without fast-math, while 8 independent accumulators let LLVM vectorize the `aw` sweep of the right reflectors.
- **Transposed V1 layout inside `bidiag_host`**: V1 is kept as `V1ᵀ` (`v1t[j*n+i] = V1[i,j]`), so both the `vw = V1ᵀ w` sweep and the rank-1 update stream contiguous rows and vectorize. This replaced a scalar 134M-op reduction (on 512x512) over stride-n columns and dropped a full transposition pass in `svd_host_seq`.

For context, the same workload with torch 2.11 (cu128) on the same machine:

| size | torch CPU (MKL) | torch GPU (cuSOLVER) |
|---|---|---|
| 1x128x128 | 1.49 ms | 1.86 ms |
| 1x512x512 | 28.8 | 15.4 |
| 8x512x512 | 227 | 124 |

As noted in the docs, the pure-Rust host implementation is not competitive with tuned native libraries for large matrices; on 128x128 it is within ~2x of MKL and it does not depend on any external LAPACK.

## Accuracy

Max reconstruction error `A - U S Vt` is identical on CUDA and ndarray (~4e-6 on a random 128x128), unchanged from the host pipeline. Stress suite (rank-deficient, clustered singular values, scales 1e20/1e-30, zeros, f16/bf16, 256x256 vs torch f64) all green.

## Testing

- `burn-backend` unit tests: 22 passed (host pipeline moved there with its tests).
- `burn-backend-tests` svd: 27/27 on ndarray and CUDA (`--test-threads=1`, same as the GPU CI runner).
- Stress + torch-reference tests (ignored by default): all green.
- `cargo clippy --no-deps -- --deny warnings`: clean.

The `missing_docs` warning on `sort.rs` that appeared earlier was caused by my accidental re-export and is gone; nothing in `sort.rs` itself was touched.

On the GPU-kernel idea from the review: I prototyped a cubecl kernel (bidiagonalization + Givens on-device). It is bit-accurate up to 128x128, but the per-step block-wide norm reduction relies on `sync_cube`, which misbehaves beyond ~4 warps on the current cubecl, and the kernel ends up slower than the host path because of the launch overhead of ~10 kernels per step times n steps. I did not include it; the `float_svd` hook above is the wiring point for when that kernel lands (e.g. with a barrier-free warp reduction or a blocked bidiagonalization).

laggui (changes requested):
Moving the host implementation into burn-backend explains much of the diff, even if it wasn't mandatory for this PR. I'm fine with the backend hook and default implementation. However, the overall change still needs substantial cleanup.

The backend tests duplicate existing assertion infrastructure and include custom factor-alignment helpers, large Torch fixtures, ignored benchmarks, and stress/debug harnesses. Please reuse `TensorData::assert_approx_eq` where applicable and focus the tests on singular values, reconstruction, orthogonality, shapes, and a few edge cases. Direct comparison of U/V factors against torch is brittle because singular vectors are not unique.

Please reduce this PR to the default implementation and focused contract tests. We probably don't need to keep the ignored debug harness and stress tests.

sehaxe:
sure thank you!

sehaxe:
also i've been working on cubecl kernel lately should i open pull request for it after this one is merged?

sehaxe:
i cleaned up the tests per your review: torch fixtures and the sign-alignment helpers are gone ignored benches/stress harnesses are gone everything runs on `TensorData::assert_approx_eq` now. 31 contract tests, 0 ignored, test file went from 1032 down to 491 lines.

also profiled and reworked the host pipeline since givens application alone was ~70% of runtime so it got faster as a bonus:

| size | before | after |
|---|---|---|
| 128x128 | 2.83 ms | 2.19 ms |
| 256x256 | 17.1 ms | 11.0 ms |
| 512x512 | 119.5 ms | 75.2 ms |

small sizes unchanged. the simd layer behind this adds ~400 lines to the default implementation. i personally think its worth it for the speed, but if you want the diff leaner we could either trim it to f32 only (~150 lines less, f64 goes through the scalar path) or remove it completely (~400 lines less, performance back to what is currently pushed) your call

laggui (commented):
Thanks for cleaning up the contract tests. I want to clarify the intended scope before more work is added, because this keeps shifting 😅 

The default implementation in `burn-backend` should remain a straightforward, portable fallback. Architecture-specific SIMD, unsafe AVX2 kernels, runtime feature detection, and custom thread scheduling do not belong there. CPU optimizations should be implemented separately in a CPU backend such as `burn-flex`, through the backend override.

We should keep this PR focused so it can be merged. CubeCL and CPU-optimized implementations would be welcome as separate follow-up PRs.

sehaxe:
sorry for making it complicated than it should've been. Thanks for your time and effort!

laggui (approved):
Figured I'd take it over the line so it can land.

I updated with main and simplified the fallback SVD implementation by returning tensors directly, aligning the backend hook with other synchronous tensor operations, reducing temporary allocations, and writing outputs directly.

I'm thinking of moving the different `burn-tensor` modules (e.g. `linalg`, `signal`) to extension crates. This would separate the core operations, keep the backend contract clean, while still opening the door for future ops and backend implementations in these domains.

PR #5262, opened 2026-07-30
feat(quant)!: support two-level quantization scales

ThierryCantin-Demers:
Adds two-level quantization: per-block scales that are themselves normalized by a single per-tensor scale, so the block scales can live in a narrow dtype while the per-tensor scale absorbs the tensor's dynamic range. NVFP4 is the inspiration and the validation target, not the spec, so nothing here hardcodes E2M1 values, ue4m3 block scales, or a block size of 16.

The scheme comes from cubecl#1514; what this PR adds is everything burn needs to carry a second scale: calibration, the serialized byte layout, and the ndarray, flex and cube backends.

Before, one scale per block of 32 stored as f16, with the scale precision living in a scheme-wide field:

```rust
    .with_value(QuantValue::Q8S)
    .with_level(QuantLevel::block([32]))
    .with_param(QuantParam::F16);

let tensor = tensor.quantize_dynamic(&scheme);
```

After, each level owns the dtype its scales are stored in, and a second level is one more setter:

```rust
    .with_value(QuantValue::Q8S)
    .per_block([32], ScaleDtype::F16);

// Two levels: ue4m3 block scales, normalized by one per-tensor f32 scale.
let two_level = QuantScheme::default()
    .with_value(QuantValue::Q8S)
    .per_block([16], ScaleDtype::UE4M3)
    .per_tensor(ScaleDtype::F32);

let tensor = tensor.quantize_dynamic(&two_level);
```

## What changed in burn

- The per-tensor scale rides beside the block scales the whole way: `QuantizationParameters { scales, global }`, `QParams.global`, and a third allocation region in a `CubeTensor`. Backends keep the two apart on the tensor and fold them only where dequantization reconstructs the product, so a save and load round trip preserves each at its stored precision.
- Calibration returns both levels unrounded, as the one-level scales already did. Each backend rounds each level to what its dtype can store and quantizes against that product, so the values written already account for the rounding. The round-up rule (store the smallest value *not below* the scale, so a block maximum never clips) lives in cubecl and is called from both the host and the kernel path, so CPU and GPU reconstruct identically.
- `burn-std`: `params_shape` takes the scheme, `scale_to_param` becomes `scale_to_dtype`, and `global_scale_dtype` answers "does this scheme have a per-tensor scale over its blocks", which `tensor_scale()` alone cannot, since that also answers for a one-level per-tensor scheme.
- burn's `block_level` helper is gone. Swap and permute rewrite the block dimensions through the scheme's own `swap_block_dims`/`permute_block_dims`, which also fixes `burn-ir`'s `permute_quantized_dtype` silently skipping that rewrite for two-level tensors.

## Rejected rather than half-supported

- The fused elementwise kernels and the fused matmul decline a scheme with a per-tensor scale rather than dropping the factor, which would reconstruct every value short by it. The quantized matmul dequantizes a two-level operand and runs in float instead, so the path exists but buys nothing yet.
- `AbsMean` calibration. BitNet's gamma is a mean over the whole tensor or block, which a per-tensor scale cannot decompose.
- A per-tensor scale in anything but f32, and block scales already spanning f32's range, both refused where the scheme is chosen rather than at the first quantize.

## Does it earn its keep

The bar set before starting: beat what one-level already achieves at the same bits per element. That is 0.0054 relative error at 8.5 bits, from `block32` with f16 scales. f16 block scales match f32 to five decimals for a full bit per element less, so f16 is the honest baseline, not f32.

`block16` with ue4m3 block scales under an f32 per-tensor scale, 8.508 bits:

| weight std | two-level | one-level block32/f16 |
| ---------- | --------- | --------------------- |
| 1.0 | 0.00503 | 0.00539 |
| 0.1 | 0.00503 | 0.00540 |
| 0.02 | 0.00503 | 0.00539 |

The margin is about 7 percent, which on its own would be thin. The result that matters is the flatness, which holds to five decimals. Every one-level ue4m3 config degrades as weights shrink, with `block16/ue4m3` going 0.00512, 0.00710, 0.02858 across those same three magnitudes, because the block scale underflows what e4m3 can represent. Two-level does not, and a realistic `Linear` weight sits in that range. Numbers come from `report_quantization_accuracy` in the test harness, extended here to cover two-level.

## Known limit

The per-tensor scale takes the tensor's range out of the block scales, but the spread *between* blocks is still bounded. A block whose scale falls further below the largest one than the block dtype can express is stored at that dtype's smallest value and quantizes entirely to zero. UE4M3 spans about 2^18 this way, so a tensor holding a genuine outlier alongside ordinary values can lose the ordinary ones. Documented on the scheme upstream.

## Not in this PR

The matmul tile path, which is where a two-level operand would stop being a fallback. It is the largest remaining piece and belongs with the tile work upstream.

## Testing

Quantization suites green on ndarray, flex and CUDA, two-level cases included. wgpu has no e4m3, so the ue4m3 cases are gated off it.

### DIFF
--- a/burn-book/src/performance/quantization.md
+++ b/burn-book/src/performance/quantization.md
@@ -45,13 +45,12 @@ tensors and can collect their statistics, such as the min and max value when usi
 
 ```rust , ignore
 # use burn::module::Quantizer;
-# use burn::tensor::quantization::{Calibration, QuantLevel, QuantParam, QuantScheme, QuantValue};
+# use burn::tensor::quantization::{Calibration, QuantScheme, QuantValue, ScaleDtype};
 #
 // Quantization config
 let scheme = QuantScheme::default()
-    .with_level(QuantLevel::Block(32))
-    .with_value(QuantValue::Q4F)
-    .with_param(QuantParam::F16);
+    .per_block([32], ScaleDtype::F16)
+    .with_value(QuantValue::Q4F);
 let mut quantizer = Quantizer {
     calibration: Calibration::MinMax,
     scheme,
@@ -80,12 +79,25 @@ values, storage format, granularity, and how the values are scaled.
 
 ```rust
 let scheme = QuantScheme::default()
-    .with_mode(QuantMode::Symmetric)         // Quantization mode
-    .with_level(QuantLevel::block([2, 16]))  // Granularity (per-tensor or per-block)
-    .with_value(QuantValue::Q8S)             // Data type of quantized values, independent of how they're stored
-    .with_store(QuantStore::Native)          // Storage format for quantized values
-    .with_param(QuantParam::F16);            // Precision for quantization parameters
+    .with_mode(QuantMode::Symmetric)          // Quantization mode
+    .per_block([2, 16], ScaleDtype::F16)      // One scale per block of values, stored as f16
+    .with_value(QuantValue::Q8S)              // Data type of quantized values, independent of how they're stored
+    .with_store(QuantStore::Native);          // Storage format for quantized values
 ```
+
+A scheme carries up to two scale levels, each set in any order and each owning the type its scales
+are stored in. `per_tensor` is one scale for the whole tensor, which is also what a scheme with no
+level set resolves to; `per_block` is one scale per block of values.
+
+```rust
+// Block scales in ue4m3, normalized by a single per-tensor f32 scale.
+let scheme = QuantScheme::default()
+    .per_block([16], ScaleDtype::UE4M3)
+    .per_tensor(ScaleDtype::F32);
+```
+
+Two levels exist so the block scales can live in a narrow type: the per-tensor scale absorbs the
+tensor's dynamic range, and the block type only has to cover the spread between blocks.
 
 #### Quantization Mode
 
@@ -93,12 +105,15 @@ let scheme = QuantScheme::default()
 | :---------- | :------------------------------------------- |
 | `Symmetric` | Values are scaled symmetrically around zero. |
 
-#### Quantization Level
+#### Scale Levels
 
-| Level                          | Description                                                                                                  |
-| :----------------------------- | :----------------------------------------------------------------------------------------------------------- |
-| `Tensor`                       | A single quantization parameter set for the entire tensor.                                                   |
-| `Block(block_size: BlockSize)` | Tensor divided into blocks (1D, 2D, or higher) defined by block_size, each with its own quantization params. |
+| Level                        | Description                                                                                     |
+| :--------------------------- | :---------------------------------------------------------------------------------------------- |
+| `per_tensor(dtype)`          | A single scale for the entire tensor.                                                           |
+| `per_block(block, dtype)`    | Tensor divided into blocks (1D, 2D, or higher) defined by `block`, each with its own scale.     |
+
+Setting both nests the blocks inside the tensor: the per-tensor scale is the factor the block
+scales are relative to, and it has to be `F32`.
 
 #### Quantization Value
 
@@ -124,16 +139,16 @@ let scheme = QuantScheme::default()
 
 Native storage is not supported for sub-byte quantization values.
 
-#### Quantization Parameters Precision
+#### Scale Data Type
 
-| Param   | Description                                                                            |
+| Dtype   | Description                                                                            |
 | :------ | :------------------------------------------------------------------------------------- |
 | `F32`   | Full floating-point precision.                                                         |
 | `F16`   | Half-precision floating point.                                                         |
 | `BF16`  | Brain float 16-bit precision.                                                          |
 | `UE4M3` | 8-bit floating point (4 exponent, 3 mantissa). Currently supported on CPU backends only. |
 
-A narrower parameter type stores less per block, but it also has a much smaller range. `UE4M3`
+A narrower scale type stores less per block, but it also has a much smaller range. `UE4M3`
 cannot represent a value below `2^-9`, so a scale smaller than that rounds to zero and the block
 is lost. Scales stay in range when the quantized values are large enough, which in practice means
 it is not a drop-in replacement for `F32` on small-magnitude weights.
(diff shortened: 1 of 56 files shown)


crates/burn-backend/src/quantization/scheme.rs
```diff
@@ -26,23 +26,25 @@ pub fn compute_range<B: Backend>(
 
                 let num_blocks = numel / block_elems;
 
-                let params_shape = params_shape(&shape, scheme.level);
+                let params_shape = params_shape(&shape, scheme);
 
                 let blocks = B::float_reshape(tensor, Shape::new([num_blocks, block_elems]));
```
laggui: Blocks are currently treated as flat `block.num_elements()` chunks, which is only correct when each block is contiguous in row-major storage. Let's take a tensor with shape `[2, 4]`  for example:
```
[[a, b, c, d],
 [e, f, g, h]]
```

a `[2, 2]` block size means `[a, b, e, f]` and `[c, d, g, h]`, while flat chunking produces `[a, b, c, d]` and `[e, f, g, h]`.

This is a pre-existing issue, but highlighted by the block generalization.

The same pattern repeats in the `AbsMean` arm below.

Note that the same flat-chunk assumption is likely present in the Flex and NdArray quantize/dequantize implementations, so we'll probably have to update those as well. And add a regression test to exercise this. All existing multidimensional cases use blocks like `[2, 16]` on tensor `[16, 16]` (blocks span the full trailing dimension).

crates/burn-std/src/tensor/quantization.rs
```diff
@@ -197,15 +255,13 @@ impl QuantizedBytes {
...
-                "two-level quantization is not supported yet, got {:?}",
-                self.scheme.level
-            ),
-        };
-        let scale_bytes = scale_size(self.scheme.param) * num_params;
+        // Rounded up, so a trailing partial block still gets its scale.
+        let num_params = self
+            .scheme
+            .block_size()
+            .map_or(1, |block| self.num_elements.div_ceil(block.num_elements()));
```
laggui: The number of block scales must be derived per axis. For example, a `[3, 4]` tensor with `[2, 2]` has a `[2, 2]` block grid. So, 4 scales, not `ceil(12 / 4) = 3`. This requires retaining the tensor shape instead of only keeping the element count.

laggui (changes requested):
The overall direction looks good, but a few shape/layout assumptions need addressing

crates/burn-backend-tests/tests/tensor/float/quantization/ops/quantize.rs
```diff
@@ -2,22 +2,22 @@ use super::*;
...
-    QParams, QuantLevel, QuantScheme, QuantStore, QuantValue, QuantizationParameters,
-    QuantizedBytes,
+    DecodedScales, QuantScheme, QuantStore, QuantValue, QuantizationParameters, QuantizedBytes,
+    ScaleDtype,
 };
 use burn_tensor::{DType, Element, TensorData};
 
-fn get_q_params(data: TensorData) -> QParams<Vec<f32>> {
+fn get_q_params(data: TensorData) -> DecodedScales {
     let num_elements = data.num_elements();
```
laggui: Now unused

laggui: `num_elements` is no longer used

burn-book/src/performance/quantization.md
```diff
@@ -80,25 +79,41 @@ values, storage format, granularity, and how the values are scaled.
...
+| :--------------------------- | :---------------------------------------------------------------------------------------------- |
+| `per_tensor(dtype)`          | A single scale for the entire tensor.                                                           |
+| `per_block(block, dtype)`    | Tensor divided into blocks (1D, 2D, or higher) defined by `block`, each with its own scale.     |
 
-| Level                          | Description                                                                                                  |
-| :----------------------------- | :----------------------------------------------------------------------------------------------------------- |
-| `Tensor`                       | A single quantization parameter set for the entire tensor.                                                   |
-| `Block(block_size: BlockSize)` | Tensor divided into blocks (1D, 2D, or higher) defined by block_size, each with its own quantization params. |
+Setting both nests the blocks inside the tensor: the per-tensor scale is the factor the block
+scales are relative to, and it has to be `F32`.
```
laggui, line 121: I'd probably add a small note on `per_block`, right after the table. Something like:

Block dimensions are relative to the tensor rank and describe per-axis rectangles, not flat chunks of contiguous values. For example, `[16]` divides the trailing dimension into groups of 16, while `[2, 16]` divides the final two dimensions into rectangular blocks. Dynamic calibration currently requires every affected tensor dimension to be evenly divisible by its block extent.

burn-book/src/performance/quantization.md
laggui: Just noticed this pre-existing copy-pasta error, might as well fix it. It should not read as "packed into a 32-bit integer", but rather "packed into a natively supported representation"

burn-book/src/performance/quantization.md
```diff
@@ -124,16 +139,16 @@ let scheme = QuantScheme::default()
...
 | `F32`   | Full floating-point precision.                                                         |
 | `F16`   | Half-precision floating point.                                                         |
 | `BF16`  | Brain float 16-bit precision.                                                          |
 | `UE4M3` | 8-bit floating point (4 exponent, 3 mantissa). Currently supported on CPU backends only. |
 
-A narrower parameter type stores less per block, but it also has a much smaller range. `UE4M3`
+A narrower scale type stores less per block, but it also has a much smaller range. `UE4M3`
 cannot represent a value below `2^-9`, so a scale smaller than that rounds to zero and the block
 is lost. Scales stay in range when the quantized values are large enough, which in practice means
 it is not a drop-in replacement for `F32` on small-magnitude weights.
```
laggui: I think this is stale now, since this changed in #5261. It should be updated to reflect these changes and this PR. Maybe something like this?
```
A narrower scale type stores less per block, but it also has a much smaller range. The smallest positive `UE4M3` value is `2^-9`; Burn rounds a smaller positive scale up to `2^-9` rather than down to zero. The resulting scale can nevertheless be so coarse that the block's values quantize to zero. A two-level scheme lets the per-tensor `F32` scale carry the absolute magnitude, but the differences between block scales can still exceed `UE4M3`'s range and make smaller blocks quantize too coarsely.
```

laggui (approved):
Just some minor doc suggestions, otherwise LGTM.

Just need to update the cubecl/cubek revisions once https://github.com/tracel-ai/cubek/pull/498 is merged.

PR #5267, opened 2026-07-31
flex: improve add_bias perf a lot, and improve conv_plane_accumulate autovectorization

laundmo:
30-40% small convolution speedup!

### Checklist

- [ ] ~~Made sure the book is up to date with changes in this PR.~~

### Related Issues/PRs

None

### Changes

- Optimized bias_add: Added a Element-trait for addition, ElementAdd, which allows spezializing addition for some Elements without passing a function pointer through, like bias_add did previously. This change has allowed bias_add to complete disappear from my benchmark, relative to the convolutional layer.

- improved `conv_plane_accumulate` used in `conv3d_depthwise_impl` and `conv3d_small_channel_impl` by switching away from a iterator and to a range loop with unchecked indexing.

### Testing

- by running the model used in my project over it
- by running `cargo run-checks`, which showed one strange failure in f16::linalg::qr::test_qr_medium_tall even tho i can't think of anything i did affecting that.
- by running convolution benchmarks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5906,9 +5906,9 @@ dependencies = [
 
 [[package]]
 name = "macerator"
-version = "0.3.3"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "508a2f720538bb7e3ea3cb6d098615b107cb82ae387d77d906276905e9ec894b"
+checksum = "2ddeaadb76e307b1d2b4836adee77910d405b9f6c1ca031bc81a7a398e2b22f2"
 dependencies = [
  "bytemuck",
  "cfg_aliases",
@@ -5922,9 +5922,9 @@ dependencies = [
 
 [[package]]
 name = "macerator-macros"
-version = "0.1.5"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed5ce85961d618ce9794bdf822bfe96fe9dd341aa5b033b454f7a8d96e79b9b1"
+checksum = "d2a397d020af346358830d6f28579f26283003ce87af3dd79173ac9ee2d3ac28"
 dependencies = [
  "darling 0.20.11",
  "proc-macro2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,7 +153,7 @@ half = { version = "2.7.1", features = [
     "num-traits",
     "serde",
 ], default-features = false }
-macerator = { version = "0.3.3" }
+macerator = { version = "0.3.4" }
 matrixmultiply = { version = "0.3.10", default-features = false }
 ndarray = { version = "0.17.2", default-features = false }
 num-traits = { version = "0.2.19", default-features = false, features = [
--- a/crates/burn-backend/src/lib.rs
+++ b/crates/burn-backend/src/lib.rs
@@ -17,9 +17,9 @@ pub use backend::*;
 pub use burn_std::reader::*; // Useful so that backends don't have to add `burn_std` as a dependency.
 pub use burn_std::{
     AllocationProperty, BoolDType, BoolStore, Bytes, DType, DataError, DeviceHandle, Distribution,
-    DistributionSampler, DistributionSamplerKind, Element, ElementConversion, ElementEq,
-    ElementOrdered, ElementRandom, FloatDType, IntDType, Scalar, SplitPolicy, TensorData,
-    Tolerance, bf16, distribution, element, f16, stream::StreamId,
+    DistributionSampler, DistributionSamplerKind, Element, ElementAdd, ElementConversion,
+    ElementEq, ElementOrdered, ElementRandom, FloatDType, IntDType, Scalar, SplitPolicy,
+    TensorData, Tolerance, bf16, distribution, element, f16, stream::StreamId,
 };
 
 /// Shape definition.
--- a/crates/burn-flex/src/ops/conv.rs
+++ b/crates/burn-flex/src/ops/conv.rs
@@ -41,14 +41,14 @@ use super::conv_common::{add_bias, squeeze_3d_to_1d, squeeze_3d_to_2d};
 
 /// Generates a conv3d_1x1 function that uses the optimized gemm fast path.
 macro_rules! conv3d_1x1_typed {
-    ($fn_name:ident, $T:ty, $dtype:expr, $zero:expr, $one:expr, $add_fn:expr) => {
+    ($fn_name:ident, $T:ty, $dtype:expr, $zero:expr, $one:expr) => {
         fn $fn_name(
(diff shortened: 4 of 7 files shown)


laundmo:
i crossed out the book, but it may be worth mentioning the flex convolution fast-paths (small_channel, depthwise, 1x1, direct) and their requirements in the book. Tho thats not really that related to the pr.

laundmo:
I finally got around to running the benchmarks on both main and this PR to compare. Note: benchmarks depend on #5285

This PR, broadly everything got better, extreme gains highlighted
```diff
conv_ops                                 diff to main │     fastest    │ slowest       │ median        │ mean          │ samples │ iters
╰─ backend                                            │               │               │               │               │         │
   ╰─ conv2d_small_channel                            │               │               │               │               │         │
      ├─ downsample_1x4x128x128_k4x4_s2_to12   -4.44% │ 673.8 µs      │ 1.04 ms       │ 703.8 µs      │ 706.5 µs      │ 100     │ 100
      ├─ first_layer_4x3x224x224_k7x7_s2       +0.42% │ 21.08 ms      │ 22.21 ms      │ 21.27 ms      │ 21.39 ms      │ 100     │ 100
      ├─ mask_1x1x256x256_k3x3_to8            -32.46% │ 219.2 µs      │ 279.4 µs      │ 220.4 µs      │ 225.5 µs      │ 100     │ 100
      ├─ preproc_1x3x488x448_k5x5_to8         -40.54% │ 4.693 ms      │ 5.254 ms      │ 4.739 ms      │ 4.768 ms      │ 100     │ 100
      ├─ sobel_1x3x488x448_k1x3               -30.08% │ 255.3 µs      │ 429.9 µs      │ 272.7 µs      │ 279.8 µs      │ 100     │ 100
      ├─ sobel_1x3x488x448_k1x5               -36.77% │ 397.2 µs      │ 490.5 µs      │ 400.8 µs      │ 411.1 µs      │ 100     │ 100
      ├─ sobel_1x3x488x448_k3x1               -25.02% │ 295.1 µs      │ 362 µs        │ 304.8 µs      │ 310.7 µs      │ 100     │ 100
      ├─ sobel_1x3x488x448_k3x3               -39.90% │ 665.3 µs      │ 952.6 µs      │ 682.4 µs      │ 701.3 µs      │ 100     │ 100
      ╰─ sobel_1x3x488x448_k5x1               -24.26% │ 454.9 µs      │ 735.1 µs      │ 472.4 µs      │ 476.2 µs      │ 100     │ 100
```


<details><summary>Benchmark of main 8a933f84914fdd0bc4f2d716eff31c7b6a11b560 </summary>
<p>

```diff
conv_ops                                      fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ backend                                                  │               │               │               │         │
   ╰─ conv2d_small_channel                                  │               │               │               │         │
      ├─ downsample_1x4x128x128_k4x4_s2_to12  661.3 µs      │ 1.138 ms      │ 674.3 µs      │ 739.4 µs      │ 100     │ 100
      ├─ first_layer_4x3x224x224_k7x7_s2      20.57 ms      │ 22.42 ms      │ 21.22 ms      │ 21.3 ms       │ 100     │ 100
      ├─ mask_1x1x256x256_k3x3_to8            329.4 µs      │ 399.6 µs      │ 332.7 µs      │ 333.9 µs      │ 100     │ 100
-      ├─ preproc_1x3x488x448_k5x5_to8         7.934 ms      │ 8.875 ms      │ 7.976 ms      │ 8.02 ms       │ 100     │ 100
      ├─ sobel_1x3x488x448_k1x3               390.9 µs      │ 408.7 µs      │ 401.2 µs      │ 400.2 µs      │ 100     │ 100
      ├─ sobel_1x3x488x448_k1x5               645.7 µs      │ 654 µs        │ 650.1 µs      │ 650.2 µs      │ 100     │ 100
      ├─ sobel_1x3x488x448_k3x1               408.9 µs      │ 425.3 µs      │ 413.4 µs      │ 414.4 µs      │ 100     │ 100
      ├─ sobel_1x3x488x448_k3x3               1.115 ms      │ 1.848 ms      │ 1.138 ms      │ 1.167 ms      │ 100     │ 100
      ╰─ sobel_1x3x488x448_k5x1               625.4 µs      │ 634.6 µs      │ 629 µs        │ 628.8 µs      │ 100     │ 100
```

</p>
</details> 

In case anyone is curious about a full run of the benchmarks, here you go:
[pre-opt-conv.txt](https://github.com/user-attachments/files/30667449/pre-opt-conv.txt)
[post-opt-conv.txt](https://github.com/user-attachments/files/30667454/post-opt-conv.txt)

laundmo:
The lint causing CI failure here is annoying, because the generic does not in fact go unused: after the macerator macro expansion it is used in the macro.

laundmo:
@laggui could i get a re-run please?

laggui (changes requested):
The conv optimization LGTM!

Is `ElementAdd` necessary though? It only needs `add_bias` to avoid dispatch through a function pointer. Making `add_bias` over `F: Fn(T, T) -> T` should be sufficient, while keeping this implementation detail local to burn-flex instead of introducing a public trait for all element implementations. The `conv_plane_accumulate` work is fully independent and can remain unchanged.

Thanks 🙂

laundmo:
> Making add_bias over F: Fn(T, T) -> T should be sufficient

This would be relying on the compiler automatically making the right decision, while with ElementAdd i can ensure it does by annotating it `#[inline(always)]`

thats why i originally did it this way. Its ofc likely the compiler would inline the closure, but i don't think relying on that is nice

laggui:
Yeah I agree that the old `fn(T, T) -> T` function pointer was an optimization barrier. That's a great catch.

However, changing it to a generic `F: Fn(T, T) -> T` behaves differently than a function pointer. Rust monomorphizes generics at compile time, so a specialized copy of `add_bias` is generated at compile time for that exact closure. I believe it should yield the same as the `ElementAdd` trait method, without expanding the public trait just to achieve this.

You could always try it out to make sure

laundmo:
i know it monomorphizes, but i can't annotate a closure with #[inline(always)] so while its likely it'll decide to inline, i've left it completely up to the compiler whether it actually does it. Thats why i'm hesitant. Testing locally doesn't mean much, since it could just have decided to inline for me, but for a different project it might not.

Also, i'm adding a new trait, not expanding an existing one, so this should not have an effect on users using the Element trait.

laggui:
Yeah it will ultimately still be up to the compiler, even with `#[inline(always)]`. That said, I'd expect a monomorphized closure that simply does does `a + b` to inline reliably. It could also be a generic function instead of a closure.

Anyway, `ElementAdd` isn't a blocker. I just wanted to point out that there may be a more local way to achieve the same optimization.

laggui (approved):

PR #5269, opened 2026-07-31
fix(tensor-data): convert to bool stores by truthiness

SamuelBelanger:
Converting numeric `TensorData` into a boolean tensor used the numeric `elem()` conversion for the `U8` and `U32` bool stores. That conversion is range checked, so a negative source value panicked with `Element cannot be represented in the target type: "i32"(-1) => "u32"`, and float sources were truncated so `0.5` became `false`. `NaN` panicked as well.

The `Native` store already converted through `to_bool`, which adopts C semantics (anything non zero is true), so the same conversion behaved differently depending on the backend's bool storage: `Bool(U32)` on WebGPU, `Bool(U8)` on CUDA/Metal/Vulkan, `Bool(Native)` elsewhere. Identical user code panicked on one backend and worked on another.

This routes the bool store targets through `to_bool` too, for both the in place and clone conversion paths, so values are converted by truthiness and normalized to 0/1 (the canonical `true_val`/`false_val` form). Numeric targets are untouched: `i32 -> u32` still range checks and panics as documented.

Added `should_convert_negative_values_to_bool_store` covering int and float sources across the three stores.

### DIFF
--- a/crates/burn-std/src/data/tensor.rs
+++ b/crates/burn-std/src/data/tensor.rs
@@ -520,17 +520,32 @@ impl TensorData {
             DType::U32 => self.convert_inplace::<Current, u32>(),
             DType::U16 => self.convert_inplace::<Current, u16>(),
             DType::U8 => self.convert_inplace::<Current, u8>(),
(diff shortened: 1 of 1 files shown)


laggui (approved):
`test_qr_medium_tall` test failure is unrelated

Issue #5270, opened 2026-07-31
Quantization broken on cube backends since #5253: fusion panic on UE4M3 scale precision and flaky int8 quantize results


ThierryCantin-Demers:
## Summary

PR #5253 (UE4M3 scale params, validated end to end on CPU) introduced two kinds of quantization test failures on cube backends. CI never sees them because there is no GPU runner. Bisected to a one commit bracket: the parent commit `610991889` is clean over 6 runs, `5467d3fad` (#5253) shows both failure groups. `af6b2668f` (#5252) is not involved.

All measurements below: CUDA on an RTX 4070 Ti SUPER, current main (`ea1831b38`), 19 suite runs total across 4 revisions.

```
cargo test -p burn-backend-tests --features cuda --test tensor -- quantization
```

## Group 1: deterministic panic, unsupported fuse precision

`quantization::accuracy::error_responds_to_scale_param` and `quantization::accuracy::narrow_scale_param_degrades_without_normalization` fail on every run of every revision containing #5253. These are the new tests added by #5253 itself (`accuracy.rs`). They do not fail on accuracy: they panic before producing a result.

```
thread 'DSU-0-0' panicked at crates/burn-cubecl-fusion/src/engine/trace/block.rs:226:21:
not implemented: Unsupported fuse precision
```

The fusion trace does not know the UE4M3 scale precision, so this hits any `Fusion<CubeBackend>` device (cuda, vulkan, wgpu, metal), independent of the runtime. Possible resolutions:

1. Implement the new scale precision in the fusion trace (`block.rs:226`).
2. Make the fusion engine fall back to unfused execution for unsupported precisions instead of panicking. Arguably the right behavior regardless of this bug.
3. Until then, gate the two tests off cube backends the same way the NaN propagation tests are gated.

## Group 2: flaky symmetric int8 quantize results

Starting at the same commit, roughly one quantization test per suite run fails nondeterministically, a different one each time. Observed at #5253 (4 flakes in 6 runs, parent commit 0 in 6) and on current main (similar rate):

- `ops::quantize::should_support_quantize_symmetric_int8`
- `ops::quantize::should_quantize_symmetric_int8_permuted_batch_dims` (most frequent)
- `ops::quantize::should_quantize_symmetric_int8_transposed_8x32` and `_48x64`
- `ops::quantize::should_quantize_dequantize_symmetric_per_block_arange_16x16`
- `ops::quantize::should_quantize_dequantize_symmetric_single_with_transform`
[...]

Issue #1463, opened 2024-03-12
Does burn have a converter from PTH file


longzou:
Mostly, there are many models saved by pytorch. So, does the burn have the converter to help this?

laggui:
Yes! We have `PyTorchFileRecorder` (see [reference in the book](https://burn.dev/book/import/pytorch-model.html)).

Since PyTorch saves a (flattened) state dictionary without any model structure definition, you need to define the model in Burn to import the weights you want to reuse.

*Shameless plug:* if you're looking to port your model over to Burn, take a look at [this tutorial](https://dev.to/laggui/transitioning-from-pytorch-to-burn-45m) to port the `torchvision` ResNet weights to the Burn implementation.

PR #5279, opened 2026-07-31
fix(fusion): decline to fuse unsupported quant scale params

ThierryCantin-Demers:
Fixes the deterministic half of #5270.

## Problem

Any `Fusion<CubeBackend>` device panics when it reads a quantized tensor whose scales are `UE4M3` or `UE8M0`:

```
CallError(task panicked on device runner thread: not implemented: Unsupported fuse precision)
```

`QuantParam::UE4M3` became reachable in #5253, and the two accuracy tests it added hit this on every run on CUDA and Vulkan. The unfused path handles the scheme, so the panic is the fusion trace refusing an input rather than a missing capability elsewhere.

## Fix

`FuseType::from_quant_param` returns `Option<FuseType>`, and the two trace entry points that read quantization scales propagate `None`. That is how the builders already report an input they cannot take (see the `indexed` and `QFloat` early returns next to them), so the operation is simply not fused and falls back to unfused execution. Any future param degrades the same way instead of aborting.

`MatmulFuser` was `unwrap`ing that `Option`. It now closes both fusers instead. Mixed operands are unaffected: a non-quantized operand still yields `Some(MatmulArg::Normal(..))`, so float x float, float x quant, quant x float and quant x quant all still fuse. `None` means only "quantized with a scale param fusion cannot read".

## Verification

CUDA, RTX 4070 Ti SUPER, `cargo test -p burn-backend-tests --features cuda --test tensor -- quantization`:

- the `Unsupported fuse precision` panic is gone in 8 consecutive runs
- `quantization::accuracy::narrow_scale_param_degrades_without_normalization` passes

Vulkan, `-- quantization --skip accuracy`: 24 passed, 0 failed. `burn-fusion` (80) and `burn-cubecl-fusion` (9) lib tests pass.

## What this does not fix

The other two groups in #5270 are not burn bugs, and both need upstream work:

- The flaky symmetric int8 failures are a failed kernel *compilation* recording an error on a pooled cubecl hardware stream, which unrelated burn streams then inherit through `stream_id.value % max_streams`. Fixed by tracel-ai/cubecl#1463; with that applied the flake goes from 3 of 4 runs to 0 of 16.
- `quantization::accuracy::error_responds_to_scale_param` still fails, because the cube path rounds scales to nearest rather than up, so #5261's fix is still missing its GPU half. The reported `block16/ue4m3 = 0.01452` and `block32/ue4m3 = 0.01144` are exactly the pre-#5261 numbers.

This makes fusion decline `UE4M3` scales permanently, so those dequantizations always run unfused. `FuseType` has no e4m3 variant, and adding one plus its codegen is a feature rather than part of this fix.

### DIFF
--- a/crates/burn-cubecl-fusion/src/engine/codegen/ir.rs
+++ b/crates/burn-cubecl-fusion/src/engine/codegen/ir.rs
@@ -918,6 +918,30 @@ impl FuseType {
         }
     }
 
+    /// The type quantized values are read as, or `None` when fusion can't read that scheme.
+    ///
+    /// Same contract as [Self::from_quant_param]: callers must decline to fuse on `None`.
+    pub fn from_quant_scheme(scheme: QuantScheme) -> Option<Self> {
+        match scheme.store {
+            QuantStore::Native => match scheme.value {
+                QuantValue::Q8F | QuantValue::Q8S => Some(Self::I8),
+                QuantValue::E4M3 | QuantValue::E5M2 => None,
+                QuantValue::Q4F
+                | QuantValue::Q4S
+                | QuantValue::Q2F
+                | QuantValue::Q2S
+                | QuantValue::E2M1 => {
+                    panic!("Can't store native sub-byte values")
+                }
+            },
+            QuantStore::PackedU32(_) => Some(Self::U32),
+            QuantStore::PackedNative(_) => match scheme.value {
+                QuantValue::E2M1 => None,
+                other => panic!("{other:?} doesn't support native packing"),
+            },
+        }
+    }
+
     /// Converts the [fused element type](FuseType) into the [cubecl element type](ElemType).
     pub fn into_elem(self) -> ElemType {
         match self {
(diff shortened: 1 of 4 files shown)


laggui (approved):
Changes LGTM.

> * `quantization::accuracy::narrow_scale_param_degrades_without_normalization` passes

FYI it's failing with a compilation error on my end (NVIDIA RTX 4050 laptop).

Same for `error_responds_to_scale_param`. But I believe https://github.com/tracel-ai/cubecl/pull/1463 will fix those

ThierryCantin-Demers:
You're right, my PR description was not entirely accurate. At the cubecl rev this PR pins, `narrow_scale_param_degrades_without_normalization` fails 3/3 on a 4070 Ti SUPER with `identifier "__half2_2" is undefined` in `quantize_symmetric_packed_kernel_..._fs_e4m3_...`. Repointing only cubecl at [#1463](https://github.com/tracel-ai/cubecl/pull/1463): passes 5/5, and `-- quantization` goes to 25 passed / 1 failed. Both tests need [#1463](https://github.com/tracel-ai/cubecl/pull/1463).

PR #5281, opened 2026-08-01
Fix quantized transpose dequantization

### LINKED ISSUE #5278: Dequantizing a transposed quantized tensor reconstructs wrong values: block level on every backend, per-tensor on the cube backends
**Describe the bug**

Transposing a tensor *after* quantizing it, then dequantizing, reconstructs mostly wrong values. Roughly 85% of elements are wrong, by up to 20x, so this is not a precision issue.

Two separate defects, one shared and one specific to the cube backends:

| probe | NdArray | CUDA |
| --- | --- | --- |
| `QuantLevel::Tensor`, transpose after quantize | pass | **fail** |
| `QuantLevel::block([32])`, transpose after quantize | **fail** | **fail** |
| `QuantLevel::block([32])`, transpose twice, back to contiguous | pass | pass |

Block level is broken on both backends. Per-tensor is broken only on the cube backends, which rules out the block bookkeeping as the whole story: a per-tensor scheme has a single scalar scale with nothing to permute, so the values themselves are being read wrong.

The third row passing is the useful control. Transposing twice restores a contiguous layout and reconstructs correctly, so the scheme and scales survive the operation. What fails is dequantizing while the quantized tensor is still non-contiguous.

On CUDA the two-level case also trips the runtime validator, which suggests the read really does run off the end of the buffer rather than merely landing on the wrong element:

```
[VALIDATION dequantize_symmetric_packed_kernel_f_f32_nf_4_fs_f16_fg_f32_qs_u32_nq_1]:
Encountered OOB index in global<1> at 736, length is 256
```

**To Reproduce**

Append to `crates/burn-backend-tests/tests/tensor/float/quantization/ops/quantize.rs`:

```rust
#[test]
fn probe_per_tensor_transposed_after_quantize() {
    let tensor = TestTensorInt::arange(0..2048, &device)
        .float()
        .div_scalar(2048.)
        .reshape([32, 64]);
    let scheme = device
        .settings()
        .quantization
        .scheme
        .with_value(QuantValue::Q8S)
        .with_level(QuantLevel::Tensor);
    let output = tensor
        .clone()
        .quantize_dynamic(&scheme)
        .transpose()
        .dequantize()
        .transpose();
    tensor.into_data().assert_approx_eq::<FloatElem>(
        &output.into_data(),
        Tolerance::absolute(1e-1).set_relative(1e-2),
    );
}

#[test]
fn probe_single_level_block_transposed_after_quantize() {
    let tensor = TestTensorInt::arange(0..2048, &device)
        .float()
        .div_scalar(2048.)
        .reshape([32, 64]);
    let scheme = device
        .settings()
        .quantization
        .scheme
        .with_value(QuantValue::Q8S)
        .with_level(QuantLevel::block([32]));
    let output = tensor
        .clone()
        .quantize_dynamic(&scheme)
        .transpose()
        .dequantize()
        .transpose();
    tensor.into_data().assert_approx_eq::<FloatElem>(
        &output.into_data(),
        Tolerance::absolute(1e-1).set_relative(1e-2),
    );
}

#[test]
fn probe_block_transposed_then_contiguous() {
    let tensor = TestTensorInt::arange(0..2048, &device)
        .float()
        .div_scalar(2048.)
        .reshape([32, 64]);
    let scheme = device
        .settings()
        .quantization
        .scheme
        .with_value(QuantValue::Q8S)
        .with_level(QuantLevel::block([32]));
    // Same as above, but the transpose is undone before dequantizing.
    let output = tensor
        .clone()
        .quantize_dynamic(&scheme)
        .transpose()
        .transpose()
        .dequantize();
    tensor.into_data().assert_approx_eq::<FloatElem>(
        &output.into_data(),
        Tolerance::absolute(1e-1).set_relative(1e-2),
    );
}
```

```
cargo test -p burn-backend-tests --features cuda --test tensor probe_
cargo test -p burn-backend-tests --features ndarray --test tensor probe_
```

CUDA, block level:

```
=> Position 17: 0.008300781 != 0.11160879
   diff (rel = +9.26e-1, abs = +1.03e-1), tol (rel = +1.00e-2, abs = +1.00e-1)
=> Position 20: 0.009765625 != 0.14439669
=> Position 21: 0.010253906 != 0.16492373
1745 more errors...
```

CUDA, per-tensor level:

```
=> Position 256: 0.125 != 0
=> Position 257: 0.12548828 != 0
=> Position 259: 0.12646484 != 0.007870171
1711 more errors...
```

NdArray, block level:

```
=> Position 15: 0.0073242188 != 0.11984421
=> Position 16: 0.0078125 != 0.13778758
1633 more errors...
```

**Expected behavior**

Dequantizing reconstructs the original values regardless of whether the quantized tensor was transposed, within the quantization error.

**Why this was not caught**

The existing `should_quantize_transposed` helper transposes the *float* tensor before quantizing and again after dequantizing:

```rust
fn should_quantize_transposed<const D: usize>(tensor: Tensor<D>, scheme: QuantScheme) {
    let tensor_t = tensor.clone().transpose();
    let output = tensor_t.quantize_dynamic(&scheme).dequantize().transpose();
    ...
}
```

No quantized tensor is ever transposed, so `swap_dims` and `permute` on the quantized path have no coverage, at any level.

**Related but distinct**

- #4491 is closed and covers the inverse case, quantizing an already-transposed tensor. That path works; this is the other order.
- #4659 is reshape on the wgpu backend.

**Desktop**

CUDA on an NVIDIA GPU, and NdArray, both on Linux. Reproduced on `c1e987d6a` with a clean worktree.

subotac:
Fixes #5278.

  Ensure block quantization parameters remain aligned when quantized tensors are transposed. For CubeCL, move the packed axis to the position supported by the dequantization kernel and restore the original
  layout afterward.

  Tests:
  - NdArray quantized transpose regression tests
  - WGPU quantized transpose regression tests
  - cargo fmt --all --check
  - git diff --check


laggui (commented):
Changes LGTM, though it doesn't entirely fix the issue for cubecl backends when fusion is enabled.

The fused dequantization needs to handle a moved packed axis too. This should be handled here: https://github.com/tracel-ai/burn/blob/main/crates/burn-cubecl-fusion/src/engine/codegen/kernel.rs#L971

I think we also need to handle the `SwapDimsOpIr` and `PermuteOpIr` output packed axis.

subotac:
Thanks — you are right. Updated in 79162785: Fusion IR output dtypes now track permuted block and packed-axis metadata, and fused dequantization handles a moved packed axis. I also added regressions for moving the packed axis to the middle and first dimensions. Focused tests pass on NdArray (4/4) and WGPU+Fusion (4/4).

laggui (approved):
Thanks for addressing my comment! LGTM

I just moved the packed dim IR fixes so they're not just local to fusion, and made a small change to make sure the layout semantics are preserved.

PR #5283, opened 2026-08-03
feat(backend): promote mask_select to a backend op with dispatch

### LINKED ISSUE #5248: Promote mask_select to a backend tensor op (dispatch + overridable default)
### Feature description

Follow-up to #5238, as suggested in the review: promote `mask_select` from a pure API-level
composition to a first-class tensor op in the backend trait, with dispatch plumbing and a default
implementation, so individual backends can override it with a native or optimized version.

### Feature motivation

Today `mask_select` is implemented entirely at the API level (`argwhere` + `select` composition),
so no backend can specialize it even when a native implementation exists. `burn-tch` can override
it directly with the LibTorch `masked_select` binding, and other backends can eventually do the
same (e.g. a fused device-side implementation for cubecl). Promoting it to a trait op unlocks this
without changing the public API.

### Suggested Solution

- Add the op to the tensor ops trait(s) with a **default implementation that keeps the current
  behavior** (the `argwhere` + `select` composition), so all backends keep working unchanged.
- Wire the dispatch plumbing.
- Override the default in `burn-tch` with the native `masked_select` binding.
- The public API (`Tensor::mask_select` / `mask_select_async`) stays unchanged. The existing tests
  from #5238 (float/int/bool kinds, non-contiguous views, autodiff including the empty-mask
  backward) already pin the behavior and gradients, so they should catch any regression from the
  promotion.

Caveat for the future, probably worth its own issue: a device-side `bool_argwhere` for cubecl
(stream compaction, count-only readback) would reduce the synchronization cost for `argwhere`,
`nonzero`, and `mask_select` at once, and a fused `mask_select` kernel could build on top of it.

I'll take this on and follow up with a PR, keeping the same incremental approach as #5238.

Yashiru:
### Checklist

### Related Issues/PRs

Closes #5248. Follow-up to #5238 (the initial `mask_select` implementation), as requested in that
PR's review.

### Changes

`mask_select` was implemented in #5238 purely at the `burn-tensor` API level (composing `argwhere` +
`select`), so no backend could provide a native implementation. This PR promotes it to a backend op:

- `burn-backend`: add `float_mask_select`, `int_mask_select`, `bool_mask_select` to the ops traits,
  each with a default implementation that composes `argwhere` + `select` on the primitives (async,
  modeled on `bool_argwhere` since the output length is data-dependent). No parameter changes.
- `burn-dispatch`: add the matching wrappers so the op is routable per backend. `float` uses
  `binary_float!` (its `(float, any)` arm handles autodiff-wrapped tensors); `int`/`bool` use
  `binary_op!`.
- `burn-tensor`: add `mask_select` to the internal `BasicOps` bridge trait and route the public
  `mask_select_async` body through it. The public API signatures are unchanged. The `Float` bridge
  impl preserves the `QFloat` kind (composes `argwhere` + `q_select`), matching the previous behavior
  rather than dequantizing.
- `burn-autodiff`: `float_mask_select` is intentionally omitted so it inherits the default, which
  composes the tracked `reshape`/`select` ops (writing a pass-through here would silently break the
  gradient). This means a tracked float `mask_select` keeps composing `argwhere` + `select` even on
  a backend with a native op, which seems like the right trade: the backward pass needs the indices
  for the scatter-add anyway, so a dedicated `Backward` impl would have to recompute `argwhere`.
  `int`/`bool` get one-line pass-throughs so `Autodiff<Tch>` reaches the native op.
- `burn-tch`: override all three with the native `Tensor::masked_select`.
- `burn-tensor` (docs): reword the public `mask_select` rustdoc (`# Notes`) to describe the new backend dispatch instead of the now-stale `argwhere` fallback, and document the new backend-trait methods. No behavior change.
- `burn-backend-tests`: add 2 quantization tests covering the QFloat branch (the only new behavior the existing suite could not reach), and 2 autodiff tests covering the int/bool pass-throughs.

All other backends (ndarray, cubecl, candle, flex, fusion, router) inherit the default and need no
changes.

The public API and its behavior are unchanged, so the 14 tests added in #5238 pass without any
modification.

### Testing

- The 14 existing `mask_select` tests (float/int/bool + autodiff) pass unchanged on the default
  (ndarray) backend, which is the non-regression proof. They run as 16 cases, since the 2 autodiff
  tests are instantiated in both the `base` and `checkpointing` suites.
- Added 2 quantization tests (`should_support_mask_select_ops` and its empty-mask variant, in
  `tensor/float/quantization/ops/extended/mask.rs`, alongside the existing `mask_where`/`mask_fill`
  ones). They assert that the output is still `QFloat`, not just that the dequantized values match.
  This is the only new logic in the PR that the existing suite could not reach: replacing the
  `QFloat` branch with a dequantizing one leaves all 14 pre-existing tests green, and fails these 2.
- Added 2 autodiff tests (test_mask_select_int and test_mask_select_bool) exercising the new int/bool autodiff pass-throughs, which are only reachable on an autodiff device and had no coverage otherwise.
- The `burn-tch` override was validated locally (CI does not run `burn-backend-tests` on tch): all
  forward tensor tests pass on tch for float, int and bool, including the empty, transposed and
  narrowed cases, plus the non-empty autodiff gradient test.

**Note on one tch edge case**, for transparency: `test_mask_select_grad_empty_mask` fails on tch. This is
a pre-existing tch bug, in files this PR does not touch, and any `backward()` on a zero-sized root
hits it. The implicit sum's backward computes `float_mul(ones([0]), grad([1]))`;
`TchTensor::binary_ops_tensor` derives the output shape with an elementwise `max`, so it gets `[1]`
instead of `[0]`, concludes the rhs already has the output element count, and takes the in-place
`rhs.f_mul_(lhs)` fast path, which libtorch rejects with `output with shape [1] doesn't match the broadcast shape [0]`. It reproduces in plain torch with `torch.ones(1).mul_(torch.ones(0))`, and in burn with `tensor.select(0, empty_indices).sum()
.backward()` (no `mask_select` involved); the same case passes on ndarray. `mask_select` only
produces the empty tensor that feeds this existing weakness, and the #5238 code path behaves
identically. The fix looks like treating a `0` dim as `0` rather than `max` in `TchTensor::binary_ops_tensor`, or skipping the in-place fast paths when either operand is empty. Happy to file it separately.

**One caveat on tch coverage**: the `burn-backend-tests` harness hardcodes `IntElem = i32`, which
LibTorch rejects (it forces `I64`), so the tch runs above needed a local `i64` harness edit. That
edit is not part of this PR.

### Book

No book change: the public API and behavior are unchanged, so the existing `mask_select` entry stays
accurate.

### DIFF
--- a/crates/burn-autodiff/src/ops/bool_tensor.rs
+++ b/crates/burn-autodiff/src/ops/bool_tensor.rs
@@ -130,6 +130,13 @@ impl<B: Backend, C: CheckpointStrategy> BoolTensorOps<Self> for Autodiff<B, C> {
         B::bool_mask_fill(tensor, mask, value)
     }
 
+    async fn bool_mask_select(
+        tensor: BoolTensor<Self>,
+        mask: BoolTensor<Self>,
+    ) -> BoolTensor<Self> {
+        B::bool_mask_select(tensor, mask).await
+    }
+
     fn bool_gather(
         dim: usize,
         tensor: BoolTensor<Self>,
--- a/crates/burn-autodiff/src/ops/int_tensor.rs
+++ b/crates/burn-autodiff/src/ops/int_tensor.rs
@@ -257,6 +257,10 @@ impl<B: Backend, C: CheckpointStrategy> IntTensorOps<Self> for Autodiff<B, C> {
         B::int_mask_fill(tensor, mask, value)
     }
 
+    async fn int_mask_select(tensor: IntTensor<B>, mask: BoolTensor<B>) -> IntTensor<B> {
+        B::int_mask_select(tensor, mask).await
+    }
+
     fn int_argmax(tensor: IntTensor<B>, dim: usize) -> IntTensor<B> {
         B::int_argmax(tensor, dim)
     }
--- a/crates/burn-backend-tests/tests/autodiff/mask_select.rs
+++ b/crates/burn-backend-tests/tests/autodiff/mask_select.rs
@@ -48,3 +48,36 @@ fn test_mask_select_grad_empty_mask() {
         .into_data()
         .assert_eq(&TensorData::from([0.0, 0.0, 0.0]), false);
 }
+
+#[test]
+fn test_mask_select_int() {
+    // Int tensors are not differentiable; this covers the autodiff `int_mask_select`
+    // pass-through to the inner backend (only reachable on an autodiff device).
+    let device = AutodiffDevice::new();
+    let tensor = TestTensorInt::<2>::from_data(TensorData::from([[1, 2, 3], [4, 5, 6]]), &device);
+    let mask = TestTensorBool::<2>::from_data(
+        TensorData::from([[false, true, false], [true, false, true]]),
+        &device,
+    );
+
+    let output = tensor.mask_select(mask);
+
+    output
+        .into_data()
+        .assert_eq(&TensorData::from([2, 4, 6]), false);
+}
+
+#[test]
+fn test_mask_select_bool() {
+    // Bool tensors are not differentiable; this covers the autodiff `bool_mask_select`
+    // pass-through to the inner backend (only reachable on an autodiff device).
+    let device = AutodiffDevice::new();
+    let tensor = TestTensorBool::<1>::from_data(TensorData::from([true, false, true]), &device);
+    let mask = TestTensorBool::<1>::from_data(TensorData::from([true, true, false]), &device);
+
+    let output = tensor.mask_select(mask);
+
+    output
+        .into_data()
+        .assert_eq(&TensorData::from([true, false]), false);
+}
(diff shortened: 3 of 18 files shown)


laggui (approved):
> test_mask_select_grad_empty_mask fails on tch. This is a pre-existing tch bug, in files this PR does not touch, and any backward() on a zero-sized root hits it. 

It would be a good idea to file a separate issue for that.

I think we should also have an issue tracking perfomant cubecl implementations for argwhere / nonzero / mask_select.

Otherwise, the changes in this PR LGTM!

Yashiru:
Thanks for the review and merge! 

Filed #5287 for the tch zero-sized `backward()` bug, and I'll take #5286 for the native cubecl `argwhere` / `nonzero` / `mask_select` work if that's ok for you.

Issue #4814, opened 2026-04-16
NaN propagation in min/max/argmax/cummin/cummax diverges across backends


antimora:
## Summary

Backends disagree on how to handle `NaN` inputs to reductions. The flex backend propagates `NaN` (returns `NaN` when any input is `NaN`), matching PyTorch / NumPy / JAX / TensorFlow. The ndarray backend and the cubecl-based backends (cuda, rocm, wgpu, vulkan, webgpu, metal) follow IEEE 754 \`minimum\`/\`maximum\` semantics and silently drop NaN.

The public API docs for \`max_dim\`/\`min_dim\`/\`argmax\`/\`cummin\`/\`cummax\` don't say anything about NaN handling today, so users currently get silent behavior differences depending on which backend they pick.

## Reproducer

Tests added in #4812 surface the divergence:

- \`tensor::float::ops::maxmin::test_max_dim_nan_propagation\`
- \`tensor::float::ops::maxmin::test_min_dim_nan_propagation\`
- \`tensor::float::ops::maxmin::test_max_dim_with_indices_nan_propagation\`
- \`tensor::float::ops::arg::test_argmax_nan_propagation\`
- \`tensor::float::ops::cumulative::test_cummin_nan_propagation\`
- \`tensor::float::ops::cumulative::test_cummax_nan_propagation\`
- \`tensor::float::ops::cumulative::test_cummin_nan_at_start\`

They pass on flex, fail on ndarray and all cube-based backends. In that PR they've been gated to run only on backends that are known to propagate NaN, so CI stays green. Lifting the gate is this issue's job.

## Proposed resolution

1. Declare NaN-propagating semantics as the public-API contract (matches every ML framework users coming from PyTorch / NumPy / JAX / TF expect) and document it on the \`max_dim\`/\`min_dim\`/\`argmax\`/\`cummin\`/\`cummax\` doc blocks.
2. Update ndarray + cube-based backend implementations to propagate NaN. For scalar-reduction kernels this is a \`val.is_nan() || val > acc\` check; for cubecl kernels it's the same logic in the comparator.
3. Remove the \`#[cfg(not(any(feature = \"ndarray\", feature = \"cube\")))]\` gates around the tests.

wingertge:
This is not actually resolved on WGPU

laggui:
Just adding to this: we should probably still preserve some the current behavior, though the default operations would propagate NaN to standardize with other frameworks.

[...]

PR #5290, opened 2026-08-03
fix: standardize NaN propagation in extrema reductions

jcwal1516:
### Checklist

### Related Issues/PRs

Related to #4814.

Depends on tracel-ai/cubek#457.

### Changes

- Standardize floating-point `max`, `min`, `max_abs`, arg-extrema, and cumulative extrema NaN propagation across NdArray, Flex scalar/SIMD, and CubeCL/Cubek.
- Preserve the lowest-index result for equal arg-extrema values and for rows containing multiple or only NaNs.
- Use true floating-point extrema identities in the dependent Cubek reduction kernels while preserving the existing Burn integer paths.
- Fix whole Flex reductions over non-contiguous logical views so they do not read values outside the view.
- Retain the raw-storage Flex extrema fast path for dense permuted and flipped views only when `layout_covers_storage_once(...)` verifies that the layout visits every storage element exactly once. Partial, broadcast, overlapping, gapped, and offset views continue through `StridedIter`.
- Keep public API signatures and TopK unchanged. NaN-ignoring `nan_*` variants remain follow-up work for #4814.

The layout guard has differential coverage against `StridedIter` across contiguous, permuted, flipped, singleton, narrowed, expanded, overlapping, gapped, offset, and empty layouts.

No book change is required because this changes existing reduction behavior without adding or renaming a public API; the affected tensor method documentation is updated in source.

### Testing

- `cargo run-checks` — passed. This includes dependency audit, formatting, Clippy with warnings denied, typo checks, release builds, NdArray/Flex backend tests, documentation, and doctests.
- `cargo check -p burn-flex --no-default-features` — passed.
- `BURN_DEVICE=wgpu cargo test -p burn-backend-tests --no-default-features --features wgpu,std,fusion --test tensor nan -- --nocapture` — 24 passed.
- `BURN_DEVICE=wgpu cargo test -p burn-backend-tests --no-default-features --features wgpu,std,fusion --test tensor whole_max_min -- --nocapture` — 6 passed.
- `BURN_DEVICE=wgpu cargo test -p burn-backend-tests --no-default-features --features wgpu,std,fusion --test tensor arg_extrema_ties_use_lowest_index -- --nocapture` — 1 passed.

### DIFF
--- a/crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
+++ b/crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
@@ -1,4 +1,6 @@
 use super::*;
+#[cfg(feature = "cube")]
+use burn_fusion::inspect::FusionInspector;
 use burn_tensor::{ElementConversion, TensorData, Tolerance};
 
 #[test]
@@ -219,4 +221,49 @@ fn test_reduce_broadcasted_masked_softmax() {
         &TensorData::new(expected, [d0, q, k]),
         Tolerance::permissive(),
     );
+}
+
+#[test]
+fn test_reduce_broadcasted_max_nan() {
+    let stream = test_stream();
+    stream.executes(|| {
+        let device = Default::default();
+        let tensor = TestTensor::<2>::zeros([2, 4], &device);
+        let fused_on_read = TestTensor::<2>::from_data(
+            TensorData::from([[1.0, f32::NAN, 3.0, 2.0], [1.0, 4.0, 3.0, 2.0]]),
+            &device,
+        );
+        let fused_on_write = TestTensor::<2>::zeros([2, 1], &device);
+
+        // Materialize the inputs before recording the fused graph.
+        device.sync().unwrap();
+        #[cfg(feature = "cube")]
+        let inspector = FusionInspector::install(stream);
+
+        let output = tensor + fused_on_read.clone();
+        let output = output.max_dim(1);
+        let output = output + fused_on_write;
+        let output = output + fused_on_read;
+        let output = output + 1.0;
+        let actual = output.into_data();
+        device.sync().unwrap();
+
+        let expected = TensorData::from([
+            [f32::NAN, f32::NAN, f32::NAN, f32::NAN],
+            [6.0, 9.0, 8.0, 7.0],
+        ]);
+        actual.assert_approx_eq::<FloatElem>(&expected, Tolerance::default());
+
+        #[cfg(feature = "cube")]
+        {
+            let reports = inspector.drain();
+            assert!(
+                reports
+                    .iter()
+                    .flat_map(|report| report.fused_blocks())
+                    .any(|block| block.fuser_name() == Some("ReduceBroadcasted")),
+                "expected ReduceBroadcasted fusion, got {reports:#?}"
+            );
+        }
+    });
 }
--- a/crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
+++ b/crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
@@ -1,5 +1,4 @@
 use super::*;
-#[cfg(feature = "cube")]
 use burn_fusion::inspect::FusionInspector;
 use burn_tensor::{ElementConversion, TensorData, Tolerance};
 
@@ -237,7 +236,6 @@ fn test_reduce_broadcasted_max_nan() {
 
         // Materialize the inputs before recording the fused graph.
         device.sync().unwrap();
-        #[cfg(feature = "cube")]
         let inspector = FusionInspector::install(stream);
 
         let output = tensor + fused_on_read.clone();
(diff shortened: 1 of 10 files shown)


jcwal1516:
Ready for review after rebasing onto `146e98d0`.

Local validation passes: `cargo run-checks`, the focused WGPU NaN suite (24 tests), the whole max/min suite (6 tests), and the lowest-index tie regression.

The macOS job is expected to remain red on the two empty-mask autodiff tests. The current PR head reproduces the same zero-length `Sum` reduction failure locally, and the exact base commit fails those same tests with `518 passed; 2 failed; 6 ignored`: https://github.com/tracel-ai/burn/actions/runs/31037995539/job/92416791200. This PR does not change `mask_select` or sum reductions. The rebased PR CI is running here: https://github.com/tracel-ai/burn/actions/runs/31039367369.

crates/burn-flex/src/ops/reduce.rs
```diff
@@ -487,16 +519,10 @@ fn max_f32_reduce(tensor: &FlexTensor) -> FlexTensor {
...
-            } else {
-                StridedIter::new(tensor.layout())
-                    .map(|idx| data[idx])
-                    .reduce(|a, b| if a >= b { a } else { b })
-                    .expect("max: tensor must not be empty")
-            }
+            StridedIter::new(tensor.layout())
+                .map(|idx| data[idx])
+                .reduce(select_float_extremum::<f32, true>)
+                .expect("max: tensor must not be empty")
```
laggui: Is there a particular reason why the full-storage fast path for permuted views (i.e., non-contiguous views that still cover every element exactly once) was removed?

crates/burn-flex/src/ops/reduce.rs
```diff
@@ -528,15 +551,10 @@ fn min_f32_reduce(tensor: &FlexTensor) -> FlexTensor {
...
-            } else {
-                StridedIter::new(tensor.layout())
-                    .map(|idx| data[idx])
-                    .reduce(|a, b| if a <= b { a } else { b })
-                    .expect("min: tensor must not be empty")
-            }
+            StridedIter::new(tensor.layout())
+                .map(|idx| data[idx])
+                .reduce(select_float_extremum::<f32, false>)
+                .expect("min: tensor must not be empty")
```
laggui: Same

crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
```diff
@@ -1,4 +1,6 @@
 use super::*;
+#[cfg(feature = "cube")]
```
laggui: The tests under `fusion/` are only included for cubecl backends, so feature gating is redundant here.

crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
```diff
@@ -220,3 +222,48 @@ fn test_reduce_broadcasted_masked_softmax() {
...
+        let tensor = TestTensor::<2>::zeros([2, 4], &device);
+        let fused_on_read = TestTensor::<2>::from_data(
+            TensorData::from([[1.0, f32::NAN, 3.0, 2.0], [1.0, 4.0, 3.0, 2.0]]),
+            &device,
+        );
+        let fused_on_write = TestTensor::<2>::zeros([2, 1], &device);
+
+        // Materialize the inputs before recording the fused graph.
+        device.sync().unwrap();
+        #[cfg(feature = "cube")]
```
laggui: Same

crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
```diff
@@ -220,3 +222,48 @@ fn test_reduce_broadcasted_masked_softmax() {
...
+        let actual = output.into_data();
+        device.sync().unwrap();
+
+        let expected = TensorData::from([
+            [f32::NAN, f32::NAN, f32::NAN, f32::NAN],
+            [6.0, 9.0, 8.0, 7.0],
+        ]);
+        actual.assert_approx_eq::<FloatElem>(&expected, Tolerance::default());
+
+        #[cfg(feature = "cube")]
```
laggui: Same

laggui (commented):
Thanks! Just a couple of comments

crates/burn-flex/src/ops/reduce.rs
jcwal1516: Good catch, i removed it because data.len() == elem_count can also be true for expanded or overlapping views, where reading the full storage includes values outside the logical tensor. I was focused on fixing that and missed it also removed the fast path for real permutations.

I think the right fix is to keep the fast path, but guard it with something like layout_covers_storage_once(layout, data.len()) to verify that the layout uses every storage element exactly once. Would something like that make sense here or am I overthinking it?

crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
jcwal1516: Oh sorry I see what you mean. I was thrown off because tensor_f16 still includes this file when flex + fusion are enabled, and the test does run with cube disabled. But I see now that fusion isn’t actually applied to Flex, and xtask only runs this suite for CubeCL backends. I’ll remove the redundant gates.

crates/burn-flex/src/ops/reduce.rs
laggui: Yeah that makes sense!

jcwal1516: Implemented in `9cf7404e` for both max and min. Dense permuted and flipped views retain the raw-storage fast path only when `layout_covers_storage_once(...)` verifies that every storage element is visited exactly once. Other non-contiguous views still use `StridedIter`. I also added differential coverage for the relevant layout cases.

crates/burn-backend-tests/tests/fusion/reduce_broadcasted.rs
jcwal1516: Removed all three redundant `cube` gates in `9cf7404e`: the `FusionInspector` import, inspector installation, and report assertion are now unconditional within the fusion test module.

laggui (approved):

Issue #5297, opened 2026-08-04
Tensor::topk rejects k == dimension size; the documented example panics


4ktLuffy:
### Describe the bug

`Tensor::topk` rejects `k` equal to the size of the dimension:

```rust
pub fn topk<I: AsIndex>(self, k: usize, dim: I) -> Self {
    let dim = unwrap_dim_index(dim.try_dim_index(D), "Top K");
    assert!(self.shape()[dim] > k);
    ...
```

Asking for the top `n` of `n` elements is a normal request — it is a sort of that dimension — but it panics. Only `k > n` should be rejected, so the assertion looks like it should be `>=`.

The clearest evidence is that **the documented example for `topk` cannot run**. From `crates/burn-tensor/src/tensor/api/orderable.rs`:

```rust
let tensor = Tensor::<2>::from_data([[12.0, -2.0, 3.0], [5.0, 3.0, 6.0]], &device);
let tensor = tensor.topk(2, 0);
```

The tensor has 2 rows and the example asks for the top 2 along dim 0, so `shape[0] > k` is `2 > 2`, which is false:

```
thread '...' panicked at crates/burn-tensor/src/tensor/api/orderable.rs:235:9:
assertion failed: self.shape()[dim] > k
```

### To Reproduce

```rust
let t = TestTensor::<2>::from_data(
    TensorData::from([[12.0f32, -2.0, 3.0], [5.0, 3.0, 6.0]]), &Default::default());

t.clone().topk(1, 0);   // ok  -> [12.0, 3.0, 6.0]
t.topk(2, 0);           // panics, k == shape[0]
```

`cargo test -p burn-backend-tests --no-default-features --features ndarray --features std`

### Expected behavior

`k == shape[dim]` should return the whole dimension in sorted order. For reference, torch allows it and rejects only `k > n`:

```python
torch.topk(x, 2, dim=0).values   # [[12., 3., 6.], [5., -2., 3.]]
torch.topk(x, 3, dim=1).values   # [[12., 3., -2.], [6., 5., 3.]]
torch.topk(x, 4, dim=1)          # RuntimeError: selected index k out of range
```

### Notes

`topk_with_indices` at the same location has the same assertion, so presumably both.

I have not opened a PR because changing an assertion changes accepted inputs, and I would rather you confirm the intent first. If `>=` is the behaviour you want, I am happy to submit it with a test covering `k == n` and `k > n`.

[...]

PR #5298, opened 2026-08-04
Correct chained doc examples in orderable.rs

4ktLuffy:
## What

Three doc examples in `orderable.rs` chain two calls and show an expected output for the second that the code does not produce. In each case the second value was computed from the **original** tensor rather than from the result of the first call.

For input `[[12.0, -2.0, 3.0], [5.0, 3.0, 6.0]]`:

| example | shown | actual |
|---|---|---|
| `sort(0).sort(1)` | `[[-2.0, 3.0, 12.0], [3.0, 5.0, 6.0]]` | `[[-2.0, 3.0, 5.0], [3.0, 6.0, 12.0]]` |
| `sort_descending(0).sort_descending(1)` | `[[12.0, 3.0, -2.0], [6.0, 5.0, 3.0]]` | `[[12.0, 6.0, 3.0], [5.0, 3.0, -2.0]]` |
| `argsort_descending(0).argsort_descending(1)` | `[[0, 2, 1], [2, 0, 1]]` | `[[1, 2, 0], [0, 1, 2]]` |

The first call in each example was already correct; only the chained second value was wrong.

## Why it was not caught

These are `println!` outputs written in comments. `cargo test --doc` compiles and runs the examples, but nothing checks the printed values, so the comments can drift from behaviour without any test failing.

## Testing

Each corrected value was produced by running the example on the `ndarray` backend rather than worked out by hand:

```rust
let t = TestTensor::<2>::from_data(
    TensorData::from([[12.0f32, -2.0, 3.0], [5.0, 3.0, 6.0]]), &Default::default());
t.clone().sort(0).sort(1);                                 // [-2.0, 3.0, 5.0, 3.0, 6.0, 12.0]
t.clone().sort_descending(0).sort_descending(1);           // [12.0, 6.0, 3.0, 5.0, 3.0, -2.0]
t.argsort_descending(0).argsort_descending(1);             // [1, 2, 0, 0, 1, 2]
```

`cargo fmt --all -- --check` is clean. Documentation only, no behaviour change.

## Not included

The `topk` example in the same file is also a chained pair with a wrong second value, but it additionally panics before reaching that line — `topk(2, 0)` on a two-row tensor trips `assert!(self.shape()[dim] > k)`. That is a behaviour question rather than a documentation one, so I have raised it separately in #5297 and left the example untouched here.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/orderable.rs
+++ b/crates/burn-tensor/src/tensor/api/orderable.rs
@@ -36,7 +36,7 @@ where
     ///   // [[5.0, -2.0, 3.0], [12.0, 3.0, 6.0]]
     ///   let tensor = tensor.sort(1);
     ///   println!("{tensor}");
-    ///   // [[-2.0, 3.0, 12.0], [3.0, 5.0, 6.0]]
+    ///   // [[-2.0, 3.0, 5.0], [3.0, 6.0, 12.0]]
     /// }
     /// ```
     pub fn sort<I: AsIndex>(self, dim: I) -> Self {
@@ -70,7 +70,7 @@ where
     ///    // [[12.0, 3.0, 6.0], [5.0, -2.0, 3.0]]
     ///    let tensor = tensor.sort_descending(1);
     ///    println!("{tensor}");
-    ///    // [[12.0, 3.0, -2.0], [6.0, 5.0, 3.0]]
+    ///    // [[12.0, 6.0, 3.0], [5.0, 3.0, -2.0]]
     /// }
     /// ```
     pub fn sort_descending<I: AsIndex>(self, dim: I) -> Self {
@@ -194,7 +194,7 @@ where
     ///    // [[0, 1, 1], [1, 0, 0]]
     ///    let tensor = tensor.argsort_descending(1);
     ///    println!("{tensor}");
-    ///    // [[0, 2, 1], [2, 0, 1]]
+    ///    // [[1, 2, 0], [0, 1, 2]]
     /// }
     /// ```
     pub fn argsort_descending<I: AsIndex>(self, dim: I) -> Tensor<D, Int> {
--- a/crates/burn-tensor/src/tensor/api/orderable.rs
+++ b/crates/burn-tensor/src/tensor/api/orderable.rs
@@ -31,12 +31,12 @@ where
     /// fn example() {
     ///   let device = Default::default();
     ///   let tensor = Tensor::<2>::from_data([[12.0, -2.0, 3.0], [5.0, 3.0, 6.0]], &device);
-    ///   let tensor = tensor.sort(0);
-    ///   println!("{tensor}");
+    ///   let sorted = tensor.clone().sort(0);
+    ///   println!("{sorted}");
     ///   // [[5.0, -2.0, 3.0], [12.0, 3.0, 6.0]]
-    ///   let tensor = tensor.sort(1);
-    ///   println!("{tensor}");
-    ///   // [[-2.0, 3.0, 5.0], [3.0, 6.0, 12.0]]
+    ///   let sorted = tensor.sort(1);
+    ///   println!("{sorted}");
+    ///   // [[-2.0, 3.0, 12.0], [3.0, 5.0, 6.0]]
     /// }
     /// ```
     pub fn sort<I: AsIndex>(self, dim: I) -> Self {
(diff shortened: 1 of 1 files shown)


crates/burn-tensor/src/tensor/api/orderable.rs
```diff
@@ -36,7 +36,7 @@ where
     ///   // [[5.0, -2.0, 3.0], [12.0, 3.0, 6.0]]
     ///   let tensor = tensor.sort(1);
     ///   println!("{tensor}");
-    ///   // [[-2.0, 3.0, 12.0], [3.0, 5.0, 6.0]]
+    ///   // [[-2.0, 3.0, 5.0], [3.0, 6.0, 12.0]]
     /// }
```
laggui, line 40:
```rust
    /// fn example() {
    ///   let device = Default::default();
    ///   let tensor = Tensor::<2>::from_data([[12.0, -2.0, 3.0], [5.0, 3.0, 6.0]], &device);
    ///   let sorted = tensor.clone().sort(0);
    ///   println!("{sorted}");
    ///   // [[5.0, -2.0, 3.0], [12.0, 3.0, 6.0]]
    ///   let sorted = tensor.sort(1);
    ///   println!("{sorted}");
    ///   // [[-2.0, 3.0, 12.0], [3.0, 5.0, 6.0]]
    /// }
```

crates/burn-tensor/src/tensor/api/orderable.rs
```diff
@@ -70,7 +70,7 @@ where
     ///    // [[12.0, 3.0, 6.0], [5.0, -2.0, 3.0]]
     ///    let tensor = tensor.sort_descending(1);
     ///    println!("{tensor}");
-    ///    // [[12.0, 3.0, -2.0], [6.0, 5.0, 3.0]]
+    ///    // [[12.0, 6.0, 3.0], [5.0, 3.0, -2.0]]
     /// }
```
laggui, line 74:
```rust
   /// fn example() {
    ///    let device = Default::default();
    ///    let tensor = Tensor::<2>::from_data([[12.0, -2.0, 3.0], [5.0, 3.0, 6.0]], &device);
    ///    let sorted = tensor.clone().sort_descending(0);
    ///    println!("{sorted}");
    ///    // [[12.0, 3.0, 6.0], [5.0, -2.0, 3.0]]
    ///    let sorted = tensor.sort_descending(1);
    ///    println!("{sorted}");
    ///    // [[12.0, 3.0, -2.0], [6.0, 5.0, 3.0]]
    /// }
```

crates/burn-tensor/src/tensor/api/orderable.rs
```diff
@@ -194,7 +194,7 @@ where
     ///    // [[0, 1, 1], [1, 0, 0]]
     ///    let tensor = tensor.argsort_descending(1);
     ///    println!("{tensor}");
-    ///    // [[0, 2, 1], [2, 0, 1]]
+    ///    // [[1, 2, 0], [0, 1, 2]]
     /// }
```
laggui, line 198:
```rust
    /// fn example() {
    ///    let device = Default::default();
    ///    let tensor = Tensor::<2>::from_data([[12.0, -2.0, 3.0], [5.0, 3.0, 6.0]], &device);
    ///    let sorted = tensor.clone().argsort_descending(0);
    ///    println!("{sorted}");
    ///    // [[0, 1, 1], [1, 0, 0]]
    ///    let sorted = tensor.argsort_descending(1);
    ///    println!("{sorted}");
    ///    // [[0, 2, 1], [2, 0, 1]]
    /// }
```

laggui (changes requested):
Good catch 😅 

The original intent was likely to reuse the original tensor values, so I think we can adapt the example accordingly instead.

Just `.clone()` the original tensor for the first sorting operation, and do not shadow the tensor variable.

4ktLuffy:
Thanks — you are right, and that is the better fix. The values were correct for what the example was meant to show; the code had shadowed `tensor`, so the second call ran on the result of the first.

Reworked accordingly: the first call now uses `tensor.clone()` and the results bind to `sorted` / `indices` instead of shadowing, so the original documented output is restored and the example demonstrates each operation on the same input.

Confirmed on the ndarray backend that each matches what the docs already claimed:

    sort(1)               on the original -> [[-2.0, 3.0, 12.0], [3.0, 5.0, 6.0]]
    sort_descending(1)    on the original -> [[12.0, 3.0, -2.0], [6.0, 5.0, 3.0]]
    argsort_descending(1) on the original -> [[0, 2, 1], [2, 0, 1]]

`cargo test --doc -p burn-tensor` passes (175), fmt clean.

The `topk` example in the same file has the same shadowing, but it also panics before reaching the second call (`topk(2, 0)` on a two-row tensor trips `assert!(shape[dim] > k)`), so I have left it for #5297.

laggui (approved):
Thanks!

PR #5302, opened 2026-08-04
test(burn-tensor): run the doc examples instead of only compiling them

4ktLuffy:
The doc examples in `burn-tensor` are wrapped in `fn example() { ... }`, which nothing ever
calls. Rustdoc wraps a doctest body in `fn main() { ... }`, so `fn example` is a nested item
that is compiled and type-checked but **never executed**.

175 examples define it. None call it, and there are no hidden `# example();` lines. The tell
is the timing — `cargo test --doc -p burn-tensor` reports every one as passing in **0.00s**,
and `--nocapture` prints nothing.

This PR renames the wrapper to `fn main` in the 173 examples that use it, so rustdoc uses the
provided `main` and the bodies actually run.

### What that immediately caught

Running them for the first time produced exactly one failure:

```
---- ...api/orderable.rs - tensor::api::orderable::Tensor<D,K>::topk (line 219) stdout ----
thread 'main' panicked at crates/burn-tensor/src/tensor/api/orderable.rs:235:9:
assertion failed: self.shape()[dim] > k
```

The documented `topk` example calls `topk(2, 0)` on a tensor whose dim-0 size is 2, so
`k == shape[dim]` and the assertion rejects it. That is #5297, still open — and the reason it
had gone unnoticed is exactly this: the example never ran.

Everything else passes: **174 passed, 1 failed** before the `topk` fix below, and **175 passed,
0 failed** after it.

### The `topk` example

It had two problems: the panic above, and the same chaining defect fixed in #5298 and #5301 —
`tensor` was shadowed by the first result, so the second call ran on that rather than the
original, while the documented output was the correct answer for the original.

I changed it to `topk(1, 0)` so the example passes against today's behaviour, and de-shadowed
it in the style you asked for on #5298:

```rust
let topk = tensor.clone().topk(1, 0);
println!("{topk}");
// [[12.0, 3.0, 6.0]]
let topk = tensor.topk(1, 1);
println!("{topk}");
// [[12.0], [6.0]]
```

Both values verified by execution rather than by inspection:

```
TOPK_1_DIM0 [12.0, 3.0, 6.0]
TOPK_1_DIM1 [12.0, 6.0]
```

**This is the one judgement call in the PR, and I would rather you made it.** If you take the
`>` → `>=` change proposed in #5297, then `topk(2, 0)` becomes legal and the original example
is fine as written once de-shadowed — it is arguably the better illustration, since it shows
`k > 1`. Say the word and I will restore `k = 2` instead. I picked `k = 1` only because it is
correct under current behaviour and does not pre-empt that decision.

### Scope

Rename only — no documented value changes anywhere, and no library code is touched. 18 files,
+178/−178, of which 173 are the wrapper rename and the rest are the `topk` example.

Verified: `cargo fmt -p burn-tensor --check` clean, `cargo test --doc -p burn-tensor` green
(175 passed, 0 failed, 19 ignored). Run time moves from 0.00s to 0.06s, which is the examples
actually executing.

### DIFF
--- a/crates/burn-tensor/src/tensor/api/base.rs
+++ b/crates/burn-tensor/src/tensor/api/base.rs
@@ -38,38 +38,36 @@ use serde::{Serialize, Serializer};
 /// use burn_tensor::Tensor;
 /// use burn_tensor::Int;
 ///
-/// fn example() {
-///     let device = Default::default();
+/// let device = Default::default();
 ///
-///     let tensor = Tensor::<2>::from_data(
-///         [
-///             [3.0, 4.9, 2.0],
-///             [2.0, 1.9, 3.0],
-///             [6.0, 1.5, 7.0],
-///             [3.0, 4.9, 9.0],
-///         ],
-///         &device,
-///     );
+/// let tensor = Tensor::<2>::from_data(
+///     [
+///         [3.0, 4.9, 2.0],
+///         [2.0, 1.9, 3.0],
+///         [6.0, 1.5, 7.0],
+///         [3.0, 4.9, 9.0],
+///     ],
+///     &device,
+/// );
 ///
-///     // Slice the tensor to get the second and third rows:
-///     // [[2.0, 1.9, 3.0], [6.0, 1.5, 7.0]]
-///     // The resulting tensor will have dimensions [2, 3].
-///     let slice = tensor.clone().slice([1..3]);
-///     println!("{slice}");
+/// // Slice the tensor to get the second and third rows:
+/// // [[2.0, 1.9, 3.0], [6.0, 1.5, 7.0]]
+/// // The resulting tensor will have dimensions [2, 3].
+/// let slice = tensor.clone().slice([1..3]);
+/// println!("{slice}");
 ///
-///     // Slice the tensor to get the first two rows and the first 2 columns:
-///     // [[3.0, 4.9], [2.0, 1.9]]
-///     // The resulting tensor will have dimensions [2, 2].
-///     let slice = tensor.clone().slice([0..2, 0..2]);
-///     println!("{slice}");
+/// // Slice the tensor to get the first two rows and the first 2 columns:
+/// // [[3.0, 4.9], [2.0, 1.9]]
+/// // The resulting tensor will have dimensions [2, 2].
+/// let slice = tensor.clone().slice([0..2, 0..2]);
+/// println!("{slice}");
 ///
-///     // Index the tensor along the dimension 1 to get the elements 0 and 2:
-///     // [[3.0, 2.0], [2.0, 3.0], [6.0, 7.0], [3.0, 9.0]]
-///     // The resulting tensor will have dimensions [4, 2]
-///     let indices = Tensor::<1, Int>::from_data([0, 2], &device);
-///     let indexed = tensor.select(1, indices);
-///     println!("{indexed}");
-/// }
+/// // Index the tensor along the dimension 1 to get the elements 0 and 2:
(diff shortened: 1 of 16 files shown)


laggui (changes requested):
We need to force the device used in the doc examples. This can be done by setting the `BURN_DEVICE` env variable, just like this:

https://github.com/tracel-ai/burn/blob/c2ecf30647ba9187b8e6ad54316895ed986c5234/xtask/src/commands/test.rs#L67-L72

but for the `cargo xtask doc` command. It should be set to use flex (lightweight, portable - so should always work on the CI runner).

4ktLuffy:
Thanks — both fixed, and the clippy failure turned out to say my original approach was wrong.

**1. `BURN_DEVICE` for the doc command.** Done, following the pattern you linked: `xtask doc`
now sets `BURN_DEVICE=flex` before running, with a comment explaining why (the examples
execute, so they need a device, and flex is portable enough to always be available on the
runners).

**2. Dropped the `fn main` wrapper entirely instead of renaming to it.** `code-quality` was
failing on:

```
error: needless `fn main` in doctest
note: `-D clippy::needless-doctest-main` implied by `-D warnings`
```

which is a fair objection to what I'd done. Removing the wrapper altogether is the better fix
anyway — rustdoc supplies the `main`, the body runs, the lint is satisfied, and the rendered
examples are less noisy. So the diff is now "unwrap the example body" rather than "rename the
wrapper", which is also why it is a net reduction in lines.

**3. Restored `topk(2, 0)`** now that #5307 has landed, as promised in #5297. The example is
de-shadowed (`tensor.clone()` for the first call) so the second call operates on the original,
which is what the documented values describe.

**One thing worth flagging while I was in there:** two doc blocks in `burn-tensor` have
unterminated ```` ``` ```` fences — `numeric.rs` (`bool`) and `orderable.rs` (`min_pair`).
Rustdoc implicitly closes them at the end of the doc comment so they still compile, and I have
left them alone to keep this diff focused, but they are easy to trip over. Happy to fix them
here or in a separate PR, whichever you prefer.

### Verification

- `cargo fmt` clean; `cargo clippy -p burn-tensor --no-deps -- --deny warnings` clean
- `BURN_DEVICE=flex cargo test --doc -p burn-tensor` -> **175 passed, 0 failed**
- To confirm the examples really do execute now rather than merely compile, I injected a
  `panic!` into one example and re-ran: it failed, printing the panic. Removing it returns to
  175 passing. Before this change the same suite reported everything as passing in 0.00s
  without running any of it.

laggui:
> **One thing worth flagging while I was in there:** two doc blocks in `burn-tensor` have unterminated `` ``` `` fences — `numeric.rs` (`bool`) and `orderable.rs` (`min_pair`). Rustdoc implicitly closes them at the end of the doc comment so they still compile, and I have left them alone to keep this diff focused, but they are easy to trip over. Happy to fix them here or in a separate PR, whichever you prefer.

If you could fix it in this PR that'd be great! Since it's documentation related, might as well :slightly_smiling_face:

4ktLuffy:
Done — both fences closed in `ae248ae`.

`Tensor::bool` and `Tensor::min_pair` now terminate their blocks properly, which means both examples execute as doctests for the first time rather than merely compiling. Their documented outputs were already correct, so no values changed:

- `bool` on `[[1.0, -2.0, 3.0], [0.0, 9.0, 6.0]]` → `[[true, true, true], [false, true, true]]`
- `min_pair` → `[[1.0, -2.0, 3.0], [1.0, 2.0, 3.0]]`

`BURN_DEVICE=flex cargo test --doc -p burn-tensor` → 175 passed, 0 failed. `cargo fmt --check` and `cargo clippy --deny warnings` are clean.

laggui (approved):
Thank you!

PR #5305, opened 2026-08-05
Fix panic broadcasting a zero-sized dimension on the tch backend (#5287)

### LINKED ISSUE #5287: burn-tch: backward() on a zero-sized tensor panics in the in-place binary-op fast path
### Describe the bug

Any `backward()` whose autodiff graph contains a zero-sized (empty) root panics on the `burn-tch`
(LibTorch) backend. During the backward pass an elementwise binary op runs between an empty `[0]`
tensor and a `[1]` gradient, and LibTorch rejects it:

```
Torch("output with shape [1] doesn't match the broadcast shape [0]")
```

The same computation passes on `ndarray`.

### Root cause

`TchTensor::binary_ops_tensor` (`crates/burn-tch/src/tensor.rs`) derives the output shape
dimension-wise with `usize::max`:

```rust
out_shape[i] = usize::max(lhs_shape[i], rhs_shape[i]); // tensor.rs:280
```

For a `[0]` vs `[1]` broadcast the correct result is `[0]`, but `max(0, 1)` gives `[1]`. With
`num_elements_out == 1`, the `rhs` (the `[1]` gradient) is treated as already matching the output
element count, so the in-place fast path is taken:

```rust
|lhs, rhs| rhs.f_mul_(lhs).unwrap(), // crates/burn-tch/src/ops/base.rs:472
```

LibTorch then computes the real broadcast (`[0]`) and refuses to write it into the `[1]` buffer.

### To Reproduce

Minimal, in plain torch:

```python
torch.ones(1).mul_(torch.ones(0))  # same error
```

In burn (no `mask_select` involved), on `Autodiff<LibTorch>`:

```rust
tensor.select(0, empty_indices).sum().backward()
```

It surfaces via `test_mask_select_grad_empty_mask` (added in #5238, promoted in #5283): a `mask_select`
on an all-`false` mask produces the empty `[0]` tensor that feeds this path. `mask_select` is not the
cause; it only creates the empty tensor.

### Expected behavior

`backward()` on a zero-sized root should produce an all-zero gradient (as it does on `ndarray`),
not panic.

### Proposed fix

In `TchTensor::binary_ops_tensor`, either:

- derive a `0` output dim as `0` rather than `usize::max` when either operand's dim is `0`, or
- skip the in-place fast paths when either operand has zero elements.

### Additional context

Found while promoting `mask_select` to a backend op in #5283; filed separately as suggested in that
PR's review.

I'd be happy to take this on and open a fix PR along the lines suggested above.

hdimer:
Fixes #5287.

## Problem

On the `burn-tch` (LibTorch) backend, `TchTensor::binary_ops_tensor` derives the broadcast output shape dimension-wise with `usize::max`:

```rust
out_shape[i] = usize::max(lhs_shape[i], rhs_shape[i]);
```

For a `[0]` vs `[1]` broadcast the correct result is `[0]`, but `max(0, 1) == 1`. That overstates `num_elements_out`, so the in-place fast path runs against the 1-element operand, and LibTorch aborts when it computes the real `[0]` broadcast:

```
Torch("output with shape [1] doesn't match the broadcast shape [0]")
```

Any `backward()` whose autodiff graph contains a zero-sized (empty) tensor hits this (the same computation passes on `ndarray`). Minimal repro in plain torch: `torch.ones(1).mul_(torch.ones(0))`.

## Fix

A zero-sized dimension broadcasts to `0`, not `max`:

```rust
out_shape[i] = if lhs_shape[i] == 0 || rhs_shape[i] == 0 {
    0
} else {
    usize::max(lhs_shape[i], rhs_shape[i])
};
```

With the correct shape, `num_elements_out` is `0` for the empty case, so the fast path runs on the empty operand (`[0].mul_([1])` writing into a `[0]` buffer, which LibTorch accepts). Non-empty broadcasts are unchanged (the `else` is the original `max`).

## Test

Added `mul_broadcasts_zero_sized_dim_without_panic`, covering both operand orders (each exercises a different in-place fast path) and a multi-dim `[2,0]` case. Confirmed it aborts with the exact error above on the pre-fix code and passes after.

A note on the existing autodiff test `test_mask_select_grad_empty_mask` (added alongside #5238/#5283) that first surfaced this: it doesn't run to the broadcast on a tch/CPU build whose LibTorch lacks `I32` support (it fails earlier at the int-dtype backend config), so this regression test targets `binary_ops_tensor` directly with float tensors instead.

`cargo test -p burn-tch`, `cargo fmt`, and `cargo clippy -p burn-tch` are clean.

---

Used AI assistance on this; I reviewed, built, and tested it.

### DIFF
--- a/crates/burn-tch/src/tensor.rs
+++ b/crates/burn-tch/src/tensor.rs
@@ -277,7 +277,14 @@ impl TchTensor {
         let mut out_shape = Shape::from(vec![1usize; d_out]);
 
         for i in 0..d_out {
-            out_shape[i] = usize::max(lhs_shape[i], rhs_shape[i]);
+            // A zero-sized dim broadcasts to zero, not `max`: `[0]` vs `[1]` is
+            // `[0]`. `max` overstated the length and took the in-place fast path
+            // below, panicking in LibTorch on the shape mismatch (#5287).
+            out_shape[i] = if lhs_shape[i] == 0 || rhs_shape[i] == 0 {
+                0
+            } else {
+                usize::max(lhs_shape[i], rhs_shape[i])
+            };
         }
 
         let num_elements_out = out_shape.num_elements();
@@ -483,6 +490,24 @@ mod tests {
             &device,
             DType::QFloat(QuantScheme::default())
         ));
+    }
+
+    #[test]
+    fn mul_broadcasts_zero_sized_dim_without_panic() {
+        // Regression for #5287: broadcasting against a zero-sized dim yields 0.
+        let device = Default::default();
+        let one: TchTensor = B::float_from_data(TensorData::from([1.0]), &device);
+        let empty: TchTensor = B::float_from_data(TensorData::new(Vec::<f32>::new(), [0]), &device);
+
+        // Both operand orders (each exercises a different in-place fast path).
+        assert_eq!(B::float_mul(one.clone(), empty.clone()).shape().dims(), [0]);
+        assert_eq!(B::float_mul(empty, one).shape().dims(), [0]);
+
+        // Multi-dim: only the zero-sized dimension collapses.
+        let m: TchTensor = B::float_from_data(TensorData::new(vec![1.0f32, 2.0], [2, 1]), &device);
+        let m_empty: TchTensor =
+            B::float_from_data(TensorData::new(Vec::<f32>::new(), [2, 0]), &device);
+        assert_eq!(B::float_mul(m, m_empty).shape().dims(), [2, 0]);
     }
 
     #[test]
--- a/crates/burn-tch/src/tensor.rs
+++ b/crates/burn-tch/src/tensor.rs
@@ -287,17 +287,21 @@ impl TchTensor {
             };
         }
 
-        let num_elements_out = out_shape.num_elements();
+        // Gate the in-place fast path on shape equality, not element count: at a
+        // zero-sized output every operand with a zero dim has 0 elements and would
+        // wrongly match, taking the in-place path for a broadcast it can't do
+        // (e.g. `[1, 0] * [2, 0]`). For non-empty operands `numel == out numel`
+        // already implies equal shapes, so routing is otherwise unchanged (#5287).
 
         // Attempt to mutate lhs tensor
-        if lhs_shape.num_elements() == num_elements_out
+        if lhs_shape == out_shape
             && let Some(output) = lhs.mut_ops(|lhs| flmut(lhs, &rhs.tensor))
         {
             return output;
         }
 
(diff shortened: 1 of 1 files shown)


Yashiru:
Thanks for picking this up! I filed this issue and had a fix in progress, so I took a close look at the branch.

It fixes the reported `[0]` vs `[1]` case, but I don't think it fully closes #5287. There's a related empty-tensor case still open. The in-place guards a few lines below match on element count rather than shape:

```rust
if lhs_shape.num_elements() == num_elements_out && ...
if rhs_shape.num_elements() == num_elements_out && ...
```

At `num_elements_out == 0` that check is degenerate: every operand with a zero-sized dim has 0 elements and matches, including one that still needs broadcasting. So a valid broadcast where an empty operand has a different shape than the output still takes the in-place path and panics. I reproduced it on this branch:

```rust
// [1, 0] * [2, 0]  ->  output [2, 0]
B::float_mul(empty_1_0, empty_2_0);
// panics: Torch("output with shape [1, 0] doesn't match the broadcast shape [2, 0]")
```

It's reachable from the public API, since `binary_ops_ew_shape` allows a `1`-vs-`N` dim, so a plain `a * b` hits it. The multi-dim test added here uses `[2, 1] * [2, 0]`, where the empty operand already has the output shape, so it doesn't exercise this path.

One option that covers both cases is to gate the in-place path on shape equality instead of element count:

```rust
if lhs_shape == out_shape && let Some(output) = lhs.mut_ops(...) { ... }
if rhs_shape == out_shape && let Some(output) = rhs.mut_ops(...) { ... }
```

For non-empty broadcasts, `numel == out_numel` already implies shape equality, so this keeps the existing routing unchanged and only adds the empty case.

I already have this locally (shape-equality guards plus a test covering both operand orders and `[1, 0] * [2, 0]`). Happy to push it as a follow-up, or you're welcome to fold it in, whichever the maintainers prefer.

hdimer:
Thanks @Yashiru, great catch. You're right that the element-count guards are degenerate at a zero-sized output. I've folded in the shape-equality guards and added regression coverage for `[1, 0] * [2, 0]` in both operand orders (the rhs case needs a shared lhs to actually reach that path). Reverting the guards to the old `num_elements` check reproduces the exact panic you described, and the shape check clears it. Ready for another look.

laggui (approved):
The isolated fix in tch LGTM.

Though I think zero-sized tensors support should generally be inspected across backends. Pretty sure they also affect cubecl backends and fusion (see the macos CI failure for `test_mask_select_grad_empty_mask`). We should probably open a separate and broader issue for that.

Issue #5286, opened 2026-08-03
[burn-cubecl] missing `argwhere`, `nonzero` and `mask_select` implementations


laggui:
The current `argwhere`, `nonzero` and `mask_select` (default) implementation read the data back to the host to compute the output.

Related:
#5283

Yashiru:
Hey @laggui,
I'd like to take this one if that's good for you.

Plan: native cubecl argwhere / nonzero via a prefix sum (scan) over the boolean mask to compute the output offsets, then a compaction pass to scatter the indices. mask_select can then compose on the native argwhere, or get its own kernel. I'll start on argwhere and open a draft PR once I have something running.

laggui:
Sounds good!

PR #5309, opened 2026-08-06
feat(burn-cubecl): native argwhere/nonzero/mask_select via stream compaction

Yashiru:
### Checklist

- [x] Made sure the book is up to date with changes in this PR. (No book change: this is a backend-internal op override with no user-facing API change.)

### Related Issues/PRs

Part of #5286. Follows the discussion on #5283, where `mask_select` was promoted to a backend op with a host-read default and a native cubecl implementation was left as a follow-up.

This is a **draft**: it lands the native implementations but deliberately leaves the parallel-scan optimization for a follow-up (see the note at the bottom).

### Changes

Today the cubecl backend inherits the host-read default for `bool_argwhere`: it copies the whole mask back to the host and computes the coordinates on the CPU. This PR implements `argwhere` natively on the device using stream compaction, so the mask never leaves the GPU (except for one scalar).

A single override, `bool_argwhere` in `burn-cubecl`, is the only new op. The other two operations improve for free because they compose on it:
- `nonzero` builds on `argwhere` at the tensor layer.
- `mask_select` uses the backend default that composes `argwhere` with `select`.

**Pipeline** (all on device unless noted):
1. `into_contiguous` so a flat position maps to a reliable coordinate.
2. `bool_cast` to a 0/1 int mask, flattened to `[n]`.
3. `cumsum` (inclusive prefix sum) so each element learns its output row.
4. Read back the last prefix-sum entry, the total count. This is a single scalar and the only host synchronization. It is data dependent, which is why the op is asynchronous.
5. Allocate the `[count, ndims]` output (`[0, ndims]` when the count is zero).
6. A new compaction kernel writes each `true` element's coordinate to its row.

The count read is the only sync. It is irreducible because the output size depends on the data, but it moves a single scalar instead of the entire mask, and it fits the existing `async` signature of the trait method.

### Testing

Validated on a real NVIDIA GPU (RTX 4090, CUDA 12.8.1, driver 580.159.04) with `--features cuda`, and cross-checked on the cubecl CPU backend.

`cargo test -p burn-backend-tests --features cuda --test tensor -- argwhere nonzero mask_select`:

```
running 20 tests
test tensor::bool::ops::argwhere_nonzero::test_argwhere_empty ... ok
test tensor::bool::ops::argwhere_nonzero::test_argwhere_3d ... ok
test tensor::bool::ops::argwhere_nonzero::test_argwhere_2d ... ok
test tensor::bool::ops::argwhere_nonzero::test_argwhere_1d ... ok
test tensor::bool::ops::argwhere_nonzero::test_nonzero_1d ... ok
test tensor::bool::ops::argwhere_nonzero::test_nonzero_2d ... ok
test tensor::bool::ops::argwhere_nonzero::test_nonzero_empty ... ok
test tensor::bool::ops::argwhere_nonzero::test_nonzero_3d ... ok
test tensor::bool::ops::mask_select::test_mask_select_bool ... ok
test tensor::float::ops::mask_select::test_mask_select_2d ... ok
test tensor::float::ops::mask_select::test_mask_select_1d ... ok
test tensor::float::ops::mask_select::test_mask_select_shape_mismatch - should panic ... ok
test tensor::float::ops::mask_select::test_mask_select_both_flipped ... ok
test tensor::float::ops::mask_select::test_mask_select_3d ... ok
test tensor::float::ops::mask_select::test_mask_select_narrowed_tensor ... ok
test tensor::float::ops::mask_select::test_mask_select_empty ... ok
test tensor::float::ops::mask_select::test_mask_select_transposed_tensor ... ok
test tensor::float::ops::mask_select::test_mask_select_single_element ... ok
test tensor::int::ops::mask_select::test_mask_select_int_1d ... ok
test tensor::int::ops::mask_select::test_mask_select_int_2d ... ok

test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 1499 filtered out; finished in 1.06s
```

The forward path passes on every shape and dtype, including the empty cases.

**Note on a pre-existing failure.** The autodiff suite has two red tests on the cuda backend, `test_mask_select_grad_empty_mask` (base and checkpointing). They fail with `plane_vectorized_parallel_reduce: input reduce axis length (currently 0) should be at least k (1)`, which comes from the autotune of a `Sum` reduction over a zero length axis in the backward pass. This is unrelated to `argwhere`: it lives in the reduce path, not the index computation. I confirmed it pre-exists on the same GPU by reverting this patch to the pristine base commit and rerunning the two tests, which fail identically. So this PR neither introduces nor fixes them.

### Follow-up (not in this PR)

The prefix sum reuses the existing `numeric::cumsum`, which is a naive O(n^2) scan (see the `TODO: efficient parallel scan` in `numeric.rs`). Replacing it with a work-efficient parallel scan (plane/warp-level) is the performance half of #5286 and is being worked on as a separate change. It is deliberately excluded here so the correctness base can land and be reviewed on its own. Before and after numbers will come with that follow-up, since benchmarking against the current naive scan would not be representative of the intended performance.

### DIFF
--- a/crates/burn-cubecl/src/kernel/index/argwhere.rs
+++ b/crates/burn-cubecl/src/kernel/index/argwhere.rs
@@ -1,0 +1,119 @@
+use crate::{
+    CubeRuntime,
+    kernel::{
+        bool_cast, into_contiguous, slice,
+        utils::{address_type, decompose_linear, shape_divmod},
+    },
+    ops::{
+        base::{into_data, reshape},
+        numeric::{cumsum, empty_device_contiguous_dtype},
+    },
+    tensor::CubeTensor,
+};
+use burn_backend::cubecl::dtype_to_storage_type;
+use burn_backend::{DType, IntDType, Shape, TensorMetadata};
+use cubecl::{CubeDim, calculate_cube_count_elemwise};
+use cubecl::{prelude::*, std::FastDivmod, std::tensor::layout::linear::LinearView};
+
+/// Scatter the coordinate of every `true` element to its output row.
+///
+/// One thread per input element. `offsets` is the inclusive prefix sum of the 0/1 mask, so for a
+/// `true` element the running count `offsets[pos]` gives its 1-based row in the output; `false`
+/// elements write nothing. `row * ndims + d` indexes the contiguous `[count, ndims]` output.
+#[cube(launch_unchecked, address_type = "dynamic")]
+fn argwhere_kernel<I: Int>(
+    mask: LinearView<'_, I>,
+    offsets: LinearView<'_, I>,
+    output: &mut Tensor<I>,
+    shape: Sequence<FastDivmod<usize>>,
+    working_units: usize,
+    #[comptime] ndims: usize,
+    #[define(I)] _dtype: StorageType,
+) {
+    if ABSOLUTE_POS >= working_units {
+        terminate!();
+    }
+
+    if usize::cast_from(mask.read(ABSOLUTE_POS)) != 0 {
+        let row = usize::cast_from(offsets.read(ABSOLUTE_POS)) - 1;
+        let (_, coords) = decompose_linear(ABSOLUTE_POS, &shape);
+        #[unroll]
+        for d in 0..ndims {
+            output[row * ndims + d] = I::cast_from(*coords.index(d));
+        }
+    }
+}
+
+/// Compute the coordinates of the `true` elements, one row per element (row-major order).
+///
+/// GPU stream compaction: cast the mask to 0/1, prefix-sum it to get each element's output row,
+/// read the total count back to the host to size the `[count, ndims]` output, then scatter each
+/// `true` element's coordinate into its row. The count read is the only host synchronization; it
+/// is data-dependent, which is why the op is asynchronous.
+pub(crate) async fn argwhere<R: CubeRuntime>(
+    tensor: CubeTensor<R>,
+    out_dtype: IntDType,
+) -> CubeTensor<R> {
+    let client = tensor.client.clone();
+    let device = tensor.device.clone();
+    let dtype: DType = out_dtype.into();
+
+    // Flat position must map to a coordinate, so the mask has to be contiguous.
+    let tensor = into_contiguous(tensor);
+    let shape = tensor.shape();
+    let ndims = shape.num_dims();
+    let n = shape.num_elements();
+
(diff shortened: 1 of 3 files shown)


Yashiru:
@laggui this is the draft I mentioned in #5286: native `argwhere` / `nonzero` /
`mask_select` via stream compaction.

Current state: correctness-complete and GPU-validated (RTX 4090, 20/20 on the
tensor suite). The only failing tests are the pre-existing
`test_mask_select_grad_empty_mask` (reduce over a zero-length axis). I reproduced
them on the unpatched base and am fixing that separately in #5313.

The scan still uses the naive `numeric::cumsum`. The work-efficient plane-level
scan is a deliberate follow-up and is not part of this draft.

Before I build that optimized scan, does the overall structure look like the
direction you want? bool cast to 0/1, inclusive cumsum for the write offsets, a
single count readback, then a scatter kernel. Happy to adjust the layout now
rather than after the perf work.

crates/burn-cubecl/src/kernel/index/argwhere.rs
```diff
@@ -0,0 +1,119 @@
...
+    let ndims = shape.num_dims();
+    let n = shape.num_elements();
+
+    // 1. bool -> 0/1 int mask, keeping the original (contiguous) shape for coordinate decomposition.
+    let mask = bool_cast(tensor, dtype);
+    let coord_shape = shape_divmod(&mask);
+
+    // 2. inclusive prefix sum over the flattened mask -> per-element output rows.
+    let mask_flat = reshape(mask.clone(), Shape::new([n]));
+    let offsets = cumsum(mask_flat.clone(), 0);
```
laggui, line 73: The count and offsets should use a sufficiently wide accumulation dtype independently of the output coordinate dtype. A full true 16x16 mask with U8 output would currently return [0, 2] instead of [256, 2], even though every coordinate fits. Could we separate these dtypes and add a regression test?

laggui (commented):
Sorry I missed the notifications with the recent uptick in contributions. I usually don't prioritize draft PRs to focus on what we could potentially merge now 😅 you could say my personal algorithm is a bit biased lol.

The overall structure looks like the right direction. The scope is reasonable, and reusing it through `argwhere` for `nonzero` and `mask_select` makes sense.

I think the efficient scan should land before enabling the override unconditionally though, since the current scan implementation can substantially slow down larger inputs.

PR #5311, opened 2026-08-06
feat(module): generalize weight reparameterization abstraction

### LINKED ISSUE #2738: Support weight reparametrization
### Feature description

It would be nice if burn supported some way to apply weight reparametrization to layers like found in pytorch (e.g. [weight_norm](https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.weight_norm.html#torch.nn.utils.parametrizations.weight_norm), [spectral_norm](https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.spectral_norm.html#torch.nn.utils.parametrizations.spectral_norm)).

### Feature motivation

pytorch docs provide a decent explanation and relevant links but in short, this can stabilize training of certain networks.

In the current design this kind of api probably belongs on [`Param`](https://docs.rs/burn/latest/burn/module/struct.Param.html) since dynamically replacing fields, like pytorch does, is not really feasible. `Param` would have to switch out implementations based on a particular reparametrization.

laggui:
### Checklist

### Related Issues/PRs

Closes #2738

### Changes

- Generalized parameter reparameterization beyond LoRA (#5139) with the public `Reparameterization` and `Reparameterizer` traits. Stored as arbitrary reparameterization state in `Param` through a private type-erased bridge.
- Added `Module::apply_reparameterization` while preserving the `apply_lora` and `apply_qlora` convenience
  APIs.
- Collapsed `LoraConfig` / `LoraMapper` into a single `Lora` struct which implements the new trait

Reparameterization techniques do not need to implement their own `ModuleMapper`, as LoRA previously did. Traversal is handled centrally by the internal `ApplyReparameterization` mapper which wraps the `Reparameterizer`.

Example:
https://github.com/tracel-ai/burn/blob/3bbe22e6117db10c3b9c319ad00cec58e413776a/crates/burn-core/src/module/param/lora.rs#L24-L31
https://github.com/tracel-ai/burn/blob/3bbe22e6117db10c3b9c319ad00cec58e413776a/crates/burn-core/src/module/lora.rs#L43-L100

### Testing

Existing tests + new integration test using a `WeightNorm` implementation as reference


crates/burn-core/src/module/lora.rs
```diff
@@ -39,30 +33,23 @@ impl LoraConfig {
...
-///
-/// Each rank-2 float weight is frozen and given a trainable low-rank [adapter](LoraAdapter); other
-/// parameters are left untouched. No model or layer code needs to change — the same `Linear` (and
-/// any other module) keeps working, now producing `base + scale * (a @ b)` for adapted weights.
+/// All existing floating-point parameters are frozen. Matching rank-2 receive trainable LoRA
+/// [adapter](LoraAdapter)s; other parameters remain frozen without adapters. No model or layer
+/// code needs to change — the same `Linear` (and any other module) keeps working, now
+/// producing `base + scale * (a @ b)` for adapted weights.
 #[derive(Debug, Clone)]
 pub struct LoraMapper {
```
laggui: I didn't rename the types, but maybe we could since they are no longer directly tied to mappers (an implementation detail)?

Charles23R: Yeah, I tend to agree. Moreover, LoraMapper just becomes a wrapper over the LoraConfig. I don't think two different structures are needed here, but the naming becomes a bit awkward. Maybe just `LoraParameterization` ?

Charles23R (approved):

crates/burn-core/src/module/lora.rs
laggui: Ha you're right :sweat_smile: I'll merge that into a single struct. Might as well rename them to `Lora` and `QLora` quite simply.

PR #5313, opened 2026-08-06
fix(burn-cubecl): fill the reduction identity over a zero-length axis

### LINKED ISSUE #5310: Zero-sized (empty) tensor support is inconsistent across backends
**Describe the bug**

Operations on zero-sized (empty) tensors are not handled uniformly across backends. Some backends panic on inputs that are valid according to the public API. This issue tracks the broader problem so the individual cases can be found and fixed together, rather than one report at a time. It follows from #5287 (a tch panic when broadcasting a zero-sized dimension, fixed by #5305) and from a failure observed while working on #5286. As suggested in the #5305 review, zero-sized tensor support deserves a dedicated pass across backends.

The main open case is on the cubecl backends and fusion. `test_mask_select_grad_empty_mask` (base and checkpointing) fails because the backward of `mask_select` on an empty mask reduces (`Sum`) over a zero-length axis, and the reduce autotune has no candidate that accepts length 0:

```
plane_vectorized_parallel_reduce: An unknown error happened.
The input reduce axis length (currently 0) should be at least k (1).
```

Both autotune candidates reject it (`sum_one_shot` and `plane_vectorized_parallel_reduce`), so autotune panics in `cubecl-runtime/src/tune/local.rs` with `Should run when selected by autotune`. Under fusion this surfaces downstream as `Should have handle for tensor` at `burn-fusion/src/client.rs:200`.

This is not backend-local. It shows up on the macos CI (metal/wgpu, noted by @laggui in #5305), and I reproduced it on the cuda backend. The forward empty cases work (for example `test_mask_select_empty` passes), so the problem is specific to reductions over a zero-length axis, which the backward pass of several ops can produce.

**To Reproduce**

On the cubecl cuda backend (also fails on macos CI with metal/wgpu):

```
cargo test -p burn-backend-tests --features cuda --test autodiff -- test_mask_select_grad_empty_mask
```

Both `base` and `checkpointing` variants fail with the reduce-axis-length-0 error above. This is pre-existing and independent of any single op: reverting to a pristine `main` and rerunning the two tests reproduces the exact same failure, so the reduce path over a zero-length axis is at fault, not the caller.

For reference, the already-fixed tch case (#5287, fixed by #5305): `binary_ops_tensor` derived the output shape with `usize::max`, so `max(0, 1)` gave `1` instead of `0`, and the in-place fast path was taken for an operand that still needed broadcasting, which LibTorch rejects.

**Expected behavior**

Operations on valid empty tensors should not panic. For the cubecl reduce case, a reduction over a zero-length axis should return the reduction identity (0 for sum, 1 for product, and so on) rather than dispatching to an autotune that has no valid candidate.

**Environment**

 - burn version: `main` (0.22.0-pre.1)
 - Backends affected: cubecl (cuda confirmed, metal/wgpu via macos CI), fusion
 - Repro setup: RTX 4090, CUDA 12.8.1, driver 580.159.04, Linux; and the macos CI runner (metal/wgpu)

**Additional context**

Likely root cause for cubecl: reduce kernels and their autotune assume a reduce axis length of at least 1. A zero-length axis has no viable candidate, so autotune panics instead of returning the reduction identity. The backward of ops that reduce over a masked or selected dimension can hit this whenever the selection is empty.

Suggested direction:
- Audit zero-sized tensor handling across backends and ops, starting with reductions on cubecl (return the reduction identity for a zero-length axis instead of dispatching to autotune).
- Add empty-tensor coverage to the shared backend test suite for the ops most likely to be affected (reduce, broadcast, backward of index and mask ops).
- Track the individual backend fixes under this issue.

Related: #5287, #5305, #5286, #5309.

I am happy to take the cubecl reduce case if that is useful, since I already have the reproduction from the #5286 work.

Yashiru:
### Checklist

### Related Issues/PRs

Closes #5310. Related to #5305, which fixed the same class of panic on the
tch backend.

### Changes

Reducing a tensor over an axis of length 0 panicked on the cubecl backends:
the reduce kernels require an axis length of at least 1. The trigger in
practice is the `mask_select` backward pass on an all-false mask, which
reduces over a zero-length axis:

```
input reduce axis length (currently 0) should be at least k (1)
```

Since there is nothing to fold, the result is the reduction identity (the
neutral element of the operation). Fill it directly instead of dispatching to
a kernel that rejects length 0.

- `crates/burn-cubecl/src/kernel/reduce/empty.rs` (new): `reduce_empty_axis`
  maps each operation to its identity and launches a small fill kernel
  (`fill_reduce_identity_kernel`) that writes it into every output position.
- `crates/burn-cubecl/src/kernel/reduce/base.rs`: a zero-length guard at each
  entry point that can reach a reduce kernel. `sum()` returns `zeros([1])`,
  `reduce_dim()` calls `reduce_empty_axis`, and `reduce_dim_with_indices()`
  returns a `ReduceError`.

**Identity per operation:**

| identity | operations | reason |
| --- | --- | --- |
| 0 | sum, maxabs, any | adding nothing |
| 1 | prod, all | multiplying nothing |
| -inf | max | loses to any value |
| +inf | min | beats any value |
| NaN | mean | sum 0 / count 0 |
| error | argmax, argmin, topk, argtopk | no index exists |

**Design decisions (confirmation welcome)**:

I made the following choices deliberately and would like your sign-off before
they land. They are laid out transparently so you can redirect any of them.

- Max / min identity and NumPy divergence. The mathematical identity of `max`
  over an empty set is `-inf` (and `+inf` for `min`), which is what this PR
  fills, so every reduction that has a neutral element returns it. NumPy instead
  raises for `max` / `min` over an empty axis ("zero-size array to reduction
  operation which has no identity"), while it returns 0 / 1 for sum / prod and
  NaN for mean. Filling the extreme keeps the behavior uniform across
  operations. This is the main decision I would like confirmed: I am happy to
  switch max / min to an error instead if you prefer NumPy parity.
- Index reductions error. `argmax` / `argmin` / `topk` / `argtopk` return an
  index; over an empty axis there is no position to point at, so they return
  `ReduceError::Validation` rather than fabricating an index. This one seems
  clearly correct, but I am flagging it too in case you would rather surface it
  differently.

Extrema built from bits is an implementation detail rather than a behavioral
choice: floats have no portable infinity literal across cubecl targets, so
`+/-inf` and `NaN` are built from their IEEE-754 bits via `f32::reinterpret`,
matching how cubek forms its own `max_identity` / `min_identity`. Integers use
`E::max_value()` / `E::min_value()`, with the identity kind selected on the host
and baked into the kernel at comptime.

I will adjust any of these to your preference before this is ready to merge.

### Testing

Five value tests were added in
`crates/burn-backend-tests/tests/cubecl/reduce.rs`, covering `sum` / `prod` /
`max` / `min` / `mean` over a `[2, 0]` axis (mean asserted as NaN, which cannot
be compared by equality). The existing reduce suite and the `mask_select`
gradient tests pass unchanged.

- CPU (cubecl-cpu): 5/5 empty-axis tests, full reduce suite with no regression,
  `test_mask_select_grad_empty_mask` green.
- GPU (RTX 4090, CUDA 12.8.1): `test_mask_select_grad_empty_mask` 2 passed;
  cubecl reduce suite 22 passed, 0 failed (including the 5 empty-axis tests).
  The extreme fill kernel (`+/-inf` via `f32::reinterpret`) produces correct
  values on a real GPU.
- `cargo run-checks` passes (format, clippy, build, unit tests, doc tests,
  integration).

### DIFF
--- a/crates/burn-backend-tests/tests/cubecl/reduce.rs
+++ b/crates/burn-backend-tests/tests/cubecl/reduce.rs
@@ -350,6 +350,65 @@ fn reduction_sum_should_match_reference_backend() {
         .assert_approx_eq::<FloatElem>(&tensor_ref.clone().sum().into_data(), Tolerance::default());
 }
 
+// Reducing over a zero-length axis folds no elements, so the result is the reduction identity.
+// The reduce kernels reject a zero-length axis, so these exercise the host-side identity fill.
+
+#[test]
+fn reduction_sum_dim_empty_axis() {
+    let device = Default::default();
+    let tensor = TestTensor::<2>::zeros([2, 0], &device);
+    let expected = TestTensor::<2>::from_data([[0.0], [0.0]], &device);
+    tensor
+        .sum_dim(1)
+        .into_data()
+        .assert_eq(&expected.into_data(), false);
+}
+
+#[test]
+fn reduction_prod_dim_empty_axis() {
+    let device = Default::default();
+    let tensor = TestTensor::<2>::zeros([2, 0], &device);
+    let expected = TestTensor::<2>::from_data([[1.0], [1.0]], &device);
+    tensor
+        .prod_dim(1)
+        .into_data()
+        .assert_eq(&expected.into_data(), false);
+}
+
+#[test]
+fn reduction_max_dim_empty_axis() {
+    let device = Default::default();
+    let tensor = TestTensor::<2>::zeros([2, 0], &device);
+    let expected = TestTensor::<2>::from_data([[f32::NEG_INFINITY], [f32::NEG_INFINITY]], &device);
+    tensor
+        .max_dim(1)
+        .into_data()
+        .assert_eq(&expected.into_data(), false);
+}
+
+#[test]
+fn reduction_min_dim_empty_axis() {
+    let device = Default::default();
+    let tensor = TestTensor::<2>::zeros([2, 0], &device);
+    let expected = TestTensor::<2>::from_data([[f32::INFINITY], [f32::INFINITY]], &device);
+    tensor
+        .min_dim(1)
+        .into_data()
+        .assert_eq(&expected.into_data(), false);
+}
+
+#[test]
+fn reduction_mean_dim_empty_axis() {
+    // Mean over an empty axis is 0 / 0 = NaN, which cannot be compared with equality.
+    let device = Default::default();
+    let tensor = TestTensor::<2>::zeros([2, 0], &device);
+    let data = tensor.mean_dim(1).into_data();
+    assert!(
+        data.iter::<f32>().all(|x| x.is_nan()),
+        "mean over an empty axis should be NaN, got {data:?}"
+    );
+}
+
 #[test]
 #[ignore = "Impossible to run unless you have tons of VRAM. Also reference backend is broken."]
 fn reduction_sum_should_match_reference_backend_64bit() {
--- a/crates/burn-cubecl/src/kernel/reduce/base.rs
+++ b/crates/burn-cubecl/src/kernel/reduce/base.rs
@@ -63,6 +63,13 @@ pub fn sum<Run: CubeRuntime>(
     let client = tensor.client.clone();
     let device = tensor.device.clone();
 
+    // A sum over an empty tensor folds no elements, so the result is the additive identity 0.
+    // The reduce kernels reject a zero-length axis (they require at least one element), so
+    // return the identity directly instead of dispatching to them.
+    if tensor.shape().num_elements() == 0 {
+        return Ok(zeros_client(client, device, [1].into(), tensor.dtype));
+    }
+
     match strategy {
         SumStrategy::OneShot(cube_count) => {
             let output = zeros_client(client.clone(), device, [1].into(), tensor.dtype);
@@ -225,6 +232,12 @@ pub fn reduce_dim<Run: CubeRuntime>(
         },
     )?;
 
+    // Reducing over a zero-length axis folds no elements, so every output position is the
+    // reduction identity. The kernels reject a zero-length axis, so fill the identity directly.
+    if input.shape()[dim] == 0 {
+        return super::empty::reduce_empty_axis(output, config);
+    }
+
     let result = match strategy {
         KernelReduceStrategy::Unspecified => cubek::reduce::reduce::<Run>(
             &client,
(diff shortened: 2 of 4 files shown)


Yashiru:
@laggui apologies for the noise here. The failing checks on this run do notseem related to the change. It looks like the run hit the GitHub Actions infrastructure incident on Aug 6: most jobs failed at the setup step with "Failed to resolve action download info. Error: Service Unavailable" and the rest were canceled in the cascade (actions/runs/31119864747). code-quality (format and clippy) and prepare-checks both passed, and the no-std job was running the tests green before it got canceled, so nothing actually failed to build or test.

I do not have admin rights to re-run the jobs, and I would rather not push an empty or amended commit just to re-trigger CI so the PR stays at a single clean commit. 

Whenever you have a moment, could a maintainer please re-run the failed jobs? 
Thank you very much, and sorry for the trouble.

Yashiru:
@laggui @robertomeroni

after comparing the two on the cubecl side, I think #5333 is a better approach than this PR.

This PR fills the identity with a small device kernel. That was only justified while max and min filled `-inf` and `+inf`, since `InputScalar` goes through `ToPrimitive` and cannot carry those extremes safely. Once max and min error instead (the right call for the integer case, where `i32::MIN` or `i32::MAX` is a real value the tensor could hold), the only values left to write are 0, 1 and NaN, which `InputScalar` handles fine. 
That is exactly what #5333 does by reusing full_device in a few host-side lines, so I think my custom kernel no longer earns its complexity.

Happy to close this in favor of #5333 if you confirm @laggui.

Separately, while validating I hit something neither PR covers:
with the reduce panic fixed, `mask_select` backward on an all-false mask still fails on the cubecl-cpu backend inside `broadcast_shape`, which does not handle a zero-sized dim (assert in debug, SIGSEGV in release). I filed #5334 for it with a repro and am already working on the fix.

robertomeroni:
@Yashiru 

thanks for this :)

worth saying yours was open first and closes #5310, so I hope it didn't read as me stepping over
you. If @laggui prefers your structure I'm happy to rebase mine onto it and keep only the parts you
don't cover

I see you've opened #5335 for the broadcast bug in the meantime, left a comment there

laggui (commented):
Thanks for tackling this! I agree that #5333 is the better approach. I'll close in favor of that PR.

PR #5316, opened 2026-08-06
Burn port of bunsen:#157 TensorData API Changes

crutcher:
On-going discussion with @laggui.

See: https://github.com/zspacelabs/bunsen/issues/157

### Changes

- add typed and runtime-dtype tensor readback methods
- add fallible `TensorData` conversions and vector extraction
- add typed immutable and mutable indexed views
- introduce `TensorReadError` and structured `DataError` variants
- deprecate `to_vec` and `into_vec` in favor of explicit `try_*` variants
- split `TensorData` into focused modules

Breaking: `Tensor::try_into_scalar` now returns `TensorReadError`, and the `DataError` variants have been reworked


crates/burn-std/src/data/tensor.rs
```diff
@@ -118,6 +133,134 @@ impl TensorData {
...
     }
 
+    /// Returns an [`Index`] view wrapper of the [`TensorData`].
+    ///
+    /// The view implements [`Deref<Target=TensorData>`].
+    ///
+    /// # Example
+    /// ```rust,no_run
+    /// use bunsen::burner::tensor::*;
+    /// use burn::prelude::*;
```
laggui: The examples still refer to `bunsen` 😄 will have to fix the imports

crates/burn-std/src/data/tensor.rs
```diff
@@ -118,6 +133,134 @@ impl TensorData {
...
+    /// assert_eq!(view[&[0, 0]], 1.0);
+    /// assert_eq!(view[&[0, 1]], 2.0);
+    /// assert_eq!(view[&[1, 0]], 3.0);
+    /// assert_eq!(view[&[1, 1]], 4.0);
+    /// ```
+    ///
+    /// # Returns
+    /// `Ok(view)` on success; `Err(DataError::TypeError)` on view [`DType`]
+    /// missmatch.
+    pub fn try_index_view<E: Element>(&self) -> Result<TensorDataView<'_, E>, DataError> {
```
laggui: The return type already communicates `TensorDataView`, indexing is one capability of the view rather than its permanent scope. It could technically support iteration, slicing, etc.

For that reason I think the methods should be stripped of the "index" scope, e.g. `try_view` instead in this case.

crates/burn-std/src/data/tensor.rs
```diff
@@ -118,6 +133,134 @@ impl TensorData {
...
+    /// assert_eq!(view[&[1, 0]], 3.0);
+    /// assert_eq!(view[&[1, 1]], 4.0);
+    /// ```
+    ///
+    /// # Returns
+    /// The view.
+    ///
+    /// # Panics
+    /// If the view [`DType`] is not compatible with the [`TensorData`].
+    pub fn expect_index_view<E: Element>(&self) -> TensorDataView<'_, E> {
```
laggui: The panicking counterparts to the fallible `try_*` methods don't need to have a prefix.

```rust
// Fallible
data.try_view::<f32>()?
data.try_view_mut::<f32>()?

// Panics on incompatible storage
data.view::<f32>()
data.view_mut::<f32>()
```

crates/burn-std/src/data/tensor.rs
```diff
@@ -457,14 +600,61 @@ impl TensorData {
...
+    /// Convert the data to a new dtype.
+    ///
+    /// By contract, this is equivalent to:
+    /// `data.try_cast(E::dtype())`
+    ///
+    /// # Returns
+    /// Ok(data) on success, (Currently) panics on failure.
+    pub fn try_convert<E: Element>(self) -> Result<TensorData, DataError> {
+        self.try_cast(E::dtype())
+    }
```
laggui: I think the names should be coherent. We can stick to convert or cast, but not both.

If we go with cast, I'd opt for something like:
```rust
try_cast::<E>()
try_cast_dtype(dtype)
```

crates/burn-std/src/data/tensor.rs
```diff
@@ -457,14 +600,61 @@ impl TensorData {
...
+
+    /// Cast the data to a new dtype.
+    ///
+    /// # Returns
+    /// Ok(data) on success, (Currently) panics on failure.
+    pub fn try_cast(self, dtype: DType) -> Result<TensorData, DataError> {
```
laggui: The branches reach `convert_inplace_with` and `convert_clone_with` whose `checked::cast_slice` calls panic. Consequently, the new fallible methods like `try_cast` panic instead` of returning a `DataError`.

We need to have fallible conversion methods to propagate the failures.

crates/burn-std/src/data/tensor.rs
```diff
@@ -782,15 +982,219 @@ impl core::fmt::Display for TensorData {
...
+}
+
+impl<'a, E> DerefMut for TensorDataViewMut<'a, E>
+where
+    E: Element,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.data
+    }
+}
```
laggui: `DerefMut<Target = TensorData>` can invalidate the view's typed invariant. Right now `TensorData` fields are public, so callers could mutate the dtype or even bytes through this `DerefMut`. We should probably close the fields, but that's another issue. Even with private fields, callers can replace the entire dereferenced value:

```rust
let mut data = TensorData::new(vec![1.0f32, 2.0], [2]);
let mut view = data.expect_index_mut_view::<f32>();
// whoops
*view = TensorData::new(vec![1i32, 2, 3], [3]);
```

I'd either remove the `DerefMut` entirely (element mutation remains available through `IndexMut`), or change the `TensorDataViewMut` to hold a slice of typed values:
```rust
struct TensorDataViewMut<'a, E> {
    values: &'a mut [E],
    shape: Shape,
}
```
where `DerefMut` would be restricted to actual typed elements.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2010,6 +2010,121 @@ where
...
+    /// By contract, this will yield the same result as
+    /// `tensor.to_data().try_cast(dtype)`.
+    ///
+    /// The conversion is a no-op if the dtype is the same as the current dtype.
+    ///
+    /// # Returns
+    /// `Ok(data)`, or an error on data conversion errors.
+    pub fn try_to_data_cast(&self, dtype: DType) -> Result<TensorData, DataError> {
+        self.to_data().try_cast(dtype)
+    }
```
laggui: Following the `TensorData` cast/convert APIs, I think the runtime dtype APIs should be named

```rust
tensor.try_into_data_dtype(dtype)
tensor.try_to_data_dtype(dtype)
```

I would avoid `*_data_cast`. "Cast" sounds like the chosen mechanism (`Tensor::cast`) even though it currently isn't the case, and might eventually be "let Burn choose between backend casting and host conversion". I think `into_data_dtype` describes the result better.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -2010,6 +2010,121 @@ where
...
+    /// By contract, this will yield the same result as
+    /// `tensor.into_data().convert::<E>(dtype)`.
+    ///
+    /// The conversion is a no-op if the dtype is the same as the current dtype.
+    ///
+    /// # Returns
+    /// `Ok(data)`, or an error on data conversion errors.
+    pub fn try_into_data_as<E: Element>(self) -> Result<TensorData, DataError> {
+        self.into_data().try_convert::<E>()
+    }
```
laggui: The `try_*` methods are not fully fallible. They only catch the conversion `DataError`, but `into_data` might fail. We have `tensor.try_into_data` which could return an `ExecutionError`. Looks like we don't have the `try_to_data` counterpart, but that's an easy addition.

laggui (changes requested):
Overall a great start, see my comments below

crates/burn-std/src/data/tensor.rs
crutcher: done.

crates/burn-std/src/data/tensor.rs
crutcher: Ok, done.

crates/burn-std/src/data/tensor.rs
crutcher: done.

crates/burn-std/src/data/tensor.rs
crutcher: to match `Tensor::cast(dtype)`, I'm going with:
* `TensorData::try_cast(dtype)`
* `TensorData::try_cast_as::<E>()`

crates/burn-std/src/data/tensor.rs
crutcher: So; yes and no.

The code currently does the right thing, because it's match for failure exactly covers the target's match for `unreachable!()`. But there's very little protecting us from drift (other than a cross-test fully-enumerated match block unittest that I haven't written yet).

I could unroll them, so `try_cast` and `try_cast_as` are the actual impls; and all the current callers call those; so that the `unreachable!()` blocks aren't in the code anymore.

crates/burn-std/src/data/tensor.rs
laggui: Yeah that make sense as well

crates/burn-std/src/data/tensor.rs
laggui: I think we're discussing two different failure cases.

I initially read the checked cast as another expected conversion failure. But on second thought this should only fail when `TensorData` already violates its bytes/dtype invariant, so either through invalid raw construction (`from_bytes` / `from_bytes_vec`) or field mutation (we'll get rid of that by restricting the visibility).

It might still be worthwhile to use the fallible bytemuck APIs defensively, but this is a separate hardening concern. It doesn't have to be addressed here so I'll resolve this discussion :+1:

crates/burn-std/src/data/tensor.rs
laggui: @crutcher I'm guessing this was done locally? No changes seem to have been pushed.

crates/burn-std/src/data/tensor.rs
crutcher: stripped the Deref/DerefMut; added ...View::{shape,dtype}().

crates/burn-tensor/src/tensor/api/base.rs
crutcher: ok

crates/burn-tensor/src/tensor/api/base.rs
crutcher: Yes; i noticed the issue with the existing API, and wanted to pick ... somewhere to draw the line.

I've taken a stab at this; but I don't like the available error surfaces. DataError and ExecutionError are both too-narrow; take a look?

Also, I noticed that the into_scalar stuff was a mess; and cleaned it up.

Ideally, we'd also do TensorData::{try_,}item::<E>() / item_as::<E>. But the try_ cases have the same problem; DataError is too restrictive, and in this case the shape MetadataError isn't right.

crutcher:
```rust
   pub fn try_into_data_as<E: Element>(self) -> Result<TensorData, DataError> {
        self.try_into_data()
            .map_err(|e| DataError::TypeMismatch(format!("{e:?}")))?
            .try_cast_as::<E>()
```

crates/burn-std/src/data/tensor.rs
crutcher: Yeh; I was using this as a to-do as I worked through stuff. I triggered re-review when I was done.

crates/burn-tensor/src/tensor/api/base.rs
crutcher: @laggui I need your attention on this. We might want to migrate TensorData to have an extended DataError type?

crutcher:
Also, when we're happy with the interface; I need to flush out the tests to match.

crates/burn-tensor/src/tensor/api/base.rs
laggui: Cross-posting from [discord](https://discord.com/channels/1038839012602941528/1091796857996451942/1536341209553829960):

I'd avoid wrapping `ExecutionError` inside `DataError`. The latter should describe operations on an existing `TensorData`. And execution failure happens when syncing a device, which happens when obtaining  `TensorData` from a `Tensor`. I'd introduce a composed error instead that wraps `Execution` and `Data` errors.

I'd probably clean up the existing `DataError` categories instead of overloading `CastError` e.g.

```rust
pub enum DataError {
    DTypeMismatch {
        expected: DType,
        actual: DType,
    },
    UnsupportedConversion {
        source: DType,
        target: DType,
    },
    InvalidRepresentation(CheckedCastError),
}
```
`item()` is another discussion. We already have `into_scalar` / `try_into_scalar`. Are you suggesting a pytorch-compatible naming (which I don't think is necessary), or recoverable shape validation?

Side note: it's easier to keep the conversation focused on one channel :sweat_smile: instead of duplicating or overlapping between github and discord

crutcher:
@laggui take a look at the (terrible) `Tensor::try_into_data_as` machinery; with the double `Result`. That's where we're going to leverage the new `TensorDataError` in the next PR.

Also, I added something from `bunsen`; `WithOkOrPanic`. The `.unwrap()` call panic messages end up making the `thiserror` errors nearly worthless; this machinery gives much better panic logs.

crutcher:
`Tensor::try_into_data_as` has a stupid signature, and will until the follow up pr adding the TensorDataError type to unify DataError and ExecutionError sub-cases:
```rust
    let piv_data: Vec<f32> = piv
        .try_into_data_as::<f32>()
        .unwrap()
        .unwrap()
        .try_into_vec()
        .unwrap();
```

This also means that `tensor.into_data().try_into_vec()` hasn't been simplified (yet) like `tensor.to_data().try_into_vec()` was.

crates/burn-core/src/store/mod.rs
```diff
@@ -470,8 +470,8 @@ mod tests {
 
     fn weights(model: &Tiny) -> (Vec<f32>, Vec<f32>) {
         (
-            model.weight.val().to_data().to_vec().unwrap(),
-            model.bias.val().to_data().to_vec().unwrap(),
+            model.weight.try_to_vec_as().unwrap(),
+            model.bias.try_to_vec_as().unwrap(),
```
laggui: This just made me realize there is a potential footgun with the recent LoRA (& subsequent reparameterization) changes. When a reparameterization is applied, `Deref` bypasses it.

So removing `.val()` here is not semantics-preserving, but that was only highlighted by the changes in this PR. The `Tensor` -> vec convenience methods weren't initially part of the proposed plan though, so I'd keep them for a follow up instead. Especially with what it just highlighted.

crates/burn-std/src/result_ext.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// Extension trait for `Result<T, E>` to add `ok_or_panic` method.
+pub trait WithOkOrPanic<T> {
+    /// Unwraps the `Result`, or panics with the error message.
+    ///
+    /// This differs from the behavior of [`Result::unwrap`]
+    /// in that the Debug format of the wrapped error is used
+    /// directly as the panic message; and not escaped.
+    fn ok_or_panic(self) -> T;
+}
```
laggui: I agree with avoiding raw `unwrap()` behind public panicking APIs, but I don't think this public trait with a blanket extension on `Result` is the solution to improve the panic messages.

We can use `unwrap_or_else` with an operation-specific message in a crate private helper with `#[track_caller]`.

crates/burn-tensor/src/tensor/api/base.rs
```diff
@@ -1992,24 +2006,180 @@ where
...
+    ///
+    /// The conversion is a no-op if the dtype is the same as the current dtype.
+    ///
+    /// # Returns
+    /// Nested result, `Ok(Ok(data))` on success;
+    /// `Err(ExecutionError)` on transaction errors,
+    /// `Ok(Err(DataError))` on dtype errors,
+    pub fn try_into_data_as<E: Element>(
+        self,
+    ) -> Result<Result<TensorData, DataError>, ExecutionError> {
```
laggui: > `Tensor::try_into_data_as` has a stupid signature, and will until the follow and will until the follow up pr adding the TensorDataError type

Yeah that's quite terrible 😅 imo the `TensorDataError` / `TensorReadError` would have been a better addition to this PR than the new `Tensor` -> vec APIs.

I know you said you were going away so I can always take this over if you'd like so we can merge the PR.

laggui (changes requested):

laggui:
- Added `TensorReadError` and flattened Tensor readback errors.
- Removed the public `WithOkOrPanic` helper in favor of contextual panics.
- Made `TensorData` access and conversion properly fallible with clearer `DataError` variants.
- Made typed views validate/materialize once and prevent mutable dtype/metadata invalidation.
- Restored explicit `.val()` calls to avoid the unrelated `Param` deref regression.
- Updated documentation and tests for the new error and readback contracts.

I also reinstated `Deref` and `DerefMut` on the tensor data views, which now hold a validated typed slice directly instead of `TensorData`. It gives the view a direct invariant: this storage is valid as [E] for the view's lifetime. Indexing is then infallible and doesn't repeatedly call `as_slice()` or `as_mut_slice()`. For mutable views, it also naturally limits mutation to elements rather than `TensorData` metadata.

PR #5320, opened 2026-08-07
fix(train): flush async metrics before checkpoint/early-stop

TsaoLun:
## Problem
Training events are processed asynchronously, but checkpointing and early
stopping query the event store immediately after submitting epoch-end events.
Metrics that are finalized only at epoch end can therefore be missing when
queried, so metric-based checkpointing may skip a save (and early stopping may
see stale values).

## Solution
- Add a default no-op `flush` hook on `EventProcessorTraining`.
- Implement it as a FIFO barrier in `AsyncProcessorTraining`.
- Flush pending events before checkpointing or early stopping in the
  single-device, multi-device, and DDP strategies.
- Keep processing fully async when neither feature is enabled.

## Testing
- Unit test: `flush` waits for previously submitted training/validation events.

## Background

This became user-visible after #5218, which moved authoritative values for
global/epoch-level metrics (e.g. AUROC, AUC-PR) from per-batch `update` into
epoch-end `compute()` / `Final`, written during `EndSplit`. Checkpointing and
early stopping still query the event store immediately after enqueueing
epoch-end events on the async processor, so those finals can be missing.


crates/burn-train/src/metric/processor/async_wrapper.rs
```diff
@@ -35,6 +35,9 @@ impl<ET: Send + 'static, EV: Send + 'static, P: EventProcessorTraining<ET, EV> +
                     match msg {
                         Message::Train(event) => worker.processor.process_train(event),
                         Message::Valid(event) => worker.processor.process_valid(event),
+                        Message::Flush(callback) => {
+                            callback.send_blocking(()).unwrap();
+                        }
```
laggui, line 41: I think this should flush the wrapped processor before sending the acknowledgement

crates/burn-train/src/learner/supervised/strategies/ddp/worker.rs
```diff
@@ -126,6 +126,10 @@ impl<M: LearnerModel> DdpWorker<M> {
                 event_processor.process_train(LearnerEvent::EndEpoch(epoch));
             }
 
+            if self.checkpointer.is_some() || self.components.early_stopping.is_some() {
+                self.event_processor.lock().unwrap().flush();
+            }
```
laggui, line 140: In DDP, only the main worker runs validation, while every worker executes the flush and early-stopping query. I think we should synchronize workers before flushing so the main worker has queued the validation and epoch-end events. Otherwise, a secondary worker could reach `flush()` first and query a missing or stale value for early stopping.

laggui (commented):
Good catch!

crates/burn-train/src/metric/processor/async_wrapper.rs
TsaoLun: Done

crates/burn-train/src/learner/supervised/strategies/ddp/worker.rs
TsaoLun: Added a peer_count barrier before the flush, and moved the gating condition to the strategy since it wasn't uniform across workers.

laggui (approved):

I removed the `flush_metrics` field (redundant) and fixed a minor condition for the `interrupter.should_stop()`.

PR #5333, opened 2026-08-09
fix(reduce): return the identity for a zero-length axis, and reject the extrema

### LINKED ISSUE #5293: Reductions on empty f16 tensors panic on Metal: autotune finds no viable plan for length 0
### Describe the bug

Any reduction over an empty `f16` tensor panics on the Metal backend. `sum`, `mean`, `max` and `prod` all fail; elementwise ops on the same tensor are fine, and the identical `f32` reduction is fine.

The reduce autotune has no candidate that accepts a zero-length axis, so the plan comes back empty:

```
Can't execute the autotune plan for key: Sum(SumAutotuneKey { dtype: F16, length: 0 })
 - plan: TunePlan { priorities: [], no_groups: [], groups: {}, returned: [] }
 - results: [AutotuneResult { outcome: Err(sum_chained: An unknown error happened.
   Caused by: CallError(task panicked on device runner thread:
   Should run when selected by autotune.: plane_vectorized_parallel_reduce:
   The input reduce axis length (currently 0) should be at least k (1).)
```

This is a release build, so it is not a `debug_assert`.

### To Reproduce

Added to `crates/burn-backend-tests/tests/tensor_f16.rs` on `main` (`318b8f9a`):

```rust
#[cfg(test)]
mod empty_reduce_probe {
    use super::*;
    use burn_tensor::TensorData;

    fn empty() -> TestTensor<1> {
        TestTensor::<1>::from_data(TensorData::from([] as [f32; 0]), &Default::default())
    }

    #[test] fn r_sum()  { println!("{:?}", empty().sum().into_data()); }
    #[test] fn r_mean() { println!("{:?}", empty().mean().into_data()); }
    #[test] fn r_max()  { println!("{:?}", empty().max().into_data()); }
    #[test] fn r_prod() { println!("{:?}", empty().prod().into_data()); }
    #[test] fn r_elementwise_control() { println!("{:?}", empty().add_scalar(1.0).into_data()); }
}
```

```
cargo test -p burn-backend-tests --no-default-features \
    --features metal --features std --release --test tensor_f16 empty_reduce_probe
```

```
test empty_reduce_probe::r_elementwise_control ... ok
test empty_reduce_probe::r_mean ... FAILED
test empty_reduce_probe::r_prod ... FAILED
test empty_reduce_probe::r_max  ... FAILED
test empty_reduce_probe::r_sum  ... FAILED

test result: FAILED. 1 passed; 4 failed
```

The same probe placed in `tests/tensor.rs` (f32) passes, so this appears specific to `f16`. I could not compare against a CPU backend because `tensor_f16.rs` is gated to the accelerator features, so `ndarray` compiles the file out.

### Expected behavior

A reduction over an empty tensor should return the identity (`0` for `sum`, `1` for `prod`) or an empty/NaN result, consistent with the `f32` path, rather than panicking.

### How this surfaces in the existing suite

This is what makes `autodiff_f16::...::test_mask_select_grad_empty_mask` fail — `mask_select` with an all-`false` mask produces a zero-length tensor, and the test then calls `.sum()` on it. Two failures in the default suite trace back to this:

```
cargo test -p burn-backend-tests --no-default-features --features metal --features std --release
-> test result: FAILED. 518 passed; 2 failed
```

Both are `test_mask_select_grad_empty_mask`, in the `autodiff_f16` binary only; the `f32` `autodiff` binary is clean.

### Additional context

#4949 reports autotune + f16 failing during training, which @nathanielsimard diagnosed as a matmul algorithm missing a feature the autotuner does not identify. This looks like the same shape of problem — the autotuner returning no viable candidate for an f16 case — but a different op and trigger, so I have filed it separately rather than adding to that thread. If they turn out to share a cause, this reproduction is at least small.

Possibly also related to #4878, which addresses zero-sized tensors in `slice_assign` and `tensor/base.rs` — a different code path, but the same class.

Separately, and only in **debug** builds, the same two tests fail earlier for an unrelated reason: `broadcast_shape` in `crates/burn-cubecl/src/kernel/utils.rs` asserts every dim equals the broadcast max or 1, and a zero-sized dim satisfies neither. Worth noting that `(0,)` broadcast with `(1,)` is `(0,)` in both numpy and torch, whereas the current code takes the max and would yield `(1,)`. I tried correcting that locally and it moved the panic into `cubecl-std`'s `permuted.rs`, so it looks like zero-size handling needs work at both levels and I did not want to submit a half-fix.

### Environment

- burn `main` at `318b8f9a`
- Apple M-series, Metal backend via `burn-wgpu`, `rustc 1.97.1`
- Not checked on CUDA or ROCm; the failing kernel selection is in `cubecl`, so other backends may be affected

robertomeroni:
### Checklist

### Related Issues/PRs

Fixes #5293. Contributes to #5310.

Overlaps #5313 (open), which fixes the burn-cubecl half. It fills the type's min/max for an empty
`max`/`min` and NaN for `mean` on every dtype; this PR errors on the extrema and keeps NaN to floats,
since `i32::MIN` is a value the tensor never held and an integer `mean` has no NaN. numpy and torch raise
on both. Happy to defer and rebase if #5313 lands first.

### Changes

Reducing over a zero-length axis panicked on the cubecl backends: `validate_shapes` rejects an empty
reduced axis, so every autotune candidate failed and the panic came out of plan selection.

Folding zero elements yields the operation's identity: `Sum`/`Any` -> 0, `Prod`/`All` -> 1, `Mean` -> NaN
for floats, as in numpy and torch, so those are written straight to the output. The extrema have no
identity in a bounded type and now return `ReduceError`. 

The shared tests then failed on the CPU backends for four pre-existing reasons:

- burn-flex `max`/`min` returned the `±inf` fold seed as the extremum of nothing.
- burn-flex `argmax`/`argmin` indexed `row[0]`: `index out of bounds`.
- burn-flex `mean_dim` rejected an empty axis for every dtype; now integers only, so a float
  `mean_dim` gives NaN like `mean()` and like the other backends. Behaviour change on that path.
- burn-ndarray `arg_view` accepted `[0, 0]` but panicked on `[3, 0]`

### Testing

`cargo run-checks`: exit 0, 0 failures. It needs `--ignore-typos` for a typo in
`burn-core/src/module/param/base.rs:330`, untouched here and failing on `main` too.

20 empty-input cases added to the shared suite (sum, prod, mean, max, min, argmax, argmin, all, any) over
`[0]`, `[3, 0]`, `[0, 3]` and `[0, 0]`. The case #5310 was filed about now passes:

    cargo test --release -p burn-backend-tests --no-default-features \
        --features metal,std,autotune --test autodiff -- mask_select

    main:  The input reduce axis length (currently 0) should be at least k (1)
    here:  8 passed; 0 failed

Full `--test tensor` suites: flex 1738 passed, ndarray 1753 passed, metal `--release` 1527 passed. The
only metal failures are `quantization::accuracy` cases, which fail on unmodified `main` too and vary run
to run.

---
AI assistance was used in developing this change: investigation, validation, additional testing

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/all.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/all.rs
@@ -65,3 +65,30 @@ fn test_all_empty() {
         false,
     );
 }
+
+// Shape [3, 0]: the identity has to be written once per surviving position, which the whole-tensor
+// case above cannot cover.
+#[test]
+fn test_all_dim_empty_axis() {
+    let device = Default::default();
+    let expected = TensorData::from([[true], [true], [true]]);
+
+    expected.assert_eq(
+        &TestTensor::<2>::empty([3, 0], &device)
+            .all_dim(1)
+            .into_data(),
+        false,
+    );
+    expected.assert_eq(
+        &TestTensorInt::<2>::empty([3, 0], &device)
+            .all_dim(1)
+            .into_data(),
+        false,
+    );
+    expected.assert_eq(
+        &TestTensorBool::<2>::empty([3, 0], &device)
+            .all_dim(1)
+            .into_data(),
+        false,
+    );
+}
--- a/crates/burn-backend-tests/tests/tensor/float/ops/all.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/all.rs
@@ -45,3 +45,24 @@ fn test_all_dim_large() {
     let boolean = TestTensorBool::<2>::from_data(TensorData::new(mask, [rows, cols]), &device);
     expected.assert_eq(&boolean.all_dim(1).into_data(), false);
 }
+
+/// `all` folds with AND, whose identity is `true`, so an empty input is vacuously `true`.
+/// Matches numpy and torch; `any` (identity `false`) is covered in `any.rs`.
+#[test]
+fn test_all_empty() {
+    let device = Default::default();
+    let expected = TensorData::from([true]);
+
+    expected.assert_eq(
+        &TestTensor::<1>::empty([0], &device).all().into_data(),
+        false,
+    );
+    expected.assert_eq(
+        &TestTensorInt::<1>::empty([0], &device).all().into_data(),
+        false,
+    );
+    expected.assert_eq(
+        &TestTensorBool::<1>::empty([0], &device).all().into_data(),
+        false,
+    );
+}
--- a/crates/burn-backend-tests/tests/tensor/float/ops/all.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/all.rs
@@ -46,8 +46,7 @@ fn test_all_dim_large() {
     expected.assert_eq(&boolean.all_dim(1).into_data(), false);
 }
 
-/// `all` folds with AND, whose identity is `true`, so an empty input is vacuously `true`.
-/// Matches numpy and torch; `any` (identity `false`) is covered in `any.rs`.
+// Vacuously true: `all` folds with AND, whose identity is `true`.
 #[test]
 fn test_all_empty() {
     let device = Default::default();
--- a/crates/burn-backend-tests/tests/tensor/float/ops/any.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/any.rs
@@ -105,3 +105,30 @@ fn test_any_empty() {
         false,
     );
 }
+
+// Shape [3, 0]: the identity has to be written once per surviving position, which the whole-tensor
+// case above cannot cover.
+#[test]
+fn test_any_dim_empty_axis() {
+    let device = Default::default();
+    let expected = TensorData::from([[false], [false], [false]]);
+
+    expected.assert_eq(
+        &TestTensor::<2>::empty([3, 0], &device)
+            .any_dim(1)
+            .into_data(),
+        false,
+    );
+    expected.assert_eq(
+        &TestTensorInt::<2>::empty([3, 0], &device)
+            .any_dim(1)
+            .into_data(),
+        false,
+    );
+    expected.assert_eq(
+        &TestTensorBool::<2>::empty([3, 0], &device)
+            .any_dim(1)
+            .into_data(),
+        false,
+    );
+}
--- a/crates/burn-backend-tests/tests/tensor/float/ops/any.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/any.rs
@@ -85,3 +85,24 @@ fn test_any_dim_large() {
     let boolean = TestTensorBool::<2>::from_data(TensorData::new(mask, [rows, cols]), &device);
     expected.assert_eq(&boolean.any_dim(1).into_data(), false);
 }
+
+/// `any` folds with OR, whose identity is `false`, so an empty input is `false`. Unlike the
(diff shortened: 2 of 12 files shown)


crates/burn-backend-tests/tests/tensor/float/ops/all.rs
laggui, line 1: We should add tests for `all_dim` too

crates/burn-backend-tests/tests/tensor/float/ops/any.rs
laggui, line 1: Same for `any_dim`

crates/burn-backend/src/backend/ops/int_tensor.rs
```diff
@@ -909,6 +909,9 @@ pub trait IntTensorOps<B: Backend> {
     /// The mean of all elements in the tensor.
     fn int_mean(tensor: IntTensor<B>) -> IntTensor<B> {
         let num_elems = tensor.shape().num_elements() as i64;
+        // Unlike a float there is no `NaN` to fall back on, so reject rather than return whatever
+        // the device's division by zero gives.
+        assert!(num_elems > 0, "Cannot compute mean of empty tensor");
```
laggui: `int_mean` can be overriden by a backend implementation, so ideally this check should live a layer above in `burn-tensor` (the `Numeric` impl for `Int`) which calls `int_mean`.

crates/burn-cubecl/src/kernel/reduce/base.rs
```diff
@@ -28,6 +30,60 @@ pub struct SumAutotuneKey {
...
+        return Ok(output);
+    }
+
+    Ok(full_device_dtype(
+        output.client.clone(),
+        output.shape(),
+        output.device.clone(),
+        InputScalar::new(identity, dtype_to_storage_type(output.dtype)),
+        output.dtype,
+    ))
```
laggui: Instead of allocating a new output, we could refactor `full_device_dtype` to split the empty output creation + fill operation. And then use `fill_device_dtype(output, identity)` here instead.

e.g.

```rust
/// Creates a tensor filled with `value`
pub fn full_device_dtype<R: CubeRuntime>(
    client: ComputeClient<R>,
    shape: Shape,
    device: R::Device,
    value: InputScalar,
    dtype: DType,
) -> CubeTensor<R> {
    let empty = empty_device_dtype(client, device, shape, dtype);

    fill_device_dtype(empty, value)
}

/// Fills an existing tensor with `value`.
pub(crate) fn fill_device_dtype<R: CubeRuntime>(
    tensor: CubeTensor<R>,
    value: InputScalar,
) -> CubeTensor<R> {
    // existing implementaiton
}
```

https://github.com/tracel-ai/burn/blob/02101a2f14b0f7f1b6e172981e83fe52d0939d7d/crates/burn-cubecl/src/ops/numeric.rs#L54

laggui (changes requested):
LGTM, just some minor comments

robertomeroni:
The failing `macos-std-tests (stable)` job is the pre-existing quantization breakage tracked in #5270, not this PR. The two failures are `quantization::accuracy::error_responds_to_scale_param` and `narrow_scale_param_degrades_without_normalization`, both added by #5253.

robertomeroni:
thank you for the good suggestions @laggui 

I have implemented all 4

laggui (approved):

PR #5335, opened 2026-08-10
fix(burn-cubecl): broadcast a size-1 dim against a zero-sized dim

### LINKED ISSUE #5334: `broadcast_shape` rejects a zero-sized dimension on the cubecl backends
**Describe the bug**

`broadcast_shape` in `crates/burn-cubecl/src/kernel/utils.rs` does not accept a
dimension of size 0 broadcasting against a dimension of size 1. For each
dimension it takes the maximum size across the operands and asserts that every
operand is either that maximum or 1. When one operand has size 0 and another has
size 1, the maximum is 1 and the size-0 operand is neither 1 nor the maximum, so
the assert at utils.rs:56 fails with:

```
Broadcast dims must be size 1
```

Under NumPy broadcasting a size-1 dimension broadcasts against a size-0
dimension to give 0, so this is a valid shape that should be accepted.

The check is a `debug_assert`, so the two build profiles diverge:

- Debug: `broadcast_shape` panics with `Broadcast dims must be size 1`.
- Release: the assert is compiled out and `broadcast_shape` returns the
  per-dimension maximum, which is wrong for a zero-sized dim. This is harmless
  for a trivial op whose result is empty, but it leads to an out-of-bounds
  access and a SIGSEGV in the `mask_select` backward pass on an all-false mask.

**To Reproduce**

Minimal, reduce independent, reproduced on current `main` in a debug build on
the cpu backend (`B` is any cubecl backend, the helper is shared):

```rust
let a = Tensor::<B, 2>::zeros([2, 0], &device);
let b = Tensor::<B, 2>::zeros([2, 1], &device);
let _ = (a + b).into_data();
// debug: panics with "Broadcast dims must be size 1" at
// crates/burn-cubecl/src/kernel/utils.rs:56
```

The release SIGSEGV surfaces through the autodiff suite, but only once reduction
over a zero-length axis no longer panics first (see #5310, #5313, #5333). With
that reduce fix applied:

```
cargo test --release -p burn-backend-tests --no-default-features \
    --features cpu,std,autotune --test autodiff -- test_mask_select_grad_empty_mask
```

fails with `signal: 11, SIGSEGV`. Without the reduce fix the same test panics
earlier at the reduce stage, so the broadcast defect is not reachable through
`mask_select` on unpatched `main`.

**Expected behavior**

`broadcast_shape` should treat a size-1 dimension as broadcasting to the other
operand's size, including 0. Broadcasting `[2, 0]` and `[2, 1]` should give
`[2, 0]`, matching NumPy semantics, rather than asserting in debug or returning
an incorrect shape in release.

**Environment**

- burn: current `main` (db95b2c5), `broadcast_shape` unchanged as of this commit
- backend: cubecl-cpu (`burn-cpu`). The helper is shared across the cubecl
  backends, so the debug assert is not cpu specific. The release SIGSEGV was
  confirmed on the cpu backend.

**Additional context**

The defect is pre-existing and independent of the reduction-over-empty-axis
work. It only becomes reachable through `mask_select` once the reduce panic is
fixed, which is how it was found while validating that fix. A fix is likely a
small change in `broadcast_shape` to treat a 0-sized dim as compatible with a
1-sized dim (result 0), plus a test over shapes like `[2, 0]` against `[2, 1]`.

Yashiru:
### Checklist

### Related Issues/PRs

Fixes #5334.

### Changes

`broadcast_shape` computed each output dimension as the maximum size across the operands. Broadcasting instead stretches a size-1 dimension to the other operand's size, which may be 0, and `max(0, 1)` is 1, not 0. A broadcast involving a zero-sized dimension therefore produced the wrong output shape. In debug this tripped the shape assert. In release the assert is compiled out, so the op returned a wrong-shaped result and, through the `can_mut_broadcast` in-place path, indexed out of bounds (SIGSEGV).

- `broadcast_shape` (`kernel/utils.rs`): compute the result as the shared size that is not 1 (0 included) rather than the maximum.
- Each broadcast launcher (`binary`, `binary_float`, `binary_int`, `comparison`, `cross`, `mask_where`): in a valid broadcast a zero-sized dimension always yields an empty output, which has nothing to compute. Return the empty output directly, before the in-place and kernel paths that assume a non-empty output. This also sidesteps `can_mut_broadcast` and the layout code, which are not zero-aware either.

Reproducible directly on `main`, with no dependency on other changes:

```rust
let a = Tensor::<B, 2>::zeros([2, 0], &device);
let b = Tensor::<B, 2>::zeros([2, 1], &device);
let _ = (a + b).into_data();
// debug: panics in kernel/utils.rs. release: wrong shape and out-of-bounds read.
```

### Testing

New `tests/cubecl/broadcast.rs`: broadcasting a zero-sized dimension against a size-1 dimension for `add` (trailing, leading, both zero), `comparison`, and `mask_where` (three operands), plus a regular `1 -> N` broadcast checked by value to guard against regressions.

- 7/7 pass on the cpu backend in both debug and release, so the release out-of-bounds path is exercised.
- Full cubecl test suite: 199 passed, 0 failed. `cargo fmt --check` and `cargo clippy -p burn-cubecl` are clean.
- `cargo run-checks` passes with `--ignore-typos`. The only failure without the flag is a pre-existing typo in `burn-core/src/module/param/base.rs:330`, untouched here and failing on `main` too.

### DIFF
--- a/crates/burn-backend-tests/tests/cubecl/broadcast.rs
+++ b/crates/burn-backend-tests/tests/cubecl/broadcast.rs
@@ -1,0 +1,72 @@
+use super::*;
+use burn_tensor::Shape;
+
+// A size-1 dimension broadcasts to the other operand's size, including 0. Regression for #5334,
+// where broadcast_shape used the per-dimension maximum and turned a 0 into 1, panicking in debug
+// and reading out of bounds in release.
+
+#[test]
+fn broadcast_add_zero_against_one_trailing() {
+    let device = Default::default();
+    let a = TestTensor::<2>::zeros([2, 0], &device);
+    let b = TestTensor::<2>::zeros([2, 1], &device);
+    let out = a + b;
+    assert_eq!(out.shape(), Shape::new([2, 0]));
+    // Materialize to exercise the kernel launch and read-back on the empty result.
+    let _ = out.into_data();
+}
+
+#[test]
+fn broadcast_add_one_against_zero_trailing() {
+    let device = Default::default();
+    let a = TestTensor::<2>::zeros([2, 1], &device);
+    let b = TestTensor::<2>::zeros([2, 0], &device);
+    assert_eq!((a + b).shape(), Shape::new([2, 0]));
+}
+
+#[test]
+fn broadcast_add_zero_against_one_leading() {
+    let device = Default::default();
+    let a = TestTensor::<2>::zeros([0, 3], &device);
+    let b = TestTensor::<2>::zeros([1, 3], &device);
+    assert_eq!((a + b).shape(), Shape::new([0, 3]));
+}
+
+#[test]
+fn broadcast_add_zero_against_zero() {
+    let device = Default::default();
+    let a = TestTensor::<2>::zeros([2, 0], &device);
+    let b = TestTensor::<2>::zeros([2, 0], &device);
+    assert_eq!((a + b).shape(), Shape::new([2, 0]));
+}
+
+#[test]
+fn broadcast_comparison_zero_against_one() {
+    let device = Default::default();
+    let a = TestTensor::<2>::zeros([2, 0], &device);
+    let b = TestTensor::<2>::zeros([2, 1], &device);
+    assert_eq!(a.greater(b).shape(), Shape::new([2, 0]));
+}
+
+#[test]
+fn broadcast_mask_where_zero_against_one() {
+    let device = Default::default();
+    let input = TestTensor::<2>::zeros([2, 0], &device);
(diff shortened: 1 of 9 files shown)


robertomeroni:
nice!

One thing for later, not a blocker: can_mut_broadcast still decides via `shape_lhs < shape_rhs`
(tensor/base.rs:275), which is false for an rhs dim of size 0 -- only a size-1 dim broadcasts away, so
`shape_lhs != shape_rhs && shape_rhs != 1` is the property. Your guards make it unreachable from all six
callers today, so nothing is broken; it just leaves the predicate wrong for the next caller that skips
the guard

laggui (approved):

Issue #5131, opened 2026-07-06
[🤖] Dispatch: Int/Bool results of autodiff-tracked ops report a plain device, then silently re-lift to autodiff on `.float()`


swfsql:
**Describe the bug**
On the Dispatch backend, taking an autodiff-tracked float and running a kind-changing op followed by a cast back to float, e.g. `x.argmax(1).float()` or `x.greater_equal_elem(1.0).float()` results in the tensor being promoted back to autodiff, where:

- The Int/Bool intermediate reports `device().is_autodiff() == false`,
- Yet the `.float()` result is a genuinely autodiff-tracked tensor again.

The tensor's reported device and its actual routing disagree, so code that checks `is_autodiff()` to keep gradient-free work (metrics, logging) off the autodiff backend is silently defeated. Downstream this surfaces as *"tensors are not on the same backend"* panics when the recast result meets a genuinely plain tensor, or as unbounded autograd-node accumulation when it doesn't.

**To Reproduce**
<details>
<summary>Reproduction</summary>

- `int_into_float_relifts_to_autodiff` and `bool_into_float_relifts_to_autodiff` assert the *expected* behavior and **fail** on the pinned rev (the final assert sees the re-lifted autodiff tensor).
- `plain_origin_does_not_lift` (control) passes: the same chain from a plain-device float stays plain.
- `workaround_inner_before_the_cast_chain` passes: calling `.inner()` on the float *before* the `float→int/bool→float` chain strips the marker at the source, so the whole chain stays plain.

```toml
[dependencies.burn]
git = "https://github.com/tracel-ai/burn.git"
rev = "d028234e2ccd75c5ca57d1c71ae74456e0e15c4f" # 2026-06-25 main
default-features = false
features = ["flex", "autodiff", "optim", "std"]
```

```rs
#[cfg(test)]
mod autodiff_implicit_recast {
    use burn::prelude::*;

    fn ad_device() -> Device {
        Device::default().autodiff()
    }

    fn sample(device: &Device) -> Tensor<2> {
        Tensor::<2>::from_floats([[0.0f32, 1.0], [2.0, 0.5]], device)
    }

    #[test]
    fn int_into_float_relifts_to_autodiff() {
        let x = sample(&ad_device()); // (autodiff)
        assert!(x.device().is_autodiff()); // ok (autodiff)
        let x_idx = x.argmax(1); // float->int
        assert!(!x_idx.device().is_autodiff()); // ok (plain)
[...]

PR #5340, opened 2026-08-10
fix(tensor): preserve autodiff device for non-float tensors

### LINKED ISSUE #5337: Tensor::one_hot() drops autodiff from backend when original tensor is autodiff
I have a custom loss function used in training (currently using libtorch->CUDA backend).  When computing the  one hot encoding for the targets (which is initially a classification label index and lives on an Autodiff variant of the libtorch->CUDA backend), Autodiff is dropped from the computed tensor, causing subsequent operations with other "on-device" tensors to panic due to the incompatibility between autodiff and non-autodiff variants.

```
pub fn forward(&self, logits: Tensor<2>, targets: Tensor<1, Int>) -> Tensor<1> {
        let [_batch_size, num_classes] = logits.dims();

        // probabilities with clamping via epsilon at both the min and max
        let probs = softmax(logits.clone(), 1).clamp(self.eps, 1.0 - self.eps);

        // one-hot encode target indices -> [batch_size, num_classes]
        let targets_one_hot = targets.one_hot::<2>(num_classes);

        // reveals autodiff dropped from one hot tensor
        eprintln!("targets_one_hot: {}", targets_one_hot);

       // remainder of function removed for brevity
}
```

tianrking:
### Checklist

### Related Issues/PRs

Fixes #5337.

### Changes

`DispatchTensor::device` only reconstructed an autodiff device for quantized float tensors. Integer labels still reported the inner backend device, so operations such as `one_hot(...).float()` left the autodiff backend and could not be combined with logits on the tape.

Preserve the autodiff capability whenever checkpointing metadata is present, including for integer and boolean tensors. The regression test covers both regular and checkpointing autodiff backends and verifies that gradients still reach the logits after multiplying by one-hot labels.

### Testing

- `cargo test -p burn-backend-tests --test autodiff --no-default-features --features std,ndarray bridge:: -- --nocapture` (4 passed)
- `cargo clippy -p burn-dispatch --no-default-features --features std,autodiff,ndarray -- -D warnings`

### DIFF
--- a/crates/burn-dispatch/src/tensor.rs
+++ b/crates/burn-dispatch/src/tensor.rs
@@ -413,10 +413,13 @@ impl TensorMetadata for DispatchTensor {
 
         #[cfg(feature = "autodiff")]
         if let Some(checkpointing) = &self.checkpointing {
-            // Non-float tensors travel beside the tape untracked, but their device
-            // must retain the autodiff capability so tensors derived from them can
-            // join the graph (for example, an integer one-hot tensor cast to float).
-            if !matches!(device, DispatchDevice::Autodiff(_)) {
+            // Int, bool, and quantized tensors travel beside the tape untracked, but
+            // their device must retain the autodiff capability so tensors derived
+            // from them can join the graph (for example, an integer one-hot tensor
+            // cast to float). Plain float gradients can also carry checkpointing
+            // metadata copied from their source tensor, but remain on the inner
+            // backend and must continue to report that device.
+            if !self.dtype().is_float() && !matches!(device, DispatchDevice::Autodiff(_)) {
                 device = DispatchDevice::autodiff(device);
             }
             if let DispatchDevice::Autodiff(device) = &mut device {
--- a/crates/burn-dispatch/src/tensor.rs
+++ b/crates/burn-dispatch/src/tensor.rs
@@ -411,19 +411,12 @@ impl TensorMetadata for DispatchTensor {
         #[allow(unused_mut)]
         let mut device = self.kind.device();
 
-        // TODO: should int and bool kinds return an autodiff device?
-        // It would be much easier once there is a single underlying primitive type, which
-        // we can wrap with Autodiff in all cases.
-
         #[cfg(feature = "autodiff")]
         if let Some(checkpointing) = &self.checkpointing {
-            // A packed (quantized) tensor is float-kind data that travels beside
-            // the tape untracked, so its device is the autodiff one: whatever is
-            // built from it — its dequantized form, LoRA factors over it — must
-            // land on the tape.
-            if matches!(self.dtype(), DType::QFloat(_))
-                && !matches!(device, DispatchDevice::Autodiff(_))
-            {
+            // Non-float tensors travel beside the tape untracked, but their device
+            // must retain the autodiff capability so tensors derived from them can
+            // join the graph (for example, an integer one-hot tensor cast to float).
+            if !matches!(device, DispatchDevice::Autodiff(_)) {
                 device = DispatchDevice::autodiff(device);
             }
             if let DispatchDevice::Autodiff(device) = &mut device {
--- a/crates/burn-backend-tests/tests/autodiff/bridge.rs
+++ b/crates/burn-backend-tests/tests/autodiff/bridge.rs
@@ -37,5 +37,6 @@ fn one_hot_preserves_autodiff_device() {
     let loss = (logits.clone() * one_hot).sum();
     let grads = loss.backward();
 
-    assert!(logits.grad(&grads).is_some());
+    let logits_grad = logits.grad(&grads).expect("logits should have gradients");
+    assert!(!logits_grad.device().is_autodiff());
 }
--- a/crates/burn-backend-tests/tests/autodiff/bridge.rs
+++ b/crates/burn-backend-tests/tests/autodiff/bridge.rs
@@ -23,3 +23,19 @@ fn test_full_precision() {
     assert!(x1_grad.is_some());
     assert!(x2_grad.is_some());
 }
+
+#[test]
+fn one_hot_preserves_autodiff_device() {
+    let device = AutodiffDevice::new();
+    let targets = TestTensorInt::<1>::from_data([0, 1], &device);
+    assert!(targets.device().is_autodiff());
+    let one_hot = targets.one_hot::<2>(2).float();
+
+    assert!(one_hot.device().is_autodiff());
+
+    let logits = TestTensor::<2>::ones([2, 2], &device).require_grad();
+    let loss = (logits.clone() * one_hot).sum();
+    let grads = loss.backward();
+
+    assert!(logits.grad(&grads).is_some());
+}


swfsql:
This may be related to https://github.com/tracel-ai/burn/issues/5131, but on a different angle? (I wanted it to not be autodiff, this case wants it to be autodiff?)

tianrking:
Yes, it is the same underlying inconsistency viewed from the other side. In #5131 the Int/Bool tensor retains checkpointing metadata but reports a plain device; #5337 then hits that mismatch when `one_hot(...).float()` interacts with autodiff logits. This PR takes the first consistency option described in #5131: when checkpointing metadata is present, report the autodiff-capable device. That makes a later float conversion unsurprising and fixes the one-hot interop, but it intentionally does not make metric-style casts stay plain. If non-differentiable results should instead discard checkpointing metadata, that would be the alternative design and this fix should change direction; I am happy to align with the preferred semantics.

fuzing:
@tianrking and @swfsql........ some additional comments on this.

For some additional context, my original issue [#5337](https://github.com/tracel-ai/burn/issues/5337) didn't fully elaborate the full use-case...... hopefully the following more detailed loss function elaborates.....

```
pub fn forward(&self, logits: Tensor<2>, targets: Tensor<1, Int>) -> Tensor<1> {
        let [_batch_size, num_classes] = logits.dims();

        // probabilities with clamping via epsilon at both the min and max
        let probs =  softmax(logits.clone(), 1).clamp(self.eps, 1.0 - self.eps);

        // the one_hot operation results in a tensor with dropped autodiff
        let targets_one_hot = targets.one_hot::<2>(num_classes).float();

        let alpha_broadcast = self.alpha.clone().unwrap().unsqueeze_dim::<2>(0);

        let log_probs = probs.clone().log();

        let focal_weight = (probs.neg().add_scalar(1.0)).powf_scalar(self.gamma);

        // panics due to the fact that targets_one_hot now lives on a non-autodiff backend!
        let loss = alpha_broadcast * focal_weight * log_probs * targets_one_hot;

       // remainder of function removed for brevity
}
```

Given that, in my particular case (i.e. a loss function), the one_hot() result doesn't actually need autodiff (not used in backprop, but for intermediate calculations only), then dropping the autodiff would be ok as long as the tensor can still be used in calculations alongside other autodiff tensors (Although I would disagree with myself in the context of model.forward, where I'd argue that autodiff should remain).  Further to @tianrking 's reference to semantics, this certainly could go either way in terms where the resultant tensor lands (autodiff/non-autodiff), but clearly there should be consistent behavior across tensor types (Float/Int/Bool/Etc.), and, IMHO, whatever is chosen should not be a surprise for people coming from other frameworks.

swfsql:
For my case, I have many tensors that go through that, and other frameworks make it non-autodiff (if I'm not mistaken) - since this makes it explicit that there is no actual backprop possible in that point (but take this with a grain of salt).

In the function, I preemptively make them `inner()` before the operation (so everything autodiff-related is stripped), but only if I'm in "autodiff-mode" (as the same function is used both by non-autodiff and autodiff paths), and calling `inner()` on a non-autodiff tensor panics. Finally, I get that information by asking some tensors what they are (are you autodiff mate?). I think I also juggle a little between inference vs autodiff devices as well.

fuzing:
@swfsql  - that all makes sense.  I'd reiterate that burn should probably mimic whatever other frameworks do, but would always defer to the burn core team as to philosophy (they have a better idea as to the granular inner-workings, and the roadmap).   If it can be determined definitively that other frameworks drop the autodiff then burn should probably follow suit (else the other way around :smiley:)

laggui (commented):
Hey, thanks everyone for your inputs.

I think the distinction here is between a tensor being differentiable and belonging to an autodiff-capable device/backend.
  
Int and bool tensors are non-differentiable in Burn, just as in other frameworks: they do not record gradients and cannot be backpropagation targets. However, before backend generics were removed from the public `Tensor` API, these tensors could still have the concrete backend `Autodiff<B>`. Their primitives were pass-throughs, but their backend context remained explicit. This is in line with other frameworks like PyTorch (though its default execution mode has autograd recording enabled globally).
  
This PR preserves that existing contract. An int/bool tensor reporting an autodiff device does not mean backpropagation is possible through it; it means that a later conversion to float remains compatible with other tensors in the same autodiff computation. This is important for values such as masks, indices, and one-hot labels.

I agree though that the current API makes this confusing because `device().is_autodiff()` can easily be mistaken for "this tensor is gradient tracked". Those are different concerns/questions.

- `is_autodiff()` should mean that the tensor belongs to an autodiff-capable execution context
- `is_require_grad()` indicates gradient retention for float tensors
- Int/bool tensors never carry gradients, even when `is_autodiff()` is true

PR #5319 recently added `tensor.no_grad()` which always leaves the autodiff context, though I think a better name could be `tensor.without_autodiff()`.

So I think your expectation that the intermediate is non-differentiable is correct. Where I would differ is that non-differentiable should not automatically mean leaving the autodiff backend context. We should make that distinction explicit and provide a simple API for users who intentionally want to leave it. Perhaps a better way to provide this scope. But that's a larger design decision and shouldn't block this PR.

@swfsql does that make sense?

I'd accept the direction of the current PR, though we might have to validate it doesn't break any other assumptions in the current dispatch logic.

swfsql:
@laggui Yes, I think it does. Some usecases I had were about metrics indeed (in which I actually prefer running things on inference-only devices, even if they came from autodiff ops), and others were when using backprop-breaking ops that converts to float, where this can be either in a inference-only or autodiff call path.

So I think I was mistaken in thinking that autodiff implied backprop. Yup, two separated things.

So either way, I think the surprising behavior is a non-autodiff tensor becoming autodiff from a float call, so I also think the current PR is the right direction.

Edit: i.e. if I care about metrics on inference-only devices, I can call `inner()` _after_ the `float()` call just fine - I'd have to no longer make that call ahead-of-time.
Edit2: I noticed this because the status quo somewhat pushed AIs, when facing the panic, to try to work-around the panic itself, by avoiding making the backprop-breaking ops themselves (they usually went to weird sub-optimal workarounds).

tianrking:
I tracked the crates failures to a narrower dispatch regression: checkpointing metadata is also copied onto extracted plain-float gradients. Reconstructing an autodiff device for those gradients made operations such as `zeros_like()` create autodiff-wrapped optimizer state, which Adan then tried to combine with inner-backend gradients.

Commit `3dd9cea3` now reconstructs the autodiff-capable device only for non-plain-float dtypes (Int/Bool/QFloat), preserving the intended one-hot behavior while keeping extracted float gradients on the inner device. The regression test now asserts both sides of that contract.

Local validation:
- `cargo test -p burn-optim --lib` (101 passed)
- `cargo test -p burn-backend-tests --test autodiff --no-default-features --features std,ndarray bridge:: -- --nocapture` (4 passed)
- `cargo clippy -p burn-dispatch --no-default-features --features std,autodiff,ndarray -- -D warnings`

The new workflow run is currently marked `action_required` with no jobs created, so a maintainer will need to approve and run it.

crates/burn-dispatch/src/tensor.rs
```diff
@@ -413,10 +413,13 @@ impl TensorMetadata for DispatchTensor {
...
-            // must retain the autodiff capability so tensors derived from them can
-            // join the graph (for example, an integer one-hot tensor cast to float).
-            if !matches!(device, DispatchDevice::Autodiff(_)) {
+            // Int, bool, and quantized tensors travel beside the tape untracked, but
+            // their device must retain the autodiff capability so tensors derived
+            // from them can join the graph (for example, an integer one-hot tensor
+            // cast to float). Plain float gradients can also carry checkpointing
+            // metadata copied from their source tensor, but remain on the inner
+            // backend and must continue to report that device.
+            if !self.dtype().is_float() && !matches!(device, DispatchDevice::Autodiff(_)) {
```
laggui, line 422: Good catch regarding float gradients. I think we should clear the checkpointing instead at the `grad()` and `grad_remove()` boundary, but I'll investigate that in a follow-up. But I'll approve & merge this as an immediate fix.

I think burn-dispatch would benefit from a refactor so the autodiff context is more robust and not implicitly carried by the checkpointing field.

laggui (approved):

crates/burn-dispatch/src/tensor.rs
fuzing: Everyone's efforts are very much appreciated - thank you for your responsiveness, and for the lively discussion :thinking:

laggui: Thanks for filing the issue!

PR #5349, opened 2026-08-12
feat(pack): stream tensors on demand when writing burnpack files

### LINKED ISSUE #5219: burn_pack::Writer requires eager materialization of all tensor bytes before writing
## Summary

`burn_pack::Writer` requires every tensor's `Bytes` to already be resident in memory before writing starts. `write_to_file` / `into_bytes` / `size()` all call `plan()` → `build_descriptors()` first, which reads `tensor.bytes.len()` for every `Tensor` in the `Vec<Tensor>` to compute the aligned offset table — before any I/O happens. So the caller must materialize every tensor's data up front just to build the `Vec<Tensor>` that goes into `Writer::new`.

This is asymmetric with the reader: the README advertises tensor data as "read lazily from files," which is true — `Reader::from_file` reads only the header+metadata up front, and each tensor's bytes are a `Bytes::view` window read from disk on first access. There's no equivalent for writing.

It's also a regression from the pre-#5064 `burn_store::BurnpackWriter`, which held `Vec<TensorSnapshot>` and called `snapshot.to_data()` **inside** the per-tensor write loop — materialize, write, drop, repeat — bounding peak memory to the largest single tensor instead of the whole tensor set.

## Impact

`BurnpackStore::collect_from` has the same shape: it maps every `TensorSnapshot` through `bridge::snapshot_to_tensor` (which calls `to_data()`, materializing) before building `Vec<Tensor>`. Any caller saving a large `Module` — or, in our case, [burn-onnx](https://github.com/tracel-ai/burn-onnx), which writes `.bpk` files during codegen straight from a `Vec<TensorSnapshot>` — must hold the entire tensor set in host memory at once. burn-onnx's model-check suite includes multi-GB models (`stable-diffusion-xl`, `qwen`), so this meaningfully raises peak RSS during codegen.

We worked around it in burn-onnx with a hand-rolled streaming writer. Doing that required duplicating `burn-pack`'s private on-disk layout types (`Metadata`, `TensorDescriptor` are `pub(crate)`) — not something a downstream crate should have to maintain.

## Root cause

- `build_descriptors()` reads `tensor.bytes.len()` for every tensor to compute offsets, before any bytes are written.
- `Tensor.bytes: Bytes` is a concrete, already-resolved field — no lazy-provider variant.
- `write_tensor_data`'s `WRITE_CHUNK_SIZE` streaming only bounds one *already-resident* tensor's write (useful for device-resident tensors); it doesn't address how many tensors are resident simultaneously going into the write loop.

## Suggested fix

A two-phase `Writer`: register each tensor's metadata (name, dtype, shape, byte length, param_id) without its bytes to build the offset table, then supply bytes per tensor via a callback/lazy provider during the actual write pass — mirroring what the old `BurnpackWriter` did with `TensorSnapshot`.

## Environment

Confirmed against `main` @ `f2321dde9549f5b9b2296b34876e168ebe4efcaf` (2026-07-22) — `writer.rs`, `reader.rs`, and `base.rs` are unchanged since #5064 (2026-06-16).

antimora:
### Checklist

### Related Issues/PRs

Closes #5219

### Changes

`burn_pack::Writer` required every tensor's bytes resident before writing: the offset table was computed from `bytes.len()` over the whole set, so saving a model meant holding all of it in host memory at once.

- Whether a tensor's bytes are resident or produced on demand is now a property of `Tensor` itself. `Tensor::deferred` takes a byte length plus a provider; `Writer` still operates on `Vec<Tensor>` and draws each provider once, in write order, dropping each tensor's bytes before the next. `BurnpackStore` file saves are therefore bounded by the largest single tensor rather than the whole module.
- `Writer::write_to_file_atomic` builds the container in a scratch sibling and renames it into place, so a provider failing mid-write cannot truncate an existing file. `write_to_file` still writes in place; only `BurnpackStore`, which has deferred tensors, uses the atomic path. See #5399.
- Fixes `TensorSnapshot::data_len()` for quantized dtypes: Q4/Q2 under `QuantStore::Native` computed zero value bytes, and a spurious alignment round-up inflated shapes whose value bytes are not a multiple of 4. Latent for safetensors offsets, fatal for a writer that reserves space from it.
- Re-exports `burn_pack` from burn-store so downstream crates (burn-onnx) can write `.bpk` straight from a `Vec<TensorSnapshot>` instead of duplicating private layout types.

### Testing

New suites: `burn-pack/tests/lazy_writer.rs` (laziness, both length guards, in-place vs atomic file behavior including rename failure), `burn-pack/tests/streaming_memory.rs` (counts live heap bytes to pin the one-tensor-at-a-time bound), and `burn-store/tests/{direct_writer,burnpack_store}.rs` (snapshot-driven writes through the re-export, file mode, overwrite, adapter saves). Quantized `data_len` is swept across Q8/Q4/Q2 with shapes hitting both historical failure modes.


crates/burn-store/src/tensor_snapshot.rs
```diff
@@ -244,27 +244,28 @@ impl TensorSnapshot {
...
         match self.dtype {
             DType::QFloat(scheme) => {
-                // Calculate value bytes using scheme's packing information
+                // Each storage element packs `num_quants` values into `size_bits_stored` bits.
+                // Round that width up to whole bytes: a `QuantStore::Native` Q4 or Q2 value is
+                // narrower than a byte but still occupies one, and dividing down reports zero.
                 let num_storage_elements = num_elements.div_ceil(scheme.num_quants());
                 let value_bytes =
-                    num_storage_elements * (scheme.size_bits_stored() / BITS_PER_BYTE);
+                    num_storage_elements * scheme.size_bits_stored().div_ceil(BITS_PER_BYTE);
```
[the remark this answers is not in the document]
antimora: Good catch, fixed in 79304bda7. The storage-shape computation at allocation divides only the last dimension's extent, so packing is per line and the flat div_ceil under-counts any non-divisible packed extent. data_len now computes (num_elements / last) * last.div_ceil(num_quants) for packed stores, with Native unchanged at one storage element per value. No current backend can materialize the counterexample (cubecl rejects non-divisible extents at allocation and the test backend rewrites packed schemes to Native), so the fix is pinned by a hand-computed formula test asserting the [3, 3] Q4 PackedU32 case at 12 value bytes. One note: the allocation code packs the last dimension regardless of the scheme's packing_dim field (nothing in the workspace reads it), so data_len mirrors the allocation rather than the field.

crates/burn-pack/src/writer.rs
```diff
@@ -388,10 +491,165 @@ impl Sink for BufferSink<'_> {
...
+impl ScratchFile {
+    /// Create a scratch file alongside `destination` and open a sink onto it.
+    ///
+    /// The guard and its file are handed back together so neither can exist without the
+    /// other: no path is reserved that nothing will clean up, and no file is created that no
+    /// guard is watching.
+    fn create(destination: &Path) -> Result<(Self, FileSink), Error> {
+        let path = Self::path_beside(destination)?;
+        let file = File::create(&path)
+            .map_err(|e| Error::IoError(format!("cannot create '{}': {e}", path.display())))?;
```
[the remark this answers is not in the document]
antimora: Agreed, File::create through a pre-planted symlink would truncate its target, and the pid+counter name is predictable. Fixed in 79304bda7: the scratch file is now opened with File::create_new, which refuses an existing path and does not follow a symlink, and AlreadyExists (stale leftover, pid reuse, planted link) is answered by retrying with the next counter value. The retry loop itself has no direct test since the process-global counter is not observable from outside, but the exclusive open is the load-bearing part.

crates/burn-pack/src/writer.rs
```diff
@@ -35,19 +35,27 @@ const fn align_offset(offset: u64, alignment: u64) -> u64 {
...
     pub(crate) metadata: BTreeMap<String, String>,
     /// Typed scalars keyed by name
     pub(crate) scalars: BTreeMap<String, Scalar>,
 }
 
-impl Writer {
+impl<T: TensorEntry> Writer<T> {
     /// Create a new writer
-    pub fn new(tensors: Vec<Tensor>) -> Self {
+    pub fn new(tensors: Vec<T>) -> Self {
```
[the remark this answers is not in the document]
antimora: The inference break is real but narrower than the comment suggests: it only affects Writer::new(vec![]) where nothing constrains the element type, which in this workspace is exactly one test line, and the E0283 help names the fix. Splitting into new(Vec<Tensor>) plus from_entries would put the primary new use, Writer::new(snapshots), behind the less discoverable name permanently to avoid a one-time turbofish in a pre-1.0 crate, and generic code over Writer<T> would lose the uniform constructor. Keeping the single generic new; I have updated the PR description to state the empty-vec caveat instead of claiming unqualified compatibility.

laggui (commented):
The PR fixes the right behavioral problem, but I think `TensorEntry` and `Writer<T>` put the abstraction boundary in the wrong place.

`burn_pack::Tensor` is already the format-level representation of a tensor. Whether its bytes are immediately available or produced on demand feels like a property of `Tensor` (or possibly `Bytes`, but that's a bigger change) rather than a separate writer input abstraction.

What if `Tensor` held an eager-or-deferred byte source internally? That would let `Writer` continue operating on `Vec<Tensor>` while preserving deferred, one-at-a-time materialization.

Some direct access to the public `Tensor::bytes` field might need to move behind an accessor, but the overall `burn-pack` API would remain non-generic and largely unchanged.

antimora:
@laggui agreed, and done. `TensorEntry` and `Writer<T>` are gone; `Tensor` now holds a private `Source` enum (`Resident(Bytes)` | `Deferred { len, provider }`), built with `Tensor::new` or `Tensor::deferred`, and `Writer` is back to a plain `Vec<Tensor>`.

The symmetry argument is the one that should have led me there originally: `Reader` has always handed back tensors whose bytes are read lazily from the source, so a separate trait for the write side made the two directions asymmetric for no reason.

It also removed a bug class rather than defending against one. Under the trait, `name()` was called once while planning and again while writing, with nothing guaranteeing the two agreed - review found that a disagreement could silently place tensors at wrong offsets. A plain `String` field cannot disagree with itself. The `Writer::new(vec![])` turbofish is gone too.

One prerequisite worth flagging, since it lands in burn-store's public API rather than burn-pack's: a deferred `Tensor` has to stay `Send`, because `OptimizerRecord` holds `Vec<Tensor>` and burn-train's `Checkpoint` is `Send + 'static`. That meant `TensorSnapshot::data_fn` moving from `Rc` to `Arc`. Mechanical (the captured tensors were already `Send + Sync`), but it constrains every future `data_fn`, so say the word if you'd rather it went another way. The pointer and its bound are `cfg`-selected so `thumbv6m-none-eabi` still builds.

On your `Bytes` parenthetical: bigger for a concrete reason, since `burn_std::Bytes` is a re-export of `cubecl_environment::bytes`, so it would need an upstream change. Worth noting `Bytes` is already lazy about the *copy* (file- and device-backed buffers report their length without materializing); `Source::Deferred` defers a step earlier, the call that produces the `Bytes` at all. Converting a whole model eagerly would submit every device readback up front even if each returned a lazy buffer.

Filed #5378 for the related convergence between `TensorSnapshot` and `Tensor` now that both are metadata plus a lazy byte source, plus #5379 and #5380 for two smaller things review turned up.

crates/burn-pack/src/writer.rs
```diff
@@ -388,10 +482,182 @@ impl Sink for BufferSink<'_> {
...
+    /// On rename failure the finished container is deleted by the guard, so the error names
+    /// it: the bytes were written and then discarded, which is worth saying out loud.
+    fn persist(mut self, sink: FileSink, destination: &Path) -> Result<(), Error> {
+        sink.finish()?;
+
+        std::fs::rename(&self.path, destination).map_err(|e| {
+            Error::IoError(format!(
+                "cannot move the completed container '{}' onto '{}': {e}",
+                self.path.display(),
+                destination.display()
```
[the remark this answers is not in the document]
antimora: Half right, and the half that lands is worth fixing - done in 4334fa013.

The stated mechanism does not hold though. "A process failure during this rename" cannot violate anything: `std::fs::rename` is a single `MoveFileExW(.., MOVEFILE_REPLACE_EXISTING)` call (std `sys/fs/windows.rs:1322`), and a process dying cannot tear a kernel call. It either took effect or it did not, and neither leaves a partial file. That is exactly what the doc scoped the guarantee to, and it was accurate on Windows.

What Windows' missing guarantee actually concerns is system-level failure, and there the docs really were too strong - one sentence over from where you pointed. The old text said that off Unix, power loss "can revert the destination to the old container", which asserts the old file survives. Without a documented atomic replace it may not. That now says the destination holds one or the other and does not promise which, and names `MoveFileEx` as the reason.

You were right that the terse restatements were unscoped. burn-store's README said saves "land atomically" with no qualifier, and the book's "an interrupted save" reads as covering power loss; both now say process-level explicitly and point at the API docs for the rest. burn-pack's README already said "a failed write", which was accurate, but it led with an unqualified "all-or-nothing", so that is scoped too.

Not taking the first half of the suggestion. A platform-specific atomic replace means `ReplaceFileW`, plus a fallback for when the destination does not exist yet, and it would harden only the system-crash case on the one platform where the directory sync the rest of the guarantee rests on is unavailable anyway. That is a lot of platform-specific surface for a narrow gain, and it should be its own change with its own testing rather than a rider on this one. Scoping the claim to what the primitive actually provides is the proportionate fix.

crates/burn-pack/src/tensor.rs
```diff
@@ -1,22 +1,86 @@
...
+#[cfg(target_has_atomic = "ptr")]
+pub trait ByteSource: Fn() -> Result<Bytes, Error> + Send + Sync + 'static {}
+#[cfg(target_has_atomic = "ptr")]
+impl<T> ByteSource for T where T: Fn() -> Result<Bytes, Error> + Send + Sync + 'static {}
+
+/// Without atomic CAS there are no threads, so the bound is just the closure itself.
+#[cfg(not(target_has_atomic = "ptr"))]
+pub trait ByteSource: Fn() -> Result<Bytes, Error> + 'static {}
+#[cfg(not(target_has_atomic = "ptr"))]
+impl<T> ByteSource for T where T: Fn() -> Result<Bytes, Error> + 'static {}
```
laggui: Is the public `ByteSource` trait necessary? It appears to exist only as a stable trait-alias workaround for the target-dependent closure bounds. We could instead `cfg` the two `Tensor::deferred` signatures and delegate both to a private helper taking `Provider`. That would keep the callback implementation private and avoid adding another public trait/re-export. The same applies to `DataSource` in `TensorSnapshot`.

crates/burn-pack/src/writer.rs
```diff
@@ -388,10 +482,182 @@ impl Sink for BufferSink<'_> {
...
+            ))
+        })?;
```
laggui, line 612: I don't think the scratch-and-rename behavior should implicitly replace `Writer::write_to_file` for every caller.

The new mid-write failure mode comes specifically from deferred tensors: `TensorSnapshot -> PackTensor` stores a provider closure that calls `snapshot.to_data()` only when the writer reaches that tensor. `BurnpackStore` uses this path, so a provider can fail after writing has started, making atomic replacement valuable there.

`ModuleRecord`, however, already stores materialized `TensorData` and converts it with `Tensor::new`. It has no deferred provider, yet `ModuleRecord::save` inherits the additional disk-space requirement (space for both copies).

Could we preserve the existing in-place `write_to_file` behavior and add an explicit `write_to_file_atomic` method (or a `WriteMode` option)? `BurnpackStore` could use the atomic path for deferred tensors, while callers with resident data could retain the existing behavior or opt into atomic replacement deliberately.

crates/burn-pack/src/tensor.rs
```diff
@@ -45,12 +109,186 @@ impl Tensor {
...
+    /// The tensor's raw little-endian bytes, leaving it intact.
+    ///
+    /// Named `to_` rather than `bytes` because it can cost the whole tensor: a `Deferred` one
+    /// re-runs its provider (nothing is cached), and a `Resident` one is deep-copied unless
+    /// its [`Bytes`] can share their backing. Prefer [`into_bytes`](Self::into_bytes), or
+    /// [`into_parts`](Self::into_parts) when the metadata is wanted too.
+    pub fn to_bytes(&self) -> Result<Bytes, Error> {
+        match &self.source {
+            Source::Resident(bytes) => Ok(bytes.clone()),
+            Source::Deferred { provider, .. } => provider(),
```
laggui: The deferred `len` is discarded here, and likewise in `into_bytes` and `into_parts`. Only `Writer::materialize` currently checks the provider’s output length, so materialization paths that bypass the writer do not enforce the source invariant.

`Source` should probably resolve its bytes and validate that `bytes.len() == len` for every deferred materialization path, e.g.

```rust
impl Source {
    fn to_bytes(&self) -> Result<Bytes, Error> {
        match self {
            Self::Resident(bytes) => Ok(bytes.clone()),
            Self::Deferred { len, provider } => Self::validate_deferred(*len, provider()?),
        }
    }

    fn into_bytes(self) -> Result<Bytes, Error> {
        match self {
            Self::Resident(bytes) => Ok(bytes),
            Self::Deferred { len, provider } => Self::validate_deferred(len, provider()?),
        }
    }

    fn validate_deferred(expected: usize, bytes: Bytes) -> Result<Bytes, Error> {
        if bytes.len() != expected {
            return Err(Error::TensorBytesSizeMismatch(format!(
                "deferred source has inconsistent length (expected {}, got {})",
                expected,
                bytes.len()
            )));
        }

        Ok(bytes)
    }
}
```

and `source.to_bytes/into_bytes` can be reused

laggui (commented):

Just a couple of final comments, after which the PR should be good to go

crates/burn-pack/src/tensor.rs
antimora: Agreed, done in 6f27ee4fd. Both traits are gone; `Tensor::deferred` and `TensorSnapshot::data_fn` are each two cfg'd signatures delegating to a private constructor taking `Provider` / `DataFn`, and neither `ByteSource` nor `DataSource` is exported any more.

Worth noting they were worse than just an extra public trait: because the bound was cfg'd, `ByteSource` was a *public trait whose shape changed with the target*, and it was declared `pub` inside a private module and never re-exported, so downstream could not name it while it sat in `Tensor::deferred`'s public signature. rustdoc rendered the bound as unlinkable text.

The cost is the contract doc duplicated across the two cfg arms, only one of which ever renders. I trimmed it so the duplicated part is short and the full contract lives on the type.

crates/burn-pack/src/tensor.rs
antimora: Good catch, and worse than it looks from the diff: `byte_len()` keeps reporting the declared number, so an unvalidated path hands back bytes of a different length than the tensor claims, silently. `bridge.rs`'s `From<PackTensor> for TensorSnapshot` goes through `to_bytes()`, so it was one of them.

Fixed in 6f27ee4fd, essentially as you wrote it: `Source` resolves and validates its own bytes for both `to_bytes` and `into_bytes`, `into_parts` goes through the same path, and `Tensor` just delegates.

One consequence worth flagging: this makes `Writer::materialize`'s own length check unreachable, since `placement.len` and the source's `len` are the same private field and nothing can mutate between planning and writing. I removed it rather than leaving a dead branch; `materialize` now only annotates the error with the tensor name via `in_tensor`. `Placement::len` became unused and went with it. The existing mid-write guard test still passes, now failing inside `Source` instead.

crates/burn-pack/src/writer.rs
antimora: Agreed, done in 6f27ee4fd. `write_to_file` writes in place again, exactly as before this PR; `write_to_file_atomic` does the scratch-and-rename, and `BurnpackStore` is the only caller that uses it. A test pins that in-place writing really does truncate on a failing provider, so the two cannot quietly converge.

Two things I found while checking the cost, both of which support the split more than the diff shows.

The disk-space point is not the main one. `FileSink::finish` does `sync_all()` on the whole container, so every save fsynced the full file, and that applied whether or not the destination existed. The second-copy requirement only bites when overwriting, which `FileCheckpointer` never does since it writes `{name}-{epoch}.bpk`.

The fsync is also less costly than it sounds, in the caller that matters: `AsyncCheckpointer::save` returns as soon as it hands the record to a channel, so the write and its fsync run on the spawned thread and training never blocks on them.

So the cost landed mainly on someone calling `ModuleRecord::save("model.bpk")` synchronously over an existing file. Which is also the caller who loses their previous checkpoint when the in-place write fails partway. That is a real gap, just not one this PR should be deciding: filed #5399 to argue atomic-by-default on its own merits, with the numbers above, and noted there the open question of whether the rename and the fsync should be separable, since they are different guarantees at very different prices.

crates/burn-store/src/tensor_snapshot.rs
```diff
@@ -1,11 +1,32 @@
...
+
+#[cfg(target_has_atomic = "ptr")]
+use alloc::sync::Arc;
+// `alloc::sync` needs atomic CAS. A target without it has no threads to share a snapshot
+// across, so neither the atomic pointer nor the `Send + Sync` bound on [`DataFn`] is
+// needed there, and both are dropped below.
+#[cfg(not(target_has_atomic = "ptr"))]
+use alloc::rc::Rc as Arc;
 use burn_core::module::ParamId;
 use burn_core::tensor::quantization::quantized_data_len;
```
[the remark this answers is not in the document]
antimora: You are right, and thank you for the specific citation - it is what made this checkable. Fixed in 7ca7a2125.

`CubeTensor::quantized_storage` does exactly what you describe:

```rust
shape[packed_dim] = shape[packed_dim].div_ceil(scheme.num_quants());
```

Only the packed extent is rounded up; the other dimensions stay. `quantized_data_len` flattened first, which under-counts whenever that extent is not a multiple of `num_quants` - your `[3, 3]` Q4 `PackedU32` example is 3 words, not 2.

`storage_elements` now builds the same storage shape the allocation uses, and `quantized_data_len` multiplies from that. Covered across a two-dimensional non-divisible shape, a higher-rank one (`[2, 5, 9]`), and a divisible one that must still agree with the flattened count. The existing sweep against `to_data()` is unchanged, so the materializable schemes stay pinned to real bytes rather than to the formula.

Worth recording how this got here, since it is the second time this arithmetic has been wrong. The previous merge pulled in `quantized_data_len` from upstream and dropped a local test that had asserted the per-line layout. I concluded no per-line rule existed after grepping only burn-std - the rule is in burn-cubecl, where the allocation happens - so I resolved in favour of the shared helper and deleted the test that was telling me otherwise. That test is restored, with the multidimensional case you asked for.

On severity: latent while the number only fed safetensors offsets, fatal here. A deferred tensor declares its length from `data_len()` before the writer reserves space, so an under-count fails the save outright rather than writing something subtly wrong.

crates/burn-pack/src/writer.rs
```diff
@@ -85,6 +93,11 @@ impl Writer {
     /// - Custom allocators
     /// - Pinned memory for GPU transfers
     ///
+    /// On failure the buffer's contents are unspecified: a deferred [`Tensor`] produces its bytes
+    /// during the write, so an entry that fails partway leaves everything before it already
+    /// copied in. Callers reusing a buffer across writes cannot treat an error as "nothing
+    /// happened". [`write_to_file`](Self::write_to_file) has no such caveat.
```
laggui: I think this should point to `write_to_file_atomic` right?

There are similar stale references in the `ScratchFile`/`FileSink` docs.

Otherwise, LGTM. Just need to rebase with main which fixed the clippy lints and CI should be happy.

laggui (approved):

crates/burn-pack/src/writer.rs
antimora: Right on both counts, fixed in b28030bea.

`write_into`'s caveat is not that `write_to_file` escapes it, it is that only the atomic path does, so it now points at `write_to_file_atomic` and says plainly that `write_to_file` has the same caveat, on the destination rather than on the caller's buffer.

Three more were stale for the same reason: `ScratchFile` and `FileSink::finish` exist only for the atomic path, and `Layout` is now shared by both writers. Swept the rest of the tree too; the remaining `write_to_file` references are either call sites or descriptions of the in-place behavior, which are correct as they stand.

Synced with main in 1e2973266. One conflict, in `tests/common/mod.rs`: the clippy pass rewrote `read_f32` to `as_chunks::<4>()` over `&tensor.bytes`, and this branch replaced that public field with an accessor. Kept your iteration, reached through `to_bytes()`.

Worth flagging one thing that is not in the original scope: 7ca7a2125 touches `burn-std`. `quantized_data_len` flattened the element count before dividing by `num_quants`, but a packed store rounds up only the packed extent and leaves the other dimensions intact, the way `CubeTensor::quantized_storage` builds the storage shape. So a `[3, 3]` Q4 `PackedU32` tensor is `3 * ceil(3 / 8) = 3` words, not `ceil(9 / 8) = 2`. That was latent while the number only fed safetensors offsets; it is fatal for a deferred tensor, which declares its length from `data_len()` before the writer reserves space, so an under-count fails the save outright. Happy to split it into its own PR if you would rather it not ride along.

PR #5355, opened 2026-08-13
Bump cubecl and cubek revs for the pliron codegen fixes

nathanielsimard:
Adapt the fused quant matmul to QuantizedView::new's per-tensor scale parameter (cubecl#1479): two-level schemes are rejected where the scale layout is built, so there is never a second factor to fold in here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2224,7 +2224,7 @@ dependencies = [
 [[package]]
 name = "cubecl"
 version = "0.11.0-pre.2"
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -172,7 +172,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
```
laggui: We can't change the windows crates versions. It breaks windows CI (version mismatch with the wgpu downstream requirements).

If you could just keep the cubecl/cubek changes

PR #5360, opened 2026-08-13
Validate TensorData byte length on deserialization

### LINKED ISSUE #5353: TensorData deserialization skips length validation, causing heap out-of-bounds write and read when loading a crafted weight file
**Describe the bug**
`TensorData` (in `burn-std`, `burn-backend` in 0.21) guarantees `bytes.len() == numel(shape) * dtype.size()`, but this invariant is enforced only by the `TensorData::new` / `quantized` constructors. The derived `Deserialize` impl — used by every file recorder (`NamedMpkFileRecorder`, `PrettyJsonFileRecorder`, `BinFileRecorder`) — and the public `from_bytes` / `from_bytes_vec` constructors skip the check, and `burn-ndarray`'s `NdArrayTensor::from_data_owned` then builds the array with `unsafe { ArrayD::from_shape_vec_unchecked(shape, vec) }` without validating length.

A weight file whose `shape` field is larger than its `bytes` payload therefore deserializes into a tensor that claims N elements over a smaller buffer, and any subsequent operation reads and writes out of bounds. The out-of-bounds length is fully controlled by the file's `shape` field.

**Vulnerable source code** (verbatim, burn 0.21.0)

`burn-backend` `src/data/tensor.rs` — the length invariant is declared, but only `new`/`quantized` enforce it:

```rust
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct TensorData {
    pub bytes: Bytes,
    #[serde(with = "shape_inner")]
    pub shape: Shape,
    pub dtype: DType,
}
```

```rust
fn check_data_len<E: Element>(data: &[E], shape: &Shape) {
    let expected_data_len = Self::numel(shape);
    let num_data = data.len();
    assert_eq!(
        expected_data_len, num_data,
        "Shape {shape:?} is invalid for input of size {num_data:?}",
    );
}
```

```rust
pub fn from_bytes_vec<S: Into<Shape>>(bytes: Vec<u8>, shape: S, dtype: DType) -> Self {
    Self {
        bytes: Bytes::from_bytes_vec(bytes),
        shape: shape.into(),
        dtype,
    }
}
```

`burn-core` `src/record/tensor.rs` — the deserialization entry point shared by all file recorders calls the derived `Deserialize`, which never runs `check_data_len`:

```rust
let data = TensorData::deserialize(deserializer).map_err(|e| {
```

`burn-ndarray` `src/tensor.rs` — `into_vec` validates only the dtype (not the length), and the array is built with the unchecked constructor:

```rust
fn from_data_owned(data: TensorData) -> NdArrayTensor {
    let shape = data.shape.to_vec();

    macro_rules! execute {
        ($data: expr, [$($dtype: pat => $ty: ty),*]) => {
            match $data.dtype {
                $( $dtype => {
                    match data.into_vec::<$ty>() {
                        Ok(vec) => unsafe { ArrayD::from_shape_vec_unchecked(shape, vec) }.into_shared(),
                        Err(err) => panic!("Data should have the same element type as the tensor {err:?}"),
                    }.into()
                }, )*
                other => unimplemented!("Unsupported dtype {other:?}"),
            }
        };
    }
```

**To Reproduce**
Dependencies: `burn = { package = "burn-core", version = "=0.21.0" }`, `burn-ndarray = "=0.21.0"`.

Minimal `src/main.rs`:

```rust
use burn::tensor::{DType, Tensor, TensorData};
use burn_ndarray::{NdArray, NdArrayDevice};

type B = NdArray<f32, i64, i8>;

fn main() {
    let device = NdArrayDevice::Cpu;
    // 4 bytes of payload, but the shape claims 1000 f32 elements. This is what a
    // weight file whose `shape` field was enlarged (without changing `bytes`)
    // deserializes into.
    let bad = TensorData::from_bytes_vec(vec![0u8, 0, 128, 63], [1000usize], DType::F32);
    let t = Tensor::<B, 1>::from_data(bad, &device); // from_shape_vec_unchecked, no length check
    let x = Tensor::<B, 1>::zeros([1000], &device);
    let _ = t.add(x); // SIMD kernel reads/writes 1000 f32 over the 4-byte buffer
}
```

Steps:

1. Build with **release** profile: `cargo build --release` (important — see note below).
2. Run under Valgrind: `valgrind --num-callers=25 ./target/release/<binary>`.

Result (raw):

```
==3604999== ERROR SUMMARY: 131 errors from 16 contexts (suppressed: 0 from 0)
==3604999== Invalid write of size 8
==3604999==    at ... run_vectorized::imp_fastcall ... (mod.rs:552)
==3604999==    by ... binary_simd_same ... (src/ops/simd/binary.rs:227)
==3604999==    by ... try_binary_simd ... (src/ops/simd/binary.rs:184)
==3604999==    by ... NdArrayMathOps ... add ... (src/ops/base.rs:546)
==3604999==  Address 0x4ac85e0 is 0 bytes inside a block of size 4 alloc'd
```

The 16 contexts split as 8 `Invalid read of size 32` + 8 `Invalid write of size 8`.

The same path is reached end-to-end via the file recorders: serialize a one-element f32 weight with `shape = [1]`, edit the recorded `shape` field to `[1000]` (leaving the 4-byte payload intact), then `Recorder::load` → `Tensor::from_data`. Verified against `NamedMpkFileRecorder` (`.mpk`, 131 errors), `PrettyJsonFileRecorder` (`.json`, 131 errors), and `BinFileRecorder` (`.bin`, 30 errors).

**Expected behavior**
Loading a weight file whose `shape` and `bytes` disagree should fail with a validation error (e.g. `check_data_len` on deserialization, or the checked `ArrayD::from_shape_vec` returning an error), never silently construct a mismatched array that later performs out-of-bounds heap reads/writes.

- OS: Linux x86-64 (Ubuntu 22.04, kernel 5.15)
- Version: burn 0.21.0 / rustc 1.96.0-nightly / valgrind 3.18.1 / ndarray 0.17.2

**Additional context**
- **Classification**: memory-safety / soundness violation reachable from safe Rust (PoC is `#![forbid(unsafe_code)]`); CWE-787 (out-of-bounds write) + CWE-125 (out-of-bounds read). Suggested severity: High.
- **Affected versions**: 0.21.0 (tested). The same code path is present un-fixed on `main` (`burn-std/src/data/tensor.rs:23` derived `Deserialize`; `burn-ndarray/src/tensor.rs:682` `from_shape_vec_unchecked`) and in `0.22.0-pre.1`.
- **Root cause**: `TensorData`'s derived `Deserialize` bypasses `check_data_len`; `from_data_owned` calls `ArrayD::from_shape_vec_unchecked` with an unvalidated shape/length.
- **Suggested fix**: in `from_data_owned`, use the checked `ArrayD::from_shape_vec(shape, vec)` and propagate the error (or assert `vec.len() == shape.numel()` before the `unsafe` call); additionally add a manual `Deserialize` for `TensorData` that runs `check_data_len`.
- **Note on reproduction**: debug builds (`cargo run` without `--release`) trip ndarray's `debug_assert!` in `from_shape_vec_unchecked` and panic cleanly *before* any out-of-bounds access, so the memory corruption is only observable in a **release** build under Valgrind. A plain release run without Valgrind exits `0` (silent corruption).
- **Clean control**: loading the *un-tampered* file (shape `[1]`, 4 bytes) and running the same op reports `0 errors from 0 contexts`.

robertomeroni:
### Checklist

- [x] Made sure the book is up to date with changes in this PR. (no book change needed: the documented invariant is unchanged, it is now enforced)

### Related Issues/PRs

Fixes #5353. Conflicts textually with #5316, which touches both files but keeps `from_shape_vec_unchecked` and the derived `Deserialize`.

### Changes

`TensorData` documents `bytes.len() == numel(shape) * dtype.size()`, but only `new`/`quantized` enforce it. The derived `Deserialize` did not, so a file whose `shape` field was enlarged without changing its payload loaded as a tensor claiming more elements than its buffer holds, and `from_data_owned`'s `ArrayD::from_shape_vec_unchecked` then let ops read and write out of bounds - from safe Rust.

`TensorData`'s `Deserialize` now validates the length through a wire struct, covering the serde paths (`burn-remote`, msgpack/JSON records). The element count uses `checked_mul`, so an overflowing shape is rejected rather than wrapping into a count that matches the payload. `QFloat` is excluded - `DType::size()` is not the stored width of quantized data. burnpack and safetensors instead use `from_bytes`, which is unchecked by design, so `from_data_owned` switches to the checked `ArrayD::from_shape_vec` (already used elsewhere in the crate) to cover those callers. `burn-ndarray` is deprecated but is still the `unsafe` site; `burn-flex` needs no change, since the same input trips a slice bound in `as_slice` (`tensor.rs:248`). The check is O(rank) and never scans the data.

### Testing

Three regression tests, each confirmed to fail on the unfixed code by reverting the fix and keeping the test. Miri on the unfixed code reports the UB directly (with `-C debug-assertions=off`, since ndarray's own `debug_assert!` otherwise panics first), and is clean after:

```
error: Undefined Behavior: constructing invalid value of type &[f32]: encountered a dangling
reference (going beyond the bounds of its allocation)
  --> ndarray-0.17.2/src/impl_views/conversions.rs:43:27
```

---
AI assistance was used in developing this change. Claude explored the load paths and ran the Miri/ASan reproductions and test suites; the diff was reviewed line by line before submission.


laggui (approved):
Sorry for the delay, that one slipped through.

Changes LGTM, though I think the broader `TensorData` contract should be tightened in follow-up work. I'll open an issue and assign myself to look into that.

PR #5364, opened 2026-08-14
fix(module): build LoRA factors at an explicit dtype over a packed base

nathanielsimard:


crates/burn-core/src/module/param/lora.rs
```diff
@@ -26,6 +26,15 @@ impl Reparameterization for LoraAdapter {
...
         let delta = self.delta().reshape(base.shape());
+        // The factors set the precision the compose runs at. A dense base was
+        // matched to them when they were built; a packed base dequantizes to
+        // the device default, which is not theirs whenever the model computes
+        // at another precision.
+        let base = if base.dtype() != delta.dtype() {
+            base.dequantize().cast(delta.dtype())
+        } else {
+            base
+        };
```
laggui, line 37: https://github.com/tracel-ai/burn/blob/67dbe6ea15f0fdb40043fe4316e437ff0eb8427c/crates/burn-tensor/src/bridge/ops/float.rs#L22-L36

In the mixed QFloat/Float addition case, the QFloat operand is already dequantized directly to the Float operand's dtype. This explicit dequantize-and-cast is redundant, we can remove it and the test should still pass.

PR #5367, opened 2026-08-14
Improve benchmark defaults and py script to compare benchmarks

laundmo:
### Checklist

### Related Issues/PRs

N/A

### Changes

- Moved the divan call and memory print into a `common` benchmark run function
- Changed sample-count away from divan default, 100 -> 600
- Added max time limit, so long-running benchmarks use less samples
- Added a python script which saves divan benchmark results and compares them to the previous run.

TODO: verify how the diff script works with memory profiling enabled

### Testing

Ran benchmarks
For example, heres me testing whether the CONV_PLANE_OH_OUTER_THRESHOLD does anything by setting it to 0 during the second run:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ examples/**/Cargo.lock
 *profiler.json
 # linker maps
 *.map
+# samply profiling files
+*.json.gz
+# benchmarking helper files
+.divan_diff.json
--- a/crates/burn-backend-tests/README.md
+++ b/crates/burn-backend-tests/README.md
@@ -150,13 +150,14 @@ cargo bench-flex --bench matmul -- square
 
 ### Comparing backends
 
-Each run targets one backend. To compare (e.g. Flex vs NdArray), run twice and diff the output:
+Each run targets one backend. To compare (e.g. Flex vs NdArray), run twice while piping into divan-diff.py (python uv script):
 
 ```sh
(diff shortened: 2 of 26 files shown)


laundmo:
I use the python script regularily, and its very useful.

No clue about the CI failures, they seem unrelated.

laggui (commented):
> No clue about the CI failures, they seem unrelated.

Definitely unrelated. They should be fixed if you update with main.

> TODO: verify how the diff script works with memory profiling enabled

Is this why this is still a draft?

At first glance, the shared benchmark runner and updated defaults look useful. For the comparison helper, would it make sense to use a Rust utility in xtask to avoid requiring Python/uv? I also noticed the parser retains the previous group name when indentation decreases, which might mix benchmark baselines. Maybe the helper can be split into a follow-up so the runner/default changes can land independently.

laundmo:
> >    TODO: verify how the diff script works with memory profiling enabled
> 
> Is this why this is still a draft?

Yup.

That, and as you pointed out, the python/uv dependency. I've chosen that because i find it far simpler to write scripts like this in it, but yea, its not ideal.

laggui:
No prob, feel free to request a review whenever you feel it's ready.

I purposefully left my previous comment as a "review" so you should be able to request one directly.

Issue #5332, opened 2026-08-09
No feature on `burn` forwards `burn-ndarray/multi-threads`, so `default-features = false` silently disables CPU threading


laurigates:
### Feature description

`burn-ndarray`'s default feature set is `["std", "simd", "multi-threads"]`, and
`multi-threads` is what enables threaded GEMM:

```toml
multi-threads = ["rayon", "ndarray/rayon", "matrixmultiply/threading"]
```

The workspace declares the backend with `default-features = false`:

```toml
burn-ndarray = { path = "crates/burn-ndarray", version = "=0.22.0-pre.1", default-features = false }
```

so `burn` never picks those defaults up, and re-exposes them selectively:
`std`, `simd`, and the four BLAS passthroughs (`accelerate`, `blas-netlib`,
`openblas`, `openblas-system`). There is none for `multi-threads` — grepping
`crates/burn/Cargo.toml` for `rayon` or `thread` returns nothing.

A dependant enabling burn's `default` feature therefore gets threading via
`burn-ndarray?/default`. One setting `default-features = false` — the usual way
to pull a single backend without the others — gets a single-threaded ndarray
backend and no feature to turn threading back on. The only route is a direct
`burn-ndarray` dependency declared purely for its features.

Nothing in the dependant's manifest hints at this: it shows features being
*added*, never that a default set was dropped upstream.

#### Reproduction

```toml
[dependencies]
burn = { version = "0.21", default-features = false, features = ["std", "ndarray", "autodiff"] }
```

Matmul-heavy CPU work then saturates exactly one core.

### Feature motivation

Single matmul, batch 1, seq 512, hidden 2560. Batch 1 is deliberate: per-batch
parallel iteration has nothing to spread there, so only threaded GEMM inside
the matmul can help. Linux x86_64, 24 cores, release.

| features | ms/matmul | GFLOPS |
|---|---|---|
| none (the reproduction above) | 94.2 | 71 |
| `simd` | 78.8 | 85 |
| `multi-threads` | 32.0 | 210 |
| both | 31.2 | 215 |

**Not a many-core-only win.** Same binary, both features on, capping
`RAYON_NUM_THREADS` / `MATMUL_NUM_THREADS`:

| threads | ms/matmul | speedup |
|---|---|---|
| 1 | 90.6 | 1.0x |
| 2 | 48.3 | 1.9x |
| 4 | 32.9 | 2.8x |
| 8 | 31.7 | 2.9x |
[...]

PR #5369, opened 2026-08-14
refactor(burn): route backend features through the dispatch chain

laurigates:
### Checklist

`cargo run-checks` passes end-to-end (xtask v5: format, typos, `cargo audit`,
`clippy -- --deny warnings`, tests, doctests) on macOS/aarch64 — see Testing. The
open questions at the bottom remain open; they are design calls for you rather than
blockers.

Rebased onto `9d6784e7b`. #5377 (graph capture backend) had since edited the same
feature tables: it made the `ir = ["burn-core/ir"]` change independently and had
already dropped `burn-ir` and `burn-dispatch` from `burn`'s optional dependencies,
so this PR now removes the remainder of that block. `capture` and `extension`
already route through `burn-core`, matching the shape this PR establishes.
`Cargo.lock` was regenerated from the merged manifests rather than hand-resolved.

The book does not document any of the feature flags this touches (`grep` over
`burn-book` for `accelerate|openblas|x86-v4|apple-amx` returns nothing), so there
is nothing to update there. `crates/burn/src/lib.rs` does carry a feature list, and
that is updated.

### Related Issues/PRs

Addresses the `# TODO: clean up from root; backends should be exposed from the
extension` in `crates/burn/Cargo.toml`, and the follow-up @laggui asked for in
https://github.com/tracel-ai/burn/issues/5332#issuecomment-5296266686:

> This cleanup must be addressed. Then, backend dependencies should be configured
> through the dispatch chain. And the `simd`/`rayon` should propagate through that
> rather than directly from `burn` to `burn-flex` imo.

### Changes

**1. `burn` no longer declares the backend crates.** `crates/burn/src` is a single
`lib.rs` that re-exports `burn_core::*`; it never names `burn_cuda`, `burn_flex`,
`burn_ndarray`, `burn_wgpu`, `burn_tch`, `burn_cpu`, `burn_rocm`, `burn_ir`,
`burn_autodiff`, `burn_dispatch` or `burn_router`. Every one of those was already
being enabled a second time down the existing chain
(`burn-core → burn-tensor → burn-dispatch`), so e.g.

```toml
cuda = ["burn-cuda", "burn-core/cuda", "burn-vision?/cuda"]
```

enabled `burn-cuda` twice. The direct arm and the optional dependency are dropped;
`burn-core/cuda` continues to reach it.

**2. Backend-tuning features move to `burn-dispatch` and are forwarded up.** Eight
features on `burn` reached into backend crates directly and had no home in the
dispatch chain, because `burn-dispatch` carried no equivalents. They are now
defined there and forwarded through `burn-tensor` → `burn-core` → `burn`:
`simd`, `x86-v4`, `apple-amx`, `accelerate`, `blas-netlib`, `openblas`,
`openblas-system`, `template`.

**3. New `rayon` feature**, which is what #5332 asked for. `burn-flex`'s `rayon`
and `burn-ndarray`'s `multi-threads` were unreachable from `burn` at all — there
was no passthrough, weak or strong.

This also fixes the weak-reference defect the issue was about, structurally rather
than case by case. `burn-flex` is a **non-optional** dependency of `burn-dispatch`,
so at that level `burn-flex/rayon` needs no `?` and cannot go inert. The old
`simd = ["burn-ndarray?/simd", "burn-flex?/simd"]` at the `burn` level keyed off
whether *`burn`'s own optional `burn-flex`* was enabled, i.e. off the `flex`
feature — so `--features ndarray,simd` compiled a Flex backend with SIMD silently
switched off, even though Flex was in the graph either way. `burn-ndarray` stays
optional and correctly keeps the weak `?` form.

**4. `router` feature and the `burn-router` dependency are removed.** Nothing in
`burn` re-exports a router type and no feature in the chain forwards it, so
`--features router` compiled `burn-router` into the graph with no features enabled
and nothing depending on it (verified below). `burn-remote` depends on
`burn-router` directly and non-optionally and is unaffected. This is the one
deletion I am least certain about — happy to restore it if it is load-bearing in a
way I could not see from the workspace.

### Testing

**Resolution diff, before vs after.** For each feature combination, the fully
resolved package + feature set of the `burn` graph
(`cargo tree -p burn --no-default-features --features <combo> --prefix none -f "{lib} :: {f}"`),
diffed between `main` and this branch:

| combo | packages before → after | delta |
|---|---|---|
| *(default)* | 327 → 327 | none beyond `burn`'s own feature list |
| `cuda` / `rocm` / `tch` / `cpu` | 279/286/210/383 → same | none beyond `burn`'s own feature list |
| `wgpu,template` / `metal` / `vulkan` / `webgpu` | 274/277/285/274 → same | none beyond `burn`'s own feature list |
| `flex` / `flex,x86-v4` / `flex,apple-amx` | 114/114/139 → same | none beyond `burn`'s own feature list |
| `ndarray,accelerate` / `ndarray,openblas` | 130/167 → same | none beyond `burn`'s own feature list |
| `autodiff,fusion,ir` | 115 → 115 | none beyond `burn`'s own feature list |
| `std` / `extension` / `remote` | 140/115/361 → same | byte-identical |
| **`ndarray,simd`** | **134 → 137** | **`simd` now also reaches `burn-flex`** |

"none beyond `burn`'s own feature list" means the only lines that changed are the
implicit `burn-<crate>` feature names disappearing from `burn`'s own resolved
feature set, e.g. `burn :: burn-cuda,cuda` → `burn :: cuda`. No package appears or
disappears, and no other crate's feature set changes.

The one substantive delta is `ndarray,simd`, and it is the bug: on `main`,
`burn-flex` resolves with **no** features there while `burn-ndarray` gets `simd`.
On this branch both get it, pulling in `aligned-vec`, `equator`, `equator-macro`
and `gemm/wasm-simd128-enable`. That is intended, but it is a behavior change for
anyone building `ndarray` + `simd` today, so flagging it explicitly.

**New feature reaches both CPU backends** (`--no-default-features`):

| features | `burn-flex` | `burn-ndarray` |
|---|---|---|
| `flex` | *(none)* | — |
| `flex,rayon` | `rayon` | — |
| `flex,simd` | `simd` | — |
| `ndarray,rayon` | `rayon` | `multi-threads,rayon` |
| `rayon` *(no backend named)* | `rayon` | — |

The last row is the `default-features = false` case from the issue.

**`router` verification.** On `main`, diffing `--features ir` against
`--features router` shows the entire delta is one package, `burn_router ::` with an
empty feature set, and nothing in the tree depends on it. No manifest, example or
workflow in the workspace names the `router` feature.

**Compilation.** `cargo check -p burn` (default features) and, with
`--no-default-features`, each of `flex`, `flex,simd`, `flex,rayon`,
`flex,rayon,simd,std`, `rayon`, `ndarray,rayon,std`, `ndarray,accelerate,std`,
`flex,x86-v4,std`, `std` — all clean on macOS/aarch64. `cargo run-checks` passes
end-to-end (format, typos, `cargo audit`, `clippy -- --deny warnings`, 3,637 tests
plus doctests, 0 failures). No CUDA/ROCm/LibTorch host here, so those arms are
resolution-checked only, and xtask v5's `validate` does not exercise the Metal
backend suite.

One arm fails, **identically on `main`**: `--no-default-features --features
ndarray,simd` (i.e. `simd` without `std`) fails in `burn-ndarray`'s own SIMD code
with `cannot find type Vec in this scope` — a missing `alloc` import, pre-existing
and unrelated to this PR. `ndarray,simd,std` is clean on both sides. Mentioning it
because this PR makes that combination *reachable* in a new way (`simd` now also
reaches `burn-flex`, which compiles fine without `std`), not because it changes it.

The two in-workspace consumers of a touched feature are
`examples/custom-cubecl-kernel` and `examples/custom-wgpu-kernel`, both using
`template`; `wgpu,template` resolves identically before and after.

### Open questions

1. **`simd` fan-out.** I kept one `simd` at `burn-dispatch` that fans out to both
   CPU backends, matching what the `burn`-level feature was written to mean. The
   alternative is per-backend names. Your call — this is a one-line change.
2. **Should `rayon` imply `std`?** I argued for it in the issue, then didn't do it:
   `burn-flex`'s own `rayon = ["dep:rayon", "gemm/rayon"]` doesn't imply `std`, and
   `--no-default-features --features rayon` compiles clean, so forcing `std` would
   have been an opinion rather than a fix. Easy to add if you'd rather it be
   explicit.
3. **Removing the optional deps removes their implicit features**, so
   `burn = { features = ["burn-ndarray"] }` stops resolving. I doubt anyone does
   that, and `ndarray` is the documented spelling, but it is technically breaking.
4. **`router`** — see above.
5. Still open from the issue and unrelated to this diff:
   `burn-dispatch/build.rs` sets `default_backend` when no backend feature is
   enabled, and `burn-flex` is non-optional there — so a `default-features = false`
   consumer naming no backend gets Flex today, not only after `ndarray` is removed.
   With this PR they can at least ask for `rayon`. Is that reading right?


laurigates:
CI: 22 of 23 checks pass. The one failure, `macos-std-tests (stable)`, is
pre-existing on `main` and unrelated to this change.

```
called `Result::unwrap()` on an `Err` value:
  UnsupportedDType { device: "Metal(DefaultDevice)", dtype: Bool(U8) }
  at crates/burn-backend-tests/tests/common/backend.rs:16:10
error: test failed, to rerun pass `-p burn-backend-tests --test autodiff`
```

Rather than assume, I checked it against the base: `macos-std-tests (stable)` fails
on each of the four most recent `main` runs, and on b6e27bdca — the commit this
branch is based on — the signature is identical down to the file and line, the
`UnsupportedDType { device: "Metal(DefaultDevice)", dtype: Bool(U8) }`, and the test
binary. Nothing to do here as far as I can tell, but flagging it in case it isn't
already known.

Two passing jobs cover the parts I couldn't check locally, and they were the two
places a feature-unification mistake would have hidden:

- **`linux-no-std-tests` (stable and prev)** — this PR drops the `burn-autodiff?/std`
  arms from `burn`'s `std` and `doc` features, on the grounds that `burn-dispatch`
  already carries them and the chain reaches it. no_std green on both toolchains is
  the check for that.
- **`linux-std-tests (stable, examples)`** — `examples/custom-cubecl-kernel` and
  `examples/custom-wgpu-kernel` both use `burn`'s `template` feature, which moved to
  `burn-dispatch` here.

The remaining gap is the same one noted in the description: `cargo run-checks` hasn't
been run end-to-end locally, and I have no CUDA/ROCm/LibTorch host, so those arms are
resolution-checked rather than compiled. The `linux-std-tests (backends)` and
`windows-std-tests (backends)` jobs passing covers more of that than I could.

laggui (approved):
> `router` feature and the `burn-router` dependency are removed

That's fine, historically it re-exported the types from `burn-router`. With the backend generics removal, those were removed. External implementations should just depend on the `burn-router` crate directly.

PR #5374, opened 2026-08-16
fix(tensor): prevent softplus from overflowing to inf

### LINKED ISSUE #5373: softplus overflows to inf for large inputs, producing NaN gradients
**Describe the bug**

`activation::softplus` computes `exp` before `log`:

https://github.com/tracel-ai/burn/blob/main/crates/burn-tensor/src/tensor/activation/base.rs#L229-L232

```rust
let tensor = (tensor.mul_scalar(beta).exp() + 1).log();
tensor.div_scalar(beta)
```

`exp` overflows above ~88.7 in `f32`, so `log(inf) = inf`. Because `beta` scales the input, a `beta` of 5 overflows from `x ≈ 18` upwards — well inside a normal activation range.

The forward `inf` is the visible symptom, but the backward pass is the more serious one: the saturated element yields a **`NaN` gradient**, which then propagates into the shared parameter gradients and corrupts the whole update step, not just that element.

`SoftMarginLoss` inherits this. It is defined as `softplus(-target * logit)`, and a confidently *wrong* prediction is exactly the case that saturates — so the loss returns `inf` and the backward pass returns `NaN` precisely when the model is most wrong.

(`mish` also calls `softplus`, but happens to survive: `tanh(inf) = 1`, so `mish(x) -> x`, which is the correct limit.)

**To Reproduce**

Measured on the `ndarray` backend, `f32`:

```rust
// forward
softplus([20, 50, 100], beta = 1.0)   ->  [20.0, 50.0, inf]      // expected [20.0, 50.0, 100.0]
softplus([20, 50, 100], beta = 5.0)   ->  [inf,  inf,  inf]      // expected [20.0, 50.0, 100.0]

// SoftMarginLoss, logits [-100, -20, -5] with targets [1, 1, 1]
->  [inf, 20.0, 5.0067153]                                        // expected [100.0, 20.0, 5.0067153]

// backward
softplus([-100, 0, 100], beta = 1.0).sum().backward()
->  grad [3.8e-44, 2.06e-9, 0.5, 1.0, 1.0, NaN]                   // expected sigmoid(x), no NaN
```

There is a second, quieter issue in the same expression. The `1 + exp(beta * x)` intermediate rounds to exactly `1.0` in `f32` once `beta * x < ~-16`, so the negative tail underflows to zero:

```rust
softplus([-20, -50], beta = 1.0)  ->  [0.0, 0.0]   // expected [2.06e-9, 1.93e-22]
```

**Expected behavior**

Softplus should saturate to the identity for large `beta * x` and to `exp(beta * x) / beta` for very negative `beta * x`, with a finite `sigmoid(beta * x)` gradient throughout. PyTorch guards this with a `threshold` parameter (default `20`), above which it returns the input linearly; `burn`'s `softplus` has no such parameter.

Existing coverage in `crates/burn-backend-tests/tests/tensor/float/activation/softplus.rs` only tests values in `[-1, 1]`, and there is no autodiff test for `softplus` at all, which is why neither the `inf` nor the `NaN` was caught.

Notably, `burn` already handles this exact class of problem for `log_sigmoid`, using the log-sum-exp trick with an explanatory comment:

https://github.com/tracel-ai/burn/blob/main/crates/burn-backend/src/backend/ops/activation.rs#L201-L217

So this looks like an oversight in `softplus` rather than a deliberate tradeoff.

**Additional context**

- Backend: `ndarray` (the bug is in the backend-agnostic `burn-tensor` composition, so it affects all backends)
- Dtype: `f32`
- OS: macOS

Mikyx-1:
### Checklist

> `cargo run-checks` was not run in full: this machine ran out of disk space building the
> workspace (`polars`, `torch-sys`). `fmt`, `clippy` and the test suites for every crate touched
> were run against the `ndarray` backend — see **Testing** below. Please let CI confirm the
> remaining backends.

### Related Issues/PRs

Fixes #5373.

### Changes

`activation::softplus` computed `exp` before `log`:

```rust
let tensor = (tensor.mul_scalar(beta).exp() + 1).log();
tensor.div_scalar(beta)
```

`exp` overflows above ~88.7 in `f32`, so `log(inf) = inf`. Because `beta` scales the input, a
`beta` of 5 overflowed from `x ≈ 18` upwards — well inside a normal activation range.

The forward `inf` is the visible symptom, but the backward pass is the more serious one: the
saturated element produced a **`NaN` gradient**, which propagates into the shared parameter
gradients and corrupts the whole update step, not just that element.

`SoftMarginLoss` inherited this. It is `softplus(-target * logit)`, and a confidently *wrong*
prediction is exactly the case that saturates — so it returned `inf`/`NaN` precisely when the
model was most wrong. (`mish` also calls `softplus` but happened to survive, since
`tanh(inf) = 1` gives `mish(x) -> x`, the correct limit.)

**The fix.** Above a `threshold` on `beta * x`, the result is taken to be the identity,
matching `torch.nn.functional.softplus`. Two details matter beyond the plain `where`:

- **The input to `exp` is capped.** Both branches are evaluated regardless of the mask, so
  without the cap the discarded branch still reaches `inf` and contributes `inf * 0 = NaN`
  to the backward pass — reintroducing the exact bug being fixed. The same mask that selects
  the final result is reused to fill those elements before the `exp`.
- **`log1p` replaces `log(1 + ..)`.** The `1 + exp(beta * x)` intermediate rounds to exactly
  `1.0` in `f32` once `beta * x < ~-16`, so the negative tail underflowed to zero. This
  matches PyTorch's own kernel, which also uses `log1p`.

A `threshold` larger than the dtype can evaluate is lowered to the largest one that dtype
supports — `ln(f16::MAX)` is only ~11.09, below the default of 20, so on `f16` the requested
threshold alone would not keep `exp` in range. Substituting the identity earlier is safe:
the two already agree to within the dtype's precision at that magnitude.

The gradient at `x = 0` is exactly `0.5`. Worth noting for reviewers, since the textbook
`max(z, 0) + log1p(exp(-|z|))` formulation is a trap here: it is stable, but `max`/`abs` have
ambiguous subgradients at exactly `z = 0`, returning `0` or `1` instead of `0.5` — and `x = 0`
is common in practice (zero-init bias, masked positions). With the threshold at 20 the cap
is inactive near the origin, so no such ambiguity arises.

### API

Not a breaking change. `softplus(tensor, beta)` keeps its signature and is now always safe —
it delegates to a new `softplus_with_threshold(tensor, beta, threshold)` with the default of
`20.0`, exposed as `activation::DEFAULT_SOFTPLUS_THRESHOLD`. Callers who want a different
threshold reach for the explicit variant; nobody else has to know overflow thresholds exist.

`SoftplusConfig` gains a `threshold` field with `#[config(default = "20.0")]`, so the
`Softplus` module and any serialized config are unaffected — old configs still deserialize,
with the default filled in. `ModuleDisplay` deliberately still prints only `beta`, leaving
display output unchanged.

### Testing

The gap that let this through was the test suite itself: existing coverage only exercised
values in `[-1, 1]`, and there was **no autodiff test for `softplus` at all**. Added:

- `tests/tensor/float/activation/softplus.rs` — overflow, underflow, and threshold behaviour
- `tests/autodiff/softplus.rs` (new file) — gradient is `sigmoid(beta * x)`, including the
  saturated case that previously returned `NaN`, and `0.5` at the origin
- `crates/burn-nn/src/loss/soft_margin.rs` — a confidently-wrong-prediction case

Before → after, on `ndarray`/`f32`:

| input | before | after |
|---|---|---|
| `softplus([20, 50, 100], beta=1)` | `[20.0, 50.0, inf]` | `[20.0, 50.0, 100.0]` |
| `softplus([20, 50, 100], beta=5)` | `[inf, inf, inf]` | `[20.0, 50.0, 100.0]` |
| `softplus([-20, -50], beta=1)` | `[0.0, 0.0]` | `[2.06e-9, 1.93e-22]` |
| `SoftMarginLoss([-100, -20, -5], t=1)` | `[inf, 20.0, 5.0067153]` | `[100.0, 20.0, 5.0067153]` |
| `grad softplus([-100, 0, 100])` | `[3.8e-44, 0.5, NaN]` | `[0.0, 0.5, 1.0]` |

Suites run against `ndarray`, all passing:

```
cargo test -p burn-backend-tests --features ndarray --test tensor     # 1783 passed
cargo test -p burn-backend-tests --features ndarray --test autodiff   #  532 passed
cargo test -p burn-nn                                                 #  455 passed
cargo test -p burn-train                                              #  209 passed
```

`cargo fmt --check` and `cargo clippy --all-targets` are clean for `burn-tensor`, `burn-nn` and
`burn-train`. Backends other than `ndarray` have not been run locally; the new tests live in the
shared harness, so CI should cover them.

### DIFF
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -4,7 +4,7 @@ use burn_dispatch::Dispatch;
 use crate::check::TensorCheck;
 use crate::check::unwrap_dim_index;
 use crate::ops::BridgeTensor;
-use crate::{AsIndex, Tensor, check, s};
+use crate::{AsIndex, DType, Tensor, check, s};
 
 /// Applies the rectified linear unit function element-wise
 /// as described in the paper [Deep Learning using Rectified Linear Units (ReLU)](https://arxiv.org/pdf/1803.08375).
@@ -213,6 +213,20 @@ pub fn softmin<const D: usize>(tensor: Tensor<D>, dim: impl AsIndex) -> Tensor<D
     Tensor::new(softmin_impl(tensor.primitive, dim))
 }
 
+/// Largest argument for which `exp` is still finite in the given float dtype, rounded down.
+///
+/// `f32`, `flex32` and `bf16` all carry `f32`'s exponent range, so they share a bound.
+fn max_finite_exp_arg(dtype: DType) -> f64 {
+    match dtype {
+        // ln(f16::MAX) = ln(65504) ~= 11.09
+        DType::F16 => 11.0,
+        // ln(f64::MAX) ~= 709.78
+        DType::F64 => 709.0,
+        // ln(f32::MAX) ~= 88.72
+        _ => 88.0,
+    }
+}
+
 /// Applies the SoftPlus function element-wise.
 ///
 #[cfg_attr(
@@ -232,10 +246,16 @@ $$
 ///   rather than evaluated. This is what keeps `exp` from overflowing: the naive form returns
 ///   `inf` in the forward pass and `NaN` in the backward pass from `beta * x > ~88` upwards in
 ///   `f32`. Pass `20.0` unless you have a reason not to, matching the default of
-///   `torch.nn.functional.softplus`. Values in `[20, 80]` are all equivalent in exact
-///   arithmetic; a `threshold` above `~88` cannot prevent the overflow it exists to prevent,
-///   and one below `~20` starts to visibly round the curve near the origin.
+///   `torch.nn.functional.softplus`. Values below `~20` start to visibly round the curve near
+///   the origin. A `threshold` too large for the dtype to evaluate is lowered to the largest
+///   one that dtype supports, since the two are indistinguishable at that magnitude anyway.
 pub fn softplus<const D: usize>(tensor: Tensor<D>, beta: f64, threshold: f64) -> Tensor<D> {
+    // `exp` saturates once its argument passes `ln(MAX)` for the dtype, which is only ~11.09
+    // in `f16` — far below the `f32`-oriented default of 20, so the requested threshold alone
+    // does not always keep `exp` in range. Substituting the identity earlier is safe: softplus
+    // and the identity already agree to within the dtype's precision at that magnitude.
+    let threshold = threshold.min(max_finite_exp_arg(tensor.dtype()));
+
     let scaled = tensor.clone().mul_scalar(beta);
 
     // The saturated elements are masked out below, but they are still evaluated here, so the
--- a/crates/burn-tensor/src/tensor/activation/base.rs
+++ b/crates/burn-tensor/src/tensor/activation/base.rs
@@ -218,16 +218,18 @@ pub fn softmin<const D: usize>(tensor: Tensor<D>, dim: impl AsIndex) -> Tensor<D
 pub const DEFAULT_SOFTPLUS_THRESHOLD: f64 = 20.0;
 
 /// Largest argument for which `exp` is still finite in the given float dtype, rounded down.
-///
-/// `f32`, `flex32` and `bf16` all carry `f32`'s exponent range, so they share a bound.
 fn max_finite_exp_arg(dtype: DType) -> f64 {
     match dtype {
+        // Flex32 computes with f16-like limits; see `FloatDType::finfo`.
         // ln(f16::MAX) = ln(65504) ~= 11.09
-        DType::F16 => 11.0,
+        DType::F16 | DType::Flex32 => 11.0,
         // ln(f64::MAX) ~= 709.78
         DType::F64 => 709.0,
-        // ln(f32::MAX) ~= 88.72
-        _ => 88.0,
+        // ln(f32::MAX) ~= 88.72; bf16 has the same exponent range.
+        DType::F32 | DType::BF16 => 88.0,
+        // Quantized operations are evaluated through a floating-point representation. Use the
+        // most conservative floating-point bound; other kinds can't reach this float API.
+        _ => 11.0,
     }
 }
 
@@ -245,8 +247,8 @@ $$
 ///
 /// The SoftPlus function is a smooth approximation of the ReLU function.
 ///
-/// For values where `beta * x > 20`, returns `x` directly to avoid overflow when evaluating
-/// the exponential. Use [`softplus_with_threshold`] to pick a different threshold.
+/// Uses a default threshold of `20.0` for numerical stability. Use
+/// [`softplus_with_threshold`] to pick a different threshold.
 ///
 /// # Arguments
 ///
@@ -260,8 +262,8 @@ pub fn softplus<const D: usize>(tensor: Tensor<D>, beta: f64) -> Tensor<D> {
 /// See [`softplus`] for the function itself, which uses the default threshold of `20.0`.
 ///
 /// For values where `beta * x > threshold`, returns `x` directly to avoid overflow when
-/// evaluating the exponential. A `threshold` too large for the dtype to evaluate is lowered
-/// to the largest one that dtype supports.
+/// evaluating the exponential. A threshold beyond the dtype's finite `exp` range is lowered to
+/// the largest safe value.
 ///
 /// # Arguments
 ///
@@ -272,10 +274,9 @@ pub fn softplus_with_threshold<const D: usize>(
     beta: f64,
     threshold: f64,
 ) -> Tensor<D> {
-    // `exp` saturates once its argument passes `ln(MAX)` for the dtype, which is only ~11.09
-    // in `f16` — below the `f32`-oriented default of 20, so the requested threshold alone does
-    // not always keep `exp` in range. Substituting the identity earlier is safe: softplus and
-    // the identity already agree to within the dtype's precision at that magnitude.
+    // `exp` overflows at a much lower input in f16 and Flex32 than in the wider formats, so the
+    // requested threshold alone does not always keep it in range. Substituting the identity at
+    // the dtype's limit is safe because Softplus has already rounded to it by that magnitude.
     let threshold = threshold.min(max_finite_exp_arg(tensor.dtype()));
 
     let scaled = tensor.clone().mul_scalar(beta);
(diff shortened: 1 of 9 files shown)


crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -226,9 +240,39 @@ $$
...
+///
+/// # Arguments
+/// - `threshold`: the value of `beta * x` above which the result is taken to be `x` directly
+///   rather than evaluated. This is what keeps `exp` from overflowing: the naive form returns
+///   `inf` in the forward pass and `NaN` in the backward pass from `beta * x > ~88` upwards in
+///   `f32`. Pass `20.0` unless you have a reason not to, matching the default of
+///   `torch.nn.functional.softplus`. Values below `~20` start to visibly round the curve near
+///   the origin. A `threshold` too large for the dtype to evaluate is lowered to the largest
+///   one that dtype supports, since the two are indistinguishable at that magnitude anyway.
+pub fn softplus<const D: usize>(tensor: Tensor<D>, beta: f64, threshold: f64) -> Tensor<D> {
```
laggui: Rust doesn't support default values, but in this case `softplus(tensor, beta)` should always be safe without users having to know about overflow thresholds. We can have this instead:

```rust
pub fn softplus<const D: usize>(tensor: Tensor<D>, beta: f64) -> Tensor<D> {
    softplus_with_threshold(tensor, beta, 20.0)
}
```

Essentially, introduce `softplus_with_threshold` and make 20.0 the default with the current `softplus`. This will also reduce API changes for current consumers of `softplus(tensor, beta)`.

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -226,9 +240,39 @@ $$
...
-    tensor.div_scalar(beta)
+///
+/// # Arguments
+/// - `threshold`: the value of `beta * x` above which the result is taken to be `x` directly
+///   rather than evaluated. This is what keeps `exp` from overflowing: the naive form returns
+///   `inf` in the forward pass and `NaN` in the backward pass from `beta * x > ~88` upwards in
+///   `f32`. Pass `20.0` unless you have a reason not to, matching the default of
+///   `torch.nn.functional.softplus`. Values below `~20` start to visibly round the curve near
+///   the origin. A `threshold` too large for the dtype to evaluate is lowered to the largest
+///   one that dtype supports, since the two are indistinguishable at that magnitude anyway.
```
laggui: This mixes API documentation, implementation details, recommendations and debugging history.

We should keep the documentation concise, e.g.

```rust
  /// For values where `beta * x > threshold`, returns `x` directly to avoid
  /// overflow when evaluating the exponential.
  ///
  /// # Arguments
  ///
  /// - `beta`: Controls the sharpness of the approximation to ReLU.
  /// - `threshold`: The value of `beta * x` above which the linear approximation is used.
```

crates/burn-tensor/src/tensor/activation/base.rs
```diff
@@ -226,9 +240,39 @@ $$
...
+    // `1 + exp(beta * x)` intermediate rounds to exactly `1.0` in `f32` and the result
+    // underflows to zero.
+    let softplus = scaled
+        .clone()
+        .clamp_max(threshold)
+        .exp()
+        .log1p()
+        .div_scalar(beta);
+
+    softplus.mask_where(scaled.greater_elem(threshold), tensor)
```
laggui: The threshold comparison can be reused for both sanitizing the input to exp and selecting the final result instead of using `clamp_max` and `greater_elem` separately.

```rust
let mask = scaled.clone().greater_elem(threshold);
let scaled_safe = scaled.mask_fill(mask.clone(), threshold);

let nonlinear = scaled_safe.exp().log1p().div_scalar(beta);

nonlinear.mask_where(mask, tensor)
```

laggui (changes requested):
Thanks for flagging the issue and addressing it!

Just a couple of comments:

crates/burn-tensor/src/tensor/activation/base.rs
Mikyx-1: Done — `softplus(tensor, beta)` keeps its original signature and delegates to a new `softplus_with_threshold(tensor, beta, 20.0)`, so existing callers are untouched and this is no longer a breaking change. All five in-tree call sites are back to the two-argument form; only `Softplus`/`SoftplusConfig` (which already defaulted `threshold` to 20.0) goes through the explicit variant.

The default is exposed as `activation::DEFAULT_SOFTPLUS_THRESHOLD` so the two crates don't each hardcode `20.0` — happy to inline it if you'd rather not add the const to the public API.

crates/burn-tensor/src/tensor/activation/base.rs
Mikyx-1: Agreed, trimmed to your wording. The two now read:

```rust
/// The SoftPlus function is a smooth approximation of the ReLU function.
///
/// For values where `beta * x > 20`, returns `x` directly to avoid overflow when evaluating
/// the exponential. Use [`softplus_with_threshold`] to pick a different threshold.
///
/// # Arguments
///
/// - `beta`: Controls the sharpness of the approximation to ReLU.
pub fn softplus<const D: usize>(tensor: Tensor<D>, beta: f64) -> Tensor<D>

/// Applies the SoftPlus function element-wise, with an explicit stability threshold.
///
/// See [`softplus`] for the function itself, which uses the default threshold of `20.0`.
///
/// For values where `beta * x > threshold`, returns `x` directly to avoid overflow when
/// evaluating the exponential. A `threshold` too large for the dtype to evaluate is lowered
/// to the largest one that dtype supports.
///
/// # Arguments
///
/// - `beta`: Controls the sharpness of the approximation to ReLU.
/// - `threshold`: The value of `beta * x` above which the linear approximation is used.
pub fn softplus_with_threshold<const D: usize>(tensor: Tensor<D>, beta: f64, threshold: f64) -> Tensor<D>
```

The rationale for the clamp and for `log1p` stayed as inline comments in the body, since both look removable without them.

crates/burn-tensor/src/tensor/activation/base.rs
Mikyx-1: Good call — applied as suggested:

```rust
let scaled = tensor.clone().mul_scalar(beta);
let mask = scaled.clone().greater_elem(threshold);

let nonlinear = scaled
    .mask_fill(mask.clone(), threshold)
    .exp()
    .log1p()
    .div_scalar(beta);

nonlinear.mask_where(mask, tensor)
```

One comparison instead of a comparison plus a `clamp_max`, and the two now can't drift apart. `mask_fill` zeroes the gradient on the filled elements the same way `clamp_max` did, so the saturated-gradient autodiff test still gives `[0.0, 0.5, 1.0]`.

Mikyx-1:
Thanks for the review — all three addressed in 5697149, replies inline.

- `softplus(tensor, beta)` keeps its signature and delegates to a new `softplus_with_threshold`, so this is no longer a breaking change and the five in-tree call sites are back to two arguments.
- Docs trimmed to the API contract; the clamp and `log1p` rationale stayed as inline comments in the body.
- The threshold comparison is computed once and reused for `mask_fill` and `mask_where`.

The branch is also rebased onto current `main`. Re-ran on `ndarray`: tensor 1783, autodiff 532, `burn-nn` 455, `burn-train` 209 — all passing, `fmt`/`clippy` clean.

swfsql:
Recently I have made a similar change to a softplus subset (fixed Beta), and one insight was that dtypes have precision limits that are usually below the overflows. The commit related to this is this one, in case you find it useful: https://github.com/swfsql/burn-mamba/commit/bf288e95bc9020e2d50fbe8501f024b3dbb1cf08
(Note that forward-only could have a lower threshold, but I decided to stick to the backward-aware one; also, bf16 and fp16 should have a lower threshold than `20`).
Edit: AI [chat ctx](https://claude.ai/share/2203eff6-d2f9-435f-bbec-93fdf9f93072) about forward-only.

laggui (approved):
LGTM! Thanks

Just added the missing threshold to the module display, and fixed the flex32 limit.

PR #5376, opened 2026-08-17
fix(burn-tch): preserve parent storage through permute/swap_dims

### LINKED ISSUE #5375: burn-tch: permute/swap_dims/flip mint a fresh storage Arc for a shared buffer, defeating can_mut — libtorch panics on in-place logical_and_
**Describe the bug**

burn-tch tracks in-place mutability with a per-buffer Arc (`Storage::can_mut`, `src/tensor.rs`). View-producing ops are supposed to inherit the parent's storage handle via `TchTensor::from_existing` so the refcount reflects the sharing — but `permute`, `swap_dims` and `flip` call `TchTensor::new` instead (`src/ops/base.rs:652-668` in burn-tch 0.21.0), minting a fresh Arc for a buffer still shared with the parent.

`can_mut()` then wrongly approves in-place execution on the shared buffer. A binary op between a tensor and a view of itself takes libtorch's in-place path (`logical_and_`) over overlapping memory, and libtorch's overlap assert kills it:

```
called `Result::unwrap()` on an `Err` value: Torch("unsupported operation: some
elements of the input tensor and the written-to tensor refer to a single memory
location. Please clone() the tensor before performing the operation.
Exception raised from assert_no_partial_overlap at ATen/MemoryOverlap.cpp:97 ...
at::_ops::logical_and_::call(...)
```

**To Reproduce**

20-line runnable crate: https://github.com/andoriyu/repros/tree/master/burn-tch-permute-alias

```rust
let mask: Tensor<B, 3, Bool> =
    Tensor::<B, 3>::random([1, 72, 72], Distribution::Uniform(0.0, 1.0), &device)
        .greater_elem(0.5);
let permuted = mask.clone().permute([0, 2, 1]);
let out = mask.bool_and(permuted); // panics here
```

`mask & mask^T` is the attention-mask pattern burn-import generates for a FastConformer encoder, which is how this was found: the generated model aborted deterministically at the first chunk.

**Expected behavior**

The AND computed out of place (as on every other backend), since the operand is a view sharing the output's buffer.

 - OS: NixOS (Linux 6.18, x86_64)
 - Burn version: 0.21.0 (burn-tch 0.21.0)
 - Other: tch 0.22.0, libtorch 2.9.0 (CPU build)

andoriyu:
### Related Issues/PRs

Fixes #5375

### Changes

`permute` and `swap_dims` return views over the parent's buffer, but built the child with `TchTensor::new`, which mints a fresh storage `Arc` for memory that is still shared. `Storage::can_mut` counts references on that `Arc`, so it reported an aliased buffer as exclusively owned.

A binary op between a tensor and a view of itself then took libtorch's in-place path over overlapping memory and tripped libtorch's own overlap assert:

```
unsupported operation: some elements of the input tensor and the written-to
tensor refer to a single memory location. Please clone() the tensor before
performing the operation.
(assert_no_partial_overlap at ATen/MemoryOverlap.cpp:97)
```

The fix captures the parent's storage before the shadowing binding and builds the child with `TchTensor::from_existing`, which compares data pointers and shares the handle when the buffer is aliased:

```rust
pub fn permute(tensor: TchTensor, axes: &[usize]) -> TchTensor {
    let storage = tensor.storage.clone();
    let tensor = tensor
        .tensor
        .permute(axes.iter().map(|x| *x as i64).collect::<Vec<_>>());
    TchTensor::from_existing(tensor, storage)
}
```

**`flip` is deliberately not changed**, although the issue names it. Unlike NumPy's, libtorch's `flip` allocates rather than returning a view, so the child legitimately owns its buffer and `can_mut() == true` is correct there. I found this while writing the regression test — an early version asserted the aliasing invariant for all three ops and failed on `flip`; the assertion was wrong, not the code. The test now asserts the opposite for `flip` so the distinction is pinned rather than rediscovered later.

### Tests

Two regression tests in `crates/burn-tch/src/tensor.rs`, both failing before the change:

- `view_ops_preserve_parent_storage` — `can_mut()` must be false for a live `permute`/`swap_dims` view, and true for `flip`.
- `bool_and_over_permuted_alias_does_not_panic` — reproduces the reported crash directly via `x & xᵀ`, the attention-mask pattern `burn-import` generates.

### Validation beyond the unit tests

The workload that surfaced this is a `burn-import`-generated FastConformer streaming ASR encoder, whose generated code contains `tile1_out1.bool_and(tile1_out1.permute([0, 2, 1]))`. We had been carrying a hand patch on the generated file (a `Bool -> Int -> Bool` round trip to break the alias). With this change **the hand patch is removed** and the model runs unmodified:

| host | lane | result |
| --- | --- | --- |
| EPYC 7302P | `tch-cpu` f32 | 30/30 exact, 0 token edits vs onnxruntime |
| Ryzen 9 7950X3D | `tch-cpu` f32 | 30/30 exact, 0 token edits |
| RX 7900 XTX | `tch-rocm` f32 | 30/30 exact, 0 token edits |

"30/30 exact" is token-for-token equality against an onnxruntime reference implementation over a 30-clip corpus, so the change is exercised on both the CPU and HIP backends with no numerical difference.

### Checklist

- [x] Confirmed that `cargo run-checks` command has been executed. (passes, 05:32, no failures)
- [x] Made sure the book is up to date with changes in this PR. (no user-facing behaviour change; nothing in the book describes this internal invariant)

### Note

`crates/burn-tch`'s own unit tests were run against libtorch 2.11 with `LIBTORCH_BYPASS_VERSION_CHECK=1`, since that is the toolchain I have built locally; the aliasing logic is version-independent. On that setup the pre-existing `should_support_dtypes` test fails on `main` as well (it asserts I32 support, which differs between 2.9 and 2.11) — unrelated to this change, and it is unaffected by it.

### DIFF
--- a/crates/burn-tch/src/ops/base.rs
+++ b/crates/burn-tch/src/ops/base.rs
@@ -650,15 +650,22 @@ impl TchOps {
     }
 
     pub fn swap_dims(tensor: TchTensor, dim1: usize, dim2: usize) -> TchTensor {
+        // `transpose` returns a view sharing the parent's buffer, so the child
+        // must inherit the parent's storage handle. `TchTensor::new` would mint a
+        // fresh `Arc` for still-shared memory and `can_mut()` would then approve
+        // an in-place op that writes over the parent.
+        let storage = tensor.storage.clone();
         let tensor = tensor.tensor.transpose(dim1 as i64, dim2 as i64);
-        TchTensor::new(tensor)
+        TchTensor::from_existing(tensor, storage)
     }
 
     pub fn permute(tensor: TchTensor, axes: &[usize]) -> TchTensor {
+        // A view over the parent's buffer — see `swap_dims`.
+        let storage = tensor.storage.clone();
         let tensor = tensor
             .tensor
             .permute(axes.iter().map(|x| *x as i64).collect::<Vec<_>>());
-        TchTensor::new(tensor)
+        TchTensor::from_existing(tensor, storage)
     }
 
     pub fn flip(tensor: TchTensor, axes: &[usize]) -> TchTensor {
--- a/crates/burn-tch/src/tensor.rs
+++ b/crates/burn-tch/src/tensor.rs
@@ -453,6 +453,7 @@ fn f_copy_data<T: TchElement>(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::ops::TchOps;
     use burn_backend::ops::FloatTensorOps;
     use burn_backend::{Backend, quantization::QuantScheme, read_sync};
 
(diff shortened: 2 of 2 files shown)


laggui (approved):
Thanks for flagging and addressing the issue!

PR #5377, opened 2026-08-17
feat(capture): add graph capture backend producing `GraphIr`

laggui:
### Checklist

### Related Issues/PRs

### Changes

- add `burn-capture` for non-executing operation graph capture
- consolidate the previous `burn-router` graph representation into `burn_ir::GraphIr` 
- expose capture through the public Burn device and backend APIs
- update router and remote graph caching to consume the shared `GraphIr`

### Testing

Unit tests


laggui:
@antimora I addressed the capture lifecycle and router concerns flagged by copilot.

The session-aware registration comment was valid, and I refactored things. I think it's much cleaner now too. Capture devices are now reusable, while each `capture_scope` installs an isolated client/session through a registration guard:

```rust
let device = Device::capture();

let graph = device.capture_scope(|scope| {
    let output = forward(input);
    scope.complete(input_ids, output_ids)
})?;
```

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+        // Bind this reusable device to the new session client for exactly this scope. BackendRouter
+        // operations call `get_client`, which now finds this client instead of trying the channel's
+        // intentionally unsupported unscoped initialization path.
+        let registration = register_scoped_client::<CaptureChannel>(self, client.clone())
+            .ok_or(CaptureError::AlreadyActive)?;
+        let guard = CaptureScopeGuard {
+            client,
+            _registration: registration,
+        };
+        guard.complete(capture(CaptureScope { _private: () }))
```
antimora: The closure runs before this line, and `CaptureScope::complete` only collects two Vecs (it holds no handle on the session at all), so the guard takes `state.operations` after the fact.

Two consequences. Declare the boundary first, keep recording, return the token, and you get a graph with three ops and an empty boundary plus an `Ok`. And a tensor that outlives the scope keeps its client clone, so ops on it land in the drained session and vanish silently.

A `closed` flag on `CaptureState`, set here and checked in `register_op`/`create_empty_handle`/`register_tensor_data`, covers both. Keep `Drop` exempt or `late_tensor_drops_stay_in_the_completed_session` breaks.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+            .map(|id| state.resolve_alias(id))
+            .collect();
+        let known: HashSet<_> = state
+            .operations
+            .iter()
+            .flat_map(OperationIr::nodes)
+            .map(|tensor| tensor.id)
+            .chain(state.values.keys().copied())
+            .collect();
+        for (boundary, ids) in [("input", &inputs), ("output", &outputs)] {
```
antimora, line 198: `known` is every tensor mentioned anywhere plus the `values` keys, so this catches duplicates and ids that appear nowhere, but never the converse: a tensor that's read and never produced can be missing from both boundaries and from `values` and still return `Ok`. Easy to hit by mixing in a tensor from a previous scope, and the consumer then sees a dangling read.

`GraphIr::new` already computes produced-vs-referenced, so reusing it here is cheap. Same spot could also reject a `ReadWrite`-consumed tensor declared as an output, which `GraphIr::new` deliberately excludes from `outputs` but this accepts.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+        self.state().lock().values.insert(id, data);
+        RouterTensor::new(id, shape, dtype, self.clone())
+    }
+
+    fn device(&self) -> Self::Device {
+        self.device
+    }
+    fn seed(&self, _seed: u64) {}
+    fn dtype_usage(&self, _dtype: DType) -> DTypeUsageSet {
+        DTypeUsage::general()
```
antimora: `_dtype` is ignored, so `supports_dtype` (which is `dtype_usage(..).is_superset(general())`) answers yes for F64, BF16 and QFloat on a capture device.

That's the guard `Device::supports_dtype` docs tell people to check before selecting a reduced precision, so model code that follows the advice picks the most aggressive dtype under capture and bakes it into the exported graph. `Accelerated` also never gets reported, so acceleration checks silently answer false.

antimora (commented):
Went through this properly and re-checked all of Copilot's threads against current HEAD. Most are genuinely fixed. The registration guard is nicely done, and the panic-unwinding test plus the rebinding tests are the right ones to have.

Two things I'd fix before merge, and they're one root cause: **nothing ever closes the session.** `complete` just collects two Vecs and the guard harvests the ops afterwards. So you can declare the boundary, keep recording, and get back a graph full of ops with an empty boundary. Same reason ops on a tensor that outlived its scope disappear into the drained session. A `closed` flag on `CaptureState`, set in `complete` and checked in `register_op`, kills both (keep the `Drop` exemption so the existing tests stay green).

Two more that I can't put inline:

- `capture` isn't in any aggregate feature set, so **8 of the 21 new tests never run in CI**. The dispatch, burn-tensor and burn-level modules all compile out; only burn-capture's own unit tests run. That's also most of the missing patch coverage, since the whole `Capture` arm of the dispatch macros looks like dead code to codecov.
- `burn-capture` has no job in `publish.yml` and `publish-burn-dispatch` doesn't list it in `needs`. burn-dispatch declares it optional at `=0.22.0-pre.2`, so `cargo publish -p burn-dispatch` will fail at the next release cut.

Checked locally at 167325174: 15/15 burn-capture tests pass, clippy clean on burn-capture/burn-ir/burn-router, `cargo check -p burn --features capture,ir --no-default-features` passes. `cargo test -p burn-capture --no-default-features` doesn't compile (two tests use `std::panic::catch_unwind` in a no_std crate) but the lib is fine and the crate isn't in `NO_STD_CRATES`, so CI won't hit it today.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+        let shape = data.shape.clone();
+        let dtype = data.dtype;
+        self.state().lock().values.insert(id, data);
+        RouterTensor::new(id, shape, dtype, self.clone())
+    }
+
+    fn device(&self) -> Self::Device {
+        self.device
+    }
+    fn seed(&self, _seed: u64) {}
```
antimora: Silently dropped, and there's no `Seed` variant in burn-ir so it can't be recorded either. `Device::seed` documents that "at least this device will be seeded", which isn't true here.

A captured graph with dropout or random ops replays with different values and nothing indicates the seed was lost. burn-remote forwards it to the server. Panicking would at least be honest about it.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+    }
+}
+
+/// Final result of a capture scope.
+#[derive(Clone, Debug)]
+pub struct CapturedGraph {
+    /// Ordered operation graph with caller-declared boundaries.
+    pub graph: GraphIr,
+    /// Concrete values supplied through tensor initialization, keyed by tensor ID.
+    pub values: HashMap<TensorId, TensorData>,
```
antimora: `hashbrown::HashMap` in a type that's public all the way out to `burn::tensor::CapturedGraph`, and hashbrown isn't re-exported from burn, burn-tensor, burn-ir or burn-std, so a downstream consumer can't name the type of this field without adding their own version-matched hashbrown.

`BTreeMap` would fix that and give exporters a deterministic initializer order (`TensorId` is `Ord`), which matters if anyone golden-files a captured graph.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+            aliases: HashMap::new(),
+        }
+    }
+}
+
+impl CaptureState {
+    /// Resolve a fusion alias to the tensor ID that owns the computation or initializer.
+    fn resolve_alias(&self, mut id: TensorId) -> TensorId {
+        while let Some(source) = self.aliases.get(&id) {
+            id = *source;
```
antimora, line 433: Unbounded walk with no cycle guard. `register_alias(id, id)` stores `id -> id` and this spins forever while holding the spin lock, so it wedges the session rather than panicking. Same loop in `AliasVisitor::visit_tensor_mut`.

Not reachable from `alias_handle` today since that always mints a fresh id, but `register_alias` is a public trait method. `if source != new_id { .. }` in `register_alias` closes the whole class.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,841 @@
...
+        assert_eq!(captured.graph.outputs, [output_id]);
+        assert_eq!(captured.graph.operations.len(), 1);
+        assert_eq!(
+            captured.values.get(&input_id),
+            Some(&TensorData::from([1.0f32, 2.0]))
+        );
+    }
+
+    #[test]
+    fn late_tensor_drops_stay_in_the_completed_session() {
```
antimora: This test can't fail. `register_op` discards every `OperationIr::Drop` unconditionally, so the assert holds regardless of which session the drop routed to, which is the thing the name says it's checking.

Swapping the `drop()` for a real op on the stale tensor (say `float_add(stale, fresh)`) makes it catch the actual bug.

crates/burn-router/src/fusion.rs
```diff
@@ -459,8 +415,9 @@ fn estimate_saved_pct(ops: &[OperationIr]) -> u64 {
         }
     }
 
-    let (inputs, outputs) = classify_boundary(ops);
-    let bindings = (inputs.len() + outputs.len()) as u64 * PAIR + dims.len() as u64 * PER_DIM;
+    let graph = GraphIr::new(ops.to_vec());
```
antimora: `classify_boundary` took a slice; this clones the whole op list. `score()` calls it from `fuse()` on every appended operation, so building an n-op block is O(n^2) deep clones. `OperationIr` is 616 bytes and `CustomOpIr` carries a String plus Vecs, so a 512-op block is roughly 131k clones just to count boundary sizes.

A borrowing classifier that `GraphIr::new` also calls would keep both callers happy.

crates/burn-router/src/fusion.rs
```diff
@@ -385,55 +387,9 @@ impl<R: RouterChannel> RouterGraphExecution<R> {
...
-    let outputs = compute_produced
-        .iter()
-        .filter(|id| !consumed.contains(id))
-        .copied()
-        .collect();
-    (inputs, outputs)
-}
-
-/// Estimate the % of serialized bytes that caching this op-graph saves per replay.
+// Estimate the % of serialized bytes that caching this op-graph saves per replay.
```
antimora: Lost a slash here, so the summary line isn't part of `estimate_saved_pct`'s docs anymore and the rendered block opens on the blank `///`. The blank line before it went too. Line 392 also runs long now.

crates/burn-router/src/fusion.rs
```diff
@@ -369,12 +369,14 @@ pub struct RouterGraphExecution<R: RouterChannel> {
...
-    graph: Vec<OperationIr>,
+    graph: GraphIr,
 }
 
 impl<R: RouterChannel> RouterGraphExecution<R> {
     fn new(graph: Vec<OperationIr>, device: R::Device) -> Self {
-        let (input_ids, output_ids) = classify_boundary(&graph);
+        let graph = GraphIr::new(graph);
+        let input_ids = graph.inputs.clone();
+        let output_ids = graph.outputs.clone();
```
antimora: These two are verbatim clones of `graph.inputs`/`graph.outputs` now and never mutated, so it's two sources of truth for one fact. Reading `self.graph.inputs`/`.outputs` at the use sites lets both fields and both clones go.

Related: `RouterGraphExecutionState` serializes the boundary too, and `from_state` discards it and recomputes from `state.graph.operations`, so a cache built to save bytes is paying for dead ones.

crates/burn-dispatch/src/ops/tensor.rs
```diff
@@ -740,3 +740,39 @@ impl FloatTensorOps<Self> for Dispatch {
...
 }
+
+#[cfg(all(test, feature = "capture", any(feature = "flex", default_backend)))]
+mod tests {
+    use super::*;
+    use burn_backend::ops::IntTensorOps;
+    use burn_capture::CaptureDevice;
+
+    #[test]
+    fn capture_tensor_movement_is_one_way() {
```
antimora, line 751: Both `float_to_device` calls target the same device, and that short-circuits on device equality, so the new `(Capture, Capture)` arm never actually runs its body. Bool movement isn't covered either.

Using two capture devices would exercise it, and that's also the only way to reach `CaptureBridge` and `CaptureChannel::register_tensor` at all. Worth doing: a computed tensor there panics from `get_tensor_handle` rather than erroring, and an initialized one silently copies into the target session with no `Init` op, unlike `float_from_data`.

crates/burn/Cargo.toml
```diff
@@ -131,6 +131,7 @@ cpu = ["burn-cpu", "burn-core/cpu", "burn-vision?/cpu"]
 
 # Backend extensions
 extension = ["burn-core/extension"]
+capture = ["extension", "ir", "burn-core/capture"]
```
antimora: Does `extension` need to be here? `cargo check -p burn --features capture --no-default-features --all-targets` passes without it, and it pulls in burn-backend-extension and burn-backend for anyone enabling `capture`.

crates/burn-ir/src/graph.rs
```diff
@@ -1,8 +1,94 @@
...
 use burn_backend::Slice;
+use hashbrown::HashSet;
 use serde::{Deserialize, Serialize};
 
+/// An ordered operation graph with an explicitly classified tensor boundary.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct GraphIr {
+    /// Operations in execution order.
+    pub operations: Vec<OperationIr>,
+    /// Tensors read by the graph but not produced by a compute operation.
```
antimora: These field docs describe what `new` computes, but capture fills the struct literally with caller-declared boundaries that don't have to satisfy them, so a consumer can't rely on this on a captured graph.

Bigger version of the same thing: `RouterGraphExecution` puts a *relative* graph in `GraphIr` (positional ids, dims that are indices into a table) while capture puts a concrete one. Same type, no marker, and mixing them type-checks. `burn_router::Graph` already is the relative form, so making it wrap `GraphIr` and return `GraphIr` from binding would separate the two and drop the duplicated boundary fields as a side effect.

crates/burn-router/src/graph.rs
```diff
@@ -108,7 +132,7 @@ struct ReplayVisitor<'a> {
     ranges: &'a [Slice],
 }
 
-impl IrVisitorMut for ReplayVisitor<'_> {
+impl IrVisitorMut for BindingVisitor<'_> {
     fn visit_tensor_mut(&mut self, tensor: &mut TensorIr) {
         tensor.id = *self
```
antimora, line 155: These intermediates get bit 63 set via `alloc_intermediate_id`, and capture pushes bound operations straight into the graph it hands the user, so they end up in `CapturedGraph` and any exporter sees `TensorId`s above 2^63.

The comment on `INTERMEDIATE_ID_BIT` still says they're purely server-internal and never referenced by the client, which isn't true now that there's a non-executing consumer.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,421 @@
...
+            id: device_id.index_id,
+        }
+    }
+
+    fn to_id(&self) -> DeviceId {
+        DeviceId {
+            type_id: u16::MAX,
+            index_id: self.id,
+        }
+    }
```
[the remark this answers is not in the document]
antimora: Fixed. `CAPTURE_DEVICE_TYPE_ID` is `u8::MAX` now, which fits `INTERNAL_ID_MASK`, and there are round-trip tests on both the burn-capture and burn-dispatch side.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,421 @@
...
+            }
+            if let Some(id) = ids.iter().find(|id| !known.contains(*id)) {
+                return Err(CaptureError::UnknownBoundary {
+                    boundary,
+                    tensor: *id,
+                });
+            }
+        }
+        let operations = core::mem::take(&mut state.operations);
+        let values = core::mem::take(&mut state.values);
```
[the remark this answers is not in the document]
antimora: Half fixed, so I'd leave this open. Old ops no longer leak forward: each scope gets its own session and `register_op` filters `Drop`. The inverse is still there though. A tensor that outlives its scope keeps its client clone, so ops on it land in the drained session and vanish, and `complete` still returns `Ok`. Details on capture.rs:66.

Also `late_tensor_drops_stay_in_the_completed_session` can't fail as written, since `register_op` drops every `Drop` regardless of which session it routed to.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,421 @@
...
+        DTypeUsage::general()
+    }
+
+    fn register_and_execute_graph(
+        &self,
+        _graph_id: GraphId,
+        relative_graph: Vec<OperationIr>,
+        _bindings: GraphBindings,
+    ) {
+        self.state.lock().operations.extend(relative_graph);
```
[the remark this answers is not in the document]
antimora: Fixed. `register_and_execute_graph` binds through `graph.bind_operations(bindings)` before recording.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    fn execute_graph(&self, _graph_id: GraphId, _bindings: GraphBindings) {
+        panic!("cached graph replay is not supported during graph capture")
```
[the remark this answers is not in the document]
antimora: Fixed. `execute_graph` is implemented, caches the relative `Graph` and rebinds per invocation. `cached_graph_invocations_are_bound_and_recorded` covers two invocations and checks the intermediates differ between them.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    fn register_alias(&self, new_id: TensorId, src_id: TensorId) {
+        let mut state = self.state.lock();
+        if let Some(value) = state.values.get(&src_id).cloned() {
+            state.values.insert(new_id, value);
+        }
```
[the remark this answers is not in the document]
antimora: Fixed. `register_alias` plus an `AliasVisitor` that rewrites recorded ops, with tests for both the computed and the initialized alias case.

One nit left over: `resolve_alias` walks an unbounded loop, so a self-alias would spin forever under the spin lock. Not reachable from `alias_handle` since it always mints a fresh id.

crates/burn-dispatch/src/macros.rs
```diff
@@ -46,10 +47,11 @@ macro_rules! backend_matrix {
...
             [LibTorch, feature = "tch"] => [[Cpu, feature = "cpu"], [Cuda, feature = "cuda"], [Metal, feature = "metal"], [Rocm, feature = "rocm"], [Vulkan, feature = "vulkan"], [Wgpu, feature = "wgpu"], [WebGpu, feature = "webgpu"], [Flex, any(feature = "flex", default_backend)], [NdArray, feature = "ndarray"], [Remote, feature = "remote"]];
-            [Remote, feature = "remote"] => [[Cpu, feature = "cpu"], [Cuda, feature = "cuda"], [Metal, feature = "metal"], [Rocm, feature = "rocm"], [Vulkan, feature = "vulkan"], [Wgpu, feature = "wgpu"], [WebGpu, feature = "webgpu"], [Flex, any(feature = "flex", default_backend)], [NdArray, feature = "ndarray"], [LibTorch, feature = "tch"]]
+            [Remote, feature = "remote"] => [[Cpu, feature = "cpu"], [Cuda, feature = "cuda"], [Metal, feature = "metal"], [Rocm, feature = "rocm"], [Vulkan, feature = "vulkan"], [Wgpu, feature = "wgpu"], [WebGpu, feature = "webgpu"], [Flex, any(feature = "flex", default_backend)], [NdArray, feature = "ndarray"], [LibTorch, feature = "tch"]];
+            [Capture, feature = "capture"] => [[Flex, any(feature = "flex", default_backend)]]
```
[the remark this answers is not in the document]
antimora: Fixed. Capture isn't in the conversion-pairs matrix, so no generated cross arm can shadow the trailing `(Capture(_), _)` panic.

The test that covers it doesn't quite exercise the new same-device arm though, since both `float_to_device` calls target the same device and that short-circuits. Noted on ops/tensor.rs:751.

crates/burn-capture/src/capture.rs
```diff
@@ -0,0 +1,421 @@
...
+    id: u16,
+}
+
+impl CaptureDevice {
+    /// Start a new isolated capture session.
+    pub fn capture() -> (Self, GraphCapture) {
+        let device = Self::default();
+        let capture = GraphCapture {
+            client: get_client::<CaptureChannel>(&device),
+        };
```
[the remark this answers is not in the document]
antimora: Client leak is fixed, the registration guard removes the entry on drop and there's a test for the panic-unwinding path.

Device ids are still never recycled, so the 65,536 cap stands. That's documented now and fine as a decision, but worth noting `Device::capture()` mints a fresh device on every call and the `Device` API offers no way to reuse one, so a service that captures per request will hit it.

crates/burn-capture/src/capture.rs
laggui: I didn't restrict the dtype support, capture is a recorder so all representable non-quantized dtypes report general support now. `QFloat` is currently rejected because router also doesn't implement the quantized operations. `Accelerated` is not a property of the capture, it performs no acceleration.

crates/burn-dispatch/src/ops/tensor.rs
laggui: Good catch.

I made capture explicitly one-way, so captured tensors cannot move to another device or session, and added a two-device regression test. `CaptureBridge` remains only as a defensive, unsupported path.

PR #5387, opened 2026-08-19
TensorData::{try_}(random,zeros,ones,full)_dtype(dtype).

crutcher:
Flush out TensorData construction mechanics.
* Scalar::dtype().

### Checklist

- [X] Confirmed that `cargo run-checks` has been executed.
- [N/A] Made sure the book is up to date with changes in this PR.

### DIFF
--- a/crates/burn-std/src/data/tensor/base.rs
+++ b/crates/burn-std/src/data/tensor/base.rs
@@ -1,6 +1,7 @@
 use core::f32;
 
 use alloc::format;
+use alloc::vec;
 use alloc::vec::Vec;
 use bytemuck::checked::CheckedCastError;
 use rand::Rng;
@@ -127,6 +128,16 @@ mod shape_inner {
 }
 
 impl TensorData {
+    /// Returns the rank (the number of dimensions).
+    pub fn rank(&self) -> usize {
+        self.shape.rank()
+    }
+
+    /// Returns the total number of elements of the tensor data.
+    pub fn num_elements(&self) -> usize {
+        self.shape.num_elements()
+    }
+
     /// Creates a new tensor data structure.
     pub fn new<E: Element, S: Into<Shape>>(value: Vec<E>, shape: S) -> Self {
         // Ensure shape is valid
@@ -182,7 +193,7 @@ impl TensorData {
 
     // Check that the input vector contains a correct number of elements
     fn check_data_len<E: Element>(data: &[E], shape: &Shape) {
-        let expected_data_len = numel(shape);
+        let expected_data_len = shape.num_elements();
         let num_data = data.len();
         assert_eq!(
             expected_data_len, num_data,
@@ -237,16 +248,6 @@ impl TensorData {
         }
     }
 
-    /// Returns the rank (the number of dimensions).
-    pub fn rank(&self) -> usize {
-        self.shape.len()
-    }
-
-    /// Returns the total number of elements of the tensor data.
-    pub fn num_elements(&self) -> usize {
-        numel(&self.shape)
-    }
-
     /// Populates the data with random values.
     pub fn random<E: Element, R: Rng, S: Into<Shape>>(
         shape: S,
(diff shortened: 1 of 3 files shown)


crates/burn-std/src/data/tensor/base.rs
```diff
@@ -238,79 +248,123 @@ impl TensorData {
...
+        Self::new(data, shape)
     }
 
     /// Populates the data with the given value
-    pub fn full_dtype<E: Into<Scalar>, S: Into<Shape>>(
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing the populated `TensorData` if successful, or a `DataError` if the conversion fails.
+    pub fn try_full_dtype<E: Into<Scalar>, S: Into<Shape>>(
```
laggui, line 361: The fallible APIs seem overstated. The only error is unsupported `QFloat`, and `UnsupportedConversion` is misleading because no conversion occurs. I suggest keeping only the current `_dtype` construction APIs. explicitly documenting that `QFloat` panics.

crates/burn-std/src/data/tensor/base.rs
```diff
@@ -324,25 +378,36 @@ impl TensorData {
...
+        dtype: DType,
+    ) -> Self {
+        Self::try_full_dtype(shape, fill_value, dtype)
+            .unwrap_or_else(|err| panic!("Failed to create tensor data: {}", err))
     }
 
     // Unchecked, used to overwrite the dtype
-    pub(super) fn into_bool_u32(mut self) -> Self {
-        self.dtype = DType::Bool(BoolStore::U32);
+    pub(super) fn unchecked_cast(mut self, dtype: DType) -> Self {
```
laggui, line 409: `unchecked_cast` implies a value or representation conversion, but the method only replaces metadata. Can we rename this to something like `with_dtype_unchecked`?

crates/burn-std/src/data/tensor/base.rs
```diff
@@ -939,4 +1008,174 @@ mod tests {
...
+        check::<i16>();
+        check::<i8>();
+
+        check::<u64>();
+        check::<u32>();
+        check::<u16>();
+        check::<u8>();
+
+        check::<bool>();
+    }
```
laggui, line 1084: The current `check::<E>(dtype)` calls inside one test obscure which dtype failed and stop at the first failure. We should keep reusable helpers, but we can generate the individual tests with a macro e.g.

```rust
fn check_zeros_dtype<E: Element + PartialEq>(dtype: DType) {
    let data = TensorData::zeros_dtype([2, 2], dtype);
    assert_eq!(data.dtype, dtype);
    assert_eq!(data.shape.dims(), [2, 2]);
    assert_eq!(data.try_into_vec::<E>().unwrap(), vec![0.elem::<E>(); 4]);
}

macro_rules! test_zeros_dtype {
    ($name:ident, $element:ty, $dtype:expr) => {
        #[test]
        fn $name() {
            check_zeros_dtype::<$element>($dtype);
        }
    };
}

test_zeros_dtype!(zeros_dtype_f32, f32, DType::F32);
test_zeros_dtype!(zeros_dtype_i32, i32, DType::I32);
...
```

crates/burn-std/src/data/tensor/base.rs
```diff
@@ -939,4 +1008,174 @@ mod tests {
...
+            );
+        }
+    }
+
+    #[test]
+    fn test_zeros() {
+        fn check<E: Element + PartialEq>() {
+            let data = TensorData::zeros::<E, _>(&[2, 2]);
+            assert_eq!(data.shape.dims(), [2, 2]);
+            assert_eq!(&data.try_into_vec::<E>().unwrap(), &vec![0.elem::<E>(); 4]);
```
laggui, line 1067: The test should validate the requested dtype, not just shape
```rust
assert_eq!(data.dtype, requested_dtype);
```

though this might raise an existing failure mode with flex32 construction, which currently constructs f32 elements. For this particular dtype, we can leave the test as ignore for now and address this pre-existing issue in another PR.

laggui (changes requested):
I don't think we need the fallible APIs given the contract, otherwise just some minor comments.

PR #5392, opened 2026-08-19
fix(store): respect PyTorch tensor strides

### LINKED ISSUE #5383: burn-store PytorchReader ignores tensor strides (non-contiguous .pth loads scrambled)
**Describe the bug**
PytorchReader parses the stride tuple (args[3] in rebuild_tensor_v2) but
discards it, reading every tensor's storage linearly. A .pth whose weights were
saved non-contiguous therefore loads silently scrambled — the model runs but
produces garbage.

**To Reproduce**
1. Save a conv weight tensor non-contiguous (channels-last: shape [o,i,k,k],
   strides (27,1,9,3)) into a .pth.
2. Load it with PytorchStore.
3. Compare against the original: the loaded head weights have correlation 0.24
   (scrambled).

**Expected behavior**
Capture the stride and scatter the storage into the logical layout (or
.contiguous() the data before returning), matching torch's semantics.

 - OS: Fedora Linux
 - burn: 0.21.0 (burn-store)

**Additional context**
Workaround: pre-process the state dict with
{k: v.contiguous() for k, v in sd.items()}.

original4422:
### Checklist

- [x] Made sure the book is up to date with changes in this PR (no public API or book change; corrupt-checkpoint behavior is called out below).

### Related Issues/PRs

Fixes #5383.

### Changes

- Parse and validate serialized tensor strides in both PyTorch rebuild paths.
- Materialize non-contiguous views in logical row-major order while preserving the existing contiguous fast path and dtype conversion.
- Validate tensor views lazily against the actual storage bytes, so one malformed tensor does not prevent checkpoint metadata from loading.
- Propagate missing or unreadable storage errors instead of silently synthesizing zero-filled tensor data.
- Support materializing zero-element tensors while preserving rank-zero scalar behavior.
- Add an offline PyTorch fixture covering a permuted offset view and an expanded zero-stride view, plus focused malformed-metadata and legacy rebuild coverage.

### Release note

Corrupt PyTorch checkpoints whose storage is missing or truncated now return a contextual data error. They are no longer accepted by padding or replacing tensor values with zeros.

### Testing

- `cargo run-checks` (initial implementation)
- `cargo xtask check lint` (current update)
- `cargo test -p burn-store` (250 unit tests and 35 doctests passed; 17 doctests ignored)
- `cargo clippy -p burn-store --all-targets -- -D warnings`
- Independent reader probe against a PyTorch-generated non-contiguous tensor


crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -617,22 +785,35 @@ fn rebuild_tensor_impl(
...
+                    if offset_bytes >= data.len() {
+                        return Ok(TensorData::new(
+                            vec![0.0f32; num_elements],
+                            shape_clone.clone(),
+                        ));
+                    }
+                    &data[offset_bytes..]
+                };
                 let available_elements = data_slice.len() / element_size;
                 let elements_to_read = num_elements.min(available_elements);
```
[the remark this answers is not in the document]
original4422: Thanks — confirmed and fixed in b9daaa598. The contiguous fast path now validates and slices the exact checked byte range before dtype conversion, so an offset at/past the entry or a short entry returns DataError instead of zero-filling. Valid zero-element tensors still request zero bytes. I added focused regressions for both malformed cases. Validation: cargo test -p burn-store (238 unit tests and 35 doctests passed, 17 doctests ignored), cargo clippy -p burn-store --all-targets -- -D warnings, and cargo fmt --all -- --check.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -564,6 +752,17 @@ fn rebuild_tensor_impl(
         ));
     };
 
+    let num_elements = tensor_num_elements(&shape)?.max(1);
```
antimora: `tensor_num_elements(&[])` already returns 1 for rank-0 scalars via the empty-product identity, so `.max(1)` changes nothing there. Its only effect is on shapes containing 0, where it is wrong.

For `shape = [0, 3]`: `num_elements` becomes 1, `contiguous_tensor_bytes` correctly yields 0 bytes, then `values.resize(num_elements, 0.0)` pushes it back to 1 and `TensorData::check_data_len` (`burn-std/src/data/tensor/base.rs:188`) asserts. `catch_unwind` flattens that into `PanicError("Panic occurred while loading tensor data")`, which names neither the tensor nor the cause.

I tested dropping it:

```rust
let num_elements = tensor_num_elements(&shape)?;
```

`empty.pt` then loads correctly (`to_data()` returns an empty slice), scalars are unchanged (`[42.0]`), and all 239 tests pass. It also makes the three new `shape.contains(&0)` special cases at lines 514, 536 and 560 unnecessary, and lets you delete the "Empty tensors cannot be loaded with to_data()" note at `tests/reader/mod.rs:354`.

This panics on `main` too, so it is not a regression. Flagging it because this PR adds the zero-dim handling and stops one line short of finishing it.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -564,6 +752,17 @@ fn rebuild_tensor_impl(
...
+    let storage_extent = tensor_storage_extent(&shape, &stride, storage_offset)?;
+    if let Some(total) = storage_total_elements
```
antimora: This runs during pickle parsing, so a failure aborts `PytorchReader::new` for the entire checkpoint, and the message names no tensor.

I patched `tensor_2d.pt` to declare storage numel 4 while its `data/0` member genuinely holds 6 floats:

```
Pickle parsing error: Invalid data in pickle file:
Tensor view requires 6 storage elements but storage contains 4.
```

The tensor data is fully present. Only the declared count disagrees, and for the zip format the member length is the source of truth, not index 4 of the persistent id. Every other tensor in the file is lost with it.

Real `torch.save` output always agrees, so this only fires on hand-written or converted pickles. Suggest moving the check into the load closure as a `TensorSnapshotError::DataError`, which is exactly how b9daaa5 already scopes the truncation failure: same validation, blast radius of one tensor, and the real bound is already enforced there against `data.len()` anyway.

If you prefer to keep it eager, please add `shape`, `stride`, `storage_offset` and `storage_key` to the message. They cannot name the tensor (this runs before the pickle assigns a key) but they do identify it uniquely to someone holding a 500-tensor file.

antimora (commented):
Reviewed the stride handling end to end, and verified findings against the branch rather than by reading alone: disassembled the new fixture with `pickletools` and independently re-derived its expected values, probed the helper functions directly, and built corrupted `.pt` files to exercise the failure paths.

The core fix is correct. The gather handles offset, permute and zero-stride broadcast uniformly, `is_contiguous_stride`'s size-1-dim skipping matches PyTorch's own `is_contiguous`, and the new fixture is a genuine regression test that fails on `main`. Baseline is green: 238 lib tests plus 37 in `pytorch-tests`, `clippy --all-targets -D warnings` clean, `cargo fmt --check` clean.

Replacing `*i as usize` with `usize::try_from` also closes real wrapping bugs, and the fallback this removes was worse than merely lenient: it returned `vec![0.0f32; n]` regardless of dtype, so a truncated i64 tensor came back as f32 and failed later with a confusing dtype error.

Three things I would like to see before merge, all of them one-liners or net line reductions:

1. Drop `.max(1)` at `pickle_reader.rs:755`. Verified: `empty.pt` then loads correctly, scalars are unchanged, 239/239 tests pass, and the three new `shape.contains(&0)` special cases become unnecessary.
2. Propagate the read error at line 819 instead of returning zeros. A deleted storage member currently loads as an all-zeros tensor while a storage truncated by 8 bytes is a hard error.
3. Scope the storage-extent check at line 757 to one tensor. As written it aborts `PytorchReader::new` for the whole checkpoint with a message that names no tensor.

Details inline, along with some smaller doc and test notes.

One process note: the truncation change is user-visible. Anyone whose checkpoint previously "loaded" with zero padding now gets an error, so it probably deserves a release note even though the checklist marks no user-facing change.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -617,24 +817,29 @@ fn rebuild_tensor_impl(
         Rc::new(move || {
             // Load data only when needed
             if let Ok(data) = data_source_clone.read(&data_file_key) {
```
antimora: This `if let Ok(...)` discards every `io::Error` from `read`, and the `else` at line 1002 returns an all-zeros tensor. After this PR that produces an odd asymmetry. Same file, same tensor:

| corruption | result |
| --- | --- |
| none | `Ok` -> `[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]` |
| storage truncated by 8 bytes | `Err(DataError("... byte range 0..24 extends beyond 16 available storage bytes"))` |
| storage member deleted from the zip | `Ok` -> `[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]` |

Partial corruption is now a hard error, total corruption is not. Swallowed here: deleted or renamed file, permission denied, corrupt zip central directory, CRC failure, and `LegacyMultiStorageSource`'s own error, which carries a `// NO FALLBACK!` comment and a precise message that is then thrown away.

Zeros are the worst possible sentinel for weights, because zero is a valid weight: the model loads "successfully" and inference returns plausible garbage with no warning and no return-value signal.

Pre-existing, but commit b9daaa5 is titled "reject truncated contiguous tensors" and adds `?`-propagating errors inside this exact block. The closure already returns `Result<TensorData, TensorSnapshotError>`:

```rust
let data = data_source_clone.read(&data_file_key).map_err(|e| {
    crate::TensorSnapshotError::DataError(format!(
        "Failed to read storage '{data_file_key}' for tensor with shape {shape_clone:?}: {e}"
    ))
})?;
```

and the ~55-line zeros `else` branch with its twelve dtype arms goes away. Net line reduction.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -472,13 +488,181 @@ fn storage_type_to_dtype(storage_type: &str) -> Result<DType> {
...
+    if is_contiguous_stride(shape, stride) {
+        return Ok(None);
+    }
+
+    let byte_len = num_elements.checked_mul(element_size).ok_or_else(|| {
+        crate::TensorSnapshotError::DataError(
+            "Tensor byte length calculation overflows usize".to_string(),
+        )
+    })?;
+    let mut reordered = Vec::with_capacity(byte_len);
```
antimora: This allocates before validating any index, whereas `contiguous_tensor_bytes` validates against `data.len()` first and never allocates on the failure path.

The new `storage_extent > total` check at line 757 does not bound this. `storage_extent` sums `(dim - 1) * stride`, so a dim with stride 0 contributes nothing: a pickle declaring shape `(2^21, 2^21)` with stride `(0, 0)` over a 1-element storage has `extent == total == 1` and passes, then dies in the allocator. `Vec::with_capacity` routes to `handle_alloc_error`, which aborts rather than unwinds, so the `catch_unwind` in `to_data()` does not contain it. That makes it an uncatchable process kill where the contiguous path would have produced a recoverable `DataError`.

`main` aborts on the same input via its own `Vec::with_capacity(num_elements)` in the dtype arms, so this is pre-existing in kind. Worth closing here because this is the hardening pass, and because the extent check reads as protection against exactly this while not providing it. Comparing `num_elements` against `data.len() / element_size` up front is one line and matches what the contiguous path already does.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -472,13 +488,181 @@ fn storage_type_to_dtype(storage_type: &str) -> Result<DType> {
...
+/// Returns `None` for an already-contiguous layout so the existing zero-copy slice path is used.
+fn reorder_tensor_bytes(
+    data: &[u8],
+    shape: &[usize],
+    stride: &[usize],
+    storage_offset: usize,
+    element_size: usize,
+    num_elements: usize,
+) -> std::result::Result<Option<Vec<u8>>, crate::TensorSnapshotError> {
+    if shape.len() != stride.len() {
```
antimora: Unreachable. `rebuild_tensor_impl` calls `tensor_storage_extent` at line 756 unconditionally, and that rejects a rank mismatch before this closure is ever constructed. The rank check there precedes its zero-shape early return, so it catches this even for shapes containing 0 (I confirmed with a direct probe).

So there are now two rank checks with two different error types, and only the `PickleError::InvalidData` one is reachable and covered by the unit test at line 1809.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -472,13 +488,181 @@ fn storage_type_to_dtype(storage_type: &str) -> Result<DType> {
...
+        )
+    })?;
+    let mut reordered = Vec::with_capacity(byte_len);
+
+    for linear_index in 0..num_elements {
+        let mut remaining = linear_index;
+        let mut storage_index = storage_offset;
+
+        for (&dim, &step) in shape.iter().zip(stride).rev() {
+            let coordinate = remaining % dim;
```
antimora, line 622: `remaining % dim` divides by zero when `dim == 0`. That is currently unreachable only because `is_contiguous_stride` returns `true` for any shape containing 0 (line 536) and short-circuits to `Ok(None)` above.

Worth not leaving this function's safety resting on an unrelated predicate's early return. Note that `.max(1)` at line 755 forces `num_elements` to 1 for a zero-element shape, so this loop would execute exactly once and divide by zero if that early return were ever relaxed. Dropping `.max(1)` makes `num_elements == 0` and the loop body naturally unreachable, which is the cleaner guarantee.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -617,24 +817,29 @@ fn rebuild_tensor_impl(
...
+                } else {
+                    contiguous_tensor_bytes(
+                        &data,
+                        &shape_clone,
+                        storage_offset,
+                        element_size,
+                        num_elements,
+                    )?
+                };
+                let elements_to_read = data_slice.len() / element_size;
```
antimora: `elements_to_read` now always equals `num_elements`: the reorder path pushes exactly `element_size` bytes exactly `num_elements` times, and `contiguous_tensor_bytes` returns exactly `elements_to_read * element_size` bytes or `Err`.

That makes the `values.resize(num_elements, ...)` call in all thirteen dtype arms dead, and the surviving `// Pad with zeros if needed` at line 857 now misdescribes the code: partial reads are no longer tolerated, which is the whole point of b9daaa5. Suggest deleting both, otherwise they will silently re-mask truncation the moment anyone relaxes `contiguous_tensor_bytes`.

The one input that still reaches the resize is a zero-element shape, and there it is what causes the panic (see the comment on line 755).

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -605,10 +805,10 @@ fn rebuild_tensor_impl(
         let source = source
             .lock()
             .unwrap_or_else(|poisoned| poisoned.into_inner());
-        let bytes_needed = match storage_total_elements {
-            Some(total) => total * dtype.size(),
-            None => (storage_offset + shape.iter().product::<usize>()) * dtype.size(),
-        };
+        let storage_elements = storage_total_elements.unwrap_or(storage_extent);
```
antimora: For zero-stride views this is smaller than the old `storage_offset + product`. Using this PR's own `expanded` fixture (shape `[2, 3]`, stride `[0, 1]`, offset 1): old = 1 + 6 = 7, new = extent = 4.

The new number is the more correct storage extent, but it is not used as an extent here. `track_storage_usage` feeds `try_build_storage_map`, which lays storages out sequentially (`current_offset += 8 + size`), so under-sizing one storage shifts the file offset of every storage after it. The result is not an error, it is misaligned bytes reinterpreted as floats.

Narrow in practice: reachable only when `storage_total_elements` is `None`, and `track_storage_usage` takes a per-key max so multiple views widen the estimate. Flagging it because the failure mode is silent wrong data and this PR changed the number.

Tangent worth considering separately: that map skips the 8-byte little-endian element count preceding each storage (`current_offset + 8`) rather than reading it back and asserting it matches the assumed size. That u64 is ground truth sitting in the file and would turn this entire class of misalignment into a hard error.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -472,13 +488,181 @@ fn storage_type_to_dtype(storage_type: &str) -> Result<DType> {
...
+        crate::TensorSnapshotError::DataError(format!(
+            "Contiguous tensor byte range {byte_start}..{byte_end} extends beyond {} available storage bytes",
+            data.len()
+        ))
+    })
+}
+
+/// Gather a non-contiguous storage view into logical row-major byte order.
+///
+/// Returns `None` for an already-contiguous layout so the existing zero-copy slice path is used.
```
antimora: There is no zero-copy path here. `data` comes from `data_source_clone.read(...)`, which returns an owned `Vec<u8>` (`lazy_data.rs:243`), and every dtype arm then copies element by element via `from_le_bytes`. The contiguous path saves one intermediate buffer relative to the reorder path, it does not avoid a copy.

Suggest: "Returns `None` for an already-contiguous layout, letting the caller slice the storage bytes directly instead of building a reordered buffer."

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -472,13 +488,181 @@ fn storage_type_to_dtype(storage_type: &str) -> Result<DType> {
...
+/// Return the number of logical elements, rejecting shape arithmetic overflow.
+fn tensor_num_elements(shape: &[usize]) -> Result<usize> {
+    shape.iter().try_fold(1usize, |total, dim| {
+        total.checked_mul(*dim).ok_or_else(|| {
+            PickleError::InvalidData("Tensor shape element count overflows usize".to_string())
+        })
+    })
+}
+
+/// Return one past the largest storage index referenced by a strided tensor.
```
antimora: Not accurate for this function's own zero-shape branch at line 514, which returns `storage_offset`. That is the first index the view would occupy, not one past the largest, and an empty view references no index at all.

Both callers actually use the return value as "minimum storage elements this view requires" (the bound check at 757 and the `unwrap_or` at 808), and under that framing `storage_offset` is the right answer. Suggest restating the contract as the requirement: "Return the minimum number of storage elements this view requires, that is, one past the largest index it reads." A `# Errors` note about the rank mismatch rejection would help too.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -472,13 +488,181 @@ fn storage_type_to_dtype(storage_type: &str) -> Result<DType> {
...
+                PickleError::InvalidData("Tensor storage extent overflows usize".to_string())
+            })
+        })?;
+
+    max_index.checked_add(1).ok_or_else(|| {
+        PickleError::InvalidData("Tensor storage extent overflows usize".to_string())
+    })
+}
+
+fn is_contiguous_stride(shape: &[usize], stride: &[usize]) -> bool {
```
antimora, line 546: This and `contiguous_tensor_bytes` are the only new helpers without a doc comment, and they hold the least self-evident logic.

The `dim > 1` guard on line 542 deliberately ignores the stride of size-1 dims. That matches PyTorch (a size-1 dim cannot be stepped, so `is_contiguous` ignores whatever is in its stride slot), and it is the only reason the `expanded` fixture with stride `[0, 1]` is classified correctly. Without a note it reads like a bug.

Lines 545-547 also return `false` on `checked_mul` overflow, that is, "I could not determine contiguity" is reported as "not contiguous". Harmless today, since line 755 already rejected any shape whose full product overflows and `expected` is a product of a suffix of those dims. It becomes a trap if this predicate is ever reused to pick a zero-copy fast path.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -352,20 +352,39 @@ fn parse_storage_arg(arg: &Object, fn_name: &str) -> Result<(Vec<u8>, Option<Vec
...
-                    "shape must contain ints".to_string(),
-                )),
+                Object::Int(i) => usize::try_from(*i).map_err(|_| {
+                    PickleError::InvalidData(format!(
+                        "{}: {} must contain non-negative ints",
+                        fn_name, field_name
+                    ))
+                }),
+                _ => Err(PickleError::InvalidData(format!(
+                    "{}: {} must contain non-negative ints",
```
antimora: This arm fires when the element is not an `Object::Int` at all, so "non-negative" is misleading: a shape tuple containing a string now reports "must contain non-negative ints". `main` said "shape must contain ints" here.

Keeping the sign wording on the `try_from` arm at line 363 and using `"{}: {} must contain ints"` here restores the distinction.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -395,17 +414,16 @@ fn rebuild_tensor(
...
 
     let (storage_info, storage_tuple) = parse_storage_arg(&args[0], "rebuild_tensor")?;
-    let storage_offset = match &args[1] {
-        Object::Int(offset) => *offset as usize,
-        _ => 0,
-    };
-    let shape = parse_shape_arg(&args[2], "rebuild_tensor")?;
+    let storage_offset = parse_storage_offset(&args[1], "rebuild_tensor")?;
+    let shape = parse_size_tuple_arg(&args[2], "rebuild_tensor", "shape")?;
+    let stride = parse_size_tuple_arg(&args[3], "rebuild_tensor", "stride")?;
```
antimora, line 418: Worth noting that `_rebuild_tensor` (PyTorch < 1.6) has no fixture anywhere in this crate. I scanned every `.pt` and `.pth` under `crates/burn-store`, including `simple_legacy.pt`, `legacy_with_offsets.pt`, `legacy_shared_storage.pt`, `legacy_uncloned_views.pt` and `tar_float32.tar`: all of them dispatch to `_rebuild_tensor_v2`.

On `main` this function never touched `args[3]` despite the `args.len() < 4` guard. It is now a hard parse, so a pre-1.6 file whose `args[3]` is not a tuple of non-negative ints fails the entire load where it previously succeeded. Nothing in the suite can tell you whether this path works. A direct unit call to `rebuild_tensor` with a synthesized `Object::Tuple` covering one contiguous and one permuted stride would be enough.

crates/burn-store/src/pytorch/pickle_reader.rs
```diff
@@ -1592,4 +1797,40 @@ mod tests {
         let err = result.unwrap_err();
         assert!(matches!(err, PickleError::InvalidData(msg) if msg.contains("exceeded 1M nodes")));
     }
+
+    #[test]
+    fn test_tensor_stride_metadata_validation() {
```
antimora: This bundles six assertions across three functions and two error types into one test. The rest of the crate splits per behavior (`keyremapper.rs` has eight narrowly named tests for one feature, `filter.rs` has fourteen), and the only other unit test in this file, `test_memo_bomb_mitigation`, tests one thing.

Two concrete costs rather than style preference: the first failing assertion hides the other five, so a regression in `contiguous_tensor_bytes` stays invisible until the earlier ones are fixed, and `contains("beyond")` is reused for three different messages from two different functions, so the assertions cannot tell you which failure actually occurred.

The name is also slightly off, since lines 1818-1834 assert data-range validation at read time rather than metadata parsing. Splitting into something like `stride_rejects_negative_values`, `storage_extent_rejects_rank_mismatch`, `storage_extent_rejects_overflow`, `reorder_rejects_out_of_bounds_stride` and `contiguous_rejects_truncated_storage` would match both the file and the crate.

crates/burn-store/src/pytorch/tests/reader/mod.rs
```diff
@@ -226,6 +226,30 @@ fn test_4d_tensor() {
     assert_eq!(values.len(), 24);
 }
 
+#[test]
+fn test_non_contiguous_tensors() {
```
antimora, line 230: Good regression test. The values are hand-derivable rather than captured, and it fails loudly on `main`, which reads `storage[5..29]` linearly and returns `[5, 6, 7, ...]` instead of `[5, 9, 13, ...]`. I disassembled the fixture with `pickletools` and independently re-derived both arrays: they match exactly, and both tensors do alias one 32-element storage, so this incidentally proves the new extent check does not false-positive on shared storage.

Gaps worth closing while you are here:

- Non-f32 through the reorder path. The fixture is f32 only, so `reorder_tensor_bytes` only ever runs with `element_size == 4`. The dtype arms differ structurally: F64/I64/U64 use `copy_from_slice` over an 8-byte window, while I8/U8/Bool bypass `element_size` arithmetic entirely. A gather bug visible only at element size 1, 2 or 8 passes today. Adding a permuted i64 and a permuted u8 to the same fixture is nearly free.
- Issue #5383's actual reproducer is a 4D channels-last conv weight, shape `[o, i, k, k]` with strides `(27, 1, 9, 3)`. The 3D permute here is structurally equivalent, but a fixture matching the reported case would be a more direct regression test.
- A rank-0 view at a non-zero offset (`torch.arange(10)[5]`, shape `()`, stride `()`, offset 5). This works, I checked, but the arithmetic only comes out right via the empty-fold identity, so one assertion is cheap insurance.

crates/burn-store/src/pytorch/tests/reader/test_data.py
```diff
@@ -82,6 +82,16 @@ def save_test_file(filename, data, description):
...
 save_test_file("tensor_4d.pt", {"tensor": tensor_4d}, "4D tensor shape (2, 3, 2, 2)")
 
+# Non-contiguous tensors sharing storage, including an offset view and zero stride
+backing = torch.arange(32, dtype=torch.float32)
+permuted = backing[5:29].reshape(2, 3, 4).permute(0, 2, 1)
+expanded = backing[1:4].view(1, 3).expand(2, 3)
+save_test_file(
+    "non_contiguous.pt",
+    {"permuted": permuted, "expanded": expanded},
+    "Non-contiguous tensors with storage offsets and strides [12, 1, 4] and [0, 1]",
```
antimora: The description records the strides but not the storage offsets (5 and 1), nor the fact that both tensors deliberately share one 32-element storage. Those are the load-bearing properties: if someone edits the slices, the test's meaning changes silently.

Something like `"Non-contiguous views of one 32-element storage: offsets 5 and 1, strides [12, 1, 4] and [0, 1]"` captures it.

Separately, worth calling out that the placement of this block is correct and non-obvious. `torch.arange` consumes no RNG draws, so inserting it after `torch.manual_seed(42)` but before the `randn` fixtures leaves every downstream fixture byte-identical on regeneration. A `randn`-based fixture dropped in here would have silently rewritten all of them.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. `num_elements` now keeps the empty product result without `.max(1)`, the shape-based special cases were removed from contiguity and byte slicing, and `test_empty_tensor` now materializes and asserts an empty f32 slice. Rank-zero scalar coverage remains green.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. The declared storage-count check no longer runs during pickle parsing. The lazy tensor closure validates the view against the actual storage bytes, so the failure is scoped to `to_data()` for that tensor. A regression declares 4 elements while the ZIP member contains 32 and verifies the valid 24-element view still loads.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. Storage read errors are mapped to a contextual `DataError` containing the storage key and tensor shape, and the full zero-filled fallback was deleted. `missing_storage_returns_contextual_error` was RED before the change because it returned six f32 zeros and is now green.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: I did not add `num_elements <= data.len() / element_size`: legitimate zero-stride views violate that relation (the existing expanded fixture has 6 logical elements backed by 3 stored elements). b0f55ed1c instead validates the maximum referenced storage index against actual bytes before reordering and uses fallible `try_reserve_exact` for the intermediate buffer. A general cap on materialized logical tensor size would need a repository-wide resource policy, so I am leaving this thread open for maintainer preference.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. The duplicate rank check was removed from `reorder_tensor_bytes`; rank and extent validation now happens once inside the lazy tensor closure through `tensor_storage_extent`, so it returns a per-tensor `DataError` rather than an eager pickle error.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. Dropping `.max(1)` makes `num_elements == 0` for every zero-element shape, so the reorder loop is naturally empty and cannot execute `% dim`. The unrelated `shape.contains(0)` contiguity shortcut was removed, and the empty end-to-end regression is green.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. `elements_to_read`, all thirteen `resize` calls, and the stale zero-padding comment were removed. Both paths now produce an exact `num_elements * element_size` slice or return an error, and conversion iterates exactly `num_elements`.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. Legacy storage-map allocation size is now separate from strided view extent: it uses the declared total when present and otherwise preserves the old checked `storage_offset + num_elements` estimate, so a zero-stride view cannot shift later storages earlier. Reading and validating the legacy length prefixes is a broader legacy-format change and was not mixed into this PR.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. The documentation now says the contiguous path slices storage bytes directly instead of building an intermediate reordered buffer; it no longer calls the path zero-copy.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. The helper contract now describes the minimum required storage elements, clarifies the empty-view and size-one-dimension behavior, and documents rank-mismatch and arithmetic-overflow errors.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. `is_contiguous_stride` now documents PyTorch's size-one-dimension rule and its conservative overflow result, while `contiguous_tensor_bytes` documents exact slicing and its arithmetic/storage error contract.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. Non-integer tuple members now report `must contain ints`, while failed signed-to-usize conversion retains `must contain non-negative ints`. Both cases have separate exact-message tests.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c with direct synthesized `Object` calls to legacy `_rebuild_tensor`: one contiguous stride and one permuted stride both load through the real fixture data source. I also added a rank-zero nonzero-offset case. All three pass through `to_data()`, not only metadata construction.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Addressed in b0f55ed1c. The bundled test was split into narrow tests for negative/type metadata, rank mismatch, extent overflow, strided out-of-bounds access, truncated contiguous storage, and offset-past-storage. Assertions now match the complete error messages.

crates/burn-store/src/pytorch/tests/reader/mod.rs
original4422: I added focused coverage in b0f55ed1c for reordered element widths 1, 2, and 8, plus a rank-zero view at offset 5 through the legacy rebuild path. I did not add another 4D fixture: the checked gather is rank-generic and the existing 3D permutation exercises the same non-contiguous coordinate mapping without duplicating the binary fixture. Full burn-store validation remains green.

crates/burn-store/src/pytorch/tests/reader/test_data.py
original4422: Addressed in b0f55ed1c. The generator description now records the shared 32-element storage, offsets 5 and 1, and both strides. A nearby comment explains that `arange` consumes no RNG draws and therefore preserves downstream seeded fixtures; the committed fixture binary itself did not need regeneration.

laggui (commented):
@antimora I believe all concerns have been addressed.

LGTM, pending merge conflicts are resolved.

original4422:
@laggui Thank you. I merged current upstream main (b8693f3b9f826d0ada194754c171b129fd78b3e4) into this branch in 64e992ee00130926ab9974f92aaac660146cb72d and resolved the only conflict in crates/burn-store/src/pytorch/pickle_reader.rs by retaining the upstream burn_pack::Tensor / bridge::deferred migration while porting the cumulative stride reordering, storage validation, empty/scalar, and legacy-storage behavior.

Validation on the resulting exact head:
- cargo test -p burn-store: 260 unit tests, 6 integration tests, and 36 doctests passed; 17 repository-configured doctests ignored
- cargo clippy -p burn-store --all-targets -- -D warnings: passed
- cargo xtask check lint: passed across the workspace
- cargo fmt --all -- --check and git diff --check: passed
- focused non-contiguous test: 1/1 passed; private pickle-reader tests: 14/14 passed

GitHub now reports the PR mergeable against that main head, with the new current-head CI running. Could you please take another look and confirm whether it can be merged once those checks finish?

crates/burn-store/src/pytorch/pickle_reader.rs
antimora: Reopening this one. `try_reserve_exact` does not close it.

I ran the case end to end: shape `(2^21, 2^21)`, stride `(0, 0)`, 32-element storage. Extent is 1 so the bounds check passes, `try_reserve_exact` for 17592186044416 bytes succeeds, and the gather loop then runs 2^42 iterations. I killed it at 100% CPU and 34% of system RAM.

Isolated probe on macOS:

```
PROBE_RESERVE: SUCCEEDED for 17592186044416 bytes (cap=17592186044416)
```

`try_reserve_exact` only reports what the allocator rejects at reservation time. macOS, and Linux with overcommit, hand back address space lazily, so the failure moves to first touch, which here is the per-element `extend_from_slice`. The guard turns an immediate abort into slow RAM exhaustion.

Your objection to `num_elements <= data.len() / element_size` was right and my suggestion was wrong: `expand` legitimately produces more logical elements than stored ones, as your own `expanded` fixture shows. There is no bound derivable from metadata alone, so this does need the resource policy you described.

Not blocking, and not something to fix here. `main` hits the same wall on the same input via `values.resize(num_elements, 0.0)`, so severity is unchanged. Tracked in #5506.

The one thing worth changing is the wording above, since as written it reads as resolved and the next person will trust it.

antimora (approved):
Rechecked at 64e992ee. Verified against a build rather than by reading: 260 unit tests, 6 integration and 36 doctests pass, `cargo clippy -p burn-store --all-targets -- -D warnings` clean, `cargo fmt --check` clean, and `git merge-tree` against current main (9542363f8) merges clean.

Every thread is addressed except the allocation one, where I replied inline. Short version: `try_reserve_exact` does not fire on macOS or on Linux with overcommit, so that input still exhausts RAM. Severity is unchanged from main and it does not belong in this PR. Tracked in #5506. Worth adjusting the reply wording there though, since it currently reads as resolved.

Two things I checked independently rather than taking on trust.

The merge resolution. No `TensorSnapshot` leftovers, the upstream `zeros()` helper from a467575b5 is fully removed along with its call sites and leaves no dangling references, and every removed line traces to the stride work.

Issue #5383's actual reproducer, which you skipped a fixture for. I synthesized shape `[1, 3, 3, 3]` with strides `(27, 1, 9, 3)` against the fixture storage and hand-derived the expected gather:

```
[0, 3, 6, 9, 12, 15, 18, 21, 24, 1, 4, 7, 10, 13, 16, 19, 22, 25, 2, 5, 8, 11, 14, 17, 20, 23, 26]
```

Exact match, so the gather really is rank-generic and skipping the extra binary fixture is fine.

Dropping the zero-fill fallback is the most valuable part of this change. A missing storage member loading as plausible all-zero weights is the kind of failure that costs someone a week.

Approving.

crates/burn-store/src/pytorch/pickle_reader.rs
original4422: Thanks for testing this end to end and for the correction. To clarify my earlier reply: `try_reserve_exact` does **not** prevent the zero-stride overcommit / first-touch exhaustion case, so this PR does not resolve that resource-exhaustion risk. The severity remains unchanged from `main`, and the repository-wide materialization policy is tracked in #5506. I’m leaving this thread open for that follow-up.

PR #5394, opened 2026-08-20
Fix non-finite padding in Flex conv fast paths

### LINKED ISSUE #5385: conv2d: burn-flex returns a finite value or NaN at padded positions for the same convolution, depending on channel count
**Describe the bug**

`burn-flex` returns either a finite value or `NaN` at padded output positions for the same
convolution, depending on how many channels the tensors have. The channel count selects between
internal fast paths, and those paths disagree about whether padded positions take part in the
multiply-accumulate.

It only becomes observable when a weight is non-finite, since `0 * w` and "skip this position"
differ only when `w` is `inf` or `NaN`. But the disagreement is between code paths inside one
crate, which is not a question of interpretation.

A 3x3 convolution with padding 1 over a 4x4 input, `groups = 1`, one output channel, a single
`-inf` weight and everything else finite. `out[0]` is the top-left output, the position where the
`-inf` weight lands entirely in the pad:

```
channels_in = 1  ->  out[0] = 14.0
channels_in = 2  ->  out[0] = 92.0
channels_in = 3  ->  out[0] = 234.0
channels_in = 4  ->  out[0] = 440.0
channels_in = 5  ->  out[0] = NaN
channels_in = 6  ->  out[0] = NaN
```

`channels_out` has its own cliff, holding `channels_in = 2`:

```
channels_out = 15 ->  out[0] = 92.0
channels_out = 16 ->  out[0] = 92.0
channels_out = 17 ->  out[0] = NaN
channels_out = 18 ->  out[0] = NaN
```

4 and 16 are `SMALL_CHANNEL_IN_THRESHOLD` and `SMALL_CHANNEL_OUT_THRESHOLD` in
`crates/burn-flex/src/ops/conv.rs`, both documented as "picked empirically" from benchmark data.

**To Reproduce**

```rust
use burn::backend::Flex;
use burn::tensor::module::conv2d;
use burn::tensor::{ops::ConvOptions, Tensor, TensorData};

type B = Flex;

// 3x3 convolution, padding 1, over a 4x4 input, groups = 1.
// One weight element is -inf; every other weight and the whole input are finite.
fn conv_with_one_inf_weight(channels_in: usize, channels_out: usize) -> Vec<f32> {

    let x_data: Vec<f32> = (0..channels_in * 16).map(|i| 1.0 + i as f32).collect();
    let x: Tensor<B, 4> =
        Tensor::from_data(TensorData::new(x_data, [1, channels_in, 4, 4]), &device);

    let mut w_data = vec![1.0f32; channels_out * channels_in * 9];
    w_data[0] = f32::NEG_INFINITY;
    let w: Tensor<B, 4> =
        Tensor::from_data(TensorData::new(w_data, [channels_out, channels_in, 3, 3]), &device);

    conv2d(x, w, None, ConvOptions::<2>::new([1, 1], [1, 1], [1, 1], 1))
        .to_data()
        .to_vec::<f32>()
        .unwrap()
}

fn main() {
    for channels_in in 1..=6 {
        println!("channels_in = {channels_in}  ->  out[0] = {:?}",
                 conv_with_one_inf_weight(channels_in, 1)[0]);
    }
    for channels_out in [15usize, 16, 17, 18] {
        println!("channels_out = {channels_out} ->  out[0] = {:?}",
                 conv_with_one_inf_weight(2, channels_out)[0]);
    }
}
```

The divergence is confined to positions that read the pad. Printing the whole 4x4 output for the
two shapes either side of the `channels_in` threshold:

```
channels_in = 4                          channels_in = 5
[440.0, 672.0, 696.0, 472.0]             [NaN, NaN,  NaN,  NaN ]
[708.0, -inf,  -inf,  -inf ]             [NaN, -inf, -inf, -inf]
[804.0, -inf,  -inf,  -inf ]             [NaN, -inf, -inf, -inf]
[568.0, -inf,  -inf,  -inf ]             [NaN, -inf, -inf, -inf]
```

Every position where the `-inf` weight sees real input is `-inf` on both, as it should be. Only
row 0 and column 0, where that weight reads the pad, differ.

**Expected behavior**

I do not have a view on which convention is correct, and either could be defended:

| reading | a padded position computes | result with a `-inf` weight |
|---|---|---|
| the input is genuinely extended with zeros | `0 * w`, accumulated | `NaN` |
| only real input positions are accumulated | a sum over an empty set | a finite value |

What should not happen is that one crate answers both ways for the same convolution because a
channel count crossed a cache-tuning constant. Whichever reading burn intends, the three padded
paths in `burn-flex` should agree.

**Screenshots**

n/a

- OS: macOS 15 (aarch64, Apple Silicon); `wgpu` on Metal
- Browser: n/a
- Version: `burn` 0.21.0, `burn-flex` 0.21.0, `tch` 0.22.0 / libtorch 2.9.0, `burn-wgpu` 0.21.0.
  The `burn-wgpu` figures below were taken with autotune both disabled and enabled; they agree.

n/a

**Additional context**

*Where it comes from.* `conv3d_typed!` in `crates/burn-flex/src/ops/conv.rs` dispatches in order:
1x1, depthwise, small-channel, direct, then the generic tiled im2col plus gemm. `is_1x1_conv` and
`should_use_direct_conv` both require `padding == [0, 0, 0]`, so with padding exactly three paths
are reachable: depthwise, small-channel, and generic. Taking each in turn, same convolution, same
single `-inf` weight, `out[0]` again:

| path | shape | `out[0]` |
|---|---|---|
| depthwise | `c_in = c_out = groups = 2, 8, 32, 64` | `14.0` at every one |
| small-channel | `c_in = 2, c_out = 2` / `c_in = 4, c_out = 16` | `92.0` / `440.0` |
| generic | `c_in = 5, c_out = 1` / `c_in = 2, c_out = 17` | `NaN` / `NaN` |

Depthwise stays finite at 64 channels, far past both thresholds, so this is three distinct paths
rather than one threshold seen twice.

The depthwise and small-channel paths share `conv_plane_accumulate`, which uses `valid_out_range`
to compute the output range each kernel position can legally touch and skips the rest:

```rust
let (oh_start, oh_end) = oh_ranges[kh];
// Skip kernel rows that fall outside the padded image at this oh.
if oh < oh_start || oh >= oh_end {
    continue;
}
```

The generic path materializes the zeros instead. `im2col_3d_tile` writes into a buffer allocated as
`vec![zero; col_len * tile_size]` and leaves padded entries at their initial value:

```rust
} else {
    // Padding: skip channels_per_group positions (already zero)
    col_offset += channels_per_group;
}
```

The subsequent `gemm` runs over the full `col_len`, so those zeros are multiplied by their weights,
and `0 * -inf` is `NaN`. Both quoted snippets are byte-identical at tag `v0.21.0` and on `main`.
#5267 does rework `conv_plane_accumulate`, but only for SIMD and unchecked indexing; the skip is
unchanged.

*Cross-backend behaviour, as supporting evidence.* Running the same sweep on the other two
backends, at all ten points rather than a single shape:

| backend | `channels_in` 1..6 | `channels_out` 15..18 |
|---|---|---|
| `burn-flex` | finite, finite, finite, finite, **NaN**, **NaN** | finite, finite, **NaN**, **NaN** |
| `burn-tch` | `NaN` at all six | `NaN` at all four |
| `burn-wgpu` | finite at all six | finite at all four |

So `burn-tch` and `burn-wgpu` are each internally consistent and simply choose opposite
conventions; only `burn-flex` changes its answer with the shape. `burn-tch` does not implement the
convolution itself, it forwards to libtorch
([`crates/burn-tch/src/ops/module.rs:82`](https://github.com/tracel-ai/burn/blob/main/crates/burn-tch/src/ops/module.rs#L82)),
so that `NaN` is PyTorch's own result. PyTorch's `conv2d` documentation describes padding only as
"implicit paddings on both sides of the input" and says nothing about whether those values
participate in the multiply-accumulate, so I do not think there is an upstream rule to appeal to.
That is what made the cross-backend version of this report a question. The `burn-flex` result is
not a question, because it disagrees with itself.

*How it was found.* Differential testing across `burn-flex`, `burn-tch` and `burn-wgpu` over
randomly generated single-operation cases, reduced by hand. The channel sweep was written
afterwards, once reading `conv.rs` suggested the answer should depend on which path was taken.

*Scope.* A non-finite weight is already a bug somewhere upstream, so this is narrow. It still seems
worth fixing, because the failure is silent and shape-dependent: the same model produces an
all-`NaN` feature map at one channel width and a mostly-finite one at another, which is the
difference between a training run that dies loudly and one that quietly learns nothing.

Tested on `f32` only, and on macOS aarch64 only. I have not checked other dtypes, `conv1d`/`conv3d`,
or the `simd` feature.

*Prior art.* burn has treated non-finite handling as in scope before: #2095 fixed `mask_fill`
turning `f32::NEG_INFINITY` into `NaN` on `burn-ndarray`, #3341 added `is_finite` and fixed
`is_close`/`all_close` for infinities, and #4697 fixed `attention_fallback` producing `NaN` on
fully-masked rows. All three are merged pull requests. #5310, on zero-sized tensor support being
inconsistent across backends, is the same genre of complaint. #5284 is an open issue of mine in
that genre too, cross-backend disagreement on a non-finite result, and is unrelated in mechanism.

I searched the tracker for `conv2d NaN`, `conv padding NaN`, `padding infinity`, `conv2d padding`,
and then for this report's actual claim: `small_channel`, `fast path different result`,
`dispatch inconsistent`, `conv shape dependent`, `im2col padding`, `same op different result`.
Nothing covers a fast-path selection changing a result.

tandede:
### Checklist

### Related Issues/PRs

Closes #5385.

### Changes

Flex's depthwise and small-channel convolution paths skip padded input positions, while the generic im2col path materializes them as zeros. Those operations are equivalent for finite weights, but not for non-finite weights: IEEE arithmetic requires `0 * inf` and `0 * NaN` to produce NaN. As a result, changing only the channel count could change padded outputs between finite/infinite values and NaN.

This change detects non-finite weights once per padded fast-path convolution and marks only the output positions where an omitted zero-padding term would participate. The existing vectorized accumulation remains unchanged for finite weights, and convolutions without padding avoid the scan entirely.

The regression covers the small-channel, generic, and depthwise dispatch paths. It also verifies that a finite depthwise channel is not affected by a non-finite weight in another channel.

### Testing

- `cargo test -p burn-flex` (391 passed)
- `cargo test -p burn-flex --no-default-features --features std` (331 passed)
- `BURN_DEVICE=flex cargo test -p burn-backend-tests --no-default-features --features flex,std --test tensor conv2d` (19 passed)
- `cargo clippy -p burn-flex --all-targets -- -D warnings`
- `cargo run-checks`

### DIFF
--- a/crates/burn-flex/src/ops/conv.rs
+++ b/crates/burn-flex/src/ops/conv.rs
@@ -935,6 +935,7 @@ fn conv_plane_accumulate<
     pad_w: usize,
     dilation_h: usize,
     dilation_w: usize,
+    has_non_finite_weights: bool,
     oh_ranges: &[(usize, usize)],
     ow_ranges: &[(usize, usize)],
 ) {
@@ -948,6 +949,50 @@ fn conv_plane_accumulate<
             out_plane, in_plane, w_plane, kernel_h, kernel_w, in_w, out_w, stride_h, stride_w,
             pad_h, pad_w, dilation_h, dilation_w, oh_ranges, ow_ranges,
         );
+    }
+
+    // The vectorized paths skip padded positions. That is equivalent to
+    // multiplying materialized zeros only while every weight is finite; IEEE
+    // arithmetic requires `0 * inf` and `0 * NaN` to produce NaN. Preserve
+    // that result after the fast accumulation without changing its hot loops.
+    if has_non_finite_weights && (pad_h != 0 || pad_w != 0) {
+        propagate_non_finite_padding(out_plane, w_plane, kernel_w, out_w, oh_ranges, ow_ranges);
+    }
+}
+
+/// Mark outputs where a skipped padding term would multiply a non-finite weight.
+fn propagate_non_finite_padding<T: num_traits::Float + Copy>(
+    out_plane: &mut [T],
+    w_plane: &[T],
+    kernel_w: usize,
+    out_w: usize,
+    oh_ranges: &[(usize, usize)],
+    ow_ranges: &[(usize, usize)],
+) {
+    if out_plane.is_empty() || out_w == 0 {
+        return;
+    }
+
+    debug_assert_eq!(out_plane.len() % out_w, 0);
+    let out_h = out_plane.len() / out_w;
+    let nan = T::nan();
+
+    for (kh, &(oh_start, oh_end)) in oh_ranges.iter().enumerate() {
+        for (kw, &(ow_start, ow_end)) in ow_ranges.iter().enumerate() {
+            if w_plane[kh * kernel_w + kw].is_finite() {
+                continue;
+            }
+
+            for oh in 0..out_h {
+                let out_row = &mut out_plane[oh * out_w..(oh + 1) * out_w];
+                if oh < oh_start || oh >= oh_end {
+                    out_row.fill(nan);
+                } else {
+                    out_row[..ow_start].fill(nan);
+                    out_row[ow_end..].fill(nan);
+                }
+            }
+        }
     }
 }
 
@@ -1175,6 +1220,8 @@ where
 
     let x_data: &[T] = x.storage();
     let w_data: &[T] = weight.storage();
+    let has_non_finite_weights =
+        (pad_h != 0 || pad_w != 0) && w_data.iter().any(|value| !value.is_finite());
 
     let in_spatial = in_h * in_w;
     let out_spatial = out_h * out_w;
@@ -1211,6 +1258,7 @@ where
             pad_w,
             dilation_h,
             dilation_w,
+            has_non_finite_weights,
             &oh_ranges,
             &ow_ranges,
         );
@@ -1398,6 +1446,8 @@ where
 
     let x_data: &[T] = x.storage();
     let w_data: &[T] = weight.storage();
+    let has_non_finite_weights =
+        (pad_h != 0 || pad_w != 0) && w_data.iter().any(|value| !value.is_finite());
 
     let in_spatial = in_h * in_w;
     let out_spatial = out_h * out_w;
@@ -1437,6 +1487,7 @@ where
                 pad_w,
                 dilation_h,
                 dilation_w,
+                has_non_finite_weights,
                 &oh_ranges,
                 &ow_ranges,
             );
@@ -2674,6 +2725,7 @@ mod tests {
             /* pad_w */ 0,
             /* dilation_h */ 1,
             /* dilation_w */ 1,
+            /* has_non_finite_weights */ false,
             &oh_ranges,
             &ow_ranges,
         );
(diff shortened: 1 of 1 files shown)


tandede:
The CI workflow is currently waiting for maintainer approval to run (`action_required` for this fork PR). I do not have repository permission to approve it.

laggui (approved):
LGTM.

The added weight scan should be negligible for typical kernels (e.g. 3x3), with meaningful overhead limited to unusually large, heavily padded kernels over tiny inputs. But that shouldn't be blocking.

PR #5395, opened 2026-08-20
fix(fusion): reject a padded reference layout

### LINKED ISSUE #5388: NHWC relayout tests fail on CUDA + fusion
**Describe the bug**
fusion_test_nhwc_relayout_broadcast_different_shapes ... ok
fusion_test_nhwc_relayout_inplace ... FAILED
fusion_test_elementwise_operation_followed_by_interpolate_nearest ... FAILED
fusion_test_elementwise_operation_followed_by_max_pool2d ... FAILED
fusion_test_nhwc_relayout_broadcast_scalar ... FAILED
fusion_test_elementwise_max_pool2d_vectorized_widths ... FAILED
fusion_test_elementwise_max_pool2d_noncontiguous_input ... FAILED

6 tests fail in `burn-backend-tests` running on CUDA + fusion

I haven't dug deeper than that (Does it reproduce on other backends, etc.).
 
**To Reproduce**
`cargo test -p burn-backend-tests --no-default-features --features "cuda,fusion,std" --test tensor fusion::nhwc_relayout`

**Expected behavior**
All tests pass.

 - OS: Linux
 - Version: main branch

ThierryCantin-Demers:
### Checklist

> `cargo run-checks` has not been run on this branch. The targeted testing that was
> run is listed under Testing below, including the CUDA suite from the issue, which
> is the one that reproduces the bug.

### Related Issues/PRs

Closes #5388.

### Changes

CUDA pitches its allocations, padding rows so they align. A tensor backed by such a
buffer has strides that leave gaps, and the fusion launcher was not accounting for
that when it picked a block's reference layout.

Under `RefLayoutSetting::Any`, a candidate input became the block's *concrete*
reference unconditionally, adopting its strides as the layout every other tensor in
the block is indexed against. A reference tagged `IsRef` / `SameAsRef` has its buffer
offset read straight off the flat thread position, while the launch is sized from the
logical element count. Those two agree only when the reference fills its buffer
without gaps. Given a padded reference they disagree, and the kernel walks a prefix of
the buffer, so elements past the first row of padding are read and written at the
wrong offsets.

The fix adopts a candidate as a concrete reference only when its layout is gap-free,
and otherwise falls back to a virtual contiguous reference, which indexes every tensor
through full stride multiplication and is correct regardless of padding.

The new predicate is `is_dense` rather than the existing `is_contiguous` on purpose.
The kernel recovers coordinates from a position with `pos / ref_strides[i] % shape[i]`,
a mixed-radix decomposition that is valid for *any* permutation of contiguous strides,
not just row-major ones. NHWC relayout deliberately produces dense-but-permuted
strides, so gating on `is_contiguous` would have rejected the layouts this feature
exists to produce and sent them all down the slow path. `is_dense` also rejects
broadcast dimensions (stride 0 on a dimension larger than 1), which could previously
be adopted as a concrete reference here.

### Testing

Run on an RTX 4070 Ti SUPER:

```
cargo test -p burn-backend-tests --no-default-features --features "cuda,fusion,std" --test tensor fusion::nhwc_relayout
```

All 7 tests pass. Six of them fail on `main`, which is the set reported in #5388.

Also run:

- `cargo test -p burn-std --lib tensor::tests`, covering `is_dense` against contiguous,
  permuted, pitched-row, unit-dimension and rank-mismatch inputs.
- `cargo clippy -p burn-std --lib` and `cargo fmt -p burn-std -- --check`, both clean.

`is_dense` returns `false` for a shape and strides of differing length rather than
indexing past the end of the shorter slice. That routes the caller to the virtual
reference layout, which stays correct for any input, where truncating to the shorter
slice would have reported a malformed layout as dense and picked the concrete path.

### DIFF
--- a/crates/burn-cubecl-fusion/src/engine/launch/output.rs
+++ b/crates/burn-cubecl-fusion/src/engine/launch/output.rs
@@ -18,7 +18,7 @@ use burn_ir::{TensorId, TensorIr};
 use burn_std::Shape;
 use burn_std::{
     Strides,
-    tensor::{ReshapeAction, contiguous_strides, is_contiguous, reshape_action},
+    tensor::{ReshapeAction, contiguous_strides, is_contiguous, is_dense, reshape_action},
 };
 use cubecl::{Runtime, client::ComputeClient};
 
@@ -269,7 +269,16 @@ impl<'a, R: Runtime> OutputPlanner<'a, R> {
                     };
 
                     match block.settings.ref_layout {
-                        RefLayoutSetting::Any => set_ref_as_concrete(block_plan),
+                        RefLayoutSetting::Any => {
+                            // A padded reference would make the kernel walk a prefix of the
+                            // buffer: the launch is sized from the logical element count while a
+                            // position indexes the buffer.
+                            if is_dense(&reference.global_ir.shape, &reference.handle.strides) {
+                                set_ref_as_concrete(block_plan)
+                            } else {
+                                set_ref_as_virtual(block_plan)
+                            }
+                        }
                         RefLayoutSetting::SameAsBlock { .. } => {
                             // Skip set ref.
                         }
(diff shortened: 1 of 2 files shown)


laggui (approved):

laggui:
Validated locally

PR #5398, opened 2026-08-20
chore: tracel github actions v11

syl20bnr:
Add timeouts on job steps to avoid long running steps doing nothing.


laggui (approved):
Whoops, missed it when merging

PR #5406, opened 2026-08-21
Perf/fusion layout propagation

nathanielsimard:


laggui (commented):
Small edge case: a valid zero stride on a singleton reference dimension is later used as a divisor during fused indexing.

I noticed this potential bug and tested locally. It seems CubeCL CPU is the only one that deterministically fails with SIGFPE (division by zero).

/edit: pushed the regression test + fix. Otherwise LGTM!

PR #5407, opened 2026-08-21
fix(optim): keep tensor data in optimizer records

### LINKED ISSUE #5379: burn-optim: OptimizerRecord should hold TensorData, not burn_pack::Tensor
## Summary

`OptimizerRecord` uses the on-disk transport type as its in-memory representation:

```rust
// crates/burn-optim/src/optim/module/record/mod.rs:23
pub(crate) tensors: Vec<burn_pack::Tensor>,
```

`ModuleRecord` does the opposite, and is the better model:

```rust
// crates/burn-core/src/store/mod.rs:85
tensors: Vec<RecordTensor>,     // holds TensorData
```

It builds `burn_pack::Tensor` only at the save boundary, in `pack_tensors()`.

The consequence is a full round trip through the transport type inside burn-optim's own memory: `to_record` destructures `TensorData` into `Tensor::new` (`module_optimizer.rs:223`, `lbfgs.rs:545`), and `load_record` immediately reassembles it. Nothing is serialized in between.

## Why it matters

**1. It forces a `Send` chain across three crates.** `burn-train`'s `Checkpoint` is `Sized + Send + 'static` (`crates/burn-train/src/checkpoint/base.rs:33`) and `OptimizerRecord` implements it, so `burn_pack::Tensor: Send`. After #5219 `Tensor` can hold a provider closure, so that propagates: `Tensor: Send` → `ByteSource: Send + Sync` → `DataSource`/`TensorSnapshot: Send + Sync`. Every snapshot construction site in burn-store now carries a bound that exists because of a burn-train constraint reached through a type burn-store never sends anywhere. The pytorch and safetensors load paths pay it too, and neither touches burn-pack or a thread.

**2. It puts two `expect`s in load paths.** `load_record` has to reach into pack tensors that only ever hold resident bytes:

```rust
// lbfgs.rs and module_optimizer.rs
let (name, dtype, shape, _, bytes) = tensor.into_parts().expect("record tensors are resident");
```

The invariant does hold today — `tensors` is `pub(crate)`, both producers use `Tensor::new`, and `Reader` bounds-checks descriptors so a malformed `.bpk` fails earlier. But it is enforced by field privacy, which a future refactor can break silently, and the type now admits a state the domain forbids.

**3. It was the origin of a real performance bug.** During #5219 review, a change to use `Tensor::bytes()` at these sites deep-copied every optimizer state tensor on load, because `Bytes::clone` only shares when the allocation controller implements `duplicate()` and these carry plain heap buffers. Caught and reverted, but the unpacking is what created the opportunity.

## Proposal

Mirror `ModuleRecord`: hold `(name, param_id, TensorData)` (or a small `RecordTensor`-alike), build `burn_pack::Tensor` inside `into_writer()`, destructure in `from_reader()`.

That:

- terminates the `Send` chain at `TensorData`, so the bound on `TensorSnapshot`/`DataSource` can be reconsidered on its own merits
- makes "record tensors are resident" a type-level fact rather than a comment, deleting both `expect`s
- removes the in-memory round trip through the transport type
- stops `burn_pack::Tensor` leaking through `OptimizerRecord`'s shape

Laziness is not lost: a file-backed `Bytes` inside a `TensorData` stays unread until access, same as today.

Roughly 30 lines in `crates/burn-optim/src/optim/module/record/mod.rs` plus the two `load_record` sites.

Context: #5219. Flagged independently by three review passes on that work.

original4422:
### Checklist

- [x] Made sure the book is up to date with changes in this PR (no documentation changes are needed for this internal refactor).

### Related Issues/PRs

Fixes #5379.

### Changes

`OptimizerRecord` previously stored `burn_pack::Tensor` transport objects internally. Module optimizers and LBFGS therefore converted state to the transport representation in memory and later required those tensors to be resident before converting them back.

This change adds a crate-private optimizer record tensor that retains the name, optional parameter id, and `TensorData`. Burnpack tensors are now constructed only when writing and are converted back to `TensorData` immediately when reading. Module optimizer and LBFGS state move that data directly, removing the resident-tensor expectations without changing the public API or dependencies.

Regression tests cover both the in-memory serialization boundary and file-backed burnpack round trips.

### Testing

- `CARGO_INCREMENTAL=0 cargo test -p burn-optim` (103 unit tests passed; 1 doctest passed and 3 existing doctests ignored)
- Focused optimizer-record, module-optimizer, and LBFGS round-trip tests
- `CARGO_INCREMENTAL=0 cargo run-checks`

### DIFF
--- a/crates/burn-optim/src/optim/lbfgs.rs
+++ b/crates/burn-optim/src/optim/lbfgs.rs
@@ -3,12 +3,12 @@
 use burn_core as burn;
 
 use super::GradientsParams;
-use crate::{LearningRate, OptimizerRecord};
+use crate::{LearningRate, OptimizerRecord, RecordTensor};
 use crate::{RecordState, StateSink, StateSource};
 use burn::config::Config;
 use burn::module::{AutodiffModule, Module, ModuleMapper, ModuleVisitor, Param};
 use burn::store::RecordError;
-use burn::tensor::{Bytes, Device, Tensor, TensorData};
+use burn::tensor::{Bytes, Device, Tensor};
 use serde::{Deserialize, Serialize};
 
 use alloc::vec;
@@ -541,8 +541,10 @@ impl LBFGS {
         let tensors = sink
             .tensors
             .into_iter()
-            .map(|(name, data)| {
-                burn_pack::Tensor::new(name, data.dtype, data.shape, None, data.bytes)
+            .map(|(name, data)| RecordTensor {
+                name,
+                param_id: None,
+                data,
             })
             .collect();
         let scalars = sink.scalars.into_iter().collect();
@@ -562,9 +564,7 @@ impl LBFGS {
         let device = Device::default();
         let mut source = StateSource::new(record.scalars);
         for tensor in record.tensors {
-            let (name, dtype, shape, _, bytes) =
-                tensor.into_parts().expect("record tensors are resident");
-            source.insert_tensor(name, TensorData::from_bytes(bytes, shape, dtype));
+            source.insert_tensor(tensor.name, tensor.data);
         }
         if let Some(state) = LBFGSState::state_unflatten("", &mut source, &device) {
             self.state = state;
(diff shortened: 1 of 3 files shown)


original4422:
Hi @louisfd, when you have a chance, could you please review this PR and let me know if any changes are needed? Thanks!

laggui (approved):
LGTM.

This brings `OptimizerRecord` in line with `ModuleRecord`, and also makes the `Send + Sync` bounds previously added for deferred tensors unnecessary. I'll address that cleanup separately.

PR #5414, opened 2026-08-21
fix(vision): keep even filter2d output size

### LINKED ISSUE #5371: burn-vision: filter2d returns a larger image than the input for even-sized kernels
**Describe the bug**

`filter2d` in `crates/burn-vision/src/filter.rs` pads `k / 2` on both sides of each spatial axis:

```rust
let images = images.pad((kw / 2, kw / 2, kh / 2, kh / 2), border);
```

That adds `2 * (k / 2)` pixels, which is `k` for an even kernel and `k - 1` for an odd one. The convolution that follows (stride 1, dilation 1, no conv padding) consumes `k - 1`. So the output keeps the input size for odd kernels, but is **one pixel wider and taller for even ones**.

The function's own doc comment states the intended behaviour, and it disagrees with the code in two places:

> The output keeps the input's spatial size for odd kernel dimensions; even dimensions pad **asymmetrically toward the top/left**.

> \# Returns — The filtered batched images, **same shape as the input**.

The asymmetric padding described there was never implemented — the padding is symmetric.

**To Reproduce**

```rust
let images = Tensor::<4>::from([[[[1., 2., 3., 4.], [5., 6., 7., 8.],
                                  [9., 10., 11., 12.], [13., 14., 15., 16.]]]]);
let kernel = Tensor::<2>::from([[1., 1.], [1., 1.]]);
let out = filter2d(images, kernel, PadMode::Edge);
assert_eq!(out.dims(), [1, 1, 4, 4]);
```

```
  left: [1, 1, 5, 5]
 right: [1, 1, 4, 4]
```

Reproduced on `main` at b6e27bdca, via `cargo test -p burn-vision --lib --no-default-features --features "flex,std"`.

**Expected behavior**

`[1, 1, 4, 4]`, per the doc comment.

**Additional context**

Introduced in #5339 (07b7b5706). `BoxBlur` and `MedianBlur` are unaffected — both assert an odd `kernel_size` — but `filter2d` is public and reachable directly. The existing tests only exercise odd kernels, which is why this wasn't caught.

The fix looks like one line, matching the documented intent:

```rust
let images = images.pad((kw / 2, (kw - 1) / 2, kh / 2, (kh - 1) / 2), border);
```

With that change all 24 existing `burn-vision` lib tests still pass. Happy to open a PR if the documented "same shape, pad toward top/left" semantics are what you want — the alternative reading is that the docs should change instead and even kernels are simply meant to grow the image.

Soundcreates:
### Checklist

### Related Issues/PRs

- Fixes https://github.com/tracel-ai/burn/issues/5371

### Changes

`filter2d` previously padded `kw / 2` and `kh / 2` on both sides. For even kernels, that adds one pixel of excess total padding per affected dimension, producing an output larger than the input. The fix pads a total of `kernel_size - 1` pixels per dimension, placing the extra pixel on the top/left as required for same-size output. Odd and rectangular kernels retain their existing behavior.

### Testing

- `cargo test -p burn-vision --lib filter` (10 passed)
- `cargo fmt --all`
- `cargo run-checks` (passed; backend suite: 1327 passed, 0 failed; existing unrelated warnings remain)

### DIFF
--- a/crates/burn-vision/src/filter.rs
+++ b/crates/burn-vision/src/filter.rs
@@ -41,9 +41,13 @@ pub fn filter2d(images: Tensor<4>, kernel: Tensor<2>, border: PadMode) -> Tensor
         .expand([channels, 1, kh, kw])
         .cast(FloatDType::from(images.dtype()));
 
-    // "Same"-size output: pad half the kernel on each side. The `(l, r, t, b)`
-    // tuple pads `(t, b)` along H and `(l, r)` along W.
-    let images = images.pad((kw / 2, kw / 2, kh / 2, kh / 2), border);
+    // "Same"-size output: the total padding is one less than the kernel size,
+    // with any extra pixel placed on the top/left. The `(l, r, t, b)` tuple
+    // pads `(t, b)` along H and `(l, r)` along W.
+    let images = images.pad(
+        (kw / 2, (kw - 1) - kw / 2, kh / 2, (kh - 1) - kh / 2),
+        border,
+    );
 
     conv2d(
         images,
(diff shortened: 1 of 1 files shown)


laggui (approved):
Thanks for the fix!

/edit: CI failures are unrelated to these changes

PR #5423, opened 2026-08-24
fix(cubecl): restrict GEMV CPU autotune priority to 1D vector workloads

### LINKED ISSUE #5419: burn-cpu: stack overflow in internal worker thread on tall [N,3] x [3,3] matmul (N >= ~5.6M)
**Describe the bug**

On the CubeCL CPU backend (`burn-cpu`), a tall `[N, 3] x [3, 3]` f32 `matmul` aborts the process with a stack overflow once `N` reaches roughly 5.6M rows. The failure is not a panic that can be caught: the overflowing thread is an internal worker (`thread '<unknown>'`), so the process dies with `fatal runtime error: stack overflow, aborting` (SIGABRT).

Two details that may help narrow it down:

- The crash is preceded by ~170s of work. The same shape at `N = 5,000,000` completes in 2.3s, so the runtime is grossly superlinear right before the abort, which suggests something recursive/exploding in the matmul launch path rather than a merely large allocation.
- Because the overflow happens on an internally spawned thread, users cannot work around it by enlarging their own stack. I verified this by running the same call inside a thread built with `stack_size(1 GiB)`: it still aborts, and the reported thread is still `<unknown>`.

The `flex` backend runs the identical workload at `N = 10,000,000` in 0.21s, so this looks specific to the CubeCL CPU path.

**To Reproduce**

`Cargo.toml`:

```toml
[dependencies]
burn = { version = "0.21", default-features = false, features = ["std", "cpu", "flex"] }
```

`src/main.rs`:

```rust
use burn::backend::{Cpu, Flex};
use burn::tensor::{backend::Backend, Tensor, TensorData};
use std::time::Instant;

fn run<B: Backend>(label: &str, n: usize, device: &B::Device) {
    let lhs = Tensor::<B, 2>::from_data(TensorData::new(vec![1.0f32; n * 3], [n, 3]), device);
    let rhs = Tensor::<B, 2>::from_data(
        TensorData::new(vec![0.8f32, -0.6, 0.0, 0.6, 0.8, 0.0, 0.0, 0.0, 1.0], [3, 3]),
        device,
    );

    println!("[{label}] n = {n}: matmul [{n},3] x [3,3]");
    let t0 = Instant::now();
    let data = lhs.matmul(rhs).into_data();
    // Touch the bytes: ops are lazy, so this is what forces execution.
    let first = data.as_slice::<f32>().expect("f32")[0];
    println!(
        "[{label}] n = {n}: OK in {:.2}s (first element = {first})",
        t0.elapsed().as_secs_f64()
    );
}

fn main() {
    let backend = std::env::args().nth(1).unwrap_or_else(|| "cpu".into());
    let n: usize = std::env::args()
        .nth(2)
        .and_then(|a| a.parse().ok())
        .unwrap_or(5_592_405);

    match backend.as_str() {
        "cpu" => run::<Cpu>("cpu", n, &Default::default()),
        "flex" => run::<Flex>("flex", n, &Default::default()),
        other => panic!("unknown backend '{other}'"),
    }
}
```

Results on my machine:

| Command | Result |
|---|---|
| `cargo run --release -- cpu 5000000` | OK, 2.28s |
| `cargo run --release -- cpu 5592405` | **abort: stack overflow** (after ~174s) |
| `cargo run --release -- cpu 6000000` | **abort: stack overflow** |
| `cargo run --release -- cpu 10000000` | **abort: stack overflow** |
| `cargo run --release -- flex 10000000` | OK, 0.21s |

Failure output:

```
[cpu] n = 5592405: matmul [5592405,3] x [3,3]

thread '<unknown>' (1290331) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The threshold sits somewhere in `5,000,000 < N < 5,592,405` rows; I did not bisect it exactly. (I initially suspected the total element count crossing 2^24, but `N = 5,592,405` is 16,777,215 elements — just *under* 2^24 — and it still aborts, so that hypothesis is wrong.)

**Expected behavior**

The matmul completes, as it does on `flex` and as it does on the same backend at `N = 5,000,000`.

**Desktop:**
 - OS: Linux 6.12.94-123.192.amzn2023.x86_64 (x86_64)
 - CPU: AMD EPYC 7R13 (96 threads)
 - rustc: 1.98.0 (88d9e12ae 2026-08-18)
 - burn: 0.21.0, features `["std", "cpu", "flex"]`

**Additional context**

Also reproduces on **0.22.0-pre.2**, so it is present on the current pre-release line. The 0.22 port of the same program (no backend generics, `Tensor<2>` + `Device::cpu()`) aborts identically at `N = 10,000,000`, with the device reported as `Device<Cpu(CpuDevice)>`:

```rust
use burn::prelude::{Device, Tensor, TensorData};

let device = Device::cpu();
let lhs = Tensor::<2>::from_data(TensorData::new(vec![1.0f32; n * 3], [n, 3]), &device);
let rhs = Tensor::<2>::from_data(
    TensorData::new(vec![0.8f32, -0.6, 0.0, 0.6, 0.8, 0.0, 0.0, 0.0, 1.0], [3, 3]),
    &device,
);
let data = lhs.matmul(rhs).into_data();
let _ = data.as_slice::<f32>().unwrap()[0];
```

One small note from porting that repro: on 0.22 `into_data()` alone returned in ~0ms without materializing anything; execution only happened once I read the bytes via `as_slice`. That is why the reproducer touches the first element — without it the benchmark silently measures nothing.

Encountered while building a point-cloud library where transforming an N-point cloud is exactly this `[N,3] x [3,3]` shape; I switched the CPU path to `flex` as a workaround.

Sadik00789:
Closes #5419

### Description
Fixes a worker thread stack overflow abort (`SIGABRT`) when executing tall matrix multiplications on CPU (e.g., `[N, 3] x [3, 3]` where `N >= 5.6M`).

### Root Cause
In `crates/burn-cubecl/src/kernel/matmul/tune/base.rs`, the `gemv` autotune group returned `PRIORITY_MAX` whenever `num_cpu_cores.is_some()`, prior to verifying `key.analysis.kind`. Because `num_cpu_cores` is always populated on CPU backends, general $M \times N$ matrix multiplications with extreme aspect ratios were assigned max priority for 1D vector routines (`SimpleVecMat`, `DoubleVecMat`, `GemvUnitPerpendicular`). 

For `[5.6M, 3] x [3, 3]`, this generated $5.6\text{M}$ independent 1-element micro-cubes along the first dimension, exhausting the 64 MiB call stack during task partition bisection on the internal CPU worker thread.

### Solution
* Gated the `num_cpu_cores` priority elevation inside the `gemv` autotune group so it only activates when `key.analysis.kind` is strictly `MatmulKind::MatVec` or `MatmulKind::VecMat`.
* General tall matrices now route to standard tiled GEMM routines (`CpuGemmStrategy`), executing cleanly in $\approx 3.4$s without call stack exhaustion.
* Added regression test `matmul_tall_matrix_should_not_overflow_stack` in `crates/burn-backend-tests/tests/cubecl/matmul.rs`.

### DIFF
--- a/crates/burn-backend-tests/tests/cubecl/matmul.rs
+++ b/crates/burn-backend-tests/tests/cubecl/matmul.rs
@@ -76,3 +76,28 @@ fn matmul_transposed_operands_single_row_should_match_reference() {
     out.into_data()
         .assert_approx_eq::<FloatElem>(&out_ref.into_data(), Tolerance::rel_abs(1e-4, 1e-5));
 }
+
+// Tall matrix matmul with extreme aspect ratio ([N, 3] x [3, 3]) should not
+// misclassify as GEMV on CPU or cause worker thread stack overflow.
+// https://github.com/tracel-ai/burn/issues/5419
+#[test]
+fn matmul_tall_matrix_should_not_overflow_stack() {
+    let device = Device::default();
+
+    let n = 5_592_405;
+    let lhs: Tensor<2> = Tensor::ones(Shape::new([n, 3]), &device);
+    let rhs: Tensor<2> = Tensor::from_floats(
+        [[0.8, -0.6, 0.0], [0.6, 0.8, 0.0], [0.0, 0.0, 1.0]],
+        &device,
+    );
+
+    let out = lhs.matmul(rhs);
+
+    // Verify row calculation without materializing the full 5.6M row tensor comparison
+    let first_row = out.slice([0..1, 0..3]);
+    let expected = Tensor::<2>::from_floats([[1.4, 0.2, 1.0]], &device);
+
+    first_row
+        .into_data()
+        .assert_approx_eq::<FloatElem>(&expected.into_data(), Tolerance::rel_abs(1e-4, 1e-5));
+}
--- a/crates/burn-backend-tests/tests/cubecl/matmul.rs
+++ b/crates/burn-backend-tests/tests/cubecl/matmul.rs
@@ -80,6 +80,7 @@ fn matmul_transposed_operands_single_row_should_match_reference() {
 // Tall matrix matmul with extreme aspect ratio ([N, 3] x [3, 3]) should not
 // misclassify as GEMV on CPU or cause worker thread stack overflow.
 // https://github.com/tracel-ai/burn/issues/5419
+#[cfg(feature = "cpu")]
 #[test]
 fn matmul_tall_matrix_should_not_overflow_stack() {
     let device = Device::default();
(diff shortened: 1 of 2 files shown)


laggui (approved):
Looks good overall.

The unconditional CPU GEMV priority predates the dedicated CPU GEMM strategy, so restricting it to actual MatVec/VecMat workloads is the right dispatch change. I validated [N,3] @ [3,3] and added a test for [N,3] @ [3,1] at the reported scale.


One correction: I couldn't reproduce the stack overflow with the old selector on my machine, nor verify the proposed cause. CubeCL CPU schedules tasks per cube unit and iterates over the cube grid inside the generated kernel, so the explanation involving scheduler task partitioning does not appear to match the current implementation.

Nonetheless, the new fast-path selection results in faster execution time for these tests.

PR #5429, opened 2026-08-24
perf(conv): route strided pointwise convolutions through the matmul path

ThierryCantin-Demers:
A strided pointwise convolution had no matmul path. `check_pointwise` requires `stride == 1`, so `conv_im2col_1x1` declined every strided 1x1, and nothing else could take them either: the remaining forward candidates need accelerated tiles. On CPU that left `conv_direct` as the only implementation, and autotune logs exactly one candidate for those keys.

A strided pointwise reads a regular subset of its input, so multiplying the spatial strides gives a view holding precisely the positions it reads. The contiguous copy `reshape_input` already performs then gathers them, and the convolution stays the single matmul it always was. No new kernel.

The gradient paths keep the strict `check_pointwise`; only the forward gets `check_pointwise_strided`, which is the same check minus the stride. Padding still declines, since padded positions are not in the input for a view to reach.

## Measurements

ResNet-50, batch 1, Ryzen 5700X, burn's cubecl CPU backend, fusion and autotune on. Both arms measured on an idle machine with autotune re-run, minutes apart:

| | before | after | |
|---|---|---|---|
| whole forward | 236.71 ms | 219.08 ms | **-7.4%** |
| layer4 group | 49.34 ms | 34.10 ms | -31% |
| downsample 1024→2048 | 15.49 ms | 2.73 ms | **-5.7x** |
| downsample 512→1024 | 4.18 ms | 2.14 ms | -2.0x |
| downsample 256→512 | 2.52 ms | 2.33 ms | -8% |

The whole model re-measured within 1% at the end of each run, so the machine held still.

Autotune's own candidate timings on the 1024→2048 key agree: `conv_im2col_1x1` 2.657 ms against `conv_direct` 15.592 ms.

The gain is shape-dependent, and the same kernel behaves very differently across machines: on a laptop `conv_direct` already tuned that shape at 3.50 ms rather than 15.14, so the win there is 1.5x rather than 5.7x. Autotune picks per shape, so a case where the gather costs more than it saves keeps the direct kernel.

## Correctness

`test_conv2d_1x1_strided` is new: the suite had no 1x1 with a stride, so nothing exercised this path. Run under `autotune-checks`, which compares every candidate's output, the whole `conv2d` module suite passes on the cpu backend (20 tests). Golden-logit parity on a pretrained ResNet-50 is unchanged on both cubecl-cpu and flex.

## One caveat worth knowing

This makes an existing candidate viable rather than adding one, so the tunable set's checksum does not change and **cached autotune results stay valid**. My first benchmark after the change showed no improvement at all, because the cache was still serving the `conv_direct` winner. Anyone upgrading past this keeps the slow kernel until their autotune cache is cleared or re-tuned.

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/module/conv2d.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/module/conv2d.rs
@@ -42,6 +42,31 @@ fn test_conv2d_simple() {
 }
 
 #[test]
+fn test_conv2d_1x1_strided() {
+    let test = Conv2dTestCase {
+        batch_size: 1,
+        channels_in: 2,
+        channels_out: 2,
+        kernel_size_1: 1,
+        kernel_size_2: 1,
+        padding_1: 0,
+        padding_2: 0,
+        stride_1: 2,
+        stride_2: 2,
+        dilation_1: 1,
+        dilation_2: 1,
+        groups: 1,
+        height: 4,
+        width: 4,
+    };
+
+    test.assert_output(TestTensor::from([[
+        [[16., 18.], [24., 26.]],
+        [[49., 59.], [89., 99.]],
+    ]]));
+}
+
+#[test]
 fn test_conv2d_simple_implicit() {
     let test = Conv2dTestCase {
         batch_size: 1,
--- a/crates/burn-cubecl/src/kernel/conv/im2col.rs
+++ b/crates/burn-cubecl/src/kernel/conv/im2col.rs
@@ -1,5 +1,8 @@
 use burn_backend::cubecl::dtype_to_storage_type;
-use burn_backend::{DType, ops::ConvOptions};
+use burn_backend::{
+    DType,
+    ops::{ConvOptions, conv::calculate_conv_output_sizes},
+};
 use burn_std::{Metadata, Shape};
 use core::iter;
 use cubecl::{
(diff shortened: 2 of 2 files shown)


ThierryCantin-Demers:
## Re-measured on the current base

The numbers below replace the originals, which were taken before this branch was rebased onto main. Same machine, idle, both arms with autotune re-run, minutes apart. ResNet-50, batch 1, Ryzen 5700X, burn's cubecl CPU backend, fusion and autotune on.

| | before | after | |
|---|---|---|---|
| whole forward | 177.28 ms | 164.46 ms | **-7.2%** |
| layer4 group | 43.82 ms | 30.15 ms | -31% |
| downsample 1024→2048 | 14.29 ms | 2.63 ms | **-5.4x** |
| downsample 512→1024 | 5.19 ms | 2.63 ms | -2.0x |
| downsample 256→512 | 2.25 ms | 2.87 ms | +0.6 ms |
| downsample 64→256 (stride 1) | 1.174 ms | 1.174 ms | unchanged |

The whole model re-measured within 0.7% at the end of each run, so the machine held still.

Two things worth reading off this. The stride-1 downsample is identical to three decimal places, which is the control: it never touches this path. And the smallest strided case, 256→512, is 0.6 ms *slower*, where the gather costs more than the matmul saves; autotune keeps whichever wins per shape, so that case simply stays on the direct kernel in a normal run with tuning enabled.

The whole model is also much faster than the original measurement on both arms (177 ms against 237 ms), which is the newer base rather than anything here.

laggui (approved):

PR #5486, opened 2026-08-25
fix(quant): serialize ue8m0 scales instead of refusing them

nathanielsimard:


crates/burn-std/src/tensor/quantization.rs
```diff
@@ -15,6 +15,7 @@ pub const QPARAM_ALIGN: usize = core::mem::align_of::<f32>();
 use alloc::vec::Vec;
 use core::any::TypeId;
 use cubecl_common::e4m3;
+use cubecl_common::quant::scheme::{f32_to_ue8m0, ue8m0_to_f32};
```
laggui, line 18: Looks like this is linked/blocked by https://github.com/tracel-ai/cubecl/pull/1557/?

PR #5493, opened 2026-08-26
Fix/frozen layers

nathanielsimard:
## Freezing reaches the layers that have no parameters

Freezing a subtree with `no_grad`/`freeze_group` is expressed by clearing `require_grad`
on parameters, so layers with no parameters of their own — dropout, gaussian noise, RReLU —
had nothing to read and kept behaving as if training. Since a frozen subtree stays on the
training device (that's where the rest of the graph lives), the device alone can't tell the
two apart, and a frozen dropout kept perturbing activations it wasn't supposed to touch.

This adds a `TrainingFlag`: a small `Module` carrying a `ParamId`, held by those three
layers and cleared by `no_grad`, `freeze_group` and `valid` (and set by `from_inner`/`train`).
It carries an id so a `ParamGroup` built from a subtree names it the way it names a parameter.
`BatchNorm` needs no flag — it now reads its own `gamma` and takes the inference path when
frozen, so it stops recomputing batch statistics and writing them into its running state.

Breaking: `Dropout`, `GaussianNoise` and `RRelu` gain a public `training` field, so struct
literals need updating; `Config::init()` is unaffected.

### DIFF
--- a/crates/burn-core/src/module/param/training.rs
+++ b/crates/burn-core/src/module/param/training.rs
@@ -21,15 +21,16 @@ use crate::module::{
 ///   trained, and a frozen dropout still perturbing its activations is not what that means. This
 ///   is the case the device alone cannot answer, because a frozen subtree stays on the training
 ///   device — that is where the rest of the graph lives.
-/// - [`valid`](AutodiffModule::valid) clears it, and [`from_inner`](AutodiffModule::from_inner) —
-///   which is what [`train`](Module::train) is — sets it.
+/// - [`freeze_group`](Module::freeze_group) clears it for the layers the group names, and
+///   [`unfreeze_group`](Module::unfreeze_group) sets it again. A group matches by
+///   [`ParamId`](crate::module::ParamId) or by module path, and a flag carries both, so the
+///   group-scoped traversal reaches it the way it reaches a parameter.
+/// - [`valid`](AutodiffModule::valid) clears it for the duration of the inference module it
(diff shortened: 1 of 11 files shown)


crates/burn-core/src/module/param/training.rs
```diff
@@ -0,0 +1,188 @@
...
+///   [`ParamId`](crate::module::ParamId) or by module path, and a flag carries both, so the
+///   group-scoped traversal reaches it the way it reaches a parameter.
+/// - [`valid`](AutodiffModule::valid) clears it for the duration of the inference module it
+///   builds, and [`from_inner`](AutodiffModule::from_inner) — which is what
+///   [`train`](Module::train) is — restores whatever the caller last asked for.
+///
+/// It starts set, because a module is built before it is placed and the alternative default would
+/// silently disable dropout for every run that never calls `train()`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TrainingFlag {
```
laggui, line 35: Reporting from our offline discussion: I feel like `TrainingFlag` conflates two concerns: parameter-level indentity/trainability for traversal and group freezing, and the layer's current runtime state. The fact that it owns a `ParamId` and implements `Module` suggests the underlying abstraction is broader.

I proposed a generalization to `Param<Flag>` in #5498: `Param` provides identity, traversal, grouping, and persistent trainability, while `Flag` represnts the current effective boolean state. The field name in the modules (e.g. batchnorm's `training`) provides the layer-specific meaning.

PR #5494, opened 2026-08-26
fix(store): preserve extensionless burnpack paths

### LINKED ISSUE #5482: burn-store: BurnpackStore::auto_extension(false) is ignored on save
## Summary

`BurnpackStore::auto_extension(false)` is documented as "uses the exact path provided without modification", but on save the file still gets `.bpk` appended. `burn_pack::Writer::write_to_file_atomic` appends the canonical extension unconditionally for a path that has none, and it has no way to know the store was told not to.

## Detail

`BurnpackStore::process_path` (`crates/burn-store/src/burnpack.rs:318-333`) honours the flag and returns the path untouched. It then hands that path to `Writer::write_to_file_atomic`, whose `resolve_path` (`crates/burn-pack/src/writer.rs:210-216`) appends the extension anyway:

```rust
fn resolve_path(path: &Path) -> std::path::PathBuf {
    if path.extension().is_none() {
        path.with_extension(crate::EXTENSION)
    } else {
        path.to_path_buf()
    }
}
```

The documented example in `burnpack.rs:226-234` therefore does not hold:

```rust
// With auto_extension disabled
let store = BurnpackStore::from_file("model")
    .auto_extension(false);  // -> "model"
```

Load happens to work by accident: `Reader::from_file` (`crates/burn-pack/src/reader.rs:63-67`) also appends `.bpk` when the path has no extension and does not exist, so it finds the file the writer actually created. The round trip succeeds while both ends ignore what the caller asked for.

## Reproduction

```rust
let path = dir.path().join("model");   // no extension
let mut store = BurnpackStore::from_file(&path).auto_extension(false);
model.save_into(&mut store).unwrap();

println!("exists as requested: {}", path.exists());
println!("exists with .bpk:    {}", dir.path().join("model.bpk").exists());
```

```
exists as requested: false
exists with .bpk:    true
```

## Suggested fix

Same seam as the companion issue on the `overwrite` guard, and worth fixing together. Either give `burn_pack::Writer` a path form that is taken literally (so the store can honour the flag), or drop `auto_extension` from `BurnpackStore` and document that `.bpk` is always appended to an extension-less path.

Note `Reader::from_file`'s appending is conditional on the file not existing, so it is a fallback rather than a rewrite; the writer's is unconditional. Making them agree would help on its own.

rioyu123:
### Checklist

- [x] Made sure the book is up to date with changes in this PR (no book changes required).

### Related Issues/PRs

Fixes #5482.

### Changes

`BurnpackStore` resolved its configured path first, but `burn-pack` applied its own extension fallback again. As a result, `.auto_extension(false)` still wrote to `model.bpk`, the overwrite guard checked a different path from the writer, and an exact load could fall back to the sibling `.bpk` file.

- Preserve the existing default extension behavior.
- Allow `Writer` to keep an extensionless path for both direct and atomic writes.
- Add an exact-path `Reader` entry point so higher-level callers do not reapply fallback behavior.
- Make `BurnpackStore` apply its extension policy exactly once for save, overwrite checks, and load.
- Use the canonical extension constant and include the requested path in open errors.
- Add regressions for default behavior, exact direct/atomic writes, exact loads, overwrite protection, and end-to-end store round trips.

### Testing

- `cargo run-checks` (Rust 1.96.0)
- `cargo test -p burn-pack -p burn-store`
- `cargo check -p burn-store --no-default-features`
- `cargo clippy -p burn-pack -p burn-store --all-targets -- -D warnings`

### DIFF
--- a/crates/burn-pack/src/lib.rs
+++ b/crates/burn-pack/src/lib.rs
@@ -106,7 +106,8 @@
 //!
 //! ## Feature Flags
 //!
-//! - `std`: Enables file I/O ([`Reader::from_file`] / [`Writer::write_to_file`]) (default)
+//! - `std`: Enables file I/O ([`Reader::from_file`], [`Reader::from_file_exact`],
+//!   [`Writer::write_to_file`], and [`Writer::write_to_file_atomic`]) (default)
 
 extern crate alloc;
 
--- a/crates/burn-pack/src/reader.rs
+++ b/crates/burn-pack/src/reader.rs
@@ -66,7 +66,23 @@ impl Reader {
             path.to_path_buf()
         };
 
-        let mut file = File::open(&path).map_err(io_err)?;
+        Self::from_resolved_file(&path)
+    }
+
+    /// Load a pack from a file using the exact path provided.
+    ///
+    /// Unlike [`from_file`](Self::from_file), this never appends [`crate::EXTENSION`] when the
+    /// requested path has no extension and does not probe for a fallback path. It is useful when
+    /// a higher-level caller has already applied its own path policy.
+    #[cfg(feature = "std")]
+    pub fn from_file_exact<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+        Self::from_resolved_file(path.as_ref())
+    }
+
+    #[cfg(feature = "std")]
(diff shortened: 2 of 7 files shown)


laggui (approved):
LGTM!

PR #5495, opened 2026-08-26
feat(optim): add LAMB optimizer

Mikyx-1:
## Summary
- add a bias-corrected Layer-wise Adaptive Moments optimizer
- support optional trust-ratio disabling, weight decay, gradient clipping, device migration, and burnpack state persistence
- validate updates against an independent PyTorch reference, including zero-norm and checkpoint-resume cases

## Testing
- cargo test -p burn-optim lamb (4 passed)
- cargo fmt --all -- --check


laggui (approved):
The implementation looks correct and consistent with other optimizers.

Thanks!

PR #5496, opened 2026-08-26
fix(store): return deserialization error instead of panicking in nested Deserializer

### LINKED ISSUE #5484: burn-store: nested deserializer panics on malformed pickle data instead of returning an error
## Summary

`PytorchReader::load_config` returns `Result<D>`, but the deserializer behind it unwraps on values that came out of the file. A config field whose pickle type does not match the Rust type panics instead of returning a deserialization error.

## Detail

Every scalar visitor in `crates/burn-store/src/nested/de.rs` has the same shape (`:151-235`):

```rust
fn deserialize_i32<V>(self, visitor: V) -> Result<V::Value, Self::Error> {
    visitor.visit_i32(self.value.unwrap().as_i32().unwrap().to_owned())
}
```

Two unwraps per call: `self.value` may be `None` for an absent field, and `as_i32()` returns `None` when the stored `NestedValue` is a different variant. The affected methods are `deserialize_string`, `deserialize_bool`, `deserialize_i16`, `deserialize_i32`, `deserialize_i64`, `deserialize_u8`, `deserialize_u16`, `deserialize_u64`, `deserialize_f32`, `deserialize_f64`, `deserialize_str` and `deserialize_byte_buf`.

The neighbouring container methods do it correctly - `deserialize_map` (`:115-130`) and the struct arm (`:95-98`) both return `de::Error::custom`. So the error channel exists and is used a few lines away; the scalars just skip it.

There are also bare `panic!`s in the sequence visitors (`:421`, `:435`, `:449`, `:463`) and at `:688`.

This is reached from a public API taking a user-supplied file path:

```rust
let config: ModelConfig = PytorchReader::load_config("model.pth", Some("config"))?;
```

A checkpoint whose config was written by a different version of the producing script - a field that went from int to string, or that was removed - takes down the process rather than returning `Err`.

## Reproduction

```rust
#[derive(Debug, Deserialize)]
struct Config { hidden_size: i32 }

let mut map = HashMap::new();
map.insert("hidden_size".to_string(), NestedValue::String("768".to_string()));
let de = Deserializer::<DefaultAdapter>::new(NestedValue::Map(map), false);

let outcome = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| Config::deserialize(de)));
println!("{}", if outcome.is_ok() { "returned" } else { "PANICKED" });
```

```
thread '...' panicked at crates/burn-store/src/nested/de.rs:158:56:
called `Option::unwrap()` on a `None` value
PANICKED
```

## Suggested fix

Replace each `unwrap` pair with an error carrying what was expected and what was found, in the style already used by `deserialize_map`:

```rust
fn deserialize_i32<V>(self, visitor: V) -> Result<V::Value, Self::Error> {
    let value = self.value.ok_or_else(|| de::Error::custom("expected i32, found nothing"))?;
    let n = value.as_i32().ok_or_else(|| {
        de::Error::custom(format!("expected i32 but got {value:?}"))
    })?;
    visitor.visit_i32(n)
}
```

A small helper taking the accessor and a type name would keep the twelve methods from repeating it.

Sadik00789:
Fixes #5484

### Description
Replaces direct `.unwrap()` calls and bare `panic!` invocations across `crates/burn-store/src/nested/de.rs` with proper `de::Error` propagation:

1. **`extract_scalar` Helper:** Added a private helper to safely extract and validate `self.value` variants, returning informative errors (e.g., `expected i32 but got NestedValue::String(...)`) when the variant mismatches or the field is absent.
2. **Scalar Visitor Refactoring:** Refactored all twelve scalar visitor methods (`deserialize_i32`, `deserialize_string`, `deserialize_bool`, `deserialize_byte_buf`, etc.) to route through `extract_scalar` without panicking.
3. **Panic Site Cleanup:** Replaced unhandled `panic!` invocations in `VecSeqAccess` and enum variant validation with structured `de::Error` returns.
4. **Unit Tests:** Added unit tests verifying that mismatched types and missing fields return clean `Err` results instead of terminating the process.

### Validation
- `cargo test -p burn-store` passed (249 tests passing, including new regression cases).
- `cargo fmt --check` and `cargo clippy -p burn-store -- -D warnings` passed.

### DIFF
--- a/crates/burn-store/src/nested/de.rs
+++ b/crates/burn-store/src/nested/de.rs
@@ -40,6 +40,21 @@ impl<A: BurnModuleAdapter> Deserializer<A> {
             phantom: std::marker::PhantomData,
         }
     }
+
+    fn extract_scalar<T>(
+        self,
+        expected: &'static str,
+        extractor: impl FnOnce(&NestedValue) -> Option<T>,
+    ) -> Result<T, Error> {
+        let value = self
+            .value
+            .ok_or_else(|| custom_err(format!("expected {expected}, found nothing")))?;
+
+        match extractor(&value) {
+            Some(val) => Ok(val),
+            None => Err(custom_err(format!("expected {expected} but got {value:?}"))),
+        }
+    }
 }
 
 impl<'de, A: BurnModuleAdapter> serde::Deserializer<'de> for Deserializer<A> {
@@ -141,11 +156,8 @@ impl<'de, A: BurnModuleAdapter> serde::Deserializer<'de> for Deserializer<A> {
     where
         V: Visitor<'de>,
     {
-        match self.value {
-            Some(NestedValue::Bool(val)) => visitor.visit_bool(val),
-            Some(other) => Err(custom_err(format!("expected bool but got {other:?}"))),
-            None => Err(custom_err("expected bool, found nothing")),
-        }
+        let val = self.extract_scalar("bool", |v| v.clone().as_bool())?;
+        visitor.visit_bool(val)
     }
 
     fn deserialize_i8<V>(self, _visitor: V) -> Result<V::Value, Self::Error>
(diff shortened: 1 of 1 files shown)


crates/burn-store/src/nested/de.rs
```diff
@@ -232,10 +267,15 @@ impl<'de, A: BurnModuleAdapter> serde::Deserializer<'de> for Deserializer<A> {
     where
         V: Visitor<'de>,
     {
-        let bytes = self.value.unwrap().as_bytes().unwrap();
-        match bytes.try_into_vec::<u8>() {
-            Ok(bytes) => visitor.visit_byte_buf(bytes),
-            Err(bytes) => visitor.visit_bytes(&bytes),
+        match self.value {
+            Some(NestedValue::Bytes(bytes)) => match bytes.try_into_vec::<u8>() {
```
laggui, line 271: Is dropping the `NestedValue::U8s` support intentional? I don't think there are any producers for that variant anymore since #5083, but `as_bytes()` path still handled it.

For now I'd preserve the existing behavior since the variant hasn't been removed.

laggui (commented):
LGTM, just a minor comment

crates/burn-store/src/nested/de.rs
Sadik00789: Good catch! Restored the NestedValue::U8s arm in deserialize_byte_buf and added a unit test to prevent future regressions.

laggui (approved):
Thanks for addressing my comment.

Windows CI failure is unrelated, and fixed on main.

PR #5498, opened 2026-08-26
feat(module): generalize `Param` with module-owned flags

laggui:
### Checklist

### Related Issues/PRs

Proposed generalization of #5493

`TrainingFlag` solves the immediate problem, but its abstraction is too specific. The underlying need is broader: some module-owned state must have an identity, participate in module traversal and parameter-group selection, and follow freeze/inference lifecycle transition, even when it is not a tensor.

Generalizing it to `Param<Flag>` gives a cleaner model:

> `Param<T>` is an identified, traversable module-owned value; tensor parameters are its differentiable specialization.

`Flag` is a reusable boolean value without embedded Dropout, BatchNorm, or training semantics. Wrapping it in `Param` provides identity and lifecycle integration, while the  field holding it provides its meaning.

  ### Changes

- New `Param<Flag>` used for Dropout, BatchNorm, GaussianNoise, and RRelu training behavior
- Introduce `ParameterValue` for values stored in `Param<T>`, while `Parameter` retains tensor-specific gradient, device, shape, and loading capabilities
- Extend module traversal and parameter-group selection to flags, including through type-erased reparameterizations, so freezing reaches matched tensor parameters and control
state
- Generalize `Param`'s internal lifecycle state from `require_grad` to `is_active`, allowing tensor gradient state and flag enablement to survive `valid()` -> `train()`  transitions
- Keep flags device-independent and outside serialized records; loading preserves the destination flag configuration and identity

### Tests
New tests covering freeze/unfreeze behavior, validation round trips, container traversal, device independence, records, and capability separation.


nathanielsimard (approved):
LGTM

antimora:
Post-merge review notes. Nothing here needs reverting, but a few are worth follow-ups. Verified against `c63690db2`; the empirical bits below were measured, not read off the source.

## `valid()` disables flags in the value, and only `train()` restores

`Param<Flag>::valid()` stores `Flag(false)` and parks the real state in `is_active`. `is_active` is `pub(crate)` and nothing on a forward path reads it: all four layers read `is_enabled()`. `to_device` and `fork` on `Param<Flag>` are the identity, so the disable survives every transition except `from_inner`.

```
autodiff baseline:         flag=true
after valid():             flag=false
after valid().fork(ad):    flag=false
after valid().to_device(): flag=false
after valid().train():     flag=true
```

`Learner::fit` returns `model.valid()` (`learner/supervised/strategies/base.rs:157`), and since tensors are backend erased that value has the same type as the trainable model. A two stage finetune that resumes from a `LearningResult` without going through `train()` will train with dropout off, noise off, RRelu on the fixed midpoint slope, and batch norm reading frozen running statistics. No error, no warning, and `is_active` is not reachable from outside the crate to inspect.

Before this PR the device was the only input, so moving back onto an autodiff device was enough to restore training behavior.

Two ways out: leave the flag value alone in `valid()` and let the existing `is_autodiff()` guard handle inference, since it already fires on all four forward paths; or make `is_active` public and document `train()` as the only restore path.

## `valid()` is no longer a no-op on a plain device module

`valid_on_a_plain_device_module_is_no_op` (`module/base.rs:643`) still passes because `SimpleLinear` carries no flag, but the contract it names is now false:

```
plain device baseline: flag=true
after valid():         flag=false
```

So a defensive `.valid()` before inference permanently changes later training behavior.

## BatchNorm display was missed

`BatchNorm::custom_content` (`norm/batch.rs:215`) never adds `training`, so a frozen batch norm prints identically to a trainable one. Dropout, RRelu and GaussianNoise all got this in b4d495f. Batch norm is where it matters most, since the flag also gates the `running_mean` / `running_var` writes, so "why did my running stats stop moving" has no answer in the printed model. The same three line match is now copy pasted in three places and needs a fourth; a small helper on `Param<Flag>` would stop the next one being forgotten.

## Docs that no longer match behavior

- `no_grad` (`module/base.rs:144`) says "each layer with no tensor of its own - dropout, noise, randomized activations". BatchNorm owns `gamma` and `beta`, is affected, and has the largest numeric change of the four.
- `freeze_group` / `unfreeze_group` (`base.rs:166`, `:180`) still describe pure `require_grad` operations. `unfreeze_group` re-arming a flag someone disabled by hand is documented nowhere.
- `burn-book/src/building-blocks/module.md:58` maps `no_grad()` onto PyTorch `require_grad_(False)`. It is now that plus `.eval()`, which PyTorch keeps as a separate axis.
- `map_flag`'s doc offers "a device move" as an example of a mapper with no opinion about control state. `to_device` is generated per field by the derive (`codegen_struct.rs:88`) and never goes through `ModuleMapper` at all. `ModuleSharder` would be a real example.

## LoRA freezes every base weight but leaves flags on

`apply_lora` / `apply_qlora` set `require_grad(false)` on every base float param, but they run through `ApplyReparameterization`, which implements only `map_float` and inherits the identity `map_flag`. So the two "freeze the whole model" entry points in the crate now disagree, and the one that disagrees is the partial finetuning path the commit messages cite as motivation. Defensible for dropout, less so for batch norm running statistics. Worth being explicit either way.

## `list_param_ids` omits flags

`ParamIdCollector` in `param/group.rs` collects flags; the identically named one behind the public `list_param_ids` (`param/visitor.rs:12`) does not. `ParamGroup::from_ids(list_param_ids(&m))` therefore freezes tensors but no flags.

Related: path and regex groups do reach flags correctly. I checked that `from_path("inner.training")` disables only the flag and `from_path("inner.weight")` only the tensor, in both the freeze and unfreeze directions. But a group that exhaustively lists tensor paths, or `from_regex(r"(weight|bias)$")`, silently misses `*.training`, and `freeze_group` reports no match count either way. A serialized `ParamGroup` also half matches after `load_record`, since tensor ids come back from the record while flag ids are regenerated by `init()`.

## Test gaps

- `unfreeze_group`'s `training = true` arm is never exercised. The only unfreeze test runs on `SimpleLinear`, which has no flag. Changing `base.rs:189` to `training = false` leaves the entire suite green while inverting the sole recovery path from `no_grad()`.
- Path based flag selection is untested. Every flag and group test uses id matching, which ignores paths, or asserts a non-match that would pass vacuously if the path were computed wrong.
- The new `self.is_active = require_grad;` on the lazy branch of `set_require_grad` (`param/base.rs:564`) is unreached. It is a real fix, since the field went stale on lazy params before, and it deserves a test on a param that has not been initialized yet.
- `parameter_capabilities_are_split_by_value_kind` is tautological: all three bounds are compile prerequisites, so it cannot fail while the build succeeds. It never asserts the half its name promises, that `Flag` is not a `Parameter`, which needs a `compile_fail` case.

Smaller items: `Flag::with(self, value)` discards `self`, so it is `Flag::new` with an extra move, and becomes silent data loss the day `Flag` gains a second field. `Param<Flag>`'s `no_grad` override produces exactly what the default `map_flag` path already gives.

## What holds up

Records genuinely exclude flags: `Collector` and `ModuleRecordMapper` both omit the hooks, old checkpoints load unchanged, and the store test pins both destination polarities plus id preservation. The sealing is airtight, `mod sealed` being `pub(crate)`. The `DynModuleVisitor` / `DynModuleMapper` hooks are correctly not defaulted, so downstream impls break loudly rather than silently dropping flags. And every remaining `is_autodiff()` branch in burn-nn now consults a flag, with no layer missed.

Trait split itself is the right call. Splitting out `ParameterValue` so a flag stops pretending to have a device and a shape is what makes the rest of this possible.

PR #5502, opened 2026-08-26
Refactor interpolate autotuning with cubek strategies and roofline bounds

SamuelBelanger:
### Changes

This PR refactors the CubeCL backend's interpolation kernel autotuner to delegate launch geometry selection to `cubek`, benchmarking runtime strategies (`MaximizeThroughput` vs. `MinimizeLatency`) with roofline model bounds rather than manually sweeping tile sizes and geometry heuristics.

#### Key Improvements:

1. **Strategy-Based Autotuning over Geometry Enumeration**:
   - `cubek` now infers the optimal launch geometry from device properties and problem dimensions.
   - The autotuner now benchmarks the two primary runtime strategies:
     - `maximize_throughput` (`InterpolateStrategy::MaximizeThroughput`)
     - `minimize_latency` (`InterpolateStrategy::MinimizeLatency`)
   - Removed obsolete manual tile-size sweeps, plane/row box bounds, and fallback groups.

2. **Roofline Bounds Integration**:
   - Integrated `autotune_bounds::with_bounds` using `InterpolateCost` and `roofline_bounds`.
   - Costs both arithmetic intensity (tap count and trigonometric filter operations like Lanczos3) and memory bandwidth, enabling the tuner to short-circuit early once roofline theoretical limits are approached.

3. **Autotune Key & Shape Correction**:
   - Corrected key generation by constructing `InterpolateForwardProblem` from permuted NHWC shapes instead of raw NCHW layout (which previously mapped width into channel count).
   - Properly populates `InterpolateAutotuneKey` with filter mode, `align_corners`, channel depth, and spatial dimensions.

4. **Streamlined `InterpolateStrategy` API**:
   - Replaced `GlobalMemory` / `SharedMemory` variants with `InterpolateStrategy::Specific(CubekInterpolateStrategy)` and `InterpolateStrategy::Autotune`.
   - Default non-autotune strategy is set to `Specific(CubekInterpolateStrategy::MaximizeThroughput)`.

5. **Dependency Updates**:
   - Updated `cubecl` and `cubek` workspace dependencies in `Cargo.toml` and `Cargo.lock`.

### Testing

- Verified autotune key generation and problem shape permutation.
- Ran backend interpolate unit tests across supported interpolation modes (nearest, bilinear, bicubic, lanczos3).


crates/burn-cubecl/src/kernel/interpolate/base.rs
```diff
@@ -9,30 +9,23 @@ use crate::{
...
-/// Strategy used to select which interpolate implementation to run.
+/// Strategy used to select how interpolation runs.
 pub enum InterpolateStrategy {
-    /// Default interpolate strategy.
-    GlobalMemory(GlobalMemoryStrategy),
-
-    /// Use shared memory for caching tiles of the input and output.
-    SharedMemory(SharedMemoryStrategy),
+    /// Run the kernel with the given geometry.
+    Config(InterpolateConfig),
```
louisfd: Not `Config`, maybe `Specific` or something

crates/burn-cubecl/src/kernel/interpolate/tune.rs
```diff
@@ -112,25 +96,286 @@ pub fn interpolate_autotune<R: CubeRuntime>(
...
+///
+/// A cube that walks past the output it covers, and a neighbourhood around the predicted
+/// geometry. The neighbourhood is what keeps the measured set small: the box holds dozens of
+/// geometries, and sweeping it from a corner spends the whole tuning budget before reaching the
+/// ones that matter.
+///
+/// What a device refuses is not repeated here. cubek returns those before any launch work, and
+/// autotune reads an `Err` as "drop this candidate", so the rule stays in the one place that
+/// owns it.
+fn worth_measuring(
```
louisfd: Too many private functions

crates/burn-cubecl/src/kernel/interpolate/base.rs
```diff
@@ -43,11 +36,10 @@ impl Default for InterpolateStrategy {
...
         return InterpolateStrategy::Autotune;
 
-        // if autotune is disabled, default to global memory with a 16x16 tile size
+        // A guess, not a measurement: it sits mid-box in cubek's benchmark catalogue and has
+        // no recorded sweep behind it. Replace it once cubek publishes one.
         #[cfg(not(feature = "autotune"))]
-        InterpolateStrategy::GlobalMemory(GlobalMemoryStrategy {
-            tile_size: TileSize::new(16, 16),
-        })
+        InterpolateStrategy::Config(InterpolateConfig::new(Residence::InPlace, 4, 2, 1))
```
louisfd: Missing strategy / blueprint. Here you would give a strategy instead of config, like MaximizeThroughput, and then the CubeK selector will take this as input along with hardware properties, and decide on values like your 4,2,1

louisfd (changes requested):
Missing a selector: (Strategy, HardwareProperties) -> Blueprint

crates/burn-cubecl/src/kernel/interpolate/base.rs
louisfd: Vectorization must be maximized, coalescing too. 
Then try to fill SMs, etc.

louisfd: Plane per cube: CPU=core, GPU~=4 or 8 usually (autotune between both), 16 for memory bound

Cargo.lock
```diff
@@ -3249,7 +3250,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
```
laggui: The windows crates downgrade are causing the CI failures (mismatch with wgpu's transitive dep). Unsure why the cubek bump changed these, but we should revert the changes. I only highlighted this change but there seem to be a couple in the lockfile.

laggui (approved):
LGTM, pending the `Cargo.lock` changes mentioned below.

PR #5504, opened 2026-08-27
feat: add GroupNorm backend primitive and CubeCL kernel

bungogood:
## Summary

- add a portable `ModuleOps::group_norm` primitive with independently optional scale and bias parameters
- route GroupNorm through tensor, NN, IR, Dispatch, Fusion, and Router layers
- add a CubeCL kernel using F32 Welford accumulation and one workgroup per batch/group
- support contiguous, permuted, sliced, and empty inputs
- validate affine shape, dtype, and device consistently before dispatch
- preserve existing BF16 and F64 precision semantics through the portable fallback
- fall back on runtimes without the required 256-thread workgroups
- cover forward behavior, layouts, dtypes, edge cases, and gradients

## Autodiff

`Autodiff<B>` intentionally uses the portable differentiable decomposition, while non-autodiff CubeCL backends use the specialized kernel for supported dtypes. Gradient tests cover input, gamma, and beta; all four affine combinations; multiple batches and groups; non-contiguous input; empty input; and gradient checkpointing. The CUDA autodiff composition was exercised on an NVIDIA GPU. This PR does not claim optimized training performance or add a custom GroupNorm backward operation.

## Performance

On a representative FP16 residual CNN workload with 21 GroupNorm calls, the specialized kernel reduced batch-1024 model latency from 37.345 ms to 20.756 ms and CUDA launches from 264 to 129 on an RTX 3070. Maximum FP16 error remained approximately `2.1e-2`. No GroupNorm epilogue fusion was enabled.

## Testing

- `cargo +nightly test -p burn-backend-tests --no-default-features --features cpu,std --test tensor group_norm -- --test-threads=1`
- `cargo +nightly test -p burn-backend-tests --no-default-features --features cpu,fusion,std --test tensor group_norm -- --test-threads=1`
- `cargo +nightly test -p burn-backend-tests --no-default-features --features cpu,std --test autodiff group_norm -- --test-threads=1`
- `cargo +nightly test -p burn-backend-tests --no-default-features --features cpu,fusion,std --test autodiff group_norm -- --test-threads=1`
- `cargo +nightly test -p burn-backend-tests --no-default-features --features cuda,std --test tensor group_norm -- --test-threads=1`
- `cargo +nightly test -p burn-backend-tests --no-default-features --features cuda,std --test autodiff group_norm -- --test-threads=1`
- `cargo +nightly check -p burn-cubecl -p burn-nn -p burn-fusion -p burn-router -p burn-dispatch`
- `cargo +nightly fmt --all -- --check`


crates/burn-cubecl/src/kernel/group_norm.rs
```diff
@@ -0,0 +1,349 @@
...
+            let shape = tensor.shape(dim);
+            offset += spatial_index % shape * tensor.stride(dim);
+            spatial_index /= shape;
+        }
+        offset
+    }
+}
+
+#[cube(launch, address_type = "dynamic")]
+fn group_norm_kernel<F: Float>(
```
laggui, line 41: Instead of hardcoding `f32` accumulation, I'd keep accumulation precision separately from the tensor precision, e.g. `<F: Float, A: Float>`. I think this could additionally support f64 specialization, which currently uses the fallback path.

crates/burn-cubecl/src/kernel/group_norm.rs
```diff
@@ -0,0 +1,349 @@
...
+    let gamma = gamma.unwrap_or_else(|| input.clone());
+    let beta = beta.unwrap_or_else(|| input.clone());
+    let spatial_size = shape[2..].iter().product::<usize>();
+    let channels_per_group = num_channels / num_groups;
+
+    let hardware = &client.properties().hardware;
+    assert!(
+        hardware.max_cube_dim.0 >= 256,
+        "group_norm: runtime must support 256 threads per workgroup"
+    );
```
laggui, line 313: Hmmm what if we moved away from the fixed 256-unit assumption, and instead select a power-of-two workgroup size from a small set based on `group_size` and the `max_cube_dim` and `max_units_per_cube`. That comptime size would be used for the shared allocations.

crates/burn-dispatch/src/ops/module.rs
```diff
@@ -676,6 +676,21 @@ impl ModuleOps<Self> for Dispatch {
...
+        beta: Option<FloatTensor<Self>>,
+        num_groups: usize,
+        epsilon: f64,
+    ) -> FloatTensor<Self> {
+        multi_op!(
+            inputs[(tensor, float)],
+            opt_inputs[(gamma, float), (beta, float)],
+            => Float,
+            B::group_norm(tensor, gamma, beta, num_groups, epsilon)
+        )
```
laggui, line 691: This is stale with #5520 merged. But that's on me for the delayed review 😅 `multi_op!` should no longer be required, a simple passthrough `B::group_norm(tensor, gamma, beta, num_groups, epsilon)` should be sufficient now.

crates/burn-fusion/src/ops/module.rs
```diff
@@ -1592,6 +1592,52 @@ impl<B: FusionBackend> ModuleOps<Fusion<B>> for Fusion<B> {
             .output()
     }
 
+    fn group_norm(
+        tensor: FloatTensor<Fusion<B>>,
+        gamma: Option<FloatTensor<Fusion<B>>>,
+        beta: Option<FloatTensor<Fusion<B>>>,
+        num_groups: usize,
+        epsilon: f64,
+    ) -> FloatTensor<Fusion<B>> {
```
laggui, line 1601: Now that `GroupNorm` is represented as an opaque fusion operation, downstream elementwise operations can no longer fuse into its final normalization/affine stage as they could with the decomposed fallback. The dedicated kernel likely compensates for this, so I don't think it's blocking. Your benchmarks used fusion right (enabled by default)?

laggui (changes requested):
Thanks for the contribution! Noticed a couple of things during my first pass.

PR #5509, opened 2026-08-27
fix(module): keep the param mapper across valid() and from_inner

nathanielsimard:


laggui (approved):
LGTM. I just added a unit test to exercise the regression

PR #5325, opened 2026-08-07
feat(tensor): support mul indexing updates on NdArray and Flex

mattisonchao:
## Motivation

`scatter` and `select_assign` accept `IndexingUpdateOp`, but the plain indexed update APIs did not support `Mul`, even though multiply-style updates are available through `scatter_nd`. This adds consistent multiply updates for float and integer tensors on the Flex and NdArray CPU backends, including float autodiff support.

The backward formulas require unique indices. Duplicate-index behavior is therefore documented explicitly so callers do not rely on backend-specific update ordering or undefined gradients.

## Modifications

- Implement `Mul` for float and integer `scatter` and `select_assign` on Flex and NdArray.
- Share Flex's add and multiply index traversal while retaining its specialized 2D paths.
- Add float autodiff gradients for `scatter(..., Mul)` and `select_assign(..., Mul)` without depending on backend-specific `Assign` support.
- Document `Mul` behavior and duplicate-index semantics on the public tensor APIs.
- Add forward and autodiff coverage for both supported CPU backends.

PR #5512, opened 2026-08-28
feat(cubecl): support mul indexing updates

mattisonchao:
## Motivation

PR #5325 added multiply-style scatter and select updates for NdArray and Flex, but CubeCL still supports only additive numeric updates for these APIs. CubeCL already has a reusable MulOp and uses it for scatter_nd, so the regular indexed update kernels can support IndexingUpdateOp::Mul without introducing new kernel math.

This brings CubeCL backends to parity for float and integer tensors while preserving the existing boolean OR behavior.

## Modifications

- Share the CubeCL scatter and select_assign launch paths across binary operation families.
- Dispatch MulOp for float and integer scatter and select_assign updates.
- Preserve the existing AddOp and boolean OrOp paths.
- Enable the shared forward and autodiff multiply-update tests for CubeCL backends.
- Add CubeCL reference tests for multiply updates across multiple workgroups using unique indices.

### DIFF
--- a/crates/burn-backend-tests/tests/autodiff/gather_scatter.rs
+++ b/crates/burn-backend-tests/tests/autodiff/gather_scatter.rs
@@ -98,7 +98,7 @@ fn test_scatter_assign_grad() {
         .assert_eq(&TensorData::from([[1., 1.], [1., 1.]]), false);
 }
 
-#[cfg(any(feature = "flex", feature = "ndarray"))]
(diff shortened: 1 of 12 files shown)


laggui (approved):

laggui:
I had to fix the conflicts with #5511 for the feature gating, simply removed it now that all tested backends support it. Thank you!

PR #5513, opened 2026-08-28
fix(router): support rfft and irfft

Mikyx-1:
### Checklist

### Related Issues/PRs

None.

### Changes

BackendRouter currently panics when rfft or irfft is called even though the corresponding IR operations and server-side execution already exist.

This change registers both operations through the router and returns their output tensors. It also adds an end-to-end WebSocket regression test that verifies the RFFT spectrum and IRFFT reconstruction over RemoteBackend.

### Testing

- cargo run-checks
- cargo test -p burn-router
- cargo test -p burn-remote test_fft_over_websocket -- --nocapture


laggui (approved):
LGTM, thanks!

PR #5515, opened 2026-08-28
Fix Dice metric aggregation across batches

### LINKED ISSUE #5514: Fix incorrect epoch aggregation in DiceMetric
### Bug

DiceMetric currently passes each batch Dice coefficient to NumericMetricState. That state averages batch values by batch size, but Dice is a ratio of foreground statistics and is not decomposable as a sample-weighted average.

For example, a perfect batch containing one foreground pixel followed by a batch containing four false-positive pixels produces an epoch score near 0.5 with per-batch averaging. Computing Dice over the accumulated statistics gives (2 * 1) / (1 + 5) = 1 / 3.

### Expected behavior

Accumulate the intersection, prediction sum, and target sum across batches, then compute the running and final epoch values from the totals. The displayed batch value should remain the Dice score for the latest batch.

This provides global micro-aggregation across the epoch and handles varying batch sizes and foreground support correctly. Regression coverage should include uneven foreground support, varying batch sizes, empty masks, background exclusion, and state reset.

Mikyx-1:
## Checklist

### Related Issues/PRs

Fixes #5514.

### Changes

- Accumulate the raw Dice numerator and denominator across batches instead of averaging batch ratios by batch size.
- Preserve the latest batch score while computing running and final values with global micro-aggregation.
- Document and validate epsilon behavior, including the existing perfect-score convention for two empty masks.
- Add regression coverage for uneven foreground support, varying batch sizes, empty masks, background exclusion, reset behavior, and invalid epsilon values.

The varying-batch-size reference uses the equivalent PyTorch tensor formula over concatenated predictions and targets.

### Testing

- `cargo run-checks`
- `cargo test -p burn-train --features vision metric::vision::dice::tests --lib`
- `cargo test -p burn-train --features vision --lib`

### DIFF
--- a/crates/burn-train/src/metric/vision/dice.rs
+++ b/crates/burn-train/src/metric/vision/dice.rs
@@ -1,9 +1,6 @@
-use crate::metric::{MetricAttributes, MetricName, SerializedEntry};
+use crate::metric::{MetricAttributes, MetricName, NumericEntry, SerializedEntry, format_float};
 
-use super::super::{
-    Metric, MetricMetadata,
-    state::{FormatOptions, NumericMetricState},
-};
+use super::super::{Metric, MetricMetadata, state::FormatOptions};
 use burn_core::{
     prelude::Tensor,
     tensor::{Int, s},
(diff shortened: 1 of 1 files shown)


laggui (approved):
LGTM. We could potentially generalize the `DiceMetricState` into something like `RatioMetricState`, since it's not dice-specific.

PR #5517, opened 2026-08-28
Fix ROUGE-L final value panic

### LINKED ISSUE #5516: Fix ROUGE-L panic when retrieving the final value
### Bug

`RougeLScore` implements `Numeric::final_value` with `todo!()`. The metric therefore panics when Burn retrieves its final numeric value at epoch completion.

The existing implementation computes a mean ROUGE-L F1 per batch and passes the batch size to `NumericMetricState`. Its weighted aggregation is equivalent to the macro average over all sequences, so no additional corpus state is required for the current metric definition.

### Expected behavior

`final_value()` should return the final entry from `NumericMetricState`, consistent with other sample-averaged metrics, without panicking.

Regression coverage should verify the final value across unequal batch sizes and after clearing the metric state.

Mikyx-1:
## Checklist

### Related Issues/PRs

Fixes #5516.

### Changes

- Return the final `NumericMetricState` entry from `RougeLScore::final_value` instead of panicking with `todo!()`.
- Remove the stale aggregation TODO. Each batch value is the mean of per-sequence ROUGE-L F1 scores, and `NumericMetricState` weights it by batch size, producing the dataset-wide macro average.
- Add regression coverage for final values across unequal batch sizes and after clearing state.

### Testing

- `cargo run-checks`
- `cargo test -p burn-train metric::rouge::tests --lib`
- `cargo test -p burn-train --lib`

### DIFF
--- a/crates/burn-train/src/metric/rouge.rs
+++ b/crates/burn-train/src/metric/rouge.rs
@@ -25,6 +25,8 @@ fn lcs_length(a: &[i32], b: &[i32]) -> usize {
 }
 
 /// ROUGE-L metric based on longest common subsequence.
+///
+/// Reports the macro average of the per-sequence ROUGE-L F1 scores.
 #[derive(Clone)]
 pub struct RougeLScore {
     name: MetricName,
--- a/crates/burn-train/src/metric/rouge.rs
+++ b/crates/burn-train/src/metric/rouge.rs
@@ -162,7 +162,6 @@ impl Metric for RougeLScore {
     }
 }
 
-// TODO: can be computed per batch, but epoch-level value should be aggregated properly (not just mean of each batch)
 impl Numeric for RougeLScore {
     fn value(&self) -> Option<NumericEntry> {
         Some(self.state.current_value())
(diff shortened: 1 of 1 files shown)


crates/burn-train/src/metric/rouge.rs
```diff
@@ -162,7 +162,6 @@ impl Metric for RougeLScore {
     }
 }
 
-// TODO: can be computed per batch, but epoch-level value should be aggregated properly (not just mean of each batch)
```
laggui, line 165: That comment was stale, likely meant to signal micro/corpus-style score.

But we should absolutely document that `RougeLScore` reports the macro average of per-sequence ROUGE-L F1 scores. Under that definition, the current implementation is accurate.

Can you make this change?

laggui (commented):
LGTM, just a minor documentation request before merge

crates/burn-train/src/metric/rouge.rs
Mikyx-1: Thanks for the clarification. I’ve updated the `RougeLScore` documentation to explicitly state that it reports the macro average of per-sequence ROUGE-L F1 scores. The existing implementation remains unchanged, and all ROUGE metric tests pass.

laggui (approved):

PR #5532, opened 2026-08-30
feat(tensor): support assign updates on Flex, Tch and CubeCL

mattisonchao:
## Motivation
`IndexingUpdateOp::Assign` is already part of the public indexing API and supported by NdArray and autodiff, but regular `scatter` and `select_assign` calls still panic on Flex, Tch, and CubeCL. Supporting overwrite updates consistently removes that backend-dependent limitation.

## Modifications
- add overwrite update paths for float and integer tensors on Flex
- use LibTorch scatter and index-copy primitives for Tch assignment updates
- wire CubeCL indexing kernels to the existing `AssignOp`
- enable the shared forward and autodiff assignment tests for all backends


laggui (commented):
LGTM, just need to fix the conflicts with main (likely due to #5523)

laggui (approved):

PR #5534, opened 2026-08-30
Fix zero-safe prod gradients

### LINKED ISSUE #5533: prod and prod_dim gradients produce NaN for zero inputs
**Describe the bug**

The autodiff backward implementations for `prod` and `prod_dim` compute the local derivative as `product / input`. When a reduction contains a zero, the forward product is zero and the zero position evaluates `0 / 0`, producing `NaN`.

For example, the gradient of `prod([2, 0, 3, 4])` should be `[0, 24, 0, 0]`, but the current backward pass returns `NaN` at the zero position. The same issue occurs independently in each reduction lane for `prod_dim`.

This affects both the base and balanced-checkpointing autodiff strategies.

**To reproduce**

Enable and run the existing ignored regressions:

- `should_diff_prod_single_zero`
- `should_diff_prod_multiple_zeros`

in `crates/burn-backend-tests/tests/autodiff/aggregation.rs`.

The issue is present on `main` at `06682dd43`.

**Expected behavior**

The derivative with respect to each input should be the product of all other inputs:

- No zeros: use the ordinary product quotient.
- Exactly one zero: only the zero position receives the product of all nonzero values.
- Multiple zeros: every derivative is zero.

For `prod_dim`, these rules should be applied independently per reduction lane while preserving upstream-gradient broadcasting.

**Additional context**

The related `cumprod` zero-gradient issue #3864 is closed, but its fix does not cover the `prod` and `prod_dim` reductions.

Mikyx-1:
### Checklist

### Related Issues/PRs

Closes #5533.

### Changes

- Add a shared zero-safe backward implementation for `prod` and `prod_dim`.
- Replace zeros with ones before computing the nonzero product and count zeros per reduction.
- Select the correct derivative for reductions containing zero, one, or multiple zeros without dividing by zero.
- Preserve upstream-gradient broadcasting for global and dimension-wise reductions.
- Enable the existing zero regressions and add `prod_dim` coverage for mixed zero counts and non-unit upstream gradients.

### Testing

- `cargo run-checks`
- `cargo test -p burn-backend-tests --no-default-features --features ndarray,std --test autodiff aggregation:: -- --nocapture`
- `cargo check -p burn-autodiff --no-default-features`

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -79,29 +79,19 @@ fn prod_backward<B: Backend>(
     let at_most_one_zero = B::float_lower_elem(zero_count, 1.5.into(), bool_dtype);
     let one_zero = B::bool_and(B::bool_not(no_zero.clone()), at_most_one_zero);
 
-    let ones = B::float_ones(shape.clone(), &device, dtype.into());
-    let expand = |tensor| {
-        let tensor = if dim.is_none() {
-            unsqueeze_like::<B>(tensor, shape.clone())
-        } else {
-            tensor
-        };
-        B::float_mul(ones.clone(), tensor)
-    };
-
-    let grad = expand(grad);
-    let nonzero_product = expand(nonzero_product);
     let no_zero = B::bool_expand(no_zero, shape.clone());
     let one_zero = B::bool_expand(one_zero, shape.clone());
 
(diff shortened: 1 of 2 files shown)


laggui (approved):
LGTM, I just removed the redundant float expansions and added regression tests.

The generic zero-safe backward still performs extra reductions and masking for zero-free inputs, so it may affect backward performance. But potential optimizations could be could be benchmarked and potentially captured in a `B::float_prod_backward` backend op.

PR #5536, opened 2026-08-30
fix(store): exclude adapter-matched tensors from unused

### LINKED ISSUE #5483: burn-store: tensors matched through an adapter's alternative name are reported unused
## Summary

When the applier finds a tensor through `ModuleAdapter::get_alternative_param_name`, it applies it but never records the name it was found under. `ApplyResult::unused` is computed from exactly that record, so every successfully applied alternative-name tensor is also reported as unused.

Loading any PyTorch normalization layer produces two phantom "unused" entries per layer.

## Detail

`crates/burn-store/src/applier.rs:177-188`:

```rust
if let Some(alt_name) = adapter.get_alternative_param_name(param_name, module_type) {
    let mut alt_path_stack = self.path_stack.clone();
    *alt_path_stack.last_mut().unwrap() = alt_name.clone();
    let alt_path = alt_path_stack.join(".");

    tensor = self.tensors.get(&alt_path).cloned();

    // Don't mark the alternative path as visited - only the original Burn path
    // should be tracked. The alternative path is just for lookup.
}
```

Keeping the Burn path out of `visited_paths` is right for the `missing` computation. But `into_result` (`:100-105`) reuses `visited_paths` for a second purpose:

```rust
let mut unused: Vec<String> = self
    .tensors
    .keys()
    .filter(|path| !self.visited_paths.contains_key(*path) && !self.skipped.contains(*path))
    .cloned()
    .collect();
```

`self.tensors` is keyed by *source* name, so `norm.weight` is looked up here, found absent from `visited_paths`, and reported unused - even though it was consumed a moment earlier as `norm.gamma`.

## Reproduction

```rust
#[derive(Module, Debug)]
struct NormModel { norm: BatchNorm }

let tensors = vec![
    from_data(TensorData::from([1.0f32; 4]), "norm.weight".into(), None),
    from_data(TensorData::from([0.0f32; 4]), "norm.bias".into(), None),
    from_data(TensorData::from([0.0f32; 4]), "norm.running_mean".into(), None),
    from_data(TensorData::from([1.0f32; 4]), "norm.running_var".into(), None),
];

let result = model.apply(tensors, None, Some(Box::new(PyTorchToBurnAdapter)), false);
println!("applied: {:?}", result.applied);
println!("unused:  {:?}", result.unused);
```

```
applied: ["norm.gamma", "norm.beta", "norm.running_mean", "norm.running_var"]
unused:  ["norm.bias", "norm.weight"]
```

Both listed tensors were applied.

## Suggested fix

Track consumption separately from visitation. A `consumed: HashSet<String>` recording the source key each successful match actually used (the original path or the alternative one), with `unused` filtering on that instead of on `visited_paths`. `missing` keeps using `visited_paths` as it does today.

This also shows up in `ApplyResult`'s "Did you mean?" suggestions (`crates/burn-store/src/apply_result.rs:130-161`), which draw candidates from `unused` - a missing tensor can currently be offered a suggestion that names a tensor already loaded somewhere else.

apoorvdarshan:
### Checklist

### Related Issues/PRs

Closes #5483

### Changes

- Track consumed source tensor paths separately from visited Burn target paths.
- Mark an adapter's alternative source path as consumed when it supplies a tensor.
- Preserve the existing target-oriented missing-tensor behavior.
- Extend the normalization adapter regressions to verify both alternative-path consumption and genuine unused-tensor reporting.

### Testing

- `cargo test -p burn-store --lib normalization_renaming -- --nocapture`
- `cargo test -p burn-store --lib` (260 passed)
- `cargo run-checks`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4364,7 +4364,7 @@ dependencies = [
  "log",
  "presser",
  "thiserror 2.0.19",
- "windows 0.61.3",
+ "windows 0.62.2",
 ]
 
 [[package]]
--- a/crates/burn-store/src/applier.rs
+++ b/crates/burn-store/src/applier.rs
@@ -37,6 +37,8 @@ pub struct Applier {
     errors: Vec<ApplyError>,
     /// Track visited paths with their container stacks (in dot notation) to find missing tensors
     visited_paths: HashMap<String, String>,
+    /// Track source tensor paths matched during application to find unused tensors
+    consumed_paths: HashSet<String>,
     /// Skip enum variant names when matching paths
     /// When true, "feature.BaseConv.weight" will also try to match "feature.weight"
     skip_enum_variants: bool,
@@ -73,6 +75,7 @@ impl Applier {
             skipped: HashSet::new(),
             errors: Vec::new(),
             visited_paths: HashMap::new(),
+            consumed_paths: HashSet::new(),
             skip_enum_variants,
         }
     }
@@ -100,7 +103,7 @@ impl Applier {
         let mut unused: Vec<String> = self
             .tensors
             .keys()
-            .filter(|path| !self.visited_paths.contains_key(*path) && !self.skipped.contains(*path))
+            .filter(|path| !self.consumed_paths.contains(*path) && !self.skipped.contains(*path))
             .cloned()
             .collect();
         // Sort for stable output order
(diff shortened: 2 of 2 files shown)


antimora (approved):
LGTM

Cargo.lock
```diff
@@ -4364,7 +4364,7 @@ dependencies = [
  "log",
  "presser",
  "thiserror 2.0.19",
- "windows 0.61.3",
+ "windows 0.62.2",
```
laggui, line 4367: I fixed the windows dep for gpu-allocator, which was causing the windows CI failure. Unrelated to this PR, it was broken on main in the recent commits.

PR #5086, opened 2026-06-17
Feat/param group

Charles23R:
### Checklist

- [ ] Confirmed that `cargo run-checks` command has been executed.
- [ ] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

_Provide links to relevant issues and dependent PRs._

### Changes

_Summarize the problem being addressed and your solution._

### Testing

_Describe how these changes have been tested._

PR #5537, opened 2026-08-31
feat(module): separate gradient control from module freezing

### LINKED ISSUE #5524: Follow-ups from #5498: valid() permanently disables module-owned training flags
Follow-ups from #5498, which generalized `Param` with module-owned flags. Full detail and repros in https://github.com/tracel-ai/burn/pull/5498#issuecomment-5457074772.

### Behavior

`Param<Flag>::valid()` writes `Flag(false)` into the value and keeps the real state in `is_active`, which is `pub(crate)` and read by nothing on any forward path. `to_device` and `fork` are the identity for flags, so only `train()` restores.

```
after valid():             flag=false
after valid().fork(ad):    flag=false
after valid().to_device(): flag=false
after valid().train():     flag=true
```

`Learner::fit` returns `model.valid()` and tensors are backend erased, so resuming a finetune from a `LearningResult` without `train()` trains with dropout off, noise off, RRelu on the fixed slope, and batch norm on frozen running statistics, silently.

- [ ] Decide the fix: either stop mutating the flag value in `valid()` and rely on the existing `is_autodiff()` guard, which already fires on all four forward paths, or make `is_active` public and document `train()` as the only restore path.
- [ ] `valid()` is no longer a no-op on a plain device module. `valid_on_a_plain_device_module_is_no_op` (`module/base.rs:643`) passes only because `SimpleLinear` has no flag.
- [ ] Decide whether `apply_lora` / `apply_qlora` should disable flags. They freeze every base weight but run through `ApplyReparameterization`, which only implements `map_float`, so the two "freeze everything" paths disagree.
- [ ] `list_param_ids` (`param/visitor.rs:12`) omits flags while `ParamGroup`'s collector includes them, so `ParamGroup::from_ids(list_param_ids(&m))` freezes tensors but no flags.

### Display

- [ ] `BatchNorm::custom_content` (`norm/batch.rs:215`) never adds `training`, so a frozen batch norm prints identically to a trainable one. Add a helper on `Param<Flag>` rather than a fourth copy of the same match.

### Docs

- [ ] `no_grad` (`module/base.rs:144`) excludes batch norm from its list of affected layers, though it is affected and changes the most.
- [ ] `freeze_group` / `unfreeze_group` (`base.rs:166`, `:180`) still describe pure `require_grad` operations.
- [ ] `burn-book/src/building-blocks/module.md:58` maps `no_grad()` onto PyTorch `require_grad_(False)`; it is now that plus `.eval()`.
- [ ] `map_flag`'s doc cites "a device move" as a mapper with no opinion. `to_device` is generated per field by the derive and never uses `ModuleMapper`.

### Tests

- [ ] `unfreeze_group`'s `training = true` arm is unexercised. Setting `base.rs:189` to `training = false` leaves the suite green while inverting the only recovery path from `no_grad()`.
- [ ] No test proves a path or predicate group can select a flag. Existing coverage uses id matching, which ignores paths, or a non-match that passes vacuously.
- [ ] The new `self.is_active = require_grad;` on the lazy branch of `set_require_grad` (`param/base.rs:564`) is unreached; needs a param that has not been initialized yet.
- [ ] `parameter_capabilities_are_split_by_value_kind` cannot fail while the build succeeds, and never asserts that `Flag` is not a `Parameter`. Needs a `compile_fail` case or removal.

laggui:
### Checklist

### Related Issues/PRs

Follow-up to #5498
Closes #5524

### Changes

Improved the explicit contract for `no_grad()` / `freeze()` with the following APIs:

| API                                  | Tensor gradients             | Training flags       |
| :----------------------------------- | :--------------------------- | :------------------- |
| `no_grad()`                            | Disable all floating params                  | Unchanged            |
| `set_require_grad_group(group, false)` | Set matched floating params              | Ignored, even if matched            |
| `freeze()` / `unfreeze()`                   | Disable/enable all                  | Disable/enable all          | 
| `freeze/unfreeze_group(group)`  | Disable/enable matched              | Disable/enable matched      |
| `valid()` / `train()`                    | Validation/training lifecycle | Temporarily disables/restores saved state |

- `no_grad()` is now strictly gradient-only. It was erroneously left over from #5493 (the base for #5498).
- Added `freeze()` and `unfreeze()` as whole-module counterparts to the existing group operations
- `set_require_grad_group` explicitly states that flags are ignored, even if present in a group

- Added `training` to `BatchNorm` display
- LoRA/QLoRA explicitly freeze base tensor parameters while preserving control flags
- `LearningResult::model` documents that it is returned in validation mode and requires `.train()` before follow-up training
- `map_flag` documentation no longer incorrectly lists device moves as mapper operations
- Removed `list_param_ids` (should have been removed by the introduction of `ParamGroup` in #5086)

### Testing

Unit tests


laggui:
@antimora thanks for the detailed issue, all points should be addressed except the shared `Param<Flag>` display helper. The display behavior itself is fixed and tested; I left the small match local because the display behavior belongs to each module's formatting logic, not an inherent responsibility of the generic Flag type.

crates/burn-core/src/module/base.rs
```diff
@@ -141,31 +189,99 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
-    /// Set `require_grad` to `false` for every parameter in the given group, leaving the rest
-    /// of the module untouched.
+    /// Unfreeze the whole module tree.
     ///
-    /// This is the group-scoped counterpart to [no_grad](Module::no_grad): where `no_grad` freezes
-    /// the whole module tree, `freeze_group` freezes only the parameters matched by `group`.  
+    /// Every floating-point tensor parameter starts requiring gradients, and every module-owned
+    /// training flag is enabled. This sets all values to their active state rather than restoring
+    /// a previously captured configuration.
+    fn unfreeze(self) -> Self {
```
antimora, line 285: `unfreeze()` re-enables grad on LoRA base weights, which turns a LoRA finetune back into a full finetune.

On an autodiff device:

```
after apply_lora   base require_grad = false
after unfreeze     base require_grad = true
after set_rg(true) base require_grad = true
base has gradient after backward: true
```

`Param::val()` materializes `base + scale * (a @ b)`, so the thawed base lands in the graph and the optimizer updates it. Same for `set_require_grad(true)` and `set_require_grad_group(g, true)`.

Heads up when writing the regression test: this only reproduces on an autodiff device. `set_require_grad(true)` is a no-op on a plain backend, so a test written against `test_device()` passes either way.

Options: re-freeze the base in the reparameterizer after a map, or give `Param` a "not trainable by construction" bit that `set_require_grad(true)` honors.

crates/burn-core/src/module/base.rs
```diff
@@ -141,31 +189,99 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
+    /// Only floating-point tensor parameters within `group` are affected. Other parameter values
+    /// matched by the group, including module-owned control flags, are ignored; unmatched tensor
+    /// parameters are also left unchanged. This means a group created with
+    /// [`ParamGroup::ids_from_module`] can select an entire subtree while this method changes only
+    /// its tensor gradient state.
+    ///
+    /// Use [`freeze_group`](Module::freeze_group) or
+    /// [`unfreeze_group`](Module::unfreeze_group) when matched training flags should be changed
+    /// together with tensor gradients.
+    fn set_require_grad_group(self, group: ParamGroup, require_grad: bool) -> Self {
```
antimora, line 214: This turns BatchNorm running stats into optimizer targets.

`RunningState::map` routes buffers through `map_float` (`param/running.rs:81`), so every enable-grads traversal reaches `running_mean` / `running_var`:

```rust
let m = BatchNormConfig::new(3).init(&device).freeze();
let g = ParamGroup::ids_from_module(m.clone());
let m = m.set_require_grad_group(g, true);   // flags ignored, per the contract
// training flag: false  -> eval branch reads the running stats
// running_mean has gradient: true
// running_var  has gradient: true
```

`GradientsParamsConverter::visit_float` only gates on `is_require_grad()`, so those register under the buffer's `ParamId` and the optimizer steps them like weights.

Grads-on + flag-off wasn't reachable before this PR: `unfreeze_group` always enabled the flag too, which routes to the train branch where the stats aren't in the loss graph. It's also the standard "freeze BN stats, keep training gamma/beta" recipe.

One-line fix in `running.rs:87`:

```rust
*tensor = tensor_out.set_require_grad(false);
```

crates/burn-core/src/module/base.rs
```diff
@@ -141,31 +189,99 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
         module!(
             map = self,
             ops = |tensor: Tensor<D>| tensor.set_require_grad(false),
             training = false
         )
     }
 
-    /// Set `require_grad` to `false` for every parameter in the given group, leaving the rest
-    /// of the module untouched.
+    /// Unfreeze the whole module tree.
```
antimora, line 273: `unfreeze()` on a validation-mode module half-applies, and you get a model that trains nothing:

```
after valid                 gamma rg=false training=false
after valid+unfreeze        gamma rg=false training=true
after valid+unfreeze+train  gamma rg=false training=true
```

The flag write is host-side so it always lands, but `set_require_grad(true)` no-ops on the inner backend, so `is_active` stays false and `train()` restores false. End state is a module on the autodiff device with dropout armed and weights that get no gradients. No error, the loss just doesn't move.

`LearningResult::model` is now documented as being in validation mode, so "unfreeze then train" is a natural thing for someone to write.

The other order works and is what the tests cover: `freeze().valid().train().unfreeze()`.

Either route `unfreeze` through `from_inner`, or state in the doc that these setters need an autodiff module.

crates/burn-core/src/module/base.rs
```diff
@@ -141,31 +189,99 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
+    /// unchanged, so dropout, noise, randomized activations and batch-normalization running
+    /// statistics retain their current behavior. Use [`freeze`](Module::freeze) to disable both
+    /// gradients and training behavior.
     ///
     /// # Warnings
     ///
     /// This should not be used for inference, use [valid](AutodiffModule::valid) when using
     /// AD modules. This is mostly useful when performing partial finetuning, which is updating only
     /// a small fraction of the parameters instead of finetuning all of them.
     fn no_grad(self) -> Self {
```
antimora, line 243: The warning below still says `no_grad` is "mostly useful when performing partial finetuning", but `freeze`'s new doc claims that role 16 lines down. After this PR they aren't interchangeable: `no_grad` leaves BatchNorm mutating its running stats, which is what the comment at `nn/modules/norm/batch.rs:112` warns about.

Suggest pointing partial finetuning at `freeze` here, and keeping `no_grad` to "gradients only, layer behavior untouched".

crates/burn-core/src/module/base.rs
```diff
@@ -176,11 +292,12 @@ pub trait Module: Clone + Send + core::fmt::Debug {
...
-    /// Set `require_grad` to `true` for every parameter in the given group, leaving the rest
-    /// of the module untouched.
+    /// Unfreeze every module-owned value in the given group, leaving the rest of the module
+    /// untouched.
     ///
-    /// The inverse of [freeze_group](Module::freeze_group): it re-enables gradient tracking for the
-    /// parameters matched by `group`, e.g. to unfreeze a previously frozen module.
+    /// The inverse of [freeze_group](Module::freeze_group): it re-enables gradient tracking for
+    /// matched tensor parameters and enables matched training flags. This also re-enables a flag
+    /// that was disabled directly before the group operation.
```
antimora: "disabled directly before the group operation" is ambiguous: temporal (immediately prior) or by-value (set directly)? Either reading, the macro arm is an unconditional `flag.with_value(true)` for matched flags, so the sentence doesn't add anything over the previous clause.

The caveat that does matter is on `unfreeze` but missing here: matched flags are set to enabled, not restored to a previously captured value. That bites harder in the group form, where an `ids_from_module` group will arm a `Dropout` the user deliberately turned off.

crates/burn-core/src/module/base.rs
```diff
@@ -582,10 +706,18 @@ pub trait ModuleMapper {
 
 /// Module with auto-differentiation backend.
 pub trait AutodiffModule: Module + Send + core::fmt::Debug {
-    /// Returns the same module, but on the inner backend without auto-differentiation.
+    /// Returns the same module on the inner backend without auto-differentiation.
+    ///
+    /// Tensor gradient requirements and module-owned training flags are disabled in the returned
+    /// value while their requested training state is retained internally. Use [`Module::train`]
+    /// to restore that state; [`Module::to_device`] and [`Module::fork`] preserve the disabled
```
antimora: "`to_device` and `fork` preserve the disabled validation state" holds for flags but not tensors, and reordering doesn't help:

```
valid -> train (no move)       weight rg=true
valid -> train -> to_device    weight rg=false
valid -> to_device -> train    weight rg=false
fresh -> to_device(ad)         weight rg=false   <- never validated
```

`to_device` clears require_grad on any module, not just a validated one (`Param::map` recomputes `is_active` from the moved tensor, which comes back detached). That part is pre-existing and not this PR's problem, but the sentence reads as advice, and there's no ordering a user can follow to make it come out right.

burn-book/src/building-blocks/module.md
```diff
@@ -50,26 +50,35 @@ the `Module` derive, you need to be careful to achieve the behavior you want.
...
+
+| Burn API                                            | PyTorch Equivalent                               |
+| --------------------------------------------------- | ------------------------------------------------ |
+| `module.devices()`                                  | N/A                                              |
+| `module.fork(device)`                               | Similar to `module.to(device).detach()`          |
+| `module.to_device(device)`                          | `module.to(device)`                              |
+| `module.set_require_grad(enabled)`                  | `module.requires_grad_(enabled)`                 |
+| `module.set_require_grad_group(group, enabled)`     | N/A                                              |
+| `module.no_grad()`                                  | `module.requires_grad_(False)`                   |
+| `module.freeze()`                                   | `module.requires_grad_(False)` + `module.eval()` |
```
antimora: These two rows conflict with lines 88-89, which map `valid()` to `eval()` and `train()` to `train()`. PyTorch's `train()`/`eval()` don't touch `requires_grad` at all, and `freeze()` here is a sticky per-flag write that `Module::train()` does not undo. `a_frozen_dropout_comes_back_frozen_from_train` asserts exactly that.

A PyTorch user reading this table will write `module.freeze().train()` and expect training behavior back.

Suggest N/A for both rows plus a sentence, and a note that Burn's `train()` is a backend transition, not a mode switch.

crates/burn-nn/src/modules/dropout.rs
```diff
@@ -94,6 +94,66 @@ mod tests {
...
+    fn no_grad_only_disables_tensor_gradients() {
+        let model = trainable_dropout().no_grad();
+
+        assert!(!model.linear.weight.is_require_grad());
+        assert!(model.dropout.training.is_enabled());
+    }
+
+    #[cfg(feature = "std")]
+    #[test]
+    fn require_grad_group_only_changes_matched_tensors() {
```
antimora, line 127: This uses `ids_from_module`, which matches the whole model, so it proves "flags ignored even if matched" but not the other half of the doc promise at `base.rs:210`, that unmatched tensor params are left alone. Needs a two-submodule model with a group over one half.

Two more gaps in the same area:

- `set_require_grad(true)` (whole-module, non-group) is never called with `true` anywhere in the tree, so that macro arm only ever sees `false`.
- `set_require_grad_group` is only tested with id groups, so the `enter_module`/`exit_module` path tracking in the new macro arm is dead under test. Delete both bodies and the suite still passes.

crates/burn-core/src/module/param/mod.rs
```diff
@@ -20,4 +19,3 @@ pub use id::*;
 pub use lora::*;
 pub use reparameterization::*;
 pub use running::*;
```
antimora, line 22: `list_param_ids` was public in v0.21.0 and nothing in this PR needs it gone. The only in-repo caller was the burn-optim test, which got rewritten to avoid it.

`ParamGroup::ids_from_module` isn't a drop-in: it returns a group, not `Vec<ParamId>`, so anyone feeding `GradientsParams::from_params` has no replacement.

Either split it out or call it out as a breaking removal in the PR body.

antimora (commented):
Reviewed the contract split. The separation is the right call, and `no_grad` losing the flag behavior isn't a released break (`Param<Flag>` landed in #5498, after v0.22.0-pre.3), so this restores published semantics rather than changing them.

Three correctness issues, all in the state this PR newly makes reachable: trainable tensors plus a disabled training flag. Details inline, each reproduced locally.

1. `unfreeze()` / `set_require_grad(true)` thaws LoRA base weights, and a gradient reaches the base.
2. The same setters make BatchNorm `running_mean`/`running_var` optimizer targets.
3. `valid().unfreeze().train()` returns a module with flags armed and weights untrainable.

(1) and (2) share a root cause: nothing marks a float param as "never a gradient target", so any enable-grads traversal hits reparameterized bases and running-stat buffers alike.

Rest is docs and test gaps. Style is clean, no em dashes added, and the tests pass locally (burn-core 76, burn-nn 476, burn-optim 108, test_derive_module 20).

crates/burn-core/src/module/base.rs
laggui: This should be an intentional contract: `unfreeze()` and `set_require_grad(true)` explicitly enabled every selected parameter. Preventing LoRA bases from being enbabled would make these APIs silently ignore parameters and prevent intentional full fine-tuning.

The behavior is now documented though, including an example using `ParamGroup::all().exclude(lora_base_group)` to unfreeze everything except LoRA bases.

crates/burn-core/src/module/base.rs
laggui: Good catch. The underlying issue was pre-existing, though it might have only been triggered by a targeted `unfreeze_group` or custom mapper. The behavior is now always the same: `RunningState` is never an optimizer target, so we always force `set_require_grad(false)`.

crates/burn-core/src/module/base.rs
laggui: Addressed. `is_active` now records the requested gradient state even on a validation/plain backend, and `unfreeze` plus the group setters operate on `Param` directly.

`valid().unfreeze().train()` now correctly restores gradients while enabling training flags. Mapping, loading, forks, and device moves also preserve this state. I added regression tests for this lifecycle tracking.

crates/burn-core/src/module/base.rs
laggui: `no_grad` and `freeze` are both useful for partial fine-tuning, the main difference being that freeze also affects module-owned training flags, no just tensor-backed parameters for gradient updates. I updated the docs and added a `Gradient and training state` section to each.

crates/burn-core/src/module/base.rs
laggui: Updated. The ambiguous sentence was removed. `unfreeze_group()` now explicitly states that matched flags are set to enabled. The same contract is documented for `unfreeze()`

crates/burn-core/src/module/base.rs
laggui: The state-loss issue is fixed: `Param::map` preserves the configured gradient setting. So now `valid().to_device(target).train()` restores gradients. `to_device(ad)` still produces intermediary, non-optimizable tensor parameters by design. That behavior hasn't changed. To move an active model for continued training, `fork(target)` should be used instead.

burn-book/src/building-blocks/module.md
laggui: Updated.

crates/burn-nn/src/modules/dropout.rs
laggui: Addressed.

- `require_grad_path_group_only_changes_matched_tensors` covers path tracking and an unmatched tensor
- `set_require_grad_during_validation_is_fully_applied_by_train` covers whole-module `set_require_grad(true)`
- `global_gradient_activation_enables_lora_base_weights` verifies the enable path on an autodiff device

crates/burn-core/src/module/param/mod.rs
laggui: Fair, I'll keep it in for now. I adjusted the docs to explicitly state that it lists the IDs of all *tensor-backed* parameters, excluding flags. This preserves the previous behavior explicitly.

laggui:
@antimora thanks for the review! I addressed your comments and especially clarified the documented contract.

PR #5539, opened 2026-08-31
feat(tensor)!: add multi-axis vector norm variants and update empty max_abs_dims semantics

### LINKED ISSUE #5527: linalg: vector norms are single-axis only; add multi-axis variants
### Feature description

Every public norm in `burn::tensor::linalg` reduces exactly one axis (`crates/burn-tensor/src/tensor/linalg/vector_norm.rs`, rev `4004a44`):

```rust
pub fn vector_norm<const D: usize>(x: Tensor<D>, norm: impl Into<Norm>, dim: impl AsIndex) -> Tensor<D>
pub fn lp_norm<const D: usize>(x: Tensor<D>, p: f64, dim: impl AsIndex) -> Tensor<D>
pub fn l1_norm<const D: usize, K>(x: Tensor<D, K>, dim: impl AsIndex) -> Tensor<D, K>
pub fn l2_norm<const D: usize>(x: Tensor<D>, dim: impl AsIndex) -> Tensor<D>
pub fn max_abs_norm<const D: usize, K>(x: Tensor<D, K>, dim: impl AsIndex) -> Tensor<D, K>
```

There is no multi-axis form, so a norm over more than one axis has to be written out by hand.

### Feature motivation

ONNX `GlobalLpPool` reduces every spatial axis at once: for input `[N, C, D1..Dk]` the output is `[N, C, 1, ..., 1]`, each entry being `sum(|x|^p over D1..Dk)^(1/p)`. That is a vector norm over axes `2..rank`.

Implementing it in burn-onnx (tracel-ai/burn-onnx#417) meant reimplementing burn's own dispatch ladder, because `lp_norm` takes a single axis. They are the same code.

burn, `lp_norm_impl`:

```rust
match p {
    1.0 => l1_norm_impl(x, dim),                        // x.abs().sum_dim(d)
    2.0 => l2_norm_impl(x, dim),                        // x.square().sum_dim(d).sqrt()
    p if is_even_integer(p) => lp_signed_norm(x, .., d),// x.powi_scalar(p).sum_dim(d).powf_scalar(1/p)
    _ => lp_norm_base(x, p, dim),                       // x.abs().powf_scalar(p).sum_dim(d).powf_scalar(1/p)
}
```

burn-onnx codegen, emitting the same thing over `sum_dims`:

```rust
match whole {
    Some(1) => quote! { x.sum_dims(&#dims) },                                     // abs already applied
    Some(2) => quote! { x.square().sum_dims(&#dims).sqrt() },
    Some(n) => quote! { x.powi_scalar(#n).sum_dims(&#dims).powf_scalar(#inv_p) },
    None    => quote! { x.powf_scalar(#p).sum_dims(&#dims).powf_scalar(#inv_p) }, // abs already applied
}
```

That includes rederiving `is_even_integer` (`p.fract() == 0.0 && (p as i64) % 2 == 0`) in order to skip the `abs`, which burn documents internally as "This lets us skip the abs".

The practical cost is that improvements to burn's norms do not reach this path. burn-onnx's `LpNormalization` calls `linalg::l1_norm` / `l2_norm` and inherits them, because LpNormalization is single-axis by definition. `GlobalLpPool` cannot, so it does not.

### Suggest a Solution

Multi-axis siblings following the existing `sum_dim` / `sum_dims` convention in `numeric.rs`:

```rust
pub fn lp_norm_dims<const D: usize, I: AsIndex>(x: Tensor<D>, p: f64, dims: &[I]) -> Tensor<D>
pub fn l1_norm_dims<const D: usize, K, I: AsIndex>(x: Tensor<D, K>, dims: &[I]) -> Tensor<D, K>
pub fn l2_norm_dims<const D: usize, I: AsIndex>(x: Tensor<D>, dims: &[I]) -> Tensor<D>
pub fn vector_norm_dims<const D: usize, I: AsIndex>(x: Tensor<D>, norm: impl Into<Norm>, dims: &[I]) -> Tensor<D>
```

The `*_impl` functions already contain the logic; they would take `dims: &[usize]` and call `sum_dims` instead of `sum_dim`. Like `sum_dims`, these keep the reduced axes at size 1, which is what GlobalLpPool needs.

`torch.linalg.vector_norm` already accepts `dim` as an int or a tuple of ints, so the semantics are settled.

### Alternatives

Folding the single-axis `lp_norm` over each axis in turn **is** mathematically equivalent, so this is an ergonomics and efficiency request rather than a capability gap. Intermediates are non-negative so the inner `abs` is a no-op, and the roots telescope: `((sum_j |x_ij|^p)^(1/p))^p == sum_j |x_ij|^p`. Checked numerically for p = 1, 2, 3 and 2.5, agreeing to ~1e-15.

The cost of folding is an extra `powf_scalar` and root per axis instead of one of each overall, plus the reimplementation above if a caller wants burn's specializations. Writing the ladder out by hand, as burn-onnx does today, avoids the extra passes but duplicates the logic.

### Related

- #1172 (L2 / Lp pooling layer) covers the windowed `nn` module, PyTorch's `LPPool2d`. Related but distinct: this is the functional multi-axis norm, and the global pool is a full-extent reduction rather than a sliding window. A `lp_norm_dims` would be a reasonable building block for it.
- #1538 lists "don't have matrix/vector norm" among its motivations. `vector_norm` has since landed, so that part is stale; the multi-axis gap is what remains for this use case.

Sadik00789:
## Description
Closes #5527.

This PR adds multi-axis variants for all vector norm operations in `burn::tensor::linalg`, following the `sum_dim` / `sum_dims` convention in `numeric.rs`. This enables multi-axis norm reductions while preserving reduced dimensions at size 1 (`keepdim = true`), unblocking multi-axis pooling operations in `burn-onnx` (such as `GlobalLpPool`).

Additionally, empty dimension reductions (`dims = &[]`) follow ONNX `noop_with_empty_axes=true` semantics, ensuring elementwise norm transformations (such as $|x|$ for $L_1/L_2/L_\infty$ or non-zero indicators for $L_0$) execute without collapsing axes.

## Changes
- **`crates/burn-tensor`**:
  - Refactored internal `*_impl` functions (`l0_norm_impl`, `l1_norm_impl`, `l2_norm_impl`, `lp_signed_norm`, `lp_norm_base`, `max_abs_norm_impl`, `min_abs_norm_impl`, `lp_norm_impl`, `vector_norm_impl`) to operate over `dims: &[usize]`.
  - Replaced manual fold reduction in `max_abs_norm_impl` with `x.max_abs_dims(dims)`.
  - Updated `max_abs_dims` in `api/orderable.rs` to return `self.abs()` when `dims.is_empty()`.
  - Ensured empty dimension slices (`dims: &[]`) execute elementwise norm transformations without collapsing dimensions (matching ONNX `noop_with_empty_axes=true` semantics), and documented this across all `*_dims` docstrings.
  - Added public multi-axis APIs: `vector_norm_dims`, `lp_norm_dims`, `l0_norm_dims`, `l1_norm_dims`, `l2_norm_dims`, `max_abs_norm_dims`, and `min_abs_norm_dims`.
  - Converted existing single-axis functions to zero-cost forwarders calling their corresponding `*_dims` variants with `&[dim]`.
  - Updated `cosine_similarity.rs` and re-exports in `linalg/mod.rs`.
- **`crates/burn-backend-tests`**:
  - Added integration tests covering spatial multi-axis reduction shape retention (`[2, 3, 4, 5]` -> `[2, 3, 1, 1]`), even integer $p$ specialization without `abs`, non-integer and odd integer $p$, negative dimensions indexing (`&[-2, -1]`), and single-vs-multi parity.
  - Added regression test cases for empty axes (`&[]`) covering negative values, zeros, odd/even $p$-norms, and multi-dimensional shapes.

## Validation
- `cargo check -p burn-tensor` (clean)
- `cargo test -p burn-tensor --lib linalg` (clean)
- `cargo test --package burn-backend-tests --test tensor -- float::linalg::vector_norm` (14 passed)
- `cargo fmt --check` (clean)
- `cargo clippy -p burn-tensor -p burn-backend-tests --all-targets -- -D warnings` (clean)

### DIFF
--- a/crates/burn-tensor/src/tensor/api/orderable.rs
+++ b/crates/burn-tensor/src/tensor/api/orderable.rs
@@ -774,6 +774,9 @@ where
     /// // [[9.0]]
     /// ```
     pub fn max_abs_dims<I: AsIndex>(self, dims: &[I]) -> Self {
+        if dims.is_empty() {
+            return self.abs();
+        }
         dims.iter()
             .fold(self, |tensor, &dim| tensor.max_abs_dim(dim))
     }
--- a/crates/burn-tensor/src/tensor/api/orderable.rs
+++ b/crates/burn-tensor/src/tensor/api/orderable.rs
@@ -773,6 +773,13 @@ where
     /// println!("{tensor}");
     /// // [[9.0]]
     /// ```
+    ///
+    /// # Empty dimensions
+    ///
+    /// Unlike ordinary reductions such as [`Tensor::max_dims`], this is a composite
+    /// operation equivalent to `self.abs().max_dims(dims)`. If `dims` is empty, no
+    /// reduction is performed, but the elementwise absolute-value operation is
+    /// still applied.
     pub fn max_abs_dims<I: AsIndex>(self, dims: &[I]) -> Self {
         if dims.is_empty() {
             return self.abs();
--- a/crates/burn-tensor/src/tensor/linalg/vector_norm.rs
+++ b/crates/burn-tensor/src/tensor/linalg/vector_norm.rs
@@ -4,6 +4,7 @@ use crate::{
     AsIndex, ElementConversion,
     kind::{Numeric, Ordered},
 };
+use alloc::vec::Vec;
 #[allow(unused_imports)]
 use num_traits::float::Float;
 /// Specifies the type of norm to compute.
@@ -98,6 +99,36 @@ impl From<f64> for Norm {
     }
 }
 
+/// Computes the vector norm of a tensor along specified dimensions.
+///
+/// Generic dispatch wrapper over specialized / optimized norms.
+///
+/// See:
+/// - [torch.linalg.vector_norm](https://pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html)
+/// - [numpy.linalg.vector_norm](https://numpy.org/doc/stable/reference/generated/numpy.linalg.vector_norm.html)
+///
+/// # Arguments
+///
+/// * `x` - The input tensor.
+/// * `norm` - The selected norm.
+/// * `dims` - The dimensions to compute the norm over.
+///   Negative dimensions are supported and count from the end.
+///
+/// # Returns
+///
+/// The vector norm of the input tensor.
+pub fn vector_norm_dims<const D: usize, I: AsIndex>(
+    x: Tensor<D>,
+    norm: impl Into<Norm>,
+    dims: &[I],
+) -> Tensor<D> {
+    let dims: Vec<usize> = dims
+        .iter()
+        .map(|&d| unwrap_dim_index(d.try_dim_index(D), "Vector Norm"))
+        .collect();
+    vector_norm_impl(x, norm, &dims)
+}
+
 /// Computes the vector norm of a tensor along a specified dimension.
 ///
 /// Generic dispatch wrapper over specialized / optimized norms.
@@ -121,8 +152,43 @@ pub fn vector_norm<const D: usize>(
     norm: impl Into<Norm>,
     dim: impl AsIndex,
 ) -> Tensor<D> {
-    let dim = unwrap_dim_index(dim.try_dim_index(D), "Vector Norm");
-    lp_norm_impl(x, norm.into().to_exponent(), dim)
+    vector_norm_dims(x, norm, &[dim])
+}
+
+fn vector_norm_impl<const D: usize>(
+    x: Tensor<D>,
+    norm: impl Into<Norm>,
+    dims: &[usize],
+) -> Tensor<D> {
+    lp_norm_impl(x, norm.into().to_exponent(), dims)
+}
+
+/// Computes the general ``L(p)`` norm of a tensor along specified dimensions.
+///
+/// Uses the specialized implementations for:
+/// * 0.0
+/// * 1.0
+/// * 2.0
+/// * 2 * N for integral N,
+/// * f64::INFINITY,
+/// * f64::NEG_INFINITY,
+///
+/// # Arguments
+///
+/// * `x` - The input tensor.
+/// * `p` - The exponent of the Lp norm.
+/// * `dims` - The dimensions to compute the norm over.
+///   Negative dimensions are supported and count from the end.
+///
+/// # Returns
+///
+/// The ``L(p)`` norm of the input tensor.
+pub fn lp_norm_dims<const D: usize, I: AsIndex>(x: Tensor<D>, p: f64, dims: &[I]) -> Tensor<D> {
+    let dims: Vec<usize> = dims
+        .iter()
+        .map(|&d| unwrap_dim_index(d.try_dim_index(D), "Lp Norm"))
+        .collect();
+    lp_norm_impl(x, p, &dims)
 }
 
 /// Computes the general ``L(p)`` norm of a tensor along a specified dimension.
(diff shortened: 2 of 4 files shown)


crates/burn-tensor/src/tensor/linalg/vector_norm.rs
```diff
@@ -301,14 +497,44 @@ where
...
 }
 
-fn max_abs_norm_impl<const D: usize, K>(x: Tensor<D, K>, dim: usize) -> Tensor<D, K>
+fn max_abs_norm_impl<const D: usize, K>(x: Tensor<D, K>, dims: &[usize]) -> Tensor<D, K>
 where
     K: Ordered,
 {
-    x.max_abs_dim(dim)
+    dims.iter()
+        .fold(x.abs(), |tensor, &dim| tensor.max_dim(dim))
```
laggui: We already have `tensor.max_abs_dims`, we can reuse it here

laggui (changes requested):
I don't think we should return the input unchanged for empty dimensions.

This bypasses the norm itself: for `x = [-2, 0]`, L1/L2/L_inf should at least produce `[2, 0]` and L0 `[1, 0]`. The current positive, non-zero test input hides most of these differences.

Right now, no-op matches none of the existing semantics. PyTorch treats empty dims as a full reduction over all dimensions, like ONNX `noop_with_empty_axes=false`. ONNX `noop_with_empty_axes=true` reduces no dimensions, but the other norm operations still execute.

Given the current `sum_dims(&[])` behavior, I think we should match the `noop_with_empty_axes=true` behavior and explicitly document it.

crates/burn-tensor/src/tensor/linalg/vector_norm.rs
Sadik00789: Addressed review feedback in the latest push:

- **Refactored `max_abs_norm_impl`**: Replaced manual fold reduction with `x.max_abs_dims(dims)`.
- **Aligned with ONNX `noop_with_empty_axes=true`**: Removed `dims.is_empty()` early returns that bypassed the norm transformation. When `dims` is empty, reduction primitives now preserve dimensions while applying elementwise operations ($|x|$ for $L_1, L_2, L_\infty, L_p$ and the indicator mask for $L_0$).
- **`orderable.rs` Consistency**: Updated `max_abs_dims` to return `self.abs()` when `dims.is_empty()`.
- **Documentation**: Documented empty-dimension behavior across all `*_dims` docstrings.
- **Tests**: Added test cases covering negative values, zeros, odd/even $p$-norms, and multi-dimensional tensors under empty `dims`.

crates/burn-tensor/src/tensor/api/orderable.rs
```diff
@@ -774,6 +774,9 @@ where
     /// // [[9.0]]
     /// ```
     pub fn max_abs_dims<I: AsIndex>(self, dims: &[I]) -> Self {
+        if dims.is_empty() {
+            return self.abs();
+        }
```
laggui, line 786: That's breaking change in behavior for empty dims, but I'm ok with that. It aligns with the behavior that empty dims skips only the reduction.

That said, it should be noted in the method's docs. Something like:
```rust
/// # Empty dimensions
///
/// Unlike ordinary reductions such as [`Tensor::max_dims`], this is a composite
/// operation equivalent to `self.abs().max_dims(dims)`. If `dims` is empty, no
/// reduction is performed, but the elementwise absolute-value operation is
/// still applied.
```

laggui (commented):
LGTM! Thanks for addressing my comments. Just one small doc change missing.

crates/burn-tensor/src/tensor/api/orderable.rs
Sadik00789: Added the doc note to max_abs_dims explaining the composite behavior and empty dimension semantics. Ready for final review!

laggui (approved):

Issue #199, opened 2023-03-06
Better error messages for incorrect shapes


antimora:
### Feature description

The compiler or runtime should explain in a clear language what is wrong and offer a solution, similar what Rust compiler does.

### Feature motivation

It's very common to miscalculate/misconfigure NN with incorrect shapes. Sometimes it's part of the process. However, the Burn's error messages are inconsistent across backends and NN modules. Different backends bubble up errors differently. For example, torch tells what size is expected vs given, but ndarray just says wrong size. Additionally, it's very hard to know exactly which line is causing the problem. All these errors will be especially difficult for beginners. 

### (Optional) Suggest a Solution

Compiler does an excellent job for dimension mismatches. It would be great if compiler could identify the issue early. And if it's possible, then the runtime should check the inputs before passing to the backends, perhaps in debug mode or always.

antimora:
from @nathanielsimard:

> > How much of your work relates to this issue: https://github.com/burn-rs/burn/issues/199? by @antimora 
> 
> It's strongly related to the issue, but it's not a complet solution yet.

https://github.com/burn-rs/burn/pull/283

ax1s-x1zz:
I'd like to take a shot at improving the ndarray backend's shape mismatch error messages for this issue.

## Current state

Looking at the current code, the ndarray backend already reports some shape problems clearly (e.g. matmul's `"Dimensions are incompatible for matrix multiplication: LHS columns ({}) != ({})"`), but three cases still panic with little or no context:

1. **matmul batch broadcast** (`output_shape`) — `panic!("Dimensions differ and cannot be broadcasted.")` — doesn't say which dimension or the values.
2. **matmul `cross`** — `panic!("Tensors are not broadcastable along dimension {}")` — says the dimension but not the conflicting sizes.
3. **`expand`** (broadcast to a target shape) — `.expect("The shapes should be broadcastable")` — doesn't show the source/target shapes.
4. **`reshape`** — `panic!("Shape should be compatible shape={dim:?}: {err:?}")` — shows the target but not the original shape; the non-standard-layout path just calls `.unwrap()` with no message at all.

## Proposed change (ndarray only, small & focused)

Update these four messages to include the offending dimension and both expected/actual values:

- matmul broadcast: `"Dimensions are incompatible for broadcasting along dimension {i}: LHS ({l_dim}) vs RHS ({r_dim})"`
[...]

PR #5542, opened 2026-09-01
feat(ndarray): improve shape mismatch error messages

ax1s-x1zz:
## Summary

Improves the ndarray backend's shape mismatch error messages so users can pinpoint the mismatched layer without digging through tensor shape state. Closes out the ndarray half of #199.

## Changes

- **matmul broadcast** (`output_shape`): `"Dimensions differ and cannot be broadcasted."` → now reports the offending dimension and both sizes:
  `"Dimensions are incompatible for broadcasting along dimension {i}: LHS ({l_dim}) vs RHS ({r_dim})"`
- **matmul `cross`**: `"Tensors are not broadcastable along dimension {}"` → now includes the conflicting lhs/rhs values:
  `"Tensors are not broadcastable along dimension {i}: lhs ({l}) vs rhs ({r})"`
- **`expand`**: `.expect("The shapes should be broadcastable")` → now shows the source and target shapes:
  `"Tensors are not broadcastable: cannot expand {src_shape:?} to {target_shape:?}"`
- **`reshape`**: reports the original and target shapes; also gives the non-standard-layout path a real message instead of a bare `.unwrap()`:
  `"Cannot reshape {orig_shape:?} to {target_shape:?}: {err:?}"`

## Why

These are the most common shape mistakes when wiring up a model (mismatched layer dims, bad broadcast). The old messages only said "cannot be broadcasted" or "shape should be compatible" without the actual sizes, so a beginner has to reconstruct what went wrong from context.

## Testing

- `cargo build -p burn-ndarray` — passes
- `cargo test -p burn-ndarray` — 38 tests pass
- `cargo test -p burn-ndarray matmul` — 4 tests pass, including the updated `test_output_shape_non_broadcast` which now asserts the new message

Ref: #199

### DIFF
--- a/crates/burn-ndarray/src/ops/base.rs
+++ b/crates/burn-ndarray/src/ops/base.rs
@@ -551,7 +551,13 @@ where
     pub(crate) fn expand(tensor: SharedArray<E>, shape: Shape) -> SharedArray<E> {
         tensor
             .broadcast(shape.into_dimension())
-            .expect("The shapes should be broadcastable")
+            .unwrap_or_else(|| {
+                panic!(
+                    "Tensors are not broadcastable: cannot expand {:?} to {:?}",
+                    tensor.shape(),
+                    shape
+                )
+            })
             // need to convert view to owned array because NdArrayTensor expects owned array
             // and try_into_owned_nocopy() panics for broadcasted arrays (zero strides)
             .into_owned()
--- a/crates/burn-ndarray/src/ops/matmul.rs
+++ b/crates/burn-ndarray/src/ops/matmul.rs
@@ -163,7 +163,10 @@ fn output_shape(lsh: &[usize], rsh: &[usize]) -> (Shape, Strides, Strides, Strid
             l_strides.push(cur_l_stride);
             r_strides.push(0);
         } else {
-            panic!("Dimensions differ and cannot be broadcasted.");
+            panic!(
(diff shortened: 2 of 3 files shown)


laggui (changes requested):
Thanks for the contribution. I think #199 is now stale and likely too broad. It conflates tensor-operation validation with higher-level NN shape contracts and needs to be reframed accordingly.

Burn has since moved toward validating public tensor contracts in the tensor API before backend dispatch (`TensorCheck` plus validation performed during argument conversion), although some gaps remain.

Updating messages only for the ndarray backend (now deprecated) doesn't really follow that direction. A focused improvement such as adding matmul batch-broadcast validation to `TensorCheck` would be more useful, for example.

There is also a performance concern: `shape().to_vec()` adds a heap allocation to every successful ndarray reshape solely to support the error path.

ax1s-x1zz:
Thanks for the detailed review — this makes sense.

Two of your points are clear to me and I agree:
1. **ndarray is deprecated**, so backend-local message fixes have limited value.
2. **`shape().to_vec()` allocates on every successful reshape** — that was a real perf regression on the hot path.

I'll close this PR and refocus on the direction you suggested: moving matmul batch-broadcast validation into `TensorCheck` so it applies at the public tensor API layer (all backends benefit, not just ndarray). I'll investigate where matmul's batch-dimension broadcast is currently validated and propose a focused change there.

PR #5545, opened 2026-09-01
perf(conv): accumulate direct convolution channels in vector registers on CPU

ThierryCantin-Demers:
The direct convolution's channel loop multiplied a vector of input channels by a vector of
weights and folded the product into `sum` one lane at a time, so every tap ended in `NIn`
adds on the same lane of `sum`. For a 512 channel convolution that is a serial chain as long
as the whole reduction, and the multiply never fuses into an add.

Each output channel now accumulates into its own vector across the channel loop, and the
lanes fold into `sum` once per tap. The adds in the loop are independent, so it becomes a
chain of multiply-adds.

## Where it applies

The accumulating form reads the input vector once per output channel where the old form
shared one read, so it is taken only where the chain it breaks is actually paid:

```rust
let accumulate_lanes = input.client.properties().hardware.plane_size_max == 1
    && vector_size_in > 1
    && weight.meta.shape()[dim_c] > vector_size_in as usize;
```

A plane of one unit has nothing to hide a dependency chain behind. Today that is the CPU
runtime; CUDA, HIP and wgpu report their warp or subgroup size, so every GPU compiles the
same kernel as before. The other two conditions keep the old body for a single lane and for
a channel loop that runs once, which is where stems and depthwise convolutions land.

## Results

cubecl-cpu on a Ryzen 5700X, medians of four runs per arm:

| model | batch | before | after | |
|---|---|---|---|---|
| ResNet-50 | 1 | 165.1 ms | 145.1 ms | -12.1% |
| ResNet-50 | 8 | 818.7 ms | 604.5 ms | -26.2% |
| MobileNetV3-Small | 1 | 37.0 ms | 37.0 ms | flat |
| MobileNetV3-Small | 8 | 64.7 ms | 64.1 ms | flat |

Per layer, on the 5700X and on a Xeon E5-2620 v4, every convolution wider than one vector
runs 2 to 3.5x faster and the gated ones are unchanged. Peak resident memory is unchanged.
Summation order changes, so results move within float rounding: max absolute logit
difference on ResNet-50 is 7.6e-6, the same as before.

## Testing

`conv` and `conv_transpose` suites green on cubecl-cpu, CUDA (Ada and Turing) and Vulkan
(RDNA1 and Intel integrated). The GPU kernels are unchanged, verified by diffing the
compiled CUDA source of the gated variant against main's.

### DIFF
--- a/crates/burn-cubecl/src/kernel/conv/direct.rs
+++ b/crates/burn-cubecl/src/kernel/conv/direct.rs
@@ -365,12 +365,12 @@ pub fn conv_direct<R: CubeRuntime, const N: usize>(
     // Use channels_per_group instead of in_channels to avoid issues here
     let vector_size_in = max_vector_size(&weight);
 
-    // Two things have to hold for a vector accumulator to pay. There has to be a vector: at
-    // one lane, accumulating into it is exactly as serial as accumulating into `sum`, and
-    // only the extra reads are left. And the channel loop has to run more than once, or
-    // there is nothing to amortize the reduction over.
-    let accumulate_lanes =
-        vector_size_in > 1 && weight.meta.shape()[dim_c] > vector_size_in as usize;
+    // The vector accumulator breaks a dependency chain that only a single-unit plane pays in
+    // full: a wide plane hides it behind its other lanes and is left with the extra input read
+    // per output channel. It also needs a vector, and a channel loop that runs more than once.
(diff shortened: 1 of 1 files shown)


laggui (approved):
LGTM! The lane-wise accumulation seems to bring a nice improvement to CPU performance.

> Summation order changes, so results move within float rounding

Seems to be within expected floating-point tolerance, so all good.

PR #5554, opened 2026-09-03
feat(tensor): add assert_shape! and debug_assert_shape! macros

antimora:
### Checklist

### Related Issues/PRs

Inspired by [burn-contracts](https://github.com/crutcher/burn-contracts) by @crutcher. Validated in my own model code before porting.

### Changes

Adds `assert_shape!` and `debug_assert_shape!` to `burn-tensor` and the prelude. Each slot is a `usize` expression the axis must equal, `_`, or one `..` for any number of axes. Rank is checked at compile time through `let dims: [usize; N] = Tensor::dims(..)`, so a wrong rank or a non-tensor argument does not compile; with `..` it becomes a runtime minimum, which is what a method generic over `const D: usize` needs. `debug_assert_shape!` compiles out like `debug_assert!`.

```rust
let [batch_size, seq_length, _] = input.dims();
assert_shape!(input, [_, _, self.d_model]);              // field, `_` skips an axis
assert_shape!(mask, [batch_size, seq_length]);           // names from dims()
assert_shape!(flat, [batch_size * seq_length, 256]);     // any usize expression
assert_shape!(patch, [3 * 2, 2]);                        // literals
assert_shape!(hidden, [.., self.d_model]);               // any rank, works for Tensor<D>
assert_shape!(image, [_, self.num_channels, ..]);        // channels first, any spatial rank
// panics with: assert_shape!(mask, [batch_size, seq_length]): axis 1 expected 128, got 64 (dims [8, 64])
```

The macros are `macro_rules!` wrappers in `burn-tensor` over proc macros in `burn-derive`, which becomes a `burn-tensor` dependency (publish order updated). `burn-nn` losses, attention, positional and rotary encoding, the norm layers, `Linear` and the feed-forward block use them at their boundaries; multi-head attention now requires key and value to share the query's batch size, as documented. Generic-rank losses keep hand-written checks. Examples and the book (new Forward Contract section) follow.

### Testing

27 integration tests in debug and release, 9 doctests with 4 compile-fail cases, new `should_panic` tests in `burn-nn`, and `cargo run-checks`.

### DIFF
--- a/burn-book/src/building-blocks/module.md
+++ b/burn-book/src/building-blocks/module.md
@@ -62,13 +62,9 @@ available from the prelude, check those sizes at the boundary of a method:
 - `assert_shape!` checks the axes of a tensor. It stays on in release builds.
 - `debug_assert_shape!` is the same check, compiled out unless debug assertions are enabled.
 
-Each takes a tensor and a bracketed pattern with one slot per axis:
-
-| Slot                   | Meaning                                               |
-| ---------------------- | ----------------------------------------------------- |
-| `=expr`                | The axis must equal this in-scope `usize` expression. |
-| `80` (integer literal) | The axis must equal this value.                       |
-| `_`                    | Skip the axis.                                        |
+Each takes a tensor and a bracketed pattern with one slot per axis. A slot is either `_`, which
+skips the axis, or any `usize` expression the axis must equal: a name from `dims()`, a config field,
+a literal.
 
 The pattern length is the expected rank. Since `tensor.dims()` returns `[usize; D]`, a pattern whose
 length differs from `D` is a compile error rather than a runtime panic, and so is passing anything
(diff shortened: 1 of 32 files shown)


antimora:
Removing `unpack_shape` after more thought

burn-book/src/building-blocks/module.md
```diff
@@ -29,6 +29,106 @@ impl PositionWiseFeedForward {
...
+///
+/// # Shapes
+///
+/// - input: `[batch_size, seq_length, d_model]`
+/// - output: `[batch_size, seq_length, d_model]`
+pub fn forward(&self, input: Tensor<3>) -> Tensor<3> {
+```
+
+Use the same convention for your own modules. It is the first thing a caller reads, and it is the
+reference for the assertions below.
```
laggui, line 53: One limitation worth documenting is that exact patterns can't be used in functions generic over `const D: usize`, like the `PositionWiseFeedForward` example immediately above this section.

A useful follow-up could support `assert_shape!(input, [.., self.d_model])` where `..` skips any number of leading dimensions. Exact patterns could retain their compile-time rank check.

laggui (commented):
LGTM overall! I also agree with the `unpack_shape` removal, I think it requires more consideration.

Just one small note/suggestion below

burn-book/src/building-blocks/module.md
antimora: Implemented in 23157b669: one `..` per pattern stands for any number of axes, slots after it index from the end, and the rank check becomes a runtime minimum while exact patterns keep the compile-time one. Applied it to `PositionWiseFeedForward`, `LayerNorm` and `RmsNorm`, and documented the limitation in the book section.

antimora: Follow-up in e90ead81a: applied `..` to the remaining rank-generic modules too, group, instance, batch and local response norm, rotary encoding and `Linear`, replacing the hand-written rank and channel checks. The channels-first form reads well: `assert_shape!(input, [_, self.num_channels, ..])`.

laggui (approved):
Nice 🎉

PR #5555, opened 2026-09-03
feat(tensor): validate matmul batch broadcast in TensorCheck

ax1s-x1zz:
## Summary

Follow-up to the closed #5542, addressing the review feedback from #199.

Instead of improving error messages only on the (now deprecated) ndarray backend, this moves the matmul **batch-dimension broadcast** validation into `TensorCheck`, the public tensor API check layer that runs *before* backend dispatch. Every backend (including ndarray) benefits from consistent errors.

## Current gap

`TensorCheck::matmul` only validated:
- device equality, and
- the inner dimension (`lhs[..., K] @ rhs[..., K]`).

An incompatible batch dimension (neither equal nor `1`) was **not** caught at the tensor API layer, so it fell through to backend-specific panics with inconsistent messages.

## Change

In `TensorCheck::matmul`, add a broadcastability loop over the leading `D - 2` batch dimensions, mirroring the existing `cross()` check. On failure it produces a proper `Tensor Operation Error`:

```
Operation: 'Matmul'
  Reason:
    1. Tensors are not broadcastable along batch dimension 0: 4 and 2.
       Lhs shape [4, 5, 3], rhs shape [2, 3, 7].
```

## Testing

- `cargo test -p burn-backend-tests --features ndarray --test tensor` — 1794 passed, 0 failed
- Added regression test `float_should_panic_when_batch_dimensions_are_not_broadcastable` in `burn-backend-tests/tests/tensor/float/ops/matmul.rs`
- Existing broadcast-success cases (`test_float_matmul_broadcast_1`, `test_float_matmul_broadcast_4d`, etc.) still pass, confirming valid broadcasts are unaffected

Ref: #199

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/matmul.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/matmul.rs
@@ -466,26 +466,12 @@ fn float_should_panic_when_inner_dimensions_are_not_equal() {
 }
 
 #[test]
-#[should_panic]
+#[should_panic(expected = "Tensors are not broadcastable along batch dimension")]
 fn float_should_panic_when_batch_dimensions_are_not_broadcastable() {
     let device = Default::default();
-    // Inner dims match (2 == 2) but batch dims [4, 5] vs [2, 3] cannot broadcast.
-    let tensor_1 = TestTensor::<3>::from_data(
-        [
-            [[1.0, 7.0], [2.0, 3.0], [1.0, 5.0], [2.0, 8.0], [3.0, 4.0]],
-            [[1.0, 7.0], [2.0, 3.0], [1.0, 5.0], [2.0, 8.0], [3.0, 4.0]],
-            [[1.0, 7.0], [2.0, 3.0], [1.0, 5.0], [2.0, 8.0], [3.0, 4.0]],
-            [[1.0, 7.0], [2.0, 3.0], [1.0, 5.0], [2.0, 8.0], [3.0, 4.0]],
-        ],
-        &device,
-    );
-    let tensor_2 = TestTensor::<3>::from_data(
-        [
-            [[4.0, 7.0], [2.0, 3.0], [1.0, 5.0]],
-            [[4.0, 7.0], [2.0, 3.0], [1.0, 5.0]],
-        ],
-        &device,
-    );
+    // Inner dims match (3 == 3) but batch dims [4] vs [2] cannot broadcast.
+    let tensor_1 = TestTensor::<3>::zeros([4, 5, 3], &device);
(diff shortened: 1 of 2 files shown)


ax1s-x1zz:
Hi @laggui — here's the follow-up based on your review of #5542. 

Instead of touching only the deprecated ndarray backend, I added the matmul batch-dimension broadcast validation to `TensorCheck` (the public tensor API check layer), so all backends get a consistent `Tensor Operation Error` before reaching backend dispatch. The perf concern from the last review (`shape().to_vec()` on the hot path) is also avoided since this version only reads shapes that `TensorCheck` already has.

Happy to adjust the message wording or scope if you'd like it narrower.

crates/burn-backend-tests/tests/tensor/float/ops/matmul.rs
```diff
@@ -464,3 +464,29 @@ fn float_should_panic_when_inner_dimensions_are_not_equal() {
 
     tensor_3.into_data().assert_eq(&expected, false);
 }
+
+#[test]
+#[should_panic]
+fn float_should_panic_when_batch_dimensions_are_not_broadcastable() {
```
laggui, line 470: The current test is a false positive, the inputs have shapes `[4, 5, 2]` and `[2, 3, 2]` so the existing inner dimension validation already makes this panic.

We should use `expected = "Tensors are not broadcastable along batch dimension"` here, and update the dims to something like `[4, 5, 3]` and `[2, 3, 7]` so the test specifically covers this change.

We can also avoid creating the large data literals by using `Tensor::zeros` and/or `Tensor::ones` with the provided shapes.

laggui (commented):
LGTM! Thanks for addressing my comments on the previous PR.

Just need to update the test

ax1s-x1zz:
@laggui — updated the test as suggested. It now uses shapes `[4,5,3]` and `[2,3,7]` so the inner dimensions match (3 == 3) and only the batch dimension is incompatible, asserts `expected = "Tensors are not broadcastable along batch dimension"` to make sure this specific check is what panics, and builds the tensors with `zeros` instead of data literals.

Verified locally:
- `cargo test -p burn-backend-tests --features ndarray --test tensor` — 1794 passed, 0 failed
- The test now fails if the batch-broadcast check is removed (previously it passed for the wrong reason via the inner-dimension check)

laggui (approved):

ax1s-x1zz:
@laggui Thanks for the approval. Looks like the macOS CI runner failed with an infrastructure error (exit status 126: cannot execute binary file). Could you please trigger a re-run for the checks when you have a moment?

laggui:
The failure is unrelated, no need to re-run :+1:

PR #5558, opened 2026-09-03
Fix invalid path filter regex handling

### LINKED ISSUE #5480: burn-store: PathFilter::with_regex swallows a regex error, so a bad pattern silently saves nothing
## Summary

`PathFilter::with_regex` discards `Regex::new`'s error. A filter whose only pattern failed to compile becomes an empty filter, and an empty filter matches nothing - so a save with a typo'd pattern writes an empty container and returns `Ok(())`.

## Detail

`crates/burn-store/src/filter.rs:67-73`:

```rust
pub fn with_regex<S: AsRef<str>>(mut self, pattern: S) -> Self {
    if let Ok(regex) = Regex::new(pattern.as_ref()) {
        self.regex_patterns.push(regex);
    }
    // TODO: Consider returning Result to handle regex compilation errors
    self
}
```

`with_regexes` (`:77-88`) does the same per pattern. The dropped error then meets `matches_with_container_path_str` (`:146-155`):

```rust
if self.match_all { return true; }
// If no filters are configured, match nothing
if self.is_empty() { return false; }
```

so "the pattern did not compile" and "the user asked for nothing" become indistinguishable, and the second answer is silence.

`BurnpackStore::collect_from` passes the filter straight into `module.collect(...)`, so every parameter is filtered out and the writer is handed an empty tensor list.

The same crate already treats this class of error the other way: `SafetensorsStore::with_key_remapping` (`crates/burn-store/src/safetensors/store.rs:298-310`) and `BurnpackStore::with_remap_pattern` (`crates/burn-store/src/burnpack.rs:294-298`) both `.expect("Invalid regex pattern")`.

## Reproduction

```rust
let mut store = BurnpackStore::from_file(&path).with_regex(r"[");  // not a valid regex
let saved = model.save_into(&mut store);
println!("save with invalid regex: {:?}", saved.is_ok());

let mut reader = BurnpackStore::from_file(&path);
println!("tensors written: {:?}", reader.keys().unwrap());
```

```
save with invalid regex: true
tensors written: []
```

## Suggested fix

Pick one of the two behaviours the crate already uses and apply it consistently. Either:

- return `Result<Self, regex::Error>` from `with_regex`/`with_regexes` (matching `KeyRemapper::add_pattern`, which already does this), or
- panic on an invalid pattern (matching `with_key_remapping`).

The `Result` form is better since these are builder methods on a user-facing configuration and a bad pattern is a recoverable input error, but either beats the current silent drop.

Independently: it may be worth having the stores refuse to write a container with zero tensors when a filter was configured, since "the filter matched nothing" is almost never what the caller meant.

Mikyx-1:
### Checklist

### Related Issues/PRs

Closes #5480.

### Changes

`PathFilter::with_regex` and `with_regexes` previously discarded regex compilation errors. A malformed pattern therefore became an empty filter and could silently produce an empty checkpoint.

Compile each pattern eagerly and fail with the existing `Invalid regex pattern` message. This keeps the builder API compatible and matches the behavior already used by the store remapping builders. The panic contract is now documented on both public methods.

Regression tests cover malformed single patterns and malformed patterns in a collection.

### Testing

- `cargo run-checks`
- `cargo test -p burn-store`

### DIFF
--- a/crates/burn-store/src/burnpack.rs
+++ b/crates/burn-store/src/burnpack.rs
@@ -259,6 +259,10 @@ impl BurnpackStore {
     }
 
     /// Add regex pattern to filter
+    ///
+    /// # Panics
+    ///
+    /// Panics if `pattern` is not a valid regular expression.
     #[cfg(feature = "std")]
     pub fn with_regex(mut self, pattern: &str) -> Self {
         let filter = self.filter.unwrap_or_default();
--- a/crates/burn-store/src/filter.rs
+++ b/crates/burn-store/src/filter.rs
@@ -63,16 +63,22 @@ impl PathFilter {
     }
 
     /// Add a regex pattern for matching paths
+    ///
+    /// # Panics
+    ///
+    /// Panics if `pattern` is not a valid regular expression.
     #[cfg(feature = "std")]
     pub fn with_regex<S: AsRef<str>>(mut self, pattern: S) -> Self {
-        if let Ok(regex) = Regex::new(pattern.as_ref()) {
-            self.regex_patterns.push(regex);
-        }
-        // TODO: Consider returning Result to handle regex compilation errors
+        let regex = Regex::new(pattern.as_ref()).expect("Invalid regex pattern");
+        self.regex_patterns.push(regex);
         self
     }
 
     /// Add multiple regex patterns
+    ///
+    /// # Panics
+    ///
+    /// Panics if any pattern is not a valid regular expression.
     #[cfg(feature = "std")]
     pub fn with_regexes<I, S>(mut self, patterns: I) -> Self
     where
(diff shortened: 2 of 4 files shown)


laggui (commented):
The new panic behavior also propagates through the public wrapper methods, as noted by the Copilot review.

The panic documentation should be reflected there as well.

Mikyx-1:
@laggui Thanks for the review. I added `# Panics` documentation to all public forwarding APIs (`BurnpackStore`, `PytorchStore`, `SafetensorsStore`, and `PathFilter::from_regex_patterns`) in 903fe8a80. Could you please take another look?

laggui (approved):

CI failures are unrelated to these changes, they come from some other shenanigans on main 😅

PR #5564, opened 2026-09-04
feat(tensor): apply TensorCheck to remainder, powi, powf, hypot, atan2

### LINKED ISSUE #5562: Some element-wise binary ops skip the `binary_ops_ew` TensorCheck
## Some element-wise binary ops skip the `binary_ops_ew` TensorCheck

Following up on the close of #199 — @laggui noted that `TensorCheck` should enforce local tensor-operation preconditions (ranks, broadcasting, matmul dims, devices) and that some gaps remain.

While scanning the public tensor API, I found that **most element-wise binary ops call `check!(TensorCheck::binary_ops_ew(...))`** (device + broadcastability), e.g.:

```rust
pub fn add(self, other: Self) -> Self {
    check!(TensorCheck::binary_ops_ew("Add", &self, &other));
    Self::new(K::add(self.primitive, other.primitive))
}
```

...but several others go straight to the backend without that check:

- `remainder` — `Self::new(K::remainder(...))`
- `powi` — `Self::new(K::powi(...))`
- `hypot` — `Self::new(hypot_impl(...))`
- `atan2` — `Tensor::new(K::atan2(...))`

So a device or broadcast mismatch is caught with a clear `Tensor Operation Error` for `add`/`sub`/`mul`/`div`, but for `remainder`/`powi`/`hypot`/`atan2` it falls through to backend-specific panics with inconsistent messages.

## Proposal

Add `check!(TensorCheck::binary_ops_ew("<Op>", &self, &other))` to these four ops, matching the existing pattern. This is a small, focused change with the same shape as the recently merged matmul batch-broadcast validation (#5555).

I can open a PR with this if it seems reasonable.

ax1s-x1zz:
## Summary

Closes #5562. Element-wise binary ops like `add`, `sub`, `mul`, and `div` already validate device and broadcast compatibility via `check!(TensorCheck::binary_ops_ew(...))`. Several others went straight to the backend without that check, so a device or broadcast mismatch fell through to backend-specific panics instead of a consistent `Tensor Operation Error`.

## Change

Add `check!(TensorCheck::binary_ops_ew(...))` to five ops, matching the existing pattern:

- `remainder` (numeric)
- `powi` (numeric, int)
- `powf` (float) — same gap as `powi`, flagged in review
- `hypot` (float)
- `atan2` (float)

## Testing

- Regression test per op asserting the `Tensor Operation Error` message: `should_panic_{remainder,powi,powf,hypot,atan2}_incompatible_shapes`
- **Verified each test fails if the check is removed** (not passing for the wrong reason)
- Verified valid broadcasts still pass (e.g. existing `powf_broadcasted`, new `should_support_remainder_broadcast`)
- `cargo test -p burn-backend-tests --features ndarray --test tensor` — 1819 passed, 0 failed

Ref: #5562

### DIFF
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -170,6 +170,7 @@ where
     ///
     /// `y = x2 % x1`
     pub fn remainder(self, other: Self) -> Self {
+        check!(TensorCheck::binary_ops_ew("Remainder", &self, &other));
         Self::new(K::remainder(self.primitive, other.primitive))
     }
 
@@ -711,6 +712,7 @@ where
     /// // [[1, -8, 81], [5, 81, 216]]
     /// ```
     pub fn powi(self, other: Self) -> Self {
+        check!(TensorCheck::binary_ops_ew("Power", &self, &other));
         Self::new(K::powi(self.primitive, other.primitive))
     }
 
--- a/crates/burn-tensor/src/tensor/api/numeric.rs
+++ b/crates/burn-tensor/src/tensor/api/numeric.rs
@@ -712,7 +712,7 @@ where
     /// // [[1, -8, 81], [5, 81, 216]]
     /// ```
     pub fn powi(self, other: Self) -> Self {
-        check!(TensorCheck::binary_ops_ew("Power", &self, &other));
+        check!(TensorCheck::binary_ops_ew("Powi", &self, &other));
         Self::new(K::powi(self.primitive, other.primitive))
     }
 
--- a/crates/burn-backend-tests/tests/tensor/float/ops/hypot.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/hypot.rs
@@ -2,6 +2,21 @@ use super::*;
 use burn_tensor::{TensorData, Tolerance};
 
 #[test]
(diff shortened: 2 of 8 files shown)


crates/burn-tensor/src/tensor/api/numeric.rs
```diff
@@ -711,6 +712,7 @@ where
     /// // [[1, -8, 81], [5, 81, 216]]
     /// ```
     pub fn powi(self, other: Self) -> Self {
+        check!(TensorCheck::binary_ops_ew("Power", &self, &other));
```
laggui:
```suggestion
        check!(TensorCheck::binary_ops_ew("Powi", &self, &other));
```

laggui (approved):
LGTM!

Looks like the additional powi test flagged an existing issue on Metal:
```
  thread 'tensor::int::ops::powi::should_support_powi_broadcast' (38668) panicked at crates/burn-backend-tests/tests/common/../tensor/int/ops/powi.rs:13:24:
  Tensors are not eq:
    => Position 2: 8 != 9
    => Position 5: 26 != 27
  
  
  failures:
      tensor::int::ops::powi::should_support_powi_broadcast
```

Cube's `int_powi` currently use the default fallback which converts ints to floats, calls powf, and then converts back to ints by truncation. The results appear to get truncated.

Not related to this PR though, I'll investigate and address in a follow-up.

PR #5569, opened 2026-09-04
fix(nn): exclude pad tokens from cross-entropy normalization

### LINKED ISSUE #5568: CrossEntropyLoss includes pad tokens in the normalization denominator
**Describe the bug**

`CrossEntropyLoss` masks padded samples out of the numerator, but still includes them in the normalization denominator. Adding a pad token therefore reduces the reported mean loss, even though the configuration documents that pad tokens are prevented from impacting the loss calculation.

This affects both the default and label-smoothed paths:

- Without class weights, the masked loss is still averaged over the full batch with `.mean()`.
- With class weights, the denominator is `weights.sum()` before the gathered weights for padded targets are masked.

Relevant source on current `main`:

- Documented behavior: https://github.com/tracel-ai/burn/blob/ef521c08967a6a98320c2174eba7fc97804ef91e/crates/burn-nn/src/loss/cross_entropy.rs#L18-L22
- Smoothed normalization: https://github.com/tracel-ai/burn/blob/ef521c08967a6a98320c2174eba7fc97804ef91e/crates/burn-nn/src/loss/cross_entropy.rs#L153-L167
- Default normalization: https://github.com/tracel-ai/burn/blob/ef521c08967a6a98320c2174eba7fc97804ef91e/crates/burn-nn/src/loss/cross_entropy.rs#L185-L195

**To reproduce**

```rust
use burn::nn::loss::CrossEntropyLossConfig;

fn main() {

    let one_logits = Tensor::<2>::from_data(TensorData::from([[0.0f32, 0.0]]), &device);
    let one_targets = Tensor::<1, Int>::from_data(TensorData::from([0i64]), &device);
    let one_loss = CrossEntropyLossConfig::new()
        .with_pad_tokens(Some(vec![1]))
        .init(&device)
        .forward(one_logits, one_targets)
        .into_scalar::<f32>();

    let padded_logits = Tensor::<2>::from_data(
        TensorData::from([[0.0f32, 0.0], [0.0, 0.0]]),
        &device,
    );
    let padded_targets = Tensor::<1, Int>::from_data(TensorData::from([0i64, 1]), &device);
    let padded_loss = CrossEntropyLossConfig::new()
        .with_pad_tokens(Some(vec![1]))
        .init(&device)
        .forward(padded_logits, padded_targets)
        .into_scalar::<f32>();

    println!("one real sample: {one_loss}");
    println!("same sample plus ignored pad: {padded_loss}");
}
```

Observed on current `main`:

```text
one real sample: 0.6931472
same sample plus ignored pad: 0.3465736
```

The second input contains the exact same real sample plus one target configured as padding. Its loss is halved because the padded sample contributes zero to the numerator but remains in the batch-size denominator.

**Expected behavior**

Both inputs should produce `0.6931472`. Appending targets listed in `pad_tokens` should not change the loss computed from the valid samples.

For the unweighted paths, normalization should use the number of non-padded samples. For the weighted paths, it should use the sum of gathered class weights for non-padded targets only. The behavior of an entirely padded batch should also be defined and tested.

**Impact**

Padding variable-length token sequences to a common batch length is routine. The current behavior makes loss magnitude depend on how much padding a batch contains, which can distort training metrics and gradient scale between batches.

**Additional context**

- Reproduced using the CPU backend against upstream `main` commit `ef521c08967a6a98320c2174eba7fc97804ef91e`.
- The issue is in the loss reduction logic and is backend-independent.
- The same denominator pattern is present in all four combinations of default/smoothed and weighted/unweighted loss.
- No matching upstream issue or pull request was found.

Mikyx-1:
### Checklist

- [x] Made sure the book is up to date with changes in this PR (no book changes required).

### Related Issues/PRs

Closes #5568.

### Changes

- Exclude padded targets from both the loss numerator and normalization denominator.
- Share the corrected reduction across default, weighted, label-smoothed, and probability-input paths.
- Define an entirely padded batch as `NaN` and handle empty or multiple pad-token lists consistently.
- Add regression coverage for forward loss and gradients, ensuring padded targets have zero gradient without scaling valid gradients.

### Testing

- `cargo run-checks`
- `cargo test -p burn-nn`
- `cargo check -p burn-nn --no-default-features`
- `cargo clippy -p burn-nn --all-targets`

### DIFF
--- a/crates/burn-nn/src/loss/cross_entropy.rs
+++ b/crates/burn-nn/src/loss/cross_entropy.rs
@@ -19,8 +19,9 @@ pub struct CrossEntropyLossConfig {
     /// Create padded cross entropy.
     ///
     /// Prevents pad tokens from impacting loss calculation.
+    ///
+    /// An entirely padded batch returns NaN, since the mean has no valid samples.
     pub pad_tokens: Option<Vec<usize>>,
-
     /// Create weighted cross-entropy.
     ///
     /// The loss of a specific sample will simply be given by: weight * log(p(x)) * 1,
@@ -150,7 +151,7 @@ impl CrossEntropyLoss {
         let tensor = tensor
             * Self::compute_smoothed_targets([batch_size, nr_classes], targets.clone(), alpha);
 
-        match &self.weights {
+        let (tensor, weights) = match &self.weights {
             Some(weights) => {
                 let tensor = tensor
                     * weights
@@ -158,14 +159,13 @@ impl CrossEntropyLoss {
                         .reshape([1, nr_classes])
                         .repeat_dim(0, batch_size);
                 let weights = weights.clone().gather(0, targets);
-                let tensor = Self::apply_mask_2d(tensor, mask);
-                tensor.sum().neg() / weights.sum()
+                (tensor, Some(weights))
             }
-            None => {
-                let tensor = Self::apply_mask_2d(tensor, mask);
-                tensor.sum_dim(1).mean().neg()
-            }
-        }
+            None => (tensor, None),
+        };
+
+        let tensor = tensor.sum_dim(1).squeeze_dim::<1>(1);
+        Self::reduce_mean(tensor, weights, mask)
     }
 
     fn forward_default(&self, logits: Tensor<2>, targets: Tensor<1, Int>) -> Tensor<1> {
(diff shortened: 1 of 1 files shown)


laggui (approved):
Looks good! Thanks for addressing the issue.

PR #5574, opened 2026-09-04
fix(rl): preserve deterministic mode in async batches

### LINKED ISSUE #5567: AsyncPolicy does not preserve per-request deterministic mode in mixed batches
**Describe the bug**

`Policy::action` accepts a `deterministic` flag on each invocation, and `AsyncPolicy` stores that flag on each queued `ActionItem`. However, `PolicyInferenceServer::flush_actions` combines all queued flags with `.all(...)` and calls the inner policy once with that single value.

As a result, when a batch contains both deterministic and stochastic requests, the whole batch is evaluated stochastically. A caller that passed `deterministic = true` therefore receives a stochastic action because of an unrelated concurrent request.

The existing `test_async_policy_deterministic_behaviour` test confirms and currently codifies this behavior: a mixed batch containing one `true` and one `false` request expects both returned actions to be stochastic.

Relevant code:

- https://github.com/tracel-ai/burn/blob/main/crates/burn-rl/src/policy/async_policy.rs#L74-L87
- https://github.com/tracel-ai/burn/blob/main/crates/burn-rl/src/policy/async_policy.rs#L414-L443
- https://github.com/tracel-ai/burn/blob/main/crates/burn-rl/src/policy/base.rs#L53-L61

**To reproduce**

1. Create an `AsyncPolicy` with an autobatch size of 2.
2. Concurrently call `action` once with `deterministic = true` and once with `deterministic = false`.
3. Use an inner policy whose deterministic and stochastic paths return distinguishable actions.
4. Observe that both requests use the stochastic path.

This is also reproduced by the existing unit test:

```console
cargo test -p burn-rl test_async_policy_deterministic_behaviour
```

**Expected behavior**

Each `Policy::action` invocation should preserve its own `deterministic` value regardless of which other requests happen to share its autobatch.

One possible implementation is to partition queued action requests into deterministic and stochastic sub-batches, run the inner policy once for each non-empty group, and route each result back to its original caller. The regression test should then expect the deterministic request to take the deterministic path and the stochastic request to take the stochastic path.

If collapsing mixed requests to stochastic mode is an intentional restriction, it should instead be made explicit in the API documentation because it changes the semantics of an individual `Policy::action` call based on concurrent callers.

**Impact**

When training/exploration and deterministic evaluation share an `AsyncPolicy`, evaluation can silently become stochastic. Results can then depend on request timing and batch composition.

**Additional context**

- Confirmed on upstream `main` at commit `1f96ddaacc21a1230b67f4419c3042a2602725f9`.
- The focused existing test passes and demonstrates the behavior.
- The `.all(...)` behavior has been present since the initial `burn-rl` merge in #4447.
- No matching issue or pull request was found by searching the upstream repository.

Ultronen:
### Checklist

- [x] Made sure the book is up to date with changes in this PR. No public API or documentation behavior changes.

### Related Issues/PRs

Fixes #5567.

### Changes

AsyncPolicy previously collapsed the per-request deterministic flags in a mixed autobatch with all(...). One stochastic request therefore changed the behavior of unrelated deterministic callers.

This change:

- stably partitions queued action requests into at most two deterministic and stochastic sub-batches;
- processes the mode of the first queued request first;
- maps both actions and contexts back through each original item index before sending results;
- updates the mixed-mode regression and adds coverage for homogeneous batches plus interleaved action/context mapping.

A mixed queue now requires two inner policy calls, so calls from interleaved modes are grouped while each group retains request order. This is the batching tradeoff required to preserve each caller mode without falling back to one policy call per item.

### Testing

- Regression proof against the previous implementation: the deterministic request returned the stochastic [0] result instead of the expected [1].
- cargo test -p burn-rl: 9 tests passed; doc tests passed.
- cargo fmt --all -- --check.
- cargo clippy -p burn-rl --all-targets -- -D warnings.
- cargo run-checks: full official gate passed, including Flex backend tests with 1,371 passed and 16 ignored.

### AI-assisted development

Codex assisted with investigation and implementation. I reviewed the complete diff, verified the request/result mapping and batching tradeoff, and ran the project full required checks.

### DIFF
--- a/crates/burn-rl/src/policy/async_policy.rs
+++ b/crates/burn-rl/src/policy/async_policy.rs
@@ -75,27 +75,38 @@ where
         if self.len_actions() == 0 {
             return;
         }
-        let input: Vec<_> = self
-            .batch_action
-            .iter()
-            .map(|m| m.inference_state.clone())
-            .collect();
-        // Only deterministic if all actions are requested as deterministic.
-        let deterministic = self.batch_action.iter().all(|item| item.deterministic);
-        let (actions, context) = self
-            .inner_policy
-            .action(P::Observation::batch(input), deterministic);
-        let actions: Vec<_> = actions.unbatch();
+        let batch_action = core::mem::take(&mut self.batch_action);
+        let first_deterministic = batch_action[0].deterministic;
 
-        for (i, item) in self.batch_action.iter().enumerate() {
-            item.sender
-                .send(ActionContext {
-                    context: vec![context[i].clone()],
-                    action: actions[i].clone(),
-                })
-                .expect("Autobatcher should be able to send resulting actions.");
+        // Keep each request's deterministic mode while retaining autobatching within each group.
+        // Process the first queued mode first to avoid unnecessarily reordering policy calls.
+        for deterministic in [first_deterministic, !first_deterministic] {
+            let (indices, input): (Vec<_>, Vec<_>) = batch_action
+                .iter()
+                .enumerate()
+                .filter(|(_, item)| item.deterministic == deterministic)
+                .map(|(index, item)| (index, item.inference_state.clone()))
+                .unzip();
+
+            if input.is_empty() {
+                continue;
+            }
+
+            let (actions, context) = self
+                .inner_policy
+                .action(P::Observation::batch(input), deterministic);
+            let actions: Vec<_> = actions.unbatch();
+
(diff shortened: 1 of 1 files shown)


laggui (approved):
LGTM! The current partitioning looks like the right scoped fix.

It may be worth replacing the scalar `deterministic: bool` with an explicit action mode that can represent per-observation choices. That would make call sites clearer and allow clear support for heterogeneous batches.

PR #5578, opened 2026-09-05
fix(train): exclude padded samples from accuracy aggregation

### LINKED ISSUE #5577: Accuracy metrics weight padded samples during epoch aggregation
**Describe the bug**

AccuracyMetric and TopKAccuracyMetric correctly exclude pad-token targets from each batch accuracy, but both pass the full batch size to NumericMetricState. That state uses the count as the weight when computing the epoch mean, so padding still affects epoch aggregation.

For example:

- Batch 1: one valid correct sample and nine padded samples gives 100%.
- Batch 2: ten valid incorrect samples gives 0%.
- The current epoch result is 50%, because both batches are weighted as ten samples.
- The correct result is 1 / 11 = 9.09%, because only valid samples should contribute.

A fully padded batch also computes 0 / 0 and can poison the running epoch value with NaN.

**To Reproduce**

Create AccuracyMetric::new().with_pad_token(...) or TopKAccuracyMetric::new(...).with_pad_token(...), then update it with two equal-sized batches containing different numbers of padded targets. Compare final_value() with total correct non-padding predictions divided by total non-padding targets.

Relevant code:

- crates/burn-train/src/metric/acc.rs
- crates/burn-train/src/metric/top_k_acc.rs
- crates/burn-train/src/metric/state.rs

**Expected behavior**

Epoch aggregation should weight each batch by batch_size - num_pad. A batch containing no valid targets should not alter the running epoch aggregate.

**Additional context**

Confirmed on upstream main at d16f7ba2e.

Mikyx-1:
### Checklist

- [x] Made sure the book is up to date with changes in this PR; no book changes are needed.

### Related Issues/PRs

Fixes #5577

### Changes

- Weight AccuracyMetric and TopKAccuracyMetric epoch aggregation by non-padding samples.
- Prevent zero-valid-sample batches from changing the running numeric aggregate.
- Add multi-batch and fully padded regression coverage.

### Testing

- cargo run-checks
- cargo test -p burn-train metric:: --lib

### DIFF
--- a/crates/burn-train/src/metric/acc.rs
+++ b/crates/burn-train/src/metric/acc.rs
@@ -52,20 +52,20 @@ impl Metric for AccuracyMetric {
 
         let outputs = outputs.argmax(1).reshape([batch_size]);
 
-        let accuracy = match self.pad_token {
+        let (num_matches, num_pad) = match self.pad_token {
             Some(pad_token) => {
                 let mask = targets.clone().equal_scalar(pad_token as i64);
                 let matches = outputs.equal(targets).float().mask_fill(mask.clone(), 0);
-                let num_pad = mask.float().sum();
+                let num_pad = mask.int().sum().into_scalar::<i64>() as usize;
 
-                let acc = matches.sum() / (num_pad.neg() + batch_size as f32);
-
-                acc.into_scalar::<f64>()
+                (matches.sum().into_scalar::<f64>(), num_pad)
             }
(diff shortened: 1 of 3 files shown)


laggui (approved):
Thanks for the fix, it looks correct.

Weighting by `valid_count` fixes the multi-batch aggregation, while the zero-count guard prevents an all-padding batch's undefined (NaN) accuracy from contaminating subsequent epoch state.

Issue #5522, opened 2026-08-28
Implement remaining IndexingUpdateOp variants for scatter/select_assign across backends


antimora:
## Feature description

Follow-up to #4477. `IndexingUpdateOp` now has all five variants (`Assign`, `Add`, `Mul`, `Min`, `Max`) and both `Tensor::scatter` and `Tensor::select_assign` take it, but most backends still only implement a subset and fall through to `unimplemented!()` at runtime.

`scatter_nd` / `select`-family aside, here is the current state on `main` for the element-wise ops:

### `scatter` (element-wise)

| Backend | Assign | Add | Mul | Min | Max |
|---|---|---|---|---|---|
| ndarray | yes | yes | yes | no | no |
| flex | no | yes | yes | no | no |
| cubecl (wgpu/cuda/metal/rocm) | no | yes | yes | no | no |
| tch | no | yes | yes | no | no |
| autodiff | yes | yes | yes | no | no |
| bool (bridge) | no | yes (`or`) | no | no | no |

### `select_assign`

| Backend | Assign | Add | Mul | Min | Max |
|---|---|---|---|---|---|
| ndarray | yes | yes | yes | no | no |
| flex | no | yes | yes | no | no |
| cubecl | no | yes | yes | no | no |
| tch | no | yes | yes | no | no |
| bool (bridge) | no | yes (`or`) | no | no | no |

### `scatter_nd` (for contrast, already complete)

All five reductions are implemented in ndarray, flex and cubecl. This is the shape the element-wise ops should reach.

So the concrete gaps are:

- **`Assign`** missing in `burn-flex`, `burn-cubecl` and `burn-tch` for both `scatter` and `select_assign` (ndarray has it).
- **`Min` / `Max`** missing in *every* backend for both `scatter` and `select_assign`.

The default trait bodies in `burn-backend/src/backend/ops/tensor.rs` (`float_scatter`, `float_select_assign`) forward only `Add`, so a backend that does not override them silently inherits Add-only support and panics at runtime rather than failing to compile.

## Feature motivation

This is what blocks the ONNX side from finishing tracel-ai/burn-onnx#486.

ONNX `ScatterElements` has five reduction modes that map one-to-one onto `IndexingUpdateOp`. `ScatterND` can already lower all five to `scatter_nd` because that op is complete. `ScatterElements` cannot:

| ONNX reduction | Lowering today in burn-onnx |
|---|---|
| `add` | `scatter(..., Add)` |
[...]

PR #5582, opened 2026-09-06
feat(tensor): implement Min and Max scatter/select_assign across backends

ax1s-x1zz:
Resolves the remaining gap in #5522: element-wise `scatter` / `select_assign` only supported `Assign`, `Add` and `Mul`, and every backend hit `unimplemented!` for `Min` / `Max` — even though `scatter_nd` already implements all five variants.

### Changes

- **ndarray**: add `scatter_min` / `scatter_max` and `select_assign_min` / `select_assign_max` primitives (per-element compare, same shape/index handling as the existing `_mul` variants).
- **flex**: `scatter_min` / `scatter_max` and `select_min` / `select_max` helpers sharing the existing `scatter_update` / `select_update` walkers.
- **cubecl**: new `scatter_min` / `scatter_max` and `select_assign_min` / `select_assign_max` entries reusing the existing `BinaryMinOp` / `BinaryMaxOp` kernels (already wired for `scatter_nd`).
- **tch**: `scatter_min` / `scatter_max` via `scatter_reduce(..., "amin"/"amax")` and `select_assign_min` / `select_assign_max` via `index_reduce_(..., "amin"/"amax")`, mirroring the existing `"prod"` paths.
- **autodiff**: backward passes for `Min` / `Max` scatter and select_assign, mirroring the `scatter_nd` Min/Max gradient (winner masks from comparisons, ties contribute to both inputs per the cummin/cummax convention; unique indices required, as with `scatter_nd`).
- Dispatch matches in all backends now enumerate every `IndexingUpdateOp` variant, so an unsupported combination fails at compile time instead of panicking at runtime.

### Tests

- Forward: float and int `scatter` / `select_assign` with `Min` / `Max`, 1D and 2D, in `burn-backend-tests/tests/tensor/{float,int}/ops/`.
- Autodiff: data-win and values-win gradient cases for `Min` / `Max` `scatter` and `select_assign`.

### Verification

- `cargo test -p burn-backend-tests --features ndarray`: 1839 tensor + 572 autodiff tests pass.
- Confirmed the autodiff tests fail with the previous `unimplemented!` when the Min/Max backward is removed.
- `cargo clippy` clean for `burn-ndarray`, `burn-flex`, `burn-autodiff`, `burn-cubecl`.

Note: `burn-tch` compiles against libtorch, which is unavailable in my local environment; the changes are a direct mirror of the existing `"prod"` scatter/`index_reduce` paths that already use `"amin"/"amax"` in `scatter_nd`.

### DIFF
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1284,7 +1284,6 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         _checkpointer: &mut Checkpointer,
                     ) {
                         let (dim, data, values, indices, is_max) = ops.state;
-                        let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));
 
                         let device = data.device();
                         let data_shape = data.shape();
@@ -1292,7 +1291,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         let settings = get_device_settings::<B>(&device);
                         let bool_dtype = settings.bool_dtype;
 
-                        let data_at_idx = B::float_gather(dim, data, indices_4lhs.clone().unwrap());
+                        let data_at_idx = B::float_gather(dim, data.clone(), indices.clone());
 
                         let (data_won_bool, values_won_bool) = if is_max {
                             (
@@ -1318,22 +1317,24 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         let values_won_float =
                             B::bool_into_float(values_won_bool, data_dtype.into());
 
-                        let ones = B::float_ones(data_shape, &device, data_dtype.into());
-                        let data_mask = B::float_scatter(
-                            dim,
-                            ones,
-                            indices_4lhs.unwrap(),
-                            data_won_float,
-                            IndexingUpdateOp::Assign,
-                        );
-
                         binary::<B, _, _>(
                             ops.parents,
                             ops.node,
                             grads,
-                            |grad| B::float_mul(grad, data_mask),
                             |grad| {
-                                let g_idx = B::float_gather(dim, grad, indices_4rhs.unwrap());
+                                let ones =
+                                    B::float_ones(data_shape.clone(), &device, data_dtype.into());
+                                let data_mask = B::float_scatter(
+                                    dim,
+                                    ones,
+                                    indices.clone(),
+                                    data_won_float,
+                                    IndexingUpdateOp::Assign,
+                                );
+                                B::float_mul(grad, data_mask)
+                            },
+                            |grad| {
+                                let g_idx = B::float_gather(dim, grad, indices.clone());
                                 B::float_mul(g_idx, values_won_float)
                             },
                         );
@@ -2118,7 +2119,6 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         let tensor: FloatTensor<B> =
                             checkpointer.retrieve_node_output(tensor_state);
                         let values: FloatTensor<B> = checkpointer.retrieve_node_output(value_state);
-                        let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));
 
                         let device = tensor.device();
                         let tensor_shape = tensor.shape();
@@ -2126,8 +2126,7 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         let settings = get_device_settings::<B>(&device);
                         let bool_dtype = settings.bool_dtype;
 
-                        let data_at_idx =
-                            B::float_select(tensor, dim, indices_4lhs.clone().unwrap());
+                        let data_at_idx = B::float_select(tensor, dim, indices.clone());
 
                         let (data_won_bool, values_won_bool) = if is_max {
                             (
@@ -2153,22 +2152,27 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                         let values_won_float =
                             B::bool_into_float(values_won_bool, tensor_dtype.into());
 
-                        let ones = B::float_ones(tensor_shape, &device, tensor_dtype.into());
-                        let data_mask = B::float_select_assign(
-                            ones,
-                            dim,
-                            indices_4lhs.unwrap(),
-                            data_won_float,
-                            IndexingUpdateOp::Assign,
-                        );
-
                         binary::<B, _, _>(
                             ops.parents,
                             ops.node,
                             grads,
-                            |grad| B::float_mul(grad, data_mask),
                             |grad| {
-                                let g_idx = B::float_select(grad, dim, indices_4rhs.unwrap());
+                                let ones = B::float_ones(
+                                    tensor_shape.clone(),
+                                    &device,
+                                    tensor_dtype.into(),
+                                );
+                                let data_mask = B::float_select_assign(
+                                    ones,
+                                    dim,
+                                    indices.clone(),
+                                    data_won_float,
+                                    IndexingUpdateOp::Assign,
+                                );
+                                B::float_mul(grad, data_mask)
+                            },
+                            |grad| {
+                                let g_idx = B::float_select(grad, dim, indices.clone());
                                 B::float_mul(g_idx, values_won_float)
                             },
                         );
--- a/crates/burn-autodiff/src/ops/tensor.rs
+++ b/crates/burn-autodiff/src/ops/tensor.rs
@@ -1259,7 +1259,113 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C> 
                     )),
                 }
             }
-            other => unimplemented!("float_scatter with {other:?} update is not implemented"),
+            IndexingUpdateOp::Min | IndexingUpdateOp::Max => {
+                // Unique indices are required for this backward formula; duplicate indices have
+                // undefined behavior (same caveat as float_scatter_nd Min/Max).
+                // Forward (Max): out[.., idx, ..] = max(data[.., idx, ..], values[.., i, ..]).
+                // Backward, with ties contributing to both sides (matches the cummin/cummax
+                // convention):
+                //   data_at_idx  = gather(data, dim, idx)
+                //   data_won     = data_at_idx >= values   (Max) / <= values (Min)
+                //   values_won   = values >= data_at_idx   (Max) / <= data_at_idx (Min)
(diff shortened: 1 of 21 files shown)


crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1259,7 +1259,113 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                        let (dim, data, values, indices, is_max) = ops.state;
+                        let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));
+
+                        let device = data.device();
+                        let data_shape = data.shape();
+                        let data_dtype = data.dtype();
+                        let settings = get_device_settings::<B>(&device);
+                        let bool_dtype = settings.bool_dtype;
+
+                        let data_at_idx = B::float_gather(dim, data, indices_4lhs.clone().unwrap());
```
laggui: `indices_4lhs` is `None` when only `values` requires gradients, so this unwrap panics. We should keep indices available independently of parent tracking, and add a values-only gradient test.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1960,8 +2066,152 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                        let [indices_4lhs, indices_4rhs] = duplicate(&ops.parents, Some(indices));
+
+                        let device = tensor.device();
+                        let tensor_shape = tensor.shape();
+                        let tensor_dtype = tensor.dtype();
+                        let settings = get_device_settings::<B>(&device);
+                        let bool_dtype = settings.bool_dtype;
+
+                        let data_at_idx =
+                            B::float_select(tensor, dim, indices_4lhs.clone().unwrap());
```
laggui: Same here

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1259,7 +1259,113 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
...
+                                ),
+                                B::float_lower_equal(values, data_at_idx, bool_dtype),
+                            )
+                        };
+
+                        let data_won_float = B::bool_into_float(data_won_bool, data_dtype.into());
+                        let values_won_float =
+                            B::bool_into_float(values_won_bool, data_dtype.into());
+
+                        let ones = B::float_ones(data_shape, &device, data_dtype.into());
```
laggui: Could the destination mask be built inside the destination-gradient closure? Right now it allocates a full-sized tensor even when only values requires gradients.

crates/burn-backend-tests/tests/tensor/float/ops/gather_scatter.rs
```diff
@@ -132,6 +132,34 @@ fn should_scatter_mul_1d() {
         .assert_eq(&TensorData::from([10.0, 140.0, 30.0, 2000.0]), false);
 }
 
+#[test]
+fn should_scatter_min_1d() {
```
laggui, line 136: We should add repeated-index cases for Min/Max scatter and select_assign. The tests currently only cover unique destinations.

crates/burn-ndarray/src/ops/base.rs
```diff
@@ -293,6 +293,120 @@ where
...
+        }
+        output
+    }
+
+    pub fn scatter_max<I: NdArrayElement>(
+        dim: usize,
+        mut tensor: SharedArray<E>,
+        mut indices: SharedArray<I>,
+        mut value: SharedArray<E>,
+    ) -> SharedArray<E>
```
laggui, line 329: Could Min and Max share a private update helper? These functions duplicate the same shape/index traversal and only in the comparison.

laggui (changes requested):
Thanks for taking this on!

Overall great start, just have a couple of comments

crates/burn-autodiff/src/ops/tensor.rs
ax1s-x1zz: Fixed. `indices` now lives in the backward state and is passed to both gradient closures directly, so it is available regardless of which parent requires gradients. Added `test_scatter_max_grad_values_only` / `test_scatter_min_grad_values_only` to cover the values-only case.

crates/burn-autodiff/src/ops/tensor.rs
ax1s-x1zz: Fixed the same way in `select_assign`: indices are kept in the state and used directly in the closures, and `test_select_assign_{max,min}_grad_values_only` cover the values-only path.

crates/burn-autodiff/src/ops/tensor.rs
ax1s-x1zz: Done — the `ones`/`data_mask` construction now happens inside the data-gradient closure, so it is only built when the data parent actually requires gradients.

crates/burn-backend-tests/tests/tensor/float/ops/gather_scatter.rs
ax1s-x1zz: Added repeated-index forward tests for Min/Max scatter and select_assign (float and int, 1D): `should_scatter_{min,max}_1d_{,_}repeated_indices`, `should_select_assign_{min,max}_1d_repeated_indices`, and the int counterparts.

crates/burn-ndarray/src/ops/base.rs
ax1s-x1zz: Refactored: `scatter_min`/`scatter_max` now share a private `scatter_extreme` helper and `select_assign_min`/`select_assign_max` share `select_assign_extreme`, with only the comparison closure differing.

crates/burn-autodiff/src/ops/tensor.rs
```diff
@@ -1259,7 +1259,114 @@ impl<B: Backend, C: CheckpointStrategy> FloatTensorOps<Self> for Autodiff<B, C>
                     )),
                 }
             }
-            other => unimplemented!("float_scatter with {other:?} update is not implemented"),
+            IndexingUpdateOp::Min | IndexingUpdateOp::Max => {
+                // Unique indices are required for this backward formula; duplicate indices have
+                // undefined behavior (same caveat as float_scatter_nd Min/Max).
```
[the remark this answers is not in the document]
ax1s-x1zz: This mirrors the existing `float_scatter_nd` Min/Max contract: duplicate indices are documented as undefined for that backward formula too, while the forward reduction itself stays well-defined (Min/Max is commutative, so the result is the reduction over all contributions regardless of order). Enforcing uniqueness would require materializing and comparing the indices on the backward path, which the existing `scatter_nd` implementation deliberately avoids. I kept the documented contract for parity; happy to add a shared check separately if maintainers prefer.

crates/burn-flex/src/ops/gather_scatter.rs
```diff
@@ -411,6 +411,48 @@ pub fn scatter_mul<E: Element + Pod + Default + Copy + core::ops::Mul<Output = E
...
+        tensor,
+        dim,
+        indices,
+        value,
+        "scatter_min",
+        |target, value| {
+            if value < *target {
+                *target = value;
+            }
+        },
```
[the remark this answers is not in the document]
ax1s-x1zz: Documented. The flex (and ndarray) Min/Max helpers now note that comparisons follow IEEE semantics — an incoming NaN never replaces the current value — matching the existing `scatter_nd` Min/Max reductions in the same modules.

crates/burn-ndarray/src/ops/base.rs
```diff
@@ -293,6 +293,88 @@ where
...
+            shape_tensor[ndims - 1],
+            shape_indices[ndims - 1],
+            shape_value[ndims - 1],
+        );
+        let batch_size = Self::gather_batch_size(&shape_tensor, shape_indices);
+
+        if shape_value != shape_indices {
+            panic!(
+                "Invalid dimension: the shape of the index tensor should be the same as the value \
+                 tensor: Index {:?} value {:?}",
```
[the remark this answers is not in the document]
ax1s-x1zz: Updated the panic in the shared ndarray helper to describe the shape mismatch and name the operation, instead of the misleading `Invalid dimension` prefix.

crates/burn-ndarray/src/ops/base.rs
```diff
@@ -1121,6 +1203,56 @@ where
...
+        value: SharedArray<E>,
+        replace: F,
+    ) -> SharedArray<E>
+    where
+        F: Fn(&mut E, &E) -> bool,
+    {
+        let mut output_array = tensor.into_owned();
+
+        for (index_value, index) in indices.into_iter().enumerate() {
+            let mut view = output_array.index_axis_mut(Axis(dim), index.elem::<i64>() as usize);
```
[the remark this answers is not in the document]
ax1s-x1zz: Added explicit preconditions to `select_assign_extreme` (dim bounds, `indices` must be 1D, value rank and axis length) with clear messages, mirroring the shape check in `scatter_extreme`.

ax1s-x1zz:
Addressed the Copilot review:

- **NaN semantics**: documented that the flex and ndarray Min/Max `scatter`/`select_assign` helpers use IEEE comparisons (an incoming NaN never replaces the current value), consistent with the existing `scatter_nd` Min/Max reductions.
- **Panic message**: the shared ndarray scatter Min/Max helper now reports the indices/value shape mismatch explicitly instead of starting with `Invalid dimension`.
- **Preconditions**: `select_assign_extreme` now validates dim bounds, 1D indices, value rank and axis length up front with clear messages.
- **Duplicate indices**: left as documented behavior (undefined for the Min/Max backward, same caveat as `float_scatter_nd`); the forward reduction remains well-defined for duplicates. Added a note in the thread.

Changes in `8fc5ed2`; `cargo test -p burn-backend-tests --features ndarray` passes (1846 tensor + 580 autodiff) and clippy/fmt are clean for the touched crates.

crates/burn-autodiff/src/ops/tensor.rs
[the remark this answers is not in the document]
laggui: That's fine. I'd just make a small documentation change in the public `Tensor::scatter` / `select_assign` methods. It should explicitly mention min/max in the duplicate-index notes.

laggui (approved):
LGTM, thanks for addressing my comments.

Just a minor note, but I'll approve in advance.

crates/burn-autodiff/src/ops/tensor.rs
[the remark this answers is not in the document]
ax1s-x1zz: Done in c85cd76. `Tensor::scatter` and `Tensor::select_assign` now mention Min/Max in their duplicate-index notes: the forward reduction is defined (the minimum/maximum over all contributions), while the backward gradients are undefined. Thanks for the review!

ax1s-x1zz:
Applied the documentation note from @laggui in c85cd76: `Tensor::scatter` and `Tensor::select_assign` now explicitly cover `Min`/`Max` in their duplicate-index notes (forward reduction defined, backward gradients undefined). No code changes.

Issue #5206, opened 2026-07-21
Feature/design: general einsum (tensor contraction) in burn-tensor


Bellman281:
### Feature description

Add a general `einsum` (Einstein-summation / tensor contraction) to `burn-tensor`.
Burn currently has no `einsum` — there's no such symbol anywhere in the crate — so
contractions that don't map cleanly onto `matmul` have to be hand-written as
`permute`/`reshape`/`matmul`/`sum` chains at every call site.

### Feature motivation

`einsum` is the main missing primitive for porting transformer/DETR-style vision
models to Burn. A native Burn inference port of SAM3 needs it in three places:

- MaskFormer mask prediction — `bqc,bchw->bqhw`
- ViT-Det decomposed relative-position attention — `bhwc,hkc->bhwk`, `bhwc,wkc->bhwk`
- RoPE frequency outer products

It's high-leverage well beyond SAM — most attention/DETR ports hit one of these
contractions. Notably, everything *else* those models need for inference already exists
in Burn (gelu, grid_sample, nonzero, the attention/conv/norm primitives), so `einsum`
is effectively the one gap standing between Burn and a large class of model ports.

### (Optional) Suggest a Solution

**Implementation:** parse the subscript equation and lower it to existing Burn ops
(`permute` → `reshape` → broadcast → `matmul` → `sum`). No backend changes required.

**Key design tradeoff — output rank vs. Burn's type system.** `einsum`'s output rank is
determined by the equation, which is a *runtime* string, but Burn tensors carry rank as
a compile-time `const D`. So the signature can't infer the output rank. Options:

1. Caller supplies the output rank as a const generic, validated against the parsed
   equation at runtime: `einsum::<D_OUT, ...>(equation: &str, operands: ...) -> Tensor<D_OUT>`.
2. A declarative macro that expands to typed ops.
3. Fixed-arity helpers for the common 2-operand case only.

I'd lean toward option 1, but this is really a question for the maintainers since it
touches Burn's conventions — hence this issue rather than a surprise PR.

**Scope / caveats for the future.** I'd suggest a first PR limited to **two-operand**
contractions (covers all the SAM cases and the majority of transformer usage), no
ellipsis or diagonals initially, with the supported subset clearly documented and room
to extend to n-operand/ellipsis later.

**Questions:**
1. Preferred API shape — const-generic output rank (option 1), a macro, or something
   that fits Burn's conventions better?
2. Is a documented 2-operand subset an acceptable first step, or do you want fuller
[...]

Issue #4706, opened 2026-03-31
Feature request: native einsum tensor operation


antimora:
## Summary

Burn currently lacks a native `einsum` tensor operation. The `burn-onnx` crate (ONNX importer) works around this by decomposing einsum equations into sequences of `permute`, `reshape`, and `matmul` calls at code-generation time. While this works for many common patterns, a native einsum API on `Tensor` would be more ergonomic for users and could enable backend-specific optimizations.

## Motivation

- **ONNX compatibility**: The ONNX `Einsum` operator supports arbitrary Einstein summation notation. The current decomposition in `burn-onnx` handles 2-input cases but more complex patterns (trace, >2 inputs, ellipsis broadcasting) require increasingly elaborate lowering logic
- **User ergonomics**: Einsum is a widely used API in NumPy, PyTorch (`torch.einsum`), and TensorFlow. Users coming from these frameworks expect it
- **Backend optimization**: Backends could implement einsum natively (e.g., cuBLAS has optimized contraction routines) rather than relying on the decomposed form

## Proposed API

```rust
// Explicit form
let c = Tensor::einsum("ij,jk->ik", &[a, b]);

// Implicit form (output inferred alphabetically)
let c = Tensor::einsum("ij,jk", &[a, b]);

// Batch matmul
let c = Tensor::einsum("bij,bjk->bik", &[a, b]);

// Trace
let t = Tensor::einsum("ii->", &[a]);

// Multi-input
let d = Tensor::einsum("ij,jk,kl->il", &[a, b, c]);
```

## Context

This came up while working on tracel-ai/burn-onnx#276 (completing full ONNX Einsum support). The decomposition approach works but a native op would simplify the importer and benefit Burn users directly.

nathanielsimard:
That would be cool if we can parse the einsum string using a macro/proc-macro to have compiler feedback. But we should also have one where the einsum is declared at runtime dynamically.

jsonmona:
Would be great if other einops functions like rearrange, reduce, and repeat are also supported.

antimora:
Landed in #5583. Closing.

Issue #2480, opened 2024-11-11
Implement something like `torch.einsum` for `Tensor`s


TheLostLambda:
### Feature description

Something like https://pytorch.org/docs/stable/generated/torch.einsum.html, with something similar done in Rust already here: https://docs.rs/einsum-codegen/latest/einsum_codegen/

### Feature motivation

It's a more compact way to write out a number of tensor operations that come up throughout ML work.

### (Optional) Suggest a Solution

It seems like a good place for writing a macro! It looks like there are some details on the algorithm here: https://docs.rs/einsum-codegen/latest/einsum_codegen/#einsum-algorithm

antimora:
@crutcher, Not sure if this falls under `linalg`, but just in case fyi

BhavyeMathur:
Hey! I recently implemented einsums for my personal Rust ML library. The relevant code can be found [here](https://github.com/BhavyeMathur/redstone-ml/blob/master/src/linalg/einsum.rs#L403). Could I be assigned this issue?

My implementation was modelled after NumPy and works as follows:

1. Parse the `einsum` input and output arguments, keeping a track of unique indices, the count of each index, etc.
2. Check if a view into the tensor can be returned, and do so if possible
3. Check if any dimensions can be collapsed for faster iteration
4. If we have 2 or 3 iteration axes (+some required conditions), dispatch to a specialised einsum loop
5. Conduct an unspecialised einsum loop

In both the specialised and unspecialised loop, we make use of a 'sum of products' function selected according to the memory layout of the tensor. More can be found [here](https://github.com/BhavyeMathur/redstone-ml/blob/master/src/linalg/sum_of_products.rs). These loops are where we can provide GPU/CPU acceleration.

The architecture of Burn seems quite different and I'm unsure if I can just implement something as described above or if there are other things I must keep in mind as well (such as kernel fusion, the autodiff engine, etc). I'm quite new to Burn so any advice on how to go about implementing this would be greatly appreciated!

laggui:
> Hey! I recently implemented einsums for my personal Rust ML library. The relevant code can be found [here](https://github.com/BhavyeMathur/redstone-ml/blob/master/src/linalg/einsum.rs#L403). Could I be assigned this issue?

Of course! I just assigned it to you. Feel free to remove the assignment if ever you change your mind 🙂 

[...]

PR #5583, opened 2026-09-06
feat(tensor): add einsum with runtime and macro APIs

Mikyx-1:
Burn currently requires users to write reshape/permute/matmul/reduction chains for general tensor contractions. This adds native einsum through a literal macro and a runtime equation API, including mixed-rank operands and automatic differentiation.

```rust
let masks: Tensor<4> = einsum!("bqc,bchw->bqhw", queries, features);
let result = Tensor::<1>::einsum(&equation, [matrix.into(), vector.into()]);
```

### Checklist

### Related Issues/PRs

Related to #2480, #4706, and #5206. Implements both literal and runtime equations as discussed in #4706, with a validated equation and shape-independent execution plan shared between the macro and runtime API.

### Changes

- Add a dependency-free `no_std + alloc` parsing and planning crate, `burn-einsum`, shared by `burn-derive` and `burn-tensor`. The plan specifies diagonal permutations, alignment, reductions, matrix groups, and output ordering; ellipsis axes are symbolic blocks.
- Add `einsum!` with compile-time equation/operand-count validation and checks for statically determined input/output ranks. The macro consumes the shared plan at compile time and emits diagonal, permute, reshape, reduction, multiplication, and matmul operations directly. Only dimension arithmetic, ellipsis widths, broadcasting, and empty-dimension branches remain runtime decisions. Generated code includes an explanation of its lowering.
- Add `Tensor::einsum` and `EinsumOperand` for runtime equations and operands of different ranks. Runtime equations build the same plan and use one private executor call; there is no staged `Execution`/`contract_next()` interface.
- Support explicit/implicit output, case-sensitive labels, repeated-label diagonals, ellipsis broadcasting and reduction, and any number of operands.
- Port the alignment and left-to-right pairwise contraction algorithm from [my Python reference](https://github.com/Mikyx-1/pytorch-einsum-reference/tree/0451b648390965f341f5f9f8fe69c2f64a5259a5): retain axes until their final participating operand, perform early reductions, and lower contractions to existing Burn operations. No new backend operation is introduced.
- Preserve early contractions when later operands have singleton dimensions, avoiding large broadcast intermediates. Use multiplication/reduction for empty contractions because some existing matmul implementations cannot handle empty dimensions. This keeps both inputs connected to autodiff. Use explicit integer widths for diagonal indices to avoid overflow with narrow device defaults.
- Add backend, gradient/checkpointing, parser/planner, generated-operation, macro hygiene, and compile-fail tests, plus book documentation and a runnable demo.

Float and Int operands must share their kind, dtype, and device. Scalar outputs use Burn's `[1]` convention; empty input subscripts accept shape `[1]`. Quantized operands, Python's sublist call format, contraction-path optimization, and ONNX importer migration are outside this change. Backend execution was validated on Flex and NdArray; GPU/fusion execution has not been tested locally.

### Testing

All checks below passed:

- `CARGO_INCREMENTAL=0 cargo run-checks`: formatting, typos, dependency audit, workspace linting, host no-std check, and the full Flex backend suite.
- Einsum backend coverage: 48 forward tests and 20 autodiff/checkpointing tests per backend, through the full Flex suite and a focused NdArray run.
- `cargo test -p burn-einsum -p burn-derive`: parser/planner and proc-macro tests, including exact compiled permutations and symbolic ellipsis layouts.
- Tensor macro integration and isolated diagonal-index regression tests: 22 tests, including crossed broadcasting, singleton contractions, reordered outputs, symbolic diagonals, empty tensors, and gradient connections. A separate metadata regression verifies that trailing singleton weights allow contraction before a large broadcast intermediate.
- `cargo test -p burn-tensor --features flex,autodiff --doc einsum`: 2 executable examples and 5 compile-fail tests.

Run the demo:

```sh
cargo run -p burn-tensor --example einsum --features flex,autodiff
```

It demonstrates matrix multiplication, a runtime equation, trace, MaskFormer-style mask prediction, and a dot-product gradient. Expected output is documented in `crates/burn-tensor/examples/README.md`.


laggui (changes requested):
Thanks for taking this on!

This follows the proposed direction of lowering einsum to existing burn operations, but the macro currently emits parsed labels and unrolled `contract_next()` calls. Layout construction and lowering still happen at runtime.

Could we move the shape-independent planning into the shared representation, as discussed in #5206? For example, `ij,jk->ik` already determines the contracted axis and output ordering, even though their sizes are only known at runtime. The macro could use that plan to generate lowering code, while retaining runtime checks and branches where broadcasting or empty dimensions affect the selected operations.

If runtime lowering is intentional, a single executor call seems simpler than exposing the staged execution interface.

Mikyx-1:
Thanks for the guidance, @laggui! Pushed c45dfe15f. I moved the shape-independent planning into `burn-einsum` and updated the macro to generate the diagonal, permutation, reshape, reduction, and matmul operations directly from that plan. This includes symbolic ellipsis axes and fixed output ordering; broadcasting and empty dimensions retain the necessary runtime branches. I also preserved early contraction when later operands have singleton dimensions.

The runtime API uses the same plan through one private executor call, and the staged `Execution`/`contract_next()` interface is removed. Added regressions cover generated operations, broadcasting, diagonals, empty tensors, and gradients.

The repository checks and focused Flex/NdArray tests pass. Could you please take another look when you have a chance? Thank you!

crates/burn-tensor/src/tensor/api/einsum/execution.rs
```diff
@@ -0,0 +1,660 @@
...
+        value = value.diagonal(
+            axis(diagonal.first, local_width),
+            axis(diagonal.second, local_width),
+            &axes(&diagonal.permutation, local_width),
+            &axes(&diagonal.restore, local_width),
+        );
+    }
+    let value = value.permute(&axes(&input.permutation, local_width));
+    let shape = alignment_shape(&value.shape(), &input.axes, plan.ellipsis, width);
+    value.reshape(shape)
```
laggui, line 387: The empty-contraction case fails with fusion enabled, as the macOS CI failure show.

This seems to be a pre-existing issue in `burn_ir::ShapeOpIr::reshape` that was simply exposed here. For `ik,kj->ij` with inputs `[2, 0]` and `[0, 3]`, alignment correctly reshapes the left operand to `[2, 1, 0]`. However, the IR builder calls `input.shape.reshape(shape).unwrap()`, which interprets zeros a "copy the source dimension" and indexes past the source rank.

Could we fix the builder to preserve the supplied concrete shape and validate equal element counts, instead of resolving reshape arguments again? This can stay scoped to `burn-ir`.

```rust
pub fn reshape(input: TensorIr, shape: Shape, new_id: impl FnOnce() -> TensorId) -> Self {
    assert_eq!(
        input.shape.num_elements(),
        shape.num_elements(),
        "Reshape must preserve the number of elements"
    );
    Self::create(input, shape, new_id)
}
```

with an additional regression test for `[2, 0]` -> `[2, 1, 0]`.

This could land separately as a prerequisite fix if you want.

laggui (commented):
Thanks for addressing my comment

crates/burn-tensor/src/tensor/api/einsum/execution.rs
Mikyx-1: @laggui Fixed in ca53dfa2e4117c80b47a698a877346c68b307f39. I reproduced the out-of-bounds panic before applying your suggested fix. The builder now preserves the concrete shape and asserts equal element counts.

The change stays in `burn-ir`, with regression coverage for `[2, 0] -> [2, 1, 0]`, moving an empty dimension, ordinary reshapes, and rejecting unequal element counts (including `[2, 3] -> [0, 3]`, which previously treated zero as a copy instruction).

Validation:
- All 27 `burn-ir` tests pass.
- Full Flex backend suite: 3,881 passed.
- CUDA fusion: the reported empty-contraction case passes; all 21 reshape-filtered tensor tests and both reshape gradient tests pass.
- Workspace Clippy, focused `burn-ir` Clippy, formatting, typos, and host no-std checks pass.

Two validation limitations remain:
- CUDA fusion einsum runs passed 47/48 tensor tests and 18/20 gradient tests. The remaining failures report unsupported strides when reading empty tensors. The empty-ellipsis failure also reproduces with the original builder; a temporary diagnostic reading `Tensor::<2>::zeros([2, 0], &device).into_data()` reproduces the same error without any reshape call.
- `cargo run-checks` stops at the existing locked `rustls 0.23.43` advisory, RUSTSEC-2026-0285. I ran the subsequent lint, no-std, and Flex stages separately.

Pushed for another look; macOS/Metal verification is left to CI. Thanks for identifying the IR issue.

laggui (approved):
LGTM! I only made some small documentation/API improvements to bring this over the line.

PR #5585, opened 2026-09-06
fix(tensor): avoid cosine similarity denominator underflow

### LINKED ISSUE #5584: cosine_similarity returns NaN for zero vectors with default epsilon
### Description

`burn_tensor::linalg::cosine_similarity` can return `NaN` for two zero vectors with the default epsilon. The epsilon is intended to prevent division by zero, but the product of the two clamped norms underflows to zero.

Observed in the implementation on upstream main at `d16f7ba2e`:
`crates/burn-tensor/src/tensor/linalg/cosine_similarity.rs`.

### Minimal example

Using `f32` tensors on the Flex backend:

```rust
use burn_tensor::{Tensor, linalg};

let x = Tensor::<1>::from_data([0.0_f32, 0.0], &device);
let y = Tensor::<1>::from_data([0.0_f32, 0.0], &device);
let result = linalg::cosine_similarity(x, y, 0, None);
println!("{:?}", result.into_data());
```

Run with the `flex` feature enabled and `BURN_DEVICE=flex`.

For `x = y = [0, 0]`, the default epsilon is `f32::MIN_POSITIVE`, approximately `1.17549435e-38`:

1. Both L2 norms and the dot product are zero.
2. Each norm is clamped to epsilon.
3. Multiplying the clamped norms computes `epsilon * epsilon`, approximately `1.38e-76`, which rounds to zero in `f32`.
4. The result is `0 / 0 = NaN`.

The same denominator-underflow problem can occur with an explicitly supplied positive epsilon that is sufficiently small.

### Expected behavior

With a positive representable epsilon, zero vectors should produce a finite zero under the regularized cosine-similarity formula. Although the mathematical cosine of a zero vector is undefined, this API explicitly supplies epsilon to avoid division by zero.

### Proposed fix

Normalize each input separately, then compute the dot product:

```text
Current:  dot(x, y) / (max(norm(x), eps) * max(norm(y), eps))
Proposed: dot(x / max(norm(x), eps), y / max(norm(y), eps))
```

For this example, each normalized vector is `[0 / eps, 0 / eps] = [0, 0]`, so the dot product is zero. This preserves the positive-epsilon formula in exact arithmetic while avoiding the product of two tiny denominators.

Add regression coverage for two zero vectors, either input being zero, and explicit epsilon handling.

### Confirmed reproduction

A backend regression test containing the zero-vector example was run against the original implementation on Flex (`f32`):

```sh
BURN_DEVICE=flex cargo test -p burn-backend-tests --no-default-features \
  --features flex,std --test tensor test_cosine_similarity_zero_vectors -- --nocapture
```

It fails with:

```text
Tensors are not eq:
  => Position 0: NaN != 0
```

Mikyx-1:
For two zero `f32` vectors, `cosine_similarity(..., None)` returns `NaN`: the product of the two default epsilon-clamped norms underflows to zero, leaving `0 / 0`.

Normalize each input by its own clamped L2 norm before taking the dot product. This returns zero for the reported case and preserves the existing positive-epsilon formula in exact arithmetic.

### Checklist

- [x] Made sure the book is up to date with changes in this PR (no book changes needed).

### Related Issues/PRs

Fixes #5584.

### Changes

- Avoid multiplying the two clamped norms in `linalg::cosine_similarity`.
- Add regression coverage for two zero vectors and either input being zero.
- Check that a custom epsilon still clamps each norm independently, including when only one norm is below epsilon.

### Testing

- Confirmed the new zero-vector regression fails on the original implementation with `Position 0: NaN != 0`.
- `cargo run-checks`: passed (3,775 tests passed; 39 ignored).
- `BURN_DEVICE=flex cargo test -p burn-backend-tests --no-default-features --features flex,std --test tensor --test tensor_f16 cosine_similarity`: passed (16 tests across f32/f16).
- `BURN_DEVICE=cuda cargo test -p burn-backend-tests --no-default-features --features cuda,fusion,std --test tensor --test tensor_f16 cosine_similarity`: passed (16 tests across f32/f16, RTX 3060).


laggui (commented):
Looks good. We could add another test that exercises denominator underflow with an explicitly small eps, like `1e-30`, and two zero vectors.

Otherwise, just need to merge with main. #5572 migrated the linalg ops into its own extension crate.

Mikyx-1:
@laggui Thanks for the review! I've merged the latest main and added the two-zero-vector regression with an explicit `eps = 1e-30` in the migrated `burn-linalg` tests. I confirmed the new test fails with `NaN != 0` on the original implementation and passes with the fix.

`cargo run-checks` passed (3,604 tests passed, 43 ignored), and all 9 cosine similarity tests passed on CUDA with fusion enabled.

Could you take another look when you have a chance? Thanks!

laggui (approved):

Issue #4362, opened 2026-01-21
tensor: move pooling args into a PoolOptions struct with asymmetric padding


antimora:
### Summary

Pool operations (max_pool1d, max_pool2d, avg_pool1d, avg_pool2d) currently use individual parameters at the functional level, while asymmetric padding is handled at the module level. This is inconsistent with how convolution operations work after #4263.

### Current State

- Conv operations: Use ConvOptions<N> struct with padding_out field for asymmetric padding at the functional level
- Pool operations: Use individual parameters; asymmetric padding handled via explicit tensor.pad() at the module level only

### Proposed Change

1. Create a PoolOptions<N> struct (or separate MaxPoolOptions/AvgPoolOptions) similar to ConvOptions:

```rust 
pub struct PoolOptions<const N: usize> {
    pub kernel_size: [usize; N],
    pub stride: [usize; N],
    pub padding: [usize; N],
    pub padding_out: Option<[usize; N]>,
    pub dilation: [usize; N],  // for max_pool
    pub ceil_mode: bool,
    pub count_include_pad: bool,  // for avg_pool
}
```
2. Update functional pool APIs in burn-tensor/src/tensor/module.rs to use the options struct
3. Move asymmetric padding handling from module level to functional level

### Benefits

- Consistency with conv operations
- Users calling functional pool APIs directly get asymmetric padding support
- Cleaner API with fewer parameters

### Related

- PR #4263: Added asymmetric padding support for conv and pool operations (module-level for pools)

PR #5592, opened 2026-09-07
fix(nn): honor count_include_pad for asymmetric average pooling

### LINKED ISSUE #5450: AvgPool1d/AvgPool2d ignore count_include_pad=false when padding is asymmetric
**Describe the bug**

`AvgPool1d` and `AvgPool2d` silently ignore `count_include_pad(false)` whenever the padding is
asymmetric. The padded cells end up in the divisor anyway, so the values near the padded edges
come out too small.

The symmetric path is correct, so this only shows up when `top != bottom` or `left != right` --
which includes every `PaddingConfig2d::Same` on an extent where the total padding is odd, not
just explicitly asymmetric `Explicit(..)` configs.

**To Reproduce**

Drop this in as a test. The symmetric control passes, both asymmetric cases fail. Expected values
are from ONNX Runtime's `AveragePool` with `count_include_pad=0` and the same kernel/stride/pads.

```rust
use burn::nn::pool::{AvgPool1dConfig, AvgPool2dConfig};
use burn::nn::{PaddingConfig1d, PaddingConfig2d};

#[test]
fn avg_pool2d_asymmetric_honours_count_include_pad() {
    let input = Tensor::<1, Int>::arange(1..26, &device)
        .reshape([1, 1, 5, 5])
        .float();

    let output = AvgPool2dConfig::new([2, 2])
        .with_strides([2, 2])
        .with_padding(PaddingConfig2d::Explicit(0, 0, 1, 1)) // asymmetric
        .with_count_include_pad(false)
        .init()
        .forward(input);

    let expected = TensorData::from([[[
        [4.0f32, 6.0, 7.5],
        [14.0, 16.0, 17.5],
        [21.5, 23.5, 25.0],
    ]]]);
    output.to_data().assert_eq(&expected, true);
}

#[test]
fn avg_pool1d_asymmetric_honours_count_include_pad() {
    let input = Tensor::<1, Int>::arange(1..6, &device)
        .reshape([1, 1, 5])
        .float();

    let output = AvgPool1dConfig::new(2)
        .with_stride(2)
        .with_padding(PaddingConfig1d::Explicit(0, 1)) // asymmetric
        .with_count_include_pad(false)
        .init()
        .forward(input);

    let expected = TensorData::from([[[1.5f32, 3.5, 5.0]]]);
    output.to_data().assert_eq(&expected, true);
}

#[test]
fn avg_pool2d_symmetric_honours_count_include_pad_control() {
    // Control: symmetric padding takes the other branch and is correct.
    let input = Tensor::<1, Int>::arange(1..17, &device)
        .reshape([1, 1, 4, 4])
        .float();

    let output = AvgPool2dConfig::new([3, 3])
        .with_strides([3, 3])
        .with_padding(PaddingConfig2d::Explicit(1, 1, 1, 1)) // symmetric
        .with_count_include_pad(false)
        .init()
        .forward(input);

    let expected = TensorData::from([[[[3.5f32, 5.5], [11.5, 13.5]]]]);
    output.to_data().assert_eq(&expected, true);
}
```

Result:

```
test avg_pool2d_symmetric_honours_count_include_pad_control ... ok
test avg_pool2d_asymmetric_honours_count_include_pad ... FAILED
test avg_pool1d_asymmetric_honours_count_include_pad ... FAILED

  => Position 2: 3.75 != 7.5
  => Position 5: 8.75 != 17.5
  => Position 6: 10.75 != 21.5
  => Position 7: 11.75 != 23.5
  => Position 8: 6.25 != 25
  => Position 2: 2.5 != 5
```

| Case | burn | expected |
| --- | --- | --- |
| 2D, `Explicit(0, 0, 1, 1)` | `[4, 6, 3.75, 14, 16, 8.75, 10.75, 11.75, 6.25]` | `[4, 6, 7.5, 14, 16, 17.5, 21.5, 23.5, 25]` |
| 2D, `Same` (same pads) | identical to the row above | identical to the row above |
| 1D, `Explicit(0, 1)` | `[1.5, 3.5, 2.5]` | `[1.5, 3.5, 5.0]` |
| 2D, `Explicit(1, 1, 1, 1)` (control) | `[3.5, 5.5, 11.5, 13.5]` | same, correct |

Concretely for the 2D case, the top-right window covers input values 5 and 10 plus two padded
cells. With `count_include_pad(false)` the divisor should be 2, giving 7.5. burn divides by 4 and
returns 3.75.

**Expected behavior**

`count_include_pad(false)` should exclude padded cells from the divisor regardless of whether the
padding happens to be symmetric.

**Root cause**

`AvgPool2d::forward` handles asymmetric padding by materialising the padding as zeros and then
pooling with no padding, but still forwards `count_include_pad` to that inner call:

https://github.com/tracel-ai/burn/blob/9cdb20df7ab3ff5572518cb0dae370a3cf85ed02/crates/burn-nn/src/modules/pool/avg_pool2d.rs#L115-L127

```rust
if top != bottom || left != right {
    let padded = input.pad((left, right, top, bottom), PadMode::Constant(0.0));
    // Use zero padding for the pool operation since we already padded
    avg_pool2d(
        padded,
        self.kernel_size,
        self.stride,
        [0, 0],
        self.count_include_pad,   // <-- no padding left to exclude
        self.ceil_mode,
    )
}
```

Once the zeros are real tensor elements, `avg_pool2d` cannot tell them apart from genuine data, so
`count_include_pad` has nothing to act on and the flag becomes a no-op. `AvgPool1d::forward` has
the same shape of bug:

https://github.com/tracel-ai/burn/blob/9cdb20df7ab3ff5572518cb0dae370a3cf85ed02/crates/burn-nn/src/modules/pool/avg_pool1d.rs#L112-L124

**Additional context**

- Found while working on ONNX import in [burn-onnx](https://github.com/tracel-ai/burn-onnx). ONNX
  `AveragePool` defaults to `count_include_pad=0`, and asymmetric pads are routine in exported
  models (any `auto_pad=SAME_UPPER` where the total padding is odd), so imported models hit this
  by default rather than as an edge case. It is silent: shapes are right and only the edge values
  are wrong.
- The existing `// TODO: Move asymmetric padding to functional level via PoolOptions` comment
  right above the affected branch points at #4362. That refactor looks like it would fix this as a
  side effect, since it would let the pooling kernel know where the padding is instead of baking
  it into the tensor. Filing this separately because the correctness bug is worth tracking on its
  own and may want a fix sooner than the refactor.
- #636 added `count_include_pad=False` support; #4263 later added asymmetric padding. The bug is
  in the interaction, not in either feature alone.
- Version: `0.22.0-pre.2` at `9cdb20df7` (also present on `main` at time of filing).
- OS: macOS 15 (Darwin 25.5.0), CPU backend (`burn/flex`). Not backend-specific: the bug is in the
  module layer above the backend.

DivyamTalwar:
### Checklist

### Related Issues/PRs

Closes #5450.

The broader pooling API refactor tracked by #4362 remains out of scope.

### Changes

`AvgPool1d` and `AvgPool2d` materialize asymmetric padding before invoking the backend pool operation. At that point the synthetic zeros are ordinary tensor elements, so the backend can no longer exclude them from the divisor when `count_include_pad=false`. Edge values were therefore too small even though output shapes were correct.

For the asymmetric exclude-pad path, this change pools a singleton validity mask with the same kernel, stride, and ceil-mode settings, then divides the pooled input by the resulting valid-element fraction. The mask contains ones for every real input position—including legitimate zero-valued inputs—and zeros only for materialized padding. Its singleton batch/channel dimensions broadcast over the input, and its dtype and device match the input. A zero validity fraction is replaced with one before division so a fully padded window remains finite at zero.

Symmetric padding and `count_include_pad=true` continue through their existing paths. The change stays within `burn-nn`; it adds no dependency or backend/API change. The asymmetric exclude-pad path now performs one additional mask pooling operation.

No book update is needed because the public API and documented configuration semantics are unchanged; this corrects the implementation of the existing `count_include_pad` behavior. The diff contains only the two pooling modules and their inline regressions.

AI tools assisted with implementation and review. In line with Burn's change-ownership policy, responsibility for understanding, explaining, and maintaining the change remains with the PR author.

### Testing

Tested commit: `112e6733df92acfaac9ddbc53154674d19b13d7b`.

#### Red evidence

The regressions were exercised before applying the validity-mask correction:

- `BURN_DEVICE=flex cargo +1.96.1 test -p burn-nn --features burn-core/flex asymmetric_padding_excludes -- --nocapture` exited 101 with 0 passed and 4 failed. The 1D explicit and odd-`Same` cases each produced `2.5 != 5` at position 2; the 2D explicit case included `3.75 != 7.5`, `8.75 != 17.5`, `10.75 != 21.5`, `11.75 != 23.5`, and `6.25 != 25`; the 2D gradient case produced `0.25 != 0.5` at padded edges.
- `BURN_DEVICE=flex cargo +1.96.1 test -p burn-nn --features burn-core/flex ceil_mode_excludes -- --nocapture` exited 101 with `1 != 2` at position 0.
- `BURN_DEVICE=flex cargo +1.96.1 test -p burn-nn --features burn-core/flex fully_padded_windows -- --nocapture` exited 101 with `1 != 4` at position 1.
- With validity normalization temporarily disabled, `BURN_DEVICE=flex cargo +1.96.1 test -p burn-nn --features burn-core/flex asymmetric_padding_mask_broadcasts_without_excluding_input_zeros -- --nocapture` exited 101 with `2 != 4`, `3 != 6`, and `4 != 8` at positions 1, 3, and 5. This final regression covers two batches, two channels, genuine input zeros, and cancelling nonzero values.
- The `count_include_pad=true` control passed before the fix: 1 passed, 0 failed.

#### Green evidence

- `BURN_DEVICE=flex cargo +1.96.1 test -p burn-nn --features burn-core/flex avg_pool` — exit 0; 23 passed, 0 failed.
- `BURN_DEVICE=flex cargo +1.96.1 test -p burn-nn --features burn-core/flex` — exit 0; 514 unit tests and 6 integration tests passed with 0 failures; doctest runs completed successfully.
- `cargo +1.96.1 run-checks` — exit 0 on the tested commit. Formatting, typo scanning, dependency audit, workspace Clippy with warnings denied, no-std checks, and release Flex backend tests completed successfully.
- Independent exact-SHA verification also passed `git diff --check HEAD^ HEAD`, `cargo +1.96.1 fmt --all -- --check`, and the 23-test focused average-pooling command above.

Rust 1.96.1 was used because the installed default Rust 1.93.0 cannot build the current dependency set: `cubecl-zspace 0.11.0-pre.3` and `hi_sparse_bitset 0.9.0` require Rust 1.95 or newer.

Validation used Flex on the host CPU. No GPU hardware/runtime backend matrix was run, and the new pooling regressions were not parameterized across the full supported dtype matrix.

### DIFF
--- a/crates/burn-nn/src/modules/pool/avg_pool1d.rs
+++ b/crates/burn-nn/src/modules/pool/avg_pool1d.rs
@@ -110,18 +110,45 @@ impl AvgPool1d {
         // See: https://github.com/tracel-ai/burn/issues/4362
         // Handle asymmetric padding by applying explicit pad operation first
         if left != right {
+            let valid = if self.count_include_pad {
+                None
+            } else {
+                let device = input.device();
+                Some(
+                    Tensor::<3>::ones([1, 1, length], (&device, input.dtype()))
+                        .pad((left, right, 0, 0), PadMode::Constant(0.0)),
+                )
+            };
             // Burn's pad takes (left, right, top, bottom) for the last two dimensions
             // For 1D (NCL format), we only pad L (last dim), so top/bottom = 0
             let padded = input.pad((left, right, 0, 0), PadMode::Constant(0.0));
             // Use zero padding for the pool operation since we already padded
-            avg_pool1d(
+            let output = avg_pool1d(
                 padded,
                 self.kernel_size,
                 self.stride,
                 0,
                 self.count_include_pad,
                 self.ceil_mode,
-            )
+            );
+
+            if let Some(valid) = valid {
+                // Materialized padding is indistinguishable from input to the backend. Pooling a
+                // validity mask with the same settings recovers the fraction of real values in
+                // each window, including partial windows created by ceil mode.
+                let valid = avg_pool1d(
+                    valid,
+                    self.kernel_size,
+                    self.stride,
+                    0,
+                    false,
+                    self.ceil_mode,
+                );
+                let empty = valid.clone().equal_elem(0.0);
+                output / valid.mask_fill(empty, 1.0)
+            } else {
+                output
+            }
         } else {
             // Symmetric padding
             avg_pool1d(
@@ -139,6 +166,7 @@ impl AvgPool1d {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use burn::tensor::{Device, TensorData, Tolerance};
     use rstest::rstest;
 
     #[test]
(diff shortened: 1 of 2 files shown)


laggui (approved):
LGTM as a targeted fix.

We should tackle #4362 now that asymmetric padding support was also added for convolutions in #5521

Issue #4855, opened 2026-04-21
burn-flex: LU on tall matrices panics in mask_blend_fill_u32 (index OOB)


antimora:
## Summary

Three `lu` tests in `burn-backend-tests` panic with an out-of-bounds index in `mask_blend_fill_u32` (`crates/burn-flex/src/simd/portable.rs:791`) when the Flex backend's SIMD `mask_fill_f32` path runs LU decomposition on rectangular matrices.

The panic signature is consistent: the reported slice length matches the *square* submatrix size `min(rows, cols) * min(rows, cols)`, not the input tensor's total element count, which points at a shape/broadcast mismatch inside LU's mask-fill call.

## Failing tests

```
tensor::float::linalg::lu::test_lu_2d_tall            len = 25     (5 * 5)
tensor::float::linalg::lu::test_lu_medium_tall        len = 16384  (128 * 128)
tensor::float::linalg::lu::test_lu_500x300_block_dispatch   len = 90000  (300 * 300)
```

Input shapes:
- `test_lu_2d_tall`: `[8, 5]`
- `test_lu_medium_tall`: `[256, 128]`
- `test_lu_500x300_block_dispatch`: `[500, 300]`

Panic:
```
thread '...' panicked at crates/burn-flex/src/simd/portable.rs:791:35:
index out of bounds: the len is N but the index is N
```

The wide equivalents (`test_lu_2d_wide`, `test_lu_medium_wide`, `test_lu_300x500_block_dispatch`) pass, so the bug is specific to tall rectangular inputs.

## Suspected location

`mask_blend_fill_u32` iterates `for i in 0..tensor.len()` and indexes `mask[i]` (line 791 / 783). The only callers are `mask_fill_f32` / `mask_fill_f64` in `crates/burn-flex/src/ops/mask.rs`, which pass `tensor.storage::<f32>()` and `mask.bytes()` after `broadcast_binary` + `to_contiguous`. The symptom (`tensor.len() > mask.len()`) suggests one of those post-broadcast lengths is wrong when LU produces an asymmetric mask/tensor pair (e.g., mask covering the square factor while tensor covers a larger intermediate).

## Repro

```
cargo test -p burn-backend-tests --test tensor tensor::float::linalg::lu
```

[...]

PR #5603, opened 2026-09-08
perf(flex): write slice_assign in place when the destination is uniquely owned

### LINKED ISSUE #5575: burn-flex: slice_assign always copies the whole tensor, even when uniquely owned
**Describe the bug**
On flex backend of burn 0.22.0-pre.3, slice_assign is O(tensor size) per call regardless of how small the assigned slice is.

This contradicts the documented behaviour in the burn-flex README:

> **Arc-based Copy-on-Write**: O(1) tensor cloning with automatic COW semantics.
> **In-place mutation when uniquely owned.**

burn-ndarray backend do not have this problem — I think it checks for unique ownership before falling back to a copy.

**To Reproduce**
## Reproduction

Cargo.toml
```toml
[package]
name = "burn_issue"
version = "0.1.0"
edition = "2024"

[dependencies]
burn = { version = "0.22.0-pre.3", features = ["flex", "ndarray"]}
```` 

src/main.rs
```rust
use burn::prelude::*;
use std::time::Instant;
 
fn main() {
    let capacity: usize = std::env::args().nth(1).unwrap().parse().unwrap();
    let device = Device::default();
 
    let mut store = Tensor::<2>::zeros([capacity, 4], &device);
    let row = Tensor::<2>::zeros([1, 4], &device);
 
    // warm up
    for i in 0..1_000 {
        let idx = i % capacity;
        store.inplace(|a| a.slice_assign(idx..idx + 1, row.clone()));
    }
 
    let n = 200_000;
    let t0 = Instant::now();
    for i in 0..n {
        let idx = i % capacity;
        store.inplace(|a| a.slice_assign(idx..idx + 1, row.clone()));
    }
    let dt = t0.elapsed();
    println!("{capacity}: {:.1} ns/push", dt.as_nanos() as f64 / n as f64);
}
```
 
```
BURN_DEVICE=flex    cargo run --release -- 1000
BURN_DEVICE=flex    cargo run --release -- 10000
BURN_DEVICE=flex    cargo run --release -- 100000
BURN_DEVICE=ndarray cargo run --release -- 1000
BURN_DEVICE=ndarray cargo run --release -- 10000
BURN_DEVICE=ndarray cargo run --release -- 100000
```

**Expected behavior**
As the number of capacity (size of tensor) increases, the execution time increases for flex backend, while ndarray backend stays constant.  
My result was:
|size of tensor|flex|ndarray|
|:---:|:---:|:---:|
|1000|699.1|477.3|
|10000|3010.3|478.6|
|100000|29359.3|491.5|

**Desktop:**
 - OS: Linux (Ubuntu 26.04)

hdimer:
Fixes #5575.

On the flex backend, `slice_assign` copies the entire destination tensor on every
call, so it costs O(destination) even when the destination is uniquely owned. On
the reporter's benchmark a small row write into a `[capacity, 4]` store scales
with `capacity`, which contradicts burn-flex's documented "in-place mutation when
uniquely owned".

## Cause

`slice_write_impl` takes its destination by value and then does

```rust
let mut tensor = tensor.to_contiguous();
```

`to_contiguous` borrows, so on its already-contiguous fast path it has to hand
back `self.clone()`, a second `Arc` handle onto the same buffer. The `let` only
*shadows* the parameter; the parameter is never moved, so it stays alive to the
end of the function still holding its handle. `Arc::strong_count` is therefore 2
by the time `storage_mut()` reaches `Arc::make_mut`, and the copy-on-write check
copies the whole buffer regardless of who actually owned the tensor.

## Fix

Turned the contiguous copy into a consuming `into_contiguous(self)`, whose fast
path returns `self` rather than a clone, and pointed `slice_write_impl` at it.
`to_contiguous(&self)` stays as it was and now delegates (`self.clone()
.into_contiguous()`), so the fast-path predicate has a single definition and the
two cannot drift. The cold path pays one extra refcount bump ahead of a full
buffer copy.

No second handle is created, so `storage_mut()` sees the real refcount and writes
in place when the caller owned the tensor. A genuinely shared destination still
takes the copy-on-write copy, exactly as before.

I kept `into_contiguous` crate-private since `slice_write_impl` is its only
caller; happy to make it `pub` for parity with `to_contiguous` if you'd prefer.

Two scope notes:

- I checked the crate's other `let t = t.to_contiguous()` sites.
  `slice_write_impl` is the only one that shadows an owned binding and then
  mutates through `storage_mut`; the rest are read-only or already take
  `mut tensor` directly, so the extra handle costs them nothing. This stays a
  one-concern change.
- `into_contiguous` keeps its whole-buffer length check, which is there (#4855)
  for callers that read `storage()` by length. `slice_write_impl` is not one of
  them, so after review it uses the crate's usual in-place predicate instead
  (`is_unique() && layout().is_contiguous() && layout().start_offset() == 0`, as
  in `bool.rs:194` and `unary.rs:81`) and a uniquely-owned contiguous *prefix
  view* destination is now written in place as well. That covers the
  `store.narrow(..)` then `slice_assign` pattern. A *shared* prefix view is still
  compacted: letting `Arc::make_mut` take that copy would duplicate the parent's
  whole buffer rather than just the view. One consequence worth naming is that
  `slice_assign` can now return a tensor whose buffer is longer than its shape,
  and which holds the parent allocation alive, exactly as `slice`, `narrow` and
  `unary_op_typed` already do.

## Measurements

20 000 `slice_assign` calls writing a `[1, 4]` row into a `[capacity, 4]` store,
release build:

| capacity | before | after |
| ---: | ---: | ---: |
| 1 000 | 2 255 ns/op | 706 ns/op |
| 10 000 | 13 446 ns/op | 641 ns/op |
| 100 000 | 122 990 ns/op | 624 ns/op |

## Testing

Two new tests pinning both halves of the ownership contract:

- `test_slice_assign_unique_destination_writes_in_place` asserts the returned
  tensor's buffer address is unchanged. It fails on `main` and passes with the
  fix.
- `test_slice_assign_shared_destination_copies` asserts a cloned destination
  still gets the copy-on-write copy and the other holder is left untouched. This
  one passes either way by design; it exists so the guarantee the fix relies on
  can't be quietly broken later.

Both follow the crate's existing `test_clone_is_cheap` / `test_cow_on_mutation`
pointer-comparison idiom.

Three more cover the prefix-view path added in review:

- `test_slice_assign_unique_prefix_view_writes_in_place` fails on the first
  commit (the buffer pointer moves), and asserts the rows past the logical prefix
  are untouched.
- `test_slice_assign_shared_prefix_view_compacts` fails `40 != 25` if the
  `is_unique()` conjunct is dropped.
- `test_slice_assign_unique_non_prefix_destinations_are_normalized` pins the
  other two conjuncts. Before it, dropping either `is_contiguous()` or
  `start_offset() == 0` left all 395 tests green: nothing in the suite hands
  `slice_assign` an offset or transposed destination, and the 1-D and 2-D
  branches compute indices from canonical strides with no `start_offset` term.

`cargo test -p burn-flex` is green (396 tests), as are `cargo xtask check format`
and `cargo xtask check lint` across the workspace.

I used an AI assistant while investigating and writing this. I have read, tested
and understood every line, and can explain the reasoning behind it.

### DIFF
--- a/crates/burn-flex/src/ops/slice.rs
+++ b/crates/burn-flex/src/ops/slice.rs
@@ -233,7 +233,19 @@ fn slice_write_impl<E: Element + bytemuck::Pod>(
     slices: &[Slice],
     source: WriteSource<'_, E>,
 ) -> FlexTensor {
-    let mut tensor = tensor.into_contiguous();
+    // A uniquely-owned contiguous prefix view needs no copy: every write
+    // below is computed from `dst_layout`, whose canonical strides and zero
+    // start offset keep it inside the logical prefix even when the buffer is
+    // longer. `into_contiguous` additionally demands a right-sized buffer
+    // (#4855) for callers that read `storage()` by length; this one doesn't.
+    let mut tensor = if tensor.is_unique()
+        && tensor.layout().is_contiguous()
+        && tensor.layout().start_offset() == 0
+    {
+        tensor
+    } else {
+        tensor.into_contiguous()
+    };
     let dst_layout = tensor.layout().clone();
     let ndims = dst_layout.num_dims();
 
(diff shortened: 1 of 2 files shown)


antimora:
The diagnosis matches what I found (shadowing keeps a second `Arc` alive through `storage_mut`), and the consuming `into_contiguous(self)` is a good way to fix it: it removes the extra handle at the source instead of adding an `is_unique()` special case in `slice_write_impl`, and `to_contiguous` delegating to it keeps one definition of the fast-path predicate. I verified the fast path is semantically unchanged (returns the same tensor) and the cold path is the same copy as before plus one refcount bump. The alias test is the right guard for the COW contract.

One follow-up worth considering, either here or separately: the whole-buffer length check in the fast path is there to protect callers that read `storage().len()` (the `mask_fill_*` SIMD kernels, #4855). `slice_write_impl` is not one of those callers. Every write it does is computed from `dst_layout` (canonical strides, offset 0), so it stays inside the logical prefix even when the buffer is longer. That means `slice_write_impl` could accept a unique contiguous prefix view without copying:

```rust
let mut tensor = if tensor.is_contiguous() && tensor.layout().start_offset() == 0 {
    tensor
} else {
    tensor.into_contiguous()
};
```

That would cover the `store.narrow(..)` then `slice_assign` pattern you mention in the scope note. If you'd rather keep this PR to the one concern, that is fine too; I can open a follow-up.

hdimer:
Took it, with one change: I kept a uniqueness check in front of the predicate.

Your version skips the copy for any contiguous, offset-0 destination. For a uniquely-owned prefix view that is exactly right, and it is the case you were after. For a *shared* prefix view it makes things worse: the fast path hands the copy to `Arc::make_mut`, which duplicates the whole backing buffer instead of the logical prefix, and the result carries that oversized buffer forward.

Measured on an `[8, 5]` store narrowed to `[5, 5]` (25 logical elements over a 40-element buffer), assigning into row 0:

| destination | this PR before | your predicate | with `is_unique()` |
| --- | --- | --- | --- |
| unique prefix view | copy, 40 -> 25 | in place, no copy | in place, no copy |
| shared prefix view | copy of 25 | copy of 40 | copy of 25 |

The shared row is what made me add the check. `store.narrow(..)` then `slice_assign` with `store` still alive has to copy either way, and the copy should be the size of the view rather than of the store: narrowing a `[1_000_000, 4]` store to `[8, 4]` would go from copying 32 elements to copying 4M of them, per call.

I know you liked that the original fix avoided an `is_unique()` special case. This one isn't doing that job. It doesn't decide whether we have the refcount bug, it decides which copy to pay when we can't write in place, and `storage_mut`'s COW check is still what enforces correctness. It also lands on the crate's existing in-place idiom, `is_unique() && layout().is_contiguous() && layout().start_offset() == 0`, which `bool.rs:194` and `unary.rs:81` already use. `unary_op_typed` is the precedent for your actual point, incidentally: same predicate, no whole-buffer check, writes `storage[..n]` and returns an oversized-buffer tensor.

```rust
let mut tensor = if tensor.is_unique()
    && tensor.layout().is_contiguous()
    && tensor.layout().start_offset() == 0
{
    tensor
} else {
    tensor.into_contiguous()
};
```

Three tests now, and I checked each one discriminates:

- `test_slice_assign_unique_prefix_view_writes_in_place` fails on the previous commit (the buffer pointer moves). It also asserts the rows past the logical prefix are untouched, which is the property your argument rests on.
- `test_slice_assign_shared_prefix_view_compacts` fails `40 != 25` without the `is_unique()` conjunct.
- `test_slice_assign_unique_non_prefix_destinations_are_normalized` covers the other two conjuncts. Worth flagging: before I added it, dropping *either* `is_contiguous()` or `start_offset() == 0` left all 395 tests green. Nothing in the suite ever hands `slice_assign` an offset or transposed destination, and the 1-D and 2-D branches compute indices from canonical strides with no `start_offset` term, so a unique `narrow(0, 1, 5)` destination under a weakened predicate writes to the wrong rows and reports success. The test pins both.

One behaviour change worth naming: `slice_assign` can now return a tensor whose buffer is longer than its shape, where it always returned a compacted one before, which also means it holds the parent allocation alive. `slice`, `narrow` and `unary_op_typed` already do both, and every consumer that keys off `storage().len()` calls `to_contiguous` first (`mask.rs:57`, `sort.rs:112`, `float.rs:62`), so I didn't find a caller this breaks. `into_contiguous` keeps its whole-buffer check for everyone else, so #4855 stays fixed.

`cargo test -p burn-flex` is green (396), as are `cargo xtask check format` and `cargo xtask check lint`.

On CI: the red `macos-std-tests` is `int::ops::powi::should_support_powi_broadcast` failing `8 != 9` and `26 != 27`, which is that job's Metal backend. This PR only touches burn-flex, and main's own macOS run is red too (5 failures in burn-linalg det/lu), so I don't think it's mine. Both codecov checks pass.

laggui (approved):

PR #5617, opened 2026-09-08
perf(flex): optimize views, copies, in-place ops, and Rayon parallelism (#5613)

### LINKED ISSUE #5613: burn-flex: performance follow-ups from audit (avoidable copies, serial backward passes, dead code)
Collected from an audit of `crates/burn-flex`. Each item is independent; splitting into separate PRs is fine. Attention parallelism is tracked separately.

## Avoidable copies

- **`mask_fill` / `mask_where` never write in place** (`crates/burn-flex/src/ops/mask.rs:25`, `:161`). Both call `to_contiguous()` on every operand and allocate a fresh output even when `tensor` is unique and contiguous. `mask_where` also materializes a broadcast `value` (e.g. a scalar expanded to the full shape) through `broadcast_three` (`mask.rs:184`). The collapsed loop nest in `zip.rs` used by `binary_op_typed` handles in-place plus broadcast already and could be reused. `mask_fill` is on the hot path for attention masks and `triu`/`tril`.
- **`Layout::reshape` rejects contiguous views with a non-zero offset** (`crates/burn-flex/src/layout.rs:297`), so `x.slice(...).reshape(...)` copies. Verified: the output buffer pointer differs. Preserving `start_offset` is safe; all callers go through `contiguous_offsets()` / `storage()` with offsets.
- **Negative-step `slice` copies element by element** through a recursive walker (`crates/burn-flex/src/ops/slice.rs:27` to `:140`). The layout already supports negative strides (flip uses them), so `Layout::slice` could emit `stride * step` with an adjusted offset and the whole copy path can be deleted.
- **`scatter_update` / `select_update` always `to_vec()` the target** (`crates/burn-flex/src/ops/gather_scatter.rs:463`, `:911`) even when it is unique. `embedding_backward` (`crates/burn-flex/src/ops/module.rs:797`) allocates zeros and then copies them again through this path.
- **`float_cast` round-trips through `Vec<f64>`** (`crates/burn-flex/src/ops/float.rs:834`): two allocations and two passes per cast.

## Missing or unthrottled parallelism

- N-D `gather` and the element-wise N-D `select` path use `into_par_iter` with no size threshold (`crates/burn-flex/src/ops/gather_scatter.rs:218`, `:606`); the 2D fast paths gate on `PARALLEL_THRESHOLD`. Small tensors pay rayon dispatch for nothing.
- conv_transpose col2im is fully serial over batch and groups (`crates/burn-flex/src/ops/conv_transpose.rs:273`), with per-element bounds checks in the inner loop. The analytic `valid_out_range` trick from `conv.rs` applies here too.
- All pool backward passes are serial (`crates/burn-flex/src/ops/pool.rs:1227`, `:1402`, `:1565`), as are bilinear and bicubic interpolate backward (`crates/burn-flex/src/ops/interpolate.rs:963`, `:1041`). Forward passes for the same ops are parallel over `(batch, channel)`.
- Softmax for f64 / f16 / bf16 spawns one rayon task per row (`crates/burn-flex/src/ops/activation.rs:416`); the f32 path chunks 64 rows per task.
- Batched `matmul_2d_i32` / `matmul_2d_i64` have no parallelism at all (`crates/burn-flex/src/ops/matmul.rs:539`, `:735`).

## Dead code

- `add/sub/mul/div_shared_row_inplace_f32` (`crates/burn-flex/src/simd/portable.rs:119` to `:224`, re-exported from `simd/mod.rs:26`) have no callers since the `SharedRow` fast path was removed from `binary_op_f32`.
- `try_broadcast_cmp_f32` (`crates/burn-flex/src/ops/comparison.rs:171` to `:275`) is subsumed by `try_zip_cmp_f32`, which runs right after it and covers the same stride patterns.

Sadik00789:
Resolves #5613

### Overview
This PR addresses performance bottlenecks, avoidable buffer copies, serial backward passes, memory layout edge cases, and cold-path regressions identified across `crates/burn-flex`.

---

#### 1. Dead Code & Precision Cleanup
* Pruned unused portable SIMD kernels (`add/sub/mul/div_shared_row_inplace_f32`) and their re-exports from `simd/mod.rs`.
* Removed redundant `try_broadcast_cmp_f32` (subsumed by `try_zip_cmp_f32`) and dead helpers in `zip.rs`.
* Avoided intermediate double-rounding in `cast_slice` by converting `f64 -> f16 / bf16` directly via `from_f64`.

#### 2. Zero-Copy Views & Negative-Step Slicing
* **`Layout::reshape`**: Removed `start_offset != 0` rejection check; preserved `start_offset` on contiguous reshaped views.
* **Negative-step slicing**: Deleted the recursive copy walker in `slice.rs`; delegated negative steps directly to layout strides with an early underflow guard for empty slices (`n == 0`).

#### 3. `to_contiguous` Cold Path Acceleration
* **Size-1 dimension contiguity**: Squeezed unit dimensions in `Layout::is_contiguous` so shapes like `[1, N]` with stride `[1, N]` (transposed from `[N, 1]`) remain contiguous and zero-copy.
* **Batched transpose tiling**: Added `copy_batched_2d_tiled` for 3D+ tensors with transposed trailing 2D dimensions to avoid falling back to `StridedIter`.
* **Negative-stride flips**: Relaxed `all_positive` gating in `collapse_for_copy`; added `copy_2d_negative` with reverse-row `copy_from_slice` for outer flips and reverse iteration for inner flips.
* **Tuned parallel 2D tiling**: Introduced `COPY_2D_PARALLEL_THRESHOLD` (1M elements / ~4 MB) to prevent thread pool overhead on cache-resident copies. Tasks are dynamically chunked to 1 coherent slice per thread, with stride-aware inner loop ordering (`row_stride <= col_stride`) preserving cache line locality.
* **Allocation pruning**: Replaced heap-allocated multi-index `Vec` in `StridedIter` with a fixed stack buffer `[usize; 8]`.

#### 4. Avoidable Allocations & In-Place Mutation
* **Dense storage contract**: Consolidated contiguity and non-broadcast guarantees into `Layout::is_dense_unique_storage` across `mask.rs` and `gather_scatter.rs`.
* **In-place masking**: Allowed in-place mutation in `mask_fill` for unique contiguous buffers with any `start_offset >= 0` (guarded against broadcast stride 0).
* **Streaming `mask_where`**: Replaced eager `broadcast_three` with a newly implemented `Zip3Nest` (`collapse_for_zip3` / `zip3_apply_inplace` / `zip3_map`) supporting heterogeneous types `(M, T, T)`.
* **In-place updates**: Enabled in-place mutation in `scatter_update` and `select_update` for unique contiguous targets; prevents buffer duplication in `embedding_backward`.
* **Direct casting**: Replaced intermediate `Vec<f64>` round-trips in `float_cast` with direct `Dst::from_elem(*v)` streaming.

#### 5. Parallelism, Gating & Reductions
* Gated N-D `gather`, N-D `select`, and multi-group `conv_transpose` behind `PARALLEL_THRESHOLD` to eliminate thread dispatch overhead and multi-buffer allocations on small workloads.
* Corrected row pitch calculation in `reduce_dim_f32` and `reduce_last_dim_f32` by computing `outer_stride` analytically, fixing incorrect reductions on swapped/permuted singleton dimensions.
* Parallelized `col2im` over `(batch, groups)` onto disjoint memory slices with analytic `valid_in_range` bounds.
* Parallelized backward passes for pooling (`max`, `avg`, `adaptive`) and interpolation (`bilinear`, `bicubic`) across `(batch, channel)` slices.
* Standardized `f64`, `f16`, and `bf16` softmax row dispatch to `.par_chunks_mut(64 * last)`.
* Parallelized outer loops in `matmul_2d_i32`, `matmul_2d_i64`, and batched integer matmuls.

---

### Verification
* **Unit tests**:
  * `cargo test -p burn-flex`: 412 passed, 0 failed.
  * `cargo test -p burn-flex --no-default-features --features std`: 374 passed, 0 failed.
* **Linter & Formatting**:
  * `cargo clippy -p burn-flex --all-targets -- -D warnings` clean.
  * `cargo clippy -p burn-flex --no-default-features --features std --all-targets -- -D warnings` clean.
  * `cargo fmt --check -p burn-flex` clean.
* **Integration suites**: All 1,185 integration tests in `burn-backend-tests --features flex` passing clean.
* **Benchmark results**: Confirmed 8.1x speedup on int matmul, 7.3x on non-last axis hooks, 5.4x on conv_transpose, with 0% regression on `convnext_reshape` composite benchmarks.
* Added regression unit tests for:
  * Zero-sized negative-strided view copies.
  * Stride-aware parallel 2D tiled copies across both row-major and column-major patterns.
  * Reductions on swapped and permuted singleton dimensions (`[3, 4, 1].swap_dims(1, 2)` and `[1, 3, 4].permute([1, 0, 2])`).
  * In-place masking, scatter updates, and pointer equality on sliced-then-reshaped zero-copy views.


Sadik00789:
Addressed all three review items in the latest commit:

1. **Zero-sized negative views (`tensor.rs`)**: Added an early return when `n == 0` prior to checking rank/stride logic in `copy_contiguous`, preventing buffer overruns/underflows on empty negative-strided views. Added `test_to_contiguous_zero_sized_negative_stride`.
2. **Stride-aware tile ordering (`tensor.rs`)**: Updated `copy_2d_tiled` in both the serial and parallel Rayon chunk runners to conditionally place contiguous dimension loops innermost (`col` inner for row-major/offset views, `r` inner for column-major), keeping reads and writes aligned with cache lines. Added `test_to_contiguous_2d_row_stride_lt_col_stride`.
3. **`conv_transpose` parallel gating (`conv_transpose.rs`)**: Gated Rayon multi-group parallelization behind `(total_groups * out_channels_per_group * out_spatial) >= PARALLEL_THRESHOLD`. Workloads beneath the threshold stay on the single-buffer serial path to eliminate task spawn latency and duplicate scratchpad allocations.

All 408 crate unit tests, `burn-backend-tests` integration suites, Clippy, and formatting checks are passing clean.

antimora:
## Benchmark audit

Ran a full A/B of this branch against its parent (`5ecfb3896`) using the `burn-backend-tests` benches on the flex backend (`--features flex`, so simd + rayon are both active). 408 benches across all 22 suites, 5 interleaved base -> PR passes on an M3 Max (16 cores), scored as min-of-medians.

Net result is strongly positive. One reproducible regression is worth fixing before merge.

### Distribution

| band | benches |
|---|---|
| faster > 2x | 5 |
| faster 10-50% | 31 |
| faster 3-10% | 58 |
| within +/-3% | 278 |
| slower 3-10% | 31 |
| slower > 10% | 5 |

Three of the five in the bottom band are noise (bimodal benches where base happened to catch a fast outlier), and the other two are the same bug. Filtering to results where the PR's *worst* pass still loses to base's *best* pass leaves 16 strictly faster and 3 strictly slower.

### Regression: the new rayon path in `copy_2d_tiled`

| bench | base | PR |
|---|---|---|
| `convnext_reshape_layer_norm_composite/c384_30x28` | 438-477 us | 549-613 us (+25%) |
| `convnext_reshape_layer_norm_composite/c192_61x56` | 913-992 us | 1010-1093 us (+11%) |

Consistent across all 5 passes with no overlap between the two sides.

Profiling both builds on `c384_30x28`: the *fastest* sample is effectively identical (409 us base, 415 us PR) while the median diverges. Base runs the entire bench on the main thread and never wakes the rayon pool. PR runs 16 workers whose stacks are dominated by `crossbeam_deque::Stealer::steal` and `WorkerThread::wait_until_cold`, underneath `burn_flex::tensor::copy_2d_tiled::{closure}`.

Cause: the bench does `[1,384,30,28].permute([0,2,3,1])`, and `collapse_for_copy` folds that view to 2D `[840, 384]` with strides `[1, 840]`. That makes `n = 322_560`, just over `PARALLEL_THRESHOLD` (262_144), so the new branch at `crates/burn-flex/src/tensor.rs:802` splits a 1.3 MB copy into `TILE * cols = 6144`-element chunks: roughly 53 tasks of 24 KB each. Fork, steal and join cost more than the copy does.

Verified by changing only that one condition and rebuilding:

```diff
  #[cfg(feature = "rayon")]
- if n >= crate::ops::PARALLEL_THRESHOLD {
+ if false && n >= crate::ops::PARALLEL_THRESHOLD {
```

| build | `c384_30x28`, 3 passes |
|---|---|
| base (parent commit) | 451.7 / 450.8 / 453.0 us |
| PR as submitted | 578.0 / 564.1 / 548.9 us |
| PR, rayon branch disabled | 459.5 / 456.1 / 458.4 us |

Two things to address:

1. Chunk size. The threshold counts elements, but this is a bandwidth-bound shuffle where 256K elements is far too small to pay for a fork and join. Sizing chunks as `n / current_num_threads()` rounded to whole tile-rows, instead of a fixed `TILE * cols`, would keep the parallel path useful at large sizes without shredding small ones.
2. Loop order. The serial path picks its nesting from `row_stride <= col_stride` so the inner loop always walks the smaller stride, and its doc comment explains why. The parallel branch always nests col-outer / row-inner, so whenever `row_stride > col_stride` it uses exactly the cache-hostile order the serial code exists to avoid. This bench does not hit that case, so it is latent rather than measured.

### Minor: removing `try_broadcast_cmp_f32`

`greater_broadcast/broadcast_256x256` moves from 3.9-4.0 us to 4.1-4.3 us, consistently across passes. Patterns 1 and 3 of the deleted function are genuinely subsumed by `try_zip_cmp_f32`, but pattern 2 is not when the broadcast operand's inner stride is not 1: the collapsed nest rejects that stride pair and the op falls through to the generic `compare_typed`. Small (4% on a 4 us op), and I did not isolate it the way I isolated the copy regression.

### Wins

| bench | base | PR | speedup |
|---|---|---|---|
| `int_square/matmul_512x512` | 17.94 ms | 2.22 ms | 8.1x |
| `non_last_axis_f32/trait_hook_dim1_of_3` | 5.36 ms | 731.8 us | 7.3x |
| `conv_transpose2d_8x64x14x14_to_28x28` | 3.54 ms | 657.7 us | 5.4x |
| `int_square/matmul_256x256` | 1.85 ms | 508.3 us | 3.6x |
| `conv_transpose1d_8x128x64_to_128` | 422.0 us | 152.3 us | 2.8x |
| `permuted_input_drilldown/p2_mean_dim_last_on_permuted` | 1.07 ms | 604.2 us | 1.8x |
| `conv_transpose2d_gan/dcgan_layer3_1x128x8x8_to_16x16` | 320.9 us | 201.6 us | 1.6x |
| `conv2d_depthwise/depthwise_3x3_4x32x56x56` | 188.4 us | 119.5 us | 1.6x |
| `conv_transpose3d_1x32x4x4x4_to_8x8x8` | 239.5 us | 152.5 us | 1.6x |

### Not covered by these benches

"No regression" does not extend to these, since nothing in the suite exercises them:

- There is no `mask_fill` or `mask_where` bench anywhere in `burn-backend-tests`, so the new in-place mask paths and the `Zip3Nest` streaming rewrite are entirely unmeasured. Worth noting that the in-place `mask_fill_f32` path replaces a SIMD kernel with a scalar branching loop, and is gated on `is_unique()`.
- `slice_ops` only covers positive steps, so the new zero-copy negative-stride slicing is unmeasured, along with the `copy_2d_negative` paths in `to_contiguous` that it now feeds.
- `benches/common/mod.rs` pins the default dtypes to f32/i32, so the f64, f16 and bf16 softmax row-chunking changes never execute.
- `StridedIter`'s heap fallback for rank above 8 is not reached by any bench.

### Reproducing

```bash
# on this branch
gh pr checkout 5617
cargo bench -p burn-backend-tests --features flex \
    --bench norm_ops -- convnext_reshape --min-time 0.6

# on the parent commit
git checkout HEAD^
cargo bench -p burn-backend-tests --features flex \
    --bench norm_ops -- convnext_reshape --min-time 0.6
```

Compare the `median` column for `c384_30x28`. Expect roughly 440-480 us on the parent and 550-615 us here.

One gotcha if you run the built bench binaries directly rather than through cargo: divan defaults to test mode and prints the bench tree with no timings. Pass `--bench` explicitly.

Sadik00789:
Thanks for the detailed benchmark and profiling breakdown @antimora! The speedups across integer matmul (8.1x), trait hooks (7.3x), and conv_transpose (5.4x) are great to see.

Your diagnosis of the `convnext_reshape` regression was spot on. I've addressed all points in the latest commit:

1. **Dedicated Copy Threshold (`COPY_2D_PARALLEL_THRESHOLD`)**: Raised the parallel activation floor from 256K to 1,048,576 elements (~4 MB for f32). Cache-resident shuffles like `c384_30x28` (~1.3 MB) now stay strictly on the serial path, eliminating Rayon worker wakeups and `crossbeam` work-stealing contention.
2. **Thread-Proportional Chunking**: Replaced static `TILE * cols` chunking with dynamic task sizing:

    let total_tile_rows = rows.div_ceil(TILE);
    let tile_rows_per_task = total_tile_rows.div_ceil(num_threads).max(1);

   This limits scheduling overhead to at most one coherent chunk per thread rather than partitioning the buffer into dozens of 24 KB micro-tasks.
3. **Stride-Aware Loop Nesting**: Mirrored the serial path's `row_stride <= col_stride` check inside the parallel worker so the innermost loop consistently advances along the smaller/unit-stride dimension for both row-major and column-major sources.
4. **Regression Unit Tests**: Added `test_to_contiguous_2d_parallel_row_stride_le_col_stride` and `test_to_contiguous_2d_parallel_row_stride_gt_col_stride` to guard both stride orderings under the parallel path.

Regarding `try_broadcast_cmp_f32`: given that Pattern 2's fallback delta was ~4% on a 4 µs micro-op, consolidating onto `try_zip_cmp_f32` feels like the right maintainability trade-off, though I can restore Pattern 2 as a fast path if you prefer.

All 410 crate unit tests, backend test suites, and Clippy lints are passing cleanly.

crates/burn-flex/src/layout.rs
```diff
@@ -101,16 +101,26 @@ impl Layout {
...
             return true;
         }
 
+        if self.strides.iter().any(|&s| s <= 0) {
+            return false;
+        }
+
         let mut expected_stride = 1isize;
         for i in (0..self.shape.num_dims()).rev() {
+            if self.shape[i] == 1 {
```
antimora, line 114: This is correct in isolation. A size-1 dim always contributes index 0, so its stride never gets stepped.

The problem is that `reduce_dim_f32` uses `is_contiguous()` as a proxy for "strides are canonical", and this makes that proxy wrong. `reduce.rs:864` only normalizes when `!tensor.is_contiguous()`, and `reduce.rs:1122` then derives the row pitch from `strides[dim - 1]`. When a size-1 dim sits next to the reduced dim, that stride is meaningless and rows get read as a sliding window.

Verified on `1c2ee37e3` against an independently constructed contiguous reference (not `to_contiguous()`, which is a no-op here for exactly this reason):

```
[3,4,1].swap_dims(1,2) -> shape [3,1,4] strides [4,1,1], is_contiguous() == true
  logical contents read back correctly: [0,1,2,...,11]
  sum_dim(2)   got [6, 10, 14]       want [6, 22, 38]
  mean_dim(2)  got [1.5, 2.5, 3.5]   want [1.5, 5.5, 9.5]
  prod_dim(2)  got [0, 24, 120]      want [0, 840, 7920]

[1,3,4].permute([1,0,2]).sum_dim(2)  panics out of bounds at reduce.rs:1144
```

The same test passes on `5ecfb3896`, so this is a regression from the PR. f32 only, since f64/f16/bf16 go through `reduce_dim_generic` which uses flat index arithmetic and is fine. But f32 is the default dtype and `swap_dims` / `permute` across a size-1 axis is everyday code, so this should not merge as is.

I would fix the consumer rather than back this change out, since the new answer is the right answer. Either compute the outer stride from the shape (`shape[dim+1..].iter().product()`), or make `needs_copy` test stride canonicality instead of contiguity. A `debug_assert_eq!(outer_stride, (dim_size * inner_size) as isize)` in `reduce_last_dim_f32` would stop it recurring.

crates/burn-flex/src/layout.rs
```diff
@@ -101,16 +101,26 @@ impl Layout {
 
     /// Check if this layout is contiguous (row-major, positive strides).
     pub fn is_contiguous(&self) -> bool {
-        if self.shape.num_dims() == 0 {
+        if self.shape.num_dims() == 0 || self.num_elements() == 0 {
             return true;
         }
 
+        if self.strides.iter().any(|&s| s <= 0) {
```
antimora, line 108: This gate runs before the size-1 skip below and partly cancels it. `ops/expand.rs:107` pushes stride 0 for every prepended dim regardless of its size, so the layouts that most commonly carry a size-1 dim never reach line 114:

```
[1,2,3] strides [0,3,1]  (expand-prepended)      is_contiguous = false   <- logically contiguous
[1,4]   strides [1,1]    (from [4,1] transpose)  is_contiguous = true
[2,1,3] strides [3,3,1]  (plain contiguous)      is_contiguous = true
```

So the squeeze only pays off for the positive-stride transpose case you tested. Relaxing this to `s < 0` would extend it to expand-produced layouts, but I would hold off until the reduce issue on line 114 is resolved, since it widens the set of layouts that report contiguous.

crates/burn-flex/src/tensor.rs
```diff
@@ -642,12 +785,16 @@ fn copy_inner_contiguous_run<E: Copy>(
     }
 }
 
+/// Minimum number of elements required to trigger parallel 2D tiled copy.
+/// Avoids thread fork/steal overhead on ~1MB cache-resident tensors.
+pub(crate) const COPY_2D_PARALLEL_THRESHOLD: usize = 1_048_576;
```
antimora, line 793: Confirmed this does the job. Base vs `1c2ee37e3`, 3 interleaved passes, all 22 suites, min-of-medians:

| bench | 5ecfb3896 | 1c2ee37e3 |
|---|---|---|
| c384_30x28 | 441-453 us | 441-461 us |
| c192_61x56 | 925-967 us | 927-948 us |
| c96_122x112 | 1858-1901 us | 1733-1776 us |

Ranges overlap now instead of separating cleanly, and `c96` came out faster. The wins all held: int matmul 8.1x, `conv_transpose2d_8x64x14x14` 5.6x, `non_last_axis_f32/trait_hook_dim1_of_3` 5.7x. Nothing new separated as slower beyond the `greater_broadcast` case already noted.

crates/burn-flex/src/tensor.rs
```diff
@@ -558,6 +578,129 @@ fn collapse_for_copy(shape: &[usize], strides: &[isize]) -> CollapsedLayout {
...
+    let rank = shape.len();
+    let rows = shape[rank - 2];
+    let cols = shape[rank - 1];
+    let r_st = strides[rank - 2];
+    let c_st = strides[rank - 1];
+    let slice_len = rows * cols;
+
+    let batch_shape = &shape[..rank - 2];
+    let batch_strides = &strides[..rank - 2];
+    let batch_count: usize = batch_shape.iter().product();
```
antimora: `batch_count` is only read inside the `#[cfg(feature = "rayon")]` block below, so this warns under `cargo check -p burn-flex --no-default-features --features std`. Same for `n` at 649 and 807, `total_groups` at `conv_transpose.rs:301`, and `out_dtype` at `mask.rs:230` and `:564`. Six new warnings in a config that was clean on `5ecfb3896`.

Worth fixing because that is the config CI actually builds. `Cargo.toml:202` declares `burn-flex` with `default-features = false`, and both the xtask path and the `test-flex` / `bench-flex` aliases pass `--no-default-features --features flex,std`. Moving each binding inside its cfg block clears the warning and skips the work when rayon is off.

The same fact has a larger consequence: none of the new rayon code is compiled by the backend test suite, and `cargo bench-flex` would not have reproduced the copy regression either, since it disables rayon. Enabling `flex-rayon` in one CI job would be worth doing separately.

crates/burn-flex/src/ops/mask.rs
```diff
@@ -6,6 +6,17 @@ use burn_std::{Bytes, bf16, f16};
 
 use crate::{FlexTensor, Layout};
 
+/// Check if a layout is contiguous and not broadcasted.
+#[inline]
+fn is_contiguous_non_broadcast(layout: &Layout) -> bool {
```
antimora: This is duplicated verbatim at `ops/gather_scatter.rs:18`. Both copies gate `storage_mut()`, so a drift between them is a bad write rather than a compile error.

It is also now equivalent to `layout.is_contiguous()` for any non-empty layout: the `s <= 0` check added at `layout.rs:108` already rejects every zero stride, so the second conjunct can only change the answer when `num_elements() == 0`, where it is vacuous either way.

Reads better as a single method on `Layout` carrying the actual contract, something like "storage[start_offset..start_offset + num_elements()] is exactly this tensor, one slot per element, so it is safe to overwrite given unique ownership".

crates/burn-flex/src/ops/gather_scatter.rs
```diff
@@ -12,6 +12,20 @@ use rayon::prelude::*;
 
 use crate::{FlexTensor, Layout};
 
+#[cfg(feature = "rayon")]
+const PARALLEL_THRESHOLD: usize = 256 * 1024;
```
antimora: This shadows `ops/mod.rs:19`, which is already `pub(crate)` with the same value and a comment explaining the choice. `ops/sort.rs:18` just does `use super::PARALLEL_THRESHOLD;`.

Retuning the shared constant would silently leave gather, select and scatter on the old value with no compile error and no test failure, and the only symptom would be a benchmark drifting months later. If these paths want a different fan-out point, `reduce.rs:1768`'s `EXTREMUM_PARALLEL_THRESHOLD` is the pattern to follow: different name, different value, comment saying why.

crates/burn-flex/src/zip.rs
```diff
@@ -320,6 +302,270 @@ where
...
+            self.c_strides[d],
+        )
+    }
+
+    pub fn a_is_dense(&self) -> bool {
+        Self::is_dense(&self.a_strides[..self.ndim], &self.shape[..self.ndim])
+    }
+
+    #[allow(dead_code)]
+    pub fn b_is_dense(&self) -> bool {
```
antimora: `b_is_dense` and `c_is_dense` are both `#[allow(dead_code)]` with no callers. Worth dropping until something needs them.

Stepping back, `Zip3Nest` is `ZipNest` with a third stride array threaded through every line, and the two have already drifted inside this PR: `ZipNest::for_each_run` was rewritten to the `while d > 0` form while `Zip3Nest::for_each_run` kept the old `loop`, and `zip_apply_inplace` has a `debug_assert_eq!(l_st, 1, ...)` that `zip3_apply_inplace` does not. A `Nest<const N: usize>` would cover both, and `CollapsedLayout` at N=1, with identical codegen since it monomorphizes. Not blocking, but the drift is already real rather than hypothetical.

crates/burn-flex/src/ops/float.rs
```diff
@@ -830,49 +830,59 @@ impl FloatTensorOps<Flex> for Flex {
...
-            DType::BF16 => {
-                let src: &[bf16] = tensor.storage();
-                src.iter().map(|&v| v.to_f32() as f64).collect()
-            }
-            _ => panic!("float_cast: unsupported source dtype {:?}", src_dtype),
-        };
+        fn cast_slice<Src: Element + bytemuck::Pod, Dst: Element + bytemuck::Pod>(
+            src: &[Src],
+        ) -> Vec<Dst> {
+            src.iter().map(|&v| Dst::from_elem(v)).collect()
```
antimora, line 859: For f32 and f64 targets this is equivalent to the old path, but f64 -> f16 and f64 -> bf16 now round twice. `ToElement::to_f16`'s default is `f16::from_f32(self.to_f32())` (`burn-std/src/element/cast.rs:94`) and f64 does not override it, where the old code called `f16::from_f64(v)` directly. So the change is genuinely direct for the f32/f64 pairs but not for the half targets.

There is no cast test for any half target in the suite, so a 1 ULP shift would go unnoticed. Either special-case the f64 source or add a line saying the double rounding is accepted.

crates/burn-flex/src/ops/mask.rs
```diff
@@ -259,60 +519,160 @@ pub fn mask_where_bf16(tensor: FlexTensor, mask: FlexTensor, value: FlexTensor)
...
     {
         mask_where::<i64>(tensor, mask, value)
     }
 }
 
-/// Mask where for bool tensors (SIMD-accelerated).
+/// Mask where for bool tensors (SIMD-accelerated when out-of-place contiguous, in-place when unique).
 pub fn mask_where_bool(tensor: FlexTensor, mask: FlexTensor, value: FlexTensor) -> FlexTensor {
-    // Preserve the input tensor's bool dtype for the output.
     let out_dtype = burn_std::BoolDType::from(tensor.dtype());
```
antimora: `out_dtype` is computed from `tensor`, but the in-place path added below returns `mask_where::<u8>(...)`, which can return `value` carrying `value.dtype()`. Previously every path went through `make_bool_tensor(..., out_dtype)`. With a `Bool(Native)` tensor and a `Bool(U8)` value, the output store would follow whichever operand happened to be uniquely owned, which depends on refcounts.

I did not execute this one, so it may be unreachable in practice, but the comment you removed just above used to say the input dtype was deliberately preserved. `mask_fill_bool` is unaffected since its in-place arm returns `tensor`.

`out_dtype` is also unused in the `not(simd)` build, which is the one CI compiles.

antimora (changes requested):
Ran a full review over `1c2ee37e3`, re-verifying everything after your two new commits rather than replaying the earlier findings.

The threshold / chunk-sizing / loop-nesting fix works. Re-benchmarked base vs `1c2ee37e3`, 3 interleaved passes across all 22 suites: the copy regression is gone and the wins held. Numbers inline on `tensor.rs`. The empty-view panic is fixed too - `swap_dims(1,2)` plus an empty slice plus `to_contiguous()` passes now, and so does the negative-step-at-end-of-buffer case. Mirroring the stride check into the parallel branch was the right call.

One blocking issue is left, and it predates those two commits: the size-1 skip in `Layout::is_contiguous` makes f32 `sum_dim` / `mean_dim` / `prod_dim` return wrong numbers on swapped or permuted layouts. It is silent, it is the default dtype, and the backend suite does not catch it. Details on `layout.rs:114`.

Verification I ran on `1c2ee37e3`: `cargo test -p burn-flex` 410 pass, `cargo clippy -p burn-flex --all-targets -- -D warnings` clean, `cargo test -p burn-backend-tests --features flex` 1185 pass, and `--no-default-features` with `std`, `std,simd` and `std,rayon` all pass. The rest of the inline comments are smaller.

Worth adding as regression tests once the reduce fix lands:

```
[3,4,1].swap_dims(1,2).sum_dim(2)
[1,3,4].permute([1,0,2]).sum_dim(2)
[3,2,4,5].transpose(1,2).slice(.., .., .., 2..2).to_contiguous()
[120].narrow(0,120,0).slice(&[Slice::new(0, None, -1)]).to_contiguous()
```

The last two pass now. Pinning them would keep the fixes from regressing.

Sadik00789:
Addressed all feedback items from the review in the latest commit:

1. **Reduce Dim Stride Derivation (`reduce.rs`)**:
   - In `reduce_dim_f32` and `reduce_last_dim_f32`, stopped deriving row pitch from `strides[dim - 1]`. The contiguous fast path now computes `outer_stride` directly from the tensor shape as `(dim_size * inner_size) as isize` (or `dim_size as isize` for the trailing dim), paired with `debug_assert_eq!` guards.
   - Swapped and permuted singleton layouts (e.g. `[3, 4, 1].swap_dims(1, 2)`) now compute correct reductions without out-of-bounds reads.
   - Added regression unit tests `test_sum_dim_swap_dims_size_one` and `test_sum_dim_permute_size_one`.

2. **Deduplication & Layout Storage Contract (`layout.rs`, `gather_scatter.rs`, `mask.rs`)**:
   - Consolidated `is_contiguous_non_broadcast` into `Layout::is_dense_unique_storage(&self) -> bool`, capturing the exact contract that the underlying buffer slice represents unique, non-broadcasted elements from `start_offset`.
   - Replaced duplicate helper functions across `mask.rs` and `gather_scatter.rs`.
   - Replaced shadowed `PARALLEL_THRESHOLD` in `gather_scatter.rs` with `use super::PARALLEL_THRESHOLD;`.

3. **Precision & Dead Code (`float.rs`, `zip.rs`)**:
   - Updated `cast_slice` to convert directly from `f64` using `f16::from_f64` and `bf16::from_f64`, preventing the 1 ULP shift caused by intermediate double-rounding through `f32`.
   - Pruned unused `b_is_dense` and `c_is_dense` methods from `zip.rs`.

4. **CFG Hygiene & Compiler Warnings**:
   - Gated `batch_count`, `total_groups`, and parallel branches behind `#[cfg(feature = "rayon")]`.
   - Gated `COPY_2D_PARALLEL_THRESHOLD` with `#[cfg(any(feature = "rayon", test))]`.
   - Gated `out_dtype` in `mask.rs` behind `#[cfg(feature = "simd")]`.
   - Verified that `cargo check -p burn-flex --no-default-features --features std` and `cargo clippy -p burn-flex --no-default-features --features std --all-targets -- -D warnings` compile completely clean with 0 warnings.

All 412 unit tests on the default feature set and all 374 unit tests on `--no-default-features --features std` pass cleanly.

antimora (approved):
Approving. Re-verified everything on `cac6c8e40` by running it rather than reading it.

The reduction fix is correct. `sum_dim`, `mean_dim` and `prod_dim` on `[3,4,1].swap_dims(1,2)` now match an independently constructed contiguous reference, the `permute([1,0,2])` out-of-bounds is gone, and the middle-dim and batched sites the fix also touched are right. Deriving the outer stride from the shape when the layout is contiguous was the right call, and it keeps the `is_contiguous()` change itself intact.

f64 casts are single-rounded now. I built four values sitting just above an f16 tie, where `f16::from_f32(v as f32)` and `f16::from_f64(v)` genuinely disagree, and the cast matches `from_f64` on all four.

The rest, all on `cac6c8e40`: 412 unit tests, 1185 and 1640 in `burn-backend-tests --features flex`, `--no-default-features` with `std`, `std,simd` and `std,rayon` all green, clippy clean, and the six cfg warnings are down to zero.

One caveat worth stating plainly: the benchmark sweep that confirmed the copy regression was cured ran on `1c2ee37e3`, not on this head. `reduce_dim_f32` picked up an extra `is_contiguous()` call since then. That is O(ndims) against a full reduction so I do not expect it to show, but I have not measured it.

Two follow-ups, neither blocking.

The new `debug_assert_eq!` cannot fire. It asserts the value assigned in the same branch:

```rust
let outer_stride = if tensor.is_contiguous() { dim_size as isize } else { ... };
if tensor.is_contiguous() { debug_assert_eq!(outer_stride, dim_size as isize); }
```

Asserting the shape-derived stride against the actual `strides[dim - 1]` is what would catch this class of bug coming back.

None of the new parallel paths have a threshold-straddling test, and CI never compiles them anyway: `test-flex` and `bench-flex` both pass `--no-default-features --features flex,std`, so rayon is off in the suite that would exercise them. `ops/sort.rs:571-712` has the pattern for the tests. Worth a separate issue rather than more scope here.

Thanks for turning these around so quickly.

laggui (commented):
Manually addressed some edge cases for reductions, empty transposed conv outputs and bool dtype preservation in masking.

PR #5618, opened 2026-09-08
fix(optim): keep a parameter's checkpointing strategy across an update

Marc-AnthonyG:
### Checklist

### Related Issues/PRs

None.

### Changes

The module optimizer takes a parameter off the tape to update it and lifts the result back with `Tensor::from_inner`, which assigns the default gradient-checkpointing strategy. A model trained under balanced checkpointing therefore came out of its first optimizer step with every updated parameter on the default strategy while its inputs and any frozen parameter kept the balanced one, and the next forward refused to merge the two. Training with gradient checkpointing could not get past its first step. L-BFGS lifts its parameters back the same way when it unflattens them and had the same problem.

Both now lift through `from_inner_with_strategy`, which restores the strategy the parameter carried before the update.

### Testing

`step_keeps_the_gradient_checkpointing_strategy` (`crates/burn-optim/src/optim/module/module_optimizer.rs`) steps a model on a device with balanced checkpointing, checks the updated weight's strategy against the device's, and steps it again, pinning both the fix and the second step no longer panicking.

### Reproduction

https://github.com/Marc-AnthonyG/burn-reproduction/tree/main/cases/optimizer-keeps-checkpointing-strategy

Run with `cargo run --release -p optimizer-keeps-checkpointing-strategy --features cuda`. It links burn from this branch and from `main` in the same process and runs two optimizer steps on a model trained under balanced gradient checkpointing against each tree, reporting whether a step panics and what checkpointing strategy the updated weight carries afterward. Expect the `before` tree to panic on the second step and the `after` tree to complete both steps, reporting the balanced strategy throughout.


crates/burn-optim/src/optim/module/lift.rs
```diff
@@ -0,0 +1,20 @@
...
+pub fn from_inner_with_strategy<const D: usize>(
+    inner: Tensor<D>,
+    strategy: Option<GradientCheckpointingStrategy>,
+) -> Tensor<D> {
+    let tensor = Tensor::from_inner(inner);
+    match strategy {
+        Some(strategy) => tensor.with_gradient_checkpointing_strategy(strategy),
+        None => tensor,
+    }
+}
```
laggui: I wouldn't make this public. For now this is an optimizer-internal reconstruction public, not sure it requires a user-facing API.

crates/burn-optim/src/optim/module/module_optimizer.rs
```diff
@@ -421,6 +421,7 @@ impl ModuleMapper for ModuleOptimizerMapper<'_> {
 
         let tensor = if let Some((grad, device)) = grad {
             let is_require_grad = tensor.is_require_grad();
+            let checkpointing = tensor.gradient_checkpointing_strategy();
             #[cfg(feature = "std")]
             let is_distributed = tensor.is_distributed();
```
laggui: Suggestion: replace `from_inner_with_strategy` with a capture/restore for parameter context.

e.g.
```rust
pub(crate) struct ParameterContext {
    require_grad: bool,
    #[cfg(feature = "std")]
    distributed: bool,
    checkpointing: Option<GradientCheckpointingStrategy>,
}

let context = ParameterContext::capture(&tensor);
// operations on `tensor.inner()` ...
let tensor = context.restore(inner, id);
```

laggui (commented):
The fix LGTM! But I have some minor changes to suggest

laggui (approved):

PR #5620, opened 2026-09-08
perf(fusion): let a padded tensor vote for the block's layout

Marc-AnthonyG:
### Checklist

### Changes

A fused block picks the dimension order of its output by a byte-weighted vote among its inputs, so that a block fed a channels-last tensor iterates channels-last and reads it linearly instead of strided. The vote asked each input for its `dim_order`, which requires the tensor to fill its buffer exactly. A tile-aligned or pitched allocator does not do that: it leaves a gap under the innermost dimension, so 48 elements held innermost on a 64-element tile carry stride 64 where a dense tensor would carry 48.

Such a tensor is not dense, so it could not vote. With no voters the block kept a contiguous reference, wrote its output vectorized and read its large channels-last input one element at a time. Nothing failed; the layout propagation simply never fired on a backend whose allocator pitches rows, for every channel count that is not tile-aligned. A backend with dense allocations took the fast path all along, which is why the gap only showed up as a difference between backends.

`nested_dim_order` accepts a gap (padding) and still rejects an overlap, and the vote now uses it. That is enough because the vote only decides the order the output is allocated in, and the output is allocated dense in that order either way; a padded input whose strides then differ from the reference is read through the strided path, since `SameAsRef` compares strides exactly. `dim_order` is unchanged wherever a buffer is reinterpreted as a flat run of elements.

Measured on CUDA (one L4), warm, median of ten: an elementwise normalization plus activation over a channels-last view of a [4, 48, 384, 384] tensor went from 6.28x the contiguous case to 1.10x; a convolutional encoder pass 1.58x faster; a full forward pass 1.6x faster; a training step 1.11x faster.

### Testing

Six unit tests cover the new `nested_dim_order` behaviour in `burn-cubecl-fusion`'s layout helper: `a_dense_tensor_nests_in_the_order_it_is_dense_in`, `padding_under_the_innermost_dimension_keeps_the_nhwc_order`, `padding_above_the_innermost_dimension_is_nesting_too`, `overlapping_dimensions_are_not_an_order_under_either`, `a_broadcast_dimension_still_cannot_vote`, and `size_one_dimensions_neither_pad_nor_constrain_nesting`.

### Reproduction

https://github.com/Marc-AnthonyG/burn-reproduction/tree/main/cases/padded-tensors-vote-layout

Run with `cargo run --release -p padded-tensors-vote-layout --features cuda`,

### DIFF
--- a/crates/burn-backend-tests/tests/fusion/mod.rs
+++ b/crates/burn-backend-tests/tests/fusion/mod.rs
@@ -10,6 +10,8 @@ mod fusion_shape;
 mod inplace;
 mod int_bitwise;
 mod nhwc_relayout;
+#[cfg(feature = "cube")]
+mod padded_layout;
 mod reduce_broadcasted;
 mod reduce_logical;
 
--- a/crates/burn-cubecl-fusion/src/engine/launch/layout.rs
+++ b/crates/burn-cubecl-fusion/src/engine/launch/layout.rs
@@ -36,6 +36,34 @@ pub type DimOrder = Shape;
 /// arbitrary and carries no traffic — but they keep a position in the returned
 /// order so it stays a permutation of `0..rank`.
 pub fn dim_order(shape: &[usize], strides: &[usize]) -> Option<DimOrder> {
+    dim_order_inner(shape, strides, Padding::Rejected)
+}
+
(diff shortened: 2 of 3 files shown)


laggui (approved):
LGTM! Just fixed the wording in some of the docs and added tests.

PR #5621, opened 2026-09-08
perf(cubecl): scatter-add runs one unit per value with atomic adds

Marc-AnthonyG:
### Changes

The scatter kernel ran one unit per output row and walked the scatter axis serially. That leaves the device nearly idle whenever the values far outnumber the output rows: a histogram of a million values into a few thousand bins ran on a thousand units each looping over every value, and cost the same whether the batch was 32 or 128.

For `Add`, on a type the device adds atomically, a second kernel now runs one unit per value and applies each update with an atomic add. The serial kernel stays for `Assign`, `Mul` and `Or`, whose order matters or which have no atomic form, and for types without atomic support. On floats the order the adds land in is not fixed, so a sum's rounding can differ between runs where the serial kernel was deterministic.

Measured on an L4 (CUDA), a 1-D convolution bank whose counting scatters 1.6 million values into 8,192 slots sixteen times a pass, same session, medians of 20 with the first pass of each binary discarded: the training step went from 95.9 to 81.2 ms, repeated 81.5 ms (-15%). In isolation the two scatters of one dilation went from 1.45 ms to about 0.3 ms.

### Testing

`should_scatter_add_many_values_into_few_slots` (`tests/tensor/float/ops/gather_scatter.rs`) scatters two thousand ones into eight slots, so every slot is hit hundreds of times, and checks the counts come out exact.

### Reproduction

https://github.com/Marc-AnthonyG/burn-reproduction/tree/main/cases/scatter-add-atomic

Run with `cargo run --release -p scatter-add-atomic --features cuda`.

##Metal
Screenshot 2026-09-08 at 2 37 53 PM

## CUDA
Screenshot 2026-09-08 at 3 21 29 PM

### DIFF
--- a/crates/burn-backend-tests/tests/cubecl/scatter.rs
+++ b/crates/burn-backend-tests/tests/cubecl/scatter.rs
@@ -3,6 +3,25 @@ use burn_tensor::Distribution;
 use burn_tensor::{Device, IndexingUpdateOp, Tolerance};
 
 #[test]
+fn scatter_add_handles_nan_and_infinity() {
+    let device = Device::default();
+    let tensor = TestTensor::<1>::from_data([f32::NAN, f32::INFINITY, 0.0], &device);
+    let values = TestTensor::<1>::from_data([1.0, f32::NEG_INFINITY, f32::NAN], &device);
+    let indices = TestTensorInt::from_ints([0, 1, 2], &device);
+    let actual = tensor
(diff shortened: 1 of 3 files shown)


laggui (approved):
LGTM. I just added a condition so CPU still uses the existing kernel given that it might not benefit from the extra parallelism.

PR #5622, opened 2026-09-08
perf(fusion): settle only the unfusable head of a block

Marc-AnthonyG:
### Changes

When a dependent block is committed and no fuser has a fusion worth executing, `Block::optimize` returned the whole block as unfused operations. A block is a dependency chain, so everything after the operation that closed the fusers ran one kernel each, although a fresh fuser would have fused it: those operations only sat in that block because they depend on its head. The common heads are a lone elementwise op followed by a view of its own output, a convolution backward, or a reduction (a single fused op scores zero), which is what a backward pass is made of. On a convolutional network's training step the fusion log showed 3,588 one-op segments per step and 3,238 elementwise ops executing unfused, 2,504 of them in runs of two or more consecutive fusable ops.

Now only the head is settled: the operations up to and including the one that closed the last fuser, plus whatever follows that no fresh fuser would start on, probed by feeding it to a clone of each pristine fuser. The rest goes back to the search, either as a hole re-optimized in the same plan or as the next segment, exactly as the tail of a block that did have a fusion already did. The stream optimizer passes its pristine fusers down for the probe, and each block records the registration index at which every fuser closed.

The operation that closed the last fuser is always settled with the head, even when a fresh fuser would start on it, so a block that closes on its second operation keeps executing as one segment, as before. Leaving that operation to the search would win a fusion in some cases at the price of more segments; that is a separate decision, not taken here.

### Testing

Three tests cover the behaviour: `unfusable_head_leaves_the_tail_to_the_next_search` (an unfusable head leaves its fusable tail to the next search instead of running it unfused), `unfusable_head_tail_is_fused_as_a_hole` (that tail is fused as a hole once an independent block follows it), and `op_that_closed_the_last_builder_is_settled_with_the_head` (the operation that closed the last fuser stays with the head even when a fresh fuser would take it). One existing expectation was updated: an in-block reorder is still refused, but the second operation is no longer dragged along unfused with it.

### Reproduction

https://github.com/Marc-AnthonyG/burn-reproduction/tree/main/cases/fusion-settle-unfusable-head

Run with `cargo run --release -p fusion-settle-unfusable-head --features cuda`.

### DIFF
--- a/crates/burn-fusion/src/search/block.rs
+++ b/crates/burn-fusion/src/search/block.rs
@@ -29,6 +29,9 @@ pub struct Block<O> {
     /// reason about the original dependency graph through the recursive merging.
     constituents: SubGraph,
     ordering: Vec<usize>,
+    /// For each builder, the index into `operations` of the operation on which it reported
+    /// [Closed](FuserStatus::Closed); `None` while it is still open.
+    closed_at: Vec<Option<usize>>,
     /// The start position in the relative execution stream.
     pub start_pos: usize,
     /// The end position in the relative execution stream.
@@ -139,6 +142,7 @@ impl<O: NumOperations> Block<O> {
             freed: HashSet::new(),
             constituents: SubGraph::empty(),
             ordering: Vec::new(),
+            closed_at: vec![None; builders.len()],
             start_pos: usize::MAX,
             end_pos: usize::MIN,
         }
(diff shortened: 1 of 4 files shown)


laggui (approved):
Looks good.

Settling only the unfusable prefix gives the tail another chance to fuse and fits the existing tail and hole handling.

PR #5623, opened 2026-09-08
perf(autodiff): reduce a broadcast gradient over all dims at once

Marc-AnthonyG:
### Changes

Adds `FloatTensorOps::float_sum_dims` to sum several dimensions while keeping each reduced dimension at length one. The default implementation reduces one dimension at a time, preserving compatibility with existing backends.

Broadcast-gradient reduction and `Tensor::sum_dims` now use this operation. Fusion and Router preserve multi-axis requests as `NumericOperationIr::SumDims`, allowing the underlying backend to optimize them together. Fusion forwards single-axis requests to the existing `SumDim` path.

CubeCL folds reduced dimensions that are densely adjacent in memory into one axis using metadata changes, then reduces that axis. Each pass selects the largest foldable group and produces a contiguous output before processing the remaining dimensions.

For dense channels-last activations, batch and spatial dimensions can fold into one reduction. For contiguous NCHW tensors, height and width fold together, followed by batch reduction. Other layouts may require additional passes, up to one per requested dimension. Singleton dimensions do not split foldable groups, and empty reduction axes retain the existing sum-identity behavior.

Multi-axis reduction fusion with surrounding elementwise operations is deferred to a separate PR.

Measured on CUDA (one L4), same session, warm, first pass discarded, median of twenty, on a convolutional classifier with channels-last activations: a training step went from 369.7 ms to 362.0 ms; reduce launches per step fell by 40% and the time spent in them by 30% (115 ms to 80 ms), most of which had been overlapped with other work. The forward pass, which does not reach this path, was unchanged.

### Testing

Nine tests compare multi-axis sums with sequential reductions or explicit expected values, covering:

- Channels-last and contiguous layouts.
- Nonadjacent reduction axes.
- Single-axis and all-axis reductions.
- Reduced and unreduced zero-length dimensions.
- Singleton dimensions between foldable axes.
- Broadcast dimensions with repeated values.

### Reproduction

https://github.com/Marc-AnthonyG/burn-reproduction/tree/main/cases/sum-dims-single-reduction

Run with `cargo run --release -p sum-dims-single-reduction --features cuda`.

# METAL
Screenshot 2026-09-08 at 2 34 37 PM

# CUDA
Screenshot 2026-09-08 at 3 20 44 PM

### DIFF
--- a/crates/burn-cubecl/src/kernel/reduce/base.rs
+++ b/crates/burn-cubecl/src/kernel/reduce/base.rs
@@ -7,9 +7,8 @@ use crate::{
 };
 use burn_backend::cubecl::{dtype_to_elem_type, dtype_to_storage_type, elem_type_to_dtype};
 use burn_backend::{DType, TensorMetadata};
-use burn_std::Shape;
-use burn_std::tensor::{ReshapeAction, reshape_action};
 use burn_std::{BoolDType, Metadata};
+use burn_std::{Shape, Strides};
 use cubecl::{AutotuneKey, client::Client, features::AtomicUsage, ir::Type, prelude::InputScalar};
 use cubek::reduce::{
     ReduceDtypes, ReduceError, ReduceStrategy, ReduceWithIndicesDtypes,
@@ -194,13 +193,14 @@ pub fn reduce(
 /// a stride change alone, and reduced in a single launch — for a channels-last
 /// tensor that is every non-channel dimension at once.
 ///
-/// The folded axis is put *first* and the kept dimensions after it, so the
-/// kept innermost dimension stays innermost in memory: the reduction then
-/// coalesces across it and writes its output vectorized along it, instead of
-/// walking a strided last axis one element at a time. Where the reduced
-/// dimensions do not all sit together, the outermost is reduced on its own
-/// first; its output is contiguous, so whatever is left then folds, and no
-/// layout takes more than two launches.
+/// Only a run that memory already holds together folds, so each pass takes the
+/// largest such run and reduces that, largest first because it is the pass that
+/// leaves the least behind. The folded axis is presented *first* and the
+/// remaining dimensions after it in logical order, which puts the reduction's
+/// output back in logical order and contiguous, so the next pass starts from a
+/// tensor whose memory order is its logical one. A layout holding every reduced
+/// dimension together therefore takes a single launch, and one that scatters them
+/// takes no more launches than reducing them one at a time would have.
 pub fn reduce_dims(
     input: CubeTensor,
     output_dtype: Option<DType>,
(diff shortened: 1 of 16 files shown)


crates/burn-cubecl/src/kernel/reduce/base.rs
```diff
@@ -182,6 +184,132 @@ pub fn reduce(
...
+    let mut shape = input.meta.shape().clone();
+    let reduced: Vec<usize> = (0..rank).filter(|dim| dims.contains(dim)).collect();
+
+    // A dimension already of length one is reduced by being left alone, and its
+    // stride is arbitrary, so keeping it among the dimensions to fold would let
+    // it break a run of dimensions that do fold.
+    let mut left: Vec<usize> = reduced
+        .iter()
+        .copied()
+        .filter(|dim| shape[*dim] > 1)
```
laggui, line 227: This drops zero-length axes as well as singleton axes. `sum_dims(&[0, 1])` on a tensor with shape `[0, 3]` should produce `[1, 1]` containing zero, consistent with the existing behavior and tests. We need to handle empty axes correctly.

crates/burn-cubecl/src/kernel/reduce/base.rs
```diff
@@ -182,6 +184,132 @@ pub fn reduce(
...
+
+    let elements = |run: &[usize]| run.iter().map(|dim| shape[*dim]).product::<usize>();
+    let mut largest: Vec<usize> = Vec::new();
+    let mut run: Vec<usize> = Vec::new();
+
+    for dim in memory_order {
+        if !left.contains(&dim) {
+            run.clear();
+            continue;
+        }
```
laggui, line 293: Singleton dimensions remain in `memory_order` and clear the run here. We should skip them during this scan.

crates/burn-fusion/src/ops/tensor.rs
```diff
@@ -1451,6 +1496,33 @@ impl<B: FusionBackend> FloatTensorOps<Self> for Fusion<B> {
...
+        let desc = ReduceDimsOpIr::create(tensor.into_ir(), dims.to_vec(), || {
+            client.create_empty_handle()
+        });
+
+        client
+            .register(
+                streams,
+                OperationIr::NumericFloat(
+                    desc.out.dtype,
+                    NumericOperationIr::SumDims(desc.clone()),
```
laggui, line 1480: `sum_dims(&[axis])` now emits `SumDims` (which is technically correct), but the reduction fuser in burn-cubecl-fusion doesn't recognize the op, losing existing fusion opportunities. When only a single axis is provided with `sum_dims` we could forward to `float_sum_dim` for now. Adding support for `SumDims` in burn-cubecl-fusion would be even better, but not strictly required.

crates/burn-router/src/interpreter.rs
```diff
@@ -829,6 +829,11 @@ impl<B: BackendIr> TensorInterpreter<B> {
                         tensor, axis
                     ))
                 }
+                NumericOperationIr::SumDims(desc) => {
+                    let input = handles.get_float_tensor::<B>(&desc.input);
+                    let output = B::float_sum_dims(input, &desc.axes);
+                    handles.register_float_tensor::<B>(&desc.out.id, output);
+                }
```
laggui, line 836: `BackendRouter` currently still inherits the default `float_sum_dims` fallback, so normal calls never emit `SumDims`. We need to add the override in the `FloatTensorOps` implementation.

laggui (changes requested):
LGTM overall, just a couple of things missing

/edit: oh, and we should fix the no-std issue (using alloc vec)

crates/burn-fusion/src/ops/tensor.rs
Marc-AnthonyG: I have this pr started for that: https://github.com/Marc-AnthonyG/burn/pull/7. However in my bench project it seems like the change is a regression in term of performance. Feel free to take a look if you want, but wasn't planning on pushing it.

laggui (approved):

PR #5625, opened 2026-09-08
perf(cubecl): walk elementwise kernels in operands' memory order

Marc-AnthonyG:
### Changes

The elementwise kernels take every operand through a linear view that walks the logical dimension order, and allocate their output contiguous in that order. A convolution hands its consumer a channels-last tensor presented in NCHW, so any elementwise operation that does not fuse (a gradient's `grad * x` in a backward pass, a scalar multiply, an activation) walked it strided, and because the logical last dimension is not the contiguous one it could not vectorize either. Now that the fusion planner (the previous PR in this stack) keeps such tensors channels-last, every unfused consumer pays this cost on every activation.

An elementwise operation does not care what a dimension means, only that its operands are walked in step. `in_memory_order` hands the operands to the kernel permuted into the order the bulk of them occupy memory in, a byte-weighted vote among the operands of the output's shape, with ties going to the logical order and padding under a dimension tolerated, and permutes the result back. The output is then allocated dense in that order: an operand already in it is walked linearly and vectorized, one that disagrees is walked strided exactly as before. A permute is a change of metadata, so no data moves, and a quantized operand is left alone. This applies to the float and numeric binary, scalar and unary launchers.

The stack's first commit moves the `dim_order`, `nested_dim_order` and `is_contiguous_order` helpers from `burn-cubecl-fusion` to `burn-std`, since these kernels need to ask the same question about a tensor's strides and `burn-cubecl` cannot depend on the fusion crate; the fusion planner re-exports them from the same path it always did, with no behaviour change.

### Testing

Six tests compare each kernel on a channels-last operand against the same kernel on a contiguous copy, on the plain backend and under fusion: `binary_with_both_operands_alive_matches_the_contiguous_answer`, `binary_consuming_a_permuted_operand_matches_the_contiguous_answer`, `binary_with_a_broadcast_operand_matches_the_contiguous_answer`, `a_permuted_and_a_contiguous_operand_together_match_the_contiguous_answer`, `scalar_on_a_permuted_operand_matches_the_contiguous_answer`, and `unary_on_a_permuted_operand_matches_the_contiguous_answer`.

### Reproduction

https://github.com/Marc-AnthonyG/burn-reproduction/tree/main/cases/elementwise-memory-order

Run with `cargo run --release -p elementwise-memory-order --features cuda`.

## Metal
Screenshot 2026-09-08 at 3 25 47 PM

## Cuda
Screenshot 2026-09-08 at 3 23 54 PM

### DIFF
--- a/crates/burn-backend-tests/tests/tensor/float/ops/mod.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/mod.rs
@@ -105,6 +105,7 @@ mod negative_dims;
 mod one_hot;
 mod padding;
 mod permute;
+mod permuted_operands;
 mod powf;
 mod powf_scalar;
 mod prod;
--- a/crates/burn-backend-tests/tests/tensor/float/ops/permuted_operands.rs
+++ b/crates/burn-backend-tests/tests/tensor/float/ops/permuted_operands.rs
@@ -6,7 +6,7 @@
 //! changes what it computes.
 
 use super::*;
-use burn_tensor::Tolerance;
+use burn_tensor::{TensorData, Tolerance};
 
 /// A `[2, 3, 4, 5]` tensor held channels-last in memory, and a contiguous copy
 /// of the same values.
@@ -14,6 +14,7 @@ fn permuted_and_contiguous() -> (TestTensor<4>, TestTensor<4>) {
     let device = Default::default();
     let permuted = TestTensorInt::arange(0..(2 * 3 * 4 * 5), &device)
         .float()
+        .div_scalar(120.0)
         .reshape([2, 4, 5, 3])
         .permute([0, 3, 1, 2]);
     let contiguous = TestTensor::<4>::from_data(permuted.to_data(), &device);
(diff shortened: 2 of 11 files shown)


laggui (approved):
Looks good with the fixes to preserve vectorization for singleton dimensions and fall back to logical order when memory-order votes tie.

There is one performance trade-off when an unfused operation cannot reuse its input buffer and its output is subsequently flattened:
```rust
// x has logical NCHW shape [2, 8, 512, 512], stored channels-last.
let y = x.clone().mul_scalar(2.5); // Retaining x forces a fresh output.
let flat = y.reshape([2 * 8 * 512 * 512]);
// x goes out of scope or `drop(x)`
```

Previously, the new output was logically contiguous, so flatten only changed metadata. The new output preserves channels-last storage, which can make multiplication faster but requires a copy for flatten.

Retained inputs can also arise through branching or autodiff, without an explicit clone here. I think this trade-off is acceptable for the intended convolution consumers: it is not a general flatten regression, and in-place execution already incurred the same layout conversion before this change.

PR #5634, opened 2026-09-09
fix(flex): guard sum fast path with layout_covers_storage_once

### LINKED ISSUE #5600: burn-flex: sum over expanded/overlapping views is wrong when numel equals buffer length
## Summary

`sum` over a non-contiguous view sums the entire raw buffer whenever `numel == buffer_len`, which is not sufficient for the view to be a bijection onto storage.

## Repro

```rust
use burn_flex::{FlexTensor, ops};
use burn_backend::TensorData;
use burn_std::Shape;

let t = FlexTensor::from_data(TensorData::new(vec![0.0f32, 1.0, 2.0, 3.0], vec![1, 4]));
// view [2,2] strides [0,1] over the first two elements: numel 4 == buffer len 4
let v = ops::expand::expand(t.narrow(1, 0, 2), Shape::from(vec![2, 2]));
let r: Vec<f32> = ops::reduce::sum(v).into_data().try_into_vec().unwrap();
assert_eq!(r, vec![2.0]); // got [6.0]
```

The same shortcut misfires on overlapping `unfold` views whose element count happens to equal the buffer length.

## Cause

`sum_f32` at `crates/burn-flex/src/ops/reduce.rs:90`:

```rust
if data.len() == elem_count {
    sum_f32_contiguous(data)
```

Equal length does not imply every storage element is visited exactly once. `max_f32_reduce` / `min_f32_reduce` in the same file already use `layout_covers_storage_once`, which checks the stride structure properly.

## Fix

Replace the length check with `layout_covers_storage_once(tensor.layout(), data.len())`.

antimora:
### Checklist

### Related Issues/PRs

Fixes #5600

### Changes

`sum_f32` reduced the raw buffer whenever `numel == buffer_len`. That is necessary but not
sufficient for a view to be a bijection onto storage: a broadcast or overlapping view reads some
elements twice and others never while the counts still match. An expanded `[2, 2]` view with
strides `[0, 1]` over a 4-element buffer summed to 6.0 instead of 2.0.

Now guarded by `layout_covers_storage_once`, which `max`/`min` already use. Permuted and flipped
tensors keep the fast path; broadcast, overlapping, gapped and partial views fall back to strided
iteration.

### Testing

Two regression tests in the shared aggregation suite, so they run on every backend:

- `test_sum_expanded_over_full_buffer`, the broadcast view from the issue (fails without the fix,
  `6 != 2`).
- `test_sum_overlapping_unfold`, `unfold(0, 3, 2)` over six elements, which overlaps at index 2 and
  skips index 5 (fails without the fix, `15 != 12`). Thanks to @laggui for the case.

Aggregation suite green on `flex`, `flex-simd`, `flex-rayon` and `ndarray`. `cargo run-checks`
passes.


laggui (approved):
The fix looks correct, but I think we could actually include a regression test for the overlapping case without narrowing: `[0, 1, 2, 3, 4, 5].unfold(0, 3, 2)` has six logical elements but sums to 12 without the fix.

Could you add the test?

```rust
#[test]
  fn test_sum_overlapping_unfold() {
      // Unfold into [[0, 1, 2], [2, 3, 4]]: six logical elements over a
      // six-element buffer, but index 2 is repeated and index 5 is unused.
      // Summing the raw buffer would incorrectly give 15.0.
      let tensor = TestTensor::<1>::from([0.0, 1.0, 2.0, 3.0, 4.0, 5.0]);
      let unfolded: TestTensor<2> = tensor.unfold(0, 3, 2);
      let output = unfolded.sum();

      output
          .into_data()
          .assert_eq(&TensorData::from([12.0]), false);
  }
```

I'll approve in advance

antimora:
Good catch, and my reasoning in the PR description was wrong. I had claimed `windows * size == buffer_len` forces `windows == 1`, but I only worked the algebra for `step == 1`. With `step = 2`, `unfold(0, 3, 2)` over six elements gives `floor((6-3)/2) + 1 = 2` windows and `2 * 3 = 6` logical elements, matching the buffer while overlapping at index 2 and skipping index 5.

Added your test verbatim in ac12fda28. Confirmed it fails without the fix (`15 != 12`) and passes with it. Aggregation suite is green on `flex`, `flex-simd`, `flex-rayon` and `ndarray`, and `cargo run-checks` passes.

I also corrected the claim in the PR description so it does not mislead later.

Note the branch now includes a merge with main to pick up #5635, which should clear the `storage_tiled` failures on the Linux and Windows runners. The macOS `burn-linalg` LU `block_dispatch` NaN failures reproduce on main without this PR, so they look unrelated.

laggui:

> The macOS `burn-linalg` LU `block_dispatch` NaN failures reproduce on main without this PR, so they look unrelated.

Yeah I just noticed the issue this morning, it seems to be due to the reduction strategy used without autotune. The move to burn-linalg actually revealed this bug given that the CI doesn't enable autotune when testing the crate :sweat_smile: will investigate the root cause.

PR #5640, opened 2026-09-10
fix(flex): validate gather_nd/scatter_nd coordinate bounds

### LINKED ISSUE #5606: burn-flex: gather_nd/scatter_nd skip per-coordinate bounds checks
## Summary

`gather_nd` and `scatter_nd` in burn-flex do not validate per-coordinate indices. Out-of-range coordinates that still land inside the flat buffer silently read from or write to a neighboring row.

## Repro

```rust
use burn_flex::{FlexTensor, ops};
use burn_backend::TensorData;
use burn_backend::tensor::IndexingUpdateOp;

let data = FlexTensor::from_data(TensorData::new((0..6).map(|i| i as f32).collect::<Vec<_>>(), vec![2, 3]));
let idx = FlexTensor::from_data(TensorData::new(vec![0i64, 5], [1, 2])); // column 5 of a 3-wide tensor
let r: Vec<f32> = ops::gather_scatter::gather_nd::<f32>(data, idx).into_data().try_into_vec().unwrap();
// returns [5.0] (element at flat offset 5 = row 1, col 2) instead of panicking

let data = FlexTensor::from_data(TensorData::new(vec![0.0f32; 6], vec![2, 3]));
let idx = FlexTensor::from_data(TensorData::new(vec![0i64, 4], [1, 2]));
let vals = FlexTensor::from_data(TensorData::new(vec![9.0f32], vec![1]));
let r: Vec<f32> = ops::gather_scatter::scatter_nd::<f32>(data, idx, vals, IndexingUpdateOp::Assign).into_data().try_into_vec().unwrap();
// [0, 0, 0, 0, 9, 0]: wrote to row 1, col 1
```

Negative indices are cast straight to `usize` (`gather_scatter.rs:1035`, `:1110`), so they produce huge offsets and a confusing slice panic instead of an index error.

## Fix

Use the existing `checked_index(raw, data_shape[j])` helper per coordinate, as every other gather/scatter/select kernel in the file already does.

stack3mpty:
### Related Issues/PRs

Resolves #5606.

### Changes

`gather_nd` and `scatter_nd` in `burn-flex` did not validate each coordinate against the corresponding tensor dimension before computing the flattened offset.

As a result, an out-of-bounds coordinate could still resolve to a valid flat buffer position and silently access an element from another row. Negative indices were also cast directly to `usize`, resulting in misleading downstream panics.

This PR:

* Uses the existing `checked_index` helper for every coordinate in `gather_nd` and `scatter_nd`.
* Ensures both positive out-of-bounds and negative indices are rejected consistently.
* Adds regression tests covering these cases.

### Testing

Added regression tests for:

* An out-of-bounds coordinate in `gather_nd`.
* A negative coordinate in `scatter_nd`.

A validation PR was also run against the fork CI. Code quality, documentation, and Linux checks passed. The Windows crates job failed while compiling `wgpu-hal` because multiple incompatible versions of the `windows` / `windows-core` crates were present in the dependency graph; this failure is unrelated to the `burn-flex` changes in this PR.

### DIFF
--- a/crates/burn-flex/src/ops/gather_scatter.rs
+++ b/crates/burn-flex/src/ops/gather_scatter.rs
@@ -1032,7 +1032,7 @@ pub fn scatter_nd<
     for n in 0..num_indices {
         let mut base_offset = 0usize;
         for j in 0..k {
-            let idx_val = idx_data[n * k + j] as usize;
+            let idx_val = checked_index(idx_data[n * k + j], data_shape[j]);
             base_offset += idx_val * strides[j];
         }
 
@@ -1107,7 +1107,7 @@ pub fn gather_nd<E: Element + Pod + Default + Copy>(
     for n in 0..num_indices {
         let mut base_offset = 0usize;
         for j in 0..k {
-            let idx_val = idx_data[n * k + j] as usize;
+            let idx_val = checked_index(idx_data[n * k + j], data_shape[j]);
             base_offset += idx_val * strides[j];
         }
         let out_offset = n * slice_size;
@@ -1315,6 +1315,7 @@ pub fn scatter_or(
 mod tests {
     use super::*;
     use burn_backend::TensorData;
+    use burn_backend::tensor::IndexingUpdateOp;
 
     #[test]
     fn test_gather_with_i32_indices() {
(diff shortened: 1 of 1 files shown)


laggui (approved):
LGTM. Thanks for fixing.

Copilot's type and overflow concerns are already handled by `read_indices` The tests can stay in flex because bounds-checking is backend specific here.

PR #5642, opened 2026-09-10
test(cpu): cover avg_pool1d gradients with multiple channels

### LINKED ISSUE #5308: cpu: `avg_pool1d` backward returns uninitialised memory at even channel counts — a sentinel value from a freed allocation leaks verbatim into the gradient
On the `cpu` backend, `avg_pool1d`'s backward pass does not write its output buffer for some
channel counts. The caller receives whatever previously occupied that allocation.

This is not "the gradient is numerically wrong" — the buffer is never written at all, so the
values you get back depend on allocation history rather than on the input.

Unlike #5304, **this reproduces against the published crate**, not only against `main`.

## Evidence

Fill a large allocation with a recognisable value and drop it, so the allocator is likely to
hand that block to the pooling kernel. Then run the backward pass:

```text
c=1  sentinel_leaked=false  grad=[0.333, 0.667, 0.333, 0.667, 0.333, 0.333]   correct
c=2  sentinel_leaked=TRUE   grad=[1234.5, 1234.5, 1234.5, 1234.5, 1234.5, 1234.5]
c=3  sentinel_leaked=false  grad=[0.333, 0.667, 0.333, 0.667, 0.333, 0.333]   correct
c=4  sentinel_leaked=TRUE   grad=[1234.5, 1234.5, 1234.5, 1234.5, 1234.5, 1234.5]
```

`1234.5` is a value this computation never touches. It comes from an unrelated tensor that was
dropped before the pool ran. Reproduced identically across 3 separate runs.

Corroborating signal from a second angle: when the leftover memory is not float data, wrong
outputs come back as `1.401298e-45`, `2.8026e-45`, `4.2039e-45`. Decoded, those are f32 bit
patterns `0x1`, `0x2`, `0x3` — small integers reinterpreted as floats, consistent with the same
buffer never being written.

`ndarray` is correct for every channel count tested.

## Reproduction

```toml
[dependencies]
burn-tensor = { version = "0.22.0-pre.1", default-features = false, features = ["autodiff", "std", "cpu"] }
```

```rust
use burn_tensor::module::avg_pool1d;
use burn_tensor::{Tensor, TensorData};

fn device() -> burn_tensor::Device {
    burn_tensor::Device::default().autodiff()
}

/// Churn the allocator with a recognisable value, then let it go.
fn poison() {
    let big = vec![1234.5f32; 8192];
    let t = Tensor::<2>::from_data(TensorData::new(big, [64, 128]), &device());
    let _: Vec<f32> = (t.clone() * t).sum().to_data().to_vec().unwrap();
}

fn main() {
    for c in [1usize, 2, 3, 4] {
        poison();
        let d: Vec<f32> = (0..c * 6).map(|i| i as f32 * 0.1 + 0.5).collect();
        let x = Tensor::<3>::from_data(TensorData::new(d, [1, c, 6]), &device()).require_grad();
        let g = avg_pool1d(x.clone(), 3, 2, 1, true, false).sum().backward();
        let grad: Vec<f32> = x.grad(&g).unwrap().to_data().to_vec().unwrap();
        println!("c={c} grad={grad:?}");
    }
}
```

Expected for every channel: `[1/3, 2/3, 1/3, 2/3, 1/3, 1/3]` (padded length 8, kernel 3,
stride 2 → windows at 0, 2, 4).

## Scope

Swept `avg_pool1d` and `max_pool1d` over channels × length × kernel × stride × padding and
compared against `ndarray` at `1e-4`: **210 of 431 configurations differ**, spread across the
plain forward, the grad-tracked forward, and the backward.

`max_pool1d`'s **forward** is affected too, so this is not confined to the backward pass. Two of
burn's own tests already fail on this backend:

```text
cargo test -p burn-backend-tests --no-default-features --features cpu,std \
  --test tensor --release max_pool1d

failures:
    tensor::float::module::maxpool1d::test_max_pool1d_simple
    tensor::float::module::maxpool1d::test_max_pool1d_with_dilation
```

Both are the 2-channel cases. One assertion diff reads `0.8221 != -0.4708`, where `-0.4708` is
input data belonging to a *different test in the same file* — the same contamination from
another angle.

One more detail that may help localise it: the plain and `require_grad` paths do not agree.
Autodiff needs argmax indices and so calls the `_with_indices` variant, and some configurations
are correct plain but wrong when tracked.

## Versions

- **Reproduces** on published `burn-cpu` / `burn-tensor` `0.22.0-pre.1` from crates.io
  (resolving `cubecl 0.11.0-pre.1`) — verified via a standalone crate with no connection to a
  burn checkout.
- **Reproduces** on `main` (which pins cubecl `b2703c28`).
- **Not checked** against `0.21.0`, the current stable release — I did not want to claim
  impact I had not measured. Happy to check if that would be useful.
- macOS / Apple Silicon, release profile.

## Why the test suite does not catch it

Because the buffer is never written, the contents depend on what the allocator hands over. In an
isolated single test the block is often freshly zeroed, so the gradient looks like a plausible
(if wrong) all-zeros. Run inside a fuller suite and the same case returns leftover tensor data
instead. That makes the failure look intermittent when it is in fact perfectly deterministic
given allocation history.

I would be glad to contribute a regression test using the sentinel technique above, since it
detects the condition regardless of what the leftover memory happens to contain.

lorenzozanee:
### Checklist

### Related Issues/PRs

Fixes #5308

### Changes

Adds a focused CPU autodiff regression test for `avg_pool1d` with even channel counts. The test poisons allocator contents and verifies every input-gradient element against the pooling derivative.

### Testing

Focused CPU autodiff `avg_pool1d`, tensor `avg_pool1d`, tensor `max_pool1d`, formatting, and diff checks pass locally; the full `cargo run-checks` command was not run.

### DIFF
--- a/crates/burn-backend-tests/tests/autodiff/avgpool1d.rs
+++ b/crates/burn-backend-tests/tests/autodiff/avgpool1d.rs
@@ -1,6 +1,52 @@
 use super::*;
 use burn_tensor::module::avg_pool1d;
-use burn_tensor::{Shape, Tolerance};
+use burn_tensor::{Shape, TensorData, Tolerance};
+
+#[test]
+fn test_avg_pool1d_backward_writes_every_channel() {
+    let device = AutodiffDevice::new();
+    let poison =
+        TestTensor::<2>::from_data(TensorData::new(vec![1234.5; 8192], [64, 128]), &device);
+    let _ = (poison.clone() * poison).sum().into_data();
+
+    for channels in [2, 4] {
+        let x = TestTensor::<3>::from_data(
+            TensorData::new(
+                (0..channels * 6)
+                    .map(|index| index as f32 * 0.1 + 0.5)
+                    .collect(),
+                [1, channels, 6],
+            ),
+            &device,
+        )
+        .require_grad();
+        let output = avg_pool1d(x.clone(), 3, 2, 1, true, false);
+        let grads = output.sum().backward();
+        let actual = x.grad(&grads).unwrap();
+        let expected = TestTensor::<3>::from_data(
+            TensorData::new(
+                vec![
+                    1.0 / 3.0,
+                    2.0 / 3.0,
+                    1.0 / 3.0,
+                    2.0 / 3.0,
+                    1.0 / 3.0,
+                    1.0 / 3.0,
+                ]
+                .into_iter()
+                .cycle()
+                .take(channels * 6)
+                .collect(),
+                [1, channels, 6],
+            ),
+            &device,
+        );
+
+        expected
+            .to_data()
+            .assert_approx_eq::<FloatElem>(&actual.into_data(), Tolerance::default());
+    }
+}
 
 #[test]
 fn test_avg_pool1d_simple() {


laggui (changes requested):
Hey, thanks for looking into this.

`test_avg_pool1d_complex` already checks the same two-channel case and expected gradients. The additions here are the four-channel case and allocator poisoning, but the poisoning doesn't guarantee that the gradient buffer reuses that memory.

Since #5308 predates the pliron CPU refactor,  I think it should be revisited to check if it still reproduces on current main. It might have already been resolved.

PR #5646, opened 2026-09-10
refactor(conv): call cubek's direct convolution routine

ThierryCantin-Demers:
The direct convolution kernel moves to `cubek-convolution`, beside the depthwise routine this crate already delegates to and the accelerated ones. What a tensor library owns stays here: making the operands contiguous along the channel, and allocating the result.

422 lines become 87, the same shape as the depthwise wrapper next to it.

**No behaviour change: the kernel moved verbatim.** It is byte-identical to this crate's `direct.rs` on `main`.

Why move it. cubek scores its benchmarks against measured device ceilings (`measure_peak_throughput`, `measure_memory_curve`) and its convolution crate already carries `Conv2dCost::traffic()` and `compute_key()`. A kernel here can reach none of that, so measuring this one has meant a bespoke harness and hand-computed theoretical peaks. It also puts the kernel next to the `compute_bound` / `memory_bound` autotune bounds that price a candidate against both roofs.

Checks: clippy clean with `-D warnings`; 35 conv2d tests pass on CPU and 35 on wgpu, built against the cubek branch through a local `[patch]`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2757,7 +2757,7 @@ dependencies = [
 [[package]]
 name = "cubek"
 version = "0.3.0-pre.3"
-source = "git+https://github.com/tracel-ai/cubek?rev=3de52af6cd1bc7dacf7898a863048267543f249e#3de52af6cd1bc7dacf7898a863048267543f249e"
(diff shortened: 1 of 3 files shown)


laggui (approved):
LGTM! Just need to update the rev once the cubek PR is merged

PR #5650, opened 2026-09-11
fix(flex): honor asymmetric padding in conv3d

### LINKED ISSUE #5601: burn-flex: conv3d ignores asymmetric padding (end padding dropped)
## Summary

`Flex::conv3d` ignores asymmetric padding. `conv1d` and `conv2d` route through `pad_asymmetric_conv_input` first; `conv3d` does not, and the kernel only reads `padding_begin()`.

## Repro

```rust
use burn_backend::ops::{ConvOptions, ModuleOps};
use burn_backend::TensorData;

let x = FlexTensor::from_data(TensorData::new(vec![1.0f32, 2.0, 3.0, 4.0], vec![1, 1, 1, 1, 4]));
let w = FlexTensor::from_data(TensorData::new(vec![1.0f32, 1.0], vec![1, 1, 1, 1, 2]));
let opts = ConvOptions::<3>::new_with_padding([1, 1, 1], [(0, 0), (0, 0), (0, 1)], [1, 1, 1], 1);
let out = Flex::conv3d(x, w, None, opts);
// shape [1,1,1,1,3] values [3,5,7]; expected [1,1,1,1,4] values [3,5,7,4]
```

The equivalent `conv2d` call with `[(0, 0), (0, 1)]` returns `[1,1,1,4]` = `[3,5,7,4]`.

## Cause

`crates/burn-flex/src/ops/module.rs:248` calls `conv::conv3d_*` directly. Inside, `conv3d_impl` and the fast paths use `options.padding_begin()` for both the output-size computation and the im2col bounds, so the end padding is dropped.

## Fix

Add `let (x, options) = pad_asymmetric_conv_input::<Flex, 3>(x, options);` at the top of `conv3d`, matching `conv1d`/`conv2d`.

onenewcode:
### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/5601

### Changes

`Flex::conv3d` dropped end padding. `conv1d` and `conv2d` already call `pad_asymmetric_conv_input` first; `conv3d` did not, and the kernel only reads `padding_begin()`.

This adds the same helper call at the top of `Flex::conv3d`:

```rust
let (x, options) = pad_asymmetric_conv_input::<Flex, 3>(x, options);
```

### Testing

**Issue repro** (`cargo test` locally: pass):

```rust
let x = FlexTensor::from_data(TensorData::new(vec![1.0f32, 2.0, 3.0, 4.0], vec![1, 1, 1, 1, 4]));
let w = FlexTensor::from_data(TensorData::new(vec![1.0f32, 1.0], vec![1, 1, 1, 1, 2]));
let opts = ConvOptions::<3>::new_with_padding([1, 1, 1], [(0, 0), (0, 0), (0, 1)], [1, 1, 1], 1);
let out = Flex::conv3d(x, w, None, opts);
// shape [1,1,1,1,4], values [3, 5, 7, 4]
```

### DIFF
--- a/crates/burn-flex/src/ops/module.rs
+++ b/crates/burn-flex/src/ops/module.rs
@@ -251,6 +251,7 @@ impl ModuleOps<Flex> for Flex {
         bias: Option<FloatTensor<Flex>>,
         options: ConvOptions<3>,
     ) -> FloatTensor<Flex> {
+        let (x, options) = pad_asymmetric_conv_input::<Flex, 3>(x, options);
         match x.dtype() {
             DType::F32 => conv::conv3d_f32(x, weight, bias, &options),
             DType::F64 => conv::conv3d_f64(x, weight, bias, &options),


antimora:
Can we also checking the test?

onenewcode:
@antimora Thanks for the review.

I put the issue [#5601](https://github.com/tracel-ai/burn/issues/5601) regression on a separate branch so this PR stays as the one-line `Flex::conv3d` fix:

https://github.com/onenewcode/burn/tree/test/flex-conv3d-asymmetric-padding

You can use that branch to run the tests:

```bash
cargo test -p burn-flex --test conv3d_asymmetric_padding
```

I didn't add them here because I don't think a one-line change needs a large test addition. If maintainers still want the tests in this PR, I can add them.

laggui (commented):
Thanks for addressing the fix!

Could you include the test in this PR? The value of a test isn't tied to the number of lines in the fix imo. The test makes sure the reported bug stays fixed.

A small test using the issue's repro and asserting the output shape and values would be enough.

onenewcode:
@laggui Added the regression test in this PR.

It uses the issue #5601 repro against `Flex::conv3d` and asserts shape `[1, 1, 1, 1, 4]` and values `[3, 5, 7, 4]`.

laggui (approved):
Thanks for addressing my comment

PR #5652, opened 2026-09-11
fix(flex): handle remainder edge cases

### LINKED ISSUE #5610: burn-flex: remainder edge cases (float rounding to divisor, inf divisor, int overflow)
## Summary

burn-flex implements Python-style remainder as `((a % b) + b) % b` for both float (`crates/burn-flex/src/ops/float.rs:167`) and signed int (`crates/burn-flex/src/ops/int.rs:558`). The double-modulo form has edge cases that PyTorch's formulation does not.

## Float

```rust
use burn_backend::ops::FloatTensorOps;
use burn_backend::TensorData;

let a = FlexTensor::from_data(TensorData::new(vec![-1e-20f32, -1.0, 5.0], vec![3]));
let b = FlexTensor::from_data(TensorData::new(vec![1.0f32, f32::INFINITY, -3.0], vec![3]));
let r: Vec<f32> = Flex::float_remainder(a, b).into_data().try_into_vec().unwrap();
// got  [0.0, NaN, -1.0]
// torch [1.0, inf, -1.0]
```

- `-1e-20 % 1.0 = -1e-20`, then `+ 1.0` rounds to exactly `1.0`, then `% 1.0` gives `0.0`. PyTorch returns `1.0` (well, `1.0 - 1e-20` rounded), i.e. the result stays in `[0, b)`.
- `a % inf = a`, then `a + inf = inf`, then `inf % inf = NaN`.

## Int

For signed ints, `(a % b) + b` can overflow: `remainder(i64::MAX - 1, i64::MAX)` panics in debug and wraps in release. `i64::MIN % -1` also overflows before the fix-up runs.

## Fix

Use the single-modulo form PyTorch uses:

```rust
let r = a % b;
if r != 0 && ((r < 0) != (b < 0)) { r + b } else { r }
```

For ints, use `wrapping_rem` and `wrapping_add`.

onenewcode:
### Checklist

### Related Issues/PRs

Fixes https://github.com/tracel-ai/burn/issues/5610

### Changes

burn-flex remainder used `((a % b) + b) % b` for float and signed int. That double-modulo form is what #5610 reports.

The fix is the single-modulo form from the issue, inlined at the existing `binary_op` / `int_binary_op` (and scalar) call sites. Integers use `wrapping_rem` / `wrapping_add`.

```rust
let r = a % b;
if r != 0 && ((r < 0) != (b < 0)) { r + b } else { r }
```

### Testing

**test_float_remainder_inf_and_tiny** — issue float repro, three pairs in one tensor:

```rust
let a = FlexTensor::from_data(TensorData::new(vec![-1e-20f32, -1.0, 5.0], [3]));
let b = FlexTensor::from_data(TensorData::new(vec![1.0f32, f32::INFINITY, -3.0], [3]));
let r: Vec<f32> = Flex::float_remainder(a, b).into_data().try_into_vec().unwrap();
// old   [0.0, NaN, -1.0]
// torch [1.0, inf, -1.0]
assert_eq!(r, vec![1.0, f32::INFINITY, -1.0]);
```

- `-1e-20 % 1.0`: `-1e-20 + 1.0` rounds to `1.0`, then `1.0 % 1.0` was `0.0`. Expected `1.0`.
- `-1.0 % inf`: `-1.0 + inf` then `inf % inf` was `NaN`. Expected `inf`.
- `5.0 % -3.0`: remainder keeps the divisor's sign. Expected `-1.0`.

**test_int_remainder_overflow** — issue int overflow cases, two pairs in one tensor:

```rust
let a = FlexTensor::from_data(TensorData::new(vec![i64::MAX - 1, i64::MIN], [2]));
let b = FlexTensor::from_data(TensorData::new(vec![i64::MAX, -1], [2]));
// remainder(MAX-1, MAX) -> MAX-1  (old: (a % b) + b overflows)
// remainder(MIN, -1)    -> 0      (old: MIN % -1 overflows)
assert_eq!(values, vec![i64::MAX - 1, 0]);
```

### DIFF
--- a/crates/burn-flex/src/ops/float.rs
+++ b/crates/burn-flex/src/ops/float.rs
@@ -1248,6 +1248,8 @@ mod tests {
     fn test_float_remainder_inf_and_tiny() {
         use burn_backend::ops::FloatTensorOps;
 
+        // `((a % b) + b) % b` rounds `-1e-20 + 1.0` to 1.0 then wraps to 0.0,
+        // and `-1.0 % inf` becomes NaN. 5 % -3 keeps the divisor's sign.
         let a = crate::FlexTensor::from_data(TensorData::new(vec![-1e-20f32, -1.0, 5.0], [3]));
         let b =
             crate::FlexTensor::from_data(TensorData::new(vec![1.0f32, f32::INFINITY, -3.0], [3]));
--- a/crates/burn-flex/src/ops/float.rs
+++ b/crates/burn-flex/src/ops/float.rs
@@ -169,8 +169,22 @@ impl FloatTensorOps<Flex> for Flex {
         binary_op(
             lhs,
             rhs,
-            |a, b| ((a % b) + b) % b,
-            |a, b| ((a % b) + b) % b,
+            |a, b| {
+                let r = a % b;
+                if r != 0. && (r < 0.) != (b < 0.) {
+                    r + b
+                } else {
+                    r
+                }
+            },
+            |a, b| {
+                let r = a % b;
+                if r != 0. && (r < 0.) != (b < 0.) {
+                    r + b
+                } else {
+                    r
+                }
+            },
             None,
         )
     }
@@ -181,8 +195,22 @@ impl FloatTensorOps<Flex> for Flex {
         scalar_op(
             lhs,
             rhs_val,
-            |a, b| ((a % b) + b) % b,
-            |a, b| ((a % b) + b) % b,
+            |a, b| {
+                let r = a % b;
+                if r != 0. && (r < 0.) != (b < 0.) {
+                    r + b
+                } else {
+                    r
+                }
+            },
+            |a, b| {
+                let r = a % b;
+                if r != 0. && (r < 0.) != (b < 0.) {
+                    r + b
+                } else {
+                    r
+                }
+            },
         )
     }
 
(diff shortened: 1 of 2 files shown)


crates/burn-flex/src/ops/float.rs
```diff
@@ -1216,6 +1244,22 @@ mod tests {
...
+        // and `-1.0 % inf` becomes NaN. 5 % -3 keeps the divisor's sign.
+        let a = crate::FlexTensor::from_data(TensorData::new(vec![-1e-20f32, -1.0, 5.0], [3]));
+        let b =
+            crate::FlexTensor::from_data(TensorData::new(vec![1.0f32, f32::INFINITY, -3.0], [3]));
+        let data: Vec<f32> = Flex::float_remainder(a, b)
+            .into_data()
+            .try_into_vec()
+            .unwrap();
+        assert_eq!(data, vec![1.0, f32::INFINITY, -1.0]);
+    }
```
laggui: Although the fix specifically targets `Flex`, we could move this to the `burn-backend-tests` instead to exercise the behavior across backends, including flex.

laggui (commented):
LGTM, thanks for addressing the issue. Just one comment regarding the float test.

The additional tests requested by Copilot could also be added, though not strictly required.

onenewcode:
### Cubecl and `should_support_remainder_inf_and_tiny`

Cubecl cannot pass  Remainder still goes through `Vector::mod_floor`:

```rust
impl<T: Numeric, N: Size> BinaryOp<T, N> for RemainderOp {
    fn execute(lhs: Vector<T, N>, rhs: Vector<T, N>) -> Vector<T, N> {
        Vector::mod_floor(lhs, rhs)
    }
}
```

That is `a - b * floor(a / b)`. For `-1.0 % inf`:

```
-1.0 / inf     = -0.0
floor(-0.0)    = 0
inf * 0        = NaN
-1.0 - NaN     = NaN
```

The test expects `inf` (Python / PyTorch / Flex). Tiny (`-1e-20 % 1.0`) and `5 % -3` are fine with `mod_floor`; only the inf divisor hits IEEE `inf * 0`. Fusion rem uses the same `mod_floor`, so a fused remainder would also return NaN.

This PR does not change cubecl.

laggui (approved):
Thanks for addressing my comment! I feature-gated the appropriate test given the CubeCL behavior. It should be addressed separately.

PR #5653, opened 2026-09-11
fix: correct Flex average-pooling gradients with ceil mode

### LINKED ISSUE #5604: burn-flex: avg_pool2d_backward divisor disagrees with forward in ceil_mode with count_include_pad
## Summary

With `count_include_pad = true` and `ceil_mode = true`, burn-flex `avg_pool2d` forward divides an overhanging last window by the clipped window size, but `avg_pool2d_backward` divides by the full kernel volume. The gradient does not match the forward.

## Repro

```rust
use burn_backend::ops::ModuleOps;
use burn_backend::TensorData;

// width 5, kernel 2, stride 2, ceil_mode -> 3 outputs; last window holds one valid element
let x = FlexTensor::from_data(TensorData::new(vec![1.0f32, 2.0, 3.0, 4.0, 5.0], vec![1, 1, 1, 5]));
let fwd: Vec<f32> = Flex::avg_pool2d(x.clone(), [1, 2], [1, 2], [0, 0], true, true).into_data().try_into_vec().unwrap();
// [1.5, 3.5, 5.0]  -> last output = x[4] / 1

let grad = FlexTensor::from_data(TensorData::new(vec![1.0f32; 3], vec![1, 1, 1, 3]));
let bwd: Vec<f32> = Flex::avg_pool2d_backward(x, grad, [1, 2], [1, 2], [0, 0], true, false).into_data().try_into_vec().unwrap();
// [0.5, 0.5, 0.5, 0.5, 0.5] -> x_grad[4] should be 1.0
```

## Cause

Forward (`crates/burn-flex/src/ops/pool.rs:763`) uses `pad_count = padded_len_avgpool(...)`, which clips the window to `input + 2*pad`. Backward (`pool.rs:1432`) uses `kernel_volume` unconditionally. The two agree only when no window overhangs, i.e. when `ceil_mode` is false.

## Fix

Compute the same clipped divisor in the backward pass (the backward already has `stride`, `padding`, `kernel_size` and the input shape; only the ceil-mode overhang needs the `padded_len_avgpool` treatment).

slobodaapl:
### Checklist

- [x] Made sure the book is up to date with changes in this PR. (no changes needed)

### Related Issues/PRs

Fixes #5604 by using the same clipped padded divisor in average-pooling backward as in forward when ceil_mode is enabled.

### Changes

Changed files:

- crates/burn-flex/src/ops/pool.rs: corrected the shared 3D backward divisor, which also fixes 2D backward. Added coverage for windows overhanging all three spatial dimensions
- crates/burn-backend-tests/tests/autodiff/avgpool2d.rs: added coverage for the reported width-five example, actual padding, both count_include_pad settings, and nonuniform upstream gradients
- Cargo.toml and Cargo.lock: temporarily pin CubeK to the fix in tracel-ai/cubek#648

### Testing

- Verified the regressions fail with the old divisor and pass with the fix on Flex. Upstream gradients [2, 4, 8] now produce the expected input gradient [1, 1, 2, 2, 8]. The same failure was reproduced on CUDA and now passes with the updated CubeK pin
- CUDA pooling tests passed against the pinned CubeK commit
- Focused pooling and autodiff tests passed, including checkpointing
- Mandatory cargo run-checks passed

### Caveats

I've found that the same defect also exists in Burn’s previous CubeK pin, namely across all CUDA, ROCm, WGPU, and CubeCL CPU backends.

I ran the new test with the CUDA backend and reproduced the issue there before fixing it in CubeK. This PR now temporarily pins that CubeK fix, and the CUDA pooling tests pass with it.

I've submitted tracel-ai/cubek#648 to fix the issue there. I'll keep this PR as a draft until it is merged and will update the pin to the final commit SHA before opening it for final review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2757,7 +2757,7 @@ dependencies = [
 [[package]]
 name = "cubek"
 version = "0.3.0-pre.3"
-source = "git+https://github.com/tracel-ai/cubek?rev=3de52af6cd1bc7dacf7898a863048267543f249e#3de52af6cd1bc7dacf7898a863048267543f249e"
+source = "git+https://github.com/tracel-ai/cubek?rev=a1dff403c3a6a8ac2f0f03fd5a4afd05d94e45ad#a1dff403c3a6a8ac2f0f03fd5a4afd05d94e45ad"
 dependencies = [
  "cubecl",
  "cubek-attention",
@@ -2775,7 +2775,7 @@ dependencies = [
 [[package]]
 name = "cubek-attention"
 version = "0.3.0-pre.3"
-source = "git+https://github.com/tracel-ai/cubek?rev=3de52af6cd1bc7dacf7898a863048267543f249e#3de52af6cd1bc7dacf7898a863048267543f249e"
+source = "git+https://github.com/tracel-ai/cubek?rev=a1dff403c3a6a8ac2f0f03fd5a4afd05d94e45ad#a1dff403c3a6a8ac2f0f03fd5a4afd05d94e45ad"
 dependencies = [
  "bytemuck",
  "cubecl",
@@ -2789,7 +2789,7 @@ dependencies = [
 [[package]]
 name = "cubek-convolution"
 version = "0.3.0-pre.3"
-source = "git+https://github.com/tracel-ai/cubek?rev=3de52af6cd1bc7dacf7898a863048267543f249e#3de52af6cd1bc7dacf7898a863048267543f249e"
+source = "git+https://github.com/tracel-ai/cubek?rev=a1dff403c3a6a8ac2f0f03fd5a4afd05d94e45ad#a1dff403c3a6a8ac2f0f03fd5a4afd05d94e45ad"
 dependencies = [
  "bytemuck",
  "cubecl",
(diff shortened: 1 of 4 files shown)


antimora:
Is this ready for a review?

slobodaapl:
@antimora Technically ready for review, but as per instructions in cubek repo, I pinned the commit to the PR commit sha I have open there tracel-ai/cubek#648 so this cannot be merged until that PR is merged and I've updated burn's cubek dependency pin to the main branch's merged commit sha in the cubek repo.

laggui:
@slobodaapl looks like the cubek PR was merged. We can update the cubek revision accordingly. I'm about to publish the pre-release for burn, so I might update this myself in the next 10-15 minutes and merge the PR. The changes look good otherwise.

slobodaapl:
@laggui sure no prob, go for it. Assuming I can leave the cargo pin and conflicts to you then? Or should I go ahead and add a commit?

laggui (approved):

laggui:
> @laggui sure no prob, go for it. Assuming I can leave the cargo pin and conflicts to you then? Or should I go ahead and add a commit?

Just fixed that! You won't have to change it

Issue #5615, opened 2026-09-08
burn-flex: minor PyTorch parity differences (NaN in max_pool, half scalar rounding, clamp bounds)


antimora:
Small semantic differences from PyTorch found while auditing burn-flex. None is clearly a bug; each is a design choice that should either be made deliberately (and documented) or aligned. Filed together so they are not lost.

1. **max_pool ignores NaN unless it is the first element in the window.** `crates/burn-flex/src/ops/pool.rs:285` uses `val > max_val`, which is false for NaN. PyTorch propagates NaN through max pooling. Cheap fix: `if max_idx < 0 || val > max_val || val.is_nan()`.

2. **Half-precision scalar ops pre-round the scalar to f16/bf16.** `scalar_op` (`crates/burn-flex/src/ops/binary.rs:294`) converts the f64 scalar to the half type, then computes in f32. PyTorch keeps the scalar as an f32 "opmath" value. Example: `f16(3.0) * 0.1` gives `0.2998` here and `0.3000` in torch (one ulp). The existing test `test_scalar_f16_non_representable` pins the current behavior, so this is intentional, but the comment says "matching dtype semantics" which is not what torch does.

3. **Half-precision `compare_elem` rounds the scalar too** (`crates/burn-flex/src/ops/comparison.rs:324`). `x >= 0.1` for the f16 value nearest 0.1 returns true here and false in torch.

4. **Fully-masked attention rows return zeros** (`crates/burn-flex/src/ops/attention.rs:584`, `:912`). This matches `attention_fallback` in burn-backend, which deliberately clamps to avoid NaN, so no change needed; noting for completeness since torch SDPA returns NaN.

5. **`float_clamp` panics when `min > max` or either bound is NaN** because it uses `f32::clamp`, which asserts. `torch.clamp(x, min=1, max=0)` returns `max` for every element.

6. **`grid_sample` nearest mode uses `round()` (half away from zero)**; PyTorch uses `nearbyint` (half to even). Only differs at exact `.5` coordinates.

huahuadeliaoliao:
I can take item 5.

[...]

PR #5658, opened 2026-09-12
fix(flex): propagate NaN through relu and clamp_min/clamp_max

### LINKED ISSUE #5609: burn-flex: relu / clamp_min / clamp_max turn NaN into the bound
## Summary

burn-flex `relu`, `float_clamp_min` and `float_clamp_max` map NaN to the bound instead of propagating it, because they use `f32::max` / `f32::min`, which return the non-NaN operand. `float_clamp` uses `f32::clamp` and does propagate NaN, so the crate is inconsistent with itself and with PyTorch.

## Repro

```rust
use burn_backend::ops::{ActivationOps, FloatTensorOps};
use burn_backend::{Scalar, TensorData};

let t = FlexTensor::from_data(TensorData::new(vec![f32::NAN], vec![1]));
let r: Vec<f32> = Flex::relu(t.clone()).into_data().try_into_vec().unwrap();          // [0.0]
let r: Vec<f32> = Flex::float_clamp_min(t.clone(), Scalar::from(0.0f32)).into_data().try_into_vec().unwrap(); // [0.0]
let r: Vec<f32> = Flex::float_clamp(t, Scalar::from(0.0f32), Scalar::from(1.0f32)).into_data().try_into_vec().unwrap(); // [NaN]
```

`torch.relu(nan)`, `torch.clamp_min(nan, 0)` both return NaN.

Locations: `crates/burn-flex/src/ops/activation.rs:24` (relu), `crates/burn-flex/src/ops/float.rs:617` and `:627` (clamp_min / clamp_max).

## Why it matters

Silently replacing NaN with 0 in relu hides divergence during training and makes NaN checks after activations useless.

## Fix

Use `if x.is_nan() || x > min { x } else { min }` (or `x.max(min)` guarded by an `is_nan` check) so NaN propagates. Same for the f64 closures.

Liberxue:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. *(all steps but `audit`, see Testing)*

### Related Issues/PRs

Fixes #5609. Filed #5659 for the same gap on the cube backends.

### Changes

`f32::max` / `f32::min` return the non-NaN operand, so `relu`, `float_clamp_min` and
`float_clamp_max` mapped NaN to the bound. `float_clamp` uses `f32::clamp` and propagates it, so the
crate disagreed with itself and with PyTorch. Testing `is_nan` before the comparison makes all four
agree.

`relu_backward` had the same hole: it masked on `out > 0.0`, false for NaN, so a NaN output lost its
gradient, while the trait default masks with `float_lower_equal_elem(output, 0)` and keeps it. That
fix is its own commit (204cfec) and can be split out if you would rather keep this PR to the forward
ops.

Only NaN behaves differently. Every other value takes the same branch as before, so callers in
`burn-nn` and the examples are unchanged. `relu(-0.0)` now returns `+0.0` instead of whichever zero
`f32::max` happened to pick.

`!(x <= 0.0)` is equivalent and compiles to the same symbol, but `clippy::neg_cmp_op_on_partial_ord`
rejects it under `--deny warnings`, so the `is_nan()` spelling stays.

### Testing

Tests are in `burn-backend-tests`, per the rule above each flex test module. Gating follows what the
backends actually do:

| op | flex | ndarray | cuda | metal |
| --- | --- | --- | --- | --- |
| `relu`, `clamp_min`, `clamp_max` | pass | pass | pass | pass |
| `clamp` (two-sided) | pass | pass | fail | not run |

The first row is ungated; two-sided `clamp` is gated to the CPU backends, as are the f64 variants.
Reverting either source fix makes the matching tests fail.

`cargo audit` cannot reach the RustSec database from here, so the remaining steps were run
separately: fmt, clippy with `--deny warnings`, typos, doc build, doc tests (238 passed), no-std
build and test (2978 passed), and the backends and examples shards. Flex suites in release: 4103
passed, 0 failed.

### DIFF
--- a/crates/burn-flex/src/ops/activation.rs
+++ b/crates/burn-flex/src/ops/activation.rs
@@ -21,7 +21,13 @@ use crate::{Flex, FlexTensor, Layout};
 
 impl ActivationOps<Flex> for Flex {
     fn relu(tensor: FloatTensor<Flex>) -> FloatTensor<Flex> {
-        unary_op(tensor, |x: f32| x.max(0.0), |x: f64| x.max(0.0))
+        // `max` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
+        unary_op(
+            tensor,
+            |x: f32| if x.is_nan() || x > 0.0 { x } else { 0.0 },
+            |x: f64| if x.is_nan() || x > 0.0 { x } else { 0.0 },
+        )
     }
 
     fn relu_backward(output: FloatTensor<Flex>, grad: FloatTensor<Flex>) -> FloatTensor<Flex> {
@@ -1516,5 +1522,22 @@ mod tests {
         let t = flex_f32(vec![1.0, 2.0, 3.0, 4.0], &[1, 4]);
         let gamma = flex_f64(vec![1.0; 4], &[4]);
         let _ = crate::ops::activation::layer_norm(t, gamma, None, 1e-5);
+    }
+
+    // Regression test for #5609. `f32::max` / `f64::max` return the non-NaN
+    // operand, so relu used to map NaN to 0 and hide divergence during training.
+    #[test]
+    fn test_relu_propagates_nan() {
+        use burn_backend::ops::ActivationOps;
+
+        let t = FlexTensor::from_data(TensorData::from([f32::NAN, -1.0, 0.0, 2.0]));
+        let out: Vec<f32> = crate::Flex::relu(t).into_data().try_into_vec().unwrap();
+        assert!(out[0].is_nan());
+        assert_eq!(out[1..], [0.0, 0.0, 2.0]);
+
+        let t = FlexTensor::from_data(TensorData::from([f64::NAN, -1.0f64, 0.0, 2.0]));
+        let out: Vec<f64> = crate::Flex::relu(t).into_data().try_into_vec().unwrap();
+        assert!(out[0].is_nan());
+        assert_eq!(out[1..], [0.0, 0.0, 2.0]);
     }
 }
--- a/crates/burn-flex/src/ops/activation.rs
+++ b/crates/burn-flex/src/ops/activation.rs
@@ -22,7 +22,9 @@ use crate::{Flex, FlexTensor, Layout};
 impl ActivationOps<Flex> for Flex {
     fn relu(tensor: FloatTensor<Flex>) -> FloatTensor<Flex> {
         // `max` returns the non-NaN operand, which would map NaN to the bound.
-        // Test `is_nan` first so NaN propagates, as PyTorch does.
+        // Testing `is_nan` first lets NaN propagate, as PyTorch does. `!(x <= 0.0)` is
+        // equivalent and compiles to the same code, but trips
+        // `clippy::neg_cmp_op_on_partial_ord`.
         unary_op(
             tensor,
             |x: f32| if x.is_nan() || x > 0.0 { x } else { 0.0 },
@@ -1522,22 +1524,5 @@ mod tests {
         let t = flex_f32(vec![1.0, 2.0, 3.0, 4.0], &[1, 4]);
         let gamma = flex_f64(vec![1.0; 4], &[4]);
         let _ = crate::ops::activation::layer_norm(t, gamma, None, 1e-5);
-    }
-
-    // Regression test for #5609. `f32::max` / `f64::max` return the non-NaN
-    // operand, so relu used to map NaN to 0 and hide divergence during training.
-    #[test]
-    fn test_relu_propagates_nan() {
-        use burn_backend::ops::ActivationOps;
-
-        let t = FlexTensor::from_data(TensorData::from([f32::NAN, -1.0, 0.0, 2.0]));
-        let out: Vec<f32> = crate::Flex::relu(t).into_data().try_into_vec().unwrap();
-        assert!(out[0].is_nan());
-        assert_eq!(out[1..], [0.0, 0.0, 2.0]);
-
-        let t = FlexTensor::from_data(TensorData::from([f64::NAN, -1.0f64, 0.0, 2.0]));
-        let out: Vec<f64> = crate::Flex::relu(t).into_data().try_into_vec().unwrap();
-        assert!(out[0].is_nan());
-        assert_eq!(out[1..], [0.0, 0.0, 2.0]);
     }
 }
--- a/crates/burn-flex/src/ops/activation.rs
+++ b/crates/burn-flex/src/ops/activation.rs
@@ -33,12 +33,14 @@ impl ActivationOps<Flex> for Flex {
     }
 
     fn relu_backward(output: FloatTensor<Flex>, grad: FloatTensor<Flex>) -> FloatTensor<Flex> {
-        // grad * (output > 0): zero the gradient where output was zero
+        // Zero the gradient where the output was zero, but keep it for a NaN output:
+        // the trait default masks with `float_lower_equal_elem(output, 0)`, which is
+        // false for NaN.
         binary_op(
             output,
             grad,
-            |out: f32, g| if out > 0.0 { g } else { 0.0 },
-            |out: f64, g| if out > 0.0 { g } else { 0.0 },
+            |out: f32, g| if out.is_nan() || out > 0.0 { g } else { 0.0 },
+            |out: f64, g| if out.is_nan() || out > 0.0 { g } else { 0.0 },
             None,
         )
     }
--- a/crates/burn-flex/src/ops/float.rs
+++ b/crates/burn-flex/src/ops/float.rs
@@ -617,20 +617,24 @@ impl FloatTensorOps<Flex> for Flex {
     fn float_clamp_min(tensor: FloatTensor<Flex>, min: Scalar) -> FloatTensor<Flex> {
         let min32 = min.to_f32().unwrap();
         let min64 = min.to_f64().unwrap();
+        // `max` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
         unary::unary_op(
             tensor,
-            move |x: f32| x.max(min32),
-            move |x: f64| x.max(min64),
+            move |x: f32| if x.is_nan() || x > min32 { x } else { min32 },
+            move |x: f64| if x.is_nan() || x > min64 { x } else { min64 },
         )
     }
 
     fn float_clamp_max(tensor: FloatTensor<Flex>, max: Scalar) -> FloatTensor<Flex> {
         let max32 = max.to_f32().unwrap();
         let max64 = max.to_f64().unwrap();
+        // `min` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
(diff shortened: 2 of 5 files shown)


crates/burn-flex/src/ops/activation.rs
```diff
@@ -21,7 +21,13 @@ use crate::{Flex, FlexTensor, Layout};
...
     fn relu(tensor: FloatTensor<Flex>) -> FloatTensor<Flex> {
-        unary_op(tensor, |x: f32| x.max(0.0), |x: f64| x.max(0.0))
+        // `max` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
+        unary_op(
+            tensor,
+            |x: f32| if x.is_nan() || x > 0.0 { x } else { 0.0 },
+            |x: f64| if x.is_nan() || x > 0.0 { x } else { 0.0 },
+        )
     }
```
antimora, line 33: Can rewrite this into a branchless loop? Like using this trick:

```rust
unary_op(
    tensor,
    |x: f32| if !(x <= 0.0) { x } else { 0.0 },
    |x: f64| if !(x <= 0.0) { x } else { 0.0 },
)
```

LLVM lowers this to a single cmpnless/ucomiss plus a select (andps/blendvps), not a jump

crates/burn-flex/src/ops/float.rs
```diff
@@ -617,20 +617,24 @@ impl FloatTensorOps<Flex> for Flex {
...
         let min64 = min.to_f64().unwrap();
+        // `max` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
         unary::unary_op(
             tensor,
-            move |x: f32| x.max(min32),
-            move |x: f64| x.max(min64),
+            move |x: f32| if x.is_nan() || x > min32 { x } else { min32 },
+            move |x: f64| if x.is_nan() || x > min64 { x } else { min64 },
         )
```
antimora, line 626: same
```rust
unary::unary_op(
    tensor,
    move |x: f32| if !(x <= min32) { x } else { min32 },
    move |x: f64| if !(x <= min64) { x } else { min64 },
)
```

crates/burn-flex/src/ops/float.rs
```diff
@@ -617,20 +617,24 @@ impl FloatTensorOps<Flex> for Flex {
...
+        // `min` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
         unary::unary_op(
             tensor,
-            move |x: f32| x.min(max32),
-            move |x: f64| x.min(max64),
+            move |x: f32| if x.is_nan() || x < max32 { x } else { max32 },
+            move |x: f64| if x.is_nan() || x < max64 { x } else { max64 },
         )
     }
```
antimora, line 639: Same.

```rust
unary::unary_op(
    tensor,
    move |x: f32| if !(x >= max32) { x } else { max32 },
    move |x: f64| if !(x >= max64) { x } else { max64 },
)
```

antimora (changes requested):

crates/burn-flex/src/ops/activation.rs
Liberxue: Thanks — that spelling is the one I would prefer too, but `cargo clippy -- --deny warnings`, which `cargo run-checks` runs, rejects it:

```
error: the use of negated comparison operators on partially ordered types produces code that is
hard to read and refactor, please consider using the `partial_cmp` method instead
  --> crates/burn-flex/src/ops/activation.rs:28:25
```

`clippy::neg_cmp_op_on_partial_ord` fires on all eight closures (relu, relu_backward, clamp_min, clamp_max, f32 and f64 each).

The good news is that the goal is already met. Compiling both spellings side by side with `-O -C target-cpu=native`, LLVM folds them into one symbol:

```
relu_suggested = relu_mine
```

Both lower to `vcmpnleps` + `vandps` over `ymm` registers — the branchless select you described, no jump. So this is purely about which spelling passes lint. I have kept `x.is_nan() || x > 0.0` and noted the equivalence in a comment. Happy to switch to `!(x <= 0.0)` with a targeted `#[allow(clippy::neg_cmp_op_on_partial_ord)]` if you would rather have the shorter form.

One behavioural note on the clamp pair: with `is_nan()` first, a NaN *bound* propagates (`clamp_min(5.0, NaN)` gives NaN, as PyTorch does), while `!(x <= min)` would return 5.0. NaN inputs behave the same either way. Mentioning it since #5615 tracks clamp bounds separately.

crates/burn-flex/src/ops/float.rs
Liberxue: Same as above — `clippy::neg_cmp_op_on_partial_ord` denies the negated form under `--deny warnings`, and the `is_nan()` spelling compiles to the identical symbol, so it is already branchless. Details in the relu thread, including the note about NaN bounds.

crates/burn-flex/src/ops/float.rs
Liberxue: Same as above — `clippy::neg_cmp_op_on_partial_ord` denies the negated form under `--deny warnings`, and the `is_nan()` spelling compiles to the identical symbol, so it is already branchless. Details in the relu thread, including the note about NaN bounds.

Liberxue:
Pushed two commits addressing the review.

**`cf9a2c5` — tests moved.** The reviewer is right and my earlier reasoning in the description was wrong: both flex ops files carry an explicit rule to put public-behaviour tests in `burn-backend-tests`, and `ops/maxmin.rs` already shows how to gate a backend-specific one. The tests now live in `tensor/float/activation/relu.rs` and `tensor/float/ops/clamp.rs`.

I measured rather than guessed at the gating, running each assertion on three backends here:

| op | flex | ndarray | cuda |
| --- | --- | --- | --- |
| `relu` | pass | pass | pass |
| `clamp_min` | pass | pass | pass |
| `clamp_max` | pass | pass | pass |
| `clamp` (two-sided) | pass | pass | **fail** |

So `relu`, `clamp_min` and `clamp_max` are ungated. Two-sided `clamp` still maps NaN to a bound on the cube backends — reproduced on CUDA — so it is gated to `flex`/`ndarray` with a comment, as are the f64 variants, following the `maxmin.rs` precedent. Happy to open a separate issue for the cube-backend `clamp` gap if that is useful.

**`204cfec` — `relu_backward`.** Flagged in the automated review and it holds up: flex masked on `out > 0.0`, false for NaN, so a NaN output had its gradient zeroed, while the trait default masks with `float_lower_equal_elem(output, 0)` and keeps it. With the forward pass now propagating NaN, the backward pass was still swallowing it. Test added in `autodiff/relu.rs`, ungated since flex is the only backend overriding `relu_backward`.

This one is beyond the letter of #5609, so say the word and I will drop that commit into its own PR.

Validation, same as before: `cargo run-checks` steps run individually because `cargo audit` cannot reach the RustSec advisory database from here — fmt, clippy with `--deny warnings`, typos, no-std, and the Flex suites in release: 4103 passed, 0 failed. Plus the ndarray and CUDA runs above.

crates/burn-flex/src/ops/activation.rs
```diff
@@ -21,7 +21,13 @@ use crate::{Flex, FlexTensor, Layout};
 
 impl ActivationOps<Flex> for Flex {
     fn relu(tensor: FloatTensor<Flex>) -> FloatTensor<Flex> {
-        unary_op(tensor, |x: f32| x.max(0.0), |x: f64| x.max(0.0))
+        // `max` returns the non-NaN operand, which would map NaN to the bound.
+        // Test `is_nan` first so NaN propagates, as PyTorch does.
+        unary_op(
+            tensor,
+            |x: f32| if x.is_nan() || x > 0.0 { x } else { 0.0 },
+            |x: f64| if x.is_nan() || x > 0.0 { x } else { 0.0 },
```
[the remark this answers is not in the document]
Liberxue: Confirmed and fixed in 204cfec. Flex masked on `out > 0.0`, false for NaN, while the trait default masks with `float_lower_equal_elem(output, 0)` and keeps the gradient. Regression test in `autodiff/relu.rs`; reverting the fix makes it fail in both the base and checkpointing variants.

crates/burn-flex/src/ops/activation.rs
```diff
@@ -1517,4 +1523,21 @@ mod tests {
         let gamma = flex_f64(vec![1.0; 4], &[4]);
         let _ = crate::ops::activation::layer_norm(t, gamma, None, 1e-5);
     }
+
+    // Regression test for #5609. `f32::max` / `f64::max` return the non-NaN
+    // operand, so relu used to map NaN to 0 and hide divergence during training.
+    #[test]
+    fn test_relu_propagates_nan() {
```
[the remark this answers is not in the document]
Liberxue: Right, and the placement rule is explicit in the comment above the test module. Moved in cf9a2c5. Gating was measured rather than assumed: relu passes on flex, ndarray and cuda, so it is ungated; the f64 variant follows the `maxmin.rs` precedent.

crates/burn-flex/src/ops/float.rs
```diff
@@ -1327,4 +1331,65 @@ mod tests {
         let t = Flex::float_random(shape, dist, &device, FloatDType::F16);
         assert_eq!(t.dtype(), DType::F16);
     }
+
+    // Regression tests for #5609. `f32::max` / `f32::min` return the non-NaN
+    // operand, so clamp_min / clamp_max used to replace NaN with the bound while
+    // `float_clamp` (built on `f32::clamp`) propagated it.
+    #[test]
+    fn test_float_clamp_min_propagates_nan() {
```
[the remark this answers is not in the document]
Liberxue: Moved in cf9a2c5. `clamp_min` and `clamp_max` pass on flex, ndarray and cuda, so they are ungated. Two-sided `clamp` still maps NaN to a bound on the cube backends — reproduced on CUDA — so that one stays gated to the CPU backends with a comment.

Liberxue:
Filed the cube-backend gap separately: #5659. On CUDA, `clamp(0.0, 1.0)` turns NaN into 1.0 while `clamp_min` / `clamp_max` keep it, because burn-cubecl only overrides the two-sided op and that one goes through `cubecl::prelude::clamp`. That is why `clamp_nan_propagation` here is gated to flex / ndarray; the gate can come off once #5659 is fixed.

crates/burn-flex/src/ops/activation.rs
laggui: The branchless rewrite isn't necessary then. And as you mentioned, it would change the NaN bound behavior. We should probably add a test (Flex only for now) for NaN bounds in `clamp_min` and `clamp_max`. The existing tests cover NaN input elements; these would cover a NaN scalar bound:

```
// NaN input: only that element remains NaN.
clamp_min([NaN, -1, 5], 0) ->  [NaN, 0, 5]

// NaN bound: every output element becomes NaN.
clamp_min([-1, 0, 5], NaN) -> [NaN, NaN, NaN]
clamp_max([-1, 0, 5], NaN) -> [NaN, NaN, NaN]
```

laggui (approved):
The implementation LGTM.

Pending the NaN bound regression tests mentioned below, but I'll approve in advance.

laggui:
Thanks!

Issue #5602, opened 2026-09-08
burn-flex: u64 right shift is arithmetic; shift widths masked at 64 bits for all int dtypes


antimora:
## Summary

Integer binary ops in burn-flex widen every dtype to `i64` before applying the closure (`int_binary_op` / `int_scalar_op` in `crates/burn-flex/src/ops/binary.rs:357`). Two consequences for shifts:

1. `bitwise_right_shift` on `u64` values above `i64::MAX` is an arithmetic shift.
2. Shift amounts are masked at 64 bits for every dtype, so narrower types do not wrap at their own width.

## Repro

```rust
use burn_backend::ops::IntTensorOps;
use burn_flex::{Flex, FlexTensor};
use burn_backend::{Scalar, TensorData};

let a = FlexTensor::from_data(TensorData::new(vec![u64::MAX], [1]));
let b = FlexTensor::from_data(TensorData::new(vec![1u64], [1]));
let r: Vec<u64> = Flex::bitwise_right_shift(a, b).into_data().try_into_vec().unwrap();
assert_eq!(r[0], u64::MAX >> 1); // got u64::MAX

let a = FlexTensor::from_data(TensorData::new(vec![u64::MAX], [1]));
let r: Vec<u64> = Flex::bitwise_right_shift_scalar(a, Scalar::from(1i64)).into_data().try_into_vec().unwrap();
assert_eq!(r[0], u64::MAX >> 1); // got u64::MAX

let a = FlexTensor::from_data(TensorData::new(vec![1i32], [1]));
let b = FlexTensor::from_data(TensorData::new(vec![33i32], [1]));
let r: Vec<i32> = Flex::bitwise_left_shift(a, b).into_data().try_into_vec().unwrap();
// got 0; ARCHITECTURE.md says shifts use wrapping_shl at type width, which would give 2
```

## Cause

`int.rs:801` does `a.wrapping_shr(b as u32)` on the widened `i64`. For `u64` the cast `a as i64` is negative, so `>>` sign-extends. `int_div` / `int_remainder` already special-case `U64` with `binary_op_typed` for the same reason; the shifts (and `bitwise_*_scalar` on U64) do not.

## Fix

Dispatch shifts per dtype through `binary_op_typed` / `scalar_op_typed` (like `int_div` does for U64) so each type uses its own `wrapping_shl` / `wrapping_shr`. That also fixes the width masking.

PR #5661, opened 2026-09-13
fix(flex): make u64 right shift logical

yuefdev:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. (I didn't run the full `run-checks`, only the burn-flex and Flex backend-test subsets listed under Testing.)
- [x] Made sure the book is up to date with changes in this PR. (No book changes needed. `crates/burn-flex/ARCHITECTURE.md` is updated.)

### Related Issues/PRs

Refs #5602. This fixes part 1 of the issue only (see "Open question" below).

### Changes

`bitwise_right_shift` and `bitwise_right_shift_scalar` went through `int_binary_op` / `int_scalar_op`, which widen every dtype to `i64`. `u64` values above `i64::MAX` turn negative there, so the right shift was arithmetic and filled with ones:

```rust
u64::MAX >> 1          // expected 0x7FFF_FFFF_FFFF_FFFF, got u64::MAX
(1u64 << 63) >> 63     // expected 1, got u64::MAX
```

Both ops now take a dedicated `u64` path (`binary_op_typed` / `scalar_op_typed` with `u64::wrapping_shr`), the same way `int_div` / `int_remainder` already special-case `U64`. The tensor version broadcasts first, like `int_div`. Other dtypes, and `u64` left shift (bit-identical through `i64`), are unchanged.

### Open question: shift-width masking (part 2 of #5602)

I left the "mask at the operand's own width" part out on purpose:

- `test_int_shift_masks_to_64_not_operand_width` pins the current behavior, and #5619 documented it in `ARCHITECTURE.md`.
- For shift amounts in `[bits, 63]`, the current i64 widening already matches PyTorch, which returns `0` for `<<` and sign-fills for `>>` when the shift is ≥ the bit width. For example, `1i32 << 33` gives `0` today, as in PyTorch. Per-dtype `wrapping_shl` would give `2`.
- Where it still differs from PyTorch is shift amounts `>= 64`, since `wrapping_*` masks them back.

If you'd like PyTorch semantics for out-of-range shifts across all dtypes (explicit `0` / sign fill instead of wrapping), I'm happy to do that in a follow-up PR. It would change the pinned test, so I didn't want to do it without agreement.

### Testing

- New tests `test_u64_right_shift_is_logical` (tensor + scalar) and `test_u64_right_shift_broadcast`. Both fail on `main` with the sign-extended values and pass with this change.
- `cargo test -p burn-flex`: 431 passed.
- `cargo test -p burn-backend-tests --features flex --test tensor`: 1669 passed, 20 ignored.
- `cargo fmt --check` and `cargo clippy -p burn-flex --all-targets -- -D warnings`: clean.

AI disclosure: I wrote this fix and the tests with help from Claude Code. I reviewed the change and can explain it.


laggui (approved):
LGTM.

Regarding the open question, we can leave it out. Burn currently has no explicit contract for out-of -range shift counts across backends, so preserving `Flex`'s existing behavior makes sense for this PR.

We can address that separately by documenting the valid range (`0 <= count < dtype bit width`) and treating results outside as unspecified, without promising warpping or zero/sign-fill behavior.

> If you'd like PyTorch semantics for out-of-range shifts across all dtypes

PyTorch [documents right shifts with negative or oversized counts as undefined](https://docs.pytorch.org/docs/2.14/generated/torch.bitwise_right_shift.html), so observed zero/sign-fill results shouldn't be treated as a guaranteed compatibility contract.

PR #5662, opened 2026-09-13
fix(flex): propagate NaN through max pooling

yuefdev:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. (I ran the burn-flex and Flex backend-test subsets listed under Testing, not the full `run-checks`.)

### Related Issues/PRs

Refs #5615. This covers item 1 only (max_pool ignores NaN unless it is the first element in the window). The other items are design questions or overlap with #5658, so I left them alone.

### Changes

The max-pool kernel in `burn-flex/src/ops/pool.rs` only updated the running max when `val > max_val`. That is false for NaN, so a NaN was dropped unless it happened to be the first element of the window:

```
window [1, NaN, 5, 6]  -> 6    (PyTorch: NaN)
window [3, 4, NaN, 8]  -> 8    (PyTorch: NaN)
```

It now uses the same rule as PyTorch's CPU kernel (`aten/src/ATen/native/cpu/MaxPoolKernel.cpp`), `(val > maxval) || isnan(val)`. The output is NaN and the returned index points at the last NaN in the window, so `max_pool2d_with_indices_backward` routes the gradient to that element, as in PyTorch.

All max_pool 1d/2d/3d variants and dtypes share this one kernel (bf16 goes through the f32 path), so this is a one-line change. The only other change is a `num_traits::Float` bound so the generic kernel can call `is_nan()`.

The new backend test is gated to `flex`. burn-ndarray has the same `val > max_val` comparison, and I haven't checked the cube backends.

### Testing

- New `test_max_pool2d_with_indices_nan_propagation` in `burn-backend-tests`. It checks values and indices for F32, F64, F16 and BF16, including a window with two NaNs. It fails on `main` and passes with this change.
- `cargo test -p burn-flex`: 429 passed.
- `cargo test -p burn-backend-tests --features flex --test tensor`: 1670 passed, 20 ignored. The pool tests also pass with `flex-rayon`.
- `cargo fmt --check` and `cargo clippy -p burn-flex --all-targets -- -D warnings`: clean.


laggui (approved):
Thanks!

PR #5663, opened 2026-09-13
fix(flex): round nearest grid_sample ties to even

yuefdev:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. (I ran the burn-flex and Flex backend-test subsets listed under Testing, not the full `run-checks`.)

### Related Issues/PRs

Refs #5615. This covers item 6 only (`grid_sample` nearest mode uses half away from zero, PyTorch uses half to even). #5662 covers item 1.

### Changes

Nearest-mode `grid_sample_2d` chose the source pixel with `px.round()`, which rounds a coordinate that lands exactly between two pixels away from zero. PyTorch uses `std::nearbyint` in `GridSampler.cpp`, and its vectorized CPU kernel uses `Vectorized::round()`. Both round ties to even. So Flex picked a different pixel than PyTorch at every exact `.5`:

| grid (x, y), 2x4 input, align_corners=false | pixel coord | before | PyTorch / after |
| --- | --- | --- | --- |
| (-0.5, -0.5) | x = 0.5 | x = 1 | x = 0 |
| (0.5, -0.5) | x = 2.5 | x = 3 | x = 2 |
| (-0.75, 0.0) | y = 0.5 | y = 1 | y = 0 |

The fix uses `libm::roundeven`. burn-flex already depends on `libm`, and `roundeven` has been available since 0.2.12 while the workspace requires 0.2.15. I used it instead of `f64::round_ties_even` so the `no_std` build keeps working. Bilinear mode is unchanged.

Flex is currently the only backend that implements nearest grid sampling (ndarray and cubecl panic on it), so the new test is gated to `flex`.

### Testing

- New `should_grid_sample_2d_nearest_round_half_to_even` in `burn-backend-tests`, with ties on both axes. It fails on `main` and passes with this change.
- `cargo test -p burn-flex`: 429 passed.
- `cargo test -p burn-backend-tests --features flex --test tensor`: 1670 passed, 20 ignored.
- `cargo check -p burn-flex --no-default-features`, `cargo fmt --check` and `cargo clippy -p burn-flex --all-targets -- -D warnings`: clean.


yuefdev:
The macOS failure is an unexecutable CubeCL build script (exit 126), also seen in #5666. Could the runner or cached build artefacts be checked?

laggui (changes requested):
LGTM! But I'd keep the CI changes out of this PR.

macos failure is unrelated to your changes, I can still merge if red.

laggui (approved):
Thanks for addressing so quickly!

PR #5664, opened 2026-09-13
Fix: burn-store: PyTorch reader drops tensors stored inside lists

### LINKED ISSUE #5595: burn-store: PyTorch reader drops tensors stored inside lists
## Summary

`PytorchReader` only collects tensors reachable through dicts. A tensor inside a list or tuple is dropped with no error and no trace.

```python
torch.save({"weights": [w1, w2]}, "m.pt")
```

`PytorchReader::new("m.pt")` succeeds with `len() == 0`. Through `PytorchStore::apply_to` the user then gets "tensor not found" for every parameter, with nothing indicating the file did hold tensors.

This is `extract_tensors` in `crates/burn-store/src/pytorch/pickle_reader.rs`: the walk descends `Object::Dict` and ignores everything else. Pre-existing behavior; #5593 made unsupported tensor *calls* an error but left this path alone because it needs a naming decision.

## Options

1. Index list elements: `weights.0`, `weights.1`. This matches how burn names `Vec<Module>` parameters and how `nn.ModuleList` keys look in a state_dict, so such a file would load into the natural burn module without remapping.
2. Error naming the path, e.g. "tensor at weights[0] is not reachable through dict keys". Strict, no naming policy needed, but refuses files that option 1 would load.

Option 1 seems right. The remaining question is tuples: same treatment, or ignore.

Surfaced by the review of #5593.

SIDDARTHAREDDY8:
### Checklist

- [ ] Made sure the book is up to date with changes in this PR. (Not applicable: no user-facing behavior change, only a loader fix + regression test.)

> `cargo run-checks` tests with `Flex` by default. Use `cargo run-checks --backend <backend>` when targeting another backend. Consider running additional tests relevant to the crates changed.

### Related Issues/PRs

Fixes #5595.

### Changes

Implemented the issue #5595 fix (option 1) in `crates/burn-store/src/pytorch/pickle_reader.rs`: `extract_tensors` walker now descends into `Object::List` and `Object::Tuple` in addition to `Object::Dict`, naming elements by index (`weights.0`, `weights.1`) via the existing `path.join(.)` mechanism — burn Vec<Module> parameter naming, matching `nn.ModuleList` state_dict keys, so such files load into the natural burn module without remapping. Non-tensor values in lists are still ignored (indices preserved, not renumbered). Added regression test `tensors_inside_lists_and_tuples_get_indexed_names` (CONTRIBUTING requires one for bug fixes) using the repo existing fixture-backed rebuild helpers: asserts `torch.save({"weights": [w1, w2]})` exposes `weights.0`/`weights.1`, a tuple exposes `bias.0`, a dict-in-list exposes `weights.2.w`, and the tensor data round-trips intact ([0.0,1.0,2.0,3.0]).

### Testing

No Rust toolchain existed on this machine; installed stable rustup. Full `cargo test -p burn-store` is not runnable: the crates.io API is 403-blocked through the egress proxy, so the workspace dependency tree cannot be fetched. Instead, compiled the REAL `Object` enum + REAL modified `extract_tensors` (extracted verbatim from the final file) with rustc 1.98.1 against a minimal PackTensor stub, and ran 4 behavioral tests: issue-5595 list scenario (`weights.0`/`weights.2`, int skipped), nested dict-in-list paths (`layers.0.w`, `layers.1.0`), dict-only behavior unchanged (`a.b`), opaque values still ignored. All 4 passed. rustfmt (edition 2024, max_width=100) is clean on the changed file. The in-file regression test itself could not be executed (needs the burn workspace), but it uses only helpers already exercised by existing tests in the same module. `cargo run-checks` checklist item reflects the run-checks steps that could be executed here (fmt + targeted behavioral compile tests); CI will cover the rest.


laggui (approved):
Changes look good to me, thanks!

PR #5665, opened 2026-09-13
fix(burn-ndarray): sign(NaN) must be 0, not its hidden sign bit

SRaswan:
### Checklist

- [ X ] Confirmed that `cargo run-checks --backend ndarray` has been executed.
- [ X ] Made sure the book is up to date with changes in this PR.

### Related Issues/PRs

Regression introduced by #4573 ("move sign back to mathOps"). Fund with an external differential fuzzer ([burnmark](https://github.com/SRaswan/burnmark)) that compares this backend against burn's LibTorch on randomly generated autograd programs.

### Changes

`sign_op` in burn-ndarray returned 1 or -1 for NaN input instead of 0.

`num_traits`'s `Signed::is_positive()` for floats is `is_sign_positive()`, which is just a raw sign-bit check, not an actual numeric comparison. For NaN, it returns true or false depending on whichever sign bit the NaN carries in passing, not the `sign(NaN) == 0` convention (`f32::signum()`, PyTorch).

Some cases where this matters is something like `abs()`'s backward pass, which is `grad * sign(input)`, so backward through any NaN produces a finite gradient instead of propagating the NaN. 

Example program: 
```
x = -0.5 
x.log().abs().backward() 
grad = -2.0
```
However, LibTorch is -0.0.

Fix: restored the comparison logic, adding `PartialOrd` to just this function rather than the whole impl block. This already supports the other natural types (f32, f64, i32, ...).

### Testing

Few tests I did:
1. Added a NaN case to the existing sign() float test (`should_support_sign_ops_float_nan`).
2. Added `should_diff_abs_of_nan`, an autodiff test that triggered the bug on an actual backward path.
3. `cargo run-checks --backend ndarray`
4. Ran the original `abs(log(x))` divergence through the differential fuzzer again against this patch. NdArray now matches LibTorch exactly (-0.0 both sides, was -2.0 NdArray vs -1.0 LibTorch).


laggui (changes requested):
Thanks for addressing this gap. The `NdArray` fix looks good, but these shared tests also run on `Flex`, which currently returns NaN for NaN inputs. That causes the new tests to fail.

Could we update burn-flex too so it's consistent across backends? It would also be useful to document this shared contract and test both NaN sign bits (`-f32::NAN`).

laggui (approved):

PR #5666, opened 2026-09-13
test(pack): check streaming memory with a drop hook, not a global allocator

### LINKED ISSUE #5380: burn-pack: test the streaming memory bound with a Drop hook instead of a global allocator
## Summary

`crates/burn-pack/tests/streaming_memory.rs` verifies the peak-memory guarantee that the deferred writer exists for: `Writer::write_tensors` drops each tensor's `Bytes` before asking the next provider for its own. Without that, saving a large model would hold everything at once.

It measures this with a process-wide `#[global_allocator]` that counts live heap bytes above a 256 KiB threshold, and asserts the count is zero on entry to each provider.

It works, and it fails for the right reason (verified by making the writer retain each `Bytes` — the test reports `tensor t1: 1048576 bytes of an earlier tensor were still live`). But it observes the property one layer out from where it lives.

## The problem with the current approach

- **A size threshold couples the assertion to unrelated allocations.** Anything in the binary allocating ≥ 256 KiB makes the test fail for a reason unrelated to the guarantee. The mitigation is the separate test binary, which is why the file exists at all.
- **"No large allocation is live" is weaker than "the previous `Bytes` was dropped."** It infers the event rather than observing it.
- **It replaces the allocator for a whole binary** to check one loop in `writer.rs`.

## Proposal

`Bytes` supports a custom backing, and it is reachable from burn-pack today — `burn-std` glob re-exports `cubecl_environment::bytes::*`, so `burn_std::AllocationController` and `Bytes::from_controller` are both in scope. (Verified: a `Box<dyn burn_std::AllocationController>` compiles against burn-pack with no new dependency.)

A test controller wrapping a `Vec<u8>` with a `Drop` impl that pushes into a shared log records exactly the event under test:

```rust
struct Tracked { data: Vec<u8>, dropped: Arc<Mutex<Vec<usize>>>, id: usize }

impl Drop for Tracked {
    fn drop(&mut self) { self.dropped.lock().unwrap().push(self.id); }
}
// impl AllocationController for Tracked { ... }
// Bytes::from_controller(Box::new(Tracked { .. }), len)
```

Then a provider for tensor *N* asserts that tensor *N-1*'s id is already in the log. That is the guarantee stated directly, with:

- no global state and no threshold heuristic
- no separate test binary — it can live in a `#[cfg(test)] mod tests` inside `writer.rs`, next to the loop it constrains
- no dependence on allocation sizes

Cost: one `unsafe` block (`from_controller` is unsafe) and four trait methods, comparable to the `unsafe impl GlobalAlloc` it replaces.

## Not urgent

The current test is correct and proven to catch the regression. This is about putting the assertion at the layer the guarantee lives, so it stays robust as the writer changes.

Two reviewers disagreed on this during #5219 review — one judged the global allocator proportionate after checking alternatives, the other argued for the local hook. Filing so the decision is recorded rather than re-litigated.

Context: #5219.

yuefdev:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. (I ran the burn-pack checks listed under Testing, not the full `run-checks`.)

### Related Issues/PRs

Closes #5380. Context: #5219.

### Changes

This implements the "local hook" option from #5380. The streaming guarantee is that `write_tensors` drops each deferred tensor's bytes before it asks the next provider for its own. `tests/streaming_memory.rs` checked that with a process-wide counting `#[global_allocator]` and a 256 KiB threshold, in its own test binary. It is replaced by a `#[cfg(all(test, feature = "std"))] mod tests` in `writer.rs`, next to the loop:

- Each deferred tensor's bytes are backed by a small owner type whose `Drop` pushes the tensor's index into a shared log.
- Each provider asserts that the log already holds every earlier index, in write order, before producing its own bytes.
- After the write, the log must hold all indices. That also guards against a vacuous pass, since an owner only exists once its provider has run.

One difference from the sketch in the issue: I didn't implement `AllocationController` with `Bytes::from_controller`. The owner goes through `bytes::Bytes::from_owner` and `Bytes::from_shared` instead. That is safe code, with no `unsafe` block and no trait methods to implement. `burn-std` always enables cubecl's `shared-bytes` feature, and burn-ndarray's tests already build `Bytes` this way. The only new thing is `bytes` as a dev-dependency of burn-pack, and it is already a workspace dependency.

The test now runs the check through all three sinks: `write_to_file` (as before), `write_to_file_atomic` (the path `BurnpackStore` uses) and `into_bytes`.

Note that a `from_shared` backing supports `Bytes::view`, so these tensors take the chunked path in `write_tensor_data` rather than the single-shot path a plain `Vec` takes. The drop point being checked is the same for both: `data` goes out of scope at the end of each loop iteration.

### Testing

- I checked that the test catches the regression. With `write_tensors` changed to push each `data` into a `Vec` instead of dropping it, it fails with `tensor t1 was requested while an earlier tensor's bytes were still live`. It passes on the unmodified writer.
- `cargo test -p burn-pack`: all pass.
- `cargo check -p burn-pack --no-default-features`, `cargo fmt --check` and `cargo clippy -p burn-pack --all-targets -- -D warnings`: clean.


yuefdev:
The macOS job stops because the cubecl-environment build script cannot execute (exit 126). The log points to a runner or build-artifact issue: https://github.com/tracel-ai/burn/actions/runs/34780498011

laggui (approved):

Issue #5506, opened 2026-08-27
PyTorch reader: broadcast views in untrusted checkpoints can exhaust memory


antimora:
### Describe the bug

A `.pt` whose metadata declares a large logical shape backed by a zero-stride view over a tiny storage makes `burn-store` materialize the full logical tensor. Nothing bounds the result, so loading the file drives the process into RAM exhaustion.

Reproducer: shape `(2^21, 2^21)`, stride `(0, 0)`, offset 0, over a 32-element `FloatStorage`. The storage extent is 1, so the bounds check in `rebuild_tensor_impl` passes. `num_elements` is 2^42, and the gather in `reorder_tensor_bytes` then tries to build a 17.6 TB buffer.

As a unit test in `pickle_reader.rs` (do not run it unattended, it will eat your RAM):

```rust
let object = rebuild_tensor(
    rebuild_args("FloatStorage", "0", 32, 0, &[1 << 21, 1 << 21], &[0, 0]),
    &mut HashMap::new(),
    &fixture_data_source(),
)
.unwrap();
let _ = bridge::into_data(tensor_from_object(object));
```

I stopped a local run at 100% CPU and 34% of system RAM.

### Why the existing guard does not catch it

#5392 added `try_reserve_exact` for the reorder buffer. That is the right shape of fix but it does not fire on macOS, or on Linux with overcommit:

```
PROBE_RESERVE: SUCCEEDED for 17592186044416 bytes (cap=17592186044416)
```

`try_reserve_exact` only reports what the allocator rejects at reservation time. Address space is handed back lazily, so the failure moves to first touch, which here is the per-element `extend_from_slice` inside the loop. The process does not abort, it grows RSS until something kills it.

### Why the obvious check is wrong

`num_elements <= data.len() / element_size` does not work. `expand` legitimately produces more logical elements than stored ones: the `expanded` tensor in `non_contiguous.pt` has 6 logical elements over 3 stored. There is no bound derivable from view metadata alone.

### Prior state

This predates #5392. `main` reaches the same state through `values.resize(num_elements, 0.0)` in the dtype arms, so #5392 changed the failure mode (abort to slow exhaustion) rather than the severity.

### Options

1. A configurable ceiling on materialized tensor bytes, defaulting to something derived from the archive size.
2. Commit the reorder buffer up front (`resize` rather than `reserve`) so allocator failure surfaces before the loop instead of during it.
[...]

PR #5667, opened 2026-09-14
fix(burn-dataset): cap MNIST item counts at the split size

### LINKED ISSUE #5651: burn-dataset: a hostile MNIST header can request a 3.3 TB allocation
`MnistDataset::read_images` sizes its buffer from a count it reads out of the file, before reading any payload:

```rust
let size = u32::from_be_bytes(buf);                                        // L184 -- from the file
let mut buf_images: Vec<u8> = vec![0u8; WIDTH * HEIGHT * (size as usize)]; // L186
```

`WIDTH * HEIGHT` is 784, so a `size` of `u32::MAX` asks for 3,367,254,359,280 bytes (~3.29 TiB) up front. `read_labels` (L212-214) has the same shape with a 4 GiB ceiling.

I reproduced it with the body of `read_images` copied verbatim into a small binary:

| input | result |
|---|---|
| 4,112-byte file, header declares `u32::MAX` items | `memory allocation of 3367254359280 bytes failed` → process aborts |

It is an allocation error, not a panic. `catch_unwind` cannot intercept it, and nothing reaches the caller.

**Why the header is untrusted.** `download_file` skips the network whenever the file already exists, so a poisoned cache directory gets used as-is. On the normal path, anything that can influence the gunzipped response (mirror, proxy, TLS interception) controls these bytes.

**The repo already has the fix pattern.** `burn-store/src/pytorch/storage.rs` bounds pre-allocations with `read_exact_len(len, capacity_bound)`, and its doc comment says it outright: *"A length read from an untrusted file could be anything..."*. #5506 fixed the same class of bug in the PyTorch reader.

Proposed patch, keeping the signature and moving the failure before the allocation:

```rust
let expected_len = (WIDTH * HEIGHT) as u64 * size as u64;
let available = f.metadata().unwrap().len().saturating_sub(16);
assert!(
    expected_len <= available,
    "image file declares {size} items ({expected_len} bytes), but only {available} bytes follow the header"
);
let mut buf_images: Vec<u8> = vec![0u8; expected_len as usize];
```

Checked both ways: the hostile file now panics with that message and no large allocation, and an honest file still loads as before.

Opening this before a PR as CONTRIBUTING asks. Disclosure: I used AI assistance to review the code and write the probe, and I can explain every line of the change. Happy to send it as a PR if this looks worth taking.

li-jin-quan:
### Checklist

      Not run — it needs `typos`, `cargo-audit` and the workspace backend-test suite (CUDA/libtorch/wgpu). I did run the crate's own suite on the msvc target; see Testing.
      No public API or behaviour change for well-formed files, so nothing to document.

### Related Issues/PRs

Fixes #5651.

### Changes

`MnistDataset::read_images` sizes its buffer from a count it reads out of the file, before reading any payload:

```rust
let size = u32::from_be_bytes(buf);                                        // from the file
let mut buf_images: Vec<u8> = vec![0u8; WIDTH * HEIGHT * (size as usize)]; // L186
```

`WIDTH * HEIGHT` is 784, so a `size` of `u32::MAX` asks the allocator for ~3.29 TiB up front. `read_labels` has the same shape with a 4 GiB ceiling.

The failure is an allocation error, not a panic, so `catch_unwind` cannot intercept it and nothing reaches the caller. For a corrupt or truncated file this is a hard abort instead of a recoverable error.

#### The length check alone is not a bound

The first version of this PR compared the implied payload against the length the file reports:

```rust
let expected_len = (WIDTH * HEIGHT) as u64 * size as u64;
let available = f.metadata().unwrap().len().saturating_sub(16);
assert!(expected_len <= available, "...");
```

That check is worth keeping — it catches a truncated file — but it cannot bound an allocation, because `metadata().len()` returns the logical length, and a sparse file reports whatever length it was given while occupying a few blocks.

Measured on an NTFS file, `FSCTL_SET_SPARSE` set, 16-byte header declaring `u32::MAX`, then extended to 3.4 TB:

| what | value |
|---|---|
| `metadata().len()` | 3,400,000,000,000 bytes |
| allocated on disk | 131,072 bytes |
| implied payload (`784 * u32::MAX`) | 3,367,254,359,280 bytes |
| `expected_len <= available` | **passes** |

So the assertion holds and the next statement is a 3.3 TB zeroed allocation. On Linux the same file is one `truncate -s 3T` away.

#### What this PR does now

The count is capped by what the split is documented to hold, and the length check stays on as a truncation check:

```rust
const TRAIN_ITEMS: u64 = 60_000;
const TEST_ITEMS: u64 = 10_000;

fn split_item_ceiling(split: &str) -> u64 {
    if split == "train" { TRAIN_ITEMS } else { TEST_ITEMS }
}

let ceiling = split_item_ceiling(split);
assert!(
    size as u64 <= ceiling,
    "image file declares {size} items, but the {split} split holds at most {ceiling}"
);

let expected_len = (WIDTH * HEIGHT) as u64 * size as u64;
let available = f.metadata().unwrap().len().saturating_sub(16);
assert!(
    expected_len <= available,
    "image file declares {size} items ({expected_len} bytes), but only {available} bytes follow the header"
);
```

`read_labels` gets the same two checks against its 8-byte header. Signatures are unchanged and no dependency is added.

A ceiling taken from the dataset definition is the only thing that holds here: the length comes from the filesystem and the count comes from the file, so neither one is trusted input. The largest payload the ceiling admits is 60,000 × 784 = 47 MB, which is the train split itself.

This follows the pattern already in the repo: `burn-store/src/pytorch/storage.rs` bounds pre-allocations with `read_exact_len`, and its comment says a length read from an untrusted file could be anything. #5506 fixed the same class of bug in the PyTorch reader.

### Testing

Six tests in `mnist.rs`:

- `read_images_rejects_a_count_past_the_split_ceiling` — the `u32::MAX` header
- `read_labels_rejects_a_count_past_the_split_ceiling` — same for labels
- `read_images_rejects_a_count_the_file_cannot_hold` — within the ceiling, but more than the file holds
- `read_labels_rejects_a_count_the_file_cannot_hold`
- `read_images_accepts_a_well_formed_file` — the checks must not reject an honest file
- `read_images_accepts_the_test_split` — 10,000 items stays under the ceiling

Both directions checked: with the ceiling assertion removed, `read_images_rejects_a_count_past_the_split_ceiling` fails, with the panic landing on the length check instead — so the test pins the ceiling and not the length check.

**Also ran the crate's own suite, on the target CI builds:**

```
cargo test -p burn-dataset --features vision --lib    # x86_64-pc-windows-msvc
test result: ok. 70 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

`cargo fmt -p burn-dataset -- --check` is clean, and clippy says nothing about `vision/mnist.rs`; the handful of warnings in the crate are pre-existing and elsewhere.

**What is still missing, and why.** `cargo run-checks` is out of reach here — though not for the reason I first gave in this PR, which I got wrong. Under the hood it runs `typos` and `cargo-audit` (neither installed), then clippy across the workspace, then the full workspace release backend-test suite, which needs CUDA, libtorch and a wgpu-capable GPU. So the box below stays unchecked: under-claiming beats ticking something I did not earn. CI remains the first full-workspace run, and I will fix whatever it finds.

Disclosure: I used AI assistance to review the code and write the tests. I understand every line and can explain it during review.


laggui (approved):

li-jin-quan:
Thanks for merging — and for sorting out the fmt. Appreciated.

PR #5674, opened 2026-09-14
Expose Tolerance fields

crutcher:

Client code can't currently read tolerance.


laggui (changes requested):
Could we expose accessors instead of making the fields public? That would let clients read the tolerances while preserving the validation currently enforced by the constructors and setters.

Also, the unrelated unfold test should be removed from this PR. Feel free to open a separate issue for that, and accompanying PR if you have a fix.

PR #5676, opened 2026-09-14
perf(nn): use a dedicated BatchNorm training op with closed-form backward

Marc-AnthonyG:
### Changes

BatchNorm in training mode computed the batch mean and variance out of
tensor operations, then normalised with `batch_norm`. On an autodiff
backend that records a reduction, a broadcast and a square per statistic,
and the backward replays each of them over the full activation.

`ModuleOps::batch_norm_train` computes the statistics inside the op and
returns them with the output, so the module still updates its running
statistics; the autodiff backend differentiates it as a single node
through `batch_norm_train_backward`, the closed-form batch-norm gradient.
The default implementation is the previous primitives plus `batch_norm`,
so backends need not implement anything.


crates/burn-backend/src/backend/ops/modules/base.rs
```diff
@@ -128,6 +155,99 @@ pub trait ModuleOps<B: Backend> {
...
+        let channels = shape[1];
+        let count = (shape.num_elements() / channels) as f64;
+        let reduced = every_dim_but_the_channels(shape.num_dims());
+
+        let mean = B::float_div_scalar(B::float_sum_dims(x.clone(), &reduced), count.into());
+        let centered = B::float_sub(x.clone(), mean.clone());
+        let variance = B::float_div_scalar(
+            B::float_sum_dims(B::float_mul(centered.clone(), centered), &reduced),
+            count.into(),
+        );
```
laggui: We should preserve the numerical stability of the previous `mean_dim` implementation. Computing `sum_dims` / `count` can overflow in lower precision (e.g. f16) before division. With `float_mean_dim` the backends use higher precision accumulation dtypes.

crates/burn-backend/src/backend/ops/modules/base.rs
```diff
@@ -128,6 +155,99 @@ pub trait ModuleOps<B: Backend> {
...
+        // The batch statistics depend on every element of the input, which is
+        // what the two centred terms account for.
+        let centred_grad = B::float_sub(
+            output_grad,
+            B::float_div_scalar(beta_grad.clone(), count.into()),
+        );
+        let projected = B::float_mul(
+            normalized,
+            B::float_div_scalar(gamma_grad.clone(), count.into()),
+        );
```
laggui: Similar comment here. We should compute the averaged terms with `float_mean_dim`, using the same per-channel reshape as the forward pass.

crates/burn-tensor/src/tensor/module.rs
```diff
@@ -47,6 +47,37 @@ pub fn batch_norm<const D: usize>(
...
+///
+/// `input` has shape `[batch, channels, ...]`; `gamma` and `beta` have shape
+/// `[channels]`. On an autodiff device this is a single operation with the
+/// closed-form batch-norm gradient.
+pub fn batch_norm_train<const D: usize>(
+    input: Tensor<D>,
+    gamma: Tensor<1>,
+    beta: Tensor<1>,
+    epsilon: f64,
+) -> (Tensor<D>, Tensor<1>, Tensor<1>) {
```
laggui: I think it would be more explicit to return a public `BatchNormTrainOutput<D>` (similar to the backend output struct) instead of a tuple. It makes the three outputs explicit (each field is named) and discoverable.

crates/burn-tensor/src/tensor/module.rs
```diff
@@ -47,6 +47,37 @@ pub fn batch_norm<const D: usize>(
     )))
 }
 
+/// Applies batch normalization with the statistics of the batch itself, and
+/// returns the normalized input along with the batch mean and the biased batch
+/// variance, both per channel.
+///
+/// `input` has shape `[batch, channels, ...]`; `gamma` and `beta` have shape
+/// `[channels]`. On an autodiff device this is a single operation with the
+/// closed-form batch-norm gradient.
```
laggui: I think we should expand the docs a bit to make the autodiff behavior clear and explicit.

e.g.

```rust
/// Applies batch normalization using statistics computed from the input batch.
///
/// `input` has shape `[batch, channels, ...]`; `gamma` and `beta` have shape
/// `[channels]`.
///
/// Returns the normalized input and the batch mean and biased (population)
/// variance, both with shape `[channels]`. Statistics are computed over every
/// dimension except the channel dimension. The returned variance excludes
/// `epsilon`.
///
/// This function does not update running statistics. For normalization using
/// explicitly supplied statistics, use [`batch_norm`].
///
/// # Autodiff
///
/// This function can be used with or without autodiff. It does not enable
/// gradient tracking.
///
/// On an autodiff-enabled device, gradients through the normalized output
/// account for the dependence of the batch statistics on the input. The returned
/// mean and variance remain on the same device but are detached.
```

laggui (changes requested):
LGTM overall, just a couple of improvements we could make

PR #5679, opened 2026-09-15
perf(autodiff): avoid redundant traversal in checkpoint topological sort

haydenflinner:
Memoize visited nodes with a HashSet instead of Vec::contains, and expand each ancestor once instead of per-branch. Adds a perf repro test for demonstration, which can probably be deleted before merge?

### Checklist

- [✅] Confirmed that `cargo run-checks` has been executed.
- [✅] Made sure the book is up to date with changes in this PR.

### Changes

Swap from a vector to a hashset to avoid searching through again and again during checkpointing. I ran into this while implementing an LSTM-CTC model.

### Testing

My actual model results:
```
┌───────┬─────────────────┬────────┬─────────┐
│ steps │     before      │ after  │ speedup │
├───────┼─────────────────┼────────┼─────────┤
│ 20    │ 0.267s          │ 0.155s │ 1.7x    │
├───────┼─────────────────┼────────┼─────────┤
│ 80    │ 2.81s           │ 0.730s │ 3.9x    │
├───────┼─────────────────┼────────┼─────────┤
│ 320   │ 46.6s           │ 4.619s │ 10.1x   │
├───────┼─────────────────┼────────┼─────────┤
│ 1280  │ (didn't finish) │ 46.1s  │ —       │
└───────┴─────────────────┴────────┴─────────┘
```

The included similar reproducer results:
```
Results

Ran the same unrolled-recurrence shape (5 layers × T timesteps × sigmoid/tanh/mul/add gates, batch=16/hidden=64) through BalancedCheckpointing (the .gradient_checkpointing() path that exercises topological_sort), before and after, in release mode:

┌──────────┬────────────────────┬───────────────┬─────────┐
│  steps   │ before (quadratic) │  after (fix)  │ speedup │
├──────────┼────────────────────┼───────────────┼─────────┤
│ 20       │ 0.010s             │ 0.010s        │ —       │
├──────────┼────────────────────┼───────────────┼─────────┤
│ 80 (4x)  │ 0.068s (6.8x)      │ 0.044s (4.4x) │ 1.5x    │
├──────────┼────────────────────┼───────────────┼─────────┤
│ 320 (4x) │ 1.655s (24.3x)     │ 0.360s (8.2x) │ 4.6x    │
└──────────┴────────────────────┴───────────────┴─────────┘
```


crates/burn-backend-tests/tests/autodiff/perf_repro.rs
laggui, line 1: This shouldn't be part of the regular test suite without asserting correctness or performance. We could move it under the `benches/`, but in this case I think it's mainly a repro for the current changes, so I think I'd just leave it out entirely.

laggui (changes requested):
The traversal change looks correct and appropriately scoped.

LGTM, with the perf reproducer removed as suggested below

laggui (approved):
Thanks!

PR #5684, opened 2026-09-15
fix(train): handle tied scores in average precision

### LINKED ISSUE #5683: AucPrMetric reports order-dependent Average Precision for tied scores
## Describe the bug

`AucPrMetric` can report different Average Precision values for the same samples in a different order when prediction scores tie. In a two-sample example, it reports either 100% or 50% even though both predictions are 0.5.

Reproduced on `b7946a625c56f65e5f51d435c60cb204657141ec`, Linux x86_64, Flex.

## Reproduction

```rust
use burn_core::{
    data::dataloader::Progress,
    tensor::{Device, Tensor},
};
use burn_train::metric::{
    AucPrMetric, ConfusionStatsInput, Metric, MetricMetadata, Numeric,
};

fn main() {
    let device = Device::flex();
    let metadata = MetricMetadata {
        progress: Progress {
            items_processed: 2,
            items_total: 2,
            unit: None,
        },
        iteration: Some(0),
        lr: None,
    };
    for labels in [[[true], [false]], [[false], [true]]] {
        let input = ConfusionStatsInput::new(
            Tensor::from_floats([[0.5], [0.5]], &device),
            Tensor::from_data(labels, &device),
        );
        let mut metric = AucPrMetric::binary();
        metric.update(&input, &metadata);
        metric.compute();
        println!("{:?}: {}%", labels, metric.final_value().current());
    }
}
```

Actual results:

```text
[[true], [false]]: 100%
[[false], [true]]: 50%
```

Both should be 50%. A score threshold includes both tied samples together: precision is 1/2 when recall reaches 1. I also checked `sklearn.metrics.average_precision_score` in scikit-learn 1.9.0, which returns 0.5 for both orders.

## Cause and proposed fix

`average_precision` in `crates/burn-train/src/metric/auc_pr.rs` sorts scores, then evaluates cumulative precision at every sample rank. That treats tied scores as distinct thresholds and makes the result depend on their ordering.

Compute precision at the end of each equal-score group, using the whole group when assigning its recall contribution. This should make results invariant to sample order and agree with the documented threshold-based Average Precision definition, including macro and micro reductions.

I am preparing a focused fix with regression tests for tied groups, batch boundaries, and an independent reference calculation.

Mikyx-1:
### Changes

With predictions `[0.5, 0.5]`, `AucPrMetric` reported 100% AP for labels `[1, 0]` but 50% after reversing the samples. Both orders now return 50%, since a threshold includes all equally scored samples together.

Compute each positive sample's precision at the end of its equal-score group. A reverse cumulative minimum propagates group-end indices, then a gather selects the appropriate precision. The calculation stays in tensor operations and works per column before macro reduction or after flattening for micro reduction.

Also correct the existing multilabel micro fixture: its repeated scores across columns give AP `0.550135118149824`, independently verified with scikit-learn, rather than the previous rank-dependent value.

### Related Issues/PRs

Fixes #5683.

### Testing

- `cargo run-checks`: passed formatting, typos, dependency audit, workspace lint, no-std checks, and 3,776 backend/linalg/transfer tests (43 existing skips).

- Confirmed the new two-sample and exhaustive-reference regressions fail before the fix.
- Added exhaustive coverage of 1,215 four-sample score/label combinations against an independent threshold-counting reference, plus order/batch invariance, macro/micro ties, singleton, and no-positive cases.
- `cargo test -p burn-train --lib`: 220 passed.
- `cargo test -p burn-train --no-default-features --lib`: 180 passed.
- `cargo clippy -p burn-train --no-default-features --lib --tests`: passed (existing unused-method warning in `learner/summary.rs`).
- Additional local public-API comparison against scikit-learn 1.9.0: 600 cases, seed 5683, 1–64 samples, 1–6 classes, reversed rows, random batch splits, and absent-positive classes. Maximum absolute AP difference: `6.75e-7`.

### Checklist

- [x] Confirmed that `cargo run-checks` has been executed successfully.
- [x] Made sure the book is up to date with changes in this PR; no book changes are needed for this metric correctness fix.


laggui (approved):
Thanks for reporting & addressing the issue. Changes LGTM!

PR #5688, opened 2026-09-15
feat(tensor): device identity and one entry per physical GPU

ThierryCantin-Demers:
`Device::enumerate(DeviceType::Cuda | DeviceType::Vulkan)` lists every NVIDIA card twice, once per runtime, with nothing saying they are the same card. This PR adds a view with one entry per physical GPU.

## Changes

- **`Device::enumerate_physical()`** returns one `PhysicalGpu { physical, name, devices }` per card, with every device that runs on it. Devices are grouped with cubecl's `PhysicalDevice::is_same_card`, and `PhysicalDevice::fill_from` keeps the PCI address and LUID from whichever runtime reported them. A card no runtime can key, such as an Apple GPU, is an entry of its own; a device that is no card, such as the CPU, is left out. Available when the build enables a cubecl runtime.
- **`Device::identity()`** returns cubecl's `DeviceIdentity` for a device, `None` for backends that have none. An autodiff device answers for the device it wraps.
- `Devices` derives `Debug` and `Clone`; `AdapterLuid`, `PciAddress`, `PciVendor` and `PhysicalDevice` are re-exported.
- The book's Enumerating Devices section covers it.

## Testing

On a machine with two NVIDIA and two AMD GPUs, the ignored test `a_card_reached_by_two_runtimes_is_listed_once` finds each NVIDIA card once, with both its CUDA and Vulkan device, and no card listed twice, under both `--features cuda,wgpu` and `cuda,vulkan`. It ran against main's cubecl, which includes tracel-ai/cubecl#1652.

## Not included

Picking a wgpu adapter by PCI address directly (take the Vulkan device from the card's `devices` instead), and live data such as utilization or free memory.


crates/burn-tensor/src/device.rs
laggui: Just need to update the feature gating here since `String` is also used with `PhysicalGpu`

laggui (approved):
LGTM, pending a minor fix.

PR #5702, opened 2026-09-16
feat(core): pipeline trait to split a model by layers across devices

ThierryCantin-Demers:
Burn keeps a whole copy of the model on every device. `SingleDevice`, `MultiDevice` and
`DistributedDataParallel` all replicate the model and split the batch, and `Module::to_device` moves
a model to one device. So a model too large for a single card has no path at all, and DDP makes it
worse rather than better, since it wants a full replica per device.

This adds pipeline parallelism: the model is cut by whole layers, and each device holds a run of
consecutive ones.

## Changes

A model implements `Pipeline` once, stating its forward pass as segments (`forward_input`, then
`forward_block` for each block in order, then `forward_output`) and, in a `PipelineLayout`, which
submodules each segment runs. A module tree does not say what order forward runs in, so the layout
is what makes the split expressible.

`PipelinePlacement` gives every segment a device, either evenly across some devices or as stages of
chosen sizes. `place` forks each parameter onto the device of the segment that owns it and returns a
`PlacedPipeline`, whose `forward` runs each segment there and moves the carry between devices.

Built on #5739, which is what makes this useful rather than just possible: a parameter that has not
initialized yet only takes its segment's device, so a model built from a config never exists whole
on one device, and weights loaded after `place` go straight to their own stage. That is what lets a
model too large for one card load at all.

`PlacedPipeline` is itself a `Module` and dereferences to the model, so records, devices, training
flags and the model's own methods work as they do on one device.

Two choices worth flagging for review:

- **`Deref` on `PlacedPipeline`** is against the Rust API guidelines, which reserve it for smart
  pointers. Without it, placing a model costs you every inherent method you wrote on it.
- **`forward` takes no placement.** It reads each segment's device back off the parameters, so it
  cannot be handed a placement the weights were not placed with. It is resolved once, at `place`,
  and again only after `Module::map`, which a caller can point anywhere.

## Testing

`crates/burn-core/tests/pipeline_placement.rs` runs the split on two fixed CPU backends, Flex and
NdArray, so placement and the split forward are covered in CI on any machine with no GPU. It checks
that lazy parameters take their segment's device without initializing, and that a model split across
two devices computes what the same weights compute on one.

Unit tests cover the layout's ownership and its refusals, the even placement, and `place`'s two
panics.

Run on a four GPU box (two NVIDIA, two AMD):

- `--features cuda,vulkan` gives a four way split, the NVIDIA cards through CUDA and the AMD cards
  through Vulkan, with predictions matching the same weights on one device. `enumerate_physical`
  matters here: the two runtimes report six devices for four cards, and without the grouping two
  stages would land on one physical GPU.
- `--features vulkan` gives the same four way split with every card through Vulkan.

The example compiles under each of `cuda`, `rocm`, `vulkan`, `metal`, `webgpu`, `wgpu`, `cuda,vulkan`
and the CPU default.

## Not included

- **No scheduling.** The stages run one after another, so this buys capacity, not speed. Overlapping
  them needs a schedule that splits a batch into microbatches, such as GPipe or 1F1B. The layout and
  placement here are what such a schedule would drive.
- **No training.** A follow up PR adds the gradient tests and a training example. Gradients already
  cross the device boundary correctly; nothing in the trait changes for it.
- **No tensor parallelism**, the other half of model parallelism.
- **KV cached generation does not fit yet.** `forward_block` takes `&self`, and burn's cached
  attention takes `&mut MhaCache` from the caller, so there is nowhere to thread it. That wants a
  sibling trait rather than a change here, since adding an associated type to `Pipeline` would break
  every implementation (associated type defaults are unstable).
- **A module's bare tensors do not follow it.** `Module::map` is a no-op on a bare `Tensor` while
  `to_device` moves it, so a precomputed table such as `RotaryEncoding`'s stays where the model was
  built. The book documents the workaround; a real fix wants a `map_tensor` hook on `ModuleMapper`,
  which is a burn-core change.
- **`Learner` integration.** `LearnerModel` also wants `Display` and `TrainStep`, and `TrainStep`
  cannot be blanket forwarded, since a model's own `step` calls its own `forward`.

The book gains a "Pipeline Parallelism" section, and `examples/pipeline` runs a split model with no
GPU needed.


examples/pipeline/src/model.rs
```diff
@@ -0,0 +1,104 @@
...
+    fn forward(&self, hidden: Tensor<2>) -> Tensor<2> {
+        let normed = self.norm.forward(hidden.clone());
+        self.activation.forward(self.linear.forward(normed)) + hidden
+    }
+}
+
+impl Pipeline for Model {
+    type Input = Tensor<2>;
+    type Output = Tensor<2>;
+    type Carry = Tensor<2>;
```
nathanielsimard: I think signal would be a better name

examples/pipeline/src/model.rs
```diff
@@ -0,0 +1,104 @@
...
+            .input(&self.input)
+            .blocks(&self.blocks)
+            .output(&self.output)
+    }
+
+    fn forward_input(&self, features: Tensor<2>) -> Tensor<2> {
+        self.input.forward(features)
+    }
+
+    fn forward_block(&self, index: usize, hidden: Tensor<2>) -> Tensor<2> {
```
nathanielsimard: It's more of a stage than a block. What stage should be executed.

```rust
forward_start, forward_end, forward_hidden
```

examples/pipeline/src/model.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl Block {
+    fn forward(&self, hidden: Tensor<2>) -> Tensor<2> {
+        let normed = self.norm.forward(hidden.clone());
+        self.activation.forward(self.linear.forward(normed)) + hidden
+    }
+}
+
+impl Pipeline for Model {
```
nathanielsimard: I think a better name would reflect the fact that we distribute the model layers onto multiple devices. So LayerParallelism.

crates/burn-core/src/module/pipeline/base.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// parameter not initialized yet only takes the device, so it initializes there and a record
+    /// loaded afterwards loads there. One that a live clone of the model shares initializes where
+    /// it is and is copied instead, so drop other clones first. Only parameters move: a tensor a
+    /// module holds directly stays where the model was built.
+    ///
+    /// # Panics
+    ///
+    /// Panics when `placement` does not give a device to every block of the layout, or when a
+    /// parameter belongs to no segment, since nothing would say which device it goes to.
+    fn place(self, placement: &PipelinePlacement) -> PlacedPipeline<Self> {
```
nathanielsimard: If we go with Layer, that could become LayerPlacement.

crates/burn-core/src/module/pipeline/placed.rs
```diff
@@ -0,0 +1,93 @@
+use burn_tensor::Device;
+
+use super::{Pipeline, PipelinePlacement, PipelineStage};
+use crate::module::{Devices, Module, ModuleMapper, ModuleVisitor};
+
+/// A [placed](Pipeline::place) model, holding where each of its segments runs so a split forward
+/// reads it rather than working it out again on every call. Dereferences to the model, and is
+/// itself a [`Module`], so records, devices and training flags work as they do on the model alone.
+#[derive(Clone, Debug)]
+pub struct PlacedPipeline<M: Pipeline> {
```
nathanielsimard: That should be a `DistributedLayeredModel`.

crates/burn-core/src/module/pipeline/base.rs
```diff
@@ -0,0 +1,295 @@
...
+/// let predictions = model.forward(features);
+/// ```
+pub trait Pipeline: Module {
+    /// What `forward_input` consumes.
+    type Input: Module;
+    /// What `forward_output` produces, left on the output segment's device.
+    type Output;
+    /// Everything a segment passes to the next: the hidden state, and whatever rides along with
+    /// it, such as masks.
+    type Carry: Module;
```
nathanielsimard: The trait should have 4 associated types: Input, Output, Layer: Module, and Signal.

crates/burn-core/src/module/pipeline/base.rs
```diff
@@ -0,0 +1,295 @@
...
+                path: Vec::new(),
+            },
+        );
+
+        PlacedPipeline::new(model, placement.clone())
+    }
+}
+
+/// `layout` says which segment owns a parameter, `placement` says where that segment runs.
+struct PlaceOnOwningSegment<'a> {
```
nathanielsimard: The mapper should be aware of the layer type from the trait above

crates/burn-core/src/module/pipeline/base.rs
```diff
@@ -0,0 +1,295 @@
...
+    type Carry: Module;
+
+    /// Which submodules each segment runs.
+    fn layout(&self) -> PipelineLayout;
+
+    /// Everything before the first block: embeddings, input projection.
+    fn forward_input(&self, input: Self::Input) -> Self::Carry;
+
+    /// One block. `index` is the block's position in the layout.
+    fn forward_block(&self, index: usize, carry: Self::Carry) -> Self::Carry;
```
nathanielsimard: I think the trait should not have 3 forward functions, but 3 `getters`. So the trait should have more associated types;

```rust
trait LayerParalelism {
   type InputLayer;

   fn layer_input(&self) -> Self::InputLayer;
   fn layer_output(&self) -> Self::OutputLayer;
   fn layer_hidden(&self, index: usize) -> Self::HiddenLayer;
}
```

Each layer should implement their own forward pass and their own "to_device/fork" function. No need for a module mapper.

examples/pipeline/src/model.rs
```diff
@@ -0,0 +1,104 @@
...
+    prelude::*,
+};
+
+/// One input projection, a stack of identical blocks, one output head: the shape pipeline
+/// parallelism exists for, where the blocks are what gets shared out across devices.
+#[derive(Module, Debug)]
+pub struct Model {
+    input: Linear,
+    blocks: Vec<Block>,
+    output: Linear,
```
nathanielsimard: If we follow the new trait structure. The model should be:

```rust
pub struct Model {
   embedding: ModelEmbedding,
   hiddens: Vec<Block>,
   head: ModelHead
}
```

PR #5705, opened 2026-09-16
fix(core): let an init_mapper parameter train on an autodiff device

ThierryCantin-Demers:
A `Linear` with `LinearLayout::Col` built on an autodiff device panics the first time it is used:

```
Can't convert a non leaf tensor into a tracked tensor
```

The same happens to any lazy parameter wrapped with `Param::init_mapper` that initializes on an autodiff device and requires grad, unless the mapper returns its input unchanged. The wrapped parameter initializes as a tracked leaf, the mapper's result is computed from it and so is no longer a leaf, and marking that result as requiring grad panics. This PR maps an untracked value instead, so the mapped value is itself the leaf that trains.

## Changes

- **`Param::init_mapper`**: the mapper receives the initial value with gradient tracking off, and its result is set to require grad as before. Cases that already worked behave the same: without autodiff or a gradient requirement, turning tracking off leaves the value as it was, and a mapper that returns its input still yields a trainable leaf with the same value.

## Testing

- `a_lazy_param_with_an_init_mapper_trains_on_an_autodiff_device` (burn-core) and `col_layout_trains_on_an_autodiff_device` (burn-nn) panic on main with the error above and pass with the fix. The first also checks the mapped value is the leaf that receives the gradient.
- On GPUs, with and without `fusion`: two NVIDIA cards through CUDA, and all four cards (two NVIDIA, two AMD) through Vulkan. On each device an `init_mapper` parameter initializes with the right value and gradient, and a `Col` linear trains with the weight gradient matching the analytic one.
- `cargo test` on burn-core with and without `autodiff`, burn-nn, burn-optim and burn-store; clippy with `-D warnings` on burn-core and burn-nn; `cargo check -p burn-core --no-default-features`.


laggui (approved):
Good catch!

PR #5717, opened 2026-09-17
fix(flex): handle NaN clamp bounds

Liberxue:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. *(fmt, clippy and the Flex suite, see Testing)*

### Related Issues/PRs

Follows #5658, which fixed the one-sided ops and left this one alone.

### Changes

`float_clamp` used `f32::clamp`, which panics when either bound is NaN, so
`clamp(x, NaN, 1.0)` brought the process down while `clamp_min` and `clamp_max`
returned NaN. Composing the comparisons those two already use makes the three agree,
and matches PyTorch: a NaN bound makes every output element NaN.

### Testing

`cargo test -p burn-backend-tests --features flex`: 1704 passed, 0 failed.

`clamp_nan_bound_propagation` covers a NaN min, a NaN max and both. Reverting the fix
panics in `f32::clamp` on all three.

`cargo fmt --all --check` and `cargo clippy -p burn-flex --all-targets -- --deny warnings`
are clean. `cargo audit` is not reachable from here, so that step was skipped.


crates/burn-flex/src/ops/float.rs
```diff
@@ -664,10 +664,18 @@ impl FloatTensorOps<Flex> for Flex {
...
-            move |x: f32| x.clamp(min32, max32),
-            move |x: f64| x.clamp(min64, max64),
+            move |x: f32| {
+                let x = if x.is_nan() || x < max32 { x } else { max32 };
+                if x.is_nan() || x > min32 { x } else { min32 }
+            },
+            move |x: f64| {
+                let x = if x.is_nan() || x < max64 { x } else { max64 };
+                if x.is_nan() || x > min64 { x } else { min64 }
+            },
```
laggui, line 685: Could we handle NaN bounds explicitly and retain the existing clamp otherwise? e.g.

```rust
 if min32.is_nan() || max32.is_nan() {
     f32::NAN
 } else {
     x.clamp(min32, max32)
 }
```

This fixes the NaN-bound panic while preserving signed zero and the existing panic for min > max.

laggui (changes requested):

crates/burn-flex/src/ops/float.rs
Liberxue: Done in 303141c4.

You were right on both counts. Replacing `f32::clamp` outright meant `clamp(-0.0, 0.0, 1.0)` returned `+0.0` instead of `-0.0`, and the `min <= max` assert was gone. Now only the NaN bounds are taken out first, and everything else goes back to `x.clamp()`.

Added a signed-zero test as well. It fails on the previous version.

laggui (approved):
LGTM! Just need to address the merge conflicts.

The f64 coverage as suggested by copilot can also be added, though this is not a blocker.

Liberxue:
Merged main in to resolve the conflict with #5718, which had removed the backend gate on `clamp_nan_propagation` while this branch still carried it. Kept main's ungated version and dropped the stale comment.

Also took the Copilot note on the f64 closure, in its own commit. `float_clamp` reaches f32 and f64 separately and the bound test only ran the default element type, so the f64 branch had no coverage. It now runs both, following the `.cast(dtype)` pattern already used by `clamp_min_max_nan_propagation_f64`. Reverting only the f64 closure makes the test fail, in `f64::clamp`'s `min <= max` assert.

Merge rather than rebase so the approval stands and the new test is visible on its own.

PR #5718, opened 2026-09-17
fix(cubecl): propagate NaN through two-sided clamp

### LINKED ISSUE #5659: Cube backends: two-sided clamp maps NaN to a bound while clamp_min/clamp_max propagate it
**Describe the bug**

On the cube backends, two-sided `Tensor::clamp` replaces NaN with a bound instead of propagating it. The one-sided `clamp_min` and `clamp_max` on the same backend are correct, so the three disagree with each other.

**To Reproduce**

Verified on CUDA (`--features cuda`), for both F32 and F64:

```rust
let tensor = TestTensor::<1>::from([f32::NAN, -1.0, 2.0]).cast(dtype);

tensor.clone().clamp_min(0.0)   // [NaN, 0.0, 2.0]   correct
tensor.clone().clamp_max(1.0)   // [NaN, -1.0, 1.0]  correct
tensor.clamp(0.0, 1.0)          // [1.0,  0.0, 1.0]  NaN became the upper bound
```

`torch.clamp(nan, 0, 1)` returns NaN.

**Analysis**

`burn-cubecl` does not override `float_clamp_min` / `float_clamp_max`, so they fall back to the trait default, which masks with `float_lower_elem` / `float_greater_elem` — false for NaN, so NaN survives.

It does override the two-sided op (`crates/burn-cubecl/src/ops/tensor.rs:798`), which dispatches to `kernel::clamp` (`crates/burn-cubecl/src/kernel/clamp.rs:26`) and from there to `cubecl::prelude::clamp`. That lowering is where the NaN is lost, so the fix likely belongs in cubecl rather than in burn.

**Expected behavior**

`clamp`, `clamp_min` and `clamp_max` agree with each other and with PyTorch: NaN in, NaN out.

**Desktop**

- OS: Ubuntu 26.04 LTS, x86_64
- burn: `09f9b3f78b9210fd91f64c96dae50950bc16b7d4`
- CUDA 12.4, Tesla V100
- rustc 1.97.1

**Additional context**

Same class of bug as #5609, which covered burn-flex. While preparing #5658 I ran the NaN assertions across three backends:

| op | flex (after #5658) | ndarray | cuda |
| --- | --- | --- | --- |
| `relu` | pass | pass | pass |
| `clamp_min` | pass | pass | pass |
| `clamp_max` | pass | pass | pass |
| `clamp` | pass | pass | **fail** |

Because of this, #5658 leaves `clamp_nan_propagation` in `burn-backend-tests` gated to `flex` / `ndarray`. Once this is fixed, that gate can be removed so the assertion covers every backend, alongside the existing ungated `clamp_min` / `clamp_max` ones. #5615 tracks a separate question about the clamp bounds themselves.

I have not checked wgpu, metal, vulkan or rocm, but they share the same cubecl lowering, so they are likely affected too.

Liberxue:
### Checklist

- [ ] Confirmed that `cargo run-checks` has been executed. *(fmt, clippy and the tensor suites, see Testing)*

### Related Issues/PRs

Fixes #5659.

### Changes

burn-cubecl overrides only `float_clamp`, and `cubecl::prelude::clamp` lowers to
`max(min(x, max), min)`, which returns the non-NaN operand, so NaN came out as the
upper bound. `clamp_min` and `clamp_max` take the trait default and mask with a
comparison, so they kept it, leaving the three ops disagreeing on one backend.

Compare and select instead, in the same order as the trait default. Comparisons
against NaN are false, so the element survives.

`clamp_nan_propagation` is ungated now that f32 passes everywhere. F64 still fails on
the cube backends, so that case moves to the existing f64 test behind the CPU-backend
gate rather than staying uncovered.

### Testing

`cargo test -p burn-backend-tests --features cuda -- clamp relu` on a Tesla V100,
CUDA 12.4: 25 passed, 0 failed. `clamp([NaN, -1, 2], 0, 1)` returns `[NaN, 0, 1]`,
where before it returned `[1, 0, 1]`.

Flex and ndarray suites pass locally, 1703 and 1699. `cargo fmt --all --check` and
`cargo clippy -p burn-cubecl --all-targets -- --deny warnings` are clean. `cargo audit`
is not reachable from here, so that step was skipped.


crates/burn-cubecl/src/kernel/clamp.rs
```diff
@@ -23,11 +23,13 @@ pub(crate) fn clamp(
...
-                Vector::new(options.min_value.get::<T>()),
-                Vector::new(options.max_value.get::<T>()),
-            )
+            // clamp lowers to max(min(x, max), min), which returns the non-NaN operand and
+            // so mapped NaN to a bound. Comparisons against NaN are false, so it survives.
+            let min_value = Vector::new(options.min_value.get::<T>());
+            let max_value = Vector::new(options.max_value.get::<T>());
+
+            let clamped = select(input > max_value, max_value, input);
+            select(clamped < min_value, min_value, clamped)
```
laggui: Could we split the Float and Int clamp implementations, keeping native clamp for integers and applying the NaN-preserving logic only to floats?

The same fix is needed in fusion

https://github.com/tracel-ai/burn/blob/8f08fdad05e734918efd695b11ebe13d92a8714e/crates/burn-cubecl-fusion/src/engine/codegen/kernel.rs#L982-L987

laggui (changes requested):

crates/burn-cubecl/src/kernel/clamp.rs
Liberxue: Done in ae1bf292.

Split the kernel in two: `clamp_float` keeps the compare-and-select, `clamp_int` keeps `cubecl::prelude::clamp`. `int_clamp` now lowers exactly as it did on main, and only `float_clamp` pays for the selects.

Fixed the fused clamp as well, using the same comptime float check that `cumulative_max` already uses in `burn-cubecl/src/ops/numeric.rs`, so integers there keep the native lowering too.

Also added an int clamp test, since there was none anywhere in the suite.

laggui (approved):
Thanks, LGTM!

PR #5724, opened 2026-09-17
fix(remote): leave the process-wide log subscriber to the program

ThierryCantin-Demers:
An application that hosts a compute server beside its own endpoints gets either a log subscriber it never asked for or no logs at all, and cannot choose the level either way. Serving over WebSocket installed a global `tracing` subscriber from inside burn-communication, with a filter fixed at info that ignores `RUST_LOG`; serving over Iroh installed none, so a server built on `IrohRemoteProtocol` reported nothing at all, which is how a stalled transfer stayed invisible while debugging a split model.

The same application also had to know to mark the runtime async by hand, or a session's tensor read parked a blocking device to host copy on an executor worker.

## Changes

- **The turnkey server installs the subscriber**, through `ServerLogging`, so both transports report the same way and `RUST_LOG` picks the level. A composed server installs its own, which is the program's call.
- **burn-communication installs nothing**, and drops its `tracing-subscriber` and `tracing-core` dependencies.
- **`IrohRemoteProtocol::new` marks the runtime async**, like `RemoteServerBuilder` already did: anything hosting it runs on one.

## Testing

- `cargo test -p burn-remote`: 25 lib tests and the 4 Iroh ones pass.
- `RUST_LOG=info cargo test -p burn-remote test_to_device_over_websocket` still prints the server's lines, and they stop when `RUST_LOG` says so, which the fixed filter would not have done.
- A GPU server composed out of `IrohRemoteProtocol` and an HTTP server: its own subscriber now survives, and its trace lines are what located a transfer stall across two servers.


crates/burn-remote/src/server/logging.rs
```diff
@@ -0,0 +1,32 @@
...
+            .with_filter(
+                EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
+            )
+            // wgpu logs a line per resource at info level, which buries everything else.
+            .with_filter(filter_fn(|metadata| {
+                let wgpu = metadata
+                    .module_path()
+                    .is_some_and(|path| path.starts_with("wgpu"));
+                !(wgpu && *metadata.level() >= Level::INFO)
+            }));
```
laggui: Could we move the wgpu suppression into the fallback `EnvFilter::new("info,wgpu=warn")` and remove this unconditional `filter_fn`? Right now this suppresses events even when `RUST_LOG` explicitly enables them.

e.g.
```rust
let filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info,wgpu=warn"));
let layer = tracing_subscriber::fmt::layer().with_filter(filter);
```

I believe this should achieve the same default configuration, but still allow an explicit `RUST_LOG` to replace the whole fallback.

laggui (approved):
LGTM! Just one suggestion, but I'll approve in advance

PR #5729, opened 2026-09-18
Fix gradient accumulation at epoch boundaries

### LINKED ISSUE #5695: Gradient accumulation drops final partial window and advances LR scheduler per microbatch
**Describe the bug**

The supervised single-device, multi-device (both optimizer modes), and DDP epoch loops discard the final partial gradient-accumulation window when the number of training batches is not divisible by the accumulation interval.

The same loops call `learner.lr_step()` per microbatch rather than per optimizer update. With an accumulation interval of `k`, the first `k - 1` scheduler values are unused and non-constant schedules advance approximately `k` times too quickly. DDP additionally advances the scheduler `peer_count` times for each local microbatch.

Consequences:

- `batch_count % accumulation_interval` batches do not affect the model each epoch.
- If `accumulation_interval > batch_count`, an epoch can process and report every batch without performing any optimizer update.
- LR schedules are compressed whenever accumulation is enabled.
- `grads_accumulation(0)` is accepted and silently behaves like an update every batch.

Affected code:

- `crates/burn-train/src/learner/supervised/strategies/single/epoch.rs`
- `crates/burn-train/src/learner/supervised/strategies/multi/epoch.rs`
- `crates/burn-train/src/learner/supervised/strategies/ddp/epoch.rs`

Confirmed on upstream `main` at `52c101f455e7228f35051f11f544d672373938be`.

**To reproduce**

This integration test uses the existing `crates/burn-train/tests/common` toy model and dataloaders. The loader produces two training batches, while the accumulation interval is three.

```rust
mod common;

use std::sync::{
    Arc,
    atomic::{AtomicUsize, Ordering},
};

use burn_core::tensor::Device;
use burn_optim::lr_scheduler::{LrScheduler, LrSchedulerRecord};
use burn_train::{SupervisedTraining, logger::InMemoryMetricLogger};
use common::*;

#[derive(Clone)]
struct CountingScheduler {
    calls: Arc<AtomicUsize>,
}

impl LrScheduler for CountingScheduler {
    fn step(&mut self) -> f64 {
        self.calls.fetch_add(1, Ordering::SeqCst);
        0.1
    }

    fn to_record(&self) -> LrSchedulerRecord {
        LrSchedulerRecord::new()
    }

    fn load_record(&mut self, _record: LrSchedulerRecord) {}
}

#[test]
fn partial_accumulation_window_is_dropped() {
    let device = Device::flex().autodiff();
    let model = ToyModel::new(&device);
    let before = model.weight.val().try_into_vec_as::<f32>().unwrap();

    let optim = burn_optim::SgdConfig::new().init();
    let calls = Arc::new(AtomicUsize::new(0));
    let scheduler = CountingScheduler {
        calls: calls.clone(),
    };
    let learner = burn_train::Learner::new(model, optim, scheduler);

    // 4 items / batch size 2 = 2 training batches.
    let (dl_train, dl_valid) = make_dataloaders();
    let dir = tempfile::tempdir().unwrap();

    let result = SupervisedTraining::new(dir.path(), dl_train, dl_valid)
        .num_epochs(1)
        .grads_accumulation(3)
        .with_metric_logger(InMemoryMetricLogger::new())
        .with_application_logger(None)
        .launch(learner);

    let after = result
        .model
        .weight
        .val()
        .try_into_vec_as::<f32>()
        .unwrap();

    // Both assertions currently pass:
    assert_eq!(before, after, "both batches' gradients were discarded");
    assert_eq!(calls.load(Ordering::SeqCst), 2);
}
```

Run with:

```console
cargo test -p burn-train --test grad_accum_repro --no-default-features -- --nocapture
```

Observed: the test passes. Both batches are reported as processed, the model is unchanged, and the scheduler advances twice despite there being zero optimizer updates.

**Expected behavior**

At natural epoch completion, a non-empty partial accumulation window should produce one final optimizer update. The scheduler should advance exactly once per optimizer update, including that final update.

For the reproduction above:

- The model weights should change.
- There should be one optimizer update.
- The scheduler should advance once.

Interrupted epochs and dataloader errors should not necessarily flush; the final flush can be limited to natural exhaustion.

**Minimal fix**

- Validate that the accumulation interval is greater than zero.
- Move `lr_step()` immediately before actual optimizer calls.
- After natural dataloader exhaustion, drain and apply accumulated gradients when `accumulation_current > 0`.
- Apply the same behavior to single-device, both multi-device optimizer modes, and DDP.

Mikyx-1:
Fixes #5695.

Training now flushes a non-empty partial gradient-accumulation window when an epoch reaches natural dataloader exhaustion. The single-device loop, both multi-device optimizer modes, and DDP all advance the learning-rate scheduler immediately before each optimizer update, so accumulated microbatches no longer compress the schedule. Interrupted epochs and dataloader errors leave partial windows unapplied.

The supervised training builder also rejects a zero accumulation interval. Regression coverage verifies that a two-batch epoch with an interval of three updates the model once, advances the scheduler once, and rejects zero.

Validation:
- `cargo fmt --check`
- `cargo test -p burn-train --no-default-features`


crates/burn-train/src/learner/supervised/strategies/ddp/epoch.rs
```diff
@@ -122,11 +121,13 @@ impl<M: LearnerModel> DdpTrainEpoch<M> {
                     if accumulation <= accumulation_current {
```
Charles23R: Instead of the `naturally_exhausted` logic, we could just add a check here to always optimize on the last step. Avoids the "naturally exhausted" logic and code duplication. Here, we can do that by checking `progress.items_processed == progress.items_total`, or better yet, a `is_completed` method on `Progress` that checks exactly that.

Charles23R (changes requested):

crates/burn-train/src/learner/supervised/strategies/ddp/epoch.rs
Mikyx-1: Good suggestion, thanks! I added `Progress::is_completed()` and now flush the partial accumulation window directly on the final step, removing the extra state and duplicated cleanup logic.

crates/burn-train/src/learner/supervised/strategies/single/epoch.rs
```diff
@@ -133,8 +138,15 @@ impl<M: LearnerModel> SingleDeviceTrainEpoch<M> {
...
+                naturally_exhausted = false;
                 break;
             }
         }
+
+        if naturally_exhausted && accumulation_current > 0 {
+            let grads = accumulator.grads();
+            learner.lr_step();
+            learner.optimizer_step(grads);
+        }
```
laggui: This update happens after the last `ProcessedItem` event, so its learning rate is never reported. This also applies to the multi-device and DDP loops.

Non-blocking for this PR. It could be addressed in a follow-up by reporting LR through a dedicated `LearnerEvent::OptimizerStep` event.

laggui (approved):
Looks good for the current built-in loaders that use a fixed batch size strategy.

One caveat: `Progress::is_completed()` doesn't necessarily account for custom strategies (dataloaders currently allow this extension), where reaching the item total doesn't necessarily mean the final batch has been yielded. For example, if one were to implement a buffered batching strategy.

This can be addressed separately in a follow-up.

Charles23R:
@laggui I guess the way the progress is computed currently is just plain wrong. We should have a `items_emitted` counter that increments when yielding the items. Even then, I see it's technically possible to drop items in the strategy, thus never reaching the "completed" state (unless introducing an API change in `BatchStrategy`).

The other way i was considering is by using peek() and peekable(), but that would materialize the next item before it is needed (if n_worker=0 in the loader) which might annoy some users. Unless we add a peek method to the DataLoader trait itself, which is another API change. What do you think?

laggui:
Agreed, tracking emitted items would fix the buffering case. But the emitted count won't necessarily match the dataset size (e.g. `drop_last` style strategy that discards incomplete batches). For more dynamic strategies, the final emitted count might not even be known upfront.

Perhaps the progress contract should explicitly allow estimates, with completion tracked separately rather than inferred from the counters :thinking: 

I'd avoid introducing lookahead just for this, given the extra batch materialization as you mentioned.

> I guess the way the progress is computed currently is just plain wrong

I'd call it a limitation rather than incorrect for the current usage. The progress counter measures how many dataset items the loader has consumed, which is accurate for the built-in fixed batch size strategy. It just doesn't generalize to every possible strategy.

Charles23R (approved):

PR #5734, opened 2026-09-18
fix(cubecl): size the mask_where launch from the broadcast output

### LINKED ISSUE #5730: cpu: `mask_where` with a broadcast input ignores the mask outside the first block
**Describe the bug**

On the cpu backend (`burn-cpu`, cubecl MLIR runtime), `mask_where` returns the wrong values when the input tensor is broadcast against a larger mask. Only the first block along the broadcast dimension is correct. In every later block the mask behaves as if it were all `true`, so the whole block takes `value`.

Both the float and the int test fail:

```
tensor::float::ops::mask::should_support_mask_where_broadcast
tensor::int::ops::mask::should_support_mask_where_broadcast_int
```

**To Reproduce**

```
BURN_DEVICE=cpu cargo test --release -p burn-backend-tests \
  --no-default-features --features cpu,std --test tensor -- \
  --test-threads=1 mask_where_broadcast
```

`--test-threads=1` only avoids an unrelated hang (tracel-ai/cubecl#1668). The failure itself is not a concurrency artifact.

The existing test, unchanged:

```rust
// tensor [1, 2, 2] broadcast against a [4, 2, 2] mask and value
let tensor = TestTensorInt::<1>::arange(2..6, &device).reshape([1, 2, 2]);
let mask = TestTensorBool::<3>::from_bool(TensorData::from([
    [[true, false], [false, true]],
    [[false, true], [true, false]],
    [[false, false], [false, false]],
    [[true, true], [true, true]],
]), &device);
let value = TestTensor::<3>::ones([4, 2, 2], &device);

let output = tensor.float().mask_where(mask, value);
```

Expected:

```
[[1, 3], [4, 1]]
[[2, 1], [1, 5]]
[[2, 3], [4, 5]]
[[1, 1], [1, 1]]
```

Reported mismatches, by flat index:

```
Position 4:  1 != 2
Position 7:  1 != 5
Position 8:  1 != 2
Position 9:  1 != 3
Position 10: 1 != 4
```

Block 0 is correct and block 3 is all `value` as expected. Blocks 1 and 2 come back as all `1.0`, which is `value`, even though their masks contain `false`. That is what reading the mask with the broadcast input's layout would produce.

**Expected behavior**

The mask and value are read at their own shape; only the input is broadcast. All four blocks should match the data above, as they do on flex and ndarray.

**Desktop**
 - OS: Linux x86_64
 - Backend: `burn-cpu` (cubecl MLIR CPU runtime), burn `0.22.0-pre.3`

Liberxue:
`mask_where` sizes its launch from `input.meta.num_elements()`, but the output takes the broadcast shape from `broadcast_shape(&[&input, &mask, &value])`. When the input is broadcast the output holds more elements than the launch covers, and the tail is never written.

What the tail reads back as depends on the strategy. `mask_where_auto` picks `InplaceRhs` when the value tensor can hold the broadcast result, which makes the output *be* the value tensor, so the unwritten elements come back as `value` whatever the mask says.

This moves `broadcast_shape` above the launch sizing and counts the output instead.

## Tests

`should_support_mask_where_broadcast` and its int twin already cover this, so no new test. `should_support_mask_where_broadcast_value_small` passes either way, since its input and output hold the same number of elements.

Verified on the cpu backend, same machine and commit, with and without the change:

| | passed | failed |
|---|---|---|
| without | 1478 | 16 |
| with | 1480 | 14 |

The difference is exactly the two tests above. The other 14 failures are unrelated and present either way: 10 `einsum_*`, and 4 zero-sized tensor cases tracked in #5310.

## Other backends

This is not cpu-specific. The cpu runtime just has no slack: it reports `plane_size_max: 1` and `num_cpu_cores: Some(n)`, so `calculate_plane_count_per_cube` returns `min(cores, working_units)` and the launch comes out at exactly `working_units`. The failing test writes 4 of its 16 elements.

GPU runtimes report `num_cpu_cores: None` and a plane size of 32, so the same call returns at least one full plane and the launch rounds up to at least 32 units. That covers this test's 16 elements, which is why it passes there. The slack is a fixed few hundred units though, not proportional to the broadcast, so a larger broadcast still under-launches: 1000 input elements broadcast to 100000 gives `cube_dim` 256 and `num_cubes` 4, so 1024 units for 100000 positions.

I have not run this on a GPU. The above is read off `CubeDim::new` and the runtime properties, not measured.

Fixes #5730


Liberxue:
Superseded by #5767, which landed the same change — `broadcast_shape` moved above the launch sizing, `working_units` counted from the output — along with `mask_fill` and the in-place reuse decision. Nothing left to merge here.

laggui:
@Liberxue sorry I missed this! It could have landed before my PR, which extended to mask_fill along with additional tests

Liberxue:
> @Liberxue sorry I missed this! It could have landed before my PR, which extended to mask_fill along with additional tests

No worries! That makes sense. Glad the fix was picked up in #5767 and extended to mask_fill with the additional tests. Thanks for taking care of it!

Liberxue:
> > @Liberxue sorry I missed this! It could have landed before my PR, which extended to mask_fill along with additional tests
> 
> @laggui  No worries ～ That makes sense. Glad the fix was picked up in #5767 and extended to mask_fill with the additional tests. Thanks for taking care of it!

PR #5738, opened 2026-09-18
fix(std): ignore unit axes when checking contiguity

Liberxue:
`is_contiguous` compares every stride against the canonical ones. A dimension of size one is visited once, so its stride is never applied and can hold anything: a permute leaves whatever the axis carried before, and broadcast views use 0. `[2, 3, 1]` with strides `[3, 1, 3]`, which is what `permute([0, 2, 1])` produces, was reported as strided.

`is_dense`, right below it, already documents and handles this.

The cost is a reshape taking the copy path for a tensor that did not need one. On the cpu backend that copy path also returns wrong data, so the misclassification showed up as wrong values rather than as lost time.

## Tests

`test_reshape_analysis_broadcasted_batch`, `test_reshape_analysis_unsqueeze_split` and `test_reshape_analysis_split` used `[32, 1, 1, 1]` with strides `[1, 32, 32, 32]`, which is contiguous: only the first axis is indexed and it steps by one. They reached the branches they name only because of this bug. They keep their coverage through genuinely strided inputs, and a new test covers the short circuit.

| cpu backend | passed | failed |
|---|---|---|
| without | 1481 | 13 |
| with | 1486 | 8 |

The five are einsum tests returning wrong values. flex (1737), ndarray (1733) and `burn-std` (170) are unchanged and green. The seven pre-existing `is_contiguous` unit tests still pass, including `is_contiguous_4d_unit_shape`, which is about unit shapes and still returns false for the right reason.

Of the eight that remain, two are tracel-ai/cubecl#1670, two are #5730, and the rest are zero-sized tensor cases under #5310.


laggui (approved):
Changes LGTM.

Can we update the `is_contiguous` docs to clarify the size-one dimensions changes? 

As a follow-up, we should consolidate this with CubeCL's `is_contiguous`, which is publicly exported. That way the definitions stay consistent.

PR #5739, opened 2026-09-18
feat(core): a lazy parameter initializes on the device it moves to

ThierryCantin-Demers:
A model built lazily and then moved initializes every parameter where it was built before copying it over: `to_device` and `fork` on a parameter that has not initialized yet run its initializer first. Loading weights right after the move, the usual way to put a pretrained model on a GPU, therefore randomly initializes the whole model on the first device and transfers it, only to overwrite it. Splitting a model across several devices (#5702) pays that for every parameter, through host memory when the devices are on different backends.

With this PR, `to_device` and `fork` on a parameter that has not initialized yet only change the device it will initialize on, and a record loaded afterwards goes straight to that device.

## Changes

- **`to_device` and `fork` retarget a lazy parameter** (float, int and bool) instead of initializing it. A parameter whose lazy state a clone shares still initializes where it is and is then moved or copied, since every clone must resolve to the same value.
- **`init_mapper` initializes its base on the device it is asked for**, so a mapped lazy parameter follows a move too.
- **`Param::uninitialized` states its contract**: the initializer must create the value on the device it receives, which is no longer always the device given at construction. burn's own initializers already do.

## Testing

- `Param` tests: a record loaded after a move lands on the new device without the initializer ever running, and a lazy int or bool parameter moved before initializing initializes with the new device; these three fail on main. A forked lazy parameter keeps its gradient requirement, one with an `init_mapper` follows the move, and one shared with a clone initializes before moving, so both keep one value.
- `cargo xtask validate` passes: format, typos, audit, lint, the no-std and backend feature checks, and the workspace's tests on Flex.
- On two GPUs, through CUDA (an RTX 2060 and a GTX 1660 SUPER) and through Vulkan: a lazy parameter moved, forked onto autodiff, loaded after a move, shared with a clone, and of int and bool kind, each ends on the second card, initializing there or loading there without initializing, and the clone keeps its value on the first.

## Behavior change

A custom `Param::uninitialized` initializer that ignores its device argument, for instance by capturing a device instead, now builds its value on that captured device even after the parameter moved.


crates/burn-core/src/module/param/base.rs
```diff
@@ -529,6 +535,30 @@ impl<T: Parameter> Param<T> {
...
+                    uninitialized.device = device.clone();
+                    true
+                }
+                None => false,
+            });
+
+        if retargeted {
+            self
+        } else {
+            self.map(move_value)
```
laggui, line 566: When the base is shared or already initialized, the fallback calls `Param::map` which consumes the raw base and drops its reparameterization. It should be preserved IMO.

crates/burn-core/src/module/param/base.rs
```diff
@@ -529,6 +535,30 @@ impl<T: Parameter> Param<T> {
...
+    ) -> Self {
+        let retargeted = Arc::get_mut(&mut self.state)
+            .and_then(|state| state.initialization.as_ref())
+            .is_some_and(|initialization| match initialization.write().as_mut() {
+                Some(uninitialized) => {
+                    uninitialized.device = device.clone();
+                    true
+                }
+                None => false,
+            });
```
laggui, line 561: We should preserve the source autodiff and checkpointing settings when re-targeting. Replacing the entire device makes a lazy trainable param lose gradients when forked onto a plain device, while initialized or cloned params retain them.

crates/burn-core/src/module/param/tensor.rs
```diff
@@ -585,4 +586,121 @@ mod tests {
...
+        let param: Param<Tensor<2>> = Param::uninitialized(
+            ParamId::new(),
+            |_, _| panic!("the moved parameter initialized before loading"),
+            device.clone(),
+            false,
+            [2, 3].into(),
+        );
+
+        let moved = param.to_device(&device.clone().autodiff());
+        assert!(moved.lazy_device().is_autodiff());
```
laggui: Shouldn't this remain non-autodiff? `to_device` and `fork` change placement, thus a destination should only change where initialization happens. For example, accessing a parameter's value beforehand shouldn't change whether moving it enables or disables gradients 🤔

laggui (changes requested):
The optimization makes sense, but I think we need to make sure the behavior is consistent.

crates/burn-core/src/module/param/base.rs
ThierryCantin-Demers: `to_device` and `fork` for `Param<Tensor<D>>` take the reparameterization off before calling `map_to_device` and reattach it through `to_device_dyn`/`fork_dyn`, so there is nothing left for `map` to drop. That is unchanged from main, where the same take and reattach wrapped `self.map` directly, and the already initialized path still falls through to that exact call. Int and bool params never carry one.

Not visible from this hunk though, so I added `a_moved_param_keeps_its_reparameterization` covering both branches.

laggui: Yeah that's right, but I meant the call inside `init_mapper`'s deferred initializer, which doesn't detach and reattach the reparameterization. That's on me though, reading back the comment it was a bit underspecified.

I added a LoRA test that registers the mapper while the base is lazy, then initializes the base through a clone before evaluating the mapped value. It passes on main but produces 2 instead of 6 here because the adapter is dropped. Applying the same take/move/reattach pattern inside that initializer fixes it.

laggui (approved):

crates/burn-tensor/src/device.rs
```diff
@@ -670,8 +670,10 @@ impl Device {
         }
     }
 
-    /// Applies `source`'s autodiff context to this device.
-    pub(crate) fn with_autodiff_context_from(self, source: &Self) -> Self {
+    /// Applies `source`'s autodiff association and gradient-checkpointing strategy to this device,
+    /// discarding this device's own.
+    #[must_use]
+    pub fn with_autodiff_context_from(self, source: &Self) -> Self {
```
laggui, line 677: I'd also keep this doc hidden for now
```suggestion
    #[must_use]
    #[doc(hidden)]
    pub fn with_autodiff_context_from(self, source: &Self) -> Self {
```

PR #5743, opened 2026-09-19
Fix quiet softmax for negative-infinity slices

### LINKED ISSUE #5742: quiet_softmax returns NaN for all-negative-infinity slices
I think this is a bug: when every value in a slice is `-inf`, `quiet_softmax` returns NaNs instead of exact zeros. The stabilization step subtracts the slice maximum, so an all-`-inf` slice evaluates `-inf - (-inf)` and produces NaNs. The backward pass should also keep gradients finite.

There is also a gap in the regression coverage: `test_quiet_softmax_grad` calls `activation::softmax` instead of `activation::quiet_softmax`. This means the test exercises the wrong function and does not detect the non-finite `quiet_softmax` behavior.

Mikyx-1:
## Summary
- replace the detached stabilization maximum only for all-`-inf` slices so quiet softmax returns exact zeros instead of NaNs
- preserve the stable result for mixed finite/`-inf` slices on every selected axis
- correct the miswired quiet-softmax autodiff test, which previously called ordinary softmax
- cover f32, f16, both axes, and finite backward gradients

## Validation
- f32 activation suite: 65 passed
- f16 activation suite: 65 passed
- softmax autodiff suite in base and checkpointing modes: 6 passed
- `cargo clippy -p burn-tensor --all-targets --features flex -- -D warnings`
- `cargo clippy -p burn-backend-tests --test autodiff --features flex -- -D warnings`
- `cargo check -p burn-tensor --no-default-features`
- `git diff --check`

Closes #5742


laggui (approved):
Thanks!

PR #5747, opened 2026-09-19
docs(pytorch-reader): sync README with lib.rs

### LINKED ISSUE #5710: pytorch-reader: README lists three containers, lib.rs documents four
`crates/pytorch-reader/README.md` says the crate reads three containers:

> It reads every container `torch.save` has written: the ZIP archive of PyTorch 1.6 and later, the legacy pickle stream of 0.1.10 through 1.5, and the TAR archive before that.

`src/lib.rs` documents four, adding the plain pickle that `from_reader` and `load_config` handle. The README is what crates.io shows, so it should match; a one-clause addition.

While there: the README's code block is not compiled (lib.rs does not `include_str!` it), so its bare `?` at top level is fine today, but it would break the moment someone wires the README in as the crate doc. Wrapping it the way the lib.rs example is (`# Ok::<(), Box<dyn std::error::Error>>(())`) costs nothing and removes the trap.

superroket169:
Adds the plain-pickle container to the README's list (lib.rs already documents four; README only listed three) and closes the README's top-level code block the same way lib.rs's example does. Closes #5710.


antimora (approved):
LGTM

laggui (approved):

PR #5752, opened 2026-09-20
fix: hold the legacy checkpoint file open instead of reopening by path on every read

### LINKED ISSUE #5713: pytorch-reader: LegacySource reopens the checkpoint by path on every read
## Summary

`LegacySource::read_storage` calls `File::open(&self.path)` on every read. `ZipSource` (after the fix for #5711) holds its handles from open until drop and never reopens by path. The two containers therefore behave differently when the file goes away: a ZIP reader keeps reading the file it opened, a legacy reader fails with `NotFound` (`test_os_errors_keep_their_kind` pins this).

The reopen is also a correctness hazard under atomic replacement. A training loop that renames the next epoch's checkpoint over the path while an evaluator holds a reader makes that reader read the new file at the old offsets. When the layout matches (same model), the `stored_numel` check passes and the tensor comes back as a torn mix of old and new bytes, with no error.

## Fix sketch

`LegacySource { file: File, state }` and `new(file)`; `load_legacy` already opens the file on its first line and can hand it over. `read_storage` becomes two positional reads (the `i64` count at `offset`, then `len` bytes at `offset + 8 + start`) through the `read_exact_at` helper that the #5711 fix adds to `storage.rs`. `finish` already validates `end <= file_len`, so the allocation is bounded by the same argument `ZipSource::read_stored` uses. About 15 lines net.

## Caveat

`test_os_errors_keep_their_kind` uses the legacy reopen to get an OS-level error at read time; it exists to check that `build_tensor` preserves OS error kinds instead of folding them into `InvalidData`. That path needs a different trigger once the file is held open.

SIDDARTHAREDDY8:
### Checklist

### Related Issues/PRs

Fixes #5713.

### Changes

`LegacySource::read_storage` reopened the checkpoint by path on every read. A reader whose file was unlinked failed with `NotFound`, and worse, an atomic replacement of the checkpoint could make a held reader read a torn mix of old and new bytes at the old offsets with no error. `ZipSource` already holds its handles from open until drop; this brings the legacy container to the same behavior:

- `LegacySource` now holds an open `File` handle instead of the path; `new(file)` takes it from `load_legacy`, which hands over a clone of the handle it already opened for the metadata pickles.
- `read_storage` became two positional reads through `read_exact_at`: the `i64` element count at `offset`, then `len` bytes at `offset + 8 + start`. Offsets are identical to the old seek/read/seek/read sequence, and `finish`'s `end <= file_len` validation bounds the allocation exactly as before.
- Reworked `test_os_errors_keep_their_kind`: shrinking the file after open is the new OS-error trigger (it asserts `UnexpectedEof` keeps its kind through `build_tensor` instead of being folded into `InvalidData`), and added `test_legacy_reads_after_the_file_is_removed` (unix), the legacy counterpart of the zip test.

This change was drafted with AI assistance; every changed line was reviewed by the author against the current sources on main.

### Testing

- `cargo run-checks` could not be run locally (no Rust toolchain in this environment); the `pytorch-reader` suite must run in CI. The two positional reads were checked line-by-line to reproduce the old sequence at identical offsets, and the allocation stays bounded by the same `finish` validation.


antimora:
Thank you, but we already have https://github.com/tracel-ai/burn/pull/5737 opened. I should have assigned the ticket to myself.

PR #5754, opened 2026-09-20
Fix fmod_scalar for infinite dividends

### LINKED ISSUE #5753: fmod_scalar(infinity) should return NaN for infinite dividends
## Bug

`Tensor::fmod_scalar` returns the input unchanged whenever the scalar is infinite. This is correct for finite dividends, but incorrectly preserves infinite dividends.

```rust
let tensor = Tensor::<1>::from_data([f32::INFINITY, f32::NEG_INFINITY, 5.0], &device);
let result = tensor.fmod_scalar(f32::INFINITY);
```

Current: `[∞, -∞, 5]`  
Expected: `[NaN, NaN, 5]`

Under IEEE 754-2019 (§5.3.1 and §7.2), a finite value remainder infinity is unchanged, while an infinite dividend is invalid and produces a quiet NaN by default. See [IEEE 754-2019](https://standards.ieee.org/ieee/754/6210/).

Mikyx-1:
### Checklist

### Related Issues/PRs

Closes #5753.

### Changes

- Return NaN for infinite dividends when the scalar divisor is infinite.
- Preserve finite dividends unchanged.
- Add a regression test for `[+∞, -∞, 5] % ∞`.

### Testing

Added a backend test for both positive and negative infinite divisors. Not run locally because `cargo` is unavailable in the current environment.


laggui (approved):
LGTM, thanks!

PR #5760, opened 2026-09-21
perf(cubecl): NHWC direct kernel for conv_transpose2d

### LINKED ISSUE #5759: conv_transpose2d has no NHWC kernel, so the conv data gradient flips layout and loses coalescing (6.9x slower than forward at equal FLOPs)
### Describe the bug

Every kernel in the CubeCL convolution stack works in NHWC, except the transposed
convolution. `conv_transpose2d_direct` is the one that is still NCHW, and the data
gradient has to flip the layout back and forth around it. The kernel carries its own
TODO saying so:

```rust
// crates/burn-cubecl/src/kernel/conv/backward_data/fallback.rs
// We don't yet have NHWC kernels for conv_transpose so need to do this.
// Should eventually use NHWC kernels instead
let out_grad = permute_nhwc_to_nchw(out_grad);
let weights  = permute_nhwc_to_nchw(weights);
```

The layout flip is the smaller half of the cost. The real cost is the memory pattern the
NCHW kernel is stuck with: the output decomposes as `[batch, oc_out, out_y, out_x]`, so
`out_x` varies fastest and neighbouring threads read unrelated input elements, while the
`in_c` reduction strides by a whole feature map -- the axis that changes slowest in
memory. There is no reuse of any kind, so a single conv1d data gradient reads on the
order of gigabytes for well under a GFLOP of arithmetic.

This is the only route available. With `dilation = 1, kernel = 5, 128 -> 128`, every
other `dgrad_autotune` candidate declines the shape (`dgrad_im2col_1x1` takes pointwise
only, the MMA variants report no available tile size, TMA is unimplemented for data
backprop), so the fallback wins by default and its quality sets dgrad performance
outright.

### To Reproduce

Forward and dgrad do the same number of multiply-accumulates at `dilation = 1`, so any
gap between them is layout and memory behaviour, not arithmetic.

```rust
use std::time::Instant;
use burn::{prelude::*, tensor::{Distribution, module::conv1d, ops::ConvOptions}};

fn main() {
    let plain = Device::default();
    let device = plain.clone().autodiff();
    let (batch, steps) = (32, 160);

    for (label, cin, cout, k) in [
        ("64->64    k5", 64, 64, 5),
        ("128->128  k5", 128, 128, 5),
        ("128->128  k3", 128, 128, 3),
        ("128->128  k1", 128, 128, 1),
    ] {
        let options = ConvOptions::new([1], [k - 1], [1], 1);
        let x = Tensor::<3>::random([batch, cin, steps], Distribution::Default, &device)
            .require_grad();
        let w = Tensor::<3>::random([cout, cin, k], Distribution::Default, &device);

        let time = |mut op: Box<dyn FnMut()>| -> f64 {
            for _ in 0..3 { op(); plain.sync().unwrap(); }
            let mut s = Vec::new();
            for _ in 0..10 {
                let t = Instant::now();
                op();
                plain.sync().unwrap();   // Burn dispatches asynchronously
                s.push(t.elapsed());
            }
            s.sort_unstable();
            s[s.len() / 2].as_secs_f64() * 1000.0
        };

        let (x1, w1, o1) = (x.clone(), w.clone(), options.clone());
        let fwd = time(Box::new(move || {
            let _ = conv1d(x1.clone(), w1.clone(), None, o1.clone());
        }));
        let (x2, w2, o2) = (x.clone(), w.clone(), options.clone());
        let both = time(Box::new(move || {
            conv1d(x2.clone(), w2.clone(), None, o2.clone()).sum().backward();
        }));
        println!("{label:<14} {fwd:>9.3} {:>9.3} {:>7.1}x", both - fwd, (both - fwd) / fwd);
    }
}
```

Only the input is tracked, so `backward()` runs the data gradient alone.

### Expected behavior

Same FLOPs, comparable time. The data gradient reads slightly more than the forward pass,
so a ratio somewhat above 1x is expected -- not 6.9x.

Observed, Apple M4 / wgpu, `burn 0.22.0-pre.3`:

```
layer            forward ms     dgrad ms  dgrad/fwd
128->128  k5          4.030       27.687       6.9x
```

The `k1` row is a useful control: at kernel 1 the gradient is routed to
`dgrad_im2col_1x1` instead and never touches this path, and the ratio is far better. The
cliff between `k1` and `k3`/`k5` is the cost of this fallback.

`CUBECL_DEBUG_OPTION=profile` shows the dispatch is a single `ConvTranspose2dDirectKernel`
plus two `CopyKernelPack` -- no matmul, no implicit GEMM.

### Desktop

- OS: macOS 15 (aarch64, Apple M4, 10 cores)
- Backend: wgpu 30.0.1 / Metal
- Version: burn `0.22.0-pre.3`, cubecl `0.11.0-pre.3`

This path has no backend branch in `burn-cubecl`, so it is not Metal-specific.

### Additional context

The forward pass already shows what the fix looks like: `kernel/conv/direct.rs` is an
NHWC direct convolution. The transposed convolution needs the same treatment -- decompose
the output as `[batch, out_y, out_x, oc_out]` so neighbouring threads differ only in the
output channel, and reduce over the contiguous channel axis.

onenewcode:
### Checklist

> `cargo run-checks --backend metal` passes format and typos. Its audit step could not
> fetch the RustSec advisory database from my network; lint and tests were run separately
> and are listed under Testing.

### Related Issues/PRs

Closes #5759.

### Changes

Every kernel in the CubeCL convolution stack works in NHWC except the transposed
convolution, so `conv_data_backward_fallback` flipped the layout back and forth around
it. The kernel's own TODO asked for this.

**The kernel is now NHWC.** The output decomposes as `[batch, out_y, out_x, oc_out]`, so
neighbouring threads differ only in the output channel: they read the same input element
at each step, which the cache broadcasts, and their weight reads are consecutive
addresses, so they coalesce. Under NCHW `out_x` varied fastest instead, so neighbouring
threads read unrelated input and the `in_c` reduction strided by a whole feature map --
the same multiply-accumulates over far more traffic.

**The channel axis is vectorized**, which compounds both effects: a thread owns `line`
channels, so one input element feeds `line` multiply-accumulates from a single register
and the weights beside it arrive in one instruction. The width comes from
`max_vector_size_many` over the weight and output; asking the weight also keeps a vector
from straddling a group boundary for free, since its channel axis is the per-group count.

**The NCHW kernel is deleted rather than kept alongside.** `conv_transpose2d_direct` now
permutes and calls the NHWC one, the way `conv_forward` already does. One kernel instead
of two near-duplicates, and callers of the public `conv_transpose2d` get the speedup too,
not only the data gradient.

**Also fixes a correctness bug the rewrite exposed.** The window end was derived from its
start plus one fixed width, which is empty whenever the stride outruns the dilated kernel
extent -- every output came back as bias alone. `contributing_range` now computes both
ends exactly. There is a regression test that fails on `main`.

### Testing

**conv1d data gradient**, dilation 1, batch 32, 160 steps, Apple M4 / wgpu. Forward and
dgrad do identical FLOPs here, so the ratio is layout and memory behaviour alone:

| layer | before | after |
| --- | --- | --- |
| `64->64 k5` | 2.3x | 0.7x |
| `128->128 k5` | **6.9x** | **0.3x** |
| `128->128 k3` | 3.4x | 0.6x |
| `128->128 k1` | 0.4x | 0.2x |

**Forward transposed convolution**, old kernel against new in a single build so the
numbers are directly comparable, with outputs asserted equal before timing:

| case | old ms | new ms | ratio | new + forced contiguous |
| --- | --- | --- | --- | --- |
| `64->64 k3 s2` | 40.285 | 2.520 | 0.06x | 0.07x |
| `128->64 k4 s2` | 27.287 | 2.429 | 0.09x | 0.09x |
| `256->128 k4 s2` | 13.735 | 1.462 | 0.11x | 0.11x |
| `32->32 k3 s1` | 25.748 | 2.050 | 0.08x | 0.09x |
| `64->64 k3 groups=4` | 10.590 | 1.079 | 0.10x | 0.14x |

The last column prices behaviour change (1): still 7x faster with a contiguous copy
forced on the result.

**Correctness.** New in-crate tests check the kernel against a host-side scatter
reference -- deliberately the opposite direction from the kernel's gather, and using the
shared `calculate_conv_transpose_output_size` rather than restating the launcher's
formula, so the oracle does not agree with its own mistakes. 13 cases covering bias,
stride, padding, dilation, groups, depthwise, pointwise, per-axis-asymmetric options, a
unit axis, ragged channel counts, and stride past the dilated extent. They exercise
vector widths 1, 2 and 4, so the vectorized indexing and the group/vector interaction are
genuinely covered rather than collapsing to the scalar path.

New `conv_backward_data.rs` covers the data gradient against the ndarray reference across
1D/2D/3D with groups, stride, dilation and asymmetric padding. It seeds a random upstream
gradient rather than the uniform one `sum()` alone produces, since a uniform gradient
hides index errors that only permute terms within the reduction.

Suites run on Metal: `burn-cubecl` lib 27, `autodiff` 110, `tensor` 48, `tensor_f16` 48,
`autodiff_f16` 110 -- all green. The kernel tests also pass under `--features cpu`, a
different runtime with different vector widths.


SamuelBelanger (approved):
Looks good! The fix and the tests are solid.

I ran conv_transpose_ops against the base commit on an RTX 2060 (CUDA). Small shapes are about 35% faster, the larger ones are unchanged (probably autotune picking col2im there), and nothing regressed.

laggui (commented):
The NHWC kernel looks good, but removing the col2im from input-gradient selection might introduce a performance regression in convolution bakward wrt to the input specifically.

I tested locally (RTX 4050), and for input an `[4,256,32,32]` with a 3x3 kernel and stride 2 it seems to have slowed down from 0.77->1.30 ms in f32 and 0.44 ->0.62 ms in f16.

Can you retain the col2im autotune candidate in this case?

onenewcode:
@laggui Good catch, and thanks for the numbers. 

Fixed in 79cf115: the fallback now goes through `conv_transpose2d` with its default (autotune) strategy again。

laggui (approved):

PR #5761, opened 2026-09-21
Reduce Flex ConvTranspose time and memory use

### LINKED ISSUE #5758: Limit ConvTranspose column workspace in the Flex backend
The Flex backend allocates the full column matrix before it adds the values to the ConvTranspose output. The 1D, 2D, and 3D operations share this code.

For each batch and group, this buffer has
`out_channels_per_group * kernel_spatial * input_spatial` elements.
For example, an F32 input of `[1, 64, 368, 368]`, 64 output channels, and a 2x2 kernel needs 132.25 MiB for columns. This does not include the input, output, or GEMM work buffers. Parallel groups can each allocate a column buffer.

I would like to compute the matrix in tiles, add each tile to the output, and reuse the column buffer. The aim is to reduce temporary memory and memory traffic. The change would stay in the Flex ConvTranspose implementation and use the current GEMM kernels.

I will compare complete operator calls with 1 and 4 threads. The tests will cover 1D/2D/3D, overlap, padding, dilation, groups, tile boundaries, and small inputs. I will also check output error and peak live allocations. Tile size and GEMM packing costs need measurements before choosing a limit.

This is separate from the batch/group parallelism and bounds checks covered by #5613 and #5617.

Source: [`conv_transpose.rs` at 3a93fbf](https://github.com/tracel-ai/burn/blob/3a93fbfc0f6517c9165031f8b2fbbe8523ff5c29/crates/burn-flex/src/ops/conv_transpose.rs).

Hosi121:
### Checklist

- [x] Made sure the book is up to date with changes in this PR. The public API is unchanged.

### Related Issues/PRs

Fixes #5758.

### Changes

Flex ConvTranspose builds a full column matrix and then adds its values to the output. This change reuses an 8 MiB column buffer per active batch/group. It also runs col2im across output channels when batch/group work does not fill the Rayon pool. The shared code covers 1D, 2D, and 3D.

The GEMM input keeps its original row stride. Tiles run in reverse order to keep the col2im addition order. Each channel task owns a separate output slice. Small tiles and one-thread calls use the serial loop. Channel work also applies below the tile limit. One input position is the minimum tile, so its columns can exceed the limit. The arithmetic count and GEMM kernels are unchanged.

The first two commits separate input tiles from channel work. The table measures those steps through `99d11a0` against main `3a93fbf`: Intel Core Ultra 7 255H, Linux under WSL2, F32, default Flex features, **four threads**. Values are medians of five alternating process runs per case. Batch size is 1; input and output channel counts are equal.

| Input and kernel | Main ms | Tiles only ms | Tiles + channels ms | Peak live MiB, main → tiles + channels |
|---|---:|---:|---:|---:|
| 1D, C64, L2048, K8/S4/P2 | 1.353 | 1.311 | 1.037 | 6.13 → 6.13 |
| 2D, C64, 368x368, K2/S2/P0 | 196.412 | 171.964 | 85.909 | 264.56 → 140.31 |
| 2D, C32, 256x256, K4/S2/P1 | 79.318 | 64.769 | 32.580 | 160.06 → 40.06 |
| 2D, C64, 128x128, K3/S1/P1 | 16.359 | 11.341 | 5.998 | 40.14 → 12.14 |
| 3D, C16, 32x32x32, K3/S2/P1 | 22.001 | 16.294 | 8.013 | 69.29 → 23.29 |

Peak live allocations include output and temporary buffers during a warmed call. They exclude existing input, weights, and retained allocations. They are not process RSS. Channel work does not add column buffers.

Controls varied on this host. The grouped control was 19.6% slower than main with one thread and 6.2% slower with four threads in this run. A second run with all cases in each process found 7.1% and 0.6% lower time. I do not claim a stable gain or regression for these controls. Both runs, all cases, and one-thread results are in the report.

Inputs and weights are generated nonzero data. Model inference and ARM/GPU performance were not measured.

The third commit lowers the channel work limit when GEMM already uses Rayon and there are at least two output channels per thread. It keeps at least 128K column elements in total and 32K per thread; eight or more threads keep the old 256K limit. The group-count guard stays in place. Removing it increased time and peak allocations in the tests.

Against `99d11a0`, two smaller 1D F32 cases use 19.7% and 29.1% less time at four threads. Three F16 cases use 29.9–42.0% less time. The F32 14x14 case is 4–5% slower at four threads. A small F32 1D case had large changes between runs, so I do not claim a stable gain there. The [work-limit report](https://github.com/Hosi121/burn/tree/bench/flex-conv-transpose-tiles/flex-conv-transpose-bench/relaxation) includes all cases, ranges, rejected rules, and results for 1, 2, 4, 6, and 8 threads.

### Testing

- `cargo run-checks`: 3,994 passed; 43 ignored.
- Release Flex unit tests: 418 passed. F16/F32/F64 tests compare serial and four-thread paths, including tile tails, batches, groups, and the lower work limit.
- Direct scalar-reference tests cover tile bounds, padding, output padding, dilation, bias, empty channels, and addition order.
- The typed entry points are tested below, at, and above the 8 MiB column-buffer limit for F16, F32, and F64. All nine cases match a direct scalar reference.
- All 12 complete F32 probe outputs at `99d11a0` match main bit for bit. All 18 additional F16/F32/F64 outputs for the work-limit change match `99d11a0` bit for bit.
- All 19 public ConvTranspose benchmark smoke cases passed. All-target Clippy, formatting, and diff checks passed.

[Source, full results, and reproduction steps](https://github.com/Hosi121/burn/tree/bench/flex-conv-transpose-tiles/flex-conv-transpose-bench).


Hosi121:
The two missing lines call the tiled implementation from the typed entry point. The earlier unit tests used smaller buffer limits and called the internal implementation directly.

Commit `93ac657` adds tests for F16, F32, and F64 with inputs just below, at, and above the 8 MiB column-buffer limit. All output values match a direct scalar reference. The cases above the limit also check a final tile with one input position. This change only adds tests; the production code is unchanged.

`cargo run-checks`, all 418 Flex unit tests, and all-target Clippy passed. The new Codecov report is pending.

The project coverage warning is separate: the previous report increased from 70.81% to 70.88%, below the fixed 80% target. Both Codecov checks are configured as informational and passed for `8248538`. I kept the coverage settings unchanged.

laggui (approved):
LGTM, just did a minor cleanup of the tests

Hosi121:
Thanks for the review, cleanup, and merge!

